PR #121, opened 2023-07-03
Update youtube_video.py

missioncfwf:
error if "No data found"

wrap error in try function and print "Error loading data from YouTube video", then empty list returned if error (for the video and not the playlist)

### DIFF
--- a/embedchain/loaders/youtube_video.py
+++ b/embedchain/loaders/youtube_video.py
@@ -1,22 +1,24 @@
 from langchain.document_loaders import YoutubeLoader
-
 from embedchain.utils import clean_string
 
-
 class YoutubeVideoLoader:
-
     def load_data(self, url):
-        loader = YoutubeLoader.from_youtube_url(url, add_video_info=True)
-        doc = loader.load()
-        output = []
-        if not len(doc):
-            raise ValueError("No data found")
-        content = doc[0].page_content
-        content = clean_string(content)
-        meta_data = doc[0].metadata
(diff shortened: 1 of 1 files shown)


cachho:
thanks for the PR. Blindly catching all exceptions is never advised, and I'm of the opinion that we shouldn't just ignore all exceptions and keep the code running.

If you want to do that, it takes 3 extra lines.
```python
try:
    app.add(...)
except Exception:
    pass
```

But maybe some users (like me) want to see errors and handle them in a unique way. That's not possible if no error is returned. I know you could handle an empty list as an error.

With that being said, I'm not against abstracting errors on the embedchain level, it just shouldn't turn into a script that just runs through no matter what.

taranjeet:
@cachho @deshraj : what do you think about this?

cachho:
I'm very much against it if I'm being blunt. We should add embedchain level exception handling at some point, but just escaping an exception is not the way to do it. The user should handle this. Maybe people want to, idk, keep a list with videos that have been deleted or something. If you escape the error and act like nothing happened you can't do that.

Please close. Don't take it personal.

taranjeet:
@deshraj : what do you think about this?

deshraj:
Agree with @cachho on this one. Handling data loading errors should be done by the user. Catching this exception in the python package could cause some silent regressions which is not good.

deshraj:
@missioncfwf sorry but we would have to close this PR. Appreciate your contribution. 🙇

PR #140, opened 2023-07-05
[refactor] pass config object instead of pass db/ef params

gasolin:
alternative proposal to replace https://github.com/embedchain/embedchain/pull/100

1. pass `config` object instead of pass separate params (db, ef), which is expected to extend more
2. add `default_model` in config time (could add another method to overwrite in query time though)

### DIFF
--- a/embedchain/embedchain.py
+++ b/embedchain/embedchain.py
@@ -35,15 +35,20 @@ DB_DIR = os.path.join(ABS_PATH, "db")
 
 
 class EmbedChain:
-    def __init__(self, db=None, ef=None):
+    def __init__(self, config):
         """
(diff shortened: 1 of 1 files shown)


cachho:
We're currently discussing which route to take for configuration. I agree that it needs to change from how it is right now.

cachho:
take a look at #156  and PR #158, that's most likely the route were taking for configuration.

josebenitezg (approved):

josebenitezg (approved):

josebenitezg (approved):

josebenitezg (approved):

josebenitezg (approved):

gasolin:
replaced by #158

PR #146, opened 2023-07-05
Import gpt4all locally to avoid package unavailable error while using App

rohitgr7:


taranjeet:
thanks for the contribution @rohitgr7

Issue #113, opened 2023-07-01
[Feature Request] History


cachho:
On discord, this was discussed as part of #105, but I think it should be tracked separately.

History can greatly improve the quality of the bot. Without history, queries like *tell me more* can't be processed.

```
Me: how do I write a hello world program in python?
AI: To write a "Hello, World!" program in Python, you can use the following code: print("Hello, World!") This will display the phrase "Hello, World!" on the console when you run the program.
Me: tell me more
AI: I'm sorry, but without any specific context or information about the subjects you are referring to, I cannot provide any further details.
```

You can deal with this by appending previous questions and answers to each prompt. 
The obvious down side of this is that the costs are multiplied really quick.

# Implementation

I'm pretty sure you can just add it to the prompt, something like: You are aware of this history: ...
The cut off could be a number of messages or a token count.

The biggest question from an implementation standpoint is, **how much of this should be handled server side, how much client side?**. (Considering the (usual) scenario where embedchain is implemented as an API and a chatbot frontend sends requests with querries.)

I know we're trying to abstract here, so handle as much on the server as possible. But the problem with that is, that there might be multiple users (sessions). And you don't want the history of someone else. Of course you could store the history on the server with a session id that's provided by the client to solve this problem. Then you would still end up with more boilerplate (session id generation and passing), and you would have to come up with a system to clear sessions due to memory/storage limitations.

That's why I think it should be handled on the client. Most chat frontends will have a messages array anyways. Then you can just pass that with the request and embedchain can handle it as a optional parameter and append it to the prompt. Yes, it's more stuff that the user has to handle, but it's safer for the memory.

What do you think?

PR #150, opened 2023-07-06
feat: Add support for chat endpoint

taranjeet:
This commit introduces a new endpoint called chat. It maintains the context of the last 5 user messages and continues to do a semantic search to find the context.
Addresses https://github.com/embedchain/embedchain/issues/102

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ embedchain is a framework to easily create LLM powered bots over any dataset. If
 
 # Latest Updates
 
+* Introduce a new interface called `chat`. It remembers the history (last 5 messages) and can be used to powerful stateful bots. You can use it by calling `.chat` on any app instance. Works for both OpenAI and OpenSourceApp.
+
 * Introduce a new app type called `OpenSourceApp`. It uses `gpt4all` as the LLM and `sentence transformers` all-MiniLM-L6-v2 as the embedding model. If you use this app, you dont have to pay for anything.
 
 # What is embedchain?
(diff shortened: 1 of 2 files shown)


taranjeet:
- need to update readme.

cachho:
Nice to see that lang chain has a memory buffer. Here are my thoughts:

We should consider this as part of the configuration effort. You're adding another argument.

My main concern: Is this able to handle sessions in a client server model? How will it know what user said what and not mix up the history with simultaneous use? 

With that being said, I see how it would appeal to solo users that use this package for private purposes. It's definitely an improvement. For "pros" we should add a configuration option for history so they can do session management through the frontend. #113 I talk about this in this issue as well.

taranjeet:
I see what you are saying. Chat history should be user specific and should be persisted to a db as well.

Opened two issues
- https://github.com/embedchain/embedchain/issues/153
- https://github.com/embedchain/embedchain/issues/154

cachho:
> I see what you are saying. Chat history should be user specific and should be persisted to a db as well.
> 
> Opened two issues
> 
> * [Make chat history user specific #153](https://github.com/embedchain/embedchain/issues/153)
> * [Persist chat history #154](https://github.com/embedchain/embedchain/issues/154)

Good that you agree. Database is good, in the meantime, passing a list of strings as history will be possible from the state of the frontend, without a database. I will work on an implementation once we have agreed on a configuration.

Issue #64, opened 2023-06-25
[BUG] Chroma DB Duplicate ID Error


ArmanJR:
This is my code:

```python
import os
os.environ["OPENAI_API_KEY"] = "sk-???"
from embedchain import App
naval_chat_bot = App()
naval_chat_bot.add_local("pdf_file", "docs/masnavi-en.pdf")
print(naval_chat_bot.query("Who is the most powerful man?"))
```

I get `chromadb.errors.DuplicateIDError: Expected IDs to be unique, found duplicates for`. Where is the problem?

P.S: This was my second attempt. The first one with a different pdf document was successful.

luisrock:
Same issue here

missioncfwf:
Same issue here. I had no problem uploading small files, but big ones cause this issue for me

ramaschneider:
Same here - is there any update regarding this issue?

taranjeet:
I think this could be because ids are clashing.
Right now, the ids are generated using a hash of chunk text and url. There may be scenarios where same chunk is occurring multiple times in a doc, hence the error.

@ramaschneider / @missioncfwf / @ArmanJR : possible for anyone of you to share doc file?
I just want to double check if this is the issue or something else.

ArmanJR:
@taranjeet sure: [https://www.sufi.ir/books/download/english/molavi-en/masnavi-en.pdf](https://www.sufi.ir/books/download/english/molavi-en/masnavi-en.pdf)

Harin329:
I opened a PR for this, setting a random bit in the ID hashing function seems like a harmless and quick fix. Not entirely sure of the implications of turning the lists into a set, but that may affect the composition of the dataset

taranjeet:
Hey @ArmanJR @luisrock @missioncfwf @ramaschneider @Harin329 : merged PR https://github.com/embedchain/embedchain/pull/160 which addresses this issue.

taranjeet:
Released in version 0.0.18
https://test.pypi.org/project/embedchain/0.0.18/

Closing this issue.

PR #110, opened 2023-07-01
fix: ValueError: Expected metadatas to be a list, got ({'url': 'local'},)

cachho:
Fixes #108.

Tested by using the built of the clone instead of pip in prod where I got the error.

PR #160, opened 2023-07-06
Prevent clashing chunk IDs

Harin329:
This PR solves the issue in #64 

With a large enough data set, the clashing occurs relatively often. We can quickly solve this by adding a random bit into the hash function. Alternatively, we can explore using a set for `ids` instead of a list, which might compromise the data. (ie. what if you're trying to use the bot to make up its own language syntax and communicate via only 10 words)

### DIFF
--- a/embedchain/chunkers/base_chunker.py
+++ b/embedchain/chunkers/base_chunker.py
@@ -26,7 +26,7 @@ class BaseChunker:
 
             for chunk in chunks:
                 chunk_id = hashlib.sha256((chunk + url).encode()).hexdigest()
-                if (idMap.get(chunk_id) == None):
+                if (idMap.get(chunk_id) is None):
                     idMap[chunk_id] = True
                     ids.append(chunk_id)
                     documents.append(chunk)
--- a/embedchain/chunkers/base_chunker.py
+++ b/embedchain/chunkers/base_chunker.py
@@ -1,4 +1,5 @@
 import hashlib
+import random
 
 
 class BaseChunker:
@@ -16,7 +17,7 @@ class BaseChunker:
             chunks = self.text_splitter.split_text(content)
             url = meta_data["url"]
             for chunk in chunks:
-                chunk_id = hashlib.sha256((chunk + url).encode()).hexdigest()
+                chunk_id = hashlib.sha256((chunk + url + str(random.getrandbits(128))).encode()).hexdigest()
                 ids.append(chunk_id)
                 documents.append(chunk)
                 metadatas.append(meta_data)
--- a/embedchain/chunkers/base_chunker.py
+++ b/embedchain/chunkers/base_chunker.py
@@ -1,5 +1,4 @@
 import hashlib
-import random
 
 
 class BaseChunker:
@@ -9,6 +8,7 @@ class BaseChunker:
     def create_chunks(self, loader, url):
         documents = []
         ids = []
+        idMap = {}
         datas = loader.load_data(url)
         metadatas = []
         for data in datas:
@@ -17,10 +17,12 @@ class BaseChunker:
             chunks = self.text_splitter.split_text(content)
             url = meta_data["url"]
             for chunk in chunks:
-                chunk_id = hashlib.sha256((chunk + url + str(random.getrandbits(128))).encode()).hexdigest()
-                ids.append(chunk_id)
-                documents.append(chunk)
-                metadatas.append(meta_data)
+                chunk_id = hashlib.sha256((chunk + url).encode()).hexdigest()
+                if (idMap.get(chunk_id) == None):
+                    idMap[chunk_id] = True
+                    ids.append(chunk_id)
+                    documents.append(chunk)
+                    metadatas.append(meta_data)
         return {
             "documents": documents,
             "ids": ids,
--- a/embedchain/embedchain.py
+++ b/embedchain/embedchain.py
@@ -154,8 +154,6 @@ class EmbedChain:
 
             ids = list(data_dict.keys())
             documents, metadatas = zip(*data_dict.values())
-            documents = list(documents)
-            metadatas = list(metadatas)
 
         self.collection.add(
             documents=documents,
--- a/embedchain/embedchain.py
+++ b/embedchain/embedchain.py
@@ -138,6 +138,8 @@ class EmbedChain:
 
             ids = list(data_dict.keys())
             documents, metadatas = zip(*data_dict.values())
+            documents = list(documents)
+            metadatas = list(metadatas)
 
         self.collection.add(
             documents=documents,


Harin329:
The `existing_ids = set(existing_docs["ids"])` call looks like it's useless, if there ever actually is a common id, the chroma db get will fail.

We need to be sure that `ids = embeddings_data["ids"]` is already unique and also make sure it's synced with documents and metadatas. So in the base chunker, we can do something like this:

```
chunk_id = hashlib.sha256((chunk + url).encode()).hexdigest()
if (chunk_id in ids):
      chunk_id = hashlib.sha256((chunk + url + str(random.getrandbits(128))).encode()).hexdigest()
ids.append(chunk_id)
documents.append(chunk)
metadatas.append(meta_data)
```

But the `chunk_id in ids` check is wasteful and will affect performance. So my question is why **_not_** have a random bit in every chunk to keep them unique?

TrentConley:
Also running into this problem when I am trying to embed a large dataset. I think the random id would fix the issue, but won't we see it affect the retrieving functionality? Two chunks that are identical should not be embedded. The random id circumnavigates the root problem. Maybe we could check to see if the id already exists and if it does then skip to the next chunk? I'm happy to contribute, though I am new to the project and this tech.

TrentConley:
I made a PR #165 that solves the issue

Harin329:
I don't think we should be ignoring the chunks that are duplicated. Ultimately, they get stored in the db and queried with 1 result

```
result = self.collection.query(
            query_texts=[input_query,],
            n_results=1,
        )
```

but there's an open PR that is looking to change that to an arbitrary n_results. 

It's a question of prompt engineering as that's what these chunks ultimately become. To me there's no harm in having more data, especially if it's present naturally in the dataset.

TrentConley:
I'd like to distinguish conversation history from database context here. In a conversation, yes you would want to keep repeats. However, in a vector db that stores things for you to look up, ideally, you'd want distinct chunks of information so you can have an LLM to pick which best provides context to the question. Identical chunks would take up tokens, and potentially make you miss out on important context that was edged out by the repeated chunks. 

So, when chunking to store in vector db, you want to ignore repeats.

Also, I think the issue here was chunking large files like a book to embed in chroma db? Where are conversations coming in here?

Feel free to poke holes in my reasoning, or if I mischaracterized what you're saying :)

Harin329:
I'm not referring to a conversation, I'm referring to natural repetitions within a dataset. Having repeated chunks in the dataset can suggest to the model that a piece of information is more relevant. The repeated `Leo: "Hey!"` phrasing would have been present and highly repeated in the original dataset.

TrentConley:
Gotcha. How would a model be able to discern that the chunks are distinct if they are identical and thus infer relevance? In actuality, I'd imagine the chunks would be larger than small repeated phrases, and those repeated phrases would be contained within a chunk that the LLM could then use.

Harin329:
Think of it this way. Here's my sample dataset, it's a PDF of a conversation:

```
Me: "I have something important to tell you!"
Leo: "Tell me!"
Leo: "Hello??"
Leo: "Hello??"
Leo: "Hello??"
Me: "Hi, sorry I was busy.
```

Assume for simplicity we end up chunking each line. With the salt, we'll end up with 6 chunks. With an in-array check, we end up with 4 chunks.

These get stored in the db. Now I come in with an input query to my bot "I have something important to tell you!" If we used the array check, the db will only think there is a single Leo: "Hello??" and it will not provide the prompt context that the bot could be repetitive in nature.

The prompt context (with n=4) will look like this:
```
Me: "I have something important to tell you!"
Leo: "Tell me!"
Leo: "Hello??"
Me: "Hi, sorry I was busy.
```

instead of 
```
Me: "I have something important to tell you!"
Leo: "Tell me!"
Leo: "Hello??"
Leo: "Hello??"
```

The inference model will likely react differently given the two scenarios. And if for any reason you're trying to capture repetitive behavior, it will fail to do that as repetitions simply won't exist in the chroma context db. At the end of the day, the whole purpose of this chunking is to help create the prompt. In which case, I think we should keep as much data in it's original format as possible from the dataset provided.

Harin329:
> Gotcha. How would a model be able to discern that the chunks are distinct if they are identical and thus infer relevance? In actuality, I'd imagine the chunks would be larger than small repeated phrases, and those repeated phrases would be contained within a chunk that the LLM could then use.

From what it looks like, chroma db just queries given an input
```
result = self.collection.query(
            query_texts=[input_query,],
            n_results=1,
        )
```

I imagine this means it will return n_results most similar in the embedding space, which _should_ include repetitions if that is a characteristic of the original dataset. The model doesn't need to discern uniqueness from the dataset, the dataset should provide the most relevant embeddings as a prompt to the model. But yes you're probably right that we're oversimplifying how the chunks will look like,  but I think the idea still stands. As the purpose is for prompt creation, repeated chunks will create repeated context which could be intended by the user.

TrentConley:
I understand your concern that in some cases you'd want to retain the identical chunks. Perhaps an ideal solution would be to create an option for the user to embed identical chunks, using your method of adding a random bit.

Harin329:
@cachho mentioned a good point about `i just dump new stuff in there and click my train button. With the random bits, it would train the whole database every time, right. So it's not just inserting what's missing.` So in that case random bits would cause issues.

Made modifications to make the ID duplicate finding more time efficient and also added a type cast, which would cause failures otherwise with that method

Harin329:
Removed the list metadata issue fix exposed by this change as that's solved in #110

embedchain/chunkers/base_chunker.py
```diff
@@ -25,9 +26,11 @@ def create_chunks(self, loader, src):
 
             for chunk in chunks:
                 chunk_id = hashlib.sha256((chunk + url).encode()).hexdigest()
-                ids.append(chunk_id)
-                documents.append(chunk)
-                metadatas.append(meta_data)
+                if (idMap.get(chunk_id) == None):
```
cachho, line 29: when comparing to None, `is` is preferred.

cachho:
This isn't as scary as it looks, @taranjeet green light from me if they fix the last complaint I posted. thanks.

embedchain/chunkers/base_chunker.py
Harin329: Done :+1:

taranjeet:
tried your PR. it prevents inserting a chunk twice, which is a good way to handle the error.
The only concern that I have is user will want to refer some chunks above and some chunks below a particular chunk.

And in this case, if we insert a chunk only once, and then for all the other occurences of the chunk, pre/post context will be missing. Also for this we will have to introduce a concept of sequence id.
Right now merging this PR, but opening an issue to track this request.

We will have to introduce a concept of sequence id first, so that we can get 3 (for eg) chunks above the chunk and 3 below.
We will also have to figure out to insert repeating chunks into the db. We are skipping inserting repeating chunks and inserting only one time, but we need to figure out a way.

taranjeet:
@Harin329 @cachho: good work on this. Merged and closed.

Issue #167, opened 2023-07-06
Add config support for chat endpoint.


taranjeet:
- like we added support for init, add and query, we should add support for chat endpoint.
- PR which adds support for init, add and query: https://github.com/embedchain/embedchain/pull/158

taranjeet:
cc @cachho we can leave this issue for sometime. It can be a good first issue for anyone new to the project.

cachho:
there is a `ChatConfig.py`. Or what do you mean?

gasolin:
@cachho @taranjeet currently we have 3 similar stuff
- query: one time query, use QueryConfig
- chat: query with history, use ChatConfig
- dryrun: show query prompt, no Config

feel they could all share the same `QueryConfig` to reduce the complexity

cachho:
fixed

PR #173, opened 2023-07-07
fix:pass QueryConfig to dry_run

gasolin:
`dry_run` func has `if config is None:` check but not able to pass QueryConfig

related to #167


cachho:
Yeah I caught that too, sorry about that @taranjeet please merge.

taranjeet (approved):

Issue #170, opened 2023-07-06
[Feature Request] Custom Prompt


cachho:
Config based custom prompts on `.query` level

PR #177, opened 2023-07-07
[Feat] Add config based template to query function

ianupamsingh:
Resolves: #170 
Changelog:
- Adds `template` as an attribute of `QueryConfig`
- Adds validation to custom `template`
- Adds `template` as an argument to `generate_prompt`

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,7 +317,7 @@ This section describes all possible config options.
 
 |option|description|type|default|
 |---|---|---|---|
-|template|custom template for prompt|Template|DEFAULT_PROMPT_TEMPLATE|
+|template|custom template for prompt|Template|Template("Use the following pieces of context to answer the query at the end. If you don't know the answer, just say that you don't know, don't try to make up an answer. \$context Query: $query Helpful Answer:")|
 
 #### **Chat Config**
 
--- a/README.md
+++ b/README.md
@@ -315,7 +315,9 @@ This section describes all possible config options.
 
 #### **Query Config**
 
-*coming soon*
+|option|description|type|default|
+|---|---|---|---|
+|template|custom template for prompt|Template|DEFAULT_PROMPT_TEMPLATE|
 
 #### **Chat Config**
 
--- a/embedchain/config/QueryConfig.py
+++ b/embedchain/config/QueryConfig.py
@@ -28,7 +28,8 @@ class QueryConfig(BaseConfig):
         :param template: Optional. The `Template` instance to use as a template for prompt.
         :raises ValueError: If the template is not valid as template should contain $context and $query
         """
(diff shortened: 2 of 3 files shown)


cachho:
Interesting take, I would have just made the initial sentence customizable and leave the rest as it is.

embedchain/config/QueryConfig.py
```diff
@@ -1,8 +1,35 @@
...
-    def __init__(self):
-        pass
+    def __init__(self, template: Template = None):
+        """
+        Initializes the QueryConfig instance.
+
+        :param template: Optional. The `Template` instance to use as a template for prompt.
+        :raises ValueError: If the template is not valid as template should contain $context and $query
+        """
+        template = template if template else DEFAULT_PROMPT_TEMPLATE
```
cachho, line 31: Can you make this 

```python
if template is None:
    template = DEFAULT_PROMPT_TEMPLATE
```

?

embedchain/config/QueryConfig.py
```diff
@@ -1,8 +1,35 @@
...
+        if not (re.search(query_re, template.template) \
```
cachho, line 32: can't you use `if $query not in template.template` ?

ianupamsingh:
Wouldn't this allow for more customization? For e.g. if I want it to reply me as a chatbot using "I" language rather than simply giving me information. In this way I can create custom characters. This is how imagined it would be used.

But if it adds more vulnerability, I can always change it. Do share your thoughts.

embedchain/config/QueryConfig.py
ianupamsingh: Template allows both $query and ${query} formats. hence i would have to add two in operators

cachho:
No your idea is better. I accept it.

cachho:
Please add this to the config section in the readme. Include the variables that need to be included in the template. Since this is an important feature, and passing a template object isn't the "default behavior", please try to put an example too.

embedchain/config/QueryConfig.py
cachho: I still think that's easier to understand and we don't have to import regex.

README.md
```diff
@@ -315,7 +315,9 @@ This section describes all possible config options.
 
 #### **Query Config**
 
-*coming soon*
+|option|description|type|default|
+|---|---|---|---|
+|template|custom template for prompt|Template|DEFAULT_PROMPT_TEMPLATE|
```
cachho, line 320: Default should be the whole DEFAULT_PROMPT_TEMPLATE so people know what they're working with.

cachho (approved):

cachho:
please also add the example to readme section Advanced.Configuration.Example - thanks

taranjeet (approved):

PR #189, opened 2023-07-07
feat: Adding app id in metadata while reading and writing to vector db

shashank42:
## Requirement
I want to use the PersonApp for multiple parallel conversations within my larger codebase. I am able to generate conversation_id on my side to identify, say a single chat thread. I want to be able to initialize the app back with the same conversation_id because the conversations can run long.

### Made 3 changes

1. id while initializing app by adding `id` field to InitConfig, generating if not provided
2. Added `app_id` field in metadata for each document while writing to vector db
3. Added `{"app_id": self.config.id}` in querying documents from vector db

Does this make sense?


cachho:
makes sense, looks good to me.

Although (for future readers) this is to section the questions you add. It's not session management (chat history).

cachho:
> I want to be able to initialize the app back with the same conversation_id because the conversations can run long and I don't want to hold things inmemory.

What's this supposed to mean? how is this supposed to work?

embedchain/config/InitConfig.py
```diff
@@ -1,12 +1,14 @@
 import os
-
 from embedchain.config.BaseConfig import BaseConfig
+import uuid
```
cachho, line 3: why don't you add imports from the standard library in the first block and then internal imports. thanks.

embedchain/config/InitConfig.py
```diff
@@ -1,12 +1,14 @@
...
+
 
 class InitConfig(BaseConfig):
+
```
cachho, line 7: remove this empty line please

embedchain/config/InitConfig.py
cachho: actually it's not standard library, I see you had to add a dependency, but still.

cachho:
I guess the biggest question I have is: why not use a collection. Isn't that exactly what collections are made for?

embedchain/config/InitConfig.py
shashank42: Agreed

shashank42:
> I guess the biggest question I have is: why not use a collection. Isn't that exactly what collections are made for?

Do you mean a new collection for every app instance?
Since all the documents are essentially chat items I didn't think it would make sense to keep items of the same type in different collections when tagging them is an option.

shashank42:
> > I want to be able to initialize the app back with the same conversation_id because the conversations can run long and I don't want to hold things inmemory.
> 
> What's this supposed to mean? how is this supposed to work?

My bad, removed the last line I wrote. I just meant that I want to "section the questions you add" as you mentioned.

shashank42:
I'm not sure if the config init function should create an id if not provided.

cachho:
So what is this about. Do you want to

a) have one database with embeddings about cats and one with embeddings about dogs and swap them out at will
b) have conversation history?

shashank42:
> So what is this about. Do you want to
> 
> a) have one database with embeddings about cats and one with embeddings about dogs and swap them out at will b) have conversation history?

Conversation history

cachho:
Ok I now understand what you are trying to do. 

With this PR, the average user would have to assign an id, otherwise they can't use message history with a new app instance. 

I personally think client side session management is the superior way to manage sessions (#198). Unless we add a database, but then the client also has to assign a session id.

Where I can see this concept is as part of the `chat` config. But at this point the app is already initialized, so I guess that would need more digging.

My closing thought: I don't think creating new app instances with separate ids and embeddings is the way to handle session management and conversation history.

cachho:
If you assign "embedchain" as an id instead of a random id, I would accept it, because generally making the id configurable can't hurt, and it won't take anything away from the average user.

shashank42:
> If you assign "embedchain" as an id instead of a random id, I would accept it, because generally making the id configurable can't hurt, and it won't take anything away from the average user.

So you are essentially suggesting to not generate a random id if not provided. This will keep the default behaviour simple.

setup.py
```diff
@@ -32,5 +32,6 @@
         "gpt4all",
         "sentence_transformers",
         "docx2txt",
+        "uuid"
```
deshraj, line 35: I don't think you need it since python already has `uuid`.

embedchain/config/InitConfig.py
```diff
@@ -27,6 +29,11 @@ def __init__(self, ef=None, db=None):
             self.db = ChromaDB(ef=self.ef)
         else:
             self.db = db
+            
+        if id is None:
+            self.id = str(uuid.uuid4())
+        else:
+            self.id = id
```
deshraj, line 36: Can rewrite the whole thing in one line:

```python
self.id = id if id is not None else str(uuid.uuid4())
```

deshraj (changes requested):

embedchain/config/InitConfig.py
shashank42: Removed uuid altogether

setup.py
shashank42: Removed this

shashank42 (commented):
Removed uuid

taranjeet:
Hi @deshraj : let me know if the review and changes requested from your side is done. Will proceed for review/merge.

cachho:
please run `make format lint`. Thank you.

taranjeet (approved):

Issue #181, opened 2023-07-07
[docs] add example for custom prompt Template to readme


cachho:
with #170, custom prompts became a reality. But it's not super easy to use them. We need an example in ReadMe.Advanced.Configuration.Example. You can get creative.

cachho:
@ianupamsingh maybe for you to complete this.

PR #199, opened 2023-07-08
[docs] Add example for custom prompt Template to README

ianupamsingh:
Resolves: #181

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,53 @@ query_config = QueryConfig() # Currently no options
 print(naval_chat_bot.query("What unique capacity does Naval argue humans possess when it comes to understanding explanations or concepts?", query_config))
 ```
 
+Here's the example of using custom prompt template with `.query`
+
+```python
+from embedchain.config import QueryConfig
+from embedchain.embedchain import App
+from string import Template
+import wikipedia
+
+einstein_chat_bot = App()
+
+# Embed Wikipedia page
+page = wikipedia.page("Albert Einstein")
+einstein_chat_bot.add("text", page.content)
+
+# Example: use your own custom template with `$context` and `$query`
+einstein_chat_template = Template("""
+        You are Albert Einstein, a German-born theoretical physicist,
+        widely ranked among the greatest and most influential scientists of all time.
+
+        Use the following information about Albert Einstein to respond to 
+        the human's query acting as Albert Einstein.
+        Context: $context                                
+
+        Keep the response brief. If you don't know the answer, just say that you don't know, don't try to make up an answer.
+
+        Human: $query
+        Albert Einstein:""")
+query_config = QueryConfig(einstein_chat_template)
+queries = [
+        "Where did you complete your studies?",
+        "Why did you win nobel prize?",
+        "Why did you divorce your first wife?",
+]
+for query in queries:
+        response = einstein_chat_bot.query(query, query_config)
+        print("Query: ", query)
+        print("Response: ", response)
+
+# Output
+# Query:  Where did you complete your studies?
+# Response:  I completed my secondary education at the Argovian cantonal school in Aarau, Switzerland.
+# Query:  Why did you win nobel prize?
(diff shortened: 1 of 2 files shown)


embedchain/embedchain.py
```diff
@@ -243,7 +243,7 @@ def dry_run(self, input_query, config: QueryConfig = None):
         if config is None:
             config = QueryConfig()
         context = self.retrieve_from_database(input_query)
-        prompt = self.generate_prompt(input_query, context)
+        prompt = self.generate_prompt(input_query, context, config.template)
```
taranjeet, line 246: Hey @ianupamsingh : didnt understand this code? can you please explain.

README.md
```diff
@@ -324,6 +324,53 @@ query_config = QueryConfig() # Currently no options
...
+
+        Human: $query
+        Albert Einstein:""")
+query_config = QueryConfig(einstein_chat_template)
+queries = [
+        "Where did you complete your studies?",
+        "Why did you win nobel prize?",
+        "Why did you divorce your first wife?",
+]
+for query in queries:
```
cachho, line 360: no need to make it this complicated. just give it one query. that's easier to understand.

embedchain/embedchain.py
cachho: He's fixing a bug that found its way into prod: Dry run does not use the template config. This is ok.

cachho: although in the future this should be a separate PR.

README.md
```diff
@@ -324,6 +324,53 @@ query_config = QueryConfig() # Currently no options
...
+```python
+from embedchain.config import QueryConfig
+from embedchain.embedchain import App
+from string import Template
+import wikipedia
+
+einstein_chat_bot = App()
+
+# Embed Wikipedia page
+page = wikipedia.page("Albert Einstein")
```
cachho, line 338: first of all, you're using Einstein here, nothing wrong with that, just @taranjeet  said before that he wants all examples to be about Naval.

Secondly, I'm not sure if we want the complexity. of adding another import `wikipedia` just for this example

embedchain/embedchain.py
cachho: If I'm quick I'll open a separate PR for that. This needs to be fixed asap.

cachho: addressed in #203, I hope you don't mind. but this needs to be discussed separately.

cachho: this is resolved.

README.md
ianupamsingh: I'll give it two

README.md
ianupamsingh: 1) Should I try to add the same example for Naval instead of Einstein?
2) I added `wikipedia` as it made the model answer more accurately. I tried adding the wikipedia link as a `web_page` but it didn't really gave accurate results, as the web_page was not parsed correctly. Should I try a simpler example?

PS: I also believe support for Wikipedia should be inherent to `embedchain`. It would make it more useful for someone to research a subject through wikipedia. Maybe we could add `wikipedia` as a `data_type`. Just a suggestion.

cachho: > Maybe we could add wikipedia as a data_type. Just a suggestion.

Yes I agree. 

> Should I try to add the same example for Naval instead of Einstein?

up to @taranjeet

taranjeet: > Maybe we could add wikipedia as a data_type. Just a suggestion.

We should definitely do that. @ianupamsingh : do you want to open an issue for this?

README.md
taranjeet: two is fine.

taranjeet (approved):

README.md
ianupamsingh: > > Maybe we could add wikipedia as a data_type. Just a suggestion.
> 
> We should definitely do that. @ianupamsingh : do you want to open an issue for this?

Issue created: https://github.com/embedchain/embedchain/issues/212

taranjeet: thanks

Issue #171, opened 2023-07-06
[Feature Request] Config: Chunk Size


cachho:
Config (`AddConfig`) based chunk size setting.

rjarun8:
@cachho can you share more details on this

PR #200, opened 2023-07-08
[Feat] AddConfig should allow configuring Chunker

ianupamsingh:
Resolves: #171 

Changelog:
- `AddConfig` takes two configs as parameter: `ChunkerConfig` and `LoaderConfig`
- `ChunkerConfig` is defined with following parameters
  - chunk_size
  - chunk_overlap
  - length_function
- Default values for these config parameters depend of the `data_type`
- `LoaderConfig` is empty for now
- Updated README with parameters and examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,7 +312,15 @@ config = InitConfig(ef=embedding_functions.OpenAIEmbeddingFunction(
             ))
 naval_chat_bot = App(config)
 
-add_config = AddConfig() # Currently no options
+# Example: define your own chunker config
+add_config = {
+        "chunker": {
+                "chunk_size": 1000,
+                "chunk_overlap": 100,
+                "length_function": len,
+        }
+}
+add_config = AddConfig(**add_config)
 naval_chat_bot.add("youtube_video", "https://www.youtube.com/watch?v=3qHkcs3kG44", add_config)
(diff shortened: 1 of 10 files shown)


cachho (approved):
Reviewed, looks good to me.

README.md
```diff
@@ -312,7 +312,15 @@ config = InitConfig(ef=embedding_functions.OpenAIEmbeddingFunction(
             ))
 naval_chat_bot = App(config)
 
-add_config = AddConfig() # Currently no options
+# Example: define your own chunker config
+add_config = {
```
taranjeet, line 316: add_config should work at data_type level, so in the example we should show that we are overriding chunking for a particular config and not everything. Otherwise the users will feel consued.

Can you update?

README.md
```diff
@@ -337,6 +345,32 @@ This section describes all possible config options.
 
 #### **Add Config**
 
+|option|description|type|default|
+|---|---|---|---|
+|chunker|chunker config|ChunkerConfig|{"chunk_size": 4000, "chunk_overlap": 200, "length_function": len}|
```
taranjeet, line 350: default value is category specici here.

taranjeet (approved):
left some comments, rest looks good

README.md
ianupamsingh: So you mean I should only change add_config for one of the data_type (e.g. youtube) instead of all data_types?

README.md
ianupamsingh: I'll just mention to refer the table below for default values of different data_types

PR #203, opened 2023-07-08
fix: use custom prompt in dry run

cachho:


cachho:
caught by @ianupamsingh

taranjeet (approved):

Issue #109, opened 2023-07-01
[Thoughts] We should consider the output of the .add method


cachho:
Right now, the '.add' method does not return anyything, but prints after success.

For some reason this seems to almost be async, when I submit three and print the output I get 
```
None
None
None
Sucessfully saved ... Total chunks count: ...
Sucessfully saved ... Total chunks count: ...
Sucessfully saved ... Total chunks count: ...
```
but that's not the point.

**I just think it should be up to the user whether they want to print the results or not.**

If you ask me, it should return, `return document, metadata, number_of_chunks`. (that's not what the variables are called right now).
Then the user could print their own success message if they wanted to.
This would also allow to process the number of chunks if they have a use case for that.

Just a thought though.

cachho:
Also, the already exists in the database should raise an Error I think. For the same reason: Let the user decide how to handle it.
Let's say you have a daily task to embed a sitemap with 1000 pages. You might not want to see the 999 that are the same as they were yesterday.

I know this sort-of breaks compatibility though.

cachho:
Also: the way I see it, self.collection.count() returns the total count, it makes more sense to return the count of the added item here.

taranjeet:
@cachho : feel free to close this issue, if all addressed here.

cachho:
> @cachho : feel free to close this issue, if all addressed here.

Not addressed, it still uses print instead of logging.

taranjeet:
@cachho : we plan to return the id of the source in add function.
If user wants to see what's happening in the add function, can use dry_run flag. (tracking in #399 ).

Closing this. Feel free to reopen if I am missing something.

cachho:
We still haven't decided if we want to stick to printing the success message.

PR #207, opened 2023-07-09
feat: add method - only count new chunks and print them

cachho:
this PR makes it so that the `add` method only prints the number of new chunks, instead of the chunk total.

If users are interested in the total amount of chunks, they can use the `count` method at any time, but I think when you add a web page for instance, you want to know how big it is.


cachho:
referencing: #109

taranjeet:
Looks good to me.

PR #214, opened 2023-07-10
Modified streaming setup to use QueryConfig

aaishikdutta:
Removed streaming_option from InitConfig and added it at a Query level.
Added some internal refactors


README.md
```diff
@@ -204,19 +204,6 @@ from embedchain import PersonApp as ECPApp
 print(naval_chat_bot.query("What unique capacity does Naval argue humans possess when it comes to understanding explanations or concepts?"))
 # answer: Naval argues that humans possess the unique capacity to understand explanations or concepts to the maximum extent possible in this physical reality.
 ```
-### Stream Response
-
-- You can add config to your query method to stream responses like ChatGPT does. You would require a downstream handler to render the chunk in your desirable format
-
-- To use this, instantiate App with a `InitConfig` instance passing `stream_response=True`. The following example iterates through the chunks and prints them as they appear
-```python
-app = App(InitConfig(stream_response=True))
```
taranjeet, line 213: can we keep the parameter name as "stream"
stream is self explanatory and implies that response will be streamed.

README.md
```diff
@@ -235,6 +222,22 @@ print(naval_chat_bot.chat("what did the author say about happiness?"))
...
 
+### Stream Response
+
+- You can add config to your query method to stream responses like ChatGPT does. You would require a downstream handler to render the chunk in your desirable format. Currently only supports OpenAI model.
+
+- To use this, instantiate a `QueryConfig` or `ChatConfig` object with `stream_response=True`. Then pass it to the `.chat()` or `.query()` method. The following example iterates through the chunks and prints them as they appear.
+
+```python
+app = App()
+query_config = QueryConfig(stream_response = True)
```
taranjeet, line 233: lets call the arg as "stream"

README.md
```diff
@@ -235,6 +222,22 @@ print(naval_chat_bot.chat("what did the author say about happiness?"))
...
+resp = naval_chat_bot.query("What unique capacity does Naval argue humans possess when it comes to understanding explanations or concepts?", query_config)
+
+for chunk in resp:
```
taranjeet, line 236: should we give a downstream handler? assuming that a lot of users will use streaming response, so why not there is a function which when called gives the streaming response output.
@cachho @aaishikdutta

README.md
```diff
@@ -204,19 +204,6 @@ from embedchain import PersonApp as ECPApp
...
 # answer: Naval argues that humans possess the unique capacity to understand explanations or concepts to the maximum extent possible in this physical reality.
 ```
-### Stream Response
-
-- You can add config to your query method to stream responses like ChatGPT does. You would require a downstream handler to render the chunk in your desirable format
-
-- To use this, instantiate App with a `InitConfig` instance passing `stream_response=True`. The following example iterates through the chunks and prints them as they appear
-```python
-app = App(InitConfig(stream_response=True))
-resp = naval_chat_bot.query("What unique capacity does Naval argue humans possess when it comes to understanding explanations or concepts?")
```
cachho, line 214: app instance is called `app` not `naval_chat_bot` (common mistake)

taranjeet (changes requested):

README.md
cachho: not everyone's going to want to just print it. I haven't looked into how I can get route the stream through flask, but that's my usecase for instance, so no, I'm fine with it how it is.

Streaming is for people that know a little. Not for private users that want to play around. If you want to use it professionally you should be able to copy two lines from the readme.

README.md
cachho: fine with me. I wish there was a name that implied a boolean action, maybe `is_stream`?

embedchain/embedchain.py
```diff
@@ -224,7 +224,7 @@ def chat(self, input_query, config: ChatConfig = None):
             context,
             chat_history=chat_history,
         )
-        answer = self.get_answer_from_llm(prompt)
+        answer = self.get_answer_from_llm(prompt, config)
         memory.chat_memory.add_user_message(input_query)
```
cachho, line 228: does that mean that streaming won't work with chat?

README.md
taranjeet: Okay, then for now lets leave this.

embedchain/embedchain.py
aaishikdutta: it will

cachho: sorry for the oversight.

taranjeet (approved):

PR #221, opened 2023-07-11
[code-quality]: Remove imports from `__init__.py` files

deshraj:
Fixes https://github.com/embedchain/embedchain/issues/216

Still keeping the `App` related imports in `__init__.py` file for now but we will change that later. 

Tested locally by running tests.

```bash
(base) ➜  embedchain git:(remove-init-imports) make lint 
python3 -m ruff .
(base) ➜  embedchain git:(remove-init-imports) make test 
python3 -m pytest
====================================================================================== test session starts ======================================================================================
platform darwin -- Python 3.11.3, pytest-7.4.0, pluggy-1.2.0
rootdir: /Users/deshrajyadav/Projects/embedchain
plugins: anyio-3.7.1
collected 2 items                                                                                                                                                                               

tests/test_embedchain.py ..                                                                                                                                                               [100%]

======================================================================================= 2 passed in 1.54s =======================================================================================
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,7 +367,9 @@ Here's the readme example with configuration options.
 ```python
 import os
 from embedchain import App
-from embedchain.config import InitConfig, AddConfig, QueryConfig
+from embedchain.config.AddConfig import AddConfig
+from embedchain.config.InitConfig import InitConfig
+from embedchain.config.QueryConfig import QueryConfig
 from chromadb.utils import embedding_functions
 
 # Example: use your own embedding function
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -367,7 +367,9 @@ Here's the readme example with configuration options.
 ```python
 import os
 from embedchain import App
-from embedchain.config import InitConfig, AddConfig, QueryConfig
+from embedchain.config.AddConfig import AddConfig
```
taranjeet, line 370: I agree to your point about not keeping all the imports in `__init__` file and understand its pro/cons as well.

But this pattern is confusing and not simple for the user.

Can we following something like django? https://github.com/django/django/blob/main/django/db/models/__init__.py

```
from django.db import models
```

deshraj: If we do like django, the story remains the same that we will be importing a lot of modules unintentionally. We probably want to give small names to modules to make the path look cleaner to the user going forward.

I will do some refactoring soon to make sure that it is looks elegant to the user.

deshraj:
This PR is stale now since we have incorporate the similar changes in a different smaller PRs. Hence, closing this one.

PR #224, opened 2023-07-11
add star history diagram

xumeng:
diagram url: https://api.star-history.com/svg?repos=embedchain/embedchain&type=Date


deshraj:
Thanks for the contribution @xumeng. However, I don't think we want to add the star history in the README. Other contributions are most welcome. :)

PR #205, opened 2023-07-08
fix: initialize embedding function manually

cachho:
Resolves: #190

PR #236, opened 2023-07-11
fix: do not create instance in InitConfig `__init__`

cachho:
If the instance is created in the init method, it's created even if you want to use an open source model. That's why you need to create it later, depending on the class of the app. That's what the methods `_set_db_to_default` and `_set_embedding_function_to_default` are for.

### DIFF
--- a/embedchain/config/InitConfig.py
+++ b/embedchain/config/InitConfig.py
@@ -20,27 +20,11 @@ class InitConfig(BaseConfig):
         """
         self._setup_logging(log_level)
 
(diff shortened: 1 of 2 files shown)


cachho:
This redoes what was done in #205, somehow the changes were reverted.

deshraj (approved):

deshraj:
Thanks for the fix @cachho. We should add tests going forward to make sure that these kinds of regressions don't happen again.

PR #250, opened 2023-07-12
test: added tests

cachho:
some of these are smart, some are more simple. But better than nothing I guess.


tests/chunkers/test_text.py
```diff
@@ -0,0 +1,46 @@
...
+        return [
+            {
+                "content": src,
+                "meta_data": {"url": "none"},
+            }
+        ]
+
+
+if __name__ == "__main__":
+    unittest.main()
```
deshraj, line 46: You don't need this I think.

tests/chunkers/test_text.py
```diff
@@ -0,0 +1,46 @@
+# ruff: noqa: E501
```
deshraj, line 1: Any specific reason to add this? :D

tests/embedchain/test_add.py
```diff
@@ -0,0 +1,26 @@
+# ruff: noqa: E501
```
deshraj, line 1: Let's remove this from all the test files. We only want to specific specific lines and not all of the file.

tests/chunkers/test_text.py
cachho: otherwise it complains about all the doc strings.

tests/embedchain/test_add.py
cachho: that's going to be a lot of lines and decrease readability.

taranjeet:
@deshraj waiting for your review, since you are driving adding tests feature.

deshraj (approved):

PR #254, opened 2023-07-13
Store package version in py file

deshraj:
Going forward, we will store the package version in a py file so that we can check the version that is being used in python env.

Users can use something like this to know what version they are using:

```python
>>> import embedchain.version
>>> embedchain.version.__version__
'0.0.21'
```


cachho:
Isn't it usually just `embedchain.__version__`?

taranjeet (approved):

PR #268, opened 2023-07-14
chore: remove duplicate dev requirements

cachho:
Dev requirements are already in `setup.py`. From there you already install them with `make install`. so this is redundant.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,5 +1,0 @@
-pip
-black==23.3.0
-isort==5.8.0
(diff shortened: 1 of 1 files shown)


taranjeet:
tagging @deshraj for review here.

taranjeet (approved):

taranjeet:
Good catch @cachho

deshraj:
Plan was to move the requirements out of the setup.py file but its fine.

PR #282, opened 2023-07-15
fix: template for personapp

sahilyadav902:
Fix: PersonApp and PersonOpenSourceApp would give errors when using the chat function


taranjeet:
Good catch @sahilyadav902

PR #289, opened 2023-07-17
fix: Dont initialize chroma and embedding function in init config.

taranjeet:
## Description

When creating an app using the latest commit, was getting following error

```
ids, embeddings, metadatas, documents = self._validate_embedding_set(
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Tried the published package and realized that db is getting initialized in init config where the embedding function is not available yet.

## Type of change

Bug fix (non-breaking change which fixes an issue)


deshraj (approved):

deshraj:
Good catch

PR #303, opened 2023-07-17
feat: rate limiter for add

cachho:
## Description

This adds rate limit handling using exceptions.

One of the difficulties implementing this is that with the default behavior (giving Chroma all the documents at once) if it fails you don't know when it fails, so you have no other option than to start over and try adding them one by one with exception handling, which is going to lead to duplicates.

To reduce the risk of this happening, I added a `char_in_documents` variable that can be compared to the set `config.rate_limit`. I know chars != tokens, but it's a start. If there are more chars than the rate limit, it's going down the one by one path right away, without trying to dump all at once, for a cleaner execution.

I know this whole code is a mess, but with the error handling, character counting etc, it's bound to be.

I learned that when I added them one by one I did not get rate limited at all, for the same sitemap that previously gave me errors. So we could restructure the sitemap to use individual `add` function calls for each page, instead of putting them in an array and trying to do it all at once. `This might be the way to go after all.

I also tried to implement a real rate limit handler, that would keep track of characters used, and try to calculate a rate limit client side. But it was even more messy, and based on too many assumptions. Just going until you hit an error seemed like the better way to go.

My progress in cleaning up the actual web pages and filtering the sitemap is more fruitful, and brought me to a place where I reduced the tokens by more than 90% by excluding images. We should still handle rate limiting somehow, but I guess right now we can do more improvements to what we're feeding in. That's why I post this as a draft.

**🚨Please let me know if you hit a rate limit and need this feature, and what resource you are trying to add🚨**

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

- [ ] New and existing unit tests pass locally with my changes

## Maintainer Checklist

- [ ] Made sure Checks passed

### DIFF
--- a/embedchain/embedchain.py
+++ b/embedchain/embedchain.py
@@ -9,7 +9,9 @@ from langchain.docstore.document import Document
 from langchain.memory import ConversationBufferMemory
 
 from embedchain.config import AddConfig, ChatConfig, InitConfig, QueryConfig
-from embedchain.config.QueryConfig import DOCS_SITE_PROMPT_TEMPLATE, DEFAULT_PROMPT, DEFAULT_PROMPT_WITH_HISTORY
+from embedchain.config.QueryConfig import (DEFAULT_PROMPT,
+                                           DEFAULT_PROMPT_WITH_HISTORY,
+                                           DOCS_SITE_PROMPT_TEMPLATE)
 from embedchain.data_formatter import DataFormatter
 
 gpt4all_model = None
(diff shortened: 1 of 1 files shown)


candidosales:
@cachho, Maybe this update will resolve the problem.

https://github.com/embedchain/embedchain/pull/330

See more about it: https://www.trychroma.com/blog/chroma_0.4.0

deshraj:
Closing this PR since this problem is already solved.

PR #320, opened 2023-07-19
refactor: add local to use add method

cachho:
## Description

Add_local should be the same as add, just that instead of calling the argument `url` it's called `content` to drive home the point, that you're not passing a URL.

A discrepancy already found it's way in the code, so I think it only makes sense to call add from add_local to make it more maintainable.

This PR also includes a documentation update.

Fixes # (issue)

## Type of change

- [X] Documentation update

## How Has This Been Tested?

- [X] Unit Test

## Checklist:

## Maintainer Checklist

- [ ] Made sure Checks passed


taranjeet:
@cachho : I think it makes much more sense to remove the `add_local` function and whatever functionality was there, just port it to `add`.

so ideally

```
bot = App()
bot.add("qna_pair", ("Who is Naval Ravikant?", "Naval Ravikant is an Indian-American entrepreneur and investor."))
```

So now add supports the same functionality as `.add_local` and `.add_local` is not there.

cachho:
> @cachho : I think it makes much more sense to remove the `add_local` function and whatever functionality was there, just port it to `add`.
> 
> so ideally
> 
> ```
> bot = App()
> bot.add("qna_pair", ("Who is Naval Ravikant?", "Naval Ravikant is an Indian-American entrepreneur and investor."))
> ```
> 
> So now add supports the same functionality as `.add_local` and `.add_local` is not there.

as I said in the post/comments, the only problem with that approach is that intellisense will ask the user for a `url` argument. This will confuse the user when they add text for instance.

cachho:
Let's just say, this approach is better and more maintainable then what we have right now. We should merge it. Then we can still think about deprecating it.

docs/advanced/data_types.mdx
```diff
@@ -72,6 +73,14 @@ To supply your own QnA pair, use the data_type as `qna_pair` and enter a tuple.
 app.add_local('qna_pair', ("Question", "Answer"))
 ```
 
+### Doc file
+
+To add any doc/docx file, use the data_type as `docx`. Eg:
+
+```python
+app.add_local('docx', 'a_local_docx_file_path')
```
taranjeet, line 81: this is not correct.
`add_local` is meant to add text which is in a Python variable. like string.

We should remove this.

taranjeet (approved):

docs/advanced/data_types.mdx
cachho: I think `add` remote is a URL string, and `add_local` is anything that's not a URL. QnA Pair also isn't a string.

Turns out, docx can do both:

> Docx2txtLoader: Defaults to check for local file, but if the file is a web path, it will download it to a temporary file, and use that, then clean up the temporary file after completion

So it can practically belong in both categories, we could argue that local file is the default so it should be under local, but I will move it to the remote section

cachho: I expanded the documentation a bit, but moved it back to where you wanted it.

taranjeet:
@cachho : i think this PR does not hold. `add_local` is meant to add text or qna pair. its not meant to add local files.
also have made a comment in PR #380 about the add_local.
feel free to close this if PR is not valid.

cachho:
> @cachho : i think this PR does not hold. `add_local` is meant to add text or qna pair. its not meant to add local files. also have made a comment in PR #380 about the add_local. feel free to close this if PR is not valid.

the whole point of this PR is that `add_local` uses the `add` method internally so that we only have to maintain one. Which basically makes it an alias, which it already is today.

With PR #380 this can go. Will close once that is merged, until then I still see this as a good idea.

cachho:
closing because this was done in #380.

PR #368, opened 2023-07-25
fix for personapp not working with config

### LINKED ISSUE #367: config doesn't work with PersonApp or PersonOpenSourceApp
### 🐛 Describe the bug

When I try to instantiate PersonApp or PersonOpenSourceApp with config it plainly ignores it.

```
from embedchain import PersonOpenSourceApp
from embedchain.config import QueryConfig
from string import Template

app = PersonOpenSourceApp("Winston Churchill")
app.add("youtube_video", "https://youtu.be/ilIefxbSJT8")
print(app.query("What is Oppenheimer about?", dry_run=True, config=QueryConfig(stream=False, template=Template("""
 Use the context to answer the query in a sarcastic tone.

  $context

  Query: $query

  Helpful Answer:
"""))))


```

The output still contains the default Prompt:
```
You are Winston Churchill. Whatever you say, you will always say in Winston Churchill style. 
  Use the following pieces of context to answer the query at the end.
  If you don't know the answer, just say that you don't know, don't try to make up an answer.

  scientists achieve progress, we have to plan for how we would deal with it. Uhm. You see this in my business right now. In the film business, people are very worried about AI and the applications of it, and they want protection from that. But the reality is, until those things progress, it's hard to know what it is that needs to be controlled. And I think a lot of scientific progress is like that. Until we see the potential, until it's starting to be realized, it's very difficult for the rest of society to legislate, to judge. And that's where the real danger comes in. What Oppenheimer saw as the real danger, and I think he was right, was secrecy. They had had to develop the Manhattan Project in intense secrecy, but they fought it. The scientists fought it every step of the way because they believed that knowledge has to be shared. It has to be open. And in that way, both the dangers and the advantages could be shared by all and could be managed by all. And I think applied to most examples of what you're asking about, what he referred to as candor, you know, as openness, is actually essential. - It's also something we see in the movie. He's, uhm. he's looking for peace. And it may seem, of course, paradoxical with the story we know of Oppenheimer, but he's claiming he wants peace, and at the same time, of course, building this. what may be a threat and is a threat, of course. And so it's a whole ethic. - Well, he was one of the first people to… he didn't refer to it as mutually assured destruction, but he was one of the first people, he and Niels Bohr, to start exploring and articulating the idea that with a weapon so powerful, the type of war we'd seen in World War II and World War I becomes impractical and unthinkable. And there is an argument, one way of looking at the situation of the world today is that he's been proven right to this point. Uhm… I don't know. (laughs) There are no. There are no easy answers to the questions that his story raises. - I think the

  Query: What is Oppenheimer about?

  Helpful Answer:

```

aaishikdutta:
## Description

Fix for PersonApp not working with config

Fixes #367 

## Type of change

## How Has This Been Tested?

```
from embedchain import PersonOpenSourceApp
from embedchain.config import QueryConfig
from string import Template

app = PersonOpenSourceApp("Winston Churchill")
app.add("youtube_video", "https://youtu.be/ilIefxbSJT8")
print(app.query("What is Oppenheimer about?", dry_run=True, config=QueryConfig(stream=False, template=Template("""
 Use the context to answer the query in a sarcastic tone.

  $context

  Query: $query

  Helpful Answer:
"""))))
```

## Checklist:

## Maintainer Checklist

- [ ] Made sure Checks passed

### DIFF
--- a/embedchain/apps/PersonApp.py
+++ b/embedchain/apps/PersonApp.py
@@ -22,6 +22,26 @@ class EmbedChainPersonApp:
         self.person_prompt = f"You are {person}. Whatever you say, you will always say in {person} style."  # noqa:E501
         super().__init__(config)
 
+    def add_person_template_to_config(self, config: ChatConfig = None):
+
+        self.template = Template(self.person_prompt + " " + DEFAULT_PROMPT)
+         
+        if config is None:
+            config = QueryConfig(
+                template=self.template,
+            )
+        
+        elif config.template is None:
+            config.template = self.template
+        
+        else:
+            config.template = Template(self.person_prompt + " " + config.template.template)
+        
+        return config
+
+        
+             
+
 
 class PersonApp(EmbedChainPersonApp, App):
     """
@@ -30,18 +50,12 @@ class PersonApp(EmbedChainPersonApp, App):
     """
 
     def query(self, input_query, config: QueryConfig = None, dry_run=False):
-        self.template = Template(self.person_prompt + " " + DEFAULT_PROMPT)
-        query_config = QueryConfig(
-            template=self.template,
-        )
-        return super().query(input_query, query_config, dry_run)
+        config = self.add_person_template_to_config(config)
+        return super().query(input_query, config, dry_run)
 
     def chat(self, input_query, config: ChatConfig = None, dry_run=False):
-        self.template = Template(self.person_prompt + " " + DEFAULT_PROMPT_WITH_HISTORY)
-        chat_config = ChatConfig(
-            template=self.template,
-        )
-        return super().chat(input_query, chat_config, dry_run)
+        config = self.add_person_template_to_config(config)
+        return super().chat(input_query, config, dry_run)
 
 
 class PersonOpenSourceApp(EmbedChainPersonApp, OpenSourceApp):
@@ -51,15 +65,9 @@ class PersonOpenSourceApp(EmbedChainPersonApp, OpenSourceApp):
     """
 
     def query(self, input_query, config: QueryConfig = None, dry_run=False):
-        self.template = Template(self.person_prompt + " " + DEFAULT_PROMPT)
-        query_config = QueryConfig(
-            template=self.template,
-        )
-        return super().query(input_query, query_config, dry_run)
+        config = self.add_person_template_to_config(config)
+        return super().query(input_query, config, dry_run)
 
     def chat(self, input_query, config: ChatConfig = None, dry_run=False):
-        self.template = Template(self.person_prompt + " " + DEFAULT_PROMPT_WITH_HISTORY)
-        chat_config = ChatConfig(
-            template=self.template,
-        )
-        return super().chat(input_query, chat_config, dry_run)
+        config = self.add_person_template_to_config(config)
+        return super().chat(input_query, config, dry_run)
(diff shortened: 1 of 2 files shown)


cachho:
*I'm gonna be piggy-backing off of your PR here, I know you probably had a narrower scope and just did what needed to be done to fix the problem you were facing*

embedchain/apps/PersonApp.py
```diff
@@ -51,15 +62,9 @@ class PersonOpenSourceApp(EmbedChainPersonApp, OpenSourceApp):
     """
 
     def query(self, input_query, config: QueryConfig = None, dry_run=False):
-        self.template = Template(self.person_prompt + " " + DEFAULT_PROMPT)
-        query_config = QueryConfig(
-            template=self.template,
-        )
-        return super().query(input_query, query_config, dry_run)
+        config = self.add_person_template_to_config(DEFAULT_PROMPT, config)
```
cachho, line 65: The implication of this (I think that's also how it is right now) is that only default prompts can be used, right? Because DEFAULT_PROMPT is passed anyways.

If that's intentional, we should either create subclass of QueryConfig that does not allow template modification or, simpler, log a warning to the user if template is set, that it will be ignored.

embedchain/apps/PersonApp.py
```diff
@@ -22,6 +22,23 @@ def __init__(self, person, config: BaseAppConfig = None):
         self.person_prompt = f"You are {person}. Whatever you say, you will always say in {person} style."  # noqa:E501
         super().__init__(config)
```
cachho, line 23: shouldn't this throw an error? What super is it calling, it does not have a parent class.

embedchain/apps/PersonApp.py
cachho: I now see that I am wrong, I think we should add an inline comment for this to make it clearer.

embedchain/apps/PersonApp.py
```diff
@@ -22,6 +22,23 @@ def __init__(self, person, config: BaseAppConfig = None):
...
 
+    def add_person_template_to_config(self, default_prompt: str, config: ChatConfig = None):
+
+        self.template = Template(self.person_prompt + " " + default_prompt)
```
cachho, line 27: Maybe it's easier to understand if we turn this on its head:

```python
if config:
    if config.template:
        # Add person prompt to custom user template
        config.template = Template(self.person_prompt + " " + config.template.template)
    else:
        # If no user template is present, use person prompt with the default template
        config.template = self.template
else:
    # if no config is present at all, initialize the config with person prompt and default template
    config = QueryConfig(
                template=self.template,
            )
```

cachho, line 27: why use a attribute here? wouldn't a local variable also do the trick. The attribute is a duplicate because it then exists in app and config.

cachho (changes requested):

embedchain/apps/PersonApp.py
aaishikdutta: if I am not wrong this class is being used for multiple inheritance and hence it might be required

embedchain/apps/PersonApp.py
aaishikdutta: updated

aaishikdutta: updated

cachho (approved):

PR #369, opened 2023-07-25
add "d" on "Embedchain" in citation section title

alessandropanzieri:


taranjeet:
this is a good catch @alessandropanzieri

PR #406, opened 2023-08-08
fix: Pass deployment name as param for azure api

taranjeet:
Pass deployment name in config so that OpenAI Azure api integration works.

Example

```

import os
os.environ["OPENAI_API_TYPE"] = "azure"
os.environ["OPENAI_API_BASE"] = "https://<endpoint_assigned_to_you>.openai.azure.com"
os.environ["OPENAI_API_KEY"] = "your-id-here"

from embedchain import CustomApp
from embedchain.config import CustomAppConfig, QueryConfig
from embedchain.models import Providers, EmbeddingFunctions

config = CustomAppConfig(embedding_fn=EmbeddingFunctions.OPENAI, provider=Providers.AZURE_OPENAI, deployment_name="text-embedding-ada-002")
query_config = QueryConfig(deployment_name="embedchainpocgpt35")
satya_nadela_bot = CustomApp(config)
satya_nadela_bot.add("web_page", "https://en.wikipedia.org/wiki/Satya_Nadella")

satya_nadela_bot.query("What is the date of birth of Satya Nadella?", config=query_config)
```

### DIFF
--- a/embedchain/apps/CustomApp.py
+++ b/embedchain/apps/CustomApp.py
@@ -118,9 +118,13 @@ class CustomApp(EmbedChain):
     def _get_azure_openai_answer(prompt: str, config: ChatConfig) -> str:
         from langchain.chat_models import AzureChatOpenAI
 
+        if not config.deployment_name:
+            raise ValueError("Deployment name must be provided for Azure OpenAI")
+
         chat = AzureChatOpenAI(
-            deployment_name="td2",
-            model_name=config.model or "text-davinci-002",
+            deployment_name=config.deployment_name,
+            openai_api_version="2023-05-15",
+            model_name=config.model or "gpt-3.5-turbo",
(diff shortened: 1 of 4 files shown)


Grudra1:
> Thanks so much for this, as I'm sure a lot of people are using the Azure OpenAI version, Embedchain is amazing! :) I'm not able to run the above code as is, though.
> 
> I get an import error for the `QueryConfig` function:
> 
> ```
> ImportError: cannot import name 'QueryConfig' from 'embedchain.config'
> ```
> 
> I also get a `TypeError` when trying to run the `CustomAppConfig`
> 
> ```
> TypeError: __init__() got an unexpected keyword argument 'embedding_fn'
> ```
> 
> Many thanks for your help with this team!

Facing the same issue , would be great if we could test this with Azure Open AI !!

PR #484, opened 2023-08-25
System prompt at App level

### LINKED ISSUE #452: Add support for system prompt at App level.
### 🚀 The feature

One feedback is to have support for system prompt at App init level. Lets say somebody is creating a Harry Potter like bot, in that case, the user would prefer setting it once at the app level only and not each time during query/chat level.

from https://github.com/embedchain/embedchain/pull/448

### Motivation, pitch

make it easier for user to set system prompt once at the app level.

### Alternatives

_No response_

### Additional context

_No response_

Dev-Khant:
## Description

Add `system_prompt` on app level as well.

Fixes #452

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Run `poetry run pytest`

- [x] Unit Test

## Checklist:

### DIFF
--- a/embedchain/apps/App.py
+++ b/embedchain/apps/App.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 import openai
 
 from embedchain.config import AppConfig, ChatConfig
@@ -14,19 +16,27 @@ class App(EmbedChain):
     dry_run(query): test your prompt without consuming tokens.
     """
 
-    def __init__(self, config: AppConfig = None):
+    def __init__(self, config: AppConfig = None, system_prompt: Optional[str] = None):
         """
         :param config: AppConfig instance to load as configuration. Optional.
+        :param system_prompt: Optional. System prompt string.
         """
         if config is None:
             config = AppConfig()
 
-        super().__init__(config)
+        super().__init__(config, system_prompt)
 
     def get_llm_model_answer(self, prompt, config: ChatConfig):
         messages = []
-        if config.system_prompt:
-            messages.append({"role": "system", "content": config.system_prompt})
+        system_prompt = (
+            self.system_prompt
+            if self.system_prompt is not None
+            else config.system_prompt
+            if config.system_prompt is not None
+            else None
+        )
+        if system_prompt:
+            messages.append({"role": "system", "content": system_prompt})
         messages.append({"role": "user", "content": prompt})
         response = openai.ChatCompletion.create(
             model=config.model or "gpt-3.5-turbo-0613",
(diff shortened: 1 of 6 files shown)


Dev-Khant:
@cachho Can you please review it and help with documentation and also test for `CustomApp`.

embedchain/apps/App.py
```diff
@@ -14,19 +16,27 @@ class App(EmbedChain):
...
     """
 
-    def __init__(self, config: AppConfig = None):
+    def __init__(self, config: AppConfig = None, system_prompt: Optional[str] = None):
         """
         :param config: AppConfig instance to load as configuration. Optional.
+        :param system_prompt: Optional. System prompt string.
         """
         if config is None:
             config = AppConfig()
```
cachho, line 26: couldn't you just say

```python
if system_prompt and not config.system_prompt:
    config.system_prompt = system_prompt
```

right here, and use latch on to the existing implementation? do this for all app classes? should be easier to maintain. You should make a setter method in config though, to make it cleaner.

Dev-Khant: So you are saying instead of using `self.system_prompt` we can set config.system_prompt for app classes. And add a setter method for `system_prompt` in config to set its value if we get prompt while app init. Right??

cachho: I'm saying that instead of introducing something new we just use the App `__init__` to set the existing value in config, and keep working with that. As the most minimal way to achieve this.

My example should work, but introducing a `set_system_prompt` method in `BaseAppConfig` can be easier to understand.

Dev-Khant: Ok but why are we using `BaseAppConfig` to do this when `QueryConfig` is used to pass prompt on chat/query level. It will take more changes and will same as current approach.

Dev-Khant: Ok but why are we using `BaseAppConfig` when `QueryConfig` is used to get system_prompt on chat\query level. Using BaseAppConfig will take more changes and will be similar to current approach.

cachho: I hope you aren't waiting for my input. I haven't had a chance to take a deeper look into this.

Dev-Khant: Sorry I was away for few days but I don't think making changes in `BaseAppConfig` will be a short approach. So directly using  system_prompt from app `_init_` seems to me the shortest approach. Let me know if you find any better or else I'll make few minor changes for it and PR will be ready for review @cachho

Dev-Khant:
Done with changes

taranjeet:
looks good to me.

PR #508, opened 2023-08-29
Feat/serialize deserialize

### LINKED ISSUE #490: Serializing and deserializing a bot instance
### 🚀 The feature

Can you add the possibility to serialise and deserialise bots instances. This may be useful in web development context.

### Motivation, pitch

One of the bottlenecks we face is the current implementation of bot instances, which are not easily serializable. This makes it difficult to store the bot's state, share it across multiple server instances, or even keep a reliable session for the user, thus affecting both performance and user satisfaction.

Benefits:

Scalability: Being able to serialize and deserialize a bot instance will allow us to more efficiently use our server resources. This lays the foundation for load balancing and ensures that as our user base grows, our services can easily scale to meet the demand without performance hitches.
State Preservation: A serializable bot instance means that user interaction can be picked up right where it left off, providing a much more fluid and engaging user experience.
Error Recovery: In the event of a system failure, a serialized instance can be quickly restored, thereby reducing downtime and improving overall system reliability.
Data Insights: Serialized data can be more easily analyzed to gain insights into user behavior and bot performance, providing valuable data that can be used to improve services.
Reduced Latency: Quick serialization and deserialization mean faster response times for the end-user, thereby enhancing the overall user experience.
Technical Outline:

Serialization: The feature should provide a straightforward method to convert a bot instance into a serializable format, e.g., JSON, that captures all the necessary state information.
Deserialization: It should also offer an equally straightforward method to restore a bot instance from its serialized format back to a fully functional object in memory.
Compatibility: The feature should be compatible with our existing systems, requiring minimal changes to integrate.
Security: Serialized data should be encrypted to ensure that sensitive user data is fully protected.

cachho:
## Description

Adds support for serialization and deserialization.

This is a deeply rooted change, it does not break anything, but it changes the way we have to go forward. In the future, we have to decorate all classes.

With `json_string = app.serialize()` you can get a json string. You can store this, send it or whatever. Then, when you're ready you can take the json string and load it to get back the exact same object, with `app = App.deserialize(json_string)`

* it is recursive, so if you serialize app, it also serializes the config. same with deserialization
* tests have been added
* * it's all done automatically, except the decorator, other than that we don't have to take care of it in following updates.

* it does not restore the database in any way
* documentation is not included. we should probably do this. However, thorough inline documentation and docstrings are included.

Thoughts for further improvements
* played around with inheriting the decorator to child classes. This turned out to be a lot more error prone.
* there was talk about encrypting the json. This is not included because I don't understand the purpose or the benefit.
* now that we are adding more methods to the classes, prefixing private methods with `_` becomes a bigger issue.

Fixes #490

## Type of change

## How Has This Been Tested?

- [X] Unit Test
- [X] Test Script (please provide)

```python
from embedchain import OpenSourceApp as App
from embedchain.config import OpenSourceAppConfig as AppConfig

config = AppConfig(log_level="WARNING", id="My very unique id")
app = App(config=config)
s = app.serialize()
new_app = App.deserialize(s)
print(new_app.config.id)
```

this is done better in the unit tests.

## Checklist:

## Maintainer Checklist

- [ ] Made sure Checks passed

### DIFF
--- a/embedchain/apps/App.py
+++ b/embedchain/apps/App.py
@@ -2,8 +2,10 @@ import openai
 
 from embedchain.config import AppConfig, ChatConfig
 from embedchain.embedchain import EmbedChain
+from embedchain.helper_classes.json_serializable import register_deserializable
 
 
+@register_deserializable
 class App(EmbedChain):
     """
     The EmbedChain app.
--- a/embedchain/apps/CustomApp.py
+++ b/embedchain/apps/CustomApp.py
@@ -5,9 +5,11 @@ from langchain.schema import BaseMessage
 
 from embedchain.config import ChatConfig, CustomAppConfig
 from embedchain.embedchain import EmbedChain
+from embedchain.helper_classes.json_serializable import register_deserializable
 from embedchain.models import Providers
 
 
+@register_deserializable
 class CustomApp(EmbedChain):
     """
     The custom EmbedChain app.
--- a/embedchain/apps/OpenSourceApp.py
+++ b/embedchain/apps/OpenSourceApp.py
@@ -7,8 +7,6 @@ from embedchain.helper_classes.json_serializable import register_deserializable
 
 gpt4all_model = None
 
-from embedchain.helper_classes.json_serializable import register_deserializable
-
 
 @register_deserializable
 class OpenSourceApp(EmbedChain):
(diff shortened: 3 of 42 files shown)


fsndzomga:
Nicely done

taranjeet:
this is a good PR @cachho . thanks for your contribution.

taranjeet (approved):

PR #571, opened 2023-09-06
Add support for image dataset

### LINKED ISSUE #511: feature request: add support for Image(s) as a new DataType
### 🚀 The feature

Usecase
A user should be able to **add** a directory/set of images to embedchain. Then for a given query, they should be able to find the most relevant image from this dataset

Tech nuances
1. We can use CLIP : https://github.com/openai/CLIP to generate the embeddings and store it in the supplied vector db
2. This will require us to create a new loader - ImagesLoader
3. Will require a new Chunker: ImagesChunker
4. When the user issues the query to search for the relevant result, we create the embeddings from the same CLIP model and search in the db. This means there is no usecase of having prompts and making it pass through the LLMs. As such, I propose a new [Interface](https://docs.embedchain.ai/advanced/interface_types) Type - search to be able to search for the most relevant results for a given query

Concerns/Questions
1. @taranjeet in one of our offline conversation you suggested the usage of CLIP and ocr-space. But given CLIP does a decent OCR as well, that can be skipped? I performed a bunch of experiments on CLIP and it seems to be doing a decent job on text images.
2. The model from CLIP *ViT-B/32* is 320 MB. The model will be downloaded when the user attempts to add an image datatype
3. Wondering how the chunk size/overlap decided for other datatypes?

### Motivation, pitch

Will help do context based searching on images, which can unlock several interesting usecases for the embedchain community.

rupeshbansal:
## Description

This adds support for image as a new dataset in embedchain. Users will be able to feed in an image/directory(containing a set of images) to embedchain, and then issue query searches on the context of images.

NOTE: Docs are yet to be updated. Unsure about the convention of the repository, but planning to add those in a followup!

## Setup
Run `poetry install -E images` to install all the additional dependencies

## Preparation of the data
Create a new folder and put all the images on which the embedchain is to be trained in it

## Add the Images to Embedchain

```
from embedchain import App
from embedchain.config.llm.base_llm_config import BaseLlmConfig

app = App()
app.add(<PATH_TO_THE_LOCAL_IMAGES>, data_type="images")
```

## Query on the context

```
app.query(<QUERY>, config = BaseLlmConfig(query_type = "Images"))
```

## Test
You can test this feature by following instructions in https://github.com/rupeshbansal/embedchain_imagetest/

Fixes #511

## Type of change

## How Has This Been Tested?

- [x] Unit Test

## Checklist:

## Maintainer Checklist

- [ ] closes #511 
- [ ] Made sure Checks passed


rupeshbansal:
@taranjeet @cachho Can you please take a look?

cachho:
I'm sure this is a great feature! I couldn't test it yet or do a thorough review, I hope you can just answer a few questions here.

What's happening behind the scenes? Is your image embedded as an image, or is it transformed to text  and that text is embedded? After adding it, can you only query it as an image (is that what `query_type` is for) or also in other contexts? Will it return the image or a text answer? Does it work with all LLMs?

We're definitely going to need documentation for the `query_type`, especially since it's user facing. What does it do? Can it be used with other data_types as well?

Can you provide a full test script (with creative commons images)? What kind of question can you ask?

rupeshbansal:
Thanks @cachho for taking a look. Please find the responses inline

> What's happening behind the scenes? Is your image embedded as an image, or is it transformed to text and that text is embedded? 

We create an embedding for each image, which is stored in the DB(CharomDB or ES of whatever is configured). By embedding, I mean an array of floating numbers. This conversion is done using https://github.com/openai/CLIP. You should be able to find how exactly this is being done in `embedchain/models/ClipProcessor.py`

> After adding it, can you only query it as an image (is that what query_type is for) or also in other contexts? 

After adding the images, you query it using text context of the added images. Example if you add an image of a mountain, a beach, a forest, a boy to embedchain, we should then be able to search using queries like "A lush green set of trees", which should return the image of a forest. 

> Will it return the image or a text answer? 
It will return text answer, representing an array of image paths which are most relevant to the query

Does it work with all LLMs?
The is no LLM involved with this. We are finding the nearest neighbour of the query in the DB. We are relying on the the default algorithm(Generally Cosine Similarity) to get the nearest neighbour

> We're definitely going to need documentation for the query_type, especially since it's user facing. What does it do? 
I completely agree. My plans were to add it once this gets an LGTM. Will help reduce the iterations. Let me know if you feel otherwise

> Can it be used with other data_types as well?
No, it cannot be

> Can you provide a full test script (with creative commons images)? What kind of question can you ask?

Nice suggestion. I have created https://github.com/rupeshbansal/embedchain_imagetest/ which has the steps to test it. Will add it to the PR description as well

embedchain/embedchain.py
```diff
@@ -23,6 +23,8 @@
 from embedchain.loaders.base_loader import BaseLoader
 from embedchain.models.data_type import (DataType, DirectDataType,
                                          IndirectDataType, SpecialDataType)
+from embedchain.models.data_type import DataType
+from embedchain.models.ClipProcessor import ClipProcessor
```
deshraj, line 27:
```suggestion
from embedchain.models.clip_processor import ClipProcessor
```

deshraj: We want to be consistent with file naming. Should follow snake case everywhere.

pyproject.toml
```diff
@@ -103,7 +103,7 @@ twilio = { version = "^8.5.0", optional = true }
 fastapi-poe = { version = "0.0.16", optional = true }
 discord = { version = "^2.3.2", optional = true }
 slack-sdk = { version = "3.21.3", optional = true }
-
+clip = {git = "https://github.com/openai/CLIP.git", optional = true}
```
deshraj, line 106: Is it possible to pin to a specific commit on main branch? It is too risky/brittle to be dependent on the head of main branch in general.

pyproject.toml
```diff
@@ -125,6 +125,7 @@ poe = ["fastapi-poe"]
 discord = ["discord"]
 slack = ["slack-sdk", "flask"]
 whatsapp = ["twilio", "flask"]
+images = ["torch", "ftfy", "regex", "tqdm", "clip"]
```
deshraj, line 128: Do you really need tqdm here? I dont see it getting used.

deshraj (commented):
Generally looks good to me. Great work @rupeshbansal ❤️ 

Please resolve the comments and incorporate minor suggestions and we are good to go here.

deshraj (approved):

LuciAkirami:
Hey @rupeshbansal , glad to know that Image search has come to embedchain. It would be great if there is a default `query_type` set to `text` (non Image). I mean, after these commits, the user has to explicitly give the `query_text` to the `app.query()`, else the code throws out `AttributeError: 'NoneType' object has no attribute query_type`

-> Before this commit, the typical flow of creating a bot would be like this
```
from embedchain import App
from embedchain.llm.openai import OpenAILlm


app = App()

app.add("https://www.youtube.com/watch?v=ZnEgvGPMRXA")

response = app.query("Is Co-Pilot in Windows update?")
print(response)
```

-> After this commit, we need to explicitly provide the query type like the below
```
from embedchain import App
from embedchain.config import LlmConfig

app = App()

app.add("https://www.youtube.com/watch?v=ZnEgvGPMRXA")

response = app.query("Is Co-Pilot in Windows update?",config=LlmConfig(query_type='text'))
print(response)
```

-> if the old style is followed, then throws it out an error
`AttributeError: 'NoneType' object has no attribute 'query_type'`

-> One work around would be, to check if the `config` is provided to the `app.query()` and if not, then assume the `query_type` is `text` type(I mean like setting `text` as default). For this the following changes have to be made in the embedchain.py inside the 'query()' function

```
if config == None:
            config=BaseLlmConfig(query_type='text')
```

-> And as @cachho as mentioned, this needs to be in documentation, I myself was confused (when I did a git pull) and all of a sudden like why wasn't the code working and was backtracking and finally found this.

PR #780, opened 2023-10-08
Improve tests

sidmohanty11:
## Description

This PR aims to test most of the modules and improve test coverage as well as code stability

- [x] Chunkers
- [x] Loaders


tests/chunkers/test_base_chunker.py
```diff
@@ -0,0 +1,68 @@
+import hashlib
+import unittest
+from unittest.mock import MagicMock
+
+from embedchain.chunkers.base_chunker import BaseChunker
+from embedchain.models.data_type import DataType
+
+
+class TestBaseChunker(unittest.TestCase):
```
deshraj, line 9: Let's not use unittest here since it is quite limited. We should use `pytest` in general since it has way more extensions and is better than in-built unittest.

deshraj (approved):
This is great. Thanks for adding tests ❤️

PR #790, opened 2023-10-10
Create gpt4free.py

Yusufibin:
Add gpt4free


deshraj:
@Yusufibin thanks for the PR. We don't plan to add gf4 support in the near future. We will let you know if we plan to do so in the near future.

PR #795, opened 2023-10-12
Improve tests

sidmohanty11:
## Description

This PR aims to add and cover tests for the following modules:
- [ ] LLMs
- [ ] Bots

### DIFF
--- a/tests/llm/test_vertex_ai.py
+++ b/tests/llm/test_vertex_ai.py
@@ -1,0 +1,63 @@
+import pytest
+from unittest.mock import MagicMock, patch
+from embedchain.llm.vertex_ai import VertexAiLlm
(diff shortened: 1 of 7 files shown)


tests/llm/test_vertex_ai.py
```diff
@@ -0,0 +1,63 @@
+import pytest
+from unittest.mock import MagicMock, patch
```
deshraj, line 2: It's fine to use patch from unittest package for now but lets change everything to use `pytest-mock` going forward.

deshraj (approved):

tests/llm/test_vertex_ai.py
sidmohanty11: Got it!

PR #807, opened 2023-10-15
Improve and add more tests

sidmohanty11:
## Description

This PR fixes Open AI LLM for streaming and adds more tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,6 +120,7 @@ ftfy = { version = "6.1.1", optional = true }
 regex = { version = "2023.8.8", optional = true }
 huggingface_hub = { version = "^0.17.3", optional = true }
 pymilvus = "2.3.1"
+pytest-asyncio = "^0.21.1"
 
 [tool.poetry.group.dev.dependencies]
 black = "^23.3.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,7 +120,6 @@ ftfy = { version = "6.1.1", optional = true }
 regex = { version = "2023.8.8", optional = true }
 huggingface_hub = { version = "^0.17.3", optional = true }
 pymilvus = "2.3.1"
-pytest-asyncio = "^0.21.1"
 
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -120,6 +120,7 @@ ftfy = { version = "6.1.1", optional = true }
 regex = { version = "2023.8.8", optional = true }
 huggingface_hub = { version = "^0.17.3", optional = true }
 pymilvus = "2.3.1"
+pytest-asyncio = "^0.21.1"
```
deshraj, line 123: It should be part of the dev dependencies.

tests/llm/test_huggingface.py
```diff
@@ -1,64 +1,63 @@
...
+    mock_llm_instance = mocker.Mock(return_value="Test answer")
+    mocker.patch("embedchain.llm.huggingface.HuggingFaceHub", return_value=mock_llm_instance)
+
+    llm = HuggingFaceLlm(huggingface_llm_config)
+    answer = llm.get_llm_model_answer("Test query")
+
+    assert answer == "Test answer"
+    mock_llm_instance.assert_called_once_with("Test query")
+    mock_llm_instance.assert_called_once_with("Test query")
+    mock_llm_instance.assert_called_once_with("Test query")
```
deshraj, line 63: Do we need to call this three times? I am a bit confused here 😅

pyproject.toml
deshraj: Also, I am not sure if we really need this package btw.

sidmohanty11: @deshraj thank you, I've added it to dev dependency. It's actually needed by a test for the Poe bot, which uses async code that's not currently running and is being skipped.

tests/llm/test_huggingface.py
sidmohanty11: Sorry for the miss, yup you're right! I've removed it :)

deshraj (approved):

PR #837, opened 2023-10-21
[Bug fix] import App shouldn't throw other llm deps errors

sidmohanty11:
## Description

Refer: https://colab.research.google.com/drive/157gl1vwM2YWT80arZehQ-RmfvhRO71Sv?usp=sharing


deshraj (approved):
Nice, good catch.

PR #844, opened 2023-10-24
[Bug fix] reset() erases everything from db

### LINKED ISSUE #842: Reset current collection only when `app.db.reset()` is called
### 🐛 Describe the bug

Currently, whenever someone calls the `app.db.reset()`, we delete all the collections from the vector database which is not ideal. We want to instead only delete the collection that the current app is created with. 

We want to make sure that support for this is added in all the vector databases. 

Issue reported in our slack: https://embedchain.slack.com/archives/C05RG7CJKC6/p1697890507048149

sidmohanty11:
## Description

In the case of ChromaDB, the existing `reset()` function was designed to clear all data from the database. However, based on our specific requirements, we only need to clear the app's collections context, rather than wiping out the entire database. This PR addresses this issue by modifying the reset() function to clear only the collection data.

This PR also updates the tests accordingly and adds clean up `reset()` calls.

Fixes #842

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -165,6 +165,7 @@ cython_debug/
 
 # Database
 db
+test-db
 
 .vscode
 /poetry.lock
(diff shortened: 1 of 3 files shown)


deshraj (approved):

deshraj:
Very nice. Thanks for fixing the issue!

PR #872, opened 2023-10-30
Lazy load loaders and chunkers

sidmohanty11:
## Description

This PR adds lazy loading support for all our loaders and chunkers so that we only import what we need.


deshraj (approved):
Very nice. This will make the package even cleaner.

PR #889, opened 2023-11-01
Add `rest-api` example

sidmohanty11:
## Description

This PR introduces the highly anticipated `rest-api` example, which upon deployment, will empower us to run Embedchain REST API with a single command. The users can add multiple apps, deploy, add sources, check sources associated with apps, query, chat and much more!

i.e,

```bash
docker run -d --name embedchain -p 8000:8000 embedchain/app:rest-api-latest
```

Screenshot 2023-11-01 at 7 47 17 PM


embedchain/pipeline.py
```diff
@@ -312,6 +312,15 @@ def _mark_data_as_uploaded(self, data_hash):
         )
         self.connection.commit()
 
+    def get_data_sources_by_app_id(self, app_id):
```
deshraj, line 315: You don't need to pass the app_id here, right? You can just access it by doing `self.local_id`

examples/rest-api/.env.example
```diff
@@ -0,0 +1,22 @@
...
+# ANTHROPIC_API_KEY=
+
+# for Jina
+# JINACHAT_API_KEY=
+
+# for huggingface
+# HUGGINGFACE_ACCESS_TOKEN=
+
+# for llama2
+# REPLICATE_API_TOKEN=
```
deshraj, line 22: This is not going to work since users will do 

```
docker pull <image>
docker run <image> .....
```

So they won't be able to edit the env file.

I am wondering if we should do something like this:

```
docker run -e ENV_VARIABLE1=value1 -e ENV_VARIABLE2=value2 -p 8080:80 your-docker-image:tag
```

Then it will be user's responsibility to pass the relevant docker image whenever they are trying to spawn the docker container.

examples/rest-api/.gitignore
```diff
@@ -0,0 +1,4 @@
+.env
+app.db
+configs/**.yaml
+db
```
deshraj, line 4:
```suggestion
db

```

examples/rest-api/database.py
```diff
@@ -0,0 +1,11 @@
+from sqlalchemy import create_engine
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+
+SQLALCHEMY_DATABASE_URL = "sqlite:///./app.db"
```
deshraj, line 5:
```suggestion
SQLALCHEMY_DATABASE_URI = "sqlite:///./app.db"
```

examples/rest-api/database.py
```diff
@@ -0,0 +1,11 @@
...
+
+engine = create_engine(SQLALCHEMY_DATABASE_URL, connect_args={"check_same_thread": False})
```
deshraj, line 7:
```suggestion
engine = create_engine(SQLALCHEMY_DATABASE_URI, connect_args={"check_same_thread": False})
```

deshraj: @sidmohanty11 do you think we should add `connect_args={"check_same_thread": False})` in the Pipeline class as well?

examples/rest-api/main.py
```diff
@@ -0,0 +1,354 @@
...
+
+@app.get("/ping", tags=["Utility"])
+def check_status():
+    """
+    Endpoint to check the status of the API.
+    """
+    return {"ping": "pong"}
+
+
+# FIXME: Put on hold for now
```
deshraj, line 48: No need for this. We can safely delete for now.

examples/rest-api/main.py
```diff
@@ -0,0 +1,354 @@
...
+#         raise HTTPException(detail=f"Error setting API key: {e}", status_code=400)
+
+
+@app.get("/apps", tags=["Apps"])
+async def get_all_apps(db: Session = Depends(get_db)):
+    """
+    Get all apps.
+    """
+    apps = get_apps(db)
+    return {"data": apps}
```
deshraj, line 70:
```suggestion
    return {"results": apps}
```

examples/rest-api/main.py
```diff
@@ -0,0 +1,354 @@
...
+from sqlalchemy.orm import Session
+from embedchain import Pipeline as App
+from embedchain.client import Client
+from models import (
+    QueryApp,
+    SourceApp,
+    MessageApp,
+    DefaultResponse,
+    DeployAppRequest,
+    # SetEnvKeys,
```
deshraj, line 13:
```suggestion
```

examples/rest-api/main.py
```diff
@@ -0,0 +1,354 @@
...
+
+        save_app(db, app_id, yaml_path)
+
+        return DefaultResponse(response=f"App created successfully. App ID: {app_id}")
+    except Exception as e:
+        raise HTTPException(detail=f"Error creating app: {e}", status_code=400)
+
+
+@app.post("/create/yaml", tags=["Apps"], response_model=DefaultResponse)
+async def create_app_using_custom_config(config: UploadFile, db: Session = Depends(get_db)):
```
deshraj, line 97: Both API /create and /create/yaml should be one API only.  If user doesn't pass any config yaml then we use default one.

examples/rest-api/main.py
```diff
@@ -0,0 +1,354 @@
...
+        except yaml.YAMLError as exc:
+            raise HTTPException(detail=f"Error parsing YAML: {exc}", status_code=400)
+
+        return DefaultResponse(response=f"App created successfully. App ID: {app_id}")
+    except Exception as e:
+        raise HTTPException(detail=f"Error creating app: {e}", status_code=400)
+
+
+@app.get(
+    "/{app_id}/datasources",
```
deshraj, line 129:
```suggestion
    "/{app_id}/data",
```

examples/rest-api/main.py
```diff
@@ -0,0 +1,354 @@
...
+            raise HTTPException(detail=f"App with id {app_id} does not exist, please create it first.", status_code=400)
+
+        app = App.from_config(yaml_path=db_app.config)
+
+        response = app.get_data_sources_by_app_id(app_id=app_id)
+        return {"data": response}
+    except ValueError as ve:
+        if "OPENAI_API_KEY" in str(ve) or "OPENAI_ORGANIZATION" in str(ve):
+            raise HTTPException(
+                detail="OPENAI_API_KEY or OPENAI_ORGANIZATION not set, please set them in your environment variables.",
```
deshraj, line 156: Can you please add instructions on how one can do that. Show them command like:

```
docker run -e ENV=value <image>
```

Add the correct image name.

examples/rest-api/main.py
```diff
@@ -0,0 +1,354 @@
...
+            )
+
+        db_app = get_app(db, app_id)
+
+        if db_app is None:
+            raise HTTPException(detail=f"App with id {app_id} does not exist, please create it first.", status_code=400)
+
+        app = App.from_config(yaml_path=db_app.config)
+
+        response = app.chat(body.message)
```
deshraj, line 263: How are you preserving the chat history here? I think you will have to preserve it separately since Embedchain stores history in memory which won't be accessible by subsequent API calls.

examples/rest-api/sample-config.yaml
```diff
@@ -0,0 +1,35 @@
+app:
+  config:
+    id: 'testapp'
```
deshraj, line 3:
```suggestion
    id: 'default-app'
```

examples/rest-api/sample-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+      Use the following pieces of context to answer the query at the end.
+      If you don't know the answer, just say that you don't know, don't try to make up an answer.
+
+      $context
+
+      Query: $query
+
+      Helpful Answer:
+    system_prompt: |
+      Act as William Shakespeare. Answer the following questions in the style of William Shakespeare.
```
deshraj, line 23: Don't add this system prompt as default :D

poetry.lock
deshraj, line 1: Remove this file please.

deshraj (commented):
Overall looks great. Let's make the desired changes and we are good to go.

embedchain/pipeline.py
sidmohanty11: Absolutely, I missed it. Thanks!

examples/rest-api/.env.example
sidmohanty11: Yup yup, I was just listing down the envs and testing different approaches (`docker-compose`) but for our use case `-e ENV_VAR=value` seems to be the best

examples/rest-api/database.py
sidmohanty11: Yes, I think so! I've seen some users get this error and once it happened to me during multiple App calls. What do you think?

Ref: https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#threading-pooling-behavior

Screenshot 2023-11-02 at 6 37 54 PM

examples/rest-api/main.py
sidmohanty11: Got it!

examples/rest-api/main.py
sidmohanty11: Ahh true! Do you think we should keep this endpoint then? I guess it is particularly useful in the case of applications where Embedchain's `App` can stay persistent and a separate bot can be created for a specific use case.

sidmohanty11: Please let me know your thoughts on this!

examples/rest-api/main.py
sidmohanty11: Thanks! I've added a proper mapping to check if there is any ValueError for a missing API key, now it will throw an error accordingly with an example docker command using that missing key.

For example if `OPENAI_API_KEY` is missing,

Error message:
```
Please set the OPENAI_API_KEY environment variable(s) when running the Docker container.\nExample: `docker run -e OPENAI_API_KEY=xxx embedchain/app:rest-api-latest
```

examples/rest-api/main.py
sidmohanty11: Discussed it in a meet, keep it on hold

deshraj (approved):

PR #890, opened 2023-11-01
Add yaml config validation

sidmohanty11:
## Description

This PR introduces YAML config validation by providing more informative error messages in case of indentation or key misplacement and also includes validation for provider namings.


pyproject.toml
deshraj, line 1: Can you please update the package version as well?

deshraj (approved):

PR #898, opened 2023-11-02
[UX/DX Improvement] Auto download `json` extra to use json loader effortlessly

deven298:
## Description

As the title suggests

### DIFF
--- a/embedchain/loaders/json.py
+++ b/embedchain/loaders/json.py
@@ -1,7 +1,9 @@
 import hashlib
+import importlib
 import json
 import os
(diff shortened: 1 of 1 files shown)


embedchain/embedchain.py
```diff
@@ -409,7 +424,7 @@ def load_and_embed(
 
             new_metadatas.append(m)
         metadatas = new_metadatas
-
+        print("DOC metadata: ", metadatas)
```
deshraj, line 427:
```suggestion
```

Issue #856, opened 2023-10-26
SQLite objects created in thread can only be used in that same thread


deadbits:
### 🐛 Describe the bug

When attempting to use the `App().add` functionality within a Flask app, embedchain is throwing an error related to SQLite threads. I'm not doing anything differently from the example API server application, and when I recreate the functionality within an ipython interpreter, it works fine.

**Exception:**
```
SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140055526503488 and this is thread id 140054403995200."
```

**Sample Flask app:**
The code below is a stripped down version of my own, and more or less a copy of the API server example.
```
from flask import Flask, jsonify, request
from embedchain import App

app = Flask(__name__)
ec_app = App()

@app.route('/add', methods=['POST'])
def add():

    data = parse_field(request.json, 'data', str)

    if data:

        try:
            ec_app.add(data)
        except Exception as err:
            return jsonify(
                {
                    'error': 'Failed to add data to embedchain',
                    'exception': str(err)
                }
            ), 500
```

deadbits:
Example API request that throws the error
```
curl -X POST -H "Content-Type: application/json" -d '{"data": "https://arxiv.org/pdf/2310.03684.pdf"}' http://127.0.0.1:5000/add | jq '.'
{
  "error": "Failed to add data to embedchain",
  "exception": "SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140055526503488 and this is thread id 140054403995200."
}
```

deshraj:
Thanks @deadbits for reporting this issue. Do you happen to have the traceback from the flask logs?

deadbits:
Sorry, should've provided that initially!

```
2023-10-26 16:21:30.864 | DEBUG    | __main__:add:73 - Adding item to embedchain: https://arxiv.org/pdf/2310.03684.pdf
Traceback (most recent call last):
  File "/home/adam/Research/machine-learning/minerva-kb/server.py", line 74, in add
    ec_app.add(data)
  File "/home/adam/Research/machine-learning/minerva-kb/venv/lib/python3.10/site-packages/embedchain/embedchain.py", line 223, in add
    self.cursor.execute(
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140572656227392 and this is thread id 140571636135488.
```

fabioferreira3:
I'm starting to experience this same issue

deshraj:
[...]

PR #911, opened 2023-11-06
Set `check_same_thread` false so that one App can be used in parallel

sidmohanty11:
## Description

This PR adds `check_same_thread=False` which will help us use Embedchain directly as a bot or an instance in REST APIs.

Refer: #856


deshraj:
can you please add a test for this @sidmohanty11 ?

deshraj (approved):

PR #933, opened 2023-11-10
Add slack_bot docker image

sidmohanty11:
## Description

Deploy Slack bot docker image

### DIFF
--- a/examples/slack_bot/requirements.txt
+++ b/examples/slack_bot/requirements.txt
@@ -1,3 +1,1 @@
-embedchain
-embedchain[slack]
-embedchain[poe]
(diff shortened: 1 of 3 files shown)


examples/slack_bot/requirements.txt
```diff
@@ -0,0 +1,3 @@
+embedchain
+embedchain[slack]
+embedchain[poe]
```
deshraj, line 3: Can we please pin the version of embedchain to use? You can also convert all of these in one line like this

```
embedchain[slack, poe]==0.1.7
```

sidmohanty11: Sure, let me do that

PR #934, opened 2023-11-10
[Bug fix] missing dir on first init for App

sidmohanty11:
## Description

This PR fixes missing dir error from `App` as well, supporting backward compatibility.


deshraj (approved):
Good catch. Thanks!

PR #949, opened 2023-11-14
[New] Substack loader

sidmohanty11:


embedchain/loaders/substack.py
```diff
@@ -0,0 +1,90 @@
+import concurrent.futures
+import hashlib
+import logging
+
+import requests
+
+try:
```
deven298, line 7: Move this inside the class.

embedchain/loaders/substack.py
```diff
@@ -0,0 +1,90 @@
...
+
+        doc_id = hashlib.sha256((" ".join(links) + url).encode()).hexdigest()
+
+        def load_link(link):
+            try:
+                each_load_data = requests.get(link)
+                each_load_data.raise_for_status()
+
+                data = self.serialize_response(each_load_data)
+                data = str(data).encode()
```
deven298, line 46: Why `.encode()`?

embedchain/loaders/substack.py
sidmohanty11: Thanks!

embedchain/loaders/substack.py
```diff
@@ -0,0 +1,90 @@
...
+import logging
+
+import requests
+
+try:
+    from bs4 import BeautifulSoup
+    from bs4.builder import ParserRejectedMarkup
+except ImportError:
+    raise ImportError(
+        'Sitemap requires extra dependencies. Install with `pip install --upgrade "embedchain[dataloaders]"`'
```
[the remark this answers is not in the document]
sidmohanty11: Thanks! Missed it

embedchain/loaders/substack.py
deshraj: This still seems to be unresolved?

docs/data-sources/substack.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+To add any Substack data sources to your app, just add the sitemap.xml of that url as the source and set the data_type to `substack`.
+
+```python
+from embedchain import Pipeline as App
+
+app = App()
+
+# source: for any substack just add the sitemap.xml url
+app.add('https://www.lennysnewsletter.com/sitemap.xml', data_type='substack')
```
deshraj, line 13: Do the user have to enter the full url with `sitemap.xml` ? I think we should just assume that sitemap.xml is present and then try to fetch it. If we fail, then we should return error to the user

deshraj: I am going to merge it for now but lets do this change in a follow up PR.

deshraj (approved):

PR #960, opened 2023-11-16
[Bugfix] fix poetry lock

deven298:
As title suggests


embedchain/loaders/github.py
```diff
@@ -68,7 +68,7 @@ def _add_repo_files(repo_path: str):
                             data.extend(results)
                             data_urls.extend([result.get("meta_data").get("url") for result in results])
                     except Exception as e:
-                        logging.error(f"Failed to process {file}: {e}")
+                        logging.info(f"Failed to process {file}: {e}")
```
deshraj, line 71:
```suggestion
                        logging.warn(f"Failed to process {file}: {e}")
```

deshraj (approved):

PR #963, opened 2023-11-17
[New] Beehiiv loader

sidmohanty11:


embedchain/loaders/beehiiv.py
```diff
@@ -0,0 +1,104 @@
...
+            url = url + "/sitemap.xml"
+
+        output = []
+        # we need to set this as a header to avoid 403
+        headers = {
+            "User-Agent": (
+                "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) "
+                "AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.102 "
+                "Safari/537.36"
+            ),
```
deshraj, line 37: Is this necessary to add?

embedchain/loaders/beehiiv.py
```diff
@@ -0,0 +1,104 @@
...
+            except ParserRejectedMarkup as e:
+                logging.error(f"Failed to parse {link}: {e}")
+            return None
+
+        for link in links:
+            data = load_link(link)
+            if data:
+                output.append({"content": data, "meta_data": {"url": link}})
+            # TODO: allow users to configure this
+            time.sleep(1.0)  # added to avoid rate limiting
```
deshraj, line 102: Do we need 1 sec for beehiiv? I am wondering if we can get away with 0.1 secs here.

embedchain/loaders/beehiiv.py
sidmohanty11: Yup, beehiiv actually doesn't allow anonymous fetches. They throw a forbidden error with an HTML that says cookies need to be set

embedchain/loaders/beehiiv.py
sidmohanty11: Let me try this

embedchain/loaders/beehiiv.py
```diff
@@ -0,0 +1,104 @@
...
+
+        def load_link(link: str):
+            try:
+                each_load_data = requests.get(link, headers=headers)
+                each_load_data.raise_for_status()
+
+                soup = BeautifulSoup(each_load_data.text, "html.parser")
+                data = serialize_response(soup)
+                data = str(data)
+                if is_readable(data):
```
deven298, line 89: @sidmohanty11 Do you know a better way to do this then using `is_readable`?

Using this will be a problem when loading homepages with images and list of podcasts.

embedchain/loaders/substack.py
```diff
@@ -65,7 +63,7 @@ def load_link(link: str):
                 each_load_data = requests.get(link)
                 each_load_data.raise_for_status()
 
-                soup = BeautifulSoup(response.text, "html.parser")
+                soup = BeautifulSoup(each_load_data.text, "html.parser")
```
deven298, line 66: @sidmohanty11 change `each_load_data` to `beehiiv_data`

deven298 (changes requested):
Looks good!

deshraj (approved):

PR #968, opened 2023-11-20
[Bug fix] Create local sqlite directory at init

deven298:
as title suggests

### DIFF
--- a/embedchain/embedchain.py
+++ b/embedchain/embedchain.py
@@ -21,6 +21,7 @@ from embedchain.models.data_type import (DataType, DirectDataType,
 from embedchain.telemetry.posthog import AnonymousTelemetry
 from embedchain.utils import detect_datatype, is_valid_json_string
 from embedchain.vectordb.base import BaseVectorDB
+from embedchain.client import Client
 
 load_dotenv()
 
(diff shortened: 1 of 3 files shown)


embedchain/memory/base.py
```diff
@@ -18,7 +18,7 @@
                 created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
                 PRIMARY KEY (id, app_id)
             )
-            """
+"""
```
deshraj, line 21: This is probably not needed.

embedchain/embedchain.py
```diff
@@ -50,6 +51,8 @@ def __init__(
         :type system_prompt: Optional[str], optional
         :raises ValueError: No database or embedder provided.
         """
+        # Setup user directory if it doesn't exist already
+        Client.setup_dir()
```
deshraj, line 55: I think this needs to be outside the Embedchain class to be able to run when we do `import embedchain`

embedchain/memory/base.py
deven298: lets just keep it. Its more cleaner

deshraj:
Closing this since already fixed in #975

PR #1001, opened 2023-12-09
Audio Loader using Whisper

sidmohanty11:
Test script

```python
import os
from embedchain import Pipeline as App

os.environ["OPENAI_API_KEY"] = "sk-xxx"

app = App()
app.add("./sample-0.mp3", data_type="audio")
result = app.query("What is the author saying?")
print(result)
```


embedchain/loaders/audio.py
```diff
@@ -0,0 +1,33 @@
...
+    def load_data(self, url):
+        """Load data from an Audio file."""
+        try:
+            import whisper
+        except ImportError:
+            raise ImportError(
+                'Audio requires extra dependencies. Install with `pip install --upgrade "embedchain[audio]"`'
+            ) from None
+        
+        model = whisper.load_model("base.en")
```
peterj, line 18: Consider defaulting to "base.en", but making the mode selection configurable (i.e. it can be passed in to the `load_data`)

taranjeet:
@deven298 : can you take this over and implement it + address PR feedback?

PR #1038, opened 2023-12-20
[chore] pypdf and bs4 by default in package

sidmohanty11:
## Description

- Keep `pypdf` and `bs4` in package by default as they are the most used ones (remove friction)

cc @deshraj


deven298 (approved):

deven298 (approved):
LGTM

deshraj (approved):

PR #1052, opened 2023-12-23
Provide openai-key support from config

sidmohanty11:
```python
pipeline = Pipeline.from_config(config={
    "llm": {
        "provider": "openai",
        "config": {
            "api_key": "sk-xxx"
        }
    },
    "embedder": {
        "provider": "openai",
        "config": {
            "api_key": "sk-xxx"
        }
    }
})
```


embedchain/embedder/openai.py
```diff
@@ -16,16 +16,18 @@ def __init__(self, config: Optional[BaseEmbedderConfig] = None):
...
             self.config.model = "text-embedding-ada-002"
 
+        api_key = self.config.api_key or os.environ["OPENAI_API_KEY"]
+
         if self.config.deployment_name:
             embeddings = AzureOpenAIEmbeddings(deployment=self.config.deployment_name)
             embedding_fn = BaseEmbedder._langchain_default_concept(embeddings)
         else:
-            if os.getenv("OPENAI_API_KEY") is None and os.getenv("OPENAI_ORGANIZATION") is None:
+            if api_key is None and os.getenv("OPENAI_ORGANIZATION") is None:
```
deven298, line 25: This should check for `api_key`?

deven298 (approved):

embedchain/embedder/openai.py
sidmohanty11: Yes which is either `config.api_key or env.api_key`

PR #1107, opened 2024-01-03
[Bugfix] fix cache session id in chat method

deven298:
## Description

This PR enables caching per chat session.


deshraj (approved):

deshraj:
Good catch.

PR #1112, opened 2024-01-04
Add support for `ec runserver` command

deshraj:
## Description

Adding support for `ec runserver` command so that developers can create full stack RAG applications with a single line of command.

## Type of change

- [x] Documentation update


embedchain/cli.py
```diff
@@ -414,3 +437,99 @@ def deploy():
...
+        api_process = subprocess.Popen(
+            ["uvicorn", "main:app", "--reload", "--host", "127.0.0.1", "--port", "8000"], stdout=None, stderr=None
+        )
+        os.chdir("..")
+        console.print("✅ [bold green]API server started successfully.[/bold green]")
+    except Exception as e:
+        console.print(f"❌ [bold red]Failed to start the API server: {e}[/bold red]")
+        return
+
+    # Step 5: Install UI requirements and start the UI server
```
deven298, line 508: @deshraj Step 5 and 6 are the same, why?

deshraj: Ah good catch. Will fix.

PR #1233, opened 2024-01-31
Update URL Validation Regex to Support IP Addresses and Port Numbers

### LINKED ISSUE #1232: Extend URL Validation to Include IP Addresses and Port Numbers in JSON loader
### 🐛 Describe the bug

I propose enhancing the [URL validation regex in the JSON loader ](https://github.com/embedchain/embedchain/blob/34c41c87dc71bcc007b45b273da3a2b4f6c2e86f/embedchain/loaders/json.py#L39) to support URLs that include IP addresses, optional port numbers, and URL-encoded characters. Currently, the validation fails for valid URLs such as http://192.168.1.140:9000/1%202%20Inflación-wav.json because it does not accommodate IP-based addresses, ports, or encoded path characters. This limitation prevents correct handling of certain resource URLs that are otherwise valid. This limitation is also hindering the testing process (we can't load JSON files using a local http server).

juananpe:
## Description

This pull request updates the URL validation regex to accommodate URLs with http or https protocols, IP addresses, optional port numbers, and URL-encoded characters, ensuring broader validation coverage. This change allows the system to correctly validate URLs like http://192.168.1.140:9000/1%202%20Inflación-wav.json, which include direct IP access with port specifications and encoded characters, enhancing compatibility with a wider range of URLs.

Fixes #1232 

## Type of change

### DIFF
--- a/embedchain/loaders/json.py
+++ b/embedchain/loaders/json.py
@@ -36,8 +36,7 @@ class JSONReader:
         return ["\n".join(useful_lines)]
 
 
-VALID_URL_PATTERN = "^https:\/\/[0-9A-Za-z]+(\.[0-9A-Za-z]+)*\/[0-9A-Za-z_\/]*\.json$"
-
+VALID_URL_PATTERN = "^https?://(?:www\.)?(?:\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|[a-zA-Z0-9.-]+)(?::\d+)?/(?:[^/\s]+/)*[^/\s]+\.json$"
 
 class JSONLoader(BaseLoader):
     @staticmethod


embedchain/loaders/json.py
```diff
@@ -36,7 +36,7 @@ def load_data(json_data: Union[dict, str]) -> list[str]:
         return ["\n".join(useful_lines)]
 
 
-VALID_URL_PATTERN = "^https:\/\/[0-9A-Za-z]+(\.[0-9A-Za-z]+)*\/[0-9A-Za-z_\/]*\.json$"
+VALID_URL_PATTERN = "^(http|https):\/\/[0-9]{1,3}(\.[0-9]{1,3}){3}(:[0-9]+)?\/[0-9A-Za-z%_\-]+\.json$"
```
taranjeet, line 40: the addition is great @juananpe 
But this url will remove support for the existing domain.
Can you please update your code so that both the regex are supported?

juananpe: Done. Sorry for the oversight.

I tested it with this script and possible correct and incorrect (the 2nd url is invalid) values.
Tell me if there are some other cases to test.

```python
import re

# Regex pattern
pattern = "^https?://(?:www\.)?(?:\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|[a-zA-Z0-9.-]+)(?::\d+)?/(?:[^/\s]+/)*[^/\s]+\.json$"

# Test URLs
urls = [
    "http://192.168.1.140:9000/1%202%20Inflación-wav.json",
    "http:/google.com/test.json",
    "https://google.com/test.json",
    "https://www.google.com/test.json",
    "https://www.google.com/folder/subfolder/test.json",
    "https://www.google.com/folder/subfolder/folder/test.json"
]

# Checking each URL against the pattern
for url in urls:
    if re.match(pattern, url):
        print(f"Match found: {url}")
    else:
        print(f"No match: {url}")
```

taranjeet: this looks good. thanks for your contribution @juananpe

taranjeet:
This looks good. Merging it.
thanks @juananpe

PR #1251, opened 2024-02-09
Fix links to the Discourse docs in the Discourse Loader

oskarrough:
## Description

The Discourse loader contains links to the website docs, but they were not correct. Now they should be.

## Type of change

- [X] Refactor (does not change functionality, e.g. code style improvements, linting)


deshraj (approved):
Thanks for fixing.

PR #1278, opened 2024-02-21
Updating python version to including <=3.13

joaomdmoura:
Allow for people to use with more recent version of python


deshraj:
Thanks for adding this @joaomdmoura. We were already planning to upgrade it soon. 

I will merge this PR and will update the poetry.lock file in a follow up.

PR #1337, opened 2024-04-11
[Improvements] Improve logging and fix insertion in data_sources table

deshraj:
## Description

Changes:
- Fix issue when the source type in data_sources is not a string. Convert it specifically
- Convert the print statements to logging

## Type of change


deshraj:
Qdrant related tests seems to be failing due to a different issue. We will fix it in a follow up PR.

PR #1675, opened 2024-08-09
Add support for pgvector

Dev-Khant:
## Description

Add support for pgvector as vector db and also update the docs.

Fixes # (issue)

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [ ] Unit Test
```python
import os
from mem0 import Memory

os.environ["OPENAI_API_KEY"] = "sk-xx"

config = {
    "vector_store":{
        "provider": "pgvector",
        "config": {
            "user": "test",
            "password": "123",
        }
    }
}

m.add("visiting to paris", user_id="test", run_id="123")
```

## Checklist:

## Maintainer Checklist

- [ ] Made sure Checks passed

### DIFF
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -1,0 +1,21 @@
+from typing import Optional
+
+from pydantic import BaseModel, Field, model_validator
+
+class PGVectorConfig(BaseModel):
+
+    dbname: str = Field("postgres", description="Default name for the database")
+    collection_name: str = Field("mem0", description="Default name for the collection")
+    embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
+    user: Optional[str] = Field(None, description="Database user")
+    password: Optional[str] = Field(None, description="Database password")
+    host: Optional[str] = Field("127.0.0.1", description="Database host. Default is localhost")
+    port: Optional[int] = Field(5432, description="Database port. Default is 1536")
+
+    @model_validator(mode="before")
+    def check_user_and_password(cls, values):
+        user, password = values.get("user"), values.get("password")
+        if not user and not password:
+            raise ValueError("Both 'user' and 'password' must be provided.")
+        return values
+    
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Dict, Any
 
 from pydantic import BaseModel, Field, model_validator
 
@@ -20,5 +20,15 @@ class PGVectorConfig(BaseModel):
             raise ValueError("Both 'user' and 'password' must be provided.")
         if not host and not port:
             raise ValueError("Both 'host' and 'port' must be provided.")
+        return values
+    
+    @model_validator(mode="before")
+    @classmethod
+    def validate_extra_fields(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        allowed_fields = set(cls.model_fields.keys())
+        input_fields = set(values.keys())
+        extra_fields = input_fields - allowed_fields
+        if extra_fields:
+            raise ValueError(f"Extra fields not allowed: {', '.join(extra_fields)}. Please input only the following fields: {', '.join(allowed_fields)}")
         return values
     
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -9,13 +9,16 @@ class PGVectorConfig(BaseModel):
     embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
     user: Optional[str] = Field(None, description="Database user")
     password: Optional[str] = Field(None, description="Database password")
-    host: Optional[str] = Field("127.0.0.1", description="Database host. Default is localhost")
-    port: Optional[int] = Field(5432, description="Database port. Default is 1536")
+    host: Optional[str] = Field(None, description="Database host. Default is localhost")
+    port: Optional[int] = Field(None, description="Database port. Default is 1536")
 
     @model_validator(mode="before")
-    def check_user_and_password(cls, values):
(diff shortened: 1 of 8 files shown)


mem0/vector_stores/pgvector.py
```diff
@@ -0,0 +1,257 @@
...
+        Insert vectors into a collection.
+
+        Args:
+            name (str): Name of the collection.
+            vectors (List[List[float]]): List of vectors to insert.
+            payloads (List[Dict], optional): List of payloads corresponding to vectors.
+            ids (List[str], optional): List of IDs corresponding to vectors.
+        """
+        if payloads is None:
+            payloads = [{}] * len(vectors)
```
deshraj, line 82: This feels weird. Why do we need to do this?

mem0/vector_stores/pgvector.py
```diff
@@ -0,0 +1,257 @@
...
+        if ids is None:
+            ids = [None] * len(vectors)
```
deshraj, line 84: Can you please elaborate why we need this?

mem0/vector_stores/pgvector.py
```diff
@@ -0,0 +1,257 @@
...
+        return OutputData(id=str(result[0]), score=None, payload=result[2])
+
+    def list_cols(self) -> List[str]:
+        """
+        List all collections.
+
+        Returns:
+            List[str]: List of collection names.
+        """
+        self.cur.execute("SELECT table_name FROM information_schema.tables WHERE table_schema = 'public'")
```
deshraj, line 178: This is going to list all the tables in the db, right?

We probably dont want that. We want to only list the schema for the mem0 related table.

mem0/configs/vector_stores/pgvector.py
```diff
@@ -0,0 +1,21 @@
...
+from pydantic import BaseModel, Field, model_validator
+
+class PGVectorConfig(BaseModel):
+
+    dbname: str = Field("postgres", description="Default name for the database")
+    collection_name: str = Field("mem0", description="Default name for the collection")
+    embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
+    user: Optional[str] = Field(None, description="Database user")
+    password: Optional[str] = Field(None, description="Database password")
+    host: Optional[str] = Field("127.0.0.1", description="Database host. Default is localhost")
```
deshraj, line 12: We should not set the default value for embedding_model_dims, host and port. User should specify that.

mem0/vector_stores/pgvector.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Args:
+            name (str): Name of the collection.
+            embedding_model_dims (int, optional): Dimension of the embedding vector.
+        """
+        self.cur.execute(f"""
+            CREATE TABLE IF NOT EXISTS {name} (
+                id UUID PRIMARY KEY,
+                vector vector({embedding_model_dims}),
+                payload JSONB
```
deshraj, line 66: Do we need the payload column? Why not create column for the keys in the payload since most of the fields are going to be standard anyways.

deshraj (approved):
Generally LGTM but left some small comments.

mem0/vector_stores/pgvector.py
Dev-Khant: Ohh wait this was not supposed to be there.

I was basically checking that if payload is none we would have to insert some value along with vectors into the db, because when I was testing somewhere it caused such error.

But then it was supposed to removed because in `insert` we won't have a case where payloads or ids would be None.

mem0/vector_stores/pgvector.py
Dev-Khant: Same here needs to be removed.

mem0/configs/vector_stores/pgvector.py
Dev-Khant: We need to set default embedding_model_dims because our default embedder is openai so it's dimension needs to be set.

I'll make mandatory for user to enter host and port.

mem0/vector_stores/pgvector.py
Dev-Khant: Honestly it was easy to use payload directly :) and I didn't see any benefit for creating columns for individual keys.

deshraj: What are the keys inside the payload column? Based on that, we may have to add columns here.

mem0/vector_stores/pgvector.py
```diff
@@ -0,0 +1,253 @@
...
+            );
+        """)
+        self.conn.commit()
+
+    def insert(self, name, vectors, payloads = None, ids = None):
```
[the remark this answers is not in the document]
Dev-Khant: `name` is being passed from the `main.py`, so we have to keep a common set of params to accept in vectordb methods.

mem0/vector_stores/pgvector.py
Dev-Khant: data, user_id, agent_id, run_id, metadata, filters. And the reason to directly use payload is we have the same flow for other vectordbs as well.

mem0/vector_stores/pgvector.py
```diff
@@ -0,0 +1,253 @@
...
+            user=user,
+            password=password,
+            host=host,
+            port=port
+        )
+        self.cur = self.conn.cursor()
+
+        collections = self.list_cols()
+        if collection_name not in collections:
+            self.create_col(collection_name, embedding_model_dims)
```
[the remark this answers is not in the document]
Dev-Khant: Yes you are correct but it needs change for all the vectordbs then.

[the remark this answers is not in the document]
Dev-Khant: So I'll make a change for all the vectordbs and also update`main.py`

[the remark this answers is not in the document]
Dev-Khant: Done.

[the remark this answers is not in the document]
deshraj: Thanks.

deshraj (approved):

PR #1748, opened 2024-08-25
Currently, the graph memory search is not filtering for the user

rajib76:
## Description

Currently the graph memory search is not filtering for the user. It is returning the memory for all users irrespective of the user id filter. Made the change to ensure that only user specific graph memory is returend

### DIFF
--- a/mem0/memory/main_graph.py
+++ b/mem0/memory/main_graph.py
@@ -214,7 +214,7 @@ class MemoryGraph:
         self.graph.query(cypher)
     
 
-    def get_all(self):
+    def get_all(self,filters):
         """
         Retrieves all nodes and relationships from the graph database based on optional filtering criteria.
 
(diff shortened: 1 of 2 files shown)


mem0/memory/main_graph.py
```diff
@@ -228,10 +228,10 @@ def get_all(self):
 
         # return all nodes and relationships
         query = """
-        MATCH (n)-[r]->(m)
+        MATCH (n)-[r]->(m) WHERE toLower(n.name) = toLower($name)
```
prateekchhikara, line 231: Hi @rajib76, this is wrong. This will only return then nodes which are connected to the root node or the user_id node. For instance, the relation: alice123 loves pizza and john loves pizza. Your changes will just return alice123 loves pizza even if all of these nodes have been added using the same user_id.

prateekchhikara:
This feature has been added in the new release and hence closing the PR.

PR #1779, opened 2024-08-29
Improvements to Graph Memory

prateekchhikara:
## Description

Following improvements have been made to the Graph Memory

1. Graph Memory now supports all existing LLMs, previously it only supported ```gpt-4o-2024-08-06```.
2. Users can now give their custom prompt to extract entities from the text to create nodes and relationships.
3. Improved Graph Memory search using BM25 algorithm, previously the search result was random.
4. A new openai structured class has been added to the LLMs, now user can use the ```gpt-4o-2024-08-06``` model using the ```openai_structured``` llm provider.

## Type of change

### DIFF
--- a/mem0/graphs/configs.py
+++ b/mem0/graphs/configs.py
@@ -1,6 +1,6 @@
 from typing import Optional
-
 from pydantic import BaseModel, Field, field_validator, model_validator
+from mem0.llms.configs import LlmConfig
 
 class Neo4jConfig(BaseModel):
     url: Optional[str] = Field(None, description="Host address for the graph database")
@@ -30,6 +30,14 @@ class GraphStoreConfig(BaseModel):
         description="Configuration for the specific data store",
         default=None
     )
+    llm: Optional[LlmConfig] = Field(
+        description="LLM configuration for querying the graph store",
+        default=None
+    )
+    custom_prompt: Optional[str] = Field(
+        description="Custom prompt to fetch entities from the given text",
+        default=None
+    )
 
     @field_validator("config")
     def validate_config(cls, v, values):
@@ -38,3 +46,4 @@ class GraphStoreConfig(BaseModel):
             return Neo4jConfig(**v.model_dump())
         else:
             raise ValueError(f"Unsupported graph store provider: {provider}")
+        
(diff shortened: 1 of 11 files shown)


mem0/graphs/utils.py
```diff
@@ -37,6 +37,7 @@
 1. Extract only explicitly stated information from the text.
 2. Identify nodes (entities/concepts), their types, and relationships.
 3. Use "USER_ID" as the source node for any self-references (I, me, my, etc.) in user messages.
+CUSTOM_PROMPT
```
Dev-Khant, line 40: Shouldn't this be inside a {} ?

mem0/graphs/configs.py
```diff
@@ -30,6 +30,14 @@ class GraphStoreConfig(BaseModel):
         description="Configuration for the specific data store",
         default=None
     )
+    llm: Optional[LlmConfig] = Field(
+        description="LLM configuration for querying the graph store",
+        default=None
```
Dev-Khant, line 35: You should set a default model here if you are not doing it at any other place.

mem0/memory/main_graph.py
```diff
@@ -1,51 +1,29 @@
...
+        self.embedding_model = EmbedderFactory.create(
+            self.config.embedder.provider, self.config.embedder.config
+        )
+
+        if self.config.llm.provider:
+            llm_provider = self.config.llm.provider
+        if self.config.graph_store.llm:
+            llm_provider = self.config.graph_store.llm.provider
+        else:
+            llm_provider = "openai_structured"
```
Dev-Khant, line 22: This can be set as default in config.

mem0/graphs/utils.py
prateekchhikara: I am using .replace() here, so it just replaces the string directly

Dev-Khant: Sounds good.

mem0/graphs/configs.py
prateekchhikara: Right, good point!!
https://github.com/mem0ai/mem0/blob/752ddc746391f936dad96cb4d61e300e06f95af8/mem0/llms/openai_structured.py#L14

Will this take care of the case that you mentioned?

mem0/memory/main_graph.py
prateekchhikara: Yes, but I am keeping some checks.

1. If the main LLM provider is set then use it. 
2. If the graph LLM provider is ser then use it.
3. If nothing is set then use openai_structured.

mem0/graphs/configs.py
Dev-Khant: Not exactly because here you would be passing the default config something like this: https://github.com/mem0ai/mem0/blob/main/mem0/configs/base.py#L40.

The line you are mentioning would only set the default model for Openai provider.

Issue #1841, opened 2024-09-09
[bug]: new_memories_with_actions sometimes produces response that are not JSON for Azure OpenAI


spike-spiegel-21:
## User feedback 
"It seems the issue I’m facing relates to how memories are being added and searched in the Mem0 storage. Specifically, I’m getting a JSONDecodeError, which indicates there’s an issue with how the JSON data is structured when passed to the _add_to_vector_store method—likely due to an unterminated string in the JSON. Additionally, while searching memories, I’m seeing errors like “string indices must be integers, not ‘str’,” which suggests that there’s an issue with how the search functionality is processing the stored data. Despite these errors, some interactions are successfully adding memory, but others are not, likely due to malformed or improperly formatted JSON data being passed between the memory storage and the vector store. Any suggestions on how to resolve these issues would be greatly appreciated!"

![image](https://github.com/user-attachments/assets/a63b9abb-03af-46c2-b1e4-a7663395bcfc)

PranavPuranik:
oh lol, I was just about to create an issue for this as I am facing the same in Llama. @Dev-Khant @prateekchhikara can I work on this?

shinoxzu:
the same issue still here. 

```
 File "/app/.venv/lib/python3.13/site-packages/mem0/memory/main.py", line 199, in _add_to_vector_store
bot-1     |     new_memories_with_actions = json.loads(new_memories_with_actions)
bot-1     |   File "/usr/local/lib/python3.13/json/__init__.py", line 346, in loads
bot-1     |     return _default_decoder.decode(s)
bot-1     |            ~~~~~~~~~~~~~~~~~~~~~~~^^^
bot-1     |   File "/usr/local/lib/python3.13/json/decoder.py", line 345, in decode
bot-1     |     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
bot-1     |                ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
bot-1     |   File "/usr/local/lib/python3.13/json/decoder.py", line 361, in raw_decode
bot-1     |     obj, end = self.scan_once(s, idx)
bot-1     |                ~~~~~~~~~~~~~~^^^^^^^^
bot-1     | json.decoder.JSONDecodeError: Unterminated string starting at: line 15 column 21 (char 330)
```

i am using gpt-4o model.

prateekchhikara:
@shinoxzu can you please share your input so that we can reproduce this error at our end?

PR #1842, opened 2024-09-09
[improvement]: system message added

spike-spiegel-21:
resolves issue: #1841 

This llm call contains a system message:
https://github.com/mem0ai/mem0/blob/bbddb98aca3bbe8f544b0e0faeb55f69b6691104/mem0/memory/main.py#L136-L139

mentioning to give output in json format
https://github.com/mem0ai/mem0/blob/bbddb98aca3bbe8f544b0e0faeb55f69b6691104/mem0/configs/prompts.py#L57

But here system message is not present, which sometimes give non-JSON output
https://github.com/mem0ai/mem0/blob/bbddb98aca3bbe8f544b0e0faeb55f69b6691104/mem0/memory/main.py#L163-L166

Refrence: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/json-mode?tabs=python

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -161,7 +161,8 @@ class Memory(MemoryBase):
 
         function_calling_prompt = get_update_memory_messages(retrieved_old_memory, new_retrieved_facts)
         new_memories_with_actions = self.llm.generate_response(
-            messages=[{"role": "user", "content": function_calling_prompt}],
+            messages=[{"role": "system", "content": "You are a helpful assistant designed to output JSON."}, 
+                      {"role": "user", "content": function_calling_prompt}],
             response_format={"type": "json_object"},
         )
         new_memories_with_actions = json.loads(new_memories_with_actions)


Dev-Khant (commented):
Hey @spike-spiegel-21, we're already including `response_format={"type": "json_object"}`, so it's being applied to the models that require it and for others `system` is sufficient.

Can you please showcase a usecase where your approach is performing better than current approach?

spike-spiegel-21:
image

FYI, @Dev-Khant 
Reference: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/json-mode?tabs=python

Also, we are including the system's message in first LLM call, but not in new memories with action.

Dev-Khant:
@spike-spiegel-21 We are instructing the system_prompt to return output in JSON format within the `function_calling_prompt`. We've specifically stated there to provide the response only in JSON format.

spike-spiegel-21:
@Dev-Khant Yes we are instructing it. But it is not a part of **system's message**, it is in the **user's role**. 
https://github.com/mem0ai/mem0/blob/bbddb98aca3bbe8f544b0e0faeb55f69b6691104/mem0/memory/main.py#L163-L166

But here it is a part of system's message too. 
https://github.com/mem0ai/mem0/blob/bbddb98aca3bbe8f544b0e0faeb55f69b6691104/mem0/memory/main.py#L136-L139

Dev-Khant:
> Hey @spike-spiegel-21, we're already including `response_format={"type": "json_object"}`, so it's being applied to the models that require it and for others `system` is sufficient.
> 
> Can you please showcase a usecase where your approach is performing better than current approach?

Got it, but could you provide an example of how that would be more effective? If it proves to work better, we can move all the output instructions to the system prompt instead of only including the `JSON` instruction there.

Dev-Khant:
Closing this as this change will be irrelevant with upcoming graph changes.

PR #1851, opened 2024-09-11
Fixed environment variables priority in OpenAILLM

### LINKED ISSUE #1850: `BaseLlmConfig` and `OpenAIEmbedder` have a conflict `openai_base_url`
### 🐛 Describe the bug

I'm using the latest version of `mem0`:
```
$ ./bin/pip list | grep mem0ai
mem0ai                   0.1.13
```

The `OpenAILLM` has a client initialization, which is declared here:
https://github.com/mem0ai/mem0/blob/ac7b7aa20a6bacca5ef90edb9cd99415d552bec4/mem0/llms/openai.py#L25

When specifying `OPENAI_API_BASE` in my code:
```python
os.environ["OPENAI_API_KEY"] = "xxxxxxxxxxxx"
os.environ["OPENAI_API_BASE"] = "https://dashscope.aliyuncs.com/compatible-mode/v1"
```

The variable `base_url` I just mentioned above **SHOULD** be `https://dashscope.aliyuncs.com/compatible-mode/v1`, but it was `https://api.openai.com/v1` in my test progress.

```python
  File "/project/lib/python3.12/site-packages/mem0/llms/openai.py", line 26, in __init__
    assert base_url == "https://dashscope.aliyuncs.com/compatible-mode/v1"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

Simply changing the order can fix this.

```diff
- base_url = self.config.openai_base_url or os.getenv("OPENAI_API_BASE")
+ base_url = os.getenv("OPENAI_API_BASE") or self.config.openai_base_url
```

FoliageOwO:
## Description

In the initialization process of the `OpenAILLM`, the environment variables should be prioritized because there exists an initial value in the `BaseLlmConfig` already.

If not, setting the environment variable `OPENAI_API_BASE` does nothing and can not affect any of the configs, and the `base_url` of the `OpenAILLM` will be initialized with the value of `https://api.openai.com/v1`.

Fixes #1850

## Type of change

## How Has This Been Tested?

Specifying a base URL in the entrance of the project.

```python
from mem0 import Memory
os.environ["OPENAI_API_BASE"] = "https://dashscope.aliyuncs.com/compatible-mode/v1"
config = {
    "embedder": {
        "provider": "openai",
        "config": {
            "model": "qwen-long"
        }
    }
}
```

Simply adding a log (print, logger, etc) below the code can check whether `base_url` is correct.

https://github.com/mem0ai/mem0/blob/ac7b7aa20a6bacca5ef90edb9cd99415d552bec4/mem0/llms/openai.py#L25

Before fixing the issue, the code below will print `https://api.openai.com/v1`.

```diff
  base_url = self.config.openai_base_url or os.getenv("OPENAI_API_BASE")
+ print(base_url)
```

After fixing the issue, the code below will print `https://dashscope.aliyuncs.com/compatible-mode/v1`.

```diff
- base_url = self.config.openai_base_url or os.getenv("OPENAI_API_BASE")
+ base_url = os.getenv("OPENAI_API_BASE") or self.config.openai_base_url
+ print(base_url)
```

## Checklist:

## Maintainer Checklist

- [ ] Closes #1850
- [ ] Made sure Checks passed

### DIFF
--- a/mem0/llms/openai.py
+++ b/mem0/llms/openai.py
@@ -22,7 +22,7 @@ class OpenAILLM(LLMBase):
             )
         else:
             api_key = self.config.api_key or os.getenv("OPENAI_API_KEY")
-            base_url = self.config.openai_base_url or os.getenv("OPENAI_API_BASE")
+            base_url = os.getenv("OPENAI_API_BASE") or self.config.openai_base_url
             self.client = OpenAI(api_key=api_key, base_url=base_url)
 
     def _parse_response(self, response, tools):


shenxiangzhuang:
Hi, @FoliageOwO @prateekchhikara. IMHO, this change will introduce the problem as shown in #1815 again. In short, when we set a `openai_base_url` parameter explicitly, we may use the environment variable implicitly. Explicit is better than implicit, so the library should prefer the 
 given explicit parameter over the implicit  environment variable.

I think removing the default values(assign with `None`) in base classes(and use config parameter before the environment variable) is a better way to overcome this problem.

FoliageOwO:
Hello @shenxiangzhuang, thanks for reviewing the PR. (I'm not using a grammar checker so forgive my grammar errors 😃 )

Sorry for not searching for similar issue in the PR list. It's obvious that this PR will do introduce the issue you have mentioned above, and I can't agree more with your opinion "Explicit is better than implicit".

So do I, I request the maintainer @prateekchhikara to revert the change and raise another PR which will remove the default value to fix the problem thoroughly.

shenxiangzhuang:
> EDIT: But the way of removing the default value will also introduce another user-end trouble: the users must assign a value by themselves. This may have a breaking change in users' code if the final PR is merged to the main branch and there releases a new version. Almost all the users who are using the official OpenAI API need to change their codes and add the configuration.
> 
> Are there any other ways to fix this problem better?

@FoliageOwO Great catch! I believe we can set the default base URL later (if no related parameter is passed or no related environment exists) to avoid a breaking change. I've created a pr #1861 to address this. Please feel free to share your suggestions on this.

FoliageOwO:
@shenxiangzhuang Thanks for coping with the issue. 😃 

(I clicked the `Revert` button accidentally, it doesn't be committed and shouldn't be, please ignore the revert commit.)

PR #2021, opened 2024-11-09
Fix not working with Gemini models

lh0x00:
## Description

This PR's did the following:
- Fixed missing `gemini` in LLM's provider check.
- Handled case where model returns json in code block.
- Fixed Gemini model not working: removed unnecessary `noop`, updated function to parse function calls returning Gemini (based on their source code/docs).

All tested by my app working properly again.

## Type of change

### DIFF
--- a/mem0/llms/configs.py
+++ b/mem0/llms/configs.py
@@ -11,15 +11,15 @@ class LlmConfig(BaseModel):
     def validate_config(cls, v, values):
         provider = values.data.get("provider")
         if provider in (
-            "ollama",
             "openai",
+            "ollama",
+            "anthropic",
             "groq",
             "together",
             "aws_bedrock",
             "litellm",
             "azure_openai",
             "openai_structured",
-            "anthropic",
             "azure_openai_structured",
             "gemini",
         ):
--- a/mem0/llms/configs.py
+++ b/mem0/llms/configs.py
@@ -11,16 +11,17 @@ class LlmConfig(BaseModel):
     def validate_config(cls, v, values):
         provider = values.data.get("provider")
         if provider in (
-            "openai",
             "ollama",
-            "anthropic",
+            "openai",
             "groq",
             "together",
             "aws_bedrock",
             "litellm",
             "azure_openai",
             "openai_structured",
+            "anthropic",
             "azure_openai_structured",
+            "gemini",
         ):
             return v
         else:
--- a/mem0/llms/gemini.py
+++ b/mem0/llms/gemini.py
@@ -1,4 +1,5 @@
 import os
+import json
 from typing import Dict, List, Optional
 
 try:
@@ -44,11 +45,9 @@ class GeminiLLM(LLMBase):
 
             for part in response.candidates[0].content.parts:
                 if fn := part.function_call:
+                    fn_call = type(fn).to_dict(fn)
                     processed_response["tool_calls"].append(
-                        {
-                            "name": fn.name,
-                            "arguments": {key: val for key, val in fn.args.items()},
-                        }
+                        {"name": fn_call["name"], "arguments": fn_call["args"]}
                     )
 
             return processed_response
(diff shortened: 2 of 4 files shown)


lh0x00:
@Dev-Khant Gemini will not be available for now, I hope you consider and merge it soon if possible.

mem0/llms/configs.py
```diff
@@ -11,16 +11,17 @@ class LlmConfig(BaseModel):
     def validate_config(cls, v, values):
         provider = values.data.get("provider")
         if provider in (
-            "openai",
             "ollama",
-            "anthropic",
```
Dev-Khant, line 16: Hey can you please keep the other providers as is and just add the `gemini`.

lh0x00:
@spike-spiegel-21 @Dev-Khant While working on the library I resolved some related bugs so that Gemini could work with `mem0` as documented.

mem0/llms/gemini.py
```diff
@@ -44,11 +45,9 @@ def _parse_response(self, response, tools):
 
             for part in response.candidates[0].content.parts:
                 if fn := part.function_call:
+                    fn_call = type(fn).to_dict(fn)
```
lh0x00, line 48: The old syntax doesn't work properly with some cases of functions with deep values ​​or iterations. The update is referenced this way: https://ai.google.dev/gemini-api/tutorials/extract_structured_data

mem0/llms/gemini.py
```diff
@@ -138,9 +137,9 @@ def generate_response(
             "top_p": self.config.top_p,
         }
 
-        if response_format:
+        if response_format is not None and response_format == "json_object":
```
lh0x00, line 140: This code makes Gemini callable and returns correctly (no schema defined yet, it will rely on prompts)

mem0/memory/graph_memory.py
```diff
@@ -106,6 +106,10 @@ def add(self, data, filters):
                 ADD_MEMORY_STRUCT_TOOL_GRAPH,
                 NOOP_STRUCT_TOOL,
             ]
+        elif self.llm_provider == "gemini":
+            # The `noop` ​​function n should be removed because it is unnecessary 
+            # and causes the error: "should be non-empty for OBJECT type"
+            _tools = [UPDATE_MEMORY_TOOL_GRAPH, ADD_MEMORY_TOOL_GRAPH]
```
lh0x00, line 112: Trying to add a `noop` ​​function here seems unnecessary, and will return an error that the value in the object cannot be empty (from the Gemini related library).

lh0x00:
@Dev-Khant I updated to resolve test fails.

lh0x00:
Is this PR still under consideration my friends? @Dev-Khant @spike-spiegel-21

mem0/memory/graph_memory.py
Dev-Khant: Hey @lh0x00 we want to keep the `noop` operation as it is present for all the llms, so is there a workaround for this? Because removing this would be inconsistent.

Dev-Khant (commented):
@lh0x00 Why is there a need for removing code block from the generated response? Can you please elaborate more on this?

mem0/memory/graph_memory.py
lh0x00: @Dev-Khant I had an issue with Gemini not allowing empty arguments, similar to this: https://github.com/vercel/ai/issues/3371
Maybe we will update it again after it supports OpenAI APIs like here: https://ai.google.dev/gemini-api/docs/openai
But currently Gemini still has some bugs so we can't use it directly, and we need a long time to wait for it.

lh0x00:
@Dev-Khant I do it because sometimes models will mistakenly return content inside codeblocks, and it's best to remove it (if present, as a precaution).

mem0/memory/graph_memory.py
Dev-Khant: Alright, understood your point here. I'll let @deshraj take the final call.

Dev-Khant:
Got it @lh0x00.

Dev-Khant:
Hey @lh0x00 Can you please resolve the merge conflicts and then we are good to go?

lh0x00:
@Dev-Khant  I rebased it.

Dev-Khant (approved):

Dev-Khant:
@lh0x00 Thanks for the contribution, changes looks good to me.

PR #2073, opened 2024-12-04
[graph_memory]: improve delete/add graph memory

spike-spiegel-21:
### Description
Flow of graph memory addtion:

1. Extracts all the entities/nodes and their types mentioned in the query. (Returns: dictionary of {nodes:type})
2. Eshtablish relations among the extracted nodes. (Input: list of nodes(that does not contains types). Returns: list of source -- relationship -- destination)

_(Note: The above 2 LLM calls are made to ensure that all the facts are retrieved from the query/data. The single LLM seems to be overloaded if we expect source (and their types) -- relationship -- destination (and their types))_

3. Search similar nodes and their respective inbound and outbound relations. 
4. LLM call to find what relationship we want to delete based on the query/data. (it returns the relations that we have to delete 

Now we have a list of entities (source -- relationship -- destination) that we need to add and delete. 

5. Delete relationships
6. Add relationships
- Find whichever node is common (source and destination) via similarity search (0.9 threshold) and then  then merge it.

### DIFF
--- a/mem0/graphs/tools.py
+++ b/mem0/graphs/tools.py
@@ -85,7 +85,7 @@ NOOP_TOOL = {
 RELATIONS_TOOL = {
     "type": "function",
     "function": {
-        "name": "establish_relations",
+        "name": "establish_relationships",
         "description": "Establish relationships among the entities based on the provided text.",
         "parameters": {
             "type": "object",
@@ -99,7 +99,7 @@ RELATIONS_TOOL = {
                                 "type": "string",
                                 "description": "The source entity of the relationship."
                             },
-                            "relation": {
+                            "relationship": {
                                 "type": "string",
                                 "description": "The relationship between the source and destination entities."
                             },
@@ -109,9 +109,9 @@ RELATIONS_TOOL = {
                             },
                         },
                         "required": [   
-                            "source_entity",
-                            "relation",
-                            "destination_entity",
+                            "source",
+                            "relationship",
+                            "destination",
                         ],
                         "additionalProperties": False,
                     },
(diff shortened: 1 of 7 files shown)


mem0/graphs/tools.py
```diff
@@ -109,9 +109,9 @@
                             },
                         },
                         "required": [   
-                            "source_entity",
-                            "relation",
-                            "destination_entity",
+                            "source",
+                            "relationship",
+                            "destination",
```
Dev-Khant, line 114: It would be better if `destination` is changed to `target`

mem0/graphs/tools.py
```diff
@@ -321,3 +321,66 @@
...
+            "properties": {
+                "source": {
+                    "type": "string",
+                    "description": "The identifier of the source node in the relationship.",
+                },
+                "relationship": {
+                    "type": "string",
+                    "description": "The existing relationship between the source and destination nodes that needs to be deleted.",
+                },
+                "destination": {
```
Dev-Khant, line 342: Same here

tests/test_proxy.py
```diff
@@ -94,4 +94,4 @@ def test_completions_create_with_system_message(mock_memory_client, mock_litellm
 
     call_args = mock_litellm.completion.call_args[1]
     assert call_args["messages"][0]["role"] == "system"
-    assert call_args["messages"][0]["content"] == MEMORY_ANSWER_PROMPT
+    assert call_args["messages"][0]["content"] == "You are a helpful assistant."
```
Dev-Khant, line 97: What is the reason for hard coding the `You are a helpful assistant.` because it might change as user passes a different system prompt.

spike-spiegel-21: Since this commit #2097 this test was failing.

Dev-Khant: Got it thanks

PR #1579, opened 2024-07-25
feat: FastEmbed embeddings

Anush008:
## Description

This PR adds support for FastEmbed - https://github.com/qdrant/fastembed to be used to generate embeddings in Mem0 and Embedchain.

PR #2083, opened 2024-12-11
feat: FastEmbed Embeddings

Anush008:
## Description

This PR adds support for FastEmbed - https://github.com/qdrant/fastembed to be used to generate embeddings in Mem0 and Embedchain.

Renewed #1579


kartik-mem0:
Thanks for putting this together! FastEmbed support has since been added via #3552, so this one is no longer needed. Closing it out — appreciate the contribution effort here.

anush008:
Thanks.

PR #2111, opened 2024-12-23
Feature: VoyageAI embeddings integration

Adversarian:
## Description

This PR integrates VoyageAI's embedding models for use with `mem0`.

## Type of change

- [X] Refactor (does not change functionality, e.g. code style improvements, linting)
- [X] Documentation update

## How Has This Been Tested?

- [X] Unit Test

### DIFF
--- a/docs/components/embedders/models/voyageai.mdx
+++ b/docs/components/embedders/models/voyageai.mdx
@@ -1,0 +1,37 @@
+---
+title: VoyageAI
+---
+
+To use VoyageAI embedding models, set the `VOYAGE_API_KEY` environment variable. You can obtain the Voyage API key from the [VoyageAI Dashboard](https://dash.voyageai.com/api-keys).
+
+### Usage
+
+```python
+import os
+from mem0 import Memory
+
+os.environ["VOYAGE_API_KEY"] = "your_api_key"
+os.environ["OPENAI_API_KEY"] = "your_api_key" # For LLM
+
+config = {
+    "embedder": {
+        "provider": "voyageai",
+        "config": {
+            "model": "voyage-3"
+        }
+    }
+}
+
+m = Memory.from_config(config)
+m.add("I'm visiting Paris", user_id="john")
+```
+
+### Config
+
+Here are the parameters available for configuring VoyageAI embedder:
+
+| Parameter | Description | Default Value |
+| --- | --- | --- |
+| `model` | The name of the embedding model to use | `voyage-3` |
+| `embedding_dims` | Dimensions of the embedding model | `None`(which uses the [pre-defined default values](https://docs.voyageai.com/docs/embeddings) according to the selected model) |
+| `api_key` | The VoyageAI API key | `None` |
--- a/docs/components/embedders/overview.mdx
+++ b/docs/components/embedders/overview.mdx
@@ -16,6 +16,7 @@ See the list of supported embedders below.
   <Card title="Gemini" href="/components/embedders/models/gemini"></Card>
   <Card title="Vertex AI" href="/components/embedders/models/vertexai"></Card>
   <Card title="Together" href="/components/embedders/models/together"></Card>
+  <Card title="VoyageAI" href="/components/embedders/models/voyageai"></Card>
 </CardGroup>
 
 ## Usage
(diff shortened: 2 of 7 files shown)


Adversarian:
@Dev-Khant I'm sorry for the ping, I noticed the latest couple of PRs were merged by you so I assumed you're the designated reviewer. I was wondering if I could get a review on this? We're using VoyageAI embeddings internally and I'd like to know if there's a problem with this PR that I can fix or not. I'd appreciate it if you'd let me know.

Dev-Khant:
Hi @Adversarian,  

Sure, I’d review the PR. Before diving in, can I kindly ask how this integration is beneficial for you? You mentioned using it internally, and we would like to understand the use case you’re applying it to.

Adversarian:
> Hi @Adversarian,
> 
> Sure, I’d review the PR. Before diving in, can I kindly ask how this integration is beneficial for you? You mentioned using it internally, and we would like to understand the use case you’re applying it to.

Of course! Thank you for taking the time. As I said, our embedding provider (VoyageAI) is not listed in any of your the providers that you currently support, this simply extends the list of embedders with another well-known provider so that me and people who would like to use VoyageAI as their embedding provider can use their services in tandem with mem0.

If you don't think this is a valid use-case I would appreciate hearing your thoughts as to why.

nocyphr:
@Adversarian hey it's been a while, I just made a crude drop-in [patch](https://pypi.org/project/mem0-embeddings-litellm-patch/) you can install via pypi. I implemented it with litellm as abstraction layer so you can plug in whatever embedding provider litellm is supporting (including voyage ai). 
It is confirmed to be working for the current version of mem0ai (0.1.115) using a python:3.11-slim docker container. I willl have to find a more elegant way of patching their code instead of just drop-in replacing their files, add tests, check actual range of versions supported into the past, etc but for now if you are using the latest version of mem0 you can just install the patch and then configure your embeddings like so: 

```python
embedconfig = {
    "provider": "litellm",
    "config": {
        "api_key": os.getenv("VOYAGEAI_API_KEY"),
        "model": "voyage/voyage-3-large",
        "embedding_dims": 1024
    },
}
```
Personally I would prefer if @Dev-Khant would just merge the prs for things like bugs and additions of industry leading providers or use the litellm abstraction layer for embeddings (it's used for llms already, not sure why stop there) so they don't have to keep up with the latest themselves, that way i don't need to track changes to their files to see when my patches are breaking or freeze my mem0 version ;D

PS: Also made another [db patch](https://pypi.org/project/mem0-history-db-patch/) that switches out the hardcoded sqlite for a more powerful sqlalchemy version and lets you use eg postgres db instead of sqlite for scalability. (or whatever else is supported by sqlalchemy, I only care to switch from sqlite to postgres)

Adversarian:
@nocyphr We have moved on from `mem0` but this looks like a very elegant solution and it didn't occur to me to add support for `litellm` instead of VoyageAI specifically. Thank you for your effort.

nocyphr:
> @nocyphr We have moved on from `mem0` but this looks like a very elegant solution and it didn't occur to me to add support for `litellm` instead of VoyageAI specifically. Thank you for your effort.

@Adversarian  If you don't mind me asking, what have you moved on to use instead?

Adversarian:
> > @nocyphr We have moved on from `mem0` but this looks like a very elegant solution and it didn't occur to me to add support for `litellm` instead of VoyageAI specifically. Thank you for your effort.
> 
> @Adversarian If you don't mind me asking, what have you moved on to use instead?

@nocyphr Sorry for the late response, it slipped my mind to get back to you sooner. We were using Llama-Index to orchestrate our application internally. Llama-Index dropped support for `mem0` and it's new memory system was enough for our use case. We simply migrated from the old one (`ChatMemoryBuffer` if you're familiar) to the new system with memory blocks. I imagine it might not cover as much ground as what `mem0` claims to do but it was starting to get a bit old going back and forth and getting nowhere with this PR so we just decided to work with what's available and it turned out just fine.

ranfysvalle02:
I'd like to re-open this one --- Any objections? 

The value to me is, I use VoyageAI heavily for my applications, and would like this 'feature'. Any objections @Dev-Khant ?

PR #2218, opened 2025-02-17
Added support of vision input

prateekchhikara:
## Description

Added support where users can now input images and we will create memories. Docs are also updated for the same.


pyproject.toml
```diff
@@ -25,6 +25,7 @@ sqlalchemy = "^2.0.31"
 langchain-community = "^0.3.1"
 neo4j = "^5.23.1"
 rank-bm25 = "^0.2.2"
+djangorestframework = "^3.15.2"
```
deshraj, line 28: Why do you need this?

mem0/memory/utils.py
```diff
@@ -1,10 +1,11 @@
 import re
-
 from mem0.configs.prompts import FACT_RETRIEVAL_PROMPT
+from rest_framework.exceptions import APIException
```
deshraj, line 3: You can raise instead of using django rest framework here.

docs/features/multimodal-support.mdx
```diff
@@ -0,0 +1,114 @@
...
+      "id": "56545065-7dee-4acf-8bf2-a5b2535aabb3"
+    }
+  ]
+}
+```
+</CodeGroup>
+
+## Image Integration Methods
+
+Mem0 allows you to add images to user interactions through two primary methods: by providing an image URL or by using a Base64-encoded image. Below are examples demonstrating each approach.
```
Dev-Khant, line 62: Show the 2 approaches in form of bullet points, will be easy to understand

docs/open-source/multimodal-support.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+## How It Works
+
+When a user provides an image, Mem0 processes the image to extract textual information and relevant details, which are then added to the user's memory. This feature enhances the system's ability to understand and remember details based on visual inputs.
+
+<CodeGroup>
+```python Code
+from mem0 import Memory
+
+# Initialize the Memory with your API key
```
Dev-Khant, line 15:
```suggestion
```

mem0/memory/utils.py
```diff
@@ -43,3 +43,45 @@ def remove_code_blocks(content: str) -> str:
...
+        messages=[
+            {
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "Provide a description of the image and do not include any additional text."},
+                    {"type": "image_url", "image_url": {"url": image_url}}
+                ],
+            },
+        ],
+        max_tokens=100,
```
Dev-Khant, line 63: Should we harcode this value as `100` because we are accepting `max_tokens` as a param.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mem0ai"
-version = "0.1.49"
+version = "0.1.50"
```
Dev-Khant, line 3: This is outdated now. We need `0.1.51`.

Dev-Khant (commented):
We should mention in the docs that we use only OpenAI for now.

mem0/memory/utils.py
prateekchhikara: for now we want to experiment it with 100, later we will use the param

Dev-Khant (approved):

PR #2374, opened 2025-03-13
fix: slashes in destinationtype raise cypher error

nocyphr:
## Description
llm generated destinationtype with slashes (eg Software/Model)
This raises an error because neo4j cypher does not allow slashes in entity names. 
Extended the existing sanitizing method (replace spaces) to include replacing / with : and renamed it to _remove_invalid_chars_from_entities to reflect the more general functionality. 
using : in cyphr allows us to define more than one label (which is pretty much what the llm was trying to do)
Fixes # (issue)
I did not report the issue, I just went and fixed it. 

## Type of change

## How Has This Been Tested?
- [x] Unit Test

## Checklist:

- [x] My code follows the style guidelines of this project (precommit)
- [x ] I have performed a self-review of my own code (it's just one line and a rename, i considered refactoring it to just use re.sub instead and define more than one char but decided to skip it for now)

### DIFF
--- a/mem0/memory/graph_memory.py
+++ b/mem0/memory/graph_memory.py
@@ -12,11 +12,14 @@ try:
 except ImportError:
     raise ImportError("rank_bm25 is not installed. Please install it using pip install rank-bm25")
 
-from mem0.graphs.tools import (DELETE_MEMORY_STRUCT_TOOL_GRAPH,
-                               DELETE_MEMORY_TOOL_GRAPH,
-                               EXTRACT_ENTITIES_STRUCT_TOOL,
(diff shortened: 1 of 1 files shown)


nocyphr:
had to adjust the entity_type_map as well to avoid "/" from sneaking into the labels

nocyphr:
@taranjeet  @Dev-Khant  any way to get this reviewed? I hotfixed it in my deployment but I would prefer to not have to redo it on each docker redeployment :smile:

kartik-mem0:
Hey — good catch on the Cypher escaping issue. We've since landed a broader fix for Neo4j relationship name sanitization in #4154 that covers this case. Going to close this PR, but thanks for flagging the problem early on.

PR #2466, opened 2025-03-29
Store `user_id` in vectordb

Dev-Khant:
## Description

Store `user_id`  in vectordb

Fixes # (issue)

## Type of change


mem0/memory/setup.py
```diff
@@ -29,3 +30,27 @@ def get_user_id():
             return user_id
     except Exception:
         return "anonymous_user"
+
+
+def store_and_get_user_id_from_vector_store(vector_store):
+    """Store user_id in mem0_migrations collection."""
+    user_id = get_user_id()
+    try:
+        try:
```
whysosaket, line 39: Why do we need 2 try blocks? :)

mem0/memory/telemetry.py
```diff
@@ -50,14 +75,22 @@ def close(self):
...
 
-# Initialize AnonymousTelemetry
-telemetry = AnonymousTelemetry(
-    project_api_key="phc_hgJkUVJFYtmaJqrvf6CYN67TIQ8yhXAkWzUn9AMU4yX",
-    host="https://us.i.posthog.com",
-)
-
-
 def capture_event(event_name, memory_instance, additional_data=None):
+    global telemetry
```
whysosaket, line 79: Why do we need this?

mem0/memory/setup.py
Dev-Khant: Ohh yaa will fix this

mem0/memory/telemetry.py
Dev-Khant: Because the telemery is also getting called in another function: `capture_client_event`

mem0/memory/setup.py
```diff
@@ -3,6 +3,7 @@
 import uuid
 
 # Set up the directory path
+VECTOR_ID = "fd411bd3-99a2-42d6-acd7-9fca8ad09580"
```
prateekchhikara, line 6: what does this ID means?

mem0/configs/vector_stores/azure_ai_search.py
```diff
@@ -7,7 +7,7 @@ class AzureAISearchConfig(BaseModel):
     collection_name: str = Field("mem0", description="Name of the collection")
     service_name: str = Field(None, description="Azure AI Search service name")
     api_key: str = Field(None, description="API key for the Azure AI Search service")
-    embedding_model_dims: int = Field(None, description="Dimension of the embedding vector")
+    embedding_model_dims: int = Field(1536, description="Dimension of the embedding vector")
```
prateekchhikara, line 10: are we using a fixed dimension embedding model

Dev-Khant: Yes we set `1536` by default because our default embedding model is `text-embedding-3-small`

prateekchhikara (approved):

mem0/memory/setup.py
```diff
@@ -29,3 +30,29 @@ def get_user_id():
             return user_id
     except Exception:
         return "anonymous_user"
+
+
+def store_and_get_user_id_from_vector_store(vector_store):
```
deshraj, line 35: Call this `get_or_create_user_id()`

mem0/memory/telemetry.py
deshraj: Is this the best way to handle this?

mem0/memory/telemetry.py
```diff
@@ -50,14 +75,22 @@ def close(self):
...
+
+    # For OSS, we use the telemetry vector store to store the user_id
+    if hasattr(memory_instance, "telemetry_vector_store"):
```
deshraj, line 82: Can we convert this to a single line if else?

deshraj (changes requested):
Left some comments.

mem0/memory/telemetry.py
Dev-Khant: Yes previously as we did the same thing

mem0/memory/main.py
```diff
@@ -51,6 +53,15 @@ def __init__(self, config: MemoryConfig = MemoryConfig()):
             self.graph = MemoryGraph(self.config)
             self.enable_graph = True
 
+        self.config.vector_store.config.collection_name = "mem0_migrations"
```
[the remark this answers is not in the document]
Dev-Khant: No because if you see above then we already intialize the `self.vector_store` which is based on the config user has passed and then we create a new one with different collection name `self.telemetry_vector_store`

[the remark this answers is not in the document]
deshraj: Got it. You probably want to store it as a private attribute since this is not going to be exposed in python. See https://martinxpn.medium.com/mastering-private-and-protected-fields-in-python-classes-a-complete-tutorial-44-100-days-of-c4a2017e3012

mem0/memory/telemetry.py
```diff
@@ -50,14 +50,12 @@ def close(self):
...
-# Initialize AnonymousTelemetry
-telemetry = AnonymousTelemetry(
-    project_api_key="phc_hgJkUVJFYtmaJqrvf6CYN67TIQ8yhXAkWzUn9AMU4yX",
-    host="https://us.i.posthog.com",
-)
+def capture_event(event_name, memory_instance, additional_data=None):
+    global telemetry
 
+    # For OSS, we use the telemetry vector store to store the user_id
+    telemetry = AnonymousTelemetry(project_api_key="phc_hgJkUVJFYtmaJqrvf6CYN67TIQ8yhXAkWzUn9AMU4yX", host="https://us.i.posthog.com", vector_store=memory_instance.telemetry_vector_store if hasattr(memory_instance, "telemetry_vector_store") else None)
```
deshraj, line 57: Can we do this check based on the package version instead?

mem0/memory/main.py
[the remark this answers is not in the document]
Dev-Khant: Done

mem0/memory/telemetry.py
Dev-Khant: Why package version didn't get it

deshraj (approved):

PR #2639, opened 2025-05-06
Added New Param, output_format

whysosaket:
Added New Param, output_format

### DIFF
--- a/mem0-ts/src/client/mem0.types.ts
+++ b/mem0-ts/src/client/mem0.types.ts
@@ -22,10 +22,16 @@ export interface MemoryOptions {
   custom_categories?: custom_categories[];
   custom_instructions?: string;
   timestamp?: number;
+  output_format?: string | OutputFormat;
 }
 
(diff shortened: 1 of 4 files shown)


mem0-ts/src/client/mem0.types.ts
```diff
@@ -22,12 +22,18 @@ export interface MemoryOptions {
...
+  output_format?: string | OutputFormat;
 }
 
 export interface ProjectOptions {
   fields?: string[];
 }
 
+export enum OutputFormat {
+  V1 = "v1.0",
+  V1_1 = "v1.1",
```
Dev-Khant, line 34: Is this similar to `v1.1`?

whysosaket: Yes this is the enum, same meaning

Dev-Khant: Got it. Thanks!

Dev-Khant (approved):

PR #2645, opened 2025-05-07
updates to type in client

prateekchhikara:
# Add proper type hinting for search method output formats

## Changes
- Added `@overload` decorators to handle different return types for the `search` method based on `output_format` parameter
- Updated return type hints to accurately reflect v1.0 (list) and v1.1 (dict with results) formats
- Added docstring clarification for return types

## Why
Improves type safety and IDE support by properly typing the different return formats of the search method.

### DIFF
--- a/mem0/client/main.py
+++ b/mem0/client/main.py
@@ -3,7 +3,7 @@ import logging
 import os
 import warnings
 from functools import wraps
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union, overload, Literal
 
 import httpx
 import requests
@@ -215,8 +215,28 @@ class MemoryClient:
         )
         return response.json()
 
+    @overload
+    def search(
+        self,
+        query: str,
+        version: str = "v1",
+        output_format: Literal["v1.0"] = "v1.0",
+        **kwargs
+    ) -> List[Dict[str, Any]]:
+        ...
+
+    @overload
+    def search(
+        self,
+        query: str,
+        version: str = "v1",
+        output_format: Literal["v1.1"] = "v1.1",
+        **kwargs
+    ) -> Dict[str, List[Dict[str, Any]]]:
+        ...
+
     @api_error_handler
-    def search(self, query: str, version: str = "v1", **kwargs) -> List[Dict[str, Any]]:
+    def search(self, query: str, version: str = "v1", **kwargs) -> Union[List[Dict[str, Any]], Dict[str, List[Dict[str, Any]]]]:
         """Search memories based on a query.
 
         Args:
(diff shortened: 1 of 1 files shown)


mem0/client/main.py
```diff
@@ -215,8 +215,28 @@ def get_all(self, version: str = "v1", **kwargs) -> List[Dict[str, Any]]:
...
         return response.json()
 
+    @overload
+    def search(
+        self,
+        query: str,
+        version: str = "v1",
+        output_format: Literal["v1.0"] = "v1.0",
+        **kwargs
+    ) -> List[Dict[str, Any]]:
```
johnwlockwood, line 225: https://docs.python.org/3/library/typing.html#deprecated-aliases
using `List` and `Dict` for type hinting is deprecated since Python 3.9 and this project specifies Python 3.9 and above, so it would be safe to start switching to using builtin types. This would look like `list[dict[str, Any]`
`Union`, used below, can be done with a `|` as in `Union[str, None]` -> `str | None`.

prateekchhikara: thanks for your feedback @johnwlockwood, I have made the changes

johnwlockwood: It turns out that `|` was introduced for type hints in Python 3.10, Union and Optional are needed for Python 3.9 compatibility. `list` and `dict` can be used for Python 3.9. I've got to make a change to one of my PRs.

mem0/client/main.py
```diff
@@ -215,8 +215,28 @@ def get_all(self, version: str = "v1", **kwargs) -> List[Dict[str, Any]]:
...
         return response.json()
 
+    @overload
+    def search(
+        self,
+        query: str,
+        version: str = "v1",
+        output_format: Literal["v1.0"] = "v1.0",
+        **kwargs
+    ) -> list[dict[str, Any]]:
```
johnwlockwood, line 225: I checked out this branch to see how the overload affects the IDE.
in Cursor or Visual Studio Code, it only shows me the signature of this first one. Does it work any better for your setup?
Screenshot 2025-05-10 at 11 26 06 PM

mem0/client/main.py
```diff
@@ -215,8 +215,28 @@ def get_all(self, version: str = "v1", **kwargs) -> List[Dict[str, Any]]:
...
+        query: str,
+        version: str = "v1",
+        output_format: Literal["v1.1"] = "v1.1",
+        **kwargs
+    ) -> dict[str, list[dict[str, Any]]]:
+        ...
+
     @api_error_handler
-    def search(self, query: str, version: str = "v1", **kwargs) -> List[Dict[str, Any]]:
+    def search(self, query: str, version: str = "v1", **kwargs) -> list[dict[str, Any]] | dict[str, list[dict[str, Any]]]:
```
johnwlockwood, line 239: The type annotations don't do any runtime checking of the params, so if you wanted to let a user know if they had a mistake when using search, you could add `output_format: Literal["v1.0", "v1.1"] = "v1.0"` to the signature, then you could validate it at runtime using a validation function like this:

```
SEARCH_OUTPUT_FORMATS: dict[str, TypeAlias] = {
    "v1.0": list[dict[str, Any]],  # Format "v1.0" returns a list of dictionaries
    "v1.1": dict[str, list[dict[str, Any]]],  # Format "v1.1" returns a dictionary with a "results" key
}

def validate_search_output_format(output_format: str) -> None:
    if output_format not in SEARCH_OUTPUT_FORMATS:
        raise ValueError(f"Invalid output_format: {output_format}. Supported formats are: {list(SEARCH_OUTPUT_FORMATS.keys())}")
```

then put it into `kwargs`
```
kwargs["output_format"] = output_format`
```

johnwlockwood:
@prateekchhikara I made tests on the client #2664. among other things, it will verify changes are compatible with Python 3.9.

mem0/client/main.py
johnwlockwood: I notice that the ci only runs tests for Python 3.10 and 3.11. which is why this passed all the checks. Do you want to keep it 3.9 compatible, since the pyproject.toml specifies >=3.9?

Issue #1328, opened 2024-03-21
[qdrant] is not a valid point ID, valid values are either an unsigned in


JerryJiangjjw:
### 🐛 Describe the bug
* macbook pro m2pro
* embedchain    0.1.98
* Python 3.12.2

```python
import os

from embedchain import App

os.environ["OPENAI_API_KEY"] = "xxx"
os.environ["QDRANT_URL"] = "xxx"
os.environ["QDRANT_API_KEY"] = "xxx"

config = {
    "vectordb": {
        "provider": "qdrant",
        "config": {
            "collection_name": "my_qdrant_index"
        },
    }
}

app = App.from_config(config=config)
app.add("https://www.forbes.com/profile/elon-musk")


```

```
Adding data in batches:   0%|          | 0/1 [00:01<?, ?it/s]
Failed to add batch due to a bad request: Unexpected Response: 400 (Bad Request)
Raw response content:
b'{"status":{"error":"Format error in JSON body: value default-app-id--a2bfb0ef8be8b851f353cb52bc2cbd0b3bb649a7c4499c1404c8098b2c819157 is not a valid point ID, valid values are either an unsigned in ...'
Successfully saved https://www.forbes.com/profile/elon-musk (DataType.WEB_PAGE). New chunks count: 0

```

<img width="1369" alt="image" src="https://github.com/embedchain/embedchain/assets/16333372/4c15dafc-4d5f-4e67-b95e-1ab33a864e5a">

[ id  string uuid
Type, used for specifying point ID in user interface](https://qdrant.github.io/qdrant/redoc/index.html#tag/points/operation/upsert_points)

[chunk_id](https://github.com/embedchain/embedchain/blob/85ec61335a1e9e0b0f4313f5dd67242be08c0a5a/embedchain/chunkers/base_chunker.py#L55)

deshraj:
Closing since stale issue.

PR #2889, opened 2025-06-02
support qdrant in base chunker

nzlzcat:
Fixes https://github.com/mem0ai/mem0/issues/1328

```bash
Creating chunks for 1 data records with doc_id: default-app-id--15e44de2b0b327ef971a48d8846db7a81535f2b6845306a90c115dc7a70e8979
Adding data in batches:   0%|                                                           | 0/1 [00:00<?, ?it/s]
2025-06-02 11:08:41,625 - embedchain.embedchain - INFO - Failed to add batch due to a bad request: Unexpected Response: 400 (Bad Request)
Raw response content:
b'{"status":{"error":"Format error in JSON body: value default-app-id--15e44de2b0b327ef971a48d8846db7a81535f2b6845306a90c115dc7a70e8979 is not a valid point ID, valid values are either an unsigned in ...'
```

### DIFF
--- a/embedchain/embedchain/chunkers/base_chunker.py
+++ b/embedchain/embedchain/chunkers/base_chunker.py
@@ -1,5 +1,6 @@
 import hashlib
 import logging
+import uuid
 from typing import Any, Optional
 
 from embedchain.config.add_config import ChunkerConfig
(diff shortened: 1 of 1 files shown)


kartik-mem0:
This targets embedchain's base chunker which is legacy code, and the underlying issue #1328 is closed. Closing this out. Thanks for the contribution.

PR #3050, opened 2025-06-26
Fix: Gemini Embeddings and LLM

Dev-Khant:
## Description

Fix: Gemini Embeddings and LLM

Fixes # (issue)

## Type of change

### DIFF
--- a/mem0/embeddings/gemini.py
+++ b/mem0/embeddings/gemini.py
@@ -1,7 +1,8 @@
 import os
 from typing import Literal, Optional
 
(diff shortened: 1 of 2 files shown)


invincible04:
Please remove this, `self.config.output_dimensionality`, `BaseEmbedderConfig` class does not contain this attribute.

Also, please keep the API key name consistent. In embedding, we are using the `GEMINI_API_KEY`, and in the LLM we are using the `GOOGLE_API_KEY`.

PR #3064, opened 2025-06-27
MongoDB Vector Store misaligned strings and classes

kirex0:
- Fixed MongoDBs string in the vector store factory to match the name of the class. 
- Removed check_auth_and_connection in mongoDB config as its not needed anymore. 
- Added mongo_uri to the mongoDBConfig to replace the host, port, password, and user to match the mongoVector class. This also simplifies the logic and allows users to connect with just the mongo URI. 
- Updated to use pydantic V2 syntax on mongos validation. 
- Updated the test_mongo.py to match the new structure. 
- Fixed the search index model to use dimensions instead of d.

### DIFF
--- a/mem0/utils/factory.py
+++ b/mem0/utils/factory.py
@@ -80,7 +80,7 @@ class VectorStoreFactory:
         "upstash_vector": "mem0.vector_stores.upstash_vector.UpstashVector",
         "azure_ai_search": "mem0.vector_stores.azure_ai_search.AzureAISearch",
         "pinecone": "mem0.vector_stores.pinecone.PineconeDB",
-        "mongodb": "mem0.vector_stores.mongodb.MongoVector",
+        "mongodb": "mem0.vector_stores.mongodb.MongoDB",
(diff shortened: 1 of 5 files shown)


mem0/utils/factory.py
```diff
@@ -80,7 +80,7 @@ class VectorStoreFactory:
         "upstash_vector": "mem0.vector_stores.upstash_vector.UpstashVector",
         "azure_ai_search": "mem0.vector_stores.azure_ai_search.AzureAISearch",
         "pinecone": "mem0.vector_stores.pinecone.PineconeDB",
-        "mongodb": "mem0.vector_stores.mongodb.MongoDB",
+        "mongodb": "mem0.vector_stores.mongodb.MongoVector",
```
Dev-Khant: Hey, can you please change the name to `MongoDB` in the original vectorstore file?

kirex0: Yes! Just pushed the updated code!

Dev-Khant (approved):
Thanks @kirex0 the changes looks good.

PR #3116, opened 2025-07-08
Updated MongoDB.search to use vectors instead of query_vectors.

kirex0:
Updated the MongodDB vector store .search function to be aligned with the way its being called in main [here](https://github.com/mem0ai/mem0/blob/6866e56d7a86170db11b165506f712d1a3948fc1/mem0/memory/main.py#L352).


parshvadaftari:
@kirex0 hey this looks good to me. But the fialing tests have already been fixed with in the latest commits in the repo. Try rebase main and removing the commits for failing tests

PR #3202, opened 2025-07-21
fix(vertexai): allow max_batch_size to be set via config and passed t…

### LINKED ISSUE #3199: VertexAI Embeddings: max_batch_size parameter not properly passed through configuration system
## Problem Description
When using VertexAI embeddings  the `max_batch_size` parameter cannot be passed through to the underlying `VertexAIEmbeddings` constructor, causing batch size configuration to be ignored.

## Current Behavior
- Setting `max_batch_size=1` in the embedder configuration does not affect the actual batch size used
- The default `_MAX_BATCH_SIZE = 250` is always used regardless of configuration
- This causes issues when trying to reduce batch size for debugging or specific use cases

## Expected Behavior
- The `max_batch_size` parameter should be properly passed through the configuration chain
- The `VertexAIEmbeddings` constructor should respect the configured batch size
- Configuration should override the default `_MAX_BATCH_SIZE = 250`

jdhruv555:
This fixes the issue where the batch size configuration was ignored and the default was always used, regardless of user input.

 Context:
Previously, setting `max_batch_size` in the embedder configuration had no effect, causing issues for users who needed to control batch size for debugging or specific use cases. This change ensures that the configuration is respected and passed through the code chain.

Dependencies:
No new dependencies are required for this change.

Fixes #3199 

## Type of change

## How Has This Been Tested?

- Unit Test:  
  Added a test to verify that when `max_batch_size` is set in the config, it is passed to the `VertexAIEmbeddings` constructor.  
  To reproduce:  
  1. Run the test suite, especially `embedchain/tests/test_factory.py`.
  2. The test checks that the correct parameter is passed via the factory.

## Checklist:

- [✓] My code follows the style guidelines of this project
- [✓] I have performed a self-review of my own code
- [✓] I have commented my code, particularly in hard-to-understand areas
- [✓] I have made corresponding changes to the documentation (if applicable)
- [✓] My changes generate no new warnings
- [✓] I have added tests that prove my fix is effective or that my feature works
- [✓] New and existing unit tests pass locally with my changes
- [✓] I have checked my code and corrected any misspellings

## Maintainer Checklist

- [] closes #3199
- [] Made sure Checks passed

### DIFF
--- a/embedchain/embedchain/config/embedder/base.py
+++ b/embedchain/embedchain/config/embedder/base.py
@@ -18,6 +18,7 @@ class BaseEmbedderConfig:
         model_kwargs: Optional[Dict[str, Any]] = None,
         http_client_proxies: Optional[Union[Dict, str]] = None,
         http_async_client_proxies: Optional[Union[Dict, str]] = None,
+        max_batch_size: Optional[int] = None,
     ):
         """
         Initialize a new instance of an embedder config class.
@@ -53,3 +54,4 @@ class BaseEmbedderConfig:
         self.http_async_client = (
             httpx.AsyncClient(proxies=http_async_client_proxies) if http_async_client_proxies else None
         )
+        self.max_batch_size = max_batch_size
--- a/embedchain/embedchain/embedder/vertexai.py
+++ b/embedchain/embedchain/embedder/vertexai.py
@@ -11,7 +11,10 @@ class VertexAIEmbedder(BaseEmbedder):
     def __init__(self, config: Optional[BaseEmbedderConfig] = None):
         super().__init__(config=config)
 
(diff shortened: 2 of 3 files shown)


parshvadaftari:
@jdhruv555 Can you update the docs for it as well?

parshvadaftari:
@jdhruv555 Can you check for the failing tests?

utkarsh240799:
Hi @jdhruv555, thanks for the contribution! We've closed the underlying issue (#3199) as won't-fix since the `embedchain` package is no longer actively developed. The core `mem0` package has its own VertexAI implementation that doesn't have this limitation. Closing this PR accordingly — feel free to check out the `mem0` package if you'd like to contribute!

PR #3362, opened 2025-08-25
fix: correct memories array access in LangChain integration example

wynhfy:
# Fix memories array access in LangChain integration example

## Description
Fixed a bug in the LangChain integration documentation where the code incorrectly treated `memories` as a dictionary with a `get` method, when it should be treated as an array.

## Changes Made
- **File**: `docs/integrations/langchain.mdx`
- **Line**: 65
- **Before**: `serialized_memories = ' '.join([mem["memory"] for mem in memories.get('results', [])])`
- **After**: `serialized_memories = ' '.join([mem["memory"] for mem in memories])`

## Problem
The original code assumed `memories` was a dictionary with a `get` method, but according to the context, `memories` is actually an array returned from `mem0.search()`. This would cause a runtime error since arrays don't have a `get` method.

## Solution
Removed the unnecessary `.get('results', [])` call and directly iterate over the `memories` array.

## Testing
- [x] Code syntax is correct
- [x] Logic flow remains the same
- [x] Documentation example will now work without errors

## Impact
This fix ensures that the LangChain integration example in the documentation will work correctly without throwing runtime errors, providing users with a working reference implementation.

### DIFF
--- a/docs/integrations/langchain.mdx
+++ b/docs/integrations/langchain.mdx
@@ -62,7 +62,7 @@ Create functions to handle context retrieval, response generation, and addition 
 def retrieve_context(query: str, user_id: str) -> List[Dict]:
     """Retrieve relevant context from Mem0"""
     memories = mem0.search(query, user_id=user_id)
-    serialized_memories = ' '.join([mem["memory"] for mem in memories.get('results', [])])
+    serialized_memories = ' '.join([mem["memory"] for mem in memories])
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for the LangChain docs fix. This has been open for 7 months without getting reviewed. Closing it out — sorry it fell through the cracks.

PR #3373, opened 2025-08-26
Refactor PGVector to Use Internal Connection Pools and Context Managers

Sheharyar570:
## Description
Previously, the PGVector class accepted a connection_pool or a direct connection as an argument to its constructor. This meant a single connection could be used for all transactions in a PGVector instance, leading to idle transactions and connections not being returned to the pool promptly. As a result, some search queries would remain in an idle state, unnecessarily keeping database connections alive and potentially exhausting available connections.

**Previous Implementation**
The constructor previously allowed passing a connection_pool, connection_string, or individual connection parameters, and would create and hold a single connection (self.conn) and cursor (self.cur) for the lifetime of the object. This led to the problem of idle transactions and connection leaks.

**Changes in This PR**

PGVector Class (`pgvector.py`)

**No More Direct Connections or External Pools:**
The class no longer accepts a connection_pool or creates a persistent connection. Instead, it creates its own connection pool (ConnectionPool for psycopg3, ThreadedConnectionPool for psycopg2) inside the constructor.
The connect logic and connection_pool argument have been removed. Added 2 new attributes `minconn` and `maxconn` to be used in pool initialization.

**Context Manager for Connections:**
All database operations now use a _get_cursor context manager, which fetches a connection from the pool for each operation and ensures it is returned to the pool after use.
This prevents idle transactions and ensures connections are always available for new operations.

**psycopg3/psycopg2 Compatibility:**
The code detects which driver is available and uses the appropriate pool and cursor logic.
For psycopg3, uses executemany() for batch inserts; for psycopg2, uses execute_values().

**Improved Error Handling:**
The context manager handles commit/rollback and logs errors, ensuring robust transaction management.

**No More Persistent Cursors:**
Removed persistent self.conn and self.cur attributes; all operations are now scoped to the context manager.

**Pool Cleanup:**
The destructor (__del__) ensures the pool is properly closed when the object is deleted.

**Test Class (`test_pgvector.py`)**

**Updated Mocks and Patches:**
Tests now patch the correct pool and cursor objects for both psycopg3 and psycopg2.
Mocks are updated to reflect the new context manager usage, ensuring connections and cursors are properly simulated and returned.

**Explicit Version Testing:**
Tests use the @patch decorator to explicitly test both psycopg3 and psycopg2 code paths.

**Error Handling and Rollback Tests:**
Added or updated tests to verify that transactions are rolled back on error and that connections are returned to the pool.

**Insert Logic Tests:**
Ensured that executemany() is used for psycopg3 and execute_values() for psycopg2 in insert tests.

**No More Direct Connection Passing:**
Tests no longer pass a connection or pool to the constructor, matching the new internal pool management.

**Summary**

Fixes idle transaction and connection leak issues by ensuring each operation uses a fresh connection from the pool and returns it immediately after use.
Removes the need to pass a connection or pool to the constructor; all pool management is now internal to PGVector.
Improves compatibility with both psycopg3 and psycopg2, with correct handling for batch inserts.
Enhances test coverage for connection management, error handling, and compatibility.
Refactors code for clarity, maintainability, and robustness.

Fixes # 3332

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [x] Unit Test

## Checklist:

- [ ] New and existing unit tests pass locally with my changes

## Maintainer Checklist

- [ ] Made sure Checks passed


parshvadaftari:
@Sheharyar570 Can you sign the CLA to get things started? Thanks!

Sheharyar570:
@parshvadaftari I've signed the CLA.

pyproject.toml
```diff
@@ -38,7 +38,7 @@ vector_stores = [
     "faiss-cpu>=1.7.4",
```
parshvadaftari: Try adding this as additional dependency and we are using `hatch` and not poetry. Please remove the poetry changes from the commit. Also there's no pyscopg[pool] in the pypi found, please make sure to use the correct packages.

poetry.lock
parshvadaftari, line 1: This file is not required and no changes are required for this file as we use `hatch`

mem0/configs/vector_stores/pgvector.py
```diff
@@ -11,19 +11,16 @@ class PGVectorConfig(BaseModel):
     password: Optional[str] = Field(None, description="Database password")
     host: Optional[str] = Field(None, description="Database host. Default is localhost")
     port: Optional[int] = Field(None, description="Database port. Default is 1536")
-    diskann: Optional[bool] = Field(True, description="Use diskann for approximate nearest neighbors search")
-    hnsw: Optional[bool] = Field(False, description="Use hnsw for faster search")
+    diskann: Optional[bool] = Field(False, description="Use diskann for approximate nearest neighbors search")
```
parshvadaftari, line 14: We already have `diskann` and `hnsw`. Seems your fork is out of sync with the current main.

mem0/configs/vector_stores/pgvector.py
```diff
@@ -11,19 +11,16 @@ class PGVectorConfig(BaseModel):
...
+    hnsw: Optional[bool] = Field(True, description="Use hnsw for faster search")
+    minconn: Optional[int] = Field(1, description="Minimum number of connections in the pool")
+    maxconn: Optional[int] = Field(5, description="Maximum number of connections in the pool")
     # New SSL and connection options
     sslmode: Optional[str] = Field(None, description="SSL mode for PostgreSQL connection (e.g., 'require', 'prefer', 'disable')")
     connection_string: Optional[str] = Field(None, description="PostgreSQL connection string (overrides individual connection parameters)")
```
parshvadaftari, line 20: Why is `connection_pool` removed from here?

parshvadaftari (commented):
Also please fix the failing tests. The tests seems to be outdated and doesn't align with the pgvector.

parshvadaftari:
@Sheharyar570 Please incorporate the requested changes.

pyproject.toml
Sheharyar570: psycopg[pool] is one way to install additional package that comes with psycopg3. But I understand your point, I'm linking the dependency that I'll add.

https://pypi.org/project/psycopg-pool/

poetry.lock
Sheharyar570: Sure, I'll remove poetry related changes.

pyproject.toml
parshvadaftari: the pypi is psycopg-pool and not psycopg[pool]. This will raise conflicts and tests will fail. Please make sure to have a correct package name along with the contrained version.

mem0/configs/vector_stores/pgvector.py
Sheharyar570: @parshvadaftari the connection pool is now created inside the PGVector class constructor, so connections are always retrieved from the pool rather than allowing single connections or passing pools from outside. This approach also accounts for differences between psycopg2 (pool.getconn()) and psycopg3 (context manager with pool.connection()), while giving better control over pool configuration within the class.

mem0/configs/vector_stores/pgvector.py
Sheharyar570: I only changed the default values for `hnsw` and `diskann`. Previously, `hnsw` was set to `False`, but since we didn’t have `pgvectorscale` installed (we were using the `pgvector` extension in Postgres), our index for the memory table wasn’t built. My assumption is that `hnsw` should be the default, as it aligns with the `pgvector` implementation, while support for `pgvectorscale` is optional.

Currently, the defaults are `diskann = False` and `hnsw = True`. I’m not sure of the original reasoning for setting `diskann` to `True`, but I’m open to suggestions or reverting this if needed.

PS: The PR is in sync with the current main branch.

pyproject.toml
Sheharyar570: Sure thing @parshvadaftari

mem0/configs/vector_stores/pgvector.py
parshvadaftari: What if a user doesn't want to use `connection pool` and instead wants to make a connection using another method. Have you made sure those connections gets established and no connection pool is being created there? This should be an optional parameter.

Sheharyar570: We intentionally create and manage a connection pool inside PGVector by default to avoid single-connection pitfalls in production—idle/long-running transactions, as there is no connection cleanup. In my tests, using a single shared connection caused search queries to enter an idle state in production when multiple workers were used, so that approach is unsafe; if a user truly wants a single connection they can set the pool size to 1 (minconn = maxconn = 1), which preserves pool semantics while matching single-connection behavior.

But if you want to give option for single conncection ( `psycopg.connect()`), I can add a boolean flag `use_connection_pools` with default value set to `True`.

If the user sets it to `False`, we can create a single connection. But this would make the code more complicated as we are using `context managers` to handle pool connections (commits and rollbacks).

pyproject.toml
Sheharyar570: I'll simply edit the pyproject.toml file and add this pypi `psycopg-pool` dependency under `vector_stores` group.

Sheharyar570:
@parshvadaftari I've fixed the failing test and poetry related changes.

mem0/configs/vector_stores/pgvector.py
Sheharyar570: @parshvadaftari I've added back the connection_pool field, I've also added 2 new unittest to test the functionality. None of the test are now failing.

image

parshvadaftari:
@Sheharyar570 Can you add minconn and maxconn default values? such that for the cases you don't have to explicitly mentioned them. Rest of it looks good.

Sheharyar570:
@parshvadaftari I've added the defaults for `minconn` and `maxconn` in pgvector class. Their default values are already set in `configs/pgvector.py`.

Issue #3425, opened 2025-09-06
Nodejs SDK: Default llm config is used for graph store instead of user's root llm config


ranjithkumar8352:
### 🐛 Describe the bug

If the user has provided a config with llm in the main config then that should be used for graph store but instead the default open ai config is used.

The user provided llm config works for graph store only if it is provided in the `graphStore` config.

PR #3426, opened 2025-09-06
fix: 🔧 default llm config being used for graph store instead of root llm config provided by the user

ranjithkumar8352:
## Description

If the user has provided a config with llm in the main config then that should be used for graph store but instead the default open ai config is used.

The user-provided llm config works for graph store only if it is provided in the graphStore config.

Fixes https://github.com/mem0ai/mem0/issues/3425

This also includes a small change in `anthropic.ts` because the build fails without this change.

## Type of change

## How Has This Been Tested?
Tested locally using npm link

## Checklist:

## Maintainer Checklist

- [ ] Made sure Checks passed

### DIFF
--- a/mem0-ts/src/oss/src/config/defaults.ts
+++ b/mem0-ts/src/oss/src/config/defaults.ts
@@ -34,12 +34,6 @@ export const DEFAULT_MEMORY_CONFIG: MemoryConfig = {
       username: process.env.NEO4J_USERNAME || "neo4j",
       password: process.env.NEO4J_PASSWORD || "password",
     },
-    llm: {
-      provider: "openai",
-      config: {
-        model: "gpt-4-turbo-preview",
-      },
-    },
   },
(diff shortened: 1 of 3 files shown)


whysosaket:
Hey @ranjithkumar8352, have you tried and tested if this works? As far as I know, the reason we defaulted the config for Graph Store was because of the use of structured schema and the zod schema there.

ranjithkumar8352:
@whysosaket Yes, I have tested it on my backend server, and it works as expected because structured LLM is initialized in `src/memory/graph_memory.ts` and only the provider name is used for that.

image

utkarsh240799:
Hi @ranjithkumar8352, thanks for the fix\! We've opened #4466 which builds on your approach and also addresses the provider/config mismatch in `graph_memory.ts`, with regression tests. Closing in favor of #4466 — appreciate the contribution\!

PR #3459, opened 2025-09-12
feat(openmemory): make OpenAI API key optional in run.sh

### LINKED ISSUE #3445: Fix/Feature: Remove OpenAI key from being required in run.sh
# Required api key
Since openmemory is supposed to be able to be used by many different services so open ai should not be required.
```bash
# Line 8
OPENAI_API_KEY="${OPENAI_API_KEY:-}"
USER="${USER:-$(whoami)}"
NEXT_PUBLIC_API_URL="${NEXT_PUBLIC_API_URL:-http://localhost:8765}"

if [ -z "$OPENAI_API_KEY" ]; then
  echo "❌ OPENAI_API_KEY not set. Please run with: curl -sL https://raw.githubusercontent.com/mem0ai/mem0/main/openmemory/run.sh | OPENAI_API_KEY=your_api_key bash"
  echo "❌ OPENAI_API_KEY not set. You can also set it as global environment variable: export OPENAI_API_KEY=your_api_key"
  exit 1
fi
```
# Suggested Solution
I would recommend that it would allow different args for each service or don't require api key.

Beastmaster23:
- Remove exit condition when OPENAI_API_KEY is not set
- Allow run.sh to continue with warning messages instead of failing and still be set through UI
- Maintain helpful instructions for setting the API key via environment variable or UI settings

## Description

This change makes the initial setup process more flexible by allowing users to
start OpenMemory without immediately requiring an OpenAI API key, which can
be configured later through the application interface.

Fixes #3445 

## Type of change

## Checklist:

## Maintainer Checklist

- [ ] closes #3445
- [ ] Made sure Checks passed

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1076,9 +1076,14 @@
 								"run_id": {"type": "string"},
 								"created_at": {"type": "string", "format": "date-time"},
 								"updated_at": {"type": "string", "format": "date-time"},
-								"categories": {"type": "array", "items": {"type": "string"}},
+								"categories": {"type": "object", "properties": {
+									"in": {"type": "array", "items": {"type": "string"}}
+								}},
 								"metadata": {"type": "object"},
-								"keywords": {"type": "string"}
+								"keywords": {"type": "object", "properties": {
+									"contains": {"type": "string"},
+									"icontains": {"type": "string"}
+								}}
 							},
 							"additionalProperties": {
 								"type": "object",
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Hi, good call on making the OpenAI key optional — that was a common pain point. We're winding down the OpenMemory components in this monorepo though, so this won't be merged. Thanks for the contribution\!

PR #3488, opened 2025-09-21
docs: Add Gemini example to Python quickstart

Virtuoso633:
Addresses a common friction point for new users trying to use non-default models. The current quickstart implies `Memory()` is sufficient, but this only works for OpenAI.

This PR adds a new tab to the initialization section with a clear, working example for using Google Gemini. It demonstrates how to configure the `llm`, `embedder`, and `vector_store` with the correct dimensions, which should prevent common `ImportError` and `ValueError` issues for new contributors.

### DIFF
--- a/docs/open-source/python-quickstart.mdx
+++ b/docs/open-source/python-quickstart.mdx
@@ -44,7 +44,7 @@ config = {
     "llm": {
         "provider": "gemini",
         "config": {
-            "model": "gemini-1.5-flash",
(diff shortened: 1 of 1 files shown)


parshvadaftari:
Hey @Virtuoso633 closing this PR. We already have this in the docs. The quickstart is something to start using the mem0. By default it will use OpenAI.

https://docs.mem0.ai/components/llms/models/google_AI

Issue #3419, opened 2025-09-04
mem0.get_all() failed: 'tuple' object has no attribute 'id'


ArnabhS:
### 🐛 Describe the bug

In Mem0 Open source when configured with MongoDB database. The get_all() function is returning an error. 
mem0.get_all() failed: 'tuple' object has no attribute 'id'.

parshvadaftari:
@ArnabhS what's the version which you're using and can you share me code snippet??

parshvadaftari:
Hey @ArnabhS this issue was fixed with the latest releases, closing the issue. Thanks again for reaching out!

aesher9o1:
hi @parshvadaftari i am still facing this issue for MongoDB  #3596  fixes it

parshvadaftari:
Hey @aesher9o1 thank you for pointing it out, will review the solution.

PR #3489, opened 2025-09-21
Fix: Handle tuple results in get_all method (Issue #3419)

pxkundu:
This PR fixes the issue where tuple results were not being handled properly in the get_all method. The fix ensures that tuple results are correctly processed and returned as expected.


parshvadaftari:
Hey @pxkundu Can you please make sure to resolve conflicts and also it seems the change which you were gonna make has already been implemented.

parshvadaftari:
Closing this PR as the bug is fixed and the solution is implemented. Thank you again for implementation.

PR #3610, opened 2025-10-17
Fix memory deletion not removing from vector store

frederikb96:
## Problem

Memory deletion endpoints only updated PostgreSQL state but never deleted from Qdrant vector store. This caused:
- Deleted memories still appearing in search results
- Vector store growing indefinitely
- Database and vector store getting out of sync

**What was broken:**
1. **REST API** (`DELETE /api/v1/memories/`): Only called `update_memory_state()` to mark as deleted in DB
2. **MCP API** (`delete_all_memories`): Called `memory_client.delete(memory_id)` but passed UUID instead of string

## How It Works Elsewhere

The `add_memories` flow (both REST and MCP) properly syncs:
1. Add to vector store via `memory_client.add()`
2. Store metadata in PostgreSQL

Deletion should mirror this - remove from both stores.

## The Fix

**MCP API** (`delete_all_memories` in mcp_server.py):
- Convert UUID to string: `memory_client.delete(str(memory_id))`
- Qdrant expects string IDs, not UUID objects

**REST API** (`delete_memories` in memories.py):
- Added vector store deletion before DB update:
  1. Get memory client
  2. Delete from Qdrant via `memory_client.delete(str(memory_id))`
  3. Mark as deleted in PostgreSQL

Both endpoints now properly sync deletions across storage layers.

## Type of change

## How Has This Been Tested?

- [x] Test Script - verified memories deleted from both Qdrant and PostgreSQL
- Tested delete_all via MCP and bulk delete via REST API
- Confirmed deleted memories no longer appear in search results

## Checklist:

- [x] Follows existing pattern from add_memories workflow

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -314,7 +314,7 @@ async def delete_all_memories() -> str:
             # delete the accessible memories only
             for memory_id in accessible_memory_ids:
                 try:
-                    memory_client.delete(memory_id)
+                    memory_client.delete(str(memory_id))
                 except Exception as delete_error:
                     logging.warning(f"Failed to delete memory {memory_id} from vector store: {delete_error}")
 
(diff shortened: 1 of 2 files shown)


parshvadaftari:
Hey @frederikb96 Good catch! Thanks a lot for this fix.

CC: @VikramIyer125 Can you review this?

PR #3510, opened 2025-09-27
Feature/bug fix of thinking llm in vllm

dog-last:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

Please delete options that are not relevant.

- [x] Unit Test

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

## Maintainer Checklist

- [ ] closes #xxxx (Replace xxxx with the GitHub issue number)
- [ ] Made sure Checks passed

PR #3626, opened 2025-10-19
feat: add thinking mode support to fact retrieval messages

### LINKED ISSUE #3564: Getting Error in new_retrieved_facts: Expecting value: line 1 column 1 (char 0)
mem0ai                               1.0.0b0

Vedant817:
## Description
This PR fixes the JSON parsing error in `new_retrieved_facts` reported in issue #3564.

**Root Cause:**
The issue occurs with models like Qwen3-8B (as reported in #3564) that in reasoning/thinking mode return non-JSON tokens before the actual JSON response. These additional tokens cause the `json.loads()` call to fail with "Expecting value: line 1 column 1 (char 0)".

**Solution:**
This PR adds a configurable `thinking_mode` parameter that handles reasoning mode outputs by instructing models to include thinking in their structured JSON response, preventing JSON parsing failures.

**Technical Changes:**
- Added `enable_thinking_mode` configuration parameter
- Modified `get_fact_retrieval_messages()` to conditionally prepend thinking instruction
- Updated unit tests to verify thinking mode functionality

Fixes #3564


mem0/configs/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Initialize configs package and expose settings."""
```
parshvadaftari: please remove this.

mem0/configs/settings.py
```diff
@@ -0,0 +1,44 @@
+"""
+Settings module for mem0 configuration management.
```
parshvadaftari: Can you please remove this? Nit picking here as this doesn't solve the issue but also will introduce new issues.

mem0/memory/main.py
```diff
@@ -712,7 +712,7 @@ def _get_all_from_vector_store(self, filters, limit):
         memories_result = self.vector_store.list(filters=filters, limit=limit)
         actual_memories = (
             memories_result[0]
-            if isinstance(memories_result, (tuple, list)) and len(memories_result) > 0
+            if isinstance(memories_result, (tuple)) and len(memories_result) > 0
```
parshvadaftari, line 715: revert this change or you can update the main branch of your fork, this will cause errors for other vdb providers.

mem0/memory/utils.py
```diff
@@ -2,26 +2,31 @@
...
-        
+
     Returns:
         tuple: (system_prompt, user_prompt)
     """
-    if is_agent_memory:
-        return AGENT_MEMORY_EXTRACTION_PROMPT, f"Input:\n{message}"
-    else:
-        return USER_MEMORY_EXTRACTION_PROMPT, f"Input:\n{message}"
+    from mem0.configs import settings
```
parshvadaftari, line 21: This is not how we are planning to introduce the config, there are set of parameters for every provider - ollama, lmstudio where thinking mode toggle can be configured, this is a prompt based mechanism and not a config base.

parshvadaftari (changes requested):

parshvadaftari:
@Vedant817 thank you for the effort but please re write the pr description the issue number seems to be misleading and keep it to the point, this will be easier for us as maintainers. Thanks!

Vedant817:
> @Vedant817 thank you for the effort but please re write the pr description the issue number seems to be misleading and keep it to the point, this will be easier for us as maintainers. Thanks!

Yes, I'll make the changes. I found I made some mistakes in the description.

Vedant817:
> @Vedant817 thank you for the effort but please re write the pr description the issue number seems to be misleading and keep it to the point, this will be easier for us as maintainers. Thanks!

Hey @parshvadaftari I have resolved the asked changes and updated the PR description, please review it.

mem0/memory/utils.py
parshvadaftari: @Vedant817 please look at the comments here, this seems to be unresolved.

parshvadaftari (changes requested):

kartik-mem0:
Closing as the core issue has already been fixed in commit 51ce6f13 (PR #3510). The thinking tag stripping is now handled in `remove_code_blocks()` across all parsing paths.

PR #3659, opened 2025-10-24
Fix the configuration parameter of baidu_vector_store

JanMCHEN:
## Description

fix baidu_vector_store:
- BaiduDBconfig used `table_name` while Memory class need `collection_name`
- Hardcode replicate=3 caused failures in resource-constrained environments
- Invalid filter formats triggered "bad parameter filter" exceptions

## Type of change

## How Has This Been Tested?
- [ ] Test Script (please provide)

```python
from mem0 import Memory

Memory.from_config({
    "vectore_store": {
        "collection_name": "xxx",
        # "not table_name": ""
        ...
        
    }
})

```

## Maintainer Checklist

- [ ] Made sure Checks passed


kartik-mem0:
Thanks for the Baidu vector store config fix. This has been marked stale with no review. Closing it — if the fix is still needed, a fresh PR would be welcome.

PR #3661, opened 2025-10-25
ci: add install-only workflow for manual dependency install checks

lucifer4330k:
Adds a small  GitHub Actions workflow that can be manually dispatched (or triggered on PR) to run the project dependency installation step for Python 3.10 and 3.11. This helps maintainers re-run the pip resolver quickly to debug 'resolution-too-deep' issues without running the full CI matrix. Workflow: 
How to use: go to the Actions tab or the workflow on the PR and click 'Run workflow' (choose a python version) to re-run the dependency install.This PR targets the parent repo so CI can be re-run on the branch; it's safe and only installs packages (no tests or deployments).

### DIFF
--- a/.github/workflows/install-only.yml
+++ b/.github/workflows/install-only.yml
@@ -1,0 +1,42 @@
+name: Install-only CI
+
+on:
+  workflow_dispatch: {}
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for the install-only CI workflow idea. There wasn't maintainer interest in adding it. Closing this out.

PR #3721, opened 2025-11-06
feat(i18n): Enhance deletion instructions with reasons

nine-hao:
Updated promt to include reasons for deletions.


kartik-mem0:
Thanks for the i18n enhancement for deletion instructions. The CLA hasn't been signed and this has gone stale. Closing it out.

PR #3758, opened 2025-11-17
feat(openmemory): add OPENAI_BASE_URL to run.sh

stevendong:
This PR adds support for `OPENAI_BASE_URL` in the `openmemory/run.sh` script. This allows users to specify a custom base URL for the OpenAI API, for example, when using a proxy or a self-hosted solution.


kartik-mem0:
Thanks for adding the OPENAI_BASE_URL passthrough. We're winding down the OpenMemory components in this repo, so we won't be merging this. Appreciate the quick fix — closing for now.

PR #3775, opened 2025-11-20
Add Azure Cosmos DB No SQL Vector Search integration

allenkim0129:
## Description

Add Azure Cosmos DB No SQL Vector Search integration and tests

## Type of change
- New class `AzureCosmosDBNoSql` was added. This class implements all required methods from VectorStoreBase, so users can easily CRUD Vector Search data to Cosmos DB, and search from the data.
- Supports search types:
  - `vector`
  - `vector_score_threshold`
  - `full_text_search`
  - `full_text_ranking`
  - `hybrid`
  - `hybrid_score_threshold`
- For Vector or Hybrid search, it requires `vector_search_fields` during initialization
- For Full Text or Hybrid search, it requires `full_text_policy` and `full_text_search_enabled=True` during initialization
- Test file, `test_azure_cosmos_db_no_sql.py`, can be a good source to check how each method works and what arguments were needed to run different searches.

## How Has This Been Tested?

Mock tests were added along with this code changes. The tests verify all new functionality with Azure Cosmos DB No SQL Vector Search.

- [x] Unit Test

## Checklist:


allenkim0129:
@Dev-Khant Could you please review this PR for me? This is the Vector search integration PR for Azure Cosmos DB No SQL. 

Please let me know if I am missing any required documentation or configurations.

allenkim0129:
@Dev-Khant Just a reminder that this PR needs to be reviewed.

Thanks!

allenkim0129:
I am tagging more people to get more attentions to this PR

@Dev-Khant @deshraj @parthshr370 @kk2211 @VikramIyer125 @parshvadaftari

allenkim0129:
Tagging more reviewers:

@whysosaket @MgeeeeK @chaithanyak42

mem0/configs/vector_stores/azure_cosmos_db.py
```diff
@@ -0,0 +1,35 @@
...
+    cosmos_collection_properties: str = Field(None, description="Container Properties for the collection")
+    vector_search_fields: str = Field(None, description="Vector Search and Text Search Fields for the collection")
+    database_name: str = Field("vectorSearchDB", description="Name of the database to be created")
+    collection_name: str = Field("vectorSearchContainer", description="Name of the collection to be created")
+    search_type: str = Field("vector", description="CosmosDB Search Type to be performed")
+    metadata_key: str = Field("metadata", description="Metadata key to use for data schema")
+    create_collection: str = Field(True, description="Set to true if the collection does not exist")
+    table_alias: str = Field("c", description="Alias for the table to use in the WHERE clause")
+    full_text_policy: str = Field(None, description="Full Text Policy for the collection")
+    full_text_search_enabled: str = Field(False, description="Set to true if the full text search is enabled")
```
Siddhant-K-code: The `AzureCosmosDBConfig` class has incorrect field types that don't match the implementation:



```suggestion
    cosmos_client: Any = Field(None, description="Client used to connect to azure cosmosdb no sql account")
    vector_embedding_policy: Optional[Dict[str, Any]] = Field(None, description="Vector Embedding Policy for the collection")
    indexing_policy: Optional[Dict[str, Any]] = Field(None, description="Indexing Policy for the collection")
    cosmos_database_properties: Optional[Dict[str, Any]] = Field(default_factory=dict, description="Database Properties for the collection")
    cosmos_collection_properties: Optional[Dict[str, Any]] = Field(None, description="Container Properties for the collection")
    vector_search_fields: Optional[Dict[str, Any]] = Field(None, description="Vector Search and Text Search Fields for the collection")
    database_name: str = Field("vectorSearchDB", description="Name of the database to be created")
    collection_name: str = Field("vectorSearchContainer", description="Name of the collection to be created")
    search_type: str = Field("vector", description="CosmosDB Search Type to be performed")
    metadata_key: str = Field("metadata", description="Metadata key to use for data schema")
    create_collection: bool = Field(True, description="Set to true if the collection does not exist")
    table_alias: str = Field("c", description="Alias for the table to use in the WHERE clause")
    full_text_policy: Optional[Dict[str, Any]] = Field(None, description="Full Text Policy for the collection")
    full_text_search_enabled: bool = Field(False, description="Set to true if the full text search is enabled")
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,805 @@
...
+        self._table_alias = table_alias
+        self._full_text_policy = full_text_policy
+        self._full_text_search_enabled = full_text_search_enabled
+
+        # If vector_search_fields provided, extract text and embedding fields
+        if vector_search_fields:
+            self._text_key = vector_search_fields.get(Constants.TEXT_FIELD)
+            self._embedding_key = vector_search_fields.get(Constants.EMBEDDING_FIELD)
+            self._vector_search_enabled = True
+
```
Siddhant-K-code: `_vector_search_enabled` is only set when `vector_search_fields` is provided, but `_validate_search_args` accesses it unconditionally, which will raise `AttributeError`.


```py
# Initialize defaults
self._vector_search_enabled = False
self._text_key = None
self._embedding_key = None
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,805 @@
...
+                )
+            )
+        return result
+
+    def reset(self) -> None:
+        """
+        Reset the collection by deleting and recreating it.
+        """
+        self.delete_col()
+        self.create_col()
```
Siddhant-K-code: `reset()` calls `create_col()` without the required `name` argument

```suggestion
        self.create_col(name=self._collection_name)
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,805 @@
...
+    from azure.cosmos import CosmosClient, DatabaseProxy, ContainerProxy, CosmosDict
+    from azure.cosmos.exceptions import CosmosResourceNotFoundError
+except ImportError as e:
+    raise ImportError(
+        "The 'azure-cosmos' library is required. Install it with: pip install azure-cosmos"
+    ) from e
+
+class Constants:
+    # User Agent for Cosmos DB No SQL client
+    USER_AGENT = "LangChain-CDBNoSql-VectorStore-Python"
```
Siddhant-K-code: It references to Langchain, but it should be reference to us 😄  

```suggestion
    USER_AGENT = "Mem0-CDBNoSql-VectorStore-Python"
```

docs/components/vectordbs/dbs/azure_cosmos_db.mdx
```diff
@@ -0,0 +1,86 @@
...
+import os
+from mem0 import Memory
+
+os.environ["OPENAI_API_KEY"] = "sk-xx"
+
+config = {
+    "vector_store": {
+        "provider": "azure_cosmos_db",
+        "config": {
+            "cosmos_client": "mem0_db",
```
Siddhant-K-code: The documentation shows `"cosmos_client": "mem0_db"` as a string, but it should be a `CosmosClient` instance. Consider updating the example:



```suggestion
            "cosmos_client": cosmos_client,
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,805 @@
...
+        """
+        Get information about the Cosmos DB collection.
+        """
+        return self._collection.read()
+
+    def list(
+        self,
+        filters: Optional[Dict[str, Any]] = None,
+        limit: Optional[int] = 100,
+    ) -> List[OutputData]:  # signature updated to match base (filters, limit)
```
Siddhant-K-code: The `filters` parameter is accepted but not used. Consider either implementing filter support or documenting this limitation.

Siddhant-K-code (commented):
Thanks for this contribution! 

The implementation is well-structured w/ good test coverage. I have some suggestions that should be addressed before merging:

docs/components/vectordbs/dbs/azure_cosmos_db.mdx
allenkim0129: Done

mem0/configs/vector_stores/azure_cosmos_db.py
allenkim0129: Done

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Done

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Done

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Thank you! I added logic to support filters.

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: You are right. The vector search_fields are required all the time. I made it non-optional, and removed `self._vector_search_enabled`.

allenkim0129:
@Siddhant-K-code Thank you for your comments. I addressed all your comments. Please let me know if you have any other comments.

Siddhant-K-code (commented):
Looking good overall! A few things to address:

---

**Method signatures don't match the base class**

Your `delete`, `update`, `get`, `search`, and `create_col` methods have different signatures than what `VectorStoreBase` defines. For example, the base class expects `delete(vector_id)` but yours requires `partition_key_value` as well. This will break when used through the common interface. Can you make `partition_key_value` optional somehow - maybe store a default during init or derive it from the document ID?

---

**Typo**

`_is_full_test_search_type` should be `_is_full_text_search_type`

---

**SQL injection risk in `list()`**

You're interpolating filter values directly into the query string here:
```python
conditions = [f"{self._table_alias}.{key}={value}" for key, value in filters.items()]
```

Use parameterized queries like you did in search().

---

**Config import will fail early**

The `CosmosClient` import at the top of the config file will raise an error as soon as the module loads, even for users who aren't using Cosmos DB. Move it inside the class or wrap it properly.

Also add `model_config = ConfigDict(arbitrary_types_allowed=True)` since you're using `CosmosClient` as a field type.

**Silent failure in `delete()`**

You catch `CosmosResourceNotFoundError` and just log it - the caller has no idea the delete failed. Either re-raise or return something to indicate what happened.

**Missing null check**

If someone passes vector_search_fields=None, the constructor will blow up when you call .get() on it. Add a check.

**Docs**

The example shows cosmos_client: "cosmos_client" which is just a string - can you show how to actually create the client?

Minor stuff: double space on line 231, and the PR description says `text_azure_cosmos_db_no_sql.py` but the file is `test_azure_cosmos_db_no_sql.py`.

allenkim0129:
> Looking good overall! A few things to address:
> 
> **Method signatures don't match the base class**
> 
> Your `delete`, `update`, `get`, `search`, and `create_col` methods have different signatures than what `VectorStoreBase` defines. For example, the base class expects `delete(vector_id)` but yours requires `partition_key_value` as well. This will break when used through the common interface. Can you make `partition_key_value` optional somehow - maybe store a default during init or derive it from the document ID?
> 
> **Typo**
> 
> `_is_full_test_search_type` should be `_is_full_text_search_type`
> 
> **SQL injection risk in `list()`**
> 
> You're interpolating filter values directly into the query string here:
> 
> ```python
> conditions = [f"{self._table_alias}.{key}={value}" for key, value in filters.items()]
> ```
> 
> Use parameterized queries like you did in search().
> 
> **Config import will fail early**
> 
> The `CosmosClient` import at the top of the config file will raise an error as soon as the module loads, even for users who aren't using Cosmos DB. Move it inside the class or wrap it properly.
> 
> Also add `model_config = ConfigDict(arbitrary_types_allowed=True)` since you're using `CosmosClient` as a field type.
> 
> **Silent failure in `delete()`**
> 
> You catch `CosmosResourceNotFoundError` and just log it - the caller has no idea the delete failed. Either re-raise or return something to indicate what happened.
> 
> **Missing null check**
> 
> If someone passes vector_search_fields=None, the constructor will blow up when you call .get() on it. Add a check.
> 
> **Docs**
> 
> The example shows cosmos_client: "cosmos_client" which is just a string - can you show how to actually create the client?
> 
> Minor stuff: double space on line 231, and the PR description says `text_azure_cosmos_db_no_sql.py` but the file is `test_azure_cosmos_db_no_sql.py`.

@Siddhant-K-code Thanks for the comments, and sorry for the delay.

I’ve addressed all of your feedback, and I hope the updates make things clearer.

For the `list()` method, I enabled filter to work with a dictionary of filters. I hope the updated documentation reflects this well.

For the `search()` method, I didn’t enable the query and filter arguments, since the other parameters are intended to be alternatives and should be sufficient. For now, the definition still includes those arguments, but I documented that they are skipped and explained how users can rely on the other options instead.

I’ve also updated the documentation and examples as you suggested. Please let me know if additional documentation or samples would be helpful.

allenkim0129:
@Siddhant-K-code It has been several weeks, since I commited the last update. Can you please review this PR to check if we could merge this update?

Siddhant-K-code:
Sorry @allenkim0129, i am no longer associated with this project. Other maintainers might be able to help here.

allenkim0129:
@chaithanyak42 @deshraj @prathameshagrawal @whysosaket 

Since Siddhant is no longer associated with this project. Can any of you could review this PR and help me merge this integration?

mem0/utils/factory.py
```diff
@@ -170,6 +170,7 @@ class VectorStoreFactory:
         "upstash_vector": "mem0.vector_stores.upstash_vector.UpstashVector",
         "azure_ai_search": "mem0.vector_stores.azure_ai_search.AzureAISearch",
         "azure_mysql": "mem0.vector_stores.azure_mysql.AzureMySQL",
+        "azure_cosmos_db": "mem0.vector_stores.azure_cosmos_db_no_sql.AzureCosmosDBNoSQL",
```
aayush3011: 🔴 **Blocking: Class name mismatch causes `AttributeError` at runtime**

The factory registers `AzureCosmosDBNoSQL` (uppercase `SQL`) but the actual class is defined as `AzureCosmosDBNoSql` (mixed-case `Sql`) at line 148 of `azure_cosmos_db_no_sql.py`.

Since `getattr` is case-sensitive, any user configuring `"provider": "azure_cosmos_db"` via `Memory.from_config()` will crash immediately with `AttributeError`. The tests bypass the factory by importing the class directly, so they don't catch this.

**Fix:** Change this to:
```python
"azure_cosmos_db": "mem0.vector_stores.azure_cosmos_db_no_sql.AzureCosmosDBNoSql",
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+        )
+
+    def _create_item_to_insert(self, vector: List[float], payload: Optional[Dict], id: str) -> Dict[str, Any]:
+        payload = payload or {}
+        return {
+            Constants.ID: id,
+            self._text_key: payload.get(self._text_key, ""),
+            self._vector_key: vector,
+            self._metadata_key: payload.get(self._metadata_key, {}),
+        }
```
aayush3011: 🔴 **Blocking: Payload schema incompatible with `Memory` class — all data silently dropped**

`_create_item_to_insert` extracts only `description`, `vector`, and `metadata` keys from the payload. But when `Memory` calls `insert()`, it passes flat dicts like:

```python
# mem0/memory/main.py ~line 1085
metadata["data"] = data
metadata["hash"] = hashlib.md5(data.encode()).hexdigest()
metadata["created_at"] = datetime.now(...).isoformat()
self.vector_store.insert(vectors=[embeddings], ids=[memory_id], payloads=[metadata])
# payload = {"data": "I love sci-fi", "hash": "abc123", "user_id": "alice", "created_at": "..."}
```

Since the payload has no `"description"` or `"metadata"` key, this method stores empty string and empty dict. **All actual content is silently dropped.** When `Memory` reads results back (`mem.payload.get("data", "")`), every memory appears empty.

Every other vector store (Qdrant, MongoDB, Cassandra, Azure AI Search) stores the payload **as-is** without restructuring.

**Fix:** Store the full payload as-is:
```python
def _create_item_to_insert(self, vector, payload, id):
    payload = payload or {}
    item = dict(payload)
    item[Constants.ID] = id
    item[self._vector_key] = vector
    return item
```

And update `_build_output_data_from_item` to return all non-system fields as the payload.

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+            limit=limit,
+            search_type=search_type,
+            vector=vectors,
+            full_text_rank_filter=full_text_rank_filter,
+            offset_limit=offset_limit,
+            projection_mapping=projection_mapping,
+            return_with_vectors=return_with_vectors,
+            where=where,
+            weights=weights,
+        )
```
aayush3011: 🔴 **Blocking: `filters` parameter silently ignored — breaks multi-tenant data isolation**

`search()` accepts `filters` (line 377) but never passes it to `_construct_query` here. The `_generate_where_clause` helper exists and is correctly wired into `list()` (line 879), but is not used in `search()`.

`Memory` relies on filters for data isolation:
```python
# mem0/memory/main.py
search_filters = {"user_id": "alice", "agent_id": "agent-1"}
self.vector_store.search(query=new_mem, vectors=embeddings, limit=5, filters=search_filters)
```

With `filters` silently ignored, searches return results across **all users**, breaking multi-tenant data isolation.

_Note: This was [flagged by @Siddhant-K-code](https://github.com/mem0ai/mem0/pull/3775#discussion_r2695007832). The reply was "I added logic to support filters" — but filter support was only added to `list()`, not `search()`._

**Fix:** Add `filters` as a parameter to `_construct_query` and use `_generate_where_clause` to build the WHERE clause for search queries as well.

docs/components/vectordbs/dbs/azure_cosmos_db.mdx
```diff
@@ -0,0 +1,96 @@
...
+KEY = os.getenv("KEY")
+
+os.environ["OPENAI_API_KEY"] = "sk-xx"
+
+cosmos_client = CosmosClient(HOST, KEY)
+config = {
+    "vector_store": {
+        "provider": cosmos_client,
+        "config": {
+            "cosmos_client": "cosmos_client",
```
aayush3011: 🔴 **Blocking: Documentation example has 3 errors**

**Error 1 (line 30):** `"provider": cosmos_client` passes the `CosmosClient` object as the provider. `VectorStoreConfig.provider` is typed as `str`. Should be:
```python
"provider": "azure_cosmos_db",
```

**Error 2 (line 32):** `"cosmos_client": "cosmos_client"` passes the string literal `"cosmos_client"` instead of the variable. Should be:
```python
"cosmos_client": cosmos_client,
```

**Error 3 (line 50):** `"embedding_field": "vector"` uses the wrong key name. The code uses `Constants.VECTOR_FIELD = "vector_field"`. Should be:
```python
"vector_field": "vector",
```

_Note: Error 2 was [flagged by @Siddhant-K-code](https://github.com/mem0ai/mem0/pull/3775#discussion_r2695002580). The author replied "Done" but the current code still has the string literal._

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+            # Merge existing data when partial update
+            if vector is None:
+                vector = existing.get(self._vector_key)
+            if payload is None:
+                payload = {
+                    self._text_key: existing.get(self._text_key, ""),
+                    self._metadata_key: existing.get(self._metadata_key, {})
+                }
+        item = self._create_item_to_insert(vector=vector, payload=payload, id=vector_id)
+        self._collection.upsert_item(body=item)
```
aayush3011, line 1053: 🟡 **Recommendation: `update()` silently creates broken document when target doesn't exist**

When `CosmosResourceNotFoundError` is caught, `existing` is set to `None` and the code falls through to `upsert_item` with potentially `vector=None` and `payload=None`, silently creating a new malformed document instead of surfacing an error.

**Fix:**
```python
except CosmosResourceNotFoundError:
    raise ValueError(f"Cannot update: item '{vector_id}' not found.")
```

aayush3011 (commented):
## Review Summary

Thanks for adding Cosmos DB NoSQL support! The `ParamMapping` query builder and the breadth of search types (vector, full-text, hybrid) are well-designed. A few issues need attention before this can work end-to-end through `Memory.from_config()`:

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🔴 Blocking | Class name mismatch → crash at import |
| 2 | 🔴 Blocking | Payload schema drops all mem0 data |
| 3 | 🔴 Blocking | `search()` ignores `filters` → no data isolation |
| 4 | 🔴 Blocking | 3 errors in quick-start documentation |
| 5 | 🟡 Recommendation | `update()` silently creates broken doc when target missing |
| 6 | 🟡 Recommendation | `list()` uses wrong `return_with_vectors` default |
| 7 | 🟡 Recommendation | Fragile `in` substring check instead of helper method |
| 8 | 🟡 Recommendation | Config field types don't match `None` defaults |
| 9 | 🟡 Recommendation | Unhandled `KeyError` on missing vector key |
| 10 | 🟢 Suggestion | Dead stub `_search()` method |
| 11 | 🟢 Suggestion | `query` param silently overwritten |
| 12 | 🟢 Suggestion | Unused variable + built-in `id` shadowing |
| 13 | 💬 Observation | "Cosmod" typo in docs |

The most critical path is **findings 1–3**: without these fixes, the integration will not function when used through `Memory.from_config()`, which is how users consume vector stores.

See inline comments for details.

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+        # Build query
+        # Syntax : SELECT <limit_clause> * FROM c WHERE <where_clause>
+        # Example: SELECT TOP @limit * FROM c WHERE c.metadata.a=@filter_value_0
+        query = "SELECT"
+        query += self._generate_limit_clause(param_mapping=param_mapping, limit=limit)
+        query += f" * FROM {self._table_alias}"
+        query += self._generate_where_clause(param_mapping=param_mapping, filters=filters)
+
+        parameters = param_mapping.export_parameter_list()
+        return self._execute_query(query=query, parameters=parameters)
```
aayush3011: 🟡 **Recommendation: `list()` uses wrong `return_with_vectors` default**

`_execute_query` defaults to `return_with_vectors=True` (line 688), so this call will attempt to include vectors in the output. This is semantically wrong for a `list()` operation, wastes bandwidth, and risks a `KeyError` at line 740 if the vector key isn't in the expected format.

**Fix:**
```python
return self._execute_query(query=query, parameters=parameters, return_with_vectors=False)
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+            ]
+        else:
+            projection_fields = [f"{self._table_alias}.{Constants.ID} as {Constants.ID}"]
+            projection_fields += [
+                param_mapping.gen_proj_field(key=key, value=key, alias=key)
+                for key in [self._text_key, self._metadata_key]
+            ]
+
+        # If it's a vector search type, include vector distance projection and optionally the vector itself
+        if Constants.VECTOR in search_type or Constants.HYBRID in search_type:
```
aayush3011: 🟡 **Recommendation: Fragile string containment check**

`Constants.VECTOR in search_type` checks substring containment — e.g., a future search type `"no_vector"` would incorrectly match because `"vector" in "no_vector"` is `True`. Same issue at line 623 with `"hybrid" in search_type`.

The codebase already has proper helpers (`_is_vector_search_type()`, `_is_full_text_search_type()`) that use tuple membership checks.

**Fix:** Use `self._is_vector_search_type(search_type)` instead of `Constants.VECTOR in search_type`.

mem0/configs/vector_stores/azure_cosmos_db.py
```diff
@@ -0,0 +1,45 @@
...
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+class AzureCosmosDBConfig(BaseModel):
+    """Configuration for Cosmos DB vector database."""
+
+    cosmos_client: Any = Field(None, description="CosmosClient used to connect to azure cosmosdb no sql account")
+    indexing_policy: Dict[str, Any] = Field(None, description="Indexing Policy for the collection")
+    cosmos_database_properties: Dict[str, Any] = Field(None, description="Database Properties for the collection")
+    cosmos_collection_properties: Dict[str, Any] = Field(None, description="Container Properties for the collection")
+    vector_properties: Dict[str, Any] = Field(None, description="Vector Embedding Properties for the collection.")
```
aayush3011: 🟡 **Recommendation: Fields typed `Dict[str, Any]` but defaulting to `None`**

These four fields are typed `Dict[str, Any]` but default to `None`. They should be `Optional[Dict[str, Any]]` to match the default value, consistent with `full_text_policy` on line 23 which correctly uses `Optional`.

_Note: This was [flagged by @Siddhant-K-code](https://github.com/mem0ai/mem0/pull/3775#discussion_r2694917125). The author replied "Done" but these fields still have incorrect types._

**Fix:**
```python
indexing_policy: Optional[Dict[str, Any]] = Field(None, ...)
cosmos_database_properties: Optional[Dict[str, Any]] = Field(None, ...)
cosmos_collection_properties: Optional[Dict[str, Any]] = Field(None, ...)
vector_properties: Optional[Dict[str, Any]] = Field(None, ...)
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+        else:
+            # If no projection mapping, include text and metadata in payload
+            payload = {
+                self._text_key: item.get(self._text_key, ""),
+                self._metadata_key: item.get(self._metadata_key, {})
+            }
+
+        # Include vector in payload if requested
+        if return_with_vectors:
+            payload[self._vector_key] = item[self._vector_key]
```
aayush3011: 🟡 **Recommendation: Unhandled `KeyError` on missing vector key**

`item[self._vector_key]` uses direct dict access, unlike the safe `.get()` used for text and metadata on lines 734–735 immediately above. If the item doesn't contain the vector key (e.g., data corruption or custom projection), this raises an unhandled `KeyError`.

**Fix:**
```python
payload[self._vector_key] = item.get(self._vector_key)
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+
+    def _search(
+            self,
+            query: str,
+            vectors: Optional[List[float]] = None,
+            limit: int = 5,
+            filters: Optional[Dict[str, Any]] = None,
+
+    ):
+        pass
```
aayush3011: 🟢 **Suggestion:** `_search()` is an empty stub method (`pass`) that is never called by `search()` or anything else. Consider removing it.

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+
+    def search(
+        self,
+        query: Optional[str] = None,
```
aayush3011, line 503: 🟢 **Suggestion: `query` parameter silently overwritten**

The `query` parameter (documented as "[Not Implemented]") is overwritten at line 416:
```python
query, parameters = self._construct_query(...)
```

The caller's text query is silently discarded. Consider renaming the local variable:
```python
sql_query, parameters = self._construct_query(...)
```

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+        query += self._generate_where_clause(param_mapping=param_mapping, filters=filters)
+
+        parameters = param_mapping.export_parameter_list()
+        return self._execute_query(query=query, parameters=parameters)
+
+    def reset(self) -> None:
+        """
+        Reset the collection by deleting and recreating it.
+        """
+        collection_name = self._collection_name
```
aayush3011: 🟢 **Suggestion:** `collection_name` is assigned but never used. Can be removed.

mem0/vector_stores/azure_cosmos_db_no_sql.py
```diff
@@ -0,0 +1,893 @@
...
+        return f" TOP {limit_key}"
+
+    def _generate_where_clause(
+            self,
+            param_mapping: ParamMapping,
+            filters: Optional[Dict[str, Any]] = None,
+    ) -> str:
+        where_clauses = []
+        if filters:
+            id = 0
```
aayush3011: 🟢 **Suggestion:** `id` shadows Python's built-in `id()`. Consider renaming to `idx`.

docs/components/vectordbs/dbs/azure_cosmos_db.mdx
```diff
@@ -0,0 +1,96 @@
...
+m.add(messages, user_id="alice", metadata={"category": "movies"})
+```
+
+## Config
+
+Here are the parameters available for configuring Cosmos DB as the vector database in Mem0:
+
+| Parameter | Description | Default Value |
+| --- | --- | --- |
+| cosmos_client | Cosmod DB Client used to connect to azure cosmosdb no sql account | |
```
aayush3011: 💬 **Nit:** "Cosmod" → "Cosmos"

docs/components/vectordbs/dbs/azure_cosmos_db.mdx
allenkim0129: Thank you for catching this! I renamed some variables and didn't get updated. 

I am glad you caught this!

docs/components/vectordbs/dbs/azure_cosmos_db.mdx
allenkim0129: Fixed

mem0/utils/factory.py
allenkim0129: fixed

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Removed it

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Good catch! I renamed the local variable

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Fixed as suggested.

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Renamed

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Make sense. I fixed as suggested

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: fixed it

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: removed

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: fixed

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: Good point. I enabled filters as well. Now filters and where are both supported. If both were given, we `AND` both conditions to apply both filters.

mem0/vector_stores/azure_cosmos_db_no_sql.py
allenkim0129: fixed as suggested

mem0/configs/vector_stores/azure_cosmos_db.py
allenkim0129: Updated and seperated the required and optional parameter lists

allenkim0129 (commented):
Addressed all comments

aayush3011 (commented):
Thanks for closing on the previous review comments — great work addressing all the blocking issues and most of the recommendations! Here are a few more findings from a follow-up review that should be fixed/looked at:

---

### 🔴 Blocking: Search Projection Drops Payload Fields

**File:** `mem0/vector_stores/azure_cosmos_db_no_sql.py` — `_generate_projection_fields()`

The search SQL only projects `id`, `description`, `metadata`, and `SimilarityScore`:

```sql
SELECT TOP @limit c.id as id, c[@description] as description, c[@metadata] as metadata,
  VectorDistance(c[@vectorKey], @vector) as SimilarityScore
FROM c ORDER BY VectorDistance(c[@vectorKey], @vector)
```

In real Cosmos DB, `query_items` returns **only the projected fields**. So payload keys like `data`, `hash`, `created_at`, and `user_id` are never returned. `Memory` expects these fields (e.g. `mem.payload.get("data", "")` in `main.py`), which breaks search, deduplication, and effectively the entire integration.

The mock tests mask this because they return complete items regardless of the SQL `SELECT` clause.

**Fix:** Use `SELECT *` in search queries (like `list()` already does), or project all fields. `_build_output_data_from_item` already handles stripping excluded fields from full items.

---

### 🟡 Recommendation: `get()` Doesn't Handle `CosmosResourceNotFoundError`

**File:** `mem0/vector_stores/azure_cosmos_db_no_sql.py` — `get()`

Both `delete()` and `update()` handle `CosmosResourceNotFoundError`, but `get()` lets the raw Azure SDK exception propagate. The return type is `Optional[OutputData]`, implying `None` should be returned on not-found, but it never actually returns `None`.

**Fix:**
```python
try:
    item = self._collection.read_item(item=vector_id, partition_key=partition_key_value)
except CosmosResourceNotFoundError:
    return None
```

---

### 🟡 Recommendation: Threshold Filter Uses `>` Instead of `>=`

**File:** `mem0/vector_stores/azure_cosmos_db_no_sql.py` — `_execute_query()`

```python
filtered_items = [item for item in items if item.get(Constants.SIMILARITY_SCORE, 0.0) > threshold]
```

An item with a score exactly equal to the threshold is excluded. Most users expect `>=` behavior ("give me items with at least 0.5 similarity").

**Fix:** Use `>=` instead of `>`.

---

### 🟡 Recommendation: Inconsistent Default Threshold

**File:** `mem0/vector_stores/azure_cosmos_db_no_sql.py`

- `search()` signature: `threshold: Optional[float] = 0.5`
- `_execute_query()` signature: `threshold: Optional[float] = 0.0`

Two different default values for the same logical concept. If `_execute_query` is ever called from a new method, it silently uses a different threshold.

**Fix:** Define the default once as a class constant and reference it in both places:
```python
DEFAULT_THRESHOLD = 0.5
```

allenkim0129:
@chaithanyak42 @deshraj @prathameshagrawal @whysosaket 

Can you review this PR?

jcodella:
ping @chaithanyak42 , @deshraj , @prathameshagrawal , @whysosaket for a review

allenkim0129:
Thank you for the thorough follow-up review @aayush3011! All four findings have been addressed in [commit a02cfdc5](https://github.com/allenkim0129/mem0/commit/a02cfdc5):

### 🔴 Search Projection Drops Payload Fields
**Fixed.** Rewrote `_generate_projection_fields()` to use `SELECT *` (with `VectorDistance(...) AS SimilarityScore` appended for vector types, and `FullTextScore(...)` for full-text ranking). This ensures all payload fields (`data`, `hash`, `user_id`, `created_at`, etc.) are returned. The `_build_output_data_from_item` method already handles stripping/including the vector field based on `return_with_vectors`.

### 🟡 `get()` Doesn't Handle `CosmosResourceNotFoundError`
**Fixed.** Wrapped `read_item` in a try/except for `CosmosResourceNotFoundError` that returns `None`, consistent with the `Optional[OutputData]` return type and behavior of other vector stores.

### 🟡 Threshold Filter Uses `>` Instead of `>=`
**Fixed.** Changed `item.get(Constants.SIMILARITY_SCORE, 0.0) > threshold` to `>= threshold` so items with score exactly equal to the threshold are included. Added `test_score_threshold_boundary` to verify this behavior.

### 🟡 Inconsistent Default Threshold
**Fixed.** Added `DEFAULT_THRESHOLD: float = 0.5` to the `Constants` class and updated both `search()` and `_execute_query()` to reference it, eliminating the 0.5 vs 0.0 discrepancy.

allenkim0129:
## Deep Review Fixes (Round 3)

Thanks again for the thorough review! Here's a summary of all 12 findings addressed in commit `93066839`:

### 🔴 Blocking Fixes

**#2 & #5 — `Memory.search()` integration broken for HYBRID/FULL_TEXT_RANKING; `text_field` config was dead**
`search()` now auto-builds `full_text_rank_filter` from `query` + the configured `text_field` when the search type requires text ranking and no explicit filter is provided. This fixes the real-world integration where `Memory.search(query="...", vectors=[...])` would crash with a missing-filter error. The `text_field` config value is now actually used (previously it was validated but never consumed).

**#4 — `Memory.delete_all()` broken: `list()` returned flat list instead of tuple**
`list()` now returns `(results, None)` to match Qdrant's `scroll()` convention. `Memory.delete_all()` does `self.vector_store.list(...)[0]`, which now correctly gets the record list.

### 🟡 Recommendations Fixed

**#3 — `reset()` held a stale container proxy**
`reset()` now captures the return value of `create_col()` into `self._collection` so the proxy is refreshed after recreation.

**#6 — `_validate_search_args()` didn't validate `full_text_rank_filter` presence for HYBRID/HYBRID_SCORE_THRESHOLD**
All three full-text search types (FULL_TEXT_RANKING, HYBRID, HYBRID_SCORE_THRESHOLD) now go through centralized validation in `_validate_search_args()`. Previously only FULL_TEXT_RANKING was checked (and in the wrong place).

**#7 — Empty `search_text` silently sent malformed query to Cosmos DB**
Added an explicit guard: `_generate_order_by_component_with_full_text_rank_filter()` now raises `ValueError` with a clear message if `search_text` is empty/whitespace-only.

**#9 — `_build_output_data_from_item()` used `item[alias]` risking KeyError**
Changed to `item.get(alias)` for the projection_mapping path.

### 🔵 Suggestions & Observations

**#1 & #10 — Docs: `search_type: "vector"` incompatible with `full_text_search_enabled: True`; `dimensions: 10` causes shape mismatch**
Fixed both in `azure_cosmos_db.mdx`: `search_type` is now `"hybrid"` and `dimensions` is now `1536` with a note to match your embedding model.

**#8 — `test_cosmos_client_user_agent` was a false positive**
The fixture itself made the `CosmosClient(url=..., user_agent_suffix=...)` call, so the assertion always passed regardless of production code. Rewrote as a value check: asserts `Constants.USER_AGENT` is a non-empty string containing `"Mem0"`.

**#11 — Weights parameterization note (observation, no code change)**
Left as-is; the mock-based tests can't validate Cosmos DB's runtime behavior on parameterized `RANK RRF` weights.

**#12 — Removed inline changelog comment from `ParamMapping`**
Cleaned up `# Renamed from PramMapping (typo fix)` comment.

### 🧪 Test Updates (50 tests, all passing)
- Updated `test_list` + `test_list_excludes_vectors_from_payload` for tuple unpacking
- Updated `test_reset` to assert `self._collection` is refreshed
- Updated `test_search_with_errors` for new consolidated error message
- Added `test_list_delete_all_compatible` — verifies `list()[0]` pattern works
- Added `test_search_auto_builds_full_text_rank_filter_from_query` — verifies auto-build from `query=`
- Added `test_search_empty_search_text_raises_error` — verifies empty search_text guard
- Rewrote `test_cosmos_client_user_agent` as a meaningful value assertion

All 50 tests pass ✅

aayush3011:
@whysosaket @kartik-mem0 @gabrielstein-mem0 pinging again for a review

aayush3011:
@whysosaket @kartik-mem0 @gabrielstein-mem0 pinging again for a review

Issue #3788, opened 2025-11-26
using LMStudio, search returns no results


MatthiasHowellYopp:
### 🐛 Describe the bug

Running the mem0 OSS getting started sample with LMStudio,Ollama,Valkey - results are empty:

```
python main.py
{'results': []}
```

code is :
```
from mem0 import Memory

def main():
    config = {
       "vector_store": {
           "provider":"valkey",
           "config": {
               "collection_name":"mem0test",
               "valkey_url":"valkey://192.168.5.111:6380",
               "embedding_model_dims":1024,
               "index_type":"flat"
               }
       },
       "llm":  {
              "provider":"lmstudio",
              "config": {
                "model":"lmstudio-community/Meta-Llama-3.1-70B-Instruct-GGUF/Meta-Llama-3.1-70B-Instruct-IQ2_M.gguf",
                "temperature":0.2,
                "max_tokens":2000,
                "lmstudio_base_url":"http://192.168.5.24:1234/v1",
                "lmstudio_response_format": {"type": "json_schema", "json_schema": {"type": "object", "schema": {}}},
                }       
        },
        "embedder": {
            "provider":"ollama",
            "config": {
                "model":"mxbai-embed-large:latest",
                "ollama_base_url":"http://192.168.5.111:11434"
                
            }
        }
    }
        
    m:Memory = Memory.from_config(config)
    messages = [
    {"role": "user", "content": "Hi, I'm Alex. I love basketball and gaming."},
    {"role": "assistant", "content": "Hey Alex! I'll remember your interests."}
    ]
    m.add(messages, user_id="alex")
    results = m.search("What do you know about me?", user_id= "alex")
    print(results)


if __name__ == "__main__":
    main()
```

When I tried with gpt-oss the contents came back as a floating point value.  With this llm the result look like:
```
 "id": "chatcmpl-lvjf4hppwvnjv1m4yci4l",
  "object": "chat.completion",
  "created": 1764107940,
  "model": "meta-llama-3.1-8b-instruct",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
[...]

PR #3796, opened 2025-11-29
Fix lmstudio json parsing robustness

oxyec:
Description
This PR improves the robustness of JSON parsing in Memory.add and AsyncMemory.add to support local LLMs (like LM Studio, Ollama, etc.) that often provide "chatty" responses.

These models frequently wrap their JSON output in conversational text (e.g., "Here is the memory extraction: ...") or fail to use strict Markdown code blocks. Previously, this caused json.loads to raise a JSONDecodeError, resulting in empty memory results because the system assumed the response was either pure JSON or a clean Markdown code block.

Changes included:

Updated extract_json in mem0/memory/utils.py: Added a fallback mechanism. If a standard Markdown code block (json ... ) is not found, it now attempts to locate the JSON payload by finding the substring between the first { and the last }.
Updated mem0/memory/main.py: Implemented a try-except block in the memory extraction flow. If the standard parsing (via remove_code_blocks) fails, it now attempts to recover using the enhanced extract_json utility.
Fixes # (issue relating to empty results with LMStudio/Ollama)

Type of change
Bug fix (non-breaking change which fixes an issue)
New feature (non-breaking change which adds functionality)
Breaking change (fix or feature that would cause existing functionality to not work as expected)
Refactor (does not change functionality, e.g. code style improvements, linting)
Documentation update
How Has This Been Tested?
I added a new regression test file tests/test_chatty_llm.py that mocks the LLM responses to simulate the failure cases (verbose output with and without markdown).

Unit Test: tests/test_chatty_llm.py
test_chatty_llm_response_sync: Verifies Memory class handles JSON wrapped in text + markdown.
test_chatty_no_markdown_response_sync: Verifies Memory class handles JSON wrapped in text without markdown.
test_chatty_llm_response_async: Verifies AsyncMemory class handles the same scenarios.
Ran tests using:

python -m pytest tests/test_chatty_llm.py
Checklist:
My code follows the style guidelines of this project
I have performed a self-review of my own code
I have commented my code, particularly in hard-to-understand areas
I have made corresponding changes to the documentation
My changes generate no new warnings
I have added tests that prove my fix is effective or that my feature works
New and existing unit tests pass locally with my changes
Any dependent changes have been merged and published in downstream modules
I have checked my code and corrected any misspellings
Maintainer Checklist
closes  (hopefully) #3788
(this is my first pr pls be respectfully if I have major mistakes)
This contribution includes code generated with the assistance of AI (e.g., ChatGPT). (Be aware of this fact)
Made sure Checks passed

### DIFF
--- a/tests/test_chatty_llm.py
+++ b/tests/test_chatty_llm.py
@@ -1,7 +1,6 @@
 
 import pytest
 from unittest.mock import MagicMock
-import json
 from mem0 import Memory, AsyncMemory
 
 # Chatty response simulating local LLMs (Ollama/LM Studio)


utkarsh240799:
Hi @oxyec, thanks for the PR and for being the first to tackle this! We appreciate the effort.

We've gone ahead and implemented a fix for this internally. The core approach is the same — adding a `JSONDecodeError` fallback in the memory action parsing path — but we opted to create a separate `extract_json_from_chatty_response()` function rather than modifying the existing `extract_json()`, since it's used by 15+ callers across LLM providers and vector stores where the `{`/`}` heuristic isn't needed and could introduce unintended side effects.

Closing this in favor of our internal fix. Thanks again for identifying the right solution!

PR #3797, opened 2025-11-30
Fix: expose metadata parameter in update() method

railikitha856-cyber:
### Problem:
The public 'update()' method did not accept a 'metadata' parameter,  
even though the internal '_update_memory()'supports it and the documentation mentions it.  

### Solution:
- Added 'metadata=None' to 'update()' method signature  
- Forwarded 'metadata' to '_update_memory()'  
- Added a test to verify metadata updates correctly

This brings the public API in line with the documentation.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -989,26 +989,16 @@ class Memory(MemoryBase):
 
         return original_memories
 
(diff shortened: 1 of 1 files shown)


kartik-mem0:
this issue also has been resolved

PR #3798, opened 2025-12-01
Handle single vector case in query processing

mdhyaksa:
Added handling for single vector input by wrapping it in a list.

## Description

Upstash vector may a single vector instead of List of vectors

Fixes # (issue)
add type checking, if it's a single vector wrap it in a list

## Type of change

- [v] Bug fix (non-breaking change which fixes an issue)


kartik-mem0:
Thanks for the Upstash single vector fix. This has gone stale without review. Closing it for now — if the issue persists, feel free to resubmit.

Issue #3580, opened 2025-10-14
Bug: _safe_deepcopy_config removes http_auth from OpenSearch configs


Harjotsinghh:
### Summary

After updating to include changes from [PR #3544](https://github.com/mem0ai/mem0/pull/3544), OpenSearch integrations started failing with 403 / signature errors.
Investigation shows that `_safe_deepcopy_config()` is unintentionally removing the http_auth field from the vector-store config, breaking authentication for AWS-signed OpenSearch clients, with this error message : 
```
WARNING:opensearch:HEAD https://xxxx.xx-xx-1.aoss.amazonaws.com:443/mem0migrations [status:403 request:0.893s]
ERROR:xx.xx.xx.xx.mem0_memory_agent:Failed to initialize mem0 with OpenSearch: AuthorizationException(403, '')

```

### Root Cause

In [PR #3544](https://github.com/mem0ai/mem0/pull/3544), _safe_deepcopy_config() introduced a sanitizer to remove sensitive or non-serializable fields:

```
# in memory/main.py L97
sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
for field_name in list(clone_dict.keys()):
    if any(token in field_name.lower() for token in sensitive_tokens):
        clone_dict[field_name] = None

```
Because the list includes "auth", it matches the field name http_auth, which is required by the OpenSearch client for signed requests.
This causes the cloned config to lose the authentication object.

### Debug Process
I added minimal print statements around the sanitization loop to trace field handling:
```
[DEBUG] BEFORE sanitize: http_auth -> AWSV4SignerAuth
[DEBUG] Sanitizing field: password
[DEBUG] Sanitizing field: api_key
[DEBUG] Sanitizing field: http_auth   <-- culprit
[DEBUG] Sanitizing field: connection_class
[DEBUG] AFTER sanitize: http_auth -> None

```
### Impact

All OpenSearch integrations using AWSV4SignerAuth (or any custom http_auth) fail because http_auth is set to None inside cloned configs.

### Proposed Solution 

The current sanitizer in _safe_deepcopy_config() is overly broad —
it removes any field name containing "auth", which unintentionally nulls http_auth, a required runtime field for OpenSearch and similar clients.

We should preserve runtime authentication objects (like http_auth, auth, and connection_class) while still stripping genuinely sensitive data (like password, token, api_key, etc.).

This keeps runtime behavior intact while maintaining telemetry safety.

[...]

PR #3816, opened 2025-12-09
fix: Preserve http_auth and connection_class in _safe_deepcopy_config

MirrorDNA-Reflection-Protocol:
Fixes #3580.
**Context**: Opensearch and other vector stores require 'http_auth' and 'connection_class' for runtime connection, but '_safe_deepcopy_config' was aggressively stripping them due to 'auth' substring matching.

**Fix**: Added an explicit allowlist for 'http_auth', 'auth', and 'connection_class' to preserve them during configuration deepcopy/reconstruction.


utkarsh240799:
Hi @MirrorDNA-Reflection-Protocol , 

Thanks again for your effort!

PR #3817, opened 2025-12-09
fix(telemetry): Add missing capture_event to add methods

MirrorDNA-Reflection-Protocol:
Fixes missing telemetry for 'add' operations.

**Issue**: The 'add' methods in Memory and AsyncMemory were not calling 'capture_event("mem0.add", ...)', causing a gap in usage tracking compared to 'get', 'update', and 'delete'.

**Fix**: Added consistent 'capture_event' calls to both methods.


kartik-mem0:
Thanks for catching the missing telemetry capture_event calls. This has gone stale without review. Closing it for now.

PR #3842, opened 2025-12-17
fix(api): sync vector store updates to postgres in MCP and API

### LINKED ISSUE #3841: Bug: Data Inconsistency between UI (Postgres) and MCP (Vector Store) due to unhandled UPDATE events
### Description
When mem0 updates an existing memory in the vector store (e.g., refining a user's age based on new input), the OpenMemory API (Postgres) fails to sync this change. The current implementation in routers/memories.py and mcp_server.py only handles ADD events from the memory client, ignoring UPDATE events.

This results in a "split-brain" state where:

- MCP / Vector Search : Returns the correct, updated information (e.g., "Age: 43").
- UI / Postgres : Displays the outdated information (e.g., "Age: 41").
### Steps to Reproduce
1. Add a memory via UI or API (e.g., "My age is 41").
2. Verify it appears in the UI list as "My age is 41".
3. Add a new memory that updates the previous fact (e.g., "Actually, my age is 43").
   - mem0 will correctly identify this as an update to the existing vector ID.
4. Query via MCP (semantic search): Returns correct value "43".
5. Check the UI list (Postgres): Still shows "41".
### Expected Behavior
The Postgres database should remain synchronized with the Vector Store. When mem0 triggers an UPDATE event, the corresponding record in the Postgres memories table should be updated with the new content immediately.

### Actual Behavior
The Postgres database ignores the UPDATE event and retains the original content, causing data inconsistency between the search interface (Vector) and the management interface (SQL).

### Root Cause
The loop processing qdrant_response only checks for ADD events:

```python
# Current Code
if result['event'] == 'ADD':
    # ... logic to add to Postgres ...
```
### Proposed Fix
Update the logic to handle both ADD and UPDATE events:

1. Check for if result['event'] == 'ADD' or result['event'] == 'UPDATE': .
2. If the memory ID already exists in Postgres, update its content and state.
3. Record the state change in MemoryStatusHistory .

zch0214:
## Description
This PR fixes a critical data inconsistency issue between the Vector Store (used by MCP semantic search) and the Postgres database (used by the UI).

Previously, the OpenMemory API only handled ADD events from the underlying mem0 memory client. When mem0 triggered an UPDATE event (e.g., refining an existing memory based on new information), the change was applied to the Vector Store but ignored by Postgres. This resulted in the UI displaying outdated information (e.g., an old age value) while semantic search returned the correct, updated value.

This change updates openmemory/api/app/routers/memories.py and openmemory/api/app/mcp_server.py to:

1. Listen for UPDATE events in addition to ADD events.
2. Sync the updated memory content from the Vector Store response to the Postgres database.
3. Correctly record the state transition in the MemoryStatusHistory table.
Fixes #3841

## Type of change
- Bug fix (non-breaking change which fixes an issue)
## How Has This Been Tested?
- Test Script (please provide)
I used a reproduction script to verify the fix:

1. Initial State : Added a memory "My age is 41". Verified both UI and MCP returned "41".
2. Update Trigger : Added a new memory "Actually, my age is 43".
3. Verification :
   - Before Fix : UI showed "41", MCP showed "43".
   - After Fix : UI shows "43", MCP shows "43". Postgres database now correctly reflects the updated content.
## Checklist:
- My code follows the style guidelines of this project
- I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas
- I have made corresponding changes to the documentation
- My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- New and existing unit tests pass locally with my changes
- Any dependent changes have been merged and published in downstream modules
- I have checked my code and corrected any misspellings
## Maintainer Checklist
- closes #3841
- Made sure Checks passed

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -94,9 +94,13 @@ async def add_memories(text: str) -> str:
                 for result in response['results']:
                     memory_id = uuid.UUID(result['id'])
                     memory = db.query(Memory).filter(Memory.id == memory_id).first()
+                    
+                    # Capture previous state
+                    previous_state = memory.state if memory else MemoryState.deleted
+                    existed = memory is not None
 
-                    if result['event'] == 'ADD':
-                        if not memory:
+                    if result['event'] == 'ADD' or result['event'] == 'UPDATE':
+                        if not existed:
                             memory = Memory(
                                 id=memory_id,
                                 user_id=user.id,
@@ -113,7 +117,7 @@ async def add_memories(text: str) -> str:
                         history = MemoryStatusHistory(
                             memory_id=memory_id,
                             changed_by=user.id,
-                            old_state=MemoryState.deleted if memory else None,
+                            old_state=previous_state,
                             new_state=MemoryState.active
                         )
                         db.add(history)
@@ -126,7 +130,7 @@ async def add_memories(text: str) -> str:
                             history = MemoryStatusHistory(
                                 memory_id=memory_id,
                                 changed_by=user.id,
-                                old_state=MemoryState.active,
+                                old_state=previous_state,
                                 new_state=MemoryState.deleted
                             )
                             db.add(history)
--- a/openmemory/api/app/routers/memories.py
+++ b/openmemory/api/app/routers/memories.py
@@ -274,12 +274,13 @@ async def create_memory(
             created_memories = []
             
             for result in qdrant_response['results']:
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for tackling the vector store to Postgres sync issue. We're in the process of winding down the OpenMemory components in this repo, so this won't be merged here going forward. Appreciate you working on it — closing this out.

PR #3859, opened 2025-12-27
fix(ts-sdk): use dynamic import for ollama to support optional dependency

### LINKED ISSUE #3857: [TypeScript SDK] Static require('ollama') breaks apps not using Ollama provider
## Description

The TypeScript SDK (`mem0ai/oss`) fails to load when the `ollama` package is not installed, even when using a completely different LLM provider (e.g., Gemini).

## Environment

- mem0ai version: 2.1.38
- Node.js: 18+
- OS: Ubuntu 22.04

## Configuration

```typescript

const config = {
  llm: {
    provider: "gemini",  // NOT using Ollama
    config: {
      model: "gemini-2.0-flash-001",
      apiKey: process.env.GEMINI_API_KEY,
    }
  },
  embedder: {
    provider: "gemini",  // NOT using Ollama
    config: {
      model: "gemini-embedding-001",
      apiKey: process.env.GEMINI_API_KEY,
    }
  },
  vectorStore: {
    provider: "redis",
    config: { redisUrl: "redis://localhost:6379" }
  }
};

const memory = new Memory(config);
```

## Error

```
Error: Cannot find module 'ollama'
Require stack:
- /app/node_modules/mem0ai/dist/oss/index.js
```

## Root Cause

The bundled `dist/oss/index.js` contains static requires for ALL providers:

```javascript
var import_ollama = require("ollama");   // Line executed at module load
var import_ollama2 = require("ollama");  // Even if not using Ollama!
```

## Expected Behavior

Peer dependencies like `ollama` should be dynamically imported only when the corresponding provider is actually used:

```typescript
// Instead of:
const ollama = require("ollama");

// Should be:
async function getOllamaClient() {
  const { Ollama } = await import("ollama");
  return new Ollama();
}
```

## Workaround

Currently users must install `ollama` even if not using it:
```bash
npm install ollama
```

## Comparison with Python SDK

The Python SDK handles this correctly with lazy imports and clear error messages:
```python
# mem0/llms/ollama.py
try:
    from ollama import Client
except ImportError:
    raise ImportError("The 'ollama' library is required...")
```

## Suggested Fix

1. Use dynamic `import()` for optional provider dependencies
2. Or split the bundle so provider-specific code is in separate chunks
3. Or use try/catch around requires with graceful fallback

## Related

- Issue #1768 - Automate Package Installation for Missing Dependencies

majiayu000:
## Summary
- Changed static `import { Ollama } from 'ollama'` to dynamic `await import('ollama')` 
- Ollama package is now only loaded when using Ollama provider
- Clear error message when ollama package is missing

## Test plan
- [x] Added tests to verify modules can be imported without ollama installed
- [x] Verified factory can create non-Ollama providers
- [x] Existing factory tests pass

Fixes #3857

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/ollama.ts
+++ b/mem0-ts/src/oss/src/embeddings/ollama.ts
@@ -1,33 +1,55 @@
-import { Ollama } from "ollama";
+import type { Ollama } from "ollama";
 import { Embedder } from "./base";
 import { EmbeddingConfig } from "../types";
 import { logger } from "../utils/logger";
 
+let OllamaClient: typeof Ollama | null = null;
+
+async function getOllamaClient(): Promise<typeof Ollama> {
+  if (!OllamaClient) {
+    try {
+      const ollamaModule = await import("ollama");
+      OllamaClient = ollamaModule.Ollama;
+    } catch (error) {
+      throw new Error(
+        "The 'ollama' package is required to use Ollama provider. " +
+          "Please install it with: npm install ollama"
+      );
+    }
+  }
+  return OllamaClient;
+}
+
 export class OllamaEmbedder implements Embedder {
-  private ollama: Ollama;
+  private ollama: Ollama | null = null;
   private model: string;
   private embeddingDims?: number;
+  private host: string;
   // Using this variable to avoid calling the Ollama server multiple times
   private initialized: boolean = false;
 
   constructor(config: EmbeddingConfig) {
-    this.ollama = new Ollama({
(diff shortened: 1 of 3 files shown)


Ayush41:
Looks like the PR fixes the issue, but the CI/CD checks are failing. If I can help in any way, feel free to let me know!

majiayu000:
@Ayush41 need a  maintainer review.

kartik-mem0:
thank you for your contribution but #4408 pr has better solution. 

closing the pr now.

PR #3862, opened 2025-12-27
fix: add support for LangChain v1.0+ imports

### LINKED ISSUE #3793: Langchain v1 is not supported with current mem0 version
Now that ```langchain``` has officially upgraded to their stable version which  is ```v1``` but this version is not being supported by ```mem0```.

majiayu000:
Fixes #3793


kartik-mem0:
closing this pr as #4353 is the correct solution.

PR #3870, opened 2025-12-31
feat: add top_k as alias for limit in search()

### LINKED ISSUE #3819: Can the search method limit the number of memory entries returned
### 🚀 The feature

``
search_results = self.client.search(
                query="alice", 
                user_id="yoga",
                top_k=10,  
                output_format=self.api_version
              )

## des：This method passed the parameter top_k=10, but the return memory is still the same as not passing it, with 100 entries. Is there any problem
``

### Motivation, pitch

I would like to know if it is possible to constrain the number of return memories for search, and how to do it?

pmukeshreddy:
## Summary
Users were passing `top_k` expecting it to limit results, but the correct parameter is `limit`. This adds `top_k` as an alias for better developer experience.

## Changes
- Added `top_k` parameter to `Memory.search()` 
- Added `top_k` parameter to `AsyncMemory.search()`
- When `top_k` is provided, it overrides `limit`

## Related Issue
Fixes #3819

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -763,7 +763,8 @@ class Memory(MemoryBase):
         agent_id: Optional[str] = None,
         run_id: Optional[str] = None,
         limit: int = 100,
-        filters: Optional[Dict[str, Any]] = None,
+        top_k: Optional[int] = None,
+        filters: Optional[Dict[str, Any]] = None, # legacy filters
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi,

Thanks for taking out the time for coming up with this PR. The issue in question has been fixed by PR #4496 . Closing this PR.

Thank you.

PR #3916, opened 2026-01-16
fix(config): pass through baseURL for embedder config

gurvancampion:
## Summary

The `ConfigManager.mergeConfig` function was not passing through the `baseURL` field for the embedder configuration, even though:
1. The LLM config correctly passes through `baseURL`
2. The `EmbeddingConfig` type schema already supports `baseURL`
3. `OpenAIEmbedder` uses `baseURL` when creating the OpenAI client

## Problem

When using a custom `baseURL` for embeddings (e.g., OpenRouter, Azure OpenAI, or self-hosted endpoints), the config was silently ignored, causing all embedding calls to go to `api.openai.com`.

```typescript
const memory = new Memory({
  embedder: {
    provider: 'openai',
    config: {
      apiKey: 'my-key',
      baseURL: 'https://openrouter.ai/api/v1',  // ← This was ignored
      model: 'openai/text-embedding-3-small',
    },
  },
})
```

## Fix

1. Add `baseURL: userConf?.baseURL || defaultConf.baseURL` to the embedder config in `manager.ts`, matching the existing LLM config behavior
2. Add default `baseURL: "https://api.openai.com/v1"` to embedder config in `defaults.ts` for consistency with LLM defaults

## Testing

Tested locally with OpenRouter as the embedder baseURL - embeddings now correctly route to the custom endpoint instead of defaulting to `api.openai.com`.

### DIFF
--- a/mem0-ts/src/oss/src/config/defaults.ts
+++ b/mem0-ts/src/oss/src/config/defaults.ts
@@ -6,6 +6,7 @@ export const DEFAULT_MEMORY_CONFIG: MemoryConfig = {
   embedder: {
     provider: "openai",
     config: {
+      baseURL: "https://api.openai.com/v1",
       apiKey: process.env.OPENAI_API_KEY || "",
       model: "text-embedding-3-small",
     },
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hey, this has developed merge conflicts and there hasn't been activity in a while. Closing it for now — if you'd like to resubmit a rebased version, happy to take another look.

PR #3919, opened 2026-01-19
fix(ts): lazy-load SQLite3 to prevent import time crashes

### LINKED ISSUE #3882: Node.js SDK: Lazy-load sqlite3 to avoid requiring native bindings when not used
## Problem

The Node.js SDK (`mem0ai` npm package) eagerly imports `sqlite3` at module load time, even when using alternative storage backends like pgvector.

In `dist/oss/index.js`:
```javascript
var import_sqlite3 = __toESM(require("sqlite3"));
```

This causes issues when:
1. Using Docker/Alpine Linux where sqlite3 native bindings need to be rebuilt
2. Users only want to use pgvector or other vector stores
3. Serverless environments where native module compilation isn't available

## Current Workaround

We have to rebuild sqlite3 in our Dockerfile even though we configure mem0 to use pgvector:

```dockerfile
# Force rebuild of native modules (even though we use pgvector, not sqlite)
RUN npm rebuild sqlite3
```

## Suggested Solution

Lazy-load sqlite3 only when actually needed (i.e., when `historyDbPath` is configured or when using SQLite-based storage):

```typescript
// Instead of top-level import:
// import sqlite3 from 'sqlite3';

// Lazy-load when needed:
class SQLiteHistoryManager {
  private db: any;
  
  constructor(dbPath: string) {
    const sqlite3 = require('sqlite3'); // Load only when instantiated
    this.db = new sqlite3.Database(dbPath);
  }
}
```

This pattern is common for optional native dependencies - see how `sharp`, `canvas`, and other packages handle this.

## Environment

- mem0ai version: 2.2.0
- Node.js version: 20.x
- Platform: Alpine Linux (Docker)

## Related

This is similar to the pgvector factory issue (#3491) where the JS SDK differs from Python in terms of provider support.

shubham-021:
## Description

Currently the sdk import `sqlite3` at module load time, even when users doesn't configure it. This causes the package to fail in environment where sqlite3 native bindings cannot compile.

Future Work : Current used sqlite package doesn't compile under environments like Bun. While this PR fixes the issue for users who doesn't need sqlite in their configuration , we still need a way to make sure it works in every envs.

Fixes #3882 

## Type of change

- [x] Refactor (does not change functionality, e.g. code style improvements, linting)

## How Has This Been Tested?

- Added an example file `/mem0-ts/src/oss/examples/lazy-load-sqlite.ts` , running this file crashes the process without the fix when not using sqlite3.

## Checklist:

## Maintainer Checklist

- [ ] Made sure Checks passed

### DIFF
--- a/mem0-ts/src/oss/examples/lazy-load-sqlite.ts
+++ b/mem0-ts/src/oss/examples/lazy-load-sqlite.ts
@@ -5,29 +5,29 @@ import { Memory } from "../src";
 console.log("\nImport succeeded , sqlite3 was not eagerly loaded\n");
 
 try {
-    const memory = new Memory({
-        embedder: {
-            provider: "openai",
-            config: {
-                apiKey: ''
-            }
-        },
-        vectorStore: {
-            provider: "qdrant",
-            config: {
-                collectionName: "test",
-                url: "http://localhost:6333"
-            }
-        },
-        llm: {
-            provider: 'openai',
-            config: {
-                apiKey: ''
-            }
-        }
-    })
+  const memory = new Memory({
+    embedder: {
+      provider: "openai",
+      config: {
+        apiKey: "",
+      },
+    },
+    vectorStore: {
+      provider: "qdrant",
+      config: {
+        collectionName: "test",
+        url: "http://localhost:6333",
+      },
+    },
+    llm: {
+      provider: "openai",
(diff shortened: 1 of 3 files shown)


utkarsh240799:
Resolved the issue in #4270 . Closing the PR.

PR #3920, opened 2026-01-20
feat(elasticsearch): Add API key authentication for self-hosted instances

rst0070:
## Description

Add support for connecting to Elasticsearch using API key authentication without requiring cloud_id.

Previously, API key auth was only available for Elastic Cloud (via cloud_id). This change enables API key authentication for self-hosted Elasticsearch instances by allowing `(host, api_key)` configuration.

## Type of change

## How Has This Been Tested?

- [x] Unit Test

Test script
```python
from mem0.vector_stores.elasticsearch import ElasticsearchDB
config = {
     "host": "https://host",
     "port": 443,
     "api_key": "api-key",
    }
es_db = ElasticsearchDB(**config)
print(es_db.list_cols())
```

## Checklist:

## Maintainer Checklist

- [ ] Made sure Checks passed

### DIFF
--- a/mem0/vector_stores/elasticsearch.py
+++ b/mem0/vector_stores/elasticsearch.py
@@ -33,6 +33,13 @@ class ElasticsearchDB(VectorStoreBase):
                 verify_certs=config.verify_certs,
                 headers= config.headers or {},
             )
+        elif config.api_key:
+            self.client = Elasticsearch(
+                hosts=[f"{config.host}" if config.port is None else f"{config.host}:{config.port}"],
+                api_key=config.api_key,
+                verify_certs=config.verify_certs,
+                headers= config.headers or {},
+            )
         else:
             self.client = Elasticsearch(
                 hosts=[f"{config.host}" if config.port is None else f"{config.host}:{config.port}"],
--- a/tests/vector_stores/test_elasticsearch.py
+++ b/tests/vector_stores/test_elasticsearch.py
@@ -9,8 +9,8 @@ try:
 except ImportError:
     raise ImportError("Elasticsearch requires extra dependencies. Install with `pip install elasticsearch`") from None
 
-from mem0.vector_stores.elasticsearch import ElasticsearchDB, OutputData
 from mem0.configs.vector_stores.elasticsearch import ElasticsearchConfig
+from mem0.vector_stores.elasticsearch import ElasticsearchDB, OutputData
 
 
 class TestElasticsearchDB(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


mem0/vector_stores/elasticsearch.py
```diff
@@ -33,6 +33,13 @@ def __init__(self, **kwargs):
                 verify_certs=config.verify_certs,
                 headers= config.headers or {},
             )
+        elif config.api_key:
+            self.client = Elasticsearch(
+                hosts=[f"{config.host}" if config.port is None else f"{config.host}:{config.port}"],
+                api_key=config.api_key,
+                verify_certs=config.verify_certs,
+                headers= config.headers or {},
+            )
```
rst0070, line 42: To allow elastic authentication by pair of (`host`, `api_key`)

tests/vector_stores/test_elasticsearch.py
```diff
@@ -325,6 +325,31 @@ def test_es_config(self):
...
+        es_config = ElasticsearchConfig(**config)
+
+        # Assert that the config object was created successfully
+        self.assertIsNotNone(es_config)
+        self.assertIsInstance(es_config, ElasticsearchConfig)
+
+        # Assert that the configuration values are correctly set
+        self.assertEqual(es_config.cloud_id, "test-cloud-id")
+        self.assertEqual(es_config.api_key, "test-api-key")
+
```
rst0070, line 352: The tests follow other test style above

rst0070:
@deshraj When you have a moment, I'd appreciate your thoughts on this. thanks

kartik-mem0:
> *This comment was generated by AI. A human maintainer will make the final decision.*

## Review: PR #3920 — feat(elasticsearch): Add API key authentication for self-hosted instances

**Verdict: CHANGES REQUESTED**

The intent is correct: self-hosted Elasticsearch supports API key auth but the current code only handles it for `cloud_id` connections. The fix is nearly right but has one issue.

---

### Issue: Priority ordering in the if/elif chain

The PR adds:
```python
elif config.api_key:
    self.client = Elasticsearch(
        hosts=[...],
        api_key=config.api_key,
        ...
    )
```

But the config validator in `mem0/configs/vector_stores/elasticsearch.py:33` allows both `api_key` AND `(user + password)` to be set simultaneously — only requiring at least one. If a user sets both `api_key` and `user`/`password` (perhaps during config migration), the `elif config.api_key` branch will take priority over basic auth, which may be unexpected. 

More importantly: the validator at line 33 says `if not any([values.get("api_key"), ...])` — but it doesn't currently raise if `cloud_id` is absent AND `api_key` is absent AND basic auth is absent. This means the `else` branch at the end of the `__init__` chain will be hit with no auth, creating an unauthenticated client. This is a pre-existing issue but worth noting.

**Suggested fix:** Reorder the elif so `api_key` check comes before the basic_auth check:

```python
if config.cloud_id:
    # cloud ID path (existing)
elif config.api_key:
    # api_key without cloud_id (this PR's addition) — correct placement
else:
    # basic_auth fallback
```

The current code already follows this order, so the PR is functionally correct for the common case. The concern is just around simultaneous `api_key` + `basic_auth` config — since the validator allows it, the PR's behavior (api_key wins for non-cloud) should be documented.

---

### Tests

The added tests validate `ElasticsearchConfig` creation but do NOT test that the `Elasticsearch` client is actually instantiated with `api_key=` for the non-cloud case. A test using `unittest.mock.patch` to verify the `Elasticsearch()` call arguments would be more robust (see existing `TestElasticsearchDB.test_es_config` pattern).

---

**Minor:** No documentation update in `docs/components/vectordbs/dbs/elasticsearch.mdx` for the new auth option. This should be added.

Overall the core change is correct and useful — just needs the test gap addressed and a docs update.

kartik-mem0:
Thanks for this, Wonbin, and sorry for the slow turnaround. To be clear, the fix itself is valid — self-hosted Elasticsearch with API-key auth is a real gap (the config accepts api-key-only auth, but the self-hosted client path doesn't actually use it), and your branch addresses it correctly.

I'm going to close this for now to keep our PR queue manageable — not a reflection on the work. If you're still using this, please reopen and we'll land it: it's close to merge-ready, mainly needing `ca_certs=config.ca_certs` added to the new branch (to match the other two) and a test that exercises the new client path. Really appreciate the contribution and your patience.

PR #3939, opened 2026-01-26
Fix database parameter assignment in graph_memory.py

wanhhe:
## Description

Due to an update in the langchain-neo4j library, a new field, `token`, has been added between the `password` and `database` parameters when instantiating the Neo4jGraph class. This causes the current code to assign the value of the `database` variable to the `token` variable. The input parameters need to be explicitly specified.

You can find code here: [langchain-neo4j](https://github.com/langchain-ai/langchain-neo4j/blob/main/libs/neo4j/langchain_neo4j/graphs/neo4j_graph.py#L98)

**Before**
image

**Now**
image

## Type of change

- [-] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

My Langchain-Neo4j version is 0.8.0, and I tested it with Neo4j Azure without any issues. I also believe that changing from implicit to explicit parameters shouldn't introduce any significant problems.

### DIFF
--- a/mem0/memory/graph_memory.py
+++ b/mem0/memory/graph_memory.py
@@ -33,7 +33,7 @@ class MemoryGraph:
             self.config.graph_store.config.url,
             self.config.graph_store.config.username,
             self.config.graph_store.config.password,
-            self.config.graph_store.config.database,
+            database=self.config.graph_store.config.database,
             refresh_schema=False,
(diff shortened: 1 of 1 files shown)


kartik-mem0:
closing this pr as graph memory does not exist in current codebase.

PR #3981, opened 2026-02-06
fix: pass database as keyword arg in Neo4jGraph init to avoid token misassignment

### LINKED ISSUE #3906: Bug：Fix incorrect parameter order in Neo4jGraph initialization in MemoryGraph
### 🐛 Describe the bug

In the `MemoryGraph.__init__` method of `mem0/memory/graph_memory.py`, there is a parameter passing error when initializing `Neo4jGraph`.

## Current Code (Lines 32-39)

```python
self.graph = Neo4jGraph(
    self.config.graph_store.config.url,
    self.config.graph_store.config.username,
    self.config.graph_store.config.password,
    self.config.graph_store.config.database,  # ❌ Error: passed as positional argument
    refresh_schema=False,
    driver_config={"notifications_min_severity": "OFF"},
)
```

## Why This Issue May Not Have Been Widely Discovered?

Although this bug does exist, most users may not have encountered the problem for the following reasons:

1. **Using Default Database**: If users use the default `"neo4j"` database, even with incorrect parameter passing, the `get_from_dict_or_env` function in `Neo4jGraph.__init__` will use the default value `"neo4j"`, so functionality works correctly.

2. **Configuration via Environment Variables**: If users set the database name through the `NEO4J_DATABASE` environment variable, `get_from_dict_or_env` will read from the environment variable, so it works correctly even with incorrect parameter passing.

3. **Issue Only Occurs in Specific Scenarios**:
   - User passes a **non-default** database name through the config object (e.g., `"my_database"`)
   - **AND** the `NEO4J_DATABASE` environment variable is not set
   - In this case, the database will be incorrectly treated as token, and the actual database will use the default value `"neo4j"`

## Problem Analysis

The correct parameter order for `Neo4jGraph.__init__` is:
```python
def __init__(
    self,
    url: Optional[str] = None,
    username: Optional[str] = None,
    password: Optional[str] = None,
    token: Optional[str] = None,      # 4th parameter
    database: Optional[str] = None,     # 5th parameter
    timeout: Optional[float] = None,
    sanitize: bool = False,
    refresh_schema: bool = True,
    *,
    driver_config: Optional[Dict] = None,
    enhanced_schema: bool = False,
) -> None:
```

**Problem**:
- `database` is passed as the 4th positional argument, but the 4th parameter should be `token`
- This causes the `database` value to be incorrectly passed to the `token` parameter
- The actual `database` parameter will use the default value `"neo4j"` (see `Neo4jGraph.__init__` lines 150-152)

**Impact**:
- If the user configures a database name other than `"neo4j"`, it will be ignored
- Always uses the default `"neo4j"` database, even if a different database name is configured

## Suggested Fix

```python
self.graph = Neo4jGraph(
    url=self.config.graph_store.config.url,
    username=self.config.graph_store.config.username,
    password=self.config.graph_store.config.password,
    token=None,  # Explicitly pass token=None
    database=self.config.graph_store.config.database,  # Use keyword argument
    refresh_schema=False,
    driver_config={"notifications_min_severity": "OFF"},
)
```

Or a more concise approach (since token defaults to None):

```python
self.graph = Neo4jGraph(
    url=self.config.graph_store.config.url,
    username=self.config.graph_store.config.username,
    password=self.config.graph_store.config.password,
    database=self.config.graph_store.config.database,  # Use keyword argument
    refresh_schema=False,
    driver_config={"notifications_min_severity": "OFF"},
)
```

## Environment Information

- mem0ai version: 1.0.1
- Python version: 3.12
- langchain-neo4j version: 0.7.0

## Reproduction Steps

1. Configure mem0 to use Neo4j graph store, and set `database` to a value other than `"neo4j"` (e.g., `"my_database"`)
2. Initialize a `Memory` object and configure graph_store
3. Observe the actual database name being connected to

## Expected Behavior

Should connect to the database specified in the configuration.

## Actual Behavior

Always connects to the default `"neo4j"` database, ignoring the `database` value in the configuration.

veeceey:
## Summary
- `Neo4jGraph.__init__` signature is `(url, username, password, token, database, ...)`
- `MemoryGraph` was passing `database` as the 4th positional arg, mapping it to `token`
- This caused silent failures when using custom database names without the env var fallback

## Changes
- Changed `database` from positional to keyword argument in `Neo4jGraph()` call
- Added test verifying `database` is passed as keyword argument

## Test plan
- [x] Unit test mocks `Neo4jGraph` and verifies arg/kwarg assignment
- [x] Test passes on both default and custom database names

Fixes #3906


veeceey:
## Manual Test Results

Tested locally on macOS, Python 3.14.2.

### Bug Verification

**Before fix** — `database` was passed as a positional argument (4th position), which mapped to the `token` parameter slot in the Neo4j driver constructor:
```python
Neo4jGraph.__init__() → GraphDatabase.driver(url, auth=(user, password), database)
#                                                                        ^^^^^^^^
# 4th positional arg = token slot, NOT database
```

**After fix** — `database` is passed as a keyword argument and stored for use in sessions:
```python
Neo4jGraph(url, username, password, database="mydb")
# database correctly passed as keyword, used in session creation
```

### Unit Test

1 unit test defined verifying that `database` is passed as a keyword argument rather than a positional argument. Test passes.

veeceey:
Hi maintainers, friendly ping on this PR. Manual test results have been provided showing the bug fix. Would appreciate a review when you get a chance. Thank you!

kartik-mem0:
@veeceey thank you so much for the contribution. 

but we favoured #3913 pr over your solution.

Issue #3767, opened 2025-11-18
Type Mismatch in Langchain Vector Store Adapter's update Method


firedai:
## Description
There is a type mismatch issue in the update method of the mem0/vector_stores/langchain.py file when passing parameters to the insert method.

## Location
mem0/vector_stores/langchain.py Method: update Line: approximately line 118

## Details
When the update method calls the insert method, the vector and payload parameters are not wrapped in lists, but the insert method's parameter definitions require them to be lists:

# Current problematic code
def update(self, vector_id, vector=None, payload=None):
    """
    Update a vector and its payload.
    """
    self.delete(vector_id)
    self.insert(vector, payload, [vector_id])  # Here payload and vector are not wrapped in []

# insert method parameter definitions
def insert(
    self, vectors: List[List[float]], payloads: Optional[List[Dict]] = None, ids: Optional[List[str]] = None
):

## Suggested Fix
The vector and payload parameters should be wrapped in lists:

def update(self, vector_id, vector=None, payload=None):
    """
    Update a vector and its payload.
    """
    self.delete(vector_id)
    self.insert([vector], [payload], [vector_id])  # Fixed: added []

## Environment Information
- mem0 version: 1.0.0
- Python version: 3.10

abhay084:
Hey @firedai , raised a fix for this here, please review: https://github.com/mem0ai/mem0/pull/3772

PR #3983, opened 2026-02-06
fix: wrap vector and payload in lists in langchain adapter update method

thecaptain789:
The `insert` method expects lists for `vectors` and `payloads` parameters, but the `update` method was passing single values. This caused type mismatch errors when updating vectors in the Langchain vector store adapter.

Closes #3767


kartik-mem0:
Hi @thecaptain789, thanks for the contribution! This issue has been addressed via #4446. Closing this PR — appreciate your effort! 🙏

PR #3985, opened 2026-02-06
fix: preserve all metadata fields when updating memory

### LINKED ISSUE #3966: MongoDB Implementation Bug: update replaces metadata entirely
### 🐛 Describe the bug 

https://github.com/mem0ai/mem0/issues/3966

# Mem0 Update Payload Replacement Issue

## Problem

Mem0's MongoDB vector store `update()` method **replaces the entire `payload` field** when updating a memory, causing **complete loss of metadata** if not handled correctly.

### Root Cause

Mem0's vector store implementation does:

```python
def update(self, vector_id: str, vector: Optional[List[float]] = None, payload: Optional[Dict] = None) -> None:
    update_fields = {}
    if vector is not None:
        update_fields["embedding"] = vector
    if payload is not None:
        update_fields["payload"] = payload  # ⚠️ REPLACES ENTIRE PAYLOAD

    if update_fields:
        self.collection.update_one({"_id": vector_id}, {"$set": update_fields})
```

When Mem0's `memory.update()` is called, it internally calls the vector store's `update()` method with a new payload, which results in:

```javascript
// MongoDB operation
{"$set": {"payload": new_payload}}  // ⚠️ This REPLACES the entire payload object
```

**Result:** All existing metadata fields (`bucket_id`, `context_id`, `category`, `source`, etc.) are **completely lost** unless explicitly preserved.

## Impact

- **Metadata Loss**: All custom metadata fields are wiped when updating memory content
- **Data Integrity**: Critical fields like `bucket_id`, `context_id`, `category` disappear
- **User Experience**: UI shows "Ungrouped" memories, filters break, context is lost

### Update memory content (without metadata preservation)

**Without our fix**, if you call Mem0 directly:

```python
# ⚠️ THIS LOSES METADATA
from mem0 import Memory

mem0 = Memory.from_config({...})
mem0.update(memory_id=memory_id, data="User loves chocolate and vanilla")

# Check metadata - ALL GONE!
updated = memory_service.get(memory_id=memory_id, user_id="user123")
print(f"Metadata after update: {updated['metadata']}")
# Output: {}  # ⚠️ EMPTY! All metadata lost!
```

### 3. Verify metadata loss

```python
# Fetch the memory
memory = memory_service.get(memory_id=memory_id, user_id="user123")

# Check metadata
assert "bucket_id" not in memory.get("metadata", {}), "Metadata should be lost!"
assert "source" not in memory.get("metadata", {}), "Metadata should be lost!"
# These assertions will PASS (metadata is gone)
```

### Option 1: Patch Mem0's Vector Store (Upstream Fix)

**Ideal Solution**: Modify Mem0's MongoDB vector store to merge payload instead of replacing:

```python
# In mem0/vector_stores/mongodb.py
def update(self, vector_id: str, vector: Optional[List[float]] = None, payload: Optional[Dict] = None) -> None:
    update_fields = {}
    if vector is not None:
        update_fields["embedding"] = vector
    if payload is not None:
        # ✅ MERGE instead of REPLACE
        for k, v in payload.items():
            update_fields[f"payload.{k}"] = v
        # Instead of: update_fields["payload"] = payload
    
    if update_fields:
        self.collection.update_one({"_id": vector_id}, {"$set": update_fields})
```

**Pros:**
- Fixes the issue at the source
- Benefits all Mem0 users
- Cleaner architecture

thecaptain789:
## Summary
Fixes #3966

## Problem
When updating a memory via `memory.update()`, custom metadata fields like `bucket_id`, `context_id`, `category`, `source`, etc. are lost because `_update_memory` only preserves a whitelist of known fields (`user_id`, `agent_id`, `run_id`, `actor_id`, `role`).

## Solution
Instead of starting with an empty dict and selectively copying known fields, start with the existing payload (preserving all metadata) and then update only the managed fields (`data`, `hash`, `updated_at`).

This is simpler, more maintainable, and correctly preserves all custom metadata fields.

## Changes
- Start `new_metadata` from `deepcopy(existing_memory.payload)` instead of empty dict
- Merge in any explicitly provided metadata
- Remove the per-field preservation logic (no longer needed)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1150,24 +1150,17 @@ class Memory(MemoryBase):
 
         prev_value = existing_memory.payload.get("data")
 
-        new_metadata = deepcopy(metadata) if metadata is not None else {}
+        # Start with existing payload to preserve all metadata fields (including custom ones)
+        new_metadata = deepcopy(existing_memory.payload)
 
+        # Override with any explicitly provided metadata
+        if metadata is not None:
+            new_metadata.update(metadata)
+
+        # Update managed fields
         new_metadata["data"] = data
         new_metadata["hash"] = hashlib.md5(data.encode()).hexdigest()
-        new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(pytz.timezone("US/Pacific")).isoformat()
-
-        # Preserve session identifiers from existing memory only if not provided in new metadata
-        if "user_id" not in new_metadata and "user_id" in existing_memory.payload:
-            new_metadata["user_id"] = existing_memory.payload["user_id"]
-        if "agent_id" not in new_metadata and "agent_id" in existing_memory.payload:
-            new_metadata["agent_id"] = existing_memory.payload["agent_id"]
-        if "run_id" not in new_metadata and "run_id" in existing_memory.payload:
-            new_metadata["run_id"] = existing_memory.payload["run_id"]
-        if "actor_id" not in new_metadata and "actor_id" in existing_memory.payload:
-            new_metadata["actor_id"] = existing_memory.payload["actor_id"]
-        if "role" not in new_metadata and "role" in existing_memory.payload:
-            new_metadata["role"] = existing_memory.payload["role"]
 
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]


kartik-mem0:
Hi @thecaptain789, thank you for your work on this fix for #3966! 🙏

This issue has been resolved via #4495 (merged), which preserves custom metadata during memory updates. Since the fix is now in main, we're closing this PR.

Thanks for taking the time to contribute — we appreciate it!

PR #3986, opened 2026-02-07

thecaptain789:
## Summary
Fixes #3580

The `_safe_deepcopy_config` function was using `'auth'` as a sensitive token, which incorrectly matched `http_auth` and broke AWS-signed OpenSearch client authentication.

## Changes
- Updated sensitive token patterns to be more specific
- Split into `sensitive_tokens` (truly sensitive data like passwords, secrets, api_keys) and `runtime_tokens` (non-serializable objects like connection_class)
- Preserves `http_auth` and similar runtime authentication objects required for vector store connections

## Testing
Verified the change preserves `http_auth` field while still sanitizing truly sensitive fields like `password`, `secret`, and `api_key`.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,9 +69,16 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @thecaptain789 , 

Thanks again for your effort!

PR #3989, opened 2026-02-07
fix: correct typo 'existance' to 'existence'

thecaptain789:
Fixes a minor typo in the weaviate vectordb docstring.

- Changed 'existance' to 'existence' in the `get` method docstring


kartik-mem0:
Thanks for the typo fix. This targets embedchain which is legacy code, and the CLA hasn't been signed. Closing it out.

PR #4034, opened 2026-02-12
Fix vector update handling for None vector values

YukunWuSm:
Handle case where vector is None by retrieving the original vector from Milvus before updating.

## Description

When performing an upsert in Milvus, if the vector field is None, an error will occur. Therefore, when updating, if the vector is None, first query the original vector from Milvus before proceeding with the update.

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Local OSS test: when the memory content does not change, mem0 will create a NONE event and set the vector to None.
```python
#main.py
self.vector_store.update(
    vector_id=memory_id,
    vector=None,  # Keep same embeddings
    payload=updated_metadata,
)
```
At this point, calling the update() method will throw an error because Milvus does not allow the vector to be None.
After the fix, mem0 will first query the original vector from Milvus and then perform the update, which prevents the error.

## Checklist:

### DIFF
--- a/mem0/vector_stores/milvus.py
+++ b/mem0/vector_stores/milvus.py
@@ -181,8 +181,20 @@ class MilvusDB(VectorStoreBase):
             vector (List[float], optional): Updated vector.
             payload (Dict, optional): Updated payload.
         """
-        schema = {"id": vector_id, "vectors": vector, "metadata": payload}
-        self.client.upsert(collection_name=self.collection_name, data=schema)
+        if vector is None:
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @YukunWuSm ,

Thanks for taking the time to come up with the PR. We appreciate it.
We are going to close this PR as the we have come up with a more comprehensive PR #4568 which covers additional bugs. 

Thank you.

PR #4045, opened 2026-02-13
feat: add Anthropic structured outputs via extended responseFormat

elvismdev:
## Description

Added Anthropic structured outputs support for Claude 4.x models with a robust fallback chain that eliminates silent failures.

### Changes

- Upgraded `@anthropic-ai/sdk` peer dependency from `^0.40.1` to `^0.74.0`
- Upgraded `zod` dependency from `^3.24.1` to `^3.25.0` (SDK peer requirement)
- Extended `ResponseFormat` type in `base.ts` with optional `schema?: z.ZodType` and `jsonSchema?: Record<string, unknown> & { type: "object" }` fields
- Added `supportsStructuredOutputs(model)` — returns `true` for Claude 4.x family (opus-4, sonnet-4, haiku-4 prefixes)
- **Primary path:** `jsonSchemaOutputFormat` from `@anthropic-ai/sdk/helpers/json-schema` — has zero zod runtime dependency, so it works regardless of which zod version is resolved at runtime
- **Fallback:** `zodOutputFormat` from `@anthropic-ai/sdk/helpers/zod` — used when `jsonSchema` is not provided (backward compat)
- **Observability:** `console.warn` logged at each fallback step so failures are visible instead of silently swallowed
- Added raw JSON schema constants (`FactRetrievalJsonSchema`, `MemoryUpdateJsonSchema`) kept in sync with the existing Zod schemas
- Memory layer passes both `schema` (Zod) and `jsonSchema` (raw) at both call sites (fact extraction and memory update)

### Fallback chain

```
jsonSchemaOutputFormat(jsonSchema)   ← primary, zero zod dependency
  ↓ (on failure, with console.warn)
zodOutputFormat(schema)              ← fallback, needs zod >= 3.25.0
  ↓ (on failure, with console.warn)
prompt-based JSON                    ← last resort, no structured output
```

### Why `jsonSchemaOutputFormat` instead of just `zodOutputFormat`?

`zodOutputFormat` internally calls `z.toJSONSchema()` which only exists in zod >= 3.25.0. When the runtime resolves an older zod (e.g. in environments with shared dependencies like VS Code extensions), the call throws and was previously silently swallowed. `jsonSchemaOutputFormat` accepts a plain JSON Schema object and has no zod dependency at all, making it the more robust primary path.

### Design decisions

- Extended the existing `responseFormat` type (following Ollama's internal format translation pattern) rather than creating a separate provider. Unlike `openai_structured.ts` (which exists because OpenAI uses a different API method `beta.chat.completions.parse()`), Anthropic's structured outputs use the same `messages.create()` with an additional `output_config` parameter.
- Other providers naturally ignore the `schema` and `jsonSchema` fields — no changes required.
- The `as const` assertions on raw JSON schemas give TypeScript the literal types that `jsonSchemaOutputFormat` expects.

## Type of change

## How Has This Been Tested?

- [x] Unit Test

8 unit tests verify:
- `supportsStructuredOutputs` returns true for Claude 4.x, false for 3.x (2 tests)
- `jsonSchemaOutputFormat` used as primary path when both schemas provided
- Falls back to `zodOutputFormat` when `jsonSchema` not provided
- Falls back to `zodOutputFormat` when `jsonSchemaOutputFormat` throws
- Warns when both structured output methods fail (no `output_config` set)
- No `output_config` on unsupported model (neither helper called)
- No `output_config` when no schema provided (neither helper called)

Existing factory tests also pass (no regressions from SDK upgrade).

Run: `cd mem0-ts && npx jest --testPathPattern="anthropic-structured-outputs"`

## Checklist:

### DIFF
--- a/mem0-ts/src/oss/src/llms/anthropic.ts
+++ b/mem0-ts/src/oss/src/llms/anthropic.ts
@@ -1,4 +1,5 @@
 import Anthropic from "@anthropic-ai/sdk";
+import { jsonSchemaOutputFormat } from "@anthropic-ai/sdk/helpers/json-schema";
 import { zodOutputFormat } from "@anthropic-ai/sdk/helpers/zod";
 import { LLM, LLMResponse, ResponseFormat } from "./base";
 import { LLMConfig, Message } from "../types";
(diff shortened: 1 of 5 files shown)


kartik-mem0:
Hi — the Anthropic structured outputs feature includes major dependency upgrades and hasn't been reviewed. Closing this for now. If structured output support is still needed, a more focused PR would be easier to review.

PR #4062, opened 2026-02-17
fix: preserve runtime authentication objects in _safe_deepcopy_config

jackjin1997:
Fixes #3580. The current sanitizer in _safe_deepcopy_config is overly broad and removes 'http_auth' which is required for OpenSearch AWS-signed clients. This PR narrows the sensitive tokens list to preserve runtime auth objects while still stripping genuinely sensitive data like passwords and tokens.


kartik-mem0:

Closing this PR.

PR #4078, opened 2026-02-19
fix(oss): propagate historyDbPath into historyStore config

### LINKED ISSUE #4074: Node.js OSS SDK: historyDbPath config ignored when historyStore default takes precedence
## Bug Description

The `openclaw-mem0` plugin passes `historyDbPath` in the config to `new Memory(config)`, but the SDK's `Memory` constructor ignores it because the default `historyStore` config takes precedence.

## Root Cause

In `src/oss/src/memory.ts` (or the compiled `dist/oss/index.js`), the `Memory` constructor at ~line 4574 checks:

```js
  ? HistoryManagerFactory.create(this.config.historyStore.provider, this.config.historyStore)
  : HistoryManagerFactory.create("sqlite", defaultConfig);
```

The `defaultConfig` correctly uses `this.config.historyDbPath`, but it's only reached when `this.config.historyStore` is falsy. However, `ConfigManager.mergeConfig()` always populates `historyStore` from `DEFAULT_MEMORY_CONFIG`:

```js
historyStore: {
  ...userConfig.historyStore              // undefined when not provided by user
}
```

So `this.config.historyStore` is always truthy (from the default), and its hardcoded `"memory.db"` relative path is used instead of the user-provided `historyDbPath`.

## Impact

- The top-level `historyDbPath` config option is effectively dead — it's stored in the merged config but never actually used
- The SDK always tries to open `memory.db` relative to the process working directory
- In containerized environments (like OpenClaw's Docker setup), the working directory (`/app/`) is often read-only for the runtime user, causing `SQLITE_CANTOPEN: unable to open database file`
- The openclaw-mem0 plugin correctly reads `historyDbPath` from its config and passes it through, but the SDK silently ignores it

## Suggested Fix

Either:

1. **In the `Memory` constructor**: Use `historyDbPath` to override `historyStore.config.historyDbPath` when both are present:
```js
}
```

2. **In `ConfigManager.mergeConfig()`**: Don't populate `historyStore` from defaults when user provides `historyDbPath`:
```js
historyStore: userConfig.historyStore || (userConfig.historyDbPath
  ? { provider: "sqlite", config: { historyDbPath: userConfig.historyDbPath } }
  : DEFAULT_MEMORY_CONFIG.historyStore)
```

## Reproduction

```js
const m = new Memory({
  historyDbPath: "/tmp/my-history.db",
  // ... other config
});
// Expected: SDK uses /tmp/my-history.db
// Actual: SDK uses memory.db (relative to cwd)
```

## Environment

- `mem0ai` npm package version: 2.2.1
- Node.js: v22.x
- Running inside Docker container where cwd (`/app/`) is not writable by the runtime user

ericpearson:
## Summary

Fixes #4074

When a user provides `historyDbPath` in their config, it is silently ignored because `DEFAULT_MEMORY_CONFIG.historyStore` always populates `historyStore` with a hardcoded `historyDbPath: "memory.db"`. The `Memory` constructor then uses `historyStore` (always truthy from defaults) instead of the user's `historyDbPath`.

This change propagates the user-provided `historyDbPath` into `historyStore.config.historyDbPath` when no explicit `historyStore` override is provided.

## Changes

- `mem0-ts/src/oss/src/config/manager.ts`: When merging config, if `userConfig.historyDbPath` is set and `userConfig.historyStore` is not, inject the user's path into `historyStore.config.historyDbPath`

## Test plan

- Verify that `new Memory({ historyDbPath: "/custom/path.db" })` uses `/custom/path.db` instead of `memory.db`
- Verify that providing an explicit `historyStore` config still takes full precedence
- Verify that omitting both `historyDbPath` and `historyStore` still defaults to `memory.db`

### DIFF
--- a/mem0-ts/src/oss/src/config/manager.ts
+++ b/mem0-ts/src/oss/src/config/manager.ts
@@ -104,6 +104,14 @@ export class ConfigManager {
       historyStore: {
         ...DEFAULT_MEMORY_CONFIG.historyStore,
         ...userConfig.historyStore,
+        ...(userConfig.historyDbPath && !userConfig.historyStore
+          ? {
+              config: {
+                ...DEFAULT_MEMORY_CONFIG.historyStore.config,
+                historyDbPath: userConfig.historyDbPath,
+              },
+            }
+          : {}),
       },
       disableHistory:
         userConfig.disableHistory || DEFAULT_MEMORY_CONFIG.disableHistory,


ericpearson:
I signed the agreement but it's not updating. Can somebody please merge this?

PR #4085, opened 2026-02-20
fix(graph): use keyword arguments when initializing Neo4jGraph to pre…

wangxvdong:
fix(graph): use keyword arguments when initializing Neo4jGraph to prevent token misassignment

Previously, positional arguments were used when initializing Neo4jGraph.
Due to the constructor signature (url, username, password, token, database),
the database argument was incorrectly passed as `token`, causing Neo4j to use
bearer authentication instead of basic authentication.

This change switches to explicit keyword arguments to ensure correct parameter
mapping and prevent authentication errors.

### DIFF
--- a/mem0/memory/graph_memory.py
+++ b/mem0/memory/graph_memory.py
@@ -30,10 +30,11 @@ class MemoryGraph:
     def __init__(self, config):
         self.config = config
         self.graph = Neo4jGraph(
-            self.config.graph_store.config.url,
-            self.config.graph_store.config.username,
-            self.config.graph_store.config.password,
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for the fix, @wangxvdong. Heads up: the OSS graph layer was removed from the SDK in the v3 pipeline port (#4805) — `mem0/memory/graph_memory.py`, where `Neo4jGraph` was initialized, no longer exists on `main`, so there's nothing for this patch to apply to anymore (graph memory now lives on the hosted-platform side). Closing as obsolete — genuinely appreciate the contribution.

PR #4090, opened 2026-02-20
fix: ensure prompts include 'json' for json_object response format

### LINKED ISSUE #3559: Error 400 'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.
### 🐛 Describe the bug

I got this error : 

`BadRequestError: 400 'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.`

When i tried to set my own customPrompt.

using typescript.

mem0ai:2.1.38

hztBUAA:
## Summary
Fixes #3559

When using `custom_fact_extraction_prompt` with `response_format: { type: "json_object" }`, OpenAI's API requires that the messages contain the word "json" in some form. If a user provides a custom prompt that doesn't include this word, the API returns:

> Error 400: 'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.

This fix adds a check in the Python SDK (both sync `Memory` and async `AsyncMemory`) to detect when the custom fact extraction prompt doesn't contain "json" (case-insensitive) and automatically appends a JSON format instruction. This matches the fix already applied in the TypeScript SDK in commit 394203d1.

## Changes
- **`mem0/memory/main.py`**: Added check in both `Memory._add_to_vector_store` and `AsyncMemory._add_to_vector_store` to append JSON format instruction to `custom_fact_extraction_prompt` when "json" is not already present in the prompt
- **`tests/test_main.py`**: Updated `test_custom_prompts` to expect the appended JSON instruction when the custom prompt doesn't contain "json"

## Test Plan
- [x] Existing `test_custom_prompts` test updated and passes
- [x] All 13 tests in `tests/test_main.py` pass
- [x] Custom prompts containing "json" are not modified (no-op)
- [x] Custom prompts without "json" get the instruction appended

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -424,6 +424,8 @@ class Memory(MemoryBase):
 
         if self.config.custom_fact_extraction_prompt:
             system_prompt = self.config.custom_fact_extraction_prompt
+            if "json" not in system_prompt.lower():
+                system_prompt += "\nYou must respond with a valid JSON object containing a 'facts' key with an array of strings."
             user_prompt = f"Input:\n{parsed_messages}"
         else:
             # Determine if this should use agent memory extraction based on agent_id presence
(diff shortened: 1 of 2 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

kartik-mem0:
thank you for the contribution.

but we favored this solution #4271

PR #4091, opened 2026-02-20
fix: preserve http_auth in _safe_deepcopy_config for OpenSearch

hztBUAA:
## Summary

Fixes #3580

- `_safe_deepcopy_config()` was using an overly broad `"auth"` token in its sensitive field sanitizer, which matched `http_auth` and stripped runtime authentication objects (like `AWSV4SignerAuth`) needed for OpenSearch connections
- Replaced the broad `"auth"` and `"connection_class"` tokens with a `preserve_fields` allowlist that restores runtime objects from the original config, while still nullifying genuinely sensitive plaintext data (passwords, API keys, credentials)

### Root Cause

The sensitive_tokens tuple included `"auth"`, which matched `http_auth` via substring check (`"auth" in "http_auth"`), causing it to be set to `None`. Similarly, `"connection_class"` was explicitly listed, removing the connection class needed for OpenSearch.

### Changes

- **`mem0/memory/main.py`**: Added `preserve_fields = {"http_auth", "connection_class"}` allowlist. Fields in this set are restored from the original config object. Removed `"auth"` and `"connection_class"` from `sensitive_tokens`.
- **`tests/vector_stores/test_opensearch.py`**: Updated existing test assertions to verify `http_auth` and `connection_class` are preserved (not nullified). Added new test using the actual `OpenSearchConfig` pydantic model.

## Test plan

- [x] All 13 existing OpenSearch tests pass
- [x] New test `test_safe_deepcopy_preserves_http_auth_on_real_config` verifies the fix with the actual `OpenSearchConfig` model
- [x] Sensitive plaintext fields (`password`, `api_key`, `credentials`) are still correctly nullified

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,9 +69,15 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
-        sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
+        # Preserve runtime objects needed for connection (e.g. http_auth, connection_class)
+        # while stripping genuinely sensitive plaintext data (passwords, API keys, etc.)
+        preserve_fields = {"http_auth", "connection_class"}
+        sensitive_tokens = ("credential", "password", "token", "secret", "key")
         for field_name in list(clone_dict.keys()):
-            if any(token in field_name.lower() for token in sensitive_tokens):
+            if field_name in preserve_fields:
+                # Restore original runtime objects that may have been lost during serialization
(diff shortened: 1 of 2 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

kartik-mem0:
hi @hztBUAA can you please sign the cla?

utkarsh240799:
Hi @hztBUAA , 

Thanks again for your effort!

PR #3997, opened 2026-02-08
Fix: add pgvector support to NodeJS OSS VectorStoreFactory (fixes #3491)

veeceey:
## Summary
This PR fixes issue #3491 where pgvector was not recognized as a supported vector store provider in the NodeJS OSS implementation, despite having a complete `PGVector` class.

## Problem
Users attempting to use pgvector as their vector store provider would encounter the error:
```
Unsupported vector store provider: pgvector
```

This occurred even though:
- The `PGVector` class was fully implemented in `src/oss/src/vector_stores/pgvector.ts`
- Documentation mentioned pgvector support
- An example was provided in `src/oss/examples/vector-stores/pgvector.ts`

## Root Cause
The `PGVector` class existed but was never registered in the `VectorStoreFactory`. The factory's switch statement was missing both:
1. An import for the `PGVector` class
2. A `case "pgvector"` entry to instantiate it

## Solution
1. Added `import { PGVector } from "../vector_stores/pgvector"` to `factory.ts`
2. Added `case "pgvector": return new PGVector(config as any);` to the factory switch statement
3. Added test case `should create pgvector vector store` to verify functionality

## Testing
- Added unit test to verify `VectorStoreFactory.create("pgvector", config)` returns a `PGVector` instance
- Follows same pattern as existing vector store tests (Azure AI Search, memory)
- CI will verify no regressions

## Impact
- **Fixes**: Users can now use pgvector as documented
- **Backwards compatible**: No breaking changes
- **Simple fix**: 2 lines + test coverage

Closes #3491

PR #4097, opened 2026-02-20
fix(ts): register PGVector in VectorStoreFactory switch

etarley:
## Problem

The `PGVector` class exists at `mem0-ts/src/oss/src/vector_stores/pgvector.ts` and is documented as a supported provider (with TypeScript examples in the docs), but it was never added to the `VectorStoreFactory` switch statement.

This means any user configuring `provider: 'pgvector'` in TypeScript gets:
```
Error: Unsupported vector store provider: pgvector
```

## Fix

Two lines in `factory.ts`:
1. Add import for `PGVector`
2. Add `case "pgvector"` to `VectorStoreFactory.create()`

## Testing

Verified locally against PostgreSQL 18 with pgvector extension — initialization, insert, and vector similarity search all work correctly with the registered provider.

## Notes

The Python SDK already supports pgvector. This brings the TypeScript SDK to parity with the docs and Python SDK.


utkarsh240799:
Hi @etarley . 
Thank you for your contribution!
PR #3997  has already been merged. Closing this PR as duplicate.

PR #4101, opened 2026-02-21
fix(mem0-ts): normalize Ollama model name before local existence check

mgoulart:
## Problem

`OllamaLLM` and `OllamaEmbedder` both call `ensureModelExists()` which compares the configured model name against `ollama.list()` using strict equality:

```ts
local_models.models.find((m: any) => m.name === this.model)
```

Ollama's list API always returns names with an explicit tag — e.g. a model stored as `mymodel` appears as `mymodel:latest`. When a user configures just `mymodel` (no tag), the comparison always fails, causing the code to call `ollama.pull({ model: "mymodel" })`.

For models that exist only locally — created via `ollama cp` or a custom Modelfile — this pull fails with:

```
ResponseError: pull model manifest: file does not exist
```

The error is caught and retried (3×), adding unnecessary HTTP round-trips on every cold init, and logging confusing errors even though the model is already available and works fine.

## Fix

Normalize the configured name by appending `:latest` when no tag is present before comparing — matching how Ollama itself resolves bare names:

```ts
const normalize = (name: string) => (name.includes(":") ? name : `${name}:latest`);
if (!local_models.models.find((m: any) => m.name === normalize(this.model))) {
```

Applied to both `OllamaLLM` and `OllamaEmbedder`.

## Reproduction

```ts
// ollama cp llama3.1:8b my-local-model
// Then configure mem0 with model: "my-local-model"
// → triggers 3x failed pull attempts on every Memory() instantiation
```

## Notes

- No behaviour change for models pulled from the registry (they already include a tag)
- Fixes locally-aliased models and custom Modelfile builds

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/ollama.ts
+++ b/mem0-ts/src/oss/src/embeddings/ollama.ts
@@ -44,7 +44,12 @@ export class OllamaEmbedder implements Embedder {
       return true;
     }
     const local_models = await this.ollama.list();
-    if (!local_models.models.find((m: any) => m.name === this.model)) {
+    // Ollama list returns names with an implicit :latest tag (e.g. "llama3.1:8b"
+    // stays as-is, but "mymodel" becomes "mymodel:latest"). Normalize the
+    // configured model name the same way before comparing so locally-created
+    // aliases (via `ollama cp` or custom Modelfiles) are found without pulling.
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hi, this has merge conflicts and the CLA hasn't been signed. Closing it for now — if you'd like to resolve the conflicts and sign the CLA, feel free to open a new PR. Thanks for the contribution.

PR #4105, opened 2026-02-21
fix(qdrant): handle 409 conflict in getUserId() for migration idempotency

AlphaComposite:
## Summary

- **`getUserId()`** calls `createCollection("memory_migrations")` without handling HTTP 409 ("already exists"). When the collection exists from a prior run — or when `initialize()` and `getUserId()` race — the unhandled 409 propagates up and collapses all downstream `recall`/`capture` operations with "Bad Request".
- **`initialize()`** already handles 409 correctly (lines 340-346 in the current source), but `getUserId()` does not, and they can execute concurrently because the constructor fires `initialize()` as fire-and-forget.

## Changes

1. **Wrap `createCollection` in `getUserId()` with 409 handling** — matches the existing pattern in `initialize()`. Non-409 errors still propagate.
2. **Store the `initialize()` promise and `await` it in `getUserId()`** — eliminates the TOCTOU race between init and first use.

## Reproduction

```js
import { QdrantClient } from "@qdrant/js-client-rest";
const client = new QdrantClient({ host: "localhost", port: 6333 });

// Ensure collection exists (simulates prior run)
await client.createCollection("memory_migrations", {
  vectors: { size: 1, distance: "Cosine" },
});

// This is what getUserId() does — crashes with 409
await client.createCollection("memory_migrations", {
  vectors: { size: 1, distance: "Cosine", on_disk: false },
});
// Error: status=409, message="Conflict"
```

## Test plan

- [x] Verified `createCollection` returns 409 when collection exists
- [x] Verified patched code treats 409 as success and continues
- [x] Verified concurrent `initialize()` + `getUserId()` does not crash
- [x] Verified 5 concurrent `getUserId()` calls don't crash
- [x] Verified non-409 errors still propagate correctly
- [x] Tested against Qdrant 1.17.0 with `@qdrant/js-client-rest` 1.17.0

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/qdrant.ts
+++ b/mem0-ts/src/oss/src/vector_stores/qdrant.ts
@@ -32,6 +32,7 @@ export class Qdrant implements VectorStore {
   private client: QdrantClient;
   private readonly collectionName: string;
   private dimension: number;
+  private _initPromise: Promise<void>;
 
   constructor(config: QdrantConfig) {
     if (config.client) {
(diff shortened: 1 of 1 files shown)


kartik-mem0:

PR #4120, opened 2026-02-23
OSS Node SDK: honor historyDbPath for default sqlite history store

### LINKED ISSUE #4096: Node SDK OSS: historyDbPath config ignored, SQLite history always opens at default memory.db
## Bug

When using the Node SDK in OSS mode with a custom `historyDbPath`, the configured path is stored in the merged config but never used. The SQLite history DB always opens at the default relative path `memory.db` in `process.cwd()`.

This crashes with `SQLITE_CANTOPEN` when `process.cwd()` is not writable — which is common in daemon/service contexts where cwd is `/`.

Related: #1600 reported the same structural issue in the Python SDK (SQLiteManager always created regardless of vector store config). This is the Node SDK equivalent, but specifically about the history DB path being silently ignored.

## Environment

- `mem0ai` v2.2.3 (Node SDK, `mem0ai/oss`)
- Node v25.6.0
- Vector store: Qdrant (not using default SQLite vector store)

## Reproduction

```js
process.chdir("/");

const m = new Memory({
  version: "v1.1",
  vectorStore: {
    provider: "qdrant",
    config: { host: "localhost", port: 6333, collectionName: "memories" },
  },
  historyDbPath: "/some/writable/path/history.db",
});
// SQLITE_CANTOPEN: unable to open database file
// Opens /memory.db instead of the configured path
```

## Root cause

Two things interact in `ConfigManager.mergeConfig()` and the `Memory` constructor:

**1. `mergeConfig` always populates `historyStore`**

```js
// ConfigManager.mergeConfig
historyStore: {
  ...userConfig.historyStore              // undefined when user only sets historyDbPath
}
```

Spreading `undefined` is a no-op, so `mergedConfig.historyStore` is always the default object.

**2. Memory constructor always takes the historyStore branch**

```js
// Memory constructor
const defaultConfig = {
  provider: "sqlite",
  config: { historyDbPath: this.config.historyDbPath || ":memory:" }  // has the right path
};

  ? HistoryManagerFactory.create(this.config.historyStore.provider, this.config.historyStore)  // always taken, uses "memory.db"
  : HistoryManagerFactory.create("sqlite", defaultConfig);  // never reached
```

Because `historyStore` is always truthy after merge, the first branch always wins. The `defaultConfig` that actually reads `historyDbPath` is dead code.

## Suggested fix

If `historyDbPath` is provided, it should override `historyStore.config.historyDbPath`:

```js
}
```

Or: only spread `DEFAULT_MEMORY_CONFIG.historyStore` when the user actually provides a `historyStore` config.

## Workaround

`disableHistory: true` skips SQLite entirely and uses `DummyHistoryManager`.

vamgan:
Fixes #4096\n\nWhen users set top-level historyDbPath (without providing historyStore), the merged config always included the default sqlite historyStore with historyDbPath="memory.db". That caused Memory to open the history DB in process.cwd().\n\nThis change propagates userConfig.historyDbPath into the default sqlite historyStore config, and adds a unit test to prevent regressions.

### DIFF
--- a/mem0-ts/src/oss/src/config/manager.ts
+++ b/mem0-ts/src/oss/src/config/manager.ts
@@ -101,10 +101,24 @@ export class ConfigManager {
         ...DEFAULT_MEMORY_CONFIG.graphStore,
         ...userConfig.graphStore,
       },
-      historyStore: {
-        ...DEFAULT_MEMORY_CONFIG.historyStore,
-        ...userConfig.historyStore,
-      },
+      historyStore: (() => {
+        const historyStore = {
+          ...DEFAULT_MEMORY_CONFIG.historyStore,
+          ...userConfig.historyStore,
+        };
+
+        // Ensure top-level historyDbPath is respected even when the default
+        // sqlite historyStore is present.
+        if (
+          userConfig.historyDbPath &&
+          historyStore?.provider === "sqlite" &&
+          historyStore.config
+        ) {
+          historyStore.config.historyDbPath = userConfig.historyDbPath;
+        }
+
+        return historyStore;
+      })(),
       disableHistory:
         userConfig.disableHistory || DEFAULT_MEMORY_CONFIG.disableHistory,
       enableGraph: userConfig.enableGraph || DEFAULT_MEMORY_CONFIG.enableGraph,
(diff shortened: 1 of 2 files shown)


kartik-mem0:
@vamgan thank you for the contribution.

closing this pr in favour of this #4307.

PR #4147, opened 2026-02-26
fix: only send temperature OR topP in Bedrock Converse API

### LINKED ISSUE #4004: Bug: AWS Bedrock Converse API rejects requests with both temperature and topP for Anthropic models
## Description

When using `aws_bedrock` as the LLM provider with Anthropic Claude models, every Converse API call fails with:

```
ValidationException: The model returned the following errors: `temperature` and `top_p` cannot both be specified for this model. Please use only one.
```

This affects newer Claude models on Bedrock (Claude 4.x, Haiku 4.5, etc.) which reject requests containing both parameters.

## Root Cause

In `mem0/llms/aws_bedrock.py`, both `_generate_with_tools()` and `_generate_standard()` unconditionally send both `temperature` and `topP` in the `inferenceConfig`:

```python
"inferenceConfig": {
    "maxTokens": self.model_config.get("max_tokens", 2000),
    "temperature": self.model_config.get("temperature", 0.1),
    "topP": self.model_config.get("top_p", 0.9),
}
```

Even if the user only configures `temperature` in their config, `top_p` defaults to `0.9` via `BaseLlmConfig` and is always included.

## Steps to Reproduce

```python
from mem0 import AsyncMemory

config = {
    "llm": {
        "provider": "aws_bedrock",
        "config": {
            "model": "us.anthropic.claude-haiku-4-5-20251001-v1:0",
            "temperature": 0.1,
        },
    },
    # ... other config
}

m = await AsyncMemory.from_config(config)
await m.add("some text", user_id="test")  # Fails with ValidationException
```

## Suggested Fix

Only include `topP` in `inferenceConfig` when the user has explicitly set it (i.e., it differs from the default), or omit it for Anthropic provider models. For example:

```python
inference_config = {
    "maxTokens": self.model_config.get("max_tokens", 2000),
}

# Only set one of temperature/topP, not both (some models reject both)
temperature = self.model_config.get("temperature")
top_p = self.model_config.get("top_p")

if temperature is not None:
    inference_config["temperature"] = temperature
elif top_p is not None:
    inference_config["topP"] = top_p
```

## Environment

- mem0 version: 0.1.118+
- Model: `us.anthropic.claude-haiku-4-5-20251001-v1:0` (also affects other newer Claude models)
- AWS Region: us-east-2
- boto3 Converse API

Anandesh-Sharma:
## Summary
- Fixes `ValidationException: temperature and top_p cannot both be specified` when using newer Claude models on AWS Bedrock
- Changes inferenceConfig construction to only include one of `temperature`/`topP`: temperature takes precedence if explicitly set, otherwise topP is used
- Applied to both `_generate_with_tools()` and `_generate_standard()` methods

## Test plan
- [ ] Verify Anthropic Claude models on Bedrock no longer throw ValidationException
- [ ] Verify temperature-only config works correctly
- [ ] Verify topP-only config works correctly
- [ ] Verify default behavior (temperature=0.1) when neither is explicitly set

Fixes #4004

### DIFF
--- a/mem0/llms/aws_bedrock.py
+++ b/mem0/llms/aws_bedrock.py
@@ -497,15 +497,25 @@ class AWSBedrockLLM(LLMBase):
             if converse_tools:
                 tool_config = {"tools": converse_tools}
 
+        # Prepare inference config — only include temperature OR topP, not both,
+        # as some models (e.g., Anthropic Claude on Bedrock) reject both together.
+        inference_config = {
+            "maxTokens": self.model_config.get("max_tokens", 2000),
+        }
+        temperature = self.model_config.get("temperature")
+        top_p = self.model_config.get("top_p")
+        if temperature is not None:
+            inference_config["temperature"] = temperature
+        elif top_p is not None:
+            inference_config["topP"] = top_p
+        else:
+            inference_config["temperature"] = 0.1
+
         # Prepare converse parameters
         converse_params = {
             "modelId": self.config.model,
             "messages": formatted_messages,
-            "inferenceConfig": {
-                "maxTokens": self.model_config.get("max_tokens", 2000),
-                "temperature": self.model_config.get("temperature", 0.1),
-                "topP": self.model_config.get("top_p", 0.9),
-            }
+            "inferenceConfig": inference_config,
         }
 
         # Add system message if present (for Anthropic)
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for the Bedrock temperature/topP fix. The CLA hasn't been signed and there are multiple PRs addressing the same issue. Closing this one — we'll address the Bedrock parameter handling internally.

PR #4150, opened 2026-02-26
fix: honor historyDbPath config in Node SDK OSS mode

### LINKED ISSUE #4074: Node.js OSS SDK: historyDbPath config ignored when historyStore default takes precedence

### LINKED ISSUE #4080: historyDbPath top-level config field is silently ignored (historyStore default always takes precedence)
## Bug

The top-level `historyDbPath` config field documented for `Memory` is effectively dead code. Any value passed there is silently ignored because the default config always wins.

## Root cause

In the `Memory` constructor, there are two code paths:

```python
# This runs and sets self.config.historyDbPath correctly...
defaultConfig = {
    "provider": "sqlite",
    "config": {"historyDbPath": self.config.historyDbPath or ":memory:"}
}

# ...but then this condition is always true because DEFAULT_MEMORY_CONFIG
# always defines historyStore, so the correctly-resolved defaultConfig is never used:
if self.config.historyStore and not self.config.disableHistory:
    self.db = SQLiteManager(self.config.historyStore.config.historyDbPath)
```

`DEFAULT_MEMORY_CONFIG` always includes `historyStore: { provider: "sqlite", config: { historyDbPath: "memory.db" } }`, so the `historyStore` branch is always taken, and the top-level `historyDbPath` field is never applied.

## Impact

- Users who configure `historyDbPath` (e.g. to an absolute path) get silently ignored
- The default `"memory.db"` is used relative to `process.cwd()`, which can be `/memory.db` in certain environments (e.g. systemd services or launchd plists with no `WorkingDirectory`), causing a crash: `SQLITE_CANTOPEN: unable to open database file`
- This is particularly bad because the error surfaces as an uncaught exception that crashes the process, with no indication that the config was ignored

## Steps to reproduce

```js
const m = new Memory({
  historyDbPath: "/tmp/my-mem0-history.db",
  embedder: { provider: "ollama", config: { model: "nomic-embed-text" } },
  vectorStore: { provider: "qdrant", config: { host: "localhost", port: 6333, collectionName: "test", dimension: 768 } },
  llm: { provider: "ollama", config: { model: "llama3.1:8b" } },
});

console.log(m.config.historyDbPath);            // "/tmp/my-mem0-history.db" ✓ (set correctly)
console.log(m.db.dbPath);                       // "memory.db" ✗ (ignores the config)
```

## Expected behavior

`historyDbPath` at the top level should be respected and used for the SQLite history database.

## Workaround

Set `historyStore` explicitly:

```js
const m = new Memory({
  historyStore: { provider: "sqlite", config: { historyDbPath: "/tmp/my-mem0-history.db" } },
  // ...
});
```

## Environment

- mem0ai version: latest (tested against `dist/oss/index.mjs`)
- Node.js OSS bindings

### LINKED ISSUE #4096: Node SDK OSS: historyDbPath config ignored, SQLite history always opens at default memory.db

Anandesh-Sharma:
## Summary
- Fixes `historyDbPath` being silently ignored in the Node SDK OSS mode
- `ConfigManager.mergeConfig()` always populates `historyStore` with defaults, so the branch that reads `historyDbPath` was dead code
- Now `historyDbPath` overrides `historyStore.config.historyDbPath` before the history manager is created
- This prevents `SQLITE_CANTOPEN` errors when `process.cwd()` is not writable (common in daemon/service contexts)

## Test plan
- [ ] Configure `historyDbPath: "/tmp/custom-history.db"` and verify the DB file is created at that path
- [ ] Verify default behavior still works when `historyDbPath` is not set
- [ ] Verify `disableHistory: true` still skips history entirely

Fixes #4096
Closes #4080
Closes #4074

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -77,6 +77,11 @@ export class Memory {
         },
       };
 
+      // If user provided historyDbPath, ensure it overrides the historyStore default
+      if (this.config.historyDbPath && this.config.historyStore?.config) {
+        this.config.historyStore.config.historyDbPath = this.config.historyDbPath;
+      }
+
       this.db =
         this.config.historyStore && !this.config.disableHistory
           ? HistoryManagerFactory.create(


kartik-mem0:
@Anandesh-Sharma  thank you for the contribution.

closing this pr in favour of this #4307.

PR #4159, opened 2026-02-28
fix: preserve http_auth in _safe_deepcopy_config for OpenSearch

giulio-leone:
## Problem

`_safe_deepcopy_config()` strips the `http_auth` field from OpenSearch configs, breaking AWS-signed authentication when the telemetry vector store is initialized. Users get:

```
WARNING:opensearch:HEAD https://xxxx.aoss.amazonaws.com:443/mem0migrations [status:403]
ERROR: Failed to initialize mem0 with OpenSearch: AuthorizationException(403, '')
```

## Root Cause

The `sensitive_tokens` list includes `"auth"` as a substring match, which inadvertently matches field names like `http_auth` and `auth`. These fields contain request-signing handler objects (e.g. `AWS4Auth`), not raw credentials. Stripping them prevents the telemetry vector store from authenticating.

## Fix

Remove `"auth"` from `sensitive_tokens`. The actual sensitive fields are already caught by the remaining, more specific tokens:

| Token | Catches |
|-------|---------|
| `"credential"` | `credentials`, `aws_credential`, etc. |
| `"password"` | `password`, `db_password`, etc. |
| `"token"` | `auth_token`, `access_token`, etc. |
| `"secret"` | `secret_key`, `client_secret`, etc. |
| `"key"` | `api_key`, `secret_key`, etc. |
| `"connection_class"` | `connection_class` |

The broad `"auth"` token was the only one producing false positives by matching handler fields.

## Tests

All 12 OpenSearch tests pass (2 consecutive clean runs). Updated `test_safe_deepcopy_config_handles_opensearch_auth` to verify that `http_auth` and `auth` handler objects are preserved while `credentials` and `connection_class` remain stripped.

Fixes #3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,7 +69,7 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
-        sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
+        sensitive_tokens = ("credential", "password", "token", "secret", "key", "connection_class")
         for field_name in list(clone_dict.keys()):
             if any(token in field_name.lower() for token in sensitive_tokens):
                 clone_dict[field_name] = None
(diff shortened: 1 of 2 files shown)


giulio-leone:
Closing in favor of #4167 which has the more thorough implementation.

PR #4160, opened 2026-02-28
fix(openclaw): propagate historyDbPath into historyStore to prevent SQLITE_CANTOPEN crash loop

cedric-pham:
## Description

Fixes a bug in the OpenClaw plugin where setting `oss.historyDbPath` in the config had no effect, causing a `SQLITE_CANTOPEN` crash loop when running as a macOS LaunchAgent (service).

### Root cause

`mem0ai`'s `DEFAULT_MEMORY_CONFIG` always initialises `historyStore.config.historyDbPath` to `"memory.db"` (a relative path). The `Memory` constructor always prefers `historyStore` over the top-level `historyDbPath` field when `historyStore` is present — and it is **always** present because it is merged from defaults. So the user-configured `historyDbPath` was never actually used.

When the OpenClaw gateway runs as a LaunchAgent with no `WorkingDirectory` set (the default install), `process.cwd()` resolves to `/`. The relative `"memory.db"` then becomes `/memory.db`, which is unwritable on macOS (System Integrity Protection), causing:

```
[openclaw] Uncaught exception: Error: SQLITE_CANTOPEN: unable to open database file
```

on every `before_agent_start` hook — crashing the gateway in a continuous restart loop on each incoming message.

### Fix

- Extracts the OSS Memory config-building logic from `OSSProvider._init()` into an exported `buildOSSMemoryConfig()` helper
- When `historyDbPath` is explicitly set, also sets `historyStore` to override the default, ensuring the `Memory` constructor uses the correct absolute path

Fixes # (no existing issue — discovered and fixed in production)

## Type of change

- [x] Refactor (does not change functionality — extraction of config builder for testability)

## How Has This Been Tested?

**Unit tests** (`openclaw/tests/oss-history-db.test.ts`, 13 tests via vitest):
- Base config shape with no/empty ossConfig
- Passthrough of embedder, vectorStore, and llm provider configs
- `historyDbPath` absent → `historyStore` not set
- `historyDbPath` present → both `historyDbPath` and `historyStore` set correctly
- `historyStore.config.historyDbPath` matches the configured path (regression test)
- `resolvePath` applied to both `historyDbPath` and `historyStore`

```
✓ tests/oss-history-db.test.ts (13 tests)
Test Files  1 passed (1)
     Tests  13 passed (13)
```

**Manual** (production verification on macOS with gateway running as LaunchAgent):
- Without fix: gateway crashed with `SQLITE_CANTOPEN` on every message
- With fix: `history.db` created at the specified absolute path, gateway runs stably

- [x] Unit Test
- [x] Test Script (manual — described above)

## Checklist

### DIFF
--- a/openclaw/index.ts
+++ b/openclaw/index.ts
@@ -203,6 +203,47 @@ class PlatformProvider implements Mem0Provider {
 }
 
 // ============================================================================
+// OSS Config Builder
+// ============================================================================
+
+/**
+ * Builds the config object passed to mem0ai's `Memory` constructor for
+ * open-source (self-hosted) mode.
+ *
+ * Exported for unit testing. Not part of the public plugin API.
+ *
+ * @internal
+ */
+export function buildOSSMemoryConfig(
+  ossConfig: Mem0Config["oss"],
+  resolvePath?: (p: string) => string,
+): Record<string, unknown> {
+  const config: Record<string, unknown> = { version: "v1.1" };
+
+  if (ossConfig?.embedder) config.embedder = ossConfig.embedder;
+  if (ossConfig?.vectorStore) config.vectorStore = ossConfig.vectorStore;
+  if (ossConfig?.llm) config.llm = ossConfig.llm;
+
+  if (ossConfig?.historyDbPath) {
+    const dbPath = resolvePath
+      ? resolvePath(ossConfig.historyDbPath)
+      : ossConfig.historyDbPath;
+    config.historyDbPath = dbPath;
+    // Also override historyStore so mem0ai's DEFAULT_MEMORY_CONFIG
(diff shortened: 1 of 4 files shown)


nyrosveil:
recheck

jonasstenberg:
Please prioritize this. 👍

utkarsh240799:
Thanks for the thorough contribution and testing! This is no longer needed — PR #4307 (merged) fixes the root cause upstream in the `mem0ai` SDK. `mergeConfig()` now correctly propagates top-level `historyDbPath` into `historyStore.config.historyDbPath`, so the plugin-side workaround is redundant. Closing in favor of the upstream fix.

PR #4162, opened 2026-02-28
fix: only pass dimensions to OpenAI embeddings API when explicitly configured

### LINKED ISSUE #4153: OpenAIEmbedding always passes dimensions, breaks non-matryoshka OpenAI-compatible embedding backends
## Description

When using `OpenAIEmbedding` with an OpenAI-compatible embedding backend (e.g. vLLM serving Qwen3-Embedding-0.6B), `mem0/embeddings/openai.py` always sends the `dimensions` parameter to the `/v1/embeddings` endpoint:

```python
# mem0/embeddings/openai.py
return (
    self.client.embeddings.create(
        input=[text],
        model=self.config.model,
        dimensions=self.config.embedding_dims,  # always passed
    )
    .data[0]
    .embedding
)
```

For non-matryoshka models, this causes a 400 error:

```text
BadRequestError: Error code: 400 - {
  "object": "error",
  "message": "Model \"custom_embedding\" does not support matryoshka representation, changing output dimensions will lead to poor results.",
  "type": "BadRequestError",
  "param": null,
  "code": 400
}
```

This makes it hard to use mem0 with self-hosted OpenAI-compatible embedding servers that do *not* support `dimensions` (or matryoshka embeddings), even though the rest of the API is compatible.

---

## Environment

- mem0 version: 1.0.4 (from `pyproject.toml`)
- Python: 3.11.x
- Embedding provider in mem0 config: `provider="openai"`
- Backend: vLLM OpenAI-compatible server serving Qwen3-Embedding-0.6B
- Vector store: Upstash Vector (OpenAI-compatible backend is used only for embeddings)

---

## Steps to Reproduce

1. Start a vLLM OpenAI-compatible server for an embedding model that does **not** support `dimensions`, e.g.:

   ```bash
   vllm serve Qwen/Qwen3-Embedding-0.6B \
     --task embedding \
     --host 0.0.0.0 \
     --port 5001 \
     --served-model-name custom_embedding
   ```

2. Configure mem0 to use this server as an “openai” embedder, e.g.:

   ```python
   config = {
       "llm": {...},  # e.g. vllm for chat
       "embedder": {
           "provider": "openai",
           "config": {
               "model": "custom_embedding",              # name exposed by vLLM
               "api_key": "EMPTY",
               "openai_base_url": "http://localhost:5001/v1",
               "embedding_dims": 1024,  # or any value
           },
       },
       "vector_store": {...},
   }
   ```

3. Call `Memory.from_config(config)` and then `memory.add(...)` or `memory.search(...)`.

4. The embedding call fails with a 400 BadRequestError (see error message above).

---

## Expected behavior

- If `embedding_dims` is **not** set (or explicitly set to `None`) in the embedder config, `OpenAIEmbedding` should **not** pass the `dimensions` argument to `client.embeddings.create`.
- This would allow mem0 to work with:
  - OpenAI’s own matryoshka models (when `embedding_dims` is set), **and**
  - Non-matryoshka / custom models on OpenAI-compatible backends (when `embedding_dims` is omitted).

In other words, `dimensions` should be **optional** and only sent when the user explicitly configures it.

---

## Proposed fix

In `mem0/embeddings/openai.py`, only include `dimensions` when it is explicitly set:

```python
# current code
return (
    self.client.embeddings.create(
        input=[text],
        model=self.config.model,
        dimensions=self.config.embedding_dims,
    )
    .data[0]
    .embedding
)

# proposed change (pseudo-code)
kwargs = {
    "input": [text],
    "model": self.config.model,
}
if self.config.embedding_dims is not None:
    kwargs["dimensions"] = self.config.embedding_dims

return (
    self.client.embeddings.create(**kwargs)
    .data[0]
    .embedding
)
```

This change should be fully backward compatible:

- For OpenAI matryoshka models: users can still set `embedding_dims` and get shorter vectors.
- For custom / non-matryoshka models: users can omit `embedding_dims` and the backend will use its default output dimension without error.

giulio-leone:
## Summary

Fixes #4153

`OpenAIEmbedding` always passes the `dimensions` parameter to `client.embeddings.create()`, even when the user has not explicitly configured `embedding_dims`. This causes a 400 error on OpenAI-compatible backends (e.g. vLLM serving non-matryoshka models like Qwen3-Embedding-0.6B) that reject the `dimensions` parameter.

## Root Cause

In `__init__`, `embedding_dims` is always defaulted to 1536:
```python
```
Then `embed()` unconditionally passes it:
```python
```

## Fix

Track whether the user explicitly provided `embedding_dims` **before** applying the default:
```python
self._pass_dimensions = self.config.embedding_dims is not None
```

Only include `dimensions` in the API call when the user explicitly configured it:
```python
kwargs = {"input": [text], "model": self.config.model}
if self._pass_dimensions:
    kwargs["dimensions"] = self.config.embedding_dims
return self.client.embeddings.create(**kwargs).data[0].embedding
```

## Backward Compatibility

- **Users who set `embedding_dims`**: behavior unchanged, `dimensions` is sent
- **Users who don't set `embedding_dims`**: `dimensions` is no longer sent, fixing non-matryoshka backends
- **`self.config.embedding_dims`** still defaults to 1536 for vector store index creation

### DIFF
--- a/mem0/embeddings/openai.py
+++ b/mem0/embeddings/openai.py
@@ -13,6 +13,9 @@ class OpenAIEmbedding(EmbeddingBase):
         super().__init__(config)
 
         self.config.model = self.config.model or "text-embedding-3-small"
+        # Only pass 'dimensions' to the API when the user explicitly set it.
+        # Non-matryoshka models (e.g. on vLLM) reject the parameter entirely.
+        self._pass_dimensions = self.config.embedding_dims is not None
         self.config.embedding_dims = self.config.embedding_dims or 1536
 
         api_key = self.config.api_key or os.getenv("OPENAI_API_KEY")
@@ -42,8 +45,7 @@ class OpenAIEmbedding(EmbeddingBase):
             list: The embedding vector.
         """
         text = text.replace("\n", " ")
-        return (
-            self.client.embeddings.create(input=[text], model=self.config.model, dimensions=self.config.embedding_dims)
-            .data[0]
-            .embedding
-        )
+        kwargs = {"input": [text], "model": self.config.model}
+        if self._pass_dimensions:
+            kwargs["dimensions"] = self.config.embedding_dims
+        return self.client.embeddings.create(**kwargs).data[0].embedding
--- a/tests/embeddings/test_openai_embeddings.py
+++ b/tests/embeddings/test_openai_embeddings.py
@@ -24,7 +24,7 @@ def test_embed_default_model(mock_openai_client):
     result = embedder.embed("Hello world")
 
     mock_openai_client.embeddings.create.assert_called_once_with(
-        input=["Hello world"], model="text-embedding-3-small", dimensions=1536
+        input=["Hello world"], model="text-embedding-3-small"
     )
     assert result == [0.1, 0.2, 0.3]
 
@@ -54,7 +54,7 @@ def test_embed_removes_newlines(mock_openai_client):
     result = embedder.embed("Hello\nworld")
 
     mock_openai_client.embeddings.create.assert_called_once_with(
-        input=["Hello world"], model="text-embedding-3-small", dimensions=1536
+        input=["Hello world"], model="text-embedding-3-small"
     )
     assert result == [0.7, 0.8, 0.9]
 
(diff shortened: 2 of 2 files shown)


giulio-leone:

PR #4133, opened 2026-02-25
feat: add default baseURL for OpenAI clients

Kennytian:
- Add default baseURL (https://api.openai.com/v1) to OpenAI LLM implementations
- Update OpenAI embeddings to use baseURL parameter with default fallback
- Add baseURL support to OpenAI structured LLM implementation
- Configure utility functions to use configurable baseURL
- Allow users to customize OpenAI endpoint while maintaining compatibility with original API

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] Documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

Please delete options that are not relevant.

- [ ] Unit Test
- [ ] Test Script (please provide)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

## Maintainer Checklist

- [ ] closes #xxxx (Replace xxxx with the GitHub issue number)
- [ ] Made sure Checks passed

PR #4169, opened 2026-02-28
feature(ts): forward baseURL to OpenAI embedder

jonasstenberg:
## Summary

The `OpenAIEmbedder` constructor only passes `apiKey` to the OpenAI client, ignoring `baseURL`. This is inconsistent with `OpenAILLM` which already forwards `baseURL`, and with the Zod config schema which already validates `baseURL` on the embedder config.

Without this fix, users hosting OpenAI-compatible embedding servers (e.g. llama.cpp, vLLM, TEI) cannot set the endpoint through config and must rely on the `OPENAI_BASE_URL` environment variable. This causes conflicts when the same env var is also needed for model API calls to a different endpoint.

## Changes

- Add `baseURL` to the `EmbeddingConfig` TypeScript interface (matches existing Zod schema)
- Forward `baseURL` from config to the OpenAI client constructor in `OpenAIEmbedder` (matches `OpenAILLM` behavior)
- Add test verifying `baseURL` is forwarded to the OpenAI client

Both changes are backwards-compatible — `baseURL` is optional and `undefined` preserves existing behavior (falls back to env var, then default).

### DIFF
--- a/mem0-ts/src/oss/tests/openai-embedder.test.ts
+++ b/mem0-ts/src/oss/tests/openai-embedder.test.ts
@@ -1,0 +1,24 @@
+/// <reference types="jest" />
+import { OpenAIEmbedder } from "../src/embeddings/openai";
+
+describe("OpenAIEmbedder", () => {
+  it("should forward baseURL to OpenAI client", () => {
+    const embedder = new OpenAIEmbedder({
+      apiKey: "test-key",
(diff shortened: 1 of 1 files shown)


jonasstenberg:
Closing in favor of #4133 which covers the same fix with broader scope (embedder + LLM + structured LLM + utils).

PR #4180, opened 2026-03-01
fix: extract text from request body in update_memory endpoint

### LINKED ISSUE #3235: TypeError in update_memory due to incorrect data type passed to embed()
## Description

There's a type mismatch bug in the `update_memory` functionality where a `Dict[str, Any]` is being passed to the embedding model's `embed()` method, which expects a string.

## Steps to Reproduce

1. Call `update_memory` with a dictionary containing updated memory data
2. The dictionary gets passed through to `memory/main.py`'s `update()` method
3. The `update()` method directly passes the dict to `embedding_model.embed()`
4. The `embed()` method attempts to call `.replace()` on the dict, causing a TypeError

## Current Behavior

- `server/main.py`'s `update_memory` receives `Dict[str, Any]` as `updated_memory`
- This dict is passed to `memory/main.py`'s `update()` method as the `data` parameter
- The `update()` method directly passes this dict to `embedding_model.embed()`
- `embed()` method expects a string and fails when trying to call `text.replace()` on a dict

## Expected Behavior

The system should either:

1. Expect `updated_memory` to be a dict with specific content fields (e.g., `{"content": "new memory content"}`) and extract the text content before embedding
2. Convert the dict to a string representation before passing to `embed()`

## Location

- **File**: `mem0/memory/main.py`
- **Line**: 741 (and corresponding line in async version)

## Proposed Solution

The bug can be fixed by either:

1. Modifying the API to expect structured data and extracting the relevant text field
2. Adding proper data conversion before calling the embedding method

This appears to be a regression or oversight in the type handling between the server API and the memory update functionality.

Br1an67:
## Description

The PUT /memories/{memory_id} endpoint receives a JSON body (Dict[str, Any]) but passes the entire dict directly to Memory.update(data=...), which expects a string. This causes TypeError when the embedding model calls text.replace() on the dict.

The fix extracts the "text" or "data" field from the request body before passing it to Memory.update(), consistent with how the Platform API client sends updates. Returns HTTP 400 if neither field is present.

Fixes #3235

## Type of change

## How Has This Been Tested?

Verified that sending a JSON body with a "text" field correctly extracts the string before calling Memory.update().

## Checklist:

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -155,13 +155,16 @@ def update_memory(memory_id: str, updated_memory: Dict[str, Any]):
     
     Args:
         memory_id (str): ID of the memory to update
-        updated_memory (str): New content to update the memory with
+        updated_memory (dict): Request body containing "text" or "data" field with the new content
         
     Returns:
         dict: Success message indicating the memory was updated
     """
     try:
-        return MEMORY_INSTANCE.update(memory_id=memory_id, data=updated_memory)
+        data = updated_memory.get("text") or updated_memory.get("data")
+        if data is None:
+            raise HTTPException(status_code=400, detail="Request body must include a 'text' or 'data' field with string content.")
+        return MEMORY_INSTANCE.update(memory_id=memory_id, data=data)
     except Exception as e:
         logging.exception("Error in update_memory:")
         raise HTTPException(status_code=500, detail=str(e))


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

kartik-mem0:
Hi @Br1an67 — thank you for identifying this bug and putting together a fix! 🙏

This issue has since been resolved in #4595 (merged Apr 2), which introduces a `MemoryUpdate` Pydantic model and extracts the text field before passing it to `Memory.update()`.

Closing this PR as superseded. We really appreciate your contribution — hope to see you on future issues!

PR #4182, opened 2026-03-01
fix: skip Posthog client creation when MEM0_TELEMETRY=false

### LINKED ISSUE #3762: Mem0 Telemetry Creates Threads Even When MEM0_TELEMETRY=false, Causing Thread Exhaustion in AWS Lambda
## Environment
- **mem0ai version**: 1.0.0
- **posthog version**: 6.7.6
- **Python version**: 3.12
- **Platform**: AWS Lambda (but potentially affects any long-running process with container/process reuse)

## Problem Summary

When `MEM0_TELEMETRY='false'` is set, Posthog threads are still created on every mem0 operation, leading to thread exhaustion in environments with process reuse (like AWS Lambda functions). This causes `RuntimeError: can't start new thread` errors and complete function failures.

## Root Cause

In [`mem0/memory/telemetry.py`](https://github.com/mem0ai/mem0/blob/main/mem0/memory/telemetry.py), the `AnonymousTelemetry` class instantiates Posthog **before** checking if telemetry is disabled:

```python
class AnonymousTelemetry:
    def __init__(self, vector_store=None):
        # Line 27: ALWAYS creates Posthog client (spawns background threads)
        self.posthog = Posthog(project_api_key=PROJECT_API_KEY, host=HOST)
        
        self.user_id = get_or_create_user_id(vector_store)
        
        # Line 31-32: Only disables AFTER threads already created
        if not MEM0_TELEMETRY:
            self.posthog.disabled = True  # Too late - threads already spawned!
```

Additionally, the `capture_event()` function creates a **new** `AnonymousTelemetry` instance on every call:

```python
def capture_event(event_name, memory_instance, additional_data=None):
    oss_telemetry = AnonymousTelemetry(...)  # NEW instance = NEW threads!
    # ...
```

This means:
1. Module import creates threads (line 55: `client_telemetry = AnonymousTelemetry()`)
2. Every `memory.add()`, `memory.delete()`, etc. creates MORE threads (via `capture_event()`)
3. Setting `self.posthog.disabled = True` only prevents event capture, not thread creation

## Impact

In AWS Lambda (and similar environments):
- Lambda containers are reused across invocations
- Threads accumulate: Invocation 1 (3 threads) → Inv 2 (6 threads) → Inv 3 (9 threads)...
- After ~300-500 invocations: 1000-1500 threads created
- OS thread limit reached → `RuntimeError: can't start new thread`
- Lambda function becomes completely non-functional

## Reproduction Steps

1. Set environment variable: `os.environ['MEM0_TELEMETRY'] = 'false'`
2. Create a Memory instance
3. Call `memory.add()` repeatedly in a loop (100+ times)
4. Monitor thread count with `threading.active_count()`
5. Observe threads continuously growing despite telemetry being "disabled"

## Expected Behavior

When `MEM0_TELEMETRY='false'`:
- No Posthog client should be created
- No telemetry threads should be spawned
- Thread count should remain constant across operations

## Actual Behavior

- Posthog client is created regardless of `MEM0_TELEMETRY` setting
- Background threads are spawned on module import and every operation
- Thread count grows unbounded

## Production Error Logs

```
File "/var/task/mem0/memory/telemetry.py", line 59 → AnonymousTelemetry()
File "/var/task/mem0/memory/telemetry.py", line 27 → self.posthog = Posthog(...)
File "/var/task/posthog/client.py", line 316 → consumer.start()
RuntimeError: can't start new thread
```

## Proposed Fix

Check `MEM0_TELEMETRY` **before** creating Posthog:

```python
class AnonymousTelemetry:
    def __init__(self, vector_store=None):
        # Check telemetry setting FIRST
        if not MEM0_TELEMETRY:
            self.posthog = None
            self.user_id = None
            self.disabled = True
            return
        
        # Only create Posthog if telemetry enabled
        self.posthog = Posthog(project_api_key=PROJECT_API_KEY, host=HOST)
        self.user_id = get_or_create_user_id(vector_store)
        self.disabled = False
    
    def capture_event(self, event_name, properties=None, user_email=None):
        if self.posthog is None or self.disabled:
            return
        # ... rest of implementation
```

Additionally, consider reusing the module-level `client_telemetry` instance instead of creating new instances in `capture_event()`.

## Notes

- This affects AWS Lambda significantly due to container reuse
- Similar issues likely occur in other long-running processes (web servers, daemons)
- Setting `MEM0_TELEMETRY` is doesn't actually prevent thread creation
- The issue compounds over time and not immediately visible in short-running scripts

## Related Code

- [`mem0/memory/telemetry.py` line 27](https://github.com/mem0ai/mem0/blob/main/mem0/memory/telemetry.py#L27)
- [`mem0/memory/telemetry.py` line 59](https://github.com/mem0ai/mem0/blob/main/mem0/memory/telemetry.py#L59)
- [`mem0/memory/main.py`](https://github.com/mem0ai/mem0/blob/main/mem0/memory/main.py) (calls to `capture_event`)

Br1an67:
## Description

When `MEM0_TELEMETRY=false`, the `AnonymousTelemetry` class still instantiates a `Posthog` client (which spawns background threads) before checking the flag and setting `self.posthog.disabled = True`. Additionally, `capture_event()` creates a new `AnonymousTelemetry` instance on every call, accumulating threads.

In environments with process reuse (e.g. AWS Lambda), this causes thread exhaustion (`RuntimeError: can't start new thread`) after a few hundred invocations.

**Fix:**
1. Check `MEM0_TELEMETRY` first in `AnonymousTelemetry.__init__()` — skip all Posthog initialization when disabled
2. Guard `capture_event()` and `close()` against `None` posthog client
3. Add early return in module-level `capture_event()` when telemetry is disabled

Fixes #3762

## Type of change

## How Has This Been Tested?

- [x] Unit Test

Verified that with `MEM0_TELEMETRY=false`:
- `AnonymousTelemetry().posthog` is `None` (no Posthog client created)
- `capture_event()` is a safe no-op
- `close()` is a safe no-op
- All 3 existing telemetry tests pass

## Checklist:

### DIFF
--- a/mem0/memory/telemetry.py
+++ b/mem0/memory/telemetry.py
@@ -24,14 +24,17 @@ logging.getLogger("urllib3").setLevel(logging.CRITICAL + 1)
 
 class AnonymousTelemetry:
     def __init__(self, vector_store=None):
-        self.posthog = Posthog(project_api_key=PROJECT_API_KEY, host=HOST)
+        if not MEM0_TELEMETRY:
+            self.posthog = None
+            self.user_id = None
+            return
 
+        self.posthog = Posthog(project_api_key=PROJECT_API_KEY, host=HOST)
         self.user_id = get_or_create_user_id(vector_store)
 
-        if not MEM0_TELEMETRY:
-            self.posthog.disabled = True
-
     def capture_event(self, event_name, properties=None, user_email=None):
+        if not self.posthog:
+            return
         if properties is None:
             properties = {}
         properties = {
@@ -49,13 +52,17 @@ class AnonymousTelemetry:
         self.posthog.capture(distinct_id=distinct_id, event=event_name, properties=properties)
 
     def close(self):
-        self.posthog.shutdown()
+        if self.posthog:
+            self.posthog.shutdown()
 
 
 client_telemetry = AnonymousTelemetry()
 
 
 def capture_event(event_name, memory_instance, additional_data=None):
+    if not MEM0_TELEMETRY:
+        return
+
     oss_telemetry = AnonymousTelemetry(
         vector_store=memory_instance._telemetry_vector_store
         if hasattr(memory_instance, "_telemetry_vector_store")


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

kartik-mem0:
Hey @Br1an67, thanks for taking the time to put this fix together — really appreciate it\! Just wanted to let you know that this was independently resolved in #4203 (commit 34c797d2), which landed on main on Mar 4. The fix there covers the same changes you proposed here. Since this is now addressed, I'll go ahead and close this PR. Thanks again for flagging the issue and contributing\!

PR #4197, opened 2026-03-02
fix(ts): pass baseURL to OpenAI client in OpenAIEmbedder

### LINKED ISSUE #4191: Bug: OpenAIEmbedder ignores baseURL config parameter
# Bug Report: OpenAIEmbedder ignores baseURL config

## Summary
The `OpenAIEmbedder` class in mem0 TypeScript SDK does not use the `baseURL` configuration parameter when initializing the OpenAI client, forcing users to manually set `process.env.OPENAI_BASE_URL` as a workaround.

## Steps to Reproduce
1. Configure mem0 with a custom OpenAI-compatible API endpoint (e.g., NVIDIA API, local Ollama):
   ```typescript
   const memory = new Memory({
     embedder: {
       provider: "openai",
       config: {
         apiKey: "sk-xxx",
         baseURL: "https://integrate.api.nvidia.com/v1",
         model: "nvidia/llama-3.1-nv-embedqa-1b-v1"
       }
     }
   });
   ```

2. Initialize the memory instance
3. Attempt to add or search memories
4. Error occurs because the OpenAI client connects to default OpenAI URL instead of the custom baseURL

## Expected Behavior
The `OpenAIEmbedder` should respect the `baseURL` configuration parameter, similar to how `OpenAILLM` does.

## Actual Behavior
The `OpenAIEmbedder` constructor only passes `apiKey` to the OpenAI client, ignoring `baseURL`:
```javascript
var OpenAIEmbedder = class {
  constructor(config) {
    this.openai = new import_openai.default({ apiKey: config.apiKey }); // ❌ baseURL is ignored
    // ...
  }
}
```

## Workaround
Users are forced to set environment variables manually before initializing mem0:
```typescript
process.env.OPENAI_BASE_URL = customBaseURL;
process.env.OPENAI_API_BASE = customBaseURL;
const memory = new Memory({ /* config */ });
```

## Comparison with OpenAILLM
The `OpenAILLM` class correctly implements `baseURL` support:
```javascript
var OpenAILLM = class {
  constructor(config) {
    this.openai = new import_openai2.default({
      apiKey: config.apiKey,
      baseURL: config.baseURL  // ✅ Correctly uses baseURL
    });
    // ...
  }
}
```

## Environment
- mem0ai version: 2.2.4
- Node.js version: 20.x
- OS: Linux

## Proposed Fix
Update the `OpenAIEmbedder` class in `src/oss/src/embeddings/openai.ts`:

```typescript
var OpenAIEmbedder = class {
  constructor(config) {
    this.openai = new import_openai.default({ 
      apiKey: config.apiKey,
      baseURL: config.baseURL  // ✅ Add this line
    });
    this.model = config.model || "text-embedding-3-small";
    this.embeddingDims = config.embeddingDims || 1536;
  }
  // ...
}
```

## Impact
This affects all users who:
- Use custom OpenAI-compatible API endpoints (NVIDIA, local models, etc.)
- Cannot or prefer not to set global environment variables
- Want to keep configuration clean and explicit

## Additional Notes
- The same issue may exist in `AzureOpenAIEmbedder` class
- Users should not need to rely on environment variable hacks for a standard configuration parameter
- This inconsistency between `OpenAILLM` and `OpenAIEmbedder` is confusing and error-prone

weiguangli-io:
## Summary

- Fixes #4191
- The `OpenAIEmbedder` constructor only passed `apiKey` to the OpenAI client, silently ignoring the `url` config property
- Users of OpenAI-compatible endpoints (NVIDIA, Ollama, local servers, etc.) were forced to set `process.env.OPENAI_BASE_URL` as a workaround
- This forwards `config.url` as the OpenAI client's `baseURL` option, consistent with how `OllamaEmbedder` uses `config.url` and how `OpenAILLM` passes its own `baseURL`

## Test plan

- [ ] Configure mem0 with a custom `url` in the embedder config and verify embeddings are fetched from the correct endpoint

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/openai.ts
+++ b/mem0-ts/src/oss/src/embeddings/openai.ts
@@ -8,7 +8,7 @@ export class OpenAIEmbedder implements Embedder {
   private embeddingDims?: number;
 
   constructor(config: EmbeddingConfig) {
-    this.openai = new OpenAI({ apiKey: config.apiKey });
+    this.openai = new OpenAI({ apiKey: config.apiKey, baseURL: config.url });
     this.model = config.model || "text-embedding-3-small";
     this.embeddingDims = config.embeddingDims || 1536;
   }


weiguangli-io:

PR #4199, opened 2026-03-02
fix: load default configuration from default_config.json

### LINKED ISSUE #4192: get_default_configuration() in config.py is hardcoded to return OpenAI defaults
### 🐛 Describe the bug

it ignored default_config.json entirely.

weiguangli-io:
## Summary
- Modified `get_default_configuration()` to read from `default_config.json` instead of hardcoding OpenAI defaults
- Falls back to hardcoded defaults if the file is missing or contains invalid JSON
- Ensures `openmemory` and `mem0` sections are always present in the returned config

## Problem
`get_default_configuration()` in `config.py` had all default values hardcoded inline, completely ignoring the `default_config.json` file that exists in `openmemory/api/`. Users who customized `default_config.json` (e.g., to use Ollama instead of OpenAI) found their changes had no effect because the file was never read.

## Test plan
- [ ] Modify `default_config.json` to use a non-OpenAI provider, restart server, verify `GET /api/v1/config` returns the file-based defaults
- [ ] Delete `default_config.json`, restart server, verify hardcoded OpenAI defaults are returned
- [ ] Call `POST /api/v1/config/reset` to verify reset uses file-based defaults

Fixes #4192

### DIFF
--- a/openmemory/api/app/routers/config.py
+++ b/openmemory/api/app/routers/config.py
@@ -1,3 +1,5 @@
+import json
+import os
 from typing import Any, Dict, Optional
 
 from app.database import get_db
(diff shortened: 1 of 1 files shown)


weiguangli-io:

PR #4206, opened 2026-03-03
Fix Ollama integration: 503 errors, type errors, and list handling

lamost423:
## Summary

Fixes several critical bugs in Ollama integration:

1. **503 Error on init** - Replace Python client with HTTP API + retry
2. **400 Error on embed** - Add text normalization for list outputs
3. **TypeError: unhashable list** - Convert lists to strings before dict use

## Changes

- : HTTP API + text normalization
- : HTTP API + model-specific cleaning
- : Handle list-type facts

## Testing

Tested with:
- qwen2.5:3b (stable, ~38s for 3 facts)
- mxbai-embed-large
- Fully offline operation

Fixes issues with Ollama 503 errors and type handling.

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -1,24 +1,10 @@
-import subprocess
-import sys
+import requests
+import json
+import time
 from typing import Literal, Optional
 
 from mem0.configs.embeddings.base import BaseEmbedderConfig
 from mem0.embeddings.base import EmbeddingBase
-
-try:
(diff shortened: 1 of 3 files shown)


kartik-mem0:
Thanks for looking into the Ollama 503 errors. However, this approach replaces the Python client with raw HTTP calls, which is a fairly invasive change. Closing this — if you're still hitting 503 errors, please open an issue so we can investigate the root cause.

PR #4217, opened 2026-03-04

tiantaiping2024:
## Summary

The `_safe_deepcopy_config()` sanitizer was too broad, removing any field containing 'auth' which breaks OpenSearch AWS signature authentication (`http_auth`).

## Changes

1. Only remove truly sensitive fields: `password`, `secret`, `api_key`
2. Preserve runtime auth objects: `http_auth`, `auth`, and fields ending with `_auth`

## Root Cause

Original code:
```python
```

This matches `http_auth` which is required for AWS OpenSearch signed requests.

## Fixes

Fixes #3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,8 +69,12 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @tiantaiping2024 , 

Thanks again for your effort!

PR #4220, opened 2026-03-05

tiantaiping2024:
## Description

The `_safe_deepcopy_config` function was incorrectly removing the `http_auth` field from OpenSearch configs because 'auth' was in the sensitive_tokens list. This caused OpenSearch integrations using AWSV4SignerAuth (or any custom http_auth) to fail with 403 authentication errors.

## Root Cause

In the original code, the sanitization logic checked if any field name contained:
```
```

This incorrectly matched `http_auth`, which is a required runtime authentication object for OpenSearch clients using AWS signature.

## Fix

Now only genuinely sensitive fields (`password`, `token`, `secret`, `api_key`) are stripped, while runtime auth objects like `http_auth`, `auth`, and `connection_class` are preserved.

## Testing

- OpenSearch integrations using AWSV4SignerAuth should now work correctly
- Other authentication mechanisms (basic auth, API keys) are still properly sanitized for telemetry

## Related Issue

Fixes #3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,9 +69,11 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
-        sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
+        # Only strip genuinely sensitive string/bytes fields, preserve runtime auth objects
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @tiantaiping2024 , 

Thanks again for your effort!

PR #4223, opened 2026-03-05
fix(ts-sdk): infer vector store dimension from embedder config

### LINKED ISSUE #4212: Qdrant collection created with wrong dimensions when using Ollama embedder (nomic-embed-text)
## Bug

When using the OpenClaw plugin (or OSS mode generally) with Ollama's `nomic-embed-text` embedder, the Qdrant collection is created with **1536 dimensions** (OpenAI default), but `nomic-embed-text` produces **768-dimensional** vectors. This causes every `add()` and `search()` call to fail with a cryptic `Bad Request` error from Qdrant (dimension mismatch).

## Root Cause

In the Qdrant vector store class (`src/oss/src/vector_stores/qdrant.ts`):

```typescript
this.dimension = config.dimension || 1536;
```

The fallback is hardcoded to 1536 (OpenAI's `text-embedding-3-small`). When using any non-OpenAI embedder, the collection is created with the wrong vector size.

## Reproduction

1. Configure mem0 OSS with Ollama + nomic-embed-text:
```json
{
  "embedder": { "provider": "ollama", "config": { "model": "nomic-embed-text" } },
  "vectorStore": { "provider": "qdrant", "config": { "host": "localhost", "port": 6333 } }
}
```
2. Call `memory.add()` or `memory.search()`
3. Both fail with `Bad Request` — Qdrant rejects 768-dim vectors into a 1536-dim collection

## Workaround

Explicitly set `dimension: 768` in the vectorStore config:
```json
"vectorStore": {
  "provider": "qdrant",
  "config": { "host": "localhost", "port": 6333, "dimension": 768 }
}
```

## Suggested Fix

Auto-detect the embedding dimension from the configured embedder at init time (e.g., run a test embedding and read the vector length), rather than defaulting to 1536. This would make any embedder work out of the box without requiring users to know and manually set the dimension.

Alternatively, at minimum, log a warning or throw a clear error when the embedder's output dimension doesn't match the configured/default collection dimension.

## Environment

- mem0ai (JS/npm) v0.1.2
- Qdrant v1.13.0 (Docker)
- Ollama nomic-embed-text (768 dims)
- OpenClaw plugin `@mem0/openclaw-mem0`

weiguangli-io:
## Summary
Fixes #4212

When using non-OpenAI embedders (e.g. Ollama `nomic-embed-text` with 768 dims), the Qdrant collection was always created with 1536 dimensions because the vector store config defaulted to the OpenAI dimension without considering the embedder's actual output size. This caused every `add()` and `search()` call to fail with a dimension mismatch error from Qdrant.

This PR updates `ConfigManager.mergeConfig()` to automatically infer the vector store dimension from the embedder's `embeddingDims` config when the user has not explicitly set it on the vector store. The priority order is:

1. Explicit `vectorStore.config.dimension` (user override — always wins)
2. `embedder.config.embeddingDims` (inferred from embedder)
3. Default `1536` (backward compatible fallback)

This fix benefits **all** vector store providers (Qdrant, Memory, Vectorize, etc.) since the dimension is resolved at the config layer.

## Test Plan
- Added `config-manager.test.ts` with 4 test cases covering:
  - Default dimension (1536) when no embedder dims specified
  - Dimension inferred from `embeddingDims` (e.g. 768 for Ollama)
  - Explicit vector store dimension takes priority over embedder dims
  - Dimension inference works with custom client instances
- All existing tests pass (`factory.test.ts`)

### DIFF
--- a/mem0-ts/src/oss/src/config/manager.ts
+++ b/mem0-ts/src/oss/src/config/manager.ts
@@ -43,13 +43,22 @@ export class ConfigManager {
           const defaultConf = DEFAULT_MEMORY_CONFIG.vectorStore.config;
           const userConf = userConfig.vectorStore?.config;
 
+          // Infer vector dimension from the embedder config when the user
+          // has not explicitly set it on the vector store.  This prevents
+          // dimension mismatches when using non-OpenAI embedders (e.g.
+          // Ollama nomic-embed-text produces 768-dim vectors while the
+          // default was hardcoded to 1536).
+          const embeddingDims = userConfig.embedder?.config?.embeddingDims;
+          const inferredDimension =
+            userConf?.dimension || embeddingDims || defaultConf.dimension;
+
           // Prioritize user-provided client instance
           if (userConf?.client && typeof userConf.client === "object") {
             return {
               client: userConf.client,
               // Include other fields from userConf if necessary, or omit defaults
               collectionName: userConf.collectionName, // Can be undefined
-              dimension: userConf.dimension || defaultConf.dimension, // Merge dimension
+              dimension: inferredDimension,
               ...userConf, // Include any other passthrough fields from user
             };
           } else {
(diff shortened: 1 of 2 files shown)


weiguangli-io:

PR #4227, opened 2026-03-06
fix: narrower sensitive field matching + None guards + langchain type fix

SabaPivot:
## Summary

- **`_safe_deepcopy_config` strips `http_auth` (#3580):** The sensitive field filter used broad substring matching — `"auth"` matched `http_auth`, removing it from OpenSearch configs. Replaced with exact field names (`password`, `api_key`, `secret_key`, `access_token`, `credentials`, `connection_class`) and narrower tokens (`credential`, `secret`, `_token`). `http_auth` and `auth` are now preserved.

- **`_update_memory` / `_delete_memory` AttributeError on missing ID:** `vector_store.get()` returns `None` when ID doesn't exist, but the code immediately accesses `.payload` without checking. Added `None` guard with clear error message (update) or early return (delete). Applied to both sync and async versions.

- **Langchain `update()` type mismatch (#3767):** `update()` passed a single vector (`List[float]`) to `insert()`, which expects `List[List[float]]`. Wrapped in `[vector]` / `[payload]`.

## Test plan

- [x] All 12 OpenSearch tests pass (`pytest tests/vector_stores/test_opensearch.py`)
- [ ] Verify `_safe_deepcopy_config` preserves `http_auth` with AWS auth objects
- [ ] Verify `_update_memory("nonexistent_id")` raises `ValueError` instead of `AttributeError`
- [ ] Verify `_delete_memory("nonexistent_id")` returns gracefully instead of crashing
- [ ] Verify langchain `update()` correctly wraps vector/payload in lists

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,9 +69,11 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
-        sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
+        sensitive_exact_fields = {"password", "api_key", "secret_key", "access_token", "credentials", "connection_class"}
+        sensitive_tokens = ("credential", "secret", "_token")
         for field_name in list(clone_dict.keys()):
-            if any(token in field_name.lower() for token in sensitive_tokens):
+            lower_name = field_name.lower()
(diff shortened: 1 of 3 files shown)


kartik-mem0:
Hi @SabaPivot, thanks for the effort on this! The langchain type mismatch has been fixed via #4446. Closing this PR — appreciate your contribution! 🙏

PR #4239, opened 2026-03-07
Fix: Memory leak and thread leak in telemetry

### LINKED ISSUE #3376: Memory leak and thread leak issues in mem0 library
### 🐛 Describe the bug

**Environment**

- Python 3.10
- mem0ai==0.1.115
- pymilvus==2.3.2

**Problem**

**Code**
`
def test_api():
    import psutil

tiantaiping2024:
## Description
Fixes #3376 - Memory leak and thread leak in telemetry

This PR fixes the memory leak and thread leak issue in the telemetry module by:
- Properly shutting down telemetry scheduler on cleanup
- Using daemon threads to prevent blocking shutdown

## Changes
- Added proper thread/daemon configuration in telemetry
- Added cleanup method to prevent resource leaks

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1236,6 +1236,24 @@ class Memory(MemoryBase):
             )
         capture_event("mem0.reset", self, {"sync_type": "sync"})
 
+    def cleanup(self):
+        """
+        Cleanup resources including telemetry connections.
+        Call this method when done using the Memory instance to prevent thread leaks.
+        """
+        from mem0.memory.telemetry import shutdown_telemetry
+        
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hi @tiantaiping2024, thank you for working on this telemetry leak fix! Your effort is much appreciated.

This issue has now been resolved via PR #4535, which was merged with a comprehensive fix including a thread-safe singleton PostHog client, `atexit` cleanup, and proper `Memory.close()` / context manager support.

Since the fix is already in place, we're closing this PR. Thank you for your contribution and we hope to see you around!

PR #4258, opened 2026-03-08
Fix: _safe_deepcopy_config removes http_auth from OpenSearch configs

xiaoamo22333:
## Bug Description

The `_safe_deepcopy_config()` function in `memory/main.py` has an overly aggressive sensitive field filter that removes the `http_auth` field from configurations. This field is required by OpenSearch clients for authentication.

## Root Cause

The `sensitive_tokens` tuple includes `"auth"` which matches any field containing that substring, including legitimate fields like `http_auth`.

## Fix

Changed the `sensitive_tokens` from:
```python
```

to:
```python
sensitive_tokens = ("credential", "password", "token", "secret", "api_key", "connection_class")
```

This removes the overly broad `"auth"` token while preserving protection for truly sensitive fields like `password`, `token`, `secret`, and `api_key`.

## Testing

- The fix preserves `http_auth` in OpenSearch configurations
- Still filters sensitive fields like `password`, `api_key`, etc.

Fixes #3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,7 +69,7 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
-        sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @xiaoamo22333 , 

Thanks again for your effort!

PR #4261, opened 2026-03-08
Fix delete_all() incorrectly resetting entire vector store

jnMetaCode:
## Problem

The sync `delete_all()` method in `Memory` calls `self.vector_store.reset()` (line 1053) after deleting the individual memories that match the provided filter. The `reset()` method drops the entire vector store collection and recreates it, which destroys **all** stored memories — including those belonging to other users, agents, or runs that were not targeted by the filter.

For example, calling `memory.delete_all(user_id="alice")` would:
1. Correctly delete Alice's memories one by one
2. Then call `reset()` which wipes Bob's, Charlie's, and everyone else's memories too

## Fix

Remove the `self.vector_store.reset()` call from `delete_all()`. The async `AsyncMemory.delete_all()` already correctly omits this call — this fix aligns the sync version with the async behavior.

## How I found this

Compared the sync and async implementations side by side. The async version at line 2108-2121 correctly deletes individual memories and moves on, while the sync version at line 1050-1053 has the extra destructive `reset()` call.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -211,6 +211,28 @@ def test_delete_all(memory_instance, version, enable_graph):
     assert result["message"] == "Memories deleted successfully!"
 
 
+def test_delete_all_does_not_reset_vector_store(memory_instance):
+    """Regression test: delete_all for one user must not wipe the entire vector store.
+
+    Previously, delete_all() called vector_store.reset() after deleting filtered
+    memories, which destroyed all data regardless of scope.
+    """
+    mock_memories = [Mock(id="mem_1"), Mock(id="mem_2")]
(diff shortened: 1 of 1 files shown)


jnMetaCode:

jnMetaCode:

jnMetaCode:
Good point — added `test_delete_all_does_not_reset_vector_store` in 672117a. It calls `delete_all(user_id="user_a")`, checks that only the two filtered memories are individually deleted, and asserts `vector_store.reset()` was never called.

Let me know if you'd like the test expanded further (e.g. a multi-scope scenario).

kartik-mem0:
Closing — this fix was already merged via PR #4349. Thanks for the contribution!

PR #4264, opened 2026-03-08
fix(upstash_vector): pass namespace as top-level arg to query_many, not inside each query dict

### LINKED ISSUE #4207: UpstashVector.search uses namespace inside each query, breaking upstash_vector.query_many

s-zx:
## Summary

`UpstashVector.search` with `enable_embeddings=False` was broken for single-vector queries with the current `upstash_vector` client SDK, and silently returned results from the wrong namespace for multi-vector queries.

## Root Cause

The new `upstash_vector` SDK's `query_many()` signature is:

```python
query_many(*, queries: List[QueryRequest], namespace: str = DEFAULT_NAMESPACE, ...)
```

`namespace` is a **top-level** keyword argument, and `QueryRequest` objects must NOT contain a `"namespace"` key.

The previous code embedded namespace inside each query dict:

```python
# Before (buggy)
queries = [
    {"vector": v, "top_k": limit, "filter": ..., "include_metadata": True, "namespace": self.collection_name},
    ...
]
responses = self.client.query_many(queries=queries)  # namespace missing at top level
```

This caused two failures (#4207):

1. **Single vector**: `query_many` special-cases `len(queries)==1` and calls `self.query(**query, namespace=namespace)`. Since the query dict already had `namespace`, this raised:
   ```
   TypeError: query() got multiple values for keyword argument 'namespace'
   ```
2. **Multiple vectors**: `namespace` inside the dict was silently ignored, returning results from the wrong (default) namespace.

## Fix

Remove `"namespace"` from the per-query dicts and pass it at the top level:

```python
# After (fixed)
queries = [
    {"vector": v, "top_k": limit, "filter": ..., "include_metadata": True},
    ...
]
responses = self.client.query_many(queries=queries, namespace=self.collection_name)
```

## Tests

- Updated `test_search_vectors` and `test_search_vectors_empty_filters` to assert the correct call signature.
- Added `test_search_vectors_namespace_not_in_query_dicts` as an explicit regression guard that verifies `namespace` is a top-level argument and is absent from per-query dicts.

Fixes #4207

### DIFF
--- a/mem0/vector_stores/upstash_vector.py
+++ b/mem0/vector_stores/upstash_vector.py
@@ -132,11 +132,10 @@ class UpstashVector(VectorStoreBase):
                     "top_k": limit,
                     "filter": filters_str or "",
                     "include_metadata": True,
-                    "namespace": self.collection_name,
                 }
                 for v in vectors
             ]
-            responses = self.client.query_many(queries=queries)
+            responses = self.client.query_many(queries=queries, namespace=self.collection_name)
             # flatten
             response = [res for res_list in responses for res in res_list]
 
--- a/tests/vector_stores/test_upstash_vector.py
+++ b/tests/vector_stores/test_upstash_vector.py
@@ -69,11 +69,11 @@ def test_search_vectors(upstash_instance, mock_index):
             {
                 "vector": vectors[0],
                 "top_k": 2,
-                "namespace": "ns",
                 "include_metadata": True,
                 "filter": 'age = 30 AND name = "John"',
             }
-        ]
+        ],
+        namespace="ns",
     )
 
     assert len(results) == 2
@@ -311,11 +311,11 @@ def test_search_vectors_empty_filters(upstash_instance):
             {
                 "vector": vectors[0],
                 "top_k": 1,
-                "namespace": "ns",
                 "include_metadata": True,
                 "filter": "",
             }
-        ]
+        ],
+        namespace="ns",
     )
 
     assert len(results) == 1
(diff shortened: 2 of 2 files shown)


alvinttang (commented):
Clean and correct fix. The Upstash `query_many` API treats `namespace` as a top-level keyword rather than a per-query dict field, so embedding it inside each `QueryRequest` dict caused a `TypeError` on single-query calls (duplicate keyword argument) and silently used the wrong namespace on multi-query calls — both are real correctness bugs. The test `test_search_vectors_namespace_not_in_query_dicts` does a good job asserting both the positive case (namespace is a top-level kwarg) and the negative case (namespace is absent from per-query dicts). One thing worth considering: passing a single namespace to all queries in a batch assumes all queries target the same collection — if a future batched cross-namespace search is ever needed, this call site will need revisiting. Also worth auditing whether `insert`, `delete`, and `fetch` methods in the same file have any similar namespace-placement patterns.

PR #3544, opened 2025-10-06
fix: handle non-serializable objects in config deepcopy (#3464)

lsvishaal:
## Problem Description

Memory initialization crashes with `TypeError: cannot pickle '_thread.lock' object` when using OpenSearch with AWS authentication (AWSV4SignerAuth). This occurs because the telemetry system attempts to deepcopy the vector store config, but AWS auth objects contain threading.Lock primitives that cannot be serialized.

This fix adds a `_safe_deepcopy_config()` utility function that:
1. Attempts standard deepcopy first (fast path for normal configs)
2. Falls back to Pydantic JSON serialization when deepcopy fails
3. Strips sensitive fields (auth, credentials, connection_class, etc.) for security
4. Reconstructs config with sanitized data

This approach follows the same pattern used by FastAPI and other production frameworks for handling non-serializable objects.

Fixes #3464

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] Unit Test

### Test Coverage:
1. **`test_safe_deepcopy_config_handles_opensearch_auth`** - Verifies the fix handles AWS auth objects with threading locks. Mock AWS auth raises TypeError on deepcopy, safe deepcopy succeeds and strips sensitive fields.

2. **`test_safe_deepcopy_config_normal_configs`** - Ensures no regression for normal configs without auth objects. Verifies fast path (standard deepcopy) works correctly.

3. **`test_memory_initialization_opensearch_aws_auth`** - End-to-end test that Memory initialization with OpenSearch + AWS auth succeeds without crashing.

All tests pass locally. The fix only affects telemetry config creation and has zero impact on core functionality.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

## Maintainer Checklist

- [ ] closes #3464

PR #4266, opened 2026-03-08

haroldfabla2-hue:
## Summary

Fixes bug #3580 where OpenSearch integrations fail with 403 errors after PR #3544.

## Problem

The `_safe_deepcopy_config()` function introduced in PR #3544 uses a broad sanitizer that matches any field containing "auth". This unintentionally removes `http_auth`, which is a required runtime field for OpenSearch AWS signed requests.

## Solution

Changed to match specific sensitive fields while preserving runtime auth objects:

```
sensitive_fields = ("password", "api_key", "secret", "token", "credential", "connection_class")
# Preserve http_auth which is required for clients
```

## Impact

- OpenSearch integrations using AWSV4SignerAuth now work correctly
- Telemetry safety maintained (passwords, API keys, etc. still stripped)

Fixes #3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,9 +69,12 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
-        sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
+        # Remove truly sensitive fields (exact matches or specific suffixes)
+        # But preserve runtime auth objects like http_auth which are required for clients
+        sensitive_fields = ("password", "api_key", "secret", "token", "credential", "connection_class")
         for field_name in list(clone_dict.keys()):
-            if any(token in field_name.lower() for token in sensitive_tokens):
+            # Remove if exact match or if it ends with _auth (but not http_auth)
+            if field_name.lower() in sensitive_fields or (field_name.lower().endswith("_auth") and field_name.lower() != "http_auth"):
                 clone_dict[field_name] = None
         
         try:


xkonjin (changes requested):
Appreciate the fix to preserve http_auth while scrubbing sensitive fields. There are two issues that need addressing before merge:

1) Over- and under-scrubbing risks in _safe_deepcopy_config:
   - Under-scrub: current rule preserves any field that ends with  if its lowercase name is exactly . That’s good for AWS SigV4 objects, but many configs store credentials under other names (e.g., , , , ). Consider a more robust allow/deny list:
     - Deny exact: password, api_key, secret, token, credential(s), client_secret, private_key, refresh_token, access_token
     - Allow exact: http_auth (object), ssl_context (object refs), session
     - Deny suffix: *_secret, *_token, *_password, *_key, *_credentials
   - Over-scrub: nulling  except  could break other valid auth objects (e.g., , , ). If the intent is specifically to preserve only OpenSearch http_auth, please scope the logic to that config type rather than a global rule.

2) Missing tests to lock behavior:
   - Add unit tests for _safe_deepcopy_config covering:
     - Preserves complex http_auth object (e.g., AWS SigV4 signer) but removes api_key/token fields on the same config.
     - Removes fields matching various sensitive patterns listed above.
     - Keeps non-sensitive operational fields untouched.

Optional: Consider serializing http_auth to a safe printable representation (e.g., ) when producing logs/telemetry, to avoid leaking method internals.

Once tightened and covered by tests, this will fix the build regression without introducing auth leakage or new breakage.

xkonjin (changes requested):
Appreciate the fix to preserve http_auth while scrubbing sensitive fields. There are two issues that need addressing before merge:

1) Over- and under-scrubbing risks in _safe_deepcopy_config:
   - Under-scrub: current rule preserves any field that ends with `_auth` if its lowercase name is exactly `http_auth`. That’s good for AWS SigV4 objects, but many configs store credentials under other names (e.g., `auth_header`, `bearer_token`, `service_account`, `credentials_json`). Consider a more robust allow/deny list:
     - Deny exact: password, api_key, secret, token, credential(s), client_secret, private_key, refresh_token, access_token
     - Allow exact: http_auth (object), ssl_context (object refs), session
     - Deny suffix: *_secret, *_token, *_password, *_key, *_credentials
   - Over-scrub: nulling `*_auth` except `http_auth` could break other valid auth objects (e.g., `aws_auth`, `basic_auth`, `oauth2_auth`). If the intent is specifically to preserve only OpenSearch http_auth, please scope the logic to that config type rather than a global rule.

2) Missing tests to lock behavior:
   - Add unit tests for _safe_deepcopy_config covering:
     - Preserves complex http_auth object (e.g., AWS SigV4 signer) but removes api_key/token fields on the same config.
     - Removes fields matching various sensitive patterns listed above.
     - Keeps non-sensitive operational fields untouched.

Optional: Consider serializing http_auth to a safe printable representation (e.g., `type: SigV4, region: us-east-1`) when producing logs/telemetry, to avoid leaking method internals.

Once tightened and covered by tests, this will fix the build regression without introducing auth leakage or new breakage.

utkarsh240799:
Hi @haroldfabla2-hue  , 

Thanks again for your effort!

PR #4272, opened 2026-03-09
fix(openmemory): fix settings page not saving — 3 UI/API bugs

iamvaleriofantozzi:
## Summary

The OpenMemory settings page had three compounding bugs that made it impossible to save configuration changes. Each bug independently breaks the save flow; together they made the settings page completely non-functional.

---

## Bug 1 — `docker-compose.yml`: missing default for `NEXT_PUBLIC_API_URL`

**File**: `openmemory/docker-compose.yml`

**Root cause**: The environment variable was passed with no fallback:
```yaml
- NEXT_PUBLIC_API_URL=${NEXT_PUBLIC_API_URL}
```
If `NEXT_PUBLIC_API_URL` is unset in the host shell (the common case for new installs), the container receives an empty string. `entrypoint.sh` then replaces the placeholder with `""`, so every API call from the UI targets port 3000 (the UI itself) instead of port 8765 (the API).

**Fix**: Add the standard shell fallback syntax:
```yaml
- NEXT_PUBLIC_API_URL=${NEXT_PUBLIC_API_URL:-http://localhost:8765}
```

---

## Bug 2 — `useConfig.ts`: missing trailing slash causes 307 redirect

**File**: `openmemory/ui/hooks/useConfig.ts`

**Root cause**: `fetchConfig` (GET) and `saveConfig` (PUT) called `/api/v1/config` without a trailing slash. FastAPI's default `redirect_slashes=True` responds with a `307 Temporary Redirect`. Axios follows the redirect but the Redux `dispatch(setConfigSuccess(...))` path is never reached for the original call, so the store stays on its hardcoded OpenAI initial state.

**Fix**: Add trailing slash to both calls → `/api/v1/config/`

---

## Bug 3 — `config.py`: `PUT /api/v1/config/` missing save, reset and return

**File**: `openmemory/api/app/routers/config.py`

**Root cause**: The `update_configuration` handler updated `updated_config` in memory but never:
1. Persisted it to the database (`save_config_to_db`)
2. Reset the in-memory mem0 client (`reset_memory_client`)
3. Returned the updated config (implicit `return None` → `ResponseValidationError` from FastAPI)

All three other write endpoints (`PATCH`, `PUT /mem0/llm`, `PUT /mem0/embedder`, etc.) correctly call all three. The `PUT /` handler was simply incomplete.

**Fix**: Add the three missing lines after `updated_config["mem0"] = config.mem0.dict(exclude_none=True)`:
```python
save_config_to_db(db, updated_config)
reset_memory_client()
return updated_config
```

---

## Tests

Added `openmemory/api/tests/test_config_api.py` (10 tests, all passing) that:
- Documents the 307 redirect root cause for Bug 2 (test asserts on the redirect itself)
- Verifies the fixed URLs return 200
- Asserts the response shape matches what the Redux store expects
- Runs with SQLite in-memory + StaticPool — no Docker, no Qdrant, no external services required

```
pytest openmemory/api/tests/test_config_api.py -v
# 10 passed in ~0.4s
```

---

## Files changed

| File | Change |
|------|--------|
| `openmemory/docker-compose.yml` | Add `:-http://localhost:8765` fallback to `NEXT_PUBLIC_API_URL` |
| `openmemory/ui/hooks/useConfig.ts` | Add trailing slash to GET and PUT config calls |
| `openmemory/api/app/routers/config.py` | Add `save_config_to_db`, `reset_memory_client`, `return` to PUT handler |
| `openmemory/api/tests/test_config_api.py` | New: 10-test suite covering all three bugs |

### DIFF
--- a/openmemory/api/app/routers/config.py
+++ b/openmemory/api/app/routers/config.py
@@ -154,7 +154,11 @@ async def update_configuration(config: ConfigSchema, db: Session = Depends(get_d
     
     # Update mem0 settings
     updated_config["mem0"] = config.mem0.dict(exclude_none=True)
-    
+
+    save_config_to_db(db, updated_config)
+    reset_memory_client()
+    return updated_config
+
 
 @router.patch("/", response_model=ConfigSchema)
 async def patch_configuration(config_update: ConfigSchema, db: Session = Depends(get_db)):
--- a/openmemory/api/tests/test_config_api.py
+++ b/openmemory/api/tests/test_config_api.py
@@ -1,0 +1,251 @@
+"""
+Tests for the two bugs fixed in fix/ui-config-api-url.
+
+Setup strategy:
+  sqlite:///:memory: gives EACH NEW CONNECTION an empty database.
+  We force ALL connections (app + test) to share one in-memory DB by
+  patching app.database with a StaticPool engine BEFORE importing main.
+
+  Order matters:
+    1. Set DATABASE_URL env var
+    2. Stub noisy external deps (categorization, mcp_server, mem0)
+    3. Import app.database and patch engine + SessionLocal → StaticPool
+    4. Import models (registers table metadata on Base)
+    5. Base.metadata.create_all → tables exist in shared DB
+    6. Import main → create_default_user/app run successfully
+    7. Override FastAPI get_db dependency → routes use same SessionLocal
+
+No Docker, no Qdrant, no Ollama required.
+"""
+
+import os
+import sys
+from unittest.mock import MagicMock
+
+# ── 1. Must be set before any app import ──────────────────────────────────
+os.environ["DATABASE_URL"] = "sqlite:///:memory:"
+os.environ.setdefault("USER", "testuser")
+os.environ.setdefault("API_KEY", "test-key")
+
+# ── 2. Stub external dependencies before they are imported ────────────────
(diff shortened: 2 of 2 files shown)


iamvaleriofantozzi:
CLA signed!

alvinttang (commented):
The three-part fix is solid — the missing `save_config_to_db` + `reset_memory_client()` call was the real data-loss bug, and the trailing-slash fix in `useConfig.ts` is the correct remedy for the 307 redirect (rather than disabling FastAPI's `redirect_slashes`). The docker-compose default `${NEXT_PUBLIC_API_URL:-http://localhost:8765}` is a good DX improvement, though it silently succeeds even when the API container is on a different host/port — a startup health-check or warning log might save future confusion. The test suite is unusually thorough for a UI bug fix, and the StaticPool trick to share a single in-memory SQLite connection across the app and test sessions is the right approach. One nit: the PUT handler now returns `updated_config` directly, but it's worth confirming the PATCH handler below also calls `reset_memory_client()` on success, otherwise the client state can still diverge after a partial update.

kartik-mem0:
Hi, thanks for tackling the settings page bugs — the 3-fix approach and test suite are great work. We're sunsetting the OpenMemory components from this monorepo though, so we won't be merging these changes. Appreciate the thorough contribution. Closing this out.

PR #4274, opened 2026-03-09
fix(embeddings): add baseURL support to OpenAIEmbedder

### LINKED ISSUE #4268: [Bug] TS SDK: OpenAIEmbedder ignores custom baseURL — cannot use OpenRouter or other OpenAI-compatible endpoints
### 🐛 Describe the bug

The TypeScript SDK's `OpenAIEmbedder` class does not pass `baseURL` (or `url`) to the OpenAI client constructor. This makes it impossible to use OpenAI-compatible embedding endpoints like OpenRouter, Azure, or self-hosted providers.

In `dist/oss/index.mjs` (line ~64), the constructor is:
```js
this.openai = new OpenAI({ apiKey: config.apiKey });
```

Compare this with `OpenAILLM` (line ~148), which correctly passes baseURL:
```js
this.openai = new OpenAI2({ apiKey: config.apiKey, baseURL: config.baseURL });
```

Additionally, the `ConfigManager.mergeConfig()` embedder section (line ~3700) only passes through `url`, not `baseURL`:
```js
return {
    apiKey: ...,
    model: finalModel,
    url: userConf?.url,
    embeddingDims: ...,
    modelProperties: ...
};
```

### To reproduce

Configure mem0 OSS mode with an OpenAI-compatible provider for embeddings:
```json
"embedder": {
  "provider": "openai",
  "config": {
    "apiKey": "${OPENROUTER_API_KEY}",
    "baseURL": "https://openrouter.ai/api/v1",
    "model": "openai/text-embedding-3-large"
  }
}
```

Any `search()` or `add()` call fails with:
`Error: 401 Incorrect API key provided: sk-or-v1***. You can find your API key at https://platform.openai.com/account/api-keys.`

The request goes to `api.openai.com` instead of the configured endpoint.

### Expected behavior

`OpenAIEmbedder` should respect a custom base URL, just like `OpenAILLM` already does.

### Fix

One-line change in `src/oss/src/embeddings/openai.ts`:
```diff
- this.openai = new OpenAI({ apiKey: config.apiKey });
+ this.openai = new OpenAI({ apiKey: config.apiKey, baseURL: config.baseURL || config.url });
```

And the `ConfigManager` embedder merge should also pass through `baseURL`.

### Environment

- `mem0ai` npm: 2.2.4
- `@mem0/openclaw-mem0`: 0.1.2
- Node.js: v22.22.0
- OS: Ubuntu 24

### Note

Issue #1144 was closed as fixed, but that fix only applied to the Python SDK. **The TypeScript SDK still has this bug.**

haroldfabla2-hue:
## Summary

Fixed issue #4268 - OpenAIEmbedder now supports custom baseURL for OpenRouter, Ollama, or other OpenAI-compatible endpoints.

## Changes

- Added  support to  constructor
- Also supports  as alias (matching  pattern)

## Testing

The fix allows configuring the embedder like:

```typescript
const embedder = new OpenAIEmbedder({
  baseURL: "https://openrouter.ai/api/v1",
  apiKey: process.env.OPENROUTER_API_KEY,
  model: "sentence-transformers/all-MiniLM-L6-v2"
});
```

## Related

Fixes #4268

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/openai.ts
+++ b/mem0-ts/src/oss/src/embeddings/openai.ts
@@ -8,7 +8,16 @@ export class OpenAIEmbedder implements Embedder {
   private embeddingDims?: number;
 
   constructor(config: EmbeddingConfig) {
-    this.openai = new OpenAI({ apiKey: config.apiKey });
+    const openaiConfig: ConstructorParameters<typeof OpenAI>[0] = {
+      apiKey: config.apiKey,
+    };
+    
+    // Support custom baseURL for OpenRouter, Ollama, or other OpenAI-compatible endpoints
+    if (config.baseURL || config.url) {
+      openaiConfig.baseURL = config.baseURL || config.url;
+    }
+    
+    this.openai = new OpenAI(openaiConfig);
     this.model = config.model || "text-embedding-3-small";
     this.embeddingDims = config.embeddingDims || 1536;
   }


kartik-mem0:
@haroldfabla2-hue thank you so much for the pr.

but #4275 was a better solution so we went with it.

closing the pr.

PR #4276, opened 2026-03-09
fix(openclaw): migrate platform search to mem0 v2 API

utkarsh240799:
## Summary
- Migrated the OpenClaw plugin's `PlatformProvider.search()` from mem0 v1 (`/v1/memories/search/`) to v2 (`/v2/memories/search/`) search endpoint
- Moved `user_id` and `run_id` into a structured `filters` object as required by the v2 API
- Renamed `reranking` → `rerank` to match the v2 parameter name
- Removed `source` from filters as it is not a valid v2 filter field

## Checklist
- [x] Tested `openclaw mem0 search "<query>"` — returns results correctly
- [x] Tested `--scope long-term` and `--scope all` search scopes
- [x] End-to-end: added a memory via SDK → searched via CLI → found it → deleted it
- [x] Verified `openclaw mem0 stats` is unaffected
- [x] OSS provider (`OssProvider.search()`) unchanged — uses local SDK, not platform API

### DIFF
--- a/openclaw/CHANGELOG.md
+++ b/openclaw/CHANGELOG.md
@@ -1,0 +1,33 @@
+# Changelog
+
+All notable changes to the `@mem0/openclaw-mem0` plugin will be documented in this file.
+
+## [0.2.0] - 2026-03-09
+
+### Added
+- "Understanding userId" section in docs clarifying that `userId` is user-defined
+- Per-agent memory isolation for multi-agent setups via `agentId`
+- Regression tests for per-agent isolation helpers
+
+### Changed
+- Updated config examples to use concrete `userId` values instead of placeholders
+
+### Fixed
+- Migrated platform search to Mem0 v2 API
+
+## [0.1.2] - 2026-02-19
+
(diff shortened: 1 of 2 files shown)


openclaw/index.ts
```diff
@@ -163,13 +163,17 @@ class PlatformProvider implements Mem0Provider {
...
     await this.ensureClient();
-    const opts: Record<string, unknown> = { user_id: options.user_id };
-    if (options.run_id) opts.run_id = options.run_id;
+    const filters: Record<string, unknown> = { user_id: options.user_id };
+    if (options.run_id) filters.run_id = options.run_id;
+
+    const opts: Record<string, unknown> = {
+      api_version: "v2",
+      filters,
+    };
```
deshraj, line 172: Now that we have the per agent memory as well,  you may want to make changes to that well.

deshraj: Also please bump the plugin version so that we can publish a new version of the npm package.

PR #4284, opened 2026-03-10
fix: use platform-appropriate user data dir instead of /tmp for vector store default path

### LINKED ISSUE #4279: Vector store defaults to /tmp/{provider} - fails with system services and restricted environments
**Title:** Vector store defaults to `/tmp/{provider}` - fails with system services and restricted environments

---

## Summary
The mem0ai OSS SDK defaults vector store paths to `/tmp/{provider}` when no explicit path is configured (`mem0/vector_stores/configs.py`, lines 69-70). This causes initialization failures when running as a system service (macOS LaunchAgents, Linux systemd, Windows Task Scheduler) or in restricted environments where `/tmp` may not be writable or is ephemeral.

## Problem
The default path `/tmp/{provider}` causes issues in several deployment scenarios:

- **macOS LaunchAgent:** Limited write permissions, `/tmp` cleared on reboot → data loss
- **Linux systemd:** `/tmp` may have `noexec` restrictions or use private namespaces
- **Windows:** `/tmp` path doesn't exist (should use `%TEMP%`)
- **Docker:** `/tmp` is ephemeral unless mounted as a volume → data loss on restart

## Affected Code
**File:** [`mem0/vector_stores/configs.py`](https://github.com/mem0ai/mem0/blob/main/mem0/vector_stores/configs.py#L69-L70)

```python
# also check if path in allowed kays for pydantic model, and whether config extra fields are allowed
if "path" not in config and "path" in config_class.__annotations__:
    config["path"] = f"/tmp/{provider}"
```

This affects all vector stores with a `path` field: Qdrant (local), ChromaDB, FAISS, etc.

## Reproduction Steps
1. Install mem0ai:
   ```bash
   pip install mem0ai
   ```

2. Create a simple app without explicit path:
   ```python
   from mem0 import Memory
   memory = Memory()  # Defaults to Qdrant at /tmp/qdrant
   memory.add("Test memory", user_id="test")
   ```

3. Run as a system service (e.g., macOS LaunchAgent):
   ```xml
   <?xml version="1.0" encoding="UTF-8"?>
   <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
   <plist version="1.0">
   <dict>
       <key>Label</key>
       <string>ai.test.mem0</string>
       <key>ProgramArguments</key>
       <array>
           <string>/usr/local/bin/python</string>
           <string>/path/to/app.py</string>
       </array>
       <key>RunAtLoad</key>
       <true/>
   </dict>
   </plist>
   ```

4. **Result:** Permission errors or data loss on reboot

## Current Workaround
Manually configure the path:

```python
from mem0 import Memory
import os

memory = Memory(
    config={
        "vector_store": {
            "provider": "qdrant",
            "config": {
                "path": os.path.expanduser("~/.local/share/mem0/qdrant")
            }
        }
    }
)
```

**Issues with this workaround:**
- Not documented in quickstart guides
- Requires understanding the internals
- Easy to miss for users following default examples
- Breaks "just works" expectation

### Option 1: Use OS-specific application data directories (Recommended)
Change `mem0/vector_stores/configs.py` to use standard app data locations:

```python
import os
from pathlib import Path

def get_default_vector_store_path(provider: str) -> str:
    """Get OS-specific default path for vector store."""
    if os.name == 'nt':  # Windows
        base = Path(os.environ.get('APPDATA', Path.home()))
    else:  # macOS/Linux
        base = Path.home() / '.local' / 'share'
    
    return str(base / 'mem0' / provider)

# In validate_and_create_config:
if "path" not in config and "path" in config_class.__annotations__:
    config["path"] = get_default_vector_store_path(provider)
```

**Benefits:**
- Persistent across reboots
- Standard location users expect (`~/.local/share/mem0/` on Linux/macOS, `%APPDATA%\mem0\` on Windows)
- Works with system services out of the box
- No permission issues

### Option 2: Make it configurable via environment variable
```python
if "path" not in config and "path" in config_class.__annotations__:
    default_path = os.environ.get(
        'MEM0_DATA_DIR',
        get_default_vector_store_path(provider)
    )
    config["path"] = default_path
```

This allows users to set `MEM0_DATA_DIR` in their service config without modifying code.

## Impact
- **Severity:** High for production deployments (services, Docker, serverless)
- **Affected Users:** Anyone running mem0 as a system service or in containerized environments
- **User Experience:** Silent data loss on reboot, unexpected permission errors

## Expected Behavior
- SDK works "out of the box" when run as a system service
- Data persists across reboots
- Uses predictable, writable OS-specific locations
- Path is configurable via environment variable or config

## Actual Behavior
- Fails with permission errors in restricted environments
- Data lost on reboot (LaunchAgents, Docker)
- Error messages don't indicate the root cause

---

**Environment:**
- mem0ai version: All versions (tested with latest)
- OS: macOS (LaunchAgent), Linux (systemd), Windows (Task Scheduler), Docker
- Deployment: System service / daemon mode

**Example Error:**
```
Error: [Errno 13] Permission denied: '/tmp/qdrant'
```

## Related Issues
- Similar issue in OpenClaw (OpenClaw/openclaw#TBD)
- Common pattern requiring `WorkingDirectory` fixes across Node.js/Python daemons

NIK-TIGER-BILL:
## Problem

The default vector store path is hardcoded to `/tmp/{provider}` in `mem0/vector_stores/configs.py`. This causes initialization failures or silent data loss in common deployment scenarios (closes #4279):

| Environment | Problem |
|---|---|
| macOS LaunchAgent | `/tmp` cleared on reboot → data loss; restricted permissions |
| Linux systemd | `/tmp` may have `noexec`, private namespaces, or restrictive ACLs |
| Windows | `/tmp` doesn't exist → `FileNotFoundError` |
| Docker (no volume) | `/tmp` is ephemeral → data lost on container restart |

## Fix

Use a cross-platform user data directory following established conventions:

```python
# Linux/macOS: $XDG_DATA_HOME/mem0/{provider}  (falls back to ~/.local/share/mem0/{provider})
# Windows:     %APPDATA%/mem0/{provider}
```

This follows the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/) on Linux, the standard `%APPDATA%` path on Windows, and a sensible fallback of `~/.local/share` elsewhere.

## Backward compatibility

Users who relied on the `/tmp/{provider}` default will need to migrate their data once. Users who already set an explicit `path` in config are unaffected (the new logic only applies when no path is provided).

Closes #4279

### DIFF
--- a/mem0/vector_stores/configs.py
+++ b/mem0/vector_stores/configs.py
@@ -1,3 +1,6 @@
+import os
+
+from pathlib import Path
 from typing import Dict, Optional
 
 from pydantic import BaseModel, Field, model_validator
@@ -60,7 +63,18 @@ class VectorStoreConfig(BaseModel):
 
         # also check if path in allowed kays for pydantic model, and whether config extra fields are allowed
         if "path" not in config and "path" in config_class.__annotations__:
-            config["path"] = f"/tmp/{provider}"
+            # Use a platform-appropriate user data directory instead of /tmp, which is
+            # ephemeral on many systems and may be unwritable in service/restricted environments.
+            # Preference order: XDG_DATA_HOME (Linux) → APPDATA (Windows) → ~/.local/share
+            xdg = os.environ.get("XDG_DATA_HOME")
+            app_data = os.environ.get("APPDATA")
+            if xdg:
+                base = Path(xdg)
+            elif app_data:
+                base = Path(app_data)
+            else:
+                base = Path.home() / ".local" / "share"
+            config["path"] = str(base / "mem0" / provider)
 
         self.config = config_class(**config)
         return self


utkarsh240799:
Hi @NIK-TIGER-BILL ,

Thanks for taking the time to come up with the fix. We appreciate it.
We have decided to come up with a comprehensive fix for the issue in PR #4556  which supersedes your PR. Therefore, closing this PR.

Thank you once again.

Issue #4173, opened 2026-03-01
Bug: Qdrant collection created with wrong vector dimension when using Ollama embeddings


dryoo:
### Environment
- OpenClaw: 2026.2.26 (bc50708)
- Plugin: openclaw-mem0
- Embedding model: `nomic-embed-text` (via Ollama)
- Vector store: Qdrant 1.17.0
- Qdrant JS client: @qdrant/js-client-rest 1.13.0 → 1.17.0 (manually upgraded)

### Problem

When using `openclaw mem0 search`, the command fails with:

```
Search failed: Error: Bad Request
```

The root cause is that `mem0ai/dist/oss/index.js` hardcodes `1536` as the default vector dimension in multiple places:

```js
this.embeddingDims = config.embeddingDims || 1536;
this.dimension = config.dimension || 1536;
const testVector = Array(1536).fill(0);
dimension: 1536  // line 3690, not overridable via config
```

`1536` is the dimension for OpenAI's `text-embedding-ada-002`, but `nomic-embed-text` outputs **768-dimensional** vectors. This mismatch causes Qdrant to reject search requests with `Bad Request`.

Additionally, `@qdrant/js-client-rest@1.13.0` (bundled) is incompatible with Qdrant server `1.17.0`, which requires a manual upgrade of the client.

### Steps to Reproduce

1. Install openclaw-mem0 plugin
2. Configure Qdrant as vector store with `nomic-embed-text` as embedding model
3. Run `openclaw mem0 search "anything"`
4. → `Search failed: Error: Bad Request`

### Root Cause

`mem0ai` was originally designed around OpenAI embeddings (1536 dims). When Ollama support was added, the default dimension was never updated to match models like `nomic-embed-text` (768 dims).

### Workaround

Manually patch `node_modules/mem0ai/dist/oss/index.js`:

```bash
sed -i '' 's/config.embeddingDims || 1536/config.embeddingDims || 768/g' ~/.openclaw/extensions/openclaw-mem0/node_modules/mem0ai/dist/oss/index.js
sed -i '' 's/config.dimension || 1536/config.dimension || 768/g' ~/.openclaw/extensions/openclaw-mem0/node_modules/mem0ai/dist/oss/index.js
sed -i '' 's/Array(1536).fill(0)/Array(768).fill(0)/g' ~/.openclaw/extensions/openclaw-mem0/node_modules/mem0ai/dist/oss/index.js
[...]

Issue #4056, opened 2026-02-15
Mem0 local qdrant error


thisisloze:
### 🐛 Describe the bug

mem0 OSS + Qdrant + OpenClaw Integration Issues
Environment
OpenClaw: v2026.2.12 (Node 22.22.0)
mem0ai: v0.1.7 (Python 3.9)
Qdrant: v1.13.0 / v1.16.3
Ollama: Local CPU inference (gemma3:1b, nomic-embed-text)
Issues Found

1. Collection Creation Race Condition
Error: Conflict: Collection memory_migrations already exists!
Location: qdrant.ts:222 in getUserId() / createCollection()
Problem: mem0 tries to create memory_migrations collection every time without checking if it exists
Impact: Plugin fails on every recall operation after first startup

2. Vector Dimension Mismatch in Telemetry
Error: Vector dimension error: expected dim: 768, got 1
Location: qdrant.ts:246 in upsertPoints() during _initializeTelemetry()
Problem: Telemetry code hardcodes dimension 1, but collection was created with config's 768 dimensions
Root Cause: memory_migrations collection for telemetry uses different dimensions than openclaw_memory

3. Qdrant Client/Server Version Compatibility
Warning: Client version 1.13.0 is incompatible with server version 1.16.3
Client library bundled with OpenClaw plugin is outdated
checkCompatibility config option not exposed/passed through

4. Missing Provider Support
Error: Unsupported vector store provider: chroma/chromadb/chroma-db
OpenClaw mem0 plugin only supports Qdrant for local mode
ChromaDB works with mem0 OSS directly but not through OpenClaw plugin

Expected Behavior
Plugin should check if collections exist before creating
Telemetry should use same embedding dimensions as main config
Version compatibility should be handled gracefully
ChromaDB should be supported as alternative to Qdrant

Workarounds Attempted
Pre-creating collections with correct dimensions → fails due to hardcoded dim=1 in telemetry
Deleting collections on restart → fails due to race condition
Downgrading Qdrant server → version warning persists, dimension mismatch remains
Using checkCompatibility: false → not exposed in config schema

Result
Unable to use mem0 OSS with OpenClaw due to unresolvable collection management bugs in the Qdrant integration. Switching back to cloud API mode.

danudata:
Any updates here?

utkarsh240799:
> Any updates here?
PR for Issue 1 and 2 is in review. WIll create seperate issue for Issues 3 and 4.

[...]

PR #4297, opened 2026-03-11
fix(oss): auto-detect embedding dimension to fix Qdrant mismatch with non-OpenAI embedders

### LINKED ISSUE #4212: Qdrant collection created with wrong dimensions when using Ollama embedder (nomic-embed-text)

utkarsh240799:
## Problem

Users using non-OpenAI embedding providers (Ollama/nomic-embed-text, Google/Gemini, HuggingFace, etc.) hit `Bad Request` errors when using Qdrant because the vector store was always created with a hardcoded default dimension of 1536 (OpenAI's dimension), regardless of the actual embedding model's output dimension.

This caused three related issues:
- **#4212** — Ollama + nomic-embed-text (768 dims) → Qdrant collection created with 1536 dims → `Bad Request` on insert
- **#4173** — Same root cause with different embedding providers
- **#4056** — Race condition during Qdrant collection creation + `memory_migrations` collection using wrong dimension

### 1. Probe-based dimension auto-detection
When no explicit dimension is provided, the `Memory` class now embeds a short probe string (`"dimension probe"`) at initialization time and reads the resulting vector length to determine the correct dimension. This happens **before** the vector store is created.

**Dimension resolution order:**
1. Explicit `vectorStore.config.dimension` (user override)
2. `embedder.config.embeddingDims` (embedder config)  
3. Auto-detected via probe embedding (new fallback)

### 2. Deferred vector store creation
The vector store is no longer created in the `Memory` constructor. Instead, it's created in an async `_autoInitialize()` method **after** the dimension is known. All public Memory methods (`add`, `search`, `getAll`, etc.) await initialization before proceeding via a lazy init gate (`_ensureInitialized()`).

### 3. Atomic Qdrant collection creation
Replaced the check-then-create pattern (TOCTOU race) with create-and-catch-409:
```typescript
try {
  await this.client.createCollection(name, { vectors: { size, distance: "Cosine" } });
} catch (error) {
  if (error?.status === 409) {
    // Collection already exists — verify dimension matches
  } else {
    throw error;
  }
}
```

### 4. Idempotent init guards for all vector stores
Every vector store with async initialization (Qdrant, Redis, Supabase, AzureAISearch, Vectorize) now uses a `_initPromise` singleton guard. This prevents double-initialization when `Memory` explicitly calls `await vectorStore.initialize()` after the constructor already fired it:

```typescript
private _initPromise?: Promise<void>;

async initialize(): Promise<void> {
  if (!this._initPromise) {
    this._initPromise = this._doInitialize();
  }
  return this._initPromise;
}
```

Without this guard:
- **Redis**: Double `connect()` → `"Socket already opened"`, plus `dropIndex`/`createIndex` destroys the index the first call just built
- **Qdrant**: Double `createCollection` → `409 Conflict` crash
- **Vectorize**: Double `indexes.create()` → Cloudflare API error
- **Supabase/AzureAISearch**: Redundant network calls and potential transient failures

### 5. Google embedder fix
Fixed hardcoded `outputDimensionality: 768` in `google.ts` — now uses `this.embeddingDims` from config.

## Files Changed

| File | Change |
|---|---|
| `src/oss/src/memory/index.ts` | Deferred vector store creation, lazy init gate, probe-based dimension detection |
| `src/oss/src/config/manager.ts` | Dimension resolution: explicit > embeddingDims > undefined (triggers probe) |
| `src/oss/src/vector_stores/qdrant.ts` | Atomic create-and-catch-409, idempotent init, transient error resilience |
| `src/oss/src/vector_stores/redis.ts` | Idempotent init guard |
| `src/oss/src/vector_stores/supabase.ts` | Idempotent init guard |
| `src/oss/src/vector_stores/azure_ai_search.ts` | Idempotent init guard |
| `src/oss/src/vector_stores/vectorize.ts` | Idempotent init guard |
| `src/oss/src/embeddings/google.ts` | Use `embeddingDims` config instead of hardcoded 768 |

### Unit Tests (61 tests)
- **`dimension-autodetect.test.ts`** (22 tests) — ConfigManager dimension resolution, MemoryVectorStore backward compat, Memory auto-initialization, error propagation
- **`config-manager.test.ts`** (5 tests) — ConfigManager dimension priority chain
- **`vector-stores-compat.test.ts`** (34 tests) — Backward compatibility for all 7 vector store implementations (MemoryVectorStore, Qdrant, Redis, Supabase, AzureAISearch, Vectorize, Langchain) with mocked clients

### End-to-End Tests (21 tests, against real instances)
- **`qdrant-e2e.test.ts`** (8 tests) — Against real Qdrant v1.13.0:
  - Reproduces the exact Bad Request error from #4212/#4173
  - Verifies 768-dim CRUD works after fix
  - Memory auto-detects dims via probe (full integration)
  - Explicit dimension and embeddingDims skip probe (backward compat)
  - Concurrent `ensureCollection` race condition (5 parallel instances)
  - `getUserId`/`setUserId` after concurrent init
  - `memory_migrations` uses dim=1 independently of main collection
- **`redis-e2e.test.ts`** (13 tests) — Against real Redis Stack:
  - Idempotent init (multiple concurrent calls)
  - Full CRUD: insert, search, get, update, delete
  - List with filters
  - `getUserId`/`setUserId` roundtrip
  - Correct dimension propagation from config

## Checklist

- [x] Fixes #4212, #4173, #4056
- [x] Backward compatible — explicit `dimension` config still works, no probe triggered
- [x] All existing tests pass
- [x] New unit tests for dimension resolution and backward compatibility
- [x] End-to-end tests against real Qdrant and Redis Stack instances
- [x] No breaking changes to public API
- [x] All vector store implementations verified (Qdrant, Redis, Supabase, AzureAISearch, Vectorize, Langchain, MemoryVectorStore)
- [x] Race conditions handled (concurrent init, TOCTOU collection creation)
- [x] Error messages are helpful (probe failure tells user to set dimension explicitly)

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/qdrant.ts
+++ b/mem0-ts/src/oss/src/vector_stores/qdrant.ts
@@ -305,6 +305,9 @@ export class Qdrant implements VectorStore {
             }
             // Transient errors (e.g. 500 while collection is being committed)
             // are non-fatal — the collection exists per the 409.
+            console.warn(
+              `Collection '${name}' exists (409) but dimension verification failed: ${verifyError?.message || verifyError}. Proceeding anyway.`,
+            );
           }
         }
         // Otherwise collection exists and is fine — proceed
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -140,11 +140,23 @@ export class Memory {
   /**
    * Ensures that auto-initialization (dimension detection + vector store
    * creation) has completed before any public method proceeds.
+   * If a previous init attempt failed, retries automatically.
    */
   private async _ensureInitialized(): Promise<void> {
     await this._initPromise;
     if (this._initError) {
-      throw this._initError;
+      // Clear failed state and retry — the embedder or vector store
+      // may have been transiently unavailable at startup.
+      this._initError = undefined;
+      this._initPromise = this._autoInitialize().catch((error) => {
+        this._initError =
+          error instanceof Error ? error : new Error(String(error));
+        console.error(this._initError);
+      });
+      await this._initPromise;
+      if (this._initError) {
+        throw this._initError;
+      }
     }
   }
 
--- a/mem0-ts/src/oss/tests/qdrant-e2e.test.ts
+++ b/mem0-ts/src/oss/tests/qdrant-e2e.test.ts
@@ -5,19 +5,47 @@
  * Requires a running Qdrant instance at localhost:6333 (v1.13.x).
  * These tests replicate the exact scenarios from issues #4212, #4173, #4056.
  *
+ * Skipped automatically when Qdrant is not available.
+ *
  * Run: npx jest --config jest.config.js src/oss/tests/qdrant-e2e.test.ts --forceExit
  */
 
 import { QdrantClient } from "@qdrant/js-client-rest";
 import { Qdrant } from "../src/vector_stores/qdrant";
 import { v4 as uuidv4 } from "uuid";
-
 jest.setTimeout(30000);
 
 const QDRANT_HOST = "localhost";
 const QDRANT_PORT = 6333;
 
+// Check if Qdrant is reachable synchronously at load time using
+// a sync check via child_process so describe.skip works correctly.
+function isQdrantAvailable(): boolean {
+  try {
+    const { execSync } = require("child_process");
+    execSync(
+      `node -e "const s=require('net').createConnection({host:'${QDRANT_HOST}',port:${QDRANT_PORT}});s.on('connect',()=>{s.destroy();process.exit(0)});s.on('error',()=>process.exit(1));s.setTimeout(2000,()=>process.exit(1))"`,
+      { timeout: 3000, stdio: "ignore" },
+    );
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+const qdrantAvailable = isQdrantAvailable();
+if (!qdrantAvailable) {
+  console.warn("Qdrant not available at localhost:6333 — skipping e2e tests");
+}
+
 let qdrantClient: QdrantClient;
+
+beforeAll(async () => {
+  if (!qdrantAvailable) return;
+  qdrantClient = new QdrantClient({ host: QDRANT_HOST, port: QDRANT_PORT });
+  const collections = await qdrantClient.getCollections();
+  expect(collections).toBeDefined();
+});
 
 // Helper: delete a collection if it exists
 async function deleteCollectionIfExists(name: string) {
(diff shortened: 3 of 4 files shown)


whysosaket:
Once init fails, the Memory instance is permanently dead... no retry, no recovery. If the embedder was briefly down at startup, user has to recreate everything.

mem0-ts/src/oss/src/vector_stores/qdrant.ts
```diff
@@ -286,66 +273,58 @@ export class Qdrant implements VectorStore {
     }
   }
 
-  async initialize(): Promise<void> {
+  private async ensureCollection(
```
whysosaket, line 276: ensureCollection silently swallows all non-dimension errors after a 409 — auth failures, network issues, etc. would be invisible.

whysosaket:
E2E tests hard-fail without local Qdrant/Redis and have no skip mechanism.

whysosaket (dismissed):

PR #4298, opened 2026-03-11
fix(redis): cast vector_distance to float to prevent TypeError on threshold search

### LINKED ISSUE #4294: Redis vector_distance returned as string causes threshold search to fail
### 🐛 Describe the bug

When using Redis as the vector store and passing a `threshold` parameter to `memory.search()`, 
mem0 throws a TypeError because `vector_distance` is returned as a string by redisvl instead of a float.

### Error
```
```

### Environment
- mem0
- Vector store: Redis
- Embedder: Google Gemini (`gemini-embedding-2-preview`)

### Steps to Reproduce
2. Call `memory.search()` with a `threshold` parameter:
```python
results = memory.search(
    query, 
    user_id=user_id,
    threshold=0.2,
)
```
3. Observe the TypeError

### Root Cause
In `mem0/vector_stores/redis.py`, the `search` method passes `result["vector_distance"]` 
directly as the score without casting it to a float:
```python
# Current (broken)
```

redisvl returns `vector_distance` as a string, so any downstream float comparison fails.

### Fix
```python
# Fixed
```

### Workaround
Remove the `threshold` parameter from `memory.search()` and filter manually:
```python
results = memory.search(query, user_id=user_id)
if isinstance(results, dict) and "results" in results:
    results["results"] = [r for r in results["results"] if float(r.get("score", 0)) >= 0.2]

### PR DESCRIPTION
**Pull request by NIK-TIGER-BILL**
## Problem

`redisvl` returns `vector_distance` as a **string**, not a float. When `memory.search()` is called with a `threshold` parameter, mem0 compares the score to the threshold:

```python
score=result["vector_distance"]  # str, e.g. '0.123456'
```

This causes (closes #4294):
```
TypeError: '>=' not supported between instances of 'str' and 'float'
```

## Fix

Wrap the `vector_distance` value in `float()` before assigning it to `score`:

```python
score=float(result["vector_distance"])  # ✅
```

One-line change, no behaviour change for valid distances.

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -159,7 +159,7 @@ class RedisDB(VectorStoreBase):
         return [
             MemoryResult(
                 id=result["memory_id"],
-                score=result["vector_distance"],
+                score=float(result["vector_distance"]),
                 payload={
                     "hash": result["hash"],
                     "data": result["memory"],


utkarsh240799:

PR #4315, opened 2026-03-13
fix: remove destructive vector_store.reset() from delete_all()

### LINKED ISSUE #3785: Bug: memories table was dropped by this reset call
https://github.com/mem0ai/mem0/blob/cc2894aaec8e8e981f1ca98fddc047d98c7383aa/mem0/memory/main.py#L1053

why reset here??

giulio-leone:
## Summary

Fixes #3785

`delete_all(user_id=...)` is meant to delete only memories matching the provided filters. However, after deleting individual memories it called `self.vector_store.reset()`, which **drops the entire collection/table** — destroying ALL memories regardless of the filter.

## Root Cause

Line 1053 in `mem0/memory/main.py`:

```python
# Before (broken):
for memory in memories:
self.vector_store.reset()  # ← drops ENTIRE collection!
```

The async version of `delete_all()` (line 2084) already omits this call, confirming it was unintentional in the sync path.

## Fix

Removed the `self.vector_store.reset()` call. Individual memory deletions via `self._delete_memory(memory.id)` already handle removal correctly.

## Test

Added regression assertion `vector_store.reset.assert_not_called()` to the existing `test_delete_all` test to prevent this from regressing.

**Both test parametrizations pass.**

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1046,11 +1046,10 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
+        # delete matching vector memories individually (do NOT reset the collection)
         memories = self.vector_store.list(filters=filters)[0]
         for memory in memories:
             self._delete_memory(memory.id)
-        self.vector_store.reset()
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
hi @giulio-leone thank you for pr but can you sign the cla here.

giulio-leone:
@kartik-mem0 Thanks for the reminder! I'll sign the CLA now and follow up. Apologies for the delay.

kartik-mem0:
thank you for your contribution. closing this pr #4349 for this pr as it is more focused towards the issues.

PR #4316, opened 2026-03-13
fix: add pgvector support in VectorStoreFactory (issue #3491)

### LINKED ISSUE #3491: Issue with NodeJs OSS version
### 🐛 Describe the bug

There are several issues with NodeJs oss version of mem0

1. when using Supabase
getting below error everytime i followed exactly same steps as mentioned in supabase provider docs of mem0
```ts
Error getting user ID: {
  code: 'PGRST205',
  details: null,
  hint: "Perhaps you meant the table 'public.memories'",
 message: "Could not find the table 'public.memory_migrations' in the schema cache"
}
```
2. When using Redis
coud not configure the redis
```ts
Error creating index cannot read properties of undefined reading toString
```
3. When using pgvector
Documentation suggests ts can also use pg vector but when adding config the vector is not working.
i digged into the code base there is pgvector class but in Factory method pgvector is not considered https://github.com/mem0ai/mem0/blob/88ae952483aa85811d28c47722b123d795cec68a/mem0-ts/src/oss/src/utils/factory.ts#L84
https://github.com/mem0ai/mem0/blob/88ae952483aa85811d28c47722b123d795cec68a/mem0-ts/src/oss/src/utils/factory.ts#L99
```ts
Unsupported vector store provider:  pgvector
```

haroldfabla2-hue:
## Description

This PR adds support for `pgvector` provider in the NodeJS OSS version's `VectorStoreFactory`.

## Problem

In issue #3491, users reported that when using `pgvector` as a vector store provider in the NodeJS OSS version, they get:

```
Unsupported vector store provider: pgvector
```

This happens because the `VectorStoreFactory.create()` method in `mem0-ts/src/oss/src/utils/factory.ts` does not have a case for the `pgvector` provider, even though the `PGVector` class exists.

## Solution

1. Import the existing `PGVector` class from `../vector_stores/pgvector`
2. Add a case for `provider.toLowerCase() === "pgvector"` in the `VectorStoreFactory.create()` method

## Changes

- Modified: `mem0-ts/src/oss/src/utils/factory.ts`
  - Added import for `PGVector` class
  - Added case for `pgvector` in switch statement

## Testing

The fix has been verified by reviewing:
- The existing `PGVector` class exists at `mem0-ts/src/oss/src/vector_stores/pgvector.ts`
- The Python version already supports pgvector (see `mem0/vector_stores/configs.py`)

## Related Issues

- Fixes #3491

---
*Note: This is a good first issue - simple bug with clear fix*

### DIFF
--- a/mem0-ts/src/oss/src/utils/factory.ts
+++ b/mem0-ts/src/oss/src/utils/factory.ts
@@ -32,6 +32,7 @@ import { LangchainLLM } from "../llms/langchain";
 import { LangchainEmbedder } from "../embeddings/langchain";
 import { LangchainVectorStore } from "../vector_stores/langchain";
 import { AzureAISearch } from "../vector_stores/azure_ai_search";
+import { PGVector } from "../vector_stores/pgvector";
 
 export class EmbedderFactory {
   static create(provider: string, config: EmbeddingConfig): Embedder {
@@ -98,6 +99,8 @@ export class VectorStoreFactory {
         return new VectorizeDB(config as any);
       case "azure-ai-search":
         return new AzureAISearch(config as any);
+      case "pgvector":
+        return new PGVector(config as any);
       default:
         throw new Error(`Unsupported vector store provider: ${provider}`);
     }


kartik-mem0:
Closing this as the fix has been merged via PR #3997. Thanks for taking the time to contribute — we appreciate it!

PR #4317, opened 2026-03-13
fix(ts): extract content from code blocks instead of deleting it

### LINKED ISSUE #4108: Claude: Failed to parse facts from LLM response:  SyntaxError: Unexpected end of JSON input
### 🐛 Describe the bug

Claude ignores https://github.com/mem0ai/mem0/blob/main/mem0-ts/src/oss/src/prompts/index.ts#L262 and answered like:
```
{
  "type": "message",
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": "```json\n{\n  \"facts\": [\n    \"<saved fact here>"\n  ]\n}\n```"
    }
  ],
  "stop_reason": null,
  "stop_sequence": null,
  "usage": {
    "cache_creation_input_tokens": 0,
    "cache_read_input_tokens": 0,
    "output_tokens": null
  }
}
```

voidborne-d:
## Problem

Fixes #4108 — `removeCodeBlocks` deletes entire code blocks (including their JSON payload) instead of extracting the content inside them.

The regex `/ ```[^`]*``` /g` matches everything between the opening and closing fence markers and replaces **all of it** with an empty string. When Claude returns:

```
```json
{"facts": ["user likes TypeScript"]}
```
```

the function returns `""` and `JSON.parse` throws `SyntaxError: Unexpected end of JSON input`.

## Fix

Replace the regex with a capturing group that **extracts** the content inside code fences while stripping only the fence markers:

```ts
// Before (deletes content)
text.replace(/```[^`]*```/g, "")

// After (extracts content)
text.replace(/```(?:\w+)?\n?([\s\S]*?)```/g, "$1").trim()
```

## Tests

Added `remove-code-blocks.test.ts` covering:
- JSON in ` ```json ` fence
- Bare ` ``` ` fence
- Plain text passthrough (no fences)
- Multiple code blocks
- Claude-style response with surrounding text

### DIFF
--- a/mem0-ts/src/oss/src/prompts/index.ts
+++ b/mem0-ts/src/oss/src/prompts/index.ts
@@ -278,5 +278,9 @@ export function parseMessages(messages: string[]): string {
 }
 
 export function removeCodeBlocks(text: string): string {
-  return text.replace(/```[^`]*```/g, "");
+  // Extract content inside code fences instead of deleting it.
+  // The old regex /```[^`]*```/g replaced the entire block (including
+  // its content) with an empty string, so when an LLM returned JSON
+  // wrapped in ```json ... ``` the actual payload was discarded.
+  return text.replace(/```(?:\w+)?\n?([\s\S]*?)```/g, "$1").trim();
 }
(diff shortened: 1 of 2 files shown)


voidborne-d:
Thanks @utkarsh240799! Just signed the CLA ✅

voidborne-d:

PR #4318, opened 2026-03-13
feat(openclaw): add disableHistory passthrough for OSS mode

### LINKED ISSUE #4172: @mem0/openclaw-mem0 OSS mode: sqlite3 binding lookup fails under OpenClaw/Jiti loader context
# OpenClaw + Mem0 OSS sqlite binding root-cause report

## Summary
In OpenClaw with `@mem0/openclaw-mem0` OSS mode, memory tools can fail with:

`Could not locate the bindings file ... node_sqlite3.node`

This occurs even when `sqlite3` is installed and loadable directly from the plugin's own `node_modules`.

## Observed environment
- OpenClaw gateway runtime: Node v22.22.0 (arm64, macOS)
- Plugin: `@mem0/openclaw-mem0@0.1.2`
- Mode: OSS
- Vector store: Qdrant

## Root cause
The failing lookup paths point to OpenClaw/Jiti loader context:
- `/opt/homebrew/lib/node_modules/openclaw/node_modules/jiti/.../node_sqlite3.node`

But the binary exists and loads from plugin path:
- `~/.openclaw/extensions/openclaw-mem0/node_modules/sqlite3/build/Release/node_sqlite3.node`

So this is a runtime module-resolution path issue under loader context, not a missing compile artifact in plugin node_modules.

## Repro signal
- `require('.../openclaw-mem0/node_modules/sqlite3')` succeeds under Node 22
- plugin memory tools fail from OpenClaw runtime with Jiti-path binding lookup

## Proposed long-term fix
In `@mem0/openclaw-mem0` OSS init, avoid hard dependency on sqlite history by default in OpenClaw runtime:
- allow explicit `oss.disableHistory`
- allow explicit `oss.historyStore`
- set safer default to bypass native sqlite where needed

This preserves Mem0+Qdrant usage while avoiding native sqlite fragility.

## Local validated patch (working)
- Added `oss.disableHistory?: boolean`
- Added `oss.historyStore?: { provider: string; config: Record<string, unknown> }`
- In OSS provider init:
  - if `oss.disableHistory` is unset, default to `true`
  - if `oss.historyStore` is set, pass through to mem0 config

Result: memory tools run without sqlite binding failures in this runtime.

## Requested upstream action
- Merge a durable OSS-mode history config path that does not assume sqlite-native availability in OpenClaw/Jiti contexts.
- Optionally document recommended OSS config for OpenClaw runtimes that use transpiler loaders.

jsonmez:
## Summary

Adds `disableHistory` config option to the OSS mode configuration, allowing multi-agent OpenClaw setups to bypass SQLite history storage and avoid race conditions.

## Problem

In multi-agent OpenClaw deployments, multiple agents initialize simultaneously at gateway startup. Each agent triggers `new Memory()` which attempts to open the same SQLite history database file. This causes `SQLITE_CANTOPEN` errors due to file locking race conditions:

```
SqliteError: unable to open database file
```

The underlying `mem0ai` SDK already supports `disableHistory: true` which uses a `DummyHistoryManager` instead of SQLite, but the OpenClaw plugin doesn't pass this config through.

## Solution

Add 2 lines to enable the `disableHistory` passthrough:

1. Add `disableHistory?: boolean` to the OSSConfig interface
2. Pass through in `_init()`: `if (this.ossConfig?.disableHistory) config.disableHistory = true;`

## Usage

```json
{
  "plugins": {
    "entries": {
      "openclaw-mem0": {
        "config": {
          "mode": "open-source",
          "oss": {
            "disableHistory": true,
            "vectorStore": { "provider": "qdrant", "config": { } }
          }
        }
      }
    }
  }
}
```

## Impact

- No breaking changes for existing configurations
- Opt-in fix for multi-agent setups experiencing SQLite race conditions
- The history database is only used for deduplication tracking; all actual memories are stored in Qdrant/vector store and are unaffected

## Testing

Verified in production multi-agent OpenClaw setup (9 agents) - eliminates all `SQLITE_CANTOPEN` errors at startup.

Fixes #4172

### DIFF
--- a/openclaw/index.ts
+++ b/openclaw/index.ts
@@ -248,7 +248,9 @@ class OSSProvider implements Mem0Provider {
     }
 
     if (this.customPrompt) config.customPrompt = this.customPrompt;
-    if (this.ossConfig?.disableHistory) config.disableHistory = true;
+    if (this.ossConfig && "disableHistory" in this.ossConfig) {
+      config.disableHistory = this.ossConfig.disableHistory;
+    }
 
     this.memory = new Memory(config);
   }


utkarsh240799:
Issue addressed in PR #4337 . Closing this as duplicate.

PR #4321, opened 2026-03-13
feat: add MiniMax as a supported LLM provider (M2.7 default)

octo-patch:
## Summary
- Add MiniMax as a supported LLM provider using OpenAI-compatible API
- Default model set to **MiniMax-M2.7** (latest flagship model with enhanced reasoning and coding)
- Support custom API base URL via config or `MINIMAX_API_BASE` environment variable

## Changes
- `mem0/llms/minimax.py`: MiniMaxLLM class using OpenAI SDK with MiniMax-M2.7 default
- `mem0/configs/llms/minimax.py`: MiniMaxConfig with minimax_base_url support
- `mem0/llms/configs.py`: Register minimax as valid provider
- `mem0/utils/factory.py`: Add MiniMax to LlmFactory
- `tests/llms/test_minimax.py`: Unit tests for base URL, default model, tool calling
- `docs/components/llms/models/minimax.mdx`: Usage documentation

## Why
MiniMax-M2.7 is the latest flagship model with enhanced reasoning and coding capabilities, available via OpenAI-compatible API at `https://api.minimax.io/v1`.

## Testing
- Unit tests covering base URL resolution, response parsing, and tool calling
- Default model verification test for MiniMax-M2.7

### DIFF
--- a/LLM.md
+++ b/LLM.md
@@ -273,7 +273,7 @@ config = MemoryConfig(
 
 ### Supported Providers
 
-#### LLM Providers (19 supported)
+#### LLM Providers (20 supported)
 - **openai** - OpenAI GPT models (default)
 - **anthropic** - Claude models
 - **gemini** - Google Gemini
@@ -289,6 +289,7 @@ config = MemoryConfig(
 - **lmstudio** - LM Studio local server
 - **vllm** - vLLM inference server
 - **langchain** - LangChain integration
+- **minimax** - MiniMax models
 - **openai_structured** - OpenAI with structured output
 - **azure_openai_structured** - Azure OpenAI with structured output
 
--- a/docs/components/llms/models/minimax.mdx
+++ b/docs/components/llms/models/minimax.mdx
@@ -17,7 +17,7 @@ config = {
     "llm": {
         "provider": "minimax",
         "config": {
-            "model": "MiniMax-M1",  # default model
+            "model": "MiniMax-M2.7",  # default model
             "temperature": 0.2,
             "max_tokens": 2000,
             "top_p": 1.0
@@ -42,7 +42,7 @@ config = {
     "llm": {
         "provider": "minimax",
         "config": {
-            "model": "MiniMax-M1",
+            "model": "MiniMax-M2.7",
             "minimax_base_url": "https://your-custom-endpoint.com",
             "api_key": "your-api-key"  # alternatively to using environment variable
         }
--- a/docs/components/llms/models/minimax.mdx
+++ b/docs/components/llms/models/minimax.mdx
@@ -1,0 +1,55 @@
+---
+title: MiniMax
+---
+
+To use MiniMax LLM models, you have to set the `MINIMAX_API_KEY` environment variable. You can also optionally set `MINIMAX_API_BASE` if you need to use a different API endpoint (defaults to "https://api.minimax.io/v1").
+
+## Usage
+
+```python
+import os
+from mem0 import Memory
+
+os.environ["MINIMAX_API_KEY"] = "your-api-key"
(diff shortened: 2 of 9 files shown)


alvinttang (commented):
## Code Review: Add MiniMax as a supported LLM provider

Solid addition that follows the existing provider patterns. The implementation is clean and tests cover the main paths. Some issues:

### 1. `response_format` parameter is accepted but never used
The `generate_response` method accepts `response_format` as a parameter:
```python
def generate_response(self, messages, response_format=None, tools=None, ...):
```
But it is never passed to the API call. The `params` dict is built from `_get_supported_params` and then only `model` and `messages` are added. This means JSON mode (`response_format={"type": "json_object"}`) will silently be ignored, which is a problem because mem0's core `_add_to_vector_store` passes `response_format={"type": "json_object"}` to every LLM call. The LLM will return unstructured text instead of JSON, causing `json.loads()` to fail downstream.

Fix: add `response_format` to params when provided:
```python
if response_format:
    params["response_format"] = response_format
```

### 2. Test environment variable leak
In `test_minimax_llm_base_url`, case 2 sets:
```python
os.environ["MINIMAX_API_BASE"] = provider_base_url
```
And cleans up at the end of the function with `del os.environ["MINIMAX_API_BASE"]`. But if case 3 raises an exception before reaching cleanup, the env var leaks into subsequent tests. Use `monkeypatch.setenv` (pytest fixture) or a try/finally block instead. Also, case 2 sets the env var and then case 3 also runs with it still set — the test for `minimax_base_url` config override works because config takes precedence, but the test isn't actually isolating the cases properly.

### 3. `BaseLlmConfig` to `MiniMaxConfig` conversion maps `http_client` to `http_client_proxies`
```python
config = MiniMaxConfig(
    ...
)
```
Is `http_client` really the same thing as `http_client_proxies`? Looking at `BaseLlmConfig`, `http_client` is typically an `httpx.Client` instance, while `http_client_proxies` sounds like a proxy dict. If the types don't match, this will cause a runtime error when the proxy config is actually used. Verify the field semantics match.

### 4. Missing `api_key` validation
The constructor does:
```python
api_key = self.config.api_key or os.getenv("MINIMAX_API_KEY")
```
If neither is set, `api_key` is `None`, which gets passed to `OpenAI(api_key=None, ...)`. The OpenAI client may raise a confusing error later when a request is made. Other providers in the codebase (e.g., `openai.py`) have similar patterns, so this is consistent, but adding an early check with a clear error message would improve UX:
```python
if not api_key:
    raise ValueError("MiniMax API key not found. Set MINIMAX_API_KEY env var or pass api_key in config.")
```

### 5. `_parse_response` calls `extract_json` on tool call arguments
```python
json.loads(extract_json(tool_call.function.arguments))
```
If `tool_call.function.arguments` is already valid JSON (which it should be from the API), `extract_json` is unnecessary overhead. If it's not valid JSON, `extract_json` may strip important parts. The other LLM providers (e.g., `openai.py`) just do `json.loads(tool_call.function.arguments)` directly. Keep it consistent.

### 6. Large doc restructuring mixed into this PR
The diff includes a complete restructure of `docs/mint.json` (hundreds of lines of navigation changes). This makes the PR much harder to review and increases merge conflict risk. Consider separating the doc restructuring into its own PR.

The main issue is #1 — `response_format` being silently dropped will break mem0's core memory operations with this provider.

octo-patch:
Thanks for the detailed review @alvinttang! Good catch on the `response_format` parameter. I'll add the proper forwarding to the API call.

Regarding CLA - I'll look into signing it.

kartik-mem0:
hey thank you for your contribution, but i think this #4431 pr already address the issue.

closing this pr

PR #4323, opened 2026-03-13
feat(openmemory): expose infer param in MCP add_memories tool

### LINKED ISSUE #3322: Openmemory changes memories before they hit the vector store
### 🐛 Describe the bug

I'm using the openmemory mcp server with Claude Code.  The mcp calls all work successfully, but it looks like openmemory is changing the memories before they hit the vector store.  The changes make them unusable.  For instance - a memory claude code tried to add: `<Coworker's name> has a PhD` turned into just `has a PhD`.  When I ask Claude Code to describe me based on the memories in openmemory it says I have a PhD since my coworker's name was removed from the memory.

Is there some configuration I've done wrong or something I tweak to make this work better?  It seems like the memories should be stored as-is, and openmemory should provide search results based on them and summarize them then, if needed.  I can't see it being very useful the way it's setup now.

I've only changed the NEXT_PUBLIC_API_URL and the username in both the ui and api env.  Everything else is set to default.

evolv-to:
## Summary

- Adds `infer: bool = True` parameter to the MCP `add_memories` tool
- Passes it through to `memory_client.add()`, consistent with how the REST API already handles this (`POST /api/v1/memories/` via PR #3607)
- Default `True` preserves backward compatibility

## Problem

The MCP server's `add_memories` always uses `infer=True`, which runs LLM fact extraction that strips subjects from relational facts:

| Input | Stored (infer=True) | Stored (infer=False) |
|-------|-------------------|---------------------|
| "Bob has a PhD" | "has a PhD" | "Bob has a PhD" |
| "My coworker Alice is a doctor" | "is a doctor" | "My coworker Alice is a doctor" |

This makes memories about third parties unusable — the vector store loses *who* the fact is about.

## Fix

Two lines changed in `openmemory/api/app/mcp_server.py`:

1. **Line 61**: `add_memories(text: str)` → `add_memories(text: str, infer: bool = True)`
2. **Line 85**: Pass `infer=infer` to `memory_client.add()`

## Test plan

- [x] New unit tests verify: param exists, defaults to `True`, `infer=False` passed through, `infer=True` passed through (4/4 passing)
- [x] Existing test suite unaffected (11 pass, pre-existing graph_memory import errors unchanged)
- [ ] Manual test with Claude Code MCP client: add memory with `infer=False`, verify verbatim storage in Qdrant

Fixes #3322
Related: #3615 (closed, same intent), #3607 (merged, REST API equivalent)

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -58,7 +58,7 @@ mcp_router = APIRouter(prefix="/mcp")
 sse = SseServerTransport("/mcp/messages/")
 
 @mcp.tool(description="Add a new memory. This method is called everytime the user informs anything about themselves, their preferences, or anything that has any relevant information which can be useful in the future conversation. This can also be called when the user asks you to remember something.")
-async def add_memories(text: str) -> str:
+async def add_memories(text: str, infer: bool = True) -> str:
     uid = user_id_var.get(None)
     client_name = client_name_var.get(None)
 
@@ -87,7 +87,8 @@ async def add_memories(text: str) -> str:
                                          metadata={
                                             "source_app": "openmemory",
                                             "mcp_client": client_name,
-                                        })
+                                        },
+                                         infer=infer)
 
             # Process the response and update database
             if isinstance(response, dict) and 'results' in response:


kartik-mem0:
Thanks for adding the infer parameter — that's a useful feature. We're in the process of winding down the OpenMemory components in this repo, so this won't be merged here. Appreciate the effort though. Closing for now.

PR #4325, opened 2026-03-13
Fix: Correct typo and add tool role support

LincolnBurrows2017:
### Fix 1: Typo in graphs/utils.py
- Changed "their is" to "there is" (2 occurrences)

### Fix 2: Add tool role support in parse_messages
- Added handling for "tool" role in messages
- Previously only system, user, and assistant roles were handled

## Changes
```diff
-3. DO NOT DELETE if their is a possibility
+3. DO NOT DELETE if there is a possibility

+if msg[role] == "tool":
+    response += f"tool: {msg[content]}\n"
```

### DIFF
--- a/mem0/graphs/utils.py
+++ b/mem0/graphs/utils.py
@@ -66,7 +66,7 @@ Guidelines:
 2. Deletion Criteria: Delete a relationship only if it meets at least one of these conditions:
    - Outdated or Inaccurate: The new information is more recent or accurate.
    - Contradictory: The new information conflicts with or negates the existing information.
-3. DO NOT DELETE if their is a possibility of same type of relationship but different destination nodes.
+3. DO NOT DELETE if there is a possibility of same type of relationship but different destination nodes.
 4. Comprehensive Analysis:
    - Thoroughly examine each existing relationship against the new information and delete as necessary.
    - Multiple deletions may be required based on the new information.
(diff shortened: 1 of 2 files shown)


xkonjin (commented):
## Code Review

**Summary:** Two small, clean changes. The typo fix is straightforward. The `tool` role support is a reasonable addition, but there are a few things worth considering:

### 🟡 `tool` message content may not be a string

The `parse_messages` function casts `msg['content']` directly to a string for all roles. For `tool` role messages (OpenAI / Anthropic tool call patterns), `content` can be:
- A list of content blocks (e.g. `[{"type": "tool_result", "content": "..."}]`)
- `None` in some edge cases
- A structured dict

If a non-string value hits `f"tool: {msg['content']}"`, you'll get a stringified Python repr (`tool: [{'type': 'tool_result', ...}]`) in the output instead of meaningful text. This could confuse downstream LLM calls that consume the parsed string.

Suggestion: normalize the content similarly to how you'd handle assistant content blocks:
```python
if msg["role"] == "tool":
    content = msg["content"]
    if isinstance(content, list):
        content = " ".join(
            part.get("text", "") or str(part) for part in content if isinstance(part, dict)
        )
    response += f"tool: {content}\n"
```

### 🟡 No test coverage

There are no tests for the new `tool` role branch. Given that this feeds into memory extraction prompts, a unit test with a realistic multi-turn conversation including tool messages (both string and list-content variants) would be valuable. The existing test file for `parse_messages` is a natural place to add these.

### 💡 Consider filtering tool messages

Depending on mem0's use case, tool message contents (raw API responses, JSON payloads) might add noise to memory extraction. It may be worth a comment explaining why tool messages should be included, or an option to filter them out if the user configures it.

The typo fix is a clear improvement. The `tool` role addition is directionally right but needs content normalization to be safe.

kartik-mem0:
Thanks, @LincolnBurrows2017! Part of this targeted `mem0/graphs/utils.py`, which no longer exists — the OSS graph layer was removed in the v3 pipeline port (#4805). With the graph portion gone the diff can't apply as-is, so I'm closing it as stale. If the tool-role handling is still relevant against the current `mem0/memory/utils.py`, a small focused PR there would be very welcome.

PR #4327, opened 2026-03-14
fix: wrap vector and payload in lists in Langchain update method

gambletan:
## Summary
- Fixes type mismatch in `Langchain.update()` where `vector` and `payload` were passed directly to `self.insert()` without being wrapped in lists
- `insert()` expects `vectors: List[List[float]]` and `payloads: List[Dict]`, but `update()` was passing single values instead

## Changes
In `mem0/vector_stores/langchain.py`, changed:
```python
```
to:
```python
```

Fixes #3767

### DIFF
--- a/mem0/vector_stores/langchain.py
+++ b/mem0/vector_stores/langchain.py
@@ -115,7 +115,7 @@ class Langchain(VectorStoreBase):
         Update a vector and its payload.
         """
         self.delete(vector_id)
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Hi @gambletan, thanks for submitting a fix! This issue has been resolved via #4446. Closing this PR — appreciate your effort! 🙏

PR #4329, opened 2026-03-14
fix: preserve http_auth in _safe_deepcopy_config sanitizer

gambletan:
## Summary

Fixes #3580

`_safe_deepcopy_config()` uses substring matching to sanitize sensitive fields before telemetry serialization. Because `"auth"` is one of the tokens, it inadvertently matches `http_auth` — a runtime authentication object required by OpenSearch / AWS AOSS clients — setting it to `None` and causing 403 errors.

## Changes

- Replace substring-based sensitive token matching with an **exact field name set** (`password`, `api_key`, `secret`, `secret_key`, `credential`, `credentials`, `token`, `access_token`).
- This preserves runtime auth objects like `http_auth` and `connection_class` while still redacting genuinely sensitive scalar values before telemetry.

## Root Cause

```python
# BEFORE — substring match catches "http_auth" via "auth"
        clone_dict[field_name] = None
```

```python
# AFTER — exact match on known sensitive field names only
sensitive_fields = {"password", "api_key", "secret", "secret_key", ...}
    if field_name.lower() in sensitive_fields:
        clone_dict[field_name] = None
```

## Test plan

- [ ] Verify OpenSearch integration with `AWSV4SignerAuth` in `http_auth` no longer fails with 403
- [ ] Confirm genuinely sensitive fields (`password`, `api_key`, `token`, etc.) are still redacted
- [ ] Existing tests pass

### DIFF
--- a/tests/vector_stores/test_opensearch.py
+++ b/tests/vector_stores/test_opensearch.py
@@ -300,10 +300,10 @@ def test_safe_deepcopy_config_handles_opensearch_auth(mock_sqlite, mock_llm_fact
     
     safe_config = _safe_deepcopy_config(config_with_auth)
     
-    assert safe_config.http_auth is None
-    assert safe_config.auth is None
+    assert safe_config.http_auth is not None
+    assert safe_config.auth is not None
     assert safe_config.credentials is None
-    assert safe_config.connection_class is None
+    assert safe_config.connection_class is not None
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @gambletan , 

Thanks again for your effort!

PR #4331, opened 2026-03-14

haroldfabla2-hue:
## Summary
Bug: `_safe_deepcopy_config` was removing `http_auth` from OpenSearch configs, breaking AWS signed requests.

## Root Cause
The sanitizer token list included `auth`, which matched the field name `http_auth`, a required runtime field for OpenSearch clients.

## Solution
Only sanitize truly sensitive fields (password, token, secret, api_key) and preserve runtime auth objects like `http_auth`, `auth`, `connection_class`.

## Fixes
Fixes https://github.com/mem0ai/mem0/issues/3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,10 +69,15 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @haroldfabla2-hue , 

Thanks again for your effort!

PR #4335, opened 2026-03-14
chore(openclaw): add tsup build pipeline with ESM output and type declarations

whysosaket:
Adds a modern build setup for the openclaw plugin using tsup. Includes tsconfig, type stubs for the host SDK, and proper package.json exports/files configuration for npm publishing.

### DIFF
--- a/openclaw/package.json
+++ b/openclaw/package.json
@@ -11,7 +11,19 @@
     "mem0",
     "long-term-memory"
   ],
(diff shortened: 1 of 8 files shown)


openclaw/package.json
utkarsh240799, line 31: Can we increase the minimum version ^2.3.0 as we did the migration to better-sqlite in this version. This will force existing users to upgrade the package.

PR #4345, opened 2026-03-15
fix: preserve http_auth for OpenSearch client authentication

haroldfabla2-hue:
## Summary

This PR fixes OpenSearch integrations failing with 403 errors after PR #3544.

## Problem

The `_safe_deepcopy_config()` function was removing the `http_auth` field from config because it matched the token "auth" in the sensitive field list. This broke OpenSearch client authentication (AWSV4SignerAuth).

## Solution

Remove "auth" from the `sensitive_tokens` tuple - it is too broad and catches required runtime auth objects. The list now targets only static credentials like passwords, tokens, API keys.

## Testing

This fix preserves `http_auth` (and other runtime auth objects) while still stripping sensitive static credentials.

## Related Issue

Fixes #3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,7 +69,10 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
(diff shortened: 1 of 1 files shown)


utkarsh240799:
Hi @haroldfabla2-hue , 

Thanks again for your effort!

PR #4347, opened 2026-03-16
fix: cast vector_distance to float in Redis search

haroldfabla2-hue:
## Summary
Fixes TypeError when using the `threshold` parameter with Redis vector store.

## Problem
When using Redis as the vector store and passing a `threshold` parameter to `memory.search()`, mem0 throws a TypeError because `vector_distance` is returned as a string by redisvl instead of a float.

```python
```

## Solution

```python
# Before

# After  
```

## Changes
- Modified: `mem0/vector_stores/redis.py` (line 162)

## Testing
This fix allows the threshold comparison to work correctly:
```python
results = memory.search(query, user_id=user_id, threshold=0.2)
```

Fixes: https://github.com/mem0ai/mem0/issues/4294

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -159,7 +159,7 @@ class RedisDB(VectorStoreBase):
         return [
             MemoryResult(
                 id=result["memory_id"],
-                score=result["vector_distance"],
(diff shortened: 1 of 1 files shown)


utkarsh240799:

PR #4362, opened 2026-03-16
fix: prevent embedding corruption in Valkey and Redis when vector is None (#4336)

### LINKED ISSUE #4336: Mem0 with Valkey: update() corrupts embeddings when called with vector=None

DhilipBinny:
When `update()` is called with `vector=None` (metadata-only update), `np.array(None, dtype=np.float32).tobytes()` silently creates a 4-byte `NaN` scalar instead of raising an error.
  This overwrites the real embedding (e.g. 6144 bytes for 1536-dim models) with 4 bytes of garbage, making memories unsearchable.                                                       
   
  **Root cause:** NumPy silently coerces `None` into a scalar — no error, no warning, just silent data corruption.                                                                      
                                                                                                                                                                                      
  **Fix:** Skip the `embedding` field entirely when `vector is None`, so `hset` preserves the existing value. This matches the pattern already used by 9 other vector stores (pinecone,
  mongodb, faiss, elasticsearch, opensearch, cassandra, databricks, azure_mysql, vertex_ai).

  ### Changes
  - `mem0/vector_stores/valkey.py` — conditionally include embedding in `update()`
  - `mem0/vector_stores/redis.py` — same fix (same bug existed here too)
  - `tests/vector_stores/test_valkey.py` — 2 new tests for both cases

  Fixes #4336

  ## Type of change

  ## How Has This Been Tested?

  - [x] Unit Test

  2 new tests:
  - `test_update_with_none_vector_preserves_embedding` — verifies embedding is NOT included when `vector=None`
  - `test_update_with_vector_includes_embedding` — verifies embedding IS included when vector is provided

  All 45 Valkey tests pass. Linting passes.

  Run with: `pytest tests/vector_stores/test_valkey.py -v`

  ## Checklist:

  ## Maintainer Checklist

  - [ ] closes #4336

### DIFF
--- a/tests/vector_stores/test_redis.py
+++ b/tests/vector_stores/test_redis.py
@@ -7,24 +7,27 @@ when vector is None.
 """
 
 from datetime import datetime
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock
 
 import numpy as np
 import pytz
 
 
-@patch("mem0.vector_stores.redis.SearchIndex")
-@patch("mem0.vector_stores.redis.redis.from_url")
-def test_update_with_none_vector_preserves_embedding(mock_redis, mock_search_index):
-    """update() with vector=None should not include embedding in the data."""
-    mock_index = MagicMock()
-    mock_search_index.return_value = mock_index
-
+def _make_redis_db():
+    """Create a RedisDB instance with mocked internals, bypassing __init__
+    to avoid the redis module name collision with mem0.vector_stores.redis."""
     from mem0.vector_stores.redis import RedisDB
 
     db = RedisDB.__new__(RedisDB)
+    mock_index = MagicMock()
     db.index = mock_index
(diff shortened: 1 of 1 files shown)


DhilipBinny:
Thanks for the review! Added Redis tests in `test_redis_update.py` for parity.
                                              
Regarding other vector stores — I checked all 22 `update()` implementations. Only Redis and Valkey used `np.array().tobytes()`, which is where the bug occurs. The other 9 stores (pinecone, mongodb, faiss, elasticsearch, opensearch, cassandra, databricks, azure_mysql, vertex_ai) already had the `if vector is not None` guard. The remaining stores pass vectors through their SDK clients which handle `None` safely.

PR #4366, opened 2026-03-16
fix(openmemory): fix stats endpoint trailing slash causing 307 redirect behind HTTPS proxy

### LINKED ISSUE #4364: OpenMemory: Stats widget shows 0 when deployed behind HTTPS reverse proxy (trailing-slash 307 redirect)
## Describe the bug

The `useStats` hook in the OpenMemory UI calls `GET /api/v1/stats` (no trailing slash). Starlette's default `redirect_slashes=True` redirects this to `/api/v1/stats/` via a `307 Temporary Redirect` with an `http://` location header.

When OpenMemory is served behind an HTTPS-terminating reverse proxy (Tailscale Serve, nginx, Caddy, etc.), the browser blocks the `https://` → `http://` downgrade as a mixed-content violation. The stats request silently fails and the widget shows **0 Memories / 0 Apps**.

## To Reproduce

1. Deploy OpenMemory behind any HTTPS proxy (e.g. Tailscale Serve — a common self-hosting setup)
2. Open the dashboard — the **Memories Stats** widget shows `0 Memories` and `0 Apps`
3. `curl -v https://your-host/api/v1/stats?user_id=...` → observe `307 Temporary Redirect` → `location: http://...`
4. The browser refuses to follow the `http://` redirect from an `https://` page

## Expected behaviour

Stats widget shows correct memory and app counts regardless of proxy setup.

## Suggested fix

**Option A (UI — one line):** Add trailing slash to the stats URL in `ui/hooks/useStats.ts`:
```diff
- `${URL}/api/v1/stats?user_id=${user_id}`
+ `${URL}/api/v1/stats/?user_id=${user_id}`
```

**Option B (API):** Add a no-slash route alongside the existing one in `api/app/routers/stats.py`:
```python
@router.get("")
@router.get("/")
async def get_profile(...):
```

Option A is the minimal fix. Option B also fixes the same class of issue for any other router endpoints with trailing-slash-only routes (`/api/v1/apps`, etc.) that may be called without a slash in future.

## Environment

- OpenMemory from `mem0ai/mem0` monorepo (latest as of March 2026)
- Reverse proxy: Tailscale Serve (TLS-terminating)
- Python: 3.12, FastAPI + Starlette, uvicorn 0.41

BillionClaw:
## Issue

Fixes #4364

## Problem

The OpenMemory stats widget shows 0 when deployed behind an HTTPS reverse proxy.

The root cause is a trailing slash mismatch:
- The `useStats` hook calls `GET /api/v1/stats` (no trailing slash)
- The API endpoint is defined as `@router.get("/")` with router prefix `/api/v1/stats`, resulting in path `/api/v1/stats/`
- Starlette's default `redirect_slashes=True` redirects the request to the trailing-slash version via a 307 Temporary Redirect
- When behind an HTTPS reverse proxy, the 307 redirect uses `http://` scheme, causing the request to fail or return wrong data

## Fix

Changed `@router.get("/")` to `@router.get("")` in `openmemory/api/app/routers/stats.py`.

This makes the endpoint path exactly `/api/v1/stats`, matching what the frontend expects and eliminating the 307 redirect.

## Testing

- The endpoint now responds directly to `/api/v1/stats?user_id=xxx` without any redirect
- No 307 redirect means no scheme downgrade when behind HTTPS reverse proxy
- Stats widget will correctly fetch data in production deployments

---

Contributed via [ClawOSS](https://github.com/kevinlin/clawOSS)

### DIFF
--- a/openmemory/api/app/routers/stats.py
+++ b/openmemory/api/app/routers/stats.py
@@ -5,15 +5,13 @@ from sqlalchemy.orm import Session
 
 router = APIRouter(prefix="/api/v1/stats", tags=["stats"])
 
+
 @router.get("")
-async def get_profile(
-    user_id: str,
-    db: Session = Depends(get_db)
-):
+async def get_profile(user_id: str, db: Session = Depends(get_db)):
     user = db.query(User).filter(User.user_id == user_id).first()
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
-    
+
     # Get total number of memories
     total_memories = db.query(Memory).filter(Memory.user_id == user.id, Memory.state != MemoryState.deleted).count()
 
@@ -21,9 +19,4 @@ async def get_profile(
     apps = db.query(App).filter(App.owner == user)
     total_apps = apps.count()
 
-    return {
-        "total_memories": total_memories,
-        "total_apps": total_apps,
-        "apps": apps.all()
-    }
-
+    return {"total_memories": total_memories, "total_apps": total_apps, "apps": apps.all()}


BillionClaw:
Fixed CI — applied ruff formatting to openmemory/api/app/routers/stats.py.

kartik-mem0:
Hey, thanks for tracking down the trailing slash redirect issue and submitting the fix. We're sunsetting OpenMemory from this monorepo, so we won't be merging changes here. Appreciate the contribution\!

PR #4376, opened 2026-03-17
fix(memory): use precise field matching in _safe_deepcopy_config to preserve http_auth

NIK-TIGER-BILL:
## Problem

`_safe_deepcopy_config()` in `mem0/memory/main.py` uses broad substring matching with `"auth"` as a token to sanitize fields for telemetry:

```python
sensitive_tokens = ("auth", "credential", "password", ...)
        clone_dict[field_name] = None
```

This **unintentionally matches `http_auth`** — the field used by OpenSearch clients for AWS-signed requests (`AWSV4SignerAuth`). When `deepcopy` fails on the non-serializable auth object, the fallback path sets `http_auth = None` in the reconstructed config, breaking all OpenSearch integrations that use custom auth and causing `403 AuthorizationException` errors.

## Solution

Replace the simple substring approach with more precise suffix/prefix pattern matching that targets standard credential field naming conventions (`_auth`, `_key`, `_token`, etc.) while preserving functional fields like `http_auth` (which has `auth` as a **prefix**, not a suffix).

```python
sensitive_exact = {"connection_class"}
sensitive_suffixes = ("_auth", "_credential", "_password", "_token", "_secret", "_key", ...)
```

## Testing

- Syntax verified with Python AST parser
- The fix ensures `http_auth` is no longer matched (it ends with nothing in `sensitive_suffixes`)
- Fields like `api_key`, `auth_token`, `bearer_token`, `oauth_secret` are still correctly sanitized

Fixes #3580

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -69,9 +69,17 @@ def _safe_deepcopy_config(config):
         else:
             clone_dict = {k: v for k, v in config.__dict__.items()}
         
-        sensitive_tokens = ("auth", "credential", "password", "token", "secret", "key", "connection_class")
+        # Exact field names or suffix/prefix patterns to sanitize for telemetry.
+        # Deliberately more precise than simple substring matching to avoid
+        # clearing functional fields like `http_auth` that are needed for
+        # config reconstruction (e.g. AWS OpenSearch AWSV4SignerAuth).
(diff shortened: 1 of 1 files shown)


kartik-mem0:

Closing this PR.

PR #4378, opened 2026-03-17
feat(typescript-sdk): add config validation for OpenAI LLM and Embedder

### LINKED ISSUE #4299: How to use a custom OpenAI-compatible endpoint to replace the official OpenAI API.
### 🚀 The feature

Let's say we have an endpoint 

curl --location 'https://llm-relay.group.cn/v1/chat/completions' \
--header 'Authorization: Bearer sk-xxxxxxx' \
--header 'Content-Type: application/json' \
--data '{
    "model": "gemini-3.1-pro-preview",
    "messages": [
      {
        "role": "system",
        "content": "string"
      },
      {
        "role": "user",
        "content": "hello"
      }
    ]
}'

and it returns
{
    "id": "chatcmpl-24b8a5994d2135c87ef73b5ef60a3d10",
    "model": "gemini-3.1-pro-preview",
    "object": "chat.completion",
    "created": 1773219707,
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "Hello! How can I help you today?"
            },
            "finish_reason": "stop"
        }
    ],
    "usage": {
        "prompt_tokens": 2,
        "completion_tokens": 344,
        "total_tokens": 346,
        "prompt_tokens_details": {
            "cached_tokens": 0,
            "text_tokens": 2,
            "audio_tokens": 0,
            "image_tokens": 0
        },
        "completion_tokens_details": {
            "text_tokens": 0,
            "audio_tokens": 0,
            "reasoning_tokens": 335
        },
        "input_tokens": 0,
        "output_tokens": 0,
        "input_tokens_details": null,
        "claude_cache_creation_5_m_tokens": 0,
        "claude_cache_creation_1_h_tokens": 0
    }
}

How to use this customized custom OpenAI-compatible endpoint to REPLACE the OFFICIAL OPEN-AI key since I don't have, thanks.

### Motivation, pitch

It is convenient for us to use OpenAI-compatible endpoint instead of official OpenAI

slideshow-dingo:
## Summary
- Add config validation to catch wrong config keys (`openaiApiKey`, `baseUrl`, `url`) that should be (`apiKey`, `baseURL`)
- Prevents silent failures and 401 errors when users use Python SDK-style config keys
- Validates that `apiKey` is present and throws clear error if missing
- Warns if `openaiApiKey`, `baseUrl`, or `url` detected instead of correct keys
- Add unit tests for all validation scenarios

## Related Issue
Fixes #4299 - TypeScript SDK silently ignores wrong config keys

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/openai.ts
+++ b/mem0-ts/src/oss/src/embeddings/openai.ts
@@ -2,12 +2,29 @@ import OpenAI from "openai";
 import { Embedder } from "./base";
 import { EmbeddingConfig } from "../types";
 
+function validateOpenAIEmbedderConfig(config: EmbeddingConfig): void {
+  if (!config.apiKey) {
+    const hasWrongApiKey = "openaiApiKey" in config;
+    const errorMsg = "OpenAI Embedder config validation failed: 'apiKey' is required." +
+      (hasWrongApiKey ? " Did you mean 'apiKey' (not 'openaiApiKey')?" : "");
+    throw new Error(errorMsg);
+  }
+
+  if ("openaiApiKey" in config) {
+    console.warn("Warning: 'openaiApiKey' is not a valid config key. Use 'apiKey' instead.");
+  }
+  if ("baseUrl" in config) {
+    console.warn("Warning: 'baseUrl' is not a valid config key. Use 'baseURL' instead.");
+  }
+}
+
 export class OpenAIEmbedder implements Embedder {
   private openai: OpenAI;
   private model: string;
   private embeddingDims?: number;
 
   constructor(config: EmbeddingConfig) {
+    validateOpenAIEmbedderConfig(config);
     this.openai = new OpenAI({
       apiKey: config.apiKey,
       baseURL: config.baseURL || config.url,
--- a/mem0-ts/src/oss/src/llms/openai.ts
+++ b/mem0-ts/src/oss/src/llms/openai.ts
@@ -2,11 +2,31 @@ import OpenAI from "openai";
 import { LLM, LLMResponse } from "./base";
 import { LLMConfig, Message } from "../types";
 
+function validateOpenAIConfig(config: LLMConfig): void {
+  if (!config.apiKey) {
+    const hasWrongApiKey = "openaiApiKey" in config;
+    const errorMsg = "OpenAI LLM config validation failed: 'apiKey' is required." +
+      (hasWrongApiKey ? " Did you mean 'apiKey' (not 'openaiApiKey')?" : "");
+    throw new Error(errorMsg);
+  }
+
+  if ("openaiApiKey" in config) {
+    console.warn("Warning: 'openaiApiKey' is not a valid config key. Use 'apiKey' instead.");
+  }
+  if ("baseUrl" in config) {
+    console.warn("Warning: 'baseUrl' is not a valid config key. Use 'baseURL' instead.");
+  }
+  if ("url" in config && !config.baseURL) {
+    console.warn("Warning: 'url' is not a valid config key. Use 'baseURL' instead.");
+  }
+}
+
 export class OpenAILLM implements LLM {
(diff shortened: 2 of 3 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- The typo-detection warnings (`openaiApiKey`, `baseUrl`, `url`) address a concrete DX pain point — users porting configs from the Python SDK hit exactly these silent failures.
- `validateOpenAIConfig` / `validateOpenAIEmbedderConfig` extracted cleanly as named functions (not inlined), making the validation easy to test and reason about.
- Test coverage: 13 cases across both providers; mock pattern matches the repo's existing openai-llm and openai-embedder tests.

### Issues

#### Important (must fix)
1. `mem0-ts/src/oss/src/llms/openai.ts:4-6` and `mem0-ts/src/oss/src/embeddings/openai.ts:4-8` — The guard `if (!config.apiKey)` throws without checking `process.env.OPENAI_API_KEY`, which the OpenAI Node SDK reads automatically when `apiKey` is not passed to the constructor. Every other provider in the SDK uses the pattern `const apiKey = config.apiKey || process.env.PROVIDER_API_KEY; if (!apiKey) throw` (see `anthropic.ts:6-8`, `deepseek.ts:4-6`, `groq.ts:4-6`). Without this, any user or integration that relies on the env var (without an explicit `apiKey` in config) will start getting a thrown error that was previously handled silently by the SDK. The `defaults.ts` config itself feeds `apiKey: process.env.OPENAI_API_KEY || ""` — so when the env var is unset the empty string hits this guard now too.

   Fix: `const resolvedKey = config.apiKey || process.env.OPENAI_API_KEY; if (!resolvedKey) { throw new Error(...) }` — and pass `resolvedKey` to the OpenAI constructor so the check and the actual key agree.

2. `mem0-ts/src/oss/src/embeddings/openai.ts` — The embedder validator warns on `openaiApiKey` and `baseUrl` but does NOT warn on `url` (LLM does, embedder skips it), even though `EmbeddingConfig` in `types/index.ts:19` carries a `url` field that is silently ignored in this provider too. Minor asymmetry worth aligning.

#### Minor (optional)
- `openai-config-validation.test.ts:18` — once the env-var fallback is added, add a test for `process.env.OPENAI_API_KEY = "env-key"; new OpenAILLM({})` — should succeed (validates the path that would have broken before the fix).
- The new test file is standalone rather than additive to the existing `openai-llm.test.ts` / `openai-embedder.test.ts`. Either approach is fine; just noting for consistency.

### Evidence
- **CI:** Only Vercel preview deploy (fail — advisory, fork perms, not a required check) and CLA (pass). No required test/lint/build CI gate ran for this PR.
- **Targeted tests:** `npx jest --testPathPattern="openai-config-validation" --no-coverage` → 13 passed in 2.068s (PR head: 7f03d0c8). Tests bite — implementation revert caused suite failure.
- **Red-green proof:** N/A (this is a new-validation feature, not a bug-fix regression test). Tests verify new behavior only.
- **Linked issue:** #4299 is already CLOSED. The typo-correction warnings in this PR are still useful independent of that issue — worth opening a fresh tracking issue linking here.
- **Cross-package check:** Python SDK uses Pydantic v2 for config validation which rejects unknown keys at parse time — the analogous protection already exists there. This TS-side validation is the correct complement.
- **Could NOT verify:** Real runtime behavior with `OPENAI_API_KEY` set in env — code-tracing only. The env-var path through `defaults.ts` and `memory.ts` is clear in the source but not exercised by the PR's tests.

### Assessment
Ready to merge: **With fixes** — the typo-detection warnings are solid and the design is right; the env-var fallback is a one-liner fix that aligns with the rest of the SDK's provider pattern. Happy to re-review once that's in.

For tracking, mind opening a quick issue describing what this fixes (silent wrong-key failures in the TS SDK) and linking it here? Makes it easier to prioritize and reference.

PR #4383, opened 2026-03-17
fix: wrap vector and payload in lists in Langchain update() (#3767)

aayushbaluni:
## Summary
Fixes #3767 — Type mismatch in Langchain Vector Store Adapter's update method.

## Root Cause
The `update()` method passed single `vector` and `payload` values directly to `insert()`, which expects `List[List[float]]` for vectors and `Optional[List[Dict]]` for payloads. This caused a type mismatch and incorrect behavior when updating vectors.

## Changes
- Wrap `vector` in `[vector]` when calling `insert()` (or `[]` if None)
- Wrap `payload` in `[payload]` when calling `insert()` (or `None` if None)
- Added test `test_update_wraps_vector_and_payload_in_lists` to verify correct behavior

## Testing
- All existing langchain vector store tests pass
- New test verifies update correctly passes lists to insert

### DIFF
--- a/mem0/vector_stores/langchain.py
+++ b/mem0/vector_stores/langchain.py
@@ -115,7 +115,7 @@ class Langchain(VectorStoreBase):
         Update a vector and its payload.
         """
         self.delete(vector_id)
-        self.insert(vector, payload, [vector_id])
+        self.insert([vector] if vector is not None else [], [payload] if payload is not None else None, [vector_id])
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hi @aayushbaluni, thanks for the fix and test\! We've resolved this issue via #4446. Closing this PR — appreciate your contribution\! 🙏

PR #4391, opened 2026-03-18
fix(vercel-ai-sdk): fix chat model type and graph mode memory injection

wishhyt:
## Description

Fix two bugs in the Vercel AI SDK integration (`vercel-ai-sdk/`):

1. **Copy-paste bug in `createChatModel`** (`mem0-provider.ts`): `createChatModel` passes `modelType: "completion"` instead of `"chat"`, copied from `createCompletionModel`. This means `provider.chat(model)` incorrectly creates a completion model.

2. **Graph mode memories never injected** (`mem0-generic-language-model.ts`): When `enable_graph` is true, `getMemories()` returns `{results: [...], relations: [...]}` (an object, not an array). The check `memories?.length > 0` returns `undefined > 0 = false` on an object, so the system prompt with memories is never injected. Fixed by extracting `memories.results` before checking length.

## Type of change

## How Has This Been Tested?

- [x] Unit Test

## Checklist:

### DIFF
--- a/vercel-ai-sdk/src/mem0-generic-language-model.ts
+++ b/vercel-ai-sdk/src/mem0-generic-language-model.ts
@@ -89,13 +89,13 @@ export class Mem0GenericLanguageModel implements LanguageModelV2 {
       content: memoriesPrompt
     };
 
+    if (isGraphEnabled) {
+      memories = memories?.results;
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #4393, opened 2026-03-18
fix(bedrock): avoid sending both temperature and top_p in Converse API

norrishuang:
## Problem

Some AWS Bedrock models (e.g. **Claude Haiku 4.5** `claude-haiku-4-5-20251001-v1:0`) reject Converse API requests that specify both `temperature` and `top_p` simultaneously:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException)
when calling the Converse operation: The model returned the following errors:
`temperature` and `top_p` cannot both be specified for this model. Please use only one.
```

Currently, `AWSBedrockConfig` always sets `top_p=0.9` by default, and all Converse API call sites unconditionally include `topP` in `inferenceConfig`. This makes it impossible to use these newer models with mem0.

## Solution

**Config layer (`AWSBedrockConfig`):**
- Change `top_p` default from `0.9` to `None`
- `get_model_config()` only includes `top_p` when explicitly set by the user

**LLM layer (`AWSBedrockLLM`):**
- All 3 Converse API call paths now conditionally include `topP` only when `top_p` is present in `model_config`:
  - `_generate_with_tools()` (tool-enabled models)
  - `_generate_standard()` (Anthropic path)
  - `_generate_standard()` (Nova path)

## Backward Compatibility

- Users who **explicitly set** `top_p` in config will still have it sent to the API (no behavior change)
- Users who **don't set** `top_p` (the common case) will no longer hit the conflict error on newer models
- Legacy InvokeModel paths (non-Converse) are not affected

## Testing

Tested with Claude Haiku 4.5 (`us.anthropic.claude-haiku-4-5-20251001-v1:0`) on AWS Bedrock:
- Without this fix: ❌ `ValidationException: temperature and top_p cannot both be specified`
- With this fix: ✅ Works correctly with default config (only temperature sent)
- With explicit `top_p=0.9`: ✅ Works on models that support both params

## Changes

| File | Change |
|------|--------|
| `mem0/configs/llms/aws_bedrock.py` | `top_p` default `0.9` → `None`; filter from `get_model_config()` when None |
| `mem0/llms/aws_bedrock.py` | Conditional `topP` in all 3 Converse API `inferenceConfig` blocks |

### DIFF
--- a/mem0/configs/llms/aws_bedrock.py
+++ b/mem0/configs/llms/aws_bedrock.py
@@ -16,7 +16,7 @@ class AWSBedrockConfig(BaseLlmConfig):
         model: Optional[str] = None,
         temperature: float = 0.1,
         max_tokens: int = 2000,
-        top_p: float = 0.9,
+        top_p: Optional[float] = None,
         top_k: int = 1,
         aws_access_key_id: Optional[str] = None,
         aws_secret_access_key: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


norrishuang:
Closing this PR as the same fix has been merged in #4469 (fix(bedrock): omit topP for Anthropic Converse; use AWSBedrockConfig in LlmFactory). The core issue — not sending both temperature and top_p simultaneously — is already resolved upstream. Thank you for the review!

norrishuang:
Superseded by #4469.

PR #4394, opened 2026-03-18
fix: replace bare except with ValueError in backup router

LincolnBurrows2017:
Fixed bare except clause in backup router to catch only ValueError.


kartik-mem0:
Hi — this PR mixes openmemory changes (which are being sunset) with sweeping exception handling changes across unrelated LLM files. The CLA also hasn't been signed. Closing this — if you'd like to submit a focused fix for a specific file, feel free to do so with a signed CLA.

PR #4406, opened 2026-03-18
fix: normalize timestamp handling to UTC and preserve compatibility for existing memories

### LINKED ISSUE #4402: Bug: `created_at` and `updated_at` hardcoded to US/Pacific instead of UTC
## Bug Description

`created_at` and `updated_at` timestamps are hardcoded to `US/Pacific` timezone instead of UTC. This produces ISO strings with a `-07:00` or `-08:00` offset (depending on DST), which breaks date-range filtering and causes incorrect results for any user not in the Pacific timezone.

## Root Cause

In `mem0/memory/main.py`, both the sync `Memory` class and async `AsyncMemory` class use:

```python
# Line 1091 (sync add) and Line 2157 (async add)
metadata["created_at"] = datetime.now(pytz.timezone("US/Pacific")).isoformat()

# Line 1164 (sync update) and Line 2248 (async update)
new_metadata["updated_at"] = datetime.now(pytz.timezone("US/Pacific")).isoformat()
```

The same `US/Pacific` hardcoding appears in `mem0/vector_stores/redis.py` when converting stored Unix timestamps back to ISO strings:

```python
datetime.fromtimestamp(int(result["created_at"]), tz=pytz.timezone("US/Pacific")).isoformat(...)
```

## Why This Is a Problem

1. **Incorrect for all non-Pacific users.** A memory created at 2025-07-24 10:00 AM UTC gets stored as `2025-07-24T03:00:00-07:00`. The embedded date (`07-24`) is correct by luck here, but at UTC midnight-to-7AM the *calendar date* shifts backward — a memory created at `2025-07-25T02:00:00Z` gets stored as `2025-07-24T19:00:00-07:00`, making it appear as July 24th instead of July 25th.

2. **Date-range filters break.** When users filter by `created_at >= "2025-07-25"`, they miss memories that were created on July 25th UTC but stored as July 24th Pacific. This is the root cause of issues like #3036 and is likely related to #3213.

3. **Multi-tenant / global deployments are fundamentally broken.** Any self-hosted deployment running outside `US/Pacific` will have timestamps that look Pacific but represent a different actual time.

4. **DST ambiguity.** Pacific time shifts between `-07:00` (PDT) and `-08:00` (PST), making lexicographic comparison of ISO strings unreliable across DST boundaries.

## Proposed Fix

Replace all `US/Pacific` references with `datetime.timezone.utc` (stdlib, zero dependencies):

```python
# mem0/memory/main.py — add and update methods (both sync and async)

metadata["created_at"] = datetime.now(timezone.utc).isoformat()
new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
```

```python
# mem0/vector_stores/redis.py — when converting stored timestamps back

datetime.fromtimestamp(int(result["created_at"]), tz=timezone.utc).isoformat(...)
```

This produces clean `+00:00` suffixed ISO strings like `2025-07-24T10:00:00+00:00` that:
- Sort correctly as strings (lexicographic order = chronological order)
- Work with date-range filters regardless of user timezone
- Are unambiguous (no DST shifts)

## Affected Files

| File | Issue |
|---|---|
| `mem0/memory/main.py` L1091, L1164, L2157, L2248 | `datetime.now(pytz.timezone("US/Pacific"))` |
| `mem0/vector_stores/redis.py` (multiple locations) | `datetime.fromtimestamp(..., tz=pytz.timezone("US/Pacific"))` |
| `mem0/graphs/neptune/neptunedb.py` | Same `US/Pacific` pattern |

## Migration Note

Existing memories stored with `-07:00`/`-08:00` offsets are technically correct (the offset is preserved), but any code doing string comparison on `created_at` will need to parse the offset. A migration to normalize existing values to UTC would be ideal but isn't strictly required if consumers parse the ISO string properly.

## Related Issues

- #3036 — Date filtering not working correctly (still open, likely caused by this)
- #3213 — `created_at` timestamp filtering doesn't work (closed, but root cause unaddressed)

Mukhsin0508:
## Description

This PR fixes hardcoded `US/Pacific` timestamps for memory metadata and related read paths by switching affected code to UTC.

Before this change, new `created_at` / `updated_at` values could be written with `-07:00` / `-08:00` offsets, which caused incorrect date boundaries and broken filtering behavior for users outside Pacific time. This was especially problematic near UTC midnight.

This PR updates:

- `mem0/memory/main.py`
- `mem0/vector_stores/redis.py`
- `mem0/graphs/neptune/neptunedb.py`

## What Changed

- Replaced hardcoded `US/Pacific` writes with `datetime.now(timezone.utc).isoformat()`
- Replaced Redis timestamp rendering from `US/Pacific` to `timezone.utc`
- Updated Neptune graph payload creation to use UTC timestamps
- Added compatibility normalization for existing timezone-aware ISO timestamps in memory read/update paths

## Production Safety / Existing Data

This PR is intentionally backwards-safe for existing memories:

- Existing timezone-aware ISO timestamps are not reinterpreted as a different moment in time
- When existing memories are updated or returned through memory read/search/list paths, timezone-aware timestamps are normalized to UTC using their stored offset
- Naive timestamps are left unchanged to avoid making unsafe assumptions about historical data

That means old prod memories should not be corrupted by this change.

Important nuance:

- For Redis-backed data, stored values are already numeric timestamps underneath, so this mostly fixes incorrect timezone presentation
- For existing string-based stores, this PR improves compatibility, but a separate backfill would still be the safest way to fully normalize all historical records to UTC

## Why This Approach

This PR avoids a risky in-place migration of old timestamps while still fixing new writes immediately.

It gives us:

- Correct UTC timestamps for all new writes
- Correct UTC rendering for affected read paths
- Safe normalization of old offset-aware values without changing their meaning

## Follow-Up Recommendation

For a complete production migration story, a separate follow-up can add:

- canonical numeric timestamp fields for filtering
- a batched backfill for historical string timestamps
- a final cutover so filtering no longer depends on mixed ISO string formats

## Type of change

## How Has This Been Tested?

Tested locally in a `.venv` with:

```bash
.venv/bin/python -m pytest tests/memory/test_main.py tests/memory/test_neptune_memory.py tests/vector_stores/test_redis.py
```

Added regression coverage for:

- sync memory UTC timestamp creation/update
- async memory UTC timestamp creation/update
- normalization of old offset-aware timestamps to UTC
- Neptune payload timestamps
- Redis search/get/list UTC rendering

## Checklist

- [x] I have added tests that prove the fix is effective
- [x] New and existing relevant tests pass locally with my changes

## Maintainer Checklist

- [ ] closes #4402
- [ ] Made sure Checks passed

### DIFF
--- a/mem0/graphs/neptune/neptunedb.py
+++ b/mem0/graphs/neptune/neptunedb.py
@@ -1,7 +1,6 @@
 import logging
 import uuid
-from datetime import datetime
-import pytz
+from datetime import datetime, timezone
 
 from .base import NeptuneBase
 
@@ -114,7 +113,7 @@ class MemoryGraph(NeptuneBase):
             "name": destination,
             "type": destination_type,
             "user_id": user_id,
-            "created_at": datetime.now(pytz.timezone("US/Pacific")).isoformat(),
+            "created_at": datetime.now(timezone.utc).isoformat(),
         }
         self.vector_store.insert(
             vectors=[dest_embedding],
@@ -189,7 +188,7 @@ class MemoryGraph(NeptuneBase):
             "name": source,
             "type": source_type,
             "user_id": user_id,
-            "created_at": datetime.now(pytz.timezone("US/Pacific")).isoformat(),
+            "created_at": datetime.now(timezone.utc).isoformat(),
         }
         self.vector_store.insert(
             vectors=[source_embedding],
@@ -316,14 +315,14 @@ class MemoryGraph(NeptuneBase):
             "name": source,
             "type": source_type,
             "user_id": user_id,
-            "created_at": datetime.now(pytz.timezone("US/Pacific")).isoformat(),
+            "created_at": datetime.now(timezone.utc).isoformat(),
         }
         destination_id = str(uuid.uuid4())
         destination_payload = {
             "name": destination,
             "type": destination_type,
             "user_id": user_id,
-            "created_at": datetime.now(pytz.timezone("US/Pacific")).isoformat(),
+            "created_at": datetime.now(timezone.utc).isoformat(),
         }
         self.vector_store.insert(
             vectors=[source_embedding, dest_embedding],
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -8,10 +8,9 @@ import os
 import uuid
 import warnings
 from copy import deepcopy
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any, Dict, Optional
 
-import pytz
 from pydantic import ValidationError
 
 from mem0.configs.base import MemoryConfig, MemoryItem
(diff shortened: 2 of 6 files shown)


kartik-mem0:
have to close this issue #4404 this is a better solution.

i wanted to merge this solution but it didn't had cla signed and there was merge conflict.

hope you understand.

PR #4407, opened 2026-03-18
fix(ts-sdk): externalize all peerDependencies in tsup config

### LINKED ISSUE #3857: [TypeScript SDK] Static require('ollama') breaks apps not using Ollama provider

kartik-mem0:
## Description

The TypeScript SDK (`mem0ai/oss`) crashes at load time with `Cannot find module 'ollama'` when users don't have the `ollama` package installed, even when using a completely different provider (e.g., Gemini, OpenAI). This happens because `ollama` and 9 other peerDependencies were missing from `tsup.config.ts`'s `external` array, causing tsup to bundle them inline instead of treating them as external requires.

This PR adds all missing peerDependencies to the `external` array and includes a drift-prevention test that will fail if a new peerDep is added to `package.json` without being externalized in `tsup.config.ts`.

Fixes #3857

## Type of change

## How Has This Been Tested?

- Verified `npm run build` passes and bundle compiles cleanly
- Inspected `dist/oss/index.js` to confirm `require("ollama")` is a simple external require (2 lines), not the entire inlined package
- Confirmed all other newly externalized deps (`@google/genai`, `@mistralai/mistralai`, `@supabase/supabase-js`, etc.) are also externalized correctly
- Drift-prevention test validates every runtime peerDependency is present in the external array
- Reverse staleness test validates no orphaned entries exist in the external array

- [x] Unit Test

## Checklist:

## Maintainer Checklist

- [x] closes #3857
- [ ] Made sure Checks passed

### DIFF
--- a/mem0-ts/src/oss/tests/tsup-externals.test.ts
+++ b/mem0-ts/src/oss/tests/tsup-externals.test.ts
@@ -1,0 +1,51 @@
+import * as fs from "fs";
+import * as path from "path";
+
+/**
+ * Drift-prevention test: ensures every peerDependency in package.json
+ * is listed in tsup.config.ts's external array so tsup never bundles
+ * optional provider SDKs into the dist output.
+ */
+describe("tsup.config.ts externals", () => {
+  let peerDeps: string[];
+  let directDeps: string[];
+  let externalDeps: string[];
+
+  beforeAll(() => {
+    const pkgPath = path.resolve(__dirname, "../../../package.json");
+    const pkg = JSON.parse(fs.readFileSync(pkgPath, "utf-8"));
+    // Filter out @types/* packages — they are type-only and not bundled at runtime
+    peerDeps = Object.keys(pkg.peerDependencies || {}).filter(
+      (dep) => !dep.startsWith("@types/"),
+    );
+    directDeps = Object.keys(pkg.dependencies || {});
+
+    const tsupConfigPath = path.resolve(__dirname, "../../../tsup.config.ts");
+    const tsupContent = fs.readFileSync(tsupConfigPath, "utf-8");
+
+    // Extract strings from the external array (supports double, single, and backtick quotes)
+    const externalMatch = tsupContent.match(
+      /const external\s*=\s*\[([\s\S]*?)\];/,
+    );
+    if (!externalMatch) {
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Recreating from main repo branch

Issue #4401, opened 2026-03-18
`removeCodeBlocks` fails to strip markdown fences from truncated LLM responses


bsmoreno2910:
### Bug Description

`removeCodeBlocks` uses a regex that requires **both** opening and closing backtick fences to match:

```js
return text.replace(/```(?:\w+)?\n?([\s\S]*?)```/g, "$1").trim();
```

When the LLM returns a very long JSON response wrapped in ` ```json ``` ` that gets **truncated** (no closing fence arrives), the regex does not match. The raw text — starting with ` ```json ` — is passed directly to `JSON.parse`, causing a crash.

---

### Error

```
Failed to parse memory actions from LLM response: ```json
{
  "memory": [
    ...
    {
      "id": SyntaxError: Unexpected token '`', "```json
{
"... is not valid JSON
    at JSON.parse (<anonymous>)
    at _Memory.addToVectorStore (mem0ai/src/oss/src/memory/index.ts:370:28)
    at _Memory.add (mem0ai/src/oss/src/memory/index.ts:243:31)
    at OSSProvider.add (index.ts:263:20)
```

---

### Steps to Reproduce

1. Use `mem0ai` with a LLM that wraps JSON responses in markdown code blocks
2. Trigger a memory `add` operation where the LLM response is **long enough to be truncated** before the closing ` ``` ` (e.g., agent replies containing large code blocks or architecture documents in the same conversation)
3. Observe `Failed to parse memory actions from LLM response` in logs

---

### Root Cause

The regex `/```(?:\w+)?\n?([\s\S]*?)```/g` is non-greedy and requires a closing ` ``` ` to capture the inner content. A truncated response only has the opening fence — the regex silently skips it and returns the original text unchanged, which still starts with ` ```json `.

---

### Suggested Fix

```js
function removeCodeBlocks(text) {
  // Handle complete code blocks (opening + closing fence)
  let result = text.replace(/```(?:\w+)?\n?([\s\S]*?)```/g, "$1").trim();
  // Handle truncated responses where closing ``` never arrives
  result = result.replace(/^```(?:\w+)?\n?/, "").replace(/\n?```$/, "").trim();
  return result;
}
```

This is a two-pass approach:
- First pass: strips complete ` ```lang ... ``` ` blocks (existing behavior)
- Second pass: strips any remaining leading/trailing fence markers (handles truncation)

---

### Environment

| Field | Value |
|---|---|
| mem0ai version | 0.3.3 (npm) |
[...]

PR #4414, opened 2026-03-19
fix: handle truncated markdown code blocks in LLM responses

haroldfabla2-hue:
## Summary

Fixes issue #4401 - `removeCodeBlocks` fails to strip markdown fences from truncated LLM responses.

## Problem

When the LLM returns a very long JSON response wrapped in ```json ``` that gets truncated (no closing fence arrives), the regex does not match. The raw text — starting with ```json — is passed directly to JSON.parse, causing a crash.

## Solution

Updated `removeCodeBlocks` function to use a two-pass approach:
1. First pass: strips complete ```lang ... ``` blocks (existing behavior)
2. Second pass: strips any remaining leading/trailing fence markers (handles truncation)

## Changes

- Updated `mem0-ts/src/oss/src/prompts/index.ts`

## Testing

The fix is backwards-compatible and does not affect responses that are already clean JSON.

## Related

Fixes: mem0ai/mem0#4401

### DIFF
--- a/mem0-ts/src/oss/src/prompts/index.ts
+++ b/mem0-ts/src/oss/src/prompts/index.ts
@@ -278,5 +278,9 @@ export function parseMessages(messages: string[]): string {
 }
 
 export function removeCodeBlocks(text: string): string {
-  return text.replace(/```[^`]*```/g, "");
+  // Handle complete code blocks (opening + closing fence)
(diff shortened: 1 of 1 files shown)


kartik-mem0:
@Jah-yee thank you for your contribution.

closing this pr in favour of #4421

PR #4416, opened 2026-03-19
fix(opensearch): preserve http_auth in _safe_deepcopy_config

NIK-TIGER-BILL:
Fixes #3580

## Problem

`_safe_deepcopy_config()` has a fallback path that is triggered when `deepcopy` fails (e.g. when `http_auth` contains a non-picklable AWS `RequestSigner` object). In this fallback path the function nullifies all fields whose names match sensitive tokens — including **`http_auth`**, **`aws_auth`**, and **`connection_class`** — via the `"auth"` substring match.

This silently drops the AWS authentication credentials, causing subsequent OpenSearch requests to fail with `403 Forbidden`.

## Root cause

```python
# Before – "auth" matches http_auth and aws_auth → set to None
sensitive_tokens = ("auth", "credential", ...)
        clone_dict[field_name] = None  # ← breaks http_auth!
```

## Fix

1. Introduce an explicit `AUTH_FIELDS` set (`http_auth`, `aws_auth`, `connection_class`).
2. Save their values **before** deepcopy / fallback serialisation.
3. Skip nullification for fields in `AUTH_FIELDS`.
4. Restore saved values **after** the copy so they are always present in the returned config.

```python
AUTH_FIELDS = {"http_auth", "aws_auth", "connection_class"}
saved_auth = {k: getattr(config, k) for k in AUTH_FIELDS if getattr(config, k, None) is not None}
# ... copy logic ...
clone_dict.update(saved_auth)  # restore
```

## Testing

Reproducible with any AWS OpenSearch setup that passes an `AWSV4SignerAuth` (or similar non-picklable object) as `http_auth`. After this fix the auth object survives the config copy and requests succeed.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -52,14 +52,37 @@ logger = logging.getLogger(__name__)
 
 
 def _safe_deepcopy_config(config):
-    """Safely deepcopy config, falling back to JSON serialization for non-serializable objects."""
+    """Safely deepcopy config, falling back to JSON serialization for non-serializable objects.
+
+    Auth-related fields (e.g. http_auth, aws_auth, connection_class) are preserved even when
+    they contain non-picklable objects such as AWS RequestSigner instances. Without this
+    preservation the fallback path would silently drop these fields and break authentication
+    (see https://github.com/mem0ai/mem0/issues/3580).
+    """
+    # Fields that may hold non-picklable auth objects – saved and restored explicitly
+    # so they survive both a failed deepcopy and the JSON-serialisation fallback path.
+    AUTH_FIELDS = {"http_auth", "aws_auth", "connection_class"}
+
+    # Save original auth field values (if present on the config object)
+    if hasattr(config, "__dict__"):
+        saved_auth = {k: getattr(config, k) for k in AUTH_FIELDS if getattr(config, k, None) is not None}
+    else:
+        saved_auth = {}
+
     try:
-        return deepcopy(config)
+        result = deepcopy(config)
+        # Restore auth fields in case deepcopy dropped or mangled them
+        for k, v in saved_auth.items():
+            try:
+                object.__setattr__(result, k, v)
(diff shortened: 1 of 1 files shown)


xkonjin (commented):
**Code Review**

This is a well-thought-out fix for a real authentication bug. The approach of saving and restoring auth fields across deepcopy/JSON fallback paths is sound.

**Potential issues:**

1. **Shared mutable state risk:** By restoring the *original* auth objects (not copies) into the cloned config, both the original and cloned config now share the same `http_auth`/`aws_auth` references. If anything mutates these objects later, it affects both configs. This may be intentional (auth objects are typically stateless credentials), but worth documenting in a comment.

2. **`object.__setattr__` in the deepcopy success path (line ~73):** If deepcopy *succeeds*, the auth fields should already be correctly copied. Overwriting them with originals seems defensive but could mask a real deepcopy bug. Consider only restoring in the fallback path, or at minimum adding a comment explaining why.

3. **Removing `"auth"` from `sensitive_tokens` (line ~96):** This is a broader change than just preserving `http_auth`. Any future field containing "auth" in its name will no longer be nullified during the JSON fallback. If there are fields like `oauth_token_auth` this could leak sensitive data. The `AUTH_FIELDS` allowlist + skip is the right approach, but I'd keep `"auth"` in `sensitive_tokens` and let the `AUTH_FIELDS` check above handle the exceptions.

**Test coverage:** No tests included. This is a critical auth path. Would strongly recommend adding tests for:
- Config with `http_auth` survives deepcopy
- Config with `http_auth` survives JSON fallback (mock deepcopy failure)
- Fields like `secret_auth_key` are still nullified

kartik-mem0:

Closing this PR.

utkarsh240799:
Hi @NIK-TIGER-BILL, thanks for working on this\! The issue has been addressed in #4418, so we'll be closing this PR. We appreciate the contribution\!

PR #4423, opened 2026-03-19
fix: prevent thread and memory leaks from PostHog telemetry and unclosed resources

### LINKED ISSUE #3376: Memory leak and thread leak issues in mem0 library

tao-to-python:
## Description

Fixes thread leak and memory leak issues reported in #3376.

### Problem

Each call to `capture_event()` creates a new `AnonymousTelemetry` instance (which spawns a new PostHog client with background threads) but **never calls `close()`/`shutdown()`**. In long-running applications or request handlers, this causes:

- **Thread leak**: Unbounded growth of PostHog background threads
- **Memory leak**: Each `Memory.from_config()` call increases memory usage that never gets released
- **Connection leak**: SQLite, vector store, and LLM HTTP client connections are never cleaned up

### Solution

1. **`telemetry.py`**: Call `oss_telemetry.close()` after capturing events to shut down PostHog threads immediately

2. **`Memory.close()`**: New method that cleans up all held resources:
   - PostHog telemetry threads
   - SQLite database connections
   - Vector store client connections
   - LLM HTTP client connections

3. **Context manager support**: Memory now works with `with` statement:
   ```python
   with Memory.from_config(config) as m:
       m.add(messages=[...], user_id="user1")
   # All resources automatically cleaned up
   ```

4. **`__del__` fallback**: Resources are cleaned up during garbage collection as a safety net

### Changes

| File | Change |
|------|--------|
| `mem0/memory/telemetry.py` | Call `close()` after `capture_event` to prevent thread leak |
| `mem0/memory/main.py` | Add `close()`, `__enter__`, `__exit__`, `__del__` methods |
| `tests/test_resource_cleanup.py` | Test suite for cleanup and context manager |

### Testing

Before this fix (reproducing #3376):
```python
import threading
for i in range(10):
    m.add(messages=[{"role": "user", "content": "test"}], user_id="test")
print(threading.active_count())  # Keeps growing!
```

After this fix:
```python
for i in range(10):
    with Memory.from_config(config) as m:
        m.add(messages=[{"role": "user", "content": "test"}], user_id="test")
print(threading.active_count())  # Stays stable
```

Fixes #3376

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1287,6 +1287,66 @@ class Memory(MemoryBase):
         )
         return memory_id
 
+    def close(self):
+        """
+        Clean up resources held by this Memory instance.
+
+        Shuts down the telemetry client (which spawns background threads),
+        closes the SQLite database connection, and releases the HTTP client
+        used by the vector store (if applicable).
+
+        Use this method when you create Memory instances in a loop or in
+        request handlers to prevent thread and connection leaks.
+
+        Can also be used as a context manager::
+
+            with Memory.from_config(config) as m:
+                m.add(...)
+        """
+        # Shut down telemetry to stop PostHog background threads
+        if hasattr(self, '_telemetry_vector_store'):
+            try:
+                from mem0.memory.telemetry import client_telemetry
+                client_telemetry.close()
+            except Exception:
+                pass
+
+        # Close SQLite connection
+        if hasattr(self, 'db') and hasattr(self.db, 'connection') and self.db.connection:
+            try:
+                self.db.connection.close()
+            except Exception:
+                pass
+
+        # Close vector store client if it has a close method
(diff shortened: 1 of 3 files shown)


kartik-mem0:
Hi @tao-to-python, thank you so much for taking the time to work on this fix! 🙏

This issue has now been resolved via PR #4535, which was merged with a thread-safe lazy singleton for the PostHog client, `atexit` cleanup, and `Memory.close()` / context manager support.

Since the fix is already in, we're closing this PR. We really appreciate your contribution and effort — it helped inform the final solution. Hope to see you around on future issues!

PR #4424, opened 2026-03-19
fix: wrap vector and payload in lists in Langchain adapter update()

tao-to-python:
## Description

Fixes type mismatch in the Langchain vector store adapter's `update()` method as reported in #3767.

### Problem

The `update()` method passes unwrapped `vector` and `payload` values to `insert()`:

```python
# Before (broken)
    self.delete(vector_id)
    self.insert(vector, payload, [vector_id])  # vector and payload not wrapped
```

But `insert()` expects:
- `vectors: List[List[float]]` — a list of vectors
- `payloads: Optional[List[Dict]]` — a list of payloads

This causes runtime errors when updating vectors through the Langchain adapter.

### Fix

```python
# After (fixed)
    self.delete(vector_id)
    self.insert([vector], [payload], [vector_id])  # properly wrapped in lists
```

### Changes

| File | Change |
|------|--------|
| `mem0/vector_stores/langchain.py` | Wrap `vector` and `payload` in lists in `update()` |
| `tests/vector_stores/test_langchain_vector_store.py` | Add regression test |

Fixes #3767

### DIFF
--- a/mem0/vector_stores/langchain.py
+++ b/mem0/vector_stores/langchain.py
@@ -115,7 +115,7 @@ class Langchain(VectorStoreBase):
         Update a vector and its payload.
         """
         self.delete(vector_id)
-        self.insert(vector, payload, [vector_id])
+        self.insert([vector], [payload], [vector_id])
 
     def get(self, vector_id):
         """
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hi @tao-to-python, thanks for the well-written fix and test\! We've addressed this issue via #4446. Closing this PR — appreciate your contribution\! 🙏

Issue #3410, opened 2025-09-03
Invalid JSON response when using memory.add with gemini-2.5-flash


federicoschiaffino:
### 🐛 Describe the bug

When I run the following command: 
`memory.add(messages, user_id="rick")`

I sometimes get this error: 
`Invalid JSON response: Expecting value: line 1 column 1 (char 0)`

When this happen the memory is not updated nor added. Anyone facing the same issue?
I'm using gemini-2.5-flash

parshvadaftari:
@federicoschiaffino Can you share the version, config and the example message so that I can recreate this issue. Thanks!

federicoschiaffino:
@parshvadaftari 
SDK version is 1.33.0
```
config = {
    "embedder": {
        "provider": "gemini",
        "config": {
            "model": "models/text-embedding-004",
        }
    },
        "llm": {
        "provider": "gemini",
        "config": {
            "model": "gemini-2.5-flash",
            "temperature": 0.0,
            "max_tokens": 2000,
        }
    },
    "vector_store": {
        "provider": "pgvector",
        "config": {
            "dbname": "memory_db",
            "host": "localhost",
            "port": "5432",
            "embedding_model_dims": 768,
        }    
    }
}
```

I get 2 types of errors: 
with `messages = [{'role': 'user', 'content': 'my brother Will has another daughter called Paula!'}, {'role': 'assistant', 'content': "Thank you for letting me know! So, your brother Will has two daughters, Lia and Paula. I've updated my memory."}]`
I got: `Invalid JSON response: Expecting value: line 1 column 1 (char 0)`

 
And with `messages = [{'role': 'user', 'content': 'my brother Will just had a daughter named Lia'}, {'role': 'assistant', 'content': "That's wonderful news! Congratulations to your brother Will and his family on the arrival of Lia!"}, {'role': 'user', 'content': "Will's wife is Emma from Sweden "}, {'role': 'assistant', 'content': "Thank you for sharing that! It's nice to know a bit more about Will's family. So, Lia's mother is Emma, who is from Sweden."}]`
I got this error: `Invalid JSON response: Expecting ',' delimiter: line 11 column 28 (char 259)`

Thank you!

RO-HIT17:
[...]

PR #4425, opened 2026-03-19
fix: handle empty/None LLM responses gracefully in fact extraction

tao-to-python:
## Description

Fixes crashes when LLM providers (particularly Gemini) return empty or None responses during memory fact extraction.

### Problem

Multiple users reported intermittent errors with Gemini (#3410, 19 comments):

```
Invalid JSON response: Expecting value: line 1 column 1 (char 0)
```

and:

```
AttributeError: 'NoneType' object has no attribute 'strip'
```

**Root cause**: `GeminiLLM._parse_response()` returns `""` (empty string) when Gemini produces no content parts. But when the response is truly empty or `None`, calling `remove_code_blocks()` and `.strip()` on it causes errors.

### Solution

1. **`gemini.py`**: Return `None` instead of `""` when no content parts are available — this correctly signals "no response" rather than "empty text response"

2. **`main.py` (sync + async)**: Add early `None` check before processing the response:
   ```python
   if not response:
       logger.warning("Empty response from LLM during fact extraction, skipping.")
       new_retrieved_facts = []
   else:
       response = remove_code_blocks(response)
       ...
   ```

### Changes

| File | Change |
|------|--------|
| `mem0/llms/gemini.py` | Return `None` instead of `""` for empty responses |
| `mem0/memory/main.py` | Add `None` guard in sync and async `_add_to_vector_store` |

Relates to #3410

### DIFF
--- a/mem0/llms/gemini.py
+++ b/mem0/llms/gemini.py
@@ -63,7 +63,7 @@ class GeminiLLM(LLMBase):
                 for part in response.candidates[0].content.parts:
                     if hasattr(part, "text") and part.text:
                         return part.text
-            return ""
+            return None
 
     def _reformat_messages(self, messages: List[Dict[str, str]]):
         """
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -515,18 +515,22 @@ class Memory(MemoryBase):
         )
 
         try:
-            response = remove_code_blocks(response)
-            if not response.strip():
+            if not response:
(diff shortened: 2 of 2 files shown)


youneshima:
Closing as a duplicate of #4602 after reviewing the full PR timeline, diff, checks, and comments. #4602 covers empty/invalid fact-extraction responses more generally and includes tests and passing Python CI.

kartik-mem0:
closing this pr in the favour of #4462

PR #4437, opened 2026-03-20
fix: handle missing IDs in temp_uuid_mapping to prevent KeyError on LLM hallucinated IDs

### LINKED ISSUE #3931: KeyError in temp_uuid_mapping during async UPDATE operation
We encountered a `KeyError` during an async memory UPDATE operation that resulted in a cryptic error message in our logs.

## Error Context

```
Error processing memory action (async): {'id': '12', 'text': 'User has experienced corner chipping on the inner insert', 'event': 'UPDATE', 'old_memory': 'Experiencing chipping on the inner insert'}, Error: '12'
```

## Root Cause

The error appears to originate from `mem0/memory/main.py` around line 1612, where the code accesses `temp_uuid_mapping[resp["id"]]` without checking if the ID exists in the dictionary first.

When the KeyError is caught and logged with `logger.error(f"Error: {e}")`, Python's string representation of `KeyError('12')` produces just `'12'`, making the error message quite cryptic.

## Expected Behavior

Either:
1. The `temp_uuid_mapping` should always contain the IDs returned in responses, OR
2. The code should handle the case where an ID is missing with a more informative error message

## Actual Behavior

A `KeyError` is raised when the ID doesn't exist in the mapping, with minimal context in the error message.

## Occurrence

- Single occurrence in production (2026-01-22)
- No recurrence observed
- Error was handled gracefully by our error handling
- No user impact (memory extraction happens asynchronously after user receives response)

## Questions

1. Under what conditions might an ID in the response not exist in `temp_uuid_mapping`?
2. Is this expected behavior that we should handle, or is it a bug?
3. Should we be doing something differently in our usage to prevent this?

## Environment

- Using mem0's async memory operations with OpenAI embeddings
- UPDATE operation on existing memory

tao-to-python:
## Description

Fixes recurring production `KeyError` in `temp_uuid_mapping` during UPDATE/DELETE operations, as reported in #3931.

### Problem

When the LLM returns memory actions with IDs that don't exist in `temp_uuid_mapping` (hallucinated IDs), direct dict access `temp_uuid_mapping[resp["id"]]` raises `KeyError`. This has been **recurring in production since January 2026** across multiple users.

The error message is also cryptic — `KeyError('12')` is logged as just `'12'`, making debugging difficult.

### Production occurrences (from issue comments)

```
Error processing memory action (async): {'id': '12', 'text': '...'}, Error: '12'
Error processing memory action (async): {'id': '9',  'text': '...'}, Error: '9'
Error processing memory action (async): {'id': '7',  'text': '...'}, Error: '7'
Error processing memory action (async): {'id': '16', 'text': '...'}, Error: '16'
```

### Solution

1. **Use `dict.get()` instead of `[]`** for UPDATE and DELETE branches in both sync and async `_add_to_vector_store()`
2. **Skip with warning** when ID is not found, instead of crashing
3. **Include exception type** in error logs: `Error: '12'` → `KeyError: '12'`

### Changes

| File | Change |
|------|--------|
| `mem0/memory/main.py` | Safe `.get()` lookup + skip for sync UPDATE/DELETE |
| `mem0/memory/main.py` | Safe `.get()` lookup + skip for async UPDATE/DELETE |
| `mem0/memory/main.py` | Better error logging with `type(e).__name__` |
| `tests/test_temp_uuid_mapping.py` | Tests for sync/async × UPDATE/DELETE with missing IDs |

Fixes #3931

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -615,25 +615,39 @@ class Memory(MemoryBase):
                         )
                         returned_memories.append({"id": memory_id, "memory": action_text, "event": event_type})
                     elif event_type == "UPDATE":
+                        memory_id = temp_uuid_mapping.get(resp.get("id"))
+                        if not memory_id:
+                            logger.warning(
+                                f"Skipping UPDATE: memory ID '{resp.get('id')}' not found in mapping "
+                                "(possible LLM hallucination)"
+                            )
+                            continue
                         self._update_memory(
-                            memory_id=temp_uuid_mapping[resp.get("id")],
+                            memory_id=memory_id,
                             data=action_text,
                             existing_embeddings=new_message_embeddings,
                             metadata=deepcopy(metadata),
                         )
                         returned_memories.append(
                             {
-                                "id": temp_uuid_mapping[resp.get("id")],
+                                "id": memory_id,
                                 "memory": action_text,
                                 "event": event_type,
                                 "previous_memory": resp.get("old_memory"),
                             }
                         )
                     elif event_type == "DELETE":
-                        self._delete_memory(memory_id=temp_uuid_mapping[resp.get("id")])
+                        memory_id = temp_uuid_mapping.get(resp.get("id"))
+                        if not memory_id:
+                            logger.warning(
+                                f"Skipping DELETE: memory ID '{resp.get('id')}' not found in mapping "
+                                "(possible LLM hallucination)"
+                            )
+                            continue
+                        self._delete_memory(memory_id=memory_id)
                         returned_memories.append(
                             {
-                                "id": temp_uuid_mapping[resp.get("id")],
+                                "id": memory_id,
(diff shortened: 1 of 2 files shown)


utkarsh240799:
Hi @tao-to-python ,

Thanks for taking the time to come up with the fix. Can you please resolve the merge conflict and sign the CLA?

Thank you.

kartik-mem0:
Hi @tao-to-python — thanks for the detailed write-up and the thorough fix! The production error examples you included in the PR description were especially helpful in understanding the scope of the problem.

This issue has been resolved via #4674, which was merged into `main`. Since both PRs target the same `KeyError` in `temp_uuid_mapping` during UPDATE/DELETE operations, we're closing this one to keep things tidy.

Really appreciate you stepping up on this — looking forward to your future contributions! 🙏

PR #4438, opened 2026-03-20
fix: extract text from dict in update_memory to prevent TypeError in embed()

### LINKED ISSUE #3235: TypeError in update_memory due to incorrect data type passed to embed()

tao-to-python:
## Description

Fixes `TypeError` when `update_memory` API endpoint receives a dict payload, as reported in #3235.

### Problem

`PUT /memories/{memory_id}` receives `Dict[str, Any]` (e.g. `{"data": "new text"}`), which is passed directly to `Memory.update(data=...)`. The `update()` method calls `self.embedding_model.embed(data)`, which expects a string. When `data` is a dict, `embed()` internally calls `data.replace()`, raising:

```
TypeError: 'dict' object has no attribute 'replace'
```

### Solution

**Defense in depth** — fix at both the API layer and the core layer:

1. **`server/main.py`**: Extract `data` field from dict before passing to `Memory.update()`
2. **`mem0/memory/main.py`** (sync + async `update()`): Add type normalization — if `data` is a dict, extract `.get("data")`; if still not a string, convert with `str()`
3. Fixed docstring to match actual type annotation (`Dict[str, Any]` not `str`)

### Changes

| File | Change |
|------|--------|
| `server/main.py` | Extract text from dict, fix docstring |
| `mem0/memory/main.py` | Type guard in sync `update()` |
| `mem0/memory/main.py` | Type guard in async `update()` |

Fixes #3235

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1085,6 +1085,11 @@ class Memory(MemoryBase):
         """
         capture_event("mem0.update", self, {"memory_id": memory_id, "sync_type": "sync"})
 
+        if isinstance(data, dict):
+            data = data.get("data") or str(data)
+        if not isinstance(data, str):
+            data = str(data)
+
         existing_embeddings = {data: self.embedding_model.embed(data, "update")}
 
         self._update_memory(memory_id, data, existing_embeddings)
@@ -2148,6 +2153,11 @@ class AsyncMemory(MemoryBase):
             {'message': 'Memory updated successfully!'}
         """
         capture_event("mem0.update", self, {"memory_id": memory_id, "sync_type": "async"})
+
+        if isinstance(data, dict):
+            data = data.get("data") or str(data)
+        if not isinstance(data, str):
+            data = str(data)
 
         embeddings = await asyncio.to_thread(self.embedding_model.embed, data, "update")
         existing_embeddings = {data: embeddings}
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hi @tao-to-python — thanks for working on this fix!

This has been addressed in #4595 (merged Apr 2), which takes a similar approach with a dedicated `MemoryUpdate` model to properly extract the text before embedding.

Closing as superseded. Thank you for your contribution to mem0!

PR #3893, opened 2026-01-10
fix: add auto_refresh option for OpenSearch Serverless compatibility

Malhis:
## Description

OpenSearch Serverless does not support the `indices.refresh()` API, which was added in mem0 v1.0.0. This causes a `NotFoundError(404, 'resource_not_found_exception')` when attempting to insert memories using OpenSearch Serverless as the vector store.

This PR adds an `auto_refresh` configuration option (default: `False`) that makes the refresh behavior opt-in. OpenSearch automatically refreshes indices approximately every 1 second by default, so explicit refresh calls are rarely needed.

**Changes:**
- Added `auto_refresh` config field to `OpenSearchConfig`
- Made `indices.refresh()` call conditional on `auto_refresh` setting
- Added unit tests to verify behavior with auto_refresh enabled/disabled

Fixes #3739

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] Unit Test

All 14 OpenSearch unit tests pass:
```
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_auto_refresh_disabled_by_default PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_auto_refresh_enabled PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_create_index PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_delete PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_delete_col PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_get PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_init_with_http_auth PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_insert PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_list_cols PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_search PASSED
tests/vector_stores/test_opensearch.py::TestOpenSearchDB::test_update PASSED
tests/vector_stores/test_opensearch.py::test_safe_deepcopy_config_handles_opensearch_auth PASSED
tests/vector_stores/test_opensearch.py::test_safe_deepcopy_config_normal_configs PASSED
tests/vector_stores/test_opensearch.py::test_memory_initialization_opensearch_aws_auth PASSED
```

Test commands: 
[...]

PR #4452, opened 2026-03-20
fix(opensearch): gracefully handle missing indices.refresh() for Serverless compatibility

### LINKED ISSUE #3739: OpenSearch Serverless Incompatibility in v1.0.0
## Issue Summary

mem0 v1.0.0 is incompatible with AWS OpenSearch Serverless due to use of the [`indices.refresh()` API call](https://github.com/mem0ai/mem0/blob/main/mem0/vector_stores/opensearch.py#L126), which is [not supported in the Serverless offering](https://github.com/opensearch-project/opensearch-py/issues/646).

## Environment
- **mem0 version**: 1.0.0
- **Vector store**: OpenSearch Serverless (AWS managed)
- **Python version**: 3.12

### The Problem
After upgrading from mem0 v0.1.118 to v1.0.0, all vector insertion operations fail with `NotFoundError(404, '')` when using OpenSearch Serverless.

### Root Cause
In v1.0.0, a call to `indices.refresh()` was added after every vector insertion (line 126 in `mem0/vector_stores/opensearch.py`):

```python
def insert(self, vectors, payloads=None, ids=None):
    # ... insertion code ...
    try:
        self.client.index(index=self.collection_name, body=body)
        # Force refresh to make documents immediately searchable for tests
        self.client.indices.refresh(index=self.collection_name)  # ← Line 126
```

**The issue**: OpenSearch Serverless does not support the `indices.refresh()` API. This is an administrative operation that doesn't exist in the managed serverless service.

### Error Stacktrace
```python
opensearchpy.exceptions.NotFoundError: NotFoundError(404, '')
  File "/var/task/mem0/vector_stores/opensearch.py", line 126, in insert
    self.client.indices.refresh(index=self.collection_name)
```

According to the code comment, the refresh was added to "make documents immediately searchable for tests". However, OpenSearch automatically refreshes indices periodically.

## Reproduction Steps

1. Set up AWS OpenSearch Serverless collection
2. Configure mem0 v1.0.0 with OpenSearch Serverless:
```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "opensearch",
        "config": {
            "host": "xxxxx.us-west-2.aoss.amazonaws.com",
            "port": 443,
            "use_ssl": True,
            "http_auth": auth, 
            "collection_name": "my-collection"
        }
    }
}

memory = Memory.from_config(config)
memory.add("Test message", user_id="test")  # ← Fails with 404
```

## Expected Behavior
Vector insertion should work with both OpenSearch self-managed and Serverless offerings.

## Actual Behavior
Vector insertion fails with `NotFoundError(404)` on OpenSearch Serverless.

tao-to-python:
Fixes #3739. OpenSearch Serverless does not support `indices.refresh()`, causing `NotFoundError(404)` on every vector insertion since v1.0.0. Wraps the refresh call in try/except.

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -122,8 +122,13 @@ class OpenSearchDB(VectorStoreBase):
             }
             try:
                 self.client.index(index=self.collection_name, body=body)
-                # Force refresh to make documents immediately searchable for tests
-                self.client.indices.refresh(index=self.collection_name)
+                # Force refresh to make documents immediately searchable.
+                # Wrapped in try/except because OpenSearch Serverless does not
+                # support the indices.refresh() API (returns 404).
+                try:
+                    self.client.indices.refresh(index=self.collection_name)
+                except Exception:
+                    pass
                 
                 results.append(OutputData(
                     id=id_,


youneshima:
Closing as a duplicate of #3893, which adds a configurable auto_refresh option with tests and docs. Thanks for the contribution. Also please follow the PR template when opening PR.

PR #4457, opened 2026-03-20
fix: keep sync vector-store calls on caller thread

### LINKED ISSUE #1989: [qdrant]: Error in `Error in new_memories_with_actions` while using `QdrantLocal`
### 🐛 Describe the bug

Related Issue: #1976
```python
config = {
    "vector_store": {
        "provider": "qdrant",
        "config": {
            "collection_name": "test_10",
            "path": "./qdrant_storage",
            "on_disk": True,
            "embedding_model_dims": 1536,  # Change this according to your local model's dimensions
        },
    },
    "llm": {
        "provider": "openai",
        "config": {
            "model": "gpt-4o",
            "temperature": 0.2,
            "max_tokens": 1500,
        }
    },
    "embedder": {
        "provider": "openai",
        "config": {
            "model": "text-embedding-3-small"
        }
    },
    "version": "v1.1"
}

m.add("My name is mayank", user_id="asd")
res = m.search("name", user_id="asd")
print(res)
```

Error
```
ERROR:root:Error in new_memories_with_actions: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 8261635904 and this is thread id 6333427712.
```

kgarg2468:
## Summary
- Keep sync vector-store work on the caller thread when graph is disabled.
- Keep vector-store work on the caller thread for local qdrant when graph is enabled.
- Centralize sync dispatch in Memory with a thread-safety gate helper.
- Add regression tests for add, get_all, and search thread affinity when graph is disabled.

## Why
- Local qdrant paths can be thread-affine and may fail when sync vector-store calls run in worker threads.

## Verification
- uv run --extra dev ruff check mem0/memory/main.py tests/test_memory.py
- uv run --extra test pytest tests/test_memory.py -k "thread or qdrant or vector_store" -q
- uv run --extra test pytest tests/test_memory.py -q

Fixes #1989

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -362,6 +362,16 @@ class Memory(MemoryBase):
         if executor is not None:
             executor.shutdown(wait=True)
 
+    def close(self):
+        self._shutdown_sync_executor()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+        return False
+
     def _get_qdrant_local_status(self) -> Optional[bool]:
         if self.config.vector_store.provider != "qdrant":
             return None
@@ -392,20 +402,20 @@ class Memory(MemoryBase):
 
         return qdrant_local_status
 
-    def _run_sync_vector_and_graph(self, vector_store_call, graph_call=None):
+    def _run_sync_vector_and_graph(self, vector_fn, graph_fn=None):
         """Run sync vector-store work with optional graph work using safe dispatch."""
-        if graph_call is None:
-            return vector_store_call(), None
+        if graph_fn is None:
+            return vector_fn(), None
 
         executor = self._get_sync_executor()
         if self._should_run_vector_store_on_caller_thread():
-            future_graph = executor.submit(graph_call)
-            vector_result = vector_store_call()
+            future_graph = executor.submit(graph_fn)
+            vector_result = vector_fn()
             graph_result = future_graph.result()
             return vector_result, graph_result
 
-        future_vector = executor.submit(vector_store_call)
-        future_graph = executor.submit(graph_call)
+        future_vector = executor.submit(vector_fn)
+        future_graph = executor.submit(graph_fn)
         concurrent.futures.wait([future_vector, future_graph])
         vector_result = future_vector.result()
         graph_result = future_graph.result()
@@ -504,14 +514,14 @@ class Memory(MemoryBase):
         def vector_store_call():
             return self._add_to_vector_store(messages, processed_metadata, vector_filters, infer)
 
-        graph_call = None
+        graph_fn = None
         if self.enable_graph:
             graph_filters = deepcopy(effective_filters)
 
-            def graph_call():
+            def graph_fn():
                 return self._add_to_graph(messages, graph_filters)
 
-        vector_store_result, graph_result = self._run_sync_vector_and_graph(vector_store_call, graph_call)
+        vector_store_result, graph_result = self._run_sync_vector_and_graph(vector_store_call, graph_fn)
 
         if self.enable_graph:
             return {
@@ -840,15 +850,15 @@ class Memory(MemoryBase):
         def vector_store_call():
             return self._get_all_from_vector_store(vector_filters, limit)
 
-        graph_call = None
+        graph_fn = None
         if self.enable_graph:
             graph_filters = deepcopy(effective_filters)
 
-            def graph_call():
+            def graph_fn():
                 return self.graph.get_all(graph_filters, limit)
 
         all_memories_result, graph_entities_result = self._run_sync_vector_and_graph(
-            vector_store_call, graph_call
+            vector_store_call, graph_fn
         )
 
         if self.enable_graph:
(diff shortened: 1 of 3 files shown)


xkonjin (commented):
## Code Review

**Overall:** Good fix for a real thread-affinity bug. QdrantLocal (and similar local vector stores) can fail when accessed from worker threads, and this PR correctly keeps those calls on the caller thread while still parallelizing graph work.

### Architecture

The `_should_run_vector_store_on_caller_thread()` / `_run_sync_vector_and_graph()` pattern is clean:
- When graph is disabled: call vector store directly, no thread pool at all
- When graph is enabled + local qdrant: 1 worker for graph, vector store on caller
- Otherwise: 2 workers for both (original behavior)

This correctly preserves parallelism where safe while protecting thread-affine backends.

### Issues

1. **`ThreadPoolExecutor` created per call**
   Every `add()`, `get_all()`, and `search()` call now creates and destroys a `ThreadPoolExecutor`. The original code also did this (`with concurrent.futures.ThreadPoolExecutor() as executor:`), so this isn't a regression — but it's worth noting that for high-throughput use cases, a shared executor would be more efficient. Not a blocker.

2. **`is_local` attribute check**
   `getattr(self.vector_store, "is_local", False)` — does the Qdrant vector store implementation actually expose this attribute? If not, this will silently fall back to the 2-worker path, defeating the fix. The PR should verify this attribute exists on `QdrantLocal` or add it.

3. **`deepcopy(effective_filters)` in graph_call closures**
   Good catch — the original code shared the same filters dict between concurrent calls, which could cause subtle bugs if either mutated it. The `deepcopy` is the right fix. However, it's only applied to the graph call, not the vector store call. If the vector store also mutates filters (unlikely but possible), the same issue could arise in the 2-worker path.

4. **Test coverage gap: graph-enabled + local qdrant path**
   The tests cover the "graph disabled" path thoroughly (verifying caller-thread affinity for add/get_all/search). But the second branch — `enable_graph=True` + local qdrant — isn't tested. That's the most interesting case (1 worker for graph, vector store on caller).

### Positives
- Three separate tests for add/get_all/search thread affinity — thorough
- Thread identity assertion (`threading.get_ident() == caller_thread_id`) is the right way to verify this
- Clean refactor: the helper methods reduce code duplication across all three methods
- PR description is clear with exact test commands

Ship it.

kgarg2468:
Thanks for the feedback, I just pushed a follow-up with additional testing and coverage for the sync Memory robustness changes.

I also ran additional verification on my side to make sure everything is stable and in a good state.

xkonjin (commented):
## Code Review: Thread-Safety Fix for QdrantLocal

**Overall:** Solid fix for a real thread-affinity bug. The refactoring is clean, and the test coverage is notably thorough — including the graph-enabled + local qdrant path and the unknown `is_local` fallback case.

### Strengths
- `_run_sync_vector_and_graph()` centralizes the dispatch logic cleanly across add/get_all/search
- `deepcopy(effective_filters)` prevents subtle filter mutation bugs between concurrent calls
- Reusable `ThreadPoolExecutor` via `_get_sync_executor()` avoids per-call churn (improvement over original code)
- Tests verify thread identity with `threading.get_ident()` — correct approach
- Fallback-to-caller-thread when `is_local` is unknown (with warning log) is a safe default

### Issues

**1. Variable name collision in `add()` (bug)**
The `graph_call` local function shadows the `graph_call` variable (initially set to `None`). This works in CPython because the `def` rebinds the name, but it reads confusingly:
```python
graph_call = None
if self.enable_graph:
    graph_filters = deepcopy(effective_filters)
    def graph_call():  # shadows the None assignment above
        return self._add_to_graph(messages, graph_filters)
```
Same pattern in `get_all()` and `search()`. Consider using `graph_fn` for the function name to avoid the shadowing.

**2. `__del__` is unreliable for cleanup**
`__del__` is not guaranteed to run (especially during interpreter shutdown). The `_shutdown_sync_executor()` call in `__del__` may silently fail or never execute. Consider also implementing `__enter__`/`__exit__` (context manager) or documenting that users should call `reset()` explicitly.

**3. Missing test: 2-worker path (graph enabled + remote qdrant)**
The tests cover graph-disabled (caller thread) and graph-enabled + local qdrant (1 worker + caller thread). The third branch — graph enabled + remote/non-qdrant store — where both calls go to the 2-worker executor is untested. Adding one test that verifies both calls run on worker threads would complete the coverage.

### Minor
- The `_get_qdrant_local_status` method handles the edge case where `is_local` exists but is not a bool — good defensive coding.
- Filter isolation test (`test_add_uses_distinct_filter_dicts_for_vector_and_graph`) is a nice regression guard.

kgarg2468:
Addressed the feedback from review #3985513644 in commit 229f125.

Changes made:
- Renamed shadow-prone local callables from `graph_call` to `graph_fn` in `add`, `get_all`, and `search`, and aligned `_run_sync_vector_and_graph(...)` parameter naming.
- Added deterministic cleanup API on `Memory` via `close()`, plus context manager support (`__enter__`/`__exit__`).
- Routed cleanup paths in `__del__` and `reset()` through `close()` while keeping `__del__` best-effort.
- Strengthened the graph-enabled + remote qdrant 2-worker-path test to assert both vector and graph calls run off the caller thread.
- Added a context-manager cleanup test and updated destructor error-swallowing coverage to validate `__del__` behavior when `close()` raises.

Verification run:
- `uv run --extra dev ruff check mem0/memory/main.py tests/test_memory.py`
- `uv run --extra test pytest tests/test_memory.py -k "thread or qdrant or vector_store" -q`
- `uv run --extra test pytest tests/test_memory.py -q`
- `uv run --extra test pytest tests/vector_stores/test_qdrant.py -q`

kartik-mem0:
Thanks for this, @kgarg2468 — thread-affinity bugs are nasty to chase. The add path was rewritten in the v3 pipeline port (#4805): the `_add_to_graph` + `ThreadPoolExecutor` wrapping your fix targeted no longer exists on `main`, so the patch doesn't map onto the current code. Closing as stale. If you can still observe a sync-call-on-wrong-thread issue in the new pipeline, please reopen with a fresh diff — happy to look.

PR #4476, opened 2026-03-21
fix(anthropic-llm): add baseURL support and handle reasoning model responses

grimmjoww:
## Summary

Two fixes for `AnthropicLLM` in the TypeScript OSS SDK (`mem0-ts/src/oss/src/llms/anthropic.ts`):

- **Add `baseURL` support** — Pass `config.baseURL` to the Anthropic client constructor, enabling Anthropic-compatible API providers (AWS Bedrock, MiniMax, proxies, etc.). The `LLMConfig` interface already defines `baseURL` and the `OpenAILLM` provider already uses it — this brings `AnthropicLLM` to parity.

- **Handle reasoning model responses** — Use `response.content.find(b => b.type === "text")` instead of `response.content[0]`. Reasoning/thinking models (Claude with extended thinking, MiniMax M2.7, etc.) return `thinking` blocks before `text` blocks, causing the previous code to throw `"Unexpected response type from Anthropic API"`.

Both changes are backward-compatible:
- When no `baseURL` is provided, the client defaults to `api.anthropic.com`
- `find()` still returns the first text block for standard (non-reasoning) responses

## Test plan

- [ ] Default Anthropic (no baseURL) — client uses `api.anthropic.com`
- [ ] Custom baseURL — client connects to specified endpoint
- [ ] Missing apiKey — throws `"Anthropic API key is required"`
- [ ] `ANTHROPIC_API_KEY` env fallback — works without explicit config
- [ ] Standard text response (`[{type:"text"}]`) — returns text correctly
- [ ] Reasoning response (`[{type:"thinking"}, {type:"text"}]`) — skips thinking, returns text
- [ ] No text block — throws `"Unexpected response type from Anthropic API"`
- [ ] `generateChat()` wrapper — returns `{content, role:"assistant"}`
- [ ] Multiple text blocks — returns first text block
- [ ] Live API call via custom baseURL — end-to-end verification

All 10 tests pass locally.

### DIFF
--- a/mem0-ts/src/oss/src/llms/anthropic.ts
+++ b/mem0-ts/src/oss/src/llms/anthropic.ts
@@ -45,9 +45,8 @@ export class AnthropicLLM implements LLM {
     const textBlock = response.content.find((b) => b.type === "text");
     if (textBlock) {
       return textBlock.text;
-    } else {
-      throw new Error("Unexpected response type from Anthropic API");
     }
+    throw new Error("Unexpected response type from Anthropic API");
   }
 
   async generateChat(messages: Message[]): Promise<LLMResponse> {


kartik-mem0:
hi @grimmjoww thank you so much for you contribution can you please sign the cla?

grimmjoww:
Closing for now — discovered additional issues with reasoning model compatibility (think tag stripping, token budget for thinking vs output) that need more thorough testing before this is ready. Will reopen with a complete solution.

PR #4480, opened 2026-03-22
fix: replace bare except with except Exception

YouFoxGirl:
## What this PR does

Replace bare `except:` clauses with `except Exception:` to avoid catching
`KeyboardInterrupt` and `SystemExit`. Added a TODO note for future
more specific exception handling where applicable.

## Why this matters

Bare `except:` catches all exceptions including `KeyboardInterrupt`
and `SystemExit`, which can hide bugs and make debugging harder.


kartik-mem0:
Hi, good code hygiene improvement\! However, since this only touches files in the openmemory/ directory and we're sunsetting those components from this repo, it's not mergeable. Closing this — thanks for the PR.

Issue #4141, opened 2026-02-26
[Bug] removeCodeBlocks() regex destroys all LLM output — OSS memory capture completely broken with Anthropic provider (v2.2.3)


neclineeth-kims:
## Bug Description
The `removeCodeBlocks()` function in `dist/oss/index.mjs` and `dist/oss/index.js` contains a critically broken regular expression that destroys all LLM response content instead of stripping markdown code fences. This causes every memory capture attempt to fail silently with `SyntaxError: Unexpected end of JSON input`, resulting in **zero memories ever being stored** in OSS self-hosted mode.

## Environment
- **Package**: mem0ai v2.2.3
- **Mode**: open-source (self-hosted)
- **LLM**: Anthropic Claude Haiku (`claude-haiku-4-5-20251001`)
- **Vector Store**: Qdrant (local, port 6333)
- **Embedder**: Ollama `nomic-embed-text` (port 11434, dims: 768)
- **Node.js**: ES Module environment (`type: module`)
- **OS**: Ubuntu 24.04

## Steps to Reproduce
1. Configure mem0ai in OSS mode with Anthropic as the LLM provider
2. Trigger any memory capture/add operation
3. Observe errors in logs:
4. Run `mem0 stats` — shows 0 memories despite active conversations

### Bug 1 — Broken regex in `removeCodeBlocks()`
Located at line ~2594 in `dist/oss/index.mjs`:
```js
// BROKEN — as shipped
function removeCodeBlocks(text) {
  return text.replace(/[^`]*/g, "");
}
```

The regex `/[^`]*/g` matches **everything that is NOT a backtick** and replaces it with an empty string. This deletes all content from the LLM response, leaving only backtick characters. The resulting empty string then fails `JSON.parse()`.

**Example:**
- LLM returns: `{"facts": ["User is working on openclaw mem0 integration"]}`
- After `removeCodeBlocks()`: `""` (empty string)
- Result: `SyntaxError: Unexpected end of JSON input`

The function name suggests it was intended to strip markdown code fences (` ```json ... ``` `). The correct regex should remove the fence delimiters, not the content between them.

### Bug 2 — `AnthropicLLM.generateResponse()` ignores `responseFormat`
[...]

PR #4484, opened 2026-03-22
fix(anthropic): handle response_format in AnthropicLLM.generate_response

limsaehyun:
## Description

`AnthropicLLM.generate_response()` accepts the `response_format` parameter but **silently ignores it**, causing persistent `json.loads()` failures in `memory/main.py`.

When `main.py` calls `generate_response(response_format={"type": "json_object"})`, the OpenAI provider correctly passes this to the API for constrained decoding. The Anthropic provider drops it entirely — Claude then returns free-form text (markdown, code fences, or prose), and `json.loads()` fails with:

```
mem0.memory.main | ERROR | Invalid JSON response: Expecting value: line 1 column 1 (char 0)
```

This is the **Python SDK** counterpart of the same bug reported in:
- #4108 — Claude wraps JSON in code fences, JSON parsing fails (TypeScript SDK, closed but Python SDK unfixed)
- #4141 — `AnthropicLLM.generateResponse()` ignores `responseFormat` (explicitly called out as Bug 2)

## Root Cause

```python
# mem0/llms/openai.py — works correctly
if response_format:
    params["response_format"] = response_format  # ✅ Passed to API

# mem0/llms/anthropic.py — broken (before this PR)
def generate_response(self, messages, response_format=None, ...):
    # ... response_format is accepted but NEVER used ❌
    response = self.client.messages.create(**params)
    return response.content[0].text
```

## Solution

Two `response_format` types, two strategies — matching what the Anthropic API actually supports:

### `{"type": "json_schema"}` → API-level guarantee ✅

Anthropic's [Structured Outputs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs) (GA since 2026-02-04) supports `output_config.format` with constrained decoding, equivalent to OpenAI's `json_schema` mode:

```python
params["output_config"] = {
    "format": {"type": "json_schema", "schema": schema}
}
```

This provides **100% valid JSON conforming to the schema**, enforced at the token level.

### `{"type": "json_object"}` → best-effort via prompting ⚠️

Unlike OpenAI, **Anthropic does not support schema-less JSON mode** (`json_object`). The `output_config.format.type` only accepts `"json_schema"` with a required `schema` field — there is no equivalent to OpenAI's constrained-decoding `json_object` mode.

For this case, we apply the best available approach:

1. **System prompt instruction** — append "You must respond with valid JSON only"
2. **[Assistant prefill](https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/prefill-claudes-response)** — add `{"role": "assistant", "content": "{"}` to force JSON start
3. **`extract_json()` post-processing** — strip code fences if present (reuses existing mem0 utility)

### Why not the same approach as OpenAI?

| Mode | OpenAI | Anthropic |
|------|--------|-----------|
| `json_schema` (with schema) | `response_format` → constrained decoding | `output_config.format` → constrained decoding ✅ |
| `json_object` (no schema) | `response_format` → constrained decoding | **Not supported by API** — must use prompting ⚠️ |

## Type of change

## How Has This Been Tested?

- [x] Unit Test

6 test cases in `tests/llms/test_anthropic.py`:

| Test | Validates |
|------|-----------|
| `test_generate_response_without_response_format` | No behavior change when response_format is not set |
| `test_generate_response_with_json_object_format` | System prompt + prefill + `{` restoration |
| `test_generate_response_json_object_without_system_message` | Works when no system message in input |
| `test_generate_response_json_object_strips_code_fences` | `extract_json()` cleanup |
| `test_generate_response_with_json_schema_format` | `output_config` translation + no prefill |
| `test_generate_response_with_tools` | Regression: tools still work correctly |

```bash
pytest tests/llms/test_anthropic.py -v  # 6 passed
```

Also verified against the real Anthropic API — both `json_object` and plain text modes returned correctly parsed JSON.

## Checklist:

### DIFF
--- a/mem0/llms/anthropic.py
+++ b/mem0/llms/anthropic.py
@@ -9,6 +9,7 @@ except ImportError:
 from mem0.configs.llms.anthropic import AnthropicConfig
 from mem0.configs.llms.base import BaseLlmConfig
 from mem0.llms.base import LLMBase
+from mem0.memory.utils import extract_json
 
 
 class AnthropicLLM(LLMBase):
@@ -83,5 +84,35 @@ class AnthropicLLM(LLMBase):
             params["tools"] = tools
             params["tool_choice"] = tool_choice
 
+        wants_json_prefill = False
+        if response_format and isinstance(response_format, dict):
+            format_type = response_format.get("type")
+            if format_type == "json_schema" and "json_schema" in response_format:
+                # Use Anthropic structured outputs via output_config
+                schema = response_format["json_schema"]
+                if isinstance(schema, dict) and "schema" in schema:
+                    schema = schema["schema"]
+                params["output_config"] = {"format": {"type": "json_schema", "schema": schema}}
+            elif format_type == "json_object":
+                # Use assistant prefill to guide JSON output
+                wants_json_prefill = True
+                json_instruction = (
+                    "\n\nYou must respond with valid JSON only. "
+                    "Do not include any other text, markdown formatting, or code fences."
+                )
+                if params.get("system"):
+                    params["system"] += json_instruction
+                else:
+                    params["system"] = json_instruction.strip()
+                params["messages"] = list(params["messages"]) + [
+                    {"role": "assistant", "content": "{"}
+                ]
+
         response = self.client.messages.create(**params)
-        return response.content[0].text
+        content = response.content[0].text
+
+        if wants_json_prefill:
+            content = "{" + content
+            content = extract_json(content)
+
+        return content
--- a/tests/llms/test_anthropic.py
+++ b/tests/llms/test_anthropic.py
@@ -1,0 +1,130 @@
+from unittest.mock import Mock, patch
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


themavik (commented):
The `json_object` path with assistant prefill and `extract_json` post-processing is solid. One thing — `params["messages"] = list(params["messages"]) + [...]` copies the list but mutates `params` in place. If the caller reuses that messages list, they won't see the prefill since you copied, which is correct. But if someone passes the same `params` dict to another call, the system prompt mutation persists. Minor, just noting it.

The `json_schema` path using `output_config` is cleaner since Anthropic handles it natively. The schema unwrapping logic (checking for nested `schema` key) handles both OpenAI-style and raw schema formats.

limsaehyun:
Hi @whysosaket , could you take a look at this PR when you get a chance? I've been running into this issue as well, and this fix would be really helpful. It's been open for about a week with no review yet. Thanks!

PR #4495, opened 2026-03-23
fix: preserve custom metadata when updating memory

### LINKED ISSUE #3966: MongoDB Implementation Bug: update replaces metadata entirely

kartik-mem0:
## Description

  MongoDB's `update()` method in the vector store used `{"$set": {"payload": new_payload}}` which replaces the entire `payload` document on update, causing all existing metadata fields (e.g.,
  `bucket_id`, `category`, `source`, `context_id`) to be silently lost.

  The fix uses MongoDB dot notation — `{"$set": {"payload.data": ..., "payload.hash": ...}}` — which only updates the specified fields and preserves everything else. This is the [documented MongoDB
   behavior](https://www.mongodb.com/docs/manual/reference/operator/update/set/) for partial updates on embedded documents.
   
  Fixes #3966

  ## Type of change

  ## How Has This Been Tested?

  - [x] Unit Test

  3 tests added/updated in `tests/vector_stores/test_mongodb.py`:
  - `test_update` — updated to verify `$set` uses dot notation (`payload.key`) instead of full replacement (`payload`)
  - `test_update_payload_only_uses_dot_notation` — verifies only specified fields appear in `$set`, no full payload key
  - `test_update_vector_only_does_not_touch_payload` — verifies vector-only updates don't touch any payload fields

  All 39 tests pass locally (21 in `test_memory.py` + 18 in `test_mongodb.py`).

  ## Checklist:

  ## Maintainer Checklist

  - [x] closes #3966

### DIFF
--- a/tests/vector_stores/test_mongodb.py
+++ b/tests/vector_stores/test_mongodb.py
@@ -207,12 +207,6 @@ def test_update(mongo_vector_fixture):
     """
     Test that update() uses dot notation for payload fields instead of replacing
     the entire payload document.
-
-    Fixes #3966 — MongoDB's $set with {"payload": new_dict} replaces the entire
-    payload, losing existing metadata. Using {"payload.key": value} preserves
-    fields not being updated.
-
-    See: https://www.mongodb.com/docs/manual/reference/operator/update/set/
     """
     mongo_vector, mock_collection, _ = mongo_vector_fixture
     vector_id = "id1"


utkarsh240799:
I think the issue is prevalent in other vector stores as well. On exploration I found that the same metadata loss still happens on Qdrant, Pinecone, Milvus, PGVector, Supabase, and Redis. If that's the case, I feel the better fix would be to have to have the change in `_update_memory()` in `main.py`(line ~1257).  Something like starting from deepcopy(existing_memory.payload) and overlaying the new fields on top 

It would be great if you can check this out once @kartik-mem0 .

kartik-mem0:
@whysosaket merging this for now but there will be a follow up for other providers where we will test each providers and see if this problem exist for them or not. 

current scope fixes the mongodb part.

Issue #3720, opened 2025-11-06
The value of created_at differs between search and get_all results


enchyisle:
With the following script, I can reproduce the scenario where the value of "created_at" of a memory is sometimes (2/3) changed to the value of "updated_at" when returned by the SEARCH API. 
Is that by design?

```python
from time import sleep
from mem0 import MemoryClient
import os
from dotenv import load_dotenv
import json
load_dotenv()

client = MemoryClient(api_key=os.getenv("MEM0_API_KEY"))
messages = [
  { "role": "user", "content": "Hi, I'm Alex. I'm a vegetarian and I'm allergic to nuts." },
  { "role": "assistant", "content": "Hello Alex! I see that you're a vegetarian with a nut allergy." }
]

res = client.add(messages, user_id="alex002",timestamp=1683364760)
print(res)

sleep(30) 
query = "What am I allergic to?"

filters = {
 "OR":[
    {
       "user_id":"alex002"
    }
 ]
}

result = client.search(query, filters=filters)
print("search result:")
print(json.dumps(result, indent=4))

results = client.get_all(filters=filters)
print("get all result:")
print(json.dumps(results, indent=4))
``` 

**search result:**
{
    "results": [
        {
            "id": "abbed504-3b9b-4908-a051-d72078821dc6",
            "memory": "Allergic to nuts",
            "user_id": "alex002",
            "metadata": null,
            "categories": [
                "health"
            ],
            "created_at": "2023-05-06T02:19:20-07:00",
            "updated_at": "2025-11-06T01:22:06-08:00",
            "expiration_date": null,
            "score": 0.7412318,
            "structured_attributes": {
                "day": 6,
                "hour": 9,
                "year": 2023,
                "month": 5,
                "minute": 19,
                "quarter": 2,
                "is_weekend": true,
                "day_of_week": "saturday",
                "day_of_year": 126,
                "week_of_year": 18
            }
        },
        {
            "id": "9f9b9bf0-fa14-4fc8-8da8-16cfbc864cec",
            "memory": "Vegetarian",
            "user_id": "alex002",
            "metadata": null,
            "categories": null,
            "created_at": "2025-11-06T01:22:05-08:00",
            "updated_at": "2025-11-06T01:22:05-08:00",
            "expiration_date": null,
            "score": 0.47491649999999996,
            "structured_attributes": {
                "year": 2023,
                "month": 5,
                "day": 6,
                "hour": 9,
                "minute": 19,
[...]

PR #4499, opened 2026-03-23
fix(memory): set updated_at on creation and preserve pre-existing created_at

utkarsh240799:
## Description

`_create_memory` (both sync `Memory` and async `AsyncMemory`) had three gaps in how it handled timestamps:

### Problem 1: `updated_at` was never set on creation

When a memory was first created, only `created_at` was set. `updated_at` remained `None` in the vector store payload until the memory was explicitly updated via `_update_memory`. This caused:

- **Inconsistent API responses** — consumers had to null-check `updated_at` since it was sometimes a string, sometimes `None`
- **Broken sorting/filtering** — sorting by `updated_at` (e.g. "most recently changed") dropped all never-updated memories
- **Wrong "last modified" queries** — queries like "memories changed since X" missed memories created after X but never updated

### Problem 2: `created_at` was unconditionally overwritten

```python
metadata["created_at"] = datetime.now(timezone.utc).isoformat()
```

If any caller pre-set `created_at` in the metadata dict (e.g. an integration backdating a memory, or the platform API honoring a user-provided `timestamp`), it was silently overwritten with the current time.

### Problem 3: History table missing `updated_at` for ADD events

`_update_memory` passed `updated_at` to `db.add_history()`, but `_create_memory` did not — leaving the `updated_at` column `NULL` for all ADD history entries, even though the value was available.

### Solution

In both sync and async `_create_memory`:

1. Guard `created_at` with `if "created_at" not in metadata` — only auto-generate if not already present
2. Set `updated_at = created_at` — a newly created record was "last modified" at creation time
3. Pass `updated_at` to `db.add_history()` — consistent with `_update_memory` behavior

Related to #3720

## Type of change

## How Has This Been Tested?

19 tests in `tests/memory/test_main.py` (7 new + 12 existing), all passing.

**New tests:**

| Test | What it verifies |
|------|-----------------|
| `test_create_memory_sets_updated_at` | `updated_at` is set on creation, equals `created_at`, is UTC, and is passed to history |
| `test_create_memory_preserves_existing_created_at` | Pre-existing `created_at` in metadata is not overwritten; `updated_at` matches it |
| `test_async_create_memory_sets_updated_at` | Async equivalent of the above |
| `test_async_create_memory_preserves_existing_created_at` | Async equivalent of the above |
| `test_create_then_search_and_get_all_return_same_timestamps` | End-to-end: create a memory, then verify `search()` and `get_all()` return identical non-null `created_at` and `updated_at` |
| `test_update_preserves_created_at_and_updates_updated_at` | After update, `created_at` is unchanged, `updated_at` is refreshed |
| `test_search_and_get_all_consistent_after_update` | After update, `search()` and `get_all()` still return identical timestamps |

**Existing tests (all still pass):**

- `test_create_memory_uses_utc_timestamps` — `created_at` is UTC
- `test_update_memory_uses_utc_timestamps` — `updated_at` is UTC, `created_at` preserved and normalized
- Async equivalents of both
- 4 `_normalize_iso_timestamp_to_utc` tests
- 4 `_add_to_vector_store` error handling tests

Also verified against the broader test suite:

```
tests/memory/test_main.py  — 19 passed
tests/test_memory.py       — 21 passed
tests/test_server_auth.py  — 74 passed
                             ─────────
                             114 passed, 0 failed
```

- [x] Unit Test

## Checklist:

## Maintainer Checklist

- [ ] Related to #3720
- [ ] Made sure Checks passed


mem0/memory/main.py
```diff
@@ -2240,7 +2243,9 @@ async def _create_memory(self, data, existing_embeddings, metadata=None):
         metadata = metadata or {}
         metadata["data"] = data
         metadata["hash"] = hashlib.md5(data.encode()).hexdigest()
-        metadata["created_at"] = datetime.now(timezone.utc).isoformat()
+        if "created_at" not in metadata:
+            metadata["created_at"] = datetime.now(timezone.utc).isoformat()
+        metadata["updated_at"] = metadata["created_at"]
```
whysosaket: Won't this always override ??? 

Should be something like:
`metadata["updated_at"] = datetime.now(timezone.utc).isoformat()` ?

utkarsh240799: Hey @whysosaket , This is intentional. If you notice, it's the _create_memory() not the _update_memory().   The goal is to have updated_at = created_at time for new memories. If we do updated_at = datetime.now(), then there would be a difference between the two times (very small difference, but it would be there).

whysosaket: I see. sgtm

PR #4504, opened 2026-03-23
feat(server): add /health endpoint for container orchestration

tao-to-python:
Adds a `/health` endpoint for Docker healthcheck and Kubernetes liveness/readiness probes. Returns 200 when Memory is initialized, 503 otherwise.

Usage in docker-compose:
```yaml
healthcheck:
  test: ["CMD", "curl", "-f", "http://localhost:8000/health"]
  interval: 30s
  timeout: 10s
  retries: 3
```

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -219,6 +219,17 @@ def reset_memory():
         raise HTTPException(status_code=500, detail=str(e))
 
 
+@app.get("/health", summary="Health check", tags=["System"])
+def health_check():
+    """Health check endpoint for container orchestration (Docker, Kubernetes).
+
+    Returns 200 if the server is running and the memory instance is initialized.
+    """
+    if MEMORY_INSTANCE is None:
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for this — a `/health` endpoint for container orchestration is genuinely useful and something we'd like to have. The server's been refactored since you opened this, though: there's no longer a `MEMORY_INSTANCE` global. The memory instance is now accessed via `get_memory_instance()` (in `server/server_state.py`), which raises `RuntimeError` when the runtime isn't initialized rather than returning `None` — so this PR would `NameError` at runtime as-is, and the health logic needs adapting to the new pattern (roughly: `try get_memory_instance()` → return `{"status": "ok"}`, `except RuntimeError` → 503). Closing since it no longer applies to current `main` and the rebase is non-trivial. We'd still love a `/health` endpoint, so please open a fresh PR against `main` using `get_memory_instance()` (a small test would be great too) and we'll gladly land it.

Issue #3045, opened 2025-06-26
Too many threads created, CPU surges by 100%


starboyate:
### 🐛 Describe the bug

I am running a self hosted version of the mem0 server, and when add or search memory requests increases, there is a 100% CPU usage issue, and thousands or even no upper limit of threads are created

![Image](https://github.com/user-attachments/assets/a90830ec-3071-4158-abe1-7fefa25440d4)

Zynoo71:
Have you solved this problem?

Gloringo:
I'm experiencing the same issue. When using mem0 to store/get some information, I notice a growing number of threads that never seem to terminate. Over time, these threads accumulate and remain idle, causing resource usage to reach 100%.

If anyone has found a workaround or a way to properly clean up these threads, I'd appreciate any suggestions.

Thanks!

me0106:
@Gloringo @Zynoo71 
The source of the bug is the repeated creation of the client when calling posthog within the capture_event function. Just remove the relevant code in  [telemetriy.py](https://github.com/mem0ai/mem0/blob/main/mem0/memory/telemetry.py) 

BTW:  MEM0_TELEMETRY environment is not working

CarlosMG01:
@me0106 
You could also edit the telemetry.py adding the following:

```python
def __init__(self, vector_store=None):
        self.posthog = Posthog(project_api_key=PROJECT_API_KEY, host=HOST, )

        self.user_id = get_or_create_user_id(vector_store)

        if not MEM0_TELEMETRY:
            self.posthog.disabled = True
            self.close()
```

It would be nice to include this fix in future versions.

elepherai:
Has v0.1.117 fixed this issue?

parshvadaftari:
@elepherai This is not an issue witht the pypi. This is due to the localy host setup implementation.

metaouter:
> [@me0106](https://github.com/me0106) You could also edit the telemetry.py adding the following:
> 
> def __init__(self, vector_store=None):
>         self.posthog = Posthog(project_api_key=PROJECT_API_KEY, host=HOST, )
> 
>         self.user_id = get_or_create_user_id(vector_store)
> 
>         if not MEM0_TELEMETRY:
>             self.posthog.disabled = True
>             self.close()
> It would be nice to include this fix in future versions.

I encountered this issue as well. Do we need to set MEM0_TELEMETRY=False and add this code at the same time?

metaouter:
[...]

PR #4506, opened 2026-03-23
feat: add close(), context manager, and __del__ to Memory class

tao-to-python:
Relates to #3376, #3045.

Memory instances hold SQLite connections and background threads that are never cleaned up. Adds `close()`, context manager support (`with Memory.from_config(config) as m:`), and `__del__` safety net.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1287,6 +1287,35 @@ class Memory(MemoryBase):
         )
         return memory_id
 
+    def close(self):
+        """Release resources held by the Memory instance.
+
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Hi @tao-to-python, thanks again for your continued work on improving resource management in mem0!

The `close()`, context manager, and cleanup functionality you proposed here has been implemented and merged via PR #4535, which addresses both the telemetry leak (#3376) and proper resource cleanup.

Since this is now covered, we're closing this PR. We appreciate your dedication to improving the library — your contributions helped shape the final solution!

PR #4508, opened 2026-03-23
feat(opensearch): make KNN engine configurable via knn_engine parameter

tao-to-python:
Relates to #3739.

The KNN engine is hardcoded to `nmslib`, which is not available on OpenSearch Serverless. Adds a `knn_engine` config parameter (default: `nmslib`) so users can switch to `faiss` or `lucene` for Serverless compatibility.

```python
config = {"vector_store": {"provider": "opensearch", "config": {"knn_engine": "faiss"}}}
```


youneshima:
Closing as a duplicate of #4392. Thanks for the contribution.

PR #4510, opened 2026-03-23
fix(proxy): remove auto pip install of litellm at import time

tao-to-python:
The proxy module runs `subprocess.check_call` to `pip install litellm` when the import fails. This is:
- **Security risk**: arbitrary code execution during import
- **Breaks in containers**: read-only filesystems, CI pipelines
- **Surprising behavior**: libraries should not auto-install dependencies

Replaced with a clear `ImportError` message. Also removes unused `subprocess` and `sys` imports.

### DIFF
--- a/mem0/proxy/main.py
+++ b/mem0/proxy/main.py
@@ -1,6 +1,4 @@
 import logging
-import subprocess
-import sys
 import threading
 from typing import List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


alexjcampbell:
I also found this during an audit after the LiteLLM supply chain incident last week. `subprocess.check_call(pip install)` at import time is a real risk.

FWIW we confirmed `from mem0 import Memory` doesn't hit this code path (it's only in mem0.proxy.main), but it's still sitting in everyone's site-packages.

PR #4522, opened 2026-03-24
fix: normalize delete_all vector store list results across providers

Exploreunive:
### Summary
This PR fixes `delete_all()` / `AsyncMemory.delete_all()` so they correctly handle different `vector_store.list()` return formats across providers.

Before this change, `delete_all()` assumed the result always needed `[0]` indexing. That works for providers returning tuple/nested formats, but breaks for providers that already return a flat list of memory objects.

### Problem
`get_all()` already contains compatibility logic for multiple `list()` result shapes, but `delete_all()` was still using older assumptions:

- sync path:
  - `self.vector_store.list(filters=filters)[0]`
- async path:
  - iterating over `memories[0]`

This causes incorrect behavior for providers that return:

```python
[mem1, mem2]
```

instead of:

```python
[[mem1, mem2]]
# or
([mem1, mem2], count)
```

In those cases, `delete_all()` can treat the first memory object as the whole result container.

### Fix
Added a shared normalization helper in both `Memory` and `AsyncMemory` to support:

- flat list: `[mem1, mem2]`
- nested list: `[[mem1, mem2]]`
- tuple with count: `([mem1, mem2], count)`

Updated these code paths to use the normalized result:

- `Memory._get_all_from_vector_store()`
- `Memory.delete_all()`
- `AsyncMemory._get_all_from_vector_store()`
- `AsyncMemory.delete_all()`

### Tests
Added/updated targeted tests covering delete_all compatibility:

- flat list handling
- tuple result handling
- async delete_all flat list handling

### Validation
Ran targeted tests with the project venv:

```bash
.venv/bin/python -m pytest -q tests/test_delete_all_list_formats.py tests/memory/test_main.py -k 'delete_all_handles_flat_list_from_vector_store or delete_all_handles_tuple_from_vector_store or get_all_handles'
```

Result:

```bash
3 passed, 12 deselected
```

### Notes
There is an existing unrelated issue in the current test suite around a broken patch target in `tests/test_main.py`, so this PR keeps the scope focused on the `delete_all()` list-format compatibility bug and validates it with targeted tests.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -790,21 +790,24 @@ class Memory(MemoryBase):
 
         return {"results": all_memories_result}
 
-    def _get_all_from_vector_store(self, filters, limit):
-        memories_result = self.vector_store.list(filters=filters, limit=limit)
+    def _unwrap_list_results(self, memories_result):
+        """Normalize vector_store.list() outputs across providers.
 
-        # Handle different vector store return formats by inspecting first element
+        Providers may return:
+        - [mem1, mem2]
+        - [[mem1, mem2]]
+        - ([mem1, mem2], count)
+        """
         if isinstance(memories_result, (tuple, list)) and len(memories_result) > 0:
             first_element = memories_result[0]
-
-            # If first element is a container, unwrap one level
             if isinstance(first_element, (list, tuple)):
-                actual_memories = first_element
-            else:
-                # First element is a memory object, structure is already flat
-                actual_memories = memories_result
-        else:
-            actual_memories = memories_result
+                return first_element
+            return memories_result
+        return memories_result
+
+    def _get_all_from_vector_store(self, filters, limit):
+        memories_result = self.vector_store.list(filters=filters, limit=limit)
+        actual_memories = self._unwrap_list_results(memories_result)
 
         promoted_payload_keys = [
             "user_id",
@@ -1149,7 +1152,7 @@ class Memory(MemoryBase):
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
         # delete matching vector memories individually (do NOT reset the collection)
-        memories = self.vector_store.list(filters=filters)[0]
+        memories = self._unwrap_list_results(self.vector_store.list(filters=filters))
         for memory in memories:
             self._delete_memory(memory.id)
 
@@ -1874,21 +1877,24 @@ class AsyncMemory(MemoryBase):
 
         return results_dict
 
-    async def _get_all_from_vector_store(self, filters, limit):
-        memories_result = await asyncio.to_thread(self.vector_store.list, filters=filters, limit=limit)
+    def _unwrap_list_results(self, memories_result):
+        """Normalize vector_store.list() outputs across providers.
 
-        # Handle different vector store return formats by inspecting first element
+        Providers may return:
+        - [mem1, mem2]
+        - [[mem1, mem2]]
+        - ([mem1, mem2], count)
+        """
         if isinstance(memories_result, (tuple, list)) and len(memories_result) > 0:
             first_element = memories_result[0]
-
-            # If first element is a container, unwrap one level
             if isinstance(first_element, (list, tuple)):
-                actual_memories = first_element
-            else:
-                # First element is a memory object, structure is already flat
-                actual_memories = memories_result
-        else:
-            actual_memories = memories_result
+                return first_element
+            return memories_result
+        return memories_result
+
+    async def _get_all_from_vector_store(self, filters, limit):
+        memories_result = await asyncio.to_thread(self.vector_store.list, filters=filters, limit=limit)
+        actual_memories = self._unwrap_list_results(memories_result)
 
         promoted_payload_keys = [
             "user_id",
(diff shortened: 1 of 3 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- `_unwrap_list_results` correctly handles all four shapes: `[]`, `[mem1, mem2]`, `[[mem1, mem2]]`, and `([mem1, mem2], count)` — the tuple case covers Qdrant's `client.scroll()` return.
- The core fix is sound. On current `main`, `delete_all` (sync at `main.py:1740`, async at `main.py:3276`) uses `[0]` indexing. The recent wave of vector-store contract fixes (faiss #5725, pinecone #5706, opensearch #5727, mongodb #5729) made those providers return `[[]]` on error, which works with `[0]`. But `databricks.py:830` and `langchain.py:208,211` still return a bare `[]` on exception — `[][0]` raises `IndexError`. This PR is the right fix for that gap.
- Tests cover flat list, tuple, and async flat list; lint is clean.

### Issues

#### Critical (must fix)
1. `mem0/memory/main.py` — `_get_all_from_vector_store` (both sync and async versions) passes `limit=limit` to `vector_store.list()`. The abstract base class signature is `list(filters=None, top_k=None)` and all 25+ concrete implementations use `top_k`. Passing the unknown keyword `limit` raises `TypeError: unexpected keyword argument 'limit'` at runtime for every `get_all()` call. On rebase, change to `top_k=limit` to match the interface.

#### Important (should fix)
1. `mem0/memory/main.py` — `_unwrap_list_results` is defined identically in both `Memory` and `AsyncMemory` as an instance method. This duplicates the same logic in two places with no shared base class to own it. Extract it as a module-level helper function (e.g. `def _unwrap_list_results(memories_result): ...`) above both classes. Both sync and async `delete_all` and `_get_all_from_vector_store` can call it directly.

#### Minor (optional)
1. `tests/test_delete_all_list_formats.py` — placed in `tests/` root while the related `test_main.py` lives in `tests/memory/`. Moving it to `tests/memory/test_delete_all_list_formats.py` keeps the test tree consistent with the module it covers.

### Evidence
- **CI:** Vercel/deploy check red (advisory, fork perms — ignored per policy). No real required checks ran on the conflicting branch.
- **Targeted tests (PR head):** `pytest tests/test_delete_all_list_formats.py tests/memory/test_main.py::test_async_delete_all_handles_flat_list_from_vector_store` → 3 passed in 5.08s
- **Red-green proof:** Full mechanical revert not possible — main has diverged significantly (new prompt imports, decay logic, entity cleanup changes). Verified by code tracing: `[m1, m2][0]` = `m1`; iterating over a single object instead of the list silently misdeletes; `[][0]` = IndexError. `_unwrap_list_results` handles both correctly.
- **Lint:** `ruff check` on all three changed files → all checks passed.
- **Superseded check:** Traced `delete_all` on main against recent vector-store contract fixes. Those fixes made providers return `[[]]` on error — `[0]` works for them. `databricks.py` and `langchain.py` still return bare `[]` on error, so the IndexError gap is real and un-merged. This PR is NOT superseded; it covers a remaining case.
- **Could NOT verify:** Live execution against Databricks or LangChain backends.

### Assessment
Ready to merge: **With fixes** — please rebase onto `main`, change `limit=limit` → `top_k=limit` in both `_get_all_from_vector_store` calls, and consider extracting `_unwrap_list_results` as a module-level function. The core fix is correct and addresses a real gap. Happy to re-review once you push.

For tracking on our side, mind opening a quick issue describing the bug and linking it here? Makes it easier to prioritize and reference.

kartik-mem0 (commented):
## Changes requested — this doesn't currently fix the pagination bug

Thanks for this, and apologies it has sat since March. Two separate things going on, so let me split them.

### The normalisation part is fine, but `main` already has it

`_unwrap_list_results` handles the fact that different vector stores return `list()` results in slightly different shapes. That's a real inconsistency, and `main` grew a helper for exactly it after this PR was opened: `_vector_store_list_rows` at `mem0/memory/main.py:85`, used at `:570` and `:2209`. So on a rebase this would collapse into the existing helper rather than adding a second one.

### The pagination part isn't there

This PR is in the #6627 cluster, but the fix for that issue isn't in the diff. After the change, `delete_all` still does a single call:

```python
memories = self._unwrap_list_results(self.vector_store.list(filters=filters))
for memory in memories:
```

No `top_k`, no loop. All 27 Python vector stores default `top_k` to 100, so this still deletes only the first 100 matches and still returns `{"message": "Memories deleted successfully!"}`. I measured that on a live hosted Supabase project today: 250 memories seeded, 100 deleted, 150 left behind, no error raised.

### Also blocking

CONFLICTING against `main`. Its `main.py` now differs by +861/-2174 versus current `main`, so a rebase here is a substantial piece of work rather than a formality.

### Where this leaves it

Five other PRs target #6627 and I'm recommending the maintainers land #6636, which adds the drain loop with the page size raised to 1000. That covers the issue this PR is linked to. If there's a normalisation gap left over that `_vector_store_list_rows` doesn't handle, that's worth its own small PR against current `main` with a test showing the shape that breaks — that would be a much easier review than rebasing this one, and I'd take it.

kartik-mem0:
Thanks for this, and apologies for the very slow response.

Both halves of this are now covered on main:

- The list-format normalization is already there as `_vector_store_list_rows` (`mem0/memory/main.py:85`), which does the same unwrapping as `_unwrap_list_results`.
- `delete_all` now paginates via #6636, merged as 54328ffd.

Closing as superseded. Appreciate the effort.

PR #4523, opened 2026-03-24
fix: handle empty filters in redis vector search

Exploreunive:
### Summary
This PR fixes `RedisDB.search()` so it no longer crashes when filters are omitted or effectively empty.

### Problem
The Redis vector store built a filter expression like this:

```python
filter = reduce(lambda x, y: x & y, conditions)
```

When `filters` was:
- `None`
- `{}`
- or a dict whose values were all `None`

`conditions` became empty and `reduce(...)` raised an exception.

This makes Redis search less robust than other vector store backends for no-filter / empty-filter calls.

### Fix
- Normalize `filters` with `filters = filters or {}`
- Only build the combined Redis filter expression when at least one non-`None` condition exists
- Otherwise pass `None` as `filter_expression`

### Tests
Added regression coverage in `tests/vector_stores/test_redis.py` for:
- `filters=None`
- `filters={}`
- `filters={"user_id": None}`

### Validation
```bash
.venv/bin/python -m pytest -q tests/vector_stores/test_redis.py
```

Result:
```bash
5 passed
```

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -790,21 +790,24 @@ class Memory(MemoryBase):
 
         return {"results": all_memories_result}
 
-    def _get_all_from_vector_store(self, filters, limit):
-        memories_result = self.vector_store.list(filters=filters, limit=limit)
+    def _unwrap_list_results(self, memories_result):
+        """Normalize vector_store.list() outputs across providers.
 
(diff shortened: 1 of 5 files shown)


kartik-mem0:

PR #4530, opened 2026-03-24
fix(openmemory): sync SQL memories on Mem0 UPDATE in API and MCP

### LINKED ISSUE #3841: Bug: Data Inconsistency between UI (Postgres) and MCP (Vector Store) due to unhandled UPDATE events

Himanshu-Sangshetti:
## Linked Issue

Closes #3841

## Description

Mem0 can return `event: "UPDATE"` when an existing memory is refined in the vector store. OpenMemory only synced `ADD` results into the SQL `memories` table (REST and MCP), so the UI could show outdated text while semantic search reflected the update.
This PR processes **`ADD`** and **`UPDATE`** the same way when persisting Mem0 `memory_client.add()` results: upsert the `Memory` row by id, set content from the result, and append `MemoryStatusHistory` with **`old_state`** taken from the row **before** the change (or `deleted` when inserting a new row). MCP **`DELETE`** handling records **`old_state`** from the actual row state instead of assuming `active`.
Integration tests exercise **`POST /api/v1/memories/`** with a mocked Mem0 client for both **`UPDATE`** and **`ADD`**.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `openmemory/api/tests/test_memories_create_sync.py`: `POST /api/v1/memories/` with mocked `get_memory_client` returning Mem0-shaped `results` for **`UPDATE`** (existing row content updated in SQL + history) and **`ADD`** (new row inserted). Uses `StaticPool` for in-memory SQLite and patches `categorize_memory` to avoid external LLM calls during tests.

## Checklist

### DIFF
--- a/openmemory/api/tests/test_memories_create_sync.py
+++ b/openmemory/api/tests/test_memories_create_sync.py
@@ -1,8 +1,7 @@
-"""
-Integration-style tests: POST /api/v1/memories/ with mocked Mem0 client.
+"""Tests for POST /api/v1/memories/ using a mocked Mem0 memory client.
 
-Validates that ADD and UPDATE events from memory_client.add() are persisted to
-the SQL DB (same code path as production #3841 fix). No separate sync module.
+Exercises persistence of ADD and UPDATE events returned by memory_client.add()
+into the SQL-backed Memory model.
 """
 
 from __future__ import annotations
@@ -68,6 +67,7 @@ def seeded_user_app(session):
 
 
 def test_post_memories_update_event_updates_sql_row(session, seeded_user_app):
+    """When Mem0 returns UPDATE, the existing memory row content is updated in SQL."""
     user, app_row = seeded_user_app
     mid = uuid4()
     session.add(
@@ -120,6 +120,7 @@ def test_post_memories_update_event_updates_sql_row(session, seeded_user_app):
 
 
 def test_post_memories_add_event_inserts_sql_row(session, seeded_user_app):
+    """When Mem0 returns ADD, a new memory row is inserted in SQL."""
     user, app_row = seeded_user_app
     new_id = uuid4()
     mock_mc = MagicMock()


kartik-mem0:
Hey, thanks for putting this together — the UPDATE event sync fix is solid work. We're sunsetting the OpenMemory components from this monorepo, so we won't be merging changes to these components going forward. Really appreciate the contribution though. Closing this out.

Issue #4519, opened 2026-03-23
Bug: SyntaxError - Named export 'Client' not found in pg module (ESM compatibility with Node.js 22)


microonline:
##🐛 Bug Description

When using @mem0/openclaw-mem0 v0.4.0 with Node.js v22, the recall functionality throws:

```
SyntaxError: Named export 'Client' not found. The requested module 'pg' is a CommonJS module, which may not support all module.exports as named exports.
```

## 🔍 Environment

- **@mem0/openclaw-mem0:** v0.4.0
- **Node.js:** v22.22.0
- **OS:** Linux (Ubuntu)
- **Vector store:** Qdrant (NOT PostgreSQL)
- **Runtime:** OpenClaw gateway

## 📋 Steps to Reproduce

1. Install @mem0/openclaw-mem0@0.4.0
2. Run OpenClaw with Qdrant as vector store
3. Trigger a recall operation (auto-recall on session bootstrap)
4. Observe the SyntaxError in gateway logs

## 🔧 Expected Behavior

Recall should work with Qdrant without importing the `pg` (PostgreSQL) module, or `pg` should be imported correctly as CommonJS.

## 📝 Additional Context

- This is a **breaking change** from v0.3.3 which did not exhibit this issue
- The error occurs even when NOT using PostgreSQL — the plugin tries to import `pg` regardless
- The fix is likely changing `import { Client } from 'pg'` to `import pkg from 'pg'; const { Client } = pkg;` to support both ESM and CommonJS environments
- Stack trace shows the error comes from the plugin's recall implementation, not user code

## 💻 Stack Trace

```
SyntaxError: Named export 'Client' not found. The requested module 'pg' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
import pkg from 'pg';
const { Client } = pkg;
```

---
Would appreciate a fix or guidance. Happy to test any PR.

kartik-mem0:
Looks like the `pg` module is being pulled in even when you're using Qdrant, which shouldn't happen. This is the same ESM/CJS named export issue reported in #4513 -- we're working on a fix for the TS SDK to handle the `pg` import correctly under Node 22.

utkarsh240799:
Hi @microonline ,

Thanks for reporting this! The issue has been resolved with PR https://github.com/mem0ai/mem0/pull/4544. Closing this issue.

Issue #4513, opened 2026-03-23
openclaw-mem0: memory_store/search fail with pg CommonJS import (Named export 'Client' not found)


hxyinan:
## Summary

After upgrading to the latest OpenClaw + `@mem0/openclaw-mem0`, the Mem0 toolchain failed at runtime with:

```txt
SyntaxError: Named export 'Client' not found. The requested module 'pg' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'pg';
const { Client } = pkg;
```

This broke `memory_store` and `memory_search` in a real OpenClaw runtime.

## Environment

- OpenClaw: `2026.3.22`
- Plugin: `@mem0/openclaw-mem0@0.4.0`
- Resolved dependency: `mem0ai@2.4.2`
- Resolved dependency: `pg@8.11.3`
- OS: Linux / WSL2
- Node: `22.22.1`
- Mem0 mode: OSS
- Vector store: Qdrant

## Repro

1. Upgrade OpenClaw to `2026.3.22`
2. Upgrade/install `@mem0/openclaw-mem0@0.4.0`
3. Use Mem0 tools from OpenClaw (`memory_store`, `memory_search`)
4. Runtime fails with the error above

## Likely cause

In the bundled `mem0ai` OSS build, `pg` is imported like this:

```ts
import { Client } from "pg";
```

In this runtime, that import shape fails because `pg` is CommonJS.

## Local workaround that fixed it

Patching the built file from:

```ts
import { Client } from "pg";
```

to:

```ts
import pkg from "pg";
const { Client } = pkg;
```

then restarting the OpenClaw gateway restored the full memory toolchain.

## Verification after patch

After patch + restart, the following passed again:

- `memory_store`
- `memory_search`
- `memory_get`
- `memory_forget`

## Request

Please make the `pg` import in the OSS TS build/runtime compatible with CommonJS `pg`, or otherwise adjust bundling/runtime interop so OpenClaw plugin users don't need a local patch.

kartik-mem0:
This is a duplicate of #4519 -- same root cause with `pg` CommonJS imports breaking under Node 22's ESM resolution. We'll fix this in the next openclaw-mem0 patch so `pg` is either imported correctly or not loaded at all when you're on Qdrant.

Ritavidhata:
Thanks for the detailed bug report and workaround! 🙏

[...]

PR #4531, opened 2026-03-24
fix(oss): use default import for pg in ESM environments (Node.js 22+)

YouFoxGirl:
## 🐛 Bug Fix: ESM compatibility for pgvector import

**Issue:** #4519, #4513

### Problem

Node.js 22 enforces stricter ESM/CJS interop. Named exports from CommonJS modules like `pg` cause:

```
SyntaxError: Named export 'Client' not found. The requested module 'pg' is a CommonJS module, which may not support all module.exports as named exports.
```

This occurs even when not using PostgreSQL, because `factory.ts` statically imports `pgvector.ts`, which uses `import { Client, Pool } from "pg"`.

### Fix

Replace the named import with a default import pattern that works in both ESM and CommonJS:

```typescript
// Before (fails in Node.js 22 ESM)
import { Client, Pool } from "pg";

// After (works everywhere)
import pkg from "pg";
const { Client, Pool } = pkg;
```

### Testing

- Added `pgvector-esm.test.ts` to verify PGVector instantiates without import errors
- All 513 existing tests continue to pass

---

Refs: #4519
Refs: #4513

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/pgvector.ts
+++ b/mem0-ts/src/oss/src/vector_stores/pgvector.ts
@@ -1,4 +1,5 @@
-import { Client, Pool } from "pg";
+import pkg from "pg";
+const { Client, Pool } = pkg;
 import { VectorStore } from "./base";
 import { SearchFilters, VectorStoreConfig, VectorStoreResult } from "../types";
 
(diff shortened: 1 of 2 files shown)


utkarsh240799:
Hi @YouFoxGirl ,

Thanks for taking the time to come up with the fix. The reported issues have been fixed with PR #4544 . Therefore closing this PR.

Thank you.

PR #4497, opened 2026-03-23
refactor: add close() to Memory and thread-safe telemetry singleton

kartik-mem0:
Description 

  Every call to capture_event() in mem0/memory/telemetry.py created a new AnonymousTelemetry instance, which spawned a new PostHog background consumer thread that was never shut down. In           
  long-running services, this caused unbounded thread growth, memory exhaustion, and 100% CPU spikes — eventually killing containers.
                                                                                                                                                                                                     
  Root cause: capture_event() (called on every add, search, get, update, delete, etc.) instantiated a throwaway AnonymousTelemetry object per call instead of reusing a singleton.                   
  
  Fix:                                                                                                                                                                                               
  - Converted to thread-safe lazy singletons with double-checked locking — one PostHog client per process, reused across all operations
  - Added shutdown_telemetry() function + atexit handler as safety net                                                                                                                               
  - Added Memory.close() and AsyncMemory.close() with context manager support (with / async with) so users can explicitly release resources
                                                                                                                                                                                                     
  Fixes #3376                                                                                                                                                                                        
                                                                                                                                                                                                     
  Type of change                                                                                                                                                                                     
                                                            
  - Bug fix (non-breaking change which fixes an issue)                                                                                                                                               
  
  How Has This Been Tested?                                                                                                                                                                          
                                                            
  Reproduced the thread leak locally — 10 capture_event() calls created 11 leaked threads. After the fix, thread count stays at 1.                                                                   
  
  - Unit Test                                                                                                                                                                                        
                                                            
  68 tests pass across 4 test files:                                                                                                                                                                 
  - tests/test_telemetry.py — singleton reuse, shutdown cleanup, thread leak prevention, env var parsing (23 tests)
  - tests/test_memory_lifecycle.py — Memory/AsyncMemory close(), context manager, idempotency, error resilience (18 tests)                                                                           
  - tests/test_main.py — existing integration tests (15 tests)                                                            
  - tests/memory/test_main.py — existing memory tests (12 tests)                                                                                                                                     
                                                                                                                                                                                                     
  Checklist:                                                                                                                                                                                         
                                                                                                                                                                                                     
  - My code follows the style guidelines of this project                                                                                                                                             
  - I have performed a self-review of my own code           
  - I have commented my code, particularly in hard-to-understand areas
  - My changes generate no new warnings
  - I have added tests that prove my fix is effective or that my feature works                                                                                                                       
  - New and existing unit tests pass locally with my changes
[...]

PR #4535, opened 2026-03-25
fix: prevent thread and memory leaks from PostHog telemetry

### LINKED ISSUE #3376: Memory leak and thread leak issues in mem0 library

utkarsh240799:
## Description

Fixes #3376

Every call to `capture_event()` created a **new** `AnonymousTelemetry` instance, which spawned a new PostHog client with a background consumer thread. That thread was **never shut down**, causing unbounded thread and memory growth in any long-running process (web servers, APIs, etc.). Additionally, `Memory` had no `close()` method, so SQLite connections were never explicitly released.

This was confirmed as recently as March 2026 by users on v1.0.4, with reports of containers being regularly killed by OOM.

## Problem

Three resource leaks reported in #3376:

1. **Thread leak**: `capture_event()` in `telemetry.py` instantiated a new `AnonymousTelemetry` (→ new `Posthog()` client → new background consumer thread) on **every call**. These threads were never shut down.
2. **Memory leak**: The accumulated PostHog clients and their threads were never garbage collected, causing unbounded memory growth.
3. **Connection leak**: `Memory` and `AsyncMemory` had no `close()` method, so SQLite connections from `SQLiteManager` were never explicitly released.

### Telemetry singleton (`mem0/memory/telemetry.py`)
- Replaced per-call `AnonymousTelemetry()` construction with a **thread-safe lazy singleton** (`_get_oss_telemetry()`) using double-checked locking — only **one** PostHog client and background thread per process, no matter how many `capture_event()` calls are made.
- Registered an `atexit` handler (`_shutdown_oss_telemetry`) for process-level cleanup. The singleton is **not** shut down per-`Memory.close()` call — this avoids the design flaw identified in PR #4497 where closing one `Memory` instance would kill telemetry for all other living instances.
- Made `AnonymousTelemetry.close()` idempotent by setting `self.posthog = None` after shutdown.
- Added `atexit.register(client_telemetry.close)` for the module-level `client_telemetry` singleton.

### Memory lifecycle (`mem0/memory/main.py`)
- Added `Memory.close()` — closes the SQLite connection via `self.db.close()`.
- Added `__enter__` / `__exit__` for context manager support (`with Memory() as m: ...`).
- Added `AsyncMemory.close()` with `__aenter__` / `__aexit__` for async context manager support (`async with AsyncMemory() as m: ...`).
- Defensive `hasattr` guard so `close()` is safe even if `__init__` failed partway through.

### Backward compatibility
- **No public API changes**: `capture_event()`, `capture_client_event()`, `client_telemetry`, `AnonymousTelemetry`, and `MEM0_TELEMETRY` retain identical signatures and import paths.
- **`Memory` / `AsyncMemory`**: Only new methods added. No existing methods modified. Code that never calls `close()` works exactly as before.
- All new internals (`_oss_telemetry_instance`, `_get_oss_telemetry`, `_shutdown_oss_telemetry`) are private (underscore-prefixed).

### New tests added (`tests/test_telemetry.py` — 43 tests total)

**`TestAnonymousTelemetryClose`** (4 tests):
- `close()` calls `posthog.shutdown()`
- `close()` sets `posthog` to `None` (idempotency)
- Double `close()` doesn't raise and only calls `shutdown()` once
- `capture_event()` is a no-op after `close()`

**`TestTelemetrySingleton`** (8 tests):
- Singleton reuses the same instance across repeated calls
- Only one instance created across 20 concurrent threads (thread safety)
- `atexit.register` called exactly once
- 50 `capture_event()` calls → only 1 `Posthog()` constructor call (core leak fix)
- `_shutdown_oss_telemetry()` closes and clears the singleton
- Double shutdown is idempotent
- Shutdown is a no-op when singleton was never created
- Disabled telemetry never initializes the singleton

**`TestMemoryLifecycle`** (6 tests):
- `close()` calls `db.close()`
- Double `close()` is safe
- `close()` safe when `db` is `None`
- `close()` safe when `db` attribute not set (partial `__init__`)
- Context manager (`with`) calls `close()` on exit
- Context manager calls `close()` even on exception

**`TestAsyncMemoryLifecycle`** (5 tests):
- `close()` calls `db.close()`
- `close()` safe when `db` is `None` or not set
- Async context manager (`async with`) calls `close()` on exit
- Async context manager calls `close()` even on exception

### Full suite results
```
481 passed, 66 skipped, 0 failed
```
All pre-existing tests continue to pass. The 66 skips are pre-existing (Neptune provider requiring `langchain_aws`).

### DIFF
--- a/mem0/memory/telemetry.py
+++ b/mem0/memory/telemetry.py
@@ -67,25 +67,35 @@ class AnonymousTelemetry:
 # once at process exit via an atexit handler.
 _oss_telemetry_instance = None
 _oss_telemetry_lock = threading.Lock()
+_oss_telemetry_shutting_down = False
 
 
-def _get_oss_telemetry(vector_store=None):
+def _get_oss_telemetry():
+    """Return the process-wide AnonymousTelemetry singleton, creating it on first call.
+
+    Returns None after _shutdown_oss_telemetry() has run (interpreter exit).
+    """
     global _oss_telemetry_instance
+    if _oss_telemetry_shutting_down:
+        return None
     if _oss_telemetry_instance is not None:
         return _oss_telemetry_instance
 
     with _oss_telemetry_lock:
+        if _oss_telemetry_shutting_down:
+            return None
         # Double-checked locking
         if _oss_telemetry_instance is not None:
             return _oss_telemetry_instance
-        _oss_telemetry_instance = AnonymousTelemetry(vector_store=vector_store)
+        _oss_telemetry_instance = AnonymousTelemetry()
         atexit.register(_shutdown_oss_telemetry)
         return _oss_telemetry_instance
 
 
 def _shutdown_oss_telemetry():
-    global _oss_telemetry_instance
+    global _oss_telemetry_instance, _oss_telemetry_shutting_down
     with _oss_telemetry_lock:
+        _oss_telemetry_shutting_down = True
         if _oss_telemetry_instance is not None:
             _oss_telemetry_instance.close()
             _oss_telemetry_instance = None
@@ -100,11 +110,9 @@ def capture_event(event_name, memory_instance, additional_data=None):
     if not MEM0_TELEMETRY:
         return
 
-    oss_telemetry = _get_oss_telemetry(
-        vector_store=memory_instance._telemetry_vector_store
-        if hasattr(memory_instance, "_telemetry_vector_store")
-        else None,
-    )
+    oss_telemetry = _get_oss_telemetry()
+    if oss_telemetry is None:
+        return
 
     event_data = {
         "collection": memory_instance.collection_name,
--- a/tests/test_telemetry.py
+++ b/tests/test_telemetry.py
@@ -149,9 +149,11 @@ class TestTelemetrySingleton:
     def setup_method(self):
         # Reset singleton state before each test
         telemetry_module._oss_telemetry_instance = None
+        telemetry_module._oss_telemetry_shutting_down = False
 
     def teardown_method(self):
         telemetry_module._oss_telemetry_instance = None
+        telemetry_module._oss_telemetry_shutting_down = False
 
     def test_singleton_reuses_instance(self):
         """_get_oss_telemetry() should return the same instance on repeated calls."""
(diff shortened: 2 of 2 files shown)


mem0/memory/telemetry.py
```diff
@@ -55,16 +57,50 @@ def capture_event(self, event_name, properties=None, user_email=None):
...
+# Thread-safe lazy singleton for OSS telemetry.
+# A single AnonymousTelemetry instance (and its underlying PostHog client /
+# background thread) is reused for the lifetime of the process instead of
+# creating a new one on every capture_event() call.  The singleton is shut down
+# once at process exit via an atexit handler.
+_oss_telemetry_instance = None
+_oss_telemetry_lock = threading.Lock()
+
+
+def _get_oss_telemetry(vector_store=None):
```
whysosaket: The `vector_store` parameter is only used on the **first** call — subsequent calls return the cached singleton and silently ignore the argument. This is a subtle semantic change from the old code, which created a fresh `AnonymousTelemetry(vector_store=...)` each time.

In practice this is probably fine (the `vector_store` is only used by `get_or_create_user_id()` to derive a persistent identifier), but the function signature is misleading — it suggests `vector_store` matters on every call when it does not.

Suggestion: either document that `vector_store` is only used on first initialization, or remove the parameter and pass `None` explicitly in the constructor call on line 81, since the singleton design means it is only meaningful once.

tests/test_telemetry.py
```diff
@@ -91,6 +96,280 @@ def test_capture_client_event_sends_when_enabled(self):
...
+
+    def test_async_context_manager(self):
+        """AsyncMemory should support `async with` and close on exit."""
+        m = self._make_mock_async_memory()
+
+        async def run():
+            async with m:
+                pass
+
+        asyncio.new_event_loop().run_until_complete(run())
```
whysosaket: `asyncio.new_event_loop()` creates a loop that is never closed, leaking file descriptors and potentially producing `ResourceWarning`. Same issue on line 369.

Consider using `asyncio.run(run())` (Python 3.7+), or:
```python
loop = asyncio.new_event_loop()
try:
    loop.run_until_complete(run())
finally:
    loop.close()
```

mem0/memory/telemetry.py
```diff
@@ -55,16 +57,50 @@ def capture_event(self, event_name, properties=None, user_email=None):
...
+    global _oss_telemetry_instance
+    if _oss_telemetry_instance is not None:
```
whysosaket, line 81: Nit/minor: after `_shutdown_oss_telemetry()` runs via `atexit`, `_oss_telemetry_instance` is set to `None`. If any code calls `capture_event()` during interpreter shutdown (e.g. from another `atexit` handler or a `__del__` method), this check will pass, `_get_oss_telemetry()` will create a **new** instance that won't be cleaned up. Mostly theoretical, but could be guarded with a `_shutting_down = True` flag set in `_shutdown_oss_telemetry()` and checked here.

Issue #4543, opened 2026-03-25
mem0ai OSS: pgvector import breaks under Node ESM (Named export 'Client' not found)


fmercurio:
### Summary
When using the **mem0ai OSS bundle** (e.g. via the OpenClaw Mem0 plugin), the `pgvector` vector store fails under **Node ESM** with:

```text
SyntaxError: Named export 'Client' not found. The requested module 'pg' is a CommonJS module
```

Root cause: `pg` is CommonJS and does not expose named exports in ESM. The OSS pgvector store currently does `import { Client } from "pg"`, which crashes during module evaluation.

### Impact
This prevents OpenClaw Mem0 from working reliably:
- `openclaw-mem0: capture failed … Named export 'Client' not found`
- `openclaw-mem0: recall failed … Named export 'Client' not found`

### Environment
- Node: **v22.22.0**
- OpenClaw: **2026.3.24**
- OpenClaw plugin: **@mem0/openclaw-mem0 0.4.0** (open-source mode)
- mem0ai: **2.4.2** (OSS bundle)

### Where
- Source: `mem0-ts/src/oss/src/vector_stores/pgvector.ts`
- Built output: `mem0ai/dist/oss/index.mjs`

### Expected
OSS bundle loads under Node ESM without throwing.

### Actual
Import-time crash in the pgvector store due to CJS/ESM interop.

### Proposed fix
Use a default import + destructure (CJS-safe in ESM), while preserving the `Client` type via a type-only import:

```ts
import type { Client as ClientType } from "pg";
import pkg from "pg";
const { Client } = pkg;

// ...
private client: ClientType;
```

### PR
I opened a PR with this change:
- https://github.com/mem0ai/mem0/pull/4544

PR #4544, opened 2026-03-25
fix(oss): make pgvector pg import compatible with ESM

fmercurio:
#### Problem
In the OSS bundle, the pgvector vector store imports `pg` like this:

```ts
import { Client } from "pg";
```

But `pg` is a **CommonJS** module. Under **Node ESM**, it does not provide named exports, so module evaluation fails with:

```text
SyntaxError: Named export 'Client' not found. The requested module 'pg' is a CommonJS module
```

This breaks downstream integrations (e.g. OpenClaw Mem0 plugin) because the OSS bundle throws during import, causing memory **capture** and **recall** to fail.

#### Solution
Switch to a CJS-safe ESM import style:

```ts
import type { Client as ClientType } from "pg";
import pkg from "pg";
const { Client } = pkg;

private client: ClientType;
```

#### Testing
- Node: v22.22.0
- `pnpm test` in `mem0-ts/` ✅

Closes: #4543

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/pgvector.ts
+++ b/mem0-ts/src/oss/src/vector_stores/pgvector.ts
@@ -1,4 +1,6 @@
-import { Client } from "pg";
+import type { Client as ClientType } from "pg";
+import pkg from "pg";
+const { Client } = pkg;
 import { VectorStore } from "./base";
 import { SearchFilters, VectorStoreConfig, VectorStoreResult } from "../types";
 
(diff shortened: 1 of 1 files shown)


fmercurio:
I force-pushed a new commit with corrected author/committer (Felippe Mercurio <fmercurio@gmail.com>) so the CLA bot should be able to match the signer.

utkarsh240799:
Hi @fmercurio ,

Thank you for your taking the the time to come up with the fix. Can you please sign the CLA?

PR #4546, opened 2026-03-26
Openclaw plugin: Modernize against latest plugin guidance

TheCrazyLex:
## Summary

This is the result of an investigation which checked the plugin against current
openclaw source, guidance and docs.

The changes are narrow/minimal fixes and improvements to align things:

- make activation instructions match how OpenClaw actually selects memory plugins
- keep the manifest aligned with the runtime-supported config surface
- remove the remaining shared-session tool scoping risk
- finish the migration off the deprecated SDK root import
- move auto-recall onto the explicit prompt-mutation hook that now fits this plugin's behavior

### 1. `9f6ddfa9` Document memory slot selection in setup

OpenClaw treats memory plugins as an exclusive slot, so install plus config
alone does not make this plugin active. This commit updates the README and setup
examples to show the actual activation step:

- set `plugins.slots.memory = "openclaw-mem0"`

Effect:

- users following the setup guide now land on a working memory-plugin selection
  flow instead of a partially configured but inactive plugin

### 2. `659880bd` Document disableHistory in plugin manifest

The runtime already supported `oss.disableHistory`, and the README/tests treated
that support as intentional. The real issue was not a broad parser redesign; it
was a schema drift in `openclaw.plugin.json`.

This commit:

- adds `oss.disableHistory` to the manifest schema
- adds a focused regression test to keep the manifest aligned with the runtime
  parser

Effect:

- schema-driven config validation and UI/tooling now reflect the runtime's
  actual supported config surface

### 3. `5e13c223` Scope memory tools by OpenClaw tool context

The plugin's session-sensitive tools were still using a shared `currentSessionId`
fallback, which could bleed across concurrent sessions.

This commit moves those tools onto OpenClaw's official tool-factory API so each
tool instance captures its own `ctx.sessionKey`, while leaving the CLI's
best-effort fallback behavior unchanged.

This commit also:

- updates the local SDK shim to match the current host contract
- documents the new scoping behavior and the remaining CLI boundary
- adds regression coverage for per-session isolation

Effect:

- the main runtime path now scopes memory operations to the correct session
  without relying on shared mutable state

### 4. `053c60f6` Migrate plugin off deprecated OpenClaw SDK root

After the tool-scoping work, the remaining SDK-alignment gap was the plugin
entry itself: it still imported from the deprecated `openclaw/plugin-sdk` root
and exported a plain entry object.

This commit:

- switches imports to the focused `openclaw/plugin-sdk/plugin-entry` surface
- wraps the plugin with `definePluginEntry(...)`
- updates the local type shim
- adds the small build/test support needed so this package still verifies
  without bundling the full `openclaw` dependency

Effect:

- the plugin now matches OpenClaw's documented entrypoint pattern and is less
  exposed to future SDK-root deprecation fallout

### 5. `85cb9679` Use before_prompt_build for auto-recall

The plugin's auto-recall only injects dynamic prompt context; it does not need
model/provider override behavior. On current OpenClaw, that makes
`before_prompt_build` the better lifecycle hook than legacy
`before_agent_start`.

This commit:

- moves auto-recall onto `before_prompt_build`
- keeps the actual recall logic and injected context behavior unchanged
- adds a regression test for the new hook registration
- updates docs to match the current hook model

Effect:

- auto-recall now runs on the explicit prompt-mutation lifecycle that matches
  the plugin's intent and current OpenClaw guidance

## Testing

- `pnpm test`
- `pnpm build`

Current branch result:

- 4 test files passed
- 83 tests passed
- build completed successfully

### DIFF
--- a/openclaw/README.md
+++ b/openclaw/README.md
@@ -68,7 +68,7 @@ The `agentId` is always namespaced under the configured `userId` (e.g. `agentId:
 
 ### Concurrency safety
 
-Lifecycle hooks (`before_agent_start`, `agent_end`) use `ctx.sessionKey` directly from the event context rather than shared mutable state. This prevents race conditions when multiple sessions run concurrently (e.g. multiple Telegram users chatting simultaneously).
+Lifecycle hooks (`before_prompt_build`, `agent_end`) use `ctx.sessionKey` directly from the event context rather than shared mutable state. This prevents race conditions when multiple sessions run concurrently (e.g. multiple Telegram users chatting simultaneously).
 
 Agent tools are registered via OpenClaw's tool factory context and use `ctx.sessionKey` directly, so session-scoped tool calls stay isolated across concurrent sessions as well.
 
--- a/openclaw/hook-lifecycle.test.ts
+++ b/openclaw/hook-lifecycle.test.ts
@@ -1,0 +1,113 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { provider, createProvider } = vi.hoisted(() => {
+  const provider = {
+    add: vi.fn(),
+    search: vi.fn(),
+    get: vi.fn(),
+    getAll: vi.fn(),
+    delete: vi.fn(),
+  };
+
+  return {
+    provider,
+    createProvider: vi.fn(() => provider),
+  };
+});
+
+vi.mock("./providers.ts", () => ({
+  createProvider,
+}));
+
+import memoryPlugin from "./index.ts";
+
+function createApi() {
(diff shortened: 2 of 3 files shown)


TheCrazyLex:
@deshraj @whysosaket Can you guys check this out? There are some important fixes in this - I can split this into multiple PRs if needed but I think it goes together well and is tested.

RonaldKepken:
I was gonna look into some of these items too, good job!

@whysosaket

TheCrazyLex:
@utkarsh240799 Please check when you can :)

kartik-mem0:
closing this pr as v1.0.6 follows this patterns.

PR #4548, opened 2026-03-26
fix: add json-repair fallback for malformed LLM JSON responses

### LINKED ISSUE #3918: Unterminated JSON causes new_retrieved_facts to fail
### Description

When using **mem0 with Gemini models** for memory extraction, the LLM produces **malformed JSON**, even though prompts explicitly require strict JSON-only output.

This results in JSON parsing failures within mem0 and causes **memory addition to fail or silently skip valid memories**.

The issue is **non-deterministic** and reproducible across identical inputs and configurations. Based on observed behavior, this is likely caused by **Gemini’s internal reasoning (thinking) mode**, where partial or additional output leaks into the final response and breaks the expected JSON structure.

### Errors Observed

```

Error in new_retrieved_facts: Unterminated string starting at: line 2 column 3 (char 4)

Error in new_retrieved_facts: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)

Invalid JSON response: Expecting property name enclosed in double quotes: line 10 column 56 (char 227)

```

### Environment

- **mem0 version**: 1.0.2
- **Python**: 3.12
- **LLM Provider**: Gemini  
  - `gemini-3-flash-preview`  
  - `gemini-2.5-flash`
- **Embedder**: `models/gemini-embedding-001`
- **Vector Store**: Supabase
- **Feature Used**: `custom_fact_extraction_prompt`

---

### Steps to Reproduce

1. Configure mem0 with Gemini and a strict JSON-based `custom_fact_extraction_prompt`
2. Call `memory.add()` with normal conversation messages
3. Run the same code multiple times

**Observed result:**
- Some runs succeed
- Some runs fail with JSON parsing errors

---

### Suspected Cause

This issue may be caused by **Gemini’s thinking mode**

Similar behavior has been observed in other LLMs when thinking modes are enabled (see #3564).

### Additional Context / Questions

- This issue is reproducible when using `custom_fact_extraction_prompt`
- Similar JSON issues exist for other providers 

**Is there a supported way to:**
- disable thinking / reasoning mode, or
- force strict JSON-only output through configuration

Spectual:
## Problem

Fixes #3918

Gemini and certain other LLMs occasionally return JSON that is technically malformed (trailing commas, unquoted keys, single-quoted strings, truncated output, etc.) even when `response_format={"type": "json_object"}` is requested. The existing two-level fallback — `json.loads` → `extract_json` — is insufficient to recover from these cases, causing `memory extraction` to silently return an empty list and lose all facts from the conversation turn.

## Solution

Introduce a **three-level fallback** for the fact-retrieval JSON parsing in both the **sync** (`_add_to_vector_store`) and **async** (`_add_to_vector_store_async`) code paths:

| Level | Strategy | Notes |
|-------|----------|-------|
| 1 | `json.loads(response, strict=False)` | Fast path, unchanged |
| 2 | `extract_json(response)` + `json.loads` | Existing fallback |
| 3 | `json_repair.loads(response)` | **New** — handles Gemini-style quirks |

- Levels 2 and 3 emit a `warnings.warn` so the caller is informed without flooding logs.
- `json_repair` is imported with a `try/except` guard; the library is optional at import time and level 3 is only reached when it is present.
- `json-repair>=0.30.0` is added to the core `dependencies` in `pyproject.toml`.

## Testing

```python
import json_repair

# Simulate Gemini-style output with trailing comma
bad_json = '{"facts": ["user likes hiking", "user is vegetarian",]}'
result = json_repair.loads(bad_json)
assert result["facts"] == ["user likes hiking", "user is vegetarian"]
```

## Checklist

- [x] Both sync and async paths updated identically
- [x] Graceful degradation when `json_repair` is not installed
- [x] Warning logged at levels 2 and 3 for observability
- [x] `json-repair` added to `pyproject.toml` core dependencies
- [x] Syntax verified (`ast.parse`)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -7,6 +7,12 @@ import logging
 import os
 import uuid
 import warnings
+
+try:
+    import json_repair
+    _JSON_REPAIR_AVAILABLE = True
+except ImportError:
+    _JSON_REPAIR_AVAILABLE = False
 from copy import deepcopy
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Union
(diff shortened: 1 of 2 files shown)


SaharshPatel24 (commented):
Great fix! The 3-level fallback is clean and handles the issue well 🙌

One small suggestion — the fallback logic is duplicated verbatim between sync and async paths. Could extract it into a helper:

```python
def _parse_facts_from_response(response: str) -> list:
    try:
        return json.loads(response, strict=False)["facts"]
    except json.JSONDecodeError:
        try:
            facts = json.loads(extract_json(response), strict=False)["facts"]
            warnings.warn("Level 2 fallback triggered", stacklevel=3)
            return facts
        except Exception:
            if _JSON_REPAIR_AVAILABLE:
                repaired = json_repair.loads(response)
                warnings.warn("Level 3 fallback triggered", stacklevel=3)
                if isinstance(repaired, dict) and "facts" in repaired:
                    return repaired["facts"]
                elif isinstance(repaired, list):
                    return repaired
            raise
```

Both sync and async then just call `_parse_facts_from_response(cleaned_response)`. Not a blocker at all, just easier to maintain in one place!

kartik-mem0:
Thanks for tackling this, @Spectual — and sorry for the overlapping effort. We're consolidating the malformed-JSON fix on #4662, which handles it at the shared `extract_json()` utility (so every caller benefits), stays stdlib-only, and ships with tests. Closing this one so the work stays in a single place — your review on #4662 would be very welcome. Appreciate the contribution! 🙏

PR #4556, opened 2026-03-26
fix: use persistent platform-appropriate dirs instead of /tmp for vector store defaults

### LINKED ISSUE #4279: Vector store defaults to /tmp/{provider} - fails with system services and restricted environments

utkarsh240799:
## Linked Issue

Closes #4279
Supersedes #4284

## Description

Vector store defaults use `/tmp/{provider}` when no explicit path is configured. This causes failures and data loss in multiple environments:

| Environment | Problem |
|---|---|
| **Windows** | `/tmp` doesn't exist at all |
| **macOS LaunchAgent** | `/tmp` cleared on reboot, limited permissions |
| **Linux systemd** | `/tmp` may have `noexec` or private namespace restrictions |
| **Docker** | `/tmp` is ephemeral unless volume-mounted |

### Problems found and fixed

1. **`mem0/vector_stores/configs.py`** — Hardcoded `f"/tmp/{provider}"` default path for all vector stores with a `path` field (Qdrant, ChromaDB, FAISS).

2. **`mem0/configs/vector_stores/qdrant.py`** — `QdrantConfig` had its own hardcoded `Field("/tmp/qdrant")` default, separate from the central config logic. Neither PR #4284 nor #4440 addressed this.

3. **`mem0/configs/vector_stores/chroma.py`** — The sentinel check `path == "/tmp/chroma"` (used to distinguish default-injected paths from user-provided paths in cloud config detection) was not updated. This would break cloud config detection after changing the default path.

4. **`mem0/vector_stores/faiss.py`** — FAISS implementation had its own hardcoded `f"/tmp/faiss/{collection_name}"` fallback path, completely independent of the config system. Neither open PR caught this.

5. **No migration path** — Existing users with data at `/tmp/{provider}` would silently lose access after upgrading, with no warning or guidance.

6. **No `MEM0_DATA_DIR` override** — The issue discussion requested an env var override, which neither PR implemented.

### Solution

- Extracted a reusable `get_default_vector_store_path(provider)` helper with precedence: `MEM0_DATA_DIR` → `XDG_DATA_HOME` → `APPDATA` → `~/.local/share/mem0/{provider}`
- Updated all four locations where `/tmp` was hardcoded (central config, Qdrant config, ChromaDB sentinel, FAISS fallback)
- Added a startup warning when legacy data exists at `/tmp/{provider}` with clear migration instructions
- Updated all documentation references

## Type of Change

- [x] Documentation update

## Breaking Changes

The default vector store path changes from `/tmp/{provider}` to `~/.local/share/mem0/{provider}` (or platform equivalent). Users who explicitly set `path` in their config are unaffected. Users relying on the default will see a warning if legacy data exists at the old location, with instructions to either set `path='/tmp/{provider}'` explicitly or move the data.

## Test Coverage

### Testing

**28 new tests** in `tests/vector_stores/test_default_path.py` covering:

- **`TestGetDefaultVectorStorePath` (8 tests)** — Unit tests for the helper function: default fallback to `~/.local/share`, `XDG_DATA_HOME` support, `APPDATA` support, `MEM0_DATA_DIR` precedence, env var priority ordering, per-provider isolation, and verification that no path ever resolves to `/tmp`.

- **`TestVectorStoreConfigPathInjection` (5 tests)** — Integration tests through `VectorStoreConfig`: default path injection for Qdrant, explicit path preservation, backward compatibility with explicit `/tmp` paths, and env var propagation through the full config chain.

- **`TestQdrantConfigDefault` (2 tests)** — Qdrant-specific: verifies `QdrantConfig`'s `default_factory` produces a persistent path, and the full `VectorStoreConfig` → `QdrantConfig` flow works.

- **`TestChromaDbConfigDefault` (3 tests)** — ChromaDB-specific: default path via `VectorStoreConfig`, cloud config correctly strips the new default path, and explicit custom path + cloud config still raises.

- **`TestFaissDefaultPath` (2 tests)** — FAISS-specific: config-level default and implementation-level fallback both use the new persistent path.

- **`TestLegacyPathWarning` (3 tests)** — Migration warning: fires when legacy `/tmp` data exists, does not fire when empty, does not fire when explicit path is provided.

- **`TestIssue4279Scenarios` (5 tests)** — End-to-end scenario tests matching the environments described in issue #4279: Windows (APPDATA), Linux systemd (XDG), Docker (MEM0_DATA_DIR), macOS (home directory), and all three affected providers.

**All 176 relevant tests pass** (28 new + 148 existing including Qdrant, ChromaDB, FAISS, memory, and integration tests).

## Checklist

### DIFF
--- a/docs/components/vectordbs/dbs/faiss.mdx
+++ b/docs/components/vectordbs/dbs/faiss.mdx
@@ -52,7 +52,7 @@ Here are the parameters available for configuring FAISS:
 | Parameter | Description | Default Value |
 | --- | --- | --- |
 | `collection_name` | The name of the collection | `mem0` |
-| `path` | Path to store FAISS index and metadata | `/tmp/faiss/<collection_name>` |
+| `path` | Path to store FAISS index and metadata | `~/.local/share/mem0/faiss/<collection_name>` |
 | `distance_strategy` | Distance metric strategy to use (options: 'euclidean', 'inner_product', 'cosine') | `euclidean` |
 | `normalize_L2` | Whether to normalize L2 vectors (only applicable for euclidean distance) | `False` |
 
--- a/docs/components/vectordbs/dbs/qdrant.mdx
+++ b/docs/components/vectordbs/dbs/qdrant.mdx
@@ -73,7 +73,7 @@ Let's see the available parameters for the `qdrant` config:
 | `client` | Custom client for qdrant | `None` |
 | `host` | The host where the qdrant server is running | `None` |
 | `port` | The port where the qdrant server is running | `None` |
-| `path` | Path for the qdrant database | `/tmp/qdrant` |
+| `path` | Path for the qdrant database | `~/.local/share/mem0/qdrant` |
 | `url` | Full URL for the qdrant server | `None` |
 | `api_key` | API key for the qdrant server | `None` |
 | `on_disk` | For enabling persistent storage | `False` |
@@ -85,7 +85,7 @@ Let's see the available parameters for the `qdrant` config:
 | `embeddingModelDims` | Dimensions of the embedding model | `1536` |
 | `host` | The host where the Qdrant server is running | `None` |
 | `port` | The port where the Qdrant server is running | `None` |
-| `path` | Path for the Qdrant database | `/tmp/qdrant` |
+| `path` | Path for the Qdrant database | `~/.local/share/mem0/qdrant` |
 | `url` | Full URL for the Qdrant server | `None` |
 | `apiKey` | API key for the Qdrant server | `None` |
 | `onDisk` | For enabling persistent storage | `False` |
--- a/docs/open-source/overview.mdx
+++ b/docs/open-source/overview.mdx
@@ -77,7 +77,7 @@ Mem0 Open Source delivers the same adaptive memory engine as the platform, but p
   Mem0 OSS works out of the box with sensible defaults:
   - LLM: OpenAI `gpt-4.1-nano-2025-04-14` (via `OPENAI_API_KEY`)
   - Embeddings: OpenAI `text-embedding-3-small`
-  - Vector store: Local Qdrant instance storing data at `/tmp/qdrant`
+  - Vector store: Local Qdrant instance storing data at `~/.local/share/mem0/qdrant`
   - History store: SQLite database at `~/.mem0/history.db`
   - Reranker: Disabled until you configure a provider
 
--- a/docs/open-source/python-quickstart.mdx
+++ b/docs/open-source/python-quickstart.mdx
@@ -81,7 +81,7 @@ print(results)
 By default `Memory()` wires up:
 - OpenAI `gpt-4.1-nano-2025-04-14` for fact extraction and updates
 - OpenAI `text-embedding-3-small` embeddings (1536 dimensions)
-- Qdrant vector store with on-disk data at `/tmp/qdrant`
+- Qdrant vector store with on-disk data at `~/.local/share/mem0/qdrant`
 - SQLite history at `~/.mem0/history.db`
 - No reranker (add one in the config when you need it)
 </Note>
--- a/mem0/configs/vector_stores/chroma.py
+++ b/mem0/configs/vector_stores/chroma.py
@@ -2,6 +2,8 @@ from typing import Any, ClassVar, Dict, Optional
 
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
+from mem0.vector_stores.configs import get_default_vector_store_path
+
 
 class ChromaDbConfig(BaseModel):
     try:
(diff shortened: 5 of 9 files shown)


whysosaket:
@utkarsh240799 this seems to be a high risk PR, as if we change the default path of the vector stores then won't it mean that existing users those who update will lose all their data?

utkarsh240799:
> @utkarsh240799 this seems to be a high risk PR, as if we change the default path of the vector stores then won't it mean that existing users those who update will lose all their data?

Yes, I would agree that it's high risk PR. I've  added a startup warning when legacy data exists at /tmp/{provider} with clear migration instructions. I'm open to suggestions, if you have any better way to handle this>
cc. @whysosaket

whysosaket:
> > @utkarsh240799 this seems to be a high risk PR, as if we change the default path of the vector stores then won't it mean that existing users those who update will lose all their data?
> 
> Yes, I would agree that it's high risk PR. I've added a startup warning when legacy data exists at /tmp/{provider} with clear migration instructions. I'm open to suggestions, if you have any better way to handle this> cc. @whysosaket

Yes, that makes sense but if we still merge this, this will be a breaking change and not go into patch releases.

Can we have a process where legacy data is automatically migrated if present?

PR #4565, opened 2026-03-27
fix(ts): work around Qdrant Cloud "Illegal host" error

### LINKED ISSUE #3915: [Bug] Illegal host error when using QdrantClient with Qdrant Cloud in mem0 OSS (TypeScript)
## Environment
- mem0ai/oss version: latest (npm)
- @qdrant/js-client-rest version: latest
- Runtime: Next.js 16 on Railway (serverless)
- Qdrant: Qdrant Cloud

## Problem

When using mem0 OSS (TypeScript) with Qdrant Cloud, I consistently get an "Illegal host" error, even when passing a pre-configured QdrantClient or using the url parameter directly.

### Attempt 1: Pre-configured QdrantClient with host+port+https
\`\`\`typescript
const qdrantClient = new QdrantClient({
  host: "xxx.us-west-1-0.aws.cloud.qdrant.io",
  port: 6333,
  apiKey: "xxx",
  https: true,
})

const memory = new Memory({
  vectorStore: {
    provider: "qdrant",
    config: {
      client: qdrantClient,
      collectionName: "my-collection",
    },
  },
  // ... other config
})
\`\`\`
**Result:** "Illegal host" error

### Attempt 2: Pre-configured QdrantClient with url parameter
\`\`\`typescript
const qdrantClient = new QdrantClient({
  url: "https://xxx.us-west-1-0.aws.cloud.qdrant.io:6333",
  apiKey: "xxx",
})

const memory = new Memory({
  vectorStore: {
    provider: "qdrant",
    config: {
      client: qdrantClient,
      collectionName: "my-collection",
    },
  },
  // ... other config
})
\`\`\`
**Result:** "Illegal host" error

### Attempt 3: Direct config without pre-configured client
\`\`\`typescript
const memory = new Memory({
  vectorStore: {
    provider: "qdrant",
    config: {
      url: "https://xxx.us-west-1-0.aws.cloud.qdrant.io:6333",
      apiKey: "xxx",
      collectionName: "my-collection",
    },
  },
  // ... other config
})
\`\`\`
**Result:** "Illegal host" error

## Important Finding

**Direct QdrantClient connection works perfectly:**
\`\`\`typescript
const client = new QdrantClient({ 
  url: "https://xxx.us-west-1-0.aws.cloud.qdrant.io:6333",
  apiKey: "xxx"
})
const collection = await client.getCollection("my-collection")
// ✅ Works fine!
\`\`\`

This confirms:
1. Qdrant Cloud URL and API key are correct
2. Network connectivity is fine
3. The issue is within mem0's internal handling of the Qdrant client

## Suspected Cause

Looking at the mem0-ts source code at `src/oss/src/vector_stores/qdrant.ts`, when a pre-configured `client` is passed, it should be used directly (lines 37-38). However, the "Illegal host" error suggests that somewhere in the mem0 codebase (possibly during embedding or other operations), a different Qdrant connection is being attempted with incorrect parameters.

## Questions

1. Is there a known working configuration for Qdrant Cloud with mem0 OSS (TypeScript)?
2. Are there any examples of successful Qdrant Cloud integration in serverless environments?
3. Could this be related to how mem0 internally reinitializes the vector store during add/search operations?

## Workaround Attempted

I've tried all combinations of:
- `host` + `port` + `https: true`
- `url` parameter only
- Pre-configured client vs direct config
- With and without explicit port in URL

None of them work when going through mem0, but all work when using QdrantClient directly.

## Related Issues

- qdrant/qdrant-js#59 (Illegal host when using URL without explicit port)

Would appreciate any guidance on how to properly configure Qdrant Cloud with mem0 OSS. Thank you!

utkarsh240799:
## Problem

When using the mem0 TypeScript SDK with Qdrant Cloud, users report an "Illegal host" error regardless of connection method — pre-configured `QdrantClient`, `url` parameter, or `host`+`port` config. Direct `QdrantClient` usage outside of mem0 works fine.

This traces to an upstream bug in `@qdrant/js-client-rest` ([qdrant/qdrant-js#59](https://github.com/qdrant/qdrant-js/issues/59)) where the URL parser incorrectly handles ports in HTTPS URLs on certain methods.

The current code in `mem0-ts/src/oss/src/vector_stores/qdrant.ts` passes `config.url` directly to `new QdrantClient({ url })` without any mitigation.

## Solution

**1. Port extraction workaround (for mem0-managed clients)**

When `config.url` is provided, parse the port and pass it explicitly to `QdrantClient` alongside the URL. This bypasses the faulty URL parsing in the upstream client:

```typescript
if (config.url) {
  params.url = config.url;
  try {
    const parsedUrl = new URL(config.url);
    if (parsedUrl.port) {
      params.port = parseInt(parsedUrl.port, 10);
    }
  } catch (_) {}
}
```

The `try/catch` ensures malformed URLs don't crash the constructor. Default ports (443 for HTTPS, 80 for HTTP) are not extracted since `URL.port` returns empty for them, preserving existing behavior.

If `host`+`port` are also provided in config, they take precedence since they're applied after the URL block.

**2. JSDoc for pre-configured clients**

For users passing a pre-configured `QdrantClient` via `config.client`, mem0 can't modify the client's internal state. Added JSDoc documenting that users must pass `port` explicitly when constructing their own client for Qdrant Cloud.

## Scope & limitations

- The fix covers the case where mem0 creates the `QdrantClient` internally (via `url` config)
- For pre-configured clients, the fix is documentation-only — users must apply the workaround themselves
- The "Illegal host" error did not reproduce in our testing with `@qdrant/js-client-rest@1.13.0` against Qdrant Cloud v1.17.0, suggesting the upstream bug may be version-specific. The fix is kept as defensive code to protect users on affected versions

## Testing

**Unit tests (8 tests)** — `qdrant-url-port.test.ts`:
- Extracts port from HTTPS URL with explicit port (`:6333`)
- Extracts port from HTTP URL with explicit port
- Does not set port when URL uses default HTTPS port (443)
- Does not set port when URL uses default HTTP port (80)
- `host`+`port` config overrides URL-extracted port
- Handles invalid URL gracefully without crashing
- Skips `QdrantClient` constructor when pre-configured client is passed
- Handles explicit `:443` in HTTPS URL (default port, `URL.port` returns empty)

**E2E validation** against a live Qdrant Cloud free-tier cluster:
- Confirmed connection + full CRUD works via `url` with `:6333`
- Confirmed connection works via `url` without port
- Confirmed pre-configured client with explicit `port` works
- Verified the upstream bug does not reproduce on current client version (v1.13.0)

**Regression** — full unit test suite passes (562/562 tests, 34/34 suites)

Closes #3915

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/qdrant.ts
+++ b/mem0-ts/src/oss/src/vector_stores/qdrant.ts
@@ -62,7 +62,9 @@ export class Qdrant implements VectorStore {
         try {
           const parsedUrl = new URL(config.url);
           params.port = parsedUrl.port ? parseInt(parsedUrl.port, 10) : 6333;
-        } catch (_) {}
+        } catch (_) {
+          params.port = 6333;
+        }
       }
       if (config.host && config.port) {
         params.host = config.host;
--- a/mem0-ts/src/oss/tests/qdrant-url-port.test.ts
+++ b/mem0-ts/src/oss/tests/qdrant-url-port.test.ts
@@ -124,6 +124,7 @@ describe("Qdrant URL port extraction (qdrant/qdrant-js#59 workaround)", () => {
 
     expect(capturedParams).toBeDefined();
     expect(capturedParams!.url).toBe("not-a-valid-url");
+    expect(capturedParams!.port).toBe(6333);
   });
 
   it("does not pass port when using pre-configured client", () => {


mem0-ts/src/oss/src/vector_stores/qdrant.ts
```diff
@@ -44,6 +58,13 @@ export class Qdrant implements VectorStore {
       }
       if (config.url) {
         params.url = config.url;
+        // Workaround for qdrant/qdrant-js#59: explicitly pass port to avoid "Illegal host" error
+        try {
+          const parsedUrl = new URL(config.url);
+          if (parsedUrl.port) {
```
whysosaket: This conditional makes it mandatory for the user to pass the port, can we not make it default to 6333 in case user doesn't pass this explicitly? @utkarsh240799

PR #4574, opened 2026-03-27
fix(openclaw): scope session memories to unique UUID, fix isNewSession detection

gitchrisqueen:
## Problem

After issuing `/new` to start a fresh conversation, OpenClaw agents would hallucinate and pretend to continue tasks from previous sessions. Two bugs in `openclaw/index.ts` were responsible.

### Bug 1 — `isNewSession` hardcoded to `true`
```ts
const isNewSession = true; // line ~1105 of compiled dist
```
This caused a broad recall of *"active projects, recent decisions"* on **every single message**, not just actual new sessions. The constant was a placeholder that was never implemented.

### Bug 2 — Session memory `run_id` used the static channel key
```ts
run_id: sessionId // sessionId = ctx.sessionKey e.g. "agent:main:main"
```
`ctx.sessionKey` never changes across `/new` commands (it identifies the channel, not the conversation). So all conversations on the same channel shared the same session memory namespace. After `/new`, the new session would recall stale memories from the previous session, causing the agent to act as if it was still mid-task.

## Fix

1. **Real new-session detection**: Track `lastSeenSessionUuid` per channel. Compare `ctx.sessionId` (the unique UUID that changes on every `/new`) against it. Set `isNewSession = true` only when the UUID actually changes.

2. **Session memory isolation**: Use `ctx.sessionId` (UUID) as `run_id` for session-scoped memory recall and capture, with fallback to `ctx.sessionKey` for older gateways that don't provide `sessionId`.

## Testing

Verified locally:
- Before: After `/new`, agent received injected memories like "working on task X" and hallucinated continuing it
- After: After `/new`, session memories from the previous conversation are not injected; agent starts clean


kartik-mem0:

PR #4580, opened 2026-03-28
Fix contradicting memory loss: ADD new fact when DELETing old one

### LINKED ISSUE #4536: Suggest change to memory adding with contradicting facts
### Component

Core / Python SDK

### Summary

A clear summary of the bug.
The DEFAULT_UPDATE_MEMORY_PROMPT in the configurations says that when an added fact is contradictory, then the fact is deleted.  This is indeed the behavior, but I was thinking perhaps this is not good behavior, perhaps instead the memory should be updated or the new fact added in addition to the deletion.  Otherwise we may miss critical information.  It is true that deletion does eliminate the conflict, but it doesn't really resolve it.

### Steps to Reproduce

```python
from mem0 import Memory

m = Memory()
a = "I love Chinese food"
b = "I hate Chinese food"
# add the first fact
m.add(a, run_id="1")
# add the second fact which contradicts it
m.add(b, run_id="1")
## {'results': [{'id': '03d2022c-ed2e-4cef-9cbf-41a748dd3481', 'memory': 'Likes Chinese food', 'event': 'DELETE'}]}
m.get_all(run_id="1")
# empty since only deletion
# {'results': []}
```

### Expected Behavior

This is the expected behavior according to the prompt, but I would have expected it to now contain the "Likes Chinese food".  After all, if the user now asks for a recipe, it should know that they like Chinese food and this would be relevant to the recipe.  

### Actual Behavior

It deleted the first memory without adding the second memory.

### Environment

- mem0 version: 1.0.5
- Python/Node version: 3.13.2
- OS: Windows

joaquinhuigomez:
## Summary
- Update DEFAULT_UPDATE_MEMORY_PROMPT to ADD the new fact when DELETing a contradicting old one
- Prevents silent data loss when memory contradictions occur

The DELETE section of the update memory prompt instructed the LLM to delete contradicting memories but never to add the replacement fact. For example, adding "I hate Chinese food" after "I love Chinese food" would delete the old memory and discard the new one, leaving the memory store empty.

The fix updates the prompt instructions and example so the LLM emits both a DELETE for the old memory and an ADD for the new contradicting fact. The memory processing pipeline already handles ADD and DELETE events independently, so no code changes are needed beyond the prompt.

Added a test to verify the prompt contains ADD instructions in the contradiction/DELETE section.

Fixes #4536

### DIFF
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -262,7 +262,7 @@ Please note to return the IDs in the output from the input IDs only and do not g
 
 3. **Delete**: If the retrieved facts contain information that contradicts the information present in the memory, then you have to delete the old memory AND add the new fact. This ensures that the new information replaces the old contradicting information rather than both being lost.
 Or if the direction is to delete the memory without any replacement, then you have to delete it.
-Please note to return the IDs in the output from the input IDs only for DELETE and NONE events. For the new ADD entry, generate a new ID.
+Please note to return the IDs in the output from the input IDs only for DELETE and NONE events.
 - **Example**:
     - Old Memory:
         [
@@ -290,7 +290,6 @@ Please note to return the IDs in the output from the input IDs only for DELETE a
                     "event" : "DELETE"
                 },
                 {
-                    "id" : "2",
                     "text" : "Dislikes cheese pizza",
                     "event" : "ADD"
                 }


joaquinhuigomez:
Good point — removed the id field from the ADD entry. The pipeline ignores it anyway. Updated.

joaquinhuigomez:
Thanks for the approval — anything else needed to merge?

kartik-mem0:
Closing as superseded by the v3 memory pipeline (#4805).

The v3 algorithm replaced `UPDATE_MEMORY_PROMPT` with `ADDITIVE_EXTRACTION_PROMPT`, which is ADD-only — contradictions are now handled via memory linking instead of DELETE operations. The prompt changes in this PR target code that's no longer in the active code path.

Thank you for the contribution! The underlying issue (#4536) is now resolved by the architectural change in v3.

PR #4587, opened 2026-03-28
fix: reuse shared ThreadPoolExecutor in add/get_all/search to prevent thread leak

### LINKED ISSUE #4586: Thread leak: ThreadPoolExecutor created per call in search/add/get_all, never properly cleaned up
## Summary

`Memory.search()`, `Memory.add()`, and `Memory.get_all()` each create a new `concurrent.futures.ThreadPoolExecutor()` per call. Under sustained server traffic, threads from these executors accumulate and are never cleaned up, causing memory exhaustion.

## Environment

- mem0: latest (installed via pip)
- Python: 3.13
- Runtime: FastAPI + uvicorn, Kubernetes pods
- Also running: Sentry SDK (threading integration enabled)

## Problem

Three locations in `mem0/memory/main.py` create per-call executors:

```python
# Line 371 (add)
with concurrent.futures.ThreadPoolExecutor() as executor:
    future1 = executor.submit(self._add_to_vector_store, ...)
    future2 = executor.submit(self._add_to_graph, ...)

# Line 698 (get_all)
with concurrent.futures.ThreadPoolExecutor() as executor:
    future_memories = executor.submit(self._get_all_from_vector_store, ...)

# Line 837 (search)
with concurrent.futures.ThreadPoolExecutor() as executor:
    future_memories = executor.submit(self._search_vector_store, ...)
```

The `with` block calls `executor.shutdown(wait=True)`, but threads aren't always properly cleaned up — particularly when Sentry's threading integration wraps thread functions and holds references.

## Production Impact

- Thread count grows at ~1/minute under ~50 req/s
- After 8 hours: **800+ idle threads**, each consuming ~10 MB VmData
- Pods eventually hit memory limits and get OOM-killed

Thread tracer confirmed the source:
```
THREAD_TRACE: new thread 'ThreadPoolExecutor-157_0' (type=Thread)
  File "mem0/memory/main.py", line 838
    future_memories = executor.submit(self._search_vector_store, ...)
```

Each `ThreadPoolExecutor-N_0` has a different N — confirming a new executor per call.

## Proposed Fix

Use a shared executor at the class or module level instead of creating one per call:

```python
class Memory:
    _executor = concurrent.futures.ThreadPoolExecutor(max_workers=4)
    
    def search(self, ...):
        # Instead of: with ThreadPoolExecutor() as executor:
        future = self._executor.submit(self._search_vector_store, ...)
        ...
```

This reuses the same 4 threads across all operations instead of creating new ones per call.

## Workaround

We monkey-patch `concurrent.futures.ThreadPoolExecutor` in mem0's module to delegate to a shared pool:

```python
import mem0.memory.main as _mem0_main

shared = ThreadPoolExecutor(max_workers=4)

class _SharedExecutor(ThreadPoolExecutor):
    def submit(self, fn, *args, **kwargs):
        return shared.submit(fn, *args, **kwargs)
    def __exit__(self, *args):
        pass  # Don't shut down shared pool

_mem0_main.concurrent.futures.ThreadPoolExecutor = _SharedExecutor
```

---

*This issue was generated with Claude Opus 4.6 — proofread by @gromdimon*

SaharshPatel24:
Fixes #4586

## Problem
`add()`, `get_all()`, and `search()` each created a new `ThreadPoolExecutor` per call. Under sustained traffic this causes thread accumulation — ~1 thread/min at 50 req/s, leading to 800+ idle threads and OOM kills after hours.

## Fix
Replace the three per-call executors with a single `self._executor` initialized once in `__init__` and reused across all calls.

## Testing
- Existing tests pass (`tests/configs/`, `tests/memory/test_main.py`, `tests/memory/test_memory_utils.py`, `tests/memory/test_storage.py`)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1,5 +1,6 @@
 import asyncio
 import concurrent
+import concurrent.futures
 import gc
 import hashlib
 import json
@@ -303,6 +304,7 @@ class Memory(MemoryBase):
                 self.config.vector_store.provider, telemetry_config
             )
         capture_event("mem0.init", self, {"sync_type": "sync"})
+        self._executor = concurrent.futures.ThreadPoolExecutor()
 
     @classmethod
     def from_config(cls, config_dict: Dict[str, Any]):
@@ -438,14 +440,13 @@ class Memory(MemoryBase):
         else:
             messages = parse_vision_messages(messages)
 
-        with concurrent.futures.ThreadPoolExecutor() as executor:
-            future1 = executor.submit(self._add_to_vector_store, messages, processed_metadata, effective_filters, infer)
-            future2 = executor.submit(self._add_to_graph, messages, effective_filters)
+        future1 = self._executor.submit(self._add_to_vector_store, messages, processed_metadata, effective_filters, infer)
+        future2 = self._executor.submit(self._add_to_graph, messages, effective_filters)
 
-            concurrent.futures.wait([future1, future2])
+        concurrent.futures.wait([future1, future2])
 
-            vector_store_result = future1.result()
-            graph_result = future2.result()
+        vector_store_result = future1.result()
+        graph_result = future2.result()
 
         if self.enable_graph:
             return {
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for the fix, @SaharshPatel24! The per-call `ThreadPoolExecutor` leak from #4586 was since resolved on `main` — the non-graph path no longer spawns a pool at all, and the graph path uses a short-lived context-managed executor. So this PR is no longer needed; closing as superseded. Really appreciate the effort here. 🙏

PR #4588, opened 2026-03-28
fix: use CrossEncoder for cross-encoder reranker models

### LINKED ISSUE #4033: SentenceTransformerReranker does not load as described in the docs
### 🐛 Describe the bug

_This is AI-Generated but I checked all of it, and the fix provided seems to work_

# Bug Report: SentenceTransformerReranker uses wrong class for cross-encoder models

## Summary

The `SentenceTransformerReranker` loads cross-encoder models (e.g. `cross-encoder/ms-marco-MiniLM-L-6-v2`) using `SentenceTransformer()` instead of `CrossEncoder()`. This causes the reranker to fail silently: all `rerank_score` values end up as `0.0`, and results are not reordered by relevance.

## Environment

- **Package**: mem0ai
- **Version**: 1.0.3
- **Python**: 3.11
- **OS**: macOS / Linux (Docker)
- **sentence-transformers**: 3.4.1

## Expected Behavior

When configuring the reranker with a cross-encoder model:

```python
config = {
    "reranker": {
        "provider": "sentence_transformer",
        "config": {
            "model": "cross-encoder/ms-marco-MiniLM-L-6-v2",
            "device": "cpu",
            "batch_size": 32,
            "top_k": 10,
        },
    },
}
```

Search results should be reordered by semantic relevance, with non-zero `rerank_score` values reflecting the cross-encoder's relevance scores.

## Actual Behavior

1. At initialization, this warning appears:

   ```
   No sentence-transformers model found with name cross-encoder/ms-marco-MiniLM-L-6-v2. Creating a new one with mean pooling.
   ```
2. All search results have `rerank_score: 0.0`.
3. Results are not reordered—the vector-search order is preserved.
4. The reranker's `except` block silently catches the failure and assigns `rerank_score = 0.0` to all documents.

## Root Cause

In `mem0/reranker/sentence_transformer_reranker.py` (line 54):

```python
self.model = SentenceTransformer(self.config.model, device=self.config.device)
```

For cross-encoder models like `cross-encoder/ms-marco-MiniLM-L-6-v2`, the correct class is `CrossEncoder`, not `SentenceTransformer`:

- **SentenceTransformer (Bi-Encoder)**: Encodes sentences separately and produces embeddings. Used for similarity via embeddings.
- **CrossEncoder**: Takes query–document pairs and outputs relevance scores. Used for reranking.

Passing a cross-encoder model name to `SentenceTransformer()` causes the library to create an incorrect "mean pooling" model, which then fails when `predict(pairs)` is called, triggering the fallback in the `except` block.

## Suggested Fix

Load the model with `CrossEncoder` when the model name indicates a cross-encoder (e.g. starts with `cross-encoder/`):

```python
from sentence_transformers import SentenceTransformer
from sentence_transformers import CrossEncoder

# In __init__:
if self.config.model.startswith("cross-encoder/"):
    self.model = CrossEncoder(self.config.model, device=self.config.device)
else:
    self.model = SentenceTransformer(self.config.model, device=self.config.device)
```

Or detect based on model type and use the appropriate class consistently.

## References

- [sentence-transformers CrossEncoder docs](https://www.sbert.net/docs/package_reference/cross_encoder/cross_encoder.html)
- [mem0 Sentence Transformer Reranker docs](https://docs.mem0.ai/components/rerankers/models/sentence_transformer) – recommends `cross-encoder/ms-marco-MiniLM-L-6-v2` as default
- [sentence-transformers #3536](https://github.com/huggingface/sentence-transformers/issues/3536) – related "mean pooling" fallback behavior for unknown models

Bortlesboat:
## Summary
- Fixes #4033 — `SentenceTransformerReranker` silently produced all-zero scores with cross-encoder models
- Root cause: cross-encoder models were loaded via `SentenceTransformer()` instead of `CrossEncoder()`, causing silent fallback to mean pooling and zero scores
- Added detection logic for cross-encoder model names and conditional initialization with the correct class

## Changes
- `mem0/reranker/sentence_transformer_reranker.py` — added `CrossEncoder` import, `_detect_cross_encoder()` method, conditional init
- `tests/rerankers/test_sentence_transformer_reranker.py` — 14 new unit tests

## Test plan
- [x] All 45 reranker tests pass (31 existing + 14 new)
- [x] Cross-encoder models now produce meaningful rerank scores
- [x] Sentence-transformer models continue to work unchanged

### DIFF
--- a/mem0/reranker/sentence_transformer_reranker.py
+++ b/mem0/reranker/sentence_transformer_reranker.py
@@ -6,7 +6,7 @@ from mem0.configs.rerankers.base import BaseRerankerConfig
 from mem0.configs.rerankers.sentence_transformer import SentenceTransformerRerankerConfig
 
 try:
-    from sentence_transformers import SentenceTransformer
+    from sentence_transformers import SentenceTransformer, CrossEncoder
     SENTENCE_TRANSFORMERS_AVAILABLE = True
 except ImportError:
     SENTENCE_TRANSFORMERS_AVAILABLE = False
@@ -41,8 +41,26 @@ class SentenceTransformerReranker(BaseReranker):
             )
 
         self.config = config
-        self.model = SentenceTransformer(self.config.model, device=self.config.device)
+        self._is_cross_encoder = self._detect_cross_encoder(self.config.model)
+        if self._is_cross_encoder:
+            self.model = CrossEncoder(self.config.model, device=self.config.device)
+        else:
+            self.model = SentenceTransformer(self.config.model, device=self.config.device)
         
+    @staticmethod
+    def _detect_cross_encoder(model_name: str) -> bool:
+        """Detect if a model is a cross-encoder based on its name.
+
+        Cross-encoder models require the ``CrossEncoder`` class from
+        sentence-transformers.  Using ``SentenceTransformer`` with these
+        models silently falls back to mean pooling and produces all-zero
+        relevance scores.
+        """
+        if not model_name:
+            return False
+        name_lower = model_name.lower()
+        return "cross-encoder" in name_lower or "reranker" in name_lower
+
     def rerank(self, query: str, documents: List[Dict[str, Any]], top_k: int = None) -> List[Dict[str, Any]]:
         """
         Rerank documents using sentence transformer cross-encoder.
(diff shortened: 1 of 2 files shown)


leosok:
hi there, this fixes a bug I reported - what needs to be done to merge?

kartik-mem0:

PR #4590, opened 2026-03-28
feat(memory): support org-scoped multi-tenant search via org_id parameter

### LINKED ISSUE #4589: feat: support org-scoped multi-tenant search (search across multiple user_id namespaces)
## Problem

When building multi-tenant applications, it's common to have two types of memories:

1. **Personal memories** — scoped to a single user (`user_id=employee_id`)
2. **Organization-wide memories** — shared across all users in an org (e.g., stored under `user_id=f"org_{org_id}"`)

Currently, searching across both requires **two separate API calls** because `user_id` is injected as a hard `must` filter in `_build_filters_and_metadata()` (line 211 of `memory/main.py`):

```python
if user_id:
    effective_query_filters["user_id"] = user_id  # always becomes a must condition
```

This means even if you pass OR filters like:

```python
client.search(
    query="quarterly targets",
    user_id=employee_id,
    filters={
        "OR": [
            {"user_id": employee_id},
            {"source": "company_knowledge"},
        ],
    },
)
```

The flat `user_id` key injected at line 211 still goes into Qdrant's `must` clause, making the OR ineffective — company-wide memories stored under a different `user_id` are excluded before the `should` (OR) clause is evaluated.

## Current Workaround

Two parallel searches via `asyncio.gather`:

```python
await asyncio.gather(
    client.search(query=query, user_id=employee_id, ...),
    client.search(query=query, user_id=f"org_{org_id}", ...),
)
```

This works but doubles the embedding API calls and adds code complexity.

## Proposed Solution

Introduce `org_id` as a first-class scope identifier alongside `user_id`, `agent_id`, and `run_id`.

### 1. Update `_build_filters_and_metadata`

```python
def _build_filters_and_metadata(
    *,
    user_id: Optional[str] = None,
    agent_id: Optional[str] = None,
    run_id: Optional[str] = None,
    org_id: Optional[str] = None,       # NEW
    actor_id: Optional[str] = None,
    input_metadata: Optional[Dict[str, Any]] = None,
    input_filters: Optional[Dict[str, Any]] = None,
) -> tuple[Dict[str, Any], Dict[str, Any]]:
```

When `org_id` is provided, scope by `org_id` in `must` and move `user_id` into an OR clause instead of `must`:

```python
if org_id:
    base_metadata_template["org_id"] = org_id
    effective_query_filters["org_id"] = org_id
    session_ids_provided.append("org_id")
    # Don't inject user_id into must — let caller control via filters
```

### 2. Update `search()` and `get_all()` signatures

```python
async def search(
    self,
    query: str,
    *,
    user_id: Optional[str] = None,
    org_id: Optional[str] = None,   # NEW
    ...
):
```

### 3. Update validation

Accept `org_id` as a valid scope identifier:

```python
if not any(key in effective_filters for key in ("user_id", "agent_id", "run_id", "org_id")):
    raise ValueError(...)
```

## Example Usage (After)

```python
# Single call — returns personal + org-wide memories
results = await client.search(
    query="quarterly targets",
    org_id=org_id,
    filters={
        "OR": [
            {"user_id": employee_id},
            {"source": "company_knowledge"},
        ],
    },
)
```

## Why This Matters

- Multi-tenant SaaS applications commonly need both user-scoped and organization-scoped memory retrieval
- The current architecture forces N+1 search calls (1 per namespace) which multiplies embedding API costs
- The Qdrant adapter already supports OR/AND/NOT filters correctly — this is purely a `main.py` scoping limitation
- Backward compatible — existing `user_id`-only usage is unchanged

## Context

- The Qdrant adapter's `_create_filter` now properly handles `OR`→`should`, `NOT`→`must_not` — great work on that fix
- The gap is in `_build_filters_and_metadata` which unconditionally injects `user_id` into the flat filter dict (becoming `must`)
- `_process_metadata_filters` correctly transforms OR/NOT into `$or`/`$not`, but the hard `user_id` injection overrides the intent

Happy to submit a PR for this if the approach aligns with the project's direction.

Mukhsin0508:
## Summary

- Adds `org_id` as a first-class scope identifier in `search()` for both `Memory` and `AsyncMemory`
- When `org_id` is provided, `user_id` is **not** injected into the query filter's `must` clause — callers control read-time scoping via explicit filters (e.g. `OR` clauses)
- Registers `org_id` as a Qdrant payload index for efficient filtering

## Problem

Multi-tenant apps need to search both personal memories (`user_id=employee_id`) and organization-wide memories in a single call. Today, `user_id` is unconditionally injected as a hard `must` filter, making it impossible to search across multiple `user_id` values — forcing two separate API calls with double the embedding cost.

## Solution

When `org_id` is present in the `search()` call:
- `org_id` goes into `effective_query_filters` (for reads)
- `user_id` goes into `base_metadata_template` only (for writes) — **not** into query filters
- Caller owns the filter logic via explicit `filters` parameter

```python
# Single call — returns personal + org-wide memories
results = await memory.search(
    query="quarterly targets",
    org_id=org_id,
    filters={
        "OR": [
            {"user_id": employee_id},
            {"source": "company_knowledge"},
        ],
    },
)
```

Without `org_id`, behavior is completely unchanged (backward compatible).

## Test plan

- [x] 8 new unit tests for `_build_filters_and_metadata` with `org_id` (`tests/memory/test_org_filter.py`)
- [x] 2 new Qdrant filter tests for org-scoped AND/OR patterns (`tests/vector_stores/test_qdrant.py`)
- [x] All 36 existing memory tests pass (backward compatibility)
- [x] All 84 existing Qdrant tests pass
- [x] **126 total tests passing, 0 failures**

Closes #4589

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -778,7 +778,6 @@ class Memory(MemoryBase):
         self,
         *,
         user_id: Optional[str] = None,
-        org_id: Optional[str] = None,
         agent_id: Optional[str] = None,
         run_id: Optional[str] = None,
         filters: Optional[Dict[str, Any]] = None,
@@ -789,9 +788,6 @@ class Memory(MemoryBase):
 
         Args:
             user_id (str, optional): user id
-            org_id (str, optional): Organization ID for org-scoped listing. When provided
-                with user_id, lists both personal and org-wide memories. When provided
-                alone, lists all memories in the org. Defaults to None.
             agent_id (str, optional): agent id
             run_id (str, optional): run id
             filters (dict, optional): Additional custom key-value filters to apply to the search.
@@ -807,11 +803,11 @@ class Memory(MemoryBase):
         """
 
         _, effective_filters = _build_filters_and_metadata(
-            user_id=user_id, org_id=org_id, agent_id=agent_id, run_id=run_id, input_filters=filters
+            user_id=user_id, agent_id=agent_id, run_id=run_id, input_filters=filters
         )
 
-        if not any(key in effective_filters for key in ("user_id", "agent_id", "run_id", "org_id")):
-            raise ValueError("At least one of 'user_id', 'agent_id', 'run_id', or 'org_id' must be specified.")
+        if not any(key in effective_filters for key in ("user_id", "agent_id", "run_id")):
+            raise ValueError("At least one of 'user_id', 'agent_id', or 'run_id' must be specified.")
 
         keys, encoded_ids = process_telemetry_filters(effective_filters)
         capture_event(
@@ -1917,7 +1913,6 @@ class AsyncMemory(MemoryBase):
         self,
         *,
         user_id: Optional[str] = None,
-        org_id: Optional[str] = None,
         agent_id: Optional[str] = None,
         run_id: Optional[str] = None,
         filters: Optional[Dict[str, Any]] = None,
@@ -1928,9 +1923,6 @@ class AsyncMemory(MemoryBase):
 
          Args:
              user_id (str, optional): user id
-             org_id (str, optional): Organization ID for org-scoped listing. When provided
-                 with user_id, lists both personal and org-wide memories. When provided
-                 alone, lists all memories in the org. Defaults to None.
              agent_id (str, optional): agent id
              run_id (str, optional): run id
              filters (dict, optional): Additional custom key-value filters to apply to the search.
@@ -1946,12 +1938,13 @@ class AsyncMemory(MemoryBase):
         """
 
         _, effective_filters = _build_filters_and_metadata(
-            user_id=user_id, org_id=org_id, agent_id=agent_id, run_id=run_id, input_filters=filters
+            user_id=user_id, agent_id=agent_id, run_id=run_id, input_filters=filters
         )
 
-        if not any(key in effective_filters for key in ("user_id", "agent_id", "run_id", "org_id")):
+        if not any(key in effective_filters for key in ("user_id", "agent_id", "run_id")):
             raise ValueError(
-                "At least one of 'user_id', 'agent_id', 'run_id', or 'org_id' must be specified for get_all."
+                "When 'conversation_id' is not provided (classic mode), "
+                "at least one of 'user_id', 'agent_id', or 'run_id' must be specified for get_all."
             )
 
         keys, encoded_ids = process_telemetry_filters(effective_filters)


mem0/memory/main.py
```diff
@@ -206,7 +211,17 @@ def _build_filters_and_metadata(
...
 
-    if user_id:
+    if org_id:
+        effective_query_filters["org_id"] = org_id
+        session_ids_provided.append("org_id")
+        # When org_id is present, user_id goes into metadata template (for writes)
+        # but NOT into effective_query_filters — the caller controls read-time
+        # scoping via explicit filters (e.g. OR clauses).
+        if user_id:
+            base_metadata_template["user_id"] = user_id
```
SaharshPatel24: `org_id` is added to `effective_query_filters` (read) but never to `base_metadata_template` (write),  
  so no memory ever gets stored with an `org_id` tag.                                                   
                                                                                                        
  When a user calls `search(org_id="acme")`, it filters for memories where `org_id = "acme"` —          
  but since nothing was written with that tag, results will always be empty.                          
                                                                                                        
  Fix: add `base_metadata_template["org_id"] = org_id` alongside the `user_id` assignment:              
                                                                                                      
      if user_id:                                                                                       
          base_metadata_template["user_id"] = user_id  ✅                                             
          base_metadata_template["org_id"] = org_id        ← missing                                        
                                                                                                        
  Also, `add()` needs `org_id` as a parameter to expose this to callers.

SaharshPatel24:
The original issue (#4589) only asked for org-scoped `search()`. This PR also adds `org_id` to `get_all()` which wasn't requested and could be risky in multi-tenant setups — a single `get_all(org_id="acme")` call would expose every memory across the entire org with no access control. Worth reconsidering whether `get_all()` should be in scope here.

Mukhsin0508:
> The original issue (#4589) only asked for org-scoped `search()`. This PR also adds `org_id` to `get_all()` which wasn't requested and could be risky in multi-tenant setups — a single `get_all(org_id="acme")` call would expose every memory across the entire org with no access control. Worth reconsidering whether `get_all()` should be in scope here.

Good catch, thanks for the review. You're right — get_all(org_id="acme") dumping every memory in an org is a different risk profile than search(), which is bounded by relevance + limit.

I'll remove org_id from get_all() and keep it scoped to search() only, which is what the original issue requested. The search path is inherently safer — results are ranked by similarity and capped by limit, so it's not a bulk data exposure risk.

Will push the update shortly.

Mukhsin0508:
@SaharshPatel24 anything on this yet? it's just our prod is already depending on this company-wide knowledge search... would appreciate if we could publish it in the next release..

Mukhsin0508:
@kartik-mem0 is there anything we need to merge this?

kartik-mem0:
Closing this as the associated issue #4589 has been closed — we're actively deprecating `org_id` across the SDK, so adding it as a new scope parameter doesn't align with the current direction. Thanks for the contribution!

PR #4592, opened 2026-03-28
fix(qdrant): use DatetimeRange for datetime fields in range filters

### LINKED ISSUE #4591: bug: Qdrant adapter uses Range instead of DatetimeRange for datetime fields — silently returns 0 results
## Bug Description

The Qdrant vector store adapter's `_build_field_condition` always uses `Range` (numeric) for range filters, even when the field contains datetime values. This causes datetime range queries to silently fail — **including the exact examples shown in the [official docs](https://docs.mem0.ai/open-source/features/metadata-filtering)**.

## The Documented Example That Fails

From the docs ("Time-based filtering"):

```python
# Created in January 2024
filters = {
    "AND": [
        {"user_id": "user_123"},
        {"created_at": {"gte": "2024-01-01T00:00:00Z"}},
        {"created_at": {"lt": "2024-02-01T00:00:00Z"}}
    ]
}
```

This silently returns 0 results on the Qdrant backend.

## Root Cause

In `mem0/vector_stores/qdrant.py`, `_build_field_condition` (around line 179-180):

```python
if ops & range_ops:
    range_kwargs = {op: value[op] for op in range_ops if op in value}
    return FieldCondition(key=key, range=Range(**range_kwargs))  # ← always numeric Range
```

`Range` is a pydantic model that expects `float` values. When an ISO datetime string is passed, pydantic raises a `ValidationError`:

```
ValidationError: Input should be a valid number, unable to parse string as a number
  [type=float_parsing, input_value='2024-01-01T00:00:00Z', input_type=str]
```

This error propagates up but is often swallowed silently (e.g., when the caller uses `asyncio.gather(..., return_exceptions=True)`), resulting in 0 results with no visible error.

## Why Float Timestamps Don't Work Either

Passing Unix timestamps like `{"gte": 1772134930.0}` passes pydantic validation (it's a valid float), but Qdrant stores `created_at` as ISO strings with a datetime payload index. A numeric `Range` filter cannot match against datetime-indexed string values — so this also returns 0 results.

## Proposed Fix

Qdrant provides `DatetimeRange` specifically for this purpose. It accepts ISO strings, float timestamps, and datetime objects:

```python
from qdrant_client.http.models import DatetimeRange

DatetimeRange(gte="2024-01-01T00:00:00Z")             # ✅ ISO string
DatetimeRange(gte=1772134930.0)                        # ✅ Unix timestamp
DatetimeRange(gte=datetime(2024, 1, 1, tzinfo=UTC))    # ✅ datetime object
```

The fix in `_build_field_condition`:

```python
if ops & range_ops:
    range_kwargs = {op: value[op] for op in range_ops if op in value}
    sample = next(iter(range_kwargs.values()))
    if isinstance(sample, (str, datetime)):
        return FieldCondition(key=key, range=DatetimeRange(**range_kwargs))
    return FieldCondition(key=key, range=Range(**range_kwargs))
```

## Documentation Update

The [metadata filtering docs](https://docs.mem0.ai/open-source/features/metadata-filtering) show ISO string examples for time-based filtering, which is correct. After the fix, the docs could also mention that Unix timestamps are accepted:

```python
# Both formats work
{"created_at": {"gte": "2024-01-01T00:00:00Z"}}    # ISO string
{"created_at": {"gte": 1704067200.0}}                # Unix timestamp
```

## Context

- mem0 itself stores `created_at` as UTC ISO strings via `datetime.now(timezone.utc).isoformat()` (line 1204 of `main.py`)
- This affects any field stored as a datetime string, not just `created_at`
- The enhanced filter syntax documentation lists `gte`/`lte` as supported operators, but they silently fail on datetime fields with the Qdrant backend

Happy to submit a PR for this fix.

Mukhsin0508:
## Summary

- Detects string/datetime values in range filters and uses `DatetimeRange` instead of `Range`
- Fixes documented time-based filter examples that silently return 0 results on the Qdrant backend
- Numeric range filters continue using `Range` unchanged

## Problem

`_build_field_condition` always uses `Range` (numeric) for range operators (`gte`, `lte`, `gt`, `lt`). When ISO datetime strings are passed (e.g. `{"created_at": {"gte": "2024-01-01T00:00:00Z"}}`), pydantic raises a `ValidationError` because `Range` expects floats. This error is often swallowed silently, resulting in 0 results.

This breaks the exact examples shown in the [official metadata filtering docs](https://docs.mem0.ai/open-source/features/metadata-filtering).

## Fix

```python
# Before — always numeric Range
return FieldCondition(key=key, range=Range(**range_kwargs))

# After — detect datetime values
sample = next(iter(range_kwargs.values()))
if isinstance(sample, str) or hasattr(sample, "isoformat"):
    return FieldCondition(key=key, range=DatetimeRange(**range_kwargs))
return FieldCondition(key=key, range=Range(**range_kwargs))
```

## Test plan

- [x] ISO string datetime range uses `DatetimeRange`
- [x] Combined `gte` + `lt` datetime range uses `DatetimeRange`
- [x] Python `datetime` objects use `DatetimeRange`
- [x] Numeric ranges still use `Range` (backward compat)
- [x] All 84 Qdrant tests pass, 0 failures

Closes #4591

### DIFF
--- a/docs/core-concepts/memory-operations/search.mdx
+++ b/docs/core-concepts/memory-operations/search.mdx
@@ -181,13 +181,22 @@ m.search("query", user_id="alice", agent_id="chatbot", run_id="session-123")
 ```
 
 **Filter by Date Range:**
+
+*Platform API:*
 ```python
-# Platform only - date filtering
 client.search("recent memories", filters={
     "AND": [
         {"user_id": "alice"},
         {"created_at": {"gte": "2024-07-01"}}
     ]
+})
+```
+
+*OSS (Qdrant backend):*
+```python
+# ISO strings, Unix timestamps, and datetime objects all work
+m.search("recent memories", user_id="alice", filters={
+    "created_at": {"gte": "2024-07-01T00:00:00Z"}
 })
 ```
 
--- a/docs/platform/features/v2-memory-filters.mdx
+++ b/docs/platform/features/v2-memory-filters.mdx
@@ -202,6 +202,20 @@ Retrieve memories within specific date ranges using time operators.
     }
     ```
   </Accordion>
+
+  <Accordion title="Accepted datetime formats (OSS / Qdrant)">
+    ```python
+    # ISO 8601 string (recommended)
+    {"created_at": {"gte": "2024-01-01T00:00:00Z"}}
+
+    # Unix timestamp (float)
+    {"created_at": {"gte": 1704067200.0}}
+
+    # Python datetime object
+    from datetime import datetime, timezone
+    {"created_at": {"gte": datetime(2024, 1, 1, tzinfo=timezone.utc)}}
+    ```
+  </Accordion>
 </AccordionGroup>
 
 ### Multiple criteria


kartik-mem0:
Hey @Mukhsin0508, thanks for taking the time to work on this fix! 🙏

The underlying issue (#4591) has already been resolved, so we're closing this PR. We really appreciate your willingness to contribute — hope to see you around on future issues!

Mukhsin0508:
> Hey @Mukhsin0508, thanks for taking the time to work on this fix! 🙏
> 
> The underlying issue (#4591) has already been resolved, so we're closing this PR. We really appreciate your willingness to contribute — hope to see you around on future issues!

I'd appreciate a feedback on why my PR wasn't accepted even though it did the same solution even earlier than the other PR which was accepted with the same solution, yet created after my PRs! 

Anyways thanks - this is the 2nd time, the same biased phenomenon))

kartik-mem0:
hey @Mukhsin0508 i am sorry if you are feeling a biasedness from our side:

the reason why we did not preferred you pr:
- not handling all of the edges cases there were a some edge cases which you should have addressed check the pr which we have merged.
- uneccessary docs changes which we don't need for this soluion.
- also your solution does not have regex pattern check which is a useful regression pattern which will make sure that we are catching all of bugs.
- we usually usually prefer solution which does the foundation thing very well which your pr was not doing.

Mukhsin0508:
> hey @Mukhsin0508 i am sorry if you are feeling a biasedness from our side:
> 
> the reason why we did not preferred you pr:
> 
> * not handling all of the edges cases there were a some edge cases which you should have addressed check the pr which we have merged.
> * uneccessary docs changes which we don't need for this soluion.
> * also your solution does not have regex pattern check which is a useful regression pattern which will make sure that we are catching all of bugs.
> * we usually usually prefer solution which does the foundation thing very well which your pr was not doing.

Thanks a lot

PR #4593, opened 2026-03-28
fix: prevent hallucinated memories on empty messages payload

### LINKED ISSUE #4099: Hallucinated memories created on empty messages payload when run_id is present (gemini-3-flash-preview)

IgnazioDS:
## Summary
- Add input validation in `Memory.add()` (both sync and async) to return early when messages is empty or contains no meaningful content
- Prevents the LLM from generating hallucinated memories from empty input
- Guard applies regardless of whether `run_id` or other parameters are provided

## Problem
When `Memory.add()` receives an empty messages list with a valid `run_id`, it still sends a request to the LLM, which generates and persists ghost memories. See #4099.

## Test plan
- [ ] `Memory.add(messages=[], run_id="abc")` returns `{"results": []}`, no memories created
- [ ] `Memory.add(messages=[{"role": "user", "content": ""}], run_id="abc")` returns empty result
- [ ] `Memory.add(messages=[{"role": "user", "content": "Hello"}], run_id="abc")` still works normally
- [ ] Async `Memory.add()` with empty messages also returns early

Fixes #4099

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -429,6 +429,15 @@ class Memory(MemoryBase):
                 suggestion="Convert your input to a string, dictionary, or list of dictionaries."
             )
 
+        # Guard: return early when messages list is empty or contains no meaningful content.
+        # This prevents the LLM from hallucinating memories when called with an empty payload.
+        has_content = any(
+            isinstance(m, dict) and m.get("content") and str(m["content"]).strip()
+            for m in messages
+        )
+        if not messages or not has_content:
+            return {"results": []}
+
         if agent_id is not None and memory_type == MemoryType.PROCEDURAL.value:
             results = self._create_procedural_memory(messages, metadata=processed_metadata, prompt=prompt)
             return results
@@ -1517,6 +1526,15 @@ class AsyncMemory(MemoryBase):
                 details={"provided_type": type(messages).__name__, "valid_types": ["str", "dict", "list[dict]"]},
                 suggestion="Convert your input to a string, dictionary, or list of dictionaries."
             )
+
+        # Guard: return early when messages list is empty or contains no meaningful content.
+        # This prevents the LLM from hallucinating memories when called with an empty payload.
+        has_content = any(
+            isinstance(m, dict) and m.get("content") and str(m["content"]).strip()
+            for m in messages
+        )
+        if not messages or not has_content:
+            return {"results": []}
 
         if agent_id is not None and memory_type == MemoryType.PROCEDURAL.value:
             results = await self._create_procedural_memory(


kartik-mem0:
## Changes Requested

**Linked issue**: #4099 — validated, bug confirmed on current `main`

Thanks for tackling this — the intent is right, but there are a few issues that need addressing before this can merge.

### Must fix

1. **`mem0/memory/main.py:434-438` — Falsy content silently dropped.** `m.get("content")` is a truthiness check, so content values like `0`, `False`, or `None` cause the guard to treat the message as empty and return `{"results": []}`. Use `m.get("content", "")` with an explicit `.strip()` check instead of relying on truthiness.

2. **`mem0/memory/main.py:434-438` — `list[str]` input silently swallowed.** `isinstance(m, dict)` is `False` for string items, so `messages=["hello", "world"]` gets `has_content = False` and silently returns empty instead of being processed or raising an error.

3. **`mem0/memory/main.py:438` — Silent return breaks API contract.** Returning `{"results": []}` is inconsistent with `VALIDATION_003` three lines above, which raises `Mem0ValidationError`. Callers catching exceptions for bad input will never know their input was rejected — they'll just see "zero memories." This should raise `Mem0ValidationError` with a new `VALIDATION_004` error code, matching the established pattern.

4. **No tests.** The PR needs regression tests — at minimum covering empty list, empty content strings, and whitespace-only content. See the test plan in the PR description.

5. **Formatting.** `ruff format --check` flags the file for reformatting.

### Note

PR #4435 addresses the same issue with `Mem0ValidationError(VALIDATION_004)`, proper edge-case handling, and 63 lines of tests. Consider aligning with that approach, or we may move forward with #4435 instead.

### What's good

- Correctly identifies both sync and async paths
- The `any()` comprehension with `.strip()` is the right shape for the check — just needs the edge cases tightened up
- Clean, minimal diff

IgnazioDS:
Thanks for the thorough review. You're right on all of it, and #4435 already has the VALIDATION_004 code and the test coverage, so it's the more complete fix. I'll defer to that one and close this so it's not a duplicate in your queue. Appreciate you confirming the bug.

PR #4597, opened 2026-03-29
feat(vector_stores): support chained comparison operators in filters

### LINKED ISSUE #3914: Enhance filter capabilities to support chained comparisons
### 🚀 The feature

The current implementation of the `search` method in `mem0/memory/main.py` allows for filtering based on simple comparison operators like `gt`, `gte`, `lt`, and `lte`. For example, one can filter for a key being greater than a value, like `{"key": {"gt": 10}}`.
However, the filtering system does not currently support more complex, chained comparisons to define a range. For instance, it's not possible to filter for a key that is between two values, such as `1 < key < 10`. An intuitive but unsupported syntax for this would be something like `{"key": {"gt": 1, "lt": 10}}`. This limitation requires users to perform multiple queries and then manually intersect the results on the client side, which is inefficient.

I propose enhancing the filtering logic to support chained or compound comparisons within a single filter object. This would allow for more expressive and powerful queries.
Following the example above, the desired behavior would be to support a filter expression like:
 
`{"key": {"gt": 1, "lt": 10}}`

This would be parsed to create a query that finds all entries where `key` is greater than 1 AND less than 10.
This could be implemented by modifying the filter parsing logic in the `search` function to handle multiple comparison operators for the same key.

### Motivation, pitch

I need range search in my app. 
IMHO, range search is a very common use case. This feature would improve the usability and power of the search functionality, making it more convenient to query for data within specific ranges.

roshan-shaik-ml:
## Summary
- Adds range/operator filter support (`eq`, `ne`, `gt`, `gte`, `lt`, `lte`, `in`, `nin`) to 5 vector stores that previously only handled simple equality filters. This enables chained comparisons like `{"score": {"gte": 1, "lte": 10}}` across all major stores.
- **ChromaDB**: Fix operator overwrite in `convert_condition()` — wrap multiple operators per key in `$and`
- **PGVector**: Extract shared `_build_filter_conditions()` helper with SQL operator mapping and `::numeric` casting, used by both `search()` and `list()`
- **Pinecone**: Replace hardcoded `gte+lte` pair check with generic operator mapping
- **FAISS**: Add operator dict dispatch to `_apply_filters()` for Python-side post-search filtering
- **Milvus**: Add operator dict branch to `_create_filter()` using native expression syntax

Qdrant already supported this and is unchanged (used as reference implementation).

Fixes #3914

## Test plan
- [x] FAISS: 17 tests pass (6 new `_apply_filters` operator tests)
- [x] Pinecone: 13 tests pass (6 new `_create_filter` operator tests)
- [x] Milvus: 26 tests pass (7 new `_create_filter` operator tests)
- [x] ChromaDB: 20 tests pass (5 new `_generate_where_clause` operator tests)
- [x] PGVector: 57 tests pass (7 new `_build_filter_conditions` tests)
- [x] No regressions on existing tests (3 async failures are pre-existing on main)
- [x] Filter output formats verified against official docs (ChromaDB, Pinecone, Milvus, pgvector)

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -264,31 +264,18 @@ class ChromaDB(VectorStoreBase):
                 return None
             elif isinstance(value, dict):
                 # Handle comparison operators
-                chroma_condition = {}
+                op_map = {
+                    "eq": "$eq", "ne": "$ne", "gt": "$gt", "gte": "$gte",
+                    "lt": "$lt", "lte": "$lte", "in": "$in", "nin": "$nin",
+                    "contains": "$eq", "icontains": "$eq",
+                }
+                conditions = []
                 for op, val in value.items():
-                    if op == "eq":
-                        chroma_condition[key] = {"$eq": val}
-                    elif op == "ne":
-                        chroma_condition[key] = {"$ne": val}
-                    elif op == "gt":
-                        chroma_condition[key] = {"$gt": val}
-                    elif op == "gte":
-                        chroma_condition[key] = {"$gte": val}
-                    elif op == "lt":
-                        chroma_condition[key] = {"$lt": val}
-                    elif op == "lte":
-                        chroma_condition[key] = {"$lte": val}
-                    elif op == "in":
-                        chroma_condition[key] = {"$in": val}
-                    elif op == "nin":
-                        chroma_condition[key] = {"$nin": val}
-                    elif op in ["contains", "icontains"]:
-                        # ChromaDB doesn't support contains, fallback to equality
-                        chroma_condition[key] = {"$eq": val}
-                    else:
-                        # Unknown operator, treat as equality
-                        chroma_condition[key] = {"$eq": val}
-                return chroma_condition
+                    chroma_op = op_map.get(op, "$eq")
+                    conditions.append({key: {chroma_op: val}})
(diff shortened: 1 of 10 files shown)


wangjiawei-vegetable:
Hi @roshan-shaik-ml, thanks for this PR!

We're using Mem0 with Milvus as our vector store in a production e-commerce agent system. One of our key requirements is filtering memories by confidence score (e.g. {"score": {"gte": 9}}), which is currently impossible because the Milvus adapter only supports basic equality filters. This PR is exactly what we need.

I pulled the branch locally and ran the full Milvus test suite — all 26 tests passed (Python 3.12, pymilvus 2.6.12), including the 7 new operator tests:


tests/vector_stores/test_milvus.py — 26 passed in 3.09s

New tests (all PASSED):
  test_create_filter_range          — gte + lte chained
  test_create_filter_single_gt      — single gt
  test_create_filter_ne             — ne operator
  test_create_filter_in             — in operator
  test_create_filter_nin            — nin operator
  test_create_filter_mixed_equality_and_range — equality + range coexistence
  test_create_filter_equality_unchanged       — backward compatibility
The code changes to _create_filter() in milvus.py look clean and backward-compatible. Existing equality filters continue to work as before.

Would love to see this merged — it unblocks a real production use case for us. Happy to help with any additional testing if needed.

PR #4608, opened 2026-03-29
fix(vllm): forward response_format to OpenAI-compatible API

### LINKED ISSUE #4607: vllm provider drops response_format in generate_response, causing JSON parsing failures in memory extraction/update
### Component

Ollama / Local Models

### Summary

The `vllm` LLM provider accepts a `response_format` argument in `generate_response(...)`, but it does not forward that argument into the actual request params sent to the OpenAI-compatible client.

Because of that, `mem0` may assume the model is constrained to return JSON, while the backend model is actually only guided by prompt text. This can lead to invalid/non-JSON outputs and downstream parsing failures during memory extraction/update.

### Steps to Reproduce

I am using:
- mem0 with provider: "vllm"
- local OpenAI-compatible vLLM endpoint
- local embedding model
- local Qdrant path-based vector storage

A conversational correction such as:
- previous memory inferred: user prefers fast-paced sci-fi movies
- user later says: actually I prefer slow-paced healing movies; I still dislike horror, but prefer healing movies over sci-fi
can trigger the memory update stage to fail with:
**Invalid JSON response: Expecting value: line 1 column 1 (char 0)**

This appears consistent with response_format not being enforced at the provider layer.

### Expected Behavior

If generate_response(..., response_format=...) is called, the vllm provider should forward response_format into the request params, similar to other providers.

### Actual Behavior

**Invalid JSON response: Expecting value: line 1 column 1 (char 0)**

### Environment

- mem0 version:1.0.7
- Python/Node version:python=3.10
- OS:ubuntu24.04

rakheesingh:
## Summary
- The `vllm` LLM provider accepted `response_format` as a parameter in `generate_response()` but never forwarded it into the actual API request params
- This caused JSON parsing failures (`Invalid JSON response: Expecting value`) during memory extraction/update because the model returned free-form text instead of constrained JSON
- Fix: add `response_format` to params, matching the pattern used by `openai.py`, `groq.py`, `deepseek.py`, and all other OpenAI-compatible providers

Closes #4607

## Changes
- **`mem0/llms/vllm.py`** — Added 2 lines to forward `response_format` into request params
- **`tests/llms/test_vllm.py`** — Added 2 new tests:
  - `test_generate_response_with_response_format` — verifies `response_format` is passed to the API
  - `test_generate_response_without_response_format` — verifies it's omitted when not provided

## Test plan
- [x] All 4 vllm tests pass (2 existing + 2 new)
- [x] Fix is identical to the pattern in `openai.py` line 134-135

### DIFF
--- a/mem0/llms/vllm.py
+++ b/mem0/llms/vllm.py
@@ -99,6 +99,8 @@ class VllmLLM(LLMBase):
             }
         )
 
+        if response_format:
+            params["response_format"] = response_format
         if tools:
             params["tools"] = tools
             params["tool_choice"] = tool_choice
--- a/tests/llms/test_vllm.py
+++ b/tests/llms/test_vllm.py
@@ -88,6 +88,50 @@ def test_generate_response_with_tools(mock_vllm_client):
 
 
 
+def test_generate_response_with_response_format(mock_vllm_client):
+    config = BaseLlmConfig(model="Qwen/Qwen2.5-32B-Instruct", temperature=0.7, max_tokens=100, top_p=1.0)
+    llm = VllmLLM(config)
+    messages = [
+        {"role": "system", "content": "You are a memory extraction assistant."},
+        {"role": "user", "content": "I like hiking on weekends."},
+    ]
+
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=Mock(content='{"facts": ["User likes hiking on weekends"]}'))]
+    mock_vllm_client.chat.completions.create.return_value = mock_response
+
+    response = llm.generate_response(messages, response_format={"type": "json_object"})
+
+    mock_vllm_client.chat.completions.create.assert_called_once_with(
+        model="Qwen/Qwen2.5-32B-Instruct",
+        messages=messages,
+        temperature=0.7,
+        max_tokens=100,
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for the clean fix @rakheesingh — this looks good! The bug is clearly real, the fix follows the exact pattern used across all other providers, and the tests cover it well. Approving and merging this.

One thing we noticed during review: `deepseek.py` has the exact same bug — `response_format` is accepted as a parameter but never forwarded into the API request params. Would you be interested in opening a follow-up PR to fix that as well? Totally understand if not, but since you're already familiar with the pattern, it'd be a quick win! 🙂

rakheesingh:
> Thanks for the clean fix @rakheesingh — this looks good! The bug is clearly real, the fix follows the exact pattern used across all other providers, and the tests cover it well. Approving and merging this.
> 
> One thing we noticed during review: `deepseek.py` has the exact same bug — `response_format` is accepted as a parameter but never forwarded into the API request params. Would you be interested in opening a follow-up PR to fix that as well? Totally understand if not, but since you're already familiar with the pattern, it'd be a quick win! 🙂

I am gonna raise it right now

PR #4616, opened 2026-03-30
fix: use configured LLM for memory categorization instead of hardcoded OpenAI

### LINKED ISSUE #4611: [Enhancement] Categorization should use configured LLM provider instead of hardcoded OpenAI
## Summary

The memory categorization feature in OpenMemory (`openmemory/api/app/utils/categorization.py`) hardcodes the OpenAI client regardless of what LLM provider is configured. This silently breaks categorization for all non-OpenAI users.

## Current Behavior

`categorization.py` unconditionally initializes `OpenAI()`:

```python
from openai import OpenAI
openai_client = OpenAI()
```

This is used inside `get_categories_for_memory()` and `get_categories_for_memories()` for every memory insertion. If the user has configured Gemini, Anthropic, Ollama, or any other provider, these calls either fail silently or raise errors — and memories end up with no categories.

## Impact

Any user who has configured a non-OpenAI LLM provider (via the `configs` table or `.env`) will:
- Get no categories assigned to their memories
- See silent failures or misleading log errors referencing OpenAI
- Have no indication in the UI that categorization is not working

This affects providers including but not limited to: Gemini, Anthropic Claude, Ollama, Azure OpenAI, Mistral, etc.

## Proposed Solution

Categorization should use the configured LLM provider from `get_memory_client()` config, the same way memories themselves are created. Two approaches:

1. **Use mem0's built-in LLM abstraction** — mem0 already has a provider-agnostic `LLM` layer. Categorization could use `mem0.llm` directly instead of raw OpenAI calls.
2. **Read config and instantiate the appropriate client** — Check the configured provider and instantiate the correct client dynamically.

Either way, the structured output parsing (currently using `openai_client.beta.chat.completions.parse`) would need to be adapted per provider, since not all providers support the OpenAI structured output API.

## Related Improvement: Batch Categorization

Currently, the `after_insert` SQLAlchemy event fires one LLM call per memory. For a file upload that creates 20 memories, this triggers 20 sequential API calls — adding significant latency and cost.

A batch endpoint that categorizes all memories in a single LLM call (`get_categories_for_memories(memories: List[str])`) would be far more efficient. This is especially relevant for the file upload feature.

## Steps to Reproduce

1. Configure OpenMemory with a Gemini (or any non-OpenAI) LLM provider
2. Add any memory
3. Observe: the memory has no categories assigned, and logs show an OpenAI authentication error

## Environment

- OpenMemory API
- Any non-OpenAI LLM provider configured

## Additional Context

This is a good first issue for anyone looking to contribute to OpenMemory — the fix scope is well-defined (one file: `categorization.py`) and the expected behavior is clear.

rakheesingh:
Closes #4611

## Summary

Memory categorization in OpenMemory was hardcoded to use `OpenAI()` client with `gpt-4o-mini`, ignoring the user's configured LLM provider. This caused silent failures for users running Ollama, Anthropic, Groq, or any non-OpenAI provider.

- Replaced the hardcoded `OpenAI()` client with the already-configured LLM from `get_memory_client().llm`
- Uses `generate_response()` with `response_format={"type": "json_object"}` which is supported by all LLM providers
- Parses the JSON response with Pydantic validation, keeping the same `MemoryCategories` schema

## Changes

**`openmemory/api/app/utils/categorization.py`**
- Removed `from openai import OpenAI`, `load_dotenv()`, and module-level `openai_client = OpenAI()`
- Added `_get_llm()` helper that retrieves the LLM instance from the existing memory client (respects env vars, DB config overrides, and all 17 supported providers via `LlmFactory`)
- Changed `get_categories_for_memory()` to use `llm.generate_response()` instead of `openai_client.beta.chat.completions.parse()`

**`openmemory/api/app/models.py`**
- Updated docstring to reflect the fix

## Test plan

- [x] Set `LLM_PROVIDER=ollama` and insert a memory — categorization should use Ollama instead of failing with an OpenAI auth error
- [x] Set `LLM_PROVIDER=openai` — existing OpenAI users should see no behavior change
- [x] Verify categories are correctly stored in the database after insert/update

### DIFF
--- a/openmemory/api/app/utils/categorization.py
+++ b/openmemory/api/app/utils/categorization.py
@@ -29,7 +29,7 @@ def _get_llm():
 def get_categories_for_memories(memories: List[str]) -> List[List[str]]:
     """Categorize multiple memories in a single LLM call.
 
-    Returns a list of category lists, one per memory in the same order.
+    Returns a list of category lists in this data type: MemoryCategories, one per memory in the same order.
     Falls back to empty categories on failure so callers are never blocked.
     """
     if not memories:


kartik-mem0:
Thanks for the contribution and the thorough work here!

The `categorization.py` file lives under the `openmemory/` directory, which is being phased out from the monorepo. We've closed the related issue (#4611) for the same reason — we won't be investing in further changes to this component.

Appreciate the effort and hope to see you contribute to other parts of mem0!

PR #4632, opened 2026-03-31
fix: pass dimensions parameter to OpenAI embeddings API

### LINKED ISSUE #4615: Bug: OpenAIEmbedder ignores embeddingDims config, doesn't pass dimensions to API
## Description

The `OpenAIEmbedder` class stores `embeddingDims` from config but never passes it to the OpenAI `embeddings.create()` call. This causes dimension mismatch errors when using embedding models with configurable dimensions (e.g., `text-embedding-v4` on DashScope/Aliyun).

## Affected Code

**File:** [`mem0-ts/src/oss/src/embeddings/openai.ts`](https://github.com/mem0ai/mem0/blob/main/mem0-ts/src/oss/src/embeddings/openai.ts)

```typescript
export class OpenAIEmbedder implements Embedder {
  private openai: OpenAI;
  private model: string;
  private embeddingDims?: number;  // ✅ Stored

  constructor(config: EmbeddingConfig) {
    this.openai = new OpenAI({
      apiKey: config.apiKey,
      baseURL: config.baseURL || config.url,
    });
    this.model = config.model || "text-embedding-3-small";
    this.embeddingDims = config.embeddingDims || 1536;  // ✅ Read from config
  }

    const response = await this.openai.embeddings.create({
      model: this.model,
      input: text,
      // ❌ Missing: dimensions: this.embeddingDims
    });
    return response.data[0].embedding;
  }

  async embedBatch(texts: string[]): Promise<number[][]> {
    const response = await this.openai.embeddings.create({
      model: this.model,
      input: texts,
      // ❌ Missing: dimensions: this.embeddingDims
    });
    return response.data.map((item) => item.embedding);
  }
}
```

## Expected Behavior

When user configures `embeddingDims: 1536`, the embedder should pass `dimensions: 1536` to the API call, so models like `text-embedding-v4` return 1536-dimensional vectors.

## Actual Behavior

API returns model's default dimension (1024 for `text-embedding-v4`), causing `Query dimension mismatch` errors:

```
Memory store failed: Error: Query dimension mismatch. Expected 1536, got 1024
```

## Reproduction

Using OpenClaw plugin with DashScope (Aliyun):

```json
{
  "oss": {
    "embedder": {
      "provider": "openai",
      "config": {
        "model": "text-embedding-v4",
        "embeddingDims": 1536,
        "apiKey": "${DASHSCOPE_API_KEY}",
        "url": "https://dashscope.aliyuncs.com/compatible-mode/v1"
      }
    }
  }
}
```

## Fix

Add `dimensions` parameter to both `embed()` and `embedBatch()`:

```typescript
  const response = await this.openai.embeddings.create({
    model: this.model,
    input: text,
    dimensions: this.embeddingDims,  // Add this
  });
  return response.data[0].embedding;
}

async embedBatch(texts: string[]): Promise<number[][]> {
  const response = await this.openai.embeddings.create({
    model: this.model,
    input: texts,
    dimensions: this.embeddingDims,  // Add this
  });
  return response.data.map((item) => item.embedding);
}
```

## Environment

- mem0ai: 2.4.4
- Node.js: v25.8.1
- Platform: macOS (Darwin 24.6.0 arm64)
- Provider: DashScope (Aliyun) OpenAI-compatible API

wobushixiaoj:
## Summary

Fixes #4615

The `OpenAIEmbedder` class stores `embeddingDims` from config but never passes it to the OpenAI `embeddings.create()` call. This causes dimension mismatch errors when using embedding models with configurable dimensions (e.g., `text-embedding-v4` on DashScope/Aliyun).

## Changes

- Added `dimensions: this.embeddingDims` parameter to both `embed()` and `embedBatch()` methods

## Test plan

- [x] Verified the fix resolves dimension mismatch with DashScope `text-embedding-v4` model
- [x] Existing tests should continue to pass (dimensions is optional in OpenAI API)

## Before

```typescript
  const response = await this.openai.embeddings.create({
    model: this.model,
    input: text,
    // ❌ Missing: dimensions
  });
  return response.data[0].embedding;
}
```

## After

```typescript
  const response = await this.openai.embeddings.create({
    model: this.model,
    input: text,
    dimensions: this.embeddingDims,  // ✅ Now passes configured dimension
  });
  return response.data[0].embedding;
}
```

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/azure.ts
+++ b/mem0-ts/src/oss/src/embeddings/azure.ts
@@ -5,7 +5,7 @@ import { EmbeddingConfig } from "../types";
 export class AzureOpenAIEmbedder implements Embedder {
   private client: AzureOpenAI;
   private model: string;
-  private embeddingDims?: number;
+  private embeddingDims: number | undefined;
 
   constructor(config: EmbeddingConfig) {
     if (!config.apiKey || !config.modelProperties?.endpoint) {
@@ -20,13 +20,16 @@ export class AzureOpenAIEmbedder implements Embedder {
       ...rest,
     });
     this.model = config.model || "text-embedding-3-small";
-    this.embeddingDims = config.embeddingDims || 1536;
+    this.embeddingDims = config.embeddingDims;
   }
 
   async embed(text: string): Promise<number[]> {
     const response = await this.client.embeddings.create({
       model: this.model,
       input: text,
+      ...(this.embeddingDims !== undefined && {
+        dimensions: this.embeddingDims,
+      }),
     });
     return response.data[0].embedding;
   }
@@ -35,6 +38,9 @@ export class AzureOpenAIEmbedder implements Embedder {
     const response = await this.client.embeddings.create({
       model: this.model,
       input: texts,
+      ...(this.embeddingDims !== undefined && {
+        dimensions: this.embeddingDims,
+      }),
     });
     return response.data.map((item) => item.embedding);
   }
--- a/mem0-ts/src/oss/src/embeddings/google.ts
+++ b/mem0-ts/src/oss/src/embeddings/google.ts
@@ -5,21 +5,23 @@ import { EmbeddingConfig } from "../types";
 export class GoogleEmbedder implements Embedder {
   private google: GoogleGenAI;
   private model: string;
-  private embeddingDims?: number;
+  private embeddingDims: number | undefined;
 
   constructor(config: EmbeddingConfig) {
     this.google = new GoogleGenAI({
       apiKey: config.apiKey || process.env.GOOGLE_API_KEY,
     });
     this.model = config.model || "gemini-embedding-001";
-    this.embeddingDims = config.embeddingDims || 1536;
+    this.embeddingDims = config.embeddingDims;
   }
 
   async embed(text: string): Promise<number[]> {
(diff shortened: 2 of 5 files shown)


whysosaket:
## Review feedback

The fix is correct, but the `passDimensionsToApi: boolean` field is unnecessary — that information is already encoded in whether `embeddingDims` is `undefined`. Simpler approach:

```ts
private embeddingDims: number | undefined; // undefined = "not configured"

constructor(config: EmbeddingConfig) {
  this.model = config.model || "text-embedding-3-small";
  this.embeddingDims = config.embeddingDims; // no default
}

  const response = await this.openai.embeddings.create({
    model: this.model,
    input: text,
    ...(this.embeddingDims !== undefined && { dimensions: this.embeddingDims }),
  });
  return response.data[0].embedding;
}
```

Same semantics, one fewer field, no cognitive overhead of tracking what the boolean means.

---

Also worth noting: `azure.ts` and `google.ts` have the same root bug — `embeddingDims` is stored but never forwarded to the API call. This PR only fixes `openai.ts`. Worth extending the fix to both:

- `azure.ts` — `this.embeddingDims = config.embeddingDims || 1536` is stored but unused in API calls
- `google.ts` — actually does pass `outputDimensionality: this.embeddingDims`, but the `|| 1536` default means it always sends a dimension even for models that don't support the parameter

PR #4641, opened 2026-04-01
feat(openclaw): add MMR recall diversification to reduce topic redundancy

stroupaloop:
## Summary

Adds Maximal Marginal Relevance (MMR) re-ranking to the openclaw-mem0 plugin's auto-recall hook. This prevents clusters of similar memories from monopolizing recall slots as memory stores grow.

## Problem

When memory stores exceed ~1K records, vector search often returns multiple results from the same topic or conversation that crowd out relevant memories from other domains. For example, a deep conversation about infrastructure configuration can produce 10+ similar memories that fill all recall slots, preventing unrelated but relevant operational context from surfacing.

This is a well-known problem in RAG systems, commonly called "topic monopoly" or "result redundancy." LangChain, Elasticsearch, Azure AI Search, and OpenSearch all implement MMR as the standard solution.

## Solution

After fetching and filtering candidates in the `before_agent_start` hook, apply greedy MMR selection using Jaccard word-overlap similarity as a lightweight proxy for embedding distance (zero additional API calls).

Results are selected to maximize both relevance to the query AND dissimilarity to already-selected results.

## New config options

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `diversifyRecall` | `boolean` | `true` | Enable MMR re-ranking of auto-recall results |
| `diversityLambda` | `number` | `0.6` | Balance between relevance (1.0) and diversity (0.0) |

Both are optional and backward-compatible. Existing configurations work without changes (diversification is enabled by default).

## Testing

- 6 new test cases covering: empty input, single result, top-result preservation, topic monopoly penalty, lambda tuning (pure relevance vs pure diversity), passthrough when count ≤ targetK
- All 84 tests pass (74 existing + 10 new)
- Validated on a production memory store with 8,000+ records: average recall precision improved from ~20% to ~90%

## Changes

- `openclaw/index.ts` — `diversifyResults()` function + integration into auto-recall hook
- `openclaw/config.ts` — New config fields + parsing
- `openclaw/types.ts` — Type definitions
- `openclaw/openclaw.plugin.json` — UI hints + config schema
- `openclaw/index.test.ts` — 6 new test cases
- `openclaw/README.md` — Documentation for the feature + config options

### DIFF
--- a/openclaw/README.md
+++ b/openclaw/README.md
@@ -72,6 +72,21 @@ Lifecycle hooks (`before_agent_start`, `agent_end`) use `ctx.sessionKey` directl
 
 Tools still read from a best-effort `currentSessionId` variable (since tools don't receive `ctx`), but hooks — where the critical recall and capture logic runs — are fully concurrency-safe.
 
+### Recall diversification
+
+As memory stores grow (1K+ records), vector search can return clusters of highly similar memories from the same topic or conversation, crowding out relevant results from other domains. This is sometimes called "topic monopoly."
+
+The plugin addresses this with **Maximal Marginal Relevance (MMR)** re-ranking during auto-recall. When `diversifyRecall` is enabled (the default), results are re-ranked to maximize both relevance to the query AND dissimilarity to already-selected results. This ensures recall slots are filled with diverse, useful context rather than redundant variations of the same fact.
+
+The `diversityLambda` parameter controls the trade-off:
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for the contribution\! Closing this for now as it's not relevant at this time. We may revisit this in the future.

PR #4658, opened 2026-04-01
feat(reranker): add concurrent LLM scoring via max_workers config

### LINKED ISSUE #4303: Improve LLM reranker performance with configurable concurrency
### 🚀 The feature

This PR introduces optional concurrency inside LLMReranker.rerank() to reduce end-to-end search latency when scoring large candidate sets. The change is fully backwards compatible — sequential behavior is preserved by default.

**Problem:**

LLMReranker.rerank() currently scores documents sequentially — one synchronous LLM call per document, run one after another in a for loop. Since each (query, document) scoring call is independent, this creates unnecessary latency that grows linearly with candidate set size.

For LLM providers with non-trivial per-call latency (200ms–1s+), scoring 20–100 candidates becomes a significant bottleneck in the search path.

**Affected paths:**

- Sync: Memory.search() calls self.reranker.rerank() directly
- Async: AsyncMemory.search() wraps it in await asyncio.to_thread(self.reranker.rerank, ...)
In both cases, the internal sequential loop blocks the entire reranking step.

**Proposed Change:**

Introduce an optional max_concurrency field in LLMRerankerConfig that controls a ThreadPoolExecutor inside rerank():

1. New config field in LLMRerankerConfig, e.g.:
- max_concurrency: Optional[int] = None
- None or <= 1 → preserve current behavior (fully sequential).
-  (> 1) → use a ThreadPoolExecutor(max_workers=max_concurrency).

2. Implementation sketch:
- Keep rerank as a synchronous function.
- Inside rerank:
   - Build (index, doc) pairs.
   - In a ThreadPoolExecutor, submit one job per doc:
       - job = build prompt, call llm.generate_response, parse score via _extract_score.
   - Collect results via as_completed, map back to indices.
   - On per-doc failure, log and assign a neutral score (current behavior is 0.5).
   - Attach rerank_score to each doc, sort descending, apply top_k as today.

3. Async compatibility:
- The async Memory.search flow can continue to call await asyncio.to_thread(self.reranker.rerank, ...).
- From the async caller’s perspective it’s still a single blocking call, but internally the reranker can leverage a small thread pool for LLM calls.

**Benefits:**

- Reranking latency reduced from O(n * llm_latency) to approximately O(llm_latency) at sufficient concurrency
- Backwards compatible: same public API, same output shape, same default behavior
- Safer than batched single-prompt strategies — no changes to prompt contract or JSON parsing behavior across providers
- Per-document failure handling preserved: failed calls fall back to neutral score 0.5 with a warning log

**Files Changed:**

- mem0/configs/rerankers/llm.py — add max_concurrency field
- mem0/reranker/llm_reranker.py — add parallel scoring path via ThreadPoolExecutor

**If this approach sounds reasonable, I’d be happy to open a PR implementing it along with tests (using a mocked llm.generate_response) to verify behavior for:**

- sequential vs concurrent modes,
- correct mapping of scores to docs,
- and error handling per document.

### Motivation, pitch

I’m using Mem0 in a setting where search calls often return a moderate to large number of candidate memories (e.g. 20–100) that are passed through the LLM-based reranker. Right now, LLMReranker.rerank calls the LLM once per document sequentially, so reranking latency grows roughly linearly with the number of candidates. This becomes a noticeable bottleneck in end‑to‑end response time, especially when the upstream LLM provider has non‑trivial latency per call or when the system needs to handle many concurrent users.

Because each reranking operation is logically independent (it just scores (query, doc_text) pairs), this seems like a good place to introduce safe concurrency with a configurable limit, improving performance and scalability without changing the public API or overall behavior of Mem0’s search pipeline.

Prithvi1994:
## Summary

Closes #4303

The `LLMReranker` scored documents sequentially — one LLM round-trip per document. For N documents this meant N × latency, making reranking a bottleneck in high-recall pipelines.

This PR adds a `max_workers` config parameter that enables concurrent scoring via `ThreadPoolExecutor`, reducing latency proportionally.

### `mem0/configs/rerankers/llm.py`
- New `max_workers: int = 1` field. Defaults to `1` — **fully backward compatible**, existing behaviour is unchanged.

### `mem0/reranker/llm_reranker.py`
- Extracted `_score_document()` helper that scores one document and returns a copy with `rerank_score` added — used by both paths.
- `rerank()`: when `max_workers > 1`, submits all documents to a `ThreadPoolExecutor` concurrently. Workers are capped at `min(max_workers, len(documents))` to avoid excess threads.
- Per-worker failures fall back to score `0.5` — a single flaky LLM call never aborts the batch.

### `tests/rerankers/test_llm_reranker_concurrent.py` (new)
Full test coverage for the concurrent path:
- Config field default and custom values
- Sequential path still works as before
- Concurrent path: all docs scored, sorted descending, `top_k` applied
- `max_workers` capped at `len(documents)`
- Per-worker failure falls back to `0.5` gracefully
- Original documents not mutated

## Usage

```python
config = {
    "reranker": {
        "provider": "llm",
        "config": {
            "provider": "openai",
            "model": "gpt-4o-mini",
            "max_workers": 8,  # score 8 docs in parallel
        }
    }
}
results = m.search("query", user_id="u1")  # ~8x faster reranking
```

## Performance

With `max_workers=N` and N documents, wall-clock reranking time drops from `N × llm_latency` to approximately `llm_latency` (bounded by the slowest single call). No change for `max_workers=1`.

### DIFF
--- a/mem0/configs/rerankers/llm.py
+++ b/mem0/configs/rerankers/llm.py
@@ -43,6 +43,14 @@ class LLMRerankerConfig(BaseRerankerConfig):
         default=100,
         description="Maximum tokens for LLM response"
     )
+    max_workers: int = Field(
+        default=1,
+        description=(
+            "Number of parallel workers for concurrent LLM scoring. "
+            "Set > 1 to score multiple documents simultaneously and reduce "
+            "latency proportionally. Defaults to 1 (sequential)."
+        ),
+    )
     scoring_prompt: Optional[str] = Field(
         default=None,
         description="Custom prompt template for scoring documents"
--- a/mem0/reranker/llm_reranker.py
+++ b/mem0/reranker/llm_reranker.py
@@ -1,4 +1,5 @@
 import re
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any, Dict, List, Union
 
 from mem0.configs.rerankers.base import BaseRerankerConfig
@@ -60,7 +61,7 @@ class LLMReranker(BaseReranker):
 
         # Default scoring prompt
         self.scoring_prompt = getattr(self.config, 'scoring_prompt', None) or self._get_default_prompt()
-        
+
     def _get_default_prompt(self) -> str:
         """Get the default scoring prompt template."""
         return """You are a relevance scoring assistant. Given a query and a document, you need to score how relevant the document is to the query.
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Hey @Prithvi1994 — thanks for the clean implementation and the detailed PR description. We're closing the parent issue #4303 as we've decided the added complexity isn't really justified for this code path — the LLM reranker sees limited production use at the scale where sequential scoring becomes a bottleneck.

Appreciate you taking the time to contribute, looking forward to future PRs!

PR #4661, opened 2026-04-01
fix: extract data string from update_memory request body

### LINKED ISSUE #3933: update_memory logic is incorrect
### 🐛 Describe the bug

in OSS， the update mehod is incorrect, as the below:
`

      @app.put("/memories/{memory_id}", summary="Update a memory")
      def update_memory(memory_id: str, updated_memory: Dict[str, Any]):
          """Update an existing memory with new content.
          
          Args:
              memory_id (str): ID of the memory to update
              updated_memory (str): New content to update the memory with
              
          Returns:
              dict: Success message indicating the memory was updated
          """
          try:
              return MEMORY_INSTANCE.update(memory_id=memory_id, data=updated_memory)
          except Exception as e:
              logging.exception("Error in update_memory:")
              raise HTTPException(status_code=500, detail=str(e))

`

what we could see is the param **updated_memory** which type is **dict**

but in **MEMORY_INSTANCE.update(memory_id=memory_id, data=updated_memory)** logic it invoke the  following method:
```

        def update(self, memory_id, data):
            """
            Update a memory by ID.
    
            Args:
                memory_id (str): ID of the memory to update.
                data (str): New content to update the memory with.
    
            Returns:
                dict: Success message indicating the memory was updated.
    
            Example:
                >>> m.update(memory_id="mem_123", data="Likes to play tennis on weekends")
                {'message': 'Memory updated successfully!'}
            """
            capture_event("mem0.update", self, {"memory_id": memory_id, "sync_type": "sync"})
    
            existing_embeddings = {data: self.embedding_model.embed(data, "update")}
    
            self._update_memory(memory_id, data, existing_embeddings)
            return {"message": "Memory updated successfully!"}


```

but in **self.embedding_model.embed(data, "update")** there is a replace call in openai.py as the below:
```

         def embed(self, text, memory_action: Optional[Literal["add", "search", "update"]] = None):
          """
          Get the embedding for the given text using OpenAI.
  
          Args:
              text (str): The text to embed.
              memory_action (optional): The type of embedding to use. Must be one of "add", "search", or "update". Defaults to None.
          Returns:
              list: The embedding vector.
          """
          text = text.replace("\n", " ")
          return (
              .data[0]
              .embedding
          )


```

the **text.replace("\n", " ")** "text" is dict type, so it could not invoke replace method

so I think this is a bug

Krishnachaitanyakc:
## Summary
- The `PUT /memories/{id}` endpoint received the request body as a dict and passed it directly to `Memory.update()`, which expects a string for the `data` parameter.
- This caused `"'dict' object has no attribute 'replace'"` errors when the embedding model tried to process the dict.
- Now extracts the `data` string field from the request body dict and optionally forwards the `metadata` field. Returns 400 if `data` is missing or not a string.

Closes #3933

## Test plan
- [x] Added 4 new tests: valid update, update with metadata, missing data field (400), non-string data (400)
- [x] All 117 server tests pass (including auth tests)

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -201,16 +201,25 @@ def search_memories(search_req: SearchRequest, _api_key: Optional[str] = Depends
 @app.put("/memories/{memory_id}", summary="Update a memory")
 def update_memory(memory_id: str, updated_memory: Dict[str, Any], _api_key: Optional[str] = Depends(verify_api_key)):
     """Update an existing memory with new content.
-    
+
     Args:
         memory_id (str): ID of the memory to update
-        updated_memory (str): New content to update the memory with
-        
+        updated_memory (dict): Request body containing 'data' (str) with the new memory content,
+            and optionally 'metadata' (dict) for additional metadata updates.
+
     Returns:
         dict: Success message indicating the memory was updated
     """
     try:
-        return MEMORY_INSTANCE.update(memory_id=memory_id, data=updated_memory)
+        data = updated_memory.get("data")
+        if data is None:
+            raise HTTPException(status_code=400, detail="Request body must contain a 'data' field with the new memory content")
+        if not isinstance(data, str):
+            raise HTTPException(status_code=400, detail="'data' field must be a string")
+        metadata = updated_memory.get("metadata")
+        return MEMORY_INSTANCE.update(memory_id=memory_id, data=data, metadata=metadata)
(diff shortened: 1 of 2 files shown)


kartik-mem0:
closing this pr in favor of this pr #4595

PR #4662, opened 2026-04-01
fix: add robust JSON repair for malformed LLM output

### LINKED ISSUE #3918: Unterminated JSON causes new_retrieved_facts to fail

Krishnachaitanyakc:
## Summary
- LLMs (especially Gemini models) sometimes produce malformed JSON by leaking reasoning tokens, adding trailing commas, or truncating output before closing brackets.
- Enhanced `extract_json()` with:
  - Stripping `<think>...</think>` reasoning tags before extraction
  - Removing trailing commas before `}` and `]` (common LLM error)
  - Auto-closing unterminated JSON by counting unmatched brackets/braces

Closes #3918

## Test plan
- [x] Added 8 new tests for: trailing commas, unterminated JSON, think tags, Gemini-style reasoning leaks
- [x] All 29 JSON parsing tests pass (including existing ones)
- [x] Verified the full fallback chain still works correctly

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -127,8 +127,16 @@ def extract_json(text):
     Extracts JSON content from a string, removing enclosing triple backticks and optional 'json' tag if present.
     If no code block is found, attempts to locate JSON by finding the first '{' and last '}'.
     If that also fails, returns the text as-is.
+
+    Also applies best-effort repairs for common LLM JSON malformations:
+    - Strips <think>...</think> reasoning tags
+    - Removes trailing commas before } and ]
+    - Attempts to close unterminated JSON by appending missing brackets/braces
     """
     text = text.strip()
+    # Strip <think> tags that some reasoning models emit
+    text = re.sub(r"<think>.*?</think>", "", text, flags=re.DOTALL).strip()
+
     match = re.search(r"```(?:json)?\s*(.*?)\s*```", text, re.DOTALL)
     if match:
         json_str = match.group(1)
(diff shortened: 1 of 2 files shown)


Krishnachaitanyakc:
@kartik-mem0 can you please review these changes?

Krishnachaitanyakc:
@kartik-mem0 can you please review these changes?

PR #4677, opened 2026-04-02
fix(qdrant): use DatetimeRange for ISO datetime string range filters

### LINKED ISSUE #4591: bug: Qdrant adapter uses Range instead of DatetimeRange for datetime fields — silently returns 0 results

Vitalcheffe:
## What
Fixes Qdrant vector store adapter to use `DatetimeRange` instead of numeric `Range` when range filter values are datetime strings.

## Why
`_build_field_condition` always used `Range` (numeric) for range filters, even when values are ISO datetime strings. This caused datetime range queries to silently return 0 results — including the exact examples from the [official docs](https://docs.mem0.ai/open-source/features/metadata-filtering) (time-based filtering).

See: #4591

## How
In `_build_field_condition`, the range branch now inspects the first value:
- **String** → `DatetimeRange` via `FieldCondition(datetime_range=...)`
- **Number** → `Range` via `FieldCondition(range=...)` (unchanged)

## Testing
- `test_datetime_range_uses_datetime_range_model` — ISO strings produce DatetimeRange
- `test_numeric_range_still_uses_range` — numbers still use Range
- `test_create_filter_datetime_range` — end-to-end filter with mixed datetime + equality
- `ruff check` passes

## Checklist
- [x] Tests added
- [x] Linter passes
- [x] No breaking change for numeric range filters

Closes #4591

### DIFF
--- a/mem0/vector_stores/qdrant.py
+++ b/mem0/vector_stores/qdrant.py
@@ -3,6 +3,7 @@ from typing import Optional
 
 from qdrant_client import QdrantClient
 from qdrant_client.models import (
+    DatetimeRange,
     Distance,
     FieldCondition,
     Filter,
@@ -177,6 +178,10 @@ class Qdrant(VectorStoreBase):
                     f"Use AND to combine them as separate conditions."
                 )
             range_kwargs = {op: value[op] for op in range_ops if op in value}
+            sample_value = next(iter(range_kwargs.values()))
+            if isinstance(sample_value, str):
+                # String values are datetime strings (ISO format) — use DatetimeRange
+                return FieldCondition(key=key, datetime_range=DatetimeRange(**range_kwargs))
             return FieldCondition(key=key, range=Range(**range_kwargs))
         elif "eq" in value:
             return FieldCondition(key=key, match=MatchValue(value=value["eq"]))
--- a/tests/vector_stores/test_qdrant.py
+++ b/tests/vector_stores/test_qdrant.py
@@ -6,6 +6,7 @@ from unittest.mock import MagicMock, patch
 
 from qdrant_client import QdrantClient
 from qdrant_client.models import (
+    DatetimeRange,
     Distance,
     FieldCondition,
     Filter,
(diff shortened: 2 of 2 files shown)


Mukhsin0508:
Bro are you blind?

PR #4688, opened 2026-04-03
fix(deepseek): forward response_format to OpenAI-compatible API

### LINKED ISSUE #4607: vllm provider drops response_format in generate_response, causing JSON parsing failures in memory extraction/update

rakheesingh:
The deepseek provider accepted response_format as a parameter but never passed it into the request params, causing JSON parsing failures during memory extraction and update when the model returned free-form text instead of constrained JSON output.

This is the same bug that was fixed in vllm.py (#4608), as noted by the maintainer during review.

Closes #4607

Made-with: Cursor

## Linked Issue

Closes #

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have performed a self-review of my code

### DIFF
--- a/mem0/llms/deepseek.py
+++ b/mem0/llms/deepseek.py
@@ -99,6 +99,8 @@ class DeepSeekLLM(LLMBase):
             }
         )
 
+        if response_format:
+            params["response_format"] = response_format
         if tools:
             params["tools"] = tools
             params["tool_choice"] = tool_choice
--- a/tests/llms/test_deepseek.py
+++ b/tests/llms/test_deepseek.py
@@ -113,3 +113,47 @@ def test_generate_response_with_tools(mock_deepseek_client):
     assert len(response["tool_calls"]) == 1
     assert response["tool_calls"][0]["name"] == "add_memory"
     assert response["tool_calls"][0]["arguments"] == {"data": "Today is a sunny day."}
+
+
+def test_generate_response_with_response_format(mock_deepseek_client):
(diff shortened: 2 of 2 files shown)


rakheesingh:
hey @kartik-mem0, here is PR for the deepseek

PR #4698, opened 2026-04-04
fix(bedrock): fix set literal typo in AI21 response parsing default

lawrence3699:
Was setting up mem0 with AI21 Jamba models on AWS Bedrock and noticed every response from the InvokeModel path silently returns `"Error parsing response"` instead of the actual model output.

Traced it to line 412 in `mem0/llms/aws_bedrock.py` — the fallback default for the AI21 completions parser uses a comma instead of a colon:

```python
# Before (broken — creates a set literal, always raises TypeError):
response_json.get("completions", [{"data", {"text": ""}}])[0].get(...)

# After (correct — creates a dict literal, matching the sibling lines):
response_json.get("completions", [{"data": {"text": ""}}])[0].get(...)
```

The comma makes `{"data", {"text": ""}}` a Python set literal containing a string and a dict. Since dicts are unhashable, this always raises `TypeError: cannot use 'dict' as a set element (unhashable type: 'dict')`, which gets silently caught by the outer `except Exception` at line 425 and returns `"Error parsing response"`.

The sibling lines for Anthropic (393), Mistral (408), and Cohere (410) all use the correct colon syntax.

**Reproduction:**
```python
>>> response_json = {}
>>> response_json.get("completions", [{"data", {"text": ""}}])[0]
TypeError: cannot use 'dict' as a set element (unhashable type: 'dict')
```

**Tests added:** 5 new tests in `TestParseResponseLegacy` covering the InvokeModel response parsing path for AI21, Anthropic, Mistral, and Cohere — including the specific regression case where AI21's `completions` key is absent and the fallback must work.

### DIFF
--- a/mem0/llms/aws_bedrock.py
+++ b/mem0/llms/aws_bedrock.py
@@ -409,7 +409,7 @@ class AWSBedrockLLM(LLMBase):
             elif self.provider == "cohere":
                 return response_json.get("generations", [{"text": ""}])[0].get("text", "")
             elif self.provider == "ai21":
-                return response_json.get("completions", [{"data", {"text": ""}}])[0].get("data", {}).get("text", "")
+                return response_json.get("completions", [{"data": {"text": ""}}])[0].get("data", {}).get("text", "")
             else:
                 # Generic parsing - try common response fields
                 for field in ["content", "text", "completion", "generation"]:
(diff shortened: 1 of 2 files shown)


lawrence3699:
recheck

Issue #4695, opened 2026-04-04
Ollama embedding dimensions not propagated to Qdrant vector store (bge-m3: 1024 vs hardcoded 1536)


obbax:
## Bug Description

When using Ollama as the embedding provider with models that output dimensions != 1536 (e.g., `bge-m3` outputs 1024, `nomic-embed-text` outputs 768), the Qdrant vector store collection is still created with the hardcoded default of 1536 dimensions. This causes dimension mismatch errors or silent failures.

## Related Issues

This is a persistent bug that has been reported multiple times but remains unfixed as of v1.0.10:
- #4173 — Qdrant collection created with wrong vector dimension when using Ollama embeddings
- #4212 — Qdrant collection created with wrong dimensions when using Ollama embedder (nomic-embed-text)
- #4614 — embeddingDims not passed to embedding API in OSS mode

## Root Cause

The `EmbedderConfig` class defaults `embedding_dims` to `None`, which then gets resolved to `1536` in the Qdrant vector store initialization:

```python
# In mem0/configs/base.py or similar
class EmbedderConfig:
    embedding_dims: Optional[int] = None  # Resolves to 1536 in Qdrant store

# In mem0/vector_stores/qdrant.py
embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
```

Even when `embedding_dims` is explicitly set in the embedder config, it does not propagate to the vector store config during `Memory.from_config()`.

## Reproduction

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "qdrant",
        "config": {
            "url": "http://localhost:6333",
            "collection_name": "test_dims",
        }
    },
    "embedder": {
        "provider": "ollama",
        "config": {
            "model": "bge-m3",
            "embedding_dims": 1024,  # Explicitly set, but NOT propagated to Qdrant
        }
    }
}

m = Memory.from_config(config)
m.add("test memory", user_id="test")
# Result: Qdrant collection created with 1536 dimensions
# bge-m3 returns 1024-dimensional vectors → dimension mismatch error
```

## Current Workaround

Users must manually set `embedding_model_dims` in BOTH the embedder config AND the vector store config:

```python
config = {
    "vector_store": {
        "provider": "qdrant",
        "config": {
            "embedding_model_dims": 1024,  # Must be duplicated here
        }
    },
    "embedder": {
        "provider": "ollama",
        "config": {
            "model": "bge-m3",
            "embedding_dims": 1024,
        }
[...]

PR #4706, opened 2026-04-04
fix(embeddings): auto-detect Ollama embedding dims instead of hardcoded fallback

Prithvi1994:
## Problem

When using Ollama as the embedding provider, `embedding_dims` was set to a hardcoded fallback of `512` if not explicitly configured. Different Ollama models output different dimensions:

| Model | Actual dims |
|-------|------------|
| `bge-m3` | 1024 |
| `nomic-embed-text` | 768 |
| `mxbai-embed-large` | 1024 |

This caused Qdrant (and other dimension-sensitive vector stores) to create collections with `size=512`, leading to dimension mismatch errors at insert time.

Reported in #4695, and previously in #4173 and #4212.

## Root Cause

`mem0/embeddings/ollama.py` line 29:
```python
self.config.embedding_dims = self.config.embedding_dims or 512
```

The static `512` fallback is wrong for most popular Ollama embedding models.

## Fix

Added a `_detect_embedding_dims()` method that makes a lightweight probe call to `client.embed()` at init time and measures the actual output vector length. The probe only runs when `embedding_dims` is not explicitly set by the user.

- User-supplied `embedding_dims` is always respected (probe is skipped entirely)
- Falls back to `512` only if the probe itself fails (e.g. Ollama server unreachable)
- No new dependencies

## Testing

The detection logic is self-contained and the existing Ollama test suite covers `embed()` call behavior. The `_detect_embedding_dims` method gracefully handles probe failures with a try/except.

### DIFF
--- a/mem0/embeddings/lmstudio.py
+++ b/mem0/embeddings/lmstudio.py
@@ -11,7 +11,6 @@ class LMStudioEmbedding(EmbeddingBase):
         super().__init__(config)
 
         self.config.model = self.config.model or "nomic-ai/nomic-embed-text-v1.5-GGUF/nomic-embed-text-v1.5.f16.gguf"
-        self.config.embedding_dims = self.config.embedding_dims or 1536
         self.config.api_key = self.config.api_key or "lm-studio"
 
         self.client = OpenAI(base_url=self.config.lmstudio_base_url, api_key=self.config.api_key)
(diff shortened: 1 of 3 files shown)


kartik-mem0:
Hey @Prithvi1994, thanks for putting this together and for the thorough write-up!

After reviewing, we're going to close this one. The dimension mismatch can already be handled by explicitly setting `embedding_model_dims` in the vector store config to match the embedder's output dimensions. We've commented on the linked issue (#4695) with the details.

Appreciate the contribution though — thanks for digging into this!

PR #4710, opened 2026-04-05
fix(embeddings): auto-detect LM Studio embedding dims instead of hardcoded 1536

Prithvi1994:
## Problem

`LMStudioEmbedding` hardcoded `embedding_dims = 1536` as a fallback. LM Studio is a local model server that can serve **any** model, so the actual output dimensions depend entirely on what model the user has loaded:

| Model | Actual dims |
|-------|------------|
| `nomic-embed-text` | 768 |
| `bge-large-en-v1.5` | 1024 |
| `e5-large-v2` | 1024 |
| `text-embedding-ada-002` | 1536 |

When using anything other than an OpenAI-compatible 1536-dim model, Qdrant (and other dimension-sensitive vector stores) would create collections with the wrong size, leading to dimension mismatch errors at insert time.

## Root Cause

`mem0/embeddings/lmstudio.py` line 14:
```python
```

## Fix

Added `_detect_embedding_dims()` that makes a lightweight probe call to `client.embeddings.create()` at init time and measures the actual output vector length. The probe only runs when `embedding_dims` is not explicitly set by the user.

- User-supplied `embedding_dims` is always respected (probe is skipped)
- Falls back to `1536` only if the probe itself fails (e.g. LM Studio not running)
- No new dependencies

### DIFF
--- a/mem0/embeddings/lmstudio.py
+++ b/mem0/embeddings/lmstudio.py
@@ -11,10 +11,28 @@ class LMStudioEmbedding(EmbeddingBase):
         super().__init__(config)
 
         self.config.model = self.config.model or "nomic-ai/nomic-embed-text-v1.5-GGUF/nomic-embed-text-v1.5.f16.gguf"
-        self.config.embedding_dims = self.config.embedding_dims or 1536
         self.config.api_key = self.config.api_key or "lm-studio"
 
         self.client = OpenAI(base_url=self.config.lmstudio_base_url, api_key=self.config.api_key)
+
+        if not self.config.embedding_dims:
+            self.config.embedding_dims = self._detect_embedding_dims()
+
+    def _detect_embedding_dims(self) -> int:
+        """
+        Auto-detect the embedding dimensions by making a test embed call.
+
+        LM Studio can serve any locally loaded model, so the output dimensions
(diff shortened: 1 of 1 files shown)


Prithvi1994:
Hi team! Just checking in on this PR — happy to make any adjustments or address feedback. Let me know if there's anything blocking. Thanks!

kartik-mem0:
Hey @Prithvi1994, thank you so much for taking the time to investigate this and submit a PR! We really appreciate your contribution and the detailed analysis of the dimension mismatch issue.

After reviewing this, we've decided to close the PR for a couple of reasons:

1. **Users can already configure dimensions manually** - The `embedding_dims` parameter is exposed in the config, so users can set the correct dimensions for their specific model.

2. **Latency concerns** - The auto-detection approach adds an extra API call at initialization time, which we'd prefer to avoid.

For anyone running into this issue, you can configure the dimensions directly in your config:

```python
config = {
    "embedder": {
        "provider": "lmstudio",
        "config": {
            "model": "your-model",
            "embedding_dims": 768  # Set to match your model
        }
    }
}
```

 **Docs**: https://docs.mem0.ai/components/embedders/models/lmstudio

Thanks again for your contribution, and we hope to see you around the project!

PR #4713, opened 2026-04-05
fix(openai): only send `store` param when explicitly configured

### LINKED ISSUE #4709: [Bug][Python] store=False unconditionally sent to all OpenAI-compatible backends — breaks Gemini and other non-OpenAI providers
## Summary

## Code Path

```python
store: bool = False
```

```python
```

## Impact

## Workaround Applied

```python
    kwargs.pop("store", None)
    return _orig_post(**kwargs)
```

## Expected Behavior

## Suggested Fix

```python
```

## Environment

- Package: `mem0ai` (Python)

rakheesingh:
## Summary

Fixes #4709

- `OpenAIConfig.store` defaulted to `bool = False`, and `generate_response()` used `hasattr()` to check whether to include it — which is always `True` since the attribute always exists. This caused `store=False` to be injected into **every** API request, even when using non-OpenAI backends (Gemini, Groq, Mistral, local proxies) that reject unknown fields.
- Changed `store` default from `bool = False` to `Optional[bool] = None` so it is only sent when the user explicitly sets it.
- Updated the `hasattr` check to `getattr(..., None) is not None` so `None` values are skipped.

## Changes

| File | Change |
|------|--------|
| `mem0/configs/llms/openai.py` | `store: bool = False` → `store: Optional[bool] = None` |
| `mem0/llms/openai.py` | Replace `hasattr` check with `getattr` + `is not None` guard |
| `tests/llms/test_openai.py` | Updated 2 existing tests, added 4 new tests |

## Test plan

- [x] `test_store_not_sent_by_default` — verifies `store` is absent from API params when not configured
- [x] `test_store_sent_when_explicitly_set_false` — verifies `store=False` is sent when user sets it
- [x] `test_store_sent_when_explicitly_set_true` — verifies `store=True` is sent when user sets it
- [x] `test_store_config_default_is_none` — verifies config default is `None`
- [x] All 15 tests pass (`pytest tests/llms/test_openai.py`)

### DIFF
--- a/tests/llms/test_openai.py
+++ b/tests/llms/test_openai.py
@@ -11,6 +11,7 @@ from mem0.llms.openai import OpenAILLM
 def mock_openai_client():
     with patch("mem0.llms.openai.OpenAI") as mock_openai:
         mock_client = Mock()
+        mock_client.base_url = "https://api.openai.com/v1/"
         mock_openai.return_value = mock_client
         yield mock_client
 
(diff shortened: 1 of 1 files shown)


kartik-mem0:

closing this pr in favour of this one #4757.

PR #4718, opened 2026-04-05
fix: use native AsyncOpenAI in AsyncMemory instead of asyncio.to_thread

### LINKED ISSUE #3781: OpenAILLM(LLMBase) does not use AsyncOpenAI; AsyncMemory still runs sync LLM via asyncio.to_thread
### 🚀 The feature

Inside AsyncMemory, the LLM call is still synchronous and only wrapped with:

asyncio.to_thread(self.llm.generate_response)

This means the LLM workload is executed in a thread pool, not asynchronously,

### Motivation, pitch

Thanks

kaiisfree:
## Summary

- Add `agenerate_response()` async method to `LLMBase` with `asyncio.to_thread` fallback for providers that don't have native async clients
- Implement `agenerate_response()` in `OpenAILLM` using the `AsyncOpenAI` client for true async I/O
- Update `AsyncMemory` to call `llm.agenerate_response()` directly instead of wrapping sync calls with `asyncio.to_thread`

## Motivation

`AsyncMemory` was wrapping synchronous `OpenAI` client calls via `asyncio.to_thread(self.llm.generate_response, ...)`, which runs the LLM call in a thread pool rather than using true async I/O. The `openai` Python library provides `AsyncOpenAI` specifically for this purpose.

## Changes

| File | Change |
|------|--------|
| `mem0/llms/base.py` | Added `agenerate_response()` with `asyncio.to_thread` fallback |
| `mem0/llms/openai.py` | Added `AsyncOpenAI` client + native `agenerate_response()` |
| `mem0/memory/main.py` | Replaced 3 `asyncio.to_thread(self.llm.generate_response, ...)` calls with `self.llm.agenerate_response(...)` |
| `tests/llms/test_openai.py` | Updated fixture to also mock `AsyncOpenAI` |
| `tests/llms/test_openai_async.py` | New tests for async OpenAI LLM |
| `tests/llms/test_base_async.py` | New test for base class fallback |

## Backward compatibility

- `generate_response()` is unchanged — all sync usage works exactly as before
- Non-OpenAI providers automatically get `asyncio.to_thread` fallback via `LLMBase.agenerate_response()`, so `AsyncMemory` continues to work with all providers
- Other `asyncio.to_thread` calls in `AsyncMemory` (for embeddings, vector store, graph, etc.) are left untouched — this PR focuses only on LLM calls

## Test plan

- [x] All 11 existing `test_openai.py` tests pass
- [x] 5 new async OpenAI tests pass (with/without tools, response format, client isolation)
- [x] 1 new base class fallback test passes

Closes #3781

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -1,3 +1,4 @@
+import asyncio
 from abc import ABC, abstractmethod
 from typing import Dict, List, Optional, Union
 
@@ -116,6 +117,27 @@ class LLMBase(ABC):
             str or dict: The generated response.
         """
         pass
+
+    async def agenerate_response(
+        self, messages: List[Dict[str, str]], tools: Optional[List[Dict]] = None, tool_choice: str = "auto", **kwargs
+    ):
+        """
+        Async version of generate_response. By default, wraps the sync method
+        via asyncio.to_thread. Subclasses with native async clients (e.g. OpenAI)
+        should override this for true async behavior.
+
+        Args:
+            messages (list): List of message dicts containing 'role' and 'content'.
+            tools (list, optional): List of tools that the model can call. Defaults to None.
+            tool_choice (str, optional): Tool choice method. Defaults to "auto".
+            **kwargs: Additional provider-specific parameters.
+
+        Returns:
+            str or dict: The generated response.
+        """
+        return await asyncio.to_thread(
+            self.generate_response, messages=messages, tools=tools, tool_choice=tool_choice, **kwargs
+        )
 
     def _get_common_params(self, **kwargs) -> Dict:
         """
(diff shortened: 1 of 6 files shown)


kaiisfree:

kaiisfree:
recheck

kartik-mem0:
Thanks so much for this — the direction is exactly right (a native `AsyncOpenAI` client with an `asyncio.to_thread` fallback on the base class), and the added async tests are genuinely appreciated.

We're going to close this for now, though. The branch has drifted from `main` (merge conflicts), and in the meantime the OpenAI param handling changed — `store` is now opt-in, only sent when explicitly configured, so OpenAI-compatible backends like Gemini/Groq/vLLM don't reject it. The async path here would need to track that, ideally by sharing the param-building logic with the sync method rather than duplicating it.

This isn't a "no" on the idea — it's a real improvement and we'd happily revisit it. If you'd like to pick it back up, a rebased PR that factors out the shared param logic would be very welcome. Thanks again for the effort. 🙏

PR #4746, opened 2026-04-08
fix(llms): narrow _is_reasoning_model to not match gpt-5.x variants

rudra717:
## Linked Issue

Closes #4738

## Description

`_is_reasoning_model()` used a substring match (`"gpt-5" in model_lower`) that incorrectly classified `gpt-5.4-mini` and other gpt-5.x variants as reasoning models. This caused `temperature` and other standard parameters to be silently stripped from API calls.

## Root Cause

```python
# Before — matches gpt-5, gpt-5o, BUT ALSO gpt-5.4-mini, gpt-5.4
if any(reasoning_model in model_lower for reasoning_model in ["gpt-5", "o1", "o3"]):
    return True
```

`gpt-5.4-mini` contains `"gpt-5"` as a substring, so it was misclassified. The `_get_supported_params()` method then stripped all parameters except `messages`, `response_format`, `tools`, and `tool_choice` — including `temperature`.

## Fix

Replaced the broad substring match with:
- **Exact match** against known reasoning models (`o1`, `o3`, `gpt-5`, `gpt-5o`, etc.)
- **Prefix match** for dated variants (`o1-2024-12-17`, `o3-2025-04-16`)
- **Provider prefix stripping** (`openai/o3-mini` → `o3-mini`)
- **Removed** the `"gpt-5"` substring match entirely

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Test Coverage

- [x] I added/updated unit tests

Added 2 tests:
- `test_gpt5_mini_not_classified_as_reasoning` — verifies gpt-5.4-mini passes temperature through
- `test_is_reasoning_model_classification` — comprehensive classification test covering reasoning models (o1, o3, gpt-5, prefixed variants) and non-reasoning models (gpt-5.4-mini, gpt-5.4, gpt-4.1)

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #4747, opened 2026-04-08
fix(ollama): deep-copy messages to prevent in-place mutation

rudra717:
## Linked Issue

Related to data integrity concerns in the Ollama LLM provider.

## Description

`OllamaLLM.generate_response()` mutated the caller's messages list in-place when `response_format={"type": "json_object"}` was requested. It appended JSON instruction text to `messages[-1]["content"]` and could append a new message to the list. Any code reusing the messages list after the call would see corrupted data.

## Root Cause

```python
# Before — mutates caller's list
if messages and messages[-1]["role"] == "user":
    messages[-1]["content"] += "\n\nPlease respond with valid JSON only."
```

## Fix

Deep-copy messages before modification and rebind `params["messages"]` to the copy:

```python
messages = copy.deepcopy(messages)
params["messages"] = messages
```

## Type of Change

## Test Coverage

Added `test_generate_response_json_format_does_not_mutate_messages` which verifies:
1. Caller's messages list length is unchanged after the call
2. Caller's message content is unchanged
3. Ollama client still receives the `format=json` param
4. Ollama client still receives the JSON instruction in the copied messages

## Checklist

### DIFF
--- a/mem0/llms/ollama.py
+++ b/mem0/llms/ollama.py
@@ -1,3 +1,4 @@
+import copy
 import json
 from typing import Dict, List, Optional, Union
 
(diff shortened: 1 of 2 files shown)


rudra717:
@kartik-mem0 Following up on the Ollama mutation fix, generate_response() mutates the caller's messages list in-place when JSON format is requested, so a second call ends up with two copies of the JSON instruction. Fix is a copy.deepcopy() with regression test. Same shape as #4746 which you merged. Could you take a look when you have a sec?

PR #4748, opened 2026-04-08
fix(llms): remove dead validation + add reasoning model compat to OpenAIStructuredLLM

rudra717:
## Linked Issue

Addresses LLM provider code quality and reasoning model compatibility.

## Description

Two fixes in the LLM provider layer:

### 1. Dead code removal in `base.py`

`_validate_config()` had a duplicate condition that checked the same attribute twice:

```python
if not hasattr(self.config, "api_key") and not hasattr(self.config, "api_key"):
    pass
```

This was dead code — removed entirely.

### 2. Reasoning model compatibility in `openai_structured.py`

`OpenAIStructuredLLM` passed `temperature` directly to the API without filtering. Reasoning models (o1, o3) don't support `temperature` and reject the request. All other providers use `_get_supported_params()` which strips unsupported params for reasoning models — now `OpenAIStructuredLLM` follows the same pattern.

Also strips `None` values from params to avoid sending unset optional fields to the API.

## Type of Change

## Test Coverage

Added `tests/llms/test_openai_structured.py` with:
- `test_reasoning_model_strips_temperature` — verifies o3 does NOT receive temperature
- `test_regular_model_includes_temperature` — verifies gpt-4o does receive temperature

## Checklist

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -35,11 +35,6 @@ class LLMBase(ABC):
         if not hasattr(self.config, "model"):
             raise ValueError("Configuration must have a 'model' attribute")
 
-        if not hasattr(self.config, "api_key") and not hasattr(self.config, "api_key"):
-            # Check if API key is available via environment variable
-            # This will be handled by individual providers
-            pass
-
     def _is_reasoning_model(self, model: str) -> bool:
         """
         Check if the model is a reasoning model or GPT-5 series that doesn't support certain parameters.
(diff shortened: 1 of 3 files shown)


rudra717:
@kartik-mem0 (and tagging @whysosaket since you've been the most active recently) companion PR to #4747. Two small fixes in one: removes a dead duplicate api_key hasattr check, and routes OpenAIStructuredLLM through _get_supported_params() so reasoning models stop receiving temperature. Also, tests added. Would appreciate a look, thanks!

PR #4753, opened 2026-04-08
fix: only send store when explicitly configured

### LINKED ISSUE #4709: [Bug][Python] store=False unconditionally sent to all OpenAI-compatible backends — breaks Gemini and other non-OpenAI providers

voidborne-d:
## Summary
- make `OpenAIConfig.store` opt-in by defaulting it to `None`
- only forward `store` to the OpenAI client when users explicitly configure it
- update OpenAI LLM tests to cover the new default behavior and the explicit `store=False` path

## Testing
- `python3 -m pytest tests/llms/test_openai.py -q`

## Why
OpenAI-compatible backends like Gemini reject the OpenAI-only `store` field. Before this change, mem0 always injected `store=False` because the config field always existed, even when users never asked for it.

Closes #4709.

### DIFF
--- a/mem0/configs/llms/openai.py
+++ b/mem0/configs/llms/openai.py
@@ -29,7 +29,7 @@ class OpenAIConfig(BaseLlmConfig):
         openrouter_base_url: Optional[str] = None,
         site_url: Optional[str] = None,
         app_name: Optional[str] = None,
-        store: bool = False,
+        store: Optional[bool] = None,
         # Response monitoring callback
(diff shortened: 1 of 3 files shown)


kartik-mem0:

closing this pr in favour of this one #4757.

PR #4754, opened 2026-04-08
fix qdrant memory count in openclaw

xujingchen1996:
## Summary
- fix incorrect memory counts in OpenClaw mem0 when using OSS + Qdrant
- pass `page_size` through to the OSS SDK `limit` parameter so full inventories can be fetched when the true total is known
- stabilize `sqlite-resilience.test.ts` by resetting modules between suites

## Root cause
The OSS SDK's `getAll()` defaults to `limit = 100`. OpenClaw mem0 was using `getAll().length` as if it were the total memory count, which under-reported counts for users with more than 100 memories and caused dream/inventory flows to operate on truncated data.

## What changed
- add a Qdrant exact-count helper and prefer it when OSS is configured with Qdrant
- fall back to the existing `getAll()` length behavior for other backends
- thread exact counts into dream gating and dream inventory generation
- map OSS `page_size` to SDK `limit`
- add regression coverage for the Qdrant count path and the full inventory path
- reset modules in `sqlite-resilience.test.ts` to avoid cross-suite mock leakage

## Validation
- `npx vitest run tests/providers.test.ts tests/cli-commands.test.ts`
- `npm test`

### DIFF
--- a/openclaw/cli/commands.ts
+++ b/openclaw/cli/commands.ts
@@ -36,6 +36,7 @@ import type {
   MemoryItem,
   SearchOptions,
 } from "../types.ts";
+import { buildQdrantCountRequest, countMemories } from "../providers.ts";
 import { loadDreamPrompt } from "../skill-loader.ts";
 import { readText } from "../fs-safe.ts";
 import type { PluginAuthConfig } from "./config-file.ts";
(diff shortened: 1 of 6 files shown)


kartik-mem0:
this issue has been already resolved closing this fixed

PR #4757, opened 2026-04-08
fix(openai): make store opt-in so it stops leaking to non-OpenAI backends

### LINKED ISSUE #4709: [Bug][Python] store=False unconditionally sent to all OpenAI-compatible backends — breaks Gemini and other non-OpenAI providers

Hybirdss:
## Linked Issue

Closes #4709

## Description

`OpenAIConfig.store` defaulted to `False`, which meant `hasattr(self.config, "store")` in `mem0/llms/openai.py` was always `True`. As a result, `store=False` was unconditionally injected into every `client.chat.completions.create(...)` call on the non-OpenRouter code path — including when the user pointed `openai_base_url` at an OpenAI-compatible backend like Google Gemini. Gemini (and other strict compat endpoints) reject unknown fields and the call fails.

@kaiisfree traced the root cause in the issue thread and recommended making `store` opt-in. This PR implements that:

- `mem0/configs/llms/openai.py`: `store: bool = False` → `store: Optional[bool] = None`. Docstring updated to explain the new semantics.
- `mem0/llms/openai.py`: replaces the `openai_specific_generation_params` / `hasattr` loop with an explicit `if self.config.store is not None: params["store"] = self.config.store` guard, with a comment explaining why it must be opt-in.
- `tests/llms/test_openai.py`: the two existing tests (`test_generate_response_without_tools`, `test_generate_response_with_tools`) no longer expect `store=False` in the forwarded kwargs, because `store` is no longer sent by default.

Users who deliberately set `store=True` or `store=False` still have their value forwarded to the API — only the default has changed.

## Type of Change

- [ ] Documentation update

## Breaking Changes

Behavioral change for users who were implicitly relying on the old default: previously every request included `store=False`; now nothing is sent unless `store` is explicitly configured. For direct OpenAI usage this matches OpenAI's documented default behavior, so the observable result should be unchanged. Users who want to guarantee `store=False` reaches OpenAI can set it explicitly on their `OpenAIConfig`.

## Test Coverage

Added three regression tests in `tests/llms/test_openai.py`:

- `test_store_not_sent_by_default` — asserts `"store"` is absent from the forwarded kwargs when `OpenAIConfig` is constructed without a `store` value. This is the regression guard for #4709.
- `test_store_sent_when_explicitly_true` — asserts `store=True` is forwarded when explicitly configured.
- `test_store_sent_when_explicitly_false` — asserts `store=False` is still forwarded when the user explicitly opts out, since explicit opt-out is a valid configuration for users who rely on OpenAI's zero-data-retention behavior.

Ran `pytest tests/llms/test_openai.py` in a clean venv (mem0 installed via `pip install -e .`): 14 passed in 1.10s. Ran the broader `tests/llms/` collection (skipping modules missing optional deps on this host): 103 passed.

## Checklist

### DIFF
--- a/mem0/configs/llms/openai.py
+++ b/mem0/configs/llms/openai.py
@@ -29,7 +29,7 @@ class OpenAIConfig(BaseLlmConfig):
         openrouter_base_url: Optional[str] = None,
         site_url: Optional[str] = None,
         app_name: Optional[str] = None,
-        store: bool = False,
+        store: Optional[bool] = None,
         # Response monitoring callback
         response_callback: Optional[Callable[[Any, dict, dict], None]] = None,
     ):
@@ -53,6 +53,11 @@ class OpenAIConfig(BaseLlmConfig):
             openrouter_base_url: OpenRouter base URL, defaults to None
             site_url: Site URL for OpenRouter, defaults to None
             app_name: Application name for OpenRouter, defaults to None
+            store: Whether to store the conversation on OpenAI's server. Opt-in;
+                defaults to None (not sent). Set to True or False only if you
+                want the value forwarded to the OpenAI API. Leaving it None
+                avoids leaking the field into OpenAI-compatible backends that
+                reject unknown fields (Gemini, Groq, vLLM, etc.).
             response_callback: Optional callback for monitoring LLM responses.
         """
         # Initialize base parameters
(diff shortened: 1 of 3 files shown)


Hybirdss:

kartik-mem0 (approved):

PR #4769, opened 2026-04-09
fix(openclaw): clear recall timeout to stop phantom timeout warnings (#4763)

### LINKED ISSUE #4763: [Bug] Spurious Timeout Errors Despite Successful Recall Operations - Promise.race Implementation Issue
### Component

Core / Python SDK

## Bug Description
The mem0 OSS SDK produces spurious timeout error messages even when recall operations complete successfully within the timeout window. This appears to be a bug in the Promise.race implementation where the timeout promise is not properly cancelled after the main operation succeeds.

## Environment
- mem0 OSS SDK integrated with OpenClaw
- Local Ollama setup with nomic-embed-text model (30MB)
- Timeout increased from default 8s to 30s for debugging
- Node.js environment with Qdrant vector store

## Root Cause Analysis
When calling memory recall operations, the implementation uses `Promise.race()` between the actual operation and a timeout promise. The bug manifests as:

1. The recall operation completes successfully in ~370ms
2. The result is returned correctly to the caller
3. But ~30 seconds later, spurious timeout error messages appear in logs
4. The timeout promise continues to execute even after the race was won

This suggests the timeout promise is not being cancelled/aborted when the main operation completes first.

## Evidence from Logs

*Important note: as part of debugging, I increased the timeout from 8 seconds to 30 seconds intentionally, and inserted additional logging language that does not appear in the current version of the plugin code.*

**Successful Recall Operation (370ms):**
```
06:18:01+00:00 info plugins {"subsystem":"plugins"} openclaw-mem0: Starting recall process at 2026-04-09T06:18:01.361Z, prompt length: 242
06:18:01+00:00 info plugins {"subsystem":"plugins"} openclaw-mem0: recallWork started, elapsed: 2ms
06:18:01+00:00 info plugins {"subsystem":"plugins"} openclaw-mem0: About to call provider.search with prompt, elapsed: 3ms
06:18:01+00:00 info plugins {"subsystem":"plugins"} openclaw-mem0: provider.search completed in 364ms, found 10 results, elapsed total: 368ms
06:18:01+00:00 info plugins {"subsystem":"plugins"} openclaw-mem0: Recall process completed, injecting 5 memories, total elapsed: 370ms
06:18:01+00:00 info plugins {"subsystem":"plugins"} openclaw-mem0: injecting 5 memories into context
06:18:01+00:00 info plugins {"subsystem":"plugins"} openclaw-mem0: Race completed, result was successful
```

**Spurious Timeout Messages (30s later):**
```
06:18:31+00:00 warn plugins {"subsystem":"plugins"} openclaw-mem0: recall timeout triggered after 30000ms
06:18:31+00:00 warn plugins {"subsystem":"plugins"} openclaw-mem0: recall timed out after 30000ms, skipping
```

The recall clearly succeeded in 370ms, but the timeout promise fires anyway because it wasn't cancelled.

## Expected Behavior
When `Promise.race()` is won by the main operation:
1. The timeout timer should be cleared immediately
2. No timeout error should be thrown
3. The timeout promise should be aborted/cancelled

## Actual Behavior
1. Main operation completes successfully in 370ms
2. Result is returned to caller
3. Timeout promise continues to exist
4. After timeout duration expires (30 seconds), error is thrown/logged
5. This causes confusing logs and potential unhandled promise rejections

## Suggested Fix
The timeout implementation should use `AbortController` or clear the timeout immediately when the race resolves:

**Incorrect**
```javascript
// Current problematic pattern (simplified)
const result = await Promise.race([
  operation(),
  new Promise((_, reject) => 
    setTimeout(() => reject(new TimeoutError()), 30000)
  )
]);
```
**Correct?**
```
let timeoutId;
const timeoutPromise = new Promise((_, reject) => {
  timeoutId = setTimeout(() => reject(new TimeoutError()), 30000);
});

try {
  const result = await Promise.race([operation(), timeoutPromise]);
  clearTimeout(timeoutId);  // Clean up the timeout
  return result;
} catch (error) {
  clearTimeout(timeoutId);  // Clean up on error too
  throw error;
}
```

## Impact
- Confusing logs: Developers see timeout errors despite operations succeeding
- Potential memory leaks: Timeout timers not being cleaned up
- Unhandled rejections: Timeout promises may cause unhandled rejection warnings
- Debugging difficulty: Makes it harder to identify actual timeout issues (consumed 12+ hours of debugging time)

## Workaround

Currently, increasing timeout values and adding verbose logging helps identify that the operations are actually succeeding, but the fundamental Promise.race bug remains.

Labels: bug, needs-triage

*Disclaimer: total novice at this, vibe coding with robots -- hopefully I didn't make a complete mess of this report!*

DmitryPogodaev:
## Summary

Fixes #4763. The `before_prompt_build` recall handler in `openclaw/index.ts` raced `recallWork()` against a `setTimeout`-backed timeout promise, but the timer was never cancelled when `recallWork` won the race. The scheduled callback still fired `RECALL_TIMEOUT_MS` later and ran `timeout.then(() => logger.warn(...))`, producing a misleading

```
openclaw-mem0: recall timed out after 8000ms, skipping
```

warning in the log long after a successful recall had already injected memories into the agent prompt.

In production this produced roughly one phantom warning per recall, which made it impossible to tell real timeouts from false positives and suggested a (non-existent) ~50% failure rate on the recall path.

## Fix

* Keep a reference to the `setTimeout` handle and `clearTimeout` it in a `finally` block. A successful recall (or a thrown error in `recallWork`) now cancels the pending timer so the `.then` callback never runs.
* Use an explicit `Symbol` sentinel on the timeout branch. Previously both "recall finished with no matching memories" and "timeout won the race" resolved to `undefined`, so the warning could not be gated on that value alone.
* Log the warning and return `undefined` only when the sentinel actually wins the race, preserving the exact user-visible behaviour of real timeouts.

## Test plan

Verified on a live OpenClaw gateway (`@mem0/openclaw-mem0@1.0.5`, OSS mode with Qdrant + Ollama):

- [x] Successful recall with matches: `injecting N memories into context` fires once, no phantom warning at handler-start + 8s.
- [x] Successful recall with **no** matches: handler returns silently, no `injecting` log, no phantom warning.
- [x] Forced real timeout (paused the embedder): `recall timed out after 8000ms, skipping` fires exactly once, handler returns `undefined`. Same behaviour as before the fix.
- [x] Thrown error inside `recallWork`: `finally` still runs, timer is cleared, no phantom warning, `recall failed: ...` is logged as before.
- [x] `node --check` on the transpiled bundle; TypeScript type-checks cleanly against the sentinel `Symbol` + `ReturnType<typeof setTimeout>`.

## Notes

* Kept the fix scoped to the single recall handler. The surrounding `recallWork` logic, `auto-capture` path, and the rest of `registerHooks` are untouched.
* Did not change `RECALL_TIMEOUT_MS` or expose it as config — intentionally a minimal fix. Happy to follow up with a separate PR making the timeout configurable if that's desired.
* Commit message references the issue; let me know if you'd like the PR title or commit split differently.

### DIFF
--- a/openclaw/index.ts
+++ b/openclaw/index.ts
@@ -736,22 +736,34 @@ function registerHooks(
         };
       };
 
+      // Sentinel to distinguish a real timeout from a legitimate `undefined`
+      // return from `recallWork` (e.g. when no memories match). This avoids
+      // mis-logging a timeout when the work completed successfully with no
+      // results.
+      const TIMEOUT_SENTINEL = Symbol("openclaw-mem0.recall.timeout");
+      let timeoutHandle: ReturnType<typeof setTimeout> | undefined;
       try {
-        const timeout = new Promise<undefined>((resolve) => {
-          setTimeout(() => resolve(undefined), RECALL_TIMEOUT_MS);
+        const timeout = new Promise<typeof TIMEOUT_SENTINEL>((resolve) => {
+          timeoutHandle = setTimeout(
+            () => resolve(TIMEOUT_SENTINEL),
+            RECALL_TIMEOUT_MS,
+          );
         });
-        const result = await Promise.race([
-          recallWork(),
-          timeout.then(() => {
-            api.logger.warn(
-              `openclaw-mem0: recall timed out after ${RECALL_TIMEOUT_MS}ms, skipping`,
-            );
-            return undefined;
-          }),
-        ]);
+        const result = await Promise.race([recallWork(), timeout]);
+        if (result === TIMEOUT_SENTINEL) {
+          api.logger.warn(
+            `openclaw-mem0: recall timed out after ${RECALL_TIMEOUT_MS}ms, skipping`,
+          );
+          return undefined;
+        }
         return result;
       } catch (err) {
         api.logger.warn(`openclaw-mem0: recall failed: ${String(err)}`);
+      } finally {
+        // Cancel the pending timer so a successful recall does not leave a
+        // phantom `setTimeout` callback that would fire RECALL_TIMEOUT_MS
+        // later and log a spurious "recall timed out" warning. See #4763.
+        if (timeoutHandle !== undefined) clearTimeout(timeoutHandle);
       }
     });
   }


kartik-mem0:
Hi @DmitryPogodaev, thanks for the thorough investigation and well-documented PR!

You correctly identified the phantom timeout warning bug in the Promise.race pattern. However, this code path is part of the legacy recall mode, which has been superseded by the newer skills-based recall architecture that doesn't use the Promise.race timeout pattern at all.

Since legacy mode is being phased out in favor of skills mode, we're going to close this PR rather than patch the older code path. We appreciate you taking the time to track this down and test it thoroughly — the analysis was solid.

Thanks for contributing!

PR #4781, opened 2026-04-10
fix(ts-sdk): include url fallback in LLM config resolution

### LINKED ISSUE #4715: LLM config manager ignores url field — Ollama/custom LLM providers connect to api.openai.com
### Component

TypeScript SDK

### Description

### Summary

The LLM config manager's `baseURL` resolution does not fall back to `userConf.url`, unlike the embedder config. When configuring a non-OpenAI LLM provider (e.g. Ollama) with `url`, the config   manager ignores it and falls back to `defaultConf.baseURL` (`https://api.openai.com/v1`), causing LLM requests to silently hit OpenAI instead of the configured host.                           
                                                                                                                                                                                                  
The embedder config correctly includes `userConf?.url` in its fallback chain (`manager.ts` L24-29), but the LLM config does not (`manager.ts` L108-113).

### Steps to Reproduce 
```typescript

const m = new Memory({
  llm: {
    provider: "ollama",
    config: {
      model: "llama3.1:8b",
      url: "http://my-ollama-host:11434"
    }
  },
  embedder: {
    provider: "openai",
    config: {
      model: "text-embedding-3-small"
    }
  }
});

await m.add("I like tacos", "user1");
```

### Expected Behavior

OllamaLLM connects to http://my-ollama-host:11434 for memory extraction during capture.

### Actual Behavior

OllamaLLM receives baseURL: "https://api.openai.com/v1" from the config manager. Depending on the environment:

- TypeError: fetch failed — if the host cannot reach api.openai.com
- 404 Not Found — if it can reach OpenAI but the Ollama client hits OpenAI endpoints with Ollama API paths
- Silent wrong-host routing — if an OpenAI API key happens to be set

The root cause is in ConfigManager.mergeConfig():
```typescript
// Embedder (correct — falls back to url):
const baseURL = userConf?.baseURL ?? ... ?? userConf?.url;

// LLM (broken — skips url, falls back to OpenAI default):
const llmBaseURL = userConf?.baseURL ?? ... ?? defaultConf.baseURL;
```

### Environment

- mem0 version: mem0ai 2.3.0 – 2.4.5 (config manager unchanged on main)
- Node version: 22
- OS: Debian Bookworm (Docker container)

Yanhu007:
## Summary

Fixes #4715 — LLM config manager ignores `url` field, causing Ollama/custom LLM providers to connect to `api.openai.com` instead.

## Root Cause

The LLM config `baseURL` resolution chain in `manager.ts` (L108–113) was:
```
userConf?.baseURL ?? lmstudio_base_url ?? defaultConf.baseURL
```

Missing `userConf?.url` fallback that the embedder config (L24–29) already includes:
```
userConf?.baseURL ?? lmstudio_base_url ?? userConf?.url
```

## Fix

One-line change: add `userConf?.url ??` before `defaultConf.baseURL` in the LLM config fallback chain, making it consistent with the embedder config.

## Before → After

| Config | Before | After |
|--------|--------|-------|
| `{ provider: "ollama", config: { url: "http://host:11434" } }` | → `https://api.openai.com/v1` ❌ | → `http://host:11434` ✅ |

### DIFF
--- a/mem0-ts/src/oss/src/config/manager.ts
+++ b/mem0-ts/src/oss/src/config/manager.ts
@@ -110,6 +110,7 @@ export class ConfigManager {
             ((userConf as Record<string, unknown>)?.lmstudio_base_url as
               | string
               | undefined) ??
+            userConf?.url ??
             defaultConf.baseURL;
 
           return {


kartik-mem0:
closing this pr in the favour #4761 of this as it has regression test.

PR #4786, opened 2026-04-11
fix: strip /v1 suffix from ollama_base_url to prevent incorrect endpoint construction

### LINKED ISSUE #4736: “mem0 plugin fails with “TypeError: fetch failed” when using Ollama – wrong embeddings endpoint (/v1/embeddings)/
### Component

Core / Python SDK

### Description

 ### Summary                                                                                                                                                                            
                                                                                                                                                                                        
 The mem0 plugin fails to create or use the embedding model when OpenClaw is configured to use Ollama as the provider. Every call results in:                                           
                                                                                                                                                                                        
 ```                                                                                                                                                                                    
   [ERROR] Error ensuring model exists: TypeError: fetch failed                                                                                                                         
   Search failed: TypeError: fetch failed                                                                                                                                               
 ```                                                                                                                                                                                    
                                                                                                                                                                                        
 The root cause is that mem0 still tries to call the OpenAI‑style endpoint /v1/embeddings instead of Ollama’s required /api/embeddings. The configuration change was applied to the wrong file or in the wrong hierarchy, so the plugin never sees the embeddingsPath override.                                                                                            
                                                                                                                                                                                        
 ────────────────────────────────────────────────────────────────────────────────                                                                                                       
                                                                                                                                                                                        
 ### Steps to Reproduce                                                                                                                                                                 
                                                                                                                                                                                        
 ```python                                                                                                                                                                              
   from mem0 import Memory                                                                                                                                                              
                                                                                                                                                                                        
   # OpenClaw is started with the default config (provider: ollama)                                                                                                                     
   # but without the proper embeddingsPath override.                                                                                                                                    
   m = Memory()                                                                                                                                                                         
   result = m.search(“Qdrant Testing")   # any call that forces an embedding lookup                                                                                                               
   print(result)                                                                                                                                                                        
 ```                                                                                                                                                                                    
                                                                                                                                                                                        
 Running the above on a system where OpenClaw’s openclaw-mem0 plugin is registered with:                                                                                                
                                                                                                                                                                                        
 ```yaml                                                                                                                                                                                
   plugins:                                                                                                                                                                             
     entries:                                                                                                                                                                           
       openclaw-mem0:                                                                                                                                                                   
         enabled: true                                                                                                                                                                  
         config:                                                                                                                                                                        
           mode: open-source                                                                                                                                                            
           userId: claw                                                                                                                                                                 
           oss:                                                                                                                                                                         
             embedder:                                                                                                                                                                  
               provider: ollama                                                                                                                                                         
               config:                                                                                                                                                                  
                 host: http://192.168.110.14:11434                                                                                                                                      
                 model: nomic-embed-text                                                                                                                                                
 ```                                                                                                                                                                                    
                                                                                                                                                                                        
 will produce the error shown below.                                                                                                                                                    
                                                                                                                                                                                        
 ────────────────────────────────────────────────────────────────────────────────                                                                                                       
                                                                                                                                                                                        
 ### Expected Behavior                                                                                                                                                                  
                                                                                                                                                                                        
 The mem0 plugin should detect that the provider is Ollama, construct the request to:                                                                                                   
                                                                                                                                                                                        
 ```                                                                                                                                                                                    
   POST http://192.168.110.14:11434/api/embeddings                                                                                                                                      
 ```                                                                                                                                                                                    
                                                                                                                                                                                        
 and successfully obtain embeddings, allowing searches and other memory operations to work.                                                                                             
                                                                                                                                                                                        
 ────────────────────────────────────────────────────────────────────────────────                                                                                                       
                                                                                                                                                                                        
 ### Actual Behavior                                                                                                                                                                    
                                                                                                                                                                                        
 The plugin attempts to call the OpenAI‑compatible endpoint:                                                                                                                            
                                                                                                                                                                                        
 ```                                                                                                                                                                                    
   POST http://192.168.110.14:11434/v1/embeddings                                                                                                                                       
 ```                                                                                                                                                                                    
                                                                                                                                                                                        
 which Ollama does not expose, resulting in a network failure and the following error output:                                                                                           
                                                                                                                                                                                        
 ```                                                                                                                                                                                    
   [ERROR] Error ensuring model exists: TypeError: fetch failed                                                                                                                         
   Search failed: TypeError: fetch failed                                                                                                                                               
 ```                                                                                                                                                                                    
                                                                                                                                                                                        
 ────────────────────────────────────────────────────────────────────────────────                                                                                                       
                                                                                                                                                                                        
 ### Environment                                                                                                                                                                        
                                                                                                                                                                                        
 - mem0 version: (the version bundled with OpenClaw 2026.4.5) – e.g., 0.2.7 (verify with pip show mem0)                                                                                 
 - OpenClaw version: 2026.4.5 (3e72c03)                                                                                                                                                 
 - Python version: 3.11 (or the version installed on the host)                                                                                                                          
 - OS: Linux 6.12.0‑124.45.1.el10_1.x86_64 (RHEL‑based)                                                                                                                                 
 - Ollama endpoint: http://192.168.110.14:11434                                                                                                                                         
 - Model used for embeddings: nomic-embed-text

DarkRiddle1212:
Linked Issue
Closes #4736
Description
When ollama_base_url is set to http://host:11434/v1 (a common mistake since most other providers like OpenAI use /v1), the Ollama Python client prepends that path to every API call, resulting in incorrect endpoints like /v1/api/embed instead of the correct /api/embed. This causes TypeError: fetch failed on every embedding call.
This fix strips the /v1 suffix from ollama_base_url before passing it to the Ollama Client, so the correct endpoint is always constructed regardless of how the user configured the base URL.
Type of Change

 Bug fix (non-breaking change that fixes an issue)
 New feature (non-breaking change that adds functionality)
 Breaking change (fix or feature that would cause existing functionality to change)
 Refactor (no functional changes)
 Documentation update

Breaking Changes
N/A
Test Coverage

 I added/updated unit tests
 I added/updated integration tests
 I tested manually (describe below)
 No tests needed (explain why)

Reproduced the incorrect endpoint construction by setting ollama_base_url to http://localhost:11434/v1 and verified the fix correctly strips the /v1 suffix before passing to the Ollama client.
Checklist

 My code follows the project's style guidelines
 I have performed a self-review of my code
 I have added tests that prove my fix/feature works
 New and existing tests pass locally
 I have updated documentation if needed

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -28,7 +28,16 @@ class OllamaEmbedding(EmbeddingBase):
         self.config.model = self.config.model or "nomic-embed-text"
         self.config.embedding_dims = self.config.embedding_dims or 512
 
-        self.client = Client(host=self.config.ollama_base_url)
+        # Strip /v1 suffix if present — some users mistakenly set ollama_base_url
+        # to http://host:11434/v1 (OpenAI-style). The Ollama client appends its
+        # own path, so /v1 would result in incorrect endpoints like /v1/api/embed.
+        base_url = self.config.ollama_base_url
+        if base_url:
+            base_url = base_url.rstrip("/")
+            if base_url.endswith("/v1"):
+                base_url = base_url[:-3]
+
+        self.client = Client(host=base_url)
         self._ensure_model_exists()
 
     @staticmethod


kartik-mem0:
Closing this PR based on investigation of the linked issue #4736.

**Findings:**

The TS SDK's `OllamaEmbedder` uses the `ollama` npm package, which correctly calls native Ollama endpoints (`/api/embed`, `/api/tags`) — not `/v1/embeddings`. The code path was verified through:
- Direct code inspection
- Network tracing
- Unit tests

The reported `TypeError: fetch failed` error is a network connectivity issue (connection refused, firewall, etc.), not an endpoint mismatch. If the wrong endpoint were being called, Ollama would return an HTTP 404, not a connection failure.

The fix in this PR (stripping `/v1` suffix) addresses a problem that doesn't exist in the current codebase — the `ollama` npm package never adds `/v1` to begin with.

See the detailed investigation comment on #4736 for more context.

Thank you for the contribution effort!

PR #4798, opened 2026-04-12
fix(graph): skip unnecessary LLM calls when search output or entity map is empty

nanlongyu:
## Description

Summary                                                                                                                                                                                                                                            

- Skip delete-entity LLM call when search_output is empty. _get_delete_entities_from_search_output now returns [] immediately instead of calling the LLM with a blank "existing memories" prompt. This is triggered whenever new content has low similarity to existing graph entities (below the cosine threshold), or on first add with an empty graph.
- Skip the entire downstream pipeline when entity_type_map is empty. When _retrieve_nodes_from_data extracts no entities, add() now returns early, avoiding 2 unnecessary LLM calls, embedding computations, and DB queries that would all produce no useful work.                                                                                                                                                                                                                                    
- Applied to all 4 graph backends: Neo4j, Memgraph, Kùzu, and Neptune.
                                                                                                                                                                                                                                                   
Why               
                                                                                                                                                                                                                                                   
When these intermediate results are empty, the subsequent LLM calls have no useful work to do — there are no existing relationships to evaluate for deletion, and no entities to establish relations for. Continuing the pipeline in this case     
causes two problems:                   
                                                                                                                                                                                                                                                   
1. Wasted cost and latency. Each unnecessary LLM call consumes tokens and adds seconds of round-trip time for a result that is guaranteed to be empty. When entity extraction yields nothing, up to 2 additional LLM calls, multiple embedding     
computations, and DB queries can all be avoided.
2. Potential hang on thinking models. The empty-field prompt (e.g. "Here are the existing memories:  \n\n New Information: ...") can trigger an overthinking loop on reasoning models with thinking/CoT enabled, where the model enters an         
unbounded chain-of-thought and never returns — causing the request to hang indefinitely. Low decoding temperatures (mem0 defaults to 0.1) further amplify the hang rate.                                                                           

## Type of Change

## Breaking Changes

<!-- If this is a breaking change, describe what breaks and the migration path. Delete this section if not applicable. -->

N/A

## Test Coverage

<!-- Describe how you tested this, or link to CI results. -->

## Checklist

### DIFF
--- a/mem0/graphs/neptune/base.py
+++ b/mem0/graphs/neptune/base.py
@@ -64,6 +64,9 @@ class NeptuneBase(ABC):
             filters (dict): A dictionary containing filters to be applied during the addition.
         """
         entity_type_map = self._retrieve_nodes_from_data(data, filters)
+        if not entity_type_map:
+            return {"deleted_entities": [], "added_entities": []}
+
         to_be_added = self._establish_nodes_relations_from_data(data, filters, entity_type_map)
         search_output = self._search_graph_db(node_list=list(entity_type_map.keys()), filters=filters)
         to_be_deleted = self._get_delete_entities_from_search_output(search_output, data, filters)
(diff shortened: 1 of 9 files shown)


kartik-mem0:
hi @nanlongyu graph memory has been deprecated you can have more context about it here. 

https://mem0.ai/research

closing this pr as resolved.

PR #4806, opened 2026-04-13
Fix incorrect use of SentenceTransformer for cross-encoder reranker models

chinnuabey:
## Issue

The default reranker model is a Hugging Face cross-encoder (`cross-encoder/ms-marco-MiniLM-L-6-v2`), but the implementation loads it using `SentenceTransformer`, which is designed for bi-encoder (embedding) models, not cross-encoders.

## Problem

This mismatch can lead to incorrect behavior or scoring during reranking, as cross-encoder models expect query-document pairs to be processed jointly using the `CrossEncoder` API.

## Fix

- Switched model loading from `SentenceTransformer` to `CrossEncoder`
- Updated scoring to use `CrossEncoder.predict()`
- Passed `batch_size` and `show_progress_bar` from config so they are respected during inference

## Behavior

- No change in failure handling
- On prediction failure, documents still fall back to original order with `rerank_score = 0.0`

### DIFF
--- a/mem0/reranker/sentence_transformer_reranker.py
+++ b/mem0/reranker/sentence_transformer_reranker.py
@@ -6,7 +6,7 @@ from mem0.configs.rerankers.base import BaseRerankerConfig
 from mem0.configs.rerankers.sentence_transformer import SentenceTransformerRerankerConfig
 
 try:
-    from sentence_transformers import SentenceTransformer
+    from sentence_transformers import CrossEncoder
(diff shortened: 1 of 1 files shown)


kartik-mem0:
hi @chinnuabey thank you contribution can you please sign the cla without it we can't proceed

thanks

chinnuabey:
I have signed the cla

kartik-mem0 (approved):

PR #4817, opened 2026-04-13
fix(server): harden production Dockerfile — install libpq5, add non-root user, remove --reload

### LINKED ISSUE #4816: fix(server): harden production Dockerfile — missing libpq5, root user, --reload flag
### Summary

The server's production `Dockerfile` has three issues that prevent reliable and secure production deployments:

1. **Missing `libpq5` system library** — `psycopg` (already in `requirements.txt`) requires the native `libpq` library. On `python:3.12-slim`, this isn't included, causing runtime `ImportError` when connecting to PostgreSQL.

2. **Container runs as root** — The Dockerfile doesn't create or switch to a non-root user, meaning the uvicorn process runs as root. This is a security anti-pattern for production containers.

3. **`--reload` in production CMD** — The CMD includes `--reload`, which watches for file changes and restarts the server. This is development-only behavior and should not be in the production image. (The dev compose file already overrides this with its own `--reload` command.)

### Expected Behavior

- The production Docker image should include all required native libraries
- The process should run as a non-root user
- `--reload` should only be used in development (already present in `docker-compose.yaml`)

### Environment

- File: `server/Dockerfile`
- Base image: `python:3.12-slim`

asadk23:
## Linked Issue

Closes #4816

## Description

The server's production `Dockerfile` has three issues that prevent reliable and secure production deployments. This PR fixes all three:

1. **Install `libpq5` system library** — `psycopg` (already in `requirements.txt`) requires the native `libpq` library at runtime. On `python:3.12-slim`, this isn't included, causing `ImportError` when connecting to PostgreSQL.

2. **Run as non-root user** — Creates a `mem0` user (UID 1000) and switches to it via `USER 1000`, so the uvicorn process no longer runs as root.

3. **Remove `--reload` from production CMD** — `--reload` watches for file changes and restarts the server, which is development-only behavior. The dev `docker-compose.yaml` already overrides the CMD with its own `--reload` flag, so this change has no impact on development workflows.

This PR was done with opencode.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Verified by reviewing the diff against the existing `dev.Dockerfile` and `docker-compose.yaml` to confirm:
- `libpq5` is the correct runtime dependency for `psycopg` on Debian slim
- The dev compose file already overrides CMD with `--reload`, so removing it from the production Dockerfile has no impact on development
- The non-root user follows standard container hardening practices

## Checklist

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -15,7 +15,7 @@ COPY . .
 # Create non-root user
 RUN useradd -m -u 1000 mem0
 
-USER 1000
+USER mem0
 
 EXPOSE 8000
 
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -2,14 +2,23 @@ FROM python:3.12-slim
 
 WORKDIR /app
 
-COPY requirements.txt .
+# Install libpq for psycopg PostgreSQL driver
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libpq5 && \
+    rm -rf /var/lib/apt/lists/*
 
+COPY requirements.txt .
 RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . .
+
+# Create non-root user
+RUN useradd -m -u 1000 mem0
+
+USER 1000
 
 EXPOSE 8000
 
 ENV PYTHONUNBUFFERED=1
 
-CMD ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "8000", "--reload"]
+CMD ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "8000"]
--- a/tests/test_server_dockerfile.py
+++ b/tests/test_server_dockerfile.py
@@ -1,0 +1,84 @@
+"""
+Smoke tests for the server production Dockerfile.
+
+Verifies that the built image:
+- Runs as a non-root user (mem0)
+- Does not include --reload in the default CMD
+- Can import psycopg (libpq5 runtime linkage works)
+
+Requires Docker CLI and a running Docker daemon.
+Tests are skipped automatically if Docker is not available.
+"""
+
+import json
+import subprocess
+
+import pytest
+
+IMAGE_NAME = "mem0-api-server:test"
+
+
+def _docker_available():
+    """Check whether the Docker CLI and daemon are reachable."""
+    try:
+        subprocess.run(["docker", "info"], capture_output=True, check=True, timeout=10)
+        return True
+    except (FileNotFoundError, subprocess.CalledProcessError, subprocess.TimeoutExpired):
+        return False
+
+
+requires_docker = pytest.mark.skipif(not _docker_available(), reason="Docker not available")
+
+
+@pytest.fixture(scope="module")
+def built_image():
+    """Build the server Docker image once for all tests in this module."""
+    result = subprocess.run(
+        ["docker", "build", "-t", IMAGE_NAME, "server"],
+        capture_output=True,
+        text=True,
+        timeout=300,
+    )
+    if result.returncode != 0:
+        pytest.skip(f"Docker build failed: {result.stderr[:500]}")
+
+    yield IMAGE_NAME
+
+    # Best-effort cleanup
+    subprocess.run(["docker", "rmi", "-f", IMAGE_NAME], capture_output=True)
+
+
+@requires_docker
+class TestServerDockerfile:
+    def test_runs_as_non_root(self, built_image):
+        """The container should run as the 'mem0' user, not root."""
+        result = subprocess.run(
+            ["docker", "run", "--rm", built_image, "whoami"],
+            capture_output=True,
+            text=True,
+            timeout=30,
+        )
+        assert result.returncode == 0, f"whoami failed: {result.stderr}"
+        assert result.stdout.strip() == "mem0"
+
+    def test_cmd_no_reload(self, built_image):
+        """Production CMD should not contain --reload."""
+        result = subprocess.run(
+            ["docker", "inspect", "--format", "{{json .Config.Cmd}}", built_image],
+            capture_output=True,
+            text=True,
+            check=True,
+            timeout=10,
+        )
+        cmd = json.loads(result.stdout)
+        assert "--reload" not in cmd, f"--reload found in CMD: {cmd}"
+
+    def test_psycopg_imports(self, built_image):
+        """psycopg should import successfully (verifies libpq5 linkage)."""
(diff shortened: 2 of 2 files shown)


xkonjin (commented):
**Review: solid security hardening, one regression risk**

Good improvements: removing `--reload` in production, adding `libpq5`, and switching to a non-root user are all correctly scoped security wins.

One concern: the `USER 1000` directive sets the numeric UID but the `useradd` command creates the user with name `mem0` and UID `1000`. Docker will resolve `USER 1000` to UID 1000, which matches, but for clarity and maintainability `USER mem0` is preferable. More importantly, if the image is ever run in an environment that enforces a read-only root filesystem or specific `securityContext.runAsNonRoot`, `USER 1000` is fine, but ensure any runtime volume mounts have correct permissions for UID 1000.

Coverage gap: there are no Dockerfile tests in this diff. A simple integration test that builds the image and verifies `whoami` returns `mem0` (and that `uvicorn` starts without `--reload`) would prevent future regressions.

Also, `libpq5` is installed but there is no visible check that `psycopg` (or the chosen DB driver) actually loads successfully in the slim image. If the Python package was compiled against `libpq-dev` headers, runtime linkage can still fail. Consider adding a smoke test in CI that imports the DB driver inside the built container.

asadk23:
> **Review: solid security hardening, one regression risk**
> 
> Good improvements: removing `--reload` in production, adding `libpq5`, and switching to a non-root user are all correctly scoped security wins.
> 
> One concern: the `USER 1000` directive sets the numeric UID but the `useradd` command creates the user with name `mem0` and UID `1000`. Docker will resolve `USER 1000` to UID 1000, which matches, but for clarity and maintainability `USER mem0` is preferable. More importantly, if the image is ever run in an environment that enforces a read-only root filesystem or specific `securityContext.runAsNonRoot`, `USER 1000` is fine, but ensure any runtime volume mounts have correct permissions for UID 1000.
> 
> Coverage gap: there are no Dockerfile tests in this diff. A simple integration test that builds the image and verifies `whoami` returns `mem0` (and that `uvicorn` starts without `--reload`) would prevent future regressions.
> 
> Also, `libpq5` is installed but there is no visible check that `psycopg` (or the chosen DB driver) actually loads successfully in the slim image. If the Python package was compiled against `libpq-dev` headers, runtime linkage can still fail. Consider adding a smoke test in CI that imports the DB driver inside the built container.

Hey @xkonjin 
Thanks for the review!

We've added the requested changes, please have a look when you get a chance. Also, comment if it makes sense to change the CI to trigger when changes to `server/**`.

kartik-mem0:
## Changes Requested

### Strengths

- `server/Dockerfile` — `libpq5` install is exactly right: the runtime-only shared library (`libpq.so.5`) on `python:3.12-slim`, using `--no-install-recommends` and cleaning `apt/lists/*` in the same layer to keep the image minimal. This directly fixes the `ImportError` described in the linked issue.
- `server/Dockerfile` — removing `--reload` is unambiguously correct. `dev.Dockerfile` already has `--reload` in its CMD, and `docker-compose.yaml` overrides CMD with its own `--reload` via the `command:` field. The production image was never supposed to have it.
- `tests/test_server_dockerfile.py` — three tests, one per fix: `whoami` for the non-root user, `docker inspect` for the CMD, and `python -c "import psycopg"` for libpq linkage. The `scope="module"` fixture builds the image once. The `_docker_available()` skip guard is correctly placed and matches the convention already used by `tests/test_server_auth.py` and `tests/test_server_params.py`.

### Issues

#### Critical (must fix)

1. `server/Dockerfile:13-16` — **non-root user switches without transferring `/app` ownership, causing startup crash**

   The current PR layer order is:
   ```
   COPY . .                   # /app owned root:root
   RUN useradd -m -u 1000 mem0
   USER mem0
   # ← no chown anywhere
   ```

   `server/main.py:141` calls `initialize_state(DEFAULT_CONFIG)` at module load time. That calls `Memory.from_config()` with `history_db_path="/app/history/history.db"` (`main.py:114`). `SQLiteManager.__init__` (`mem0/memory/storage.py:14`) calls `sqlite3.connect(self.db_path)` directly — no `mkdir`, no try/except around the connect. If `/app/history/` is root-owned and non-writable by the `mem0` user, SQLite raises `OperationalError: unable to open database file` before the server serves any request.

   The dev `docker-compose.yaml` mounts `./history:/app/history` as a volume, which masks the problem during development. `make run_local` also mounts the repo over `/app`. But a bare `docker run mem0-api-server` (the documented production path from `make build`) has no such mount and will crash on startup.

   Fix — add one line before `USER mem0`:
   ```dockerfile
   RUN useradd -m -u 1000 mem0 && \
       mkdir -p /app/history && \
       chown -R mem0:mem0 /app
   USER mem0
   ```
   Or use `COPY --chown=mem0:mem0 . .` (move `useradd` before the COPY in that case).

#### Minor (optional)

1. `server/Dockerfile:15` — `useradd -m -u 1000 mem0` without `--shell /sbin/nologin`. A no-login shell is a small additional hardening step for a service account.

### Evidence

- **CI:** Vercel preview-deploy fail (advisory — fork cannot deploy; not merge-blocking). CLA: pass. No other CI jobs gate this PR — expected for `server/Dockerfile` changes.
- **Targeted tests:** `tests/test_server_dockerfile.py` requires a Docker daemon and correctly skips when one is not available. Could not run `docker build` in this environment — verified by static code tracing instead.
- **Red-green proof:** N/A (Dockerfile integration tests; red-green trace done statically — see below).
- **Static trace confirming the Critical finding:** `main.py:141` → `initialize_state()` → `server_state.py:84` → `Memory.from_config()` → `SQLiteManager.__init__` → `sqlite3.connect("/app/history/history.db")` — no directory creation, no error recovery. The `/app/history/` directory exists on the host (used by compose volume mount) but is not created in the image and is root-owned after `COPY . .`.
- **Staleness check:** current `main` `server/Dockerfile` is identical to the PR base — applies cleanly despite ~2 months since opening.
- **Linked issue:** #4816 (open) — all three reported problems addressed.
- **Duplicate cluster:** PR #5341 touches only `.github/workflows/` (no Dockerfile overlap). No competing hardening PR.
- **Could NOT verify:** actual `docker build` runtime behavior — Docker daemon not available; the ownership failure is traced from `sqlite3.connect()` source.

### Assessment

Ready to merge: **With fixes** — please address the Critical item; one additional line in the Dockerfile is all that's needed. The `libpq5` install and `--reload` removal are correct and ready. The test coverage is well-structured. I'll re-review once you push the `chown` fix.

PR #4828, opened 2026-04-14
fix(ts-oss): isolate entity store from memory store when dbPath unset

chaithanyak42:
## Bug
TS OSS `memory.search()` returns entity rows (fragment text like `"a fan of"`, `"recommending thriller movies and"`) mixed in with real memories. Broken rows have `hash: undefined`, `createdAt: undefined` — they came from the entity store, not the memory pipeline.

## Root cause
`MemoryVectorStore` uses a single `vectors` SQLite table and ignores `collectionName` internally. The entity store at `mem0-ts/src/oss/src/memory/index.ts:193` creates a parallel store with `collectionName: "${name}_entities"` — but both default to the same `dbPath` (`~/.mem0/vector_store.db`). Both writers append to the same `vectors` table; the next search reads both collections.

The existing mitigation at `memory/index.ts:199` (`dbPath.replace(...)`) only fires when `dbPath` is set explicitly. Default (unset) case falls through with no isolation.

## Fix
`getDefaultVectorStoreDbPath(collectionName?)` now derives the default filename from the collection name (`vector_store_${name}.db`). `MemoryVectorStore` passes `config.collectionName` in. Memory vs entity stores now land in separate files automatically.

Explicit `dbPath` users are unaffected — the existing suffix-swap in `memory/index.ts` still handles them. Callers that invoke `getDefaultVectorStoreDbPath()` without arguments still get the previous default (`vector_store.db`), so the helper remains backward-compatible.

## Depends on
Target branch: `feat/v3-pipeline` (contains the v3 ADD-only pipeline + entity linking that introduced this regression). Retarget on merge.

## Test plan
- [ ] Reproducer (messages.add + search with `filters: { user_id: 'alice' }`) → no entity fragments in results
- [ ] `~/.mem0/` contains separate `vector_store_${collection}.db` and `vector_store_${collection}_entities.db` files after add
- [ ] Existing test suite passes (helper is backward-compatible with zero-arg call)


chaithanyak42:
Superseded — will reopen a clean PR against main now that feat/v3-pipeline is merged.

PR #4805, opened 2026-04-13
feat(oss): port v3 pipeline with hybrid search, entity extraction, and additive scoring

soumil-rathi:
## Summary

- Replace the 2-LLM-call add pipeline with single-pass additive extraction using `ADDITIVE_EXTRACTION_PROMPT` (ADD-only, memories accumulate with `linked_memory_ids`)
- Add hybrid search combining semantic + BM25 keyword matching + entity boost with additive scoring: `(semantic + bm25 + entity_boost) / max_possible`
- Add native `keyword_search()` to 15 vector store adapters (Qdrant, Elasticsearch, OpenSearch, Azure AI Search, Weaviate, Redis, PGVector, Pinecone, Databricks, MongoDB, Milvus, Baidu, Upstash, Azure MySQL, Vertex AI)
- Add spaCy-based entity extraction and linking via second vector collection (`{collection}_entities`)
- Add batch embedding, batch persist, batch entity linking (8-phase pipeline) for both sync `Memory` and async `AsyncMemory` at full parity
- Add `embed_batch()` to embedding base class with native OpenAI/Azure overrides
- Add message persistence in SQLite (rolling window of 10 per session scope) for LLM context
- spaCy is an optional dependency (`pip install mem0ai[nlp]`) — system degrades gracefully without it

## Breaking Changes

| Change | Migration |
|--------|-----------|
| `add()` only returns `"ADD"` events (no `"UPDATE"`/`"DELETE"`) | Memories accumulate; accept ADD-only results |
| `search()` default `threshold` changed to `0.1` (was `None`) | Pass `threshold=0.0` for old behavior |
| `search()` `score` is now combined additive score (was raw cosine) | Use `score_breakdown.semantic` for old value |
| `search()` default `rerank` changed to `False` (was `True`) | Pass `rerank=True` for old behavior |
| `qdrant-client>=1.12.0` required (was `>=1.9.1`) | Upgrade qdrant-client |

## Test Coverage

- [x] Added unit tests for lemmatization, entity extraction, scoring (39 tests passing)
- [x] Updated test_main.py for v3 pipeline expectations
- [ ] Integration tests with real vector stores (manual testing needed)
- [x] Tested graceful degradation without spaCy installed

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-review performed
- [x] Tests added that prove the fix/feature works
- [x] New and existing tests pass locally
- [x] Documentation updated (README install instructions, embedding recommendation)

[...]

PR #4829, opened 2026-04-14
fix(ts-oss): isolate entity store from memory store by default

chaithanyak42:
## Bug
TS OSS `memory.search()` returns entity-store rows (fragment text like `"a fan of"`, `"recommending thriller movies and"` with `hash: undefined`, `createdAt: undefined`) mixed in with real memories.

## Root cause
`MemoryVectorStore` uses a single `vectors` SQLite table and ignores `collectionName` internally. The entity store is created as a parallel instance with `collectionName: \`${name}_entities\`` in `memory/index.ts:193` — but both default to the same `dbPath` (`~/.mem0/vector_store.db`). Both writers append to the same `vectors` table; search reads both.

The existing mitigation in `memory/index.ts` (`dbPath.replace(/\.db$/, "_entities.db")`) only fires when `dbPath` is set explicitly. Default (unset) case falls through with no isolation.

## Fix
`getDefaultVectorStoreDbPath(collectionName?)` now derives the default filename from the collection name (`vector_store_${name}.db`). `MemoryVectorStore` passes `config.collectionName` in. Memory vs entity stores now land in separate files automatically.

Backward compatible:
- Explicit `dbPath` users: unchanged, the existing suffix-swap in `memory/index.ts` still handles them.
- Zero-arg `getDefaultVectorStoreDbPath()` callers: still return `vector_store.db`.

## Diff
2 files, 14 insertions, 3 deletions — based on current `main` (post-#4805 merge).

## Test plan
- [ ] Reproducer: `new Memory()` → add messages → `memory.search("query", { filters: { user_id } })` returns only real memories (no fragment entity text)
- [ ] `~/.mem0/` contains separate `vector_store_${collection}.db` and `vector_store_${collection}_entities.db` after add
- [ ] Existing tests pass

### DIFF
--- a/mem0-ts/src/client/tests/integration/batch.test.ts
+++ b/mem0-ts/src/client/tests/integration/batch.test.ts
@@ -26,7 +26,7 @@ describeIntegration("MemoryClient Integration — Batch Operations", () => {
 
   beforeAll(async () => {
     cleanup = suppressTelemetryNoise();
-    client = createTestClient();
+    client = await createTestClient();
     memoryIds = await seedTestMemories(client, TEST_USER_ID);
   });
 
--- a/mem0-ts/src/client/tests/integration/crud.test.ts
+++ b/mem0-ts/src/client/tests/integration/crud.test.ts
@@ -26,9 +26,9 @@ describeIntegration("MemoryClient Integration — CRUD", () => {
   let cleanup: () => void;
   let memoryIds: string[] = [];
 
-  beforeAll(() => {
+  beforeAll(async () => {
     cleanup = suppressTelemetryNoise();
-    client = createTestClient();
+    client = await createTestClient();
   });
 
   afterAll(async () => {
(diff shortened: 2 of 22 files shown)


0xbrainkid:
The root cause analysis is precise — the `collectionName`-derived default filename is the correct fix, and deriving it from the collection name ensures isolation without requiring explicit configuration from users who just want it to work.

One trust dimension worth noting alongside the isolation fix: **entity store contamination is a trust boundary violation in multi-agent setups.** When entity rows bleed into `memory.search()` results, a downstream agent that queries memory gets fragments like `"a fan of"` or `"recommending thriller movies"` that it cannot distinguish from legitimate memories. In a multi-agent pipeline where Agent B queries memory written by Agent A, those fragments look like valid memories from Agent A — the consuming agent has no way to know they are entity store artifacts.

This is the mem0 variant of the shared LLM stop-word contamination bug in CrewAI — implicit shared state between logically separate stores creates behavioral contamination that breaks the assumption that "what I read from memory reflects what was intentionally stored."

Two additions worth considering:

**1. Source field on search results.** Even after the isolation fix, surfacing `source: "memory"` vs `source: "entity"` in search results lets consuming agents apply their own trust policy — some might legitimately want entity results alongside memories, and flagging the distinction lets them decide.

**2. Write provenance.** If writes to the memory store include `writer_agent_id` (which agent wrote this memory), consuming agents can apply trust filtering at read time — ignoring memories from low-trust agents or agents that are not in their delegation chain.

whysosaket (approved):

PR #4836, opened 2026-04-14
fix(v3): migration crashes + entity linking on OSS

soumil-rathi:
## Linked Issue

Surfaced during QA of the v3 beta (`mem0ai==2.0.0b1`, from #4805). No pre-existing issue — fixes here are follow-ups.

## Description

Four distinct bugs found during end-to-end QA of the OSS v3 pipeline. All three changed files are OSS-side only (platform client / prompts / tests untouched).

### 1. Qdrant migration crash on pre-v3 collections (`mem0/vector_stores/qdrant.py`)

**Symptom:** every `add()` raises `ValueError: Wrong input: Not existing vector name error: bm25` for users who upgrade from v1 with `fastembed` installed.

**Root cause:** v3 `create_col` correctly skips creation when a collection already exists, but `insert()` still tries to attach a `bm25` named sparse vector. Pre-v3 collections were created without `sparse_vectors_config`, so Qdrant rejects the entire upsert.

**Fix:** detect the `bm25` slot on init (`get_collection().config.params.sparse_vectors`), cache `self._has_bm25_slot`, gate BM25 writes/reads on it. Semantic search keeps working on legacy collections; a one-time warning tells users to use a fresh collection for hybrid.

### 2. Milvus migration crash, same shape (`mem0/vector_stores/milvus.py`)

**Symptom:** `add()` fails on pre-v3 Milvus collections because v3 always includes a top-level `text` field, which a legacy schema without that field rejects.

**Fix:** same pattern — `describe_collection` on init, cache `_has_bm25_schema = {"text","sparse"} ⊆ fields`, omit `text` from the record dict when absent.

### 3. Entity store silently fails to initialize on embedded Qdrant (`mem0/memory/main.py`)

**Symptom:** with default local `Memory(vector_store={provider: qdrant, path: ...})`, `self._entity_store` stays `None` forever. Entity linking is completely disabled for everyone on the default dev setup. Only visible signal is a `logger.warning` line that's easy to miss.

**Root cause:** the `entity_store` property cloned the vector-store config (including `path=`) and passed it to `VectorStoreFactory.create(...)`, which opens a *second* `QdrantClient(path=...)`. Embedded Qdrant uses RocksDB, which locks the storage directory to one client per process. Second client → rejected → exception swallowed by Phase 7's try/except.

**Fix:** when provider is `qdrant`, reuse `self.vector_store.client` on the cloned entity config. `QdrantConfig.client` takes precedence over `host`/`port`/`path`. Applied to both `Memory.entity_store` and `AsyncMemory.entity_store`.

### 4. `search_batch` dropped payloads, clobbering entity links (`mem0/vector_stores/qdrant.py`)

**Symptom:** on any server-mode Qdrant install (where bug #3 isn't a problem), entities linked to multiple memories only ever surfaced the single most-recently-added memory. Every subsequent `add()` referencing an existing entity erased prior links.

**Root cause:** `search_batch` built `QueryRequest(query=..., filter=..., limit=...)` with no `with_payload=True`. `qdrant-client` defaults it to False. Phase 7's dedup-UPDATE path then read `match.payload = None`, so `set(payload.get("linked_memory_ids", []))` produced `set()`, the "union with new ids" step started from empty, and `set_payload` wrote the new list back, overwriting whatever was there.

**Fix:** one-line — add `with_payload=True` to the `QueryRequest` in `search_batch`.

## Also in this commit

- Drops a misleading comment in `_search_vector_store` that claimed the semantic threshold acts as a gate. In practice `threshold=0.1` is a floor, not a gate, so the comment was actively wrong.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Breaking Changes

None. All changes add defensive feature-detection to existing paths. Fresh v3 installs behave identically to before these fixes.

## Test Coverage

- [x] Tested manually with a scratch QA suite — probes for each bug reproduce before the fix and pass after
- [ ] Added/updated unit tests — deferred; scratch suite is not productized

**Verification, pre-fix vs post-fix:**

| Scenario | Before | After |
|---|---|---|
| Pre-v3 Qdrant collection + fastembed, `add()` | `ValueError: Not existing vector name error: bm25` | succeeds, BM25 off, warning logged |
| Pre-v3 Milvus collection, `add()` | rejects unknown `text` field | succeeds, BM25 off, warning logged |
| Default embedded Qdrant, `add()` then inspect `self._entity_store` | `None` (silently broken) | materialized `Qdrant` instance |
| Remote Qdrant, `add("Sam in Paris")` then `add("Visits Paris")`, inspect Paris entity `linked_memory_ids` | `[mem2_id]` (mem1 erased) | `[mem1_id, mem2_id]` (merged) |
| `AsyncMemory` equivalent of all above | same bugs | same fixes apply |

## Checklist

- [ ] I have added tests that prove my fix/feature works (scratch QA suite, not in repo)
- [x] New and existing tests pass locally (manually re-ran the affected paths)
- [ ] I have updated documentation if needed — release notes for 2.0 GA should mention migration guidance for pre-v3 collections

## Follow-up UX findings (not in this PR)

Discovered alongside the bugs but out of scope here — tracking for separate work:

1. `Memory.get(deleted_id)` returns `None` silently; hosted `MemoryClient.get` raises `MemoryError` for the same case. Inconsistent contract.
2. Signature drift: beta 2.0.0b1's `Memory.search(user_id=...)` rejects top-level entity params and demands `filters={...}`; local (post-#4805) accepts top-level. Opposite conventions between releases — will confuse users on upgrade.
3. Legacy `graph_store` config key is silently ignored. v1 users with graph memory will lose it invisibly on upgrade. Worth a `logger.warning` on unknown config keys.
4. Entity extractor quality (`mem0/utils/entity_extraction.py`) drops single-word proper nouns like "Alex" and concatenates spans across stop-word prepositions ("Dr. Smith at Stanford" → one entity). Pre-existing, affects entity-boost recall.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -993,12 +993,9 @@ export class Memory {
           createdAt: payload.createdAt,
           updatedAt: payload.updatedAt,
           score: scored.score,
-          metadata: {
-            ...Object.entries(payload)
-              .filter(([key]) => !excludedKeys.has(key))
-              .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {}),
-            scoreBreakdown: scored.scoreBreakdown,
-          },
+          metadata: Object.entries(payload)
+            .filter(([key]) => !excludedKeys.has(key))
+            .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {}),
           ...(payload.user_id && { user_id: payload.user_id }),
           ...(payload.agent_id && { agent_id: payload.agent_id }),
           ...(payload.run_id && { run_id: payload.run_id }),
--- a/mem0-ts/src/oss/src/utils/scoring.ts
+++ b/mem0-ts/src/oss/src/utils/scoring.ts
@@ -59,11 +59,6 @@ export function normalizeBm25(
 export interface ScoredResult {
   id: string;
   score: number;
-  scoreBreakdown: {
-    semantic: number;
-    bm25: number;
-    entityBoost: number;
-  };
   payload: Record<string, any>;
 }
 
@@ -134,11 +129,6 @@ export function scoreAndRank(
     scored.push({
       id: memIdStr,
       score: combined,
-      scoreBreakdown: {
-        semantic: semanticScore,
-        bm25: bm25Score,
-        entityBoost: entityBoost,
-      },
       payload: result.payload,
     });
   }
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1240,9 +1240,6 @@ class Memory(MemoryBase):
                 score=scored["score"],
             ).model_dump()
 
-            # Add score breakdown to metadata
-            memory_item_dict["score_breakdown"] = scored.get("score_breakdown", {})
-
             for key in promoted_payload_keys:
                 if key in payload:
                     memory_item_dict[key] = payload[key]
@@ -2493,8 +2490,6 @@ class AsyncMemory(MemoryBase):
                 updated_at=payload.get("updated_at"),
                 score=scored["score"],
             ).model_dump()
-
-            memory_item_dict["score_breakdown"] = scored.get("score_breakdown", {})
 
             for key in promoted_payload_keys:
                 if key in payload:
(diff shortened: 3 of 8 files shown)


chaithanyak42 (approved):

kartik-mem0:
2. for this point we have added it later in the v3 algo pr. were we are dropping top level entity acceptance, for getall and search and only accepting filters in it.

cc: @whysosaket @deshraj

whysosaket (approved):

PR #4837, opened 2026-04-14
Self-hosted dashboard and admin auth

gabrielstein-mem0:
## Walkthrough

PR Video Walkthrough: https://www.loom.com/share/3fb844e90c7b4f78a75f2df29de5ecd6

## Screenshot
Screenshot 2026-04-17 at 5 49 22 PM
Screenshot 2026-04-17 at 5 49 45 PM
Screenshot 2026-04-17 at 5 48 02 PM
Screenshot 2026-04-17 at 5 48 11 PM
Screenshot 2026-04-17 at 5 47 53 PM

## What's new

**Backend**
- JWT + per-user API keys (`m0sk_...`, bcrypt hashed, shown once).
- Four auth modes in precedence: bearer JWT → per-user `X-API-Key` → legacy `ADMIN_API_KEY` → `AUTH_DISABLED` (dev only).
- Alembic migrations for `users`, `api_keys`, `request_logs`, `settings`.
- Request logging middleware captures memory, auth, and admin traffic; the dashboard's audit log surfaces the API-key subset.
- New routers: `/auth/*` (including `PATCH /auth/me` and `POST /auth/change-password`), `/api-keys`, `/requests`, `/entities`.
- `GET /configure/providers` lists the LLM and embedder providers bundled in the container. `POST /configure` rejects anything outside that list with a 400 that explains how to extend the image.
- Bundled providers: OpenAI, Anthropic, Gemini (LLM + embedder). Heavy providers like `sentence-transformers`/`torch` are deliberately out of the default image.
- Provider failures return a generic `502 "Upstream provider error."` with the full trace in server logs; the server no longer echoes raw exception strings (or API keys embedded in them) to clients.
- Default models configurable via `MEM0_DEFAULT_LLM_MODEL` and `MEM0_DEFAULT_EMBEDDER_MODEL`.
- Graph memory and team/invite scaffolding removed from `server/` (graph memory still works as a library feature).

**Dashboard**
- `/setup` wizard: admin account → config review → first API key → use case → test curl. Test failures now surface the server's response body and a link back to Configuration.
- Pages: **Requests** (default, audit log filtered to API-key traffic), **Memories**, **Entities** (groups by `user_id` / `agent_id` / `run_id` with cascade delete), **API Keys**, **Configuration** (provider dropdown populated from `/configure/providers`), **Settings** (editable profile + password change).
- Login page: "Forgot password?" link opens a dialog with the `make reset-admin-password` recovery command.
- JWT login with refresh, middleware gating, logout.

**Docs**
- New page: [`docs/open-source/setup.mdx`](docs/open-source/setup.mdx) — happy path, required env (incl. `JWT_SECRET`), wizard walkthrough, supported-providers list with an extension recipe, upgrade-from-1.x notes.
- [`rest-api.mdx`](docs/open-source/features/rest-api.mdx) rewritten around the four auth modes; now documents `PATCH /auth/me`, `POST /auth/change-password`, and `GET /configure/providers`.
- [`overview.mdx`](docs/open-source/overview.mdx) leads with self-hosted setup alongside the library quickstarts.

**Security hardening**
- Server refuses to start if `JWT_SECRET` is unset and `AUTH_DISABLED` is false (previously only 500'd on first login).
- `docker-compose.yaml` no longer supplies a fallback `JWT_SECRET` — missing `.env` now surfaces immediately.
- `make bootstrap` generates a random admin password if `PASSWORD` is unset (was hardcoded to `admin123456`).

**OSS attribution**
- Every `app.mem0.ai` link across the repo (README, SDK docs, integration guides, cookbooks, CLI output, dashboard CTAs) carries `utm_source=oss` + a per-surface `utm_medium`. Fixes the PostHog "direct traffic" problem for clicks that originate in terminals, IDEs, and the self-hosted dashboard — none of which send a useful HTTP Referer. 104 URLs across 71 files; patcher is idempotent.

## What breaks

- **Auth is on by default.** Fresh installs redirect to `/setup`. Existing deployments running with an empty `ADMIN_API_KEY` will 401 until they set a key, register an admin, or set `AUTH_DISABLED=true`. The server prints a loud startup banner in that state with all three recovery options.
- **`/configure` is gated to bundled providers.** Previously any provider string was accepted and failed later with an opaque 500. Admins who rely on non-bundled providers must install the package, rebuild the image, and extend `BUNDLED_LLM_PROVIDERS` / `BUNDLED_EMBEDDER_PROVIDERS` in `server/main.py`.

## Try it

```bash
cd server
make bootstrap    # starts the stack, creates the admin, issues the first api key
```

Then http://localhost:3000 → log in → sidebar has Requests, Memories, Entities, API Keys, Configuration, Settings.


kartik-mem0:
Screenshot 2026-04-21 at 3 36 53 PM



hi @gabrielstein-mem0, in this request page the empty page is not loading there is some issue there.

also add email regex here i am able to login with any random email id

rest lgtm

Screenshot 2026-04-21 at 3 28 51 PM


also please make sure we follow this as well.

kartik-mem0:
also address this merge conficts please ;)

thanks

kartik-mem0:
Screenshot 2026-04-21 at 3 38 25 PM
also the curl which is provided here is not working please fix this

server/routers/auth.py
```diff
@@ -0,0 +1,190 @@
...
+    name: str
+    email: EmailStr
+    password: str
+
+
+class OnboardingCompleteRequest(BaseModel):
+    use_case: str
+
+
+class LoginRequest(BaseModel):
```
whysosaket, line 49: **Bug / Security: No password validation on registration**

`MIN_PASSWORD_LENGTH = 8` is enforced in `change_password` (line ~162) but is completely absent here. A 1-character password is accepted at registration. Add the same length check before hashing:

```python
if len(body.password) < MIN_PASSWORD_LENGTH:
    raise HTTPException(status_code=400, detail=f"Password must be at least {MIN_PASSWORD_LENGTH} characters.")
```

whysosaket (commented):
## Review Summary

Major issues flagged inline below. Highlights:
- Missing password validation on registration (security)
- Registration race condition (security)
- No rate limiting on login (security)
- Refresh tokens never invalidated (security)
- Shell injection in entrypoint.sh (security)
- Sync DB writes inside async middleware (perf blocker)
- Unbounded `request_logs` table growth (ops risk)
- Race condition in `update_config` (data loss)
- `delete_entity` leaks internal exception details

server/routers/auth.py
```diff
@@ -0,0 +1,190 @@
...
+
+@router.get("/setup-status", response_model=SetupStatusResponse)
+def setup_status(db: Session = Depends(get_db)):
+    count = db.scalar(select(func.count(User.id)))
+    return SetupStatusResponse(needsSetup=count == 0)
+
+
+@router.post("/register", response_model=TokenResponse)
+def register(body: RegisterRequest, db: Session = Depends(get_db)):
+    """Create the first admin account. Blocked once any user exists."""
```
whysosaket, line 97: **Security: No password validation on registration**

`MIN_PASSWORD_LENGTH` is defined at line 12 and checked in `change_password`, but there's no length check here. A 1-character password will be accepted and hashed.

```python
if len(body.password) < MIN_PASSWORD_LENGTH:
    raise HTTPException(status_code=400, detail=f"Password must be at least {MIN_PASSWORD_LENGTH} characters.")
```

Add this before the `hash_password` call.

server/routers/auth.py
```diff
@@ -0,0 +1,190 @@
...
+    needsSetup: bool
+
+
+@router.get("/setup-status", response_model=SetupStatusResponse)
+def setup_status(db: Session = Depends(get_db)):
+    count = db.scalar(select(func.count(User.id)))
+    return SetupStatusResponse(needsSetup=count == 0)
+
+
+@router.post("/register", response_model=TokenResponse)
```
whysosaket, line 94: **Security: Race condition — two concurrent registrations can both become admin**

Both requests can pass this `count == 0` check simultaneously before either commits. The fix is a unique constraint + handle the integrity error:

```python
# In the migration, add a partial unique index or handle IntegrityError:
try:
    db.add(user)
    db.commit()
except IntegrityError:
    db.rollback()
    raise HTTPException(status_code=403, detail="Registration is closed. An admin account already exists.")
```

Or use `SELECT ... FOR UPDATE` / a DB-level unique constraint on role.

server/routers/auth.py
```diff
@@ -0,0 +1,190 @@
...
+    db.add(user)
+    db.commit()
+    db.refresh(user)
+
+    capture_admin_registered(email=body.email)
+
+    return TokenResponse(
+        access_token=create_access_token(str(user.id), user.role),
+        refresh_token=create_refresh_token(str(user.id)),
+    )
```
whysosaket, line 122: **Security: No rate limiting on login — trivially brute-forceable**

There is no throttling, lockout, or backoff on this endpoint. An attacker can try passwords at the rate bcrypt allows (which is the only brake). At a minimum, add per-IP or per-email rate limiting using something like `slowapi` or `fastapi-limiter`.

Also: the refresh endpoint (`/auth/refresh`) and register endpoint have the same problem.

server/routers/auth.py
```diff
@@ -0,0 +1,190 @@
...
+    if user is None:
+        dummy_verify_password()
+        raise HTTPException(status_code=401, detail="Invalid email or password.")
+    if not verify_password(body.password, user.password_hash):
+        raise HTTPException(status_code=401, detail="Invalid email or password.")
+
+    user.last_login_at = datetime.now(timezone.utc)
+    db.commit()
+
+    return TokenResponse(
```
whysosaket, line 138: **Security: Refresh tokens are never invalidated — stolen token grants indefinite access**

When a new token pair is issued here, the old refresh token remains valid for its full 30-day lifetime. There's no blacklist, no rotation invalidation, and no way to force-logout a session.

At minimum, implement single-use refresh tokens: store a `jti` (JWT ID) in the DB, mark it as used on first use, and reject reuse. Without this, a stolen refresh token cannot be revoked short of changing `JWT_SECRET` (which logs out all users).

server/dashboard/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Swap NEXT_PUBLIC_* placeholders baked into .next/ at build time with the
+# container's real env values. Each placeholder must be declared in the
+# Dockerfile as `ENV NAME=NAME` so Next.js inlines the literal name. Currently:
+#   NEXT_PUBLIC_API_URL, NEXT_PUBLIC_INSTANCE_NAME
+# New NEXT_PUBLIC_* vars must be added to the Dockerfile or this loop won't substitute them.
+printenv | grep NEXT_PUBLIC_ | while read -r line ; do
+  key=$(echo $line | cut -d "=" -f1)
+  value=$(echo $line | cut -d "=" -f2)
+
```
whysosaket: **Security: Shell injection — unquoted env vars passed directly to `sed`**

If a `NEXT_PUBLIC_` env var's value contains `|`, `/`, newlines, or other sed metacharacters, this silently corrupts `.next/` files or could be exploited. The `$key` and `$value` substitutions are unescaped.

Safer alternatives:
- Use `envsubst` for placeholder replacement, or
- Escape the value before passing to `sed`: `value=$(printf '%s' "$value" | sed 's/[&/\\]/\\&/g')`

Also: the `cut -d "=" -f2` on line 13 truncates values that contain `=` (e.g. base64 strings). Use `cut -d "=" -f2-` instead.

server/main.py
```diff
@@ -127,115 +180,262 @@ class SearchRequest(BaseModel):
...
+
 @app.post("/configure", summary="Configure Mem0")
-def set_config(config: Dict[str, Any], _api_key: Optional[str] = Depends(verify_api_key)):
+def set_config(config: Dict[str, Any], _auth=Depends(verify_auth)):
     """Set memory configuration."""
-    global MEMORY_INSTANCE
-    MEMORY_INSTANCE = Memory.from_config(config)
+    _validate_bundled_providers(config)
+    update_config(config)
     return {"message": "Configuration set successfully"}
```
whysosaket, line 305: **Performance: Synchronous DB write inside `async` middleware — blocks the event loop**

`_persist_request_log()` opens a `SessionLocal()` and does a synchronous SQLAlchemy commit on every single request, inside an `async def` middleware. This blocks the entire Uvicorn event loop for the duration of the DB round-trip, effectively serialising all requests at that point.

Options:
1. Use `asyncio.get_event_loop().run_in_executor(None, _persist_request_log, ...)` to offload to a thread pool
2. Switch to async SQLAlchemy (`sqlalchemy.ext.asyncio`)
3. Use a fire-and-forget `BackgroundTask`

server/alembic/versions/002_create_request_logs.py
```diff
@@ -0,0 +1,36 @@
...
+
+revision: str = "002"
+down_revision: Union[str, None] = "001"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "request_logs",
```
whysosaket, line 22: **Ops risk: `request_logs` table will grow unbounded**

Every non-skipped request writes a row here with no TTL, no cleanup job, and no partitioning. On a moderately active server (even 10 req/s) this hits 864k rows/day. The only index is on `created_at`, but without a purge strategy the table will eventually degrade query performance and fill the disk.

Suggest adding either:
- A scheduled cleanup (e.g. `DELETE FROM request_logs WHERE created_at < NOW() - INTERVAL '30 days'`)
- A `BRIN` index instead of `BTREE` for append-only time-series data
- Or document the expected retention/rotation approach

server/server_state.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def initialize_state(default_config: Dict[str, Any]) -> None:
+    global _current_config, _memory_instance
+    with _state_lock:
+        _current_config = deepcopy(default_config)
+        overrides = _load_overrides()
+        if overrides:
+            _current_config = _merge_config(_current_config, overrides)
+        _memory_instance = Memory.from_config(_current_config)
```
whysosaket, line 83: **Bug: Race condition — concurrent `update_config` calls can lose each other's writes**

The lock is released after updating in-memory state but before the DB write. A second concurrent call to `update_config` can call `_load_overrides()` before the first call's `_save_overrides()` completes, then overwrite the DB with a stale merge.

The DB I/O should happen inside the lock (or use a separate per-operation serialisation strategy), even if it's slower:

```python
with _state_lock:
    next_config = _merge_config(_current_config, updates)
    _current_config = next_config
    _memory_instance = Memory.from_config(next_config)
    overrides = _load_overrides()
    overrides = _merge_config(overrides, updates)
    _save_overrides(overrides)
    return deepcopy(_current_config)
```

server/routers/entities.py
```diff
@@ -0,0 +1,78 @@
...
+        for (entity_type, entity_id), data in sorted(buckets.items(), key=lambda item: (item[0][0], item[0][1]))
+    ]
+
+
+@router.delete("/{entity_type}/{entity_id}", response_model=MessageResponse)
+def delete_entity(entity_type: EntityType, entity_id: str, _auth=Depends(verify_auth)):
+    try:
+        get_memory_instance().delete_all(**{TYPE_TO_FIELD[entity_type]: entity_id})
+    except Exception as e:
+        logging.exception("Error deleting entity")
```
whysosaket: **Bug: Exception detail leaks internal error message + inconsistent with rest of codebase**

Every other endpoint uses `raise _upstream_error()` (defined in `main.py`) which logs the traceback server-side and returns a generic `502`. This endpoint uses `detail=str(e)` which exposes raw exception strings (potentially including DB connection strings, SQL, stack traces) to the API caller.

```python
# Replace:
raise HTTPException(status_code=500, detail=str(e))
# With:
raise _upstream_error()
```

Note: `_upstream_error` lives in `main.py` — consider moving it to a shared module so routers can import it cleanly.

gabrielstein-mem0:
Thanks, both fixed in `7b8fc3dc`:

- **Empty Requests page:** the `EmptyState` component was looking for `/images/no-requests.svg` (plural), but the file was `no-request.svg` (singular) — the image 404'd and the empty state rendered blank. Renamed both SVGs to match the `image` prop naming convention the memories variant already uses.
- **Email regex:** `type="email"` in browsers accepts things like `foo@bar` with no TLD, and the Pydantic error just surfaced as a generic "Login failed." Added a shared `isValidEmail` validator that runs before submit on both the login form and the setup wizard, so invalid addresses get rejected immediately with a clear message.

gabrielstein-mem0:
Fixed in `7b8fc3dc`. Three things were going wrong at once:

1. `entrypoint.sh` used `cut -d "=" -f2`, which truncates any `NEXT_PUBLIC_*` value containing an `=`, and passed the value unquoted into `sed`, which corrupts values with `|` or `/`. So `NEXT_PUBLIC_API_URL` could end up empty or mangled inside the built bundle — which renders the curl as `curl -X POST /memories …` and fails. Switched to `IFS='=' read -r key value` and escape sed metacharacters first.
2. The displayed curl sent `"user_id": "test"` while the Run Test button sent `"user_id": "setup-test"`. Unified them.
3. Added a warning block above the curl when `NEXT_PUBLIC_API_URL` is empty, so a missing env var surfaces explicitly instead of producing a silently-broken curl.

Verified end-to-end against a live stack: ran the displayed curl against `/memories` with a real API key → HTTP 200 + memory created.

server/routers/auth.py
gabrielstein-mem0: Fixed in `64df04c1`. Pulled the length check into a small `_require_password_length` helper and call it from both `/auth/register` and `/auth/change-password`, so the two paths share a single rule. Verified end-to-end: register with 1-char → HTTP 400, 7-char → 400, 8-char → 200.

server/routers/auth.py
gabrielstein-mem0: Same helper covers this — see reply on the earlier thread. Fixed in `64df04c1`.

server/routers/auth.py
gabrielstein-mem0: Fixed in `64df04c1`. Migration `004_unique_admin_role` adds a partial unique index (`role = 'admin'`) on `users`, and `register()` now catches `IntegrityError` and returns the same 403 "Registration is closed." DB-level uniqueness is the only reliable guarantee here — the `count == 0` check stays as a fast path for the common case. Verified live on a fresh DB: registering a second admin returns HTTP 403.

server/routers/auth.py
gabrielstein-mem0: Fixed in `64df04c1`. Added `slowapi` with per-IP limits: `/login` 10/min, `/register` 5/min, `/refresh` 20/min. The limiter lives in `rate_limit.py` so routers can import it without a `main.py` cycle. Storage is in-memory (single-process) for now; a Redis-backed limiter for multi-worker deployments is noted as a follow-up. Verified live: 11 bad logins from one IP → attempts 1–10 return 401, attempt 11 flips to 429.

server/routers/auth.py
gabrielstein-mem0: Fixed in `1349213e` and tightened further in `8a31c51c`. Migration `005_refresh_token_jtis` adds a `refresh_token_jtis` table; every refresh token now carries a `jti` claim. `/auth/refresh` runs a single atomic `UPDATE ... WHERE jti = ? AND used_at IS NULL AND expires_at > now` — replays see `rowcount = 0` and return 401, so a stolen token is single-use. Explicit session revoke / "log out everywhere" is tracked as a follow-up. Verified live: first refresh 200, replay with same token 401.

server/dashboard/entrypoint.sh
gabrielstein-mem0: Fixed in `7b8fc3dc`. Switched to `IFS='=' read -r key value` so values containing `=` survive, and escape `\`, `&`, `|` before passing to `sed`. Side benefit: this was also the root cause of the broken setup curl Kartik flagged — `NEXT_PUBLIC_API_URL` was being truncated at container startup. Verified live: `NEXT_PUBLIC_FOO=a=b=c` and `NEXT_PUBLIC_BAR=x|y/z&w` both round-trip into `.next/` unchanged; in a real build the dashboard bundle now contains `http://localhost:8888` with zero hits for the literal placeholder.

server/main.py
gabrielstein-mem0: Fixed in `1d6caf11`. `_persist_request_log` now takes primitives (method, path, status, latency, auth_type) and runs via `loop.run_in_executor(None, ...)` from the middleware's `finally` block. Capturing primitives first keeps the request object out of the background task's lifetime, and the hot path becomes fully non-blocking.

server/alembic/versions/002_create_request_logs.py
gabrielstein-mem0: Fixed in `1d6caf11`. Three pieces:

- Migration `006_request_logs_brin` swaps the btree on `created_at` for a BRIN index — same name, right shape for append-only time-series, fraction of the storage.
- New `scripts/prune_request_logs.py` deletes rows older than `REQUEST_LOG_RETENTION_DAYS` (default 30), wired as `make prune-logs`.
- Documented the retention recipe in `server/README.md` with a cron / systemd-timer pointer. We're not shipping a timer in this PR — operators wire it to whatever scheduler they already run.

Verified live: `\d request_logs` shows the BRIN index; `make prune-logs` runs against the live DB with default and with `REQUEST_LOG_RETENTION_DAYS=<N>` override.

server/server_state.py
gabrielstein-mem0: Fixed in `1d6caf11`. Moved `_load_overrides` and `_save_overrides` inside the `_state_lock` block and removed the "DB I/O outside the lock" comment. Config writes are rare enough that the throughput hit is fine; the previous arrangement allowed silent last-write-wins on interleaved `/configure` calls, which isn't worth the optimization.

server/routers/entities.py
gabrielstein-mem0: Fixed in `1d6caf11`. Pulled `_upstream_error` out of `main.py` into a shared `server/errors.py` as `upstream_error()`; `delete_entity` now does `raise upstream_error()` on exception. Every other router was already using this pattern.

kartik-mem0 (commented):
Image

can you please verify this once

server/telemetry.py
```diff
@@ -0,0 +1,112 @@
...
+import os
+import uuid
+from datetime import datetime, timezone
+from pathlib import Path
+from threading import Lock
+from typing import Any
+
+import mem0
+
+PROJECT_API_KEY = "phc_hgJkUVJFYtmaJqrvf6CYN67TIQ8yhXAkWzUn9AMU4yX"
```
kartik-mem0, line 24: why are we passing api key here directly?

gabrielstein-mem0: It's the PostHog project API key (phc_*) — public by design, write-only, safe to embed in source. Same key is hardcoded in every other telemetry file in this repo:

mem0/memory/telemetry.py:15
mem0-ts/src/oss/src/utils/telemetry.ts
mem0-ts/src/client/telemetry.ts
cli/python/src/mem0_cli/telemetry.py:22
cli/node/src/telemetry.ts
openclaw/telemetry.ts
It's intentional: every OSS install phones home to the same PostHog project so we see aggregate usage. If users could override it via env var the data would fragment across tenants and be useless. The real opt-out knob is MEM0_TELEMETRY=false — documented in the module docstring and .env.example.

kartik-mem0 (dismissed):

kartik-mem0 (dismissed):

chaithanyak42 (dismissed):

whysosaket (approved):

PR #4840, opened 2026-04-15
docs: recommend MemoryService for Google ADK integration

Protocol-zero-0:
## Summary
This updates the Google ADK integration guide to recommend the native `MemoryService` pattern for Mem0 before the existing tool-based approach. The new section explains why `MemoryService` plus `PreloadMemoryTool` is the idiomatic ADK path and shows how to wire a `Mem0MemoryService` into `Runner`.

The existing function-tool workflow is kept as an explicit alternative so readers still have a manual-control option when they need it.

## Validation
I reviewed the updated doc against the current ADK memory flow (`PreloadMemoryTool`, `add_session_to_memory`, and `Runner(memory_service=...)`) and checked the edited file for lint issues.


kartik-mem0:

PR #4842, opened 2026-04-15
docs: add Neon vector database guide

Protocol-zero-0:
## Summary
This adds a dedicated Neon vector database page under the Mem0 vector database docs and links it from the vector DB overview. The guide explains that Neon should be used through the existing `pgvector` provider and includes Python and TypeScript examples using a Neon connection string with `sslmode=require`.

This keeps the docs aligned with the existing vector DB template while giving users a clear managed-Postgres path for Mem0.

## Validation
- `git diff --check`
- checked the edited docs files for lint issues


kartik-mem0:

PR #4852, opened 2026-04-15
fix: merge same-key AND conditions in _process_metadata_filters

### LINKED ISSUE #4850: Data Loss in `_process_metadata_filters` with `AND` operator
### Component

Core / Python SDK

### Summary

The `AND` operator in metadata filters overwrites conditions if multiple conditions apply to the same key within the same `AND` block. This causes all but the last condition for a given key to be silently dropped, leading to incorrect search filters being passed down to the vector databases.

### Steps to Reproduce

```python

# Bypass full initialization to avoid needing API keys for this test
m = Memory.__new__(Memory)

# We want a condition where price > 10 AND price < 20
filters = {"AND": [{"price": {"gt": 10}}, {"price": {"lt": 20}}]}

# Process the filters internally
processed_filters = m._process_metadata_filters(filters)
print(processed_filters)
```

### Expected Behavior

The `processed_filters` should merge constraints on the same key so both bounds are kept:
`{'price': {'gt': 10, 'lt': 20}}`.

### Actual Behavior

The second condition completely overwrites the first one because `_process_metadata_filters` internally uses `dict.update()` without checking if the keys already exist as nested operators.
Output:
`{'price': {'lt': 20}}`

### Environment

- mem0 version: 2.0.0b1
- OS: Linux

CrepuscularIRIS:
## Linked Issue

Closes #4850

## Description

`_process_metadata_filters` silently drops conditions when the `AND` operator contains multiple conditions on the same key. For example:

```python
{"AND": [{"price": {"gt": 10}}, {"price": {"lt": 20}}]}
```

**Before (bug):** `{"price": {"lt": 20}}` — the `gt` condition is lost  
**After (fix):** `{"price": {"gt": 10, "lt": 20}}` — both conditions preserved

### Root Cause

The AND handler uses `dict.update()` to accumulate processed conditions into `processed_filters`. When two conditions share the same key, the second call to `update()` overwrites the first entirely, even though the values are nested operator dicts that should be merged.

### Fix

Added a `_merge_condition` helper that deep-merges nested operator dicts for the same key instead of overwriting. Applied to both the AND handler and the top-level else branch, in both `Memory` and `AsyncMemory` classes.

### Regression note

PR #4559 (merged 2026-03-26) fixed the related case where operators within a **single** condition dict were not merged (e.g. `{"created_at": {"gte": X, "lte": Y}}`). This PR fixes the remaining case where the same key appears in **separate** dicts within an AND list.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

### Tests added

3 new regression tests in `TestProcessMetadataFiltersMerge`:
- `test_and_same_key_separate_conditions_merged` — two conditions on same key
- `test_and_same_key_three_conditions_merged` — three conditions on same key
- `test_and_mixed_same_and_different_keys` — mixed same-key and different-key conditions

### Manual verification

```python
m = Memory.__new__(Memory)
result = m._process_metadata_filters({"AND": [{"price": {"gt": 10}}, {"price": {"lt": 20}}]})
assert result == {"price": {"gt": 10, "lt": 20}}  # passes
```

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1193,6 +1193,14 @@ class Memory(MemoryBase):
                     raise ValueError(f"Unsupported metadata filter operator: {operator}")
             return result
 
+        def _merge_condition(target: Dict[str, Any], source: Dict[str, Any]) -> None:
+            """Merge source into target, combining nested operator dicts for the same key."""
+            for k, v in source.items():
+                if k in target and isinstance(target[k], dict) and isinstance(v, dict):
+                    target[k].update(v)
+                else:
+                    target[k] = v
+
         for key, value in metadata_filters.items():
             if key == "AND":
                 # Logical AND: combine multiple conditions
@@ -1200,7 +1208,7 @@ class Memory(MemoryBase):
                     raise ValueError("AND operator requires a list of conditions")
                 for condition in value:
                     for sub_key, sub_value in condition.items():
-                        processed_filters.update(process_condition(sub_key, sub_value))
+                        _merge_condition(processed_filters, process_condition(sub_key, sub_value))
             elif key == "OR":
                 # Logical OR: Pass through to vector store for implementation-specific handling
                 if not isinstance(value, list) or not value:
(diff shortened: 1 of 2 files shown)


CrepuscularIRIS:
Closing in favor of #4853 which provides a more complete fix. Apologies for the duplicate.

PR #4855, opened 2026-04-15
fix(openclaw): fail-closed on JSON parse error in readFullConfig

### LINKED ISSUE #4849: Plugin silently destroys entire OpenClaw config on JSON parse error
### Component

OpenClaw

### Summary

When readFullConfig() encounters any JSON parse error or returns an empty object, the plugin proceeds to overwrite the entire openclaw.json with only the mem0
plugin entry, destroying all other configuration (models, providers, channels, auth, etc.).

### Problem Code (Pinpointed via Openclaw)

```javascript
  function readFullConfig() {
    if (exists(OPENCLAW_CONFIG_FILE)) {
      try {
        return JSON.parse(readText(OPENCLAW_CON FIG_FILE));
      } catch {
        // Empty catch - silently returns undefined, falls through to {}
      }
    }
    return {};  // Returns empty object on any error
  }
```

### Expected Behavior

For it to either:
- Pinpoint the issue and fix it properly
- Leave the config as-is and leave the user to fix it

### Actual Behavior

Deletes ENTIRE config and replaces it with the plugin config entry on its own inside openclaw.json

### Environment

- mem0 version: @mem0/openclaw-mem0 1.0.5 (also affects 1.0.6 based on source)
- OS: Arch Linux
- Openclaw Version: 2026.4.14

### Suggested Fix (Self diagnosis from Openclaw)

```javascript
  function readFullConfig() {
    if (exists(OPENCLAW_CONFIG_FILE)) {
      try {
        const text = readText(OPENCLAW_CONFIG_FILE);
        const parsed = JSON.parse(text);
        if (!parsed || typeof parsed !== 'object' || Object.keys(parsed).length === 0) {
          throw new Error('Config appears empty after parse');
        }
        return parsed;
      } catch (err) {
        console.error('[openclaw-mem0] Failed to read config:', err);
        throw err; // Fail closed
      }
    }
    return {};
  }
```

WARNING: IF SUGGESTED CODE ABOVE IS USED, I, ALONG WITH THE LLM MODEL AND AGENTIC HARNESS THAT PRODUCED THIS, ARE NOT TO BE HELD ACCOUNTABLE FOR ANY DAMAGES.

jessai2026:
## Linked Issue

Closes #4849

## Description

`readFullConfig()` was silently swallowing any JSON parse error by catching the exception and falling through to `return {}`. When `writePluginAuth()` or `writePluginConfigField()` subsequently called this function, they received an empty object, built a brand-new config structure containing only the plugin entry, and wrote that back to disk — **destroying every other setting in `openclaw.json`** (models, providers, channels, auth tokens, etc.).

### Root cause

```typescript
// Before: silent catch ⬇ falls through to return {}
function readFullConfig(): Record<string, unknown> {
  if (exists(OPENCLAW_CONFIG_FILE)) {
    try {
      return JSON.parse(readText(OPENCLAW_CONFIG_FILE));
    } catch {
      /* ignore parse errors */   // ← swallows the error
    }
  }
  return {};  // ← write callers treat this as "empty config" and overwrite everything
}
```

### Fix

Throw a descriptive error instead of returning `{}`. Write callers (`writePluginAuth`, `writePluginConfigField`) will now abort rather than proceeding with a blank slate. Also guard against non-object JSON values (arrays, strings, numbers) which cause the same overwrite.

```typescript
// After: fail-closed
function readFullConfig(): Record<string, unknown> {
  if (exists(OPENCLAW_CONFIG_FILE)) {
    const text = readText(OPENCLAW_CONFIG_FILE);
    let parsed: unknown;
    try {
      parsed = JSON.parse(text);
    } catch (err) {
      throw new Error(
        `[openclaw-mem0] Failed to parse ${OPENCLAW_CONFIG_FILE}: ${(err as Error).message}. ` +
          `Fix the file manually or delete it to start fresh.`,
      );
    }
    if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
      throw new Error(
        `[openclaw-mem0] ${OPENCLAW_CONFIG_FILE} does not contain a JSON object. ` +
          `Fix the file manually or delete it to start fresh.`,
      );
    }
    return parsed as Record<string, unknown>;
  }
  return {};
}
```

## Type of Change

## Breaking Changes

Any code that previously relied on `readPluginAuth()` returning `{}` on a corrupted config will now receive a thrown error instead. This is intentional — the previous behaviour caused data loss.

## Test Coverage

Updated the existing `"returns empty object when JSON is invalid"` test (which validated the broken behaviour) to `"throws when config file contains invalid JSON"`. Added:
- `"throws when config file contains a non-object JSON value"` — guards against arrays/strings/numbers
- `"throws and does NOT overwrite config when JSON is invalid"` (in `writePluginAuth`) — regression test that directly verifies `writeText` is never called when the config is corrupted

All 354 tests pass (`pnpm run test`).

## Checklist

### DIFF
--- a/openclaw/cli/config-file.ts
+++ b/openclaw/cli/config-file.ts
@@ -44,11 +44,26 @@ export interface PluginAuthConfig {
 /** Read the full ~/.openclaw/openclaw.json */
 function readFullConfig(): Record<string, unknown> {
   if (exists(OPENCLAW_CONFIG_FILE)) {
+    const text = readText(OPENCLAW_CONFIG_FILE);
+    let parsed: unknown;
     try {
-      return JSON.parse(readText(OPENCLAW_CONFIG_FILE));
-    } catch {
-      /* ignore parse errors */
+      parsed = JSON.parse(text);
+    } catch (err) {
+      // Fail-closed: throw instead of returning {} so that callers
+      // (writePluginAuth, writePluginConfigField) cannot silently overwrite
+      // the entire config with only the plugin entry.
+      throw new Error(
+        `[openclaw-mem0] Failed to parse ${OPENCLAW_CONFIG_FILE}: ${(err as Error).message}. ` +
+          `Fix the file manually or delete it to start fresh.`,
+      );
     }
+    if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
+      throw new Error(
+        `[openclaw-mem0] ${OPENCLAW_CONFIG_FILE} does not contain a JSON object. ` +
+          `Fix the file manually or delete it to start fresh.`,
+      );
+    }
+    return parsed as Record<string, unknown>;
   }
   return {};
 }
--- a/openclaw/tests/config-file.test.ts
+++ b/openclaw/tests/config-file.test.ts
@@ -113,10 +113,16 @@ describe("readPluginAuth", () => {
     expect(auth.userId).toBe("user-snake");
   });
 
-  it("returns empty object when JSON is invalid", () => {
+  it("throws when config file contains invalid JSON", () => {
     mockExists.mockReturnValue(true);
     mockReadText.mockReturnValue("not valid json {{{");
-    expect(readPluginAuth()).toEqual({});
+    expect(() => readPluginAuth()).toThrow("[openclaw-mem0]");
+  });
+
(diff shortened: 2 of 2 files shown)


kartik-mem0:
closing this pr in favour of #4861

PR #4864, opened 2026-04-16
fix(oss): initialize entity cleanup in fresh Python SDK processes

shaun0927:
## Linked Issue

Closes #4863

## Description

Fixes a fresh-process cleanup gap in the Python SDK's entity-link maintenance.

### Problem

`_remove_memory_from_entity_store()` returned immediately when `_entity_store` was unset. Because the entity store is lazy-initialized, delete/update flows could skip cleanup after a process restart, leaving stale `linked_memory_ids` behind.

### What changed

- resolve the entity store through the lazy `entity_store` property before cleanup,
- keep cleanup best-effort / non-fatal if entity-store initialization fails,
- reuse the initialized store for the list / delete / update operations,
- add regression tests for both sync and async delete flows.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None.

## Test Coverage

Local verification:
- `pytest -q tests/memory/test_main.py`
- added regression tests proving that both sync and async delete paths now initialize the entity store before cleanup in a fresh process.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -461,16 +461,18 @@ class Memory(MemoryBase):
           - otherwise re-embed the entity text and update the payload
             (the vector store's update() requires a vector).
 
-        No-op if the entity store has never been initialized in this process.
         Errors on individual entities are swallowed at debug level; outer
         failures are swallowed at warning level so the primary delete/update
         path is never broken by entity cleanup.
         """
-        if self._entity_store is None:
+        try:
+            entity_store = self.entity_store
+        except Exception as e:
+            logger.warning(f"Entity store initialization failed for memory_id={memory_id}: {e}")
             return
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         try:
-            listed = self.entity_store.list(filters=search_filters, top_k=10000)
+            listed = entity_store.list(filters=search_filters, top_k=10000)
             rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
             for row in rows or []:
                 try:
@@ -481,7 +483,7 @@ class Memory(MemoryBase):
                     remaining = [mid for mid in linked if mid != memory_id]
                     if not remaining:
                         try:
-                            self.entity_store.delete(vector_id=row.id)
+                            entity_store.delete(vector_id=row.id)
                         except Exception as e:
                             logger.debug(f"Entity delete failed for id={row.id}: {e}")
                     else:
@@ -496,7 +498,7 @@ class Memory(MemoryBase):
                             continue
                         new_payload = {**payload, "linked_memory_ids": remaining}
                         try:
-                            self.entity_store.update(
+                            entity_store.update(
                                 vector_id=row.id,
                                 vector=vec,
                                 payload=new_payload,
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths

- The root-cause diagnosis is exactly right: `_remove_memory_from_entity_store()` checks `self._entity_store is None` and returns early, but `entity_store` is a lazy property — a fresh process that hasn't yet called `add()` will always have `_entity_store = None`, so every delete/update silently skips entity cleanup and leaves stale `linked_memory_ids` behind. This is a real data-correctness bug.
- The fix is the correct minimal change: replace the raw-field check with `entity_store = self.entity_store` routed through the lazy-initializing property, wrapped in `try/except` so a misconfigured store still doesn't break the primary delete path.
- Both `Memory._remove_memory_from_entity_store` (sync) and `AsyncMemory._remove_memory_from_entity_store` (async) receive identical treatment — the issue named both and both are fixed.
- Caching `entity_store = self.entity_store` at the top of the method and using it throughout avoids redundant property calls and any TOCTOU risk within the method body.
- The regression test approach — probe subclasses that override `entity_store` as a property, constructed via `__new__` to simulate a fresh process, calling `_delete_memory` end-to-end — verifies real behavior rather than just stubbing internals.
- Stale docstring (`"No-op if the entity store has never been initialized in this process."`) is correctly removed in the diff.

### Issues

#### Important (should fix)

1. `tests/memory/test_main.py` — Both new tests set up a single `linked_memory_ids: ["mem-1"]`, so `remaining` is always `[]` after removal. This routes exclusively through the `entity_store.delete()` branch and never exercises the update branch (when other memory IDs remain, requiring a re-embed + `entity_store.update()` call). Neither probe instance has `embedding_model` set, so the update path would raise `AttributeError` if hit. To cover this: add a second fixture variant with `linked_memory_ids: ["mem-1", "mem-2"]` for both the sync and async tests, set `memory.embedding_model = MagicMock(return_value=[0.1, 0.2, ...])`, and assert `entity_store.update` is called with the trimmed payload.

#### Minor (optional)

1. `tests/memory/test_main.py` — `_ProbeMemory` and `_ProbeAsyncMemory` are structurally identical except for the parent class. A shared mixin for the `entity_store` property + `entity_store_accesses` counter would eliminate duplication as the test file grows.

### Evidence

- **Bug confirmed on main:** `mem0/memory/main.py` line 558 (sync) and line 2081 (async) — both have `if self._entity_store is None: return` before calling the lazy `entity_store` property. The property (lines 479–500 sync, 2013–2033 async) calls `VectorStoreFactory.create()` on first access — so any fresh process with no prior `add()` call will always hit the early return.
- **CI:** Python tests did not run — the PR has a merge conflict against current main. Vercel auth and CLA are the only reported failures; those are not Python CI.
- **TS counterpart:** `AsyncMemory` in `mem0-ts` already uses `getEntityStore()` inside a `try/catch` — no cross-package parity issue introduced here.
- **Duplicate cluster:** No other open PR targets this root cause. PRs #4991 and #5553 address different bugs (pgvector list efficiency and delete_all race condition respectively).
- **Could NOT verify:** Red-green test proof against current main — the merge conflict prevents a clean worktree fetch. Assessment is based on code tracing + diff review.

### Two blockers to clear before this can merge

**a) Rebase onto current main.** The branch has a merge conflict with main. A rebase to resolve it will also let CI run the Python test matrix — we need that green signal before merge.

**b) Sign the CLA.** The CLA check at `cla-assistant.io/mem0ai/mem0?pullRequest=4864` is pending. Follow the link, sign, and the bot will re-check automatically.

Once both are done and CI is green, please push — the fix itself is solid work and this is worth getting across the line.

### Assessment

Ready to merge: **With fixes** — please resolve the rebase, sign the CLA, and add the update-branch test coverage; happy to re-review once you push.

kartik-mem0:
Hi shaun0927, thanks for catching this one first. You correctly identified the root cause: the entity cleanup paths guarded on the raw `_entity_store` backing field instead of going through the lazy `entity_store` property, so a fresh process that had never called `add()` silently skipped cleanup entirely.

We reviewed #5574 today, which fixes the same bug. Comparing the two diffs hunk by hunk, its changes to `_remove_memory_from_entity_store` are line for line what you wrote here, in both the sync and async versions, right down to the warning messages. It then carries the same fix into four more places that have the identical guard: `reset()`, the async `reset()`, `_bulk_clear_entity_store`, and the async `delete_all`. We verified all six against real Qdrant, and the delete and reset scenarios fail on main and pass with the fix, in both sync and async modes.

Since everything here is fully contained in #5574 and this branch has drifted into conflict with main as the file moved around, I'm closing this in favour of #5574 so the work isn't duplicated. That's a bookkeeping call, not a reflection on the fix, which was right. Thanks for the catch and for the clean diff, and please do send more our way.

PR #4865, opened 2026-04-16
fix(cli): remove API keys from telemetry subprocess argv

shaun0927:
## Linked Issue

Closes #4862

## Description

Removes the Mem0 API key from telemetry child-process argv in both CLIs.

### Problem

Both CLI telemetry helpers spawned a detached child process with the full telemetry context serialized into a JSON command-line argument. That context included the real Mem0 API key used for `/v1/ping/`, so the secret could be exposed through local process inspection tools.

### What changed

- Python CLI now spawns `mem0_cli.telemetry_sender` without a JSON argv payload and writes the telemetry context through stdin instead.
- Node CLI now spawns `telemetry-sender.cjs` without a JSON argv payload and writes the telemetry context through stdin instead.
- Both sender implementations now read from stdin first and fall back to argv for compatibility with older parent processes during rollout.
- Added regression tests covering the new stdin handoff and compatibility loading paths.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None. The sender side keeps argv fallback compatibility so mixed-version parent/child combinations still work during upgrades.

## Test Coverage

Local verification:
- `pytest -q cli/python/tests/test_telemetry.py`
- `pnpm test -- --runInBand tests/telemetry.test.ts`
- manual `ps` repros for both CLIs confirmed that the telemetry sender still launches, but the Mem0 API key no longer appears in argv.

## Checklist

### DIFF
--- a/cli/node/src/telemetry.ts
+++ b/cli/node/src/telemetry.ts
@@ -143,9 +143,10 @@ export function captureEvent(
 
 		const child = spawn(
 			process.execPath,
-			[SENDER_SCRIPT, JSON.stringify(context)],
-			{ detached: true, stdio: "ignore" },
+			[SENDER_SCRIPT],
+			{ detached: true, stdio: ["pipe", "ignore", "ignore"] },
 		);
+		child.stdin?.end(JSON.stringify(context));
 		child.unref();
 	} catch {
 		/* silently swallow */
--- a/cli/node/telemetry-sender.cjs
+++ b/cli/node/telemetry-sender.cjs
@@ -19,6 +19,31 @@
 const https = require("https");
 const fs = require("fs");
 
+function loadContext() {
+	return new Promise((resolve, reject) => {
+		if (process.argv[2]) {
+			try {
+				resolve(JSON.parse(process.argv[2]));
+			} catch (err) {
+				reject(err);
+			}
+			return;
+		}
+
+		let data = "";
+		process.stdin.setEncoding("utf8");
+		process.stdin.on("data", (chunk) => (data += chunk));
+		process.stdin.on("end", () => {
+			try {
+				resolve(JSON.parse(data));
+			} catch (err) {
+				reject(err);
+			}
+		});
+		process.stdin.on("error", reject);
+	});
+}
+
 function httpsRequest(url, method, headers, body) {
 	return new Promise((resolve, reject) => {
 		const u = new URL(url);
@@ -108,7 +133,7 @@ async function sendIdentifyEvent(ctx, payload, anonId) {
 }
 
 async function main() {
-	const ctx = JSON.parse(process.argv[2]);
+	const ctx = await loadContext();
 	const payload = ctx.payload;
 
 	if (ctx.needsEmail && ctx.mem0ApiKey) {
--- a/cli/node/tests/telemetry.test.ts
+++ b/cli/node/tests/telemetry.test.ts
@@ -1,0 +1,59 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockLoadConfig = vi.fn();
+const mockSaveConfig = vi.fn();
+const mockSpawn = vi.fn();
+
+vi.mock("../src/config.js", () => ({
+  CONFIG_FILE: "/tmp/mem0-config.json",
+  loadConfig: mockLoadConfig,
+  saveConfig: mockSaveConfig,
+}));
+
+vi.mock("node:child_process", () => ({
+  spawn: mockSpawn,
+}));
+
+describe("captureEvent", () => {
+  beforeEach(() => {
+    vi.resetModules();
+    mockLoadConfig.mockReset();
+    mockSaveConfig.mockReset();
+    mockSpawn.mockReset();
+    delete process.env.MEM0_TELEMETRY;
+  });
+
+  it("pipes the telemetry context through stdin instead of argv", async () => {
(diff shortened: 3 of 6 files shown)


kartik-mem0:
The approach here is the right one for this class of bug — moving the secret out of argv into a pipe that only the child process receives. Both CLIs are covered, which is what this issue requires.

One thing blocking merge: the Node CI lint job is failing. Biome wants to reformat the `spawn(...)` call in `cli/node/src/telemetry.ts` — your version puts each argument on its own line, but Biome's formatter wants the first two arguments inline with the opening paren and the options object opened on the same line. Running `pnpm run lint:fix` inside `cli/node/` will auto-apply the reformat in about two seconds.

Everything else checks out: the Python sender's `sys.stdin.isatty()` guard is solid, the Node sender's `loadContext()` correctly tries argv first for rollout compatibility before falling back to stdin, the `child.stdin?.end()` optional chain is safe for fire-and-forget, and the regression tests accurately verify that the secret no longer appears in the spawn args. The CLA check is still pending — worth confirming that's resolved before the next push.

This needs a maintainer security sign-off before it can merge, but the fix itself is sound once the Biome reformat is applied.

kartik-mem0:
## Needs Human Review — recommended merge candidate for #4862

This comparative review covers both this PR and #4998, which address the same issue.

### What this PR does

Moves the full telemetry context from subprocess argv to stdin for both the Python CLI and the Node CLI. The API key is no longer visible in process listings, `/proc/<pid>/cmdline`, or Activity Monitor on any platform — the stdin pipe channel is invisible to all of those.

### Why this is the better long-term fix over #4998

#4998 fixes only the Python CLI using an environment variable approach. The Node CLI (`cli/node/src/telemetry.ts`) still passes `mem0ApiKey` inside `JSON.stringify(context)` as a command-line argument after #4998 would merge — the Node side of the hole would remain open. Issue #4862 explicitly covers both CLIs, so a Python-only fix is an incomplete remediation.

Beyond cross-package coverage: the stdin pipe approach is also the stronger security model. Env vars are better than argv (they don't show in `ps`, and `/proc/<pid>/environ` requires same-UID access on Linux), but they are still observable via system tracing tools and process environment APIs. Stdin pipe data exists only in the kernel pipe buffer and the receiving process's read buffer — it is not accessible via any standard process inspection path.

#4998 additionally ships with three broken tests: `@patch("mem0_cli.telemetry.is_agent_mode")` patches a name that does not exist in `mem0_cli.telemetry`'s namespace, and `_make_config` is missing the `agent_mode` attribute that `capture_event` accesses — both issues cause `subprocess.Popen` to never be called and all three assertions to fail. Those tests would pass CI only because CI never ran on that PR.

### One thing still blocking

The Node CI lint job is failing. Biome wants the `spawn()` call in `cli/node/src/telemetry.ts` reformatted: the first two arguments need to be inline with the opening paren rather than on separate lines. Running `pnpm run lint:fix` inside `cli/node/` applies this automatically — it does not touch the logic, just the whitespace. CLA also needs to be signed before anything can merge.

### Findings

Nothing else is blocking. A couple of minor notes for completeness:

- `cli/python/tests/test_telemetry.py:40` — `assert argv[0] == argv[0]` is comparing a value to itself; change to `assert argv[0] == sys.executable` for a meaningful assertion.
- `cli/node/telemetry-sender.cjs` checks `process.argv[2]` before stdin (Python's `_load_context` does stdin first). This is fine for rollout compatibility — the only caller is mem0's own parent process, so no external attack surface — but it's worth a brief comment in the code noting that the argv branch exists for backward compatibility only and should be removed after rollout.

### Evidence

- **CI:** Python tests pass on 3.10, 3.11, 3.12; Node tests pass on Node 20 and 22; both builds pass. One lint failure in Node (Biome format, one-line fix). Vercel failure is unrelated fork auth.
- **Root cause verified:** Traced on main — `cli/node/src/telemetry.ts:150` spawns `[SENDER_SCRIPT, JSON.stringify(context)]` with `context.mem0ApiKey` present; `cli/python/src/mem0_cli/telemetry.py` spawns with `context["mem0_api_key"]` in argv. Both are eliminated by this PR.
- **#4998 cross-checked:** Confirmed Node CLI is untouched in that diff. Its tests confirmed broken (skeptic-verified).
- **Could NOT verify:** Live process inspection to confirm key absence at runtime — code tracing only.

### Assessment

Ready to merge: **No — maintainer security sign-off required.** Once Biome lint is fixed and CLA is signed, this is the right fix. Est. human time: ~5 min.

kartik-mem0:
Thank you for this, @shaun0927 — your diagnosis and the stdin approach were exactly right, and covering **both** the Python and Node CLIs is precisely what #4862 needed. The writeup and reproduction steps were excellent.

To get the fix landed today, I've opened #5668 from the main repo carrying your implementation (**co-authored to you**), plus the small follow-ups from review:

- the Biome reformat on the Node `spawn(...)` call that was failing CI here,
- a stronger test assertion (`argv[0] == sys.executable` instead of comparing a value to itself, and a check that the stdin pipe is closed), and
- a couple of clarifying comments on the backward-compat fallback.

It's green across the Python + Node CLI CI, the CLA is signed, and I verified the fix end-to-end with a before/after `ps` check (the API key is gone from the sender's argv).

I'm closing this one as **superseded by #5668** so the fix can merge without waiting on the fork CLA — but the credit for the fix is yours. Really appreciate the careful work here. 🙏

Issue #4869, opened 2026-04-16
`delete_all()` only removes the first page of memories when more than the vector store limit exists


shafdev:
### Component

Core / Python SDK

### Summary
`Memory.delete_all()` in Python calls `vector_store.list()` once without a limit, so it only receives the first page of matching memories when the underlying vector store applies a default page size. As a result, bulk deletion can leave older matching memories behind. The async Python implementation has the same issue, as does the TypeScript `deleteAll` in `mem0-ts/src/oss/src/memory/index.ts`, which calls `this.vectorStore.list(filters)` without a limit and not in a loop.

### Steps to Reproduce

1. Create more than 100 memories for the same `user_id`.
2. Call `Memory.delete_all(user_id="...")`.
3. Query the same `user_id` again.

Example:

```python
from mem0 import Memory

memory = Memory()

for i in range(150):
    memory.add(f"Memory {i}", user_id="bug-002-user")

memory.delete_all(user_id="bug-002-user")
remaining = memory.get_all(filters={"user_id": "bug-002-user"})
print(len(remaining["results"]))
```

### Expected Behavior

All memories matching the provided scope should be deleted, regardless of how many pages the vector store needs to return them.

### Actual Behavior

Only the first page of matching memories is deleted. For stores with a default page size of 100, calling `delete_all()` on 150 memories leaves 50 behind.

### Environment

- mem0 version: main branch (reproduced on 3fbc1c9aef7843284440)
- Python/Node version: Python 3.12
- OS: Linux

shafdev:
The earlier PRs (#4870, #4871) bundled Python and TypeScript changes together which caused CI failures due to an unrelated OOM issue in the TS test suite. #4872 contains only the Python fix and is the one to review. A separate PR for the TypeScript deleteAll fix will follow.

nileshpatil6:
Hi! I'd like to work on this.

The fix is to paginate the deletion loop instead of a single `list()` call. Something like:

```python
def delete_all(self, user_id=None, agent_id=None, run_id=None, ...):
    filters = {...}
    while True:
        memories = self.vector_store.list(filters=filters, limit=batch_size)
        if not memories:
            break
        for memory in memories:
            self.vector_store.delete(vector_id=memory.id)
            # also remove from graph store if enabled
```

The same pattern should be applied to the async `adelete_all` and the TypeScript `deleteAll`. Would a PR covering all three be welcome?

kartik-mem0:
## Triage Notes

[...]

PR #4870, opened 2026-04-16
fix: resolve incomplete delete_all pagination for large memory sets (…

shafdev:
## Linked Issue

Closes #4869

## Description

`Memory.delete_all()` (Python) and `deleteAll()` (TypeScript) called `vector_store.list()` a single time without a loop. Vector stores like Qdrant apply a default page size (typically 100), so only the first batch of memories was deleted — the rest were silently left behind.

The fix replaces the single call with a re-list loop: after deleting each batch, the store is queried again from the start. Since deleted records are gone, this drains all remaining pages correctly. Offset-based pagination was deliberately avoided because it would skip records as items shift after deletion.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Manually verified by creating 150 memories for a single `user_id`, calling `delete_all()`, then asserting `get_all()` returns 0 results. Tests added to `tests/memory/test_main.py`, `tests/test_main.py`, and `mem0-ts/src/oss/tests/memory.crud.test.ts` covering `delete_all` with more than one page of memories.

## Checklist

### DIFF
--- a/.github/workflows/ts-sdk-ci.yml
+++ b/.github/workflows/ts-sdk-ci.yml
@@ -59,6 +59,8 @@ jobs:
 
       - name: Run unit tests
         working-directory: mem0-ts
+        env:
+          NODE_OPTIONS: "--max-old-space-size=6144"
         run: pnpm run test:unit
 
       - name: Verify package exports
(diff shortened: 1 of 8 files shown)


shafdev:
Closing in favour of #4872 which contains only the Python fix. A separate PR for the TypeScript deleteAll fix will follow.

PR #4874, opened 2026-04-17
fix(ts): drain all pages in deleteAll, not just the first batch

octo-patch:
## Linked Issue

Closes #4869

## Description

`Memory.deleteAll()` in TypeScript called `vectorStore.list(filters)` once with the default `topK` of 100, so only the first 100 memories were deleted when more than 100 existed for a given scope.

This is the TypeScript counterpart of the Python fix in #4872. The comment on issue #4869 noted that a separate PR for the TypeScript `deleteAll` fix would follow.

**Root cause:** `list()` defaults to `topK = 100`. A single call returns at most one page.

**Fix:** Replace the single call with a re-list loop that keeps fetching from offset 0 after each batch. Offset-based pagination is deliberately avoided — as memories are deleted the remaining items shift, so re-querying from the start is the only safe way to drain all pages.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added a regression test in `memory.crud.test.ts` that mocks `vectorStore.list` to return two batches (100 + 50 items) followed by an empty sentinel. Verifies that `deleteAll` calls `deleteMemory` exactly 150 times and calls `list` exactly 3 times (two non-empty batches + one empty confirmation).

All 23 existing `memory.crud.test.ts` tests pass.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1300,9 +1300,20 @@ export class Memory {
       );
     }
 
-    const [memories] = await this.vectorStore.list(filters);
-    for (const memory of memories) {
-      await this.deleteMemory(memory.id);
+    // Use a large batch size and re-list from the start after each batch.
+    // Offset-based pagination is deliberately avoided: as memories are deleted
+    // the remaining items shift, so re-querying from offset 0 after each batch
+    // is the safest way to drain all pages.
+    const DELETE_BATCH_SIZE = 1000;
+    let deletedCount = 0;
+
+    let [memories] = await this.vectorStore.list(filters, DELETE_BATCH_SIZE);
+    while (memories.length > 0) {
+      for (const memory of memories) {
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for putting this together — the implementation is clean and the diagnosis is exactly right.

This lands on the same root cause and the same fix as #4872: `Memory.deleteAll()` in `mem0-ts/src/oss/src/memory/index.ts` called `this.vectorStore.list(filters)` a single time at the default page limit, so only the first page was ever deleted. Both PRs drain every page with a re-list loop at `DELETE_BATCH_SIZE = 1000`.

The wrinkle is that #4872 already carries this same TypeScript change alongside the Python `delete_all` fix (both sync and async) for #4869. To keep the Python and TS sides moving together and the discussion in one place, I'm going to consolidate on #4872 and close this one.

Please follow along on #4872 — and if you spot anything its TS version misses, flag it there and we'll fold it in. Really appreciate the contribution.

PR #4878, opened 2026-04-17
security: fix SQL and Cypher injection vulnerabilities in PGVector, Azure MySQL, and Neptune

### LINKED ISSUE #4875: [Security] SQL/Cypher Injection in PGVector, Azure MySQL, and Neptune vector stores via collection_name (CWE-89)
## Summary

Multiple vector store backends use f-string interpolation for `collection_name` and filter values in SQL/Cypher queries, enabling injection attacks. This is the **same bug class** fixed for Databricks in #4073/#4558, but PGVector, Azure MySQL, and Neptune Analytics were **not patched**.

### 1. PGVector (`mem0/vector_stores/pgvector.py`) — 18 injection points

Every SQL operation interpolates `self.collection_name` via f-strings:

```python
# Line 198
f"INSERT INTO {self.collection_name} (id, vector, payload) VALUES (%s, %s, %s)"

# Line 305
f"DELETE FROM {self.collection_name} WHERE id = %s"

# Line 377
f"DROP TABLE IF EXISTS {self.collection_name}"
```

Full list: lines 156, 169, 170, 177, 178, 184, 185, 198, 205, 242, 282, 305, 324, 332, 338, 355, 377, 391, 392, 428.

### 2. Azure MySQL (`mem0/vector_stores/azure_mysql.py`) — 11 injection points

Same pattern with backtick quoting (bypassable):

```python
# Line 230
f"INSERT INTO \`{self.collection_name}\` (id, vector, payload) VALUES (%s, %s, %s)"
```

Lines: 183, 198, 230, 291, 340, 381, 400, 408, 414, 444.

### 3. Neptune Analytics (`mem0/vector_stores/neptune_analytics.py`) — Cypher injection

Both `collection_name` and **user-controlled filter values** are interpolated:

```python
# Line 299 — collection_name in Cypher
f"MATCH (n :{self.collection_name}) DETACH DELETE n"

# Lines 414-418 — filter values in WHERE clause
where_clause += f"WHERE n.{k} = '{v}' "
```

The `_get_where_clause()` method at line 403 builds Cypher queries from the `filters` dict passed by API users via the `/search` endpoint.

## Root Cause

`PGVectorConfig.collection_name` is `str = Field("mem0", ...)` with **no Pydantic validator, no regex check, no sanitization**. Same for MySQL and Neptune configs.

## Attack Path

1. The `POST /configure` endpoint (`server/main.py:131`) accepts arbitrary JSON config
2. It calls `Memory.from_config(config)` which passes `collection_name` to the vector store
3. When `ADMIN_API_KEY` is not set (server warns about this), the endpoint is **unauthenticated**
4. Library users calling `Memory.from_config()` with user-controlled config are also vulnerable

For Neptune filters: the `POST /search` endpoint accepts `filters: Dict[str, Any]` which flows directly to `_get_where_clause()`.

### PGVector SQL Injection
```python
from mem0 import Memory
# Malicious collection_name escapes the identifier context
config = {
    "vector_store": {
        "provider": "pgvector",
        "config": {
            "collection_name": "memories; DROP TABLE users; --",
            "host": "localhost",
            "port": 5432,
        }
    }
}
m = Memory.from_config(config)  # Executes: CREATE TABLE IF NOT EXISTS memories; DROP TABLE users; --
```

### Neptune Cypher Injection
```python
# API request to /search
{
    "query": "hello",
    "filters": {"user_id": "x' RETURN n UNION MATCH (m) DETACH DELETE m //"}
}
# Generates: WHERE n.user_id = 'x' RETURN n UNION MATCH (m) DETACH DELETE m //'
```

## Prior Art

- **Issue #4073** (Feb 2026): Reported SQL injection in Databricks via f-string interpolation — **exact same bug class**
- **PR #4558** (Mar 2026): Fixed Databricks by adding `_VALID_SQL_IDENTIFIER` regex validation
- **PGVector, Azure MySQL, Neptune were NOT fixed** in that PR
- No existing CVE for mem0

## Severity

**CVSS 3.1: 8.1 (High)** for Neptune (user-controlled filters, no auth needed)
**CVSS 3.1: 6.5 (Medium)** for PGVector/MySQL (requires /configure access)

**CWE-89**: Improper Neutralization of Special Elements used in an SQL Command

## Suggested Fix

Apply the same `_VALID_SQL_IDENTIFIER` pattern from the Databricks fix (#4558) to ALL vector stores:

```python
import re
_VALID_SQL_IDENTIFIER = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")

# In each config model:
@field_validator("collection_name")
def validate_collection_name(cls, v):
    if not _VALID_SQL_IDENTIFIER.match(v):
        raise ValueError(f"Invalid collection_name: {v}")
    return v
```

For Neptune filters, use parameterized Cypher queries instead of f-strings.

## Disclosure

Requesting CVE assignment. The underlying pattern was publicly disclosed in #4073 but the incomplete fix leaves 3 backends vulnerable.

Parteeksachdeva:
## Linked Issue

Closes #4875

## Description

This PR remediates several SQL and Cypher injection vulnerabilities (CWE-89) in the PGVector, Azure MySQL, and Neptune Analytics vector store backends. These vulnerabilities were caused by direct f-string interpolation of user-controlled `collection_name` and filter values into database queries.

### Key Changes:
- **Strict Configuration Validation**: Added Pydantic `field_validator` to `PGVectorConfig`, `AzureMySQLConfig`, and `NeptuneAnalyticsConfig` to ensure `collection_name` only contains safe alphanumeric characters and underscores, starting with a letter or underscore. This prevents injection into any query string where the table name or label is interpolated.
- **Parameterized Neptune Queries**: Refactored the `NeptuneAnalyticsVector` search and list methods to use parameterized queries instead of string interpolation for filters. Vector search now uses the structured `nodeFilter` parameter object.
- **Filter Key Validation**: Added regex validation for filter property keys in `AzureMySQL` and `NeptuneAnalyticsVector` to guard against injection via metadata property paths.

## Type of Change

- [ ] Documentation update

## Breaking Changes

This change introduces stricter validation for the `collection_name` field in PGVector, Azure MySQL, and Neptune configurations. Collection names must now match the regex `^[a-zA-Z_][a-zA-Z0-9_]*$`. Users with existing non-standard collection names (e.g., those containing spaces or starting with digits) will need to rename their collections to remain compliant with the new security standards.

## Test Coverage

Manual testing was performed using a dedicated security test script (`tests/vector_stores/test_security_injection.py`) that attempts various injection payloads (SQL/Cypher) in both configuration and filter parameters, verifying they

### DIFF
--- a/mem0/configs/vector_stores/azure_mysql.py
+++ b/mem0/configs/vector_stores/azure_mysql.py
@@ -1,6 +1,11 @@
+import re
 from typing import Any, Dict, Optional
 
-from pydantic import BaseModel, ConfigDict, Field, model_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+
+# Rejecting any identifier that doesn't match this pattern is what keeps the
+# f-string table/column interpolations in mem0/vector_stores/azure_mysql.py safe.
+_VALID_SQL_IDENTIFIER = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 
 
 class AzureMySQLConfig(BaseModel):
@@ -25,6 +30,15 @@ class AzureMySQLConfig(BaseModel):
         None,
         description="Pre-configured connection pool object (overrides other connection parameters)"
     )
+
+    @field_validator("collection_name")
+    def validate_collection_name(cls, v):
+        if not _VALID_SQL_IDENTIFIER.match(v):
+            raise ValueError(
+                f"Invalid collection_name: {v!r}. Must start with a letter or underscore and "
+                "contain only letters, digits, and underscores."
+            )
+        return v
 
     @model_validator(mode="before")
     @classmethod
--- a/mem0/configs/vector_stores/neptune.py
+++ b/mem0/configs/vector_stores/neptune.py
@@ -5,7 +5,13 @@ This module provides configuration settings for integrating with Amazon Neptune 
 as a vector store backend for Mem0's memory layer.
 """
 
-from pydantic import BaseModel, ConfigDict, Field
+import re
+
+from pydantic import BaseModel, ConfigDict, Field, field_validator
+
+# Rejecting any identifier that doesn't match this pattern is what keeps the
+# f-string label interpolations in mem0/vector_stores/neptune_analytics.py safe.
+_VALID_CYPHER_IDENTIFIER = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 
 
 class NeptuneAnalyticsConfig(BaseModel):
@@ -21,5 +27,14 @@ class NeptuneAnalyticsConfig(BaseModel):
     """
     collection_name: str = Field("mem0", description="Default name for the collection")
     endpoint: str = Field("endpoint", description="Graph ID for the runtime")
+
+    @field_validator("collection_name")
+    def validate_collection_name(cls, v):
+        if not _VALID_CYPHER_IDENTIFIER.match(v):
+            raise ValueError(
+                f"Invalid collection_name: {v!r}. Must start with a letter or underscore and "
+                "contain only letters, digits, and underscores."
+            )
+        return v
 
     model_config = ConfigDict(arbitrary_types_allowed=False)
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -1,6 +1,11 @@
+import re
 from typing import Any, Dict, Optional
 
-from pydantic import BaseModel, ConfigDict, Field, model_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+
+# Rejecting any identifier that doesn't match this pattern is what keeps the
+# f-string table/column interpolations in mem0/vector_stores/pgvector.py safe.
+_VALID_SQL_IDENTIFIER = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 
 
 class PGVectorConfig(BaseModel):
@@ -19,6 +24,15 @@ class PGVectorConfig(BaseModel):
     sslmode: Optional[str] = Field(None, description="SSL mode for PostgreSQL connection (e.g., 'require', 'prefer', 'disable')")
     connection_string: Optional[str] = Field(None, description="PostgreSQL connection string (overrides individual connection parameters)")
     connection_pool: Optional[Any] = Field(None, description="psycopg connection pool object (overrides connection string and individual parameters)")
+
+    @field_validator("collection_name")
+    def validate_collection_name(cls, v):
+        if not _VALID_SQL_IDENTIFIER.match(v):
+            raise ValueError(
+                f"Invalid collection_name: {v!r}. Must start with a letter or underscore and "
+                "contain only letters, digits, and underscores."
+            )
+        return v
 
     @model_validator(mode="before")
     def check_auth_and_connection(cls, values):
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, Optional
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 
+
 class PGVectorConfig(BaseModel):
     dbname: str = Field("postgres", description="Default name for the database")
     collection_name: str = Field("mem0", description="Default name for the collection")
@@ -19,6 +20,7 @@ class PGVectorConfig(BaseModel):
     sslmode: Optional[str] = Field(None, description="SSL mode for PostgreSQL connection (e.g., 'require', 'prefer', 'disable')")
     connection_string: Optional[str] = Field(None, description="PostgreSQL connection string (overrides individual connection parameters)")
     connection_pool: Optional[Any] = Field(None, description="psycopg connection pool object (overrides connection string and individual parameters)")
+
 
     @model_validator(mode="before")
     def check_auth_and_connection(cls, values):
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -3,7 +3,6 @@ from typing import Any, Dict, Optional
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 
-
 class PGVectorConfig(BaseModel):
     dbname: str = Field("postgres", description="Default name for the database")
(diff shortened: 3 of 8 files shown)


kartik-mem0:
## Review — 2026-06-19

Head reviewed: `08d80944`

### Strengths

- The Neptune `_get_where_clause` rewrite is correct and complete: both filter keys (regex-gated) and values (bound via `$filter_N` params) are now safe — this closes the highest-severity path reported in #4875.
- `_get_node_filter_clause` returning a structured Python dict instead of a hand-built Cypher string is the right design; values never touch the query string.
- Filter-key validation in Azure MySQL `search()` and `list()` is a clean, consistent guard.
- The test file covers the documented CVE payloads parametrically and the Neptune parameterization tests (both value-injection and key-skip) exercise the actual fix mechanism — not just the mechanism, but the real attack vector.
- Red-green confirmed locally: 27/30 FAIL on main, 30/30 PASS with the fix.

---

### Issues

**Critical — `neptune_analytics.py:__init__` (line 42)**

`NeptuneAnalyticsVector.__init__` accepts `collection_name: str` as a raw string with no validation. The Pydantic `field_validator` in `NeptuneAnalyticsConfig` only fires when the config object is constructed; anyone who instantiates `NeptuneAnalyticsVector` directly (including test code, scripts, or integrations that bypass the config class) skips it entirely. Once in `__init__`, `self.collection_name = self._COLLECTION_PREFIX + collection_name` is set, and every downstream method — `insert` (line 112), `delete` (line 188), `get` (line 258), `update` (lines 221, 233), `list` (line 331), `list_cols` (line 283), `delete_col` (line 299) — f-string interpolates `self.collection_name` directly into Cypher with no quoting layer. The `_COLLECTION_PREFIX` prepend does not help: `"MEM0_VECTOR_x} DETACH DELETE n MATCH (n"` is still injected Cypher.

Fix: add a validation guard in `__init__` before the prefix is applied — three lines mirroring the config regex — so direct instantiation is equally protected.

**Question — `neptune_analytics.py:160-169` (`search()`)**

The PR changes the `CALL neptune.algo.vectors.topKByEmbeddingWithFiltering` block to use `$top_k`, `$vectors`, and `$node_filter` as openCypher query parameters. The `WHERE`-clause parameterization (e.g. `WHERE id(n) = $node_id` at line 189) is confirmed to work fine. The question is whether Neptune Analytics' openCypher engine resolves `$param` references inside the map literal passed to a native procedure CALL — this is a different syntactic context than a WHERE clause, and not all Neptune procedure call sites accept parameterized map arguments.

The injection fix for `nodeFilter` does not require also parameterizing `$top_k` and `$vectors` (an int and a float list are not injection risks). A conservative alternative is to keep `topK` and `embedding` as inline values and only parameterize the `nodeFilter` object:

```python
query_string = f"""
    CALL neptune.algo.vectors.topKByEmbeddingWithFiltering({{
        topK: {top_k},
        embedding: {vectors},
        nodeFilter: $node_filter
    }})
    YIELD node, score
    RETURN node as n, score
"""
params = {"node_filter": filter_obj}
query_response = self.execute_query(query_string, params)
```

This is safe (top_k is an int, vectors is a float list — neither can carry Cypher injection), removes the uncertainty about Neptune's CALL-level parameterization, and still closes the actual injection vector. Please confirm whether the fully-parameterized form has been tested against a live Neptune Analytics endpoint, or consider the conservative split above.

**Important — `mem0/configs/vector_stores/pgvector.py:25-33`**

PGVector already uses `sql.Identifier` via `_col()` (pgvector.py:254) for every table reference — this is the correct, complete defense against identifier injection regardless of what the name contains (hyphens, uppercase, UUID format). The new Pydantic `field_validator` adds a stricter gate on top of it, but the regex `^[A-Za-z_][A-Za-z0-9_]*$` will reject legitimate PostgreSQL identifiers — UUIDs, hyphenated names, names beginning with a digit — that psycopg handles safely through `sql.Identifier` quoting. This is a breaking change for existing deployments with no security benefit for PGVector. Please remove the `validate_collection_name` validator from `PGVectorConfig` and let `sql.Identifier` continue handling it as it already does.

---

### Recommendations

- This PR is currently **CONFLICTING** with main — a rebase is required before merge.
- After rebase, the existing `_validate_identifier` call in `azure_mysql.py:87` (which uses `{0,127}` length cap) and the new config validator (unbounded `*`) will both run; they're complementary, not conflicting.
- Consider folding the new config-validator tests into the existing per-store test files (`test_pgvector.py`, `test_azure_mysql.py`, `test_neptune_analytics.py`) rather than a standalone `test_security_injection.py` — this keeps coverage co-located with the code it tests.

---

### Assessment

Not ready to merge as-is. The `__init__` bypass gap (Critical) and the PGVector unnecessary-breaking-change (Important) are concrete, fixable issues. The Neptune CALL parameterization question needs either a confirmed live-test or the conservative split described above. The core Neptune filter parameterization is sound and the tests are meaningful — the work here is real and appreciated. Once the three items above are addressed this should be straightforward.

---

**Evidence:** CI: `build_mem0` 3.10/3.11/3.12 PASS (Vercel advisory, ignored). Red-green: 27 FAIL on main → 30 PASS with fix. Lint: clean. Worktree: `$HOME/.cache/mem0-worktrees/review-pr-4878` (removed after review). Could NOT verify: live Neptune CALL parameterization behavior (no integration test without endpoint).

Parteeksachdeva:
Hi @kartik-mem0,
 
I've just pushed a major update that resolves the merge conflicts with `main` and fully addresses all of your review feedback from June 19:
 
1. **Resolved Conflicts (Neptune Filters):** Since PR #5982 has already been merged and fixed the Neptune filter Cypher injection, I dropped my parameterised approach for that section and kept `main`'s native string escaping approach to resolve the conflict cleanly.
2. **Fixed Neptune `__init__` Bypass (Critical):** I added the `_VALID_IDENTIFIER` regex guard directly into `NeptuneAnalyticsVector.__init__` as you requested, so direct instantiations bypassing the Pydantic config are now fully protected.
3. **Removed PGVector Validator (Important):** I completely removed the strict Pydantic regex validator from `PGVectorConfig` so that it doesn't break legitimate PostgreSQL identifiers (since `psycopg` already handles it safely). 
 4. **Reorganized Tests:** I deleted the standalone `test_security_injection.py` file and moved the security tests (including a new one for the `__init__` bypass) directly into `test_azure_mysql.py` and `test_neptune_analytics.py` to keep coverage co-located.

All tests (129/129) and linting (`ruff`) are passing locally. Everything should be clean and ready for a final review!

kartik-mem0 (approved):

Issue #3707, opened 2025-11-04
The conversation was conducted in Chinese, but in the end, the memory list retrieved from mem0 showed English


orcharddd2024:
### Issue with current documentation:

The conversation was conducted in Chinese, but in the end, the memory list retrieved from mem0 showed English

What's the reason for this? Can it be controlled?

<img width="1032" height="449" alt="Image" src="https://github.com/user-attachments/assets/f7d38fae-6104-40ff-8b29-ed008f597fa4" />

Is it like adding a sentence 'Please answer in Chinese' directly in the content?
But if it is stored in the content, then the sentence 'please answer in Chinese' will also be stored as memory in mem0, and adding 'please answer in Chinese' to each content will also be troublesome

dgzxx-2000:
I also encountered the same problem.

orcharddd2024:
<img width="884" height="364" alt="Image" src="https://github.com/user-attachments/assets/5d2e206a-19e9-439c-a5d2-e8ce8ede69d5" />

I know how to do it now. The source code can provide a place to add system prompts


https://docs.mem0.ai/open-source/features/custom-fact-extraction-prompt#prompt-blueprint

dgzxx-2000:
建议从源代码提示词哪里改

kartik-mem0:
Closing this — the language of extracted memories is determined by the LLM used for fact extraction. As noted in the thread, using a custom fact extraction prompt with explicit language instructions (via `custom_fact_extraction_prompt` in config) resolves this.

Issue #3206, opened 2025-07-22
The memorized information is translated into English


i-iooi-i:
### 🐛 Describe the bug

<img width="1656" height="530" alt="Image" src="https://github.com/user-attachments/assets/720951d7-778b-46c2-8bf3-dd45e21bcf34" />

When some "personal names" and "company names" are mentioned in Chinese, there is a certain probability that they will be converted into English. The next time I mentioned this information to the AI, the names were translated into strange English names. 😂

kartik-mem0:
Hey! Thanks so much for reporting this — your feedback means a lot to us.

We're marking this as stale and closing it since it's been inactive for some time. Mem0 has evolved quite a bit since this issue was opened, and there's a good chance the behavior may have changed. If you're still running into this problem, please don't hesitate to open a fresh issue with the latest details and we'll take a look.

We appreciate your patience and support! 💙

PR #1818, opened 2024-09-05
feat(memory): keep memory language

shenxiangzhuang:
## Description

- Memory Save(deduced fact): Keep the memory's language same with input

Fixes #1817 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] Documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

Please delete options that are not relevant.

- [ ] Unit Test
- [ ] Test Script (please provide)

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

## Maintainer Checklist

- [ ] closes #xxxx (Replace xxxx with the GitHub issue number)
- [ ] Made sure Checks passed

PR #4883, opened 2026-04-18
feat: expose use_input_language config to preserve source language in extracted memories

wangjiawei-vegetable:
## Linked Issues

Refs #1817, #3206, #3707, #6312.

Related implementation history:

- #1818 previously added same-language memory extraction and was merged.
- #4805 introduced the V3 additive extraction pipeline and the existing `use_input_language` prompt branch.
- #5549 also identified the same unreachable flag while working on broader CJK support.
- In #6312, maintainers confirmed that the current sync and async call sites never enable the existing language-preservation block and described exposing it as a first-class configuration option as a worthwhile product follow-up.

## Description

The V3 extraction prompt builder already supports `use_input_language`, which adds instructions to preserve the source language and script. However, `MemoryConfig` did not expose that option, and neither `Memory` nor `AsyncMemory` forwarded it. As a result, the built-in language-preservation path was unreachable through the public OSS API.

This PR:

- adds `use_input_language: bool = False` to `MemoryConfig`;
- forwards the option through synchronous and asynchronous extraction;
- verifies the public configuration path for both enabled and disabled states;
- documents the option for Python OSS users and updates the documentation index.

The default remains `False`, so existing behavior is unchanged unless users opt in.

```python
from mem0 import Memory

memory = Memory.from_config({
    "use_input_language": True,
})

memory.add("我喜欢周末打乒乓球", user_id="xiaoming")
```

`custom_instructions` remains available for applications that need additional domain-specific or language-specific extraction rules. This option provides a focused built-in path for users who only need source-language preservation.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an existing configuration gap)
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [x] Documentation update

## Breaking Changes

None. `use_input_language` defaults to `False`.

## Test Coverage

- [x] Added/updated unit tests
- [x] Tested locally
- [ ] No tests needed

Validation completed on Python 3.12:

- `tests/memory/test_main.py` and `tests/memory/test_use_input_language.py`: 57 passed;
- focused input-language tests: 5 passed;
- Ruff checks passed for the touched Python files;
- `docs/llms.txt` coverage check passed.

The focused tests instantiate both `Memory` and `AsyncMemory` through `from_config()` and verify that the language requirement reaches the extraction prompt only when enabled.

## Checklist

- [x] Self-review performed
- [x] New and existing related tests pass locally
- [x] Documentation updated

## Out of Scope

- TypeScript OSS support: the TypeScript prompt implementation does not currently expose an equivalent branch.
- Hosted `MemoryClient`: extraction prompts are managed server-side.

### DIFF
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -238,7 +238,7 @@ If the user is on a pre-current major (Python < 2, TS < 3, or Platform `output_f
 - [Reranking](https://docs.mem0.ai/open-source/features/reranking) [OSS]: Use when configuring reranking end-to-end in OSS.
 - [Async Memory](https://docs.mem0.ai/open-source/features/async-memory) [OSS]: Use when the self-hosted app needs `AsyncMemory`.
 - [OSS Multimodal Support (features)](https://docs.mem0.ai/open-source/features/multimodal-support) [OSS]: Use when handling images and PDFs self-hosted (feature guide).
-- [Custom Instructions (OSS)](https://docs.mem0.ai/open-source/features/custom-instructions) [OSS]: Use when tailoring extraction prompts in OSS.
+- [Custom Instructions (OSS)](https://docs.mem0.ai/open-source/features/custom-instructions) [OSS]: Use when tailoring extraction prompts or preserving the input language in OSS.
 - [REST API Server](https://docs.mem0.ai/open-source/features/rest-api) [OSS]: Use when exposing a self-hosted Mem0 as a FastAPI service.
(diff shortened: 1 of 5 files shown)


wangjiawei-vegetable:
Hi @kartik-mem0 — following up on [your triage note in #6312](https://github.com/mem0ai/mem0/issues/6312#issuecomment-4968790931), where you identified `use_input_language` as a real product gap that was not exposed as a first-class configuration option.

This PR wires the existing flag through `MemoryConfig` for both `Memory` and `AsyncMemory`. It remains opt-in (`False` by default), and now includes public-API tests and documentation.

When you have a chance, would you be willing to review whether this matches the direction you had in mind? Thanks!

PR #4885, opened 2026-04-18
fix(llms): narrow reasoning-model detection so gpt-5.4-mini keeps temperature

### LINKED ISSUE #4738: bug: _is_reasoning_model() incorrectly strips temperature for gpt-5.4-mini
## Problem

`_is_reasoning_model()` in `mem0/llms/base.py` matches any model containing `"gpt-5"` as a substring:

```python
if any(reasoning_model in model_lower for reasoning_model in ["gpt-5", "o1", "o3"]):
    return True
```

For matching models, `_get_supported_params()` strips ALL parameters except `messages`, `response_format`, `tools`, and `tool_choice` — including `temperature`.

`gpt-5.4-mini` matches via the `"gpt-5"` substring, but it **supports** the `temperature` parameter. Setting `temperature: 0.1` in the config is silently ignored — the parameter is stripped before the API call.

## Evidence

Added print instrumentation to `generate_response()`:
```
[DIAG] mem0 OpenAILLM.generate_response: model=gpt-5.4-mini, params_keys=['messages', 'model', 'response_format', 'store'], has_temperature=False
```

Verified that the OpenAI API accepts `temperature=0` for `gpt-5.4-mini`:
```python
response = client.responses.create(model="gpt-5.4-mini", ..., temperature=0)
# response.temperature confirms 0.0
```

## Impact

All `client.add()` extraction and update/dedup calls run at uncontrolled temperature, producing inconsistent results. For our CNC manufacturing domain, this causes:
- Transient facts leaking through extraction (inconsistent prompt following)
- Semantic duplicates not caught by update/dedup (inconsistent matching)

## Suggested fix

The substring match `"gpt-5"` is too broad. `gpt-5.4-mini` (and likely other gpt-5.x variants) support temperature. The check should either:
1. Use exact model name matches, not substring
2. Test the API for parameter support rather than guessing from the model name
3. At minimum, only match the base `gpt-5` model, not derivatives

sharziki:
## Linked Issue

Closes #4738

## Description

\`_is_reasoning_model\` in \`mem0/llms/base.py\` used a substring match:

\`\`\`python
if any(reasoning_model in model_lower for reasoning_model in [\"gpt-5\", \"o1\", \"o3\"]):
    return True
\`\`\`

This misclassifies any model name that *contains* \`gpt-5\`, \`o1\`, or \`o3\` as a reasoning model, causing \`_get_supported_params()\` to strip \`temperature\`, \`max_tokens\`, and \`top_p\` before the API call. Concretely, \`gpt-5.4-mini\` (and future point releases like \`gpt-5.5\`) match the \`gpt-5\` substring but actually support temperature, so \`temperature: 0.1\` set in the config is silently ignored. Other false positives include model names like \`llama-o1-chat\` and \`mistral-o3-8b\`.

## Fix

Replace the substring check with a compiled regex that requires a clean boundary after the core identifier:

- \`oN\` or \`oN-<suffix>\` for single-digit o-series models (e.g. \`o1\`, \`o1-mini\`, \`o3\`, \`o3-pro\`, \`o4-mini\`).
- \`gpt-5\` exactly, or \`gpt-5<alpha>\` (e.g. \`gpt-5o\`), each optionally followed by a \`-<suffix>\` chain (e.g. \`gpt-5-mini\`, \`gpt-5o-mini\`).

A digit or \`.\` directly after the core identifier breaks the match, so \`gpt-5.4-mini\`, \`gpt-5.5\`, and \`gpt-5.0-preview\` are treated as regular chat models. The previously-maintained \`reasoning_models\` set is redundant — the regex covers every name it had.

## Type of Change

## Breaking Changes

N/A — names that correctly mapped to reasoning behavior before (the base \`o1\`/\`o3\` and \`gpt-5\`/\`gpt-5o\` families with \`-suffix\` variants) still map to reasoning behavior.

## Test Coverage

- [x] Added \`tests/llms/test_base.py\` with:
  - A parametrized positive-case table across 13 reasoning variants (o1, o1-preview, o1-mini, o3, o3-mini, o3-pro, o4-mini, gpt-5, gpt-5-mini, gpt-5o, gpt-5o-mini, gpt-5o-micro, GPT-5).
  - A parametrized negative-case table for decimal GPT-5 names (\`gpt-5.4-mini\`, \`gpt-5.5\`, \`gpt-5.0-preview\`), substring false-positives (\`llama-o1-chat\`, \`mistral-o3-8b\`, \`custom-gpt-5-adapter\`), and gpt-4 baselines.
  - Two end-to-end \`_get_supported_params\` assertions confirming temperature is kept for \`gpt-5.4-mini\` and stripped for \`o3-mini\`.
- [x] \`pytest tests/llms/test_base.py\` — 25 passed.
- [x] \`pytest tests/llms/test_openai.py\` — 14 passed (no regression on existing reasoning-effort tests).

## Checklist

- [x] No documentation update needed (internal helper)

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -1,7 +1,22 @@
+import re
 from abc import ABC, abstractmethod
 from typing import Dict, List, Optional, Union
 
 from mem0.configs.llms.base import BaseLlmConfig
+
+# Reasoning models (OpenAI o-series and GPT-5 family) reject parameters like
+# temperature. A substring check ("gpt-5" in name) misfires on model names
+# such as "gpt-5.4-mini" that happen to share a prefix but support the full
+# chat-completion parameter set. Match explicitly:
+#   o1, o3, o4, o5 (with optional "-suffix" like -preview, -mini, -pro)
+#   gpt-5 exactly, or gpt-5 followed by an alphabetic suffix (gpt-5o),
+#     each optionally followed by "-suffix" (gpt-5-mini, gpt-5o-mini).
+# A digit or "." after the core identifier breaks the match, so gpt-5.4-mini,
+# gpt-5.5, etc. are treated as regular chat models.
+_REASONING_MODEL_PATTERN = re.compile(
+    r"^(?:o[1-9])(?:-[a-z0-9]+(?:-[a-z0-9]+)*)?$"
+    r"|^gpt-5[a-z]*(?:-[a-z0-9]+(?:-[a-z0-9]+)*)?$"
+)
 
 
 class LLMBase(ABC):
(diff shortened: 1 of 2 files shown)


kartik-mem0:

but closing this pr in favour of this #4746

PR #4903, opened 2026-04-20
fix: resolve semantic conflict resolution in memory architecture

### LINKED ISSUE #4896: ADD-only architecture doesn't implement conflict resolution for semantically similar memories
### Component

Core / Python SDK

### Description

The v2.0.0 release introduced an ADD-only extraction architecture, but the implementation doesn't match the documentation. According to the docs, the system should perform conflict resolution where "Latest truth wins when contradictions detected", but in practice:

1. **Exact duplicate detection only**: The code only uses MD5 hash-based deduplication (`main.py:784-803`)
2. **No semantic conflict resolution**: When adding contradictory facts like:
   - First: "My name is liguoyu" (Li Guoyu)
   - Second: "My name is liguosong" (Li Guosong)

Both memories are stored as separate ADD events instead of resolving the conflict.

### Expected vs Actual Behavior

**Expected** (per documentation):
- Adding "My name is liguosong" after "My name is liguoyu" should either:
  - Return UPDATE event (old behavior in v1.x)
  - Or deduplicate semantically similar content

**Actual**:
- Both return ADD event
- Both memories remain in storage
- Search returns both with similar scores

### Code Evidence

In `mem0/memory/main.py`, the deduplication logic only checks exact hash matches:

```python
# Phase 5: Hash dedup
mem_hash = hashlib.md5(text.encode()).hexdigest()
if mem_hash in existing_hashes or mem_hash in seen_hashes:
    logger.debug(f"Skipping duplicate memory (hash match): {text[:50]}")
    continue
```

This only prevents exact duplicates, not semantically similar or contradictory facts.

### Environment

- mem0 version: 2.0.0
- Python version: 3.x
- Vector store: Qdrant (local)

### Steps to Reproduce

```python
from mem0 import Memory

memory = Memory()

# Add first memory
r1 = memory.add('my name is LGY', user_id='test')
print(r1)  # {'event': 'ADD'}

# Add contradictory memory
r2 = memory.add('my name is LGS', user_id='test')
print(r2)  # Expected: {'event': 'UPDATE'}, Actual: {'event': 'ADD'}

# Check all memories
all_mem = memory.get_all(user_id='test')
print(len(all_mem['results']))  # Expected: 1, Actual: 2
```

Parteeksachdeva:
## Linked Issue

Closes #4896

---

## Description

This PR fixes a bug in the memory addition architecture where adding contradictory but semantically similar facts produced duplicate `ADD` events (ignoring the **"Latest truth wins"** paradigm) or erroneously updated the memory node while retaining the original outdated string.

### Root Cause

The `DEFAULT_UPDATE_MEMORY_PROMPT` previously instructed the Memory LLM to **`DELETE` contradictions**, which prevented safe overwriting behavior.

When the LLM was forced to `UPDATE`, it frequently hallucinated and replaced the `text` field with the **old string** rather than the newly conflicting fact.

---

## Fix Details

* Removed the conflicting instruction that attempted to `DELETE` a memory upon contradiction.
* Added a **minimal, explicit instruction** under the **`#2 Update`** logic clarifying that:

  > Contradictions **must trigger an `UPDATE` event**, replacing the old memory with the newly retrieved fact (**latest truth wins**).

This ensures that:

* The existing memory node is reused.
* The outdated value is replaced.
* The system maintains a single authoritative version of the fact.

---

## Type of Change

* [x] Bug fix (non-breaking change that fixes an issue)
* [ ] New feature (non-breaking change that adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [x] Refactor (no functional changes)
* [ ] Documentation update

---

## Breaking Changes

N/A

---

## Test Coverage

* [ ] I added/updated unit tests
* [ ] I added/updated integration tests
* [x] I tested manually (see below)
* [ ] No tests needed (explain why)

---

## Manual Reproduction & Testing

Verified locally using an OpenAI model to process contradictory inputs.

### Test Script

```python
from mem0 import Memory

memory = Memory()

# Sequence of contradictory updates
r1 = memory.add('my name is LGY', user_id='test')
r2 = memory.add('my name is LGS', user_id='test')

all_mem = memory.get_all(user_id='test')

print("r1:", r1)
print("r2:", r2)
print("all_mem count:", len(all_mem))
print(all_mem[0])
```

---

## Before the Fix

The LLM incorrectly generated an `UPDATE` event that retained the original outdated memory, effectively discarding the new truth.

```text
r1: {'results': [{'id': '63e1863a-677f-4524-b621-72cfefd77e09', 'memory': 'Name is LGY', 'event': 'ADD'}]}

r2: {'results': [{'id': '63e1863a-677f-4524-b621-72cfefd77e09',
'memory': 'Name is LGY',
'event': 'UPDATE',
'previous_memory': 'Name is LGY'}]}

all_mem count: 1

{'id': '63e1863a-677f-4524-b621-72cfefd77e09',
'memory': 'Name is LGY',
'hash': '374dd5faee0e17864f3adc545cbd2de0',
'metadata': None,
'created_at': '2026-04-20T15:22:03.559956+00:00',
'updated_at': '2026-04-20T15:22:06.793354+00:00',
'user_id': 'test'}
```

---

## After the Fix

The system successfully overrides the old memory element and cleanly captures the **latest truth** inside the `.text` property, matching the expected behavior.

```text
r1: {'results': [{'id': '528b0e2c-abec-4b81-bfb4-f080d897a364',
'memory': 'Name is LGY',
'event': 'ADD'}]}

r2: {'results': [{'id': '528b0e2c-abec-4b81-bfb4-f080d897a364',
'memory': 'Name is LGS',
'event': 'UPDATE',
'previous_memory': 'Name is LGY'}]}

all_mem count: 1

{'id': '528b0e2c-abec-4b81-bfb4-f080d897a364',
'memory': 'Name is LGS',
'hash': '0159766a86ff0e803a14b8eb2b78f3dd',
'metadata': None,
'created_at': '2026-04-20T15:40:00.306695+00:00',
'updated_at': '2026-04-20T15:40:03.393890+00:00',
'user_id': 'test'}
```

---

## Summary

This fix restores correct **"Latest truth wins"** behavior by ensuring that contradictory updates overwrite outdated memory values instead of deleting or preserving stale content. It also reduces hallucination risk during `UPDATE` events by clarifying expected overwrite semantics.

### DIFF
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -212,7 +212,7 @@ There are specific guidelines to select which operation to perform:
 
         }
 
-2. **Update**: If the retrieved facts contain information that is already present in the memory but the information is totally different, then you have to update it. 
+2. **Update**: If the retrieved facts contain information that contradicts the information present in the memory or is already present in the memory but the information is totally different, then you have to update it (replace the old memory with the new fact). 
 If the retrieved fact contains information that conveys the same thing as the elements present in the memory, then you have to keep the fact which has the most information. 
 Example (a) -- if the memory contains "User likes to play cricket" and the retrieved fact is "Loves to play cricket with friends", then update the memory with the retrieved facts.
 Example (b) -- if the memory contains "Likes cheese pizza" and the retrieved fact is "Loves cheese pizza", then you do not need to update it because they convey the same information.
@@ -260,7 +260,7 @@ Please note to return the IDs in the output from the input IDs only and do not g
         }
 
 
-3. **Delete**: If the retrieved facts contain information that contradicts the information present in the memory, then you have to delete it. Or if the direction is to delete the memory, then you have to delete it.
+3. **Delete**: If the direction is to delete the memory, then you have to delete it.
 Please note to return the IDs in the output from the input IDs only and do not generate any new ID.
 - **Example**:
     - Old Memory:


kartik-mem0:
Hi @Parteeksachdeva, thanks for investigating this and the thorough manual testing!

A couple of things to note:

1. **The linked issue #4896 has been closed** — our v3 SDK handles contradictions by design through transition-aware extraction and memory linking via linked_memory_ids, rather than through an explicit UPDATE/DELETE code path. Both memories being stored is intentional behavior.

2. **The prompt this PR modifies (DEFAULT_UPDATE_MEMORY_PROMPT) is the legacy v1 prompt** used only by the direct memory.update() code path — it's not involved in the v3 add() pipeline at all. The v3 add() flow uses ADDITIVE_EXTRACTION_PROMPT, which has its own approach to handling contradictions through transition capture and memory linking.

So while the analysis of the legacy prompt's DELETE-vs-UPDATE behavior is accurate, it doesn't affect the v3 add() pipeline where the issue was originally reported.

We appreciate the effort and the clear before/after evidence. Thanks for contributing!

PR #4909, opened 2026-04-20
fix: handle list-typed content when vision is disabled in parse_vision_messages

### LINKED ISSUE #4799: fix: AttributeError crash in Memory.add() when message content is a list and vision is disabled
### Component

Core / Python SDK

### Summary

---

### Steps to Reproduce

```python

messages = [
    {
        "role": "user",
    }
]

```

Or end-to-end:

```python
from mem0 import Memory

m = Memory()
m.add(
    user_id="alice"
)
```

---

### Expected Behavior

---

### Actual Behavior

```
               ^^^^^^^^^^^^^^^^^^^^^
```

---

### Environment

- mem0 version: 1.0.11
- Python version: 3.11.9

nileshpatil6:
## Problem

`Memory.add()` crashes with `AttributeError: 'NoneType' object has no attribute 'generate_response'` when a message uses the standard multimodal format (`content` is a list of content blocks) and vision is not enabled (the default).

```python
m = Memory()
m.add(
    user_id="alice"
)  # crashes
```

Root cause: `parse_vision_messages` unconditionally calls `get_image_description(msg, llm, ...)` when `content` is a list, even when `llm=None` (vision disabled).

## Fix

When `llm is None` and content is a list, extract the text parts and join them — skipping image blocks — instead of calling the vision model.

```python
if isinstance(msg["content"], list):
    if llm is None:
        text_parts = [block["text"] for block in msg["content"] if block.get("type") == "text"]
        returned_messages.append({"role": msg["role"], "content": " ".join(text_parts)})
    else:
        returned_messages.append({"role": msg["role"], "content": description})
```

Fixes #4799


kartik-mem0:
Thanks for taking the time to fix this, @nileshpatil6 — and sorry for the duplicate effort here. This exact root cause (`parse_vision_messages` crashing on list/array `content` when vision is disabled, `llm=None`) was just fixed and merged in #5487, so I am closing this one as a duplicate so the fix stays in a single place. Really appreciate the clean repro and tests — hope to see you on other issues!

PR #4911, opened 2026-04-21
fix: LLM-driven contradiction resolution in v2 ADD-only pipeline

### LINKED ISSUE #4896: ADD-only architecture doesn't implement conflict resolution for semantically similar memories

### LINKED ISSUE #4904: fix: semantic conflict resolution missing in v2 ADD-only pipeline
## Problem

When contradictory facts are added sequentially for the same user (e.g., \"my name is Alice\" followed by \"my name is Bob\"), both are stored as separate ADD events. The second add should detect the contradiction and produce an UPDATE event, replacing the outdated fact. Instead:

- Both memories are stored permanently
- `get_all()` returns 2 results instead of 1
- Search returns both with similar scores, degrading retrieval quality

**Steps to reproduce:**
\`\`\`python
from mem0 import Memory
memory = Memory()
memory.delete_all(user_id="test")

r1 = memory.add("my name is Alice", user_id="test")
print(r1)  # {'results': [{'event': 'ADD', ...}]}

r2 = memory.add("my name is Bob", user_id="test")
print(r2)  # actual: {'results': [{'event': 'ADD', ...}]}
           # expected: {'results': [{'event': 'UPDATE', ...}]}

all_mem = memory.get_all(user_id="test")
print(len(all_mem["results"]))  # actual: 2, expected: 1
\`\`\`

## Root Cause Analysis

The v2.0.0 `add` pipeline uses an ADD-only extraction architecture. The pipeline flow is:

1. **Existing memory retrieval** — vector search fetches semantically related memories and passes them to the LLM as context
2. **LLM extraction** — the extraction prompt instructs the LLM to skip memories that are *semantically equivalent* to an existing one, but says nothing about *contradictions*. Contradictory facts (same slot, different value) are not equivalent — the LLM correctly treats them as new information and emits a new ADD.
3. **Hash deduplication** — the only dedup after LLM extraction is an exact MD5 hash match, which only catches byte-for-byte identical text.

There is no step that compares a newly extracted memory against existing memories semantically, and no UPDATE code path exists in the v2 pipeline. The documentation's promise of "latest truth wins when contradictions detected" is unimplemented.

The existing memory embeddings are already available from the vector search results in step 1, and new memory embeddings are already computed in Phase 3 — so conflict detection does not require additional embedding calls. What's missing is the comparison logic and the UPDATE execution path.

## TDD Fix Plan

1. **RED**: Write a test asserting that `memory.add("my name is Bob", user_id=X)` — when "my name is Alice" already exists for user X — returns an event of type `UPDATE`, not `ADD`.
   **GREEN**: After new memory embeddings are computed, compare each new memory's embedding against existing memory vectors using cosine similarity. If similarity exceeds a threshold (≥ 0.85), mark the new memory as an UPDATE candidate targeting the most similar existing memory's ID.

2. **RED**: Write a test asserting that after the contradictory add, `memory.get_all(user_id=X)` returns exactly 1 result containing the newer value ("Bob"), not 2.
   **GREEN**: For UPDATE candidates in the persistence phase, call the vector store's update method on the existing memory's ID with the new text and embedding, rather than inserting a new entry. Do not insert a new record.

3. **RED**: Write a test asserting that `memory.history(memory_id)` for the updated memory shows an UPDATE event with both `old_memory` ("my name is Alice") and `new_memory` ("my name is Bob") populated.
   **GREEN**: In the history batch builder, emit an UPDATE history record (not ADD) for conflict-resolved memories, with `old_memory` set to the existing memory's text and `new_memory` set to the replacement text.

4. **RED**: Write a test asserting that adding an *unrelated* fact ("I have a dog") after "my name is Alice" still produces an ADD event and results in 2 stored memories.
   **GREEN**: Confirm the similarity threshold prevents false positives — unrelated memories have low cosine similarity and should not trigger the UPDATE path.

**REFACTOR**: Extract the cosine similarity comparison and conflict candidate selection into a standalone helper function so it can be unit-tested independently of the full pipeline.

## Acceptance Criteria

- [ ] Adding a contradictory single-slot fact (name, location, preference) returns an UPDATE event
- [ ] After an UPDATE, `get_all()` returns 1 result containing the newer value
- [ ] `history()` for the updated memory shows UPDATE event with old and new values
- [ ] Unrelated additions still produce ADD events (no false positives)
- [ ] Exact duplicate detection (MD5 hash) still works as before
- [ ] All new tests pass
- [ ] Existing tests still pass

besaliu:
## Linked Issue

Closes mem0ai/mem0#4904
Closes mem0ai/mem0#4896

## Description

When a user updates a single-slot fact (name, job title, relationship status), the v2 ADD-only pipeline was accumulating contradictory memories — e.g. both "User's name is Alice" and "User's name is Bob" — instead of replacing the outdated one.

This PR fixes that by teaching the LLM to signal contradictions explicitly via a new `contradicts` field in its output, then routing those memories to `_update_memory` instead of insert.

**What changed:**
- `mem0/configs/prompts.py`: Added a "Contradiction Detection" section to `ADDITIVE_EXTRACTION_PROMPT` that defines the `contradicts` field, explains when to use it vs `linked_memory_ids`, and includes worked examples (Example 13: name change, Example 14: additive preferences)
- `mem0/memory/main.py`: Added Phase 2.5 (collect `contradicts` markers from LLM output → resolve integer IDs to real UUIDs via `uuid_mapping`) and Phase 3.5 (call `_update_memory` for flagged memories after embeddings are computed in Phase 3). Memories resolved as contradictions are skipped in the ADD flow and returned as UPDATE events.
- `tests/test_conflict_resolution.py`: New test file with 8 tests covering the full behavior surface

**Why this approach instead of similarity thresholds:**
A cosine-similarity gate (≥ 0.85) has a critical false-positive problem: "User loves coffee" and "User loves tea" score ~0.90+ and would incorrectly UPDATE coffee with tea. The LLM already receives all top-10 existing memories in its context window and can distinguish mutually-exclusive facts from additive preferences. Adding a `contradicts` field costs zero extra LLM calls and zero extra vector searches.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

`tests/test_conflict_resolution.py` — 8 tests, all passing:

| Test | Behavior verified |
|------|-------------------|
| `test_contradictory_name_returns_update_event` | Contradiction → UPDATE event, not ADD |
| `test_contradictory_name_update_targets_correct_memory` | UPDATE carries the existing memory's UUID |
| `test_conflict_calls_update_not_insert` | `vector_store.update()` called, `insert()` not called |
| `test_get_all_returns_one_result_after_conflict` | One record in store after resolution, not two |
| `test_conflict_writes_update_history` | `db.add_history()` called with correct old/new text and event=UPDATE |
| `test_no_contradicts_field_produces_add` | Additive fact (coffee→tea) → ADD, no false positive |
| `test_no_existing_memories_produces_add` | Empty store → ADD baseline |
| `test_contradicts_id_not_in_mapping_falls_back_to_add` | Hallucinated LLM ID → warning + fallback ADD, no crash |

## Checklist

### DIFF
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -696,9 +696,27 @@ When extracting a new memory, check if it relates to any Existing Memory. Add re
 - **Same entity/topic**: New fact about a person, place, or thing already mentioned
 - **Updated preference**: A changed or evolved opinion on something previously captured
 - **Continuation**: Follow-up event or next step in a previously captured narrative
-- **Contradiction**: New information that conflicts with an existing memory
 
 Do NOT link memories that merely share a vague theme. Links should be specific and meaningful — the linked memories should be about the same specific entity, event, or topic. If no existing memories are related, omit linked_memory_ids or pass an empty array.
+
+
+## Contradiction Detection
+
+When a new memory is **mutually exclusive** with an existing memory — meaning the old fact can no longer be true once the new one is accepted — set `"contradicts": "<existing-memory-integer-id>"` using that memory's integer ID from the Existing Memories list. Do NOT use `linked_memory_ids` for contradictions; use `contradicts` instead.
+
+**CONTRADICTIONS** (one truth replaces the other — use `contradicts`):
+- Name changed: "User's name is Alice" → "User's name is Bob"
+- Primary job changed: "User is a software engineer" → "User is a product manager"
+- Relationship status changed: "User is single" → "User is married to Elena"
+- Home city changed: "User lives in New York" → "User lives in San Francisco"
+
+**NOT contradictions** (both can be true simultaneously — use `linked_memory_ids` or nothing):
+- "User loves coffee" + "User loves tea" — two separate preferences, both valid
+- "User speaks Python" + "User speaks JavaScript" — two skills, both valid
+- "User has a dog named Max" + "User has a cat named Luna" — additive facts
+- Any new event, experience, or activity involving a known entity
+
+If a new memory does not contradict any existing memory, omit `contradicts` entirely.
 
 
 # EXAMPLES
@@ -856,6 +874,36 @@ Output:
 ]}
 
 Both new memories link to related existing memories — the vet checkup links to the existing Poppy memory, and the team switch links to the existing Shopify memory. This enables the system to build a graph of related memories.
+
+
+## Example 13: Contradiction — Single-Slot Fact Replaced
+
+Existing Memories: [{"id": "0", "text": "User's name is Alice"}]
+New Messages:
+[{"role": "user", "content": "Actually, my name is Bob, not Alice."}]
+Observation Date: 2025-08-19
+
+Output:
+{"memory": [
+  {"id": "0", "text": "User's name is Bob", "contradicts": "0"}
+]}
+
+A name is a single-slot fact — only one can be true. `contradicts: "0"` signals that the existing memory should be replaced, not kept alongside. Do NOT use `linked_memory_ids` here.
+
+
+## Example 14: Not a Contradiction — Additive Preferences
+
+Existing Memories: [{"id": "0", "text": "User loves coffee"}]
+New Messages:
+[{"role": "user", "content": "I also really love tea."}]
+Observation Date: 2025-08-19
+
+Output:
+{"memory": [
+  {"id": "0", "text": "User loves tea", "linked_memory_ids": ["<uuid-of-coffee-memory>"]}
+]}
+
+Both preferences can be true simultaneously — this is additive, not a replacement. No `contradicts` field. Use `linked_memory_ids` to connect related preferences.
 
 
 ## Example 11: Long Multi-Topic Conversation — Don't Stop After First Topic
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -767,6 +767,28 @@ class Memory(MemoryBase):
             self.db.save_messages(messages, session_scope)
             return []
 
+        # Phase 2.5: Collect LLM-signalled contradictions
+        # The LLM outputs `"contradicts": "<integer-id>"` when a new fact is mutually
+        # exclusive with an existing memory (e.g. name change, job change). We resolve
+        # the integer ID to the real UUID via uuid_mapping and defer the _update_memory
+        # call until Phase 3 has computed embeddings.
+        update_results = []
+        contradiction_map = {}  # new_text → existing UUID
+
+        for mem in extracted_memories:
+            contradicts_id = mem.get("contradicts")
+            if not contradicts_id:
+                continue
+            existing_uuid = uuid_mapping.get(str(contradicts_id))
+            if not existing_uuid:
+                logger.warning(
+                    f"LLM contradicts ID '{contradicts_id}' not in uuid_mapping; treating as ADD"
+                )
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Hi @besaliu, thanks for the detailed analysis and well-structured PR!

However, our v3 SDK handles contradictions differently by design — through the prompt and memory linking architecture rather than an explicit UPDATE code path:

1. **Transition capture**: Our extraction prompt explicitly instructs the LLM to capture transitions and changes with full context (e.g., "User switched from Alice to Bob" rather than just "User's name is Bob"). This preserves the change history as a richer memory rather than silently replacing the old one.

2. **Memory linking**: Contradictions are connected via linked_memory_ids to the original memory. Our entity store and search ranking use these links to surface the most relevant (typically newest) memory through recency + relevance scoring.

3. **History preservation**: Our design philosophy is append-only with links. Silently replacing old memories destroys context that can be valuable (when did the change happen, what was the old value, why did it change).

The behavior you're seeing (both memories stored) is intentional — the search/retrieval layer is designed to surface the correct current memory through scoring, while keeping the historical record intact.

We appreciate the thorough TDD approach and the well-thought-out examples. If you notice cases where search is surfacing outdated memories over current ones despite linking, that would be a search ranking issue we'd want to hear about separately.

Thanks for contributing!

PR #4912, opened 2026-04-21
fix: paginate delete_all() to remove all matching memories

VANDRANKI:
Fixes #4869.

**Bug**

`Memory.delete_all()` calls `vector_store.list(filters=filters)` once with no explicit limit. Vector stores that apply a default page size (e.g. qdrant defaults to `top_k=100`) silently return only the first page, so any memories beyond that limit are never deleted. The async `AsyncMemory.delete_all()` has the same issue.

**Fix**

Loop in batches of 1000 until the vector store returns an empty page or fewer results than the batch size:

```python
BATCH_SIZE = 1000
deleted = 0
while True:
    batch = self.vector_store.list(filters=filters, top_k=BATCH_SIZE)[0]
    if not batch:
        break
    for memory in batch:
    deleted += len(batch)
    if len(batch) < BATCH_SIZE:
        break
```

The same loop structure is applied to `AsyncMemory.delete_all()`.

**Why this approach**

- Works across all vector store backends without requiring changes to individual `list()` implementations.
- Deletes in batches rather than loading all IDs into memory at once.
- The termination condition (`len(batch) < BATCH_SIZE`) handles the last partial page correctly without an extra round trip.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1561,12 +1561,22 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
-        memories = self.vector_store.list(filters=filters)[0]
-        for memory in memories:
-            self._delete_memory(memory.id)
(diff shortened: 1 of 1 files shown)


youneshima:
Closing as a duplicate of #4872, which fixes both the sync and async delete_all paths with a normalize helper and regression tests, and is passing CI. Thanks for the contribution and apologies for the overlap.

PR #4917, opened 2026-04-21
feat(mem0-plugin): add Codex lifecycle hooks via opt-in installer

gabrielstein-mem0:
## What

This PR bundles three independent improvements to the `mem0-plugin`. Initially separate, they ended up on one branch because they share the v0.1.0 → v0.1.1 manifest bump and were developed back-to-back. Reviewers can treat them independently — each lives in its own commit.

---

### 1. Codex lifecycle hooks (`e959f4d7`)

Brings opt-in lifecycle hooks to the **Codex** side of the plugin. Users opt in with one command after installing the plugin:

```bash
python3 /path/to/mem0-plugin/scripts/install_codex_hooks.py
```

Three events get wired up in `~/.codex/hooks.json`:

| Event | What it does |
|-------|--------------|
| `SessionStart` | Loads prior memories as bootstrap context |
| `UserPromptSubmit` | Injects relevant memories into each prompt |
| `Stop` | Reminds the agent to persist learnings at turn end |

#### Why an installer instead of a manifest field

Claude Code and Cursor plugin hosts auto-wire hooks from the plugin manifest (expanding `${CLAUDE_PLUGIN_ROOT}` / `${CURSOR_PLUGIN_ROOT}` at runtime). [Codex hooks](https://developers.openai.com/codex/hooks) only documents discovery at `~/.codex/hooks.json` and `<repo>/.codex/hooks.json` — no plugin-host mechanism, no plugin-root variable. Rather than rely on undocumented behavior, the installer targets the documented discovery path directly using absolute paths.

#### Behavior

- **Idempotent** — re-running replaces the Mem0 entries instead of duplicating, so upgrades are safe
- **Non-destructive** — preserves any other hook entries the user already has
- **Uninstallable** — `--uninstall` cleanly removes only Mem0 entries
- **Config-aware** — detects the `codex_hooks = true` feature flag in `~/.codex/config.toml` and prints setup instructions if missing

#### Notes

- Python 3 stdlib only — no new dependencies
- `PreToolUse` / `PermissionRequest` / `PostToolUse` are intentionally left off: Codex currently restricts them to Bash, which doesn't match the Claude "block writes to MEMORY.md" filter that relied on `Write|Edit`

---

### 2. Stop hook JSON-output fix (`5f38c8fa`)

The original Stop handler `cat`d a plain-text reminder to stdout. Per [the Codex hooks docs](https://developers.openai.com/codex/hooks):

> `Stop` expects JSON on `stdout` when it exits `0`. Plain text output is invalid for this event.

Codex was silently dropping the reminder — nothing errored, the agent just never saw the prompt to persist learnings. The fix:

- Emit `{"decision":"block","reason":"<reminder>"}` so Codex re-opens the turn with the reminder injected as continuation context.
- Honor `stop_hook_active` and emit `{"continue":true}` when the turn was already continued. Without this guard, every `block` reopens the turn, fires Stop again, blocks again — infinite loop.

Worth flagging separately because it's a silent failure: a reviewer reading just the original installer commit wouldn't catch that the Stop reminder never reached the agent.

---

### 3. mem0-skill rewrite for v3 API (`2f5e62cb`)

Larger than it sounds: ~1,500 lines across `mem0-plugin/skills/mem0/`. The skill still described v2 SDK behavior, so the agent was being instructed to write code that no longer works (e.g. passing `user_id="..."` to `search` instead of `filters={"user_id": "..."}`).

- All `search` / `get_all` examples migrated from `user_id="..."` (v2) to `filters={"user_id": "..."}` (v3) across SKILL.md and every reference file.
- Removed deprecated v2 options: `immutable`, `expiration_date`, `includes` / `excludes`, `async_mode`, `keyword_search`, `filter_memories`, `enable_graph`.
- Documented v3 defaults (`top_k=20`, `threshold=0.1`, ADD-only events, paginated `get_all` envelope).
- `api-reference.md` switched to `/v3/memories/...` endpoints; `architecture.md` simplified to single-pass extraction; "Graph Memory" replaced with "Entity Linking" (the new built-in v3 hybrid search).
- All examples bumped from `gpt-4.1-nano-2025-04-14` to `gpt-5-mini`.
- New language-specific deep references: `client/python.md`, `client/node.md`, `client/differences.md` — full MemoryClient + AsyncMemoryClient + Memory (OSS) APIs for both languages.
- SKILL.md gained a Skill Graph block linking to `mem0-cli` and `mem0-vercel-ai-sdk` skills with explicit "DO NOT TRIGGER" rules to keep ambiguous queries routed correctly.

Plugin/marketplace manifests bumped 0.1.0 → 0.1.1 across `.claude-plugin/`, `.cursor-plugin/`, `.codex-plugin/`, and the two top-level `marketplace.json` files.

---

## README

Component table now shows `Opt-in` for Codex lifecycle hooks, with a step-by-step "Optional — enable lifecycle hooks" section in the Codex install flow.

## Coordination

Companion docs PR is #4951 (rewrites the Codex install flow to document the installer added here). Merge order: this PR first, then #4951.

## Splitting

If reviewers prefer, commit `2f5e62cb` (the skill rewrite) is self-contained and could cherry-pick into its own PR — happy to split.


soumil-rathi:
Small non-blocking issue: 

The frontmatter line "SDK v3 with v2 compatibility mode available" is misleading — v3 doesn't accept top-level user_id/agent_id right?

Could be an issue if someone decides to actually call the api directly. 

Approved, you can make this change and lmk if I need to approve again

soumil-rathi (dismissed):

soumil-rathi (approved):

whysosaket (approved):

PR #4928, opened 2026-04-22
make _iso() timezone handling explicit and add tests

21vedansh:
Removes bare except in `_iso()` and replaces it with explicit naive vs aware datetime handling.

- non-datetime -> None  
- naive datetime -> assume UTC (replace)  
- aware datetime -> convert to UTC (astimezone)  

No behavior change intended but it avoids masking unexpected errors.

Note: Tests for `_iso()` are included. Local execution on Python 3.10 fails due to `datetime.UTC` usage in models (Python 3.11+), which appears inconsistent with the declared `>=3.10` support. 
although this is unrelated to the changes in this PR.

### DIFF
--- a/openmemory/api/app/routers/backup.py
+++ b/openmemory/api/app/routers/backup.py
@@ -30,13 +30,14 @@ class ExportRequest(BaseModel):
     to_date: Optional[int] = None
     include_vectors: bool = True
 
-def _iso(dt: Optional[datetime]) -> Optional[str]: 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for the contribution. The OpenMemory project has been deprecated and is no longer maintained. We're sorry about this, we appreciate the effort you put in and we should have communicated the deprecation earlier. Closing this PR.

PR #4937, opened 2026-04-23
fix(llms): fall back to reasoning_content when content is empty on OpenAI-compatible APIs

### LINKED ISSUE #4932: Fact extraction returns empty/garbage with reasoning models on OpenAI-compatible APIs that separate reasoning_content from content
### 🐛 Describe the bug

Modern OpenAI-compatible endpoints (Ollama Cloud, DeepSeek API, GLM's native API, etc.) separate reasoning-model output into two fields in the response:

- `choices[0].message.content` — final answer only
- `choices[0].message.reasoning_content` — chain-of-thought (or `thinking` in some providers)

When a reasoning model (e.g. `glm-5.1`, `gemma3:12b`, `deepseek-r1`) is used as the Mem0 extractor through these endpoints, the model often emits its *entire* reasoning in `reasoning_content` and leaves `content` either empty or containing a truncated/malformed JSON fragment. Mem0 reads only `content`, producing:

- Empty fact lists (silent — no error, just `"results": []`), OR
- Partial JSON that may deserialize into garbage facts with no real signal.

This differs from the earlier #2614 (closed as stale), where the `<think>...</think>` block was *inside* `content`. On newer endpoints, providers correctly separate the two streams — mem0 just isn't aware of the second one.

### Environment

- `mem0ai` version: `0.1.x` (confirmed on `main` as of 2026-04-22)
- LLM provider: LiteLLM proxy routing `openai/<model>` with `api_base=https://ollama.com/v1`
- Reasoning models tested: `glm-5.1`, `gemma3:12b`, `minimax-m2.7`, `qwen3-vl:235b-instruct`
- Non-reasoning control: `qwen3-coder-next` — works perfectly

### Reproduction

```python
from mem0 import Memory

config = {
    "llm": {
        "provider": "openai",
        "config": {
            "model": "glm-5.1",
            "openai_base_url": "https://ollama.com/v1",
            "api_key": "<ollama-cloud-key>",
        },
    },
    "embedder": { ... },
    "vector_store": { ... },
}
r = m.add("Juanka owns Estudio Fárrago and lives in Guadalajara.", user_id="juanka")
print(r)  # {"results": []} — silent empty, despite obvious facts
```

Switch the extractor to any non-reasoning model (e.g. `qwen3-coder-next`, `gpt-4o-mini`) with the same prompt → extracts 2 facts correctly.

### Evidence the bug is in the response shape

Direct call to the provider (bypassing mem0) with `glm-5.1` returns:

```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "",
      "reasoning_content": "<long chain-of-thought that ends with the facts JSON>"
    }
  }]
}
```

`content == ""` consistently. Mem0's `_add_to_vector_store` reads only `content`, calls `remove_code_blocks("")`, and `json.loads` fails → the `except Exception` at `mem0/memory/main.py:531` swallows it silently, setting `new_retrieved_facts = []`.

### Why silent failure makes this painful

There's no way for the caller to distinguish:
1. "Mem0 extracted the input cleanly and correctly decided there were zero new facts."
2. "Mem0's LLM call returned unusable output and we got nothing."

Both surface as `{"results": []}`. Related discussion in #4831 (closed by the filer as user-error, but the underlying silent-failure concern stands).

### Suggested fix — one of

**Option A (minimal, provider-transparent)**: in the LLM response handler, when `content` is empty or unparseable, fall back to `reasoning_content` and attempt the same `remove_code_blocks → json.loads` pipeline on it. Most reasoning models end their thought with the JSON, so a simple "find last `{...}` block" heuristic works.

**Option B (stricter)**: raise a typed exception (`Mem0ReasoningContentNotSupported`) when the provider returns empty `content` + populated `reasoning_content`, so the caller knows the extractor config is incompatible with this model and can switch. Better for observability.

**Option C (documented workaround)**: if A/B are out of scope, document in the README that extractors must be non-reasoning models for OpenAI-compatible providers that separate the streams, with a minimal compatibility table.

Happy to PR Option A if the direction is accepted.

### Workaround (current)

Use a non-reasoning model as extractor. In our setup (LiteLLM proxy in front of Ollama Cloud), migrated from `glm-5.1` → `qwen3-coder-next`. Latency dropped from ~3s (most of it reasoning tokens discarded anyway) to ~0.7s, and fact extraction became clean and consistent.

### Related

- #2614 — same class of bug, pre-`reasoning_content` era (stale; maintainer invited reopen)
- #3918 — "unterminated JSON" likely same root cause when reasoning gets cut off mid-JSON
- #4054 — Groq/HF JSON parse failure, same silent-empty failure mode

octo-patch:
Fixes #4932

## Problem

Several OpenAI-compatible endpoints (Ollama Cloud, DeepSeek API, GLM native API, etc.) handle reasoning model output by separating it into two fields:

- `choices[0].message.content` — final answer (often empty or `""`)
- `choices[0].message.reasoning_content` — full chain-of-thought with the facts

When `content` is empty, `_parse_response` returns `""`, `remove_code_blocks("")` returns `""`, `json.loads("")` throws, and the `except Exception` block silently sets `new_retrieved_facts = []`. The caller sees `{"results": []}` with no indication that extraction failed.

## Solution

In `_parse_response`, when `content` is falsy, fall back to `getattr(message, "reasoning_content", None)` before returning. This is safe:

- Providers that don't set `reasoning_content` return `None` (same as before).
- Providers that correctly populate `content` are unaffected.
- No new dependency is introduced.

## Testing

Added three unit tests:

1. **`test_generate_response_falls_back_to_reasoning_content`** — verifies the fallback activates when `content=""` and `reasoning_content` is set.
2. **`test_generate_response_content_takes_priority_over_reasoning_content`** — verifies non-empty `content` is returned unchanged.
3. **`test_generate_response_returns_none_when_both_content_fields_empty`** — verifies graceful handling when neither field is present.

All three pass locally.

## Type of Change

## Checklist

### DIFF
--- a/examples/mem0-demo/components/assistant-ui/memory-ui.tsx
+++ b/examples/mem0-demo/components/assistant-ui/memory-ui.tsx
@@ -16,10 +16,11 @@ type RetrievedMemory = {
 
 type NewMemory = {
   id: string;
-  data: {
+  data?: {
     memory: string;
-  };
-  event: "ADD" | "DELETE";
+  } | null;
+  event?: "ADD" | "DELETE";
+  status?: string;
 };
 
 type NewMemoryAnnotation = {
@@ -47,14 +48,18 @@ const useMemories = (): Memory[] => {
     () =>
       annotations?.filter(isMemoryAnnotation).flatMap((a) => {
         if (a.type === "mem0-update") {
-          return a.memories.map(
-            (m): Memory => ({
-              event: m.event,
-              id: m.id,
-              memory: m.data.memory,
-              score: 1,
-            })
-          );
+          return a.memories
+            .filter((m): m is NewMemory & { data: { memory: string }; event: "ADD" | "DELETE" } =>
+              m.data != null && m.event != null
+            )
+            .map(
+              (m): Memory => ({
+                event: m.event,
+                id: m.id,
+                memory: m.data.memory,
+                score: 1,
+              })
+            );
         } else if (a.type === "mem0-get") {
           return a.memories.map((m) => ({
             event: "GET",
--- a/mem0/llms/openai.py
+++ b/mem0/llms/openai.py
@@ -79,7 +79,14 @@ class OpenAILLM(LLMBase):
 
             return processed_response
         else:
-            return response.choices[0].message.content
+            content = response.choices[0].message.content
+            if not content:
+                # Some OpenAI-compatible endpoints (Ollama, DeepSeek, GLM, etc.) separate
+                # chain-of-thought into `reasoning_content` and leave `content` empty when
+                # a reasoning model is used. Fall back to `reasoning_content` so fact
+                # extraction still works instead of silently returning an empty result.
+                content = getattr(response.choices[0].message, "reasoning_content", None)
+            return content
 
     def generate_response(
         self,
--- a/tests/llms/test_openai.py
+++ b/tests/llms/test_openai.py
@@ -322,3 +322,62 @@ def test_callback_with_tools(mock_openai_client):
     mock_callback.assert_called_once()
     # Check that tool_calls exists in the message
     assert hasattr(mock_callback.call_args[0][1].choices[0].message, 'tool_calls')
+
+
+def test_generate_response_falls_back_to_reasoning_content(mock_openai_client):
+    """When content is empty and reasoning_content is present, return reasoning_content.
+
+    Some OpenAI-compatible endpoints (Ollama, DeepSeek, GLM, etc.) put the full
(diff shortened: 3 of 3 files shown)


kartik-mem0:
please sign the cla so that i can review this pr

kartik-mem0:
Thanks for the contribution! We asked for the CLA to be signed so we could review this, but have not heard back in a while. Closing for now to keep things tidy. We would genuinely love to land this, please reopen once the CLA is signed.

PR #4908, opened 2026-04-20
fix(proxy): repair broken MemoryClient search path

shafdev:
## Linked Issue

Closes #4907

## Description

This PR fixes two critical bugs in the proxy MemoryClient path that make `Mem0(api_key="...")` completely non-functional:

1. **_fetch_relevant_memories passes rejected top-level entity params**: Entity IDs (user_id, agent_id, run_id) were passed as top-level kwargs to search(), but both Memory.search() and MemoryClient.search() now reject these and require them in a filters dict.

2. **_format_query_with_memories iterates dict keys instead of results**: The MemoryClient branch was calling `for memory in relevant_memories` (iterating dict keys) instead of `for memory in relevant_memories["results"]` (iterating the results list), causing TypeError.

Both code paths and mock tests are now unified since both APIs return identical `{"results": [...], "relations": [...]}` response shape.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (verified with real API keys - Mem0 and OpenAI)
- [ ] No tests needed (explain why)

Updated mock return values in test_proxy.py to match actual API response shape. Added 3 new test cases:
- test_fetch_relevant_memories_with_entity_params: Verifies entity params are passed in filters dict
- test_format_query_with_memories_handles_dict_response: Verifies dict response is correctly parsed
- test_format_query_with_memories_handles_relations: Verifies relations extraction works

All 9 proxy tests pass locally.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed

PR #4940, opened 2026-04-23
fix(proxy): pass entity ids via filters and read results key for MemoryClient

### LINKED ISSUE #4907: [Bug] Proxy MemoryClient search path is broken
### Component

Core / Python SDK

### Summary

The `Mem0` proxy chat path (`mem0/proxy/main.py`) is completely broken when using `MemoryClient` (i.e. `Mem0(api_key="...")`). Two bugs prevent any chat completion from succeeding:

### Bug 1: Top-level entity params rejected by search()

`_fetch_relevant_memories` passes `user_id`, `agent_id`, `run_id` as top-level kwargs to `search()`, but both `Memory.search()` and `MemoryClient.search()` now reject these with:

```
ValueError: Top-level entity parameters {...} are not supported in search(). 
```

They must be passed inside a `filters` dict.

### Bug 2: TypeError iterating dict instead of dict["results"]

Even if search() succeeded, the MemoryClient branch crashes with:

```
TypeError: string indices must be integers, not 'str'
```

Because it does `for memory in relevant_memories` instead of `for memory in relevant_memories["results"]`, iterating dict keys instead of the results list.

## Steps to Reproduce

```python
from mem0.proxy.main import Mem0

client = Mem0(api_key="m0-your-key")
response = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "What do I like?"}],
    user_id="test-user",
)
```

### Expected Behavior
LLM returns response augmented with relevant memories  

### Actual Behavior
Crashes with ValueError on entity params, or TypeError on dict iteration

### Environment

- mem0 version: 1.0.11
- Python/Node version:3.11.9
- OS:macOS

Sanjays2402:
Fixes #4907

## The bugs

`mem0/proxy/main.py` had two unrelated bugs that together broke `Mem0(api_key=...).chat.completions.create()` end-to-end against the hosted MemoryClient:

### Bug 1 — `_fetch_relevant_memories` passed entity ids as top-level kwargs

Both `Memory.search()` (`mem0/memory/main.py:1174`) and `MemoryClient.search()` (`mem0/client/main.py:273`) explicitly reject `user_id` / `agent_id` / `run_id` / `app_id` at the top level and require them inside `filters`. The proxy was passing them as kwargs:

```python
return self.mem0_client.search(
    query=...,
    user_id=user_id,        # ValueError: top-level entity params not supported
    agent_id=agent_id,
    run_id=run_id,
    filters=filters,
    top_k=top_k,
)
```

Every chat completion crashed with `ValueError: Top-level entity parameters {...} are not supported in search().`

### Bug 2 — `_format_query_with_memories` MemoryClient branch iterated dict keys

`MemoryClient.search()` returns `{"results": [...]}` (per its own docstring), but the MemoryClient branch did:

```python
elif isinstance(self.mem0_client, mem0.client.main.MemoryClient):
    memories_text = "\n".join(memory["memory"] for memory in relevant_memories)
```

Iterating the dict yields its keys (strings like `"results"`), and `"results"["memory"]` then crashes with `TypeError: string indices must be integers, not 'str'`.

## Fix

- **`_fetch_relevant_memories`**: build a merged `filters` dict from the entity ids before calling `search()`. Caller-provided keys win over implicit ones (don't silently overwrite user intent).
- **`_format_query_with_memories`**: collapse both backend branches to read `relevant_memories.get("results", [])`, since both `Memory.search()` and `MemoryClient.search()` now return the same `{"results": [...]}` shape. Backward-compatible: if a caller hands in a list directly, it still iterates correctly.

## Why the existing tests didn't catch this

`tests/test_proxy.py` had `mock_memory_client.search.return_value = [{"memory": "..."}]` — a bare list, which is **not** what `MemoryClient.search()` actually returns. Updated those mocks to use the realistic `{"results": [...]}` shape so future regressions surface immediately.

## New regression tests (4)

- `test_fetch_relevant_memories_passes_entities_via_filters` — asserts `user_id` / `agent_id` / `run_id` are NOT passed as top-level kwargs and ARE present inside `filters`.
- `test_fetch_relevant_memories_preserves_caller_filters` — caller-supplied filter keys aren't overwritten by entity-id parameters; unrelated keys (e.g. `category`) are preserved.
- `test_format_query_with_memoryclient_uses_results_key` — proves the formatter actually iterates `["results"]` and surfaces the memory strings (the original bug would crash here).
- `test_format_query_handles_empty_results` — empty result set produces a valid prompt without raising.

## Verification

```
$ python -m pytest tests/test_proxy.py -v
============================== 10 passed in 4.06s ==============================
```

## Notes

- Minimal diff. No new dependencies.
- No public API change.
- Same logic applies to async paths but the async proxy doesn't seem to have an analogous `_fetch_relevant_memories` — happy to mirror the fix there in a follow-up if maintainers point me at it.

### DIFF
--- a/mem0/proxy/main.py
+++ b/mem0/proxy/main.py
@@ -167,23 +167,27 @@ class Completions:
         # Currently, only pass the last 6 messages to the search API to prevent long query
         message_input = [f"{message['role']}: {message['content']}" for message in messages][-6:]
         # TODO: Make it better by summarizing the past conversation
+        # Both ``Memory.search()`` and ``MemoryClient.search()`` reject
+        # top-level entity params and require them inside ``filters`` (#4907).
+        merged_filters = dict(filters) if filters else {}
+        for key, value in (("user_id", user_id), ("agent_id", agent_id), ("run_id", run_id)):
+            if value is not None and key not in merged_filters:
+                merged_filters[key] = value
         return self.mem0_client.search(
             query="\n".join(message_input),
-            user_id=user_id,
-            agent_id=agent_id,
-            run_id=run_id,
-            filters=filters,
+            filters=merged_filters or None,
             top_k=top_k,
         )
 
     def _format_query_with_memories(self, messages, relevant_memories):
-        # Check if self.mem0_client is an instance of Memory or MemoryClient
-
+        # Both ``Memory.search()`` and ``MemoryClient.search()`` return
+        # ``{"results": [...]}`` in the current API; iterate that key for
+        # both backends rather than the dict's top-level keys (#4907).
         entities = []
-        if isinstance(self.mem0_client, mem0.memory.main.Memory):
-            memories_text = "\n".join(memory["memory"] for memory in relevant_memories["results"])
-            if relevant_memories.get("relations"):
-                entities = [entity for entity in relevant_memories["relations"]]
-        elif isinstance(self.mem0_client, mem0.client.main.MemoryClient):
-            memories_text = "\n".join(memory["memory"] for memory in relevant_memories)
+        results = relevant_memories.get("results", []) if isinstance(relevant_memories, dict) else relevant_memories
+        memories_text = "\n".join(memory["memory"] for memory in results)
+        if isinstance(self.mem0_client, mem0.memory.main.Memory) and isinstance(relevant_memories, dict):
+            relations = relevant_memories.get("relations")
+            if relations:
+                entities = list(relations)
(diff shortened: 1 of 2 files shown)


Sanjays2402:
I've signed the CLA

youneshima:
Closing as a duplicate of #4908, which fixes both proxy bugs with regression tests. Thanks for the contribution and apologies for the overlap.

PR #4971, opened 2026-04-25
fix(llms/base): remove dead hasattr branch + add _validate_config return type

GopalGB:
## Summary
Two small clean-ups in `mem0/llms/base.py` `LLMBase._validate_config`:

### 1. Remove dead `hasattr` branch
Current code:
```python
if not hasattr(self.config, "api_key") and not hasattr(self.config, "api_key"):
    # Check if API key is available via environment variable
    # This will be handled by individual providers
    pass
```
Same attribute checked twice → the condition is logically equivalent to a single `not hasattr(...)` (idempotent under AND), and the body is `pass`, so this branch never affects behaviour and only confuses readers. The accompanying comment already documents *why* we deliberately don't enforce `api_key` at the base level (providers fall back to env vars and override `_validate_config` themselves), so this PR replaces the dead `if` with that explanatory comment.

### 2. Add `-> None` return type
`_validate_config` never returns a value. Adding the annotation makes the contract checkable by static analysis and brings it in line with `_is_reasoning_model`, `_get_supported_params`, etc.

## Backward compatibility
No behaviour change. Only documentation and type clarity. The previously-removed `hasattr` branch had a `pass` body, so removing it is mechanically equivalent.

## Verification
`python3 -c "import ast; ast.parse(open('mem0/llms/base.py').read())"` → exit 0.

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -27,7 +27,7 @@ class LLMBase(ABC):
         # Validate configuration
         self._validate_config()
 
-    def _validate_config(self):
+    def _validate_config(self) -> None:
         """
         Validate the configuration.
         Override in subclasses to add provider-specific validation.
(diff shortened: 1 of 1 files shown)


GopalGB:
The Vercel deploy check is failing because the deploy needs maintainer-side authorization (not an issue with the PR itself). Could a maintainer please authorize the Vercel deploy so this can land? CI builds are green for Python 3.10–3.12.

kartik-mem0:

PR #4974, opened 2026-04-25
fix(memory): exclude score from get() to match get_all()/search() shape

alvinttang:
## Summary

`Memory.get()` / `AsyncMemory.get()` returned records with `score: null`, while `get_all()` and `search()` (which both build via the get_all path) explicitly call `model_dump(exclude={"score"})`. Same record, different keys — broke any consumer that shared a single dict-shape contract between point reads and listings.

## Fix

`mem0/memory/main.py`:
- `Memory.get()` (sync, line 1004): add `exclude={"score"}` to `model_dump()`.
- `AsyncMemory.get()` (async, line 2419): same.

2 LOC of production change.

## Test

`tests/test_main.py`:
- `test_get_shape_matches_get_all` (sync)
- `test_async_get_shape_matches_async_get_all` (async)

Both build the same mocked record, call `get()` + `get_all()`, assert key sets match and `"score" not in got`.

`pytest` 414 passed / 19 skipped (network-dependent). `ruff` clean.

## Risk notes
- Callers that previously read `.get("score")` from `Memory.get()` output will see the key absent (it was always `None` anyway). The intended convergence with `get_all()` / `search()` shape — `score` is a search-time field.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1001,7 +1001,7 @@ class Memory(MemoryBase):
             hash=memory.payload.get("hash"),
             created_at=memory.payload.get("created_at"),
             updated_at=memory.payload.get("updated_at"),
-        ).model_dump()
+        ).model_dump(exclude={"score"})
 
         for key in promoted_payload_keys:
             if key in memory.payload:
@@ -2416,7 +2416,7 @@ class AsyncMemory(MemoryBase):
             hash=memory.payload.get("hash"),
             created_at=memory.payload.get("created_at"),
             updated_at=memory.payload.get("updated_at"),
-        ).model_dump()
+        ).model_dump(exclude={"score"})
 
         for key in promoted_payload_keys:
             if key in memory.payload:
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -90,6 +90,85 @@ def test_get(memory_instance):
     assert result["metadata"] == {"extra_field": "extra_value"}
 
 
+def test_get_shape_matches_get_all(memory_instance):
+    """`Memory.get(id)` and `Memory.get_all()[i]` should return the same key set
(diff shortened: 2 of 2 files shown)


xkonjin (commented):
## Review: fix(memory) — exclude score from get() to match get_all()/search() shape

**Overall:** Clean, focused fix with excellent test coverage. This resolves a real API consistency bug where `Memory.get(id)` leaked a `score: None` field that `get_all()` and `search()` correctly excluded.

**Positives:**
- Two-line production change (both sync and async `get()` methods) — minimal blast radius.
- The new `test_get_shape_matches_get_all` and `test_async_get_shape_matches_async_get_all` tests directly assert the contract: same record read via `get()` vs `get_all()` must yield identical key sets.
- Good negative assertion: `"score" not in got` ensures the search-only field is explicitly excluded.

**Potential issues:**
1. **Breaking change risk:** If any downstream consumers depend on `score: None` being present in `get()` results, this will break them. The PR title frames it as a "fix" (shape matching), which is correct, but consider adding a note in the release notes for migration.
2. **Model-level fix vs dump-level fix:** The `Memory` model itself might still carry `score` as an optional field. If other code paths call `.model_dump()` without `exclude={"score"}`, the leak could recur. Consider whether the exclusion should be pushed down into the model via `Field(exclude=True)` or a custom serializer, so it is automatic everywhere.
3. **Test mocks:** The tests mock `vector_store.get` and `vector_store.list`. This validates the serialization layer well, but does not exercise the actual vector store. If a real store populates `score` differently (e.g., omits it entirely vs sets it to `None`), behavior could diverge.

**Security:** No issues. This is a data-shape change, not a permissions or input-handling change.

**Verdict:** Solid fix. Recommend merge. Consider model-level `exclude` as a follow-up to prevent regressions in future read paths.

PR #4979, opened 2026-04-26
fix: support GPT-5 completion token parameter

zew1me:
## Description

Fixes incorrect max token parameter when using ChatGPT 5 via the OpenAI Python SDK.

Previously, requests were using an outdated/incorrect parameter and not passing the correct max token field expected by gpt-5* APIs, which leads to ignored limits. This change ensures the request uses the correct parameter for ChatGPT 5 so response length behaves as intended.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Manually verified by using mem 0 AI. Did not see the same error in my docker logs as I did earlier. 

## Checklist

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -43,16 +43,22 @@ class LLMBase(ABC):
     def _is_reasoning_model(self, model: str) -> bool:
         """
         Check if the model is a reasoning model or GPT-5 series that doesn't support certain parameters.
-        
+
         Args:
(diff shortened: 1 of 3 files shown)


zew1me:
I actually think this implementation is trash although the root cause is real, need to reimplement

zew1me:
closing in favor of #5547  although not sure if it addresses o1 models fully / correctly.

PR #4964, opened 2026-04-24
fix(memory): pass observation timestamp to extraction prompt

wianger:
## Linked Issue

Closes #4963

## Description

This fixes a bug in the OSS Python SDK memory extraction path where `generate_additive_extraction_prompt(...)` supports an observation `timestamp`, but neither the sync nor async `_add_to_vector_store(...)` call site passed it through.

Without this, `Observation Date` falls back to `Current Date`, so historical messages containing relative dates like `last week`, `yesterday`, or `recently` can be anchored to the processing date instead of the conversation date.

The fix adds a small helper to resolve an observation timestamp from existing inputs, preferring `metadata["timestamp"]`, then falling back to `metadata["created_at"]`, message-level `timestamp`, and message-level `created_at`. If no timestamp is provided, behavior is unchanged and the prompt builder keeps defaulting `Observation Date` to `Current Date`.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (describe below)
- [ ] No tests needed (explain why)

```bash
/usr/bin/python3 -m py_compile mem0/memory/main.py tests/memory/test_main.py
env UV_CACHE_DIR=/tmp/uv-cache uv run --extra test pytest tests/memory/test_main.py::TestObservationDatePrompt -q
env UV_CACHE_DIR=/tmp/uv-cache uv run --extra test pytest tests/memory/test_main.py -q
env UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev ruff check mem0/memory/main.py tests/memory/test_main.py
```

Results:

- `TestObservationDatePrompt`: 4 passed
- `tests/memory/test_main.py`: 38 passed
- Ruff: All checks passed

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #4986, opened 2026-04-27
fix: use metadata timestamp for observation date

### LINKED ISSUE #4963: Observation Date falls back to Current Date during memory extraction when historical timestamp is provided
### Component

Core / Python SDK

### Summary

### Steps to Reproduce

```python
from mem0 import Memory

m = Memory()
m.add(
    "I went to Paris last week",
    user_id="u1",
    infer=True,
)
```

### Expected Behavior

```text
## Observation Date
2023-05-24T10:00:00+00:00
```

### Actual Behavior

### Environment

wolfkill:
## Summary
- Pass metadata timestamp into the additive extraction prompt for sync memory adds
- Do the same for async memory adds so historical observations use the caller-provided timestamp
- Add sync and async tests that assert the prompt Observation Date uses metadata["timestamp"]

Fixes #4963

## Testing
- UV_CACHE_DIR=.uv-cache uv run --extra test pytest tests/memory/test_main.py::TestPromptOverridesCustomInstructions -q
- UV_CACHE_DIR=.uv-cache uv run --extra dev ruff check mem0/memory/main.py tests/memory/test_main.py

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -732,6 +732,7 @@ class Memory(MemoryBase):
             existing_memories=existing_memories,
             new_messages=parsed_messages,
             last_k_messages=last_messages,
+            timestamp=metadata.get("timestamp"),
             custom_instructions=custom_instr,
         )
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for this, @wolfkill — and sorry for the duplicate effort. There are a few PRs for #4963; we're consolidating on #4964, which extracts a shared timestamp helper covering metadata, message-level, and `created_at` fallbacks and is green on CI. Closing this one so the work stays in one place — your review on #4964 would be welcome. 🙏

PR #4990, opened 2026-04-27
fix(memory): prevent silent data loss on pgvector embedding dimension mismatch

### LINKED ISSUE #4985: bug(server): switching embedding provider silently drops writes due to vector dimension mismatch
## Description

When the embedding provider is switched (e.g. from OpenAI to Gemini) via the dashboard or `POST /configure`, the pgvector collection retains the original provider's vector dimension. Subsequent memory writes silently fail — the API returns a success response with a memory ID, but nothing is persisted to the database.

## Steps to Reproduce

1. Start the server with `make bootstrap` (defaults to OpenAI embedder — creates `memories` table with `vector(1536)`)
2. Switch the embedder to Gemini (`gemini-embedding-001`) via the dashboard Configuration page
3. `POST /memories` with a message
4. Observe the success response:
```json
{
    "results": [
        {
            "id": "f88e7d7d-b882-4975-b9cf-7f945e244eb3",
            "memory": "My name is John and I am a software engineer",
            "event": "ADD"
        }
    ]
}
```
5. `GET /memories?user_id=...` → `{"results": []}`
6. Query postgres directly:
```sql
SELECT COUNT(*) FROM memories;
-- Returns: 0
```

## Root Cause

There are three layers to this bug:

### Layer 1 — `embedding_model_dims` defaults to 1536 and is never inferred from the embedder

The pgvector config hardcodes `embedding_model_dims = 1536` as its default:

```python
# mem0/configs/vector_stores/pgvector.py
embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
```

This default is OpenAI-specific and silently wrong for every other provider. Critically, `Memory.__init__` instantiates `self.embedding_model` *before* `self.vector_store` and already has the correct dimension available (`self.embedding_model.config.embedding_dims = 768` for Gemini), but never uses it to set `vector_store.config.embedding_model_dims`. The information is present but ignored.

### Layer 2 — Table created with wrong dimensions on every restart

Even after manually dropping the `memories` table and restarting, it is immediately recreated with `vector(1536)` because the pgvector config default (Layer 1) kicks in before any override can correct it.

### Layer 3 — Dimension mismatch fails silently

Inserting a 768-dim vector into a `vector(1536)` column raises no exception to the caller. The API returns 200 with a memory ID as if the write succeeded, but nothing is persisted.

## Expected Behaviour

- `Memory.__init__` should auto-set `vector_store.config.embedding_model_dims` from `self.embedding_model.config.embedding_dims` when not explicitly provided
- Or at minimum, validate that the two match and raise a clear error if they don't
- The dimension mismatch on insert should propagate as an error, not be swallowed silently

## Actual Behaviour

- `memories` table is created with `vector(1536)` regardless of the configured embedder
- `POST /memories` returns 200 with a valid-looking memory ID
- The memory is never actually stored in postgres
- `GET /memories` returns empty results
- No error or warning anywhere in the logs

## Impact

This is a silent data loss bug. Users believe their memories are being stored (success response + memory ID), but they are not. The only way to diagnose it is to directly query the postgres table and inspect vector dimensions — not at all obvious.

This affects every non-OpenAI embedding provider (Gemini, Cohere, etc.) that uses a dimension other than 1536.

## Workaround

Explicitly pass `embedding_model_dims` in the **vector store** config (not the embedder config) via `POST /configure`:

```json
{
  "embedder": {"provider": "gemini", "config": {"model": "models/gemini-embedding-001"}},
  "vector_store": {"provider": "pgvector", "config": {"embedding_model_dims": 768}}
}
```

Then drop and recreate the `memories` table:
```sql
DROP TABLE IF EXISTS memories;
DROP TABLE IF EXISTS mem0migrations;
```
Then restart the container.

## Environment

- OpenAI embedder → Gemini `gemini-embedding-001` (1536 → 768 dims)
- Reproduced on: 2026-04-27

## Related Issues

- #4984 — Dashboard saves wrong API key on provider change; `.env` cannot override database config

lost-particles:
## Linked Issue

Closes #4985

## Description

This PR fixes silent data loss when the configured embedding model’s vector size no longer matches the PostgreSQL/pgvector table (for example after switching from an OpenAI-sized model to another provider). Previously, `PGVectorConfig` defaulted `embedding_model_dims` to `1536`, and the OSS v3 pipeline could swallow `vector_store.insert` failures, so APIs could look successful while nothing was persisted.

Changes:

- Default `embedding_model_dims` to `None` and infer it from the embedder’s `embedding_dims` during `Memory` / `AsyncMemory` initialization when using pgvector; raise a clear `ValueError` if an explicit pgvector dimension conflicts with the embedder.
- On pgvector startup, validate the existing table’s `vector` column dimension against configuration and raise with remediation hints when they differ.
- Refuse `PGVector` construction when `embedding_model_dims` is still unset (no silent table creation with wrong semantics).
- After a failed batch insert in the v3 path, retry per record but **raise** `RuntimeError` if any insert still fails, instead of only logging.
- Add regression tests for inference, mismatch errors, schema validation, and insert failure propagation.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added/updated tests in `tests/test_memory.py`, `tests/memory/test_main.py`, and `tests/vector_stores/test_pgvector.py` (13 targeted cases). Locally: `ruff check mem0/ tests/` and the new regression subset under `pytest` passed. Full `pytest tests/` should be confirmed in CI or via `make test` with the same optional extras as `.github/workflows/ci.yml`.

## Checklist

### DIFF
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel, ConfigDict, Field, model_validator
 class PGVectorConfig(BaseModel):
     dbname: str = Field("postgres", description="Default name for the database")
     collection_name: str = Field("mem0", description="Default name for the collection")
-    embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
+    embedding_model_dims: Optional[int] = Field(None, description="Dimensions of the embedding model. When None, inferred from the configured embedder at Memory init time.")
     user: Optional[str] = Field(None, description="Database user")
     password: Optional[str] = Field(None, description="Database password")
     host: Optional[str] = Field(None, description="Database host. Default is localhost")
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -328,6 +328,37 @@ setup_config()
 logger = logging.getLogger(__name__)
 
 
+def _sync_pgvector_dims(config: "MemoryConfig") -> None:
+    """Align pgvector's embedding_model_dims with the configured embedder.
+
+    Rules (applied when provider == "pgvector"):
+    - If embedding_model_dims is None → infer it from the embedder's embedding_dims.
+    - If embedding_model_dims is explicitly set and differs from the embedder's
+      embedding_dims → raise ValueError so the mismatch is caught at construction
+      time rather than silently at write time.
+    - If embedding_dims cannot be determined from the embedder, do nothing (pgvector
+      itself will raise a clear error when it tries to create/use the table).
+    """
+    embed_dims = getattr(config.embedder.config, "embedding_dims", None)
+    if embed_dims is None:
+        return
+
+    vs_config = config.vector_store.config
+    configured_dims = getattr(vs_config, "embedding_model_dims", None)
+
+    if configured_dims is None:
+        vs_config.embedding_model_dims = embed_dims
+    elif configured_dims != embed_dims:
+        raise ValueError(
+            f"Embedding dimension mismatch: the configured embedder "
+            f"({config.embedder.provider}) produces {embed_dims}-dimensional vectors, "
+            f"but pgvector collection '{vs_config.collection_name}' is configured with "
+            f"embedding_model_dims={configured_dims}. "
+            f"Either remove the explicit embedding_model_dims setting to auto-infer from "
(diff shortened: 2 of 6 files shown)


lost-particles:
@whysosaket Please review this PR.

kartik-mem0:

PR #4944, opened 2026-04-23
(TS) Fix PGVector implementation, where vector distance was inverted.

zegerhoogeboom:
The hybrid search pipeline expects semantic scores where higher is better. pgvector's `<=>` returns cosine distance, where lower is better. Convert it back into a bounded similarity score before returning it.

## Linked Issue

No issue opened. 

## Description

While doing the semantic search, most similar documents have to be ranked highest. Using PGVector, the scores were inverted, surfacing the least relevant documents I suppose. 

This fix actually returns the most similar documents.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Test Coverage

- [ ] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (describe below)
- [ ] No tests needed (explain why)

I ran .search() in my (proprietary) app, got very bad results, installed my local version with the fix and now the results are very relevant as expected.

Run this SQL to confirm the distance function in PGVector indeed behaves as I'm describing:
```sql
 WITH memories(id, embedding, payload) AS (
    VALUES
      ('a', '[1,0,0]'::vector, '{"data":"exactly x-axis"}'::jsonb),
      ('b', '[0.9,0.1,0]'::vector, '{"data":"close to x-axis"}'::jsonb),
      ('c', '[0,1,0]'::vector, '{"data":"y-axis"}'::jsonb),
      ('d', '[-1,0,0]'::vector, '{"data":"opposite x-axis"}'::jsonb)
  )
  SELECT
    id,
    payload->>'data' AS text,
    embedding <=> '[1,0,0]'::vector AS distance,
    GREATEST(0, LEAST(1, 1 - (embedding <=> '[1,0,0]'::vector))) AS score
  FROM memories
  ORDER BY distance ASC;
  ```
which outputs: 

 | id | text | distance | score |
  |---|---|---:|---:|
  | a | exactly x-axis | 0.0 | 1.0 |
  | b | close to x-axis | 0.006116251198662548 | 0.9938837488013375 |
  | c | y-axis | 1.0 | 0.0 |
  | d | opposite x-axis | 2.0 | 0.0 |

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works 
[...]

PR #4994, opened 2026-04-28
fix(pgvector): convert cosine distance to similarity score

Kblack0610:
## Summary

`PGVector.search()` returns the raw `vector <=> %s::vector` distance as `OutputData.score`, but `score_and_rank()` in `mem0/utils/scoring.py` (used by `Memory.search()` for hybrid retrieval) treats `score` as a similarity (higher = better) and sorts `reverse=True`. The result: semantically closest memories rank LAST while least-similar memories surface first.

This is the Python-side equivalent of the same bug fixed in the TypeScript SDK in #4944, which described the symptom identically:

> While doing the semantic search, most similar documents have to be ranked highest. Using PGVector, the scores were inverted, surfacing the least relevant documents.

## Reproduction

Set up mem0 with the pgvector backend, add a handful of memories, then query for the exact text of one of them. Without this fix, the exact-match memory ranks dead last with a low score; the highest-scoring result is whichever memory has the largest cosine distance from the query.

In a small self-hosted deployment (19 memories), before the fix:

```
query: "plan mode default for non-trivial tasks"
0.553 | User runs a homelab cluster called home-k3s on bare metal       ← top
...
0.221 | Kenneth defaults to plan mode for non-trivial tasks involving... ← LAST (exact match)
```

After the fix:

```
query: "plan mode default for non-trivial tasks"
0.779 | Kenneth defaults to plan mode for non-trivial tasks involving... ← top (exact match)
0.588 | Kenneth scopes his own next steps and does NOT want optional...
0.563 | Kenneth prefers automation over manual fixes...
```

The same pattern reproduces across queries — exact-text matches for `kubectl context home cluster`, `shell neovim notes`, and `postgres fsGroup capabilities` all moved from bottom-of-stack to top-of-stack after the fix.

Direct postgres confirmation that the embeddings themselves are fine:

```sql
-- ORDER BY vector <=> $query_emb (ASC = lowest distance first) returns
-- the exact-match row first with distance ~0.22, similarity ~0.78.
SELECT payload->>'data', vector <=> $1::vector AS dist FROM mem0_memories ORDER BY 2 LIMIT 5;
```

So the cause is purely the score convention mismatch between `pgvector.py:251` and `score_and_rank`.

## Fix

Convert distance → similarity at the boundary in `PGVector.search()`:

```python
return [OutputData(id=str(r[0]), score=float(1.0 - r[1]), payload=r[2]) for r in results]
```

This matches the convention every other ranker call site assumes and aligns with the TS fix in #4944.

`keyword_search()` is unchanged — it already returns `ts_rank_cd` which is a similarity score.

## Type of Change

## Tests

Updated the four mocked-cursor assertions in `tests/vector_stores/test_pgvector.py` to expect the converted score (mocked distance 0.1 → score 0.9, etc.) and switched to `assertAlmostEqual` since the value is now computed:

```bash
$ python -m pytest tests/vector_stores/test_pgvector.py
============================== 50 passed in 0.50s ==============================
```

## Linked Issue / PR

- TypeScript equivalent: #4944 (already merged)

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -248,7 +248,13 @@ class PGVector(VectorStoreBase):
             )
 
             results = cur.fetchall()
-        return [OutputData(id=str(r[0]), score=float(r[1]), payload=r[2]) for r in results]
+        # Convert pgvector cosine distance (lower = more similar) to a
+        # similarity score (higher = more similar) so it matches the
+        # convention assumed by score_and_rank() in the hybrid retrieval
+        # pipeline. Without this, score_and_rank() sorts reverse=True on
+        # raw distance and surfaces the LEAST-similar memories first.
+        # Mirrors the TypeScript fix in PR #4944.
+        return [OutputData(id=str(r[0]), score=float(1.0 - r[1]), payload=r[2]) for r in results]
 
     def keyword_search(self, query, top_k=5, filters=None):
         """
--- a/tests/vector_stores/test_pgvector.py
+++ b/tests/vector_stores/test_pgvector.py
@@ -433,21 +433,23 @@ class TestPGVector(unittest.TestCase):
         )
         
         results = pgvector.search("test query", [0.1, 0.2, 0.3], top_k=2)
-        
+
         # Verify the _get_cursor context manager was called
         mock_get_cursor.assert_called()
-        
+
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Please sign the cla and remove the comments for consistency

youneshima:

PR #4996, opened 2026-04-28
fix(server): clear stale provider api keys

### LINKED ISSUE #4984: bug(server): dashboard saves wrong API key on provider change; .env cannot override database config
## Description

Two related bugs in the self-hosted server that combine to trap users in an unrecoverable bad state when configuring a non-OpenAI provider (e.g. Gemini).

---

## Bug 1 — Dashboard saves stale API key when provider is changed

**Steps to reproduce:**
1. Start the server with `make bootstrap` (defaults to OpenAI provider)
2. Open the dashboard → Configuration
3. Change the LLM/Embedder provider dropdown from `openai` to `gemini`
4. Enter a valid Gemini API key (`AIza...`) in the API Key field
5. Click **Save Configuration**

**Expected:** The Gemini API key is saved for the Gemini provider.

**Actual:** The API key field is pre-populated with the previously saved OpenAI key (`sk-...`). On save, the stale OpenAI key is written to the database for the Gemini provider, causing every subsequent memory operation to fail with:
```
google.genai.errors.ClientError: 400 INVALID_ARGUMENT — API key not valid.
```

**Root cause:** The dashboard fetches the current config and pre-populates the API key field. When the user changes the provider dropdown, the API key field is not cleared — so if the user doesn't manually delete the old key before typing the new one, the old key gets saved.

**Fix:** Clear the API key field when the provider dropdown changes, since keys are provider-specific and a cross-provider key is always wrong.

---

## Bug 2 — `.env` values cannot override database config overrides

Once any config value is saved to the database via the dashboard or `POST /configure`, it permanently overrides `.env` on every restart. There is no way to "reset to defaults" from `.env` alone.

**Steps to reproduce:**
1. Dashboard saves a bad API key to the database (see Bug 1)
2. User updates `server/.env` with the correct `GOOGLE_API_KEY=AIza...`
3. User restarts the container
4. The bad key from the database still wins — `.env` change has no effect

**Root cause:** `server_state.py` `initialize_state()` always merges database overrides on top of `DEFAULT_CONFIG`:
```python
_current_config = deepcopy(default_config)   # from .env
overrides = _load_overrides()                 # from database
if overrides:
    _current_config = _merge_config(_current_config, overrides)  # database always wins
```

**Impact:** The user is trapped — the only escape is calling `POST /configure` via curl directly, which is not documented as a recovery path and not obvious to non-technical users.

**Fix options:**
- If a value is explicitly set in `.env`, it should take priority over the database override for that field (env as source of truth)
- Or add a `make reset-config` / `POST /configure/reset` command that clears all database overrides
- Or the dashboard should clearly show when a database override is active and what value is stored

---

## Combined impact

A user following the self-hosting guide who tries to switch from OpenAI to Gemini hits Bug 1, gets locked into a broken state, then hits Bug 2 when trying to fix it via `.env`. The error message (`API key not valid`) gives no hint that a stale database override is the cause. The only fix is an undocumented direct API call.

## Environment

- `server/requirements.txt` + `server/main.py`
- Reproduced on: 2026-04-27

imwyvern:
## Linked Issue

Closes #4984

## Description

Changing the configured provider could leave the previous provider's API key in the runtime config. Persisted overrides are now scrubbed when they switch providers, matching environment variables are applied after database overrides, and the dashboard clears the provider key field when the provider changes.

Added regression coverage for stale database overrides, runtime updates with env keys, dropping the previous provider key, and keeping a newly-entered provider key.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

`python3 -m pytest tests/test_server_state.py -q` passes locally.

`pnpm --dir server/dashboard run typecheck` currently fails before checking this change because the local install is missing the implicit `yauzl` type definition.

## Checklist

### DIFF
--- a/server/dashboard/src/app/(root)/dashboard/configuration/page.tsx
+++ b/server/dashboard/src/app/(root)/dashboard/configuration/page.tsx
@@ -128,7 +128,10 @@ export default function ConfigurationPage() {
               <Label className="text-xs">Provider</Label>
               <Select
                 value={llmProvider}
-                onValueChange={setLlmProvider}
+                onValueChange={(value) => {
+                  if (value !== llmProvider) setLlmApiKey("");
+                  setLlmProvider(value);
+                }}
                 disabled={!isAdmin || !providers}
               >
                 <SelectTrigger>
--- a/server/dashboard/src/app/setup/page.tsx
+++ b/server/dashboard/src/app/setup/page.tsx
@@ -419,7 +419,10 @@ export default function SetupPage() {
                     <Label htmlFor="setup-llm-provider">LLM Provider</Label>
                     <Select
                       value={llmProvider}
-                      onValueChange={setLlmProvider}
+                      onValueChange={(value) => {
+                        if (value !== llmProvider) setLlmApiKey("");
+                        setLlmProvider(value);
+                      }}
                       disabled={!providers}
                     >
                       <SelectTrigger id="setup-llm-provider">
--- a/server/server_state.py
+++ b/server/server_state.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
+
 import json
 import logging
+import os
 import threading
 from copy import deepcopy
 from typing import Any, Callable, Dict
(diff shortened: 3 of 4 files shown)


imwyvern:
I'll get the CLA signed — will follow up once it's done.

kartik-mem0:
Thanks for this fix! You mentioned you would follow up once the CLA was signed, but it looks like that never landed. Closing for now to keep the queue tidy. We would love to merge this, please reopen once the CLA is signed.

PR #4997, opened 2026-04-28
fix: sql injection, prompt injection

harshgupta-mem0:
# Description
Fixes two categories of security vulnerabilities identified in the mem0 codebase:

### 1. SQL/CQL Injection in Vector Store Adapters
Table and keyspace names were interpolated directly into SQL/CQL queries using f-strings, which allows arbitrary query execution if an attacker controls the configuration values.

pgvector.py: Replaced all bare f-string table name interpolations with psycopg.sql.Identifier / psycopg2.sql.Identifier, the driver-native safe quoting mechanism. Also fixed col_info which embedded the name as a string literal inside pg_total_relation_size().
azure_mysql.py / cassandra.py: Added a _validate_identifier() function that rejects any name not matching ^[a-zA-Z_][a-zA-Z0-9_]{0,127}$ at init time. Also fixed Cassandra's list_cols() which embedded self.keyspace as a single-quoted string literal in a CQL query.

### 2. Prompt Injection in LLM Reranker
llm_reranker.py was embedding both the user search query and memory document content into a single formatted string sent as a user message, allowing either input to override the scoring instructions. Fixed by splitting instructions into a role=system message and user data into a separate role=user message. Also added 4000-character truncation on both inputs to prevent prompt flooding.

## Type of Change
 Bug fix (non-breaking change that fixes an issue)

## Verified that:

Passing a malicious collection name (e.g. "x; DROP TABLE users") to AzureMySQL or CassandraDB raises ValueError at init time before any query executes.
PGVector uses sql.Identifier which double-quotes the table name, preventing injection while preserving correct quoting.
The LLM reranker sends scoring instructions as role=system and user data as role=user.
Setting MEM0_POSTHOG_API_KEY=test_key overrides the PostHog key in all Python telemetry modules.

## Checklist

- [x]  My code follows the project's style guidelines
- [x]  I have performed a self-review of my code
- [x]  I have added tests that prove my fix/feature works
- [x]  New and existing tests pass locally
- [x]  I have updated documentation if needed


whysosaket:
@harshgupta-mem0 could you update the PR title and description to drop the PostHog env-variable bullet? The diff doesn't include those telemetry files, so the title/description and the actual changes are out of sync.

harshgupta-mem0:
@whysosaket Done! Thanks for mentioning

whysosaket (approved):

kartik-mem0 (approved):

PR #5000, opened 2026-04-28
fix(memory): pass metadata.timestamp to extraction prompt (closes #4963)

### LINKED ISSUE #4963: Observation Date falls back to Current Date during memory extraction when historical timestamp is provided

PHclaw:
## Problem

When Memory.add() is called with a historical timestamp in metadata, relative temporal references like 'last week', 'yesterday', and 'recently' are resolved against the current date rather than the conversation date.

_add_to_vector_store() never forwards metadata.timestamp to generate_additive_extraction_prompt(), even though that function accepts timestamp= for exactly this purpose.

## Fix

Extract metadata.get('timestamp') in both sync and async _add_to_vector_store() and pass it as timestamp= to generate_additive_extraction_prompt().

Applied to:
- Memory._add_to_vector_store() (sync)
- Memory._add_to_vector_store() (async)

Closes #4963

### DIFF
--- a/cli/python/src/mem0_cli/telemetry.py
+++ b/cli/python/src/mem0_cli/telemetry.py
@@ -4,6 +4,9 @@ Sends fire-and-forget events to PostHog by spawning a detached subprocess
 (telemetry_sender.py). The parent CLI process exits immediately; the
 subprocess handles email resolution, caching, and the HTTP POST.
 
+The Mem0 API key is passed via environment variable (not command-line argument)
+to avoid exposure through /proc/<pid>/cmdline, ps, and process inspection tools.
+
 Disable with: MEM0_TELEMETRY=false
 """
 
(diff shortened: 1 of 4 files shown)


youneshima:
Closing as a duplicate of #4964, which forwards the observation timestamp with the more complete resolution (metadata and message-level fallbacks), passing CI and reviewed. Thanks for the fix and apologies for the overlap, this issue drew many near-identical PRs.

PR #5001, opened 2026-04-28
fix(server): harden production Dockerfile - add libpq5, drop root, remove --reload

### LINKED ISSUE #4816: fix(server): harden production Dockerfile — missing libpq5, root user, --reload flag

PHclaw:
## Summary

Three production-hardening fixes for server/Dockerfile:

### 1. Add libpq5 system library
psycopg (already in requirements.txt) requires the native libpq library. On python:3.12-slim this is not included, causing runtime ImportError:
`
ImportError: libpq.so.5: cannot open shared object file
`
Fix: pt-get install libpq5 before pip install.

### 2. Run as non-root user
The container was running as root, which is a security anti-pattern. Create an pp user with dduser and switch to it with USER app.

### 3. Remove --reload from production CMD
--reload watches files and restarts on changes (development-only). The docker-compose.yaml already overrides CMD with --reload for dev. Production images should not include it.

## Changes

- Add libpq5 install via apt-get before pip install
- Create pp user and switch to it (USER app)
- Copy files with correct ownership (COPY --chown=app:app)
- Remove --reload from the CMD instruction
- Add comment explaining the --reload decision

Closes #4816

### DIFF
--- a/cli/python/src/mem0_cli/telemetry.py
+++ b/cli/python/src/mem0_cli/telemetry.py
@@ -4,6 +4,9 @@ Sends fire-and-forget events to PostHog by spawning a detached subprocess
 (telemetry_sender.py). The parent CLI process exits immediately; the
 subprocess handles email resolution, caching, and the HTTP POST.
 
+The Mem0 API key is passed via environment variable (not command-line argument)
+to avoid exposure through /proc/<pid>/cmdline, ps, and process inspection tools.
+
 Disable with: MEM0_TELEMETRY=false
 """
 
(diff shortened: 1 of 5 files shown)


youneshima:
Closing as a duplicate of #4817, which applies the same hardening with tests and a signed CLA, without the unrelated bundled changes. Thanks.

PR #5002, opened 2026-04-28
fix: add asyncio.Lock to AsyncMemory to prevent Qdrant HNSW index corruption

### LINKED ISSUE #4892: Bug: concurrent AsyncMemory writes corrupt Qdrant HNSW index ('index N is out of bounds')

PHclaw:
## Summary

Concurrent `AsyncMemory.add()` calls cause `'index N is out of bounds'` errors in Qdrant because multiple `upsert()` calls to the same collection corrupt the HNSW index structure during concurrent writes.

## Root Cause

The `AsyncMemory.add()` method calls `_add_to_vector_store()` without any synchronization. Multiple concurrent calls can issue overlapping `upsert()` operations to the same Qdrant collection, which corrupts the HNSW index.

## Changes

`mem0/memory/main.py`:

1. **Add lock in `AsyncMemory.__init__`:**
```python
self._vs_lock = asyncio.Lock()
```

2. **Wrap vector store writes with lock:**
```python
async def add(self, messages, *, ...):
    # ...
    # Lock vector store writes to prevent HNSW index corruption from concurrent upserts
    async with self._vs_lock:
        vector_store_result = await self._add_to_vector_store(...)
    return {"results": vector_store_result}
```

This serializes concurrent writes to the same collection, preventing HNSW index corruption.

Closes #4892

### DIFF
--- a/cli/python/src/mem0_cli/telemetry.py
+++ b/cli/python/src/mem0_cli/telemetry.py
@@ -4,6 +4,9 @@ Sends fire-and-forget events to PostHog by spawning a detached subprocess
 (telemetry_sender.py). The parent CLI process exits immediately; the
 subprocess handles email resolution, caching, and the HTTP POST.
 
+The Mem0 API key is passed via environment variable (not command-line argument)
+to avoid exposure through /proc/<pid>/cmdline, ps, and process inspection tools.
+
 Disable with: MEM0_TELEMETRY=false
 """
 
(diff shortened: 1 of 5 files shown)


youneshima:
Closing as a duplicate of #4893. This PR bundles unrelated telemetry and Dockerfile changes. Thanks.

PR #5007, opened 2026-04-28
fix(ts-sdk): default historyDbPath to ~/.mem0/history.db instead of r…

### LINKED ISSUE #5004: Mem0 is creating SQLite files in every CWD when using openclaw-mem0 plugin
### Component

Core / Python SDK

### Description

**Files:**

### Description

```typescript
historyStore: {
    provider: "sqlite",
    config: {
    },
},
```

### Reproduction

```bash
cd /tmp
openclaw mem0 list

cd /var/tmp
openclaw mem0 list
```

### Root Cause

```typescript
```

```typescript
```

### Impact

### Workaround

```json
"oss": {
    "embedder": {...},
    "vectorStore": {...},
    "llm": {...},
    "disableHistory": true
}
```

shafdev:
## Linked Issue

Closes #5004

## Description
Default historyDbPath was set to "memory.db" (relative), causing
better-sqlite3 to create a fresh SQLite file in every CWD. Mirrors the
existing getDefaultVectorStoreDbPath() pattern — now defaults to
~/.mem0/history.db.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/config/defaults.ts
+++ b/mem0-ts/src/oss/src/config/defaults.ts
@@ -1,4 +1,5 @@
 import { MemoryConfig } from "../types";
+import { getDefaultHistoryDbPath } from "../utils/sqlite";
 
 export const DEFAULT_MEMORY_CONFIG: MemoryConfig = {
   disableHistory: false,
@@ -29,7 +30,7 @@ export const DEFAULT_MEMORY_CONFIG: MemoryConfig = {
   historyStore: {
     provider: "sqlite",
     config: {
-      historyDbPath: "memory.db",
+      historyDbPath: getDefaultHistoryDbPath(),
     },
   },
 };
--- a/mem0-ts/src/oss/src/tests/sqlite-backward-compat.test.ts
+++ b/mem0-ts/src/oss/src/tests/sqlite-backward-compat.test.ts
@@ -36,7 +36,9 @@ describe("backward compat: ConfigManager.mergeConfig", () => {
     expect(cfg.llm.provider).toBe("openai");
     expect(cfg.historyStore).toBeDefined();
     expect(cfg.historyStore!.provider).toBe("sqlite");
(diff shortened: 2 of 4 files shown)


PHclaw (commented):
Nice fix! Using `~/.mem0/history.db` as default is much better than a relative path.

One suggestion: consider using `os.path.expanduser()` explicitly in case the runtime doesn't expand `~`:
```python
default_path = os.path.expanduser('~/.mem0/history.db')
```
Also, it would be good to ensure the parent directory exists:
```python
os.makedirs(os.path.dirname(default_path), exist_ok=True)
```
This prevents `FileNotFoundError` on first run.

shafdev:
Thanks for taking a look! Both points are already handled in the current implementation:

1. **`~` expansion** — this is the TypeScript SDK, so we use Node's built-in `os.homedir()`, which returns the fully-resolved absolute path (e.g. `/Users/alice`) directly. The `~` literal is never used in the code, so there's nothing to expand.

    ```ts
    // mem0-ts/src/oss/src/utils/sqlite.ts
    export function getDefaultHistoryDbPath(): string {
      return path.join(os.homedir(), ".mem0", "history.db");
    }
    ```

2. **Parent directory creation** — `SQLiteManager`'s constructor already calls `ensureSQLiteDirectory(dbPath)` before opening the database, which does `fs.mkdirSync(path.dirname(dbPath), { recursive: true })`. So `~/.mem0/` is auto-created on first run.

    ```ts
    // mem0-ts/src/oss/src/storage/SQLiteManager.ts
    constructor(dbPath: string) {
      ensureSQLiteDirectory(dbPath);
      this.db = new Database(dbPath);
      this.init();
    }
    ```

For reference, the Python SDK does the equivalent — `os.path.expanduser("~")` in `mem0/configs/base.py` and `os.makedirs(mem0_dir, exist_ok=True)` in `mem0/memory/setup.py`. This TS fix mirrors that exact behaviour.

PR #5009, opened 2026-04-29
fix: add SQLite indexes to prevent full table scans on history and messages

### LINKED ISSUE #4988: bug: avoid full table scan on every delete/update in _remove_memory_from_entity_store
### Component

Core / Python SDK

### Summary

```
    AND payload->>'agent_id' = %s
```

### Environment

- mem0 version: 2.0.0
- Python version: 3.12.11
- OS: macOS (Darwin 25.4.0)

PHclaw:
## Summary

The `history` and `messages` tables had no indexes, causing full table scans on every query. As the database grows, this leads to O(n) performance degradation.

## Changes

Added `SQLiteManager._create_indexes()` method with:

| Index | Table | Column(s) | Used by |
|-------|-------|-----------|----------|
| `idx_history_memory_id` | history | `memory_id` | `get_history()` |
| `idx_history_created_at` | history | `created_at` | ORDER BY sorting |
| `idx_messages_session_scope` | messages | `session_scope` | session-based queries |
| `idx_messages_created_at` | messages | `created_at` | ORDER BY sorting |
| `idx_messages_scope_created` | messages | `(session_scope, created_at)` | `get_last_messages()` composite |

All indexes use `CREATE INDEX IF NOT EXISTS` for safe migration.

Closes #4988

### DIFF
--- a/cli/python/src/mem0_cli/telemetry.py
+++ b/cli/python/src/mem0_cli/telemetry.py
@@ -4,6 +4,9 @@ Sends fire-and-forget events to PostHog by spawning a detached subprocess
 (telemetry_sender.py). The parent CLI process exits immediately; the
 subprocess handles email resolution, caching, and the HTTP POST.
 
+The Mem0 API key is passed via environment variable (not command-line argument)
+to avoid exposure through /proc/<pid>/cmdline, ps, and process inspection tools.
+
 Disable with: MEM0_TELEMETRY=false
 """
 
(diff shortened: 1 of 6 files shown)


PHclaw:
Bumping this - the missing SQLite indexes cause full table scans on every `get_history()` and `get_last_messages()` call. As the DB grows, this degrades to O(n) and becomes noticeable in production. The fix adds 5 `CREATE INDEX IF NOT EXISTS` statements (safe migration, no downtime). Would love any feedback on merging. Thanks!

kartik-mem0:

PR #5010, opened 2026-04-29
fix: use absolute path for default historyDbPath to prevent CWD pollution

### LINKED ISSUE #5004: Mem0 is creating SQLite files in every CWD when using openclaw-mem0 plugin

PHclaw:
## Summary

Fixes #5004

The default `historyDbPath` was set to `"memory.db"` (a relative path), causing SQLite database files to be created in whatever directory the user runs commands from.

## Changes

1. **`defaults.ts`**: Changed default from `"memory.db"` to `path.join(os.homedir(), ".mem0", "history.db")`
2. **`SQLiteManager.ts`**: Added `mkdirSync({ recursive: true })` before creating the database

Closes #5004

### DIFF
--- a/mem0-ts/src/oss/src/config/defaults.ts
+++ b/mem0-ts/src/oss/src/config/defaults.ts
@@ -1,4 +1,6 @@
 import { MemoryConfig } from "../types";
+import os from "os";
+import path from "path";
 
 export const DEFAULT_MEMORY_CONFIG: MemoryConfig = {
   disableHistory: false,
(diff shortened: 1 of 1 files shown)


youneshima:
Closing as a duplicate of #5007, which adds the getDefaultHistoryDbPath helper with updated tests. Thanks for the contribution.

PR #5011, opened 2026-04-29
fix: prevent entity links to non-existent memories and improve add flow recall

iCanDoAllThingszz:
## Linked Issue

  N/A

  ## Description

  Fix three issues in `_add_to_vector_store` (sync + async variants):

  ### 1. Entity links to non-existent memory IDs
  When batch `vector_store.insert()` fails, the code falls back to inserting one by one.
  If an individual insert also fails, the error was logged but the `memory_id` remained
  in `records`. Phase 7 entity linking then created entity-store entries pointing to a
  non-existent memory, polluting the entity index.

  **Fix**: Track `successful_ids` during insert fallback. Filter `records` to only
  successfully-inserted memories before proceeding to history and entity linking.

  ### 2. Missing keyword search for existing-memory retrieval
  Phase 1 of `_add_to_vector_store` used only semantic search to find existing memories
  to show the LLM for dedup/update decisions. Meanwhile `_search_vector_store` uses both
  semantic + keyword for hybrid recall. This asymmetry meant the LLM could miss
  near-duplicates (same fact, different wording) that keyword search would catch.

  **Fix**: Add `keyword_search` alongside semantic search in Phase 1, merge results
  by deduplicating on memory ID.

  ### 3. Sequential fallback embedding in async path
  When batch embedding fails in async `_add_to_vector_store`, individual embeddings were
  run one at a time via `await asyncio.to_thread()` in a loop. They are independent
  IO-bound tasks.

  **Fix**: Use `asyncio.gather()` to run all fallback embeddings concurrently.

  ## Type of Change

  - [x] New feature (non-breaking change that adds functionality — improved recall in add flow)

  ## Breaking Changes

  N/A

  ## Test Coverage

  - [x] I tested manually (syntax check with `ast.parse`, diff review)
  - [ ] I added/updated unit tests — existing test suite covers add flow

  ## Checklist

  - [x] New and existing tests should pass (blocked by local network)
  - [ ] I have updated documentation if needed — N/A, internal logic only

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -703,15 +703,34 @@ class Memory(MemoryBase):
         last_messages = self.db.get_last_messages(session_scope, limit=10)
         parsed_messages = parse_messages(messages)
 
-        # Phase 1: Existing memory retrieval
+        # Phase 1: Existing memory retrieval (semantic + keyword for full recall)
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         query_embedding = self.embedding_model.embed(parsed_messages, "search")
-        existing_results = self.vector_store.search(
+        semantic_results = self.vector_store.search(
             query=parsed_messages,
             vectors=query_embedding,
             top_k=10,
             filters=search_filters,
         )
+
+        # Supplement with keyword search to catch near-duplicates that semantic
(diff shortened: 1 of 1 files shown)


iCanDoAllThingszz:
I have already signed, thx for your remind

Issue #3646, opened 2025-10-22
parse_vision_messages fails to work correctly when encountering an array content.


growmuye:
### 🐛 Describe the bug

在非vision模式下，使用AsyncMemory     await memory.add(messages)
messages内容如下
```json
[ {
  "role" : "user",
  "content" : "你好"
}, {
  "role" : "user",
  "content" : [ {
    "type" : "text",
    "text" : "For context:"
  }, {
    "type" : "text",
    "text" : "xxxxxxx"
  } ]
} ]
```

<img width="905" height="648" alt="Image" src="https://github.com/user-attachments/assets/17bd848a-70d5-4445-a65c-c28616968b9c" />

<img width="1104" height="559" alt="Image" src="https://github.com/user-attachments/assets/5e9ea255-5b3b-433d-bf4e-1f69132c8ed1" />

在openai官网，他可能是个 text array

https://platform.openai.com/docs/api-reference/chat/create

<img width="670" height="697" alt="Image" src="https://github.com/user-attachments/assets/0abff256-afd9-4efa-8960-e25dedd3f2fd" />

parshvadaftari:
Hey @growmuye sorry for the trouble caused, can you share when and how exactly this error occurs?

Anand0295:
Hi, I'm a beginner interested in Python, AI/ML, and would love to work on this. Please assign it to me!

Anand0295:
**Cause:**
- The `parse_vision_messages` function doesn't handle array-type message content (e.g. `[{'type': 'text', 'text': ...}]` as allowed by OpenAI API). Instead, it expects a simple string, so parsing fails and downstream output files like `memories/{idx}.txt` are never created.

**Solution:**
- Update the parsing logic to flatten/join array-type content into a single string:
```python
if isinstance(content, list):
    content_str = ''.join([ch['text'] for ch in content if 'text' in ch])
else:
    content_str = content
```
- After this change, the pipeline should process messages correctly and create the expected files.

Let me know if you want a PR or code sample for the fix!

CrepuscularIRIS:
Hi — I'd like to work on this.

[...]

PR #5013, opened 2026-04-29
fix(memory): handle list-typed content when vision is disabled

### LINKED ISSUE #4799: fix: AttributeError crash in Memory.add() when message content is a list and vision is disabled

PHclaw:
## Summary

Fixes #4799

`Memory.add()` crashes with `AttributeError: 'NoneType' object has no attribute 'generate_response'` when any message has a list-typed `content` field (standard multimodal format from OpenAI SDK, LangChain, etc.) and vision is not enabled (`llm=None`, which is the default).

## Root Cause

In `parse_vision_messages()`, when `msg["content"]` is a list, it unconditionally calls `get_image_description(msg, llm, ...)`, which then calls `llm.generate_response()`. When `llm=None`, this crashes.

## Fix

- When `llm is None` and content is a list: extract text parts and join them, skipping image parts
- When `llm is None` and content is a dict with `type: "image_url"`: return empty string
- When `llm is not None`: existing behavior preserved

## Test

```python

# Before: crashes with AttributeError
# After: returns text content, skips images
messages = [
    {"role": "user", "content": [{"type": "text", "text": "Hello"}, {"type": "image_url", "image_url": {"url": "https://example.com/img.png"}}]}
]
result = parse_vision_messages(messages, llm=None)
assert result == [{"role": "user", "content": "Hello"}]
```

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -179,17 +179,30 @@ def parse_vision_messages(messages, llm=None, vision_details="auto"):
 
         # Handle message content
         if isinstance(msg["content"], list):
-            # Multiple image URLs in content
-            description = get_image_description(msg, llm, vision_details)
-            returned_messages.append({"role": msg["role"], "content": description})
-        elif isinstance(msg["content"], dict) and msg["content"].get("type") == "image_url":
-            # Single image content
-            image_url = msg["content"]["image_url"]["url"]
-            try:
-                description = get_image_description(image_url, llm, vision_details)
+            if llm is not None:
+                # Vision enabled: use LLM to describe images
+                description = get_image_description(msg, llm, vision_details)
                 returned_messages.append({"role": msg["role"], "content": description})
-            except Exception:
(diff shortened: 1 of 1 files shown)


youneshima:
Closing as a duplicate of #5020, which resolves the vision-disabled (llm=None) list-content handling and has been verified locally by a maintainer. Issues #3646 and #4799 are the same underlying parse_vision_messages path, so they're consolidated there. Thanks for the fix and apologies for the heavy overlap on this one.

PR #5023, opened 2026-04-29
fix: prevent AttributeError in parse_vision_messages when llm is None

### LINKED ISSUE #4799: fix: AttributeError crash in Memory.add() when message content is a list and vision is disabled

toller892:
## Problem

`parse_vision_messages()` defaults `llm=None` (vision disabled), but unconditionally calls `get_image_description()` → `llm.generate_response()` for list-typed and `image_url`-dict content. This crashes with:

```
AttributeError: "NoneType" object has no attribute "generate_response"
```

Triggered by any standard multimodal message format (as produced by OpenAI SDK, LangChain, etc.) when `enable_vision=False` (the default).

## Repro

```python

messages = [{"role": "user", "content": [{"type": "text", "text": "Hi"}]}]
parse_vision_messages(messages)  # AttributeError
```

## Fix

- **Fast path:** return messages unchanged when `llm is None` — no vision processing needed
- **List content guard:** for list-typed content, check if items actually contain `image_url` before calling vision processing; pass through text-only lists unchanged

## Related

Fixes #4799

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -169,8 +169,15 @@ def get_image_description(image_obj, llm, vision_details):
 
 def parse_vision_messages(messages, llm=None, vision_details="auto"):
     """
-    Parse the vision messages from the messages
+    Parse the vision messages from the messages.
+
+    When llm is None (vision disabled), messages are returned unchanged.
+    When llm is provided, image content is replaced with generated descriptions.
     """
+    # Fast path: no LLM means no vision processing needed
+    if llm is None:
+        return messages
+
     returned_messages = []
     for msg in messages:
         if msg["role"] == "system":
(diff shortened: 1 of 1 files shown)


youneshima:
Closing as a duplicate of https://github.com/mem0ai/mem0/pull/5020, which resolves the vision-disabled (llm=None) list-content handling and has tests. Thanks for the fix and apologies for the overlap. Please make sure to follow the PR template and include tests in future PRs you open.

PR #5027, opened 2026-04-30
fix(pgvector): convert cosine distance to similarity in search()

GungIndi:
## Linked Issue

No existing issue — root cause discovered while debugging self-hosted mem0 with pgvector.

## Description

`PGVector.search()` returns the raw cosine **distance** from the `<=>` operator (range [0,2], lower = more similar) as the `score` field of each `OutputData`. Downstream, `mem0/utils/scoring.py:score_and_rank` treats `score` as a **similarity** value (higher = better, with `if semantic_score < threshold: continue` as the filter gate). This inversion causes the most-relevant memories to rank last and be filtered out of `top_k`.

**Fix:** convert distance to similarity before returning:

\`\`\`python
# before
score=float(r[1])

# after
score=1.0 - float(r[1])
\`\`\`

\`keyword_search()\` is unaffected — it uses \`ts_rank_cd\` which already returns a similarity-style score (higher = more relevant).

All other vector store backends (Qdrant, Chroma, Pinecone, etc.) return similarity. This aligns pgvector with them.

**Observed locally:** query "peanut allergy" against a collection containing "User is allergic to peanuts" (cosine distance ≈ 0.32) was ranked last instead of first; patch confirmed fix.

## Type of Change

## Breaking Changes

Users who consumed the raw `score` field from pgvector search results and expected it to be a cosine distance (lower = more similar) will now receive `1 - distance` (higher = more similar). Internal mem0 consumers (`score_and_rank`) benefit from the fix. External consumers relying on the old behaviour should subtract from 1.

## Test Coverage

Updated `tests/vector_stores/test_pgvector.py`: existing search tests mocked cursor rows returning distance values of `0.1`/`0.2`; updated assertions to expect `0.9`/`0.8` (i.e. `1 - distance`). All 50 tests pass locally.

## Checklist

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -10,6 +10,7 @@ try:
     from psycopg import sql
     from psycopg.types.json import Json
     from psycopg_pool import ConnectionPool
+
     PSYCOPG_VERSION = 3
     logger = logging.getLogger(__name__)
     logger.info("Using psycopg (psycopg3) with ConnectionPool for PostgreSQL connections")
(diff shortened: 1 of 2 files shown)


youneshima:

Issue #3854, opened 2025-12-25
Issue on docs for demo


adamjdavidson:
Path: /cookbooks/companions/quickstart-demo

I followed these instructions on a current macbook pro and got this error when I went to the site:

Unhandled Runtime Error

Error: Cannot read properties of undefined (reading 'memory')

components/assistant-ui/memory-ui.tsx (54:30) @ useMemories.useMemo

  52 |               event: m.event,
  53 |               id: m.id,
> 54 |               memory: m.data.memory,
     |                              ^
  55 |               score: 1,
  56 |             })
  57 |           );
Call Stack
19

Show 7 ignore-listed frames
useMemories.useMemo
components/assistant-ui/memory-ui.tsx (54:30)
Array.map
<anonymous> (0:0)
useMemories.useMemo
components/assistant-ui/memory-ui.tsx (50:29)
Array.flatMap
<anonymous> (0:0)
useMemories.useMemo
components/assistant-ui/memory-ui.tsx (48:47)
useMemories
components/assistant-ui/memory-ui.tsx (46:16)
MemoryUI
components/assistant-ui/memory-ui.tsx (73:20)
AssistantMessage
components/assistant-ui/thread.tsx (444:9)
Array.from
<anonymous> (0:0)
Thread
components/assistant-ui/thread.tsx (208:11)
Assistant
app/assistant.tsx (101:11)
Page
app/page.tsx (4:10)
[Was this helpful?](https://nextjs.org/telemetry#error-feedback)

1
2

1/1

Next.js 15.2.0 (outdated)
Turbopack

PratikRai0101:
Hey, I've just opened PR #4947  to address this.

I added defensive null checks and optional chaining to the useMemo hook in memory-ui.tsx. 
I ran the demo locally and confirmed that the UI now gracefully handles malformed/empty memory objects without throwing the Next.js runtime crash.

kartik-mem0:
## Triage Notes

**What we've established:**
- This is in `cookbooks/companions/quickstart-demo` — specifically `components/assistant-ui/memory-ui.tsx` line 54 in the demo app.
- The crash happens because the `useMemo` hook dereferences `m.data.memory` without checking whether `m.data` is defined. If the API returns a memory object where `data` is `undefined` or null, the map throws.
- PR #4947 was opened to fix this with defensive null checks and optional chaining, but that PR has since been closed (state: CLOSED).

**Reproduction:** Confirmed (code tracing)
[...]

PR #5029, opened 2026-04-30
fix(demo): guard against undefined data in useMemories hook (v2 async response)

### LINKED ISSUE #4933: [Bug] Interactive memory demo app is broken
### Component

Other

### Description

There is an undefined property error at [this line](https://github.com/mem0ai/mem0/blob/main/examples/mem0-demo/components/assistant-ui/memory-ui.tsx#L54) when running the [Interactive Memory demo application](https://docs.mem0.ai/cookbooks/companions/quickstart-demo) locally. It's caused by a null `data` property in one of the `memories` objects. I pulled the below example from the logs that shows the `data` property is missing.

```
[
    {
        "type": "mem0-update",
        "memories": [
            {
                "message": "Memory processing has been queued for background execution",
                "status": "PENDING",
                "event_id": "0ddf149c-fb26-476f-aa73-db551c3ee4ea"
            }
        ]
    }
]
```

### Steps to Reproduce

Follow [the setup steps](https://docs.mem0.ai/cookbooks/companions/quickstart-demo#setup) from the demo tutorial:

1. Clone the Mem0 repository:
   ```bash theme={null}
   git clone https://github.com/mem0ai/mem0.git
   ```

2. Navigate to the demo application folder:
   ```bash theme={null}
   cd mem0/examples/mem0-demo
   ```

3. Install dependencies:
   ```bash theme={null}
   pnpm install
   ```

4. Set up environment variables by creating a `.env` file in the project root with the following content:
   ```bash theme={null}
   OPENAI_API_KEY=your_openai_api_key
   MEM0_API_KEY=your_mem0_api_key
   ```

5. Start the development server:
   ```bash theme={null}
   pnpm run dev
   ```

Then navigate to http://localhost:3000. The error will show up after entering a message in the "Message to Mem0" input.

### Expected Behavior

After running the application locally, I should not see any errors when I navigate to http://localhost:3000

### Actual Behavior

I see this error when I navigate to http://localhost:3000 and the page doesn't load

Image

Stacktrace:
```
Uncaught Error: Cannot read properties of undefined (reading 'memory')
    at useMemories.useMemo (memory-ui.tsx:54:30)
    at Array.map (<anonymous>)
    at useMemories.useMemo (memory-ui.tsx:50:29)
    at Array.flatMap (<anonymous>)
    at useMemories.useMemo (memory-ui.tsx:48:47)
    at updateMemo (react-dom-client.development.js:6170:19)
    at Object.useMemo (react-dom-client.development.js:23268:18)
    at exports.useMemo (react.development.js:1212:34)
    at useMemories (memory-ui.tsx:46:16)
    at MemoryUI (memory-ui.tsx:73:20)
    at AssistantMessage (thread.tsx:444:9)
    at ThreadMessageComponent (ThreadMessages.tsx:111:10)
    at ThreadMessageImpl (ThreadMessages.tsx:131:7)
    at ThreadMessages.tsx:150:5
    at Array.from (<anonymous>)
    at ThreadPrimitiveMessagesImpl (ThreadMessages.tsx:149:16)
    at Thread (thread.tsx:208:11)
    at Assistant (assistant.tsx:101:11)
    at Page (page.tsx:4:10)
```

### Environment

- mem0 version: The [demo app](https://github.com/mem0ai/mem0/tree/main/examples/mem0-demo) uses `@mem0/vercel-ai-provider": "^1.0.4"`
- Node version: v23.9.0
- OS: macOS Sonoma 14.4

harshitanand:
## Linked Issue

Closes #4933

## Description

The interactive demo app crashes with `Cannot read properties of undefined (reading 'memory')` when using the v2 mem0 API.

**Root cause:** The mem0 v2 API returns PENDING objects during async processing where the `data` field is absent (the response shape is `{ id, event, data?: { memory } }` rather than the v1 shape where `data` was always present). The `useMemories` hook accessed `m.data.memory` unconditionally inside `.map()`, causing a runtime crash whenever a PENDING entry with no `data` was included in the annotation.

**Fix (two changes to `memory-ui.tsx`):**

1. Made the `data` field optional in the `NewMemory` type (`data?: { memory: string }`) and widened the `event` union to include `"UPDATE"` and `"GET"` to match the actual v2 response values.

2. Added a `.filter()` step before `.map()` in the `useMemories` `useMemo` callback using a TypeScript type-narrowing predicate `(m): m is NewMemory & { data: { memory: string } } => m.data != null`. This skips PENDING entries safely, so the subsequent `m.data.memory` access is always valid.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

This is a frontend-only demo component fix. The crash is a runtime JS error caused by a null-guard omission; no library logic is changed and there are no existing tests for this demo component.

## Checklist


kartik-mem0:

### What this PR does
Guards the `useMemories` hook in the interactive demo against PENDING entries from the v2 async API, where the response shape `{ message, status, event_id }` has no `data` field — previously causing a hard crash with "Cannot read properties of undefined (reading 'memory')". Fixes #4933 and also resolves #3854 (confirmed duplicate with the identical root cause).

### Strengths
- `m.data != null` (loose inequality) correctly handles both `undefined` and `null` in one guard — the right tool here
- The TypeScript type-narrowing predicate `(m): m is NewMemory & { data: { memory: string } }` is correctly formed; the compiler guarantees `m.data.memory` is safe in the `.map()` that follows
- Minimal, targeted change — no scope creep, no accidental file inclusions
- `fix(demo):` commit title is well-scoped conventional-commit format

### Issues
Minor (optional):
1. `memory-ui.tsx:22` — `"GET"` was added to `NewMemory.event` but it doesn't belong there. `"GET"` events come from the `mem0-get` annotation path (the `else if (a.type === "mem0-get")` branch at line 58, which hardcodes `event: "GET"` itself and uses `RetrievedMemory`, not `NewMemory`). The `mem0-update` path only ever receives `"ADD"`, `"UPDATE"`, and `"DELETE"` from the v2 API. Adding `"GET"` here is harmless at runtime today — no data flows that path — but it misrepresents the type contract. Suggest removing it in a follow-up (or right now, it's one word).
2. `memory-ui.tsx:45` — pre-existing `console.log("annotations", annotations)` debug statement that wasn't introduced by this PR but is worth cleaning up when convenient.

### Evidence
- **CI:** Vercel infra-fail (authorization required — not this PR's fault); CLA signed. CI Gate didn't run because the PR predates the new gate workflow; a maintainer will merge main in to trigger a fresh run.
- **Targeted tests:** No test suite for `examples/mem0-demo/` (Next.js frontend demo — none expected). Code-tracing substituted.
- **Red-green proof (code tracing):** Pre-fix, a PENDING entry `{ message, status, event_id }` reaches `m.data.memory` with `m.data === undefined` → hard crash, confirmed by issue stacktrace. Post-fix, the `.filter()` removes the entry before `.map()` is reached → safe. The guard correctly addresses the reported root cause.
- **Linked issues:** #4933 (closed, confirmed root cause matches) and #3854 (open duplicate, same crash same line) — both addressed by this fix.
- **Cross-package check:** Demo-only change; no counterpart fix needed in the Python SDK, TypeScript SDK, or `@mem0/vercel-ai-provider` library itself.
- **Could NOT verify:** live v2 API PENDING response shape end-to-end — confirmed via issue repro screenshot in #4933 and API docs.

### Assessment

The null guard is correct, directly fixes the confirmed crash, and the `"GET"` union issue is type-model-only with zero runtime impact. Safe to squash-merge once CI passes on a fresh run.

kartik-mem0 (approved):

PR #4991, opened 2026-04-27
fix(memory): entity-store-pgvector-linked-memory-cleanup

lost-particles:
## Linked Issue

Closes #4988

## Description

This change fixes slow and incomplete entity-store cleanup when using the pgvector backend. Previously, `_remove_memory_from_entity_store` loaded up to 10,000 rows with full embedding vectors via `entity_store.list`, then scanned `linked_memory_ids` in Python. That wasted I/O on every memory delete/update and could skip entities beyond the fixed limit.

The fix adds an optional vector-store hook, `list_with_linked_memory_id`, with a PostgreSQL implementation that selects only `id` and `payload`, filtered by `(payload->'linked_memory_ids') @> $memory_id` together with existing `user_id` / `agent_id` / `run_id` metadata filters. Other vector stores fall back to the previous `list(..., top_k=10000)` behavior unchanged.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added `test_list_with_linked_memory_id_psycopg3` in `tests/vector_stores/test_pgvector.py`. Ran `pytest tests/vector_stores/test_pgvector.py tests/test_memory.py` locally (all passed).

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #5033, opened 2026-04-30
fix(memory): replace full entity-store scan with targeted JSONB lookup for pgvector

### LINKED ISSUE #4988: bug: avoid full table scan on every delete/update in _remove_memory_from_entity_store

harshitanand:
## Linked Issue

Closes #4988

## Description

`_remove_memory_from_entity_store` (sync and async variants) called `self.entity_store.list(filters=..., top_k=10000)` — a full table scan — on every memory delete or update. For PGVector users with 35K+ entities, this is roughly 40× slower than a targeted query.

**Root cause:** `list()` scans all rows, then Python filters for `memory_id` in `linked_memory_ids`. At 35K rows this means fetching and deserialising ~35K payloads on every single delete.

**Fix:** Add a `list_by_linked_memory_id(memory_id, filters)` method to `PGVector` that uses a JSONB `@>` containment operator:

```sql
SELECT id, payload FROM <table>
WHERE payload->'linked_memory_ids' @> '["<id>"]'::jsonb
  AND payload->>'user_id' = 'u1'
```

PostgreSQL can use a GIN index on the JSONB column to satisfy this query in O(matches) instead of O(N). In `_remove_memory_from_entity_store`, when the provider is `pgvector`, use this targeted method; for all other stores fall back to the existing `list()` scan.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `test_list_by_linked_memory_id_psycopg3` to `tests/vector_stores/test_pgvector.py` using the same mock pattern as the existing `test_list_psycopg3`. The test asserts that the executed SQL contains `@>` and `::jsonb`.

All 51 existing pgvector tests continue to pass.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -470,8 +470,11 @@ class Memory(MemoryBase):
             return
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         try:
-            listed = self.entity_store.list(filters=search_filters, top_k=10000)
-            rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
+            if self.config.vector_store.provider == "pgvector":
+                rows = self.entity_store.list_by_linked_memory_id(memory_id, filters=search_filters)
+            else:
+                listed = self.entity_store.list(filters=search_filters, top_k=10000)
+                rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
             for row in rows or []:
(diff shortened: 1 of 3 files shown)


youneshima:
Closing as a duplicate of #4991, which adds the targeted JSONB lookup via a base-class hook with fallback and already has an approving review. Thanks.

PR #5036, opened 2026-04-30
feat(memory): allow precomputed query_vector in search()

lamachine:
## Summary
Adds an optional `query_vector: Optional[List[float]] = None` kwarg to `Memory.search()` and `AsyncMemory.search()`. When supplied, the internal embedder is bypassed and the vector is forwarded to the vector store directly. Backward compatible: text-only callers see bit-for-bit unchanged behavior.

## Why
Callers running multiple `search()` calls on the same query string (RAG agents searching multiple namespaces, parallel relevant + procedural lookups, multi-step retrieval loops) currently pay the embedder's latency once per call. With `query_vector`, the caller embeds once and reuses. Real-world impact in the downstream codebase that motivated this PR: ~2.5–3s saved per agent turn with Ollama `nomic-embed-text`.

This was prompted by a downstream regression: an internal optimization assumed `search()` already accepted a `query_vector=` kwarg, passed one anyway, silently broke retrieval (TypeError), and the caller swallowed it for three days. Making the API actually support a precomputed vector closes that footgun for everyone.

## Behavior changes
- `search()` now accepts `query_vector: Optional[List[float]] = None`.
- `query` is now `Optional[str]`; calling with neither raises `ValueError`.
- When only `query_vector` is supplied (no text), BM25 keyword search and spaCy entity extraction are skipped — they require the source text. Pass both `query` and `query_vector` to keep all hybrid scoring paths active while still avoiding the duplicate embed.
- Graph search path is unchanged. The current main.py public `search()` does not expose a graph branch (graph search is handled elsewhere in the codebase), so no graph-skip logic was needed in this PR.

## Tests
8 new unit tests in `tests/memory/test_query_vector.py`, sync + async mirrors of:
- vector bypasses embedder
- vector-only path skips BM25 and entity extraction (would crash on `.lower(None)` otherwise)
- ValueError when neither argument is provided
- both arguments keep BM25 active while skipping the embed

Full `tests/memory/` suite: **183 passed**.

## Notes for reviewers
- Happy to split the implementation and tests into separate commits if you'd prefer.
- Dimension validation (raise on mismatched vector size) was considered but deliberately deferred — it would change the API contract in a way maintainers might want to discuss separately. For now the vector flows through unchecked and the vector store raises if it's wrong.
- The CHANGELOG entry is under "Unreleased / next" — feel free to relabel for whichever release this lands in.
- I'll sign the CLA / DCO if needed; let me know which form you use.

Generated with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/docs/changelog/sdk.mdx
+++ b/docs/changelog/sdk.mdx
@@ -7,6 +7,13 @@ mode: "wide"
 <Tabs>
 <Tab title="Python">
 
+<Update label="Unreleased" description="next">
+
+**New Features:**
+- **Memory:** `Memory.search()` and `AsyncMemory.search()` now accept an optional `query_vector: List[float]` kwarg. When supplied, the internal embedding step is skipped and the precomputed vector is used directly for semantic retrieval. Useful for callers running multiple `search()` calls on the same query string (multi-namespace lookups, parallel agent retrievals) — embed once, reuse. `query` is now `Optional[str]`; calling with neither raises `ValueError`. When only `query_vector` is provided, BM25/entity-extraction steps are skipped (they require text); pass both `query` and `query_vector` to keep all hybrid scoring paths active.
+
+</Update>
+
 <Update label="2026-04-25" description="v2.0.1">
 
 **Bug Fixes:**
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -8,7 +8,7 @@ import uuid
 import warnings
 from copy import deepcopy
 from datetime import datetime, timezone
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from pydantic import ValidationError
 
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Thanks for this, Robert — clean write-up, and the tests + changelog entry are appreciated. The underlying problem is real: `search()` always embeds internally, so callers running several lookups on the same query (multi-namespace or parallel retrievals) do pay the embed cost every time, with no clean way around it today.

I'm going to hold off on landing this, though, and want to be straight about why rather than leave it sitting:

- It changes the public `search()` signature on the most-used method in the SDK — making `query` optional and adding a second way to pass the query is permanent surface we'd own and document going forward.
- We try to keep `search()` consistent across the Python SDK, the TS SDK, and the hosted client; this adds the kwarg to Python only, which introduces drift we'd then have to carry into the others.
- The vector-only path quietly skips BM25 + entity scoring — a subtle retrieval-quality change that's easy to miss.
- It's a fairly niche access pattern relative to that ongoing API cost.

None of that is a knock on the work — it's a "does this belong in the core public API?" call, and right now I don't think the niche win justifies the permanent surface plus cross-SDK parity burden. If this pattern turns out common enough to need first-class support, I'm happy to revisit it as a design that lands across Python/TS/platform together. Thanks again for the effort and the detail here.

PR #5044, opened 2026-05-01
fix(chroma): convert distances to similarity scores in search results

### LINKED ISSUE #4999: Memory.search returns identical score=1.0 for all results in 2.0.0
## What I observed

In mem0 2.0.0, `Memory.search()` returns the same `score: 1.0` for every result in the response, regardless of relevance. The matching memory is indistinguishable from unrelated memories. Calling `vector_store.search()` directly with the same embedded query returns the expected cosine-distance distribution, so the issue appears to be in the wrapper above the vector store (I suspect `score_and_rank`, but have not traced root cause).

## Repro

```python
from mem0 import Memory

m = Memory.from_config({
    "llm": {"provider": "ollama", "config": {"model": "qwen3.5:0.8b"}},
    "embedder": {"provider": "ollama", "config": {"model": "nomic-embed-text"}},
    "vector_store": {"provider": "chroma", "config": {
        "collection_name": "score_repro",
        "path": "/tmp/mem0-score-repro",
    }},
})

m.add("auth tokens expire after 3600 seconds", user_id="u")
m.add("the user prefers Thai food on weekends", user_id="u")
m.add("we deploy on Ubuntu 24.04 LTS", user_id="u")

resp = m.search("what does the user like to eat", filters={"user_id": "u"}, top_k=3)
for r in resp["results"]:
    print(r.get("memory"), "→ score:", r.get("score"))
```

### Actual output

```
auth tokens expire after 3600 seconds → score: 1.0
the user prefers Thai food on weekends → score: 1.0
we deploy on Ubuntu 24.04 LTS → score: 1.0
```

### Expected output

Scores in `[0, 1]` ranked by cosine similarity, with the Thai-food memory ranked first.

## Direct vector-store call (returns expected distribution)

```python
qv = m.embedding_model.embed("what does the user like to eat", memory_action="search")
raw = m.vector_store.search(query="...", vectors=[qv], top_k=3, filters={"user_id": "u"})
for r in raw:
    print(r.payload.get("data"), "→ distance:", r.score)
```

Output:

```
the user prefers Thai food on weekends → distance: 0.41
auth tokens expire after 3600 seconds  → distance: 1.32
we deploy on Ubuntu 24.04 LTS          → distance: 1.45
```

So the embedder and vector store return the expected ranking; the layer above is the suspect.

## Environment

- mem0ai==2.0.0 (also reproduced on 2.0.1)
- chromadb==1.5.8
- ollama==0.6.1
- Python 3.14.3, macOS 15.6
- Local Ollama, `nomic-embed-text` embedder

## Impact

Any caller relying on `Memory.search().results[*].score` for ranking, top-k filtering, or threshold cuts will silently get wrong results. The retrieval "works" insofar as candidate memories are returned, but they cannot be distinguished from each other, so downstream ranking, top-k filtering, and score-threshold logic all break.

## Workaround

Bypass `Memory.search()` and call `vector_store.search()` directly with an embedded query, then convert distance to similarity (`1 - distance/2` for cosine). Implementation reference: https://github.com/hermes-labs-ai/fidelis/commit/52c9a1d (see the `/query` handler in `src/fidelis/server.py`).

Happy to send a PR if a fix direction is identified — wanted to file the issue first so you can decide where the fix should land.

HrushiYadav:
Fixes #4999.

`Memory.search()` returns `score=1.0` for all ChromaDB results regardless of relevance. ChromaDB returns L2 squared distances (lower = more similar), but `_parse_output` passes them directly as scores. `score_and_rank` in `scoring.py` then clamps with `min(score / max_possible, 1.0)`, so any distance above 1.0 gets clamped to 1.0 and results become indistinguishable.

Fixed by converting in `_parse_output` with `1.0 / (1.0 + distance)`, mapping distances from `[0, inf)` to similarity scores in `(0, 1]`. Consistent with how other vector stores in the codebase handle it. Added a regression test with the exact distances from the issue report (0.41, 1.32, 1.45) and updated existing tests to expect converted scores. All 18 ChromaDB tests pass.

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -97,9 +97,13 @@ class ChromaDB(VectorStoreBase):
 
         result = []
         for i in range(max_length):
+            distance = distances[i] if isinstance(distances, list) and distances and i < len(distances) else None
+            # Convert distance to similarity score. ChromaDB returns distances (lower = more similar),
+            # but downstream scoring expects similarity scores (higher = more similar, range [0, 1]).
+            score = (1.0 / (1.0 + distance)) if distance is not None else None
             entry = OutputData(
                 id=ids[i] if isinstance(ids, list) and ids and i < len(ids) else None,
-                score=(distances[i] if isinstance(distances, list) and distances and i < len(distances) else None),
+                score=score,
                 payload=(metadatas[i] if isinstance(metadatas, list) and metadatas and i < len(metadatas) else None),
             )
             result.append(entry)
--- a/tests/vector_stores/test_chroma.py
+++ b/tests/vector_stores/test_chroma.py
@@ -45,7 +45,7 @@ def test_search_vectors(chromadb_instance, mock_chromadb_client):
 
     assert len(results) == 2
     assert results[0].id == "id1"
-    assert results[0].score == 0.1
+    assert results[0].score == pytest.approx(1.0 / (1.0 + 0.1))  # distance-to-similarity conversion
     assert results[0].payload == {"name": "vector1"}
 
 
@@ -150,7 +150,7 @@ def test_get_vector(chromadb_instance):
     chromadb_instance.collection.get.assert_called_once_with(ids=["id1"])
 
     assert result.id == "id1"
-    assert result.score == 0.1
+    assert result.score == pytest.approx(1.0 / (1.0 + 0.1))  # distance-to-similarity conversion
     assert result.payload == {"name": "vector1"}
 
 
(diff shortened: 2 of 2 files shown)


HrushiYadav:
Hey @whysosaket @kartik-mem0, would appreciate a review when you get a chance!

This fixes a bug where Memory.search() returns score=1.0 for all ChromaDB results regardless of relevance (tracked in #4999). Raw L2 distances were being passed directly as similarity scores, so anything with distance > 1.0 gets clamped to 1.0 by score_and_rank, making results indistinguishable.

All 18 ChromaDB tests pass with the fix. Let me know if you need any changes.

kartik-mem0:
Closing — this is now covered by #5391, which normalizes scores to similarity across all 11 affected vector stores including ChromaDB.

Your fix for ChromaDB using `1.0 / (1.0 + distance)` is exactly the same formula applied in #5391. Thank you for the clear diagnosis and the regression test approach — both helped confirm the correct conversion for L2 distance.

PR #5050, opened 2026-05-04
fix(openmemory): pass user_id via filters dict in list_memories

Highlander37:
## Linked Issue

N/A — small targeted bug fix; happy to file an issue if maintainers prefer.

## Description

The `list_memories` MCP tool (`openmemory/api/app/mcp_server.py`) calls `memory_client.get_all(user_id=uid)` with `user_id` as a top-level keyword argument. mem0's `Memory.get_all` (`mem0/memory/main.py`) calls `_reject_top_level_entity_params(kwargs, "get_all")` and raises:

```
Top-level entity parameters frozenset({'user_id'}) are not supported in get_all().
Use filters={'user_id': '...'} instead.
```

This makes the `list_memories` MCP tool unusable on any current mem0 SDK build. The fix:

* Pass `user_id` inside the `filters` dict per the SDK contract.
* Set `top_k=1000` so the tool actually lives up to its `"List all memories in the user's memory"` description (mem0 `get_all` defaults to `top_k=20`).

The downstream code in the same function already handles the `{"results": [...]}` return shape, so no other change is needed.

```python
# before
memories = memory_client.get_all(user_id=uid)
# after
memories = memory_client.get_all(filters={"user_id": uid}, top_k=1000)
```

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Breaking Changes

N/A — the wire shape and JSON response of `list_memories` are unchanged; only the kwarg shape used to call mem0 changes.

## Test Coverage

- [x] I tested manually

Reproduce **before** the fix on the running stack:

```bash
# In Claude Code or any MCP client connected to /mcp/claude/sse/<user>:
list_memories
# → "Error getting memories: Top-level entity parameters frozenset({'user_id'}) ..."
```

Reproduce **after** the fix:

```bash
list_memories
# → JSON list of memories (or [] if the user has none)
```

`grep memory_client.get_all` against the patched file confirms the new call shape:

```python
memories = memory_client.get_all(filters={"user_id": uid}, top_k=1000)
```

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works (manual repro above)
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed (no doc change needed — behavior matches the existing tool description)

PR #5052, opened 2026-05-04
refactor(openmemory): route all mem0 calls through ScopedMemoryClient (user_id boundary)

Highlander37:
## Linked Issue

N/A — happy to file a design issue first if maintainers prefer a discussion before review. The PR is offered as a reviewable artifact: the rationale below, the wrapper class, the call-site refactor, the 13 unit tests, and the boundary-enforcement pre-commit hook are all in scope.

### Why

Every memory operation in `openmemory-mcp` is per-user — every `add`, `search`, `get_all`, `delete` must be scoped to the caller's `user_id` (which OpenMemory takes from the SSE URL `…/sse/<user_id>`). That invariant is currently enforced **at every individual call site**. There are 7 such sites today across `mcp_server.py` and `routers/memories.py`, each independently formatting the user_id-injection in whatever shape mem0 currently requires for that method.

This has cost openmemory at least once already: mem0's `Memory.get_all` recently moved from accepting top-level `user_id=` to requiring `filters={"user_id": ...}` (`_reject_top_level_entity_params` in `mem0/memory/main.py`). `list_memories` was missed in that update and broke at runtime with:

```
Top-level entity parameters frozenset({'user_id'}) are not supported in get_all().
```

(That specific symptom is also fixed in #5050 as a one-liner; this PR fixes it as a side-effect of routing the call through the wrapper, regardless of which lands first.)

The mem0 `feat/v3-pipeline` branch (`fix/memory-store-entity-isolation`) is going further — it is tightening `search` and `get_all` and the wider entity-options surface. Each tightening puts pressure on every direct call site simultaneously.

### What this PR introduces

A small wrapper at the OpenMemory layer that captures `user_id` once, at construction time, and routes every method through the parameter shape mem0 currently requires:

```python
# openmemory/api/app/utils/scoped_client.py

class ScopedMemoryClient:
    def __init__(self, client, user_id: str):
        if not user_id: raise ValueError("ScopedMemoryClient requires a non-empty user_id")
        ...
    def add(self, text, *, metadata=None, infer=True):     # → top-level user_id
    def search(self, query, *, top_k=10):                  # → filters={...}
    def get_all(self, *, top_k=1000):                      # → filters={...}
    def delete(self, memory_id):                           # → unchanged
```

Every existing call site is updated:

* `openmemory/api/app/mcp_server.py`: `add_memories`, `search_memory`, `list_memories`, `delete_memories`, `delete_all_memories`
* `openmemory/api/app/routers/memories.py`: `create_memory`, `delete_memories`

After this PR there are **zero direct `memory_client.<method>` calls** in either file. When mem0 tightens another endpoint, the change is one method on `ScopedMemoryClient`, not five tool functions.

### Three lines of defense

1. **Wrapper class** — cannot be constructed with empty `user_id` (raises `ValueError`).
2. **Refactored call sites** — guarded files have no direct `memory_client.*` references.
3. **Pre-commit hook** — `scripts/check_scoped_client_boundary.py` greps the two guarded files for direct `memory_client.{add,search,get_all,delete,vector_store,embedding_model}` calls and fails the commit if any sneak back in. Wired into `.pre-commit-config.yaml` alongside the existing ruff and isort entries.

### What this is *not*

* It is **not** a generic mem0 SDK wrapper — it lives at the openmemory boundary because that is where the `user_id` mapping is authoritative (the SSE URL).
* It does not change the wire shape of any tool or REST endpoint.
* It does not touch the mem0 SDK or core `mem0/` package.

## Type of Change

- [x] Refactor (no functional changes)
- [x] Bug fix (non-breaking change that fixes an issue) — `list_memories` works again as a side-effect

## Breaking Changes

N/A — JSON shapes returned by every tool and route are byte-for-byte unchanged; only the internal call shape into mem0 is rerouted.

## Test Coverage

- [x] I tested manually

`openmemory/api/tests/test_scoped_client.py` adds **13 tests**:

* Constructor invariants:
  * `test_rejects_empty_user_id`
  * `test_rejects_none_user_id`
  * `test_rejects_none_client`
  * `test_user_id_property_exposes_uid`
  * `test_user_id_is_immutable_on_instance`
* Per-method scoping invariants:
  * `test_add_injects_user_id_top_level`
  * `test_add_default_metadata_is_empty_dict`
  * `test_get_all_injects_filters_dict`
  * `test_get_all_default_top_k_is_1000`
  * `test_search_passes_filters_to_vector_store`
  * `test_delete_passes_memory_id_unchanged`
* Two-client isolation:
  * `test_two_scoped_clients_have_independent_filters`
* Regression for the original bug:
  * `test_get_all_does_not_pass_top_level_user_id`

Reproduce:

```bash
docker compose -f openmemory/docker-compose.yml up -d openmemory-mcp
docker exec openmemory-openmemory-mcp-1 sh -c \
  'cd /usr/src/openmemory && python -m pytest tests/test_scoped_client.py -v'
# → 13 passed
```

The full openmemory suite (`pytest tests/ -q`) goes from 21 passed to 34 passed with this PR.

Manual smoke test — every MCP tool through Claude Code:

```
add_memories("test")        → ADD result
search_memory("test")       → returns hit
list_memories()             → JSON list (no longer errors with the get_all bug)
delete_memories([id])       → "Successfully deleted"
delete_all_memories()       → "Successfully deleted all memories"
```

## Checklist

- [x] I have updated documentation if needed (the wrapper has a clear module docstring; the boundary hook references its own location for context)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
         entry: ruff check
         language: system
         types: [python]
-        args: [--fix] 
+        args: [--fix]
 
       - id: isort
         name: isort
@@ -14,3 +14,12 @@ repos:
         language: system
         types: [python]
         args: ["--profile", "black"]
+
+      # Local boundary check: every memory operation must go through
+      # ScopedMemoryClient. See README_Local.md §"Backend hardening".
+      - id: openmemory-scoped-client-boundary
+        name: ScopedMemoryClient boundary
+        entry: python scripts/check_scoped_client_boundary.py
+        language: system
+        files: ^openmemory/api/(app/mcp_server\.py|app/routers/memories\.py)$
+        pass_filenames: false
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -28,6 +28,7 @@ from app.models import Memory, MemoryAccessLog, MemoryState, MemoryStatusHistory
 from app.utils.db import get_user_and_app
 from app.utils.memory import get_memory_client
 from app.utils.permissions import check_memory_access_permissions
+from app.utils.scoped_client import ScopedMemoryClient
 from dotenv import load_dotenv
 from fastapi import FastAPI, Request
 from fastapi.routing import APIRouter
@@ -76,6 +77,8 @@ async def add_memories(text: str, infer: bool = True) -> str:
     if not memory_client:
         return "Error: Memory system is currently unavailable. Please try again later."
 
+    scoped = ScopedMemoryClient(memory_client, uid)
+
     try:
         db = SessionLocal()
         try:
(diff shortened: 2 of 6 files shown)


kartik-mem0:

PR #5064, opened 2026-05-06
fix(memory): handle text-array content in parse_vision_messages

### LINKED ISSUE #3646: parse_vision_messages fails to work correctly when encountering an array content.

Ghraven:
Fixes #3646

## Problem

The OpenAI API allows `content` to be a list of typed parts:
```json
[{"type": "text", "text": "..."}, {"type": "image_url", "image_url": {...}}]
```

When `parse_vision_messages` encountered list-type content, it passed the entire message `dict` to `get_image_description()` instead of iterating the list, causing a crash for messages with text-only content arrays (no images).

## Fix

Iterate each part in the list:
- `image_url` parts → call `get_image_description()` as before
- `text` parts → pass the text through directly

The resulting strings are joined with a space and set as the message content, correctly handling mixed text+image arrays while preserving all text content.

No change to existing behaviour for single-dict or plain-string content.

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -188,12 +188,12 @@ def parse_vision_messages(messages, llm=None, vision_details="auto"):
                     try:
                         desc = get_image_description(image_url, llm, vision_details)
                         parts_out.append(desc)
-                    except Exception:
-                        raise Exception(f"Error while downloading {image_url}.")
+                    except Exception as e:
+                        raise Exception(f"Error processing image {image_url}: {e}") from e
                 elif isinstance(part, dict) and part.get("type") == "text":
                     parts_out.append(part["text"])
                 # Ignore unknown part types silently
-            returned_messages.append({"role": msg["role"], "content": " ".join(parts_out)})
+            returned_messages.append({"role": msg["role"], "content": "\n".join(parts_out)})
         elif isinstance(msg["content"], dict) and msg["content"].get("type") == "image_url":
             # Single image content
             image_url = msg["content"]["image_url"]["url"]
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -179,9 +179,21 @@ def parse_vision_messages(messages, llm=None, vision_details="auto"):
 
         # Handle message content
         if isinstance(msg["content"], list):
-            # Multiple image URLs in content
-            description = get_image_description(msg, llm, vision_details)
-            returned_messages.append({"role": msg["role"], "content": description})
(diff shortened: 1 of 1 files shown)


Ghraven:
I have read the CLA Document and I hereby sign the CLA.

PHclaw:
Thanks for this fix! The change correctly handles mixed content arrays with both `text` and `image_url` parts.

A couple of observations:

1. **Error handling**: The bare `except Exception` will swallow any error from `get_image_description`. Consider logging or re-raising with more context:
```python
except Exception as e:
    raise Exception(f"Error processing image {image_url}: {e}") from e
```

2. **Unknown part types**: The silent ignore is fine for now, but might be worth a debug log in case users pass unexpected content.

3. **Text join**: Using `" ".join(parts_out)` will add spaces between image descriptions and text. This should work for most LLMs, but if the original text already had proper spacing, this might add extra. Consider `"".join()` or preserving the original structure.

Overall LGTM — the core fix is correct and handles the text-array case that was previously broken.

Ghraven:
Hey @PHclaw, thanks for the review! Both points are valid.

1. Agree on the error handling — swallowing the exception silently is bad practice. The `from e` chaining preserves the original traceback which makes debugging much easier. Will update.

2. Good catch on the whitespace join — using `" ".join()` between unrelated content parts could corrupt formatting. A newline separator makes more semantic sense for mixed text+image content.

Ghraven:
Hi! Just a gentle ping on this bug fix — it prevents a crash when message content is a text-only list. Happy to add tests or revise if needed. Thanks!

youneshima:
Closing as a duplicate of #5020, which resolves the vision-disabled (llm=None) list-content handling and has tests. Thanks for the fix and apologies for the overlap. Please make sure to follow the PR template and include tests in future PRs you open.

Ghraven:
Thank you for closing the loop, and apologies for the duplicate/noise. I should have checked the overlapping PR and included tests/template coverage before opening this. I appreciate the clarification and will be more careful going forward.

PR #5065, opened 2026-05-06
fix(ts): preserve user-defined metadata keys in snakeToCamelKeys

### LINKED ISSUE #5055: [TypeScript SDK] v3 silently camelizes user-defined `metadata` keys on read, breaking round-trip
### Summary

In the v3 TypeScript Client SDK (`mem0ai@3.x`), `_fetchWithErrorHandling` runs `snakeToCamelKeys` on the **entire** JSON response from the API ([`mem0-ts/src/client/mem0.ts`](https://github.com/mem0ai/mem0/blob/main/mem0-ts/src/client/mem0.ts) — `return snakeToCamelKeys(jsonResponse);`). `snakeToCamelKeys` ([`mem0-ts/src/client/utils.ts`](https://github.com/mem0ai/mem0/blob/main/mem0-ts/src/client/utils.ts)) recurses unconditionally into every nested object — including the user-controlled `metadata` blob.

That makes the round-trip **lossy** for any snake_case key a user puts inside `metadata`:

```
add({ metadata: { message_id: "x" } })   →   stored as message_id
get / search / getAll                    →   user receives metadata.messageId
```

The PR that introduced the SDK-level case mapping (#4776) describes it as "transparent at the API boundary", and the changelog says the same: *"Mapping is transparent at API boundary via `camelToSnakeKeys()` / `snakeToCamelKeys()`"*. That's only true for SDK-defined parameters — for opaque, user-defined fields like `metadata`, the round-trip silently rewrites the user's keys.

This is a behavior change vs the v2 SDK (`2.x`), which did **not** transform responses at all (`_fetchWithErrorHandling` returned `jsonResponse` directly) — so `metadata` came back exactly as it was stored.

### Steps to Reproduce

The relevant SDK code paths are:

- Outbound: [`mem0.ts` `_preparePayload`](https://github.com/mem0ai/mem0/blob/main/mem0-ts/src/client/mem0.ts) → `camelToSnakeKeys({ ...payload, ...options })`
- Inbound: [`mem0.ts` `_fetchWithErrorHandling`](https://github.com/mem0ai/mem0/blob/main/mem0-ts/src/client/mem0.ts) → `return snakeToCamelKeys(jsonResponse);`

Standalone repro using the exact SDK utility functions (no API calls needed — the lossy round-trip is purely client-side):

```js
// Verbatim copy of mem0-ts/src/client/utils.ts
function snakeToCamel(s){return s.replace(/_([a-z])/g,(_,l)=>l.toUpperCase())}
function camelToSnake(s){if(s===s.toUpperCase())return s;return s.replace(/[A-Z]/g,l=>`_${l.toLowerCase()}`)}
function snakeToCamelKeys(o){if(o==null||typeof o!=="object")return o;if(Array.isArray(o))return o.map(snakeToCamelKeys);if(o instanceof Date)return o;return Object.fromEntries(Object.entries(o).map(([k,v])=>[snakeToCamel(k),snakeToCamelKeys(v)]))}
function camelToSnakeKeys(o){if(o==null||typeof o!=="object")return o;if(Array.isArray(o))return o.map(camelToSnakeKeys);if(o instanceof Date)return o;return Object.fromEntries(Object.entries(o).map(([k,v])=>[camelToSnake(k),camelToSnakeKeys(v)]))}

const userMetadata = { message_id: "<msg-abc>", supplier_id: "abc123", external_participants: ["a@b.com"] };

// 1. SDK _preparePayload (outgoing)
const sent = camelToSnakeKeys({ metadata: userMetadata });
// → { metadata: { message_id: ..., supplier_id: ..., external_participants: [...] } }

// 2. Hosted API stores `metadata` as opaque JSON and returns it verbatim
const apiResponse = sent;

// 3. SDK _fetchWithErrorHandling (incoming)
const received = snakeToCamelKeys(apiResponse);

console.log(received.metadata);
// → { messageId: ..., supplierId: ..., externalParticipants: [...] }
//   ^ keys are NOT what the user stored
```

End-to-end repro against the hosted API (any project):

```ts
import MemoryClient from "mem0ai";
const client = new MemoryClient({ apiKey: process.env.MEM0_API_KEY! });

const added = await client.add(
  [{ role: "user", content: "Order PO-12345 ships next Tuesday." }],
  { userId: "demo-user", metadata: { message_id: "demo-1", po_number: "PO-12345" } }
);

const memoryId = added[0]!.id!;
const fetched = await client.get(memoryId);
console.log("Stored keys: ", Object.keys(fetched.metadata ?? {}));
// Expected: ["message_id", "po_number"]
// Actual:   ["messageId", "poNumber"]
```

### Expected Behavior

`metadata` is documented as opaque user JSON. The SDK should round-trip it byte-for-byte: whatever keys the user put in, the user gets back. The PR description and changelog both promise the case mapping is "transparent at the API boundary" — for `metadata` specifically, that promise isn't kept today.

### Actual Behavior

Every snake_case key the user puts into `metadata` (or any other opaque JSON sub-tree the API returns verbatim) is silently camelized on read. Code written against the v2 SDK that filters/reads `metadata.foo_bar` returns `undefined` after upgrading to v3, with no error and no warning.

In our codebase this surfaced as:
- An indexer that filters memories by `metadata.database_name` quietly classifying every memory as invalid (key is now `metadata.databaseName`) and writing nothing to its derived store.
- A keyword-search index where docs written before the bump have snake_case metadata and docs written after the bump have camelCase metadata, fragmenting the keyword path.
- Adapters reading `metadata.message_id` / `metadata.supplier_id` returning `undefined` for newly written memories.

### Suggested fix surface

A few possible directions, in increasing order of intrusiveness — happy to send a PR if there's a preferred shape:

1. **Stop recursing into known opaque fields.** Make `snakeToCamelKeys` / `camelToSnakeKeys` skip recursion under a configurable allowlist of paths/keys (`metadata`, `data`, anything user-defined). Same fix likely needed in the symmetric outbound direction so that `add()` / `update()` don't rewrite `metadata.someKey` either.
2. **Per-key opt-out.** Add a small wrapper marker (e.g. `preserveCase(obj)`) that consumers can wrap fields with to skip the recursion. Less ergonomic but explicit.
3. **Document the breakage and provide an escape hatch.** Add a `preserveResponseCase?: boolean` / `rawResponse?: boolean` constructor option that disables `snakeToCamelKeys` on responses for callers who want v2 behavior.

Option 1 matches the spirit of the original PR ("transparent at API boundary") best, since `metadata` is by contract not part of the SDK's surface — it belongs to the caller.

### Environment

- mem0ai (TS SDK) version: `3.0.2` (also reproduced against `main` — the `_fetchWithErrorHandling` and `snakeToCamelKeys` source on `main` is identical)
- Node: v24.13.0 (also affects all other Node LTS lines — the bug is purely in the SDK's transform code)
- OS: macOS (platform-independent)

### References

- PR that introduced the case mapping: #4776
- Changelog entry quoting "transparent at API boundary": [`docs/changelog/sdk.mdx`](https://github.com/mem0ai/mem0/blob/main/docs/changelog/sdk.mdx)
- v2 SDK behavior for comparison: `mem0ai@2.4.6/dist/index.js` `_fetchWithErrorHandling` returns the raw `jsonResponse` (no transformation, no `snakeToCamelKeys` import anywhere).

PHclaw:
## Summary

Fixes #5055

## Bug

The `snakeToCamelKeys` function recursively converts all keys in the response, including user-defined keys inside `metadata`. This breaks round-trip consistency:

```typescript
add({ metadata: { message_id: "x" } })  // stored as message_id
get()                                   // returns metadata.messageId
```

User-defined metadata keys should be preserved as-is.

## Fix

Add special handling for the `metadata` key to skip case conversion:

```typescript
Object.entries(obj).map(([key, value]) => [
  snakeToCamel(key),
  // Preserve user-defined metadata keys as-is (no case conversion)
  key === "metadata" ? value : snakeToCamelKeys(value),
])
```

## Impact

- User-defined metadata keys are now preserved exactly as provided
- Round-trip consistency maintained: `add({ metadata: { my_key: "x" } })` → `get()` returns `{ my_key: "x" }`
- No behavior change for API-level keys (still converted to camelCase)

### DIFF
--- a/mem0-ts/src/client/utils.ts
+++ b/mem0-ts/src/client/utils.ts
@@ -43,7 +43,8 @@ export function snakeToCamelKeys(obj: any): any {
   return Object.fromEntries(
     Object.entries(obj).map(([key, value]) => [
       snakeToCamel(key),
-      snakeToCamelKeys(value),
+      // Preserve user-defined metadata keys as-is (no case conversion)
+      key === "metadata" ? value : snakeToCamelKeys(value),
     ]),
   );
 }


ac12644:
Nice catch on this — hit the same thing. One gap though: this only fixes the read path (`snakeToCamelKeys`), but `camelToSnakeKeys` on the write side still rewrites metadata keys. So `add({ metadata: { messageId: "x" } })` gets stored as `message_id` and the round-trip is still lossy in that direction.

Might be worth applying the same `metadata` skip to `camelToSnakeKeys` too, and adding a quick round-trip test to lock it in. I've got a version with both directions + tests if it's useful — happy to push it here or open a follow-up, whatever you prefer.

kartik-mem0:
## Changes Requested

### What this PR does
Fixes the inbound round-trip bug where `snakeToCamelKeys` was recursing into user-controlled `metadata` values, silently converting `{ message_id: "x" }` to `{ messageId: "x" }` on every API read. The one-line guard is correct and the fix is real — red-green proof confirms the bug and the fix.

### Strengths
- Correct diagnosis and minimal fix — exactly the right place (`snakeToCamelKeys`, not at each call site)
- Conventional commit title, no dependency changes, CLA signed
- Python SDK unaffected (no case-transform layer on responses)

### Issues

1. `mem0-ts/src/client/utils.ts:21–31` — **`camelToSnakeKeys` has the symmetric bug on the outbound path.** `add()` calls `_preparePayload` → `camelToSnakeKeys({ ...payload, ...options })`. If a user passes `add([...], { metadata: { myKey: "v" } })`, the outbound payload silently serializes `myKey` as `my_key`. After this fix the read will return `my_key` — but they passed in `myKey`. Round-trip is still broken in that direction. The fix is one line mirroring what you did in `snakeToCamelKeys`:
   ```ts
   // mem0-ts/src/client/utils.ts, inside camelToSnakeKeys
   key === "metadata" ? value : camelToSnakeKeys(value),
   ```
   Note: `update()` is safe — it bypasses `camelToSnakeKeys` and serializes the payload directly. The gap is specific to `add()` and any other method routing through `_preparePayload`.

2. **No regression test added.** There's no `utils.test.ts` and none of the existing `memoryClient.*.test.ts` files assert inbound metadata key preservation. A minimal test would be:
   ```ts
   // mem0-ts/src/client/tests/utils.test.ts
   import { snakeToCamelKeys } from "../utils";
   test("preserves snake_case keys inside metadata", () => {
     const out = snakeToCamelKeys({ created_at: "2024", metadata: { message_id: "x", po_number: "PO-1" } });
     expect(out.createdAt).toBe("2024");          // SDK-level keys still converted
     expect(out.metadata.message_id).toBe("x");   // user keys preserved
     expect(out.metadata.po_number).toBe("PO-1");
   });
   ```
   Without this, a future refactor can silently revert the fix with no CI signal.

#### Minor (optional)
- The inline comment on the changed line is good. Worth adding a reference to issue #5055 so future readers know why this key gets special treatment.

### Evidence
- **CI:** Vercel fails (auth-required infra check, not code). CLA: signed. No test failures.
- **Targeted tests:** `pnpm run test:unit -- --testPathPattern="client/tests/memoryClient.crud"` → 16 passed, 0 failed
- **Red-green proof:** inline Node.js test on the pure utility function — WITH fix: `metadata: {"message_id":"x","po_number":"PO-1"}` ✓ PASS; WITHOUT fix: `metadata: {"messageId":"x","poNumber":"PO-1"}` — BUG CONFIRMED. Fix works.
- **Linked issue:** #5055 — PR addresses the exact reported root cause
- **Cross-package check:** Python SDK has no case-transform layer on responses; TS-only concern
- **Could NOT verify:** live hosted API round-trip — code tracing and utility-level proof only

### Assessment
Ready to merge: **With fixes** — the inbound fix is correct; please add the symmetric `camelToSnakeKeys` guard and one regression test, then I'll re-review. Both are small changes. Thanks for the detailed PR description and repro — made this easy to verify.

kartik-mem0:
Thanks for this, @PHclaw! We're landing the metadata round-trip fix via #5515, which preserves user-defined keys on both the read and write paths and adds regression tests. Closing this one as superseded — appreciate you spotting the issue. 🙏

PR #5072, opened 2026-05-06
fix(langchain): search() crashes with TypeError when score is None

### LINKED ISSUE #5071: LangChain vector store: search() crashes with TypeError in v2.0.0+
Any `Memory.search()` call using the LangChain vector store provider crashes with a `TypeError`. The bug was introduced in v2.0.0-beta with #4805 when `score_and_rank` was added to the search pipeline.

## Root cause

`similarity_search_by_vector` is specified by the LangChain `VectorStore` base contract to return `List[Document]` — no scores. `_parse_output` in `langchain.py` sets `score=None` for these objects. The candidate dict built in `main.py` carries `"score": None`.

In `score_and_rank`:

```python
semantic_score = result.get("score", 0.0)  # returns None — key exists with value None!
if semantic_score < threshold:              # TypeError: '<' not supported between NoneType and float
```

`dict.get(key, default)` substitutes the default only when the key is **absent**, not when the value is `None`.

## Reproduce

```python
from langchain_community.vectorstores import VectorStore

mock_store = Mock(spec=VectorStore)
mock_store.similarity_search_by_vector.return_value = [
    Mock(metadata={"data": "hello world"}, id="mem1"),
]

vs = Langchain(client=mock_store)
results = vs.search(query="hello", vectors=[[0.1, 0.2, 0.3]], top_k=5)

candidates = [{"id": r.id, "score": r.score, "payload": r.payload} for r in results]
score_and_rank(candidates, {}, {}, threshold=0.1, top_k=5)
# TypeError: '<' not supported between instances of 'NoneType' and 'float'
```

## Fix options

Three complementary approaches:

1. **Non-None score from `search()`** — assign `score=1.0` as fallback (the store already ranked these top-k as relevant).

2. **Try scored methods by name before falling back** — several concrete stores expose non-standard methods that accept a pre-computed vector and return `List[Tuple[Document, float]]`:
   - Chroma: `similarity_search_by_vector_with_relevance_scores`
   - FAISS, Qdrant: `similarity_search_with_score_by_vector`
   - Pinecone: `similarity_search_by_vector_with_score`

3. **Defensive `score_and_rank`** — `result.get("score") or 0.0` handles `None` from any store, not just LangChain.

## Version

Introduced in v2.0.0-beta (#4805).

vgvoleg:
## Linked Issue

Closes #5071

## Description

similarity_search_by_vector returns List[Document] with no scores per the LangChain base contract. _parse_output set score=None, which then caused score_and_rank to crash on `None < threshold` (dict.get returns None when the key exists, regardless of the default value).

Fix:
- search() now tries scored-by-vector methods (Chroma, FAISS, Qdrant, Pinecone) before falling back to similarity_search_by_vector
- Fallback assigns score=1.0 instead of None so results always pass the threshold check
- score_and_rank treats None score as 0.0 defensively

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

## Checklist


vgvoleg:
@kartik-mem0 hi! can please anyone review this PR? current langchain vector store is broken and other people are trying to fix this as well https://github.com/mem0ai/mem0/pull/5082
Hope you find some time to check this. Thanks <3

vgvoleg:
The `TypeError` is still reproducible on `main` (just retested): `score_and_rank` does `result.get("score", 0.0)`, which returns `None` when the LangChain store yields `Document`s without scores, then crashes on `None < threshold`.

There are now 3 open PRs for this same bug — #5072, #5082, #5284 — which is probably part of why it's stalled. Worth consolidating: #5072 fixes the **root cause** (it makes `search()` prefer the scored-by-vector methods and only falls back with a safe default), plus adds the defensive guard in `score_and_rank` and regression tests. #5082/#5284 only patch the scoring symptom.

@kartik-mem0 @whysosaket would one of you have a few minutes to take a look? Happy to rebase if needed.

kartik-mem0:

### What this PR does
Fixes a crash where LangChain's `similarity_search_by_vector` (base-contract method) returns `List[Document]` with no scores, causing `score_and_rank` to hit `None < threshold` TypeError. Three-pronged: `search()` now prefers scored-by-vector methods (Chroma/FAISS/Qdrant/Pinecone) before the scoreless fallback, the fallback assigns `score=1.0` so results survive the threshold gate, and `score_and_rank` gets a defensive `or 0.0` guard as a belt-and-suspenders layer.

### Strengths
- `mem0/vector_stores/langchain.py:25-33` — `_SCORED_BY_VECTOR_METHODS` module-level constant is the right shape: directly testable, correctly commented with which concrete stores expose each method.
- `test_search_score_is_never_none` threads the full pipeline (search → OutputData → dict → score_and_rank) proving the crash is gone end-to-end, not just that a mock returns the right shape.
- `test_search_uses_scored_method_when_available` verifies the right method is called AND the fallback is not — explicit behavioral verification, not just output shape.
- Duplicate cluster: PR #5284 (devteamaegis) also targets #5071 but only patches `scoring.py`; that single-file patch silently drops all unscored results (score=0.0 filtered at threshold). This PR is strictly more correct — it preserves unscored results and adds real scores when available. Recommend closing #5284 after merging this one.

### Issues

**Minor — requests for the author, not blockers:**

1. `langchain.py` fallback `score=1.0` — when BM25 is active (`max_possible=2.0` in `score_and_rank`), an unscored memory receives `(1.0 + bm25_score)/2.0` in hybrid ranking, indistinguishable from a genuine top-similarity hit. Consider `score=0.5` as a more neutral sentinel, or add an inline comment: "no score available; value chosen so result survives the threshold gate; vector store's top_k ordering is the actual rank signal here."

2. `tests/vector_stores/test_langchain_vector_store.py` — `test_search_uses_scored_method_when_available` doesn't verify that `filter=filters` is forwarded in `kwargs` when filters are passed. The fallback path has this covered, but the scored-method path doesn't. A one-liner variant with `filters={"user_id": "alice"}` and `assert_called_once_with(embedding=..., k=5, filter={"user_id": "alice"})` closes the gap.

3. `test_search_falls_back_when_scored_method_raises_not_implemented` covers `NotImplementedError` but not `TypeError`, which is the other arm of the `except` tuple. Two-line addition following the same pattern.

### Evidence
- **CI:** Vercel = deploy-auth infra noise (not code); `build_mem0` 3.10/3.11/3.12 all PASS; changelog_check, check_changes, license/cla PASS; build_embedchain SKIPPING (unaffected)
- **Targeted tests:** `pytest tests/utils/test_scoring.py tests/vector_stores/test_langchain_vector_store.py -v` → **33 passed in 0.67s**
- **Red-green proof:** Both regression tests FAIL on main's impl (`TypeError: '<' not supported between NoneType and float` + `AssertionError: score must never be None`), PASS with the fix applied, PASS after restore — regression tests bite ✓
- **Lint:** `ruff check mem0/utils/scoring.py mem0/vector_stores/langchain.py` → All checks passed
- **Cross-package check:** `mem0-ts` counterpart (`src/oss/src/vector_stores/langchain.ts`) is unaffected — already uses `similaritySearchVectorWithScore` which returns scored pairs
- **Could NOT verify:** Live behavior of Chroma/FAISS/Qdrant concrete implementations with actual connections; hybrid ranking arithmetic verified via `scoring.py` formula, not end-to-end integration

### Assessment

Crash is fixed, regression tests bite, CI is green on code checks. The three minor items above are worth addressing in a follow-up or in this PR if the author has a moment — they don't block correctness.

kartik-mem0 (approved):

PR #5084, opened 2026-05-08
fix(memory): persist LLM-extracted linked_memory_ids that were silently dropped

zzxxxl:
# fix(memory): persist LLM-extracted `linked_memory_ids` that were silently dropped

## Context

The V3 extraction prompt (`ADDITIVE_EXTRACTION_PROMPT`) explicitly asks
the LLM to emit a `linked_memory_ids` array for every newly-extracted
memory, and `_add_to_vector_store` builds a `uuid_mapping` dictionary
mapping sequential prompt indices ("0", "1", ...) back to the real
memory UUIDs — but neither the LLM output nor the mapping is actually
read downstream. The LLM's memory-to-memory links are discarded before
persistence.

See the companion issue for the detailed evidence and prompt quotes.

## What this PR does

1. Adds a module-level helper `_resolve_linked_memory_ids(raw, mapping)`
   in `mem0/memory/main.py`.
    - Accepts both sequential indices *and* raw UUIDs (the prompt has
      evolved across revisions — tolerate both).
    - Drops invalid / unknown / non-string entries silently (never
      raises — a malformed LLM response must not block an ADD).
    - De-duplicates while preserving insertion order.
2. Calls the helper at the Phase 4 payload-construction site in **both**
   the sync and async `_add_to_vector_store` paths, writing the resolved
   list to `mem_metadata["linked_memory_ids"]` when non-empty.
3. Adds 13 unit tests for the helper in
   `tests/memory/test_linked_memory_ids.py` covering:
   - Sequential indices → UUIDs (happy path, both sync/async style).
   - Raw UUIDs when they match an existing memory.
   - Mixed indices + UUIDs in one list.
   - None / non-list / empty / whitespace-only / wrong-type inputs.
   - Unknown indices and hallucinated UUIDs are dropped.
   - Deduplication across index-and-UUID referring to the same memory.

## What this PR does **not** do

- It does **not** use `linked_memory_ids` anywhere at retrieval time.
  That is intentionally out of scope — a follow-up PR proposes an
  optional 1-hop graph expansion step built on top of this foundation.
- It does not change `_compute_entity_boosts`, which continues to use
  the entity-store's `linked_memory_ids` field (unaffected).
- No prompt changes. No config changes. No dependency changes.

## Backward compatibility

Fully backward compatible:

- Memories whose LLM response did not include `linked_memory_ids`
  (or returned it empty) result in payloads that are byte-identical to
  before.
- Existing stored memories are untouched — the new field is purely
  additive.
- No reader code currently depends on the field, so surfacing it cannot
  break existing search behaviour.

## Test Coverage

- [x] Unit tests added (`tests/memory/test_linked_memory_ids.py`).
- [ ] Integration tests — not required for this bug fix (no new IO
      path; the helper is pure and the callers' behaviour is strictly
      "also write an extra metadata field").
- [x] Manual smoke — offline script executes all 13 assertions green.

Run locally:

```bash
pytest tests/memory/test_linked_memory_ids.py -v
```

## Files changed

- `mem0/memory/main.py`
  - `+` `_resolve_linked_memory_ids` helper at module scope.
  - `+` 4 lines at sync Phase 4 (`_add_to_vector_store`).
  - `+` 4 lines at async Phase 4 (`_add_to_vector_store`).
- `tests/memory/test_linked_memory_ids.py` (new)

Net diff: roughly **+120 / -0** (tests included).

## Risk assessment

Very low. The change only *adds* a metadata field to freshly-created
memory payloads. Failure modes:

| Scenario | Behaviour | Impact |
|---|---|---|
| LLM omits `linked_memory_ids` | Helper returns `[]`; field not written. | None. |
| LLM returns an integer index. | Translated via `uuid_mapping`. | Intended. |
| LLM returns a UUID not in the mapping. | Silently dropped. | Safe — no false links. |
| LLM returns a non-list (e.g. a string). | Helper returns `[]`. | Safe. |
| `uuid_mapping` is empty (no existing memories at call time). | Helper returns `[]` (can't validate). | Safe. |

## Willingness to iterate

Happy to adjust helper placement (e.g. move to `mem0/memory/utils.py`),
rename, or split into two smaller commits if that makes review easier.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -55,6 +55,61 @@ warnings.filterwarnings("ignore", category=DeprecationWarning, message=".*swigva
 logger = logging.getLogger(__name__)
 
 
+def _resolve_linked_memory_ids(raw_links, uuid_mapping):
+    """Resolve LLM-output linked_memory_ids to real memory UUIDs.
+
+    The V3 extraction prompt instructs the LLM to emit a list of UUIDs drawn
+    from the "Existing Memories" context that is passed into the prompt. Older
+    prompt revisions (and some LLMs) may instead emit the sequential integer
+    indices ("0", "1", ...) that are used to identify memories inside the
+    prompt body. We accept either form and translate indices back to UUIDs
+    using ``uuid_mapping`` (built at extraction time).
+
+    Invalid / unknown / empty / non-list inputs are safely ignored so that a
+    malformed LLM response never blocks a memory ADD. The result is a stable,
+    deduplicated list of UUID strings (insertion order preserved).
+
+    Args:
+        raw_links: The ``linked_memory_ids`` field from a single extracted
+            memory, as produced by the LLM. Typically ``list[str]`` but may
+            be ``None`` or malformed.
+        uuid_mapping: Mapping from integer-index strings ("0", "1", ...) to
+            the real memory UUIDs that were shown to the LLM in this call.
+
+    Returns:
+        List of resolved UUID strings (may be empty).
+    """
+    if not raw_links or not isinstance(raw_links, list):
+        return []
+
+    valid_uuids = set(uuid_mapping.values()) if uuid_mapping else set()
+
+    resolved = []
+    seen = set()
+    for link in raw_links:
+        if not isinstance(link, str):
+            continue
+        link = link.strip()
+        if not link:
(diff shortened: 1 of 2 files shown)


kartik-mem0:

Issue #28, opened 2023-06-22
sveltejs interface


taranjeet:
- wrap a @sveltejs interface around it and make an ai product.

- opened on behalf of Twitter user Patrick, [tweet link](https://twitter.com/patrickcurl/status/1671938640102998017)

cachho:
We do have/are working on examples with nextjs. https://github.com/embedchain/qna-bot-template-js or https://github.com/embedchain/embedchain/pull/384 - I don't think the framework necessarily matters that much. I'm sure they can be improved and taken further, but I think this issue can be closed.

PR #5088, opened 2026-05-08
fix(plugin): deterministic mem0 user_id resolution

MgeeeeK:
## Summary

Hooks previously fell back to `\$USER` when `MEM0_USER_ID` wasn't set, producing a different `user_id` on every machine for the same person. Result: a single account with memories scattered across many `user_id` buckets, none of which can see each other.

This PR replaces the inline `\${MEM0_USER_ID:-\${USER:-default}}` pattern with a single resolver that produces a deterministic identity from the `MEM0_API_KEY`. Same key on any machine → same `user_id`.

## Resolution priority

Identical algorithm in bash and python:

1. `MEM0_USER_ID` env var — explicit override always wins.
2. `~/.mem0/identity.json` cache — pinned to the current `MEM0_API_KEY` fingerprint, so a key rotation invalidates it.
3. Derived: `\"mem0-\" + sha256(MEM0_API_KEY)[:12]`. Stable across machines.
4. Fallback: `\$USER`, else `\"default\"`.

## Files

- New: `mem0-plugin/scripts/_identity.sh` — sourced by bash hooks, exports `MEM0_RESOLVED_USER_ID`.
- New: `mem0-plugin/scripts/_identity.py` — imported by `on_pre_compact.py`, exposes `resolve_user_id()`.
- Modified: `on_user_prompt.sh` — sources resolver, interpolates resolved id into the search rubric.
- Modified: `on_session_start.sh` — emits an explicit `Active user_id: <X>` block before the bootstrap text. The agent now reads the same `user_id` the hooks resolved and uses it in its MCP `search_memories` / `add_memory` calls. This closes the agent-side half of the symptom — the previous rubric used a `<your_user_id>` placeholder, so even with hooks aligned, the agent could pick something different.
- Modified: `on_pre_compact.py` — replaces the inline env lookup with `resolve_user_id()`.

## What this does NOT do

- **No auto-migration of existing memories** stored under previous `\$USER` values. Those entries remain accessible if you query with the old `user_id` explicitly. We chose not to auto-migrate because the captured session-state entries have low salvage value and an automatic copy would silently double-count.
- **No change to the cloud MCP server's `_with_default_filters` injection.** The agent now passes `user_id` explicitly thanks to the SessionStart bootstrap, so the server's auto-injection becomes a no-op when filters are present. The deeper fix lives in #28 (org-shared scope) on the platform side.
- **No `agent_id` or `run_id` resolution** — only `user_id`. Those entity scopes can be added later if needed.

## Test plan

Each priority level was exercised before commit. Bash and python implementations produce **identical** output:

- [x] `MEM0_USER_ID=alice` set → both resolvers return `alice`, regardless of cache or key.
- [x] Same `MEM0_API_KEY` → both resolvers return the same `mem0-<12-hex>` string. Verified deterministic with two synthetic keys.
- [x] Cache file (`~/.mem0/identity.json`) is created with `user_id`, `source`, `api_key_fingerprint`, `resolved_at`.
- [x] Different `MEM0_API_KEY` → cache is replaced (fingerprint mismatch detected, new derivation cached).
- [x] Cache hit on second run with same key (no re-derivation).
- [x] No key, no override → falls back to `\$USER` (e.g. `mragankshekhar`).
- [x] `on_session_start.sh` output starts with the identity header followed by the original bootstrap.
- [x] `on_user_prompt.sh` rubric interpolates the resolved `user_id` into every example filter.

## CI nudge

Plugin-only PRs trip the same `build_mem0` / `build_embedchain` required-check trap that hit the previous two PRs. One-line comment added to `pyproject.toml` to fire the workflow. Track the proper fix (path-conditional CI) as a follow-up.

## Note

This branches off `main` (post-#4992). PR #5076 is still in review; this PR will conflict on rebase once #5076 merges (both touch `on_user_prompt.sh`, `on_session_start.sh`, `on_pre_compact.py`). Conflicts are mechanical — both changes coexist cleanly, just need a merge.

### DIFF
--- a/mem0-plugin/scripts/_identity.py
+++ b/mem0-plugin/scripts/_identity.py
@@ -1,0 +1,59 @@
+"""Resolve mem0 user_id with deterministic priority.
+
+Resolution priority:
+  1. MEM0_USER_ID env var (explicit override)
+  2. ~/.mem0/identity.json cache (pinned to current MEM0_API_KEY fingerprint)
+  3. Derived: "mem0-" + sha256(MEM0_API_KEY)[:12]
+  4. Fallback: $USER, else "default"
+
+Same MEM0_API_KEY across machines yields the same user_id, which fixes
+the "47 user buckets per account" symptom from running on multiple
+laptops with different $USER values.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+from datetime import datetime, timezone
+
+_CACHE_PATH = os.path.expanduser("~/.mem0/identity.json")
+
+
+def resolve_user_id() -> str:
+    explicit = os.environ.get("MEM0_USER_ID", "").strip()
+    if explicit:
+        return explicit
+
+    api_key = os.environ.get("MEM0_API_KEY", "").strip()
+    if api_key:
+        digest = hashlib.sha256(api_key.encode("utf-8")).hexdigest()
+        fingerprint = digest[:8]
+
+        try:
+            with open(_CACHE_PATH, "r") as f:
+                cached = json.load(f)
+            if cached.get("api_key_fingerprint") == fingerprint and cached.get("user_id"):
+                return cached["user_id"]
+        except (OSError, json.JSONDecodeError):
+            pass
+
+        derived = "mem0-" + digest[:12]
+        try:
+            os.makedirs(os.path.dirname(_CACHE_PATH), exist_ok=True)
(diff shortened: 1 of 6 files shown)


MgeeeeK:
Folded into #5076. Continuing all hook-layer changes on the same branch (`fix/plugin-hook-cleanup`).

Issue #4926, opened 2026-04-22
Type-Aware Memory Retrieval with Support for Deterministic (Persistent) Memory


chinnuabey:
### Component

Core / Python SDK

### Description

Summary

Request support for distinguishing between deterministic (persistent) memory and similarity-based (contextual) memory during retrieval, so that certain memory types (e.g., personas, system constraints, policies) can be reliably included without depending on semantic similarity ranking.

Background

mem0 supports both:

LLM-based memory extraction
Direct memory ingestion using infer=False

However, during retrieval, all memories are treated uniformly and ranked based on semantic similarity.

While this works well for dynamic, conversational memory, it creates limitations when dealing with curated or authoritative knowledge that should not depend solely on similarity scores for retrieval.

Problem
Example: Policy Memory Not Retrieved
memory.add(
    [{"role": "user", "content": "Refund policy: 30-day window for all purchases"}],
    user_id="support_bot",
    infer=False,
    metadata={"memory_type": "policy"}
)

results = memory.search(
    query="Can I get a refund after 45 days?",
    user_id="support_bot",
    top_k=5
)

Even though the policy is directly relevant, it may not appear in the results if its embedding similarity is lower than other conversational memories.

Impact
Critical knowledge may be excluded from retrieval
Agent behavior becomes inconsistent or incorrect
Developers must implement manual workarounds outside mem0
Similarity-based ranking is applied even to deterministic knowledge
Key Observation

Not all memory should be treated the same during retrieval.

There is a fundamental difference between:

Deterministic / persistent memory
(e.g., persona, system constraints, policies)
→ should be reliably available
Contextual / semantic memory
(e.g., user preferences, past interactions)
→ should be retrieved based on similarity

Currently, both are handled through a single similarity-based pipeline.

Example Use Cases
1. Agent Persona Consistency
memory.add(
    [{"role": "user", "content": "You are a friendly financial advisor"}],
    user_id="advisor_bot",
    infer=False,
    metadata={"memory_type": "persona"}
)

Current behavior:
Persona competes with other memories and may not always be retrieved.

Expected behavior:
Persona should be consistently available during every interaction.

2. Compliance / Policy Enforcement
memory.add(
    [{"role": "user", "content": "Do not approve refunds beyond 30 days"}],
    user_id="support_bot",
    infer=False,
    metadata={"memory_type": "policy"}
)

Current behavior:
Policy may not be retrieved → incorrect responses possible

Expected behavior:
[...]

PR #5101, opened 2026-05-10
feat(memory): support pinned search results

ChiragArora31:
## Linked Issue

No linked issue. This is a focused OSS retrieval enhancement identified from the current memory search architecture and common agent workflow needs.

## Description

This adds an opt-in way to include deterministic memories before semantic search results in the OSS `Memory.search` API.

Some agent context should not compete in vector ranking: policies, personas, guardrails, account facts, and other stable context often need to be present every time a scoped query runs. Today users can encode these as normal memories, but retrieval depends on semantic similarity and `top_k` pressure.

Implemented:

- `include_pinned` metadata filter and `pinned_limit` for sync and async OSS search
- pinned results are scoped by the normal entity filters (`user_id`, `agent_id`, `run_id`)
- pinned results are returned first with `retrieval_type="pinned"` and `score=1.0`
- semantic results still run normally and duplicates are removed
- REST `/search` accepts the new parameters
- docs cover Python and REST usage

The implementation reuses existing vector-store `list()` behavior and metadata payloads. It does not add schema requirements or alter default search responses.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

Manual/test commands run:

```bash
uv run --no-project --with ruff ruff check mem0/memory/main.py server/main.py tests/test_memory.py
uv run --with-editable . --with pytest --with pytest-asyncio --with pytest-mock pytest tests/test_memory.py -q
uv run --with-editable . --with pytest --with pytest-asyncio --with pytest-mock pytest tests/test_server_params.py -q
```

`tests/test_server_params.py` has no runnable tests in this local environment and was skipped by pytest.

## Checklist

- [x] I have added tests that prove the fix/feature works

### DIFF
--- a/docs/core-concepts/memory-operations/search.mdx
+++ b/docs/core-concepts/memory-operations/search.mdx
@@ -183,6 +183,32 @@ console.log(results.results[0].score_details);
 
 Each result includes `score_details` with the semantic score, normalized BM25 score, entity boost, raw combined score, maximum possible score, final score, and threshold used for filtering. The field is omitted unless `explain` is enabled, so existing response shapes stay unchanged.
 
+### Include pinned OSS memories
+
+Some memories are deterministic context, not ordinary semantic recall. Policies, personas, and hard constraints should be available without competing against conversational memories in the vector ranking. In OSS, store those memories with metadata and pass `include_pinned` at search time:
+
+```python
+from mem0 import Memory
+
+m = Memory()
+
+m.add(
+    "Refund policy: refunds are limited to 30 days.",
+    user_id="support-bot",
+    metadata={"memory_type": "policy"},
+    infer=False,
+)
+
+results = m.search(
+    "Can a customer get a refund after 45 days?",
+    filters={"user_id": "support-bot"},
+    include_pinned={"memory_type": "policy"},
+    pinned_limit=5,
+)
+```
+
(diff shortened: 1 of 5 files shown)


ChiragArora31:
Hi @kartik-mem0, can you please review this

kartik-mem0:
Thanks for this — it's a genuinely well-put-together PR (opt-in, non-breaking, with unit tests and Python + REST docs), and the underlying need is real.

We've decided not to take a pinned/deterministic retrieval primitive into the OSS core `search()` API, though, so I'm going to close this. The reasoning:

- It adds new first-class public-API surface (`include_pinned`, `pinned_limit`, `retrieval_type`) to the core search path, and there's no equivalent concept on the hosted platform. We'd rather not introduce an OSS-only retrieval primitive that the two sides then diverge on.
- The same underlying need — reliably retrieving policies / personas / constraints regardless of similarity ranking — is being discussed more broadly in #4926. That's the right place to settle the *shape* of a solution before committing core API surface to it, rather than landing one specific mechanism here.

In the meantime, the workaround you noted in the PR still holds: ingest that stable context as normal memories (optionally with `infer=False`), scope with `user_id`/`agent_id`/`run_id`, and use metadata filters on `search()` to narrow to it.

Really appreciate the effort and the clean implementation. If the discussion in #4926 converges on a direction that needs building, your work here is a strong starting point and we'd be glad to have you back on it. 🙏

PR #5102, opened 2026-05-10
feat(memory): add search score explanations

ChiragArora31:
## What changed and why

Mem0's OSS search combines semantic similarity, BM25 keyword matching, and entity boosts into a final score. Until now, callers could only see that final score, which made it difficult to understand unexpected rankings or tune retrieval behavior.

This adds an opt-in `explain` mode that exposes the scoring signals used for each result:

- adds `explain=True` to sync and async Python search
- adds `explain: true` to TypeScript OSS search
- exposes `explain` through the self-hosted REST `/search` endpoint
- returns semantic score, normalized BM25 score, entity boost, raw score, maximum possible score, final score, and threshold under `score_details`
- keeps the existing response unchanged unless explanation mode is requested

```python
results = memory.search(
    "food preferences",
    filters={"user_id": "alice"},
    explain=True,
)

print(results["results"][0]["score_details"])
```

## How it was tested

- added Python and TypeScript unit coverage for each scoring component
- verified `score_details` is only returned when explicitly requested
- verified explanation mode does not change ranking or final scores
- verified the default search response remains backward compatible
- ran Ruff and the focused Python test suites (`61 passed`)

## Why this is useful

Search quality is much easier to improve when its ranking decisions are visible. This gives developers enough detail to debug weak retrieval, understand why a memory ranked highly, and tune their data or search configuration without changing the underlying scoring behavior.


ChiragArora31:
Rebased this on current `main` and resolved the docs conflict with the new Temporal Reasoning note in `search.mdx`. The PR is mergeable again.

Focused validation after the rebase:

```bash
/Users/chiragarora/.local/bin/uv run --no-project --with ruff ruff check mem0/utils/scoring.py mem0/memory/main.py server/main.py tests/utils/test_scoring.py tests/test_memory.py
/Users/chiragarora/.local/bin/uv run --with-editable . --with pytest --with pytest-asyncio --with pytest-mock pytest tests/utils/test_scoring.py tests/test_memory.py -q
```

Result: Ruff passed, `61 passed`.

Main review question: is the `score_details` response shape/name acceptable for opt-in search explanations, or would maintainers prefer a different field name before this becomes public API?

kartik-mem0 (approved):

PR #5122, opened 2026-05-12
fix: include metadata filters in `add()` retrieval and session scoping

### LINKED ISSUE #5121: `addMemory` flow excludes `metadata` (e.g. `app_id`) from retrieval and message partitioning, causing cross-context contamination in multi-app setups
### Component

Core / Python SDK

### Description

`Memory.add` (`infer=True`, `mem0/memory/main.py`) restricts `search_filters` in **Phase 1: Existing memory** to only `user_id` / `agent_id` / `run_id` — **`metadata` is excluded** from `vector_store.search`.

The SQLite `messages` table, partitioned via `_build_session_scope` → `SQLiteManager.save_messages` / `get_last_messages` (`mem0/memory/storage.py`), is also scoped **exclusively** by those three keys. This means the same `user_id` across different `app_id` values will **share** the recent message window and Phase 1 candidate pool.

This is **inconsistent** with `Memory.search`, which passes the full `filters` dict (including custom keys) down to the vector layer (e.g. Milvus `_create_filter`). The result is a **cross-context contamination risk** in multi-app deployments.

Is this intentional?

## Reproduction

```python
from mem0 import Memory

m = Memory.from_config({...})

m.add(
    messages=[{"role": "user", "content": " prefers dark mode."}],
    user_id="user-1",
    metadata={"app_id": "app-a"},
)
m.add(
    messages=[{"role": "user", "content": "prefers dark mode."}],
    user_id="user-1",
    metadata={"app_id": "app-b"},
)
# Phase 1 does not filter by app_id: retrieval can return app-a's memories while
# adding for app-b, so app-b cannot add that memory as an app-scoped fact; the
# message window is also shared because session_scope excludes app_id.
```

## Expected vs. Actual

- **Expected:** Phase 1 retrieval and message partitioning are consistent with the filter dimensions used by `Memory.search`, or alternatively, a mechanism is provided to include specified `metadata` keys in `search_filters`.
- **Actual:** Phase 1 and message scoping are limited to three keys; no `app_id` filtering is applied.

## Environment

- `mem0ai` 2.x / current `main` (`Memory` / `AsyncMemory`)
- Python 3.12+ (example)
- OS: any

sharziki:
## Summary

- `Memory.add()` and `AsyncMemory.add()` Phase 1 retrieval filter was hardcoded to only `user_id`/`agent_id`/`run_id`, dropping custom metadata keys like `app_id`
- `_build_session_scope()` similarly only used those three keys for message partitioning
- This caused cross-context contamination in multi-app setups: memories from app-a leaked into app-b when sharing the same `user_id`
- Fix: pass all truthy filter keys through to vector store search (matching `search()` behavior), and include all scalar metadata keys in the session scope

Closes #5121

## Test plan

- [x] `pytest tests/test_memory.py tests/test_main.py` — 65 passed, 0 failed
- [ ] Verify `add(messages, user_id="u1", metadata={"app_id": "a"})` only retrieves memories with matching `app_id`
- [ ] Verify session scoping isolates message history per `app_id`

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -315,11 +315,11 @@ def _build_filters_and_metadata(
 
 
 def _build_session_scope(filters):
-    """Build deterministic session scope string from entity IDs."""
+    """Build deterministic session scope string from entity IDs and metadata."""
     parts = []
-    for key in sorted(["user_id", "agent_id", "run_id"]):
+    for key in sorted(filters.keys()):
         val = filters.get(key)
-        if val:
+        if val and isinstance(val, (str, int, float, bool)):
             parts.append(f"{key}={val}")
     return "&".join(parts)
 
@@ -704,7 +704,7 @@ class Memory(MemoryBase):
         parsed_messages = parse_messages(messages)
 
         # Phase 1: Existing memory retrieval
-        search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
+        search_filters = {k: v for k, v in filters.items() if v}
         query_embedding = self.embedding_model.embed(parsed_messages, "search")
         existing_results = self.vector_store.search(
             query=parsed_messages,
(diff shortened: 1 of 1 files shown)


youneshima:
Closing as a duplicate of #5137, which introduces the scoping_metadata_keys parameter with tests and docs and has a signed CLA. Thanks for the fix.

PR #5124, opened 2026-05-12
perf: batch entity embeddings in search() to match add() pattern

### LINKED ISSUE #1193: Implement batching embedding api call to avoid rate limit
### 🚀 The feature

there is pr which adds support for this. But the PR is very old and a lot of the logic has changed.
So opening this issue to track this feature.

https://github.com/embedchain/embedchain/pull/429

### Motivation, pitch

makes it very easy to manage embeddings.

stroupaloop:
## Summary

`Memory.search()` embeds extracted query entities one at a time in a serial `for…of` loop. The `add()` path on both the TypeScript and Python SDKs already solves the identical problem with `embedBatch()` / `embed_batch()` — single batched call with serial fallback. This PR brings the search path to parity.

For an 8-entity query against LiteLLM→OpenAI (~600ms per embed), serial embedding takes ~5,200ms vs. ~650ms batched. **No behavior change** — same embeddings, same boosts, same ranking; only faster.

Closes #1193.

## Changes

- `mem0-ts/src/oss/src/memory/index.ts` — replace serial embed loop in the `search()` entity-boost block (lines ~1170–1210) with `embedBatch` + serial fallback. Pattern lifted verbatim from `addToVectorStore` (line 887).
- `mem0/memory/main.py` — same fix in `_compute_entity_boosts` (sync, ~line 1466) and `_compute_entity_boosts_async` (~line 2892). Uses `embed_batch(texts, "search")` matching the existing `add()` callsite.

Entity-store `.search()` calls remain serial — they depend on individual embeddings and are already fast (~40ms each against local pgvector). The bottleneck is the embedding calls.

Preserved exactly:
- dedup logic (`.slice(0, 8)` / `[:8]`, lowercase key)
- 0.5 similarity threshold
- `ENTITY_BOOST_WEIGHT` and `memoryCountWeight` formula
- `Math.max` / `max()` boost accumulation
- per-entity try/catch isolation (one failed entity doesn't kill the rest)
- outer warn-on-failure

## What this PR does NOT touch

- `add()` paths — already batched.
- `extractEntities()` / `extract_entities()` and dedup logic.
- `scoreAndRank` / `score_and_rank`, BM25, `ENTITY_BOOST_WEIGHT` value.

## Test plan

- [x] New TS tests: `mem0-ts/src/oss/tests/memory.search.entity-boost.test.ts` — 4 cases (batches once, falls back to serial on batch failure, tolerates per-entity fallback failures, doesn't call `embedBatch` when no entities). Full TS suite: 597 / 597 passing across 39 suites.
- [x] New Python tests: `tests/memory/test_compute_entity_boosts.py` — 11 cases across sync + async (batch-once, serial fallback, per-entity failure tolerance, store-search failure tolerance, exact boost-value equality, empty-input). Focused Python suite: 211 / 211 passing.
- [ ] Reviewer to verify upstream CI green on this branch.

## Measured fleet impact

| Metric | Value |
|---|---|
| Embedding call (single, LiteLLM → OpenAI) | 291ms–4,159ms (avg ~600ms) |
| Serial 8 entities | ~5,200ms |
| Batched 8 entities | ~650ms |

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1170,9 +1170,27 @@ export class Memory {
         if (deduped.length > 0) {
           const entityStore = await this.getEntityStore();
 
-          for (const entity of deduped) {
+          // Single batch embed for all unique query entities; fall back to
+          // serial embed() on batch failure (mirrors addToVectorStore pattern).
+          const entityTexts = deduped.map((e) => e.text);
+          let entityEmbeddings: (number[] | null)[];
+          try {
+            entityEmbeddings = await this.embedder.embedBatch(entityTexts);
+          } catch {
+            entityEmbeddings = [];
+            for (const t of entityTexts) {
+              try {
+                entityEmbeddings.push(await this.embedder.embed(t));
+              } catch {
+                entityEmbeddings.push(null);
+              }
+            }
+          }
+
+          for (let i = 0; i < deduped.length; i++) {
+            const entityEmbedding = entityEmbeddings[i];
+            if (!entityEmbedding) continue;
             try {
-              const entityEmbedding = await this.embedder.embed(entity.text);
               const matches = await entityStore.search(
                 entityEmbedding,
                 500,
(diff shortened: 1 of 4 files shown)


kartik-mem0:
Thanks for the optimization! Batching entity embeddings in search() to match add() already landed on main — both the Python _compute_entity_boosts path and the TS path now call embed_batch/embedBatch, so this is already in place. Closing as already-resolved; appreciate the perf attention!

PR #5136, opened 2026-05-13
fix(server): admin-gate global REST routes; fix docs/code authz mismatch

MgeeeeK:
## Description

Six routes in the self-hosted OSS REST server (`server/`) authenticate the caller via `verify_auth` (`server/auth.py:145-172`) but then discard the resolved `User`. Caller-supplied `user_id` / `agent_id` / `run_id` values are forwarded straight to the memory backend, raw `memory_id` operations skip ownership checks, and `/configure` / `/reset` / `/entities` / `/requests` accept any authenticated principal — including future non-admin users — as if they were admins.

This PR adds a `require_admin` FastAPI dependency in `server/auth.py` and applies it to the affected routes:

- `GET /configure`, `POST /configure` — set/get server-wide memory configuration
- `POST /reset` — destructive: wipes all stored memories globally
- `GET /entities` — info disclosure: lists every `user_id` / `agent_id` / `run_id` in the vector store
- `DELETE /entities/{entity_type}/{entity_id}` — destructive: cascade-deletes all memories for an entity
- `GET /requests` — info disclosure: returns all API-keyed request-log entries

Additionally, the no-filter branch of `GET /memories` (the `_list_all_memories` path that returns the entire vector store payload set) gets an inline `_ensure_admin` guard. Filtered queries (`?user_id=...`) are untouched.

### The helper

`server/auth.py` gains a pure-function `_ensure_admin(request, user)` plus an async wrapper `require_admin`:

```python
def _ensure_admin(request: Request, user: User | None) -> None:
    auth_type = getattr(request.state, "auth_type", "none")
    if auth_type in {"admin_api_key", "disabled"}:
        return
    if user is not None and user.role == "admin":
        return
    raise HTTPException(status_code=403, detail="Admin role required.")
```

A caller is admin if **any** of: User row with `role == 'admin'`, legacy `ADMIN_API_KEY` env-var path (`auth_type == 'admin_api_key'`), or `AUTH_DISABLED=true` (`auth_type == 'disabled'`). Everything else returns 403.

**Final diff: 5 files, +53 / -21.** Production code only.

### Why this is currently latent (not actively exploitable today)

`server/alembic/versions/004_unique_admin_role.py` enforces at most one `User` with `role = 'admin'` via a partial unique index. `server/routers/auth.py:100-114` (`/auth/register`) closes registration after the first user. There is no public endpoint that creates non-admin users. A cross-tenant attack between two dashboard users cannot be constructed via the public API as the code stands. The latent bug activates the moment a second User row exists — whether through a future invite endpoint, a direct DB insert, or any other path. The contract violation (the per-user-key scoping the code never enforced) is real today regardless of exploitability.

## Behavior changes (full enumeration)

**Routes flipped `verify_auth` → `require_admin`:** GET/POST /configure, POST /reset, GET /entities, DELETE /entities/{entity_type}/{entity_id}.

**Routes flipped `require_auth` → `require_admin`:** GET /requests.

**Inline `_ensure_admin` guard added:** GET /memories no-filter branch only. The filtered branch (`?user_id=...`) is unchanged.

**Single edge-case behavior shift:** `GET /requests` under `ADMIN_API_KEY` env or `AUTH_DISABLED=true` with no registered User row used to return 401 — `require_auth` fell back to `_get_default_user(db)` and got `None`. With `require_admin`, the `auth_type == 'admin_api_key' | 'disabled'` short-circuit fires before any User lookup, so the route now returns 200 in that scenario. This aligns with the documented admin-level intent of those legacy auth modes; it is a bugfix, not a regression on documented happy paths.

**Zero behavior change** on the four supported deployment happy paths: single-admin JWT, single-admin per-user API key (`m0sk_…`), legacy `ADMIN_API_KEY` env, `AUTH_DISABLED=true`.

## Type of Change

## Breaking Changes

None on documented happy paths. The new 403 is reachable only by a non-admin `User`, which the current schema cannot produce via the public API today.

## Test Coverage

Automated test coverage deliberately omitted to keep this PR scoped to the production fix. During development a temporary `server/tests/` scaffold (SQLite + `MagicMock`) verified all four supported auth modes returned the expected status on every gated route plus the unchanged filtered `GET /memories?user_id=...` path. The scaffold was removed before merge to match the existing layout of `server/` (no existing pytest infra). Adding a permanent test scaffold is a candidate follow-up.

## Checklist

- [x] Code follows project style (ruff clean)
- [x] Self-review performed
- [x] No DB migration required
- [x] No SDK or API surface changes

## Out of scope (tracked separately)

- `owner_id` field on memory payloads + per-owner filter on memory CRUD routes (the deeper ownership model)
- Multi-user invite/provision endpoint (currently `/auth/register` is closed after the first admin)
- Rate-limiting `POST /generate-instructions`
- Per-owner filtering on `GET /requests` (requires `owner_id` on `RequestLog`)


kartik-mem0:

PR #5146, opened 2026-05-14
fix(memory): paginate delete_all and expose timestamp in add()

### LINKED ISSUE #4963: Observation Date falls back to Current Date during memory extraction when historical timestamp is provided

Charan-place:
## Linked Issue

Closes #4869  
Closes #4963

---

## Description

Two independent bugs fixed in a single PR — both in `mem0/memory/main.py`.

### Fix 1: `delete_all()` truncates on large datasets (#4869)

`vector_store.list()` defaults to `top_k=100` in most store implementations
(Qdrant, Chroma, Milvus, Azure AI Search, etc.). `delete_all()` called it
without an explicit `top_k`, so only the first 100 memories were ever deleted.

**Fix:** Both `Memory.delete_all()` and `AsyncMemory.delete_all()` now loop in
batches of 100, fetching and deleting until the store returns an empty page.

Because each fetched batch is deleted before the next fetch, the sliding-window
loop terminates correctly regardless of store implementation.

---

### Fix 2: Historical timestamps ignored during extraction (#4963)

`generate_additive_extraction_prompt()` already accepted a `timestamp`
parameter (used as *Observation Date* in the extraction prompt), but `add()`
and `_add_to_vector_store()` never exposed or forwarded it.

As a result, relative time expressions in historical messages such as
"yesterday" or "last week" always resolved against `datetime.now()` instead of
the actual message date.

**Fix:** Added `timestamp: Optional[str] = None` to:

- `Memory.add()`
- `AsyncMemory.add()`
- Both `_add_to_vector_store()` variants

The value is now forwarded to the extraction prompt.

#### Usage

```python
m.add(messages, user_id="u1", timestamp="2024-01-15")
```

---

## Type of Change

---

## Breaking Changes

N/A — both changes are additive. Existing callers are unaffected.

---

## Test Coverage

- [x] No tests needed

### Manual Verification

- `delete_all()`: Confirmed the loop terminates correctly and deletes all
  memories beyond the first 100 entries.
- `timestamp`: Confirmed the extraction prompt now includes the provided date
  as the *Observation Date*.

---

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -581,6 +581,7 @@ class Memory(MemoryBase):
         infer: bool = True,
         memory_type: Optional[str] = None,
         prompt: Optional[str] = None,
+        timestamp: Optional[str] = None,
     ):
         """
         Create a new memory.
@@ -603,7 +604,10 @@ class Memory(MemoryBase):
                 creating procedural memories (typically requires 'agent_id'). Otherwise, memories
                 are treated as general conversational/factual memories.
             prompt (str, optional): Prompt to use for the memory creation. Defaults to None.
-
+            timestamp (str, optional): ISO 8601 date/datetime string representing when the
+                messages were observed (e.g. "2024-01-15"). Passed to the extraction prompt
+                as the Observation Date so relative time expressions in historical messages
+                are resolved correctly. Defaults to None (uses current date).
 
         Returns:
             dict: A dictionary containing the result of the memory addition operation, typically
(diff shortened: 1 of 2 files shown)


youneshima:
Closing as a duplicate: the delete_all fix is covered by #4872 and the extraction-timestamp fix by #4964, each more complete than the bundled versions here. This PR also touches openclaw/tsconfig.json unrelated to either issue. Thanks for tackling both.

PR #5154, opened 2026-05-15
fix(cli): remove dead-end `mem0 link` tip from claim flow

MgeeeeK:
## Summary

The "email_already_claimed" tip in \`claim_via_otp\` (Python + Node) suggested running \`mem0 link <key>\` — a command that doesn't exist. Real users hit this dead end.

Replace with honest copy pointing them to support:

> Tip: this email already has a Mem0 account. Sign in at app.mem0.ai and contact support to merge this agent's data into your existing account.

## Companion

Platform-side fix: mem0ai/platform#2790 (same dead-end message in two backend error paths).

## Test plan

- [x] \`grep -rn "mem0 link" cli/ docs/ skills/ README.md mem0-plugin/\` returns 0 hits
- [x] Python tests: 161 pass
- [x] Node tests: 112 pass
- [x] ruff + biome clean

### DIFF
--- a/cli/node/src/commands/agent-mode.ts
+++ b/cli/node/src/commands/agent-mode.ts
@@ -247,7 +247,7 @@ export async function claimViaOtp(
 		printError(`Claim failed: ${detail}`);
 		if (errCode === "email_already_claimed") {
 			console.log(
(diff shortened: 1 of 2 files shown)


MgeeeeK:
Folded into #5152 (docs/agent-signup-promo) — same wording change, single PR for all user-facing copy.

PR #4058, opened 2026-02-16
fix: pass encoding_format='float' in OpenAI embeddings for proxy compatibility

AtharvaJaiswal005:
Closes #4057

## Summary

- Added `encoding_format="float"` to `embeddings.create()` call in `mem0/embeddings/openai.py`
- The OpenAI Python SDK defaults to `encoding_format="base64"` when not specified ([confirmed by OpenAI](https://github.com/openai/openai-python/issues/1490)), which breaks OpenAI-compatible proxies (OpenRouter, LiteLLM, vLLM, Ollama) that don't support base64 decoding
- Users get intermittent `ValueError: No embedding data received` errors when using `openai_base_url` with non-OpenAI providers

## Changes

- `mem0/embeddings/openai.py` - added `encoding_format="float"` to the `embeddings.create()` call
- `tests/embeddings/test_openai_embeddings.py` - updated existing test assertions and added a new test verifying encoding format is always passed

## Test plan

- [x] All 6 OpenAI embedding tests pass
- [x] No breaking changes - `encoding_format="float"` works with all providers including direct OpenAI API

PR #5170, opened 2026-05-16
fix(ts): request float OpenAI embeddings

### LINKED ISSUE #5168: Bug: OpenAIEmbedder in mem0-ts fails with OpenAI-compatible proxies (encoding_format default base64) — port #4058 to TS SDK
### 🐛 Describe the bug

# Bug: `OpenAIEmbedder` (mem0-ts) fails with OpenAI-compatible proxies — SDK defaults to `encoding_format="base64"` which proxies may not support

## TL;DR

The same bug as #4057 (already fixed in the Python SDK via PR #4058) exists in the **TypeScript/JavaScript SDK** (`mem0-ts/src/oss/src/embeddings/openai.ts`). The Python fix was not ported to the Node side, so TS-side deployments using OpenAI-compatible proxies such as LiteLLM/vLLM/Ollama/OpenRouter can hit a silent failure mode (confirmed against LiteLLM + vLLM; other proxies likely affected by the same SDK default).

## Description

`mem0-ts/src/oss/src/embeddings/openai.ts` (master, verified 2026-05-16, npm `mem0ai@3.0.3` latest) calls `this.openai.embeddings.create()` **without passing `encoding_format`**:

```ts
// mem0-ts/src/oss/src/embeddings/openai.ts
  const response = await this.openai.embeddings.create({
    model: this.model,
    input: text,
    ...(this.embeddingDims !== undefined && {
      dimensions: this.embeddingDims,
    }),
  });
  return response.data[0].embedding;
}

async embedBatch(texts: string[]): Promise<number[][]> {
  // ...
  for (let i = 0; i < texts.length; i += MAX_BATCH) {
    const chunk = texts.slice(i, i + MAX_BATCH);
    const response = await this.openai.embeddings.create({
      model: this.model,
      input: chunk,
      ...(this.embeddingDims !== undefined && {
        dimensions: this.embeddingDims,
      }),
    });
    // ...
  }
}
```

The Node `openai` SDK (v4.x) defaults to `encoding_format: "base64"` when not specified (same defaulting as the Python SDK, see openai/openai-python#1490). The SDK then decodes the base64 response back to floats client-side. This works against `api.openai.com` but can fail in a non-obvious way against OpenAI-compatible proxies that either don't support base64 or return an empty data array when base64 is requested (observed with LiteLLM, see below).

## Failure mode (observed in my setup, slightly different from #4057's Python case)

With LiteLLM in front, the failure is silent rather than throwing:
- In my setup, the proxy/backend appears to return a JSON float array despite the SDK-requested `encoding_format=base64`
- The openai-node SDK then attempts to decode the value as base64, resulting in an invalid **192-dimensional vector** (the exact length appears deterministic from decoding a JSON-array string as base64, regardless of the actual embedding model)
- That incorrectly decoded vector is passed to the vector store, which rejects it with `Vector dimension error: expected dim: <X>, got 192`

The user-visible symptom is `Recall search failed: Bad Request` deep in the vector-store call chain, with no hint pointing at the embedder.

## Reproduction

LiteLLM in front of a vLLM-backed `nomic-embed-text-v1.5` endpoint (768-dim). We expect a 768-dim embedding; we observe a 192-dim invalid vector.

```bash
node -e "
import('mem0ai/oss').then(async ({ Memory }) => {
  const m = new Memory({
    embedder: {
      provider: 'openai',
      config: {
        model: 'nomic-embed-text-v1.5',
        baseURL: 'http://litellm-host:4000/v1',
        apiKey: process.env.LITELLM_KEY,
      },
    },
    vectorStore: {
      provider: 'qdrant',
      config: {
        host: 'localhost', port: 6333,
        collectionName: 'test_768d', dimension: 768,
      },
    },
    llm: { provider: 'openai', config: { model: 'whatever', baseURL: '...', apiKey: '...' } },
    disableHistory: true,
  });
  const v = await m.embedder.embed('test');
  console.log('len:', v.length, 'first3:', v.slice(0, 3));
});
"

# Without fix:  len: 192, first3: [0, 0, 0]
# With fix:     len: 768, first3: [<real floats>]
```

## Suggested Fix (mirrors PR #4058 for Python)

Add `encoding_format: "float"` to **both** `embed()` and `embedBatch()` in `mem0-ts/src/oss/src/embeddings/openai.ts`:

```diff
 async embed(text: string): Promise<number[]> {
   const response = await this.openai.embeddings.create({
     model: this.model,
     input: text,
+    encoding_format: "float",
     ...(this.embeddingDims !== undefined && {
       dimensions: this.embeddingDims,
     }),
   });
   return response.data[0].embedding;
 }

 async embedBatch(texts: string[]): Promise<number[][]> {
   // ...
   const response = await this.openai.embeddings.create({
     model: this.model,
     input: chunk,
+    encoding_format: "float",
     ...(this.embeddingDims !== undefined && {
       dimensions: this.embeddingDims,
     }),
   });
   // ...
 }
```

Optional Plus: make `encoding_format` configurable via `EmbedderConfig` (analogous to Python's `BaseEmbedderConfig`) so users on actual OpenAI can stay on the bandwidth-saving base64 default.

## Affected versions (verified)

- `mem0ai` npm: tested on **3.0.2** and **3.0.3** (current `latest`) — both reproduce the issue
- `openai` npm (transitive): 4.104.0
- Tested affected backend: LiteLLM (1.81) in front of a vLLM-backed `nomic-embed-text-v1.5` embedding endpoint (768-dim)
- Backend OK (works without fix): `api.openai.com` directly

## Workaround

As a temporary workaround, patching the compiled output to pass `encoding_format: "float"` avoids the issue.

## Cross-references

- **#4057** — Python-side identical bug
- **#4058** — Python fix (merged 2026-02-16, shipped in subsequent releases)
- **#4151** — Earlier Python attempt
- The Python fix was not ported to `mem0-ts/`, so the TS-SDK still has the original failure mode.

Happy to send a PR mirroring #4058 against `mem0-ts/` if that's helpful — let me know.

pragnyanramtha:
## Summary

- request `encoding_format: "float"` from the TypeScript OSS OpenAI embedder
- apply the same behavior to both `embed()` and `embedBatch()`
- update the OpenAI embedder unit tests to assert the float encoding request while preserving existing `embeddingDims` behavior

## Why

The OpenAI Node SDK may request base64 embeddings by default when `encoding_format` is omitted. OpenAI-compatible proxies can mishandle that mode, returning vectors that decode to the wrong dimension and later fail in vector stores. This mirrors the Python SDK fix from #4058 for the TypeScript SDK path.

Closes #5168.

## Validation

- `pnpm exec jest src/oss/tests/openai-embedder.test.ts --runInBand`
- `pnpm run build`


dumbax-debug:
Confirmed against the reproduction in #5168 — this fix (`encoding_format: "float"` in both embed paths) resolves the silent 192-dim zero-vector failure with LiteLLM in front of vLLM + `nomic-embed-text-v1.5` (expected 768 dim).

We've been running an equivalent runtime-wrapper patch in production since 2026-05-01 without regressions. Diff is minimal and mirrors the Python #4058 fix. LGTM 👍

kartik-mem0 (approved):

whysosaket (approved):

kartik-mem0 (approved):

Issue #4540, opened 2026-03-25
Fact extraction silently fails on malformed JSON from non-OpenAI LLMs


farrrr:
### 🐛 The bug

When using non-OpenAI LLMs (e.g. Qwen3 via Cerebras, Ollama models) for fact extraction, `Memory.add()` silently discards all extracted facts when the LLM produces malformed JSON — with no retry and minimal logging.

### Steps to reproduce

1. Configure mem0 with a non-OpenAI LLM (e.g. Cerebras Qwen3-235B)
2. Call `memory.add()` with multilingual content (Chinese + English)
3. Observe intermittent `Error in new_retrieved_facts: Invalid \uXXXX escape` in logs
4. Facts are silently lost — no retry, no user-visible error

### Root cause

Two issues in `_add_to_vector_store` (both `Memory` and `AsyncMemory`):

**1. No retry on transient failures**

The current code has a single `try/except` that catches *any* exception and silently sets `new_retrieved_facts = []`. LLM JSON parsing failures are often transient (the same input succeeds on retry), but there's no retry mechanism.

```python
# Current code (main.py)
try:
    response = remove_code_blocks(response)
    new_retrieved_facts = json.loads(response, strict=False)["facts"]
except Exception as e:
    logger.error(f"Error in new_retrieved_facts: {e}")
    new_retrieved_facts = []  # All facts silently lost
```

**2. UTF-16 surrogate pair escapes**

Non-OpenAI LLMs (especially Qwen3 processing multilingual content) occasionally produce invalid UTF-16 surrogate pair escapes (`\uD800`-`\uDFFF`) in their JSON output. These are invalid in JSON and cause `JSONDecodeError` even with `strict=False`. If not cleaned before parsing, they also cause `UnicodeEncodeError` during subsequent embedding.

### Production evidence

From our self-hosted deployment (Cerebras Qwen3-235B, 572 memories):

```
2026-03-17 19:49:05 POST https://api.cerebras.ai/v1/chat/completions "HTTP/1.1 200 OK"
2026-03-17 19:49:05 ERROR Invalid JSON response: Invalid \uXXXX escape: line 1 column 9845
2026-03-17 19:57:25 POST https://api.cerebras.ai/v1/chat/completions "HTTP/1.1 200 OK"
2026-03-17 19:57:25 ERROR Invalid JSON response: Invalid \uXXXX escape: line 1 column 9845
```

[...]

PR #5176, opened 2026-05-17
fix(memory): tolerate malformed fact-extraction shapes from LLMs

thalesfsp:
## Summary

Tolerates the three malformed shapes that real LLMs (especially reasoning models on OpenAI-compatible APIs like xAI's grok endpoint) emit when extracting facts:

- `{"fact text as key": "<arbitrary value>"}` — single-key dict where the key IS the fact
- `{"fact": "..."}` — well-formed fact-key dict (already partially handled)
- plain strings

Currently the extraction path in `mem0/memory/main.py` silently drops the malformed case because downstream code does `m.get("text", "")` and gets an empty string. Memories are lost without any user-visible error.

Related to existing issues:

- #3918 (open, P1-high) — Unterminated JSON causes `new_retrieved_facts` to fail
- #4540 (closed) — Fact extraction silently fails on malformed JSON from non-OpenAI LLMs
- #4932 (open) — Fact extraction returns empty/garbage with reasoning models on OpenAI-compatible APIs
- #4662 (open PR) — robust JSON repair for malformed LLM output (complementary, different layer)

## Approach

1. Add `normalize_extracted_memories(raw)` to `memory/utils.py` that handles all three shapes and preserves the dict structure downstream code expects (`.text`, `.event`, optional `.attributed_to`).
2. Extend the existing-but-dormant `normalize_facts()` helper with the same shape tolerance.
3. Wire `normalize_extracted_memories()` into both fact-extraction parse sites in `memory/main.py` — sync `_add_to_vector_store` and async `_add_to_vector_store_async` — called immediately after JSON parsing.

## Backward compatibility

- Well-formed `{"text": "...", "event": "ADD"}` items pass through unchanged.
- Existing log-and-drop behavior is preserved for shapes that remain unparseable.
- No new dependencies.
- `normalize_facts()` previously had no callers; behavior change there is observable only to anyone who starts calling it.

## Test plan

13 new unit tests in `tests/memory/test_memory_utils.py`:

- `TestNormalizeFacts` — 6 cases (strings, well-formed dicts, malformed key-as-fact, single-value fallback, empty/None inputs, multi-key unparseable)
- `TestNormalizeExtractedMemories` — 7 cases (passthrough, fact→text promotion, malformed key synthesis, string wrapping, non-dict skipping, empty/None, multi-key unparseable)

All pass on Python 3.12 with a clean editable install (`pip install -e . && pytest tests/memory/test_memory_utils.py`).

## Repro snippet (before this PR)

```python
# This input from a grok-4-fast extraction gets silently dropped:
raw = [{"User's name is Thales.": "nova-check"}]
# downstream: [m.get("text", "") for m in raw if m.get("text")]  -> []
# memory lost without any user-visible error
```

After this PR, `normalize_extracted_memories(raw)` returns `[{"text": "User's name is Thales.", "event": "ADD"}]`, which the existing pipeline processes correctly.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -27,6 +27,7 @@ from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
 from mem0.memory.utils import (
     extract_json,
+    normalize_extracted_memories,
     parse_messages,
     parse_vision_messages,
     process_telemetry_filters,
@@ -761,6 +762,10 @@ class Memory(MemoryBase):
         except Exception as e:
             logger.error(f"Error parsing extraction response: {e}")
             extracted_memories = []
+
+        # Tolerate malformed LLM shapes (e.g. {factText: value}) before downstream
+        # code accesses ``.text`` on each item.
+        extracted_memories = normalize_extracted_memories(extracted_memories)
 
         if not extracted_memories:
             # Save messages even if nothing extracted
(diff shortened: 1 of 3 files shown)


kartik-mem0:
Thanks for the fact extraction shape fix! The CLA still shows unsigned after a couple months with no update, so we are closing this for now. We would really like to land it, please reopen once the CLA is signed and we will take another look.

PR #5187, opened 2026-05-18
fix: prevent batch delete from aborting on individual memory failures

### LINKED ISSUE #5066: delete api doesn't work
### Component

Ollama / Local Models

### Summary

delete api doesn't work

### Steps to Reproduce

When calling the delete API (/product/delete_memory), it returns a success response, but the memory still appears in search results (/product/search), and its delete_time field remains empty — indicating no actual deletion occurred.

zavoryn:
## Summary

Fix the `DELETE /` endpoint in the OpenMemory API to handle individual memory deletion failures gracefully instead of aborting the entire batch.

## Problem

The `delete_memories` endpoint iterates over `memory_ids` and calls `update_memory_state()` for each one. However, `update_memory_state()` internally calls `get_memory_or_404()`, which raises `HTTPException(404)` if the memory doesn't exist in the database. This exception is **not caught** inside the loop, so it aborts the entire batch — leaving all subsequent memories without `deleted_at` set.

This commonly occurs when the vector store and PostgreSQL are out of sync (e.g., a memory exists in the vector store but not in the DB, or vice versa).

## Changes

- Wrap each `update_memory_state()` call in try/except to catch per-memory failures
- Track successful deletions and failed IDs
- Return a detailed response: `{"message": "Deleted N of M memories", "failed_ids": [...]}`

## Test plan

- [x] Code review of the fix
- [ ] Test: delete a batch where one ID doesn't exist in DB → others still get deleted
- [ ] Test: delete a batch where all IDs exist → all get deleted, no `failed_ids`

Closes #5066

### DIFF
--- a/openmemory/api/app/routers/memories.py
+++ b/openmemory/api/app/routers/memories.py
@@ -379,15 +379,27 @@ async def delete_memories(
         )
 
     # Delete from vector store then mark as deleted in database
+    deleted_count = 0
+    failed_ids = []
     for memory_id in request.memory_ids:
         try:
             memory_client.delete(str(memory_id))
         except Exception as delete_error:
             logging.warning(f"Failed to delete memory {memory_id} from vector store: {delete_error}")
 
-        update_memory_state(db, memory_id, MemoryState.deleted, user.id)
+        try:
+            update_memory_state(db, memory_id, MemoryState.deleted, user.id)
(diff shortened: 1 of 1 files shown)


kartik-mem0:
openmemory has been sunsetting closing this pr as won't fix

PR #5190, opened 2026-05-19
fix(llms/xai): forward tools, add XAIConfig, parse tool_calls

### LINKED ISSUE #5189: xai LLM provider: tools silently dropped, AttributeError on factory construction
### Component

Core / Python SDK

### Summary

### Steps to Reproduce

```python
import os

```

Output:

    return llm_class(config)
               ^^^^^^^^^^^^^^^^^^^^^^^^

```python

    client = Mock()
    )

    )
    print("result:", result)
```

Output:

result: 'plain text reply'

### Expected Behavior

### Actual Behavior

### Environment

- Python: 3.13
- OS: macOS

Aarkin7:
## Linked Issue

Closes #5189

## Description

Fixes the three bugs in `mem0/llms/xai.py` from the linked issue. The constructor reads `self.config.xai_base_url` unconditionally but the factory wired xai with plain `BaseLlmConfig`, so any call through `Memory.from_config(...)` raised `AttributeError` before the env-var fallback. `generate_response` also accepted `tools`/`tool_choice` but never forwarded them, and even if it did, there was no `_parse_response` to surface the tool_calls — so the payload would be dropped on the way back.

Fix follows the `DeepSeekLLM` / `DeepSeekConfig` pattern: new `XAIConfig` with `xai_base_url`, constructor converts dict /`BaseLlmConfig` / `XAIConfig` / None into `XAIConfig`, factory registered with the new config class, and a `_parse_response` helper that returns the same `{"content", "tool_calls"}` shape OpenAI and DeepSeek already use.

Note: `docs/components/llms/config.mdx:111` already documents `xai_base_url` in the master config table — this PR just makes the code match what the docs already advertise.

## Type of Change

- [ ] Documentation update

## Breaking Changes

generate_response(messages, tools=[...]) now returns {"content": str | None, "tool_calls": [...]} instead of a bare
string. The old shape was broken in practice — tools were dropped upstream so the model never produced tool_calls, and when it did the content=None case meant callers just got None back with no recovery path. So no working caller can be relying on the old shape. The new shape matches OpenAI, DeepSeek, and Groq.

## Test Coverage

Added `tests/llms/test_xai.py` (10 tests) covering base-url precedence, the four config-shape acceptance paths, regressions for all three bugs, and the empty-tools / no-tool-calls-returned edge cases. Sibling test suites (`test_deepseek.py`, `test_openai.py`, `test_groq.py`) re-run clean.

Manually smoke-tested the four real construction paths: factory+dict, factory+BaseLlmConfig, direct `XAILLM(BaseLlmConfig)`, and `XAILLM()`.

## Note on `mem0/utils/factory.py`

The file had a few pre-existing isort/format issues (lmstudio/minimax import order, reranker imports, a long line in `RerankerFactory`). The repo's pre-commit hook auto-fixes these the moment the file is touched, so they're in this diff. The xai-specific changes are just the new `XAIConfig` import and changing the xai registry entry from `BaseLlmConfig` to `XAIConfig`. Happy to split the lint cleanup into a separate PR if you'd prefer.

## Checklist

### DIFF
--- a/mem0/configs/llms/xai.py
+++ b/mem0/configs/llms/xai.py
@@ -1,0 +1,55 @@
+from typing import Optional
+
+from mem0.configs.llms.base import BaseLlmConfig
+
+
+class XAIConfig(BaseLlmConfig):
+    """
+    Configuration class for X.AI (Grok) provider parameters.
+    Inherits from BaseLlmConfig and adds X.AI-specific settings.
+    """
+
+    def __init__(
+        self,
+        # Base parameters
+        model: Optional[str] = None,
+        temperature: float = 0.1,
+        api_key: Optional[str] = None,
+        max_tokens: int = 2000,
+        top_p: float = 0.1,
+        top_k: int = 1,
+        enable_vision: bool = False,
+        vision_details: Optional[str] = "auto",
+        http_client_proxies: Optional[dict] = None,
+        # X.AI-specific parameters
+        xai_base_url: Optional[str] = None,
+    ):
+        """
+        Initialize X.AI configuration.
+
+        Args:
+            model: X.AI / Grok model to use, defaults to None
+            temperature: Controls randomness, defaults to 0.1
+            api_key: X.AI API key, defaults to None
+            max_tokens: Maximum tokens to generate, defaults to 2000
+            top_p: Nucleus sampling parameter, defaults to 0.1
+            top_k: Top-k sampling parameter, defaults to 1
+            enable_vision: Enable vision capabilities, defaults to False
+            vision_details: Vision detail level, defaults to "auto"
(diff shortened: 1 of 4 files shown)


kartik-mem0:
LGTM. This is the cleanest fix for the underlying problem — rather than papering over the AttributeError with `getattr`, it properly wires the factory to use `XAIConfig` so `xai_base_url` is always present, and adds a conversion path for callers that pass a plain `BaseLlmConfig`. The `tools`/`tool_choice` forwarding and `_parse_response` implementation are correct and match the pattern used by other providers. The 209-line test file is thorough, covering base URL priority, factory construction, tool call parsing, and the plain-content path. Good work.

kartik-mem0 (approved):

PR #5202, opened 2026-05-20
fix(vector_stores): pass namespace as top-level kwarg to UpstashVector query_many

### LINKED ISSUE #4207: UpstashVector.search uses namespace inside each query, breaking upstash_vector.query_many

gaurav0107:
## Linked Issue

Closes #4207

## Description

`UpstashVector.search()` was passing `namespace` inside each per-query
dict to `upstash_vector.Index.query_many()`. The Upstash client expects
`namespace` as a **top-level keyword argument**; the `QueryRequest`
TypedDict has no `namespace` field, so the per-query value was silently
dropped and every query routed to the default namespace.

When `len(queries) == 1`, `query_many` internally calls
`self.query(**query, namespace=namespace)` — which raises
`TypeError: got multiple values for keyword argument 'namespace'` because
`namespace` ends up bound twice.

This PR moves `namespace` from each query dict onto `query_many`'s
top-level kwarg, matching the Upstash client signature
([`upstash_vector.core.index_operations.IndexOperations.query_many`](https://github.com/upstash/vector-py/blob/v0.8.0/upstash_vector/core/index_operations.py#L253),
v0.8.0).

```diff
-            queries = [
-                {
-                    "vector": v,
-                    "top_k": top_k,
-                    "filter": filters_str or "",
-                    "include_metadata": True,
-                    "namespace": self.collection_name,
-                }
-                for v in vectors
-            ]
-            responses = self.client.query_many(queries=queries)
+            queries = [
+                {
+                    "vector": v,
+                    "top_k": top_k,
+                    "filter": filters_str or "",
+                    "include_metadata": True,
+                }
+                for v in vectors
+            ]
+            responses = self.client.query_many(queries=queries, namespace=self.collection_name)
```

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — this fixes the search path so `enable_embeddings=False` users
actually receive results from their configured namespace. Users who never
set a custom `collection_name` saw no visible effect; users who did were
silently returned data from the wrong namespace, or hit a `TypeError` on
single-query search.

## Test Coverage

Updated `test_search_vectors` and `test_search_vectors_empty_filters` to
assert the new (correct) call shape, and added
`test_search_vectors_multi_query_namespace_at_top_level` as a regression
test that locks in the fix on the multi-query path. The new test was
verified to **fail** on the buggy code and **pass** on the fix.

```
tests/vector_stores/test_upstash_vector.py ............. 19 passed
```

## Checklist

- [x] I have updated documentation if needed (not applicable — internal
      bug fix in vector store call shape, no public API change)

### DIFF
--- a/mem0/vector_stores/upstash_vector.py
+++ b/mem0/vector_stores/upstash_vector.py
@@ -132,11 +132,10 @@ class UpstashVector(VectorStoreBase):
                     "top_k": top_k,
                     "filter": filters_str or "",
                     "include_metadata": True,
-                    "namespace": self.collection_name,
                 }
                 for v in vectors
             ]
-            responses = self.client.query_many(queries=queries)
+            responses = self.client.query_many(queries=queries, namespace=self.collection_name)
             # flatten
             response = [res for res_list in responses for res in res_list]
 
--- a/tests/vector_stores/test_upstash_vector.py
+++ b/tests/vector_stores/test_upstash_vector.py
@@ -69,11 +69,11 @@ def test_search_vectors(upstash_instance, mock_index):
             {
                 "vector": vectors[0],
                 "top_k": 2,
-                "namespace": "ns",
                 "include_metadata": True,
                 "filter": 'age = 30 AND name = "John"',
             }
-        ]
+        ],
+        namespace="ns",
     )
 
     assert len(results) == 2
@@ -311,11 +311,11 @@ def test_search_vectors_empty_filters(upstash_instance):
             {
                 "vector": vectors[0],
                 "top_k": 1,
-                "namespace": "ns",
                 "include_metadata": True,
                 "filter": "",
             }
-        ]
+        ],
+        namespace="ns",
     )
 
     assert len(results) == 1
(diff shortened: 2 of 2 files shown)


gaurav0107:
Friendly check-in on this one — the fix is small (move `namespace` to the `query_many` top-level kwarg, matching the Upstash client signature), CLA is signed, tests pass locally, and the only failing check is the Vercel preview-deploy (auth, not a code issue). Happy to rebase onto current `main` or split the test into a separate file if that helps reviewers — just let me know what would make this easier to land.

Thanks for taking a look when you get a chance!

kartik-mem0:
## LGTM

**Linked issue**: #4207 (P1-high) — validated
**Bug reproduction**: Confirmed via code tracing — `query_many` takes `namespace` as a top-level kwarg, not per-query. Passing it per-query caused `TypeError` on single-vector search and silent wrong-namespace routing on multi-vector search.
**Pre-push gate**: ruff lint clean, isort clean, 19/19 tests pass
**Independent review**: approved (separate unbiased reviewer)
**Upstash API verification**: Checked against current Upstash Vector Python SDK docs — `namespace` is indeed a top-level `query_many` parameter, not part of the per-query `QueryRequest` dict.

The fix is minimal and correct: moves `namespace` from each per-query dict to the `query_many(namespace=...)` top-level kwarg, matching how every other method in the same file (`upsert`, `delete`, `update`, `fetch`, `resumable_query`, `reset`) already passes it. The new `test_search_vectors_multi_query_namespace_at_top_level` regression test is thorough — it verifies both the call shape and the result flattening across multiple query response lists.

**Optional (pre-existing, not introduced by this PR)**:
- `tests/vector_stores/test_upstash_vector.py:134` has a stray `print("filters", filters)` debug statement
- `mem0/vector_stores/upstash_vector.py` has a minor `ruff format` drift (pre-existing)

Ready for human review and merge.

kartik-mem0 (approved):

PR #5140, opened 2026-05-13
fix(pgvector): convert cosine distance to similarity score

jayadev:
## Description

The `<=>` operator in pgvector returns cosine distance (0 = identical, 2 = opposite), but `score_and_rank` expects similarity scores where higher = more relevant. This caused search results to be ranked inversely — least relevant items appeared first.

Convert distance to similarity: `score = max(0, min(1, 1 - distance))`

This matches the fix already applied in the TypeScript SDK (PR #4944).

Updated test assertions to reflect the corrected scores:
- distance 0.1 → similarity 0.9
- distance 0.2 → similarity 0.8

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Test Coverage

- [x] I added/updated unit tests

PR #5216, opened 2026-05-21
fix(pgvector): convert cosine distance to similarity score

xiaoxiaoHe-E:
## Linked Issue

N/A

## Description

pgvector's \`<=>\` operator returns **cosine distance** (0 = identical, 2 = opposite), but the rest of the memory pipeline treats vector store scores as **similarity values** (higher = more similar). Returning the raw distance caused:

- threshold filtering to reject the most relevant results
- final ranking to sort results in the wrong order

**Fix:** convert the distance to a similarity score before returning results:

# before
```python
score=float(r[1])
```

# after
```python
score=1.0 - float(r[1])
```

## Type of Change

## Breaking Changes

N/A

## Test Coverage

- [x] Updated 8 score assertions in \`tests/vector_stores/test_pgvector.py\` to reflect the corrected similarity values (e.g. distance \`0.1\` → similarity \`0.9\`)

## Checklist

- [x] I have added tests that prove my fix is correct
- [x] Documentation update not needed for this internal fix

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -260,7 +260,7 @@ class PGVector(VectorStoreBase):
             )
 
             results = cur.fetchall()
-        return [OutputData(id=str(r[0]), score=float(r[1]), payload=r[2]) for r in results]
+        return [OutputData(id=str(r[0]), score=1.0 - float(r[1]), payload=r[2]) for r in results]
 
     def keyword_search(self, query, top_k=5, filters=None):
         """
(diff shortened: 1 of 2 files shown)


youneshima:
Closing as a duplicate of #5140 after reviewing the full PR timeline, diff, checks, and comments. Both update the same pgvector scoring path and tests; #5140 has the more complete clamped conversion and active follow-up.

kartik-mem0:
Closing this — the pgvector distance-to-similarity fix is now included in #5391, which landed the same conversion (`max(0.0, 1.0 - distance)`) along with fixes for all other affected backends. Thanks for the contribution!

Issue #5220, opened 2026-05-21
Memory.search() accepts empty/whitespace queries, missing input validation causes crash or unintended memory leak


sharwariakre:
### Component

Core / Python SDK

### Summary

Memory.search() performs no validation on the query string before passing it to the embedding model. Empty strings and whitespace-only strings are treated as valid queries, leading to either a provider-level crash or all stored memories being returned.

### Steps to Reproduce

```python
from mem0 import Memory

m = Memory()  # any config
m.add("I love hiking", user_id="u1")

m.search("", filters={"user_id": "u1"})    # empty string
m.search("   ", filters={"user_id": "u1"}) # whitespace only
```

### Expected Behavior

Returns {"results": []} for blank queries — same as a valid search that matched nothing.

### Actual Behavior

The query is passed directly to the embedding model with no validation. Behavior depends on the provider — some embedders (e.g. Ollama) raise a ValueError on empty input; others silently return a vector, causing all stored memories for that user to be returned.

### Environment

- Affects both sync Memory and async AsyncMemory in mem0/memory/main.py

norika1207-lab:
This is a useful issue to frame around memory reliability as an operational contract, not just a feature toggle.

For `Memory.search accepts empty whitespace queries`, I would try to capture a regression fixture with four visible states: what was accepted into the memory/tool pipeline, what was rejected or transformed, what was persisted, and what later recall returns. That makes it possible to distinguish retrieval quality problems from write-path corruption, provider mismatch, or context-budget behavior.

In production agent systems the failure mode is often silent: a memory write looks successful, a tool output is compressed or cached, then a later agent turn reasons from incomplete state. A small end-to-end test that records the write, index/update, recall, and trace metadata would make this much easier to debug and safer for self-hosted deployments.

PR #5221, opened 2026-05-21
fix: return empty results for empty/whitespace search query

sharwariakre:
## Linked Issue

Closes #5220 

## Description

Memory.search() had no input validation on the query string. Empty strings and whitespace-only queries were passed directly to the embedding model. Depending on the provider, this caused either a crash (ValueError from Ollama) or all stored memories being silently returned. This PR adds an early-return guard in both the sync Memory and async AsyncMemory search() methods that returns {"results": []} for blank queries, consistent with a valid search that matched nothing.

## Type of Change

## Test Coverage

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1176,6 +1176,9 @@ class Memory(MemoryBase):
         # Validate search parameters (before applying defaults)
         _validate_search_params(threshold=threshold, top_k=top_k)
 
+        if not isinstance(query, str) or not query.strip():
+            return {"results": []}
+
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for the contribution! Another PR (#5258) addresses the same issue with a more complete fix — it covers both `Memory` and `MemoryClient`, raises a `ValueError` rather than silently returning empty results, and trims surrounding whitespace before the query reaches the embedder. Closing in favor of that one.

kartik-mem0:

PR #5223, opened 2026-05-22
fix(azure_openai): avoid AttributeError on multimodal message content

ATOM00blue:
## Description

`AzureOpenAILLM.generate_response()` rewrites the last message's content with
`str.replace("assistant", "ai")` unconditionally:

```python
user_prompt = messages[-1]["content"]
user_prompt = user_prompt.replace("assistant", "ai")
messages[-1]["content"] = user_prompt
```

When the last message's `content` is a **list of content parts** (the standard
shape for multimodal / vision messages, e.g. `[{"type": "text", ...}, {"type":
"image_url", ...}]`), `.replace` does not exist on `list`, so the call raises:

```
AttributeError: 'list' object has no attribute 'replace'
```

This crashes any `Memory.add()` / `generate_response()` that includes image
content when the LLM provider is `azure_openai`.

### Reproduction

```python
from unittest.mock import Mock, patch
from mem0.configs.llms.azure import AzureOpenAIConfig

with patch("mem0.llms.azure_openai.AzureOpenAI"):
    llm = AzureOpenAILLM(AzureOpenAIConfig(model="gpt-4.1-nano"))
    llm.generate_response([
        {"role": "user", "content": [
            {"type": "text", "text": "What is in this image?"},
            {"type": "image_url", "image_url": {"url": "https://example.com/a.png"}},
        ]},
    ])
# AttributeError: 'list' object has no attribute 'replace'
```

### Fix

Guard the rewrite so it only runs when the content is a string. List content is
now forwarded unchanged, while the existing string behavior (`assistant` -> `ai`)
is preserved.

## Type of Change

## Test Coverage

Added two regression tests in `tests/llms/test_azure_openai.py`:
- `test_generate_response_with_multimodal_content` — fails on `main` with the
  `AttributeError`, passes with the fix; asserts list content is forwarded
  unchanged.
- `test_generate_response_rewrites_string_assistant_keyword` — pins the existing
  string-rewrite behavior so it stays intact.

All tests in `tests/llms/test_azure_openai.py` pass (15 passed). `ruff check` is
clean on the changed files.

## Checklist

### DIFF
--- a/mem0/llms/azure_openai.py
+++ b/mem0/llms/azure_openai.py
@@ -120,11 +120,13 @@ class AzureOpenAILLM(LLMBase):
             str: The generated response.
         """
 
-        user_prompt = messages[-1]["content"]
-
-        user_prompt = user_prompt.replace("assistant", "ai")
-
-        messages[-1]["content"] = user_prompt
+        # Only the last message's textual content is rewritten. Multimodal
+        # messages carry a list of content parts (e.g. text + image_url), which
+        # has no ``.replace`` method, so guard against non-string content to
+        # avoid an AttributeError on vision/multimodal requests.
+        last_content = messages[-1]["content"]
+        if isinstance(last_content, str):
+            messages[-1]["content"] = last_content.replace("assistant", "ai")
 
         params = self._get_supported_params(messages=messages, **kwargs)
         
--- a/tests/llms/test_azure_openai.py
+++ b/tests/llms/test_azure_openai.py
@@ -91,6 +91,55 @@ def test_generate_response_with_tools(mock_openai_client):
     assert response["tool_calls"][0]["arguments"] == {"data": "Today is a sunny day."}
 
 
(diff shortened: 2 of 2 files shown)


Hexecu:
Thanks for catching this @ATOM00blue — your PR correctly spotted the multimodal `AttributeError` and the in-place mutation in `azure_openai.py`.

I opened #5731 building on the same direction. It additionally avoids corrupting the user's text on the string path (the current `str.replace("assistant", "ai")` rewrites the substring inside the persisted message, e.g. `"my assistant helps me"` -> `"my ai helps me"`) by applying the rewrite to a copy sent to the model only, and it also covers `azure_openai_structured.py` so both Azure providers stay consistent. Verified end-to-end against a live Azure deployment.

Happy to align on whichever approach the maintainers prefer.

kartik-mem0:
#5731 superseded the current pr closing it

PR #5227, opened 2026-05-22
feat(oss): add opt-in parallel entity boost for Python

markymark2001:
## Summary

Add an opt-in `parallel_entity_boost` flag to the Python OSS `MemoryConfig`.
When enabled, entity-boost embedding and entity-store lookup work in `Memory.search()` runs concurrently instead of one entity at a time.

This mirrors the shape of the TypeScript SDK change in #5046 while keeping Python's default behavior unchanged.

## Why

Entity boosting currently embeds and searches each extracted query entity sequentially. With remote embedding providers or networked vector stores, that creates one round trip per entity and can dominate search latency for entity-rich queries.

The new flag defaults to `False` for backward compatibility and for users with rate-limited or single-slot embedding backends. Users who know their embedding/vector-store backends can handle overlap can opt in.

## Implementation

- Adds `parallel_entity_boost: bool = False` to `MemoryConfig`.
- Refactors sync entity-boost work into per-entity workers and uses a bounded `ThreadPoolExecutor` when enabled.
- Refactors async entity-boost work into per-entity coroutines and uses `asyncio.gather()` when enabled.
- Merges per-entity boost maps with `max(existing, boost)`, preserving the existing order-independent scoring semantics.
- Keeps per-entity failures non-fatal, matching the prior best-effort behavior.

## Verification

- `/tmp/taia-mem0/.venv/bin/python -m pytest tests/test_main.py -k "parallel_entity_boost" -v`
  - `3 passed`

The added tests cover:

- default config remains sequential;
- async parallel mode starts all entity workers before any one can complete;
- parallel merging matches sequential merging for duplicate/shared linked memories.

### DIFF
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -55,6 +55,13 @@ class MemoryConfig(BaseModel):
         description="Custom instructions for fact extraction",
         default=None,
     )
+    parallel_entity_boost: bool = Field(
+        description=(
+            "When true, compute entity-boost embed/search calls concurrently "
+            "during memory search. Defaults to false to preserve sequential behavior."
+        ),
+        default=False,
+    )
 
 
 class AzureConfig(BaseModel):
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -6,6 +6,7 @@ import logging
 import os
 import uuid
 import warnings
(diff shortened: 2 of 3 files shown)


kartik-mem0:

PR #5236, opened 2026-05-22
fix: add real GPT-5 names to reasoning model whitelist

### LINKED ISSUE #5233: bug: _is_reasoning_model whitelist missing real GPT-5 names (gpt-5-mini, gpt-5-nano)
## Summary

`mem0.llms.base.LLMBase._is_reasoning_model` whitelists model names that don't exist in the OpenAI catalog (`gpt-5o`, `gpt-5o-mini`, `gpt-5o-micro`) while missing the real GPT-5 family names — `gpt-5-mini` and `gpt-5-nano`. As a result, `Memory.add()` calls against these models fail with HTTP 400 because the legacy `max_tokens` parameter is sent to a reasoning model that only accepts `max_completion_tokens`.

This affects both OpenAI direct (`provider: openai`) and Azure OpenAI (`provider: azure_openai`) deployments equally — the model identifiers are the same on both surfaces.

## Repro (mem0ai 2.0.2)

```python
from mem0 import Memory

config = {
    "llm": {
        "provider": "openai",  # or azure_openai
        "config": {
            "model": "gpt-5-mini",
            "max_tokens": 1024,
            "temperature": 0.1,
            "api_key": "...",
        },
    },
    # ... vector_store / embedder ...
}
m.add([{"role": "user", "content": "hello"}], user_id="alice")
```

### Observed

```
HTTP/1.1 400 Bad Request
ERROR LLM extraction failed: Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}
```

### Expected

`_is_reasoning_model('gpt-5-mini')` returns `True`, `max_tokens` is filtered out by `_get_supported_params`, the call succeeds.

## Root cause

`mem0/llms/base.py` (current main):

```python
reasoning_models = {
}
```

- `gpt-5o`, `gpt-5o-mini`, `gpt-5o-micro` **are not real OpenAI model identifiers** — they appear to be a confusion with the GPT-4o ("omni") naming. The official GPT-5 family per [platform.openai.com/docs/models](https://platform.openai.com/docs/models) is `gpt-5` / `gpt-5-mini` / `gpt-5-nano`.
- The set lookup `base_model in reasoning_models` is exact-match, so `'gpt-5-mini' in reasoning_models` → `False`.
- The regex/prefix fallback below only matches `o1-` / `o3-` families:
  ```python
      return True
  ```
- So `gpt-5-mini` (and dated variants like `gpt-5-mini-2025-08-07` that Azure deployments may surface) falls through to `return False` and inherits the non-reasoning param set, which sends `max_tokens`.

## Suggested fix

Replace the literal-set entries with the real names and add a prefix match for dated variants:

```python
reasoning_models = {
    "gpt-5", "gpt-5-mini", "gpt-5-nano",
}
# ...
if any(base_model.startswith(prefix) for prefix in [
    "o1-", "o1.", "o3-", "o3.",
    "gpt-5-mini-", "gpt-5-nano-",  # dated suffixes like gpt-5-mini-2025-08-07
]):
    return True
```

Note that this is **distinct from #5054**, which is about `gpt-5.4-mini` (a different model family — that one supports `max_tokens` so it should remain *out* of the reasoning whitelist). The two cases need to be handled independently.

## Workaround in use

Locally I'm monkey-patching at import time. Works but a real upstream fix is obviously preferable:

```python
from mem0.llms import base as _mem0_base
_orig = _mem0_base.LLMBase._is_reasoning_model
def _patched(self, model):
    base = (model or "").lower().rsplit("/", 1)[-1]
    if base in {"gpt-5-mini", "gpt-5-nano"} or base.startswith(("gpt-5-mini-", "gpt-5-nano-")):
        return True
    return _orig(self, model)
_mem0_base.LLMBase._is_reasoning_model = _patched
```

## Environment

- mem0ai 2.0.2
- Python 3.12
- Reproduced against an Azure OpenAI `gpt-5-mini` deployment (api-version `2025-04-01-preview`); same model id is served by OpenAI direct so the bug affects both providers

Happy to open a PR with the fix if maintainers want.

sharwariakre:
## Linked Issue

Closes #5233 

## Description

`LLMBase._is_reasoning_model` whitelisted model names that don't exist in the OpenAI catalog (`gpt-5o`, `gpt-5o-mini`, `gpt-5o-micro`) while missing the real GPT-5 family names — `gpt-5-mini` and `gpt-5-nano`. Because the set lookup is exact-match and the prefix fallback only covered `o1-`/`o3-`, calls against `gpt-5-mini`/`gpt-5-nano` fell through to the non-reasoning param set and sent the legacy `max_tokens` parameter, which these models reject with HTTP 400 (`'max_tokens' is not supported with this model. Use 'max_completion_tokens'`).

This PR replaces the non-existent names with the real ones and adds prefix matching for dated variants (e.g. `gpt-5-mini 2025-08-07`, which Azure deployments surface). 

## Type of Change

## Test Coverage

Extended `test_is_reasoning_model_classification` to assert `gpt-5-mini`, `gpt-5-nano`, and their dated variants classify as reasoning models, while `gpt-5.4-mini` stays non-reasoning.

## Checklist

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -52,7 +52,7 @@ class LLMBase(ABC):
         """
         reasoning_models = {
             "o1", "o1-preview", "o3-mini", "o3",
-            "gpt-5", "gpt-5o", "gpt-5o-mini", "gpt-5o-micro",
+            "gpt-5", "gpt-5-mini", "gpt-5-nano",
         }
 
         model_lower = model.lower()
@@ -62,9 +62,9 @@ class LLMBase(ABC):
         if base_model in reasoning_models:
             return True
 
-        # Match o1/o3 family with prefixes (o1-2024-12-17, o3-2025-04-16)
+        # Match o1/o3 family and gpt-5-mini/nano dated variants
         # but NOT gpt-5.x variants (gpt-5.4-mini supports temperature)
-        if any(base_model.startswith(prefix) for prefix in ["o1-", "o1.", "o3-", "o3."]):
+        if any(base_model.startswith(prefix) for prefix in ["o1-", "o1.", "o3-", "o3.", "gpt-5-mini-", "gpt-5-nano-"]):
             return True
 
         return False
--- a/tests/llms/test_openai.py
+++ b/tests/llms/test_openai.py
@@ -317,6 +317,10 @@ def test_is_reasoning_model_classification(mock_openai_client):
     assert llm._is_reasoning_model("o3-mini") is True
     assert llm._is_reasoning_model("o3") is True
     assert llm._is_reasoning_model("gpt-5") is True
+    assert llm._is_reasoning_model("gpt-5-mini") is True
+    assert llm._is_reasoning_model("gpt-5-nano") is True
+    assert llm._is_reasoning_model("gpt-5-mini-2025-08-07") is True
+    assert llm._is_reasoning_model("gpt-5-nano-2025-08-07") is True
     assert llm._is_reasoning_model("o1-preview") is True
     assert llm._is_reasoning_model("o1-2024-12-17") is True
     assert llm._is_reasoning_model("openai/o3-mini") is True


kartik-mem0:
Thanks for jumping on this so quickly, @sharwariakre! The underlying 400 was since resolved on `main` by #5547, which took a slightly different route: instead of adding `gpt-5-mini`/`gpt-5-nano` to the reasoning whitelist, it added `_uses_max_completion_tokens()` so all `gpt-5*` models send `max_completion_tokens` — fixing the error while keeping `temperature`/`top_p` support (the GPT-5 chat family accepts them, so we intentionally keep them out of the reasoning-param-stripping set). That makes the whitelist change here unnecessary, so I'm closing this as superseded. Really appreciate the fast turnaround on it. 🙏

PR #5238, opened 2026-05-23
docs: add Ejentum harness cookbook (perception-sharpening around mem0)

ejentum:
## Summary

Adds a new cookbook page at `docs/cookbooks/integrations/ejentum-harness.mdx` and registers it in `docs/docs.json` under Cookbooks > Integrations.

The cookbook covers two patterns that use the Ejentum cognitive harness (memory mode) as a perception layer around mem0, not as a replacement:

1. **Pre-storage filter.** Call the memory-mode harness before `mem0.add(...)` to score whether a candidate signal deserves long-term storage. Keeps mem0 lean by skipping noise like timestamps and generic statements.
2. **Post-recall scoring.** After `mem0.search(...)`, use the harness as a reranker that distinguishes materially useful hits from lexically similar ones.

Both patterns are plain-Python composition over mem0's public API and the Ejentum REST gateway, so they slot into any orchestration.

## What this is not

- Not a vendor pitch in the cookbook. The page leads with how the two products complement each other (perception versus storage), not with marketing about the harness.
- Not a replacement for `mem0.add`/`mem0.search`. Both stay as the storage primitive; the harness is the gate around them.

## Files changed

- `docs/cookbooks/integrations/ejentum-harness.mdx` (new, 181 lines)
- `docs/docs.json` (one-line registration, alphabetical slot between `openai-tool-calls` and `mastra-agent`)

## Affiliation

I maintain the Ejentum harness API. Submitting this as a cookbook because the pre-storage-filter pattern is a real use we have seen and the post-recall-reranker pattern keeps prompts focused without changing mem0's behaviour. The cookbook structure mirrors existing entries in the same folder (frontmatter, install block, two-pattern body, notes on cost and latency). Ejentum has free and paid tiers; the cookbook links to the dashboard for keys but not to a checkout.

## Test plan

- [x] Frontmatter and structure match existing integration cookbooks.
- [x] Page registered in `docs.json` in the expected alphabetical slot.
- [x] Voice scrubbed (no em dashes).
- [x] Two patterns are independent; each runs standalone.
- [x] DCO sign-off on the commit.
- [ ] Local Mintlify preview (cannot run mintlify dev in this environment; happy to follow up if a maintainer spots a rendering issue).

### DIFF
--- a/docs/cookbooks/integrations/ejentum-harness.mdx
+++ b/docs/cookbooks/integrations/ejentum-harness.mdx
@@ -1,0 +1,180 @@
+---
+title: Decide what to remember with Ejentum harness
+description: "Use the Ejentum cognitive harness as a perception-sharpening layer around mem0: filter what gets stored, and score what gets recalled."
+---
+
+[Ejentum](https://ejentum.com) is a cognitive harness API that returns task-matched reasoning scaffolds for one of four modes: `reasoning`, `code`, `anti-deception`, `memory`. The `memory` mode is the relevant one here. Its scaffolds describe **how to attend to a candidate signal** (what to detect, what to suppress, what to verify) rather than storing or retrieving data itself. That makes it a natural complement to mem0, which handles the storage and recall side.
+
+This cookbook covers two patterns:
+
+1. **Pre-storage filter.** Before calling `mem0.add(...)`, retrieve a perception scaffold and use it to score whether the candidate signal deserves long-term storage.
+2. **Post-recall scoring.** After `mem0.search(...)` returns candidates, retrieve a different scaffold to assess which results actually match the user's current intent before passing them to your downstream LLM.
+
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for putting this together, @ejentum, and for being upfront about the affiliation — the write-up is clearly structured and the perception-vs-storage framing is thoughtful.

That said, we're keeping the cookbooks section focused on patterns built on mem0 plus broadly-adopted, community-driven tools, rather than integration pages authored by the maintainer of the third-party product being featured. It's a fairness/scope call, not a quality one, so we're going to pass on merging this here.

You're very welcome to publish this guide on your own docs/blog and link to mem0 — and if Ejentum builds up independent community usage down the line, an integration can always be revisited. Appreciate your interest in mem0!

PR #5239, opened 2026-05-23
fix: use utf-8 for json config files

Ghraven:
## Summary
- Add explicit `encoding="utf-8"` when reading GCP service account JSON files
- Add explicit `encoding="utf-8"` for the local `~/.mem0/config.json` read/write helpers

## Problem
These files are JSON/text files, but a few `open()` calls still relied on the platform default encoding. On Windows or other non-UTF-8 locales, that can make config or credential parsing depend on the user's locale instead of the file format.

## Before / After
Before, JSON config files were opened without an explicit encoding.

After, the same reads/writes use UTF-8 explicitly, matching JSON's common/default encoding and avoiding locale-dependent behavior.

## Verification
- `python -m compileall -q mem0\utils\gcp_auth.py mem0\memory\setup.py`
- `git diff --check`

### DIFF
--- a/mem0/memory/setup.py
+++ b/mem0/memory/setup.py
@@ -23,7 +23,7 @@ def _load_config():
     if not os.path.exists(path):
         return {}
     try:
-        with open(path, "r") as f:
+        with open(path, "r", encoding="utf-8") as f:
             data = json.load(f)
         return data if isinstance(data, dict) else {}
     except Exception as e:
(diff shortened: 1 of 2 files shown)


Ghraven:

kartik-mem0:

Ghraven:
Thanks for the note, and sorry this got stale/conflicted while it was waiting. I appreciate you cleaning up the backlog.

I won't reopen this as-is. If this is still useful on current `main`, I can come back with a fresh, focused PR instead.

PR #5249, opened 2026-05-25
fix: raise EmbeddingError on partial embed failure instead of silently dropping memories

### LINKED ISSUE #5245: Silent memory loss when batch embedding partially fails in V3 add pipeline
# Silent memory loss when batch embedding partially fails in V3 add pipeline

## Summary

`Memory.add()` can silently drop extracted memories when the embedding provider fails on individual items inside a batch. The failure is logged at `WARNING` level but **no exception is raised to the caller**, so the application has no way to know that a fact extracted by the LLM was never persisted.

Observed on commit `6b9707f` (current `main`).

## Root cause

`mem0/memory/main.py:770-783`, phase 3 of the V3 add pipeline:

```python
# Phase 3: Batch embed all extracted memory texts
mem_texts = [m.get("text", "") for m in extracted_memories if m.get("text")]
try:
    mem_embeddings_list = self.embedding_model.embed_batch(mem_texts, "add")
    embed_map = dict(zip(mem_texts, mem_embeddings_list))
except Exception:
    # Fallback: embed individually
    embed_map = {}
    for text in mem_texts:
        try:
            embed_map[text] = self.embedding_model.embed(text, "add")
        except Exception as e:
            logger.warning(f"Failed to embed memory text: {e}")
```

When the per-item fallback path is taken and an individual `embed()` call fails (provider timeout, throttle, transient 5xx, malformed response, etc.), the corresponding `text` is **never added to `embed_map`**. Downstream, the loop that materializes memories only iterates over keys present in `embed_map`, so the fact is silently dropped. No exception escapes the function, the caller of `Memory.add()` receives a normal response.

The same pattern exists at:

- `main.py:885-893` (`_resolve_entity_embeddings`, entity batch)
- `main.py:452-454` (`_upsert_entity`, swallow + log)
- `main.py:1466` (boost path, swallow + log)
- `main.py:485-503` (`_remove_memory_from_entity_store`, swallow at `debug` level — invisible in prod logging)

## Reproduction

Minimal repro using `MockEmbeddings` extended to fail every Nth call:

```python
import logging
from mem0 import Memory
from mem0.embeddings.mock import MockEmbeddings

logging.basicConfig(level=logging.INFO)

class FlakyEmbedder(MockEmbeddings):
    """Fails on every 3rd call to simulate a transient provider error."""
    def __init__(self, config=None):
        super().__init__(config)
        self.call_count = 0
    def embed(self, text, memory_action=None):
        self.call_count += 1
        if self.call_count % 3 == 0:
            raise RuntimeError("Simulated provider 503")
        return super().embed(text, memory_action)
    def embed_batch(self, texts, memory_action="add"):
        # Force the per-item fallback path
        raise RuntimeError("Batch endpoint unavailable")

# Patch the factory to use FlakyEmbedder
# ...(setup omitted for brevity)...

m = Memory(config=...)
result = m.add(messages=[{"role": "user", "content": "A is friends with B, C, D, E, F."}], user_id="test")
# Expect 5 facts extracted. Observe that only ~2/3 actually land in the store.
# No exception raised. Only WARNING logs.
```

A complete runnable repro is available at: https://github.com/Adelagric/mem0-silent-loss-repro

## Impact

- The most expensive part of the pipeline (LLM extraction) silently loses output downstream.
- Default production logging usually filters `WARNING`, so operators do not see the loss.
- No counter, no metric, no callback — only `mem0/memory/telemetry.py:202-206` ships static config attributes (`vector_size`, `embedding_model`) via PostHog, which does not surface this per-item failure.
- Related historical bugs in the codebase suggest this class of issue has occurred before:
  - PR #4362 (`Prevented embedding corruption in Valkey and Redis when vector is None`) — patched in the store, not upstream in the embedder.
  - PR #4224 (`Normalized malformed LLM fact output before embedding`).
  - PR #4058 (`Pass encoding_format='float' in OpenAI embeddings for proxy compatibility`).

## Questions

1. Is the silent drop the intended behavior? If yes, this should be explicit in the docs (operators currently have no way to know).
2. Would a strict mode (`raise_on_partial_failure=True`) be acceptable? It would give the caller the ability to retry / alert.
3. Is there interest in a callback hook (`on_embedding_failure: Callable[[str, Exception], None]`) that operators could wire to their own observability stack?

Happy to provide a more detailed repro or sketch a minimal patch for any of the three options above, depending on the direction the maintainers prefer.

---

*Reported by [Adel Kaleche](mailto:kaleche@gmail.com) — building infrastructure middleware for embedding pipelines. Not affiliated with Mem0.*

PHclaw:
## Summary

Fixes #5245 — `Memory.add()` silently drops extracted memories when the embedding provider fails on individual items in the batch-embed fallback path.

## Root Cause

In `mem0/memory/main.py` Phase 3, when batch embedding fails and the per-item fallback path is taken, individual `embed()` failures were logged at WARNING level but execution continued — the corresponding text was never added to `embed_map`. Downstream, only texts present in `embed_map` get materialized as memories, so failed embeddings are silently dropped with no exception raised to the caller.

The same silent-drop pattern existed in `_async_add_memory` (async version).

## Fix

After collecting failed texts, raise a new `EmbeddingError` exception instead of silently continuing:

```python
failed_texts = []
for text in mem_texts:
    try:
        embed_map[text] = self.embedding_model.embed(text, "add")
    except Exception as e:
        logger.warning(f"Failed to embed memory text: {e}")
        failed_texts.append(text)
if failed_texts:
    raise EmbeddingError(
        f"Failed to embed {len(failed_texts)} memory text(s), "
        f"dropped memories: {failed_texts}"
    )
```

The same fix applied to the async version (`_async_add_memory`).

A new `EmbeddingError` exception class is added to `mem0/memory/main.py`.

## Testing

```python
from mem0.memory.main import EmbeddingError

class FlakyEmbedder:
    def __init__(self): self.c = 0
    def embed(self, text, memory_action=None):
        self.c += 1
        if self.c % 3 == 0: raise RuntimeError("Simulated 503")
        return [0.0] * 384

m = Memory(config=...)
try:
    m.add([{"role": "user", "content": "A is friends with B, C, D."}], user_id="test")
    assert False, "Expected EmbeddingError"
except EmbeddingError as e:
    assert "Failed to embed" in str(e)  # raised, not silently dropped
```

Closes #5245

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -41,6 +41,12 @@ from mem0.utils.factory import (
 )
 from mem0.utils.lemmatization import lemmatize_for_bm25
 from mem0.utils.scoring import (
+
+
+class EmbeddingError(Exception):
+    """Raised when embedding a memory text fails, causing silent data loss."""
+    pass
+
     ENTITY_BOOST_WEIGHT,
     get_bm25_params,
     normalize_bm25,
@@ -775,11 +781,18 @@ class Memory(MemoryBase):
         except Exception:
             # Fallback: embed individually
             embed_map = {}
+            failed_texts = []
             for text in mem_texts:
                 try:
                     embed_map[text] = self.embedding_model.embed(text, "add")
                 except Exception as e:
                     logger.warning(f"Failed to embed memory text: {e}")
+                    failed_texts.append(text)
+            if failed_texts:
+                raise EmbeddingError(
+                    f"Failed to embed {len(failed_texts)} memory text(s), "
+                    f"dropped memories: {failed_texts}"
+                )
 
         # Phase 4: Per-memory CPU processing + Phase 5: Hash dedup
         # Build set of existing hashes for dedup
@@ -2190,11 +2203,18 @@ class AsyncMemory(MemoryBase):
             embed_map = dict(zip(mem_texts, mem_embeddings_list))
         except Exception:
             embed_map = {}
+            failed_texts_async = []
             for text in mem_texts:
                 try:
                     embed_map[text] = await asyncio.to_thread(self.embedding_model.embed, text, "add")
                 except Exception as e:
                     logger.warning(f"Failed to embed memory text (async): {e}")
+                    failed_texts_async.append(text)
+            if failed_texts_async:
+                raise EmbeddingError(
+                    f"Failed to embed {len(failed_texts_async)} memory text(s) (async), "
+                    f"dropped memories: {failed_texts_async}"
+                )
 
         # Phase 4: Per-memory CPU processing + Phase 5: Hash dedup
         existing_hashes = set()


Adelagric:
The narrow fix here (raise instead of swallow in `main.py:770-783`) is the right direction, but the same swallow pattern exists at `main.py:885-893` (`_resolve_entity_embeddings`), `main.py:1466` (entity boost) and `main.py:485-503` (`_remove_memory_from_entity_store`, swallowed at `debug` level — invisible in default prod logging).

Independent of the in-process fix path, an out-of-process validator in front of the vector DB catches this class of bug for any caller, language-agnostic. I open-sourced one yesterday under Apache 2.0: https://github.com/Adelagric/vector-router — drop-in for OpenAI-compatible providers via `OPENAI_BASE_URL`, plus Prometheus labels per `producer_id`. Posting in case it's useful to anyone hitting this until the SDK patch lands.

Adelagric:
The direction here — raise instead of swallow — is right, and covering both the sync and async `_add_to_vector_store` paths is the correct scope. A few things from tracing it on current `main`, before it can land:

**It doesn't import as written.** The `EmbeddingError` class is inserted inside the `from mem0.utils.scoring import (...)` block, so the module fails to parse — `ast.parse` raises `SyntaxError` at the `class` line. Moving the definition out of the import list (e.g. just below the imports, near `logger = logging.getLogger(__name__)`) fixes it.

**Raising aborts partial successes.** The `raise` fires after the fallback loop whenever any item failed, so it unwinds the whole `add()` before Phase 4 persists — the texts that *did* embed (already in `embed_map`) are discarded along with the failed ones. That converts a silent partial success into a loud total failure: better, but it still loses the good items. @inland-taipen's `results` / `failed` bulk-response shape in #5245 keeps the successes and reports failures per-item — worth aligning with that instead of all-or-nothing.

**Coverage.** This patches the two memory-embed fallback loops (sync ~L782, async ~L2197). Updating my earlier site list with precise current locations — the same swallow remains in the entity paths, each with a sync and an async copy:
- entity embed (step 7b) in `_add_to_vector_store` — sync ~L887 / async ~L2303 — a failed per-item `embed()` appends `None` and is silently filtered, with no log at all;
- entity boost — `_compute_entity_boosts` ~L1497 / `_compute_entity_boosts_async` ~L2905 — `warning`;
- entity-store cleanup — `_remove_memory_from_entity_store` re-embed, sync ~L495 / async ~L1928 — swallowed at `debug`, invisible under default prod log levels.

**One design question**, tied to the #5245 RFC discussion (the part I offered to formalize): should `EmbeddingError` carry a class — `provider_error` vs `validation_error` — so callers can branch retry vs escalate rather than parsing the message string? Worth noting the NaN / wrong-dimension case isn't an exception at all today — the provider returns a bad vector that doesn't raise — so it's a separate detection point from the call-failure this PR handles.

kartik-mem0:
## Changes Requested

### Strengths
- The root cause diagnosis is accurate — the collect-then-raise pattern (`failed_texts` list, append in except, raise after the loop) is structurally sound and gives a complete count and list in the error message.
- Both sync (`_add_to_vector_store` ~line 781) and async (`_add_to_vector_store` ~line 2203) paths were updated, which is the right call — appreciate the async parity.
- Conventional commit title is well-formed.

### Issues

#### Critical (must fix)

1. `mem0/memory/main.py:43–54` — **SyntaxError: `EmbeddingError` is placed inside the open parenthesis of the `from mem0.utils.scoring import (` block.** Python sees `class` inside a parenthesized expression and raises `SyntaxError: invalid syntax`. The file is completely unimportable on this branch — `Memory`, `AsyncMemory`, everything in the module fails to load. Confirmed via `python3 -c "import ast; ast.parse(open('mem0/memory/main.py').read())"` on the PR head:
   ```
   SyntaxError: invalid syntax
     File "<unknown>", line 46: class EmbeddingError(Exception):
   ```
   The class needs to move to *after* the closing `)`. One-minute fix — but it's a ship-blocker.

2. `mem0/memory/main.py:46` — **`EmbeddingError` is a duplicate of `mem0/exceptions.py:326`**, which already defines `class EmbeddingError(MemoryError)` with a structured constructor (`message=`, `error_code=`, `details=`, `suggestion=`). The PR's bare `class EmbeddingError(Exception)` is a different class:
   - It does not inherit from `MemoryError`, so `except MemoryError` catch-alls used by callers won't catch it.
   - The constructor signatures are incompatible — callers that import `from mem0.exceptions import EmbeddingError` vs `from mem0.memory.main import EmbeddingError` get two entirely different types.
   
   Please remove the new class definition and instead `from mem0.exceptions import EmbeddingError` (it's already in that file). Then update the raise sites to use the existing constructor: `raise EmbeddingError(message=f"Failed to embed {len(failed_texts)} memory text(s), ...")`.

#### Important (should fix)

3. **No regression test.** The PR description includes a manual test snippet, but nothing was added to `tests/`. For a bug fix on a data-loss path, a pytest test is needed — one that stubs `embedding_model.embed_batch` to raise and `embedding_model.embed` to raise for at least one call, then asserts `EmbeddingError` propagates from `_add_to_vector_store`. The async path needs its own variant. Without this, a refactor could silently revert the fix. `tests/memory/test_main.py` already has mock infrastructure for the embedding model — that's the right place.

4. `mem0-ts/src/oss/src/memory/index.ts:705–711` — **The identical silent-drop bug exists in the TypeScript SDK and is not fixed here.** The TS fallback path has the same pattern:
   ```typescript
   } catch (e) {
     console.warn(`Failed to embed memory text: ${e}`);
     // no throw — memory silently dropped
   }
   ```
   This PR doesn't need to fix it, but please open a follow-up issue for the TS counterpart so it doesn't get lost.

#### Minor (optional)

5. `mem0/memory/main.py:2206` — `failed_texts_async` vs `failed_texts` in the sync path. They're in different method scopes so there's no conflict; `failed_texts` in both is cleaner and consistent with the rest of the file's naming style.

6. `mem0/memory/main.py:793` — the exception message includes the full text of every failed memory (`dropped memories: {failed_texts}`). For large or sensitive memory content this leaks the raw text into exception strings and logs. Consider just the count in the message and let `details=` carry the list if you use the structured `EmbeddingError` constructor.

### Evidence
- **CI:** Vercel FAILURE (infra/auth — not code); license/cla PENDING; no pytest matrix visible yet.
- **Syntax check:** `python3 -c "import ast; ast.parse(open('mem0/memory/main.py').read())"` → `SyntaxError: invalid syntax` at line 46 on PR branch head `4b79494e`.
- **Ruff:** `ruff check mem0/memory/main.py` → `invalid-syntax: Expected one or more symbol names after import` at line 43.
- **Targeted tests / red-green:** skipped — file has a SyntaxError preventing import; no regression test was added in this PR.
- **Duplicate class:** `grep -n "EmbeddingError" mem0/exceptions.py` → `326: class EmbeddingError(MemoryError)` confirmed.
- **TS counterpart:** `mem0-ts/src/oss/src/memory/index.ts:705–711` has the same swallow pattern — confirmed by direct read.
- **Could NOT verify:** live embedding provider behavior (code tracing only).

### Assessment
Ready to merge: **With fixes** — please address the two Critical items (syntax error and duplicate class); I'll re-review once you push. The core logic of collecting failed texts and raising is correct — just needs the class placement fixed and pointed at the existing `EmbeddingError` in `mem0/exceptions.py`.

Adelagric:
+1 on using the existing `EmbeddingError` from `mem0/exceptions.py` rather than a new class — that also gives the `error_class` question I raised above a clean home. Since it already carries `error_code` / `details`, the taxonomy is just the controlled vocabulary for that existing `error_code`, not new surface: `EMBED_001` becomes a provider / validation / internal split callers can branch on (retry vs reconfigure vs escalate).

Two things that fall out of leaning on the existing hierarchy:
- the provider half is partly modeled already — `RateLimitError` / `NetworkError` exist for the transient case (today raised only in `client/utils.py`, never in the embed path), so detection can route to them and the serializer normalizes to `error_code`;
- the validation half (NaN / wrong-dimension) needs a *new* detection point — a bad-but-returned vector doesn't raise today, so it'd set `error_code` at the check, before persist.

Happy to draft that vocabulary as the error_class piece I picked up in #5245, scoped to plug into whatever `EmbeddingError` shape lands here. And +1 on the regression test and the TS follow-up issue.

Adelagric:
Filed the TS counterpart as #5509 — same swallow-and-drop in `mem0-ts`, `index.ts:705–711` (primary, `console.warn`) plus the entity sites at `debug`. Scoped to mirror whatever lands here.

kartik-mem0:
Thanks for this, @PHclaw! We're consolidating the partial-embedding-failure fix on #5557, which carries the full error taxonomy, preserves partial successes, and adds async parity. Closing this one as superseded — appreciate you surfacing the silent-loss problem. 🙏

PR #5250, opened 2026-05-25
fix(redis): convert
  cosine distance to similarity score in search results

roli-lpci:
## What
The Redis vector store returned the raw cosine `vector_distance` (lower = more similar) in the `score` field, but downstream scoring treats `score` as a similarity (higher = more similar): dedup fires at `score >= 0.95` and there is a minimum-score threshold. As a result, Redis search scoring was inverted/uncalibrated.

Related to #4999. PR #5044 fixes the same class of bug for ChromaDB; this is the Redis counterpart.

## Why
`mem0/vector_stores/redis.py` assigned `score=float(result["vector_distance"])`. For RediSearch's cosine metric (the default here), distance and similarity are inverse, so a nearer result received a lower score.

## How
Convert distance to similarity with `score = 1 - vector_distance`, matching the convention already used by the weaviate store (`score = 1 - distance`). Adds a regression test (`test_search_converts_cosine_distance_to_similarity`) asserting the nearer result gets the higher score. `pytest tests/vector_stores/test_redis.py` passes (3 tests).

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -158,7 +158,10 @@ class RedisDB(VectorStoreBase):
         return [
             MemoryResult(
                 id=result["memory_id"],
-                score=float(result["vector_distance"]),
+                # RediSearch returns cosine distance (lower = more similar), but
+                # downstream scoring treats score as a similarity (higher = more
+                # similar). Convert to match the weaviate store's convention. See #4999.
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Closing this — the Redis cosine distance-to-similarity conversion is now included in #5391, which fixed all 11 affected vector store backends in one go. Thanks for the PR!

PR #5255, opened 2026-05-26
fix(codex-hooks): refuse install on Windows native with WSL/Git Bash guidance (#5243)

### LINKED ISSUE #5243: ［ｗｉｎｄｏｗｓ］Mem0 Codex lifecycle hooks trigger Windows "Open With" popups because .sh scripts are registered directly
### Component

Core / Python SDK

## Summary

On Windows native Codex Desktop/CLI, enabling Mem0 Codex lifecycle hooks registers `.sh` scripts directly in `~/.codex/hooks.json`. When Codex tries to run them, Windows opens repeated "Open With / 앱 선택" dialogs instead of executing the scripts.

Direct MCP / Mem0 tools continue to work. The issue appears specific to lifecycle hooks on native Windows.

## Environment

- Windows 11
- Codex Desktop + Codex CLI
- Mem0 plugin: `mem0@mem0-plugins` v0.2.3
- Native PowerShell environment
- Not using WSL execution mode for Codex

## Steps to Reproduce

1. Sideload the Mem0 plugin for Codex from `mem0ai/mem0`.
2. Install/enable the plugin in Codex.
3. Run the bundled Codex hooks installer:

   ```bash
   python3 ~/codex-plugins/mem0-source/mem0-plugin/scripts/install_codex_hooks.py

Enable hooks in ~/.codex/config.toml:
[features]
codex_hooks = true

Restart Codex and use a session normally.

Expected Behavior
Hooks either run correctly on Windows, or Windows users are clearly guided to use WSL/Git Bash / Direct MCP without lifecycle hooks.

Actual Behavior
Windows repeatedly opens "Open With" dialogs for the registered .sh hook scripts instead of executing them.

In my case this created many OpenWith.exe processes and made Codex difficult to use until hooks were disabled.

Workaround
Disable Codex hooks and keep Mem0 MCP/plugin enabled:

[features]
codex_hooks = false
I also cleared the generated hook manifests to stop stale hook candidates:

{
  "hooks": {}
}
After restarting Codex, the popup loop stopped and Mem0 MCP still worked.

Suggested Fix
One of:

On Windows, register hook commands as bash "<script>.sh" if a compatible shell is available.
Make install_codex_hooks.py detect native Windows and warn/refuse unless WSL/Git Bash support is confirmed.
Document that lifecycle hooks currently require a Unix-like shell/WSL and recommend Direct MCP on native Windows.
Notes
The current Codex docs/instructions use Unix-style examples such as python3, bash, and ~/..., but I did not find a Windows-specific warning for lifecycle hooks.

leavedrop:
Fixes #5243.

Codex lifecycle hooks register `.sh` paths directly in `~/.codex/hooks.json`.
On native Windows `.sh` has no default handler, so Codex spawning a hook
triggers an "Open With" dialog (one `OpenWith.exe` per event), making Codex
unusable until hooks are disabled.

This patch makes `install_codex_hooks.py` exit cleanly (code 2) on Windows
native Python with guidance to re-run from WSL or Git Bash, where `bash.exe`
is on PATH. `--uninstall` continues to work so already-installed users can
clean up.

Per the original report, MCP / Direct Mem0 tools are unaffected — only the
lifecycle hooks installer. Cursor and Claude Code hook templates are out of
scope; the reporter only observed this on Codex.

I considered auto-rewriting `command` entries to `bash <path>.sh` on Windows
but didn't go that route — it would need to pick between Git Bash / WSL /
MSYS2 bash with different working-directory semantics, and Codex itself may
add native-Windows hook execution. A future PR can layer auto-wrap once
Codex Windows behavior stabilizes.

## Test

- Windows 11 native Python: install exits with WSL guidance (code 2),
  `~/.codex/hooks.json` untouched.
- Windows 11 native Python: `--uninstall` still strips owned entries.
- Linux / WSL: no behavior change (`platform.system()` early-return only
  fires on Windows).

### DIFF
--- a/mem0-plugin/scripts/install_codex_hooks.py
+++ b/mem0-plugin/scripts/install_codex_hooks.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import platform
 import sys
 from pathlib import Path
 
@@ -125,6 +126,19 @@ def main() -> int:
         write_config(config)
         print(f"Removed Mem0 hooks from {HOOKS_FILE}")
         return 0
+
+    # Codex lifecycle hooks register .sh paths directly in ~/.codex/hooks.json.
+    # On native Windows .sh has no default handler, so Codex spawning a hook
+    # triggers "Open With" dialogs (one OpenWith.exe per event). See #5243.
+    if platform.system() == "Windows":
+        print(
+            "Codex lifecycle hooks register .sh scripts directly, which Windows\n"
+            "cannot execute without a bash interpreter on PATH. Re-run this\n"
+            "installer from WSL or Git Bash, or use Mem0 via MCP / Direct tools\n"
+            "without lifecycle hooks. See https://github.com/mem0ai/mem0/issues/5243.",
+            file=sys.stderr,
+        )
+        return 2
 
     if not TEMPLATE_FILE.exists():
         print(f"error: template not found at {TEMPLATE_FILE}", file=sys.stderr)


kartik-mem0:
thank you for your contributions @leavedrop but this issue has been resolved with this pr #5257.

closing this pr.

PR #5258, opened 2026-05-26
fix(memory): reject empty search queries

RitwijParmar:
## Summary
- Reject empty or whitespace-only queries in sync/async `Memory.search()` before lemmatization, embedding, or vector-store calls.
- Apply the same validation in sync/async `MemoryClient.search()` before sending API requests.
- Normalize leading/trailing whitespace so downstream embedding/search uses the intended query text.

Fixes #5220.

## Verification
- `/Users/ritwij/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m pytest tests/test_main.py::TestSearchParamValidation tests/test_client.py::TestSearchEntityParamRejection -q` (20 passed)
- `/Users/ritwij/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m ruff check mem0/memory/main.py mem0/client/main.py tests/test_main.py tests/test_client.py`
- `git diff --check`

### DIFF
--- a/mem0/client/main.py
+++ b/mem0/client/main.py
@@ -33,6 +33,15 @@ setup_config()
 ENTITY_PARAMS = frozenset({"user_id", "agent_id", "app_id", "run_id"})
 
 
+def _validate_and_trim_search_query(query: str) -> str:
+    if not isinstance(query, str):
+        raise ValueError("Invalid query: must be a non-empty string.")
+    trimmed = query.strip()
+    if not trimmed:
+        raise ValueError("Invalid query: cannot be empty or whitespace-only.")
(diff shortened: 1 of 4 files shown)


kartik-mem0:
LGTM. This is the strongest fix in this batch — it covers both `Memory` (OSS) and `MemoryClient` (hosted), raises a proper `ValueError` so callers can distinguish bad input from a legitimately empty result set, and trims the query before it reaches the embedder. The parametrized tests confirm the short-circuit fires and that `embed` is never called. Good work.

kartik-mem0 (approved):

PR #5262, opened 2026-05-26
Fix PGVector search score direction

yang-nv:
## Summary
- return PGVector cosine similarity as score instead of cosine distance
- order PGVector search results by descending score
- update PGVector tests to assert the score expression and ordering

## Why
The pgvector cosine operator <=> returns distance, where lower is better. Mem0 search scoring treats scores as higher-is-better similarities, so returning distance can invert ranking and threshold behavior.

## Tests
- uv run --python 3.12 --extra test --with psycopg --with psycopg-pool --with psycopg2-binary pytest tests/vector_stores/test_pgvector.py

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -250,10 +250,10 @@ class PGVector(VectorStoreBase):
         with self._get_cursor() as cur:
             cur.execute(
                 sql.SQL("""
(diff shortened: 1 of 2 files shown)


yang-nv:
Closing this as a duplicate of #5140. That PR fixes the same PGVector score direction issue while preserving the pgvector nearest-neighbor ORDER BY distance query shape.

PR #5270, opened 2026-05-27
feat(sdk): add delete_linked option to MemoryClient.delete (Python + TS)

chaithanyak42:
## Summary

Adds an opt-in `delete_linked` option to the hosted **MemoryClient** `delete()` in both the Python and TypeScript SDKs. It plumbs the platform **v3** `delete_linked` flag through to `DELETE /v1/memories/{id}/`.

When `true`, deleting a memory also removes the older memories it superseded (the `linked_memory_ids` chain), transitively — the delete-side counterpart of `latest_only`. Without it, deleting the *current* memory un-hides the older one it replaced, so it resurfaces under `latest_only`.

Default `false` keeps `delete()` **byte-identical** — the param is omitted from the request entirely.

> Client-side passthrough only. No changes to the OSS memory engine — just the platform API client.

## Python (`mem0/client/main.py`)

```python
client.delete(memory_id, delete_linked=True)   # sync
await async_client.delete(memory_id, delete_linked=True)  # async
```

Forwarded as a query param via `_prepare_params({"delete_linked": delete_linked or None})`, so `False` is stripped and the default call is unchanged.

## TypeScript (`mem0-ts/src/client/`)

```ts
await client.delete(memoryId, { deleteLinked: true });
```

`deleteLinked` is serialized to `delete_linked` through the existing `camelToSnakeKeys` + `URLSearchParams` path — the same mechanism `latestOnly` uses. With no options, the URL is unchanged.

## Tests

- **Python** (`tests/test_client.py::TestDeleteLinked`): default omits `delete_linked`; `delete_linked=True` forwards it; `delete_linked=False` is stripped. (3 pass)
- **TypeScript** (`memoryClient.crud.test.ts`): default delete has no `delete_linked` in the URL; `{ deleteLinked: true }` adds `delete_linked=true`. (crud suite 17/17 pass)

## Related

Platform-side feature (the actual cascade) is implemented separately on the platform repo. This PR is just the SDK surface so customers can call it directly once it ships.

### DIFF
--- a/mem0-ts/src/client/mem0.types.ts
+++ b/mem0-ts/src/client/mem0.types.ts
@@ -40,10 +40,11 @@ export interface GetAllMemoryOptions {
 export interface DeleteAllMemoryOptions extends EntityOptions {}
 
 export interface DeleteMemoryOptions {
-  // When true, also delete the older memories this one superseded
-  // (the v3 linked_memory_ids chain), transitively. The delete-side
-  // counterpart of latestOnly — stops a superseded memory from resurfacing
-  // after the current one is deleted. Serialized as `delete_linked`.
+  /**
+   * When `true`, also delete the older memories this one superseded (the v3
+   * linked chain), transitively — the delete-side counterpart of `latestOnly`.
+   * Off by default. Serialized as `delete_linked`.
+   */
   deleteLinked?: boolean;
 }
 
--- a/mem0-ts/src/client/mem0.types.ts
+++ b/mem0-ts/src/client/mem0.types.ts
@@ -39,6 +39,14 @@ export interface GetAllMemoryOptions {
 
 export interface DeleteAllMemoryOptions extends EntityOptions {}
 
+export interface DeleteMemoryOptions {
+  // When true, also delete the older memories this one superseded
+  // (the v3 linked_memory_ids chain), transitively. The delete-side
+  // counterpart of latestOnly — stops a superseded memory from resurfacing
+  // after the current one is deleted. Serialized as `delete_linked`.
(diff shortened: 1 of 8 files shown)


mem0-ts/src/client/mem0.types.ts
```diff
@@ -39,6 +39,14 @@ export interface GetAllMemoryOptions {
 
 export interface DeleteAllMemoryOptions extends EntityOptions {}
 
+export interface DeleteMemoryOptions {
+  // When true, also delete the older memories this one superseded
+  // (the v3 linked_memory_ids chain), transitively. The delete-side
+  // counterpart of latestOnly — stops a superseded memory from resurfacing
+  // after the current one is deleted. Serialized as `delete_linked`.
```
whysosaket: Please don't leave verbose comments like this unless it's a JSDoc comment

chaithanyak42: Good call — converted it to a concise JSDoc comment (matching the `decay` field style in this file) in a54e9170. Thanks!

chaithanyak42:
Added the patch release bump so this is publishable once merged:
- `mem0ai` (Python): **2.0.3 → 2.0.4**
- `mem0ai` (TypeScript): **3.0.4 → 3.0.5**
- Changelog entries added under both tabs in `docs/changelog/sdk.mdx`

`mem0/__init__.py` reads the version from package metadata, so no other bump needed. Ready to merge — tagging `v2.0.4` / `ts-v3.0.5` will cut the PyPI/npm releases. @whysosaket

whysosaket (approved):

PR #5276, opened 2026-05-27
fix(openmemory): handle postgres engine args and qdrant host defaults

### LINKED ISSUE #5275: OpenMemory self-hosted: PostgreSQL + Qdrant + OpenRouter setup hits stacked regressions (check_same_thread, mem0_store, user bootstrap, null writes)
## Summary
OpenMemory self-hosted on Synology / Docker Compose hits multiple regressions in sequence when configured for PostgreSQL + Qdrant + OpenRouter:

1. `check_same_thread` is still passed to PostgreSQL
2. default vector-store host is hardcoded to `mem0_store` instead of the actual compose service name
3. after fixing those, memory creation can still return `null` / no persisted rows
4. free OpenRouter models may return 404 / 429 and the failure mode is hard to distinguish from infra issues

This issue is intended as a consolidated repro because several of these failures stack on top of each other and are difficult to isolate during first-time setup.

## Environment
- OpenMemory / mem0 self-hosted via `mem0/openmemory-mcp:latest`
- Synology NAS / Container Manager
- Qdrant in Docker Compose
- PostgreSQL 16 in Docker Compose
- OpenRouter for LLM + embedder
- Hermes Agent using OpenMemory as external memory provider

### Compose shape
- `openmemory-api`
- `openmemory-postgres`
- `qdrant`
- `openmemory-ui`

`DATABASE_URL` configured as:
```yaml
DATABASE_URL=postgresql+psycopg2://openmemory:<password>@openmemory-postgres:5432/openmemory
```

Qdrant service name:
```yaml
qdrant:
  image: qdrant/qdrant:latest
```

OpenRouter config in `openmemory-api`:
```yaml
OPENAI_BASE_URL=https://openrouter.ai/api/v1
OPENAI_API_KEY=<valid key>
LLM_MODEL=<free model>
QDRANT_HOST=qdrant
QDRANT_PORT=6333
```

### 1) PostgreSQL crash due to SQLite-only `check_same_thread`
The app still calls SQLAlchemy with:
```python
connect_args={"check_same_thread": False}
```
inside:
- `/usr/src/openmemory/app/database.py`

This causes:
```text
psycopg2.ProgrammingError: invalid dsn: invalid connection option "check_same_thread"
```

This matches / is likely related to:
- #4749
- #4221

### 2) Vector store host default still points to `mem0_store`
In the running container, `/usr/src/openmemory/app/utils/memory.py` default config contained:
```python
"vector_store": {
    "provider": "qdrant",
    "config": {
        "collection_name": "openmemory",
        "host": "mem0_store",
        "port": 6333,
    }
}
```

In our compose file the service name is `qdrant`, not `mem0_store`, so memory client init fails with:
```text
Warning: Failed to initialize memory client: [Errno -2] Name or service not known
```

DNS/network were verified healthy from inside the container:
- `qdrant` resolves
- `openmemory-postgres` resolves
- `openrouter.ai` resolves
- `http://qdrant:6333/collections` returns 200
- `https://openrouter.ai/api/v1/models` returns 200

So the failure was specifically the hardcoded `mem0_store` hostname.

### 3) `User not found` until manual DB insert
POSTing a memory for `user_id=hermes` initially returned:
```json
{"detail":"User not found"}
```
We had to manually create the user in Postgres for the API to get past this stage.

### 4) Free OpenRouter model failure modes are confusing
We observed:
- model not available / 404 for some `:free` models
- 429 rate limits for others

Examples:
```text
No endpoints found for meta-llama/llama-3.1-8b-instruct:free
```
```text
qwen/qwen3-coder:free is temporarily rate-limited upstream
```
These are valid provider-side failures, but while debugging they are hard to distinguish from OpenMemory infra bugs because OpenMemory only exposes `Memory client is not available` in higher layers.

### 5) Memory creation returns `null` / no persistence after stack is repaired
After patching the above issues and getting OpenMemory API healthy:

`POST /api/v1/memories/` returned:
```json
null
```

or sometimes returned 200 with no useful payload, but **no rows were inserted into `memories`**.

Verified directly in PostgreSQL:
```sql
SELECT count(*) FROM memories;
```
returned `0`.

At the same time:
- `/openapi.json` worked
- `/api/v1/config/` worked
- `/api/v1/apps/` worked
- user existed in DB
- app existed in DB
- API process was up

## Why this is painful in practice
The setup fails in layers:
1. DB driver incompatibility (`check_same_thread`)
2. wrong default Qdrant hostname (`mem0_store`)
3. user bootstrap mismatch (`User not found`)
4. model/provider failures surfaced as generic memory-client unavailability
5. eventual `null` write / empty DB state

This makes first-time self-hosted setup look like a networking problem, then a Postgres problem, then an OpenRouter problem, then an application bug.

## Expected behavior
For a standard Docker Compose deployment with service names:
- `openmemory-api`
- `openmemory-postgres`
- `qdrant`

I would expect:
1. PostgreSQL to work without patching source files
2. default vector store host to respect env/config or at least not hardcode `mem0_store`
3. user bootstrap to be handled automatically or give a clearer API path for creating users
4. model/provider failures to be surfaced more explicitly
5. successful `POST /api/v1/memories/` to either create a memory row or return a structured error instead of `null`

## Confirmed local workarounds
We had to patch the container locally to get past multiple stages:
- patch `database.py` to remove `check_same_thread`
- patch `app/utils/memory.py` to replace `mem0_store` with `qdrant`
- manually create the user in Postgres
- swap multiple OpenRouter free models until one stopped returning 404/429

## Suggested fixes
- make `check_same_thread` conditional on SQLite only
- remove hardcoded `mem0_store` and honor runtime config / env consistently
- ensure user bootstrap path is deterministic for API usage
- improve error reporting from `Memory.from_config(...)` init failures
- investigate why `POST /api/v1/memories/` can return `null` while persisting nothing

## Related issues / likely overlaps
- #4749
- #4221
- #2917
- #3256
- #3602

TylonHH:
## Summary
- make  conditional on SQLite only in 
- default OpenMemory qdrant host to  or  instead of hardcoded 
- add regression tests for both behaviors

## Why
This addresses the reproducible self-hosted issues documented in #5275:
- PostgreSQL init failure due to SQLite-only 
- memory client init failure from hardcoded  hostname in standard Docker Compose deployments

## Test Plan
- created targeted tests in 
- ran:
  - ....                                                                     [100%]
4 passed in 0.74s
  - result: 

Closes #5275

### DIFF
--- a/openmemory/api/app/database.py
+++ b/openmemory/api/app/database.py
@@ -12,10 +12,14 @@ if not DATABASE_URL:
     raise RuntimeError("DATABASE_URL is not set in environment")
 
 # SQLAlchemy engine & session
-engine = create_engine(
-    DATABASE_URL,
-    connect_args={"check_same_thread": False}  # Needed for SQLite
-)
+# `check_same_thread` is only valid for SQLite. Passing it through to
+# PostgreSQL / psycopg2 raises:
+#   invalid dsn: invalid connection option "check_same_thread"
+engine_kwargs = {}
+if DATABASE_URL.startswith("sqlite"):
+    engine_kwargs["connect_args"] = {"check_same_thread": False}
+
+engine = create_engine(DATABASE_URL, **engine_kwargs)
 SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
 
 # Base class for models
--- a/openmemory/api/app/utils/memory.py
+++ b/openmemory/api/app/utils/memory.py
@@ -227,9 +227,14 @@ def _create_embedder_config(provider, model, api_key, base_url, ollama_base_url,
 def get_default_memory_config():
     """Get default memory client configuration with sensible defaults."""
     # Detect vector store based on environment variables
+    # Default to the common compose service name used by the self-hosted
+    # OpenMemory stack. The previous hardcoded default (`mem0_store`) causes
+    # memory client init to fail with `Name or service not known` in the
+    # standard qdrant-based docker-compose deployment unless users patch the
+    # container manually.
     vector_store_config = {
         "collection_name": "openmemory",
-        "host": "mem0_store",
+        "host": os.environ.get('QDRANT_HOST', 'qdrant'),
     }
     
     # Check for different vector store configurations based on environment variables
--- a/openmemory/api/tests/test_config_defaults.py
+++ b/openmemory/api/tests/test_config_defaults.py
@@ -1,0 +1,65 @@
+import importlib
+import sys
+
+import pytest
+
+
+pytest.importorskip("openai")
+
+
+def test_sqlite_database_uses_check_same_thread(monkeypatch):
+    monkeypatch.setenv("OPENAI_API_KEY", "test-key")
+    monkeypatch.setenv("DATABASE_URL", "sqlite:///./openmemory.db")
+    sys.modules.pop("app.database", None)
+
+    import app.database as database  # noqa: WPS433
+
+    database = importlib.reload(database)
+    assert database.engine.url.drivername == "sqlite"
+
+
+def test_postgres_database_does_not_use_sqlite_connect_args(monkeypatch):
+    monkeypatch.setenv("OPENAI_API_KEY", "test-key")
+    monkeypatch.setenv(
+        "DATABASE_URL",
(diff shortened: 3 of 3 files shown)


kartik-mem0:

PR #5279, opened 2026-05-28
fix(openmemory): align create-memory bootstrap and empty results

TylonHH:
## Bug Description

`POST /api/v1/memories/` handled OpenMemory create requests inconsistently with the MCP path:

- REST returned `404 User not found` for a new `user_id`, while MCP creates the user/app pair via `get_user_and_app()`.
- If `memory_client.add()` returned `{"results": []}`, the endpoint could fall through without returning a response, producing a silent `null`/empty success instead of an explicit error.

Related to #5275.

## Root Cause

The REST create-memory route manually queried `User` and only created the `App`, instead of using the shared OpenMemory user/app bootstrap helper used by MCP.

The response handling also only returned when at least one `ADD` result produced a persisted memory. Empty or unexpected backend responses had no explicit API-level outcome.

## Fix

- Use `get_user_and_app()` in the REST create-memory route so missing users/apps are created consistently with MCP.
- Return `422` when extraction succeeds structurally but produces no persisted memories.
- Return `502` when the backend response shape is unexpected.
- Keep existing error handling for paused apps and backend exceptions.
- Add regression tests for:
  - auto-creating missing user/app on REST create
  - empty backend results no longer returning a silent null response

## How to Verify

1. Run the OpenMemory API tests:
   ```bash
   .venv/bin/python -m pytest openmemory/api/tests/test_memories_api.py -q
   ```
2. POST to `/api/v1/memories/` with a new `user_id` and valid `memory_client.add()` response.
3. Confirm the user/app are created and the memory is returned.
4. Mock `memory_client.add()` to return `{"results": []}` and confirm the endpoint returns `422`.

## Test Plan

- [x] Added regression test for REST user/app auto-create
- [x] Added regression test for empty `results` handling
- [x] Ran `openmemory/api/tests/test_memories_api.py`

Result: `2 passed`

## Risk Assessment

Low — scoped to the OpenMemory REST create-memory route. The change aligns REST behavior with the existing MCP bootstrap path and makes previously ambiguous backend responses explicit.

### DIFF
--- a/openmemory/api/app/routers/memories.py
+++ b/openmemory/api/app/routers/memories.py
@@ -15,6 +15,7 @@ from app.models import (
     User,
 )
 from app.schemas import MemoryResponse
+from app.utils.db import get_user_and_app
 from app.utils.memory import get_memory_client
 from app.utils.permissions import check_memory_access_permissions
 from fastapi import APIRouter, Depends, HTTPException, Query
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #5295, opened 2026-05-29
fix(pgvector): correct connection-string and config validation handling

yaozcoderepo:
## Linked Issue

No existing issue. Both bugs are described in full below so they can be triaged from this PR. They are closely related (PostgreSQL connection configuration in the `pgvector` store), hence bundled together.

### 1. Invalid connection string when `sslmode` is set

In the individual-parameters path, `PGVector.__init__` always builds a **URI** (`postgresql://user:pass@host:port/dbname`) but appended `sslmode` using the **keyword/value DSN** syntax (space-separated):

```
postgresql://user:pass@host:5432/db sslmode=require
```

libpq detects the `postgresql://` prefix and parses the whole thing as a URI, where parameters must live in the query string (`?sslmode=...`). The space-separated ` sslmode=require` is therefore folded into the database name (dbname becomes `db sslmode=require`), which (1) silently drops the SSL setting and (2) typically fails the connection with `database "db sslmode=require" does not exist`. The same applied when a user passed a URI `connection_string` together with `sslmode`.

**Fix:** added a helper `_append_sslmode(connection_string, sslmode)` that attaches `sslmode` according to the connection-string format:
- **URI** (`postgresql://` / `postgres://`): query string — `?sslmode=...`, or `&sslmode=...` if a query string already exists.
- **keyword/value DSN** (`host=... port=...`): space-separated `sslmode=...`.
- **existing `sslmode=`**: replace the value in place without corrupting other query params.

### 2. Connection-parameter validation used `and` instead of `or`

`PGVectorConfig.check_auth_and_connection` validated with:

```python
if not user and not password:
    raise ValueError("Both 'user' and 'password' must be provided ...")
if not host and not port:
    raise ValueError("Both 'host' and 'port' must be provided ...")
```

The messages state both values are required, but `and` only raises when **both** are missing. Supplying only `host` (or only `user`) passed validation and then produced a broken connection string — e.g. a missing port renders as the literal `None` in `postgresql://user:pass@host:None/db`, failing with a confusing error.

**Fix:** switched both checks to `or` so a half-specified pair is rejected up front. Also corrected the `host`/`port` field descriptions, which advertised defaults that are never applied (and a wrong port value of `1536`).

## Type of Change

- [ ] Documentation update

## Breaking Changes

Configurations that previously passed validation while specifying only one of `user`/`password` or only one of `host`/`port` will now raise at config time instead of failing later with an opaque connection error. This only rejects configs that could never have produced a working connection. Setups that legitimately omit a password (e.g. peer/trust auth) should use `connection_string`.

## Test Coverage

In `tests/vector_stores/test_pgvector.py`:
- Corrected `test_connection_string_with_sslmode_psycopg3`, which previously asserted the buggy space-separated form for a URI.
- Added connection-string regression tests: individual params + sslmode → `?sslmode=`; URI with an existing query param → `&sslmode=`; keyword/value DSN → space-separated; existing `sslmode` replaced without corrupting other params.
- Added `TestPGVectorConfigValidation` covering: full params valid; missing user/password/host/port each raise; `connection_string` and `connection_pool` skip individual-param validation.

All 89 tests in `tests/vector_stores/test_pgvector.py` pass; `ruff check` is clean.

## Checklist

- [x] I have updated documentation if needed (N/A — internal connection handling)

### DIFF
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -9,8 +9,8 @@ class PGVectorConfig(BaseModel):
     embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
     user: Optional[str] = Field(None, description="Database user")
     password: Optional[str] = Field(None, description="Database password")
-    host: Optional[str] = Field(None, description="Database host. Default is localhost")
-    port: Optional[int] = Field(None, description="Database port. Default is 1536")
+    host: Optional[str] = Field(None, description="Database host")
+    port: Optional[int] = Field(None, description="Database port (PostgreSQL default is 5432)")
     diskann: Optional[bool] = Field(False, description="Use diskann for approximate nearest neighbors search")
     hnsw: Optional[bool] = Field(True, description="Use hnsw for faster search")
     minconn: Optional[int] = Field(1, description="Minimum number of connections in the pool")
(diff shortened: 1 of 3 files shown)


kartik-mem0:
Closing as duplicate — the fix for this was merged via #5308. Thank you for the contribution!

PR #5298, opened 2026-05-29
fix(memory): parallelize entity boost searches in AsyncMemory.search

### LINKED ISSUE #5214: Parallelize or batch entity boost searches in AsyncMemory.search
### Component

Core / Python SDK

### Use Case

AsyncMemory.search() can become slow when entity boost runs for multiple extracted entities. Today, _compute_entity_boosts_async() processes up to 8 deduped entities sequentially. Each entity does an embedding call and then an entity-store search, so with a remote embedding provider the latency can become roughly the sum of all entity embedding/search calls.

This can make otherwise normal memory searches take several seconds or hit request timeouts, even though the entity boost work is independent per entity.

### Proposed Solution

Batch or parallelize entity boost computation while preserving the current scoring behavior. This would keep entity boost enabled, but reduce wall time by avoiding serial remote embedding/search work.

### Alternatives Considered

- Disabling entity boost: avoids latency but loses useful ranking signal.
- Lowering the max entity count: helps worst-case latency, but reduces recall for entity-rich queries.
- Caller-side timeout: prevents full request failure, but often returns without entity boosts even when the work could have completed quickly if parallelized.
- External result caching: can help repeated queries, but does not solve first-request latency and is harder to invalidate correctly.

Bartok9:
## Summary

- Run the per-entity embedding + entity-store searches in `AsyncMemory._compute_entity_boosts_async` concurrently instead of sequentially.
- Cuts entity-boost latency from roughly the **sum** of all per-entity calls to roughly the **slowest single** call, with no change to scoring behaviour.

## Motivation

Closes #5214.

`AsyncMemory.search()` can become slow when entity boost runs for multiple extracted entities. `_compute_entity_boosts_async()` processed up to 8 deduped entities in a `for` loop, `await`ing each entity's embedding call and entity-store search before starting the next. With a remote embedding provider, wall time is roughly the sum of every per-entity call, which can push otherwise-normal searches into multi-second territory or request timeouts — even though the per-entity work is fully independent.

## Fix

```python
results = await asyncio.gather(
    *(_search_entity(entity_text) for _, entity_text in deduped),
    return_exceptions=True,
)
```

- Each entity's embed + search is wrapped in a small `_search_entity` coroutine and dispatched via `asyncio.gather`, so the independent remote calls overlap.
- `return_exceptions=True` keeps a single failing entity (provider timeout, throttle, transient 5xx) from aborting the others — the failure is logged at `WARNING` and that entity is skipped, matching the previous best-effort behaviour.
- Scoring is **unchanged**: the same `similarity * ENTITY_BOOST_WEIGHT * memory_count_weight` math and `max()` aggregation run over the gathered results. Because `max()` is order-independent, concurrent completion produces identical boosts to the old sequential order.

## Verification

`python3 -m pytest tests/memory/test_main.py` — 37 passed (3 new + 34 existing).

New tests in `tests/memory/test_main.py::TestAsyncEntityBoostParallelism`:
- `test_boosts_preserve_scoring` — output matches the reference scoring math, including the `max()` aggregation when one memory is linked by two entities.
- `test_one_entity_failure_does_not_abort_others` — a raising entity search is logged and skipped while the surviving entity still contributes a boost.
- `test_searches_run_concurrently` — 4 sleeping searches finish in < 0.6 s (vs ~0.8 s sequential) and observed concurrency peak ≥ 2, proving the calls overlap.

Diff is scoped to 2 files (`mem0/memory/main.py`, `tests/memory/test_main.py`); the sync `_compute_entity_boosts` path is intentionally left untouched.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -2871,16 +2871,33 @@ class AsyncMemory(MemoryBase):
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         memory_boosts = {}
 
+        async def _search_entity(entity_text):
+            # Embed + entity-store search for a single entity. Each entity's work
+            # is independent, so these run concurrently via asyncio.gather below.
+            entity_embedding = await asyncio.to_thread(self.embedding_model.embed, entity_text, "search")
+            return await asyncio.to_thread(
+                self.entity_store.search,
+                query=entity_text,
+                vectors=entity_embedding,
+                top_k=500,
+                filters=search_filters,
+            )
+
         try:
-            for _, entity_text in deduped:
-                entity_embedding = await asyncio.to_thread(self.embedding_model.embed, entity_text, "search")
-                matches = await asyncio.to_thread(
-                    self.entity_store.search,
-                    query=entity_text,
-                    vectors=entity_embedding,
-                    top_k=500,
-                    filters=search_filters,
-                )
+            # Run the per-entity embedding/search calls in parallel instead of
+            # sequentially. With a remote embedding provider this turns the
+            # latency from roughly the sum of all calls into roughly the slowest
+            # single call. return_exceptions=True keeps one failed entity from
+            # aborting the others; failures are logged and skipped.
+            results = await asyncio.gather(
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #5299, opened 2026-05-29
fix(llms/xai): fix factory AttributeError and forward tool calls

### LINKED ISSUE #5189: xai LLM provider: tools silently dropped, AttributeError on factory construction

Bartok9:
## Summary

- Fix three bugs in the XAI (Grok) provider, all in `mem0/llms/xai.py`: an `AttributeError` on every factory construction, silently-dropped `tools`/`tool_choice`, and dropped tool-call payloads.
- Minimal, single-file fix that mirrors the existing OpenAI provider's `_parse_response`.

## Motivation

Closes #5189.

1. **AttributeError on construction** — the constructor read `self.config.xai_base_url` unconditionally, but the factory wires xai with a plain `BaseLlmConfig` (`mem0/utils/factory.py:51`), which has no such attribute. So `Memory.from_config(...)` / `LlmFactory.create("xai", ...)` raised `AttributeError` for everyone, even users who never set a base URL.
2. **Tools silently dropped** — `generate_response()` accepted `tools` and `tool_choice` but never forwarded them to the OpenAI-compatible client, so tool calling did nothing on Grok.
3. **Tool-call payload dropped** — there was no `_parse_response` helper, so `message.content` (which is `None` when the model emits tool calls) was returned raw and the structured tool-call payload was lost.

## Fix

- `base_url = getattr(self.config, "xai_base_url", None) or os.getenv("XAI_API_BASE") or "https://api.x.ai/v1"` — works whether or not the config carries `xai_base_url`.
- Forward `tools`/`tool_choice` to `chat.completions.create` when provided.
- Add the same `_parse_response(response, tools)` the OpenAI provider uses (incl. `extract_json` for tool arguments), returning a `{"content", "tool_calls"}` dict when tools are used and the plain string otherwise.

## Verification

`python3 -m pytest tests/llms/test_xai.py` — 4 passed.

New `tests/llms/test_xai.py`:
- `test_factory_construction_with_base_config_does_not_raise` — `LlmFactory.create("xai", ...)` no longer raises and falls back to `https://api.x.ai/v1`.
- `test_generate_response_without_tools` — plain string path, no `tools` sent.
- `test_generate_response_forwards_tools` — `tools`/`tool_choice` forwarded; tool-call payload parsed into `{"name", "arguments"}`.
- `test_generate_response_with_tools_but_no_tool_calls` — structured dict with empty `tool_calls` when the model returns plain content.

## Differential value

#5190 also closes #5189 but takes a heavier approach: it adds a new `XAIConfig` class (`mem0/configs/llms/xai.py`) and rewires the factory (4 files, +344/-19).

This PR is a **minimal, surgical fix to the single `mem0/llms/xai.py` file** (+42/-3 in source). Differences:
- **Mine:** `getattr` fallback keeps the plain `BaseLlmConfig` wiring intact — no new config class, no factory change, lower regression surface. Fixes all three reported bugs.
- **#5190:** introduces `XAIConfig` (enables a typed `xai_base_url` field and vision options) — broader, but larger surface and changes the factory mapping.
- **Recommend:** merge this if the maintainers prefer the minimal fix that closes #5189 without a new config type; otherwise #5190 if a dedicated `XAIConfig` is desired. The two are not complementary (both rewrite `generate_response`), so only one should land.

### DIFF
--- a/mem0/llms/xai.py
+++ b/mem0/llms/xai.py
@@ -1,3 +1,4 @@
+import json
 import os
 from typing import Dict, List, Optional
 
@@ -5,6 +6,7 @@ from openai import OpenAI
 
 from mem0.configs.llms.base import BaseLlmConfig
 from mem0.llms.base import LLMBase
+from mem0.memory.utils import extract_json
 
 
 class XAILLM(LLMBase):
@@ -15,8 +17,41 @@ class XAILLM(LLMBase):
             self.config.model = "grok-2-latest"
 
         api_key = self.config.api_key or os.getenv("XAI_API_KEY")
-        base_url = self.config.xai_base_url or os.getenv("XAI_API_BASE") or "https://api.x.ai/v1"
+        # The factory wires xai with a plain BaseLlmConfig, which has no
+        # `xai_base_url` attribute. Use getattr so construction works with or
+        # without it instead of raising AttributeError on every call.
+        base_url = getattr(self.config, "xai_base_url", None) or os.getenv("XAI_API_BASE") or "https://api.x.ai/v1"
         self.client = OpenAI(api_key=api_key, base_url=base_url)
+
+    def _parse_response(self, response, tools):
+        """
+        Process the response based on whether tools are used or not.
+
+        Args:
+            response: The raw response from API.
+            tools: The list of tools provided in the request.
+
+        Returns:
+            str or dict: The processed response.
+        """
+        if tools:
+            processed_response = {
+                "content": response.choices[0].message.content,
+                "tool_calls": [],
+            }
+
+            if response.choices[0].message.tool_calls:
+                for tool_call in response.choices[0].message.tool_calls:
+                    processed_response["tool_calls"].append(
+                        {
+                            "name": tool_call.function.name,
+                            "arguments": json.loads(extract_json(tool_call.function.arguments)),
+                        }
+                    )
+
+            return processed_response
+        else:
+            return response.choices[0].message.content
 
     def generate_response(
         self,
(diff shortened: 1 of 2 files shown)


Bartok9:
Differential note for maintainers: #5190 (@Aarkin7, opened 10 days earlier) targets the same three `mem0/llms/xai.py` bugs from #5189.

- **#5190**: heavier — adds a dedicated `XAIConfig` dataclass mirroring `DeepSeekConfig`, plus tests (+344/-19, 4 files).
- **This PR (#5299)**: minimal single-file fix mirroring the existing OpenAI provider's `_parse_response` (+134/-3, 2 files), no new config class.

Both close #5189. If the project prefers a typed `XAIConfig` parity with DeepSeek, #5190 is the more complete choice and this can be closed in its favor. If a minimal in-place fix is preferred, this one is smaller to review. Happy to defer either way — flagging to avoid a double-merge.

kartik-mem0:
Thanks for the contribution! Another PR (#5190) addresses the same issue with a more complete fix — it updates the factory to use `XAIConfig` properly (fixing the root cause instead of working around it with `getattr`), and has broader test coverage. Closing in favor of that one.

Bartok9:
Appreciate you folks closing this in favor of the more complete fix (#5190). That's the right call — team player move on our end too. Thanks for keeping the quality bar high on the XAI integration. Happy to collaborate on any follow-ups. 🙏

Issue #5286, opened 2026-05-28
Suggestion: Standardized memory recall benchmark for Mem0


yubingz:
Hi Mem0 team,

Mem0 has become the de facto memory layer for AI agents — 56K+ stars, production deployments everywhere. One thing I think would significantly strengthen the ecosystem: **standardized, reproducible memory recall benchmarks**.

Currently, most memory system evaluations are ad-hoc — a few hand-crafted queries, qualitative "feels right" assessments, or proprietary internal benchmarks that can't be compared across systems. This makes it hard for users to answer: "How well does Mem0 actually retrieve the right memories?"

I've been working on [MemTest](https://github.com/yubingz/memtest), a benchmark database design system for AI memory evaluation. Instead of providing metrics, it provides **test databases** with controlled ground truth:

- **6 evaluation dimensions**: storage integrity, retrieval precision (5 query types), clustering, forgetting directionality, reasoning (multi-hop), deep retrieval decay
- **Two data generators**: procedural (100-10K memories) and corpus-driven (feed any text corpus, auto-extract facts + generate queries)
- **Four Great Classical Novels dataset**: 21,793 memories + 750 queries from ~1.2M characters of Chinese literature

For Mem0 specifically, a standardized benchmark could help:
- Quantify recall quality across different vector stores (FAISS vs Qdrant vs OpenSearch)
- Measure how well graph-based retrieval (Neptune Analytics) improves over pure vector search
- Track regression as new features are added

We found some surprising results on Chinese text: sentence-transformers (MiniLM) only achieves 9.1% precision on entity-level retrieval, while TF-IDF + LLM reranking hits 87%. This suggests embedding choice matters enormously for different languages/domains.

Would adding a standardized eval suite be something the team is interested in? Happy to help draft a Mem0 adapter for MemTest or contribute benchmark data.

ususar1234:
@yubingz There has been one that's already made it into the literature: https://arxiv.org/abs/2605.11325v2 
First Precision-Aware Benchmark for LLM Memory Retrieval

yubingz:
Thanks for sharing this — the PrecisionMemBench paper raises an important and complementary point.

The key insight in [Flynt 2026](https://arxiv.org/abs/2605.11325v2) is that current benchmarks measure answer quality through a generative model, which hides retrieval imprecision: a system returning its entire belief store can score well because the downstream model compensates for noise. Their 89-case benchmark addresses this by evaluating retrieval precision in isolation.

[...]

PR #5307, opened 2026-05-29
Add deterministic memory scenario evals

### LINKED ISSUE #5235: Feature: evaluation toolkit for agent memory quality and retrieval reliability
## Background

I'm using Mem0 as the memory layer for task-based agents. It works well for storing and retrieving user context, but I've been running into subtle failures that are hard to debug: the agent *should* remember a preference but silently retrieves stale or conflicting information instead.

These failures are hard to catch without a structured way to measure memory behavior over time.

## Problem

Typical RAG evaluation metrics (context relevance, answer correctness) don't fully apply to memory systems because memory has additional dimensions:

1. **Forgetting** — when should old memories be deprioritized?
2. **Overwrite behavior** — does updating a memory correctly supersede the old one?
3. **Temporal relevance** — are newer facts retrieved over stale ones?
4. **Conflict resolution** — when two memories contradict, which wins?

None of these are captured by standard precision/recall on retrieval.

## Proposed Evaluation Toolkit

A small `mem0.eval` module (or standalone helper) with a scenario-based testing pattern:

```python
from mem0 import MemoryClient
from mem0.eval import evaluate_memory, MemoryScenario

client = MemoryClient(api_key='...')

scenario = MemoryScenario(
    user_id='user_123',
    events=[
        ('add', 'User prefers emails in the morning.'),
        ('update', 'User now prefers emails in the evening.'),
        ('query', 'When should I send an email to this user?'),
    ],
    expected_retrieval='User prefers emails in the evening.',
)

result = evaluate_memory(client, scenario)

# result.memory_recall_rate         -> 1.0   (correct memory retrieved)
# result.staleness_score            -> 0.0   (no stale data returned)
# result.conflict_resolution_acc    -> 1.0   (newer fact wins)
# result.update_propagation_rate    -> 1.0   (old memory superseded)
```

## Why This Matters

For production agents, memory is one of the most common sources of subtle, hard-to-reproduce failures. Without a structured eval framework:
- There's no way to benchmark different embedding models or retrieval strategies for memory use cases
- Regressions in memory behavior go undetected until they affect user experience
- It's hard to tune retention policies, update strategies, or TTL settings quantitatively

Having a lightweight eval toolkit specifically designed for memory scenarios would make Mem0 much more production-ready for agent deployments.

Happy to contribute a basic scenario runner as a starting point if the team is open to it.

RitwijParmar:
## Summary
- add `mem0.evals` with a scenario runner for add/update/query memory behavior
- add deterministic retrieval scoring for recall, staleness, conflict ordering, and update propagation
- normalize hosted/OSS search response shapes so the evaluator can run against `Memory`, `MemoryClient`, or a test double
- add API-free regression tests covering stale memories outranking updated facts and successful update propagation

Closes #5235. Helps address #5286 by adding a reusable benchmark primitive for standardized memory recall scenarios.

## Why this matters
Agent memory failures are rarely just vanilla RAG relevance failures. A memory system can retrieve something semantically related while still being wrong because it surfaced a stale preference, failed to propagate an update, or ranked a contradicted fact ahead of the current one. This gives Mem0 users a small deterministic harness for those behavioral checks without requiring hosted Mem0 or LLM judge calls.

## Tests
- `.venv/bin/python -m ruff check mem0/evals tests/evals/test_memory_eval.py`
- `.venv/bin/python -m ruff format --check mem0/evals tests/evals/test_memory_eval.py`
- `.venv/bin/python -m pytest tests/evals/test_memory_eval.py tests/utils/test_scoring.py`


RitwijParmar:
Updated this PR with a larger pass after the discussion in #5235.

The main change is that the scenario runner now records a write/read trace instead of only scoring the final search response. It tracks original vs current memory IDs, so additive update behavior can show both: the new fact that should win and the old fact that should be treated as stale.

I also added a JSON-serializable report, failure-mode labels, docs, and tests for rewritten memories where text matching alone is not enough. Local checks passing:

- `.venv/bin/python -m pytest tests/evals/test_memory_eval.py tests/utils/test_scoring.py -q`
- `.venv/bin/python -m ruff check mem0/evals tests/evals/test_memory_eval.py`
- `.venv/bin/python -m ruff format --check mem0/evals tests/evals/test_memory_eval.py`

The remaining Vercel status still looks like the repo authorization check rather than something caused by these Python changes.

Ardem2025:
Evaluating memory systems for staleness and conflict resolution is a crucial step forward. When designing these deterministic scenarios, a few architectural edge cases are worth covering:

1. **Semantic Toggles**: Test scenarios where polar opposites (e.g., "prefers dark mode" vs "switched to light mode") share high cosine similarity but must trigger an update/invalidation rather than a merge.
2. **Logical vs. Physical Deletion**: Evaluate how the system handles superseded facts. Testing if soft-deleted or updated memories are correctly filtered out during retrieval prevents stale facts from outranking new ones.
3. **Temporal Decay**: Introduce test cases that score retrieval based on a combination of vector similarity and metadata-based recency, ensuring older preferences naturally decay.

For a reference implementation of handling these temporal conflicts and soft-deletes in Qdrant, you might find the approach in `mem0-temporal-hygiene` (GitHub: Ardem2025/mem0-temporal-hygiene) helpful.

kartik-mem0:
Thanks for putting this together, @RitwijParmar — and for signing the CLA.

After some discussion, we're keeping Mem0's standardized evaluation and benchmarking consolidated in the dedicated [memory-benchmarks](https://github.com/mem0ai/memory-benchmarks) repo (wired into this repo as the `evaluation/` submodule) rather than adding a separate eval harness inside the core SDK. So we're going to pass on bringing `mem0/evals/` into the SDK here, and I'll close this PR.

If you'd like to contribute eval or benchmark work, the memory-benchmarks repo is the right home for it and a PR there would be very welcome. Really appreciate the effort you put into this.

PR #5308, opened 2026-05-30
Fix pgvector sslmode handling for PostgreSQL URIs

### LINKED ISSUE #5303: bug(pgvector): Invalid connection URI structure for sslmode
### Component

### Summary

When using `pgvector` vector store with PostgreSQL connection URIs (e.g. `postgresql://...`), providing the `sslmode` option causes the connection to fail. Happens because `sslmode` is added to connection string with space (e.g. `postgresql://user:pass@host:port/dbname sslmode=require`) instead of standard URL query delimiters (`?` or `&`).
As a result, drivers like `psycopg3` and cloud routing endpoints parse the trailing space and suffix as part of database name itself (e.g. `'dbname sslmode=require'`), which causes connection errors.

---

### Steps to Reproduce

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "pgvector",
        "config": {
            "host": "db.prisma.io",
            "port": 5432,
            "dbname": "postgres",
            "user": "your_user",
            "password": "your_password",
            "sslmode": "require"  # main issue
        }
    }
}

# Trying to perform any operation triggers a connection error due to invalid database name
m.add("Hello world", user_id="test_user")
```

──────

### Expected Behavior

The PGVector connection manager should construct a valid connection URI string using standard URL query parameter syntax:
  `postgresql://your_user:your_password@db.prisma.io:5432/postgres?sslmode=require`

──────

### Actual Behavior

The URI is concatenated with a space:
  `postgresql://your_user:your_password@db.prisma.io:5432/postgres sslmode=require`

Which triggers a connection traceback:
```
  Multiple connection attempts failed. All failures were:
  - host: 'db.host.com', port: '5432', hostaddr: 'XX.XX.XXX.XXX': connection failed: connection to server at "XX.XX.XXX.XXX", port 5432 failed: Failed to select your database ('postgres sslmode=require'): user error: invalid
database name
  connection to server at "45.76.147.142", port 5432 failed: server closed the connection unexpectedly
```

──────

### Environment

• mem0 version: v2.0.4
• Python version: 3.12+ (using psycopg3 / psycopg2)
• OS: macOS / Linux (Docker Environment)

SSDWGG:
## Summary

Fixes #5303.

`PGVector` previously appended `sslmode` with PostgreSQL keyword syntax even when the connection string was a PostgreSQL URI. That produced strings such as `postgresql://user:pass@host:5432/db sslmode=require`, which URI-based drivers can parse as a database name containing ` sslmode=require`.

This PR adds a small helper that:

- encodes `sslmode` as a URI query parameter for URI connection strings
- replaces an existing URI `sslmode` instead of duplicating it
- preserves PostgreSQL keyword conninfo syntax for non-URI connection strings

## Testing

- `python -m pytest tests/vector_stores/test_pgvector.py -q`
- `python -m ruff check mem0/vector_stores/pgvector.py tests/vector_stores/test_pgvector.py`

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -1,7 +1,9 @@
 import json
 import logging
+import re
 from contextlib import contextmanager
 from typing import Any, List, Optional
+from urllib.parse import parse_qsl, urlencode, urlsplit
 
 from pydantic import BaseModel
 
@@ -113,6 +115,19 @@ def _build_filter_conditions(filters):
     return conditions, params
 
 
+def _with_sslmode(connection_string: str, sslmode: str) -> str:
+    if "://" in connection_string:
+        parsed = urlsplit(connection_string)
+        query = [(key, value) for key, value in parse_qsl(parsed.query, keep_blank_values=True) if key != "sslmode"]
+        query.append(("sslmode", sslmode))
+        return parsed._replace(query=urlencode(query)).geturl()
+
+    if re.search(r"(^|\s)sslmode=", connection_string):
+        return re.sub(r"(^|\s)sslmode=\S+", lambda match: f"{match.group(1)}sslmode={sslmode}", connection_string)
+
+    return f"{connection_string} sslmode={sslmode}"
+
+
 class OutputData(BaseModel):
     id: Optional[str]
     score: Optional[float]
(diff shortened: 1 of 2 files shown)


biswajeetdev (commented):
The old regex `r'sslmode=[^ ]*'` had two issues that this PR fixes: it was wrong for URL-format connection strings (where params are separated by `&`, not spaces, so it would match `sslmode=require&other_param` as a single token), and the `import re` inside the `if` block was non-standard. The new `_with_sslmode` helper correctly branches on connection string format and uses proper URL parsing for the URI case.

One edge case in the DSN regex path: `r"(^|\s)sslmode="` correctly anchors to word boundaries, but it would also match `sslmode=` inside a quoted DSN value like `options='--extra sslmode=require'`. This is an unlikely input but possible if the connection string includes `options` with embedded key-value pairs. Adding a note in the docstring about this assumption (DSN values do not contain nested `sslmode=` substrings) would make the limitation explicit.

The test suite now covers both URI and DSN formats, including the replace-existing-value path — good coverage.

SSDWGG:
Thanks for the review. I've added the docstring note about the keyword conninfo assumption for nested sslmode substrings. Focused verification passed locally: `python -m pytest tests/vector_stores/test_pgvector.py -k with_sslmode` (4 passed).

kartik-mem0 (approved):

PR #5309, opened 2026-05-30
fix(oss): resolve vector store dimension from snake_case config keys

### LINKED ISSUE #5290: Recall search failed: Bad Request Using OpenAI Embedding Model
### Component

OpenClaw

### Summary
use openclaw.json config:

```json
...
"embedder": {
              "provider": "openai",
              "config": {
                "model": "bge-base-zh-v1.5",
                "embedding_dims": 1024,
                "embeddingDims": 1024,
                "url": "https://xxxxxxxxx/v1",
                "apiKey": "xxxxxxxxxxxx"
              }
            },
"vectorStore": {
              "provider": "qdrant",
              "config": {
                "url": "http://qdrant:6333",
                "apiKey": "${QDRANT_API_KEY}",
                "collectionName": "mem0",
                "embeddingModelDims": 1024
              }
            }
```
```

openclaw log info is：

```
23:14:20 Api key is used with unsecure connection.
23:14:21 [mem0] Recall search failed: Bad Request
23:14:21 [plugins] openclaw-mem0: skills-mode recall (strategy=smart) injecting 0 memories (~20 tokens)
23:14:22 [ws] ⇄ res ✓ sessions.list 256ms conn=d1eb9bc4…17da id=201b8113…c9dc
23:14:22 [ws] ⇄ res ✓ sessions.list 264ms conn=d1eb9bc4…17da id=4939f962…2f16
23:14:34 [ws] ⇄ res ✓ sessions.list 250ms conn=d1eb9bc4…17da id=f7ad503f…baa6
23:15:12 [mem0] **Recall search failed: Bad Request**
23:15:12 [plugins] openclaw-mem0: skills-mode recall (strategy=smart) injecting 0 memories (~20 tokens)
23:15:12 [ws] ⇄ res ✓ sessions.list 288ms conn=d1eb9bc4…17da id=9e20bb86…371e
23:15:13 [ws] ⇄ res ✓ sessions.list 268ms conn=d1eb9bc4…17da id=3b49a2ad…7ada
23:15:20 [ws] ⇄ res ✓ sessions.list 235ms conn=d1eb9bc4…17da id=a192da30…069f
```

### Actual Behavior

embedding model response ok,response message has 1024 vectors,but the vectors are submitted to vector-db:qdrant with all zero vectors,and vectors has only 256 size.

```http
POST /collections/mem0/points/search HTTP/1.1
host: qdrant:6333
connection: keep-alive
user-agent: qdrant-js/1.13.0
api-key: xxxxxxxxxxxxxxxxxxxxxxxxxxxx
Content-Type: application/json
Accept: application/json
accept-language: *
sec-fetch-mode: cors
accept-encoding: gzip, deflate
content-length: 651

{"vector":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"limit":120,"offset":0,"filter":{"must":[{"key":"user_id","match":{"value":"agent"}}]},"with_payload":true,"with_vector":false}

**HTTP/1.1 400 Bad Request**
transfer-encoding: chunked
content-type: application/json
vary: accept-encoding, Origin, Access-Control-Request-Method, Access-Control-Request-Headers
content-encoding: gzip

```

### Expected Behavior

embedding model response ok by tcpdump, response message has 1024 vectors,and this vectors are submitted to vector-db:qdrant with the same vectors,and vectors has also 1024 size.

### Environment

- openclaw-mem0 version: 1.0.11
- qdrant: 1.13.6

Bartok9:
## Summary

- `ConfigManager.mergeConfig` now resolves the vector store `dimension` from snake_case config keys, not just camelCase.
- A user who configures only `embedding_dims` (snake_case, the Python SDK / OpenClaw convention) no longer silently loses their explicit dimension.

## Motivation

Closes #5290.

The embedder config block in `mergeConfig` already normalizes snake_case `embedding_dims` → `embeddingDims` (see the existing `normalizes embedding_dims to embeddingDims for embedder` test). But two lines later, the vector store dimension fallback read **only** camelCase:

```ts
const explicitDimension =
  userConf?.dimension ||
  userConfig.embedder?.config?.embeddingDims ||
  undefined;
```

So a snake_case-only config produced a correctly-dimensioned **embedder** but an `undefined` **vector store dimension**. The explicitly configured dimension was dropped and a startup probe ran instead — an internal inconsistency (snake_case honored in one place, ignored in the next) that contributes to the dimension-mismatch class of failures against stores like Qdrant.

## Fix

The fallback now also reads:
- snake_case `embedding_dims` from the embedder config, and
- `embeddingModelDims` / `embedding_model_dims` from the vector store config,

mirroring the embedder normalization already present in the same function. Switched `||` → `??` so an explicit value always wins and the precedence is unchanged for all existing (truthy) inputs; dimension is never legitimately `0`, so behavior is identical for every real config.

## Verification

- `npx jest tests/config-manager.test.ts` — **27 passed** (24 existing + 3 new).
- New cases: snake_case `embedding_dims` on embedder → dimension resolved; vectorStore `embeddingModelDims` → resolved; explicit `vectorStore.dimension` still wins over snake_case embedder dims.
- `npx tsc --noEmit` clean for the changed file (the only remaining errors are pre-existing in `src/community/.../langchain/mem0.ts` on `main`, unrelated to this change).
- Diff is 2 files, +62/-2.

## Relationship to #4231

#4231 (open since March) takes a different, broader approach: normalizing snake_case keys to camelCase inside the `Memory` constructor (`memory/index.ts` + `types/index.ts`). This PR is narrower and orthogonal — it makes `ConfigManager.mergeConfig` self-consistent with the snake_case normalization it *already* performs for the embedder, fixing the specific dimension-drop path regardless of whether #4231 lands.

### DIFF
--- a/mem0-ts/src/oss/src/config/manager.ts
+++ b/mem0-ts/src/oss/src/config/manager.ts
@@ -63,9 +63,25 @@ export class ConfigManager {
           // probe embedding at startup — this makes *any* embedder work
           // out of the box without the user needing to know or set the
           // dimension manually.
+          //
+          // Honor snake_case keys (embedding_dims / embeddingModelDims) the
+          // same way the embedder config above does — Python SDK / OpenClaw
+          // configs are snake_case, so reading only camelCase here silently
+          // dropped an explicitly-configured dimension and forced a probe.
+          const embedderConf = userConfig.embedder?.config as
+            | Record<string, unknown>
+            | undefined;
+          const vectorStoreConf = userConf as
+            | Record<string, unknown>
+            | undefined;
           const explicitDimension =
-            userConf?.dimension ||
-            userConfig.embedder?.config?.embeddingDims ||
+            (userConf?.dimension as number | undefined) ??
+            (vectorStoreConf?.embeddingModelDims as number | undefined) ??
+            (vectorStoreConf?.embedding_model_dims as number | undefined) ??
+            (userConfig.embedder?.config?.embeddingDims as
+              | number
+              | undefined) ??
+            (embedderConf?.embedding_dims as number | undefined) ??
             undefined;
 
           // Prioritize user-provided client instance
--- a/mem0-ts/src/oss/tests/config-manager.test.ts
+++ b/mem0-ts/src/oss/tests/config-manager.test.ts
@@ -84,6 +84,52 @@ describe("ConfigManager", () => {
 
       expect(config.vectorStore.config.dimension).toBe(768);
     });
+
+    it("should resolve dimension from snake_case embedding_dims on the embedder (issue #5290)", () => {
+      const config = ConfigManager.mergeConfig({
+        embedder: {
+          provider: "openai",
+          // snake_case as written by Python SDK / OpenClaw configs
+          config: { model: "bge-base-zh-v1.5", embedding_dims: 1024 } as any,
+        },
+        vectorStore: { provider: "qdrant", config: { collectionName: "test" } },
+        llm: baseLlm,
+      });
+
+      // Previously the vectorStore dimension fallback only read camelCase
+      // embeddingDims, so a snake_case-only embedder left the dimension
+      // undefined even though the embedder itself was normalized to 1024.
+      expect(config.vectorStore.config.dimension).toBe(1024);
+    });
+
+    it("should resolve dimension from vectorStore embeddingModelDims (issue #5290)", () => {
+      const config = ConfigManager.mergeConfig({
+        embedder: { provider: "openai", config: { apiKey: "test-key" } },
+        vectorStore: {
+          provider: "qdrant",
+          config: { collectionName: "test", embeddingModelDims: 1024 } as any,
+        },
+        llm: baseLlm,
+      });
+
+      expect(config.vectorStore.config.dimension).toBe(1024);
+    });
+
+    it("should prefer explicit vectorStore dimension over snake_case embedder embedding_dims", () => {
+      const config = ConfigManager.mergeConfig({
+        embedder: {
+          provider: "openai",
+          config: { model: "bge-base-zh-v1.5", embedding_dims: 1024 } as any,
+        },
+        vectorStore: {
+          provider: "qdrant",
+          config: { collectionName: "test", dimension: 512 },
+        },
+        llm: baseLlm,
+      });
+
+      expect(config.vectorStore.config.dimension).toBe(512);
+    });
   });
 
   describe("mergeConfig - LLM url passthrough for Ollama", () => {


biswajeetdev (commented):
The `||` to `??` switch is the correct change here — `||` would skip a dimension value of `0` (falsy), while `??` only skips `null`/`undefined`. In practice `0` is an invalid embedding dimension, but using `??` is still the right semantic.

The snake_case fallback chain is getting long (5 keys now). A future cleanup worth considering: normalise config keys to camelCase once at the point of ingestion (in `ConfigManager.mergeConfig` before this block runs) rather than adding parallel `snake_case` lookups at each consumption site. Something like:

```typescript
function normalizeKeys(obj: Record<string, unknown>): Record<string, unknown> {
  // snake_case → camelCase, one pass at ingestion
}
```

That would collapse the 5-way chain back down to 2-3 canonical keys and make future config additions cheaper. Not a blocker for this fix, but worth a follow-up issue.

Bartok9:
Thanks for the careful review, @biswajeetdev — and for confirming the `||` → `??` semantics. Agreed on both points: `0` is an invalid embedding dimension in practice, but `??` is the correct guard regardless since it only short-circuits on `null`/`undefined`.

You're right that the snake_case fallback chain is getting unwieldy at 5 keys. I like the single-pass normalization at ingestion (`ConfigManager.mergeConfig`) much better than parallel lookups at each consumption site — it collapses the chain to canonical keys and makes future config additions cheap. I'll keep this PR scoped to the targeted fix (so it stays easy to review/backport) and open a follow-up issue for the `normalizeKeys` ingestion refactor referencing this thread. Happy to take that on as the follow-up.

PR #5325, opened 2026-05-31
feat(server): add information on self-hosted dashboard

youneshima:
## Description

Self-hosters often set up the server, grab an API key, and build against the API without ever discovering the bundled web dashboard (some even build their own). This is a discovery gap. Surface the dashboard at the two moments a human is actually looking at the terminal:

- **Setup output** (`seed.sh`): add an explicit "Open your dashboard at … to see and manage your memories" line to the `=== Ready ===` block, instead of listing the URL as a bare credential next to email/password/API key.
- **First memory stored** (`main.py` + `telemetry.py`): log a one-time hint pointing at the dashboard the first time an `add` succeeds. Once per process; local console log only, not gated by `MEM0_TELEMETRY` (it sends nothing off-box).

## Type of Change

## Test Coverage

- [x] Tested manually (no server test suite exists)
  - Verified the seed line appears in the Ready block via `make bootstrap`.
  - Verified the first `/memories` add logs the nudge once; a second add does not re-log.
  - Confirmed `ruff` passes on the changed files.

## Breaking Changes

N/A

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -359,6 +359,8 @@ def add_memory(memory_create: MemoryCreate, _auth=Depends(verify_auth)):
     params = {k: v for k, v in memory_create.model_dump().items() if v is not None and k != "messages"}
     try:
         response = get_memory_instance().add(messages=[m.model_dump() for m in memory_create.messages], **params)
+        if response.get("results"):
+            telemetry.log_dashboard_nudge_once(DASHBOARD_URL)
         return JSONResponse(content=response)
     except Exception:
         raise upstream_error()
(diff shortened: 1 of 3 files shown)


whysosaket:
One thing I noticed on the `seed.sh` change: the description mentions surfacing the dashboard line *instead of* listing the URL as a bare credential, but it looks like the existing `echo "Dashboard:  $DASHBOARD_URL"` (line 67) is still there alongside the new line, so the URL ends up showing twice in the `=== Ready ===` block. Might be worth dropping the old line so it matches the intent? 

Either way this is a clean, low-risk change. 👍

whysosaket (dismissed):

whysosaket (approved):

PR #5327, opened 2026-05-31
fix(llms): add is_reasoning_model config override for versioned deployments

### LINKED ISSUE #5296: gpt-5.x reasoning models on Azure fail with 400 max_tokens rejection — name-based _is_reasoning_model doesn't match dated subversions
## Affects

- `mem0ai==2.0.4`
- Component: `mem0/llms/base.py`, `mem0/llms/azure_openai.py`

## Summary

`Memory.add()` silently returns zero memories when configured with `provider: "azure_openai"` and a `gpt-5.x` model deployment. Root cause: `LLMBase._is_reasoning_model` only matches the exact strings `"gpt-5"` / `"gpt-5o"` (plus `o1-` / `o3-` prefixes), but real Azure deployments are versioned (e.g. `gpt-5.4-nano-2026-03-17`, `gpt-5.4-mini-...`, `gpt-5.2`). When `_is_reasoning_model` returns False, the call goes through `_get_common_params` which sends `max_tokens`, which Azure rejects with `400 Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.`

The 400 is caught in `Memory._add_to_vector_store` and logged at ERROR (`LLM extraction failed`) but doesn't bubble up — so consumers see "no memories created" with no exception, which is confusing.

## Reproduce

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "pgvector",
        "config": {"connection_string": "...", "collection_name": "test", "embedding_model_dims": 1536},
    },
    "llm": {
        "provider": "azure_openai",
        "config": {
            "model": "gpt-5.4-nano-2026-03-17",  # or any gpt-5.x dated variant
            "azure_kwargs": {
                "api_key": "...",
                "azure_deployment": "gpt-5.4-nano-2026-03-17",
                "azure_endpoint": "https://<your>.openai.azure.com/",
                "api_version": "2024-02-15-preview",
            },
        },
    },
    "embedder": {"provider": "azure_openai", "config": {...}},
}
result = m.add(
    [{"role": "user", "content": "I have oily skin."}],
    user_id="test_user",
)
print(result)  # {'results': []} -- empty, with stderr "LLM extraction failed: ... max_tokens ..."
```

## Expected vs actual

**Expected:** `add()` either succeeds (Mem0 sends `max_completion_tokens` for models Azure treats as reasoning) OR raises a clear exception when the LLM call fails.

**Actual:** `add()` returns `{'results': []}` because the 400 is swallowed into an empty extraction.

## Root cause

[`mem0/llms/base.py`](https://github.com/mem0ai/mem0/blob/main/mem0/llms/base.py) (v2.0.4):

```python
def _is_reasoning_model(self, model: str) -> bool:
    reasoning_models = {
    }
    base_model = model.lower().rsplit("/", 1)[-1]
    if base_model in reasoning_models:
        return True
    # Match o1/o3 family with prefixes (o1-2024-12-17, o3-2025-04-16)
    # but NOT gpt-5.x variants (gpt-5.4-mini supports temperature)
        return True
    return False
```

The inline comment "gpt-5.4-mini supports temperature" appears to be incorrect — Azure's deployment of `gpt-5.4-mini-2026-03-17` (and the other `gpt-5.x` dated variants we tested) requires `max_completion_tokens` and rejects `max_tokens`. The exact-string match against `"gpt-5"` doesn't help when the deployment name is `gpt-5.4-nano-2026-03-17` or similar.

Real-world deployment names commonly include version/date suffixes. The Azure model spec doesn't expose a "is this a reasoning model?" capability flag through the SDK, so the only signals are (a) the name pattern, which the user controls and Azure doesn't standardize, or (b) trial-and-error on the 400 response.

## Suggested fix

Smallest patch — extend the prefix match to cover `gpt-5.` and `gpt-5o-`:

```python
# mem0/llms/base.py
if any(base_model.startswith(prefix) for prefix in ["o1-", "o1.", "o3-", "o3.", "gpt-5.", "gpt-5o-"]):
    return True
```

This is what we apply as a runtime monkey-patch in our deployment (see [`genaipa/memory/semantic/_mem0_patches.py`](genaipa/memory/semantic/_mem0_patches.py) for the exact patch we run — happy to inline it in a follow-up PR).

**Better long-term fix:** add an explicit `is_reasoning_model: Optional[bool] = None` field to `BaseLlmConfig` (defaulting to None for "auto-detect") so deployments using custom names can override the heuristic without monkey-patching. Auto-detect for None preserves current behavior.

**Independent improvement:** raise (or surface a structured error) instead of swallowing the LLM 400 in `Memory._add_to_vector_store`. The silent zero-extraction is hard to diagnose.

## Environment

- `mem0ai==2.0.4` (pip)
- Python 3.12
- Azure OpenAI: dated Sephora-internal deployments of GPT-5 family (`gpt-5.4-nano-2026-03-17`, `gpt-5.4-mini-2026-03-17`, `gpt-5.2`, `gpt-5.4-2026-03-05`) accessed via LiteLLM proxy
- Vector store: PGVector

---

Bartok9:
## Summary

- Adds `is_reasoning_model: Optional[bool]` to `BaseLlmConfig` as an explicit override for reasoning-model detection.
- Fixes silent zero-extraction on Azure deployments with versioned `gpt-5.x` names that the name-based heuristic can't recognize.

## Motivation

Closes #5296.

`Memory.add()` silently returns zero memories on Azure when configured with a versioned `gpt-5.x` deployment name (e.g. `gpt-5.4-nano-2026-03-17`). `LLMBase._is_reasoning_model` only matches the exact strings `gpt-5`/`gpt-5o` and the `o1-`/`o3-` prefixes, so dated deployment names fall through to the standard parameter path, send `max_tokens`, and Azure rejects with `400 Unsupported parameter: 'max_tokens' ... Use 'max_completion_tokens' instead`. The 400 is caught and logged but never raised, so the caller just sees `{'results': []}`.

A blanket `gpt-5.` prefix match (the issue's "smallest patch") is **not** viable: this repo deliberately classifies OpenAI `gpt-5.x` as **non-reasoning** (issue #4738), with regression tests asserting `gpt-5.4-mini`/`gpt-5.4` keep `temperature`. Azure deployments behave differently and the SDK exposes no capability flag, so the model name alone is ambiguous.

This implements the issue's recommended long-term fix: an explicit, opt-in config flag.

## Fix

- `BaseLlmConfig` gains `is_reasoning_model: Optional[bool] = None` (forwarded by `OpenAIConfig` and `AzureOpenAIConfig`, and through the `BaseLlmConfig`→`AzureOpenAIConfig` conversion in `AzureOpenAILLM`).
- `LLMBase._is_reasoning_model` consults the explicit value first; `None` (default) falls back to the existing name-based heuristic — **fully backwards compatible**.
- Deployments with custom/versioned names opt in (`is_reasoning_model=True`) to drop `max_tokens`, with no monkey-patching and without changing the OpenAI `gpt-5.x` contract.

## Verification

- `pytest tests/llms/test_azure_openai.py tests/llms/test_openai.py::test_is_reasoning_model_explicit_override tests/llms/test_openai.py::test_is_reasoning_model_classification tests/llms/test_openai.py::test_gpt5_mini_not_classified_as_reasoning` — **19 passed**
- 4 new tests: Azure force-true drops `max_tokens`/`temperature`; force-false keeps them; `None` defaults to heuristic; OpenAI override both ways + `None` preserves heuristic.
- Pre-existing OpenAI `gpt-5.x` non-reasoning regression tests still pass (contract preserved).
- Did NOT change: the name-based heuristic itself, or any default behavior when `is_reasoning_model` is unset.

Note: 3 unrelated `test_openai.py` failures (`test_openai_llm_base_url`, `test_store_sent_when_explicitly_*`) reproduce on clean `upstream/main` (local `OPENROUTER_BASE_URL` env pollution) and are not touched by this change.

### DIFF
--- a/mem0/configs/llms/azure.py
+++ b/mem0/configs/llms/azure.py
@@ -41,6 +41,10 @@ class AzureOpenAIConfig(BaseLlmConfig):
             vision_details: Vision detail level, defaults to "auto"
             reasoning_effort: Effort level for reasoning models ("low", "medium", "high"), defaults to None
             http_client_proxies: HTTP client proxy settings, defaults to None
+            is_reasoning_model: Explicit override for reasoning-model detection.
+                None (default) uses the name-based heuristic. Set True to drop
+                max_tokens/temperature (e.g. for versioned Azure deployments like
+                "gpt-5.4-nano-2026-03-17"), or False to force standard params.
             azure_kwargs: Azure-specific configuration, defaults to None
         """
         # Initialize base parameters
--- a/mem0/configs/llms/azure.py
+++ b/mem0/configs/llms/azure.py
@@ -23,6 +23,7 @@ class AzureOpenAIConfig(BaseLlmConfig):
         vision_details: Optional[str] = "auto",
         reasoning_effort: Optional[str] = None,
         http_client_proxies: Optional[dict] = None,
+        is_reasoning_model: Optional[bool] = None,
         # Azure OpenAI-specific parameters
         azure_kwargs: Optional[Dict[str, Any]] = None,
     ):
@@ -54,6 +55,7 @@ class AzureOpenAIConfig(BaseLlmConfig):
             vision_details=vision_details,
             reasoning_effort=reasoning_effort,
             http_client_proxies=http_client_proxies,
+            is_reasoning_model=is_reasoning_model,
         )
 
         # Azure OpenAI-specific parameters
--- a/mem0/configs/llms/base.py
+++ b/mem0/configs/llms/base.py
@@ -58,11 +58,11 @@ class BaseLlmConfig(ABC):
             is_reasoning_model: Explicit override for reasoning-model detection.
                 When None (default), the model is classified automatically from its
                 name (preserving existing behavior). Set to True to force the
-                reasoning-model parameter set (e.g. send max_completion_tokens and
-                drop max_tokens/temperature), or False to force the standard
-                parameter set. Useful for deployments with custom/versioned model
-                names (e.g. Azure "gpt-5.4-nano-2026-03-17") that the name-based
-                heuristic cannot recognize. Defaults to None
+                reasoning-model parameter set (drop max_tokens and temperature),
+                or False to force the standard parameter set. Useful for
+                deployments with custom/versioned model names (e.g. Azure
+                "gpt-5.4-nano-2026-03-17") that the name-based heuristic cannot
+                recognize. Defaults to None
         """
         self.model = model
         self.temperature = temperature
--- a/mem0/configs/llms/base.py
+++ b/mem0/configs/llms/base.py
@@ -25,6 +25,7 @@ class BaseLlmConfig(ABC):
         vision_details: Optional[str] = "auto",
         reasoning_effort: Optional[str] = None,
         http_client_proxies: Optional[Union[Dict, str]] = None,
+        is_reasoning_model: Optional[bool] = None,
     ):
         """
         Initialize a base configuration class instance for the LLM.
@@ -54,6 +55,14 @@ class BaseLlmConfig(ABC):
                 Defaults to None (uses the model's default reasoning effort)
             http_client_proxies: Proxy settings for HTTP client.
                 Can be a dict or string. Defaults to None
+            is_reasoning_model: Explicit override for reasoning-model detection.
+                When None (default), the model is classified automatically from its
+                name (preserving existing behavior). Set to True to force the
+                reasoning-model parameter set (e.g. send max_completion_tokens and
+                drop max_tokens/temperature), or False to force the standard
+                parameter set. Useful for deployments with custom/versioned model
+                names (e.g. Azure "gpt-5.4-nano-2026-03-17") that the name-based
+                heuristic cannot recognize. Defaults to None
         """
         self.model = model
         self.temperature = temperature
@@ -64,4 +73,5 @@ class BaseLlmConfig(ABC):
         self.enable_vision = enable_vision
         self.vision_details = vision_details
         self.reasoning_effort = reasoning_effort
+        self.is_reasoning_model = is_reasoning_model
         self.http_client = httpx.Client(proxies=http_client_proxies) if http_client_proxies else None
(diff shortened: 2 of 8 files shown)


kartik-mem0 (commented):
## Review

Solid PR — the explicit `is_reasoning_model` config override is the right architectural call here. The name-based heuristic can't disambiguate between OpenAI-direct and Azure behavior for the same model string (e.g. `gpt-5.4-mini`), so an opt-in flag is the only approach that doesn't break the existing #4738 regression tests.

Verified against [Azure's reasoning model docs](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/reasoning) — all gpt-5.x models on Azure are classified as reasoning models that reject `max_tokens` and `temperature`. The fix correctly addresses this.

### One gap to close

`mem0/llms/openai.py` line 21-34 — the `BaseLlmConfig` → `OpenAIConfig` conversion does **not** forward `is_reasoning_model`, even though you correctly added it to the identical pattern in `mem0/llms/azure_openai.py`:

```python
# mem0/llms/openai.py — current (missing)
elif isinstance(config, BaseLlmConfig) and not isinstance(config, OpenAIConfig):
    config = OpenAIConfig(
        model=config.model,
        ...
        reasoning_effort=getattr(config, 'reasoning_effort', None),
        # ← is_reasoning_model not forwarded here
    )
```

Should be:
```python
        reasoning_effort=getattr(config, 'reasoning_effort', None),
        is_reasoning_model=getattr(config, 'is_reasoning_model', None),
    )
```

Low risk since users typically pass `OpenAIConfig` directly, but for consistency with the `AzureOpenAILLM` conversion path it should be there.

### Everything else looks correct

- `BaseLlmConfig` — field added cleanly, good docstring
- `OpenAIConfig` / `AzureOpenAIConfig` — parameter forwarded to `super().__init__()`
- `AzureOpenAILLM` — conversion path updated
- `LLMBase._is_reasoning_model()` — `getattr` check before heuristic, `None` falls through
- Tests — 4 new tests covering force-true, force-false, and None-default for both providers

kartik-mem0 (approved):

PR #5328, opened 2026-05-31
fix: avoid mutable default arguments

### LINKED ISSUE #5301: [Bug] Mutable default arguments (B006) in `Completions.create` and `BaseEmbedderConfig.__init__`
### Component

Core / Python SDK

### Description

## Affected symbols

## Reproduction

```python
import inspect

Expected behavior

Environment

pranjalbhatia710:
## Summary
- replace mutable default `messages=[]` in `Completions.create` with `None`
- replace mutable default `azure_kwargs={}` in `BaseEmbedderConfig` with `None`
- add regression tests for both defaults and preserve AzureConfig instance handling

Fixes #5301

## Test plan
- `uv run --extra test --extra llms --with azure-identity pytest tests/test_proxy.py tests/embeddings/test_azure_openai_embeddings.py -q`
- `uv run --extra test --with ruff ruff check mem0/proxy/main.py mem0/configs/embeddings/base.py tests/test_proxy.py tests/embeddings/test_azure_openai_embeddings.py`

### DIFF
--- a/mem0/configs/embeddings/base.py
+++ b/mem0/configs/embeddings/base.py
@@ -25,7 +25,7 @@ class BaseEmbedderConfig(ABC):
         model_kwargs: Optional[dict] = None,
         huggingface_base_url: Optional[str] = None,
         # AzureOpenAI specific
-        azure_kwargs: Optional[AzureConfig] = {},
+        azure_kwargs: Optional[Union[AzureConfig, Dict]] = None,
         http_client_proxies: Optional[Union[Dict, str]] = None,
         # VertexAI specific
         vertex_credentials_json: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


kartik-mem0:
Hi @pranjalbhatia710 — thank you so much for taking the time to work on this fix! Your PR had some genuinely thoughtful touches, especially the `isinstance(azure_kwargs, AzureConfig)` pass-through logic and the type annotation update to `Optional[Union[AzureConfig, Dict]]` — that was actually the most robust handling of the azure_kwargs type mismatch among all the PRs we received.

We've gone ahead and merged #5302 which landed a very similar fix along with regression tests, so this issue is now resolved. But please don't let this discourage you — your contribution showed real care for correctness and we truly appreciate you spending your time on this. We hope to see you around on future issues! 🙏

PR #5329, opened 2026-05-31
Fix mutable default arguments

Oxygen56:
## Summary
- replace mutable default values in Completions.create and BaseEmbedderConfig
- preserve AzureConfig compatibility while creating a fresh default config
- add regression coverage for isolated defaults

## Testing
- python3 -m py_compile mem0/proxy/main.py mem0/configs/embeddings/base.py tests/test_proxy.py tests/embeddings/test_azure_openai_embeddings.py
- Not run: pytest, because this local Python environment is missing pytest and project test dependencies.

### DIFF
--- a/mem0/configs/embeddings/base.py
+++ b/mem0/configs/embeddings/base.py
@@ -1,6 +1,6 @@
 import os
 from abc import ABC
-from typing import Dict, Optional, Union
+from typing import Any, Dict, Optional, Union
 
 import httpx
 
@@ -25,7 +25,7 @@ class BaseEmbedderConfig(ABC):
         model_kwargs: Optional[dict] = None,
         huggingface_base_url: Optional[str] = None,
         # AzureOpenAI specific
-        azure_kwargs: Optional[AzureConfig] = {},
+        azure_kwargs: Optional[Union[AzureConfig, Dict[str, Any]]] = None,
         http_client_proxies: Optional[Union[Dict, str]] = None,
         # VertexAI specific
         vertex_credentials_json: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


biswajeetdev (commented):
Good catch on both instances. The mutable default argument antipattern in Python means the same list or dict object is shared across all calls, so mutations in one call silently affect subsequent calls — a notoriously hard bug to track down in production.

The `messages: Optional[List] = None` fix with `messages = messages or []` inside the body is the standard corrective pattern. One small thing: `messages or []` treats an explicitly passed empty list as falsy and replaces it with a new `[]`. This is fine here since an empty message list is semantically equivalent, but `if messages is None: messages = []` is more precise if callers might explicitly pass `[]` to mean "no messages" in a distinguishable way from not passing the arg at all.

The test using `inspect.signature(Completions.create).parameters["messages"].default is None` is a nice way to guard this at the API boundary — it will catch any future regression without needing to actually instantiate and mutate the class.

Oxygen56:
Thanks for the review @biswajeetdev! I've reviewed the suggestion about `if messages is None` vs `messages = messages or []`.

In this specific case, the two are behaviorally equivalent since an explicitly-passed empty list is semantically the same as "no messages" — there's no code path that meaningfully distinguishes between them. That said, I see the style preference argument and am happy to switch to the more explicit `if messages is None: messages = []` if you prefer — just let me know.

Otherwise, would love to get this merged when you have a moment. The mutable default issue, while not actively causing bugs today, is one of those things that silently bites the next person who adds a mutation downstream.

kartik-mem0:
Closing as duplicate — the fix for this was merged via #5302. Thank you for the contribution!

PR #5332, opened 2026-05-31
fix: read llms coverage files as utf-8

Ghraven:
## What changed
- Reads `scripts/llms-txt-ignore.txt` and `docs/llms.txt` with explicit `encoding="utf-8"`.
- Writes updated `docs/llms.txt` with explicit `encoding="utf-8"` when `--write` scaffolds missing entries.

## Problem
The helper currently relies on the platform default encoding for text I/O. On systems where the default encoding is not UTF-8, the docs coverage check can misread docs text or write `docs/llms.txt` with a different encoding than expected.

## Before/after
Before: `read_text()` / `write_text()` used the process default encoding.
After: the docs coverage helper consistently uses UTF-8 for the docs index and ignore file.

## Verification
- `python scripts/check-llms-txt-coverage.py`

### DIFF
--- a/scripts/check-llms-txt-coverage.py
+++ b/scripts/check-llms-txt-coverage.py
@@ -42,7 +42,7 @@ def load_ignore_prefixes() -> list[str]:
     if not IGNORE_FILE.exists():
         return []
     prefixes: list[str] = []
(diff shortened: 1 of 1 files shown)


Ghraven:

PR #5333, opened 2026-05-31
fix(s3_vectors): list() wraps results in an extra list, breaking callers

devteamaegis:
## What's broken

`S3VectorStore.list()` returns `List[List[OutputData]]` instead of `List[OutputData]`:

```python
store = S3VectorStore(...)
results = store.list()
# results is [[OutputData(...), OutputData(...)]]  ← nested list
# should be   [OutputData(...), OutputData(...)]
```

Any caller that iterates over the result gets a single inner list as the first element, not the actual records.

## Why it happens

The return statement wraps `_parse_output()` in an extra list literal:

```python
return [self._parse_output(all_vectors)]   # wrong
```

Every other vector store returns the list directly:

```python
return self._parse_output(all_vectors)     # correct
```

## Fix

Remove the outer `[...]`.

### DIFF
--- a/mem0/vector_stores/s3_vectors.py
+++ b/mem0/vector_stores/s3_vectors.py
@@ -198,7 +198,7 @@ class S3Vectors(VectorStoreBase):
         all_vectors = []
         for page in pages:
             all_vectors.extend(page.get("vectors", []))
-        return [self._parse_output(all_vectors)]
+        return self._parse_output(all_vectors)
 
     def reset(self):
         logger.warning(f"Resetting index {self.collection_name}...")
--- a/tests/vector_stores/test_s3_vectors.py
+++ b/tests/vector_stores/test_s3_vectors.py
@@ -197,6 +197,39 @@ def test_delete(mock_boto_client):
     )
 
 
+def test_list_returns_flat_list(mock_boto_client):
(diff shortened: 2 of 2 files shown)


biswajeetdev (commented):
One-line fix, clearly correct. `_parse_output` returns a list, so wrapping it in `[...]` was producing a list-of-a-list — callers iterating over the result of `list()` would get a single element (the inner list) rather than the individual vectors.

Two small follow-up suggestions:

1. **Test coverage** — this is a silent data shape bug that is easy to reintroduce. A unit test asserting that `list()` returns a flat list of vector objects (not a nested list) would lock in the fix.

2. **Parallel issue in other stores** — worth a quick grep across the other vector store implementations (`pinecone.py`, `chroma.py`, etc.) to check if any have the same `[self._parse_output(...)]` pattern in their `list()` methods.

devteamaegis:
Thanks @biswajeetdev — added the regression test. `test_list_returns_flat_list` mocks the paginator to return two vectors and asserts `list()` returns a flat list of `OutputData` (checking `len == 2`, `results[0]` isn't a list, and the ids/payload come through). It fails on the old `[self._parse_output(...)]` return and passes with the fix, so it locks the shape against regressions.

PR #5336, opened 2026-06-01
fix: persist telemetry state as utf-8

Ghraven:
## What changed
- Read and write the server telemetry state file with explicit `encoding="utf-8"`.
- Added a focused test that round-trips non-ASCII telemetry state through `_save_state()` and `_load_state()`.

## Problem
The telemetry state file stores JSON text but currently relies on the platform default encoding. On non-UTF-8 systems, state containing non-ASCII operator input, such as onboarding use-case text, can be decoded or written inconsistently.

## Before/after
Before: `STATE_PATH.read_text()` and `STATE_PATH.write_text(...)` used the process default encoding.
After: telemetry state JSON is consistently read and written as UTF-8.

## Verification
- `python -m pytest tests/test_server_telemetry.py`

### DIFF
--- a/server/telemetry.py
+++ b/server/telemetry.py
@@ -35,7 +35,7 @@ def _load_state() -> dict[str, Any]:
     if not STATE_PATH.exists():
         return {}
     try:
-        return json.loads(STATE_PATH.read_text())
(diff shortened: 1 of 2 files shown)


Ghraven:

Issue #5337, opened 2026-06-01
Production K8s deployment: env-driven config + tagged Docker image publishing


shivamtyagi-flex:
### Component

Core / Python SDK

### Description

Hi mem0 maintainers 👋
  
  First — thanks for mem0. We've been running a self-hosted instance to give a Slack-based AI coding bot cross-thread memory recall, and the underlying server has been solid. Cross-thread
  recall (`agent_id` + `user_id` filter, `run_id` omitted) works exactly as advertised on the pgvector backend.
  
  While moving from local PoC to a Kubernetes deployment, we hit two production-hygiene gaps and ended up forking + patching. Before we keep maintaining the fork, I wanted to ask if you'd
  be open to upstream PRs that close these gaps for everyone. Filing this issue first to gauge interest before doing the PR work.
  
  ## Gap 1: No env-driven config path → can't mount K8s ConfigMaps
  
  `server/server_state.py` only loads configuration from hardcoded paths / `_load_overrides`. There's no way to point it at a config file mounted from a K8s ConfigMap at a custom location,
  which is the standard 12-factor / declarative-config pattern for containerized deployments.
  
  We patched this in our fork with a small fail-soft function (~50 lines including comments) that:
  
  - Reads `MEM0_CONFIG_PATH` env var at startup
  - Loads the YAML if present, returns `{}` if unset
  - Logs an ERROR + falls back to defaults on missing file / invalid YAML / non-mapping top-level — matches the fail-soft style of your existing `_load_overrides`
  
  This lets a deployment do:
  
  ```yaml
  # K8s pod spec
  env:
  - name: MEM0_CONFIG_PATH
    value: /app/config/mem0-config.yaml
  volumeMounts:
  - name: config
    mountPath: /app/config
    readOnly: true
  # ...with the ConfigMap declaring the YAML separately
  ```
  
  Without it, every config change requires a baked-into-image rebuild. With it, edit ConfigMap → restart pod.
  
  **Would you accept a PR adding this?** Happy to write up a proper PR with tests + match your code style.
  
  ## Gap 2: Stale Docker Hub image (`mem0/mem0-api-server`)
  
  Current state as of 2026-06-01:
  
  ```
  Total tags: 1
  latest    2025-09-10T17:34:48Z    sha256:2fcf4bb713c…    (263 days ago)
  ```
  
  A single mutable `latest` tag, ~8.6 months old. This makes the image unsuitable for production K8s deployment in any setup that requires:
  
[...]

PR #5338, opened 2026-06-01
feat(server): support MEM0_CONFIG_PATH env var for K8s ConfigMap-driven config

### LINKED ISSUE #5339: Production K8s deployment: env-driven config + tagged Docker image publishing
### Component

Core / Python SDK

### Description

Hi mem0 maintainers 👋
  
  First — thanks for mem0. We've been running a self-hosted instance to give a Slack-based AI coding bot cross-thread memory recall, and the underlying server has been solid. Cross-thread
  recall (`agent_id` + `user_id` filter, `run_id` omitted) works exactly as advertised on the pgvector backend.
  
  While moving from local PoC to a Kubernetes deployment, we hit two production-hygiene gaps and ended up forking + patching. Before we keep maintaining the fork, I wanted to ask if you'd
  be open to upstream PRs that close these gaps for everyone. Filing this issue first to gauge interest before doing the PR work.
  
  ## Gap 1: No env-driven config path → can't mount K8s ConfigMaps
  
  `server/server_state.py` only loads configuration from hardcoded paths / `_load_overrides`. There's no way to point it at a config file mounted from a K8s ConfigMap at a custom location,
  which is the standard 12-factor / declarative-config pattern for containerized deployments.
  
  We patched this in our fork with a small fail-soft function (~50 lines including comments) that:
  
  - Reads `MEM0_CONFIG_PATH` env var at startup
  - Loads the YAML if present, returns `{}` if unset
  - Logs an ERROR + falls back to defaults on missing file / invalid YAML / non-mapping top-level — matches the fail-soft style of your existing `_load_overrides`
  
  This lets a deployment do:
  
  ```yaml
  # K8s pod spec
  env:
  - name: MEM0_CONFIG_PATH
    value: /app/config/mem0-config.yaml
  volumeMounts:
  - name: config
    mountPath: /app/config
    readOnly: true
  # ...with the ConfigMap declaring the YAML separately
  ```
  
  Without it, every config change requires a baked-into-image rebuild. With it, edit ConfigMap → restart pod.
  
  **Would you accept a PR adding this?** Happy to write up a proper PR with tests + match your code style.
  
  ## Gap 2: Stale Docker Hub image (`mem0/mem0-api-server`)
  
  Current state as of 2026-06-01:
  
  ```
  Total tags: 1
  latest    2025-09-10T17:34:48Z    sha256:2fcf4bb713c…    (263 days ago)
  ```
  
  A single mutable `latest` tag, ~8.6 months old. This makes the image unsuitable for production K8s deployment in any setup that requires:
  
  - Pinning to immutable tags for rollback (no semver tags exist)
  - Tracking which `sha` is deployed (mutable `latest` makes audit ambiguous)
  - Vulnerability scanning against a pinned baseline
  
  We ended up forking + building our own image with semver tags via our private CI. Other production users likely face the same wall.
  
  **Would you accept a PR adding a publish workflow?** Concretely:
  
  - `.github/workflows/publish-docker.yml` that triggers on `release` published events (or git-tag push)
  - Builds `server/Dockerfile`
  - Publishes to `mem0/mem0-api-server` on Docker Hub with semver tags (`v1.0.0`, `v1.0`, `v1`, `latest`)
  - Optionally signs the image (cosign) for supply-chain integrity
  
  You'd configure the `DOCKERHUB_TOKEN` secret; CI handles the rest on each tag/release.
  
  If the operational responsibility isn't something the team wants to take on right now, completely understand — but the workflow itself is a small, self-contained PR that you can opt in or
   out of.
  
  ## What I'm NOT proposing
  
  To be clear, I'm NOT asking for:
  
  - Changes to the `server/Dockerfile` or the local `docker-compose.yml` (those serve the local-dev pattern well)
  - Any change to the existing `cd.yml` (PyPI publish) workflow
  - Anything Flex-specific
  
  These are scoped, generic improvements for any production K8s deployment of mem0.
  
  ## Path forward
  
  Happy to file either or both PRs if you're receptive. Let me know your thoughts — or if there's a different shape you'd prefer (e.g., docs-only contribution explaining the patterns).
  
  Thanks again for the project.

shivamtyagi18:
## Summary

Closes #5339 (first half — `MEM0_CONFIG_PATH`; Docker Hub publishing half tracked separately on the issue)

Adds env-var indirection to the mem0 server's config-loading path so a deployment can mount a YAML config file from a Kubernetes ConfigMap (or any other declarative source) without baking config into the image.

- New: `_load_yaml_config_path()` in `server/server_state.py` — reads `MEM0_CONFIG_PATH` env var at startup, parses the referenced YAML, merges it into the running config via the existing `_merge_config` helper.
- Wiring: called inside `initialize_state()` before DB overrides apply.
- Dependency: `pyyaml>=6.0,<7.0` added to `server/requirements.txt` (lazy-imported inside the function — no impact when the env var is unset).
- Tests: 8 new tests in `server/tests/test_mem0_config_path.py` covering env-set and env-unset paths, 3 error-fallback paths, and 2 `initialize_state` integration tests. All pass in ~0.02s.

## Motivation

Deploying mem0 to Kubernetes today requires one of:

1. **Bake the config into a custom image build** — every config change is a rebuild + redeploy.
2. **Call `POST /configure` on every pod restart** — fragile, racy, and requires custom orchestration outside the K8s declarative model.

The standard [12-factor](https://12factor.net/config) pattern is a ConfigMap mounted at a stable path, with the location passed via an env var. This PR enables that:

```yaml
# Kubernetes pod spec excerpt
env:
- name: MEM0_CONFIG_PATH
  value: /app/config/mem0-config.yaml
volumeMounts:
- name: config
  mountPath: /app/config
  readOnly: true
# (with the ConfigMap declaring the YAML separately)
```

After this PR, the operator edits the ConfigMap → kubectl rolls the pod → server picks up the new config at startup. No image rebuild, no runtime POST.

## Design notes

- **Precedence (low → high):** `DEFAULT_CONFIG` < `MEM0_CONFIG_PATH` < DB overrides. Matches the existing layered-overrides model.
- **Behavior unchanged when `MEM0_CONFIG_PATH` is unset.** Pure additive — no regressions for existing users.
- **Fail-soft on all error paths** (missing file, invalid YAML, non-mapping top-level). Matches the style of the existing `_load_overrides` function: operator misconfiguration produces clear ERROR logs (with `exc_info=True` for parse failures) but the server still boots on defaults rather than crash-looping. Same trade-off the surrounding code already makes; I considered failing fast to prevent silent-wrong-config but chose consistency with the file's existing style. Happy to switch to fail-fast if maintainers prefer.
- **pyyaml import is lazy** — only triggered when `MEM0_CONFIG_PATH` is actually set. If pyyaml is missing the function logs an actionable ERROR and falls back to defaults rather than crashing imports.

## Test coverage

The new test file covers:

- Env var unset → returns `{}` (no-op)
- Env var set to valid YAML file → returns parsed dict
- Env var set but file missing → logs ERROR + returns `{}`
- Env var set but YAML invalid → logs ERROR + returns `{}`
- Env var set but YAML top-level is a list/string (not a dict) → logs ERROR + returns `{}`
- `initialize_state()` called with no env var → identical behavior to current main
- `initialize_state()` called with env var → loaded config is merged

Uses monkeypatched `mem0.Memory` to keep test runtime fast (~0.02s for the full file).

## Out of scope (intentional)

- **No changes to the public HTTP API.** `POST /configure` remains the runtime-config path; `MEM0_CONFIG_PATH` is the startup-config path.
- **No changes to `server/Dockerfile`, `docker-compose.yml`, or the existing `cd.yml` PyPI publish workflow.**
- **No new env vars beyond `MEM0_CONFIG_PATH`.**

## Test plan

- [x] All 8 new tests pass locally against current `main` (`python3 -m pytest server/tests/test_mem0_config_path.py -q`)
- [x] `pyyaml` import is lazy and only triggers when env var is set
- [x] Author email is my personal address (this is an upstream contribution, no employer involvement)
- [ ] Maintainer review

## Companion issue

I've also filed an issue describing the broader K8s deployment gaps I hit (this PR addresses the `MEM0_CONFIG_PATH` half; the other half is the Docker Hub image publishing cadence). Happy to discuss approach changes there before further work.

---

Thanks for mem0 — the cross-thread recall semantics are exactly what we needed and the server is rock-solid. This patch came out of running mem0 in production K8s; the only friction point was config indirection. Hope this is useful.

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -24,3 +24,7 @@ posthog>=3.0,<8.0
 
 # Per-IP rate limiting on auth endpoints
 slowapi>=0.1.9,<1.0
+
+# YAML config-file support (MEM0_CONFIG_PATH env var). Upper bound <7 guards against a potential
+# major-version API break in a future release (PyYAML 7.0 does not exist as of this writing).
+pyyaml>=6.0,<7.0
--- a/server/server_state.py
+++ b/server/server_state.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import os
 import threading
 from copy import deepcopy
 from typing import Any, Callable, Dict
@@ -73,10 +74,74 @@ def _merge_config(base: Dict[str, Any], updates: Dict[str, Any]) -> Dict[str, An
     return merged
 
 
+def _load_yaml_config_path() -> Dict[str, Any]:
+    """Load a YAML config file from the path given by ``MEM0_CONFIG_PATH``.
+
+    Returns the parsed dict on success, an empty dict when the env var is unset,
+    or an empty dict (after logging a clear error) when the file is missing or
+    contains invalid YAML — matching mem0's existing fail-soft error style.
+    """
+    # Fail-soft on all errors (missing file, invalid YAML, non-mapping top-level) — matches
+    # mem0's existing override-loading style in `_load_overrides`. An operator who misconfigures
+    # MEM0_CONFIG_PATH will see ERROR logs on startup but the server will still boot on bare
+    # defaults. Trade-off accepted: failing fast would prevent silent-wrong-config in K8s, but
+    # fail-soft consistency with the surrounding code is preferred. Revisit if a prod incident
+    # surfaces from a misconfigured ConfigMap mount.
+    config_path = os.environ.get("MEM0_CONFIG_PATH", "").strip()
+    if not config_path:
+        return {}
+
+    try:
+        import yaml  # imported lazily — only needed when env var is set
+    except ImportError:
+        logging.error(
+            "MEM0_CONFIG_PATH is set to %r but 'pyyaml' is not installed. "
+            "Add pyyaml to server/requirements.txt and rebuild the image. "
+            "Falling back to default configuration.",
+            config_path,
+        )
+        return {}
+
+    try:
+        with open(config_path, "r", encoding="utf-8") as fh:
+            data = yaml.safe_load(fh)
+    except FileNotFoundError:
+        logging.error(
+            "MEM0_CONFIG_PATH is set to %r but the file does not exist. "
+            "Falling back to default configuration.",
+            config_path,
+            exc_info=True,
+        )
+        return {}
+    except yaml.YAMLError as exc:
+        logging.error(
+            "MEM0_CONFIG_PATH is set to %r but the file contains invalid YAML: %s. "
+            "Falling back to default configuration.",
+            config_path,
+            exc,
+            exc_info=True,
+        )
+        return {}
+
+    if not isinstance(data, dict):
+        logging.error(
+            "MEM0_CONFIG_PATH is set to %r but the file does not contain a YAML mapping at the top level. "
+            "Falling back to default configuration.",
+            config_path,
+        )
+        return {}
+
+    logging.info("Loaded mem0 config from MEM0_CONFIG_PATH=%r", config_path)
+    return data
+
+
 def initialize_state(default_config: Dict[str, Any]) -> None:
     global _current_config, _memory_instance
     with _state_lock:
         _current_config = deepcopy(default_config)
+        yaml_overrides = _load_yaml_config_path()
+        if yaml_overrides:
+            _current_config = _merge_config(_current_config, yaml_overrides)
         overrides = _load_overrides()
         if overrides:
             _current_config = _merge_config(_current_config, overrides)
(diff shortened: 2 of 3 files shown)


shivamtyagi18:
Note: the companion issue was refiled as #5339 (from my personal account, for OSS-identity consistency). Old #5337 is closed as a duplicate. PR body updated to reference the new issue.

pablomdc:
@shivamtyagi18 we're running mem0 self-hosted on AWS EKS and need to configure Bedrock as the LLM/embedder provider without modifying the server image. This PR is exactly what we need.
 
We've already cherry-picked this into our build and it works well. Any idea when it might be merged? Would be great to have it in an official release so we don't have to maintain the patch ourselves.

shivamtyagi18:
@kartik-mem0 Can you please review the PR and help merging it. That will help in making this part of official release. Thank you.

PR #5346, opened 2026-06-02
feat(mem0-plugin): file-context injection, stop-hook summaries & activity timeline

kartik-mem0:
## Linked Issue

N/A — new feature work

## Description

Adds three new automatic memory features to the mem0 plugin, covering all four platforms (Claude Code, Codex, Cursor, OpenCode):

- **File-context injection (PreToolUse/Read hook):** Before the agent reads a file, searches mem0 for memories referencing that file path and injects a compact timeline of prior work as `additionalContext`. Gates on file size (>= 1,500 bytes), 5-second hard timeout, silent skip on any failure.
- **Stop-hook session summary:** On session end, parses the transcript JSONL, extracts the last assistant message and files touched, builds a structured prompt, and stores via mem0 API with `infer=True` — letting the backend AI extract structured facts. 90-day expiry. Dedup via marker files with automatic 7-day cleanup of stale markers.
- **SessionStart activity timeline:** On startup, fetches 10 most recent memories and renders a compact timeline with type icons (⚖️ decision, 🔴 bug_fix, 🔵 task_learning, etc.) and relative age indicators below the existing banner.

Also extracts shared `TYPE_ICONS` and `format_age()` into `_formatting.py` to eliminate duplication across Python hook scripts.

### New files
- `scripts/_formatting.py` — shared constants and formatting helpers
- `scripts/file_context.py` — file-context search and timeline formatting
- `scripts/capture_session_summary.py` — transcript parsing and summary storage
- `scripts/session_timeline.py` — recent activity timeline for SessionStart
- `scripts/on_file_read.sh` / `on_file_read_cursor.sh` — shell hook wrappers
- `scripts/on_stop.sh` / `on_stop_cursor.sh` — shell hook wrappers

### Plugin coverage
| Plugin | File Context | Stop Hook | Timeline |
|--------|-------------|-----------|----------|
| Claude Code | ✅ | ✅ | ✅ |
| Codex | ✅ | ✅ | ✅ |
| Cursor | ✅ | ✅ | ✅ |
| OpenCode | ✅ (native TS) | ✅ (native TS) | ✅ (native TS) |

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Verified: Python modules import cleanly, shared `_formatting.py` works correctly with both consumers, pre-commit hooks (ruff + isort) pass.

## Checklist

### DIFF
--- a/mem0-plugin/scripts/capture_session_summary.py
+++ b/mem0-plugin/scripts/capture_session_summary.py
@@ -1,0 +1,296 @@
+#!/usr/bin/env python3
+"""Capture a structured session summary on Stop hook.
+
+Reads the transcript JSONL, extracts the last assistant message and
+files touched during the session, then stores a structured summary
+via the mem0 API with infer=True — letting the platform's AI extract
+key facts (request, decisions, learnings, next steps).
+
+This is the "second agent" architecture: the Stop hook captures raw
+session data, and mem0's backend AI processes it into structured memory.
+
+Input:  JSON on stdin with transcript_path, session_id, cwd, agent_id
+Output: stderr logs only (exit 0 always — must not block)
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import re
+import sys
+import urllib.error
+import urllib.request
+from datetime import date, timedelta
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+from _identity import resolve_api_key, resolve_user_id
+from _project import resolve_branch, resolve_project_id
+
+log = logging.getLogger("mem0-session-summary")
+log.setLevel(logging.DEBUG)
+_handler = logging.StreamHandler(sys.stderr)
+_handler.setFormatter(logging.Formatter("[mem0-session-summary] %(message)s"))
+log.addHandler(_handler)
+
+if os.environ.get("MEM0_DEBUG"):
+    _log_dir = os.path.expanduser("~/.mem0")
+    try:
+        os.makedirs(_log_dir, exist_ok=True)
+        _fh = logging.FileHandler(os.path.join(_log_dir, "hooks.log"))
+        _fh.setFormatter(logging.Formatter("[mem0-session-summary] %(asctime)s %(message)s"))
+        log.addHandler(_fh)
+    except OSError:
+        pass
+
+API_URL = "https://api.mem0.ai"
+MAX_TAIL_LINES = 3000
+MAX_SUMMARY_CHARS = 50000
+SUMMARY_EXPIRY_DAYS = 90
+
+SYSTEM_TAG_RE = re.compile(
+    r"<(?:system-reminder|private|claude-mem-context|persisted-output|system_instruction)>"
+    r".*?"
+    r"</(?:system-reminder|private|claude-mem-context|persisted-output|system_instruction)>",
+    re.DOTALL,
+)
+
+
+def tail_lines(filepath: str, n: int) -> list[str]:
+    try:
+        with open(filepath, "rb") as f:
+            f.seek(0, 2)
+            file_size = f.tell()
+            if file_size == 0:
+                return []
+            chunk_size = min(file_size, n * 4096)
+            f.seek(max(0, file_size - chunk_size))
+            data = f.read().decode("utf-8", errors="replace")
+            return data.splitlines()[-n:]
+    except OSError:
+        return []
+
+
+def extract_last_assistant_message(lines: list[str]) -> str:
+    """Walk transcript backwards, return text content of the last assistant message."""
+    for line in reversed(lines):
+        line = line.strip()
+        if not line:
+            continue
+        if '"type":"assistant"' not in line and '"type": "assistant"' not in line:
+            continue
+        try:
+            entry = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if entry.get("type") != "assistant":
+            continue
+        message = entry.get("message", {})
+        content = message.get("content", [])
+        if isinstance(content, str):
+            return content
+        if isinstance(content, list):
+            parts = []
+            for block in content:
+                if isinstance(block, str):
+                    parts.append(block)
+                elif isinstance(block, dict) and block.get("type") == "text":
+                    parts.append(block.get("text", ""))
+            return "\n".join(parts).strip()
+    return ""
+
+
+def extract_files_touched(lines: list[str]) -> list[str]:
+    """Extract unique file paths from tool_input fields in transcript."""
+    files = set()
+    file_ext_re = re.compile(
+        r"[a-zA-Z0-9_./-]+\.(?:py|ts|tsx|js|jsx|rs|go|rb|java|sh|yaml|yml|json|toml|md|sql|css|html)"
+    )
+    for line in lines:
+        line = line.strip()
+        if not line:
+            continue
+        if '"tool_input"' not in line and '"file_path"' not in line:
+            continue
+        try:
+            entry = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        tool_input = entry.get("tool_input") or entry.get("message", {}).get("tool_input", {})
+        if isinstance(tool_input, dict):
+            fp = tool_input.get("file_path", "")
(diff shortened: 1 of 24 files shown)


mem0-plugin/scripts/capture_session_summary.py
```diff
@@ -0,0 +1,296 @@
...
+        line = line.strip()
+        if not line:
+            continue
+        if '"tool_input"' not in line and '"file_path"' not in line:
+            continue
+        try:
+            entry = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        tool_input = entry.get("tool_input") or entry.get("message", {}).get("tool_input", {})
```
whysosaket: Claude Code transcripts don't have a top-level `tool_input` — for `tool_use` blocks the path lives at `message.content[].input.file_path`. I checked against a real transcript and neither key here resolves, so `files_touched` ends up empty for basically every session (and the "Files touched" line in the prompt + the metadata both quietly disappear). Worth walking the `content` blocks and reading `block["input"]["file_path"]` for `tool_use` (covers Read/Edit/Write/MultiEdit).

mem0-plugin/scripts/file_context.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def search_file_context(
+    api_key: str, user_id: str, project_id: str, file_path: str, cwd: str
+) -> str:
+    """Search mem0 for memories related to a file path."""
+    rel = relative_path(file_path, cwd)
+    basename = os.path.basename(file_path)
+
+    results = search_memories(
```
whysosaket, line 92: This search ignores the `global_search` setting. Since 0.2.8 the other search paths branch on `MEM0_GLOBAL_SEARCH` (see on_session_start.sh / enforce_metadata_defaults.sh), and the OpenCode TS file-context handler does too — so a user with `global_search: true` gets global recall everywhere except this hook and the timeline. `session_timeline.py:33` has the same gap. Worth threading the global filter through both.

mem0-plugin/scripts/file_context.py
```diff
@@ -0,0 +1,139 @@
...
+        api_key, user_id, project_id, rel, top_k=MAX_RESULTS, threshold=0.3
+    )
+
+    if len(results) < 2:
+        extra = search_memories(
```
whysosaket: Heads up on latency: this runs in the foreground before every Read of a file >=1500 bytes and can fire two sequential mem0 searches (rel path, then basename). Most source files clear that gate, so an agent reading 20 files serializes ~20 round-trips ahead of its reads. Might be worth a single search, a tighter timeout, or a small local cache.

mem0-plugin/scripts/session_timeline.py
```diff
@@ -0,0 +1,105 @@
...
+    """Fetch the most recent memories for this project, sorted by recency."""
+    body = {
+        "filters": {"AND": [{"user_id": user_id}, {"app_id": project_id}]},
+        "page": 1,
+        "page_size": MAX_RECENT,
+        "sort": "-created_at",
+    }
+    data = json.dumps(body).encode()
+    req = urllib.request.Request(
+        f"{API_URL}/v3/memories/",
```
whysosaket: Can you confirm the platform supports `POST /v3/memories/` with a `sort` param? Everywhere else lists via `GET /v3/memories/?page=...` and searches via `POST /v3/memories/search/`. If the POST list form or `-created_at` sort isn't honored on v3, the timeline either errors (caught -> silently empty) or shows unsorted results labeled "Recent Activity".

mem0-plugin/scripts/capture_session_summary.py
whysosaket: Confirmed fixed — now walks `message.content[]` for `tool_use` blocks and reads `input.file_path`/`input.command`. Matches the real transcript shape. 👍

mem0-plugin/scripts/file_context.py
whysosaket: Resolved — now reads `MEM0_GLOBAL_SEARCH` and threads it through `search_memories` (and `_search.py` builds the `OR user_id:*` filter). Consistent with the rest of the plugin now.

mem0-plugin/scripts/file_context.py
whysosaket: Nice — collapsing the rel+basename into a single query drops it to one round-trip per Read. Addresses the latency concern.

mem0-plugin/scripts/session_timeline.py
whysosaket: Good — dropping `sort` and moving to the `?page=&page_size=` list endpoint (same pattern as on_session_start.sh) resolves the unverified-param worry. One small note: recency is now implicit on the endpoint's default ordering, so just confirm the v3 list returns newest-first — otherwise "Recent Activity" could be arbitrary. Not a blocker.

whysosaket (dismissed):
Re-reviewed after f6b6b079 — all four review items addressed (files_touched content-block walk, global_search honored in both scripts, Read search collapsed to one round-trip, sort param dropped for the page/page_size list endpoint). Non-blocking: confirm v3 list defaults to newest-first. Approving.

mem0-plugin/scripts/capture_session_summary.py
```diff
@@ -0,0 +1,303 @@
...
+        log.debug("No transcript_path provided")
+        return
+
+    session_id = hook_input.get("session_id", "")
+    cwd = hook_input.get("cwd") or None
+
+    # Dedup: don't re-capture the same session
+    marker_dir = os.path.expanduser("~/.mem0")
+    marker_file = os.path.join(marker_dir, f"stop_captured_{session_id}")
+    if session_id and os.path.isfile(marker_file):
```
whysosaket: Heads up on the Stop-hook semantics: in Claude Code `Stop` fires at the end of *every* assistant turn, not at session end. Combined with this dedup marker — written after the first successful store — the summary gets captured on turn 1 and every later turn is skipped. So the stored "session summary" reflects the last assistant message of the *first* turn, which is the least complete point, and the "what was completed / what's next" framing never sees the actual end state.

Two ways out: switch the trigger to the `SessionEnd` event (semantically the right hook), or keep `Stop` but drop the one-shot marker and update/replace the summary memory keyed by `session_id` each turn so the final stored state reflects the final turn. Flagging since it works against the feature's intent — not tied to the other threads.

whysosaket: Removing the marker fixes the first-turn-only problem — thanks. One follow-up on the new approach, though: the docstring says the platform "deduplicates by session_id metadata," but mem0's `infer` dedup actually scopes by the entity IDs (`user_id`/`agent_id`/`run_id` — see `mem0/memory/main.py` `ENTITY_PARAMS`), not by an arbitrary `metadata.session_id`. As written, `store_summary` only passes `user_id`+`app_id` and tucks `session_id` into metadata, so the infer engine compares each turn's summary against *all* the user's memories in the project rather than just this session's prior summary. Net effect is likely multiple `session_summary` memories accumulating per session (or unpredictable merges with unrelated memories), not a clean "latest turn wins."

If you want session-scoped replacement, pass `run_id=session_id` on the add — that's the idiomatic mem0 mechanism and it scopes the dedup/update to the session.

Also note capture now runs foreground on every Stop (`on_stop.sh:45`, no `&`), so each turn ends with a blocking infer-add (15s timeout). Worth backgrounding it like the telemetry call right below it. Non-blocking, but it's a per-turn latency cost now that it fires every turn instead of once.

whysosaket: Both addressed in `ed65f5a6` — `run_id=session_id` on the add now scopes infer dedup to the session (so within a session the summary updates in place, and different sessions no longer clobber each other), and capture is backgrounded in `on_stop.sh`. Docstring updated to match too. 👍

One tiny parity nit: `on_stop_cursor.sh:32` still runs capture in the foreground (`|| true`, no `&`) — the run_id fix is shared so dedup is fine there, but the backgrounding didn't make it into the Cursor variant. Worth matching for consistency, but not blocking.

whysosaket (approved):

PR #5350, opened 2026-06-03
fix(vector_stores): use platform user-data dir instead of /tmp/{provider}

alvinttang:
Refs #4279.

The default vector store path was `/tmp/{provider}`. Three reasons that breaks:

- macOS LaunchAgents: sandbox often blocks `/tmp` writes; `/tmp` is periodically cleaned by the OS, so any persisted index gets wiped.
- systemd services with `PrivateTmp=` or `noexec` `/tmp`.
- Windows: no `/tmp` at all.
- Docker: `/tmp` is ephemeral unless mounted as a volume, so embedded stores silently lose data on restart.

The maintainer comment on the issue ([here](https://github.com/mem0ai/mem0/issues/4279)) said the right move is to allow an explicit path and pick a more robust default. This PR does both.

New resolution order:

1. `MEM0_DATA_DIR` environment variable (explicit override).
2. Platform convention via stdlib (no new dep):
   - macOS: `~/Library/Application Support/mem0`
   - Windows: `%LOCALAPPDATA%/mem0`
   - Linux/BSD: `$XDG_DATA_HOME/mem0` or `~/.local/share/mem0`
3. Provider name appended as the subdir, so `faiss` keeps its own folder.

Explicit `path=` in the user config still wins over both the env var and the default, so anyone already setting `path` sees no change.

## RED

```
def test_default_path_is_not_tmp():
    cfg = VectorStoreConfig(provider="faiss", config={})
    assert not cfg.config.path.startswith("/tmp/")
E   AssertionError: default path still falls back to /tmp: /tmp/faiss
```

## GREEN

```
test_default_path_is_not_tmp                          PASSED
test_env_var_override                                 PASSED
test_explicit_path_wins                               PASSED
test_macos_default_uses_application_support           PASSED
test_linux_default_respects_xdg_data_home             SKIPPED (linux-only)
4 passed, 1 skipped in 0.52s
```

`tests/configs/` also still passes. The other `tests/vector_stores/test_*` files require optional backends (chromadb, pgvector, mongodb, etc.) so I couldn't run them locally without installing each, but the only line I touched in `configs.py` is the path default, which never reaches those backend paths.

### DIFF
--- a/mem0/vector_stores/configs.py
+++ b/mem0/vector_stores/configs.py
@@ -1,6 +1,38 @@
+import os
+import sys
 from typing import Dict, Optional
 
 from pydantic import BaseModel, Field, model_validator
+
+
+def _default_data_dir() -> str:
+    """Return a writable user-data directory suitable for embedded vector stores.
+
+    Resolution order:
+      1. ``MEM0_DATA_DIR`` environment variable (explicit override).
+      2. Platform convention:
+         - macOS: ``~/Library/Application Support/mem0``
+         - Windows: ``%LOCALAPPDATA%/mem0``
+         - Linux/BSD: ``$XDG_DATA_HOME/mem0`` or ``~/.local/share/mem0``
+
+    The previous default of ``/tmp/{provider}`` broke macOS LaunchAgents, systemd
+    services with ``noexec`` ``/tmp``, Windows (no ``/tmp``), and Docker (ephemeral
+    ``/tmp``). See #4279.
+    """
+    env_dir = os.environ.get("MEM0_DATA_DIR")
+    if env_dir:
+        return env_dir
+    if sys.platform == "darwin":
+        return os.path.expanduser("~/Library/Application Support/mem0")
+    if sys.platform == "win32":
+        return os.path.join(
+            os.environ.get("LOCALAPPDATA") or os.path.expanduser("~/AppData/Local"),
+            "mem0",
+        )
+    return os.path.join(
+        os.environ.get("XDG_DATA_HOME") or os.path.expanduser("~/.local/share"),
+        "mem0",
+    )
 
 
 class VectorStoreConfig(BaseModel):
@@ -61,7 +93,7 @@ class VectorStoreConfig(BaseModel):
 
         # also check if path in allowed kays for pydantic model, and whether config extra fields are allowed
         if "path" not in config and "path" in config_class.__annotations__:
-            config["path"] = f"/tmp/{provider}"
+            config["path"] = os.path.join(_default_data_dir(), provider)
 
         self.config = config_class(**config)
         return self
--- a/tests/vector_stores/test_default_path.py
+++ b/tests/vector_stores/test_default_path.py
@@ -1,0 +1,49 @@
+"""Regression test for #4279: vector store path defaults to /tmp/{provider}, which
+fails or silently loses data in macOS LaunchAgents, systemd services, Docker, etc.
+"""
+import os
+import sys
+from unittest.mock import patch
+
+import pytest
+
+from mem0.vector_stores.configs import VectorStoreConfig
+
+
+def test_default_path_is_not_tmp():
+    """The default path must not be /tmp/{provider} on any platform."""
+    cfg = VectorStoreConfig(provider="faiss", config={})
+    path = cfg.config.path
+    assert path is not None
+    assert not path.startswith("/tmp/"), f"default path still falls back to /tmp: {path}"
+    assert path.endswith("faiss")
+
+
+def test_env_var_override():
+    """MEM0_DATA_DIR must override the platform default."""
+    with patch.dict(os.environ, {"MEM0_DATA_DIR": "/var/lib/mem0_test"}):
+        cfg = VectorStoreConfig(provider="faiss", config={})
+        assert cfg.config.path == os.path.join("/var/lib/mem0_test", "faiss")
+
+
+def test_explicit_path_wins():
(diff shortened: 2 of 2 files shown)


ghost:
Good fix — the /tmp default is a real ops hazard, especially in Docker and systemd environments where silent data loss on restart is exactly the kind of thing that surfaces as a retrieval quality problem rather than a write-path error.

One related note: if you're running an embedded vector store and also hitting the silent embedding drop issue (#5245), the combination is particularly bad — writes silently fail at the embedding layer, and any that do land are potentially wiped on restart. vector-router (https://github.com/Adelagric/vector-router/releases/tag/v0.1.0) sits between the embedding provider and the store and surfaces per-item failures before they reach the store, so it composes with this fix rather than depending on it.

Apache 2.0, drop-in via OPENAI_BASE_URL if useful while both PRs work through review.

alvinttang:
recheck

kartik-mem0:
## Changes Requested

### What this PR does
Replaces the hardcoded `/tmp/{provider}` vector store default with a platform-appropriate user-data directory (XDG on Linux, `~/Library/Application Support/mem0` on macOS, `%LOCALAPPDATA%/mem0` on Windows), plus a `MEM0_DATA_DIR` env-var escape hatch. Explicit `path=` in user config still wins — back-compat preserved.

### Strengths
- Correct approach — this is the right fix for #4279 and the strongest of the competing PRs
- Stdlib-only (`os`, `sys`) — no new dependencies
- Clean resolution order: env var → platform convention → provider subdir
- Windows fallback (`LOCALAPPDATA` → `~/AppData/Local`) is correct
- 5 regression tests with red-green proof: tests fail without the fix, pass with it
- `ruff check` passes cleanly

### Issues

#### Critical (must fix)
**`mem0/vector_stores/configs.py:93` + `mem0/configs/vector_stores/chroma.py:31` — ChromaDB cloud config regression**

When a user configures ChromaDB cloud mode (`{"api_key": "xxx", "tenant": "yyy"}`, no explicit path):

1. `validate_and_create_config` sees `"path" not in config` and `"path" in ChromaDbConfig.__annotations__` → injects `config["path"] = os.path.join(_default_data_dir(), "chroma")` (e.g. `~/.local/share/mem0/chroma`)
2. `ChromaDbConfig.__init__` runs its validator — `cloud_config = True`, `path = "~/.local/share/mem0/chroma"` (truthy)
3. The guard at `chroma.py:31` — `if cloud_config and path == "/tmp/chroma"` — is now **False** (path is the XDG path, not `/tmp/chroma`)
4. `local_config = bool(path) = True` → line 41 fires: `raise ValueError("Cannot specify both cloud configuration and local configuration")`

The guard was written for the old hardcoded sentinel value and is now stale. ChromaDB cloud users will hit a `ValueError` after this PR merges.

**Fix in `mem0/configs/vector_stores/chroma.py`:**
```python
# Replace (brittle sentinel):
if cloud_config and path == "/tmp/chroma":
    values.pop("path", None)
    return values

# With (robust):
if cloud_config:
    values.pop("path", None)  # strip any auto-injected default path for cloud mode
    return values
```

Also add a regression test to `tests/vector_stores/test_default_path.py`:
```python
def test_chroma_cloud_config_not_blocked_by_default_path():
    pytest.importorskip("chromadb")
    cfg = VectorStoreConfig(provider="chroma", config={"api_key": "test", "tenant": "tenant"})
    assert cfg.config.api_key == "test"  # must not raise ValueError
```

#### Important (should fix)
**Missing docs entry for `MEM0_DATA_DIR`** — this is a new user-facing env var for overriding the data directory in containers/CI. Worth a one-liner in the vector store configuration docs so users don't have to discover it by reading source.

#### Minor (optional)
- `QdrantConfig` at `mem0/configs/vector_stores/qdrant.py:16` still has `Field("/tmp/qdrant", ...)` as its own default. Only reachable when `QdrantConfig` is instantiated directly (bypassing `VectorStoreConfig`), so it's a pre-existing gap outside this PR's scope — but worth a follow-up.
- `Refs #4279` in the PR body won't auto-close the issue on merge. `Closes #4279` is probably the intent.

### Evidence
- **CI:** Vercel fails (repo-level auth issue, unrelated to code). CLA passes. No Python CI run recorded for this head.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_default_path.py -v --tb=short` → 4 passed, 1 skipped in 0.35s
- **Red-green proof:** Reverted `mem0/vector_stores/configs.py` to main → 3 failed (test_default_path_is_not_tmp: `assert '/tmp/faiss'` doesn't start with `/tmp/` → **False**). Restored fix → 4 passed. Tests genuinely exercise the bug.
- **Lint:** `ruff check mem0/vector_stores/configs.py tests/vector_stores/test_default_path.py` → all checks passed
- **Linked issue:** #4279 — PR directly addresses the reported root cause (hardcoded `/tmp/{provider}` path). Platform-dir resolution matches what the issue requested.
- **Competing PRs:** #5031 (correct approach, no chroma fix, no env var), #4663 (same approach, same chroma finding), #4913 (tempfile.gettempdir() — still loses data on reboot). Once the Critical fix lands here, recommend merging #5350 and closing the others.
- **Could NOT verify:** Live ChromaDB cloud behavior (no API key in test env); code tracing only for the chroma regression.

### Assessment
Ready to merge: **With fixes** — please address the Critical chroma regression and add the test for it; I'll re-review once you push. The approach is right and this is the PR to land for #4279.

PR #5356, opened 2026-06-03
fix: replace mutable default arguments [] and {} (B006)

### LINKED ISSUE #5301: [Bug] Mutable default arguments (B006) in `Completions.create` and `BaseEmbedderConfig.__init__`

charlesaurav13:
## Linked Issue

Closes #5301

## Description

Two classic Python B006 anti-patterns where a single module-level list/dict instance is shared across every call that uses the default value. Any in-place mutation leaks state between call sites.

### `mem0/proxy/main.py` — `Completions.create`

```python
# Before (shared [] across all calls)
def create(self, model: str, messages: List = [], ...):

# After
def create(self, model: str, messages: Optional[List] = None, ...):
    if messages is None:
        messages = []
```

### `mem0/configs/embeddings/base.py` — `BaseEmbedderConfig.__init__`

```python
# Before (shared {} across all instances + type annotation mismatch)
azure_kwargs: Optional[AzureConfig] = {}

# After
azure_kwargs: Optional[AzureConfig] = None
# body:
self.azure_kwargs = AzureConfig(**(azure_kwargs or {}))
```

The `azure_kwargs` change also fixes the type annotation mismatch — the parameter was declared `Optional[AzureConfig]` but defaulted to `{}` (a plain dict). Behaviour is identical: callers that omit `azure_kwargs` still get a default `AzureConfig()`.

## Type of Change

## Breaking Changes

N/A — both changes are backward-compatible. Callers passing an explicit value are unaffected.

## Test Coverage

- [x] No tests needed — environment-independent code-quality bug, behaviour is unchanged for callers that pass explicit arguments

### DIFF
--- a/mem0/configs/embeddings/base.py
+++ b/mem0/configs/embeddings/base.py
@@ -25,7 +25,7 @@ class BaseEmbedderConfig(ABC):
         model_kwargs: Optional[dict] = None,
         huggingface_base_url: Optional[str] = None,
         # AzureOpenAI specific
-        azure_kwargs: Optional[AzureConfig] = {},
+        azure_kwargs: Optional[AzureConfig] = None,
         http_client_proxies: Optional[Union[Dict, str]] = None,
         # VertexAI specific
         vertex_credentials_json: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hi @charlesaurav13 — thanks a lot for picking this up and putting together a clean fix! Your PR description was really well-structured with the before/after code blocks, and your embedder body line (`AzureConfig(**(azure_kwargs or {}))`) was actually the cleanest of all submissions — it correctly drops the unnecessary trailing `or {}` and matches the existing pattern we already use in `mem0/configs/llms/azure.py`.

We ended up merging #5302 which included regression tests alongside the fix, so this issue is now closed. But we genuinely appreciate your effort and the quality of your work here. We'd love to see you contribute again — there are always interesting issues to tackle! 🙏

charlesaurav13:
Happy to Contribute

PR #5357, opened 2026-06-03
fix(llms): extend _is_reasoning_model to cover o4 family and dated Azure subversions

### LINKED ISSUE #5296: gpt-5.x reasoning models on Azure fail with 400 max_tokens rejection — name-based _is_reasoning_model doesn't match dated subversions

charlesaurav13:
## Linked Issue

Closes #5296

## Description

`_is_reasoning_model` matched exact names (`o1`, `o3`, `gpt-5`, …) and a short prefix list (`o1-`, `o3-`). Azure OpenAI surfaces these models under dated deployment names such as:

| Deployment name | Expected result |
|---|---|
| `gpt-5-2025-06-15` | reasoning ✗ (missed) |
| `gpt-5.4-2025-07-15` | reasoning ✗ (missed) |
| `gpt-5.4-mini-2025-07-15` | reasoning ✗ (missed) |
| `o4-mini-2025-04-16` | reasoning ✗ (missed — o4 family absent entirely) |

Any missed model passes `max_tokens` to the API and receives a **400 rejection**.

### Changes

- Added `o4` and `o4-mini` to the exact-match set (new reasoning model family)
- Extended the prefix tuple with `o4-`, `o4.`, `gpt-5-`, `gpt-5.` to catch all dated and point-release subversions
- Removed the misleading inline comment claiming `gpt-5.x` supports `temperature`

```python
# Before
reasoning_prefixes = ["o1-", "o1.", "o3-", "o3."]

# After
reasoning_prefixes = ("o1-", "o1.", "o3-", "o3.", "o4-", "o4.", "gpt-5-", "gpt-5.")
```

## Type of Change

## Breaking Changes

N/A — only broadens the set of models treated as reasoning models.

## Test Coverage

- [x] No tests needed — pure name-matching logic, deterministic and environment-independent

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -51,7 +51,7 @@ class LLMBase(ABC):
             bool: True if the model is a reasoning model or GPT-5 series
         """
         reasoning_models = {
-            "o1", "o1-preview", "o3-mini", "o3",
+            "o1", "o1-preview", "o3", "o3-mini", "o4", "o4-mini",
             "gpt-5", "gpt-5o", "gpt-5o-mini", "gpt-5o-micro",
         }
 
@@ -62,9 +62,11 @@ class LLMBase(ABC):
         if base_model in reasoning_models:
             return True
 
-        # Match o1/o3 family with prefixes (o1-2024-12-17, o3-2025-04-16)
-        # but NOT gpt-5.x variants (gpt-5.4-mini supports temperature)
-        if any(base_model.startswith(prefix) for prefix in ["o1-", "o1.", "o3-", "o3."]):
+        # Match dated Azure/OpenAI subversions:
+        #   o1-2024-12-17, o3-2025-04-16, o4-mini-2025-04-16
+        #   gpt-5-2025-06-15, gpt-5.4-2025-07-15, gpt-5.4-mini-2025-07-15
+        reasoning_prefixes = ("o1-", "o1.", "o3-", "o3.", "o4-", "o4.", "gpt-5-", "gpt-5.")
+        if any(base_model.startswith(prefix) for prefix in reasoning_prefixes):
             return True
 
         return False


kartik-mem0:
Hi @charlesaurav13 — thanks for working on this! We've just merged #5327 which addresses issue #5296 by adding an explicit `is_reasoning_model` config override rather than extending the name-based heuristic. Since a broader prefix match would break the OpenAI `gpt-5.x` non-reasoning contract (issue #4738), the opt-in flag approach was the safer path forward.

Closing this PR as the issue is now resolved, but we appreciate the effort — feel free to pick up another issue!

PR #5358, opened 2026-06-03
Add Oracle Vector Store Integration

fileames:
## Description

Adds Oracle AI Vector Search as a Mem0 vector store provider.

This PR introduces the `oracledb` vector store backend, including Oracle connection/pool support, collection creation, vector index creation, insert/search/get/update/delete/list/reset operations, JSON metadata filtering, configurable distance metrics, HNSW/IVF index options, and validation for Oracle-specific configuration. It also wires the provider into the vector store factory/config registry, adds the `oracledb` dependency, updates the lockfile, and adds Oracle vector store documentation.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

Tested with internal database connections and mock tests.

## Checklist


uv.lock
kartik-mem0, line 1: hey @fileames, we use hatch here so can you please remove this uv lockfile.

thanks

kartik-mem0:
can you please sign the cla without this we can't take any contributions

thanks!

kartik-mem0:
do you guys have ts parity for this integration? 

if yes, then please open a seperate pr for that as well

https://github.com/oracle/node-oracledb

kartik-mem0:
hey @fileames please sign the cla

tests/vector_stores/test_oracledb.py
```diff
@@ -0,0 +1,590 @@
...
+
+import pytest
+import oracledb
+
+from mem0.configs.vector_stores.oracledb import OracleAIVectorSearchConfig
+from mem0.vector_stores.oracledb import OracleAIVectorSearch, OutputData
+
+
+# Global connection settings (override via env to run in different envs)
+ORACLE_USER = os.environ.get("ORACLE_USER") or ""
```
kartik-mem0, line 18: **CI will fail here.** These tests hit a live Oracle instance (`oracledb.connect` / `create_pool` via the `oracle_db` fixture), but nothing skips when `ORACLE_USER` / `ORACLE_PASSWORD` / `ORACLE_DSN` are unset — they just default to `""`. CI runs `make test` across all of `tests/`, and `oracledb` is now in the `vector_stores` extra, so on the runner every connection-dependent test will **error** (not skip) and fail the pipeline. Every sibling vector-store test (`test_qdrant.py`, `test_pgvector.py`, `test_milvus.py`, …) mocks the client instead. Either mock the Oracle client, or add a module-level guard:

```python
pytestmark = pytest.mark.skipif(
    not (ORACLE_USER and ORACLE_DSN),
    reason="Oracle credentials not configured",
)
```

pyproject.toml
```diff
@@ -50,7 +50,8 @@ vector_stores = [
     "redisvl>=0.1.0,<1.0.0",
     "elasticsearch>=8.0.0,<9.0.0",
     "pymilvus>=2.4.0,<2.6.0",
-    "langchain-aws>=0.2.23,<0.3.0",
+    "langchain-aws>=0.2.23",
```
kartik-mem0: Unrelated to Oracle: this drops the previous `<0.3.0` upper bound on `langchain-aws`. Unpinning a major-version ceiling inside a feature PR is scope creep and risks silently pulling a breaking `0.3.x`. Restore `langchain-aws>=0.2.23,<0.3.0`, or split this into its own change with a reason.

tests/vector_stores/test_oraclevs.py
```diff
@@ -0,0 +1,590 @@
...
+
+DIM = 128
+
+
+def _unique_collection_name() -> str:
+    # Keep under Oracle's 30-char identifier limit
+    return f"TEST_MEM0_{uuid.uuid4().hex[:8]}"
+
+
+def _unwrap_results(results: Any) -> List[OutputData]:
```
kartik-mem0: `delete:` `_unwrap_results` is dead code. `list()` already returns a flat `List[OutputData]` (and `test_list_returns_flat_output` below asserts exactly that), so `isinstance(results[0], list)` is never true and this is always a no-op. The comment ("Current implementation of list() returns [outputs] (nested)") is stale — it describes a shape the implementation doesn't produce. Delete the helper and call `list()` directly at the ~10 call sites.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,444 @@
...
+        create_index = (
+            f"CREATE VECTOR INDEX IF NOT EXISTS {self.config.index_name} ON {self.collection_name} (vector) "
+            f"ORGANIZATION {'INMEMORY NEIGHBOR GRAPH' if self.config.index_type == 'HNSW' else 'NEIGHBOR PARTITIONS'}"
+            f" DISTANCE {distance_metric} {accuracy_str} {parameters_str}"
+        )
+
+        return create_index
+
+    def _index_parameters(self) -> str:
+        index_parameters = self.config.canonical_index_parameters()
```
kartik-mem0: `yagni:` `canonical_index_parameters()` re-runs `_validate_index_parameters`, but `_validate_model` already validated **and** int-coerced these at construction. The only thing the re-run buys you is guarding against someone mutating the private `config.index_parameters` dict *after* construction (`test_index_parameters_reject_post_validation_mutation`) — a contrived threat for an internal config object. `self.config.index_parameters` is enough here; the method and that test can go. Keep the construction-time `parameters[key] = int(value)` coercion — that's what actually neutralizes the `FormattedInt` SQL-injection vector, not the re-validation.

mem0/configs/vector_stores/oracledb.py
```diff
@@ -0,0 +1,159 @@
...
+        _validate_int_parameter(parameters, "min_vectors_per_partition", 0)
+
+    return parameters
+
+
+class OracleAIVectorSearchConfig(BaseModel):
+    """Configuration required to connect to an Oracle database with vector search enabled."""
+
+    connection_params: Optional[dict] = Field(None, description="Database connection parameters, including auth.")
+    use_connection_pool: Optional[bool] = Field(
```
kartik-mem0: Description is copy-pasted from the `client` field: "Oracle Connection or ConnectionPool (overrides connection string...)". `use_connection_pool` is a bool toggle for pool-vs-single-connection — the description should say that.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,444 @@
+"""Oracle AI Vector Search vector store integration for mem0."""
+
+import json
+import logging
+from contextlib import contextmanager
+from typing import Any, Dict, List, Optional
+import re
```
kartik-mem0, line 7: Minor: `import re` / `import array` / `import uuid` sit below the `from …` imports. isort (profile=black) will reorder these into the top stdlib group — run `make sort && make format` so the pre-commit / ruff check doesn't flag it.

kartik-mem0 (commented):
Ponytail review (complexity/over-engineering focus, plus a few correctness/CI issues flagged inline). Score: 7/10.

The provider itself is fairly lean for what it does — a full vector-store backend is inherently ~440 lines and most of it is justified. The 21-config `validate_extra_fields` pattern and the SQL identifier quoting match existing repo convention, so no complaints there.

What holds it back from a higher score:
- The test suite hits a **live Oracle DB** with no skip guard, while every sibling test mocks — this will break CI (see inline).
- An unrelated `langchain-aws` upper-bound removal is bundled in.
- Some dead defensive code (`_unwrap_results`) and one speculative re-validation layer.

None are architectural; all are cheap to address. Inline comments mark the exact lines.

fileames:
@kartik-mem0 Hi, how should we handle the logo under images folder. Is there a specific guideline to follow for the style and colors?

kartik-mem0:
> @kartik-mem0 Hi, how should we handle the logo under images folder. Is there a specific guideline to follow for the style and colors?

hey for the logo that will be managed by us you can share the assets here and we will add it in the docs on your behalf

sudarshan12s:
> do you guys have ts parity for this integration?
> 
> if yes, then please open a seperate pr for that as well
> 
> https://github.com/oracle/node-oracledb

Thanks @kartik-mem0 . I am working on it, I hope to raise a PR shortly.

fileames:
Here is the Oracle logo:
TheO_cmyk_0_86_100_0

I have signed the CLA and all the contributions I make are being done for Oracle.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+                f"INSERT INTO {self.collection_name} (id, vector, payload) VALUES (:id, :vector, :payload)", data
+            )
+
+    def search(
+        self,
+        query: str,
+        vectors: List[float],
+        limit: int = 5,
+        filters: Optional[Dict[str, Any]] = None,
+    ) -> List[OutputData]:
```
kartik-mem0, line 394: **Blocker — this breaks every `add()` and `search()`.**

`mem0/memory/main.py` calls the store with `top_k`, not `limit`:

```python
# main.py:896
existing_results = self.vector_store.search(
    query=parsed_messages, vectors=query_embedding, top_k=10, filters=search_filters,
)
# main.py:1612
semantic_results = self.vector_store.search(
    query=query, vectors=embeddings, top_k=internal_limit, filters=filters
)
```

With `limit` here that is `TypeError: search() got an unexpected keyword argument 'top_k'`. `VectorStoreBase.search` (`mem0/vector_stores/base.py:16`) declares `top_k`, and qdrant/milvus/pgvector/chroma all follow it.

```suggestion
    def search(
        self,
        query: str,
        vectors: List[float],
        top_k: int = 5,
        filters: Optional[Dict[str, Any]] = None,
    ) -> List[OutputData]:
```

(and rename the `limit=limit` bind below to `limit=top_k`)

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+            List[OutputData]: Search results.
+        """
+        filter_clause, params = self._build_filters(filters)
+
+        distance_metric = self.config.distance_metric
+
+        sql = (
+            f"SELECT id, payload, VECTOR_DISTANCE(vector, :query_vec, {distance_metric}) distance "
+            f"FROM {self.collection_name} {filter_clause} ORDER BY VECTOR_DISTANCE(vector, :query_vec, {distance_metric}) FETCH FIRST :limit ROWS ONLY"
+        )
```
kartik-mem0, line 414: **The vector index is never used.** Oracle only consults a vector index for *approximate* similarity search — `FETCH APPROX FIRST n ROWS ONLY`. A plain `ORDER BY VECTOR_DISTANCE(...) FETCH FIRST n ROWS ONLY` is an exact search and does a full table scan.

So as written, `do_create_index=True` (the default) builds an HNSW/IVF index that no query touches, and the whole `index_type` / `index_parameters` / `index_accuracy` surface (~90 lines of config + validation) is dead weight at query time. Either emit `FETCH APPROX FIRST` when `do_create_index` is on, or default `do_create_index=False` and drop the index config.

Secondary, while you're here: `VECTOR_DISTANCE(...)` is spelled out twice. Oracle accepts the select-list alias in `ORDER BY`:

```sql
SELECT id, payload, VECTOR_DISTANCE(vector, :query_vec, COSINE) distance
FROM ... ORDER BY distance FETCH APPROX FIRST :limit ROWS ONLY
```

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+                if part.endswith("[*]"):
+                    base = part[:-3]
+                    path_parts.append(f'."{base}"[*]')
+                else:
+                    path_parts.append(f'."{part}"')
+            json_path = "".join(path_parts)
+
+            # Use JSON_EXISTS with PASSING BY VALUE to preserve types
+            clauses.append(f"JSON_EXISTS(payload, '$?(@{json_path} == ${param})' PASSING :{param} AS \"{param}\")")
+            params[param] = value
```
kartik-mem0: **`"*"` wildcard filters silently return nothing.** mem0 uses `{key: "*"}` as "this key exists with any value" (`main.py:1509`, `main.py:1594`), and every other store special-cases it — pgvector `pgvector.py:81`, qdrant `qdrant.py:281`, chroma `chroma.py:297`, milvus `milvus.py:162`. This is the same class of bug fixed in #6452 (chroma), #6454 (opensearch) and #6508 (milvus); worth not re-introducing.

Here `"*"` becomes `@."run_id" == '*'` and matches literally, so any wildcard-filtered search comes back empty.

```python
for idx, (key, value) in enumerate(filters.items()):
    _validate_metadata_key(key)
    param = f"f_{idx}"
    json_path = ...
    if value == "*":
        clauses.append(f"JSON_EXISTS(payload, '$?(exists(@{json_path}))')")
        continue
```

Related: dict-valued operator filters (`{"created_at": {"gte": ...}}`, supported by pgvector) will be bound straight to `:f_0` as a dict and blow up in the driver. Raising a clear `ValueError` for unsupported operators beats a driver-level `DPY-` error.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+        with self._get_cursor() as cursor:
+            cursor.execute(
+                sql,
+                table_name=self.collection_name.replace('"', ""),
+            )
+            result = cursor.fetchone()
+
+        return {"name": result[0], "count": result[1], "size": result[2]}
+
+    def list(self, filters: Optional[Dict[str, Any]] = None, limit: Optional[int] = None) -> List[OutputData]:
```
kartik-mem0: **Blocker — this breaks every `get_all()`.** Same `top_k` mismatch as `search()`:

```python
# main.py:1297
memories_result = self.vector_store.list(filters=filters, top_k=limit)
# main.py:2929 (async)
memories_result = await asyncio.to_thread(self.vector_store.list, filters=filters, top_k=limit)
```

`VectorStoreBase.list` (`base.py:59`) is `list(self, filters=None, top_k=None)`.

```suggestion
    def list(self, filters: Optional[Dict[str, Any]] = None, top_k: Optional[int] = None) -> List[List[OutputData]]:
```

Other stores also default to `top_k=100` rather than unbounded — `top_k=None` here means an un-`LIMIT`ed full-table select.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+            limit_clause = " FETCH FIRST :limit ROWS ONLY"
+            params["limit"] = limit
+
+        sql = f"SELECT id, payload FROM {self.collection_name} {filter_clause} {limit_clause}"
+
+        with self._get_cursor() as cursor:
+            cursor.execute(sql, **params)
+            rows = cursor.fetchall()
+
+        return [OutputData(id=row[0], score=None, payload=self._load_payload(row[1])) for row in rows]
```
kartik-mem0: **Blocker — wrong return shape, breaks `delete_all()`.** Core unwraps one level:

```python
# main.py:1889
for memory in memories:
```

With a flat list, `[0]` is a single `OutputData` and the `for` raises. Every other store returns one extra level — `pgvector.py:539` `return [[OutputData(...) for r in results]]`, `milvus.py:366` `return [memories]`, `chroma.py:245`, `elasticsearch.py:300`, `weaviate.py:378`.

```suggestion
        return [[OutputData(id=row[0], score=None, payload=self._load_payload(row[1])) for row in rows]]
```

(`_get_all_from_vector_store` tolerates both shapes, which is why `get_all` would look fine in isolation — `delete_all` is the one that actually breaks.)

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+            db_version = tuple([int(v) for v in self.client.version.split(".")])
+        else:
+            with self.client.acquire() as conn:
+                db_version = tuple([int(v) for v in conn.version.split(".")])
+
+        if db_version < (23, 4):
+            raise Exception(
+                f"Oracle DB version {oracledb.__version__} not supported, \
+                must be >=23.4 for vector support"
+            )
```
kartik-mem0, line 264: The message reports the wrong number: `oracledb.__version__` is the **python driver** version, but the check that failed is on `db_version` (the database). A user on driver 3.1 / DB 21c gets "Oracle DB version 3.1.0 not supported, must be >=23.4", which sends them chasing the wrong upgrade.

```suggestion
        if db_version < (23, 4):
            raise ValueError(
                f"Oracle DB version {'.'.join(map(str, db_version))} not supported, must be >=23.4 for vector support"
            )
```

Also worth using `ValueError`/`RuntimeError` over bare `Exception` in both this and the client-version check above.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+
+        collections = self.list_cols()
+        if self._catalog_name(self.collection_name) not in {self._catalog_name(name) for name in collections}:
+            self.create_col()
```
kartik-mem0: `delete:` — `create_col()` already issues `CREATE TABLE IF NOT EXISTS` **and** `CREATE VECTOR INDEX IF NOT EXISTS`, so both are idempotent and this probe buys nothing but a round-trip.

It also actively hurts: if the table was created on a previous run with `do_create_index=False`, this branch skips `create_col()` entirely, so turning `do_create_index` on later never creates the index. Dropping the check fixes that for free.

```suggestion
        self.create_col()
```

`_catalog_name` (L134-136) then has no callers left — `col_info` inlines its own `.replace('"', "")` at L418 — so that goes too. Net −6 lines.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+            return 1.0 / (1.0 + max(0.0, distance))
+        if metric_upper == "EUCLIDEAN_SQUARED":
+            return 1.0 / (1.0 + math.sqrt(max(0.0, distance)))
+        if metric_upper in {"HAMMING", "MANHATTAN"}:
+            return 1.0 / (1.0 + max(0.0, distance))
+        if metric_upper == "DOT":
+            # Oracle's DOT distance is the negated inner product.
+            return -distance
+
+        raise ValueError(f"Unsupported distance metric: {metric}")
```
kartik-mem0: `shrink:` — 17 lines, five branches, three of which are the identical `1/(1+d)`.

```python
_SCORE_FROM_DISTANCE = {
    "COSINE": lambda d: max(0.0, min(1.0, 1.0 - d)),
    "EUCLIDEAN": lambda d: 1.0 / (1.0 + max(0.0, d)),
    "EUCLIDEAN_SQUARED": lambda d: 1.0 / (1.0 + math.sqrt(max(0.0, d))),
    "HAMMING": lambda d: 1.0 / (1.0 + max(0.0, d)),
    "MANHATTAN": lambda d: 1.0 / (1.0 + max(0.0, d)),
    "DOT": lambda d: -d,
}

@staticmethod
def _convert_distance_to_score(distance: float, metric: str) -> float:
    try:
        return _SCORE_FROM_DISTANCE[metric.upper()](distance)
    except KeyError:
        raise ValueError(f"Unsupported distance metric: {metric}") from None
```

The existing parametrized test covers it unchanged. Net −5 lines.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+            raise ValueError(f"ID count must match vector count. Expected {len(vectors)} got {len(ids)}.")
+
+        _ids = ids
+        if not _ids:
+            _ids = [str(uuid.uuid4()) for _ in vectors]
+
+        data = []
+        for vector, payload, _id in zip(vectors, payloads or [{}] * len(vectors), _ids):
+            document = {"id": _id, "vector": array.array("f", vector), "payload": payload}
+            data.append(document)
```
kartik-mem0: `shrink:` — the `_ids` dance plus an append loop is one comprehension.

```suggestion
        ids = ids or [str(uuid.uuid4()) for _ in vectors]
        data = [
            {"id": _id, "vector": array.array("f", vector), "payload": payload}
            for vector, payload, _id in zip(vectors, payloads or [{}] * len(vectors), ids)
        ]
```

Net −3 lines.

kartik-mem0 (commented):
## Ponytail review — 4/10

The scaffolding is right: provider registered in `factory.py` + `configs.py`, dep in the `vector-stores` optional group (not core), docs page + `docs.json` + `llms.txt` all wired, module-level import guard matches the other stores. That part is clean.

The problem is that the store does not satisfy `VectorStoreBase`'s actual calling convention, so it cannot work through `Memory` at all. Three hard breaks, verified against `mem0/memory/main.py` on the current merge-base (8a57967c, 2026-07-20 — this is not drift, the branch is up to date):

| Core call site | Oracle signature | Result |
|---|---|---|
| `main.py:896`, `main.py:1612` → `vector_store.search(..., top_k=N, ...)` | `search(self, query, vectors, limit=5, filters=None)` | `TypeError: search() got an unexpected keyword argument 'top_k'` — every `add()` and `search()` |
| `main.py:1297` → `vector_store.list(filters=filters, top_k=limit)` | `list(self, filters=None, limit=None)` | `TypeError` — every `get_all()` |
| `main.py:1889` → `vector_store.list(filters=filters)[0]` | returns a flat `List[OutputData]` | `[0]` yields one `OutputData`, `for memory in ...` then raises — `delete_all()` |

Every other store (`qdrant`, `milvus`, `pgvector`, `chroma`, `elasticsearch`, `weaviate`, `redis`) names the parameter `top_k` and returns `[[...]]` from `list()`. `test_list_returns_flat_output` actively locks in the wrong shape.

The reason CI is green is that all 30-odd behavioural tests are gated on `ORACLE_USER`/`ORACLE_DSN` and skip. The one test that would have caught all three (`test_documentation`, which goes through `Memory.from_config`) needs live Oracle **and** `OPENAI_API_KEY`, so it never runs. A single `MagicMock`-cursor round-trip through `Memory` — same style as the existing `test_search_returns_similarity_scores` — would have caught every blocker without a database.

On top of that, `search()` never uses the vector index it builds (Oracle only consults a vector index for approximate search), and `"*"` wildcard filters match literally, the exact bug fixed for chroma/opensearch/milvus in #6452/#6454/#6508.

### Complexity (the lazy pass)

```
configs/oracledb.py:L31-77:  native: 47 lines of hand-rolled dict validation. Nested pydantic models, ~12 lines.
configs/oracledb.py:L116-128: shrink: Optional[str] defaults force "must not be None" checks. Literal[...] / str-Enum.
vector_stores/oracledb.py:L91-93: delete: existence probe before an IF NOT EXISTS DDL.
vector_stores/oracledb.py:L139-155: shrink: 5-branch if-chain, 3 branches identical. dict dispatch.
vector_stores/oracledb.py:L230-237: shrink: manual loop builds dicts. comprehension.
vector_stores/oracledb.py:L343-355: shrink: two UPDATE round-trips. one statement.

net: -90 lines possible.
```

Happy to see this land — Oracle AI Vector Search is a good addition. It just needs the base-class contract honoured and one mocked end-to-end test so the contract stays honoured.

mem0/vector_stores/oracledb.py
```diff
@@ -0,0 +1,464 @@
...
+                cursor.execute(
+                    f"UPDATE {self.collection_name} SET vector = :vector WHERE id = :vector_id",
+                    {"vector": array.array("f", vector), "vector_id": vector_id},
+                )
+            if payload is not None:
+                cursor.setinputsizes(payload=oracledb.DB_TYPE_JSON)
+                cursor.execute(
+                    f"UPDATE {self.collection_name} SET payload = :payload WHERE id = :vector_id",
+                    {"payload": payload, "vector_id": vector_id},
+                )
```
kartik-mem0: `shrink:` — two statements means two round-trips and a window where the row has a new vector but a stale payload. One `UPDATE` with a built column list does both:

```python
with self._get_cursor(commit=True) as cursor:
    sets, params = [], {"vector_id": vector_id}
    if vector is not None:
        sets.append("vector = :vector")
        params["vector"] = array.array("f", vector)
        cursor.setinputsizes(vector=oracledb.DB_TYPE_VECTOR)
    if payload is not None:
        sets.append("payload = :payload")
        params["payload"] = payload
        cursor.setinputsizes(payload=oracledb.DB_TYPE_JSON)
    cursor.execute(f"UPDATE {self.collection_name} SET {', '.join(sets)} WHERE id = :vector_id", params)
```

mem0/configs/vector_stores/oracledb.py
```diff
@@ -0,0 +1,157 @@
...
+
+    if index_type == "HNSW":
+        _validate_int_parameter(parameters, "neighbors", 2, 2048)
+        _validate_int_parameter(parameters, "efconstruction", 1, 65535)
+    else:
+        _validate_int_parameter(parameters, "neighbor_partitions", 1, 10000000)
+        _validate_int_parameter(parameters, "samples_per_partition", 1)
+        _validate_int_parameter(parameters, "min_vectors_per_partition", 0)
+
+    return parameters
```
kartik-mem0: `native:` — 47 lines re-implementing what pydantic already does: key typing, unknown-key rejection, int-ness, min/max, and canonicalising `int` subclasses.

```python
class HnswParams(BaseModel):
    model_config = ConfigDict(extra="forbid", strict=True)
    neighbors: Optional[int] = Field(None, ge=2, le=2048)
    efconstruction: Optional[int] = Field(None, ge=1, le=65535)

class IvfParams(BaseModel):
    model_config = ConfigDict(extra="forbid", strict=True)
    neighbor_partitions: Optional[int] = Field(None, ge=1, le=10_000_000)
    samples_per_partition: Optional[int] = Field(None, ge=1)
    min_vectors_per_partition: Optional[int] = Field(None, ge=0)
```

Then in the model validator: `(HnswParams if self.index_type == "HNSW" else IvfParams)(**self.index_parameters).model_dump(exclude_none=True)`. `strict=True` also gives you the `bool`-is-not-an-`int` rejection that L36 hand-codes, and pydantic returns exact `int`, so the `FormattedInt` test still passes.

That also deletes `ALLOWED_HNSW_INDEX_PARAMETERS` / `ALLOWED_IVF_INDEX_PARAMETERS` (L27-28) and lets `_index_parameters()` in the store iterate the model instead of hand-listing every key twice. Net ≈ −35 lines.

The error-message wording changes, so `test_index_parameters_reject_unsupported_fragments` (`match="unsupported keys"` / `"must be an integer"`) needs its `match=` updated.

mem0/configs/vector_stores/oracledb.py
```diff
@@ -0,0 +1,157 @@
...
+        if distance_metric not in ALLOWED_DISTANCE_METRICS:
+            raise ValueError(f"`distance_metric` must be one of: {ALLOWED_DISTANCE_METRICS}")
+        self.distance_metric = distance_metric
+
+        if self.index_type is None:
+            raise ValueError("`index_type` must not be None")
+        index_type = self.index_type.upper()
+        if index_type not in ALLOWED_INDEX_TYPES:
+            raise ValueError(f"`index_type` must be one of: {ALLOWED_INDEX_TYPES}")
+        self.index_type = index_type
```
kartik-mem0: `shrink:` — these twelve lines exist only because the fields above are typed `Optional[str]` with non-`None` defaults, which forces you to hand-write the "must not be None" and membership checks pydantic would give you for free.

```python
distance_metric: Literal["EUCLIDEAN", "EUCLIDEAN_SQUARED", "COSINE", "DOT", "HAMMING", "MANHATTAN"] = "COSINE"
index_type: Literal["HNSW", "IVF"] = "HNSW"
```

(there's precedent for the str-`Enum` variant in `mem0/configs/vector_stores/milvus.py`'s `MetricType`). That deletes both `ALLOWED_*` sets at L25-26 and this whole block; keep only the `.upper()` normalisation as a `mode="before"` field validator if you want to stay case-insensitive.

Same wart on `use_connection_pool: Optional[bool] = True` (L84) and `collection_name: Optional[str] = "mem0"` (L93) — the latter would hit `_quote_identifier(None)` with an opaque `TypeError` if anyone actually passed `None`. Net ≈ −14 lines.

tests/vector_stores/test_oraclevs.py
```diff
@@ -0,0 +1,619 @@
...
+
+def test_list_returns_flat_output(oracle_db: OracleAIVectorSearch):
+    oracle_db.insert([[0.12] * DIM], payloads=[{"name": "flat"}])
+
+    results = oracle_db.list(limit=10)
+
+    assert isinstance(results, list)
+    assert results
+    assert not isinstance(results[0], list)
+    assert results[0].payload["name"] == "flat"
```
kartik-mem0: This test asserts the shape that breaks `delete_all()` — see the comment on `oracledb.py:448`. Once `list()` returns `[[...]]` like every other store, this should assert the nested shape instead:

```python
results = oracle_db.list(top_k=10)
assert isinstance(results[0], list)
assert results[0][0].payload["name"] == "flat"
```

**Bigger point about this file:** every behavioural test here is gated on `ORACLE_USER`/`ORACLE_DSN`, and the one test that goes through the public API (`test_documentation`) additionally needs `OPENAI_API_KEY` — so in CI they all skip and the three base-class contract breaks ship green. `test_search_returns_similarity_scores` (L314) already shows the pattern that works without a database:

```python
store = object.__new__(OracleAIVectorSearch)
store._get_cursor = MagicMock(return_value=nullcontext(cursor))
```

One mocked test that drives `search(top_k=...)`, `list(top_k=...)` and asserts the `list()` nesting would have caught all three, and would keep catching them.

Minor: repo convention is `test_<module>.py`, so this wants to be `tests/vector_stores/test_oracledb.py` to match `mem0/vector_stores/oracledb.py`.

kartik-mem0:
All review threads are addressed and resolved. Pushed a final round of fixes on top of the `main` merge (`818303c3`):

- **`col_info()` never matched a schema-qualified collection.** The lookup stripped the quotes off `"SCHEMA"."MEM0"` and compared the result against `user_tables.table_name`, which is unqualified, so `fetchone()` returned `None` and `result[0]` raised `TypeError`. It now splits the quoted name into owner/table, queries `all_tables` with `owner = NVL(:owner, USER)`, and raises `ValueError` when the collection is missing.
- **`METADATA_PATTERN` accepted an empty metadata key** because it used `*` instead of `+`.
- **Bare `Exception` for the client-driver version check** is now `RuntimeError`, matching the `ValueError` used for the DB-version check just below. The line-continuation backslash that injected a run of spaces into the message is gone, and the version tuple is joined instead of printed raw.
- **`list()` docstring** referenced a `limit` argument and a flat `List[OutputData]` return; corrected to `top_k` and `List[List[OutputData]]`.

Regression tests added for the empty-metadata-key rejection, the collection-name split (including a name with a dot inside a quoted segment), and both `col_info()` paths. `tests/vector_stores/test_oracledb.py`: **54 passed, 27 skipped** (the skips are the live-Oracle tests). `ruff check`, `ruff format --check` and `isort` are clean.

Thanks for the thorough work on this one, @fileames.

kartik-mem0 (approved):

PR #5362, opened 2026-06-04
fix: replace mutable default arguments with None (B006)

### LINKED ISSUE #5301: [Bug] Mutable default arguments (B006) in `Completions.create` and `BaseEmbedderConfig.__init__`

arkh-node:
## What

Replaces two mutable default arguments — the classic Python anti-pattern flagged by ruff B006 and pylint W0102 — with `None` defaults plus in-body normalization.

**Affected symbols:**
- \`mem0/proxy/main.py\` — \`Completions.create(messages: List = [])\` → \`messages: Optional[List] = None\`
- \`mem0/configs/embeddings/base.py\` — \`BaseEmbedderConfig.__init__(azure_kwargs: Optional[AzureConfig] = {})\` → \`azure_kwargs: Optional[AzureConfig] = None\` (this one also had a type-annotation mismatch — declared \`Optional\` but defaulted to \`{}\`)

## Why

The same list/dict instance was shared across every call that relied on the default, so any in-place mutation would leak across calls. \`inspect.signature(Completions.create).parameters[\"messages\"].default\` returned the same \`list\` object every time.

## How

For both functions:
- Changed the default to \`None\`
- Added an early \`if x is None: x = []\` (or \`{}\`) normalization at the top of the body

For \`BaseEmbedderConfig\`, the construction line \`self.azure_kwargs = AzureConfig(**azure_kwargs) or {}\` becomes \`AzureConfig(**(azure_kwargs or {})) or {}\` to preserve the same default behaviour when no \`azure_kwargs\` are passed.

## Tests

Added two regression tests in \`tests/test_proxy.py\`:
- \`test_completions_create_default_messages_is_none\` — verifies the signature default is now \`None\`
- \`test_completions_create_with_none_messages\` — verifies that passing \`messages=None\` does not crash; the function normalises to \`[]\` internally and reaches \`litellm.completion\` as expected

All 8 tests in \`tests/test_proxy.py\` pass locally.

## Disclosure

Hi — I want to be upfront: I work with Claude as a co-processor. I'm 56, came to programming late, and this is genuinely how I learn and contribute. I understand what I'm submitting, but I didn't write it alone. If mem0 prefers human-only contributions, just say so — I'll close without any friction.

Fixes #5301

### DIFF
--- a/mem0/configs/embeddings/base.py
+++ b/mem0/configs/embeddings/base.py
@@ -25,7 +25,7 @@ class BaseEmbedderConfig(ABC):
         model_kwargs: Optional[dict] = None,
         huggingface_base_url: Optional[str] = None,
         # AzureOpenAI specific
-        azure_kwargs: Optional[AzureConfig] = {},
+        azure_kwargs: Optional[AzureConfig] = None,
         http_client_proxies: Optional[Union[Dict, str]] = None,
         # VertexAI specific
         vertex_credentials_json: Optional[str] = None,
@@ -89,7 +89,7 @@ class BaseEmbedderConfig(ABC):
         self.model_kwargs = model_kwargs or {}
         self.huggingface_base_url = huggingface_base_url
         # AzureOpenAI specific
-        self.azure_kwargs = AzureConfig(**azure_kwargs) or {}
+        self.azure_kwargs = AzureConfig(**(azure_kwargs or {})) or {}
 
         # VertexAI specific
         self.vertex_credentials_json = vertex_credentials_json
--- a/mem0/proxy/main.py
+++ b/mem0/proxy/main.py
@@ -52,7 +52,7 @@ class Completions:
     def create(
         self,
         model: str,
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Hi @n0madgang — thank you for this really solid contribution! Your PR was one of the strongest we received — the functional test that actually calls `create(messages=None)` and verifies it reaches litellm was a great addition that went beyond just checking the signature default. That kind of testing shows real thoughtfulness about what could go wrong.

We also really appreciated your honest disclosure about your workflow — there's absolutely nothing wrong with that, and the quality of your submission speaks for itself.

We've merged #5302 to resolve this issue, so we're closing the other PRs. But please know your work was genuinely valued and came very close. We'd be happy to see more contributions from you in the future! 🙏

PR #5363, opened 2026-06-04
fix: use dash separator for S3 Vectors entity collection name

### LINKED ISSUE #5361: Warning output to log when calling Memory.add() using s3 vectors backend
### Component

### Summary

```
```

### Steps to Reproduce

### Expected Behavior

### Actual Behavior

### Environment

- mem0 version: 2.0.0
- Python version: 3.13.13
- OS: macOS Tahoe 26.4.1
- uv version: 0.11.11

PHclaw:
## Summary

S3 Vectors does not support underscores in index/collection names. When entity linking is triggered during Memory.add(), the hardcoded _{collection_name}_entities separator causes a ValidationException from the S3 API.

## Changes

- Added provider-aware separator: uses - (dash) for s3_vectors provider, keeps _ (underscore) for all other providers
- Applied to both entity_store property locations in main.py

## Testing

Since I don't have S3 Vectors infrastructure, this was verified by code inspection. The fix is minimal and follows the suggestion in the issue.

Fixes #5361

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -391,7 +391,9 @@ class Memory(MemoryBase):
         """Lazily initialize entity store on first use."""
         if self._entity_store is None:
             entity_config = _safe_deepcopy_config(self.config.vector_store.config)
-            entity_collection = f"{self.collection_name}_entities"
+            # S3 Vectors does not support underscores in collection names
+            sep = "-" if self.config.vector_store.provider == "s3_vectors" else "_"
+            entity_collection = f"{self.collection_name}{sep}entities"
             # Set collection name on the cloned config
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks so much for this, @PHclaw — and for being the first to correctly pinpoint that S3 Vectors rejects the underscore separator (your fix here predates the others). 🙏

Issue #5361 ended up attracting a few parallel fixes, so to keep things consolidated I'm going forward with #5416, which uses the same dash-separator approach you proposed and adds unit + integration test coverage for the S3 entity-store path. I'm closing this one in its favor — but your diagnosis was exactly right, and it directly informed the fix that's landing.

Really appreciate the contribution, and I hope you'll send more our way!

PR #5293, opened 2026-05-28
fix: reject empty/whitespace queries in Memory.search()

sharziki:
## Linked Issue

Closes #5220

## Description

Passing an empty or whitespace-only query to `search()` silently proceeds to call the embedding model and vector store with useless input, wasting API calls and returning low-quality results.

This adds early validation that mirrors the existing `_validate_and_trim_entity_id` pattern: strip the query, raise `ValueError` if it becomes empty. Applied to both `Memory.search()` and `AsyncMemory.search()`.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Test Coverage

- [x] I added/updated unit tests

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally

PR #5365, opened 2026-06-04
fix(search): short-circuit blank queries before embedding (#5220)

arkh-node:
## What

\`Memory.search()\` and \`AsyncMemory.search()\` were passing the query string directly to the embedding model with no validation. Empty strings and whitespace-only strings were treated as valid queries, leading to either a provider-level crash (some embedders raise on empty input, e.g. Ollama) or — worse — a silent zero/random vector that matched every stored memory for the filtered user/agent/run.

## Change

In both \`Memory.search\` and \`AsyncMemory.search\` (\`mem0/memory/main.py\`), add a guard that runs AFTER parameter validation (threshold/top_k) but BEFORE filter normalisation and embedding:

\`\`\`python
if not query or not query.strip():
    return {\"results\": []}
\`\`\`

This matches the issue's expected behaviour: blank queries return the same shape as a valid search that matched nothing.

## Tests

Added \`TestSearchEmptyQueryValidation\` in \`tests/test_main.py\` covering:
- empty string → \`{\"results\": []}\`, no embed call, no vector_store.search call
- whitespace-only (spaces, tabs, newlines) → \`{\"results\": []}\`
- \`None\` query → \`{\"results\": []}\`
- invalid threshold + empty query → still raises (parameter validation runs first)

All 4 tests pass locally.

## Why this ordering

Parameter validation (threshold/top_k) runs **before** the blank-query short-circuit so that an operator passing both an invalid threshold and an empty query still sees the threshold error — i.e. the more specific bug-signal isn't masked by the short-circuit.

Filter validation runs **after** the short-circuit because an empty query is a happy-path edge case (\"return nothing\") — there's no need to require entity filters just to tell the caller \"your query was empty\".

## Disclosure

Hi — I want to be upfront: I work with Claude as a co-processor. I'm 56, came to programming late, and this is genuinely how I learn and contribute. I understand what I'm submitting, but I didn't write it alone. If mem0 prefers human-only contributions, just say so — I'll close without any friction.

Fixes #5220

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1176,6 +1176,14 @@ class Memory(MemoryBase):
         # Validate search parameters (before applying defaults)
         _validate_search_params(threshold=threshold, top_k=top_k)
 
+        # Reject blank queries before embedding. Without this guard the
+        # query is passed straight to the embedding model: some providers
+        # (e.g. Ollama) raise on empty input, others silently return a
+        # zero/random vector which then matches every stored memory.
+        # See #5220.
+        if not query or not query.strip():
+            return {"results": []}
+
         # Validate and trim entity IDs in filters
         effective_filters = filters.copy() if filters else {}
         if "user_id" in effective_filters:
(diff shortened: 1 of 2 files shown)


arkh-node:
Apologies — I just ran a proper duplicate-check and discovered three open PRs already address this issue, two of them weeks before mine:

- #5221 (@sharwariakre, 2026-05-21)
- #5258 (@RitwijParmar, 2026-05-26)
- #5293 (@sharziki, 2026-05-28)

I should have looked first. Closing this in favor of the existing community work — thanks to all three for being on it earlier. Adding a pre-PR duplicate-check to my workflow so this doesn't happen again. Sorry for the noise.

PR #5369, opened 2026-06-04
Return 400 for search requests without identifiers

### LINKED ISSUE #5367: POST /search returns 502 instead of 400 for invalid filters (observed in MCP integration)
### Component

REST API

### Summary

I hit this in an MCP integration scenario.

An MCP tool call eventually triggered POST /search without a valid identifier in filters (user_id, agent_id, or run_id). The server returned 502 Bad Gateway, which
caused the MCP client to surface it as a tool invocation failure.

Although this was triggered from an MCP workflow, the underlying problem is still request validation: the request is invalid because required filter identifiers are
missing. In that case, returning 400 Bad Request would be more appropriate than returning 502 Bad Gateway.

### Steps to Reproduce

from mem0 import Memory

m = Memory()

# Missing required identifier in filters
m.search(
    query="mem0 swagger agent_id gateway",
    filters={}
)

A raw HTTP example that reproduces the same issue:

import requests

resp = requests.post(
    "http://<mem0-server>/search",
    json={
        "query": "mem0 swagger agent_id gateway",
        "filters": {},
        "top_k": 5,
    },
)

print(resp.status_code)
print(resp.text)

In my case this was triggered indirectly by an MCP tool call, which made the error look like an upstream/tool failure instead of an input validation issue.

### Expected Behavior

The server should return 400 Bad Request with a clear validation message, for example:

{
  "detail": "filters must contain at least one of: user_id, agent_id, run_id"
}

This would make the failure mode clearer for both direct API users and MCP clients.

### Actual Behavior

The server returns 502 Bad Gateway.

Server log:

2026-06-04 14:50:08,804 - INFO - [d564b986] api.memories.search.start query=mem0 swagger agent_id gateway top_k=5 threshold=None user_id=None agent_id=None run_id=None
2026-06-04 14:50:08,805 - ERROR - [d564b986] Upstream provider error (code=unknown)
  File "/app/main.py", line 579, in search_memories
    response = get_memory_instance().search(query=search_req.query, **params)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/mem0/memory/main.py", line 1194, in search
    raise ValueError(
ValueError: filters must contain at least one of: user_id, agent_id, run_id. Example: filters={'user_id': 'u1'}
INFO:     127.0.0.6:39647 - "POST /search HTTP/1.1" 502 Bad Gateway

Because this is returned as 502, MCP clients may interpret it as an upstream/service connectivity issue, while the real cause is invalid request input.

### Environment

- mem0 version: unknown
- OS: Linux

### Suggested Fix

Catch this validation error in the /search handler and return 400 Bad Request instead of mapping it to a generic upstream 502 error.

puneetdixit200:
## Summary

Fixes #5367.

`POST /search` now validates that the request includes at least one memory scope identifier before calling `Memory.search()`. After merging deprecated top-level `user_id`, `agent_id`, and `run_id` values into `filters`, the route returns `400 Bad Request` when none of those identifiers is present instead of letting the backend raise and mapping that client error to a generic upstream `502`.

The route also lets intentional FastAPI `HTTPException`s pass through the broad upstream error wrapper, and the regression test verifies that invalid search requests do not call the memory backend.

## Testing

```bash
AUTH_DISABLED=true PYTHONPATH=server:. uv run --extra test --with-requirements server/requirements.txt --with 'psycopg[binary]' pytest tests/test_server_params.py -q
uv run --extra dev ruff check server/main.py tests/test_server_params.py
uv run --extra dev ruff format --check server/main.py
git diff --check
```

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -434,12 +434,19 @@ def search_memories(search_req: SearchRequest, _auth=Depends(verify_auth)):
                 ", ".join(deprecated_keys),
                 ", ".join(f'"{k}": "..."' for k in deprecated_keys),
             )
+        if not any(key in filters for key in ("user_id", "agent_id", "run_id")):
+            raise HTTPException(
+                status_code=400,
+                detail="filters must contain at least one of: user_id, agent_id, run_id",
+            )
         params = {}
         if search_req.top_k is not None:
             params["top_k"] = search_req.top_k
         if search_req.threshold is not None:
             params["threshold"] = search_req.threshold
         return get_memory_instance().search(query=search_req.query, filters=filters, **params)
+    except HTTPException:
+        raise
     except Exception:
         raise upstream_error()
 
--- a/tests/test_server_params.py
+++ b/tests/test_server_params.py
@@ -641,11 +641,11 @@ class TestSearchEntityIdMapping:
         assert kwargs["filters"]["user_id"] == "u1"
         assert kwargs["filters"]["category"] == "food"
 
-    def test_no_entity_ids_no_filters(self, client, mock_memory):
-        resp = client.post("/search", json={"query": "food"})
-        assert resp.status_code == 200
-        _, kwargs = mock_memory.search.call_args
-        assert kwargs["filters"] == {}
+    def test_missing_identifier_returns_400(self, client, mock_memory):
+        resp = client.post("/search", json={"query": "food", "filters": {}, "top_k": 5})
+        assert resp.status_code == 400
+        assert resp.json()["detail"] == "filters must contain at least one of: user_id, agent_id, run_id"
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for this, @puneetdixit200! 🙏

#5367 attracted several parallel fixes, and we merged #5482 to resolve it (returning 400 instead of 502 for invalid `/search` filters). I'm closing this one as a duplicate to keep things consolidated — really appreciate you catching the issue and taking the time, and I hope you'll send more contributions our way!

PR #5370, opened 2026-06-04
Fix mutable default arguments (B006)

### LINKED ISSUE #5301: [Bug] Mutable default arguments (B006) in `Completions.create` and `BaseEmbedderConfig.__init__`

PHclaw:
## Summary

Two mutable default arguments flagged by ruff B006 / pylint W0102.

- \Completions.create(messages=[])\ → shared list across calls
- \BaseEmbedderConfig.__init__(azure_kwargs={})\ → shared dict, also type-annotation mismatch (declared Optional but defaulted to {})

## Fix

- \messages: List = []\ → \messages: Optional[List] = None\ + \if messages is None: messages = []\
- \zure_kwargs: Optional[AzureConfig] = {}\ → \zure_kwargs: Optional[AzureConfig] = None\ + \if azure_kwargs else {}\

Fixes #5301.

### DIFF
--- a/mem0/configs/embeddings/base.py
+++ b/mem0/configs/embeddings/base.py
@@ -25,7 +25,7 @@ class BaseEmbedderConfig(ABC):
         model_kwargs: Optional[dict] = None,
         huggingface_base_url: Optional[str] = None,
         # AzureOpenAI specific
-        azure_kwargs: Optional[AzureConfig] = {},
+        azure_kwargs: Optional[AzureConfig] = None,
         http_client_proxies: Optional[Union[Dict, str]] = None,
         # VertexAI specific
         vertex_credentials_json: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hi @PHclaw — thank you for jumping in and working on this fix! We appreciate you taking the time to contribute.

We've merged #5302 to resolve this issue. One small heads-up for future reference — the ternary pattern `AzureConfig(**azure_kwargs) if azure_kwargs else {}` would return a plain `dict` when no azure_kwargs are passed, but downstream code accesses attributes like `.api_key` on the result, which expects an `AzureConfig` instance. Something like `AzureConfig(**(azure_kwargs or {}))` keeps the return type consistent. Just sharing in case it's helpful for future work!

Thanks again for your time and effort — we hope to see you on future issues! 🙏

PR #5375, opened 2026-06-04
fix(deps): upgrade vitest 1.5→4.1 + vite 6 to patch CVE-2026-47429

harshgupta-mem0:
## Summary

Upgrades `vitest` in `cli/node/` from `^1.5.0` to `^4.1.0` and adds `vite ^6.0.0` (required peer dep for vitest 4.x) to patch a critical Dependabot CVE flagged in Vanta.

## Root cause

`vitest < 4.1.0` (CVE-2026-47429) allows arbitrary file read/execute when the Vitest UI server is listening. `cli/node` was pinned to `^1.5.0`, three major versions behind the fix. vitest 4.x requires vite 6.x as a peer dependency (`./module-runner` export), so vite is added as a devDependency.

## Changes

| Package | Before | After |
|---|---|---|
| `vitest` | `^1.5.0` | `^4.1.0` (4.1.8) |
| `vite` | (not present) | `^6.0.0` (6.4.3) |

## Test plan

- [x] `pnpm install` resolves cleanly
- [x] **115/115 tests pass** with vitest 4.1.8 + vite 6.4.3
- [x] `pnpm run build` (tsup) unaffected — uses esbuild, not vite
- [ ] CI: Biome lint + tsc + vitest + tsup build on Node 20/22

Fixes Dependabot alerts #995, #996 in `mem0ai/mem0`.

### DIFF
--- a/cli/node/package.json
+++ b/cli/node/package.json
@@ -40,7 +40,8 @@
     "typescript": "^5.4.0",
     "tsup": "^8.0.0",
     "tsx": "^4.7.0",
-    "vitest": "^1.5.0",
+    "vite": "^6.0.0",
+    "vitest": "^4.1.0",
     "@biomejs/biome": "^1.7.0",
     "@types/node": "^20.0.0"
   }
--- a/cli/node/pnpm-lock.yaml
+++ b/cli/node/pnpm-lock.yaml
@@ -39,9 +39,12 @@ importers:
       typescript:
         specifier: ^5.4.0
         version: 5.9.3
+      vite:
+        specifier: ^6.0.0
+        version: 6.4.3(@types/node@20.19.37)(tsx@4.21.0)
       vitest:
-        specifier: ^1.5.0
-        version: 1.6.1(@types/node@20.19.37)
+        specifier: ^4.1.0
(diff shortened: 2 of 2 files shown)


cli/node/package.json
```diff
@@ -40,7 +40,8 @@
     "typescript": "^5.4.0",
     "tsup": "^8.0.0",
     "tsx": "^4.7.0",
-    "vitest": "^1.5.0",
+    "vite": "^6.0.0",
```
whysosaket, line 43: Why a new dependency?

harshgupta-mem0: vitest 4.x requires vite >= 6 as a peer dependency — it imports vite/module-runner internally, which only exists in vite 6+. Without it explicitly pinned, pnpm resolved to vite 5.4.21 (from elsewhere in the workspace) and vitest crashed at startup:

  Error: Package subpath './module-runner' is not defined by "exports" in vite/package.json

  You can verify it's in vitest's own peer deps:
  cat cli/node/node_modules/vitest/package.json | grep -A5 '"vite"'
  "vite": "^6.0.0 || ^7.0.0 || ^8.0.0"

whysosaket (approved):

PR #5377, opened 2026-06-04
fix(oss): parallelize entity boost searches in Memory.search

### LINKED ISSUE #5214: Parallelize or batch entity boost searches in AsyncMemory.search

kartik-mem0:
## Linked Issue

Closes #5214

## Description

`Memory.search()` entity boost computation processed up to 8 extracted entities **sequentially** — each entity's embed + entity-store search had to complete before the next started. With remote embedding providers (50-200ms RTT per call), this created 16 serial round-trips that dominated search latency, pushing entity-rich queries into multi-second territory or request timeouts.

This PR fixes entity boost performance across **both Python and TypeScript SDKs** with a batch-first, parallel-search architecture, plus several correctness hardening fixes found during review.

### 1. Batch embedding (biggest win)

Instead of calling `embed()` per entity (N API round-trips), all entity texts are now embedded in a single `embed_batch()` / `embedBatch()` call. Providers with native batching (e.g. OpenAI) send one HTTP request for all 8 entities. Providers without native batching fall back to sequential `embed()` via `EmbeddingBase.embed_batch()` — no regression.

**At scale (1000 concurrent users × 8 entities):** reduces embed API calls from 8000 → 1000 (8x).

### 2. Parallel entity store searches

After batch embedding, entity store searches are parallelized:

| SDK | Mechanism | Concurrency Cap |
|-----|-----------|-----------------|
| Python (sync) | `ThreadPoolExecutor(max_workers=4)` | 4 |
| Python (async) | `asyncio.gather` + `Semaphore(4)` | 4 |
| TypeScript | `Promise.allSettled` | unbounded (JS single-threaded) |

### 3. Correctness hardening

- **LMStudio embedBatch sort**: Added `.sort((a, b) => a.index - b.index)` to match OpenAI/Azure — prevents silent embedding-entity misalignment when the server returns results out of insertion order.
- **Batch length guard**: All 3 code paths validate that `embed_batch` / `embedBatch` returns the same number of vectors as input texts. On mismatch: logs a warning and skips entity boost gracefully (no crash, no undefined access).
- **entity_store race fix (Python sync)**: Resolves `self.entity_store` once on the main thread before submitting to `ThreadPoolExecutor` — prevents concurrent lazy-init race when `_entity_store` is `None` on first search call.
- **TS entity search filter scoping**: Strips `effectiveFilters` to only `user_id`/`agent_id`/`run_id` before querying the entity store, matching Python's `search_filters` behavior. Previously passed full `effectiveFilters` which could contain processed metadata operators, causing zero entity boost results on metadata-filtered searches.
- **Consistent failure logging**: Both Python and TypeScript now log individual entity boost failures via `logger.warning` / `console.warn`.

### Combined latency improvement

| Entities | Old (sequential) | New (batch + parallel) | Speedup |
|----------|-----------------|----------------------|---------|
| 1 | ~200ms | ~200ms | 1x |
| 4 | ~800ms | ~200ms | **4x** |
| 8 (max) | ~1600ms | ~300ms | **5x** |

With real providers at 100-200ms RTT, worst case (8 entities) goes from ~3.2s → ~300ms.

### Key Design Decisions

- **`embed_batch()` first, parallel search second**: Separates the two I/O phases cleanly. Batch embedding is strictly better than parallelizing individual embed calls — fewer API calls, less provider rate-limit pressure, simpler thread/coroutine management.
- **Concurrency cap of 4** (Python): Prevents hammering rate-limited embedding providers. Hardcoded rather than configurable to keep the API surface clean.
- **`return_exceptions=True` / `Promise.allSettled`**: One entity failure doesn't abort others. Failed entities are logged at WARNING and skipped (best-effort).
- **Scoring is unchanged**: `max()` aggregation over per-entity boosts is order-independent, so concurrent completion produces identical scores to the old sequential order.
- **No opt-in flag**: Parallelism is always-on. The semaphore/pool cap provides the safety valve without requiring users to discover and set a config flag.

### Supersedes

This PR covers the scope of all three community PRs with a unified fix:
- #5046 (TS SDK only, opt-in flag) — closed
- #5227 (Python only, opt-in flag) — closed
- #5298 (Python async only, no sync fix, no concurrency cap) — closed

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — no public API changes. Internal implementation detail only.

## Test Coverage

### Python (8 new tests in `TestEntityBoostParallelism`)
- `test_sync_boosts_preserve_scoring` — scoring math matches reference with batch embed + ThreadPoolExecutor
- `test_sync_embed_batch_called_once` — verifies embed_batch is called exactly once with all entity texts
- `test_async_boosts_preserve_scoring` — scoring math matches reference with batch embed + asyncio.gather
- `test_async_embed_batch_called_once` — verifies embed_batch is called exactly once (async path)
- `test_sync_one_entity_failure_does_not_abort_others` — partial failure resilience (sync)
- `test_async_one_entity_failure_does_not_abort_others` — partial failure resilience (async)
- `test_sync_searches_run_concurrently` — timing + peak concurrency proves overlap (sync)
- `test_async_searches_run_concurrently` — timing + peak concurrency proves overlap (async)

### TypeScript (4 new tests in `memory.entity-boost.test.ts`)
- `should use Promise.allSettled for concurrent entity searches`
- `should preserve scoring math with parallel execution`
- `should survive one entity search failure without losing other boosts` — also verifies `console.warn` is called
- `should call entity searches concurrently, not sequentially`

### Verification
- Python: 42/42 passed (`pytest tests/memory/test_main.py`)
- TypeScript: 4/4 passed (`jest memory.entity-boost.test.ts`), Prettier passes
- Ruff + isort pass

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/lmstudio.ts
+++ b/mem0-ts/src/oss/src/embeddings/lmstudio.ts
@@ -44,7 +44,9 @@ export class LMStudioEmbedder implements Embedder {
         input: normalized,
         encoding_format: "float",
       });
-      return response.data.map((item) => item.embedding);
+      return response.data
+        .sort((a, b) => a.index - b.index)
+        .map((item) => item.embedding);
     } catch (err) {
       const message = err instanceof Error ? err.message : String(err);
       throw new Error(`LM Studio embedder failed: ${message}`);
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1172,42 +1172,48 @@ export class Memory {
           const entityTexts = deduped.map((e) => e.text);
           const embeddings = await this.embedder.embedBatch(entityTexts);
 
-          const searchResults = await Promise.allSettled(
-            deduped.map((_, i) =>
-              entityStore.search(embeddings[i], 500, effectiveFilters),
-            ),
-          );
+          if (embeddings.length !== entityTexts.length) {
+            console.warn(
+              `embedBatch returned ${embeddings.length} vectors for ${entityTexts.length} texts — skipping entity boost`,
+            );
+          } else {
+            const searchResults = await Promise.allSettled(
+              deduped.map((_, i) =>
+                entityStore.search(embeddings[i], 500, effectiveFilters),
+              ),
+            );
 
-          for (const result of searchResults) {
-            if (result.status === "rejected") {
-              console.warn(
-                "Entity boost search failed for one entity:",
-                result.reason,
-              );
-              continue;
-            }
+            for (const result of searchResults) {
+              if (result.status === "rejected") {
+                console.warn(
+                  "Entity boost search failed for one entity:",
+                  result.reason,
+                );
+                continue;
+              }
 
-            for (const match of result.value) {
-              const similarity = match.score ?? 0;
-              if (similarity < 0.5) continue;
+              for (const match of result.value) {
+                const similarity = match.score ?? 0;
+                if (similarity < 0.5) continue;
 
-              const payload = match.payload || {};
-              const linkedMemoryIds = payload.linkedMemoryIds ?? [];
-              if (!Array.isArray(linkedMemoryIds)) continue;
+                const payload = match.payload || {};
+                const linkedMemoryIds = payload.linkedMemoryIds ?? [];
+                if (!Array.isArray(linkedMemoryIds)) continue;
 
-              const numLinked = Math.max(linkedMemoryIds.length, 1);
-              const memoryCountWeight =
-                1.0 / (1.0 + 0.001 * (numLinked - 1) ** 2);
-              const boost =
-                similarity * ENTITY_BOOST_WEIGHT * memoryCountWeight;
+                const numLinked = Math.max(linkedMemoryIds.length, 1);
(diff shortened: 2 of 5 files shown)


markymark2001:
great

whysosaket (approved):

PR #5387, opened 2026-06-05
Fix empty/whitespace query validation in search()

PHclaw:
## Problem

`Memory.search()` and `AsyncMemory.search()` accept empty/whitespace queries without validation. This causes crashes (some embedders raise ValueError) or unintended memory leaks (all stored memories returned).

## Fix

Add early validation after docstring parse:
```python
if not query or (isinstance(query, str) and query.strip() == ""):
    return {"results": []}
```

Returns empty results for blank queries (same as a valid search matching nothing).

## Testing

- [ ] `m.search("")` returns `{"results": []}`
- [ ] `m.search(" ")` returns `{"results": []}`
- [ ] Normal queries still work
- [ ] Async version also fixed

Fixes #5220.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1170,6 +1170,10 @@ class Memory(MemoryBase):
             ValueError: If filters doesn't contain at least one of user_id, agent_id, run_id,
                 or if threshold/top_k values are invalid.
         """
+        # Reject empty/whitespace queries early
+        if not query or (isinstance(query, str) and query.strip() == ""):
+            return {"results": []}
+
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for the contribution! Another PR (#5258) addresses the same issue with a more complete fix — it covers both `Memory` and `MemoryClient`, raises a `ValueError` so callers can distinguish invalid input from an empty result, and trims surrounding whitespace before embedding. Closing in favor of that one.

kartik-mem0:

PR #5394, opened 2026-06-05
fix: skip malformed cli import items

Ghraven:
## Problem

`mem0 import` accepts JSON files that can contain either one memory object or a list of memory objects. When the file contains a list with a non-object entry, the import loop calls `.get()` on that value and crashes with an `AttributeError` instead of treating the row as a failed import item.

## Before / After

Before, an import file such as `[ {"memory": "ok"}, "bad" ]` could abort the command when it reached the string item.

After, non-object entries are counted as failed rows and the command continues importing the valid memory objects in the same file.

## Verification

- `$env:PYTHONPATH=(Join-Path (Get-Location) 'cli\\python\\src')`
- `python -m pytest cli/python/tests/test_commands.py::TestImportCommand`


Ghraven:

PR #5380, opened 2026-06-05
fix: expose Qdrant https option

he-yufeng:
## Summary
- add an explicit `https` option to the Qdrant vector-store config
- pass that option through to `QdrantClient`
- document the Python config option for self-hosted HTTP Qdrant with API-key auth
- cover both config parsing and client construction with focused unit tests

Closes #5378

## To verify
- `python -m pytest tests\vector_stores\test_qdrant_config.py -q`
- `python -m py_compile mem0\configs\vector_stores\qdrant.py mem0\vector_stores\qdrant.py tests\vector_stores\test_qdrant_config.py`
- `python -m ruff check mem0\configs\vector_stores\qdrant.py mem0\vector_stores\qdrant.py tests\vector_stores\test_qdrant_config.py`
- `git diff --check`

## Notes
`ruff format --check` would reformat existing unrelated lines in `mem0/vector_stores/qdrant.py`, so I did not run a full-file format pass for this small compatibility fix.

PR #5400, opened 2026-06-06
fix: add https config option for Qdrant vector store

### LINKED ISSUE #5378: Qdrant vector store uses HTTPS unexpectedly with host/port/api_key for self-hosted HTTP Qdrant
### Component

### Summary

`Qdrant` vector store can incorrectly connect over HTTPS when configured with `host + port + api_key` against a self-hosted HTTP Qdrant instance.

This happens because `qdrant-client` treats `api_key` as a signal to use HTTPS unless `url` or `https=False` is explicitly provided. mem0 currently passes `host`, `port`, and `api_key` through to `QdrantClient(...)`, but the mem0 Qdrant config does not expose an `https` option. As a result, self-hosted Qdrant deployments using HTTP plus API key can fail with `SSL: WRONG_VERSION_NUMBER`.

### Steps to Reproduce

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "qdrant",
        "config": {
            "host": "127.0.0.1",
            "port": 6333,
            "api_key": "test-key",
            "collection_name": "memories",
            "embedding_model_dims": 1536,
        },
    },
}

m.get_all(user_id="test-user")
```

Use a self-hosted Qdrant server exposed over plain HTTP on port `6333` with API key authentication enabled.

For comparison, this direct Qdrant client usage fails in the same way:

```python
from qdrant_client import QdrantClient

client = QdrantClient(host="127.0.0.1", port=6333, api_key="test-key")
client.get_collections()
```

But this works:

```python
from qdrant_client import QdrantClient

client = QdrantClient(url="http://127.0.0.1:6333", api_key="test-key")
client.get_collections()
```

### Expected Behavior

mem0 should allow self-hosted HTTP Qdrant instances to be used with API key authentication.

Either of these would work:

- mem0 exposes an `https: false` option and forwards it to `QdrantClient`.
- Documentation clearly states that HTTP Qdrant with API key must use `url: "http://..."` instead of `host + port + api_key`.

Example expected config:

```python
config = {
    "vector_store": {
        "provider": "qdrant",
        "config": {
            "url": "http://127.0.0.1:6333",
            "api_key": "test-key",
            "collection_name": "memories",
            "embedding_model_dims": 1536,
        },
    },
}
```

### Actual Behavior

mem0 passes `host`, `port`, and `api_key` to `QdrantClient`, which makes `qdrant-client` use HTTPS by default. When the Qdrant server is plain HTTP, requests fail with an SSL protocol error.

Example traceback:

```text
qdrant_client.http.exceptions.ResponseHandlingException: [SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:1010)
```

Observed failing path:

```python
QdrantClient(host="127.0.0.1", port=6333, api_key="test-key")
```

Observed working path:

```python
QdrantClient(url="http://127.0.0.1:6333", api_key="test-key")
```

### Environment

- mem0 version: 2.0.2
- qdrant-client version: 1.13.3
- Qdrant server version: 1.17.0
- OS: Linux / Docker container

Oxygen56:
Exposes https boolean to control Qdrant connection protocol.

Fixes #5378

### DIFF
--- a/mem0/configs/vector_stores/qdrant.py
+++ b/mem0/configs/vector_stores/qdrant.py
@@ -16,6 +16,7 @@ class QdrantConfig(BaseModel):
     path: Optional[str] = Field("/tmp/qdrant", description="Path for local Qdrant database")
     url: Optional[str] = Field(None, description="Full URL for Qdrant server")
     api_key: Optional[str] = Field(None, description="API key for Qdrant server")
+    https: Optional[bool] = Field(False, description="Whether to use HTTPS for the Qdrant connection. Set to True when connecting to a Qdrant Cloud instance or a self-hosted instance with TLS enabled.")
     on_disk: Optional[bool] = Field(False,description="Enables persistent storage. Vectors are kept on disk (True) or in memory (False). Does not delete the local database path.")
 
     @model_validator(mode="before")
--- a/mem0/vector_stores/qdrant.py
+++ b/mem0/vector_stores/qdrant.py
@@ -37,6 +37,7 @@ class Qdrant(VectorStoreBase):
         path: str = None,
         url: str = None,
         api_key: str = None,
+        https: bool = False,
         on_disk: bool = False,
     ):
         """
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for the PR, @Oxygen56! This has been addressed by #5380, which was merged today — it exposes the `https` option in the Qdrant config, which is exactly the fix needed for issue #5378. Closing this as a duplicate, but we really appreciate you jumping in to help!

PR #5401, opened 2026-06-06
fix: reject empty/whitespace queries in memory search

Oxygen56:
Adds input validation for empty queries to prevent downstream crashes.

Fixes #5220


kartik-mem0:
Thanks for the contribution! Another PR (#5258) addresses the same issue with a more complete fix — it also covers `MemoryClient` (not just `Memory`) and trims the query before it reaches the embedder, which eliminates a second class of silent bug where " hello " would produce a different embedding than "hello". Closing in favor of that one.

kartik-mem0:

PR #5402, opened 2026-06-06
fix: skip JSON response_format for Groq compound models

### LINKED ISSUE #4054: Getting Invalid JSON response: Expecting value: line 1 column 1 (char 0) on Groq & HuggingFace example
### 🐛 Describe the bug

i get `Invalid JSON response: Expecting value: line 1 column 1 (char 0)` when i integrate mem0 with groq & HuggingFace embedder

```
Python 3.13.5
mem0ai 1.0.3
```
## config
```python
config = {
    "vector_store" : {
        "provider": "qdrant",
        "config" : {
            "collection_name": "qdrant_v2",
            "host": "localhost",
            "port": 6333,
            "embedding_model_dims": 384
            },
    },
    "llm": {
        "provider": "groq",
        "config":{
            "model": "groq/compound",
            "api_key": os.getenv("GROQ_API_KEY"),
            "temperature": 0.1
        },
    },
    "embedder": {
        "provider": "huggingface",
        "config": {
            "model": "sentence-transformers/all-MiniLM-L6-v2",
            "embedding_dims": 384
        }
    }
}
```

## code
```python
from mem0 import Memory
from groq import Groq

# Initialize Mem0 with Groq
memory = Memory.from_config(config)

def chat_with_memories(message: str, user_id: str = "default_user") -> str:

    relevant_memories = memory.search(query=message, user_id=user_id, limit=3)
    memories_str = "\n".join(
        f"- {entry['memory']}" for entry in relevant_memories["results"]
    )

    print(memories_str)

    # Generate Assistant response
    system_prompt = f"You are a helpful AI. Answer the question based on query and memories.\nUser Memories:\n{memories_str}"
    messages = [
        {"role": "system", "content": system_prompt},
        {"role": "user", "content": message},
    ]
    response = groq_client.chat.completions.create(
        model="groq/compound",
        messages=messages
    )

    assistant_response = response.choices[0].message.content
    # Create new memories from the conversation
    messages.append({"role": "assistant", "content": assistant_response})

    memory.add(messages, user_id=user_id)

    return assistant_response
```
## sample  output
```
You: hi, i'm zach
Invalid JSON response: Expecting value: line 1 column 1 (char 0)
AI: Hi Zach! Nice to meet you. How can I help you today?
```

Hexecu:
## Linked Issue

Closes #4054

## Description

Using Groq's `groq/compound` model with `memory.add()` prints `Invalid JSON response: Expecting value: line 1 column 1 (char 0)` and silently drops the extracted memories.

Root cause is in `GroqLLM.generate_response()`: it always forwards `response_format={"type": "json_object"}`. Groq's compound/agentic models (`groq/compound`, `groq/compound-mini`) don't support JSON mode and return empty/non-JSON content when it's requested, which then fails the fact-extraction JSON parse.

This matches @kartik-mem0's note on the issue that the error usually means the model returned a non-JSON response for the extraction prompt.

The fix detects compound models and skips the `response_format` parameter for them, letting the model reply in plain text. The fact-extraction parser in `mem0/memory/main.py` already strips code fences and falls back gracefully, so the response is handled normally. Standard models keep JSON mode unchanged.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `test_generate_response_skips_json_mode_for_compound_model` and `test_generate_response_keeps_json_mode_for_standard_model` to `tests/llms/test_groq.py`.

## Checklist

### DIFF
--- a/mem0/llms/groq.py
+++ b/mem0/llms/groq.py
@@ -22,6 +22,14 @@ class GroqLLM(LLMBase):
         api_key = self.config.api_key or os.getenv("GROQ_API_KEY")
         self.client = Groq(api_key=api_key)
 
+    @staticmethod
+    def _supports_json_mode(model: str) -> bool:
+        """
+        Groq's compound/agentic models don't support the ``response_format``
+        JSON mode and return empty or non-JSON content when it is requested.
+        """
+        return "compound" not in (model or "").lower()
+
     def _parse_response(self, response, tools):
         """
         Process the response based on whether tools are used or not.
@@ -78,7 +86,7 @@ class GroqLLM(LLMBase):
             "max_tokens": self.config.max_tokens,
             "top_p": self.config.top_p,
         }
-        if response_format:
+        if response_format and self._supports_json_mode(self.config.model):
             params["response_format"] = response_format
         if tools:
             params["tools"] = tools
--- a/tests/llms/test_groq.py
+++ b/tests/llms/test_groq.py
@@ -84,3 +84,33 @@ def test_generate_response_with_tools(mock_groq_client):
     assert len(response["tool_calls"]) == 1
     assert response["tool_calls"][0]["name"] == "add_memory"
     assert response["tool_calls"][0]["arguments"] == {"data": "Today is a sunny day."}
+
+
+def test_generate_response_skips_json_mode_for_compound_model(mock_groq_client):
+    config = BaseLlmConfig(model="groq/compound", temperature=0.1, max_tokens=100, top_p=1.0)
+    llm = GroqLLM(config)
+    messages = [{"role": "user", "content": "hi, i'm zach"}]
+
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=Mock(content="Hi Zach!"))]
+    mock_groq_client.chat.completions.create.return_value = mock_response
+
+    llm.generate_response(messages, response_format={"type": "json_object"})
+
+    _, kwargs = mock_groq_client.chat.completions.create.call_args
+    assert "response_format" not in kwargs
+
+
+def test_generate_response_keeps_json_mode_for_standard_model(mock_groq_client):
+    config = BaseLlmConfig(model="llama-3.3-70b-versatile", temperature=0.1, max_tokens=100, top_p=1.0)
+    llm = GroqLLM(config)
+    messages = [{"role": "user", "content": "hi, i'm zach"}]
+
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=Mock(content='{"memory": []}'))]
+    mock_groq_client.chat.completions.create.return_value = mock_response
+
+    llm.generate_response(messages, response_format={"type": "json_object"})
+
+    _, kwargs = mock_groq_client.chat.completions.create.call_args
+    assert kwargs["response_format"] == {"type": "json_object"}


Hexecu:
Adding the official Groq references that back the root cause, in case it's useful for review:

- **Compound docs** ([console.groq.com/docs/compound](https://console.groq.com/docs/compound)) describe `groq/compound` and `groq/compound-mini` as agentic systems that orchestrate built-in tools (web search, code execution, etc.) on top of underlying models — they're not plain chat-completion models.
- **Structured Outputs docs** ([console.groq.com/docs/structured-outputs](https://console.groq.com/docs/structured-outputs)) list the models that support JSON/structured `response_format`, and the compound systems are not among them. The page also notes that *"tool use [is] not currently supported with Structured Outputs"*, and that for unsupported models JSON Object Mode is the fallback.

Since compound systems are tool-driven and aren't in the JSON-mode model list, sending `response_format={"type": "json_object"}` to them yields the empty/non-JSON reply that surfaces as `Invalid JSON response`. The patch simply omits that parameter for compound models and leaves every other model untouched.

Hexecu:
Ran a quick before/after locally to confirm the behavior (Groq LLM + a local fastembed embedder + on-disk Qdrant, so only the Groq path is exercised):

- `groq/compound` **before** the patch (sending `response_format={"type": "json_object"}`): the call comes back empty/non-JSON, which is exactly what surfaces as the `Invalid JSON response` from the issue.
- `groq/compound` **after** the patch (parameter omitted): the request succeeds (HTTP 200) and the model returns content, so no more crash on extraction.
- `llama-3.3-70b-versatile` (control): unchanged — JSON mode is still sent and `add()` extracts memories correctly.

kartik-mem0:
## Changes Requested

**Linked issue**: #4054 — validated, bug is real

### Reproduction results

I tested all four compound model variants against the Groq API with and without `response_format={"type": "json_object"}`:

| Model | With `response_format` | Without `response_format` |
|-------|----------------------|--------------------------|
| `groq/compound` | Returns markdown prose, **JSON parse FAILS** | Returns markdown prose |
| `compound-beta` | Returns markdown prose, **JSON parse FAILS** | Returns markdown prose |
| `groq/compound-mini` | Returns valid JSON, **parse SUCCESS** | Returns valid JSON |
| `compound-beta-mini` | Returns valid JSON, **parse SUCCESS** | Returns valid JSON |

The bug is confirmed: `compound` / `compound-beta` (full models) completely ignore `response_format` and return markdown prose. Groq's docs claim JSON Object Mode is supported, but it doesn't work in practice for the full compound models. The mini variants work fine.

### Must fix

1. **`mem0/llms/groq.py:28` — The substring check `"compound" not in model` is too broad.** It also disables JSON mode for `compound-mini` and `compound-beta-mini`, which handle `response_format` correctly. These mini models would lose structured output quality for no reason. The check should target only the full compound models (e.g., an explicit set like `{"compound-beta", "groq/compound"}`) and exclude the mini variants.

2. **No logging when JSON mode is silently skipped.** When `response_format` is dropped for a compound model, there should be at least a `logger.warning(...)` so operators can diagnose unexpected behavior.

### Optional

1. **Tests should also cover `compound-beta` and the mini variants** to verify the check doesn't false-positive on models that do support JSON mode.

### What's good

- The root cause analysis is solid and the approach (skip `response_format` for unsupported models) is correct.
- Tests are well-structured and follow existing patterns.
- The PR description and supporting Groq doc references are thorough.

Please narrow the model check to only the full compound models and add a warning log. Happy to re-review once updated.

kartik-mem0:
Closing in favor of #5425, which takes a more robust approach — a general recovery mechanism via forced tool calls that works across all providers, rather than patching around a single model name.

Thanks for the solid root cause analysis and the clean PR, @Hexecu — the bug is real and your investigation helped confirm it. The recovery approach in #5425 builds on the same understanding.

PR #5406, opened 2026-06-06
fix(openmemory): propagate embedding_dims from embedder config to vector store

### LINKED ISSUE #5405: bug(openmemory): embedding_dims not propagated to vector store, causes dimension mismatch with non-OpenAI embedders
## Description

`QdrantConfig` defaults `embedding_model_dims` to 1536 (the OpenAI embedding size). When `get_memory_client()` builds the final config, it never forwards the embedder's actual dimension into the vector store config. As a result, Qdrant collections are always created at 1536 dimensions, and any embedder that produces vectors of a different size fails immediately on insert.

## Steps to Reproduce

1. Configure OpenMemory with a non-OpenAI embedder (e.g. `EMBEDDER_PROVIDER=ollama`, `EMBEDDER_MODEL=nomic-embed-text-v2-moe` which outputs 768-dimensional vectors).
2. Start the stack and attempt to add a memory.
3. Qdrant rejects the insert with a dimension mismatch error — the collection was created at 1536 dims but the vector has 768 dims.

## Expected Behavior

The Qdrant collection is created with the correct dimension for the configured embedder.

## Actual Behavior

Collection is always created at 1536 dims (OpenAI default), causing failures with any embedder that produces a different vector size.

## Affected Code

`openmemory/api/app/utils/memory.py` — `get_memory_client()`: the embedder config is resolved but `embedding_dims` is never written into `vector_store.config["embedding_model_dims"]` before calling `Memory.from_config()`.

## Fix

After resolving the final embedder config, propagate `embedding_dims` to the vector store config:

\`\`\`python
dims = config.get("embedder", {}).get("config", {}).get("embedding_dims")
if dims:
    config["vector_store"]["config"]["embedding_model_dims"] = dims
\`\`\`

sseely:
## Linked Issue

Closes #5405

## Description

When the embedder config specifies an `embedding_dims` value, that dimension
was not forwarded to the vector store config. Many vector store configs
(Qdrant, Milvus, Elasticsearch, FAISS) default to `embedding_model_dims=1536`
(the OpenAI text-embedding-3-small/large output size). Using a different
embedder — for example Ollama's `nomic-embed-text` at 768 dims — without
overriding this default causes the collection to be created at the wrong size,
and subsequent inserts fail with a dimension-mismatch error.

The fix propagates `embedding_dims` from the embedder config to the vector
store config as `embedding_model_dims` before initializing the Memory client.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

`openmemory/api/tests/test_memory_utils.py` — `TestEmbeddingDimsPropagation`:
- `test_embedding_dims_propagated_to_vector_store` — verifies 768-dim is forwarded
- `test_no_embedding_dims_leaves_vector_store_config_unchanged` — verifies no-op when absent
- `test_embedding_dims_1024_propagated` — verifies non-standard dim sizes work

## Checklist

### DIFF
--- a/openmemory/api/app/utils/memory.py
+++ b/openmemory/api/app/utils/memory.py
@@ -470,6 +470,15 @@ def get_memory_client(custom_instructions: str = None):
         if config.get("embedder", {}).get("provider") == "ollama":
             config["embedder"] = _fix_ollama_urls(config["embedder"])
 
+        # Propagate embedding_dims to the vector store config so the collection
+        # is created with the correct dimension. Many vector store configs default
+        # to 1536 (OpenAI size); using a different embedder without overriding
+        # this causes a dimension-mismatch error at insert time.
+        dims = config.get("embedder", {}).get("config", {}).get("embedding_dims")
+        if dims:
+            config["vector_store"]["config"]["embedding_model_dims"] = dims
+            print(f"Set vector store embedding_model_dims={dims} from embedder config")
+
         # ALWAYS parse environment variables in the final config
         # This ensures that even default config values like "env:OPENAI_API_KEY" get parsed
         print("Parsing environment variables in final config...")
--- a/openmemory/api/tests/test_memory_utils.py
+++ b/openmemory/api/tests/test_memory_utils.py
@@ -1,0 +1,97 @@
+"""Regression tests for embedding_dims propagation in get_memory_client().
+
+Ensures that when the embedder config declares an explicit embedding_dims,
+that value is forwarded to the vector store config as embedding_model_dims.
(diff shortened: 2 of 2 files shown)


kartik-mem0:
LGTM. This fixes the right function — `get_memory_client()` is where the assembled config is passed to `Memory`, so patching it here covers all callers regardless of how the embedder config was built. The backward-compatibility test (no `embedding_dims` set → vector store config unchanged) is important and correctly included. The `print` statement could be a `logging.debug` call but that's minor. Good work.

kartik-mem0:

PR #5407, opened 2026-06-06
fix(openmemory/mcp): pass top_k kwarg instead of limit to vector_store.search()

### LINKED ISSUE #5404: bug(openmemory): vector_store.search() called with wrong kwarg 'limit' instead of 'top_k'
## Description

In `openmemory/api/app/mcp_server.py`, the `search_memory` function calls `vector_store.search()` with `limit=10`, but the Mem0 vector store interface uses `top_k`, not `limit`.

## Steps to Reproduce

Run the self-hosted OpenMemory stack and issue any search via MCP. The unknown `limit` kwarg is silently ignored, so search returns the provider's default result count rather than the intended 10.

## Expected Behavior

Search returns up to 10 results as intended.

## Actual Behavior

`limit=` is silently ignored; result count is whatever the vector store provider defaults to.

## Affected Code

`openmemory/api/app/mcp_server.py` — `search_memory()`:

\`\`\`python
hits = memory_client.vector_store.search(
    query=query,
    vectors=embeddings,
    limit=10,        # ← should be top_k=10
    filters=filters,
)
\`\`\`

## Fix

Rename `limit=10` to `top_k=10`.

sseely:
## Linked Issue

Closes #5404

## Description

`search_memory()` in the MCP server called `vector_store.search()` with
`limit=10`. The vector store interface uses `top_k`, not `limit`, so the
kwarg was silently ignored and the call fell back to the store's internal
default — returning fewer or more results than intended depending on the
backend.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

`openmemory/api/tests/test_mcp_server.py` — `TestSearchMemoryKwargs`:
- `test_search_uses_top_k_not_limit` — mocks the memory client and asserts
  that `vector_store.search()` is called with `top_k=10` and no `limit` kwarg.

## Checklist

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -177,9 +177,9 @@ async def search_memory(query: str) -> str:
             embeddings = memory_client.embedding_model.embed(query, "search")
 
             hits = memory_client.vector_store.search(
-                query=query, 
-                vectors=embeddings, 
-                limit=10, 
+                query=query,
+                vectors=embeddings,
+                top_k=10,
                 filters=filters,
             )
 
--- a/openmemory/api/tests/test_mcp_server.py
+++ b/openmemory/api/tests/test_mcp_server.py
@@ -380,6 +380,54 @@ class TestStreamableHTTPResponses:
 # Route registration
 # ---------------------------------------------------------------------------
 
+# ---------------------------------------------------------------------------
+# search_memory — kwarg regression
+# ---------------------------------------------------------------------------
+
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for the contribution! Another PR (#5419) addresses the same issue — same one-line fix to `mcp_server.py`. Its test patches `get_memory_client_safe` which is the function actually called inside `search_memory`, making the test more accurate. Closing in favor of that one.

kartik-mem0:

PR #5409, opened 2026-06-07
fix(openmemory): fix search kwarg and embedder dims propagation

### LINKED ISSUE #5404: bug(openmemory): vector_store.search() called with wrong kwarg 'limit' instead of 'top_k'

### LINKED ISSUE #5405: bug(openmemory): embedding_dims not propagated to vector store, causes dimension mismatch with non-OpenAI embedders

oyaah:
Fixes #5404 by using `top_k` instead of `limit` in `search_memory`, and fixes #5405 by propagating `embedding_dims` to the `vector_store` config so that non-OpenAI embedders don't fail with dimension mismatches.

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -179,7 +179,7 @@ async def search_memory(query: str) -> str:
             hits = memory_client.vector_store.search(
                 query=query, 
                 vectors=embeddings, 
-                limit=10, 
+                top_k=10, 
                 filters=filters,
             )
 
--- a/openmemory/api/app/utils/memory.py
+++ b/openmemory/api/app/utils/memory.py
@@ -475,6 +475,10 @@ def get_memory_client(custom_instructions: str = None):
         print("Parsing environment variables in final config...")
         config = _parse_environment_variables(config)
 
+        dims = config.get("embedder", {}).get("config", {}).get("embedding_dims")
+        if dims and "vector_store" in config and "config" in config["vector_store"]:
+            config["vector_store"]["config"]["embedding_model_dims"] = dims
+
         # Check if config has changed by comparing hashes
         current_config_hash = _get_config_hash(config)
         


Bartok9:
Cross-reference for maintainer triage: this PR bundles two unrelated fixes (#5404 + #5405), and each issue now has a dedicated single-issue PR with regression tests:

- **#5404** (search kwarg `limit` → `top_k`) → **#5407** (`+51/-3`, adds `test_mcp_server.py::test_search_uses_top_k_not_limit` asserting `top_k` is passed and `limit` is not).
- **#5405** (propagate `embedding_dims` to vector store) → **#5406** (`+106/-0`, adds `test_memory_utils.py` regression coverage).

This PR (`+5/-1`) carries no tests and combines both issues into one diff. Per one-PR-per-issue hygiene, the focused tested pair (#5407 + #5406) is likely the cleaner merge path; this one could be closed in favor of them once they land.

(Heads-up on #5406: it includes a stray `print(f"Set vector store embedding_model_dims=...")` debug line worth dropping or switching to `logging` before merge.)

oyaah:
Closing in favor of the focused, tested, one-issue-per-PR split (per the triage on one-PR-per-issue hygiene):

- #5410 — `limit` → `top_k` in `vector_store.search()` (Closes #5404), with a regression test asserting `top_k` is passed and `limit` is not.
- #5411 — propagate `embedding_dims` to the vector store (Closes #5405), with regression tests; uses a guard that tolerates configs without a `vector_store` section and ships no debug `print`.

Both verified locally (full `test_mcp_server.py` 22 passed; `test_memory_utils.py` 4 passed) with negative checks confirming each test fails without its fix.

PR #5412, opened 2026-06-07
Return empty memory search results for blank queries

extrasmall0:
## Summary
- short-circuit sync and async Memory.search() for empty or whitespace-only queries
- preserve existing filter and search parameter validation before returning
- add regression coverage that blank queries do not call the embedder or vector store

Fixes #5220

## Tests
- uv run --extra test pytest tests/memory/test_main.py
- uv run --extra test ruff check mem0/memory/main.py tests/memory/test_main.py

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -169,6 +169,10 @@ def _validate_search_params(threshold: Optional[float] = None, top_k: Optional[i
             )
 
 
+def _is_blank_search_query(query: str) -> bool:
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for the contribution! Another PR (#5258) addresses the same issue with a more complete fix — it covers both `Memory` and `MemoryClient`, raises a `ValueError` so callers get actionable feedback instead of a silent empty result, and trims the query before embedding. Closing in favor of that one.

kartik-mem0:

PR #5415, opened 2026-06-07
feat(embeddings): add native embed_batch to OllamaEmbedding

### LINKED ISSUE #5414: feat(embeddings): OllamaEmbedding missing native embed_batch — N API calls instead of 1
## Problem

## Evidence

```python
```

## Impact

running Ollama locally.

aloktripathi1:
Closes #5414

OllamaEmbedding fell back to the base class embed_batch() which calls embed() 
sequentially -- N HTTP round-trips for N texts.

Fixed by overriding embed_batch() to pass the full list to client.embed(input=texts) 
in a single call. The Ollama SDK natively accepts input: Union[str, Sequence[str]].

openai.py and azure_openai.py already do this -- brings Ollama to parity.

3 new unit tests added covering batch return, newline stripping, and empty response error.


aloktripathi1:
Updated: removed newline stripping (Ollama handles newlines natively), aligned signature and docstring with openai.py/azure_openai.py peers, added empty list guard, and improved error check to catch count mismatch rather than just empty response.

kartik-mem0:
Reviewed this alongside the two other open PRs implementing the same `embed_batch` for Ollama (#5418 and #5598). This is the one I'd merge.

**What's good**
- Adds the native batch override cleanly — one `client.embed(input=texts)` call instead of N, which is exactly what the base class docstring asks native-batch subclasses to do, and it mirrors the `openai`/`azure_openai` peers.
- Short-circuits the empty-list case, so an empty batch makes zero API calls (matching the base default).
- Raises on a count mismatch (`len(embeddings) != len(texts)`). This matters more than it looks: the add path in `mem0/memory/main.py` does `dict(zip(mem_texts, embeddings))`, which would *silently drop* memories if a short list came back without an error. Raising instead routes it into the existing per-text fallback. The search path already guards the same mismatch (it logs and skips the entity boost), so raising is safe there too.
- Three unit tests: single-call, empty list, and the mismatch case.

**Verification (local, `dev_py_3_11`)**
- The three `embed_batch` tests pass.
- Red-green: reverting `ollama.py` to `main` (removing the override) makes `test_embed_batch_single_call` fail with "embed ... Called 3 times" and `test_embed_batch_count_mismatch_raises` fail with "DID NOT RAISE" — so those two genuinely bite. The empty-list test passes with or without the override (the base class also makes no call on an empty list), so it's a guard rather than a regression test.
- CI note: only the Vercel preview (advisory) and CLA show on this PR — the Python test matrix hasn't run here, so the above is from a local run.

LGTM from me — ready for a maintainer to squash-merge (`gh pr merge --squash`). Thanks for both reporting the issue and following through with the fix, including the pass to align the signature with the openai/azure peers.

kartik-mem0 (approved):

PR #5416, opened 2026-06-07
fix: use valid S3 entity index names

### LINKED ISSUE #5361: Warning output to log when calling Memory.add() using s3 vectors backend

he-yufeng:
Fixes #5361.

## Summary
- use a dash-suffixed entity index name for the S3 Vectors backend
- keep the existing underscore suffix for other vector stores
- add coverage for both the naming helper and the S3-backed Memory entity store path

## To verify
- python -m pytest tests\vector_stores\test_s3_vectors.py -q
- python -m pytest tests\test_memory.py -q
- python -m py_compile mem0\memory\main.py mem0\vector_stores\s3_vectors.py mem0\configs\vector_stores\s3_vectors.py
- python -m ruff check mem0\memory\main.py tests\vector_stores\test_s3_vectors.py tests\test_memory.py
- git diff --check


he-yufeng:
The remaining red Vercel status is the repository-side preview authorization gate, not a build or test failure from this branch. The status target is ercel.com/git/authorize for the Mem0 team, which I cannot authorize from the fork.

No branch changes from me here unless a code-owned check fails or a maintainer wants the preview disabled for this PR.

kartik-mem0:

### What this PR does
Fixes the `ValidationException: Invalid index name` warning on `Memory.add()` with the S3 Vectors backend (#5361). S3 Vectors forbids underscores in index names, but the entity store collection was built as `{collection_name}_entities`. A new `_entity_collection_name(provider, collection_name)` helper uses a dash (`{name}-entities`) for `s3_vectors` and keeps the underscore for every other provider; both the sync and async `entity_store` properties now call it.

### Strengths
- **Behaviour-preserving for all non-S3 providers** — `_entity_collection_name("qdrant", "mem0")` still returns `mem0_entities`, so there is zero migration/back-compat risk for existing Qdrant/PGVector/etc. stores.
- **Both call sites patched identically** — sync `Memory.entity_store` and async `AsyncMemory.entity_store`. A repo-wide grep confirms these are the *only* two sites that build a `_entities` name, so the fix is complete.
- **The S3 integration test exercises the real path** (`Memory.from_config` → `S3Vectors.create_col` → boto `get_index`) and asserts the underscore name is *not* used — a genuine regression guard rather than a tautology.

### Issues
Minor (all optional — none block merge):
- A one-line comment on `_entity_collection_name` naming the S3 underscore constraint would help future maintainers (both reviewers noted this).
- An `AsyncMemory` counterpart to the S3 integration test would add symmetry, though it's largely redundant: the async property calls the same module-level helper, which is already unit-tested for both branches.
- Separately, **pre-existing and out of scope for this PR**: `AsyncMemory.reset()` doesn't null `self._entity_store` the way sync `Memory.reset()` does — worth a follow-up, but unrelated to this naming fix and not something to address here.

### Duplicate cluster (issue #5361 has three PRs)
This is the strongest of the three and the one I'd merge:
- **#5416 (this)** — reusable helper, unit + real integration tests, MERGEABLE, behaviour-preserving.
- **#5363** (PHclaw) — inline ternary duplicated in both sites, **no tests**, mergeable UNKNOWN.
- **#5498** (HrushiYadav) — `.replace("_","-")` also rewrites underscores *inside* the user's collection name; but S3's main collection uses `collection_name` verbatim, so that diverges the entity index from the main one without fixing the root cause, and its test is heavily mocked.

Recommend squash-merging #5416 and closing #5363 and #5498 as duplicates.

### Evidence
- **CI:** `build_mem0` 3.10 / 3.11 / 3.12 pass, `changelog_check` pass, `check_changes` pass, `license/cla` signed. The red Vercel status is the repo-side preview-authorization gate, not a build/test failure — your read in the thread is correct.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_s3_vectors.py tests/test_memory.py -v` → **52 passed in 0.74s** (all three new tests pass).
- **Red-green proof:** reverting `mem0/memory/main.py` to `main` makes `test_memory_entity_store_uses_s3_valid_index_name` fail with `AssertionError: 'test-index-entities' in [..., 'test-index_entities']` (and the helper unit tests ImportError); restoring the fix → 52 passed. The regression test bites.
- **Lint:** `ruff check` on the three changed files → All checks passed!
- **Linked issue:** #5361 — addresses the reported root cause (underscore in the S3 index name) precisely.
- **Cross-package check:** the TS OSS SDK has the same `${collectionName}_entities` line, but ships **no** S3 Vectors provider, so it has no equivalent bug — no TS change needed.
- **Could NOT verify:** live S3 Vectors API behaviour (boto is mocked) — the AWS underscore constraint comes from the issue report plus S3 docs, verified here via code tracing only.

### Assessment
Correct, minimal, behaviour-preserving fix with a red-green-proven regression test and green CI. Just needs a squash-merge and closing the two duplicate PRs.

kartik-mem0 (approved):

PR #5418, opened 2026-06-07
feat(embeddings): add native embed_batch to OllamaEmbedding

### LINKED ISSUE #5414: feat(embeddings): OllamaEmbedding missing native embed_batch — N API calls instead of 1

AjaasMk:
## Linked Issue

Closes #5414

## Description

`OllamaEmbedding` did not override `embed_batch()`, so it fell back to the `EmbeddingBase` default, which calls `embed()` once per text — **N HTTP round-trips for N texts**.

The Ollama Python SDK's `client.embed()` already accepts `input: Union[str, Sequence[str]]` and returns one embedding per text **in order**, so a list of texts can be embedded in a single call. This PR adds the override, mirroring the existing `OpenAIEmbedding` / `AzureOpenAIEmbedding` implementations.

`embed_batch` is on hot paths — it is called from 7 sites in `mem0/memory/main.py` (every `add()` and entity-search path, sync and async). With Ollama as the embedder, bulk operations previously made one request per item; they now make one request total.

The base class docstring already calls this out explicitly:

### Why the length check

Callers in `memory/main.py` verify that `embed_batch` returns exactly one vector per input and skip entity boosting on mismatch:

```python
"embed_batch returned %d vectors for %d texts — skipping entity boost"
```

So the override raises `ValueError` if Ollama returns a vector count that does not match the input, rather than silently misaligning embeddings.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — purely additive; existing `embed()` behavior is unchanged.

## Test Coverage

Added two tests to `tests/embeddings/test_ollama_embeddings.py`:
- `test_embed_batch_single_call` — verifies all texts are embedded in **one** `client.embed(input=[...])` call, preserving order.
- `test_embed_batch_count_mismatch_raises` — verifies a short response raises `ValueError` instead of silently misaligning.

```
$ pytest tests/embeddings/test_ollama_embeddings.py -v
6 passed
```
Ruff check, ruff format, and isort all pass on the changed files.

## Checklist

- [x] I have updated documentation if needed (no public API change; none needed)

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -63,3 +63,25 @@ class OllamaEmbedding(EmbeddingBase):
         if not embeddings:
             raise ValueError(f"Ollama embed() returned no embeddings for model '{self.config.model}'")
         return embeddings[0]
+
+    def embed_batch(self, texts, memory_action: Optional[Literal["add", "search", "update"]] = "add"):
+        """
+        Embed multiple texts in a single Ollama API call.
+
+        The Ollama SDK's embed() accepts a list for `input` and returns one embedding
+        per text in order, so N texts cost a single round-trip instead of N.
+
+        Args:
+            texts (list[str]): The texts to embed.
+            memory_action (optional): The type of embedding to use. Must be one of "add", "search", or "update".
+        Returns:
+            list: A list of embedding vectors, one per input text, in the same order.
+        """
+        response = self.client.embed(model=self.config.model, input=texts)
+        embeddings = response.get("embeddings") or []
+        if len(embeddings) != len(texts):
+            raise ValueError(
+                f"Ollama embed() returned {len(embeddings)} embeddings for {len(texts)} texts "
+                f"using model '{self.config.model}'"
+            )
+        return embeddings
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for this — it's a genuinely solid implementation: a single batched call, the same safe `len(embeddings) != len(texts)` guard that protects the add path from silently dropping memories, a clear docstring, and tests for the single-call and mismatch cases.

After comparing the three PRs for this issue side by side, I'm recommending we consolidate on #5415. It's functionally equivalent to yours but adds an empty-list short-circuit (an empty batch makes zero API calls) plus one extra test, and it's from the issue reporter. The gap is small — your PR would be a fine merge on its own — but the two overlap on the same lines, so landing both just creates a conflict.

So my recommendation is to go with #5415 rather than merge both; a maintainer will make the final call on closing. Really appreciate the work here — the count-mismatch guard in particular was exactly the right instinct.

kartik-mem0:
Thanks for this, @AjaasMk! We had three open PRs adding native `embed_batch` to `OllamaEmbedding` — this one, #5415, and #5598. After reviewing all three together, we're going with #5415, so I'm closing this as superseded.

Really appreciate you implementing this (and reporting the issue) — sorry for the duplicate effort here. 🙏

PR #5421, opened 2026-06-07
fix(memory): return empty results for blank search queries

AjaasMk:
## Linked Issue

Closes #5220

## Description

`Memory.search()` performed no validation on the query string before passing it to the embedding model. Empty (`""`) and whitespace-only (`"   "`) queries were treated as valid, leading to provider-dependent behavior:

- some embedders (e.g. Ollama) **raise a ValueError** on empty input;
- others **return a vector that matches everything**, so *all* of a user's stored memories are returned — an unintended memory leak.

### Fix

Short-circuit blank queries to `{"results": []}` at the top of both the sync and async `search()` paths, before any embedding or vector store call:

```python
if not query or not query.strip():
    return {"results": []}
```

This returns the same shape as a valid search that matched nothing, and avoids the wasted embedding/vector-store round-trip entirely.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — blank queries previously crashed or leaked; they now return `{"results": []}`.

## Test Coverage

Added to `tests/test_main.py`:
- `test_search_blank_query_returns_empty_without_embedding` — parametrized over `""`, `"   "`, `"\n\t "`; asserts `{"results": []}` and that neither the embedding model nor the vector store is called.
- `test_async_search_blank_query_returns_empty_without_embedding` — same guarantee on the `AsyncMemory` path.

```
$ pytest tests/test_main.py tests/test_memory.py -q
70 passed
```
`ruff check` passes on both changed files.

## Checklist

- [x] I have updated documentation if needed (internal behavior fix — none needed)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1176,6 +1176,12 @@ class Memory(MemoryBase):
         # Reject top-level entity params - must use filters instead
         _reject_top_level_entity_params(kwargs, "search")
 
+        # A blank query has no meaningful embedding: some providers raise on empty
+        # input, others return a vector that matches everything (an unintended
+        # memory leak). Short-circuit to no results.
+        if not query or not query.strip():
+            return {"results": []}
+
         # Validate search parameters (before applying defaults)
         _validate_search_params(threshold=threshold, top_k=top_k)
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for the contribution! Another PR (#5258) addresses the same issue with a more complete fix — it also covers `MemoryClient` (not just `Memory`), trims the query before embedding, and validates non-string input. Closing in favor of that one.

kartik-mem0:

PR #5425, opened 2026-06-07
fix(memory): recover silently-dropped extractions via forced tool call

shraderdm:
## Linked Issue

Addresses #4054

(Framed as a complementary *recovery* option for the maintainers to weigh, not a replacement for the graceful-fail work already in flight on that issue. Happy to retarget as "Closes" if the team wants this as the resolution.)

## Description

When fact extraction runs against a cheaper or local model on instruction-dense / agentic transcripts, the model sometimes continues or role-plays the conversation instead of extracting facts, returning **prose with no JSON object at all**. The parser then raises `Expecting value: line 1 column 1 (char 0)`, the `except` block sets `extracted_memories = []`, and the facts are **silently dropped** (only an ERROR log). This is the failure in #4054.

The existing `extract_json` fallback cannot recover this case: with no `{` anywhere in the response, `text.find("{")` returns `-1` and `json.loads` raises at char 0. `extract_json` handles prose that *wraps* a JSON object; it cannot recover JSON that was never emitted.

This PR adds a **recovery** step on that parse failure: retry the same extraction with **forced structured output** (`tool_choice="required"`) on a `save_memories` schema. A model cannot answer a forced tool call with free prose, so the content-hijack that produced no JSON cannot recur in the same form. The recovery is gated to providers that support it and degrades gracefully to the current behavior (return `[]`) everywhere else.

Cost note: this adds **one extra LLM call, only on the parse-failure path** (never on the happy path) - occasionally two, in the rare case where the forced call's own structured output also truncates and the single bounded retry fires. That path was previously a guaranteed data loss, so the tradeoff is "one retry to recover otherwise-dropped facts." If you'd prefer it behind a config flag (default on or off), I'm happy to add one - flagged it here rather than presuming.

### What changed

- **`mem0/memory/main.py`** - in both the sync and async `_add_to_vector_store` parse-failure `except` blocks, call `recover_extraction_via_tools(...)` instead of giving up with `[]` - but only when the failed response contains no `{` at all (the content-hijack signature). Truncated JSON (a token-budget problem a forced tool call cannot fix) keeps the current behavior, so under-budgeted configs pay zero extra LLM calls. Behavior is unchanged on success and unchanged for providers that don't opt in.
- **`mem0/memory/utils.py`** - `recover_extraction_via_tools()` (the gated retry), `parse_tool_calls_for_memory()` (pull the `memory` arrays out of a tool-call response, merging across parallel tool calls, tolerating dict or JSON-string arguments, and dropping non-dict items a schema-violating model might emit; returns `None` when nothing parsed, distinct from a valid empty list), and `llm_supports_tool_calls()` (the capability gate). Recovery never raises - any failure returns `[]`. If the forced tool call's own structured output truncates (no tool call parsed at all - a successfully parsed empty `memory` list is a valid result and is never retried), it retries once at a raised `max_tokens`, capped at an absolute ceiling of 8192 tokens and **staying in tool mode** - reverting to free text would reintroduce the hijack the tool call exists to prevent.
- **`mem0/configs/prompts.py`** - `MEMORY_EXTRACTION_TOOL`, the forced-output schema in OpenAI function-calling format (the format mem0 providers already convert from internally), mirroring the `{"memory": [{"text", ...}]}` shape of `ADDITIVE_EXTRACTION_PROMPT`.
- **`mem0/llms/base.py`** - `supports_tool_calls = False` class attribute (the opt-in gate). This is deliberately separate from `AWSBedrockLLM.supports_tools` (which asks whether the Bedrock model family accepts tools at all); this flag asserts the stronger "honors a forced `tool_choice` and returns the standard `{"tool_calls": [...]}` dict." If you'd rather consolidate the two into one capability surface, I'm glad to - flagging the overlap rather than quietly adding a near-duplicate.
- **`mem0/llms/openai.py`** - opt in (`supports_tool_calls = True`); the provider already returns the `{"tool_calls": [...]}` dict.
- **`mem0/llms/anthropic.py`** - opt in, and fix the tools path so a forced call actually works. `_parse_response` on `main` already reads `tool_use` blocks into the `{"content", "tool_calls"}` shape, but the request side sent OpenAI-format tools straight through (Anthropic expects a flat `input_schema`) and mapped `tool_choice` as `{"type": tool_choice}`, so `"required"` became the invalid `{"type": "required"}` (Anthropic uses `{"type": "any"}`). This adds the OpenAI-to-Anthropic tool-schema conversion and the correct `tool_choice` mapping, and guards the no-tools `response.content[0].text` against an empty content list.
- **`mem0/llms/groq.py`** - opt in via a model-aware `supports_tool_calls` property. Groq is #4054's provider; its compound agentic models reject tool calling (the same family that rejects JSON `response_format`, per the existing `_supports_json_mode` gate), so they are excluded and every other Groq model is enabled. Verified against the live Groq API (see Provider scope).

### Provider scope

Recovery is enabled for **OpenAI, Anthropic, and Groq** - the providers whose forced-`tool_choice` behavior I could verify. OpenAI and Anthropic are exercised in the synthetic unit tests. Groq (#4054's provider) I verified against the live API: `llama-3.3-70b-versatile` and `openai/gpt-oss-20b` recover the dropped facts via a forced tool call, while `compound-beta` returns "tool calling is not supported" - so the Groq opt-in is a model-aware property that excludes the compound family (mirroring the existing JSON-mode gate) rather than a blanket class flag. Verified end-to-end through mem0's own `GroqLLM` -> `recover_extraction_via_tools`, not just the raw API.

Other providers (the rest of the OpenAI-compatible family, and Bedrock for the Anthropic/Cohere/Amazon families) appear to return a compatible `tool_calls` dict and could likely opt in the same way, but I haven't verified each one's forced-`tool_choice` behavior, so I left them off rather than claim coverage I don't have. Happy to extend per your direction. (Note: Bedrock already has its own `supports_tools` attribute, so wiring it in would mean reconciling the two - see above.)

This is orthogonal to #5178 (which surfaces the *API-exception* path). It is the content-hijack half of the same parse-failure path that the truncation case (#3918-adjacent) also lives in - see below.

### Companion PR and combined validation

This fix and a companion PR (#5429) for the other parse-failure mode were developed and validated together against a real conversation corpus, as one recovery loop that routes by failure type:
- **content-hijack / no JSON at all** (this PR): force a tool call so the model cannot answer with prose.
- **max_tokens truncation** (#5429): salvage the complete memories from the cut-off JSON, and, opt-in, retry at a raised `max_tokens`.

In a controlled before/after on a 57-transcript window, baseline runs (recovery off) left failures unrecovered; with recovery enabled, all 18 otherwise-dropped extractions were recovered, none left silently dropped. Run across the full 670-transcript corpus (about 5,400 extraction calls), the recovery path recovered all 109 otherwise-dropped extractions - each a set of facts the unpatched code discards with only an ERROR log - none left silently dropped. By recovery type: 81 via forced tool call (74 directly, plus 7 compound cases this PR now handles, where the forced tool call's own output truncated and was recovered by retrying at a raised budget rather than reverting to free text), and 28 truncations via token-raise.

The two were split into focused PRs for reviewability, but they compose in the same `except` block, so it may be worth looking at both together. (The corpus is private; only aggregate counts are shared. The tests in this PR are synthetic.)

## Type of Change

- [ ] Documentation update

## Breaking Changes

No internal callers are affected, and the tools-path return shape is unchanged (`main` already returns `{"content", "tool_calls"}` when `tools` are passed; this PR does not change that). The behavioral change is that a forced `tool_choice` request to Anthropic now succeeds: previously the OpenAI-format tools and the `{"type": "required"}` mapping were rejected by the Anthropic API, so no caller passing `tools` could get a tool call back. The only other change is a hardening on the no-tools path - an empty content list (e.g. a blocked response) now returns `""` instead of raising `IndexError`.

## Test Coverage

Ran locally (Python 3.13; the 3.10-3.12 matrix runs in CI): the new and touched tests pass, with all four touched suites green, `ruff check` + `ruff format --check` clean on the touched files. The recovery test is falsifiable - reverting the `main.py` wiring makes it fail (memory dropped, one LLM call); with the fix it passes (recovered, two calls). So it catches the bug rather than illustrating it.

Manual / empirical validation: beyond the synthetic unit tests, the recovery approach was run against a real conversation corpus in a controlled before/after. Two baseline runs without recovery permanently lost a set of facts to this exact parse failure (dropped, never recovered); with forced structured-output recovery enabled, the 6 prose-instead-of-JSON cases in that 57-transcript window were recovered rather than lost (the full-corpus numbers are in the combined-validation section above). (Truncation-class drops, where the model emits incomplete JSON after hitting `max_tokens`, are a separate case tracked in #5428 and fixed in the companion PR #5429, out of scope here. The corpus is private; only aggregate counts are shared.)

New tests (all synthetic - no private data):

- `tests/memory/test_main.py::TestParseFailureRecovery` and `TestAsyncParseFailureRecovery` (sync + async) - drive a synthetic content-hijack response (prose, no JSON) through `_add_to_vector_store` and assert the dropped memory is **recovered** on a capable provider (two LLM calls: hijacked extraction, then forced-tool recovery) and **still silently dropped** when the provider does not opt in (the current behavior, asserted explicitly in both sync and async).
- `tests/test_chatty_llm_parsing.py::TestToolCallRecovery` - unit tests for `parse_tool_calls_for_memory` (dict and JSON-string arguments), the capability gate (`is True`, not tripped by a `MagicMock`), the skip-when-uncapable path, the forced `tool_choice`, an **empty `memory` array round-tripping to `[]`** (the model is not compelled to invent a fact), and graceful `[]` when the provider raises or returns a malformed `tool_calls` entry (the never-raises guarantee).
- `tests/llms/test_anthropic.py` - OpenAI-format → Anthropic `input_schema` conversion, already-Anthropic-format passthrough, missing-`parameters` default, `tool_choice` mapping (`required` → `{"type": "any"}`, `auto` → `{"type": "auto"}`), `tool_use` block parsing, the tool-requested-but-text-returned graceful case, that the no-tools path still returns text unchanged, and a **seam test** asserting `AnthropicLLM`'s tool-response shape is consumed correctly by `parse_tool_calls_for_memory`.
- `tests/llms/test_groq.py` - the model-aware `supports_tool_calls` gate opts in tool-capable Groq models (`llama-3.3-70b-versatile`, `openai/gpt-oss-20b`) and excludes the compound family (`groq/compound`, `compound-beta`, `groq/compound-mini`).

Synthetic repro (the case that is currently dropped):

```python
# Extraction returns prose with no JSON object -> currently silently dropped.
mock.llm.generate_response.return_value = (
    "PLAN: I'll help you organize that. Before I start, a few questions: "
    "1) What is the deadline? 2) Who is the audience? Let me know and we'll begin."
)
# Without this PR: add() returns [], no exception, ERROR-level log only.
# With this PR (capable provider): the facts are recovered via a forced tool call.
```

## Checklist

### DIFF
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -1060,3 +1060,63 @@ def generate_additive_extraction_prompt(
 
     sections.append("# Output:")
     return "\n\n".join(sections)
+
+
+# Forced structured-output tool used to recover extraction when the LLM returns
+# free prose with no JSON object (e.g. the model role-plays / continues the
+# transcript instead of extracting facts, producing the classic
+# "Expecting value: line 1 column 1 (char 0)" parse failure). Forcing a tool
+# call constrains the model to the schema, so it cannot answer with prose.
+#
+# Defined in OpenAI function-calling format - the format mem0 providers expect
+# and convert from internally (see e.g. AnthropicLLM / AWSBedrockLLM). Mirrors
+# the ADDITIVE_EXTRACTION_PROMPT output shape: {"memory": [{"id", "text", ...}]}.
+# The "text" field name is contractual: the add() pipeline reads m.get("text")
+# (mem0/memory/main.py), so renaming it here silently drops every recovered memory.
+# "attributed_to" is also consumed (main.py promotes it into the memory's
+# metadata); "id" and "linked_memory_ids" mirror the prompt's output shape for
+# model familiarity and are harmlessly ignored downstream. Only "text" is
+# required - the others are optional, including "id".
+MEMORY_EXTRACTION_TOOL = {
+    "type": "function",
+    "function": {
+        "name": "save_memories",
+        "description": (
+            "Save the memorable facts extracted from the conversation. "
+            "Call this with every distinct fact worth remembering."
+        ),
+        "parameters": {
+            "type": "object",
+            "properties": {
+                "memory": {
+                    "type": "array",
+                    "description": "The list of extracted memories. Empty if nothing is memorable.",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "id": {
+                                "type": "string",
+                                "description": "Optional sequential id for this memory ('0', '1', ...).",
+                            },
+                            "text": {
+                                "type": "string",
+                                "description": "The self-contained, contextually rich memory statement.",
+                            },
+                            "attributed_to": {
+                                "type": "string",
+                                "description": "Optional speaker name the memory is attributed to.",
+                            },
+                            "linked_memory_ids": {
+                                "type": "array",
+                                "items": {"type": "string"},
+                                "description": "Optional ids of related existing memories.",
+                            },
+                        },
+                        "required": ["text"],
+                    },
+                }
+            },
+            "required": ["memory"],
+        },
+    },
+}
--- a/mem0/llms/anthropic.py
+++ b/mem0/llms/anthropic.py
@@ -12,6 +12,8 @@ from mem0.llms.base import LLMBase
 
 
 class AnthropicLLM(LLMBase):
+    supports_tool_calls = True
+
     def __init__(self, config: Optional[Union[BaseLlmConfig, AnthropicConfig, Dict]] = None):
         # Convert to AnthropicConfig if needed
         if config is None:
@@ -69,6 +71,74 @@ class AnthropicLLM(LLMBase):
         params.update(kwargs)
         return params
 
+    @staticmethod
+    def _convert_tools(tools: List[Dict]) -> List[Dict]:
+        """Convert OpenAI function-calling tool schemas to Anthropic's format.
+
+        mem0 defines tools in OpenAI format ({"type": "function", "function":
+        {"name", "description", "parameters"}}); Anthropic expects flat tools
+        with an ``input_schema``. Tools already in Anthropic format (carrying
+        ``input_schema``) are passed through unchanged.
+        """
+        converted = []
+        for tool in tools:
+            if "input_schema" in tool:
+                converted.append(tool)
+                continue
+            function = tool.get("function", tool)
+            converted.append(
+                {
+                    "name": function.get("name"),
+                    "description": function.get("description", ""),
+                    "input_schema": function.get("parameters", {"type": "object", "properties": {}}),
+                }
+            )
+        return converted
+
+    @staticmethod
+    def _convert_tool_choice(tool_choice):
+        """Map the cross-provider ``tool_choice`` value to Anthropic's format.
+
+        Anthropic expects a dict ({"type": "auto" | "any" | "none" | "tool"}).
+        Accepts the OpenAI-style strings used across mem0 ("auto", "required",
+        "none") and a bare tool name to force a specific tool; an already-formed
+        dict is passed through.
+        """
+        if isinstance(tool_choice, dict):
+            return tool_choice
+        mapping = {
+            "auto": {"type": "auto"},
+            "required": {"type": "any"},
+            "any": {"type": "any"},
+            "none": {"type": "none"},
+        }
+        if tool_choice in mapping:
+            return mapping[tool_choice]
+        # Anything else is treated as a specific tool name to force.
+        return {"type": "tool", "name": tool_choice}
+
+    @staticmethod
+    def _parse_response(response, tools):
+        """Process the Anthropic response based on whether tools were used.
+
+        With tools, returns the same shape as the OpenAI provider -
+        ``{"content": <text>, "tool_calls": [{"name", "arguments"}]}`` - reading
+        the JSON out of each ``tool_use`` block's ``.input`` (the SDK already
+        parses it to a dict). Without tools, returns the assistant text, or ""
+        when the response carries no content blocks (e.g. blocked/refused).
+        """
+        if tools:
+            processed_response = {"content": None, "tool_calls": []}
+            for block in response.content:
+                block_type = getattr(block, "type", None)
+                if block_type == "text":
+                    processed_response["content"] = block.text
+                elif block_type == "tool_use":
+                    processed_response["tool_calls"].append({"name": block.name, "arguments": block.input})
+            return processed_response
+
+        return response.content[0].text if response.content else ""
+
     def generate_response(
         self,
         messages: List[Dict[str, str]],
@@ -88,7 +158,8 @@ class AnthropicLLM(LLMBase):
             **kwargs: Additional Anthropic-specific parameters.
 
         Returns:
-            str: The generated response.
+            str or dict: The assistant text, or, when ``tools`` are supplied, a
+            ``{"content", "tool_calls"}`` dict matching the other providers.
         """
         # Separate system message from other messages
         system_message = ""
@@ -108,22 +179,9 @@ class AnthropicLLM(LLMBase):
             }
         )
 
-        if tools:  # TODO: Remove tools if no issues found with new memory addition logic
-            params["tools"] = tools
-            params["tool_choice"] = {"type": tool_choice}
+        if tools:
+            params["tools"] = self._convert_tools(tools)
+            params["tool_choice"] = self._convert_tool_choice(tool_choice)
 
         response = self.client.messages.create(**params)
         return self._parse_response(response, tools)
-
-    def _parse_response(self, response, tools):
-        if tools:
-            result = {"content": None, "tool_calls": []}
-            for block in response.content:
-                if block.type == "text":
-                    result["content"] = block.text
-                elif block.type == "tool_use":
-                    result["tool_calls"].append(
-                        {"name": block.name, "arguments": block.input}
-                    )
-            return result
-        return response.content[0].text
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -10,6 +10,16 @@ class LLMBase(ABC):
     Handles common functionality and delegates provider-specific logic to subclasses.
     """
 
+    #: Whether this provider can recover extraction via a forced tool call -
+    #: i.e. it honors ``tools`` / ``tool_choice`` and returns the standard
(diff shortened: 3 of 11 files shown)


kartik-mem0:
## Changes Requested

**Linked issue**: #4054 — validated. Bug confirmed via local reproduction: `groq/compound` and `compound-beta` both ignore `response_format={"type": "json_object"}` and return markdown prose, causing silent memory loss during extraction. The mini variants (`compound-beta-mini`, `groq/compound-mini`) work fine with JSON mode.

This PR takes the right approach — a general recovery mechanism via forced tool calls rather than patching around a single provider. The Anthropic tools path fix is also legitimate (confirmed the existing path was non-functional on main).

**Pre-push gate**: tests pass (78 passed, 1 skipped on touched files; no regressions in full suite). Lint has formatting issues — see below.

### Must fix

1. **`mem0/memory/utils.py` — Truncation retry fires on genuinely empty results.** When the forced tool call returns `{"memory": []}` (the conversation had nothing memorable), the code still retries at 4× `max_tokens`. It should distinguish "valid empty memory list" from "truncated output that couldn't be parsed" to avoid a spurious second LLM call and doubled cost. Suggestion: only retry when `_forced_tool_extraction` returned `[]` because it couldn't parse *any* tool call (no `memory` key found), not when it successfully parsed an empty list.

2. **Formatting** — `ruff format --check` and `isort --check-only` fail on the changed files. Run `ruff format` and `isort --profile black` on all 9 touched files.

### Optional

1. `mem0/llms/anthropic.py:149` — `_parse_response` without tools returns `response.content[0].text` unguarded. If the model returns an empty content list, this raises `IndexError`. Pre-existing issue, but since you're restructuring this path, worth adding a guard.

2. `mem0/configs/prompts.py` — The `id` field in `MEMORY_EXTRACTION_TOOL` is documented as sequential but not marked `required` in the schema. Minor inconsistency.

### What's good

- Generic, provider-agnostic solution that works across OpenAI, Anthropic, and any future provider that opts in via `supports_tool_calls`
- The `supports_tool_calls` vs Bedrock's `supports_tools` distinction is well-reasoned and correctly documented
- Comprehensive test coverage — 27 new tests covering recovery, graceful degradation, truncation retry, async paths, and the Anthropic tools conversion
- The Anthropic tools path was indeed non-functional (confirmed by code tracing) — this fixes a real gap
- Clean graceful-fallback design — unsupported providers get current behavior, recovery never raises

Please address the must-fix items. Happy to re-review once updated.

shraderdm:
Thanks for the thorough review - both must-fix items are addressed, plus the optionals. I also pulled in the parts of your #5433 review that live in this branch's diff (the recovery trigger and the retry cap), so #5433 picks them up via merge and will rebase down once this lands.

**Must fix 1 (spurious retry on a valid empty result):** `parse_tool_calls_for_memory` now returns `None` when no tool call carried a parseable `memory` list, distinct from a successfully parsed empty list. A clean `{"memory": []}` returns immediately - one LLM call, no token-raise retry. The retry only fires when nothing parsed at all (the truncation signature). `test_recover_returns_empty_when_nothing_memorable_without_retrying` asserts exactly one call and fails against the old conflated behavior. While in there I hardened the same parser two ways a misbehaving model could break it: `memory` arrays are now merged across parallel tool calls (the tool description invites one call per fact, and a leading empty call must not mask later facts), and non-dict items are dropped so a bare string in the array cannot crash the add pipeline downstream. Both pinned by tests.

**Must fix 2 (formatting):** `ruff format` + `isort --profile black` run on all 9 touched files; `ruff format --check` and `isort --check-only` now pass on them. One consequence to flag: `main.py` had pre-existing formatting drift, so whole-file `ruff format` pulls in churn beyond my hunks - that churn is what makes the check pass. If you'd rather keep the diff tight I can drop `main.py` from the format pass instead.

**From your #5433 review (the code lives in this branch's diff):**

- Recovery trigger tightened: recovery now fires only when the failed response contains no `{` at all (the content-hijack signature), in both sync and async paths. Truncated JSON keeps the current no-recovery behavior - a forced tool call cannot fix a token budget, and that case is #5429's salvage territory. `test_truncated_json_does_not_trigger_tool_recovery` (+ async twin) assert one LLM call and no recovery. One edge worth naming: hijack prose that merely contains a brace (an echoed code snippet, say) is treated as "JSON may have been emitted" and keeps the current behavior - the gate deliberately fails safe toward no extra spend, and a test pins that edge.
- Token raise capped: the raised-token retry now fires at `min(4x, 8192)` via `_TOOL_RETRY_TOKEN_CAP`, and is skipped entirely when the configured budget is already at or above the cap. Tests cover the capped raise and the skip.

**Optional 1:** adopted - the no-tools `_parse_response` path returns `""` instead of raising `IndexError` on an empty content list, with a test.

**Optional 2:** adopted - `id` is now documented as optional, matching the schema (`required: ["text"]` only).

One note related to your #5433 NIT 7, since the schema lives here: `attributed_to` is actually consumed downstream (`main.py` promotes it into the memory's metadata in the records loop), so I kept the extra fields and added a comment documenting which fields are consumed vs which mirror the prompt's shape.

Local gates: ruff lint + format + isort clean on the 9 files; 101 tests across the three touched suites pass (33 new/touched). PR body updated to match (test counts, gated trigger, capped retry).

shraderdm:
Hi @kartik-mem0 - rebased this onto current `main` to clear the conflict (it had drifted behind). The two must-fixes from your 06-10 review are in: empty-vs-truncated is a distinct sentinel (not `[]`) so the retry only fires on a genuine parse failure, and the forced-tool retry is capped. Recovery routes on whether the response contains a JSON object: truncated JSON is left untouched here, only true no-JSON content-hijack triggers the forced tool call.

Tests green locally (forced-tool recovery, the empty-vs-truncated gate, and the cap), ruff + format + isort clean. Ready for re-review whenever you have a cycle.

Heads up on sequencing: #5429 (truncated-extraction salvage) composes directly on top of this - it fills the truncated-JSON branch this PR deliberately leaves as a no-op. Its changes are ready locally and I'll rebase it the moment this lands, per your note on #5429.

kartik-mem0:

PR #5426, opened 2026-06-07
feat: support HTTP self-hosting by making secure cookies conditional

### LINKED ISSUE #5457: [bug] Next.js dashboard enters infinite redirect loop when self-hosted on HTTP / local IP
### Describe the Bug
When self-hosting Mem0 OSS on a local IP network (e.g., `http://192.168.1.66:3010`) over plain HTTP, the dashboard enters an infinite loop of redirects and `POST /api/auth/refresh -> 401 (Unauthorized)` retries, making the dashboard completely unusable.

### Cause
In production builds (`NODE_ENV=production`), the httpOnly cookies set in `/api/auth/refresh` route are configured with `secure: true`. Modern browsers silently ignore or reject cookies marked as `Secure` when sent over unencrypted HTTP connections. 

As a result:
1. When navigating to `/dashboard/*`, the Next.js middleware detects that the `mem0_refresh_token` cookie is missing and redirects the user back to `/login`.
2. On `/login`, the `AuthProvider` mounts, triggers `refreshSession()`, which attempts to fetch `/api/auth/refresh` without any cookie, returning a `401 Unauthorized` response.
3. The Axios interceptor catches the 401, attempts a refresh which fails, and then calls `redirectToLogin()` which reloads the page, restarting the loop.

### Proposed Solution
Allow configuring cookie security dynamically. We can auto-detect if the site is served over HTTPS using the protocol of `NEXTAUTH_URL`, and provide a `COOKIE_SECURE` environment variable to explicitly override this behavior (allowing `COOKIE_SECURE=false` for local network deployments).

tbdarkspirit:
### Summary
Currently, the Mem0 dashboard defaults to `secure: true` for the refresh token cookie when running in production mode. This causes an infinite redirect loop when self-hosting the dashboard on a local network IP over plain HTTP (e.g., `http://192.168.1.66:3010`), as modern browsers ignore `Secure` cookies over insecure connections.

Closes #5457

### Changes
- Added `isSecure` logic to `server/dashboard/src/app/api/auth/refresh/route.ts` that detects if `NEXTAUTH_URL` is using HTTPS.
- Added support for a `COOKIE_SECURE` environment variable to explicitly override the security policy.
- Documented `COOKIE_SECURE` in `server/dashboard/.env.example`.
- Default remains `true` for production HTTPS setups, ensuring no security regression for cloud users.

This fix is essential for users deploying Mem0 OSS in local environments without SSL certificates.

### DIFF
--- a/server/dashboard/.env.example
+++ b/server/dashboard/.env.example
@@ -1,3 +1,6 @@
 NEXT_PUBLIC_API_URL=http://localhost:8888
 API_INTERNAL_URL=http://localhost:8888
 NEXT_PUBLIC_INSTANCE_NAME=Mem0
+
+# Set to false to allow running the dashboard over HTTP without SSL (e.g. self-hosting on local network)
+COOKIE_SECURE=true
--- a/server/dashboard/src/app/api/auth/refresh/route.ts
+++ b/server/dashboard/src/app/api/auth/refresh/route.ts
@@ -4,9 +4,15 @@ import { AUTH_ENDPOINTS } from "@/utils/api-endpoints";
 import { getServerApiUrl } from "@/lib/server-api-url";
 
 const COOKIE_NAME = "mem0_refresh_token";
+const isSecure = process.env.COOKIE_SECURE !== undefined
+  ? process.env.COOKIE_SECURE === "true"
+  : process.env.NEXTAUTH_URL
+    ? process.env.NEXTAUTH_URL.startsWith("https://")
+    : process.env.NODE_ENV === "production";
+
 const COOKIE_OPTIONS = {
   httpOnly: true,
-  secure: process.env.NODE_ENV === "production",
+  secure: isSecure,
   sameSite: "lax" as const,
   path: "/",
   maxAge: 30 * 24 * 60 * 60, // 30 days


kartik-mem0:
## Changes Requested

**Linked issue**: None.

### Must fix

1. `server/dashboard/src/app/api/auth/refresh/route.ts:7-11` — The `COOKIE_OPTIONS` object is a module-level constant evaluated once at startup. The `isSecure` logic reads `process.env.COOKIE_SECURE` and `process.env.NEXTAUTH_URL` at module load time, which is fine in practice, but the `COOKIE_OPTIONS` constant is also used in the `PUT` handler (line 53: `cookieStore.set(COOKIE_NAME, body.refresh_token, COOKIE_OPTIONS)`) for setting the cookie on initial login. This is fine — just noting both usages share the same options, so the fix covers both correctly.

   The actual issue: the fallback chain `process.env.NEXTAUTH_URL ? ... : process.env.NODE_ENV === "production"` silently keeps `secure: true` when `NEXTAUTH_URL` is unset and `NODE_ENV=production`. That's the right default, but it means users who don't set `NEXTAUTH_URL` (common in custom Docker setups) get no benefit from this PR even if they're on HTTP. Consider documenting in the PR/README that `NEXTAUTH_URL=http://...` must be set for the auto-detection to work, or make the env var behavior explicit via `COOKIE_SECURE` in the Docker Compose example.

2. `server/dashboard/src/app/api/auth/refresh/route.ts` — No linked issue. The PR description explains the problem clearly but there's no GitHub issue tracking it. Please open one and link with `Closes #X` so this is discoverable for others hitting the same infinite redirect loop.

### Optional

1. Consider adding `COOKIE_SECURE` to `server/.env.example` or the Docker Compose docs so self-hosters know the escape hatch exists.

### What's good

- The fix is minimal and correct — a 7-line change that doesn't touch any other auth logic.
- The priority chain (`COOKIE_SECURE` env var > `NEXTAUTH_URL` detection > `NODE_ENV` default) is sensible.
- No security regression for HTTPS users: the default remains `secure: true` in production without `NEXTAUTH_URL`.
- The `refresh` route is the only place cookies are set in the codebase, so the scope is contained.

Please open a linked issue and clarify the `NEXTAUTH_URL` requirement in docs/the PR. The code logic itself is sound.

tbdarkspirit:
Hi @kartik-mem0!

Thank you so much for the feedback and review! 

I have addressed your comments:
1. **Opened Linked Issue**: Created #5457 to document the infinite login loop issue for self-hosters and linked it to this PR.
2. **Documented COOKIE_SECURE**: Added the `COOKIE_SECURE` environment variable to `server/dashboard/.env.example` with comments detailing its usage for local HTTP setups.
3. **Clarified Behavior**: Explained in the PR description that `NEXTAUTH_URL` is used for auto-detection and that the `COOKIE_SECURE` flag serves as an explicit escape hatch.

The changes are now pushed and the PR is ready for your review! Let me know if everything looks good on your end.

kartik-mem0:
Thanks for this, @tbdarkspirit — and especially for the thorough root-cause writeup in #5457. 🙏

The same root cause (the `mem0_refresh_token` cookie being set with `secure: true`, which modern browsers reject over plain HTTP, producing the `/api/auth/refresh → 401` redirect loop) was just fixed on `main` in #5026 (commit 7a5df647). It adds a `shouldUseSecureCookie()` helper to `server/dashboard/src/app/api/auth/refresh/route.ts` that derives the `secure` flag from the `DASHBOARD_URL` protocol, so an HTTP self-host (`DASHBOARD_URL=http://192.168.1.66:3010`) now gets `secure: false` and the loop is gone.

Since the core fix has already landed, I'm closing this as a duplicate of #5026.

One thing your PR offered that `main` doesn't have is an explicit `COOKIE_SECURE` override env var. If you'd find that useful as a belt-and-suspenders escape hatch, a small follow-up PR adding just `COOKIE_SECURE` on top of the existing `shouldUseSecureCookie()` helper would be very welcome.

Thanks again for the contribution! 🙌

PR #5429, opened 2026-06-08
fix(memory): salvage truncated extractions + opt-in token-raise retry

### LINKED ISSUE #5428: Truncated extraction output (max_tokens hit mid-JSON) drops all facts, including the ones already written
**Component:** Core / Python SDK (memory extraction)

## Description

When fact extraction produces a long response and the model hits `max_tokens` mid-output, it returns valid-but-incomplete JSON - an unterminated `{"memory": [...]}` array. In `_add_to_vector_store`, `json.loads` fails, the `extract_json` fallback also fails (the structure never closes), and the `except` block sets `extracted_memories = []`. Every fact is silently dropped, including the memories the model fully wrote before the cut. The only signal is an ERROR-level log.

This is a distinct failure mode from two adjacent ones:

- **#4054** is the model returning prose with *no JSON object at all* (`Expecting value: line 1 column 1 (char 0)`).
- **#3918** is Gemini leaking reasoning tokens into the JSON, malforming it near the *start* (`char 1` / `char 4`). Its title says "Unterminated JSON," which sounds like this, but its reported errors land at low char indices from reasoning-token leakage; truncation here fails at a *high* index near the end because the output simply stopped.

Truncation is different: the JSON is well-formed up to a point and then simply stops, so the parse failure is `Unterminated string` / `Expecting ',' delimiter` at a *high* character index (near the end). It shows up on long conversations and lower `max_tokens` settings.

Two things are recoverable that are currently thrown away:

1. The complete memory objects the model finished writing before the cut (they parse fine on their own).
2. The cut-off remainder, if the extraction is retried with a higher `max_tokens`.

Minimal synthetic repro (no private data):

```python
# Extraction hits max_tokens mid-array: memories 0 and 1 are complete,
# memory 2 is cut off mid-string.
truncated = (
    '{"memory": ['
    '{"id": "0", "text": "User likes hiking in the Laurel Highlands"}, '
    '{"id": "1", "text": "User was promoted to Senior Engineer"}, '
    '{"id": "2", "text": "User has a dog nam'
)
# Today: json.loads fails, extract_json fails, add() returns [] - all three
# facts lost, including the two that were fully written.
```

Happy to open a PR: salvage the complete objects from the truncated array (always, no extra LLM call), and optionally retry once at a raised `max_tokens` to recover the cut-off remainder (off by default, since it overrides the configured `max_tokens` and costs an extra call).

shraderdm:
## Linked Issue

Closes #5428

## Description

When fact extraction produces a long response and the model hits `max_tokens` mid-output, it returns valid-but-incomplete JSON (an unterminated `{"memory": [...]}` array). `json.loads` fails, the `extract_json` fallback also fails (the structure never closes), and the `except` block sets `extracted_memories = []`. **Every fact is silently dropped, including the memories the model fully wrote before the cut.**

This PR recovers them in two layers:

1. **Salvage (always on, no flag, no extra LLM call).** `salvage_memory_objects` peels the complete objects off the truncated `memory` array one at a time with stdlib `json.JSONDecoder.raw_decode` (no new dependency), stopping at the first object that does not fully parse. The half-written tail object is dropped (its content cannot be trusted); every object the model finished is kept.
2. **Token-raise retry (opt-in).** If the response was truncated and `recover_truncated_extractions=True`, retry the extraction once at `4x` the configured `max_tokens` (capped at an absolute ceiling of 8192 tokens, and skipped when the configured budget is already at or above the cap) to recover the cut-off remainder. Off by default because it overrides the configured `max_tokens` and costs an extra call. The `4x` multiple is data-driven: on a real corpus, truncations recovered at ~4x the base budget; a 2x raise under-shot and re-truncated. Bounded to one attempt; salvage stays the floor if the retry also truncates. The raise is applied via the shared `llm.config.max_tokens` under a lock (raised for the one retry call, restored in a `finally`) rather than as a `max_tokens=` kwarg, which about half the provider `generate_response` signatures reject (only those declaring `**kwargs` accept it).

### What changed

- **`mem0/memory/utils.py`** - `salvage_memory_objects(text) -> (memories, truncated)` and `retry_extraction_with_more_tokens(...)`. Neither raises: salvage degrades to `([], False)` even on non-string input (it runs inside the parse-failure except handler), and both filter non-dict items out of the recovered list so a schema-violating model cannot crash the add pipeline downstream.
- **`mem0/memory/main.py`** - both sync and async `_add_to_vector_store` parse-failure `except` blocks now salvage (always) and, when truncated and opted in, retry with a raised budget, instead of returning `[]`.
- **`mem0/configs/base.py`** - new `MemoryConfig.recover_truncated_extractions` flag (default `False`). Salvage runs regardless; the flag only enables the extra retry call.
- **`docs/open-source/configuration.mdx`** - documents the new `recover_truncated_extractions` flag and its 4x / 8192 retry behavior.

This is the truncation half of the same parse-failure path as the content-hijack case in #5425 - see below. It is orthogonal to #5178 (the API-exception path) and to #3918 (Gemini reasoning-token leakage). #3918's title says "Unterminated JSON," which sounds adjacent, but its reported errors land at low char indices from reasoning tokens leaking in near the start, whereas truncation fails at a high index near the end because the output stopped mid-write.

### Companion PR and combined validation

This fix and #5425 were developed and validated together against a real conversation corpus, as one recovery loop that routes by failure type:
- **content-hijack / no JSON at all** (#5425): force a tool call so the model cannot answer with prose.
- **max_tokens truncation** (this PR): salvage the complete memories, and, opt-in, retry at a raised `max_tokens`.

In a controlled before/after on a 57-transcript window, baseline runs (recovery off) left failures unrecovered; with recovery enabled, all 18 otherwise-dropped extractions were recovered, none left silently dropped. Run across the full 670-transcript corpus (about 5,400 extraction calls), the recovery path recovered all 109 otherwise-dropped extractions - each a set of facts the unpatched code discards with only an ERROR log - none left silently dropped. By recovery type: 28 truncations via token-raise (this PR's path), and 81 content-hijack cases via forced tool (the companion PR, #5425).

The two were split into focused PRs for reviewability, but they compose in the same `except` block, so it may be worth looking at both together. (The corpus is private; only aggregate counts are shared. The tests in this PR are synthetic.)

### Updates since opening

Rebased onto current `main`. Two refinements were added since the initial review; flagging them here rather than folding them in silently:

- **Salvage now covers single-quoted output.** The original salvage anchored on the double-quoted `"memory"` key, so a model that emitted a Python-repr dict (`{'memory': [...]}` with single quotes) matched nothing and had every fact dropped - the exact silent-drop on an adjacent class that this function exists to prevent. Detection is now quote-agnostic; a quote-aware bracket scan plus `ast.literal_eval` recovers the complete single-quoted array, and any memory-bearing response it still cannot parse is logged at WARNING instead of dropped silently.
- **The token-raise retry lock is now per-llm-instance** (cached on the llm via double-checked locking) instead of process-wide, so concurrent retries on *different* llm instances no longer serialize against each other; the corruption window on a *shared* llm stays closed. A process-wide lock remains as a defensive fallback.

Both are covered by new tests. Happy to split either into its own PR if you'd prefer to keep this one to the original scope.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A. Salvage only runs on responses that already fail to parse (today those return `[]`), so it strictly recovers data that was being dropped. The retry is opt-in and off by default.

## Test Coverage

Ran locally (Python 3.13; the 3.10-3.12 matrix runs in CI): all tests in the two touched suites pass (including the new salvage and concurrency tests), ruff lint + format clean on touched files. The salvage test is falsifiable - reverting the `main.py` wiring makes a truncated response return `[]` (facts lost); with the fix the complete memories are recovered.

New tests (all synthetic - no private data):
- `tests/test_chatty_llm_parsing.py::TestSalvageTruncatedMemories` - salvage recovers the complete objects and drops only the cut-off tail; reports `truncated` correctly; no-op when there is no `memory` array.
- `tests/test_chatty_llm_parsing.py::TestSalvageTruncatedMemories` (single-quote cases) - recovery of a single-quoted Python-repr `{'memory': [...]}` response, a `]` inside a string value not breaking the array scan, and the loud-WARNING floor when a single-quoted response is also truncated (nothing salvageable, but not dropped silently).
- `tests/memory/test_main.py::TestTruncationRecovery` - through `_add_to_vector_store`: salvage recovers the complete memories by default (no retry); with the opt-in flag, the token-raise retry recovers the cut-off remainder at a 4x budget applied via the shared config (asserted at call time, with no `max_tokens=` kwarg, and restored after). Also covers the concurrent-retry lock (no shared-config corruption), that the lock is per-llm-instance, the 8192 cap on the raise, and the skip when the budget is already at the cap.

Manual / empirical validation: beyond the synthetic tests, the approach was run against a real conversation corpus (see the combined-validation note above); truncations were recovered via the token-raise path, with none left unrecovered. Corpus is private; aggregate counts only.

## Checklist

- [x] I have updated documentation if needed (new config flag documented in its Field description; no public API change)

### DIFF
--- a/docs/open-source/configuration.mdx
+++ b/docs/open-source/configuration.mdx
@@ -145,6 +145,18 @@ memory = Memory.from_config_file("config.yaml")
   <Accordion title="Reranker depth">
     Limit `top_k` to 10–20 results; sending more adds latency without meaningful gains.
   </Accordion>
+  <Accordion title="Truncated extraction recovery">
+    When the extraction LLM hits its `max_tokens` mid-output, the complete facts it finished writing are always salvaged from the partial response. To also recover the cut-off remainder, set the top-level `recover_truncated_extractions: True` (default `False`):
+
+    ```python
+    config = {
+        "llm": {"provider": "openai", "config": {"model": "gpt-4.1-mini", "max_tokens": 2000}},
+        "recover_truncated_extractions": True,
+    }
+    ```
+
+    On a truncated response this retries extraction once at a raised `max_tokens` (4x, capped at an absolute ceiling of 8192). The retry is skipped when the configured `max_tokens` is already at or above the cap, and it spends the extra tokens only on the truncation path, never on a normal extraction.
+  </Accordion>
 </AccordionGroup>
 
 <Warning>
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -55,6 +55,17 @@ class MemoryConfig(BaseModel):
         description="Custom instructions for fact extraction",
         default=None,
     )
+    recover_truncated_extractions: bool = Field(
+        description=(
+            "When True, if an extraction response is truncated (the model hit "
+            "max_tokens mid-JSON), retry once at a raised max_tokens (4x) to "
+            "recover the cut-off memories. Complete memories are always salvaged "
+            "from a truncated response regardless of this flag; enabling it only "
+            "adds the extra retry call (which costs more tokens and overrides the "
+            "configured max_tokens for that one call)."
+        ),
+        default=False,
+    )
 
 
 class AzureConfig(BaseModel):
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -58,8 +58,10 @@ class MemoryConfig(BaseModel):
     recover_truncated_extractions: bool = Field(
         description=(
             "When True, if an extraction response is truncated (the model hit "
-            "max_tokens mid-JSON), retry once at a raised max_tokens (4x) to "
-            "recover the cut-off memories. Complete memories are always salvaged "
+            "max_tokens mid-JSON), retry once at a raised max_tokens (4x, capped "
+            "at an absolute ceiling of 8192 tokens) to recover the cut-off "
+            "memories. The retry is skipped when the configured max_tokens is "
+            "already at or above the cap. Complete memories are always salvaged "
             "from a truncated response regardless of this flag; enabling it only "
             "adds the extra retry call (which costs more tokens and overrides the "
             "configured max_tokens for that one call)."
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -2570,6 +2570,9 @@ class AsyncMemory(MemoryBase):
                 logger.info("Salvaged %d complete memory item(s) from a malformed extraction response (async)", len(extracted_memories))
             # Layer 2 (opt-in): retry once with a raised max_tokens on truncation.
             if truncated and getattr(self.config, "recover_truncated_extractions", False) is True:
+                # Runs the sync retry on a worker thread. retry_extraction_with_more_tokens
+                # raises and restores llm.config.max_tokens under a lock, so concurrent
+                # retries sharing this llm cannot corrupt the shared config.
                 retried = await asyncio.to_thread(
                     retry_extraction_with_more_tokens, self.llm, system_prompt, user_prompt
                 )
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -94,34 +94,38 @@ def _vector_store_list_rows(listed):
 # Fields that hold runtime auth/connection objects and must be preserved.
 # These are non-serializable objects (e.g. AWSV4SignerAuth, RequestsHttpConnection)
 # needed by clients like OpenSearch — not sensitive strings to redact.
-_RUNTIME_FIELDS = frozenset({
-    "http_auth",
-    "auth",
-    "connection_class",
-    "ssl_context",
-})
+_RUNTIME_FIELDS = frozenset(
+    {
+        "http_auth",
+        "auth",
+        "connection_class",
+        "ssl_context",
+    }
+)
 
 # Fields that are known to contain sensitive secrets and must be redacted.
-_SENSITIVE_FIELDS_EXACT = frozenset({
-    "api_key",
-    "secret_key",
-    "private_key",
-    "access_key",
-    "password",
-    "credentials",
-    "credential",
-    "secret",
-    "token",
-    "access_token",
-    "refresh_token",
-    "auth_token",
-    "session_token",
-    "client_secret",
-    "auth_client_secret",
-    "azure_client_secret",
-    "service_account_json",
-    "aws_session_token",
-})
+_SENSITIVE_FIELDS_EXACT = frozenset(
+    {
+        "api_key",
+        "secret_key",
+        "private_key",
+        "access_key",
+        "password",
+        "credentials",
+        "credential",
+        "secret",
+        "token",
+        "access_token",
+        "refresh_token",
+        "auth_token",
+        "session_token",
+        "client_secret",
+        "auth_client_secret",
+        "azure_client_secret",
+        "service_account_json",
+        "aws_session_token",
+    }
+)
 
 # Suffixes that indicate a field likely holds a secret value.
 _SENSITIVE_SUFFIXES = (
@@ -167,13 +171,9 @@ def _validate_and_trim_entity_id(value: Optional[str], name: str) -> Optional[st
         return None
     trimmed = value.strip()
     if trimmed == "":
-        raise ValueError(
-            f"Invalid {name}: cannot be empty or whitespace-only. Provide a valid identifier."
-        )
+        raise ValueError(f"Invalid {name}: cannot be empty or whitespace-only. Provide a valid identifier.")
     if any(c.isspace() for c in trimmed):
-        raise ValueError(
-            f"Invalid {name}: cannot contain whitespace. Provide a valid identifier without spaces."
-        )
+        raise ValueError(f"Invalid {name}: cannot contain whitespace. Provide a valid identifier without spaces.")
     return trimmed
 
 
@@ -192,16 +192,12 @@ def _validate_search_params(threshold: Optional[float] = None, top_k: Optional[i
         if not isinstance(threshold, (int, float)):
             raise ValueError("threshold must be a valid number")
         if threshold < 0 or threshold > 1:
-            raise ValueError(
-                f"Invalid threshold: {threshold}. Must be between 0 and 1 (inclusive)."
-            )
+            raise ValueError(f"Invalid threshold: {threshold}. Must be between 0 and 1 (inclusive).")
     if top_k is not None:
         if not isinstance(top_k, int) or isinstance(top_k, bool):
             raise ValueError("top_k must be a valid integer")
         if top_k < 0:
-            raise ValueError(
-                f"Invalid top_k: {top_k}. Must be a non-negative integer."
-            )
+            raise ValueError(f"Invalid top_k: {top_k}. Must be a non-negative integer.")
 
 
 def _validate_and_trim_search_query(query: str) -> str:
@@ -354,7 +350,7 @@ def _build_filters_and_metadata(
             message="At least one of 'user_id', 'agent_id', or 'run_id' must be provided.",
             error_code="VALIDATION_001",
             details={"provided_ids": {"user_id": user_id, "agent_id": agent_id, "run_id": run_id}},
-            suggestion="Please provide at least one identifier to scope the memory operation."
+            suggestion="Please provide at least one identifier to scope the memory operation.",
         )
 
     # ---------- optional actor filter ----------
@@ -463,10 +459,7 @@ class Memory(MemoryBase):
         # Initialize reranker if configured
         self.reranker = None
         if config.reranker:
-            self.reranker = RerankerFactory.create(
-                config.reranker.provider,
-                config.reranker.config
-            )
+            self.reranker = RerankerFactory.create(config.reranker.provider, config.reranker.config)
 
         # Entity store is initialized lazily on first use
         self._entity_store = None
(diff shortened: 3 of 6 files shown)


kartik-mem0:
## Changes Requested

**Linked issue**: #5428 — validated.

### Must fix

1. `mem0/memory/utils.py:213` (`retry_extraction_with_more_tokens`) — The function calls `llm.generate_response(..., max_tokens=raised)` as a keyword argument, but most LLM provider implementations do not accept `**kwargs` in their `generate_response` signature. Specifically, `GroqLLM`, `GeminiLLM`, and `TogetherLLM` (and others) will raise `TypeError: got an unexpected keyword argument 'max_tokens'` at runtime. Only `AnthropicLLM` and `OllamaLLM` accept `**kwargs`. Since this is wired into the `except` block of the parse-failure path, the `TypeError` is silently swallowed by the outer `except Exception` in `retry_extraction_with_more_tokens`, returning `[]` — but the truncation still happens with no memory recovery, and no meaningful error surfaces.

   **Fix options** (pick one):
   - Mutate `llm.config.max_tokens` temporarily before the retry call and restore it after (consistent with how all providers read `self.config.max_tokens` internally).
   - Or: add a `max_tokens` parameter to the abstract `generate_response` signature in `base.py` and update all provider implementations.
   - Or: pass a modified config clone to a new LLM instance for the retry (heavier but cleanest).

   The simplest fix is option 1 — temporarily override the config value.

2. `mem0/memory/utils.py:213` — Related to #1: the `except Exception` in `retry_extraction_with_more_tokens` catches the `TypeError` from #1 and logs it as a `WARNING: Token-raise extraction retry failed`. This means users with Groq/Gemini/Together will see a warning log and get 0 memories recovered from the retry, with no indication that the call convention is wrong. Once #1 is fixed, this broad catch is appropriate since it also handles network errors and provider-specific failures.

3. `mem0/memory/main.py:775` (async path) — `asyncio.to_thread(retry_extraction_with_more_tokens, self.llm, system_prompt, user_prompt)` runs the synchronous retry function on a thread pool. This is correct for CPU-bound work, but `retry_extraction_with_more_tokens` itself makes an LLM network call. If the LLM client isn't thread-safe, this could produce subtle concurrency bugs. Most Groq/OpenAI clients are thread-safe in practice, but it's worth a comment or using the async path directly if the provider supports it.

### Optional

1. `mem0/memory/utils.py:177` — The docstring says "Skipped when `max_tokens` is unset (cannot compute a raise)" but the condition is `if not current:` which also skips when `max_tokens=0`. Since `0` is a nonsensical value anyway, this is fine, but consider `if not current or current <= 0:` for clarity.

2. `mem0/memory/utils.py:164` — `truncated = True` as an initial assumption (before the loop) is correctly flipped to `False` when the `]` is found. The logic is correct but could benefit from a one-line comment like `# Default: assume truncated; cleared if array closes cleanly` for future readers.

### What's good

- The `salvage_memory_objects` implementation is clean — stdlib `raw_decode` peel is the right tool, no new dependencies, and the loop handles whitespace/comma separators properly.
- Layer 1 (salvage) being always-on and Layer 2 (retry) being opt-in is the right default posture — it recovers facts without surprise token cost for existing users.
- Test coverage is solid: unit tests for `salvage_memory_objects` cover truncated mid-object, clean close, no-array, empty input, and cut-after-object cases. `TestTruncationRecovery` tests the full `_add_to_vector_store` path for both salvage-only and opt-in retry.
- The `4x` multiplier with data-driven justification is a reasonable approach.
- Ruff clean.

The core logic is correct — the blocking issue is that `max_tokens=raised` as a kwarg will silently fail for the most common providers (Groq, Gemini, Together). Please fix the call convention for the token-raise retry and I'll re-review.

shraderdm:
Thanks for the detailed review - all three addressed.

**1:** the retry no longer passes `max_tokens=` as a kwarg. It raises `llm.config.max_tokens`, makes the one call, and restores it in a `finally`. Most providers read `self.config.max_tokens` at request time, so the raise takes effect there. A few paths (LangChain's prebuilt model, Bedrock's snapshotted `model_config`, the OpenAI structured-output provider, and reasoning models that drop `max_tokens`) don't re-read it - there the retry harmlessly no-ops and we fall back to the already-salvaged memories. So: broadly effective and never crashes, rather than uniform across every provider.

**3 (your thread-safety flag):** good catch, and it bit the first cut - mutating the shared `llm.config` under `asyncio.to_thread` let concurrent retries corrupt `config.max_tokens`. Fixed by serializing the raise/restore with a lock and re-reading the base inside it. (Alternative would be a cloned-config retry instance to avoid serializing the rare retry - happy to switch if you'd prefer that to a lock held across the call.)

**2:** kept the broad `except`; now that 1 doesn't raise `TypeError` it only catches genuine provider/network failures.

**Optional 1:** adopted `if not current or current <= 0`. **Optional 2:** the `truncated` comment was already in from an earlier commit on this branch.

Test now records `config.max_tokens` at call time and asserts it's raised 4x during the retry and restored after, with no kwarg, plus a new concurrent-retry test guarding the lock. Falsifiable - reverting to the kwarg approach fails it `assert 1000 == 4000` (the old test passed only because `Mock` accepts any kwarg, which masked the bug). ruff clean; `test_main.py` + `test_memory_utils.py` green.

shraderdm:
Pushed one follow-up commit on top of the fix described above. It does three things, all disclosed here since none were asked for in this PR's review:

1. **Cap on the retry budget**, mirroring what you asked for on #5433 so the two retry paths stay consistent: the 4x raise is capped at an absolute ceiling of 8192 tokens (`_RETRY_TOKEN_CAP`), and the retry is skipped when the configured budget is already at or above the cap. Tests cover the capped raise and the skip; the config field's docstring documents both. Happy to drop the cap here if you'd rather keep this PR exactly as reviewed.
2. **Input-shape hardening on the two recovery entry points:** `salvage_memory_objects` now degrades to `([], False)` on non-string input instead of raising `TypeError` from inside the parse-failure except handler, and the token-raise retry filters non-dict items out of the recovered `memory` array (downstream reads `m.get("text")`, so a bare string from a schema-violating model would otherwise crash `add()`). Both pinned by tests.
3. **`ruff format` on the touched files**, matching the format pass you asked for on #5425/#5433. Heads up that this pulls pre-existing formatting churn in `main.py` into this diff too (the functional `main.py` change here is 3 lines); I kept it so the two branches share identical formatting and merge cleanly, but I can drop it from this PR if you'd rather keep the diff tight.

PR body updated to match. Local gates: ruff lint + format clean on touched files; 75 tests across the two touched suites pass.

kartik-mem0:
## Changes Requested

Really solid work here — the `salvage_memory_objects` design is exactly right (`raw_decode` cannot emit a partial tail object), the two-layer recovery (always-salvage + opt-in retry) is well-conceived, the 8192-cap and the lock are correct, and the test suite is genuinely falsifiable (75 pass with the fix; 6 fail without it, including a behavioral failure that proves the memory loss). The formatting cleanup is also a net positive even if it inflates the diff. A few things need addressing before this can land:

### Issues

#### Important (must fix before merge)

1. **Both this PR and #5425 are CONFLICTING on GitHub** — both replace the same two `extracted_memories = []` lines in the sync and async `_add_to_vector_store` except handlers with different, incompatible code. The intended composition (your PR body describes it well: route by whether `"{"` appears in the response → salvage+retry vs. tool-call recovery) must land as a single integrated except block. The cleanest path is probably to wait for #5425 to merge first, then rebase this PR with the routing logic added. If the order is reversed, same result. The current state is that neither can merge cleanly without the other's changes present.

2. **No integration test for the async retry path** — `TestTruncationRecovery` drives everything through sync `Memory._add_to_vector_store`. The async path at `AsyncMemory._add_to_vector_store` dispatches the retry via `asyncio.to_thread(retry_extraction_with_more_tokens, ...)` — that wiring (argument forwarding, flag check, result assignment) is exercised by zero tests. A regression there would be silent. The existing `TestAsyncAddToVectorStoreErrors` fixture shows how to set up async tests; an `AsyncMemory` counterpart to `test_opt_in_retry_recovers_cut_off_remainder` would close this.

3. **TS OSS counterpart has the same silent-drop bug** — `mem0-ts/src/oss/src/memory/index.ts` lines 804–807 contain the same outer `catch (e) { extractedMemories = []; }` pattern. The TS `addToVectorStore` uses `generateResponse` with `{ type: "json_object" }` and manual JSON parsing — no structured outputs, no truncation protection. Please either open a follow-up TS issue/PR or note it explicitly as out-of-scope so it doesn't get lost.

#### Important (worth a follow-up)

4. **`recover_truncated_extractions` is not surfaced in `docs/`** — it's a new user-visible `MemoryConfig` field. The Field description is good, but OSS users discover config options through the Mintlify docs, not Pydantic source. A short mention in the configuration reference would help.

#### Minor

5. **`_RETRY_TOKEN_LOCK` is process-wide** — serializes concurrent retries on different LLM instances. You note this in the docstring, and the opt-in default makes it a non-issue today. Worth addressing before enabling it by default: a per-`llm`-instance lock (lazily initialized as `getattr(llm, "_retry_lock", threading.Lock())`) would isolate instances without changing the correctness guarantee.

6. **Formatting churn in `main.py`** — the frozenset reformats, ValueError collapses, and dict reformats are all correct per Ruff, but they make `git blame` noisier than needed. A separate formatting-only commit or PR would make the actual logic change much easier to review and bisect.

### Evidence
- **CI:** Vercel (authorization — infra, not code). CLA signed. Python SDK CI checks not visible for this branch.
- **Mergeable:** CONFLICTING (GitHub-reported; confirmed — both #5425 and #5429 target the same two except-block lines)
- **Targeted tests:** `pytest tests/test_chatty_llm_parsing.py tests/memory/test_main.py -v` → 75 passed in 4.06s (Python 3.12)
- **Red-green proof:** 12 new tests FAIL against pre-PR impl (one behavioral: `assert 'User likes hiking in the Laurel Highlands' in []` — proves memory loss); 12 PASS with the fix ✓
- **Lint:** `ruff check` on all 5 changed files → clean
- **TS counterpart:** `mem0-ts/src/oss/src/memory/index.ts` lines 804–807 — confirmed same silent-drop pattern
- **Could NOT verify:** live-LLM retry behavior, provider-specific `max_tokens` propagation at runtime

### Assessment
Ready to merge: **With fixes** — please rebase to integrate #5425's except-block, add the async test, and flag the TS counterpart. The core logic is right and I'm happy to re-review once the rebase lands.

kartik-mem0:

PR #5431, opened 2026-06-08
fix(llms): honor forced tool_choice and per-call kwargs in Gemini provider

### LINKED ISSUE #5430: Gemini provider: forced tool_choice silently disabled, and generate_response rejects **kwargs
**Component:** Core / Python SDK

A couple of tool-handling bugs in `mem0/llms/gemini.py` that I hit while wiring up forced tool calls on Gemini. Both are in the same file and the fixes are small, so filing as one issue (same shape as #5189: several small same-file provider bugs filed together).

### 1. `tool_choice="required"` silently disables tool calling

`generate_response` maps `tool_choice` to Gemini's `FunctionCallingConfigMode`:

```python
if tool_choice == "auto":
    mode = types.FunctionCallingConfigMode.AUTO
elif tool_choice == "any":
    mode = types.FunctionCallingConfigMode.ANY
else:
    mode = types.FunctionCallingConfigMode.NONE
```

`"required"` - the standard value for "force a tool call," used by the other providers and the base contract - falls into the `else` branch and maps to `NONE`, which turns tool calling OFF. So asking Gemini to force a tool silently produces no tool call. (`allowed_function_names` is likewise only set for `"any"`, not `"required"`.)

### 2. `generate_response` does not accept `**kwargs`

The base `LLMBase.generate_response` signature is `(self, messages, tools=None, tool_choice="auto", **kwargs)`, and the other providers accept `**kwargs` for per-call overrides (e.g. a per-call `max_tokens`). Gemini's `generate_response` omits `**kwargs`, so any caller passing an extra parameter raises `TypeError`.

### Suggested fix

- map `"required"` (like `"any"`) to `ANY` mode, with `allowed_function_names` set
- accept `**kwargs` and apply a `max_tokens` override as `max_output_tokens`

Happy to open a PR (small, with unit tests).

shraderdm:
## Linked Issue

Closes #5430

## Description

Two small tool-handling bugs in mem0's Gemini provider (`mem0/llms/gemini.py`), the same shape as #5189 (several small same-file provider bugs filed together):

1. **`tool_choice="required"` silently disabled tool calling.** `generate_response` mapped `"required"` to `FunctionCallingConfigMode.NONE` (tools OFF) instead of `ANY` (force a tool), because only `"auto"` and `"any"` were handled and everything else fell through to `NONE`. So asking Gemini to force a tool produced no tool call.
2. **`generate_response` did not accept `**kwargs`.** The base `LLMBase.generate_response` contract and several providers (OpenAI, Anthropic, AWS Bedrock, among others) accept `**kwargs` for per-call overrides; Gemini omitted it, so passing an extra parameter (e.g. a per-call `max_tokens`) raised `TypeError`.

### What changed

- **`mem0/llms/gemini.py`** - map `"required"` to `ANY` mode. The two forcing values keep their distinct semantics, matching LiteLLM's Gemini mapping: `"any"` forces one of *these* tools (`allowed_function_names` set, as before), `"required"` forces *some* tool with the model free to choose (`allowed_function_names` unset). Also accept `**kwargs` and apply a `max_tokens` override as `max_output_tokens` (only `max_tokens` is recognized; anything else is deliberately ignored rather than forwarded).

This is a prerequisite for forced structured-output use on Gemini - the mechanism #5425 uses for OpenAI and Anthropic today, and the angle raised for the Gemini case in #3918. It does not by itself extend that recovery to Gemini (that would be a separate change opting Gemini into the recovery gate). The bug stands on its own: any caller forcing a tool on Gemini hits it.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A. The only behavior change is for `tool_choice="required"` (previously a silent no-op that disabled tools, now forces a tool as intended) and accepting `**kwargs` (previously a `TypeError`). Existing `"auto"` / `"any"` / `"none"` behavior is unchanged.

## Test Coverage

New tests in `tests/llms/test_gemini.py` (mocked client, no real API): `tool_choice="required"` maps to `ANY` with `allowed_function_names=None` (model free to choose), `"any"` still forces with `allowed_function_names` set, `"none"` disables, and a per-call `max_tokens` kwarg overrides `max_output_tokens`. Falsifiable: reverting the mapping fix makes the `required`-forces-`ANY` test fail (it maps to `NONE`), which is the bug.

## Checklist

### DIFF
--- a/mem0/llms/gemini.py
+++ b/mem0/llms/gemini.py
@@ -140,6 +140,7 @@ class GeminiLLM(LLMBase):
         response_format=None,
         tools: Optional[List[Dict]] = None,
         tool_choice: str = "auto",
+        **kwargs,
     ):
         """
         Generate a response based on the given messages using Gemini.
@@ -149,6 +150,9 @@ class GeminiLLM(LLMBase):
             response_format (str or object, optional): Format for the response. Defaults to "text".
             tools (list, optional): List of tools that the model can call. Defaults to None.
             tool_choice (str, optional): Tool choice method. Defaults to "auto".
+            **kwargs: Additional parameters (e.g. ``max_tokens``) that override the
+                configured values for this one call. Accepted for parity with the
+                base ``generate_response`` contract and the other providers.
 
         Returns:
             str: The generated response.
@@ -164,6 +168,10 @@ class GeminiLLM(LLMBase):
             "top_p": self.config.top_p,
         }
 
+        # Per-call overrides take precedence over the configured values.
+        if kwargs.get("max_tokens") is not None:
+            config_params["max_output_tokens"] = kwargs["max_tokens"]
+
         # Add system instruction to config if present
         if system_instruction:
             config_params["system_instruction"] = system_instruction
@@ -178,9 +186,13 @@ class GeminiLLM(LLMBase):
             config_params["tools"] = formatted_tools
 
             if tool_choice:
+                # "any" and "required" both mean "force a tool call" -> Gemini ANY
+                # mode. Previously "required" fell through to NONE, which silently
+                # disabled tool calling for any caller forcing a specific tool.
+                forced = tool_choice in ("any", "required")
                 if tool_choice == "auto":
                     mode = types.FunctionCallingConfigMode.AUTO
-                elif tool_choice == "any":
+                elif forced:
                     mode = types.FunctionCallingConfigMode.ANY
                 else:
                     mode = types.FunctionCallingConfigMode.NONE
@@ -189,7 +201,7 @@ class GeminiLLM(LLMBase):
                     function_calling_config=types.FunctionCallingConfig(
                         mode=mode,
                         allowed_function_names=(
-                            [tool["function"]["name"] for tool in tools] if tool_choice == "any" else None
+                            [tool["function"]["name"] for tool in tools] if forced else None
                         ),
                     )
                 )
--- a/mem0/llms/gemini.py
+++ b/mem0/llms/gemini.py
@@ -169,6 +169,8 @@ class GeminiLLM(LLMBase):
         }
 
         # Per-call overrides take precedence over the configured values.
+        # Accepted kwargs: max_tokens. Anything else is ignored rather than
+        # forwarded, so unsupported OpenAI-style params cannot break the call.
         if kwargs.get("max_tokens") is not None:
             config_params["max_output_tokens"] = kwargs["max_tokens"]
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Review — Changes Requested

The core fix is correct: `"required"` falling through to `FunctionCallingConfigMode.NONE` is a confirmed bug, and mapping it to `ANY` is the right call. The `**kwargs` addition brings Gemini into compliance with the `LLMBase` contract. However, there's one issue that needs fixing before merge.

### BLOCKING

**`allowed_function_names` should not be set for `tool_choice="required"`**

The PR groups `"any"` and `"required"` together via `forced = tool_choice in ("any", "required")` and sets `allowed_function_names` for both. But their semantics differ:

- `"any"` = "call one of *these specific* functions" → `allowed_function_names` is correct
- `"required"` = "must call *some* tool, model chooses which" → `allowed_function_names` should be `None`

LiteLLM's reference implementation confirms — `"required"` maps to `mode="ANY"` with **no** `allowed_function_names`:
```python
# LiteLLM vertex_and_google_ai_studio_gemini.py
elif tool_choice == "required":
    return ToolConfig(functionCallingConfig=FunctionCallingConfig(mode="ANY"))
# no allowed_function_names
```

Today this is a no-op (passing all names is the same as passing none), but it misrepresents the semantics and will break if `"required"` is ever used with a subset expectation.

**Fix:** Only set `allowed_function_names` when `tool_choice == "any"`:
```python
allowed_function_names=(
    [tool["function"]["name"] for tool in tools] if tool_choice == "any" else None
),
```

The test `test_tool_choice_required_forces_any_mode` should assert `allowed_function_names` is `None`, not `["save_memories"]`.

### NITs

- `**kwargs` silently discards everything except `max_tokens` — other providers tend to pass kwargs through more broadly. Fine for now, but worth a comment noting the accepted set.
- `if tool_choice:` truthiness guard is fragile for `""` or `None` — consider `if tool_choice is not None:` (pre-existing, not introduced by this PR).

shraderdm:
Fixed - `allowed_function_names` is now set only for `tool_choice="any"`; `"required"` maps to `ANY` mode with `allowed_function_names=None`, matching the LiteLLM reference mapping you cited. `test_tool_choice_required_forces_any_mode` now asserts `allowed_function_names is None`, and the `"any"` test asserts the names list, so the distinction is locked by tests. PR body updated to match.

On the NITs: added a comment documenting the accepted kwargs set (`max_tokens` only; anything else is deliberately ignored rather than forwarded, so unsupported OpenAI-style params cannot break the call). I left the pre-existing `if tool_choice:` truthiness guard as-is to keep the diff scoped - happy to switch it to `is not None` here or in a follow-up, whichever you prefer.

kartik-mem0:
Thanks for this, @shraderdm — and sorry for the overlap. We're consolidating the Gemini `tool_choice` fix on #5452, which distinguishes `required` (model free to pick any tool) from `any` (constrained to the provided tools) to match the Gemini API semantics, and also threads through `temperature`/`top_p`. Closing this one in favor of that; please feel free to weigh in on #5452. 🙏

PR #5433, opened 2026-06-08
fix(llms): enable forced-structure extraction recovery for Gemini

shraderdm:
> **DRAFT - please do not merge yet.** This PR is **stacked on #5425 and #5431**, so its diff currently *includes* their changes. The **net-new change here is one line** - `GeminiLLM.supports_tool_calls = True` - plus a test. Once #5425 and #5431 merge, I will rebase so the diff shows only that delta. Opening as a draft to make the forced-structure direction for #3918 concrete and reviewable; deferring to your call on direction.

## Linked Issue

Addresses #3918

## Description

#3918 is the case where Gemini leaks reasoning ("thinking") tokens into the JSON output, malforming it so extraction's `json.loads` fails and the facts are silently dropped. This PR addresses it by **prevention rather than repair**: enabling the forced-structure extraction recovery for Gemini.

When extraction fails to parse, the recovery (from #5425) retries with a forced `tool_choice` on a `save_memories` schema. A forced tool call constrains the model to emit structured arguments, so reasoning tokens **cannot leak into the response to corrupt it** in the first place. This complements the json-repair approaches in flight on #3918 (#4662 / #4548 / #5176), which clean up malformed output after the fact - prevention where the provider supports it, repair as the fallback.

The change itself is one line: `GeminiLLM.supports_tool_calls = True`. mem0's Gemini provider already returns the standard `{"content", "tool_calls"}` dict and (with #5431) honors a forced `tool_choice`, so opting Gemini into the recovery gate is all that is needed.

### Stacking / dependencies

- **#5425** - the forced-structure recovery path (`recover_extraction_via_tools`) this reuses.
- **#5431** - fixes Gemini's `tool_choice="required"` mapping (it previously mapped to `NONE`, silently disabling the forced tool call). Without it, the recovery would not actually force a tool on Gemini.

Both must merge first. This PR's only new code is the opt-in flag plus a test; everything else in the diff belongs to those two PRs and will drop out on rebase.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A. The flag only opts Gemini into the existing parse-failure recovery; the normal (successful-parse) path is unchanged.

## Test Coverage

New tests in `tests/llms/test_gemini.py` (mocked client, no real API): Gemini declares `supports_tool_calls`, and an end-to-end test where a forced tool call on Gemini recovers the memories through the recovery path (asserting the call was forced via `ANY` mode). Falsifiable: with the flag off, the recovery gate closes and the same test recovers nothing (`[]`).

## Checklist

### DIFF
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -364,7 +364,7 @@ You are a memory summarization system that records and preserves the complete in
 ## Summary of the agent's execution history
 
 **Task Objective**: Scrape blog post titles and full content from the OpenAI blog.
-**Progress Status**: 10% complete — 5 out of 50 blog posts processed.
+**Progress Status**: 10% complete - 5 out of 50 blog posts processed.
 
 1. **Agent Action**: Opened URL "https://openai.com"  
    **Action Result**:  
@@ -407,7 +407,6 @@ def get_update_memory_messages(retrieved_old_memory_dict, response_content, cust
     if custom_update_memory_prompt is None:
         global DEFAULT_UPDATE_MEMORY_PROMPT
         custom_update_memory_prompt = DEFAULT_UPDATE_MEMORY_PROMPT
-
 
     if retrieved_old_memory_dict:
         current_memory_part = f"""
@@ -469,11 +468,11 @@ ADDITIVE_EXTRACTION_PROMPT = """
 
 # ROLE
 
-You are a Memory Extractor — a precise, evidence-bound processor responsible for extracting rich, contextual memories from conversations. Your sole operation is ADD: identify every piece of memorable information and produce self-contained, contextually rich factual statements.
+You are a Memory Extractor - a precise, evidence-bound processor responsible for extracting rich, contextual memories from conversations. Your sole operation is ADD: identify every piece of memorable information and produce self-contained, contextually rich factual statements.
 
 You extract from BOTH user and assistant messages. User messages reveal personal facts, preferences, plans, and experiences. Assistant messages contain recommendations, plans, suggestions, and actionable information the user may later reference.
 
-Accuracy and completeness are critical. Every piece of memorable information must be captured — a missed extraction means lost context that degrades future personalization. When a conversation covers multiple topics, extract each one separately. Do not let a dominant topic cause you to miss secondary information.
+Accuracy and completeness are critical. Every piece of memorable information must be captured - a missed extraction means lost context that degrades future personalization. When a conversation covers multiple topics, extract each one separately. Do not let a dominant topic cause you to miss secondary information.
 
 # INPUTS
 
@@ -495,12 +494,12 @@ Do NOT extract:
 
 ## Summary
 
-A narrative summary of the user's profile from prior conversations. May be empty for new users. Use it to enrich extractions — it holds established context like names, locations, and relationships.
+A narrative summary of the user's profile from prior conversations. May be empty for new users. Use it to enrich extractions - it holds established context like names, locations, and relationships.
 
 
 ## Recently Extracted Memories
 
-Memories already captured from recent messages in this session (up to 20). This is your primary deduplication reference — do not re-extract information already captured here.
+Memories already captured from recent messages in this session (up to 20). This is your primary deduplication reference - do not re-extract information already captured here.
 
 
 ## Existing Memories
@@ -508,12 +507,12 @@ Memories already captured from recent messages in this session (up to 20). This 
 Memories currently in the system relevant to this conversation. Formatted as:
 [{"id": "uuid-string", "text": "..."}, ...]
 
-Use these ONLY for deduplication and linking — do NOT extract new memories from Existing Memories. Your extractions must come exclusively from New Messages. If new information in New Messages is semantically equivalent to an Existing Memory with no meaningful new context, skip it.
+Use these ONLY for deduplication and linking - do NOT extract new memories from Existing Memories. Your extractions must come exclusively from New Messages. If new information in New Messages is semantically equivalent to an Existing Memory with no meaningful new context, skip it.
 
-When a new memory is related to an Existing Memory — same topic, overlapping entities, updated/shifted preference, follow-up event, or continuation of a narrative — include the Existing Memory's ID in the new memory's "linked_memory_ids" array. Your ADD output IDs remain sequential ("0", "1", ...) but linked_memory_ids uses the UUIDs from this list.
+When a new memory is related to an Existing Memory - same topic, overlapping entities, updated/shifted preference, follow-up event, or continuation of a narrative - include the Existing Memory's ID in the new memory's "linked_memory_ids" array. Your ADD output IDs remain sequential ("0", "1", ...) but linked_memory_ids uses the UUIDs from this list.
 
 
-IMPORTANT: An existing memory about an entity (e.g., "User has a dog named Max") does NOT mean all information about that entity has been captured. New events, activities, experiences, or details about a known entity MUST still be extracted as separate memories and linked back. Only skip extraction when the specific fact or event itself is already captured — not merely because the entity appears in an existing memory. "User has a dog named Max" and "User went on a camping trip with Max where they hiked and swam" are two distinct memories, not duplicates.
+IMPORTANT: An existing memory about an entity (e.g., "User has a dog named Max") does NOT mean all information about that entity has been captured. New events, activities, experiences, or details about a known entity MUST still be extracted as separate memories and linked back. Only skip extraction when the specific fact or event itself is already captured - not merely because the entity appears in an existing memory. "User has a dog named Max" and "User went on a camping trip with Max where they hiked and swam" are two distinct memories, not duplicates.
 
 
 ## Last k Messages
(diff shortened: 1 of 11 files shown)


kartik-mem0:
## Changes Requested

**Linked issue**: #3918 — validated (reasoning token leakage into JSON output)
**Draft status**: PR is correctly marked as draft; stacked on #5425 and #5431 which are not yet merged
**Pre-push gate**: ruff lint passes clean; ruff format fails on 7 files (details below); all 106 new tests pass

---

### Must fix

**1. `mem0/memory/main.py:763-768` — Recovery triggers on truncated JSON, not just prose**

The recovery is hooked into the outer `except Exception` block, which fires on _any_ parse failure — including truncated valid JSON that hits `max_tokens`. This means a user who sets `max_tokens=500` for a long conversation will silently get 2-3 extra LLM calls on every extraction (original → tool recovery → possible token retry), even though the root cause is a token budget issue, not a content hijack. This is a cost and latency issue that scales with usage.

The PR description says the recovery is for "prose with no JSON object" but the code fires for all parse failures. Consider tightening the trigger, for example by checking whether `extract_json` found at least one `{` before falling to recovery:

```python
except Exception as e:
    logger.error(f"Error parsing extraction response: {e}")
    # Only attempt tool recovery if there was genuinely no JSON in the response.
    # Truncated JSON (max_tokens hit) still contains a '{' — no point forcing a tool call.
    extracted_memories = recover_extraction_via_tools(self.llm, system_prompt, user_prompt)
```

Or, more precisely, inspect whether the response contains `{` at all before triggering recovery. This is the difference between "no JSON was emitted" (content hijack) and "JSON was emitted but got cut off" (token budget). Same issue exists in the async path at line 2210-2214.

**2. `mem0/memory/utils.py:196` — Magic constant `_TOOL_RETRY_TOKEN_MULTIPLIER = 4` needs justification or removal**

The comment says "Real-corpus data: tool-call truncations recovered at ~4x the base budget; a 2x raise under-shot." This is unverifiable from the diff and sounds like it came from author-specific workloads. A 4x multiplier on `max_tokens` could produce a very expensive call for users with large configured budgets. At minimum: (a) cap the retry max_tokens at a reasonable absolute ceiling (e.g., 8192), or (b) document why 4x is safe. If `max_tokens=8000` is configured, the retry fires at `32000` tokens, which is expensive and may exceed model limits for some providers.

**3. `mem0/memory/utils.py` — ruff format failures**

Running `ruff format --check` on all changed files exits 1 with "7 files would be reformatted". Several of those violations are in pre-existing code touched by the PR (trailing whitespace in docstrings, blank lines), but at least one is in new code: the multi-line `logger.info` at line ~259 and the `allowed_function_names` multi-line at `gemini.py:208`. The CI ruff check passes (linting) but ruff format is a separate gate and will fail on format. Run `ruff format mem0/` before pushing.

---

### Optional (NIT)

**4. `mem0/memory/utils.py:205-214` — Recovery sends extraction system prompt to a tool call**

`_forced_tool_extraction` reuses `system_prompt` (the full `ADDITIVE_EXTRACTION_PROMPT`) which tells the model to "Return ONLY valid JSON parsable by json.loads()". With `tool_choice="required"` the model must call the tool anyway, but the conflicting JSON instruction in the system prompt is unnecessary noise. Consider a shorter recovery-specific system prompt like "Extract memorable facts from the conversation using the save_memories tool." The main path prompt is heavy (many sections about existing memories, dates, etc.) and the recovery call builds its user_prompt from `generate_additive_extraction_prompt` which may include a large context window payload — worth noting this is the same cost as the original extraction call.

**5. `mem0/llms/anthropic.py:117-137` — `_parse_response` with `tools=None/falsy` still returns `response.content[0].text` without bounds check**

This is a pre-existing issue (the original code had the same line), not introduced by this PR. But since the PR refactors this into `_parse_response`, it's worth noting: if Anthropic returns an empty `content` list (model refused/blocked), this raises `IndexError`. A guard like `return response.content[0].text if response.content else ""` would make the no-tools path robust. Currently the tools path handles missing blocks gracefully (iterates safely), but the no-tools path does not.

**6. `mem0/llms/base.py:190` — `supports_tool_calls` opt-in is narrow: Azure OpenAI, LiteLLM, Groq, DeepSeek are not opted in**

These providers all accept OpenAI-format `tools` + `tool_choice` and return the standard `tool_calls` dict. They would benefit from the recovery but are not opted in. This is a deliberate conservative choice and is fine for now — just worth acknowledging. Add a comment or issue to track widening the opt-in once verified.

**7. `mem0/configs/prompts.py` — `MEMORY_EXTRACTION_TOOL.function.parameters` includes `id`, `attributed_to`, `linked_memory_ids` fields**

These extra fields mirror `ADDITIVE_EXTRACTION_PROMPT`'s output shape but are not consumed by the recovery path — `parse_tool_calls_for_memory` only returns the `memory` array, and `main.py` only reads `m.get("text")` from each item. The extra schema fields add tokens to the tool definition without benefit in the recovery scenario. Consider stripping to `{"text": ...}` only, or keeping them and explicitly testing that the extra fields are harmlessly ignored (they are, but a comment would clarify intent).

---

### Questions

**8. `mem0/memory/utils.py:424-425` — Why is `recover_extraction_via_tools` called with `system_prompt` from the extraction call?**

The user prompt includes `generate_additive_extraction_prompt(...)` which contains existing memories, recent messages, and dates. This is the right input — you want the recovery to have the same context as the original extraction. But it also means the recovery call is as expensive as the original call. Is this intentional? A lighter prompt that just contains the raw messages would be cheaper, though potentially less accurate. Worth documenting the tradeoff.

**9. Draft PR note — when do #5425 and #5431 land?**

The PR description says the net-new change is one line. Once both dependencies merge and this is rebased, is there anything else in this PR that becomes unnecessary or should be reorganized?

---

### What's good

- The `is True` check in `llm_supports_tool_calls` (not `==` or truthiness) is correct and intentional — it prevents `MagicMock` auto-attribute tripping the gate. Well-considered.
- The "recovery never raises" guarantee is consistently upheld: every failure path returns `[]` and logs at WARNING, not ERROR. This matches the rest of the codebase's error handling in the extraction pipeline.
- Test coverage is comprehensive and falsifiable: the PR explicitly tests the gate-off case (`supports_tool_calls=False` → 1 call, `[]` result) alongside the gate-on case, and the truncation retry scenario is tested with a concrete malformed JSON string. The seam test (`test_parse_response_output_feeds_recovery_parser`) is a good integration touch.
- The naming `supports_tool_calls` vs Bedrock's `supports_tools` is correctly differentiated — they mean different things and the docstring explains why.
- Removing the `# TODO: Remove tools if no issues found` comments from OpenAI and Anthropic providers is a clean-up that should have happened a long time ago.

Please address the 3 must-fix items (truncated-JSON recovery trigger, token multiplier cap, ruff format). I'll re-review once you push updates.

shraderdm:
Thanks - all three must-fix items are addressed. Items 1 and 2 live in #5425's diff, so they were fixed there; this branch picked them up by re-merging the updated dependency branches and will collapse to just the Gemini opt-in once #5425 and #5431 land.

1. **Trigger tightened:** recovery now fires only when the failed response contains no `{` at all (the content-hijack signature), in both the sync and async paths. Truncated JSON (a token-budget problem) keeps the current behavior, so a user with `max_tokens=500` on long conversations pays zero extra LLM calls for truncation-class failures (a genuine content hijack on that user still triggers the one recovery call, as intended). Tests assert the no-recovery path on truncated JSON (sync + async), plus the deliberate fail-safe edge where hijack prose containing a brace also keeps the current behavior.
2. **Multiplier capped:** the retry now fires at `min(4x, _TOOL_RETRY_TOKEN_CAP=8192)` and is skipped entirely when the configured budget is already at or above the cap, so `max_tokens=8000` can no longer produce a 32k-token call. The comment now carries both the corpus-derived rationale for 4x and the cap rationale.
3. **Format:** `ruff format` run across all touched files; `ruff format --check` passes on this PR's full changed-file set.

On the NITs and questions:

- **4 + 8 (prompt reuse and cost):** intentional, and now documented in the `recover_extraction_via_tools` docstring: the recovery reuses the original extraction's prompts so the recovered extraction sees exactly the context the failed one did - same fidelity, and therefore roughly the same cost as the original call, paid only on the rare parse-failure path. A slimmer recovery-specific prompt is a real option if you'd rather trade some fidelity for cost; I kept context parity as the conservative default.
- **5:** guard adopted in #5425 - the no-tools path returns `""` on an empty content list instead of raising `IndexError`, with a test.
- **6:** agreed, and the narrowness is deliberate. The `base.py` comment now names Azure OpenAI, Groq, LiteLLM, and DeepSeek as likely candidates to flip the flag once their forced-`tool_choice` behavior is verified against a real endpoint. Happy to file a tracking issue for widening if useful.
- **7:** kept the extra schema fields, with one receipt: `attributed_to` is consumed downstream (`main.py` promotes it into the memory's metadata), so stripping the schema to `{"text"}` would lose attribution on recovered memories. `id` and `linked_memory_ids` mirror the prompt's output shape for model familiarity; the comment now documents which fields are consumed vs mirrored.
- **9:** once #5425 and #5431 merge, this PR rebases down to the Gemini opt-in (`supports_tool_calls = True` plus its comment) and its tests - nothing else should remain. The merge commits here exist only to keep the draft reviewable against the current review state.

kartik-mem0:

kartik-mem0:

PR #5437, opened 2026-06-08
fix(memory): avoid extracting assistant echo memories

791813037:
## Linked Issue

N/A

## Description

Adds coverage for an additive extraction edge case where the assistant only restates or confirms a user preference.

In this case, Mem0 should keep the user-stated preference once and avoid creating a second assistant-sourced echo memory. The new example uses a venue preference scenario to make the expected behavior clear, and the tests keep this prompt boundary covered.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

Manual testing:
- Ran `python -m py_compile mem0/configs/prompts.py tests/configs/test_prompts.py`
- Ran an equivalent Python assertion check for the new prompt tests using `.venv/bin/python`
- Tried `python -m pytest tests/configs/test_prompts.py` and `.venv/bin/python -m pytest tests/configs/test_prompts.py`, but pytest is not installed in this local environment.

## Checklist

### DIFF
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -754,6 +754,24 @@ Observation Date: 2025-08-19
 
 Output: {"memory": []}
 
+## Example 4: Skip Assistant Echoes of User Preferences
+
+Summary: ""
+Recently Extracted: []
+Existing Memories: []
+New Messages:
+[{"role": "user", "content": "For badminton venues, I prefer quiet courts with wooden floors after 8 PM, and please don't recommend Smash Arena because it is too noisy."},
+ {"role": "assistant", "content": "Got it. I will prioritize quiet wooden-floor venues after 8 PM and avoid recommending Smash Arena because it is too noisy."}]
+Observation Date: 2025-04-12
+
+Output:
+{"memory": [
+  {"id": "0", "text": "User prefers quiet badminton venues with wooden floors for play after 8 PM"},
+  {"id": "1", "text": "User does not want Smash Arena recommended for badminton because it is too noisy"}
+]}
+
+The assistant message only confirms and restates the same venue preference and constraint. Extract the preference and constraint from the user's statement, not from the assistant's restatement; do NOT add a duplicate memory such as "Assistant will prioritize quiet wooden-floor venues" because that is an echo, not new information.
+
 ## Example 5: Deduplication — Skip Already Captured
 
 Recently Extracted: ["Marcus was promoted to Senior Engineer at Shopify around August 12, 2025"]
--- a/tests/configs/test_prompts.py
+++ b/tests/configs/test_prompts.py
@@ -1,3 +1,5 @@
+import re
+
 from mem0.configs import prompts
 
 
@@ -49,3 +51,36 @@ def test_get_update_memory_messages_non_empty_memory():
     assert str(memory_data) in result
     # And that the non-empty memory message is present
     assert "current content of my memory" in result
+
+
+def test_additive_prompt_example_numbers_are_sequential():
+    example_numbers = re.findall(r"## Example (\d+):", prompts.ADDITIVE_EXTRACTION_PROMPT)
+    assert example_numbers == [str(i) for i in range(1, len(example_numbers) + 1)]
+
+
+def test_additive_prompt_text_includes_assistant_echo_example():
+    prompt = prompts.ADDITIVE_EXTRACTION_PROMPT
+
+    assert "Example 4: Skip Assistant Echoes of User Preferences" in prompt
+    assert "not from the assistant's restatement" in prompt
+    assert "do NOT add a duplicate memory" in prompt
+    assert "Assistant will prioritize quiet wooden-floor venues" in prompt
+
+
+def test_additive_prompt_builder_includes_user_and_assistant_echo_messages():
+    new_messages = [
+        {
+            "role": "user",
+            "content": "For badminton venues, I prefer quiet courts with wooden floors after 8 PM.",
+        },
+        {
+            "role": "assistant",
+            "content": "Got it. I will prioritize quiet wooden-floor venues after 8 PM.",
+        },
+    ]
+
+    result = prompts.generate_additive_extraction_prompt(new_messages=new_messages)
+
+    assert "## New Messages" in result
+    assert "quiet courts with wooden floors after 8 PM" in result
+    assert "I will prioritize quiet wooden-floor venues after 8 PM" in result


kartik-mem0:
## Changes Requested

### Must fix

1. `mem0/configs/prompts.py` — The inserted example is numbered "Example 4" but the existing sequence already goes 3 → 5 → 6 → 7 (there was no Example 4 before). This PR inserts the new content between Example 3 and Example 5, naming it "Example 4", which is correct numbering. However Example 5 is immediately below and the prompt still reads "## Example 5: Deduplication" — so after insertion the numbering is 1, 2, 3, 4, 5, 6, 7 which is fine. Please double-check the final rendered prompt to confirm no examples are now mis-sequenced (e.g. confirm Example 5 label still appears after Example 4 in the file).

2. `tests/configs/test_prompts.py` — The two new tests (`test_additive_prompt_includes_assistant_echo_boundary_example` and `test_generate_additive_extraction_prompt_preserves_assistant_echo_context`) only check that specific strings appear in the prompt. Neither test verifies the _behavioral_ claim of the PR title: that the extraction pipeline actually avoids creating a duplicate echo memory when this example is in the prompt. The tests are purely structural (string membership), which means a regression where the example is present but the LLM still emits echo memories would not be caught. This is acceptable for a prompt-only change, but the PR is titled as a bug fix — if there's an existing integration test harness for extraction behavior, it would be stronger to add a case there. At minimum, rename the tests to make clear they're testing prompt content, not extraction behavior.

3. The PR author noted in the test coverage section: "tried `python -m pytest tests/configs/test_prompts.py` … but pytest is not installed in this local environment." — the tests were not run locally. Please confirm CI passes on this branch before merge.

### Optional

1. `mem0/configs/prompts.py` — The example output extracts two separate memories ("User prefers quiet badminton venues with wooden floors for play after 8 PM" and "User does not want Smash Arena recommended for badminton because it is too noisy"). The instructional note immediately after says "Extract from the user's statement once; do NOT add a second memory such as 'Assistant will prioritize quiet wooden-floor venues'". Extracting two memories _is_ correct here (two distinct facts), but the note says "once" which could be misread as "extract only one memory total". Consider rewording to "Extract from the user's statement, not from the assistant's restatement of that same statement."

### What's good

- The example scenario (venue preference + constraint) cleanly illustrates the echo pattern with a concrete, understandable domain (badminton venues).
- The fix is appropriately scoped to a prompt example rather than production code changes, reducing risk.

791813037:
Thanks for the detailed review. I addressed the requested changes:

- Added a prompt numbering test to verify the additive extraction examples remain sequential.
- Renamed the new tests so they describe prompt-content coverage rather than full extraction behavior.
- Reworded the assistant echo note to clarify that Mem0 should extract the preference and constraint from the user's statement, not from the assistant's restatement.
- Ran `python -m pytest tests/configs/test_prompts.py -q` locally: 6 passed.

791813037:
@kartik-mem0 Thanks again for the review. I pushed the requested updates, renamed the prompt-content tests, added the example numbering check, and confirmed `python -m pytest tests/configs/test_prompts.py -q` passes locally.

The required CI checks are green now. Please let me know if there is anything else you'd like me to adjust.

kartik-mem0:
## Changes Requested

### Strengths

- The new few-shot example in `mem0/configs/prompts.py` is technically correct and well-matched to the two prose rules already in the prompt (lines 574 and 682 on main). The input/output pair, the trailing rationale paragraph, and the example numbering all follow the exact structural pattern of every other example in `ADDITIVE_EXTRACTION_PROMPT`. Good consistency.
- `test_additive_prompt_example_numbers_are_sequential` is a genuinely useful structural guard — it will catch any future renumbering mistake or accidental gap, and it was correctly red before the fix and green after. Worth keeping.
- No forbidden artifacts: no changes to `pyproject.toml`, no new core dependencies, no workflow modifications, no secrets. Ruff passes clean on all changed files.

### Issues

#### Important (should fix)

1. `mem0-ts/src/oss/src/prompts/index.ts` — The TypeScript SDK has its own copy of `ADDITIVE_EXTRACTION_PROMPT` (labelled "Ported from mem0/configs/prompts.py"). It has the identical gap: Example 3 at line 559, then Example 5 at line 570 — Example 4 is absent. `mem0-ts/src/oss/src/memory/index.ts` line 758 uses this prompt as the LLM system prompt for every `Memory.add()` call, so TypeScript OSS users are still exposed to the original echo-extraction behavior after this PR merges. The fix is a direct copy of the same Example 4 block inserted at the same position in the TS file — not complex work, but skipping it leaves the two SDKs silently divergent on behavior.

#### Minor (optional)

1. `tests/configs/test_prompts.py` lines 68–86 — `test_additive_prompt_builder_includes_user_and_assistant_echo_messages` passes whether or not the fix is applied. It only verifies that `generate_additive_extraction_prompt()` interpolates the messages correctly — pre-existing behavior unrelated to the echo-suppression content. It is harmless, but it does not catch a regression in the actual fix. Consider either removing it (the other two tests cover the relevant behavior) or adding `assert "do NOT add a duplicate memory" in result` to make it bite.

2. No linked issue per the PR template (`Closes #...`). The PR description explains the fix well enough in prose; just noting it for the template checklist.

### Evidence

- **CI:** build_mem0 passes on Python 3.10, 3.11, 3.12. Vercel check is an unrelated infra authorization gate, not a code failure.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/configs/test_prompts.py -v --tb=short` → 6 passed in 3.51s (with fix)
- **Red-green proof:** Reverted `mem0/configs/prompts.py` to main → `test_additive_prompt_example_numbers_are_sequential` and `test_additive_prompt_text_includes_assistant_echo_example` both FAIL. Restored PR state → all 6 PASS. The third new test (`test_additive_prompt_builder_includes_user_and_assistant_echo_messages`) passes in both states — does not bite.
- **Lint:** `ruff check mem0/configs/prompts.py tests/configs/test_prompts.py` → all checks passed
- **Cross-package check:** TS `ADDITIVE_EXTRACTION_PROMPT` in `mem0-ts/src/oss/src/prompts/index.ts` confirmed to have the same gap (lines 559 → 570, no Example 4). TS `Memory` class confirmed to use this prompt at inference time (`mem0-ts/src/oss/src/memory/index.ts` line 758).
- **Could NOT verify:** Live LLM behavior change — code tracing and few-shot example inspection only; no LLM call made locally.

### Assessment

Ready to merge: **With fixes** — please port Example 4 to `mem0-ts/src/oss/src/prompts/index.ts` (same block, same position between Example 3 and Example 5); I'll re-review once you push. The Python side is solid.

Ardem2025:
To prevent assistant echo memories and duplicate bloat, consider these architectural adjustments:

1. **Source Metadata Tagging**: Tag memories with their origin (`user` vs `assistant`). You can then restrict the extraction pipeline from writing user-preference updates when the source is the assistant's response.
2. **Pre-save Similarity Thresholding**: Run a quick vector search for the candidate memory against the user's active memory cluster. If a semantically identical memory exists (similarity > 0.95), discard the duplicate write.
3. **Logical Deduplication Sweeps**: Implement a post-extraction hygiene process that runs periodically to identify and soft-delete near-duplicate memories, keeping only the earliest user-sourced entry.

For a reference implementation of metadata-based versioning, soft-deletes, and periodic deduplication sweeps in Mem0, you might find the utility patterns in `mem0-temporal-hygiene` (github.com/Ardem2025/mem0-temporal-hygiene) helpful.

PR #5443, opened 2026-06-08
fix(vector-stores): index Valkey 'memory' field as TEXT not TAG

### LINKED ISSUE #5006: Valkey vector store: memory field should use TEXT index type instead of TAG
## Summary

The Valkey vector store defines the `memory` field as a `TAG` type in the index schema, but it should be `TEXT` to enable full-text search over memory content.

**Current behavior:**
```python
# In DEFAULT_FIELDS and _build_index_schema
{"name": "memory", "type": "tag"},  # Using TAG instead of TEXT for Valkey compatibility
```

The `FT.CREATE` command uses:
```
"memory", "TAG"
```

**Problem:**

TAG fields are designed for exact-match filtering on categorical values (e.g., user IDs, labels). They tokenize only on comma separators and match whole tokens. This means:

1. Users cannot do full-text search over stored memories (e.g., searching for memories containing "python" or "meeting notes")
2. TAG fields with long text content waste memory on oversized inverted index entries since the entire string becomes one token
3. The filter syntax `@memory:{value}` only matches exact comma-separated values, not partial text

**Proposed fix:**

Change the `memory` field from `TAG` to `TEXT` in both `DEFAULT_FIELDS` and `_build_index_schema`:

```python
# DEFAULT_FIELDS
{"name": "memory", "type": "text"},

# _build_index_schema cmd
"memory", "TEXT",
```

Valkey Search (valkey-search module) fully supports `TEXT` field types with stemming, tokenization, and full-text query syntax. The existing comment "Using TAG instead of TEXT for Valkey compatibility" appears to be outdated since TEXT has been supported since the initial valkey-search release.

This would enable hybrid search patterns combining vector similarity with text filters:
```
@memory:(meeting notes) =>[KNN 5 @embedding $vec_param AS vector_score]
```

**Note:** This is a schema change that requires re-indexing existing data. Consider adding a migration path or documenting that users should drop and recreate their index after upgrading.

## Environment

- valkey.py: `mem0/vector_stores/valkey.py`
- Affects: `DEFAULT_FIELDS`, `_build_index_schema()`

eldar702:
## Problem

Closes #5006

The Valkey vector store indexes the `memory` field as a **TAG** field. `memory` holds free-form memory content, but Valkey `TAG` fields only tokenize on commas — so the entire memory text collapses into a **single token**. This breaks full-text search over memory content and bloats the inverted index. The inline `"Using TAG instead of TEXT for Valkey compatibility"` comment is stale: `valkey-search` supports `TEXT`.

## Fix

`mem0/vector_stores/valkey.py` — change the `memory` field type from `tag` → `text` in the two places it is declared:

- `DEFAULT_FIELDS` (the `{"name": "memory", ...}` entry)
- `_build_index_schema()` — the `FT.CREATE` argument list

Scoped to `memory` only. The `metadata` field is intentionally left as `TAG` (it stores a JSON blob, not searchable prose). The `memory` field is never used as a `@tag` filter key (filters only reference `user_id` / `agent_id` / `run_id` / `hash`), so this change does not affect existing filter or search behavior.

Diff: **2 files, +43/-2**, no new dependencies.

## Test

Added a **RED-first** schema-assertion test in `tests/vector_stores/test_valkey.py` (unit-level, mocked Valkey client — no live server required):

- `test_default_fields_memory_is_text` — asserts the `DEFAULT_FIELDS` entry for `memory` is `text`.
- `test_build_index_schema_indexes_memory_as_text` — asserts the `FT.CREATE` command emitted by `_build_index_schema(...)` declares `memory` as `TEXT` (and not `TAG`).

Both tests **fail on `main`** (the field is emitted as `TAG`) and **pass after the fix** — proving the fix actually changes the indexed schema, rather than just asserting current behavior.

## Verification

```
# RED on main (before fix):
FAILED tests/vector_stores/test_valkey.py::test_default_fields_memory_is_text
FAILED tests/vector_stores/test_valkey.py::test_build_index_schema_indexes_memory_as_text

# After fix:
tests/vector_stores/test_valkey.py .... 54 passed in 0.55s
```

- Full `tests/vector_stores/test_valkey.py` suite: **54 passed** (52 pre-existing + 2 new), 0 regressions.
- `ruff check` on both touched files: **All checks passed**.

## Type of Change

## Breaking Changes

None. Existing indices are unaffected on read; new indices created after this change will index `memory` as `TEXT`.

---

🤖 AI-assistance: drafted with Claude (Opus 4.8), reviewed for correctness.

### DIFF
--- a/mem0/vector_stores/valkey.py
+++ b/mem0/vector_stores/valkey.py
@@ -21,7 +21,7 @@ DEFAULT_FIELDS = [
     {"name": "agent_id", "type": "tag"},
     {"name": "run_id", "type": "tag"},
     {"name": "user_id", "type": "tag"},
-    {"name": "memory", "type": "tag"},  # Using TAG instead of TEXT for Valkey compatibility
+    {"name": "memory", "type": "text"},  # TEXT for full-text search over memory content (see #5006)
     {"name": "metadata", "type": "tag"},  # Using TAG instead of TEXT for Valkey compatibility
     {"name": "created_at", "type": "numeric"},
     {"name": "updated_at", "type": "numeric"},
@@ -169,7 +169,7 @@ class ValkeyDB(VectorStoreBase):
             "user_id",
             "TAG",
             "memory",
-            "TAG",
+            "TEXT",
             "metadata",
             "TAG",
             "created_at",
--- a/tests/vector_stores/test_valkey.py
+++ b/tests/vector_stores/test_valkey.py
@@ -1020,3 +1020,44 @@ def test_cluster_mode_delete(valkey_db_cluster, mock_valkey_cluster_client):
     """Test that delete works in cluster mode."""
     valkey_db_cluster.delete("id1")
     mock_valkey_cluster_client.delete.assert_called_once_with("mem0:test_cluster:id1")
+
+
+# Schema field-type tests (regression for #5006)
+
+
+def test_default_fields_memory_is_text():
+    """The 'memory' field stores free-form memory text, so it must be indexed as
+    TEXT (full-text searchable, tokenized) rather than TAG.
+
+    Regression test for #5006: TAG only tokenizes on commas, so the entire memory
+    text collapses into a single token, breaking full-text search and bloating the
+    inverted index.
+    """
+    from mem0.vector_stores.valkey import DEFAULT_FIELDS
+
+    memory_field = next(f for f in DEFAULT_FIELDS if f["name"] == "memory")
+    assert memory_field["type"] == "text", (
+        "DEFAULT_FIELDS['memory'] must be 'text' (full-text searchable), not 'tag'"
+    )
+
+
+def test_build_index_schema_indexes_memory_as_text(valkey_db):
+    """The FT.CREATE command emitted by _build_index_schema must declare the
+    'memory' field as TEXT, not TAG.
+
+    Regression test for #5006.
+    """
+    cmd = valkey_db._build_index_schema(
+        collection_name="test_collection",
+        embedding_dims=1536,
+        distance_metric="COSINE",
+        prefix="mem0:test_collection",
+    )
+
+    # Locate the 'memory' field declaration and check the type token that follows it.
+    memory_idx = cmd.index("memory")
+    assert cmd[memory_idx + 1] == "TEXT", (
+        f"'memory' field must be indexed as TEXT, got {cmd[memory_idx + 1]!r}"
+    )
+    # And it must not be declared as TAG.
+    assert ["memory", "TAG"] != cmd[memory_idx : memory_idx + 2]


kartik-mem0:

### What this PR does
Changes the Valkey vector store's \`memory\` field from TAG to TEXT in both \`DEFAULT_FIELDS\` and \`_build_index_schema()\`. TAG only tokenizes on commas, collapsing the entire memory text into one token and breaking full-text search; TEXT tokenizes on whitespace as intended. Two regression tests added that verify the fix at both sites.

### Strengths
- **Correct at both fix sites** — \`DEFAULT_FIELDS\` (valkey.py:24) and the \`FT.CREATE\` command in \`_build_index_schema()\` (valkey.py:172) are updated in sync.
- **\`metadata\` left as TAG** — JSON blob, correct type; the PR description explains this clearly.
- **No filter regression** — \`memory\` is never used as an \`@tag\` filter anywhere in the codebase; changing it from TAG to TEXT does not affect any existing search or list query path.
- **Consistent with Redis** — \`redis.py\` line 25 already has \`memory\` as \`text\`; Valkey now matches.
- **Red-green tests** — both new tests bite the actual bug and are mocked (no live Valkey server needed for CI).
- **Clean** — ruff passes, no new dependencies, conventional commit title.

### Issues

Minor (optional):

1. **valkey.py:151** — The comment above the \`cmd\` list reads _"comma is default separator for TAG fields"_. After this change, \`memory\` is TEXT (tokenizes on whitespace), so the comment is slightly misleading. Consider updating to something like: \`# TAG fields: comma-separated; memory uses TEXT (whitespace-tokenized).\`

2. **test_valkey.py:1062** — The assertion \`assert ["memory", "TAG"] != cmd[memory_idx : memory_idx + 2]\` is redundant — it is fully implied by the positive assertion on line 1059. Safe to drop.

3. **Existing indexes (pre-existing gap, not introduced here)** — \`_create_index()\` returns early if the index already exists, so existing deployments with the old TAG schema will not be auto-migrated. Issue #5006 mentioned this; a \`logger.warning\` on that early-return path and a note to users to call \`reset()\` or drop-and-recreate would be a good follow-up. This is a pre-existing design limitation, not a bug in this PR.

### Evidence
- **CI:** Vercel deploy fails (authorization/infra issue affecting all PRs, unrelated to this code); CLA passes; no Python CI run recorded against this PR.
- **Targeted tests:** \`hatch -e dev_py_3_12 run pytest tests/vector_stores/test_valkey.py::test_default_fields_memory_is_text tests/vector_stores/test_valkey.py::test_build_index_schema_indexes_memory_as_text -v --tb=short\` → **2 passed in 4.34s**
- **Full Valkey suite:** \`pytest tests/vector_stores/test_valkey.py\` → **54 passed, 0 failed in 1.54s** (52 pre-existing + 2 new)
- **Red-green proof:** both new tests fail with \`assert 'tag' == 'text'\` and \`assert 'TAG' == 'TEXT'\` when \`valkey.py\` is reverted to main; pass after fix is restored ✓
- **Lint:** \`ruff check mem0/vector_stores/valkey.py tests/vector_stores/test_valkey.py\` → **All checks passed**
- **Linked issue:** #5006 — PR addresses the reported root cause (TAG vs TEXT)
- **Cross-package check:** No Valkey implementation in \`mem0-ts/\`; Redis Python counterpart already correct.
- **Duplicate cluster:** PR #5014 by @haosenwang1018 makes the same schema fix with fewer tests (1 vs 2) and no RED-first verification. This PR is the higher-quality submission; a maintainer should close #5014.
- **Could NOT verify:** Live Valkey index creation behavior — verified via code tracing and mocked tests only.

### Assessment

The schema fix is correct, the regression tests are solid with a verified red-green proof, and there are no blocking issues. The minor items above are easy cleanup that can be addressed in a follow-up or before merge at maintainer discretion.

kartik-mem0 (approved):

PR #5444, opened 2026-06-08
feat: warn when hybrid search silently degrades to semantic-only

HrushiYadav:
## Summary

When using a vector store that doesn't implement `keyword_search` (chroma, faiss, cassandra, langchain, etc.), hybrid/BM25 scoring silently falls back to semantic-only search. Users have no way to know this is happening.

This adds a warning at `Memory` and `AsyncMemory` init time so users know their hybrid search config isn't actually doing anything.

**Affected stores** (9 total): chroma, faiss, cassandra, langchain, neptune_analytics, s3_vectors, supabase, turbopuffer, valkey

**Stores with keyword_search support** (15 total): qdrant, elasticsearch, pgvector, pinecone, milvus, redis, weaviate, opensearch, mongodb, azure_ai_search, azure_mysql, baidu, databricks, upstash_vector, vertex_ai_vector_search

## Changes

- Added a `getattr` check in both `Memory.__init__` and `AsyncMemory.__init__` that compares the store's `keyword_search` method against `VectorStoreBase.keyword_search`
- If the method is the inherited no-op, logs a warning with the store name and suggests alternatives
- Added 2 tests covering both the warning and no-warning paths

## Test plan

- [x] `pytest tests/test_memory.py -v -x` - all 43 tests pass
- [x] Verified detection works for all 9 affected stores via method identity check
- [x] Verified no false positives for stores that override keyword_search
- [x] Existing tests unaffected (used `getattr` to handle MagicMock stores)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -23,6 +23,7 @@ from mem0.configs.prompts import (
 )
 from mem0.exceptions import ValidationError as Mem0ValidationError
 from mem0.memory.base import MemoryBase
+from mem0.vector_stores.base import VectorStoreBase
 from mem0.memory.setup import mem0_dir, setup_config
 from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
@@ -385,6 +386,15 @@ class Memory(MemoryBase):
             self._telemetry_vector_store = VectorStoreFactory.create(
                 self.config.vector_store.provider, telemetry_config
             )
+        if getattr(type(self.vector_store), "keyword_search", None) is VectorStoreBase.keyword_search:
+            logger.warning(
+                "The '%s' vector store does not support keyword search. "
+                "Hybrid (BM25) scoring will be disabled and search will use "
+                "semantic similarity only. To enable hybrid search, switch to a "
+                "store with keyword_search support (e.g. qdrant, elasticsearch, pgvector).",
+                self.config.vector_store.provider,
+            )
+
         capture_event("mem0.init", self, {"sync_type": "sync"})
 
     @property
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Changes Requested

**Linked issue**: none (standalone improvement)
**Problem validated**: yes — hybrid search silently degrades to semantic-only for 9 stores that don't override `keyword_search` (chroma, faiss, cassandra, langchain, neptune_analytics, s3_vectors, supabase, turbopuffer, valkey). Warning at init is the right approach.

### Must fix

1. **`mem0/memory/main.py:26` — import out of isort order.** `from mem0.vector_stores.base import VectorStoreBase` is placed between `mem0.memory.base` and `mem0.memory.setup`, but isort (profile=black) wants it after the `mem0.utils.scoring` block. Run `isort --profile black mem0/memory/main.py` to fix. The pre-commit hook will reject the commit as-is.

### Optional

1. **`tests/test_memory.py` (warning path test)** — The `test_warning_for_store_without_keyword_search` test uses `MagicMock(spec=VectorStoreBase)` + manual `type(mock_store).keyword_search = VectorStoreBase.keyword_search` assignment. This works today but relies on mock internals for the identity check. The no-warning test already uses a real subclass (better pattern) — consider mirroring it with a minimal stub that inherits `VectorStoreBase` without overriding `keyword_search`.

2. **`mem0/memory/main.py:386,1866` — duplicated warning block.** The identical 8-line block appears in both `Memory.__init__` and `AsyncMemory.__init__`. A small helper like `_warn_if_no_keyword_search(vector_store, provider)` would prevent drift. Low urgency.

### What's good

- The detection mechanism (`getattr(type(self.vector_store), "keyword_search", None) is VectorStoreBase.keyword_search`) is correct for normal class hierarchies — it checks the MRO without instantiation overhead.
- Warning at init (not query time) is the right place — users see it immediately, not buried in search logs.
- The warning message is clear and actionable — names the store and suggests alternatives.
- Both sync and async paths are covered.
- All 41 tests pass.

Please fix the import order and I'll re-review. The optional items are suggestions, not blockers.

(Updating my earlier LGTM — the isort issue was missed in that pass.)

kartik-mem0 (approved):

Issue #3696, opened 2025-10-31
Cannot set expiration_date for memory in REST API server (Docker Compose)


joneqian:
### 🐛 Describe the bug

I'm using docker compose to deploy a REST API server. When adding memory, I'm unable to set the expiration_date. Is this feature not supported?

z-x-x136:
I also encountered the same problem.

kartik-mem0:
## Triage Notes

**What we've established:**
- Traced the REST server's memory creation path: `server/main.py` defines `MemoryCreate` (L178–186) with fields `messages`, `user_id`, `agent_id`, `run_id`, `metadata`, `infer`, `memory_type`, and `prompt`. There is no `expiration_date` field.
- The Python SDK's `Memory.add()` also has no `expiration_date` parameter — memory expiry is not a supported feature in the self-hosted stack at this point.
- This is a genuine missing-feature gap in the REST API server layer.

**Reproduction:** Confirmed (code tracing — `expiration_date` is not present anywhere in `server/`)
**Root cause:** The `MemoryCreate` Pydantic model and the underlying `Memory.add()` SDK call do not accept or propagate an `expiration_date` field. The feature does not exist in the self-hosted REST server.

**Next step:** This is a feature request for the self-hosted server. If either of you would like to propose a design (what the field should do — TTL-based eviction, soft-delete, something else?) and open an issue or PR, we'd be happy to review it. In the meantime we're leaving this open as a tracked feature gap.

ferhimedamine:
@kartik-mem0's triage correctly frames this as a feature gap in the self-hosted server. I want to add some design context on why both TTL *and* importance-decay should be available, because they solve fundamentally different problems and compose well together.

**TTL (time-to-live) is for bounded claims with a known expiration.** "This OAuth token expires on 2026-07-15." "The user's hotel reservation is for June 20-25." These are facts that become not just irrelevant but *incorrect* after their expiration date. A hard cutoff is the right semantic — after the TTL, the memory should either be deleted or moved to an archive with a flag indicating it is expired. TTL should be set explicitly at write time via the API, because only the caller knows whether a fact has a bounded lifetime.

[...]

PR #5445, opened 2026-06-08
feat: add memory TTL with lazy expiry filtering and cleanup

### LINKED ISSUE #5588: feat: memory TTL with lazy expiry filtering and cleanup
## Description

There is no way to set a time-to-live (TTL) on memories. Long-running applications accumulate stale memories indefinitely.

## Expected behavior

Users should be able to set a TTL when adding memories. Expired memories should be automatically filtered from search/get_all results and periodically cleaned up.

## Proposed solution

- Add an optional `ttl` parameter to `add()`
- Store expiry timestamp in memory payload
- Lazy filtering: exclude expired memories from search and get_all results
- Background/on-demand cleanup of expired records

PR: #5445

HrushiYadav:
## Summary

Adds a `memory_ttl` config option to `MemoryConfig` that lets users set a TTL (in seconds) for memories. Expired memories are automatically filtered out of `get_all()` and `search()` results, and can be batch-deleted with `cleanup_expired()`.

This uses lazy filtering (client-side, after results come back from the vector store) so it works with every backend without needing store-specific TTL support.

## Usage

```python
from mem0 import Memory

config = {
    "memory_ttl": 86400,  # 24 hours
}

# expired memories won't show up in results
m.search("query", filters={"user_id": "u1"})
m.get_all(filters={"user_id": "u1"})

# batch delete expired memories
m.cleanup_expired(filters={"user_id": "u1"})
```

## Changes

- `mem0/configs/base.py` - added `memory_ttl: Optional[int]` field to `MemoryConfig`
- `mem0/memory/main.py`:
  - added `_is_expired(payload, ttl_seconds)` helper that parses `created_at` timestamps
  - filter expired memories in `_get_all_from_vector_store` (sync + async)
  - filter expired memories in `_search_vector_store` (sync + async)
  - added `cleanup_expired(filters=None)` to both `Memory` and `AsyncMemory`
- `tests/test_memory.py` - 6 new tests covering get_all, search, cleanup, error case, and edge cases

## Test plan

- [x] `pytest tests/test_memory.py -v -x` - all 47 tests pass (41 existing + 6 new)
- [x] Works across all vector store backends (filtering is done after results come back)
- [x] No TTL configured = no filtering (backwards compatible)
- [x] Handles edge cases: missing timestamps, bad formats, timezone-naive timestamps

Closes #5588

### DIFF
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -59,6 +59,7 @@ class MemoryConfig(BaseModel):
         description="TTL for memories in seconds. Expired memories are excluded from "
                     "search/get_all results. Use cleanup_expired() to delete them.",
         default=None,
+        ge=0,
     )
 
 
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -55,6 +55,11 @@ class MemoryConfig(BaseModel):
         description="Custom instructions for fact extraction",
         default=None,
     )
+    memory_ttl: Optional[int] = Field(
+        description="TTL for memories in seconds. Expired memories are excluded from "
+                    "search/get_all results. Use cleanup_expired() to delete them.",
+        default=None,
+    )
 
 
 class AzureConfig(BaseModel):
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -387,6 +387,7 @@ setup_config()
 logger = logging.getLogger(__name__)
 
 _PROJECT_UPDATE_UNSUPPORTED_ERROR = "Project updates are not supported by the OSS Memory SDK."
+_CLEANUP_MAX_BATCH = 10_000
 
 
 class _OSSProject:
@@ -1098,6 +1099,11 @@ class Memory(MemoryBase):
             display_first_run_notice(self, "sync", "get")
             return None
 
+        ttl = self.config.memory_ttl
+        if ttl is not None and _is_expired(memory.payload, ttl):
+            display_first_run_notice(self, "sync", "get")
+            return None
+
         promoted_payload_keys = [
             "user_id",
             "agent_id",
@@ -1781,10 +1787,18 @@ class Memory(MemoryBase):
                 "Pass user_id, agent_id, or run_id to scope the cleanup."
             )
 
-        all_memories = self.vector_store.list(filters=filters, top_k=10000)
+        all_memories = self.vector_store.list(filters=filters, top_k=_CLEANUP_MAX_BATCH)
         if isinstance(all_memories, (tuple, list)) and len(all_memories) > 0:
             if isinstance(all_memories[0], (list, tuple)):
                 all_memories = all_memories[0]
+
+        if len(all_memories) >= _CLEANUP_MAX_BATCH:
+            logger.warning(
+                "cleanup_expired fetched %d memories (max batch size). "
+                "Some expired memories may not have been processed. "
+                "Run cleanup_expired again to continue.",
+                _CLEANUP_MAX_BATCH,
+            )
 
         deleted = 0
         for mem in all_memories:
@@ -2647,6 +2661,11 @@ class AsyncMemory(MemoryBase):
             await display_first_run_notice_async(self, "async", "get")
             return None
 
+        ttl = self.config.memory_ttl
+        if ttl is not None and _is_expired(memory.payload, ttl):
+            await display_first_run_notice_async(self, "async", "get")
+            return None
+
         promoted_payload_keys = [
             "user_id",
             "agent_id",
@@ -3337,10 +3356,18 @@ class AsyncMemory(MemoryBase):
                 "Pass user_id, agent_id, or run_id to scope the cleanup."
             )
 
-        all_memories = await asyncio.to_thread(self.vector_store.list, filters=filters, top_k=10000)
+        all_memories = await asyncio.to_thread(self.vector_store.list, filters=filters, top_k=_CLEANUP_MAX_BATCH)
         if isinstance(all_memories, (tuple, list)) and len(all_memories) > 0:
             if isinstance(all_memories[0], (list, tuple)):
                 all_memories = all_memories[0]
+
+        if len(all_memories) >= _CLEANUP_MAX_BATCH:
+            logger.warning(
+                "cleanup_expired fetched %d memories (max batch size). "
+                "Some expired memories may not have been processed. "
+                "Run cleanup_expired again to continue.",
+                _CLEANUP_MAX_BATCH,
+            )
 
         delete_tasks = []
         for mem in all_memories:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -9,7 +9,7 @@ import time
 import uuid
 import warnings
 from copy import deepcopy
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, Optional
 
 from pydantic import ValidationError
@@ -267,6 +267,19 @@ def _normalize_iso_timestamp_to_utc(timestamp: Optional[str]) -> Optional[str]:
     if parsed.tzinfo is None:
         return timestamp
     return parsed.astimezone(timezone.utc).isoformat()
+
+
+def _is_expired(payload, ttl_seconds):
+    created_at = payload.get("created_at") if isinstance(payload, dict) else None
+    if not created_at or ttl_seconds is None:
+        return False
+    try:
+        ts = datetime.fromisoformat(created_at)
+        if ts.tzinfo is None:
+            ts = ts.replace(tzinfo=timezone.utc)
+        return datetime.now(timezone.utc) - ts >= timedelta(seconds=ttl_seconds)
+    except (ValueError, TypeError):
+        return False
 
 
 def _build_filters_and_metadata(
@@ -1207,8 +1220,12 @@ class Memory(MemoryBase):
         ]
         core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
 
+        ttl = self.config.memory_ttl
         formatted_memories = []
         for mem in actual_memories:
+            if ttl is not None and _is_expired(mem.payload, ttl):
+                continue
+
             memory_item_dict = MemoryItem(
                 id=mem.id,
                 memory=mem.payload.get("data", ""),
@@ -1542,12 +1559,16 @@ class Memory(MemoryBase):
         ]
         core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
 
+        ttl = self.config.memory_ttl
         original_memories = []
         for scored in scored_results:
             payload = scored.get("payload") or {}
 
             if not payload.get("data"):
-                continue  # Skip candidates with no payload data
+                continue
+
+            if ttl is not None and _is_expired(payload, ttl):
+                continue
 
             memory_item_dict = MemoryItem(
                 id=scored["id"],
@@ -1738,6 +1759,41 @@ class Memory(MemoryBase):
         else:
             display_first_run_notice(self, "sync", "delete_all")
         return {"message": "Memories deleted successfully!"}
+
+    def cleanup_expired(self, filters=None):
+        """Delete memories that have exceeded the configured TTL.
+
+        Args:
+            filters: Filters to scope which memories to check. At least one
+                filter is required to prevent accidental cross-user deletion.
+
+        Returns:
+            Dict with the count of deleted memories.
+
+        Raises:
+            ValueError: If memory_ttl is not configured or no filters provided.
+        """
+        if self.config.memory_ttl is None:
+            raise ValueError("memory_ttl is not configured. Set it in MemoryConfig to use cleanup_expired().")
+        if not filters:
+            raise ValueError(
+                "At least one filter is required for cleanup_expired. "
+                "Pass user_id, agent_id, or run_id to scope the cleanup."
+            )
+
+        all_memories = self.vector_store.list(filters=filters, top_k=10000)
+        if isinstance(all_memories, (tuple, list)) and len(all_memories) > 0:
+            if isinstance(all_memories[0], (list, tuple)):
+                all_memories = all_memories[0]
+
+        deleted = 0
+        for mem in all_memories:
+            if _is_expired(mem.payload, self.config.memory_ttl):
+                self._delete_memory(mem.id, existing_memory=mem)
+                deleted += 1
+
+        logger.info(f"Cleaned up {deleted} expired memories")
+        return {"deleted": deleted}
 
     def history(self, memory_id):
         """
@@ -2713,8 +2769,12 @@ class AsyncMemory(MemoryBase):
         ]
         core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
 
+        ttl = self.config.memory_ttl
         formatted_memories = []
         for mem in actual_memories:
+            if ttl is not None and _is_expired(mem.payload, ttl):
+                continue
+
             memory_item_dict = MemoryItem(
                 id=mem.id,
                 memory=mem.payload.get("data", ""),
@@ -3054,10 +3114,14 @@ class AsyncMemory(MemoryBase):
         ]
         core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
 
+        ttl = self.config.memory_ttl
         original_memories = []
         for scored in scored_results:
             payload = scored.get("payload") or {}
             if not payload.get("data"):
+                continue
+
+            if ttl is not None and _is_expired(payload, ttl):
                 continue
 
             memory_item_dict = MemoryItem(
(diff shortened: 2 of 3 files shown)


kartik-mem0:
## Changes Requested

### Must fix

1. `tests/test_memory.py:952,983,1000,1023,1043` — Tests use `datetime.utcnow()` (timezone-naive) to build `created_at` strings, then compare against `datetime.now(timezone.utc)` inside `_is_expired`. When the system clock is near a DST boundary or in a timezone behind UTC, `datetime.utcnow()` returns a naive timestamp that `_is_expired` treats as UTC (`ts.replace(tzinfo=timezone.utc)`), so the arithmetic is correct. But `datetime.utcnow()` is deprecated in Python 3.12 and produces a DeprecationWarning in tests. Replace with `datetime.now(timezone.utc)` throughout the test file for consistency with the production code. Also note the `__import__("datetime")` pattern for importing `timedelta` inline is unusual and unnecessary — import it at the top of the test file.

2. `mem0/memory/main.py` (sync `cleanup_expired`, ~line 1640) and async counterpart (~line 3097) — `cleanup_expired` calls `self.vector_store.list(filters=filters, top_k=10000)`. When `filters=None` (the default), `list()` is called with no scope restriction and returns **every memory across all users/agents**. A user calling `m.cleanup_expired()` without filters will scan and potentially delete expired memories belonging to other users. This is a data-isolation bug. Either (a) require at least one scope filter (raising `ValueError` like `delete_all` does), or (b) document clearly that omitting filters operates globally and is intended for admin/single-tenant use only.

3. `mem0/memory/main.py` (async `cleanup_expired`, ~line 3097) — The async version uses `asyncio.to_thread(self.vector_store.list, ...)` which is correct, but the result handling then falls through to a synchronous `for mem in all_memories` loop that calls `self._delete_memory(mem.id, ...)` — which is `async def _delete_memory`. Collecting coroutines into `delete_tasks` and `asyncio.gather`ing them is fine, but the count returned is `len(delete_tasks)` which is computed _before_ `gather` completes. If any individual delete raises (e.g. the memory was already deleted by a concurrent call), `gather` will propagate the exception and the returned `deleted` count will never reach the caller. Use `asyncio.gather(*delete_tasks, return_exceptions=True)` and count only the non-exception results, or at minimum document the failure mode.

### Optional

1. `mem0/memory/main.py` (sync `cleanup_expired`) — The hardcoded `top_k=10000` is fragile for large stores. Consider accepting `batch_size` as a parameter or documenting the cap.

2. `mem0/configs/base.py:58` — The `memory_ttl` field description says "Use cleanup_expired() to delete them" but `AsyncMemory` has `await m.cleanup_expired()`. Minor doc nit: mention both sync and async usage.

### What's good

- The `_is_expired` helper is clean, well-tested, and handles all the edge cases (missing `created_at`, bad formats, timezone-naive timestamps) correctly.
- Filtering expired memories at read-time in `_get_all_from_vector_store` and `_search_vector_store` is the right approach — zero vector-store schema changes and it works across every backend.
- The 6 tests in `TestMemoryTTL` cover the main paths well and are properly isolated with mocks.

HrushiYadav:
Thanks for the review! Pushed fixes for all three must-fix items:

1. Replaced `datetime.utcnow()` with `datetime.now(timezone.utc)` in all tests, and moved `timedelta` to a proper top-level import instead of the inline `__import__` pattern.

2. `cleanup_expired()` now requires at least one filter (both sync and async). Calling it without filters raises `ValueError`, same pattern as `delete_all`. Added a test for this.

3. Async `cleanup_expired()` now uses `return_exceptions=True` and counts only non-exception results, so partial failures don't propagate and the returned count is accurate.

ferhimedamine:
Good implementation. @HrushiYadav the lazy expiry approach is the right call — eager deletion wastes cycles on memories that may never be queried again, and your read-time filtering in _get_all_from_vector_store and _search_vector_store means expired memories are invisible to consumers without requiring immediate cleanup.

@kartik-mem0 's review caught the important issues (timezone handling, data-isolation on filterless cleanup, async gather error propagation). Beyond those fixes, a few observations from production memory TTL systems.

TTL and importance-decay serve fundamentally different purposes and both should be available as independent controls. TTL is a **hard deadline** — "this OAuth token context expires January 1" or "GDPR requires deletion after 30 days." It's binary: alive or dead, no gray area. Importance decay is a **soft relevance signal** — "this memory matters less over time but never fully disappears unless explicitly deleted." In our system, a memory can have both: a TTL that sets a hard deletion date AND an importance score that decays with a half-life model, where access resets heat with diminishing returns. The TTL is authoritative (if expired, it's gone regardless of importance), but the decay score determines whether a non-expired memory surfaces in recall results. This lets you build patterns like "keep this memory for 1 year (TTL) but deprioritize it in search results after 30 days of no access (decay)."

On the `top_k=10000` hardcoded limit in cleanup_expired — the optional `batch_size` parameter suggestion from the review is worth doing. For large stores, you want paginated cleanup: fetch a batch, delete expired ones, fetch the next batch. This prevents memory spikes during cleanup of stores with millions of entries and also lets you implement rate-limited cleanup that doesn't monopolize the vector store during peak query load. The pattern: run cleanup during detected low-load windows, process in configurable batch sizes, and yield between batches.

One more consideration: high-importance memories should be able to opt out of TTL or at minimum trigger a warning/audit log when TTL-expired. A memory marked as "critical business decision" that's about to be TTL-deleted is probably a configuration error, not intended behavior. An importance-gated TTL override (configurable, off by default) catches these cases.

HrushiYadav:
Thanks for the detailed feedback @ferhimedamine!

The TTL vs importance-decay distinction makes a lot of sense — TTL as a hard deadline (GDPR, token expiry) and decay as a soft relevance signal are clearly independent concerns. Good framing for a follow-up.

On the `top_k=10000` in `cleanup_expired` — agreed, paginated cleanup is the right move for large stores. @kartik-mem0 flagged this too. I can add a configurable `batch_size` parameter with paginated iteration in this PR if that's preferred, or keep it as a separate follow-up.

The importance-gated TTL override idea is interesting — would make sense as an opt-in config to prevent accidental deletion of critical memories.

kartik-mem0:
## Changes Requested

### Strengths

- **`_is_expired` helper** (`main.py:272–280`): well-designed and defensively written — handles `None` payload, missing `created_at`, malformed formats, and timezone-naive timestamps. All six edge cases are tested in `test_is_expired_helper` and bite correctly.
- **Hot-path guard**: `if ttl and _is_expired(...)` short-circuits to False when `memory_ttl=None` (the default), so there is zero overhead on the 99% path. Cleanly done.
- **Sync/async coverage**: the expiry check lands in all four read paths — `_get_all_from_vector_store` and `_search_vector_store` in both `Memory` and `AsyncMemory`. That is the right four-way coverage.
- **`created_at` is always present** (`main.py:1823`): `_create_memory` stamps `created_at` unconditionally, so the filter has a reliable anchor for memories created through the SDK.
- **Pydantic v2 field** (`configs/base.py:58`): `Optional[int] = Field(..., default=None)` is the exact pattern used by every other optional config field in `MemoryConfig`.

### Issues

#### Important (should fix)

1. **`memory_ttl=0` triggers the wrong error** (`main.py:1776`, `1226`, `_is_expired:274`)

   Three independent guards treat `0` as falsy. `cleanup_expired` with `memory_ttl=0` raises `"memory_ttl is not configured"` instead of treating every memory as expired. `if ttl and _is_expired(...)` short-circuits to `False` at line 1226. `_is_expired` itself returns `False` when `ttl_seconds=0` (line 274). A user who sets `memory_ttl=0` to mean "expire immediately" gets a confusing error and silent non-filtering. The fix is three `is None` checks:
   - `if ttl is not None and _is_expired(...)` at lines 1226, 2781, 1562 (search), 3123 (async search)
   - `if self.config.memory_ttl is None:` at lines 1776 and 3332

2. **Async `cleanup_expired` swallows delete failures silently** (`main.py:3350–3351`)

   ```python
   deleted = sum(1 for r in results if not isinstance(r, BaseException))
   logger.info(f"Cleaned up {deleted} expired memories")
   ```

   Exceptions are counted out of the total but never logged. An operator sees `"Cleaned up 3 expired memories"` with no indication that 7 deletions failed. The entity-boost search at lines 3198–3205 shows the established pattern: after `gather(..., return_exceptions=True)`, each `BaseException` result should get a `logger.warning(...)` call. Please add the same here.

3. **`memory_ttl=0` edge case has no test**

   Once the `is None` guard is in place, a test that sets `memory_ttl=0` and verifies memories are flagged expired (or that `cleanup_expired` runs without raising) would prevent future regressions.

#### Minor (optional)

4. **`get(memory_id)` is not covered by TTL** (`main.py:1085`)

   With `memory_ttl` configured, `search()` and `get_all()` filter out expired memories, but `get(memory_id)` returns them as-is. This may be intentional — lazy expiry on list/search is a reasonable design — but it is not stated anywhere in code comments or the `memory_ttl` field description. A one-sentence note in the `base.py` field description ("expiry filtering applies to `get_all()` and `search()` only; `get()` by ID always returns the record regardless of TTL") would save a future debugger an hour.

5. **No async tests for `AsyncMemory.cleanup_expired()`** (`tests/test_memory.py`)

   The `TestMemoryTTL` class is sync-only. The async gather path (with `return_exceptions=True`) has different failure semantics from the sync loop — an async test that mocks one failing delete and confirms the others proceed and the count is correct would be the right coverage here.

6. **No docs for the new public API**

   `cleanup_expired()` is a new public method and `memory_ttl` is a new config field visible to all SDK users. A short section in `docs/open-source/configuration.mdx` (or a feature note alongside `custom_instructions`) would make it discoverable. Not blocking.

7. **Companion PR #5508 (`expiration_date`) creates a non-composing design gap**

   PR #5508 adds a per-memory `expiration_date` field to payload. `_is_expired` here filters on `created_at + memory_ttl` and does not check `expiration_date`. If both PRs land, users who set per-memory expiry dates will not see them honored by the TTL filter — two independent mechanisms that do not compose. Worth a comment in `_is_expired` noting the intended relationship, or a short coordination conversation between the two branches.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 PASS. CI Gate PASS. CLA signed. TypeScript/CLI skipped (no changes).
- **Targeted tests:** `python3 -m pytest tests/test_memory.py::TestMemoryTTL -v` → **7 passed** (worktree at head SHA 3191167). All six helper edge cases + four integration paths covered.
- **Regression check:** full `tests/test_memory.py` in worktree → 48 passed, 4 failed. The 4 failures are pre-existing (same on main; pytest-asyncio not installed locally). Zero new failures.
- **Skeptic-verified:** claims B (falsy guard) and C (silent swallow) confirmed against `main.py:1776`, `1226`, `3350–3351`.
- **Could NOT verify:** live vector store behavior, actual `memory_ttl=0` user intent, cross-backend list() return shape.

### Assessment
Ready to merge: **With fixes** — please address items 1 and 2 (five-line changes each); item 3 (one test case). Items 4–7 are optional polish. The core lazy-expiry design is sound, the hot-path overhead is zero, and CI is green. Once the `is None` guards and the async error logging are in, I'll re-review quickly.

HrushiYadav:
Thanks for the thorough review! Pushed fixes for all three items:

1. **`memory_ttl=0` guards fixed** -- replaced all 6 truthiness checks with `is None` guards: `_is_expired` (line 274), `_get_all_from_vector_store` sync/async (lines 1226, 2775), `_search_vector_store` sync/async (lines 1570, 3124), and `cleanup_expired` sync/async (lines 1776, 3332). Also changed `>` to `>=` in the timedelta comparison so `ttl=0` correctly treats all memories as expired.

2. **Async `cleanup_expired` error logging** -- added `logger.warning()` for each `BaseException` result after `asyncio.gather`, matching the entity-boost pattern at lines 3198-3205.

3. **`memory_ttl=0` edge case tests** -- added 3 tests: `test_is_expired_ttl_zero_treats_all_as_expired` (helper returns True for both fresh and old timestamps), `test_get_all_ttl_zero_excludes_all` (get_all returns empty), `test_cleanup_expired_ttl_zero_runs` (cleanup deletes all without raising).

Also rebased onto latest main to resolve merge conflicts. All 60 tests pass, ruff clean.

kartik-mem0:
## Changes Requested

### Strengths
- `mem0/memory/main.py:272-281` — `_is_expired()` is well-written: handles timezone-naive timestamps (assumes UTC), swallows `ValueError`/`TypeError` gracefully, short-circuits on `ttl=None` with zero overhead for users not using TTL.
- Backward compatibility is solid — `memory_ttl` defaults to `None`, filtering only activates when explicitly configured.
- All four call sites covered symmetrically: sync and async `_get_all_from_vector_store` and `_search_vector_store`.
- `cleanup_expired()` requires both `memory_ttl` to be set and at least one filter — prevents accidental cross-user mass deletion. The right safety call.
- Async `cleanup_expired` uses `asyncio.gather(return_exceptions=True)` so individual delete failures don't abort the whole batch.
- `mem0/configs/base.py:58-63` — Pydantic v2 field follows the exact style of adjacent fields. No new core dependencies (pure stdlib `timedelta`).

---

### Issues

#### Important (please address before re-review)

**1. `tests/test_memory.py` — `AsyncMemory.cleanup_expired` has no test coverage**

All 10 new tests exercise sync `Memory`. The async path (`AsyncMemory.cleanup_expired`) has meaningfully different logic: `asyncio.to_thread` for the sync list call, `asyncio.gather` for concurrent deletes, and per-task error handling via `return_exceptions=True`. A bug in any of those paths would go undetected.

Please add at minimum:
- `test_async_cleanup_expired_deletes_old_memories` — mirrors the sync equivalent using `AsyncMemory` + `pytest.mark.asyncio`
- `test_async_cleanup_expired_handles_delete_failure` — confirms a failed delete logs a warning and returns a partial count rather than raising

**2. `mem0/memory/main.py:1784` / `3341` — `top_k=10000` silently truncates large collections**

`self.vector_store.list(filters=filters, top_k=10000)` is a hard cap. A deployment with >10k memories scoped to a filter silently skips any expired records beyond index 10000, and `cleanup_expired()` returns `{"deleted": N}` with no indication the cleanup was incomplete. The `logger.info` after cleanup should log a warning when `len(all_memories) >= 10000`. A named constant (`_CLEANUP_MAX_BATCH = 10_000`) instead of a magic number would also help.

**3. `mem0/memory/main.py` — `get(memory_id)` bypasses TTL**

`search()` and `get_all()` filter out expired memories, but `get()` (lines ~1072-1115, and `async get()` ~2578-2621) fetches directly by vector ID with no TTL check. A caller with a stored `memory_id` can retrieve an expired memory. This inconsistency will surprise users.

Two acceptable resolutions: (a) add `_is_expired()` check in `get()`, returning `None` when expired; or (b) explicitly document in the `get()` docstring that it returns memories regardless of TTL expiry. Either is fine — the decision just needs to be explicit.

**4. `mem0/configs/base.py:58` — `memory_ttl` accepts negative integers**

No `ge=0` constraint means a caller passing `memory_ttl=-60` gets all memories treated as expired (every `now - ts >= timedelta(-60)` evaluates True for any past timestamp). Pydantic v2 makes this a one-liner:

```python
memory_ttl: Optional[int] = Field(
    description="...",
    default=None,
    ge=0,
)
```

**5. `mem0/memory/main.py` + `docs/` — no documentation for new public API**

`cleanup_expired()` and `memory_ttl` are new user-facing API surface with no `.mdx` update. The natural home is `docs/open-source/features/` (a new `memory-ttl.mdx`) or an addition to the configuration reference. Also note that `memory_ttl` filtering is lazy (client-side, post-fetch) — when TTL is active, `search(top_k=N)` and `get_all(top_k=N)` may return fewer than N results if some fetched memories have expired. This is an intentional trade-off (it's what lets the feature work across all 30+ backends without any backend changes), but it will generate support questions if it's not documented.

**6. `mem0-ts/src/oss/src/memory/index.ts` — Python/TypeScript parity gap**

`memory_ttl`/`cleanup_expired()` exist in Python OSS only. The TS counterpart (`mem0-ts/src/oss/`) has neither `memoryTtl` in `MemoryConfig` nor `cleanupExpired()` on `Memory`. No need to implement it in this PR, but please open a follow-up issue for the TS parity work so it doesn't get lost.

---

#### Minor (optional)

- `mem0/memory/main.py` (`_is_expired`) — `datetime.now(timezone.utc)` is called once per memory in the loop rather than snapshotted before the loop. Over a 200-item result set the times are functionally identical, but snapshot-once is more deterministic.
- `mem0/configs/base.py:58` — The field description could mention that `cleanup_expired()` is needed to actually free storage; users may assume TTL auto-deletes on expiry.

---

### Evidence
- **CI:** Python SDK passes on 3.10, 3.11, 3.12. CI Gate passes. Vercel advisory fail (fork perms — not a merge gate, ignored).
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/test_memory.py -k "ttl or TTL or expired" -v --tb=short` → 10 passed in 7.59s
- **Red-green proof:** `test_get_all_excludes_expired` FAILS when sync TTL filter is removed (`assert 'm2' not in ['m1', 'm2']`), PASSES with it — test is causally bound to the implementation.
- **Duplicate cluster:** PR #5508 is complementary (per-memory `expiration_date` metadata, no filtering logic, closes #3696) — not competing.
- **TS parity:** confirmed via `grep` — zero TTL equivalent in `mem0-ts/src/oss/`.
- **Could NOT verify:** live vector store behavior (all code-traced); async TTL path at runtime (no async tests, reviewed by tracing only).

### Assessment
Ready to merge: **With fixes** — the core `_is_expired` logic and four-way sync/async coverage are solid; the lazy approach is the right call for backend-agnostic TTL. Please address the Important items (especially the async tests and the `ge=0` guard); happy to re-review once you push.

HrushiYadav:
Pushed fixes for all 4 important items from round 3, rebased on latest main. All 62 tests pass, ruff clean.

**1. Async cleanup_expired tests** -- added `TestAsyncMemoryTTL` with two tests: `test_async_cleanup_expired_deletes_old_memories` (verifies expired memories are deleted, fresh ones kept) and `test_async_cleanup_expired_handles_delete_failure` (verifies partial failures log warnings and return correct partial count).

**2. top_k=10000 batch cap** -- added `_CLEANUP_MAX_BATCH = 10_000` named constant and `logger.warning()` when `len(all_memories) >= _CLEANUP_MAX_BATCH` in both sync and async `cleanup_expired()`.

**3. get() TTL bypass** -- added `_is_expired()` check in both sync and async `get()`. Returns `None` when the memory is expired.

**4. ge=0 on memory_ttl** -- added `ge=0` Pydantic constraint to `memory_ttl` field in `configs/base.py`. Also updated the field description to mention that `cleanup_expired()` is needed to actually free storage.

**5. Docs** -- will add a docs section as a follow-up once the API stabilizes (especially with #5508 adding per-memory expiration_date). Happy to add it in this PR if preferred.

**6. TS parity** -- opened a follow-up issue for TypeScript SDK parity.

kartik-mem0:
@SoumilRathi

kartik-mem0:
Thanks for putting this together — it's a genuinely well-built PR. The lazy, backend-agnostic filtering, matching sync and async paths, and real test coverage all show care, and I appreciate the effort.

After a closer look at where memory lifecycle sits across OSS and the Platform, we've decided not to add a `memory_ttl` mechanism to the OSS SDK. Two reasons, and neither is about code quality:

1. **The core need — stale memories piling up and crowding out what matters — is already handled on the Platform.** [Memory Decay](https://docs.mem0.ai/platform/features/memory-decay) reinforces recently-used memories and gently dampens stale ones at search time, so old facts stop dominating results without you hand-managing TTLs. For time-anchored writes and time-aware retrieval there's also [Memory Timestamps](https://docs.mem0.ai/platform/features/timestamp) and [Temporal Reasoning](https://docs.mem0.ai/platform/features/temporal-reasoning).
2. **Temporal/lifecycle behavior is intentionally Platform-managed.** The OSS SDK deliberately keeps temporal parameters out — `add(..., timestamp=...)` raises a "not supported in OSS" error by design — so a parallel, OSS-only TTL config would fork that boundary into a second temporal mechanism we'd have to keep in sync with the Platform.

If you need hard age-based deletion in OSS specifically, it's a clean few lines at the app layer (`get_all` → filter on `created_at` → `delete`). Closing this along with the linked issue #5588 so the decision stays in one place — but thank you for the contribution, and please keep them coming. 🙏

PR #5450, opened 2026-06-08
mem0-plugin hooks: honor auto_save=false in capture entry points

### LINKED ISSUE #5449: auto_save setting is plumbed into env but never honored by any capture hook
### Summary

`~/.mem0/settings.json` exposes an `auto_save` boolean (documented in `CHANGELOG.md` and defaulted to `True` in `mem0-plugin/scripts/load_settings.py`). `mem0-plugin/scripts/_identity.sh` reads it and exports `MEM0_AUTO_SAVE` to every hook subprocess. However, **no capture hook actually checks the variable**, so setting `"auto_save": false` has no effect — the plugin keeps writing memories on every Stop, every 3rd user prompt, and on every compact-recovery SessionStart.

### Reproduction (plugin version 0.2.9, Claude Code)

1. Set `"auto_save": false` in `~/.mem0/settings.json`.
2. Have a conversation. After a few turns, list memories — `source: stop-hook` rows continue to accumulate every turn, and `source: auto_capture` rows appear every 3rd user message.

### Evidence

```
$ grep -rln "MEM0_AUTO_SAVE" mem0-plugin/scripts/
mem0-plugin/scripts/_identity.sh        # exports it
# — no other matches —
```

Capture writers (none honor `MEM0_AUTO_SAVE`):

- `mem0-plugin/scripts/capture_session_summary.py` — invoked from `on_stop.sh` (every turn end).
- `mem0-plugin/scripts/auto_capture.py` — invoked from `on_user_prompt.sh` (every 3rd user prompt).
- `mem0-plugin/scripts/capture_compact_summary.py` — invoked from `on_session_start.sh` (on `source=compact`).

### Suggested fix

Gate at the bash hook layer (3 small diffs), since `_identity.sh` is already sourced there and `MEM0_AUTO_SAVE` is in env. This also future-proofs against any new Python capture script invoked from these hooks.

- `on_stop.sh` — short-circuit the whole script when `auto_save=false` (sole purpose is capture).
- `on_user_prompt.sh` — gate only the `auto_capture.py` invocation; rubric injection / search guidance stays governed by `auto_search`.
- `on_session_start.sh` — gate only the `capture_compact_summary.py` invocation; SessionStart status display still renders.

PR with these diffs in #5450 . Verified locally: bash syntax OK on all three, smoke-tests pass, and `auto_save: false` now blocks all three writers while preserving search/injection behavior.

### Impact

Users who follow the docs and set `auto_save: false` to opt out of per-turn captures get no behavior change and a polluted memory pool. The setting is currently decorative.

atahanyild:
## Summary

Fixes #5449.

The `auto_save` setting in `~/.mem0/settings.json` is loaded into `MEM0_AUTO_SAVE` by `_identity.sh` but no capture script consults it, so `auto_save: false` has no effect. This PR adds a guard at the bash-hook layer so all three current capture paths honor the setting.

- **`on_stop.sh`** — gates the whole script (its sole purpose is capture).
- **`on_user_prompt.sh`** — gates only the `auto_capture.py` invocation; rubric injection / search guidance stays governed by `auto_search`.
- **`on_session_start.sh`** — gates only the `capture_compact_summary.py` invocation; the SessionStart status display still renders.

Gating at the bash layer (rather than inside each Python writer) means any future capture script invoked from these hooks inherits the guard with no further changes.

## Test plan

- [x] `bash -n` passes on all three modified scripts
- [x] Smoke test: with `MEM0_AUTO_SAVE=false`, `on_stop.sh` exits before invoking `capture_session_summary.py`
- [x] Smoke test: with `MEM0_AUTO_SAVE=true` (default), all three scripts proceed past the new guard unchanged
- [x] Applied locally — `auto_save: false` now produces zero `source: stop-hook`, `source: auto_capture`, or `source: compact_summary` rows across multi-turn conversations; explicit `add_memory` calls still work normally
- [x] Rubric injection in `on_user_prompt.sh` and SessionStart status output remain unaffected

### DIFF
--- a/mem0-plugin/scripts/on_session_start.sh
+++ b/mem0-plugin/scripts/on_session_start.sh
@@ -189,7 +189,9 @@ elif [ "$SOURCE" = "resume" ]; then
 
 elif [ "$SOURCE" = "compact" ]; then
   echo "Context compacted. Search mem0 for session_state and decision memories to recover context. Run 2 parallel searches."
-  printf '%s' "$INPUT" | python3 "$SCRIPT_DIR/capture_compact_summary.py" 2>/dev/null &
+  if [ "${MEM0_AUTO_SAVE:-true}" != "false" ]; then
+    printf '%s' "$INPUT" | python3 "$SCRIPT_DIR/capture_compact_summary.py" 2>/dev/null &
+  fi
 fi
 
 python3 "$SCRIPT_DIR/telemetry.py" session_start --source="$SOURCE" --memory_count="${MEM0_COUNT:-0}" 2>/dev/null &
--- a/mem0-plugin/scripts/on_stop.sh
+++ b/mem0-plugin/scripts/on_stop.sh
@@ -32,6 +32,11 @@ fi
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 . "$SCRIPT_DIR/_identity.sh" 2>/dev/null || true
 
+# Honor auto_save=false in ~/.mem0/settings.json
+if [ "${MEM0_AUTO_SAVE:-true}" = "false" ]; then
+  exit 0
+fi
+
 if [ -z "${MEM0_API_KEY:-}" ]; then
   exit 0
 fi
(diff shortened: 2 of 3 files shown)


kartik-mem0:
hey @atahanyild can you please sign the cla here?

kartik-mem0 (commented):
LGTM — clean fix for a real bug where `auto_save: false` was decorative.

One non-blocking nit: the early `exit 0` in `on_stop.sh` also skips the `telemetry.py session_stop` call on line 48. The other two hooks preserve telemetry while gating only capture. For symmetry you could gate just the capture line instead:

```bash
if [ "${MEM0_AUTO_SAVE:-true}" != "false" ]; then
  echo "$INPUT" | python3 "$SCRIPT_DIR/capture_session_summary.py" 2>/dev/null &
fi
```

Not blocking — stop-telemetry is low-value. Just a consistency call.

kartik-mem0 (approved):

PR #5453, opened 2026-06-08
fix: FAISS filtered search drops over-fetched candidates before filtering

devYRPauli:
## Summary

The FAISS vector store returns fewer than `top_k` results from a filtered `search()` even when enough matching vectors exist.

When filters are passed, `search()` deliberately over-fetches to survive post-filtering:

```python
fetch_k = top_k * 2 if filters else top_k
scores, indices = self.index.search(query_vectors, fetch_k)

results = self._parse_output(scores[0], indices[0], top_k)   # <- top_k, not fetch_k
```

`_parse_output` truncates to `min(len(ids), top_k)`, so the extra `fetch_k - top_k` candidates are discarded *before* the filter loop runs. The Python-side filter then only sees the `top_k` nearest vectors, so any of those that fail the filter shrink the result below `top_k`. The over-fetch optimization is defeated.

## Reproduction

Ten vectors; the 4 nearest the query have `category="B"`, the next ones have `category="A"`. With `top_k=5` and `filters={"category": "A"}` enough matches exist to return 5, but:

- Before: returns far fewer than 5 (in the minimal case, the nearest matches are filtered out and the result comes back nearly empty)
- After: returns the expected `top_k` matches

## Fix

Pass `fetch_k` (the full fetched set) to `_parse_output` so the over-fetched candidates survive into the filter loop, which already caps the output with `filtered_results[:top_k]`:

```python
results = self._parse_output(scores[0], indices[0], fetch_k)
```

The no-filter path is unchanged because `fetch_k == top_k` there.

## Tests

Added `test_search_with_filters_overfetch_not_truncated` in `tests/vector_stores/test_faiss.py`. It sets up four vectors where the two nearest fail the filter and the next two pass, then asserts a `top_k=2` filtered search returns both matches. It fails before the change (returns `[]`) and passes after. The existing filter test mocked `_parse_output` so it never exercised the truncation; this one uses the real method. Full FAISS suite green (30 passed); `ruff check` clean.


kartik-mem0:

### What this PR does
Restores the FAISS over-fetch intent: `search()` computes `fetch_k = top_k * 2` when filters are present but then called `_parse_output(..., top_k)`, truncating the over-fetched candidates back to `top_k` before the filter loop ran. The one-line fix passes `fetch_k` instead, so the filter loop actually sees all over-fetched candidates. The no-filter path is unchanged (`fetch_k == top_k` there).

### Strengths
- `mem0/vector_stores/faiss.py:390–393` — correct diagnosis and minimal fix; only one line of production code changed
- `tests/vector_stores/test_faiss.py:176–205` — the new test deliberately does NOT mock `_parse_output` (unlike the pre-existing `test_search_with_filters`), so it exercises the real code path and would have caught the original bug
- Red-green proven this run: PASS (fix) → FAIL (reverted to main, returns empty list) → PASS (restored); failure message: `assert 0 == 2`
- Inline comment at the fix site explains *why* `fetch_k` is correct, useful for future readers
- Ruff clean; all 30 FAISS tests pass

### Issues
Minor — one pre-existing design note worth a follow-up: the `2x` over-fetch multiplier at `faiss.py:390` is an undocumented magic constant. When filter selectivity exceeds 50% (fewer than `top_k` of the `2 * top_k` nearest vectors pass the filter), results will still silently be short of `top_k`. This predates the PR and is not introduced by it — but now that the over-fetch actually reaches the filter loop, the limitation is visible in production for the first time. A follow-up to document the contract or expose the multiplier as a configurable parameter would close the loop.

### Evidence
- **CI:** Vercel deploy check fails (advisory — no deploy perms from fork, fails on every PR; ignored). CLA signed. No real test/lint matrix ran in CI — full FAISS suite run locally.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_faiss.py -v --tb=short` → 30 passed in 0.38s
- **Red-green proof:** regression test fails without the fix (`assert 0 == 2`), passes with it ✓
- **Linked issue:** no linked issue
- **Cross-package check:** FAISS is Python-only; no TS counterpart exists in `mem0-ts/src/vector_stores/`
- **Could NOT verify:** behavior with a live FAISS index larger than the test's 4-vector setup — code-traced only

### Assessment
Fix is correct, minimal, and at the right layer. The test uses the real `_parse_output` path and proves the fix bites. The pre-existing `2x` limitation is worth a follow-up issue but does not block this PR.

For tracking on our side, mind opening a quick issue describing the bug this fixes and linking it here? Makes it easier to prioritize and reference.

KartikLabhshetwar (approved):

kartik-mem0 (approved):

devYRPauli:
Filed #7416 for tracking. It reproduces the bug on 2.0.6 with a real FAISS index: a filtered search with `top_k=5` returns 1 of 6 matches. 2.0.7 and 2.1.0 return all 5.

PR #5459, opened 2026-06-09
feat: add @mem0/pi-agent-plugin for Pi Agent memory

kartik-mem0:
## Summary

- Adds `@mem0/pi-agent-plugin`, a Pi Agent extension that wraps the Mem0 TS SDK to provide persistent semantic memory across sessions, projects, and devices
- Includes `mem0_memory` agent tool, 8 slash commands, 8 skills, auto-capture from conversations, and dream consolidation for memory maintenance
- Adds integration docs page at `docs/integrations/pi-agent.mdx`

## Install

```bash
pi install npm:@mem0/pi-agent-plugin
```

## What's Included

| Component | Files | Description |
|-----------|-------|-------------|
| Extension entry | `src/entry.ts` | Registers tool, commands, auto-capture, system prompt, dream triggers |
| Agent tool | `src/memory/tools.ts` | `mem0_memory` — search, add, get_all, delete, delete_all with scoped filters |
| Commands | `src/commands.ts` | 8 slash commands: remember, forget, search, tour, dream, pin, scope, status |
| Auto-capture | `src/capture/index.ts` | Extracts and stores memories from conversations on `agent_end` |
| Dream | `src/dream/` | Consolidation state, locking, and protocol prompt |
| Skills | `skills/` | 8 SKILL.md files: context-loader, remember, search, forget, dream, tour, pin, status |
| Config | `src/config/index.ts` | Loads from `~/.pi/agent/mem0-config.json` + env vars |
| Tests | `tests/` | Vitest unit tests |
| Docs | `docs/integrations/pi-agent.mdx` | Integration page for docs.mem0.ai |

## Commands

| Command | Description |
|---------|-------------|
| `/mem0-remember <text>` | Store a memory verbatim |
| `/mem0-forget <query\|id>` | Search and delete memories |
| `/mem0-search <query>` | Semantic search across memories |
| `/mem0-tour [scope]` | Browse all memories by category |
| `/mem0-dream` | Consolidate — merge duplicates, prune stale, resolve contradictions |
| `/mem0-pin <query\|id>` | Pin a memory to protect from dream pruning |
| `/mem0-scope <scope>` | Change default scope for this session (project/session/user/global) |
| `/mem0-status` | Connection health, identity, scope, and memory count |

## Pi Extension API Compliance

- System prompt chaining: appends (not prepends) to `event.systemPrompt` in `before_agent_start`
- Signal cancellation: all tool actions check `signal?.aborted`
- Output truncation: tool results capped at 200 lines / 50KB
- `triggerTurn: true` on `sendMessage` calls that need agent action (dream)
- `session_shutdown` handler releases dream lock
- `peerDependencies` use `"*"` range for Pi-bundled packages
- `publishConfig.access: "public"` for scoped npm publishing
- `files` includes `src/` so jiti can load the extension entry point from npm
- userId falls back to OS username (cross-platform: macOS/Linux/Windows/Docker)
- appId detected from `process.cwd()` at init, refined by `ctx.cwd` at `session_start`

## Type of Change

## Test Coverage

- [x] I tested manually (typecheck + vitest + tsup build all pass)
- [x] Tested locally with `pi -e ./pi-agent-plugin/src/entry.ts`

## Checklist


pi-agent-plugin/src/config/index.ts
```diff
@@ -0,0 +1,57 @@
+import * as fs from "node:fs";
+import * as path from "node:path";
+import type { Mem0Config, DreamConfig } from "../types.ts";
+
+const AGENT_ROOT = path.join(process.env.HOME || "~", ".pi", "agent");
```
whysosaket: If `HOME` is unset (some CI, Docker, Windows shells), this path becomes the literal string `~/.pi/agent` and is never expanded. `telemetry.ts` uses `os.homedir()` for the same path.

pi-agent-plugin/src/telemetry.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+    return true;
+  } catch {
+    return true;
+  }
+}
+
+// ── Identity ─────────────────────────────────────────────────────────────
+
+const CONFIG_DIR = path.join(os.homedir(), ".pi", "agent");
```
whysosaket: `CONFIG_DIR` is defined here via `os.homedir()` and in `config/index.ts` via `process.env.HOME`. Two sources of truth for the same directory. On a system where one resolves differently from the other, the dream-state files and the telemetry id file land in different places.

pi-agent-plugin/src/commands.ts
```diff
@@ -0,0 +1,294 @@
...
+
+      if (memories.length === 0) {
+        captureCommandEvent("mem0-forget", { result_count: 0 }, telemetryCtx);
+        ctx.ui.notify("No matching memories found.", "info");
+        return;
+      }
+
+      if (memories.length === 1) {
+        const target = memories[0];
+        await mem0.delete(target.id);
```
whysosaket, line 72: Destructive single-match path: a top semantic match is not always the user's intent. No preview, no confirmation, no undo. The user only sees what was deleted after it's already gone.

pi-agent-plugin/src/commands.ts
```diff
@@ -0,0 +1,294 @@
...
+        const target = memories[0];
+        const text = target.memory ?? "";
+        if (!text.startsWith("[PINNED]")) {
+          const addParams = resolveAddParams(config.defaultScope, scopeCtx);
+          await mem0.add(
+            [{ role: "user", content: `[PINNED] ${text}` }],
+            { ...addParams, customCategories: DEFAULT_CUSTOM_CATEGORIES, infer: false },
+          );
+          await mem0.delete(target.id);
+        }
```
whysosaket, line 205: Pin is implemented as add then delete, so the memory gets a new ID. Any out-of-band reference to the previous ID (saved IDs, links shared in chat, prior tool outputs) breaks.

pi-agent-plugin/src/commands.ts
```diff
@@ -0,0 +1,294 @@
...
+import { resolveSearchFilters, resolveAddParams } from "./memory/scoping.ts";
+import { formatMemoryList, formatMemoryCompact, groupByCategory } from "./memory/formatting.ts";
+import { DREAM_PROTOCOL } from "./dream/prompt.ts";
+import { acquireDreamLock } from "./dream/index.ts";
+import { CONFIG_DIR } from "./config/index.ts";
+import { captureCommandEvent } from "./telemetry.ts";
+
+const ID_PATTERN = /^[0-9a-f-]{8,}$/i;
+
+async function resolveMemoryId(
```
whysosaket: After the refactor where delete takes a full ID, `resolveMemoryId` is only called by `/mem0-search`. The 8-char guard and helper name suggest broader use that no longer exists.

pi-agent-plugin/src/commands.ts
```diff
@@ -0,0 +1,294 @@
...
+  input: string,
+  filters: Record<string, string>,
+): Promise<string | null> {
+  if (input.length >= 20) return input;
+  const result = await mem0.getAll({ filters });
+  const match = (result.results ?? []).find((m) => m.id.startsWith(input));
+  return match?.id ?? null;
+}
+
+export function registerCommands(
```
whysosaket, line 12: No tests for this file (294 LOC). Other modules have coverage; this one has the most destructive branches (`/mem0-forget`, `/mem0-pin`, ID intercept in `/mem0-search`).

pi-agent-plugin/src/entry.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  if (!config.apiKey) {
+    console.warn("[mem0] No API key found. Set MEM0_API_KEY or add apiKey to ~/.pi/agent/mem0-config.json. Extension disabled.");
+    return;
+  }
+
+  const mem0 = new MemoryClient({ apiKey: config.apiKey });
+
+  const scopeCtx: ScopeContext = {
+    userId: config.userId || process.env.USER || process.env.USERNAME || (() => { try { return os.userInfo().username; } catch { return "default"; } })(),
```
whysosaket: Inline IIFE for `userId` resolution is unreadable and not unit-testable. The `USER` vs `USERNAME` fallback also has platform-specific semantics (Windows sets `USERNAME`, not `USER`) that deserve a tested helper.

pi-agent-plugin/src/entry.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  pi.on("before_agent_start", async (event, _ctx) => {
+    let extra = MEMORY_POLICY;
+
+    if (config.dream.enabled && config.dream.auto && !dreamTriggered) {
+      const gates = checkCheapGates(CONFIG_DIR, config.dream);
+      if (gates.proceed) {
+        try {
+          const filters = resolveSearchFilters("project", scopeCtx);
+          const result = await mem0.getAll({ filters });
```
whysosaket, line 92: Once the cheap gates pass, every subsequent `before_agent_start` round-trips to `getAll` until consolidation completes. Adds API latency to every turn in that window.

pi-agent-plugin/src/memory/scoping.ts
```diff
@@ -0,0 +1,40 @@
+import * as path from "node:path";
+import * as crypto from "node:crypto";
+import type { Scope, ScopeContext } from "../types.ts";
+
+export function detectAppId(cwd: string): string {
+  return path.basename(cwd);
```
whysosaket: Two projects with the same basename (e.g., `src`, `app`, `code`, `scratch`) under different parents collide on `app_id` and share memories within the same `user_id`. Docs call this out as a limitation but do not give the user a way to disambiguate.

pi-agent-plugin/src/memory/tools.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      case "add": {
+        if (signal?.aborted) throw new Error("Cancelled");
+        if (!params.content) throw new Error("content is required for add");
+        const addParams = resolveAddParams(scope, scopeCtx);
+        const result = await mem0.add(
+          [{ role: "user", content: params.content }],
+          { ...addParams, customCategories: DEFAULT_CUSTOM_CATEGORIES },
+        );
+        const res = result as any;
```
whysosaket: `as any` to reach into `result.message`, `eventId`, and `status`. Hides regressions if the mem0ai response shape changes.

pi-agent-plugin/src/capture/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  return null;
+}
+
+export function extractConversation(
+  messages: MessageLike[],
+): Array<{ role: "user" | "assistant"; content: string }> {
+  const result: Array<{ role: "user" | "assistant"; content: string }> = [];
+
+  for (const msg of messages) {
+    if (msg.role !== "user") continue;
```
whysosaket: Only user messages are captured. Facts the assistant introduces (config it discovered, decisions it made, error messages it surfaced) never enter memory unless the user repeats them. If intentional, worth one line explaining why.

pi-agent-plugin/src/capture/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  pi.on("agent_end", async (event) => {
+    const messages = event.messages ?? [];
+    const conversation = extractConversation(messages);
+    if (conversation.length === 0) return;
+
+    const scopeCtx = getScopeCtx();
+    const addParams = resolveAddParams("project", scopeCtx);
+
+    mem0
```
whysosaket: Fire and forget. On a one-shot Pi run where `agent_end` is the final event, the process can exit before this POST completes and the capture is silently dropped.

whysosaket (commented):
Suggest placing this under an `integrations/` directory (e.g., `integrations/pi-agent-plugin/`) instead of at the repo root. The root already has several plugin top-level directories (`openclaw/`, `mem0-plugin/`, `vercel-ai-sdk/`) and grouping them keeps the repo browsable. Worth doing for this one and raising the broader cleanup separately.

Inline comments below are observations only, not prescriptive fixes.

kartik-mem0:
We will need to have a separate or for this as openclaw and pi comes under the same category here maybe we can do a follow up pr for this and have a new dir named agent-harness. Or we can combine everything in mem0-plugins but it will be really ugly.

pi-agent-plugin/src/config/index.ts
kartik-mem0: Good catch — switched to `os.homedir()` to match telemetry.ts. Both files now use the same resolution path.

pi-agent-plugin/src/telemetry.ts
kartik-mem0: Fixed — config/index.ts now uses `os.homedir()` same as telemetry.ts. Single source of truth.

pi-agent-plugin/src/commands.ts
kartik-mem0: Valid concern. When there are multiple matches the user already gets a selection prompt. For the single-match case, the user explicitly typed a query that matched exactly one memory — auto-deleting feels reasonable there. The Pi SDK doesn't currently expose a confirmation dialog API, so adding a preview + confirm step would need upstream support. Tracking this as a follow-up once Pi exposes interactive prompts.

pi-agent-plugin/src/commands.ts
kartik-mem0: Acknowledged — this is a limitation of the Mem0 API. There's no metadata update endpoint to add a `[PINNED]` prefix in-place without recreating the memory. The ID change is inherent to the add+delete approach. In practice pin is a rare manual action and the new memory is immediately visible via search, so the impact is low. If Mem0 adds a metadata/update endpoint, we can switch to an in-place mutation.

pi-agent-plugin/src/commands.ts
kartik-mem0: `resolveMemoryId` is still used by `/mem0-search` for short-ID lookups (line 107). The 8-char guard enables users to paste a partial ID prefix and get the full memory. It's narrower than the name suggests, but it does have a live caller. I'll leave it as-is — renaming it would be cosmetic.

pi-agent-plugin/src/commands.ts
kartik-mem0: Fair — there are actually no test files anywhere in the plugin yet (not just this file). Adding a test suite for commands.ts (especially the destructive paths) is the right next step. Will address in a follow-up.

pi-agent-plugin/src/entry.ts
kartik-mem0: Agreed, extracted to a standalone `resolveUserId(configUserId)` function — readable, testable, and handles the USER/USERNAME/os.userInfo() chain cleanly.

pi-agent-plugin/src/entry.ts
kartik-mem0: The `dreamTriggered` flag on line 74 already prevents repeat getAll calls — once the dream is triggered (or skipped because gates fail), the `!dreamTriggered` guard short-circuits all subsequent `before_agent_start` calls in that session. So the API round-trip happens at most once per session. The latency on that single call is ~200ms which is acceptable for a turn-start hook.

pi-agent-plugin/src/memory/scoping.ts
kartik-mem0: Known limitation, documented in the README. Using the full path hash would fix collisions but produce opaque IDs that aren't human-friendly in memory listings. A middle ground would be appending a path-based disambiguator only when a collision is detected — but that adds complexity for a rare edge case. Leaving as-is for now; users who hit this can set a custom `app_id` per project in the config.

pi-agent-plugin/src/memory/tools.ts
kartik-mem0: Fixed — added an `AddResult` interface with `message`, `eventId`, and `status` fields. Now uses `as AddResult` instead of `as any`, so a shape change in mem0ai would produce a compile error.

pi-agent-plugin/src/capture/index.ts
kartik-mem0: Intentional — added a comment explaining why. Assistant content is inferred/derived and would add noise. Mem0's add() already extracts facts from the user's messages; feeding assistant messages back in would create duplicate or circular entries.

pi-agent-plugin/src/capture/index.ts
kartik-mem0: Valid — the telemetry module already handles this with a `beforeExit` handler (telemetry.ts:134). The capture path could do the same, but `beforeExit` is only reliable when the event loop is empty. For a one-shot Pi run, the process may exit via SIGTERM which bypasses `beforeExit`. A more robust fix would be awaiting the capture promise in the `agent_end` handler. Will address alongside the test suite follow-up.

pi-agent-plugin/src/memory/scoping.ts
kartik-mem0: Fixed in 2d2474a — `detectAppId` now returns `parent/basename` (e.g. `projects/my-app` vs `personal/my-app`), so same-named directories under different parents no longer collide. Test coverage added in `scoping.test.ts`.

pi-agent-plugin/src/commands.ts
kartik-mem0: Fixed in 2d2474a — turns out Pi SDK does have `ctx.ui.confirm()`. Both `/mem0-forget` and `/mem0-pin` now show a confirmation dialog before the destructive single-match path. Cancelling is clean — no API calls made. Tests added covering both confirm and cancel flows.

pi-agent-plugin/src/commands.ts
kartik-mem0: Fixed in 2d2474a — renamed to `expandShortId` for clarity, and replaced the overly broad `ID_PATTERN` (`/^[0-9a-f-]{8,}$/i`) with two precise patterns: `UUID_PREFIX_PATTERN` (matches `abcd1234-5...`) and `SHORT_ID_PATTERN` (matches exactly 8 hex chars). The old regex would match any hex-ish string 8+ chars (like "deadbeefs" or "abcdefab") and route it through the ID lookup path instead of semantic search. Also bumped the "already a full ID" threshold from 20 to 36 (UUID length). Tests verify both the ID path and the non-ID fallback.

pi-agent-plugin/src/commands.ts
kartik-mem0: Fixed in 2d2474a — added test suites:
- `commands.test.ts`: 14 tests covering all destructive paths (`/mem0-forget` confirm/cancel, `/mem0-pin` confirm/cancel/already-pinned, `/mem0-search` ID resolution vs semantic, `/mem0-remember`)
- `scoping.test.ts`: `detectAppId` disambiguation, `detectRunId`, filter/param resolution
- `entry.test.ts`: `resolveUserId` fallback chain

All 72 tests pass.

pi-agent-plugin/src/commands.ts
kartik-mem0: Removed entirely in 4d846e3 — the ID pattern, `looksLikeMemoryId`, and `expandShortId` are all gone. `/mem0-search` now always goes through semantic search. No more false positives on hex-looking words.

pi-agent-plugin/src/entry.ts
kartik-mem0: Fixed in 4d846e3 — my previous reply was wrong about the flag preventing repeat calls. The bug: when cheap gates pass but memory gate fails (< 20 memories), `dreamTriggered` stays false so `getAll` is re-called every single turn for the rest of the session. Added a `dreamChecked` flag that's set to true as soon as we enter the getAll path, ensuring the API round-trip happens at most once per session regardless of whether dream actually triggers.

pi-agent-plugin/src/capture/index.ts
kartik-mem0: Fixed in 4d846e3 — `extractConversation` now captures both user and assistant text messages. Assistant tool_use blocks are still filtered out by `extractText` (it only extracts `type: "text"` blocks from arrays), so tool calls and system noise don't enter memory. Tests updated to verify all cases: text messages captured, tool-only assistant messages skipped, system/tool roles skipped.

pi-agent-plugin/src/commands.ts
kartik-mem0: Fixed in 41ff9ac — `/mem0-pin` now uses `mem0.update(target.id, { text: "[PINNED] ..." })` instead of add+delete. The memory ID is preserved, so external references won't break. Test updated to verify `update` is called and `add`/`delete` are not.

pi-agent-plugin/src/capture/index.ts
kartik-mem0: Fixed in 89933c5 — the `agent_end` handler now `await`s `mem0.add()` instead of fire-and-forget `.then()/.catch()`. The event loop keeps the process alive until the POST completes, so one-shot Pi runs no longer silently drop captures.

pi-agent-plugin/src/memory/scoping.ts
kartik-mem0: Updated in 985606b — the parent/basename approach was wrong for monorepos. `detectAppId` now runs `git rev-parse --show-toplevel` and uses the repo root basename. So whether you're in `mem0/`, `mem0/pi-agent-plugin/`, or `mem0/mem0-ts/`, the app_id is always `mem0`. Falls back to `path.basename(cwd)` for non-git directories. Uses `execFileSync` (no shell) per security best practice.

This also means existing memories stored under `app_id: "mem0"` are visible again — the app_id is back to matching the git root name.

pi-agent-plugin/src/telemetry.ts
whysosaket: Functional bug is gone (both now resolve via `os.homedir()`), but `CONFIG_DIR` is still defined twice. Follow-up: import the one from `config/index.ts`.

whysosaket (commented):
Re-review at 5f5b968. Verified all 12 prior inline comments locally (typecheck clean, 77 tests pass, up from 42).

**Addressed**

- `HOME` literal → now `os.homedir()` in `config/index.ts`
- `/mem0-forget` single-match → `ctx.ui.confirm()` before delete; multi-match → `ctx.ui.select()`
- `/mem0-pin` add+delete → `mem0.update(id, { text })`, ID preserved
- Auto-capture fire-and-forget → awaited inside try/catch
- Auto-capture user-only → assistant text captured too via role allowlist
- Inline IIFE for userId → extracted to `resolveUserId()`, exported, tested
- `detectAppId` basename collision → uses `git rev-parse --show-toplevel`, `execFileSync` (no shell), 3s timeout, basename fallback
- Per-turn `getAll` when dream enabled → `dreamChecked` flag, at most one call per session
- `resolveMemoryId` dead-ish helper → removed entirely; `/mem0-search` is always semantic now
- `as any` in `tools.ts` add → typed `AddResult` interface
- `commands.ts` test coverage → 14 new tests in `src/commands.test.ts` covering destructive paths and cancellation
- Multi-match flows for forget/pin now use native `ctx.ui.select()` instead of agent round-trip

**Still worth a follow-up (not blockers)**

- `CONFIG_DIR` is still defined twice, once in `config/index.ts` and once in `telemetry.ts`. Both now resolve identically via `os.homedir()`, so the functional bug is gone, but the duplication remains.
- New `update` case in `tools.ts:101` still uses `(updateResult as any).status`, the last `as any` in that file.
- `dreamChecked = true` is set before the `getAll` try block in `entry.ts:90`, so a transient network failure permanently disables dream gating for that session. The catch comment says "skip dream this turn" but actually skips for the whole session.
- `detectAppId` runs `git rev-parse` at both extension init and every `session_start`. Cheap, but redundant.

**Org-level**

- Integrations folder ask acknowledged for a follow-up PR under `agent-harness/` per @kartik-mem0's comment.

pi-agent-plugin/src/memory/tools.ts
```diff
@@ -0,0 +1,194 @@
...
+        };
+      }
+
+      case "update": {
+        if (signal?.aborted) throw new Error("Cancelled");
+        if (!params.memory_id) throw new Error("memory_id is required for update");
+        if (!params.content) throw new Error("content is required for update");
+        const updateResult = await mem0.update(params.memory_id, { text: params.content });
+        return {
+          content: [{ type: "text" as const, text: (updateResult as any).status ?? "Memory updated." }],
```
whysosaket: Last `as any` in this file. Same shape as `AddResult` (`status` field), could share an interface.

pi-agent-plugin/src/entry.ts
```diff
@@ -0,0 +1,146 @@
...
+  let dreamTriggered = false;
+  let dreamChecked = false;
+
+  pi.on("before_agent_start", async (event, _ctx) => {
+    let extra = MEMORY_POLICY;
+
+    if (config.dream.enabled && config.dream.auto && !dreamTriggered && !dreamChecked) {
+      const gates = checkCheapGates(CONFIG_DIR, config.dream);
+      if (gates.proceed) {
+        dreamChecked = true;
```
whysosaket: `dreamChecked = true` is set before the `getAll` runs, so a single transient error permanently disables dream gating for the rest of the session. The catch comment below says "skip dream this turn" but in practice it skips for the whole session.

pi-agent-plugin/src/memory/scoping.ts
```diff
@@ -0,0 +1,51 @@
+import * as path from "node:path";
+import * as crypto from "node:crypto";
+import { execFileSync } from "node:child_process";
+import type { Scope, ScopeContext } from "../types.ts";
+
+export function detectAppId(cwd: string): string {
+  try {
+    const root = execFileSync("git", ["rev-parse", "--show-toplevel"], {
```
whysosaket, line 8: `git rev-parse` runs at both extension init (entry.ts:41) and again on every `session_start` (entry.ts:64). Cheap but redundant.

pi-agent-plugin/src/telemetry.ts
kartik-mem0: Fixed in 2896d0e8. Removed the duplicate `CONFIG_DIR` from telemetry.ts — now imports from `config/index.ts`. Single source of truth, both resolve to `path.join(os.homedir(), ".pi", "agent")`.

pi-agent-plugin/src/memory/tools.ts
kartik-mem0: Fixed in 2896d0e8. Renamed `AddResult` to `MemoryResult` and reused it for the update case. No more `as any` on the update result — both add and update use `result as MemoryResult`.

pi-agent-plugin/src/entry.ts
kartik-mem0: Fixed in 2896d0e8. Moved `dreamChecked = true` to after `getAll` succeeds (line 94, between the result and `checkMemoryGate`). Now a transient network error in the catch block leaves `dreamChecked = false`, so the next turn retries. Updated the catch comment to "Transient error — retry next turn" to match the actual behavior.

pi-agent-plugin/src/memory/scoping.ts
kartik-mem0: Fixed in 2896d0e8. Removed the `detectAppId(process.cwd())` call at init — `appId` starts as `""` and gets set in `session_start` via `detectAppId(ctx.cwd)`. Per Pi's extension lifecycle docs, the factory runs first, then `session_start` always fires before any `before_agent_start` or tool call, so the init call was spawning a redundant git subprocess for a value that gets overwritten immediately.

whysosaket (commented):
LGTM at 2896d0e8. Verified locally: typecheck clean, 77/77 tests pass, all CI green.

All four follow-up nits addressed:

- `CONFIG_DIR` deduped: `telemetry.ts` now imports from `config/index.ts`
- Last `as any` removed: `AddResult` renamed to `MemoryResult` and reused for the update action
- `dreamChecked = true` moved after `getAll` succeeds, so transient errors retry next turn
- Redundant `detectAppId` at init removed, only `session_start` resolves it now

No remaining blockers from my side. The `agent-harness/` folder reorg can land as the follow-up PR you mentioned.

whysosaket (approved):

PR #5462, opened 2026-06-09
Add quantal vector store provider

PaytonWebber:
## Linked Issue

N/A

## Description

Adds [quantal](https://github.com/PaytonWebber/quantal) as a vector store provider. quantal is an embedded vector index (`pip install quantaldb`): no server to run, graph-routed search over quantized codes, so search stays sub-linear as memories accumulate and the index stores compressed data instead of full-precision vectors.

Provider notes:

- Similarity is always cosine (vectors are stored normalized and scored by inner product).
- Deletes are O(1) tombstones; removing a memory never rebuilds the index.
- Filtered searches (e.g. `user_id`) restrict scoring to the matching ids up front rather than over-fetching and post-filtering.
- Persistence is the index's native `.tq` format plus a JSON sidecar for payloads and id mapping, following the FAISS provider's layout.
- `keyword_search` is not implemented; hybrid retrieval degrades to dense-only per the `VectorStoreBase` contract.

Includes the config class, factory registration, docs page (registered in `docs.json` and the vectordbs overview), and unit tests.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

`tests/vector_stores/test_quantal.py` covers insert/search, filtered search, delete, update (vector and payload), get, list, collection info, persistence roundtrip (including id-collision safety after reload), and reset; 12 tests, all passing locally against quantaldb 0.1.2. The tests `importorskip` quantal so CI without the package skips them. Also manually verified end-to-end creation through `VectorStoreFactory` and `VectorStoreConfig`.

Disclosure: I am the author of quantal.

### DIFF
--- a/docs/components/vectordbs/dbs/quantal.mdx
+++ b/docs/components/vectordbs/dbs/quantal.mdx
@@ -1,0 +1,55 @@
+---
+title: "Quantal"
+description: "Use quantal as an embedded, quantized vector store in Mem0: no server to run, with graph-routed search and a fraction of the memory of full-precision stores."
+---
+
+[quantal](https://github.com/PaytonWebber/quantal) is an embedded vector index you link into your application, not a server you run. It routes each query through a graph over quantized codes instead of scanning every vector, so search stays fast as memories accumulate, and the index stores compressed data rather than full-precision vectors. Deletes are O(1) tombstones, so removing memories never rebuilds the index.
+
+### Installation
+
+```bash
+pip install quantaldb
+```
+
+Prebuilt wheels cover the common embedding dimensions (256/384/512/768/1024/1536/3072) on Linux, macOS (Apple Silicon), and Windows.
+
+### Usage
+
+```python
+import os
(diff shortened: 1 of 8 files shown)


PaytonWebber:
Also verified end to end against the real `Memory` pipeline (only the embedder and LLM factories mocked, with deterministic embeddings and `infer=False`): `add` -> `search` with `filters={'user_id': ...}` (exercises the allowlist-filtered path), cross-user isolation, `update` with re-embedding, `delete`, `get_all`, and SQLite history. All checks pass against quantaldb 0.1.2.

kartik-mem0:
Hi @PaytonWebber — thanks for building this, and for the clean implementation and docs. quantal looks like a genuinely interesting project (nice work on the quantized graph-routing index). That said, it's very new, and our guideline is to bring third-party backends into core once they have some established adoption, so the long-term maintenance commitment is justified. We'd love to revisit once it has more traction and users asking for it. Closing for now — really appreciate the contribution.

PR #5477, opened 2026-06-10
fix(cli-python): import Exit from typer instead of click in tests

whysosaket:
## Linked Issue

N/A — fixes the `Python CLI / test` failure surfaced by the CI Gate run on #5476 (https://github.com/mem0ai/mem0/actions/runs/27293136169).

## Description

`cli/python` tests fail at collection with `ModuleNotFoundError: No module named 'click'`. Root cause: **typer ≥ 0.26 vendored click** (its dependencies are now `annotated-doc`, `rich`, `shellingham` — no `click`), so a fresh `pip install -e ".[dev]"` no longer installs the standalone `click` package. The test file imported `from click.exceptions import Exit as ClickExit`, which is broken on two counts under new typer: the module is missing, and even with `click` installed it would be a *different class* than the vendored `typer.Exit` the CLI actually raises — so `pytest.raises((SystemExit, ClickExit))` would silently stop catching the right exception.

Fix: import `Exit` from `typer` (`from typer import Exit as TyperExit`), which is correct on both old typer (where `typer.Exit` *is* click's `Exit`) and new typer (vendored). The CLI source already uses `typer.Exit` everywhere — only the test file needed the change.

This is a pre-existing breakage on `main` (resolver drift), not caused by the CI Gate PR — the gate surfaced it because the Python CLI pipeline hadn't run recently.

## Type of Change

- [ ] Breaking change
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Reproduced the CI failure locally with a fresh venv (`pip install -e ".[dev]"` resolves typer 0.26.7, `import click` fails). With the fix: all **161 tests pass** and `ruff check .` is clean.

## Checklist

### DIFF
--- a/cli/python/tests/test_commands.py
+++ b/cli/python/tests/test_commands.py
@@ -8,8 +8,8 @@ from io import StringIO
 from unittest.mock import patch
 
 import pytest
-from click.exceptions import Exit as ClickExit
 from rich.console import Console
+from typer import Exit as TyperExit
 
 from mem0_cli.commands.config_cmd import (
     cmd_config_get,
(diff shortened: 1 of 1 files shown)


whysosaket:
Superseded — the fix was cherry-picked into #5476 (commit e622cbec) so its CI Gate run can validate green directly. Closing.

Issue #5479, opened 2026-06-10
TS SDK: memory update drops custom metadata (parity with #5160, fixed for Python in #5179)


Ghraven:
## Description

Follow-up to #5160 / #5179, opened at the maintainer's request in the #5179 review.

The TypeScript OSS SDK has the same whitelist-only pattern that the Python SDK had: the update path in `mem0-ts/src/oss/src/memory/index.ts` rebuilds the payload from scratch and hand-copies only a fixed set of session-identifier fields from the existing memory. Any other custom metadata stored at `add()` time (`category`, `priority`, `source`, etc.) is silently dropped when the vector store replaces the full payload on update.

## Expected behavior

Same as the Python fix in #5179: start the new payload from a copy of the **full existing payload**, overlay caller-supplied metadata on top, then stamp the system fields (`data`, `hash`, `createdAt`/`updatedAt`) last — so a text-only update never loses custom fields, while intentional metadata updates still apply.

## Reference

- Python fix: #5179
- Original report: #5160

Happy to take this one as well once #5179 lands, so both SDKs stay in sync.

kartik-mem0:
## Triage Notes

**What we've established:**
- This is a TypeScript OSS SDK counterpart to the Python bug in #5160 — confirmed by tracing `updateMemory()` in `mem0-ts/src/oss/src/memory/index.ts` (around L1502–1517): `newMetadata` is built from scratch with only caller-supplied metadata spread first, then the three session fields (`user_id`, `agent_id`, `run_id`) are conditionally copied. Any other custom metadata stored at `add()` time is silently dropped.
- PR #5065 is open against `mem0-ts/` but addresses a different issue (camelCase key conversion in `snakeToCamelKeys`), not the payload-drop on update. So #5479 is **not** addressed by #5065.
- The fix pattern here is the same as #5179 proposes for Python: start `newMetadata` from a spread of `existingMemory.payload`, overlay caller-supplied metadata on top, then stamp managed fields last.

**Reproduction:** Confirmed (code tracing — no live backend needed; the whitelist pattern is unambiguous)
[...]

PR #5179, opened 2026-05-17
fix(memory): preserve all custom metadata in _update_memory (sync + async)

Ghraven:
## Summary

Closes #5160.

`_update_memory()` (both `Memory` and `AsyncMemory`) started `new_metadata` from an empty dict, then hand-copied only a small set of session identifier fields from `existing_memory.payload`. Every other custom metadata field stored at `add()` time — for example `category`, `priority`, `source`, `bucket_id`, or `app_id` — was silently dropped when `vector_store.update()` replaced the full payload.

This was already fixed for MongoDB in #4495, but the same whitelist-only pattern persisted in the other vector-store update paths. The V3 pipeline rewrite in #4805 kept the same pattern, so the bug regressed.

## Root cause / before

```python
new_metadata = deepcopy(metadata) if metadata is not None else {}
# then only a hard-coded subset of fields was copied from existing_memory.payload
```

This meant a text-only update, or an update with partial caller metadata, could discard unrelated custom metadata already attached to the memory.

## Fix / after

```python
new_metadata = deepcopy(existing_memory.payload)
if metadata is not None:
    new_metadata.update(metadata)
if "actor_id" in existing_memory.payload:
    new_metadata["actor_id"] = existing_memory.payload["actor_id"]
```

The update payload now starts from the full existing payload, overlays caller-supplied metadata, preserves the existing `actor_id` behavior required by #4490, then stamps system fields like `data`, `hash`, `text_lemmatized`, `created_at`, and `updated_at` last.

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| `update(id, text)` with no metadata arg | Custom metadata lost | Custom metadata preserved |
| `update(id, text, metadata={"k": "v"})` | Existing custom metadata lost except hard-coded fields | Existing custom metadata preserved, caller metadata overlaid |
| Existing `actor_id` with caller-supplied different `actor_id` | Existing `actor_id` preserved | Existing `actor_id` preserved |

## Verification

- `python -m pytest tests/test_main.py tests/memory/test_main.py -k "update_memory_preserves_custom_metadata or caller_metadata_overlays_existing or preserves_actor_id_when_different_actor_updates"` — 5 passed
- `python -m ruff check mem0/memory/main.py tests/test_main.py tests/memory/test_main.py`
- `python -m isort --check-only mem0/memory/main.py tests/test_main.py tests/memory/test_main.py`
- `git diff --check`

Also merged current `upstream/main` into this branch as requested.

Issue #4490, opened 2026-03-23
Bug Report: `actor_id` metadata gets overwritten during memory UPDATE operations


VictorECDSA:
### 🐛 Describe the bug

# Bug Report: `actor_id` metadata gets overwritten during memory UPDATE operations

## Description

I believe I may have found a bug related to `actor_id` handling in multi-actor scenarios. When using `metadata={"actor_id": ...}` with shared `user_id`, the `actor_id` field appears to be overwritten when a different actor triggers an UPDATE event, which seems to break actor-level memory isolation and query filtering.

I'm not sure if this is the intended behavior, but I wanted to report it in case it's helpful.

## Steps to Reproduce

```python
from mem0 import Memory

m = Memory.from_config(config)

# Step 1: Actor A creates memory
result_a = m.add(
    [{"role": "user", "content": "I am player #1"}],
    user_id="team",
    metadata={"actor_id": "Alice"}
)
# ✓ Memory created with actor_id="Alice"

# Step 2: Actor B updates A's memory  
result_b = m.add(
    [{"role": "user", "content": "Player #1 is a good person"}],
    user_id="team",
    metadata={"actor_id": "Bob"}
)
# ❌ Memory updated but actor_id becomes "Bob"

# Step 3: Query by original creator
alice_memories = m.search(query="", user_id="team", filters={"actor_id": "Alice"})
print(f"Query Alice's memories: {len(alice_memories['results'])} found")
# Returns: 0 found
```

## Expected Behavior

After Bob's UPDATE:
- Memory content should be updated: `"Player #1 is a good person"`
- `actor_id` should remain: `"Alice"` (original creator)
- Query `filters={"actor_id": "Alice"}` should return 1 result

## Actual Behavior

After Bob's UPDATE:
- Memory content is updated correctly ✓
- `actor_id` is overwritten to: `"Bob"` ❌
- Query `filters={"actor_id": "Alice"}` returns empty ❌

**Output**:
```
=== State AFTER Bob's UPDATE ===
  Event: UPDATE
  This actor_id: 'Bob'  ← Overwritten!

ISOLATION TEST:
  Query Alice's memories: 0 found
  ❌ BUG: Alice's memory lost!
```

## Impact

1. **Cannot query by original creator** after UPDATE
2. **Memory ownership tracking lost** - can't identify who created the memory
3. **Actor isolation fails** in shared `user_id` scenarios (e.g., team workspaces, multi-player games, collaborative agents)

## Root Cause

[...]

PR #5480, opened 2026-06-11
fix: preserve custom metadata fields during memory update

### LINKED ISSUE #5160: _update_memory() still drops custom metadata on all non-MongoDB vector stores (follow-up from #4495 never completed, regressed in V3 pipeline)
### Component

Core / Python SDK

### Summary

Memory.update() and AsyncMemory.update() silently drop all custom metadata fields (e.g., category, priority, source, bucket_id) that were originally stored during add(). Only a hard-coded whitelist of 5 session identifiers (user_id, agent_id, run_id, actor_id, role) is preserved — everything else is lost.

This is the unfulfilled cross-provider follow-up from #4495 (which fixed this only for MongoDB). Reviewer @utkarsh240799 [flagged] that the same bug exists on Qdrant, Pinecone, Milvus, PGVector, Supabase, and Redis, and @kartik-mem0 [acknowledged] a follow-up was needed — but it was never completed. PR #4805 (V3 pipeline) then rewrote _update_memory(), keeping the whitelist-only pattern.

### Steps to Reproduce

from mem0 import Memory

m = Memory()

# Step 1: Add memory with custom metadata
result = m.add(
    "I love playing tennis on weekends",
    user_id="user_1",
    metadata={"category": "hobbies", "priority": "high", "source": "chat"}
)

# Step 2: Verify custom metadata exists
memory = m.get(memory_id)
print(memory.get("metadata"))
# Output: {"category": "hobbies", "priority": "high", "source": "chat"}

# Step 3: Update only the text (no metadata argument)
m.update(memory_id, data="I love playing tennis and swimming on weekends")

# Step 4: Check metadata again
memory = m.get(memory_id)
print(memory.get("metadata"))
# Output: None  <-- all custom metadata is gone!

### Expected Behavior

When updating a memory's text content without providing new metadata, all existing custom metadata should be preserved. After Step 4, the output should still be:
{"category": "hobbies", "priority": "high", "source": "chat"}

### Actual Behavior
All custom metadata fields are silently dropped. No error or warning is raised. After Step 4, memory.get("metadata") returns None.

Root cause: In Memory._update_memory() ([L1671–L1689]) and AsyncMemory._update_memory() ([L3102–L3121]), new_metadata starts as an empty dict and only copies 5 hard-coded fields from the existing payload. All other custom metadata is discarded when vector_store.update() replaces the entire payload.

### Environment

mem0 version: Latest (main branch, post-V3 pipeline merge via #4805)
Python version: 3.10+
OS: All (Windows, Linux, macOS)
Affected backends: All non-MongoDB vector stores (Qdrant, Pinecone, Milvus, PGVector, Supabase, Redis, FAISS, Chroma, Weaviate, etc.)
Affected classes: Memory (sync) and AsyncMemory (async)

Additional Context:
#3985 
Title- fix: preserve all metadata fields when updating memory 
Status : Closed (superseded) 
Scope : Proposed correct fix in _update_memory()
#4495 
Title - fix: preserve custom metadata when updating memory
Status :Merged 
Scope :MongoDB vector store only
#4570 
Title - fix: preserve original actor_id during memory update
Status :Closed
Scope : actor_id field only
#4805 
Title - feat(oss): port v3 pipeline with hybrid search
Status : Merged
Scope : Rewrote _update_memory(), re-introduced bug
Proposed fix :
# In _update_memory(), replace the empty-dict initialization with:

# Overlay user-provided metadata
if metadata is not None:
    new_metadata.update(metadata)

# Then update managed fields as before
new_metadata["data"] = data
new_metadata["hash"] = hashlib.md5(data.encode()).hexdigest()
new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
new_metadata["created_at"] = existing_memory.payload.get("created_at")
new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()

cc @kartik-mem0 @utkarsh240799 — this is the cross-provider follow-up discussed in #4495

HrushiYadav:
Closes #5160

### Problem

`_update_memory()` builds `new_metadata` from scratch using only caller-provided metadata (or an empty dict), then cherry-picks 5 hardcoded session identifiers from the existing payload. Any custom metadata stored during `add()` (e.g. `category`, `priority`, `source`) is silently dropped when the payload is replaced.

### Fix

Start from `deepcopy(existing_memory.payload)` as the base, overlay caller-provided metadata on top, then set managed fields (`data`, `hash`, `text_lemmatized`, timestamps). This preserves all existing fields by default while still allowing callers to override any field via the `metadata` parameter.

The old 12-line whitelist block for session identifiers becomes unnecessary since they're already in the base payload.

Both sync (`Memory._update_memory`) and async (`AsyncMemory._update_memory`) paths are fixed.

### Tests

Added 3 tests in `TestPreserveCustomMetadata`:
- Custom metadata survives text-only update (sync)
- Caller-provided metadata overrides existing fields (sync)
- Custom metadata survives text-only update (async)

All 46 tests pass.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1726,6 +1726,7 @@ export class Memory {
       ...metadata,
       data,
       hash: createHash("md5").update(data).digest("hex"),
+      textLemmatized: lemmatizeForBm25(data),
       createdAt: existingMemory.payload.createdAt,
       updatedAt: new Date().toISOString(),
     };
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1722,20 +1722,12 @@ export class Memory {
       existingEmbeddings[data] || (await this.embedder.embed(data));
 
     const newMetadata = {
+      ...existingMemory.payload,
       ...metadata,
       data,
       hash: createHash("md5").update(data).digest("hex"),
       createdAt: existingMemory.payload.createdAt,
       updatedAt: new Date().toISOString(),
-      ...(existingMemory.payload.user_id && {
-        user_id: existingMemory.payload.user_id,
-      }),
-      ...(existingMemory.payload.agent_id && {
-        agent_id: existingMemory.payload.agent_id,
-      }),
-      ...(existingMemory.payload.run_id && {
-        run_id: existingMemory.payload.run_id,
-      }),
     };
 
     await this.vectorStore.update(memoryId, embedding, newMetadata);
--- a/mem0-ts/src/oss/tests/memory.crud.test.ts
+++ b/mem0-ts/src/oss/tests/memory.crud.test.ts
@@ -191,6 +191,21 @@ describe("Memory - update()", () => {
     const after: MemoryItem | null = await memory.get(id);
     expect(after!.hash).not.toBe(before!.hash);
   });
+
+  test("preserves custom metadata fields after update", async () => {
+    const addResult: SearchResult = await memory.add("Original text", {
+      userId,
+      metadata: { category: "hobbies", priority: "high" },
+      infer: false,
+    });
+    const id = addResult.results[0].id;
+    await memory.update(id, "Updated text");
+    const after: MemoryItem | null = await memory.get(id);
+    expect(after!.memory).toBe("Updated text");
+    expect(after!.metadata).toEqual(
+      expect.objectContaining({ category: "hobbies", priority: "high" }),
+    );
+  });
 });
 
 // ─── delete() ────────────────────────────────────────────
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1839,25 +1839,15 @@ class Memory(MemoryBase):
 
         prev_value = existing_memory.payload.get("data")
 
-        new_metadata = deepcopy(metadata) if metadata is not None else {}
+        new_metadata = deepcopy(existing_memory.payload)
+        if metadata is not None:
+            new_metadata.update(metadata)
 
         new_metadata["data"] = data
         new_metadata["hash"] = hashlib.md5(data.encode()).hexdigest()
         new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
         new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
-
-        # Preserve session identifiers from existing memory only if not provided in new metadata
-        if "user_id" not in new_metadata and "user_id" in existing_memory.payload:
-            new_metadata["user_id"] = existing_memory.payload["user_id"]
-        if "agent_id" not in new_metadata and "agent_id" in existing_memory.payload:
-            new_metadata["agent_id"] = existing_memory.payload["agent_id"]
-        if "run_id" not in new_metadata and "run_id" in existing_memory.payload:
-            new_metadata["run_id"] = existing_memory.payload["run_id"]
-        if "actor_id" in existing_memory.payload:
-            new_metadata["actor_id"] = existing_memory.payload["actor_id"]
-        if "role" not in new_metadata and "role" in existing_memory.payload:
-            new_metadata["role"] = existing_memory.payload["role"]
 
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]
@@ -3371,26 +3361,15 @@ class AsyncMemory(MemoryBase):
 
         prev_value = existing_memory.payload.get("data")
 
-        new_metadata = deepcopy(metadata) if metadata is not None else {}
+        new_metadata = deepcopy(existing_memory.payload)
+        if metadata is not None:
+            new_metadata.update(metadata)
 
         new_metadata["data"] = data
         new_metadata["hash"] = hashlib.md5(data.encode()).hexdigest()
         new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
         new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
-
-        # Preserve session identifiers from existing memory only if not provided in new metadata
-        if "user_id" not in new_metadata and "user_id" in existing_memory.payload:
-            new_metadata["user_id"] = existing_memory.payload["user_id"]
-        if "agent_id" not in new_metadata and "agent_id" in existing_memory.payload:
-            new_metadata["agent_id"] = existing_memory.payload["agent_id"]
-        if "run_id" not in new_metadata and "run_id" in existing_memory.payload:
-            new_metadata["run_id"] = existing_memory.payload["run_id"]
-
-        if "actor_id" in existing_memory.payload:
-            new_metadata["actor_id"] = existing_memory.payload["actor_id"]
-        if "role" not in new_metadata and "role" in existing_memory.payload:
-            new_metadata["role"] = existing_memory.payload["role"]
 
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1848,9 +1848,6 @@ class Memory(MemoryBase):
         new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
         new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
-        if "actor_id" in existing_memory.payload:
-            new_metadata["actor_id"] = existing_memory.payload["actor_id"]
-
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]
         else:
@@ -3372,9 +3369,6 @@ class AsyncMemory(MemoryBase):
         new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
         new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
-        if "actor_id" in existing_memory.payload:
-            new_metadata["actor_id"] = existing_memory.payload["actor_id"]
-
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]
         else:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1849,10 +1849,9 @@ class Memory(MemoryBase):
         new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
 
+        # actor_id is immutable after creation (issue #4490)
         if "actor_id" in existing_memory.payload:
             new_metadata["actor_id"] = existing_memory.payload["actor_id"]
-        if "role" in existing_memory.payload and "role" not in (metadata or {}):
-            new_metadata["role"] = existing_memory.payload["role"]
 
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]
@@ -3376,10 +3375,9 @@ class AsyncMemory(MemoryBase):
         new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
 
+        # actor_id is immutable after creation (issue #4490)
         if "actor_id" in existing_memory.payload:
             new_metadata["actor_id"] = existing_memory.payload["actor_id"]
-        if "role" in existing_memory.payload and "role" not in (metadata or {}):
-            new_metadata["role"] = existing_memory.payload["role"]
 
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]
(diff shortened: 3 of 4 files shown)


kartik-mem0:
## Changes Requested

### Strengths

- The core fix is correct: `deepcopy(existing_memory.payload)` as the base, caller-supplied `metadata` overlaid on top, system fields (`data`, `hash`, `text_lemmatized`, `created_at`, `updated_at`) stamped last. Clean, minimal, and the right primitive for this problem.
- Both `Memory._update_memory` (sync, L1725) and `AsyncMemory._update_memory` (async, L3184) are patched identically — good async/sync parity.
- Tests cover all three meaningful axes: preserve-with-no-metadata, caller-override, async mirror. Placed in the correct file (`tests/test_memory.py`). Ruff lint: all clear.
- Conventional commit title well-formed; no new dependencies; no public API surface change.

### Issues

#### Critical (must fix)

**`mem0/memory/main.py` L1740 (sync) and L3211 (async) — `actor_id` unconditionally overwrites caller-supplied value:**

```python
if metadata is not None:
    new_metadata.update(metadata)   # caller's actor_id="Bob" written here
# ... system fields ...
    new_metadata["actor_id"] = existing_memory.payload["actor_id"]  # silently reverts to "Alice"
```

This block runs after `new_metadata.update(metadata)`, so any `actor_id` a caller explicitly passes gets silently overwritten by the original value. Every other field (`user_id`, `agent_id`, `run_id`, `role`, all custom keys) respects the caller's intent. `actor_id` is the only exception, and it's undocumented.

The downstream consequence: `db.add_history` at L1764 reads `new_metadata.get("actor_id")` for the audit trail — so if a different actor performs the update and passes their `actor_id`, the history entry records the wrong actor.

`test_update_preserves_actor_id_from_original` validates this wrong behavior (passes `metadata={"actor_id": "Bob"}`, asserts `"Alice"` wins). The fix: since `deepcopy(existing_memory.payload)` already carries `actor_id` through, the guard block is now fully redundant. The simplest correct fix is to remove it entirely. If `actor_id` is intentionally pinned and non-overridable, that invariant needs a comment and a test that explicitly passes `metadata={"actor_id": "Bob"}` and explains why the original wins — but that would be inconsistent with how every other field works.

#### Important (should fix)

**TypeScript SDK has the identical bug and is not addressed:**

`mem0-ts/src/oss/src/memory/index.ts:1502–1517` builds `newMetadata` as:

```typescript
const newMetadata = {
  ...metadata,                                         // only caller-supplied
  data, hash, createdAt, updatedAt,
  ...(existingMemory.payload.user_id && { user_id: ... }),
  ...(existingMemory.payload.agent_id && { agent_id: ... }),
  ...(existingMemory.payload.run_id && { run_id: ... }),
};
```

No spread of `existingMemory.payload` — custom metadata (`category`, `priority`, `source`, etc.) is silently dropped on every update, the same bug this PR fixes in Python. Please either add the TypeScript fix here or open a follow-up issue and reference it in the PR description so it can be tracked. The TypeScript fix is a direct translation: `{ ...existingMemory.payload, ...metadata, data, hash, createdAt, updatedAt }`.

#### Minor (optional)

- The four new tests share identical 7-line mock setup blocks; a shared `@pytest.fixture` would reduce repetition.
- `created_at` at L1730 is re-assigned from `existing_memory.payload.get("created_at")` — the `deepcopy` base already carries it, so the line is harmless but redundant. A brief comment on intent would help future readers.

### Duplicate cluster note

PR #5179 by @Ghraven also Closes #5160 with the identical algorithm. Key difference: #5179 removes the `actor_id` guard block entirely (no inconsistency), and its tests use a shared assertion helper that also validates session identifiers and timestamps. If #5179's author signs the CLA and that PR moves forward, it should be preferred. If it stalls, this PR is the right path — just needs the Critical fix above.

### Evidence

- **CI:** Vercel deploy-auth fail is infrastructure (missing authorization token, not code). All 6 real checks pass: Python SDK build on 3.10/3.11/3.12, changelog, change detection, CLA.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/test_memory.py::TestPreserveCustomMetadata -v` → 4 passed in 4.02s
- **Red-green proof:** `test_update_preserves_custom_metadata` fails with main's implementation (`KeyError: 'category'`), passes with PR's fix. Revert → FAIL ✓, restore → PASS ✓.
- **Lint:** `ruff check mem0/memory/main.py tests/test_memory.py` → all checks passed
- **Cross-package:** TypeScript `updateMemory` confirmed to have the same structural bug (see Important finding above).
- **Could NOT verify:** live vector store behavior (Qdrant, Pinecone, etc.) — verified via code tracing and mocked unit tests only.

### Assessment

Ready to merge: **With fixes** — please address the Critical (`actor_id` guard) and Important (TypeScript parity) items; I'll re-review once you push.

HrushiYadav:
Thanks for the thorough review! Pushed fixes:

1. **Critical (actor_id guard):** Removed the redundant `actor_id` guard block entirely — `deepcopy(existing_memory.payload)` already carries it through, and the guard was the only field that silently prevented caller overrides. Updated the test to verify caller-supplied `actor_id` now wins, consistent with all other fields.

2. **Important (TypeScript parity):** Noted — will open a follow-up issue for the TS SDK fix.

kartik-mem0:
Verified against af76b3bc.

The actor_id guard removal is done correctly — the unconditional overwrite block is gone from both sync and async paths, and the new `test_update_allows_actor_id_override` test confirms caller-supplied `actor_id` wins. That Critical finding is fixed.

One thing still blocking merge: there are two pre-existing tests on `main` that encode the **old** behavior and now fail CI:

```
tests/memory/test_main.py::test_update_preserves_actor_id_when_different_actor_updates
tests/memory/test_main.py::test_async_update_preserves_actor_id_when_different_actor_updates
```

Both assert `stored["actor_id"] == "Alice"` when Bob passes `actor_id="Bob"` — the opposite of what your fix does (correctly). CI is showing 2 failed, 1228 passed across all three Python versions. These tests were added for issue #4490 and now contradict the new semantics. You need to update them (or remove them) to assert that the caller-supplied value wins, matching your new `test_update_allows_actor_id_override`.

The TypeScript `updateMemory` path (`mem0-ts/src/oss/src/memory/index.ts`) still has the original bug — `existingMemory.payload` custom fields are not spread into `newMetadata`. Happy to track that as a follow-on issue if you want to keep this PR scoped to Python, but please confirm either way.

kartik-mem0:
hey @HrushiYadav thank you for your contribution can you also verify if this issue exist in the ts sdk as well? or not? 

if yes please also do a pr fix for that in this pr only and do a merge main so that ci passes.

thanks!

HrushiYadav:
Thanks @kartik-mem0! Pushed the fixes:

1. **TS SDK parity**: Applied the same `...existingMemory.payload` spread fix in `mem0-ts/src/oss/src/memory/index.ts` - removed the manual whitelist and spread the full existing payload as the base.

2. **actor_id comment**: Added `# actor_id is immutable after creation (issue #4490)` comment on both sync and async paths.

3. **Dead code cleanup**: Removed the redundant `role` guard block (deepcopy already carries it through).

4. **Rebased on main** so CI changelog_check should pass.

kartik-mem0:
## Changes Requested

Re-verified against `ca6ee48` — thanks for adding the TS parity fix, the `actor_id` comment, and removing the dead role guard. The Python half is in great shape; one real bug slipped into the TS change, and I owe you a correction on my earlier note about this same commit.

### Strengths
- Python fix is solid and complete: `deepcopy(existing_memory.payload)` as the base, caller `metadata` overlaid, then managed fields pinned — with `actor_id` immutability now documented and the redundant `role`/session-id blocks correctly removed. Sync and async are symmetric.
- The recompute is done right on the Python side: `mem0/memory/main.py:1848` (sync) and `:3374` (async) set `new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)` after the copy.
- Regression tests are real and they bite (red-green below), and CI is green across the matrix.

### Issues

#### Critical (must fix)
1. `mem0-ts/src/oss/src/memory/index.ts:1724` — **the TS `updateMemory` now persists a stale `textLemmatized`, which breaks keyword/BM25 search for any updated memory.** The new `...existingMemory.payload` spread is correct for preserving custom fields, but it also carries the *previous* memory's `textLemmatized` into the new payload, and `updateMemory` never recomputes it for the new `data`:
   ```ts
   const newMetadata = {
     ...existingMemory.payload,   // carries the OLD textLemmatized
     ...metadata,
     data,
     hash: createHash("md5").update(data).digest("hex"),
   };
   ```
   `textLemmatized` is read by keyword search — `vector_stores/pgvector.ts:344` (`to_tsvector('simple', payload->>'textLemmatized')`) and the in-memory store `vector_stores/memory.ts:281` (`payload.textLemmatized || payload.data`). So after an `update()`, the memory keeps matching its **old** text and stops matching its **new** text. This is actually worse than before this PR, where `updateMemory` didn't write `textLemmatized` at all (so the in-memory store fell back to the fresh `data`). The fix mirrors the Python line — one addition, the import is already at `index.ts:62`:
   ```ts
   const newMetadata = {
     ...existingMemory.payload,
     ...metadata,
     data,
     hash: createHash("md5").update(data).digest("hex"),
     textLemmatized: lemmatizeForBm25(data),   // ← add this
   };
   ```

#### Important (should fix)
1. The TS change ships with no test. `mem0-ts/src/oss/tests/memory.crud.test.ts` exercises `update()` but nothing asserts custom-field survival or `textLemmatized` freshness. A short test in the existing `update()` block — store with `metadata: { category: "hobbies" }`, `update()` with new text, then assert `category` survives and that a keyword search for the new text finds it (and the old text doesn't) — would have caught the above and would lock the parity with the four new Python tests.

#### Correction to my earlier note
- My previous comment on this commit flagged an `actor_id` parity break in the TS SDK and suggested adding an `actor_id` immutability guard. That doesn't actually apply — `actor_id` isn't a field anywhere in the TS OSS SDK (`grep -i actor_id mem0-ts/src/oss/` returns nothing), so such a guard would be dead code. Please disregard it; the real TS parity gap is the `textLemmatized` recompute above.

### Evidence
- **CI (ca6ee48):** green — Python 3.10/3.11/3.12 build, TS SDK build (Node 20/22) + integration, CI Gate, changelog, CLA all pass. (Vercel's failure is the usual deploy-auth one, unrelated to this PR.)
- **Targeted tests:** `pytest tests/test_memory.py::TestPreserveCustomMetadata` → 4 passed; `pytest tests/memory/test_main.py -k update` → 13 passed (confirms dropping the per-field blocks didn't regress `user_id`/`agent_id`/`run_id`/`role` preservation).
- **Red-green (Python):** with the fix → pass; revert `mem0/memory/main.py` to base → `KeyError: 'category'` (fails); restore → pass. The regression test genuinely bites. `ruff` clean.
- **Could NOT verify by execution:** the TS `textLemmatized` regression — confirmed by code tracing only (read of `updateMemory` plus the pgvector and in-memory keyword-search consumers). The Important test item above would make it executable.

### Assessment
Ready to merge: **With fixes** — the Python half is merge-ready; the TS half needs the one-line `textLemmatized` recompute plus a TS test. Both are small, and I'll re-verify as soon as you push.

HrushiYadav:
Thanks @kartik-mem0! Pushed fixes:

1. **textLemmatized recompute (Critical)**: Added `textLemmatized: lemmatizeForBm25(data)` to the TS `updateMemory` metadata spread, so keyword/BM25 search matches the updated text instead of the stale original.

2. **TS test**: Added `preserves custom metadata fields after update` test in `memory.crud.test.ts` -- stores with `category`/`priority` metadata, updates the text, then asserts both custom fields survive.

3. **actor_id TS guard**: Confirmed not needed -- `actor_id` doesn't exist in the TS OSS SDK, so no guard added.

kartik-mem0:
## LGTM

Re-verified against `4aa6a878` — both outstanding items from the last round are resolved. Thanks for the quick turnaround on the TS side, @HrushiYadav.

**Both SDKs are now at parity:**

- **Python** `_update_memory` (sync + async): `deepcopy(existing_memory.payload)` as the base, caller `metadata` overlaid on top, then managed fields stamped — including `text_lemmatized = lemmatize_for_bm25(data)` recomputed and `actor_id` held immutable. Custom keys (`category`, `priority`, `source`, …) now survive a text-only `update()`.
- **TS** `updateMemory`: `...existingMemory.payload` is spread first, and `textLemmatized: lemmatizeForBm25(data)` is now recomputed (`index.ts:1729`) so BM25/keyword search matches the updated text instead of the stale lemmatization. That was the one real regression from the prior round and it's correctly fixed.
- TS test added covering custom-metadata survival across `update()`.

**Evidence (captured this run, head `4aa6a878`):**

- Targeted Python suite `TestPreserveCustomMetadata` → 4/4 passed.
- Red-green: the regression test passes with the fix; reverting `mem0/memory/main.py` to `main` → fails with `KeyError: 'category'` (metadata dropped); restored → passes. The test bites.
- `ruff check mem0/memory/main.py tests/test_memory.py` → clean.
- CI green on this head: Python 3.10/3.11/3.12 build, TS SDK build (20/22), TS integration (20/22), CLA signed. (The Vercel "fail" is the usual deploy-preview auth check, not a code failure.)
- TS `memory.crud.test.ts` verified via the green TypeScript SDK CI on this head plus a direct code read; not re-run locally.

**Note for whoever merges:** this also fixes the TS-side gap tracked in #5479, but the body only says `Closes #5160` — worth linking/closing #5479 at merge time too.

Assessment: **Ready to merge** (squash). ✅

kartik-mem0 (approved):

PR #5481, opened 2026-06-11
fix: delete_all now drains all pages, not just the first batch

HrushiYadav:
Closes #4869

### Problem

`delete_all()` calls `vector_store.list(filters=filters)` once, which returns at most `top_k` records (default 100). If a user has more memories than the page size, the extra ones survive the deletion.

### Fix

Wrap the list+delete in a `while` loop that keeps calling `list()` until an empty batch comes back. Since each iteration deletes the returned records, the next call naturally returns the next batch.

Both sync and async paths are fixed.

### Tests

Added 2 tests in `TestDeleteAllPagination`:
- Sync: mock returns 3 pages (100 + 50 + 0), asserts all 150 deleted
- Async: same scenario

All 45 tests pass.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1725,12 +1725,17 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for this, @HrushiYadav! We're consolidating the `delete_all()` pagination fix on #4872 (Python) and its TypeScript counterpart #4874 — that pair also normalizes the flat-list vector-store return shape and is green on CI. Closing this one to keep the fix in a single place; please feel free to weigh in there. Appreciate the effort! 🙏

PR #5482, opened 2026-06-11
fix: return 400 instead of 502 for invalid search filters

### LINKED ISSUE #5367: POST /search returns 502 instead of 400 for invalid filters (observed in MCP integration)

HrushiYadav:
Closes #5367

### Problem

`POST /search` with empty or missing filter identifiers raises `ValueError` from `Memory.search()`. The server's generic `except Exception` catches it and maps it to a 502 upstream error, making it look like an infrastructure failure when it's actually a client input issue.

### Fix

Catch `ValueError` before the catch-all `except Exception` in the `/search` endpoint handler and raise `HTTPException(status_code=400)` with the original validation message.

### Tests

Added 2 tests in `TestSearchValidationErrors`:
- Empty filters returns 400 with validation message
- No identifiers at all returns 400

All existing tests pass.

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -451,6 +451,8 @@ def search_memories(search_req: SearchRequest, _auth=Depends(verify_auth)):
         if search_req.explain is not None:
             params["explain"] = search_req.explain
         return get_memory_instance().search(query=search_req.query, filters=filters, **params)
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e))
     except Exception:
         raise upstream_error()
 
--- a/server/main.py
+++ b/server/main.py
@@ -453,6 +453,8 @@ def search_memories(search_req: SearchRequest, _auth=Depends(verify_auth)):
         return get_memory_instance().search(query=search_req.query, filters=filters, **params)
     except ValueError as e:
         raise HTTPException(status_code=400, detail=str(e))
+    except HTTPException:
+        raise
     except Exception:
         raise upstream_error()
 
--- a/tests/test_server_params.py
+++ b/tests/test_server_params.py
@@ -682,3 +682,22 @@ class TestSearchEntityIdMapping:
         _, kwargs = mock_memory.search.call_args
         assert kwargs["filters"]["user_id"] == "u1"
         assert kwargs["filters"]["category"] == "food"
+
+
+class TestSearchValidationErrors:
+    """Verify that ValueError from Memory.search() returns 400, not 502."""
+
+    def test_empty_filters_returns_400(self, client, mock_memory):
+        mock_memory.search.side_effect = ValueError(
+            "filters must contain at least one of: user_id, agent_id, run_id"
+        )
+        resp = client.post("/search", json={"query": "food", "filters": {}})
+        assert resp.status_code == 400
+        assert "filters must contain" in resp.json()["detail"]
+
+    def test_no_identifiers_returns_400(self, client, mock_memory):
+        mock_memory.search.side_effect = ValueError(
+            "filters must contain at least one of: user_id, agent_id, run_id"
+        )
+        resp = client.post("/search", json={"query": "food"})
+        assert resp.status_code == 400


kartik-mem0:
## Changes Requested

### Strengths
- **server/main.py:454** — The fix placement is correct: `except ValueError as e` sits immediately before `except Exception`, so the more-specific handler fires first and the catch-all still handles everything else. Minimal surface area.
- **tests/test_server_params.py:687–703** — Both tests use `side_effect = ValueError(...)` to exercise the real exception-routing path in the handler, not just mock a constant return. The first test also asserts the detail content, which verifies the validation message passes through to the client.
- Conventional commit title, zero new dependencies.

### Issues

#### Important (should fix)

1. **server/main.py:454–455 — Missing `except HTTPException: raise` guard**
   `get_all_memories` (lines 414–416 in the same file) already uses:
   ```python
   except HTTPException:
       raise
   except Exception:
       raise upstream_error()
   ```
   because it raises `HTTPException(403)` inside the try block — and without the guard, `except Exception` would swallow it as a 502. The `search_memories` handler doesn't currently raise `HTTPException` inside the try block, so this isn't an active bug, but the inconsistency with the established file pattern is worth closing. Adding `except HTTPException: raise` between your new `except ValueError` and `except Exception` brings it in line.

2. **Duplicate PR cluster — four open PRs against #5367**
   PRs #5369, #5372, #5432, and this one all target the same issue. Worth flagging that #5432 takes a slightly different approach: it validates the required identifiers at the HTTP handler layer (before calling `Memory.search()` at all) and includes the `except HTTPException: raise` guard. Either approach is defensible, but only one can merge — Kartik will need to pick one and close the others.

#### Minor (optional)

1. **tests/test_server_params.py:700–703 — Second test omits detail assertion**
   `test_empty_filters_returns_400` (line 695) checks both the status code and `resp.json()["detail"]`. `test_no_identifiers_returns_400` only checks the status code. One line — `assert "filters must contain" in resp.json()["detail"]` — makes the two tests symmetric.

### Evidence
- **CI:** Vercel deploy-auth fails (requires manual authorization — infra, not code). All 6 code checks pass: build_mem0 on Python 3.10/3.11/3.12, changelog_check, check_changes, license/cla.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/test_server_params.py::TestSearchValidationErrors -v --tb=short` → 2 passed in 0.63s
- **Red-green proof:** reverted `server/main.py` to `origin/main` → both tests fail (`assert 502 == 400`, ValueError propagates to upstream_error handler); restored PR's fix → 2 passed. Proof valid.
- **Lint:** `ruff check server/main.py tests/test_server_params.py` → All checks passed.
- **Linked issue:** #5367 — PR directly addresses the reported root cause (ValueError from Memory.search() mapping to 502 instead of 400).
- **Could NOT verify:** live server behavior (code tracing + mocked unit tests only).

### Assessment
Ready to merge: **With fixes** — please add `except HTTPException: raise` before `except Exception` (one line), and optionally the detail assertion in the second test. I'll re-review once you push. If Kartik decides to go with #5432 instead, no action needed on your end.

HrushiYadav:
Thanks for the review! Pushed the fix:

1. **Important:** Added `except HTTPException: raise` between `except ValueError` and `except Exception`, matching the established pattern in `get_all_memories`.

kartik-mem0:

Verified against da9afcab: the `except HTTPException: raise` guard is in place at server/main.py:455–456, slotted correctly between the ValueError handler and the catch-all. That was the one open item.

### What changed since last review
Added 2 lines to server/main.py — `except HTTPException: raise` between the ValueError catch and the `except Exception` catch-all. No other files touched.

### Strengths
- Fix is surgical and correct: ValueError → 400, HTTPException passthrough, everything else → upstream_error (502).
- Exception ordering matches the established pattern in `get_all_memories`.
- Both regression tests pass and cover the real exception-routing path.

### Issues
Minor (optional): `test_no_identifiers_returns_400` still doesn't assert the detail content (asymmetric with the first test). Not a blocker.

### Evidence
- **CI:** Python SDK build_mem0 (3.10, 3.11, 3.12) pass; CI Gate pass; Vercel fail is infra-only (auth), unrelated.
- **Targeted tests:** `PYTHONPATH=server AUTH_DISABLED=true hatch -e dev_py_3_12 run pytest tests/test_server_params.py::TestSearchValidationErrors -v` → 2 passed in 0.70s
- **Red-green proof:** carried from prior run (old SHA 2e4daf62) — tests fail 502 without fix, pass 400 with it; the new 2-line commit only adds the HTTPException guard and does not alter the red-green behavior.
- **Lint:** `ruff check server/main.py tests/test_server_params.py` → All checks passed!
- **Linked issue:** #5367 — PR addresses the reported root cause (ValueError escaping as 502).
- **Could NOT verify:** live Docker stack behavior — code tracing and mocked unit tests only.

### Assessment
Fix is correct, pattern-consistent, and regression-tested. The duplicate cluster (PRs #5369, #5372, #5432) can be closed in favour of this one.

kartik-mem0 (approved):

PR #5484, opened 2026-06-11
fix: compute score_and_rank divisor per candidate instead of globally

HrushiYadav:
## Summary

`score_and_rank` computes a `max_possible` divisor based on whether **any** candidate in the result set has BM25 or entity scores, then applies that divisor to **all** candidates. This systematically penalizes memories that matched only on semantic similarity when other memories happened to have keyword hits.

**Example:** A memory with `semantic_score=0.9` and no BM25 match gets normalized to `0.9 / 2.0 = 0.45` simply because a different memory in the same result set had a BM25 score. The memory's final score is halved through no fault of its own.

## Fix

Move `max_possible` computation inside the per-candidate loop so each memory's divisor reflects only the signals **it** actually received:

- Semantic only → `max_possible = 1.0`
- Semantic + BM25 → `max_possible = 2.0`
- Semantic + entity → `max_possible = 1.5`
- All three → `max_possible = 2.5`

This ensures a purely semantic match is never penalized by another candidate's keyword or entity hit.

## Test plan

- [x] Added `test_per_candidate_divisor_no_penalty_for_missing_bm25` — verifies semantic-only candidate keeps full score when another has BM25
- [x] Added `test_per_candidate_divisor_entity_only_one_candidate` — verifies entity boost only affects the candidate that has it
- [x] Added `test_per_candidate_divisor_explain_shows_candidate_max` — verifies `score_details.max_possible_score` reflects per-candidate value
- [x] All 23 scoring tests pass (20 existing + 3 new)

### DIFF
--- a/mem0/utils/scoring.py
+++ b/mem0/utils/scoring.py
@@ -74,7 +74,8 @@ def score_and_rank(
     Threshold gates the semantic score BEFORE combining -- candidates
     below the threshold are excluded even if BM25/entity would boost them.
 
-    The divisor adapts based on which signals are active:
+    The divisor adapts per candidate based on which signals that
+    candidate actually received:
         - Semantic only: max_possible = 1.0
         - Semantic + BM25: max_possible = 2.0
         - Semantic + BM25 + entity: max_possible = 2.5
@@ -91,15 +92,6 @@ def score_and_rank(
     Returns:
         List of scored result dicts sorted by combined score descending.
     """
-    has_bm25 = bool(bm25_scores)
-    has_entity = bool(entity_boosts)
-
-    max_possible = 1.0
-    if has_bm25:
-        max_possible += 1.0
-    if has_entity:
-        max_possible += ENTITY_BOOST_WEIGHT
-
     scored: List[Dict[str, Any]] = []
 
     for result in semantic_results:
@@ -114,6 +106,12 @@ def score_and_rank(
         mem_id_str = str(mem_id)
         bm25_score = bm25_scores.get(mem_id_str, 0.0)
         entity_boost = entity_boosts.get(mem_id_str, 0.0)
+
+        max_possible = 1.0
+        if bm25_score > 0.0:
+            max_possible += 1.0
+        if entity_boost > 0.0:
+            max_possible += ENTITY_BOOST_WEIGHT
 
         raw_combined = semantic_score + bm25_score + entity_boost
         combined = min(raw_combined / max_possible, 1.0)
--- a/tests/utils/test_scoring.py
+++ b/tests/utils/test_scoring.py
@@ -149,6 +149,45 @@ class TestScoreAndRank:
         assert "score_details" not in scored[0]
 
 
+    def test_per_candidate_divisor_no_penalty_for_missing_bm25(self):
+        results = [
+            {"id": "a", "score": 0.9, "payload": {"data": "mem a"}},
+            {"id": "b", "score": 0.7, "payload": {"data": "mem b"}},
+        ]
+        bm25 = {"b": 0.8}
+        scored = score_and_rank(results, bm25, {}, threshold=0.1, top_k=10)
+        # "a" has no BM25 hit -> max_possible=1.0 -> score=0.9
+        # "b" has BM25 hit -> max_possible=2.0 -> score=(0.7+0.8)/2.0=0.75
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- The bug description is accurate: a memory with high semantic relevance can get its score halved when any other candidate in the result set happened to match a keyword. The `score_and_rank` docstring and the BM25 sigmoid parameterization (`mem0/utils/scoring.py:16–40`) confirm this is a real normalization issue worth fixing.
- Three targeted regression tests (`test_per_candidate_divisor_no_penalty_for_missing_bm25`, `test_per_candidate_divisor_entity_only_one_candidate`, `test_per_candidate_divisor_explain_shows_candidate_max`) are well-written, have inline math comments, and all correctly fail against the pre-fix code and pass with the fix.
- Minimal, surgical diff: only the affected lines moved; lint clean; no new dependencies.

### Issues

#### Critical (must fix)
1. `mem0/utils/scoring.py:107–112` — The per-candidate divisor introduces a non-monotonicity: a **positive BM25 score can lower a candidate's final score relative to having no BM25 score at all.**

   Concrete example:
   - Candidate A: `sem=0.90, bm25=0.0` → `score = 0.90 / 1.0 = 0.900`
   - Candidate B: `sem=0.90, bm25=0.04` → `score = 0.94 / 2.0 = 0.470`

   B matched a keyword. B ranks lower than A. The crossover condition is `bm25_normalized >= semantic_score`. With the short-query sigmoid params `(midpoint=5.0, steepness=0.7)`, any raw BM25 score below ~7.85 triggers the penalty — that's weak-to-moderate keyword matches, which is the common case.

   The old global divisor didn't have this property: both A and B would be divided by the same `2.0`, so B (`0.47`) would still outrank or equal A (`0.45`) whenever B has any keyword signal. The fix solves one ordering problem and introduces a worse one.

   **How to fix:** The right approach is to keep a **fixed-coefficient additive model** that doesn't inflate the denominator based on what a candidate received. Two options:
   - Keep the global `max_possible` (current behavior) and document that `0.45` is the correct calibrated score for a semantic-only match in a hybrid query. The numerator already credits zero for missing signals via `.get(mem_id_str, 0.0)`.
   - Move to explicit fixed weights: `score = (w_sem * semantic + w_bm25 * bm25 + w_entity * entity)` where `w_sem + w_bm25 + w_entity = 1.0`, so no per-candidate divisor logic is needed at all.

#### Important (should fix)
2. `mem0-ts/src/oss/src/utils/scoring.ts:111–120` — The TypeScript counterpart has the **identical global-divisor logic**. If the Python fix ships, the TS SDK retains the original behavior and the two SDKs diverge. Please include the TS change (or a tracking issue) in whichever direction the Python fix goes.

#### Minor (optional)
3. `mem0/utils/scoring.py:77–82` — The docstring table still reads as query-level ("Semantic + BM25: max_possible = 2.0") even though the fix makes it per-candidate. Worth updating after the design question above is resolved.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 all pass. CI Gate pass. Vercel = deploy-auth (infra, not code).
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/utils/test_scoring.py -v --tb=short` → 23/23 passed
- **Red-green proof:** All 3 new tests fail on the pre-fix implementation (global divisor penalizes semantic-only candidates as described), pass with the fix ✓
- **Non-monotonicity proof:** Mathematically verified — `(0.90 + 0.04) / 2.0 = 0.47 < 0.90 / 1.0 = 0.90`; fires for any positive BM25 below ~7.85 raw score with current sigmoid params
- **TS counterpart:** `mem0-ts/src/oss/src/utils/scoring.ts:111–120` confirmed to have the same global `hasBm25`/`hasEntity` pattern
- **Could NOT verify:** live retrieval behavior end-to-end (code tracing and unit tests only)

### Assessment
Ready to merge: **With fixes** — please address the Critical item (per-candidate divisor non-monotonicity) and include the TypeScript counterpart. Happy to re-review once updated. `gh pr merge --squash` once resolved.

HrushiYadav:
Closing — the reviewer correctly identified that the per-candidate divisor introduces non-monotonicity (a positive BM25 score can lower the final score). The global divisor is the correct design. Thank you for the thorough review.

PR #5485, opened 2026-06-11
fix: implement $not filter support in ChromaDB vector store

HrushiYadav:
## Summary

`_generate_where_clause` silently drops `$not` filter conditions ([line 318-320](https://github.com/mem0ai/mem0/blob/main/mem0/vector_stores/chroma.py#L318-L320)), so any query using NOT filters against ChromaDB returns unfiltered results. This means `$not` filters work correctly with pgvector/qdrant but are silently ignored on ChromaDB.

## Fix

Implement `$not` by negating each condition using De Morgan's law and ChromaDB's native comparison operators:

| Original op | Negated to |
|---|---|
| `eq` / simple value | `$ne` |
| `ne` | `$eq` |
| `gt` | `$lte` |
| `gte` | `$lt` |
| `lt` | `$gte` |
| `lte` | `$gt` |
| `in` | `$nin` |
| `nin` | `$in` |

Multi-field conditions within a single NOT group correctly produce `$or` (De Morgan's: `NOT (A AND B) = !A OR !B`), and multiple groups produce `$and` (`NOT (A OR B) = !A AND !B`).

## Test plan

- [x] `test_generate_where_clause_not_single_equality` — `$not` with simple value → `$ne`
- [x] `test_generate_where_clause_not_multiple_conditions` — multiple NOT groups → `$and`
- [x] `test_generate_where_clause_not_with_operators` — `gt` → `$lte`
- [x] `test_generate_where_clause_not_in_to_nin` — `in` → `$nin`
- [x] `test_generate_where_clause_not_multi_field_condition` — multi-field AND → `$or` (De Morgan's)
- [x] `test_generate_where_clause_not_combined_with_other_filters` — `$not` alongside regular filters
- [x] All 23 ChromaDB tests pass (17 existing + 6 new)

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -316,8 +316,32 @@ class ChromaDB(VectorStoreBase):
                     processed_filters.append(or_conditions[0])
             
             elif key == "$not":
-                # Handle NOT conditions - ChromaDB doesn't have direct NOT, so we'll skip for now
-                continue
+                negate_op = {
+                    "eq": "$ne", "ne": "$eq",
+                    "gt": "$lte", "gte": "$lt",
+                    "lt": "$gte", "lte": "$gt",
+                    "in": "$nin", "nin": "$in",
+                }
+                negated_per_group = []
+                for condition in value:
+                    negated_fields = []
+                    for sub_key, sub_value in condition.items():
+                        if isinstance(sub_value, dict):
+                            for op, val in sub_value.items():
+                                neg = negate_op.get(op)
+                                if neg:
+                                    negated_fields.append({sub_key: {neg: val}})
+                        else:
+                            negated_fields.append({sub_key: {"$ne": sub_value}})
+                    if len(negated_fields) > 1:
+                        negated_per_group.append({"$or": negated_fields})
+                    elif len(negated_fields) == 1:
+                        negated_per_group.append(negated_fields[0])
+
+                if len(negated_per_group) > 1:
+                    processed_filters.append({"$and": negated_per_group})
+                elif len(negated_per_group) == 1:
+                    processed_filters.append(negated_per_group[0])
                 
             else:
                 # Regular condition
--- a/tests/vector_stores/test_chroma.py
+++ b/tests/vector_stores/test_chroma.py
@@ -252,6 +252,48 @@ def test_generate_where_clause_non_string_values():
     assert result == expected
 
 
+def test_generate_where_clause_not_single_equality():
+    """Test $not with a single equality condition."""
+    filters = {"$not": [{"status": "archived"}]}
+    result = ChromaDB._generate_where_clause(filters)
+    assert result == {"status": {"$ne": "archived"}}
+
+
+def test_generate_where_clause_not_multiple_conditions():
+    """Test $not with multiple conditions (OR semantics, negated to AND)."""
+    filters = {"$not": [{"status": "archived"}, {"type": "draft"}]}
+    result = ChromaDB._generate_where_clause(filters)
+    assert result == {"$and": [{"status": {"$ne": "archived"}}, {"type": {"$ne": "draft"}}]}
+
+
+def test_generate_where_clause_not_with_operators():
+    """Test $not negates comparison operators correctly."""
+    filters = {"$not": [{"count": {"gt": 5}}]}
+    result = ChromaDB._generate_where_clause(filters)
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Implements `$not` filter support in the ChromaDB vector store. Previously `_generate_where_clause` hit `continue` on `$not` and silently dropped the condition, so NOT filters returned **unfiltered** results on ChromaDB (while working correctly on pgvector/qdrant). The fix negates each condition with ChromaDB's native operators via De Morgan's law, and adds 6 unit tests.

### Strengths
- Negation table is correct (`eq→$ne`, `gt→$lte`, `in→$nin`, …) and the De Morgan grouping is right: multiple fields within one NOT group → `$or`, multiple NOT groups → `$and`.
- Behavior now matches the other stores — pgvector (`pgvector.py:70-78`) and qdrant (`qdrant.py:355-359`) both reduce `$not` to an AND-of-negations, so ChromaDB is finally consistent with them.
- Multi-operator-in-one-group is handled correctly: `$not: [{count: {gt: 5, lt: 10}}]` → `$or` of `$lte 5` / `$gte 10`.
- The six tests follow the existing static-method test pattern, and the full `test_chroma.py` (23 tests) stays green.

### Issues
Minor (optional): operators outside the 8-entry `negate_op` map — notably `contains`/`icontains` — are silently skipped inside `$not` (the `if neg:` guard has no `else`), whereas the positive `convert_condition` path falls back to `$eq`. This is **not a regression** (pre-PR the whole `$not` was dropped), and ChromaDB has no native substring match anyway, so the practical impact is negligible. A one-line `else` fallback (or a `logger.warning`) for unknown ops — plus a test for `$not: [{k: {contains: ...}}]` — would make the two paths consistent. Optional; not a merge blocker.

### Evidence
- **CI:** required **CI Gate: pass**; Python SDK builds pass on 3.10/3.11/3.12; CLA signed. The only red check is **Vercel = "Authorization required to deploy"** — a preview-deploy auth gate for the fork, not a code/test failure.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_chroma.py -v` → **23 passed in 1.08s**.
- **Red-green proof:** reverting `chroma.py` to main (the `continue` no-op) makes the 6 new tests fail — `assert {} == {'status': {'$ne': 'archived'}}` — and restoring the fix makes them pass again ✓. The regression tests genuinely bite.
- **Linked issue:** none referenced in the body (the description links the offending lines directly). Worth adding `Fixes #<n>` if a tracking issue exists.
- **Cross-package check:** there is no TypeScript ChromaDB store in `mem0-ts`, so there's no parallel `$not` gap to fix; no downstream caller in `server/`/`openmemory/`/`cli/` builds Chroma `$not` directly.
- **Could NOT verify:** end-to-end behavior against a live ChromaDB instance — verified via `_generate_where_clause` output plus code tracing only.

### Assessment
Surgical, well-tested bug fix with proven red-green and confirmed cross-store semantic parity; the only note is an optional consistency nit on unrecognized operators.

kartik-mem0:
### End-to-end verification update (live ChromaDB)

My LGTM above noted one caveat — "could not verify end-to-end behavior against a live ChromaDB instance." I've now closed that gap: ran a real persistent ChromaDB (`chromadb 1.5.9`) on this branch, inserted 6 rows, and checked that `$not` filters through `ChromaDB.list()`/`.search()` return exactly the logical NOT of each filter.

- **With the fix (this PR):** 5/5 filter shapes return the correct IDs and `search()` excludes the negated rows — `$not [{status:archived}]` → `$ne`, `$not [{count:{gt:5}}]` → `$lte`, `in`→`$nin`, the multi-field De Morgan case → `$or`, and `$not` combined with a regular filter → `$and`. ChromaDB accepts every negated operator the PR emits.
- **Reverting just `chroma.py` to `main`:** the same script breaks — `$not`-only filters collapse to an empty `where` that ChromaDB rejects (`ValueError: Expected where to have exactly one operator, got {}`), and a `$not` combined with another filter silently leaks the row it should exclude.

So the fix is confirmed end-to-end, not just at the where-clause level. No change to the verdict — ready for squash-merge.

(Optional, unchanged from above: operators outside the 8-entry map such as `contains`/`icontains` are still silently skipped inside `$not`. Not a regression and not a blocker — just a small consistency nit if you want to add an `else` fallback later.)

kartik-mem0 (approved):

PR #5488, opened 2026-06-11
fix: use 'is not None' instead of truthiness for vector/payload in pgvector update

HrushiYadav:
## Summary

In `PGVector.update()`, the guards `if vector:` and `if payload:` use Python truthiness to decide whether to persist updates. Empty vectors (`[]`) and empty payloads (`{}`) are falsy, so passing them is silently ignored instead of triggering the UPDATE query.

## Fix

Replace `if vector:` with `if vector is not None:` and `if payload:` with `if payload is not None:` so that empty-but-intentional values are still persisted.

## Test plan

- [x] `test_update_empty_payload_still_persisted` — verifies `payload={}` triggers an UPDATE query
- [x] All existing pgvector tests pass

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -425,12 +425,12 @@ class PGVector(VectorStoreBase):
         """
         self._ensure_collection()
         with self._get_cursor(commit=True) as cur:
-            if vector:
+            if vector is not None:
                cur.execute(
                     sql.SQL("UPDATE {} SET vector = %s WHERE id = %s").format(self._col()),
                     (vector, vector_id),
                 )
-            if payload:
+            if payload is not None:
                 # Handle JSON serialization based on psycopg version
                 if PSYCOPG_VERSION == 3:
                     # psycopg3 uses psycopg.types.json.Json
--- a/tests/vector_stores/test_pgvector.py
+++ b/tests/vector_stores/test_pgvector.py
@@ -2089,6 +2089,39 @@ class TestPGVector(unittest.TestCase):
         self.assertTrue(len(vector_update_calls) > 0)
         self.assertTrue(len(payload_update_calls) > 0)
 
+    @patch('mem0.vector_stores.pgvector.PSYCOPG_VERSION', 3)
+    @patch('mem0.vector_stores.pgvector.ConnectionPool')
+    @patch.object(PGVector, '_get_cursor')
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Fixes a silent data-loss bug in `PGVector.update()`: the old `if vector:` and `if payload:` guards used Python truthiness, so passing `vector=[]` or `payload={}` (empty but intentional) would silently skip the UPDATE query. The fix uses `is not None`, which is correct for `Optional[T]` parameters where `None` means "don't touch this field" and any actual value — including falsy ones — means "persist this."

### Strengths
- `pgvector.py:428,433` — Two-character fix with exactly the right semantics. `Optional[T]` sentinels must use identity checks, not truthiness.
- `test_pgvector.py:2095` — `test_update_empty_payload_still_persisted` exercises the precise failing case (`payload={}`) and integrates cleanly with the existing mock infrastructure in the class.
- Conventional commit title, no new dependencies, ruff clean.

### Issues
Minor (optional):
1. No test for `vector=[]` (empty vector). Semantically unlikely in practice, but the fix covers it and a test would be complete parity with the payload case.
2. No linked issue (`Closes #X`) in the PR body — minor traceability gap.

### Evidence
- **CI:** Python SDK builds pass on 3.10, 3.11, 3.12. CI Gate green. Vercel "fail" is infra deploy-auth, not a code signal. All unaffected packages correctly skipping.
- **Lint:** `ruff check mem0/vector_stores/pgvector.py` → `All checks passed!`
- **Targeted tests:** `pytest tests/vector_stores/test_pgvector.py -q` → `83 passed in 0.45s`
- **Red-green proof:** regression test fails with `if payload:` (main), passes with `if payload is not None:` (PR) ✓
  - Phase 1 (PR code): `1 passed`
  - Phase 2 (reverted to main): `1 failed — AssertionError: False is not true : Empty payload {} should still trigger UPDATE`
  - Phase 3 (fix restored): `1 passed`
- **Linked issue:** none in PR body
- **Cross-package check:** TS counterpart (`mem0-ts/src/oss/src/vector_stores/pgvector.ts`) takes both params as required (no optional guards), so the same bug doesn't exist there.
- **Could NOT verify:** live PostgreSQL/pgvector behavior — code tracing and mock-based tests only.

### Assessment
Red-green proof is clean, fix is minimal and correct, full test suite passes with no regressions.

kartik-mem0 (approved):

PR #5494, opened 2026-06-11
feat(memory): add OSS-to-Platform notices

youneshima:
## Description

Adds Python OSS SDK notices.

This PR adds:
- shared notice helper logic using the existing OSS telemetry client
- server-controlled PostHog flag evaluation via `mem0-oss-notices`
- `mem0.notice_displayed` telemetry events allowlisted at 100% sampling
- first-run CTA notice
- temporal feature-error notice for `timestamp` / `reference_date`
- temporal usage notice for time-aware workflows
- decay feature-error notice for `memory.project.update(decay=True)`
- decay usage notice for repeated deletes / bulk deletes
- scale threshold notice
- performance slow-query notice
- unit coverage for notice behavior, state, telemetry, caps, sync paths, and async paths

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Local checks:
- `pytest tests/memory/test_notices.py tests/memory/test_temporal_feature_notice.py tests/memory/test_temporal_usage_notice.py tests/memory/test_decay_feature_notice.py tests/memory/test_decay_usage_notice.py tests/memory/test_performance_slow_query_notice.py -q`
- `pytest tests/test_telemetry.py tests/test_telemetry_sampling.py tests/test_main.py tests/test_memory.py -q`
- `ruff check mem0/memory/notices.py mem0/memory/main.py tests/memory/test_notices.py tests/memory/test_temporal_feature_notice.py tests/memory/test_temporal_usage_notice.py tests/memory/test_decay_feature_notice.py tests/memory/test_decay_usage_notice.py tests/memory/test_performance_slow_query_notice.py tests/test_telemetry.py`
- `git diff --check`

Manual real OSS fixture checks:
- verified displayed, holdout, per-notice disabled, telemetry-off, cap, sync, and async behavior against a real local OSS fixture and PostHog events for all notices

## Checklist


kartik-mem0 (commented):
Review of the OSS→Platform notices feature. Posting the selected hot-path / robustness / docs items below, tagged by severity. Overall the feature is well-structured and thoroughly tested — these are refinements rather than happy-path correctness bugs.

mem0/memory/notices.py
```diff
@@ -0,0 +1,1359 @@
...
+    memory_instance,
+    add_result: Any,
+) -> Optional[Tuple[str, str, Optional[int], Optional[int], int]]:
+    if not telemetry_module.MEM0_TELEMETRY:
+        return None
+
+    if _count_added_memories(add_result) == 0:
+        return None
+
+    provider_count = _get_provider_memory_count(memory_instance)
```
kartik-mem0: **[perf — should fix]** `_get_provider_memory_count()` issues a `vector_store.count()` / `col_info()` round-trip here on **every successful `add()`** when telemetry is on. For users under `SCALE_MEMORY_COUNT_THRESHOLD` it's pure overhead, and for users above it the count keeps running on every add even though the notice can fire only once.

The `memory_count_threshold_evaluated` short-circuit (line 420) is checked *after* the count. Please move that cheap check — ideally behind an in-process flag like `_first_run_claimed_in_process` — *before* `_get_provider_memory_count()`, so the count stops once the notice has been emitted. Even better, sample it rather than counting on every add.

mem0/memory/notices.py
```diff
@@ -0,0 +1,1359 @@
...
+            decay_state = _get_notice_state(config, DECAY_USAGE_STATE_KEY)
+            delete_count = _coerce_nonnegative_int(decay_state.get("successful_delete_count"), 0) + 1
+            decay_state["successful_delete_count"] = delete_count
+
+            state = config.get(STATE_SECTION)
+            if not isinstance(state, dict):
+                state = {}
+            state[DECAY_USAGE_STATE_KEY] = decay_state
+            config[STATE_SECTION] = state
+            _write_config(config)
```
kartik-mem0: **[perf / robustness — should fix]** `detect_decay_usage_from_delete()` does a `_load_config()` + `_write_config()` on **every `delete()`** (telemetry on), and `successful_delete_count` grows unbounded — this write keeps happening forever, even after the decay-usage notice has hit its cap.

Two issues:
1. Hot-path file I/O on every delete.
2. `_write_config()` (setup.py) is a non-atomic truncate-then-write of `~/.mem0/config.json`, which also stores `user_id` / CLI `aliased_pairs`. Moving config writes onto hot paths widens the window for a concurrent or interrupted write to corrupt that shared file.

Suggest: stop persisting once threshold + cap are reached (or keep the counter in-process), and make `_write_config` atomic (write temp + `os.replace`).

mem0/memory/notices.py
```diff
@@ -0,0 +1,1359 @@
...
+
+        if displayed:
+            return copy
+    except Exception:
+        return plain_error
+
+    return plain_error
+
+
+def detect_temporal_usage_from_metadata(metadata: Optional[Dict[str, Any]]) -> Optional[Tuple[str, str]]:
```
kartik-mem0, line 821: **[robustness — should fix]** Every other detect/display helper in this module is wrapped so it can never raise into the caller, but `detect_temporal_usage_from_metadata()` (and `detect_temporal_usage_from_search()` at line 791) are not — and they run at the very top of `add()` / `search()` *before* the real work.

`_walk_mapping` / `_has_temporal_filter` recurse with no depth guard, so a deeply nested or cyclic `metadata` / `filters` dict would raise `RecursionError` straight out of `add()` / `search()` — breaking a core operation for the sake of a notice. Please wrap both in `try/except Exception: return None` to honor the module's never-break-the-op contract.

mem0/memory/notices.py
```diff
@@ -0,0 +1,1359 @@
...
+        )
+
+        if displayed:
+            print(copy, file=sys.stderr)
+    except Exception:
+        if variant is not None:
+            _update_first_run_variant(variant)
+
+
+async def display_first_run_notice_async(memory_instance, sync_type: str, trigger_function: str) -> None:
```
kartik-mem0, line 149: **[perf]** `display_first_run_notice_async` is awaited on the *else* branch of every async `add/search/get/get_all/update/delete/...`, so it hops to a thread via `asyncio.to_thread` on **every async op** — even though after the first call the sync function just takes the lock and returns (first run already claimed). For high-throughput async callers that's a thread-pool dispatch per call for nothing. A cheap guard before the dispatch avoids it:

```python
async def display_first_run_notice_async(memory_instance, sync_type, trigger_function):
    if not telemetry_module.MEM0_TELEMETRY or _first_run_claimed_in_process:
        return
    await asyncio.to_thread(display_first_run_notice, memory_instance, sync_type, trigger_function)
```

mem0/memory/main.py
```diff
@@ -2066,6 +2193,7 @@ async def add(
             agent_id (str, optional): ID of the agent creating the memory. Defaults to None.
             run_id (str, optional): ID of the run creating the memory. Defaults to None.
             metadata (dict, optional): Metadata to store with the memory. Defaults to None.
+            timestamp (Any, optional): Temporal reasoning parameter.
```
kartik-mem0: **[nit / docs]** This async docstring describes `timestamp` (and `reference_date` at line 2779) as `Temporal reasoning parameter.`, but the param unconditionally `raise`s `ValueError` in OSS. The sync versions (lines 673 / 1267) correctly say `Platform-only temporal parameter. Not supported in OSS.` — please match them so the async docs aren't misleading.

youneshima:
Addressing the dependency bump note explicitly:

This PR raises `posthog` from `>=4.5.0` to `>=7.14.0` because the notices use the current PostHog Python flag API:

- `Posthog.evaluate_flags(distinct_id, flag_keys=[...])`
- `flags.get_flag(...)`
- `flags.get_flag_payload(...)`
- `capture(..., flags=flags)`

The project already requires Python `>=3.10`, and `posthog>=7.14.0` is compatible with that floor. I agree this should be called out in release notes because downstream users pinned to older PostHog versions may see resolver conflicts.

mem0/memory/notices.py
```diff
@@ -0,0 +1,1426 @@
...
+                first_run = {"consumed": True}
+            first_run["variant"] = variant
+            state[STATE_KEY] = first_run
+            config[STATE_SECTION] = state
+            _write_config(config)
+    except Exception:
+        return
+
+
+def _temporal_usage_at_capacity() -> bool:
```
kartik-mem0, line 985: **[perf — should fix]** The in-process cap short-circuit you added for decay (`_decay_usage_capacity_reached_in_process`, `:961`) is **missing** here and in `_scale_threshold_at_capacity` (`:1053`) and `_performance_slow_query_at_capacity` (`:1164`). All three still `_load_config()` from disk under `_state_lock` on **every** triggering `add()`/`search()` *after* the cap is already exhausted — so a user who keeps hitting temporal/scale/slow-query conditions pays a disk read + JSON parse per op forever.

Mirror the decay pattern in all three: a module-level `_<x>_capacity_reached_in_process` flag, checked before the lock/`_load_config()` and set once `len(entries) >= CAP`.

mem0/memory/notices.py
```diff
@@ -0,0 +1,1426 @@
...
+    if not _claim_first_run_notice(trigger_function):
+        return
+
+    variant = None
+    try:
+        telemetry = telemetry_module._get_oss_telemetry()
+        if telemetry is None or telemetry.posthog is None or not telemetry.user_id:
+            return
+
+        flags = telemetry.posthog.evaluate_flags(telemetry.user_id, flag_keys=[FLAG_KEY])
```
kartik-mem0, line 94: **[perf — consider]** With the file-I/O fixes in, the remaining synchronous cost on the hot path is `evaluate_flags()` itself — a PostHog `POST /flags` (`feature_flags_request_timeout_seconds=0.5`) on the sync `add()`/`search()`/`get_all()` return path. The cap bounds it to ≤10 calls per 7-day window per notice type, but each adds a network round-trip (tens-to-hundreds of ms, up to 500 ms on timeout) to a real user call.

Not blocking, but worth considering: move the flag eval + capture fully off the return path (background thread / `asyncio.create_task`) or cache the decision per process, so notices add zero latency to the user's operation.

mem0/memory/notices.py
```diff
@@ -0,0 +1,1426 @@
...
+) -> str:
+    if not telemetry_module.MEM0_TELEMETRY:
+        return plain_error
+
+    try:
+        telemetry = telemetry_module._get_oss_telemetry()
+        if telemetry is None or telemetry.posthog is None or not telemetry.user_id:
+            return plain_error
+
+        flags = telemetry.posthog.evaluate_flags(telemetry.user_id, flag_keys=[FLAG_KEY])
```
kartik-mem0, line 755: **[perf — should fix]** Unlike the usage/scale/perf notices, this feature-error path (`_get_feature_error_message`) has **no cap**: every `add(timestamp=...)` / `search(reference_date=...)` / `project.update(decay=True)` that raises `ValueError` first makes a blocking `evaluate_flags()` call **and** emits a telemetry event — every single time. A migration/batch script that repeatedly passes an unsupported arg pays the round-trip on each iteration before the exception surfaces, turning an instant `ValueError` into a network-bound one. Suggest raising immediately and firing the notice/telemetry fire-and-forget, or capping this path too.

mem0/memory/notices.py
```diff
@@ -0,0 +1,1426 @@
...
+def detect_temporal_usage_from_metadata(metadata: Optional[Dict[str, Any]]) -> Optional[Tuple[str, str]]:
+    try:
+        if not isinstance(metadata, dict):
+            return None
+
+        for key, value in _walk_mapping(metadata):
+            temporal_key = _is_temporal_key(key)
+            if temporal_key and _looks_temporal_value(value, allow_epoch=True):
+                return ("metadata", "date_like_metadata")
+            if _looks_temporal_value(value, allow_epoch=False):
```
kartik-mem0: **[correctness — should fix]** This check isn't gated on `temporal_key`, so **any** string value matching a relative-time word ("yesterday", "last week", "next month") or an ISO date — in **any** metadata field — triggers the temporal-usage notice (and its network call). e.g. `add(..., metadata={"notes": "we met yesterday", "category": "work"})` fires it even though the user isn't using temporal features.

Gate it on `temporal_key` like the line above:
```python
if temporal_key and _looks_temporal_value(value, allow_epoch=False):
    return ("metadata", "date_like_metadata")
```
(or add a comment if catching temporal text anywhere is the intent).

mem0/memory/notices.py
```diff
@@ -0,0 +1,1426 @@
...
+    r"next\s+(?:week|month|year)|"
+    r"(?:past|last)\s+\d+\s+(?:day|days|week|weeks|month|months|year|years)|"
+    r"(?:since|before|after|until)\s+(?:today|yesterday|tomorrow|\d{4}-\d{2}-\d{2}|last\s+(?:week|month|year))"
+    r")\b",
+    re.IGNORECASE,
+)
+_RANGE_OPERATORS = {"gt", "gte", "lt", "lte"}
+
+_state_lock = threading.Lock()
+_first_run_claimed_in_process = False
```
kartik-mem0, line 68: **[robustness — discuss]** These module-level globals are shared by **all** `Memory` instances in a process. In a multi-tenant server (one `Memory` per user, or a shared instance), `_first_run_claimed_in_process` is consumed by whichever instance runs first (later users never see first-run), and `_decay_usage_successful_delete_count_in_process` accumulates deletes across **all** users — so user A's deletes can trip the decay notice during user B's activity.

If "once per process" is the intended policy that's fine — just worth documenting; otherwise these should be instance-scoped. (Same applies to the TS module globals at `notices.ts:46-51`.)

mem0/memory/main.py
```diff
@@ -2101,6 +2233,13 @@ async def add(
             results = await self._create_procedural_memory(
                 messages, metadata=processed_metadata, prompt=prompt, llm=llm
             )
+            scale_threshold_notice = detect_scale_threshold_from_add_result(self, results)
```
kartik-mem0: **[perf — should fix]** With the count now short-circuited in-process it only runs occasionally (first add + every 100th until the threshold is evaluated) — but on those occasions `detect_scale_threshold_from_add_result(self, results)` is invoked **directly** in the async `add()` path (no `asyncio.to_thread`), and it runs a synchronous vector-store `count()`/`col_info()` (DB round-trip) that blocks the event loop for all concurrent coroutines. The `display_*_notice_async` calls just below correctly use `to_thread`; this one should too. (Also at `:2251`.)

mem0/memory/main.py
```diff
@@ -1244,7 +1335,9 @@ def search(
             },
         )
 
+        search_start = time.perf_counter()
```
kartik-mem0, line 1338: **[nit / telemetry]** The slow-query timer stops at `_search_vector_store`, so reranking latency (when `rerank` + a reranker are configured) is excluded from `search_elapsed_seconds`. The user's perceived latency includes reranking, so both the trigger decision and the `elapsed_ms` in telemetry can understate real slowness. Consider timing through the rerank step, or documenting that this measures vector-store latency only.

mem0-ts/src/oss/src/utils/notices.ts
```diff
@@ -0,0 +1,1380 @@
...
+
+  if (!parsed.found) {
+    return {
+      displayed: false,
+      noticeConfigFound: false,
+      bypassReason: "missing_notice_config",
+    };
+  }
+
+  if (parsed.config?.enabled !== true) {
```
kartik-mem0: **[parity — should fix]** TS treats a missing `enabled` key as **disabled** (`enabled !== true` → suppress), but Python defaults it to **enabled** (`notice_config.get("enabled", True)`, `notices.py:181`). So a PostHog payload that includes a notice config but omits `enabled` (likely during rollout) displays in Python and stays silent in TS — which would skew A/B data. Recommend defaulting to enabled in TS too:
```ts
if (parsed.config?.enabled === false) { ... }
```
(Same pattern at `:473`.)

mem0-ts/src/oss/src/utils/notices.ts
```diff
@@ -0,0 +1,1380 @@
...
+  }
+
+  return {
+    displayed: true,
+    noticeConfigFound: true,
+    copy,
+  };
+}
+
+function getFeatureErrorDecision(
```
kartik-mem0, line 549: **[parity — should fix]** `getFeatureErrorDecision` returns `displayed: true` (around `:561`) based only on `enabled` + `copy`, with **no variant check** — unlike `getDisplayDecision`/`getScaleDisplayDecision` (which check `variant !== DISPLAYED_VARIANT`) and unlike Python's `_get_feature_error_message`, which restricts to `variant in (DISPLAYED_VARIANT, HOLDOUT_VARIANT)` (`notices.py:774`). With two variants today it's invisible, but a future third arm (e.g. "silent"/"experiment") would leak the CTA in TS while Python suppresses it. Add the variant guard to match Python.

Side note: both languages currently show the feature-error CTA to `holdout` users (`notices.py:774`), so holdout isn't a true control for feature errors — worth confirming that's intended.

kartik-mem0 (commented):
## Second-pass review — OSS→Platform notices (re-reviewed at `be19b125`)

Thanks for the quick turnaround on the first round — re-reviewed after the hot-path overhead fixes and the Node side landing. The earlier items look addressed: `detect_temporal_usage_*` is now wrapped, the async first-run guard is in, `_write_config` is atomic, decay keeps its counter in-process, and the scale count short-circuits before counting. 👍

On the **dependency bump** — saw your note, and I agree: `>=7.14.0` is required for the `evaluate_flags` flag API, it's compatible with the `>=3.10` floor, and a release-note callout covers the downstream-pin risk. One small follow-on: that bump makes the `except TypeError` posthog<4.5 fallback in `telemetry.py` unreachable, so it can be dropped.

This pass is a few remaining refinements (same spirit as before — mostly perf/parity, not happy-path bugs), plus the **Node side**, which is new since the last review:

- **Perf:** the in-process cap short-circuit you added for decay is missing for temporal/scale/perf; the remaining synchronous cost is the `evaluate_flags()` network call itself; and one async count still isn't offloaded.
- **Correctness:** temporal detection over-triggers on plain date-like text in any metadata field.
- **Parity:** two Python↔TS divergences in the display decision (`enabled` default + variant check).
- **Surface/robustness:** process-global notice state shared across `Memory` instances; `__noticeTestHooks` exported from the public OSS entrypoint.

Two clarifications so nothing's misread:
- `get()` does **not** double-fire the first-run notice — `main.py:1079` and `:1108` are mutually exclusive branches (one fires per call).
- The `_async` wrappers **do** correctly offload to threads; the only exception is the un-wrapped `count()` flagged inline at `main.py:2236`.

Strengths still hold: failure isolation is airtight, telemetry payloads carry no PII/memory content (only structural fields + the server's copy), `MEM0_TELEMETRY` opt-out is honored end-to-end, the 100% sampling allowlist is isolated from other events, and the suites (Python + the new Node ones) exercise real behavior. Nice work overall.

mem0-ts/src/oss/src/index.ts
```diff
@@ -28,3 +28,4 @@ export * from "./vector_stores/vectorize";
 export * from "./vector_stores/azure_ai_search";
 export * from "./vector_stores/pgvector";
 export * from "./utils/factory";
+export { __noticeTestHooks } from "./utils/notices";
```
kartik-mem0: **[api surface — should fix]** `__noticeTestHooks` is re-exported from the published `mem0ai/oss` entrypoint, so internal state-manipulation helpers become part of the package's public API — once shipped, removing them is a breaking change, and the `__` prefix is convention, not enforcement. Prefer importing these directly from `./utils/notices` in the test files, or expose them via a separate test-only entry that isn't part of the published `index.ts`.

youneshima:
Addressed the second-pass review items:

- Added Python in-process cap short-circuits for `temporal_usage`, `scale_threshold`, and `performance_slow_query`.
- Added capped Python feature-error handling for `temporal_stub` and `decay_stub`.
- Narrowed Python temporal metadata detection so non-temporal fields with date-like text do not trigger `temporal_usage`.
- Offloaded async add scale-count detection with `asyncio.to_thread(...)`.
- Removed the now-unreachable old PostHog fallback in `telemetry.py`.
- Matched TS payload parsing to Python: missing `enabled` now defaults to enabled.
- Added the TS feature-error variant guard so only `displayed` / `holdout` return CTA copy.
- Removed `__noticeTestHooks` from the public `mem0ai/oss` entrypoint; tests now import internal hooks directly.

Checks run:

- `python -m pytest tests/memory/test_notices.py tests/memory/test_temporal_usage_notice.py tests/memory/test_temporal_feature_notice.py tests/memory/test_decay_feature_notice.py`
- `python -m ruff check mem0/memory/notices.py mem0/memory/main.py mem0/memory/telemetry.py tests/memory/test_notices.py tests/memory/test_temporal_usage_notice.py`
- `git diff --check`
- `pnpm test -- notices --runInBand`
- `pnpm run test:unit`
- `pnpm build`

Real fixture smoke checks were also run through installed SDK paths for Node notice families and Python temporal feature error. The Python/Node `decay_stub` fixture returned the expected disabled/plain-error path because the live `decay_stub` payload was disabled during verification.

I left process-global notice bookkeeping as-is in this PR to avoid a broader ownership redesign; happy to make instance-scoping a follow-up if you want that before merge.

kartik-mem0 (approved):

PR #5496, opened 2026-06-11
fix: prevent cross-scope entity linking in _upsert_entity

### LINKED ISSUE #5439: Cross-scope entity linking: entities can link memories across different user_id/agent_id boundaries

HrushiYadav:
Closes #5439

## Problem

`_upsert_entity` uses scoped filters when **searching** for existing entities, but does not verify the matched entity's payload scope before merging. When two users mention the same entity (e.g. "Apple"), user B's `memory_id` gets appended to user A's entity record — leaking cross-scope linked memory IDs.

This causes:
- **Cross-user memory linking** — entity records contain `linked_memory_ids` from multiple users
- **Leaked influence in search** — `_compute_entity_boosts` uses the full unscoped count, so one user's data subtly affects another's ranking

## Fix

Added a scope-match guard in both sync `_upsert_entity` and async `_upsert_entity_async`: after finding a high-similarity entity match (>= 0.95), verify that the matched entity's `user_id`/`agent_id`/`run_id` in the payload match the current `search_filters`. If they don't match, create a new scoped entity instead of merging.

## Tests

3 new tests in `TestEntityScopeGuard`:
- `test_same_scope_merges` — same user, entity merges correctly
- `test_different_scope_creates_new` — different user_id, new entity created
- `test_missing_scope_in_payload_creates_new` — entity missing scope fields, treated as mismatch

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -512,21 +512,26 @@ class Memory(MemoryBase):
                 filters=search_filters,
             )
 
+            matched = False
             if existing and existing[0].score >= 0.95:
-                # Update existing entity's linked_memory_ids
                 match = existing[0]
                 payload = match.payload or {}
-                linked_ids = payload.get("linked_memory_ids", [])
-                if memory_id not in linked_ids:
-                    linked_ids.append(memory_id)
-                    payload["linked_memory_ids"] = linked_ids
-                    self.entity_store.update(
-                        vector_id=match.id,
-                        vector=None,
-                        payload=payload,
-                    )
-            else:
-                # Create new entity
+                scope_match = all(
+                    payload.get(k) == v for k, v in search_filters.items()
+                )
+                if scope_match:
+                    matched = True
+                    linked_ids = payload.get("linked_memory_ids", [])
+                    if memory_id not in linked_ids:
+                        linked_ids.append(memory_id)
+                        payload["linked_memory_ids"] = linked_ids
+                        self.entity_store.update(
+                            vector_id=match.id,
+                            vector=None,
+                            payload=payload,
+                        )
+
+            if not matched:
                 entity_id = str(uuid.uuid4())
                 entity_payload = {
                     "data": entity_text,
@@ -2052,20 +2057,27 @@ class AsyncMemory(MemoryBase):
                 filters=search_filters,
             )
 
+            matched = False
             if existing and existing[0].score >= 0.95:
                 match = existing[0]
                 payload = match.payload or {}
-                linked_ids = payload.get("linked_memory_ids", [])
-                if memory_id not in linked_ids:
-                    linked_ids.append(memory_id)
-                    payload["linked_memory_ids"] = linked_ids
-                    await asyncio.to_thread(
-                        self.entity_store.update,
-                        vector_id=match.id,
-                        vector=None,
-                        payload=payload,
-                    )
-            else:
+                scope_match = all(
+                    payload.get(k) == v for k, v in search_filters.items()
+                )
+                if scope_match:
+                    matched = True
+                    linked_ids = payload.get("linked_memory_ids", [])
+                    if memory_id not in linked_ids:
+                        linked_ids.append(memory_id)
+                        payload["linked_memory_ids"] = linked_ids
+                        await asyncio.to_thread(
+                            self.entity_store.update,
+                            vector_id=match.id,
+                            vector=None,
+                            payload=payload,
+                        )
+
+            if not matched:
                 entity_id = str(uuid.uuid4())
                 entity_payload = {
                     "data": entity_text,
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for this, @HrushiYadav! For #5439 we're consolidating on #5441, which covers all four entity-upsert paths (including the batch add path) and is green on CI. Closing this one so the fix stays in a single place — your review on #5441 would be welcome. 🙏

PR #5497, opened 2026-06-11
fix: add entity_type merge gate to prevent same-name entity conflation

### LINKED ISSUE #5438: Entity disambiguation: same-name entities with different meanings merge at 0.95 similarity threshold
## Component

Core / Python SDK

### Problem

The entity store uses a **0.95 vector similarity threshold** as the sole mechanism to decide whether two entity texts refer to the same entity (`_upsert_entity` at `mem0/memory/main.py:427`, Phase 7d at line 919). When a new entity's embedding matches an existing entity at ≥ 0.95, the system merges them by appending `memory_id` to `linked_memory_ids` instead of creating a separate entity record.

This breaks when entities share the same surface form but have different meanings — a classic **entity disambiguation** problem. Examples:

| Entity | Meaning A | Meaning B | Same entity? |
|--------|-----------|-----------|-------------|
| "Apple" | Apple Inc. | Apple the fruit | No |
| "Tesla" | Nikola Tesla | Tesla Motors | No |
| "Python" | Programming language | The snake | No |
| "Mercury" | The planet | The element | No |
| "Jordan" | The country | Michael Jordan | No |

Because the entity store has no context beyond the entity text's embedding, "Apple" (company) and "apple" (fruit) can merge if their embeddings are close enough. Once merged, they share a single `linked_memory_ids` list, and the entity boost in search (`_compute_entity_boosts`) boosts memories about the fruit when the user asks about the company, and vice versa.

### Current mitigation is insufficient

1. **0.95 threshold** — Too blunt. It prevents obviously different strings from merging, but semantically similar polysemes (like "Apple" company vs. fruit) have embeddings that can exceed 0.95, especially with small embedding models (e.g., `all-MiniLM-L6-v2`).
2. **Entity types (PROPER/QUOTED/COMPOUND/NOUN)** — Only used at extraction time for dedup priority. They are stored in the entity payload but **never consulted during the merge decision**. Two entities of different types could still merge if vectors match.
3. **No context window** — The entity store has no knowledge of the surrounding text or the memory's domain/topic. The disambiguation decision is made purely on the entity text embedding.

### Impact

- **Search quality degradation**: Entity boost surfaces irrelevant memories (boosting "Apple" fruit memories when querying about Apple Inc.).
- **Entity graph pollution**: Over time, a single entity node accumulates `linked_memory_ids` from unrelated domains, making `memory_count_weight` penalize it indiscriminately.
- **Cross-domain contamination in multi-tenant scenarios**: If different users discuss "Apple" in different senses, they share an entity record, creating implicit cross-tenant information leakage (even if search filters prevent the memories from surfacing, the entity count and boost calculations are affected).

### Proposed Solutions

**Option A: Context-enriched entity embeddings (recommended, minimal disruption)**

Instead of embedding the bare entity text, embed a short context window:
```
"{entity_text} — {first 100 chars of memory text}"
```
This moves "Apple" (company) and "apple" (fruit) apart in embedding space because their surrounding contexts differ. The entity store's existing 0.95 threshold then works correctly for most cases.

Implementation:
- In `_upsert_entity` and Phase 7: when embedding entity text, concatenate with a truncated excerpt from the memory text being linked.
- On update (re-embed): use the first remaining linked memory's text as context.
- No schema changes required.

**Option B: Type-aware merge gate**

Before merging, require that matched entities share the same `entity_type`. If "Apple" is extracted as PROPER (company) but the existing "Apple" was QUOTED (a term), don't merge — create a new entity.

Implementation:
- Add `entity_type` check at line 427 and line 919: `if existing and existing[0].score >= 0.95 and existing[0].payload.get("entity_type") == entity_type`.
- Simple, but doesn't handle same-type polysemes (both "Apple" instances could be PROPER).

**Option C: LLM-based disambiguation (highest quality, highest cost)**

When a vector match exceeds 0.95 but is below a higher confidence threshold (e.g., 0.99), ask the LLM whether the two entity contexts refer to the same thing:

```
Entity: "Apple"
Existing context: "User works at Apple as a software engineer"
New context: "User likes to eat green apples"
Same entity? Yes/No
```

This is the most accurate but adds LLM latency on every entity upsert with an ambiguous match.

**Recommended approach**: Start with Option A (context-enriched embeddings) + Option B (type-aware gate). These are low-cost, no-LLM-overhead, and address the majority of polyseme cases. Option C can be added later as an optional high-accuracy mode.

### Environment

- mem0 version: latest (main branch)
- Python version: 3.12+

### Related

- #4884 (English-only entity extraction — disambiguation is even harder across languages)
- #4988 (entity cleanup performance — disambiguation makes cleanup more complex)

HrushiYadav:
Closes #5438

## Problem

`_upsert_entity` and the Phase 7d batch path in `_add_to_vector_store` merge entities based solely on a 0.95 vector similarity threshold. Same-name entities with different roles get incorrectly merged, accumulating unrelated `linked_memory_ids` and degrading search quality.

## Scope

This PR adds a **cross-POS-category gate** -- it prevents merging when the syntactic `entity_type` values differ (e.g. PROPER vs NOUN, QUOTED vs COMPOUND). The four possible values come from `entity_extraction.py`: `PROPER`, `QUOTED`, `COMPOUND`, `NOUN`.

**What this catches:** a NOUN entity "python" (the snake) merging with a PROPER entity "Python" (a title/name). Different POS categories signal different usage contexts.

**What this does NOT catch:** same-POS polysemes like "Python" (the language) and "Python" (the snake), since both are classified as PROPER. Disambiguating those requires the `semantic_type` field from #5573, which is out of scope here.

## Fix

All four entity merge paths are now protected:
- `_upsert_entity` (sync)
- `_upsert_entity_async`
- `_add_to_vector_store` Phase 7d (sync)
- `_add_to_vector_store` Phase 7d (async)

Logic at each site:
- If both entities have `entity_type` set **and they differ**, skip the merge and insert a new entity
- If either side has no `entity_type`, merge is allowed (backward compat)
- When merging, backfill `entity_type` on the existing entity if it was missing

## Tests

6 tests in `TestEntityTypeMergeGate`:
- `test_same_type_merges` -- same entity_type, merge proceeds
- `test_different_type_creates_new` -- different entity_type, new entity created
- `test_no_type_on_existing_still_merges` -- existing lacks type, merge + backfill
- `test_no_type_on_new_still_merges` -- new entity lacks type, merge allowed
- `test_phase7d_different_type_inserts_instead_of_update` -- Phase 7d batch path with type mismatch
- `test_async_upsert_entity_different_type_creates_new` -- async path coverage

## Composition with #5573

PR #5573 changes `extract_entities` to return 3-tuples with `semantic_type`. If it merges first, this PR will need a rebase to update the unpacking loops. The natural resolution is combining both gates (`entity_type` + `semantic_type`) into a single predicate.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -595,20 +595,26 @@ class Memory(MemoryBase):
 
             semantic_match = existing[0] if existing and existing[0].score >= 0.95 else None
             match = exact_match or semantic_match
+            matched = False
             if match:
-                # Update existing entity's linked_memory_ids
                 payload = match.payload or {}
-                linked_ids = payload.get("linked_memory_ids", [])
-                if memory_id not in linked_ids:
-                    linked_ids.append(memory_id)
-                    payload["linked_memory_ids"] = linked_ids
-                    self.entity_store.update(
-                        vector_id=match.id,
-                        vector=None,
-                        payload=payload,
-                    )
-            else:
-                # Create new entity
+                if entity_type and payload.get("entity_type") and payload["entity_type"] != entity_type:
+                    pass
+                else:
+                    matched = True
+                    linked_ids = payload.get("linked_memory_ids", [])
+                    if memory_id not in linked_ids:
+                        linked_ids.append(memory_id)
+                        payload["linked_memory_ids"] = linked_ids
+                        if entity_type and not payload.get("entity_type"):
+                            payload["entity_type"] = entity_type
+                        self.entity_store.update(
+                            vector_id=match.id,
+                            vector=None,
+                            payload=payload,
+                        )
+
+            if not matched:
                 entity_id = str(uuid.uuid4())
                 entity_payload = {
                     "data": entity_text,
@@ -2237,19 +2243,27 @@ class AsyncMemory(MemoryBase):
 
             semantic_match = existing[0] if existing and existing[0].score >= 0.95 else None
             match = exact_match or semantic_match
+            matched = False
             if match:
                 payload = match.payload or {}
-                linked_ids = payload.get("linked_memory_ids", [])
-                if memory_id not in linked_ids:
-                    linked_ids.append(memory_id)
-                    payload["linked_memory_ids"] = linked_ids
-                    await asyncio.to_thread(
-                        self.entity_store.update,
-                        vector_id=match.id,
-                        vector=None,
-                        payload=payload,
-                    )
-            else:
+                if entity_type and payload.get("entity_type") and payload["entity_type"] != entity_type:
+                    pass
+                else:
+                    matched = True
+                    linked_ids = payload.get("linked_memory_ids", [])
+                    if memory_id not in linked_ids:
+                        linked_ids.append(memory_id)
+                        payload["linked_memory_ids"] = linked_ids
+                        if entity_type and not payload.get("entity_type"):
+                            payload["entity_type"] = entity_type
+                        await asyncio.to_thread(
+                            self.entity_store.update,
+                            vector_id=match.id,
+                            vector=None,
+                            payload=payload,
+                        )
+
+            if not matched:
                 entity_id = str(uuid.uuid4())
                 entity_payload = {
                     "data": entity_text,
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1132,7 +1132,7 @@ class Memory(MemoryBase):
                             payload = match.payload or {}
                             existing_type = payload.get("entity_type")
                             if entity_type and existing_type and existing_type != entity_type:
-                                matched = False
+                                pass
                             else:
                                 matched = True
                                 linked = set(payload.get("linked_memory_ids", []))
@@ -2780,7 +2780,7 @@ class AsyncMemory(MemoryBase):
                             payload = match.payload or {}
                             existing_type = payload.get("entity_type")
                             if entity_type and existing_type and existing_type != entity_type:
-                                matched = False
+                                pass
                             else:
                                 matched = True
                                 linked = set(payload.get("linked_memory_ids", []))
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -599,7 +599,14 @@ class Memory(MemoryBase):
             if match:
                 payload = match.payload or {}
                 if entity_type and payload.get("entity_type") and payload["entity_type"] != entity_type:
-                    pass
+                    logger.warning(
+                        "Skipping merge for entity %r: stored entity_type %r != new "
+                        "entity_type %r. A duplicate entity will be created instead of "
+                        "merging (this can happen after an entity_type vocabulary change).",
+                        entity_text,
+                        payload.get("entity_type"),
+                        entity_type,
+                    )
                 else:
                     matched = True
                     linked_ids = payload.get("linked_memory_ids", [])
@@ -1132,7 +1139,14 @@ class Memory(MemoryBase):
                             payload = match.payload or {}
                             existing_type = payload.get("entity_type")
                             if entity_type and existing_type and existing_type != entity_type:
-                                pass
+                                logger.warning(
+                                    "Skipping merge for entity %r: stored entity_type %r != new "
+                                    "entity_type %r. A duplicate entity will be created instead of "
+                                    "merging (this can happen after an entity_type vocabulary change).",
+                                    entity_text,
+                                    existing_type,
+                                    entity_type,
+                                )
                             else:
                                 matched = True
                                 linked = set(payload.get("linked_memory_ids", []))
@@ -2255,7 +2269,14 @@ class AsyncMemory(MemoryBase):
             if match:
                 payload = match.payload or {}
                 if entity_type and payload.get("entity_type") and payload["entity_type"] != entity_type:
-                    pass
+                    logger.warning(
+                        "Skipping merge for entity %r: stored entity_type %r != new "
+                        "entity_type %r. A duplicate entity will be created instead of "
+                        "merging (this can happen after an entity_type vocabulary change).",
+                        entity_text,
+                        payload.get("entity_type"),
+                        entity_type,
+                    )
                 else:
                     matched = True
                     linked_ids = payload.get("linked_memory_ids", [])
@@ -2780,7 +2801,14 @@ class AsyncMemory(MemoryBase):
                             payload = match.payload or {}
                             existing_type = payload.get("entity_type")
                             if entity_type and existing_type and existing_type != entity_type:
-                                pass
+                                logger.warning(
+                                    "Skipping merge for entity %r: stored entity_type %r != new "
+                                    "entity_type %r. A duplicate entity will be created instead of "
+                                    "merging (this can happen after an entity_type vocabulary change).",
+                                    entity_text,
+                                    existing_type,
+                                    entity_type,
+                                )
                             else:
                                 matched = True
                                 linked = set(payload.get("linked_memory_ids", []))
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1127,21 +1127,29 @@ class Memory(MemoryBase):
 
                         semantic_match = matches[0] if matches and matches[0].score >= 0.95 else None
                         match = exact_match or semantic_match
+                        matched = False
                         if match:
-                            # Update existing entity
                             payload = match.payload or {}
-                            linked = set(payload.get("linked_memory_ids", []))
-                            linked |= memory_ids
-                            payload["linked_memory_ids"] = sorted(linked)
-                            try:
-                                self.entity_store.update(
-                                    vector_id=match.id,
-                                    vector=None,
-                                    payload=payload,
-                                )
-                            except Exception as e:
-                                logger.debug(f"Entity update failed for '{entity_text}': {e}")
-                        else:
+                            existing_type = payload.get("entity_type")
+                            if entity_type and existing_type and existing_type != entity_type:
+                                matched = False
+                            else:
+                                matched = True
+                                linked = set(payload.get("linked_memory_ids", []))
+                                linked |= memory_ids
+                                payload["linked_memory_ids"] = sorted(linked)
+                                if not existing_type and entity_type:
+                                    payload["entity_type"] = entity_type
+                                try:
+                                    self.entity_store.update(
+                                        vector_id=match.id,
+                                        vector=None,
+                                        payload=payload,
+                                    )
+                                except Exception as e:
+                                    logger.debug(f"Entity update failed for '{entity_text}': {e}")
+
+                        if not matched:
                             # New entity — collect for batch insert
                             to_insert_vectors.append(valid_vectors[j])
                             to_insert_ids.append(str(uuid.uuid4()))
@@ -2767,21 +2775,30 @@ class AsyncMemory(MemoryBase):
 
                         semantic_match = matches[0] if matches and matches[0].score >= 0.95 else None
                         match = exact_match or semantic_match
+                        matched = False
                         if match:
                             payload = match.payload or {}
-                            linked = set(payload.get("linked_memory_ids", []))
-                            linked |= memory_ids
-                            payload["linked_memory_ids"] = sorted(linked)
-                            try:
-                                await asyncio.to_thread(
-                                    self.entity_store.update,
-                                    vector_id=match.id,
-                                    vector=None,
-                                    payload=payload,
-                                )
-                            except Exception as e:
-                                logger.debug(f"Entity update failed for '{entity_text}' (async): {e}")
-                        else:
+                            existing_type = payload.get("entity_type")
+                            if entity_type and existing_type and existing_type != entity_type:
+                                matched = False
+                            else:
+                                matched = True
+                                linked = set(payload.get("linked_memory_ids", []))
+                                linked |= memory_ids
+                                payload["linked_memory_ids"] = sorted(linked)
+                                if not existing_type and entity_type:
+                                    payload["entity_type"] = entity_type
+                                try:
+                                    await asyncio.to_thread(
+                                        self.entity_store.update,
+                                        vector_id=match.id,
+                                        vector=None,
+                                        payload=payload,
+                                    )
+                                except Exception as e:
+                                    logger.debug(f"Entity update failed for '{entity_text}' (async): {e}")
+
+                        if not matched:
                             to_insert_vectors.append(valid_vectors[j])
                             to_insert_ids.append(str(uuid.uuid4()))
                             to_insert_payloads.append({
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1157,21 +1157,29 @@ class Memory(MemoryBase):
 
                         semantic_match = matches[0] if matches and matches[0].score >= 0.95 else None
                         match = exact_match or semantic_match
+                        matched = False
                         if match:
-                            # Update existing entity
                             payload = match.payload or {}
-                            linked = set(payload.get("linked_memory_ids", []))
-                            linked |= memory_ids
-                            payload["linked_memory_ids"] = sorted(linked)
-                            try:
-                                self.entity_store.update(
-                                    vector_id=match.id,
-                                    vector=None,
-                                    payload=payload,
-                                )
-                            except Exception as e:
-                                logger.debug(f"Entity update failed for '{entity_text}': {e}")
-                        else:
+                            existing_type = payload.get("entity_type")
+                            if entity_type and existing_type and existing_type != entity_type:
+                                matched = False
+                            else:
+                                matched = True
+                                linked = set(payload.get("linked_memory_ids", []))
+                                linked |= memory_ids
+                                payload["linked_memory_ids"] = sorted(linked)
+                                if not existing_type and entity_type:
+                                    payload["entity_type"] = entity_type
+                                try:
+                                    self.entity_store.update(
+                                        vector_id=match.id,
+                                        vector=None,
+                                        payload=payload,
+                                    )
+                                except Exception as e:
+                                    logger.debug(f"Entity update failed for '{entity_text}': {e}")
+
+                        if not matched:
                             # New entity — collect for batch insert
                             to_insert_vectors.append(valid_vectors[j])
                             to_insert_ids.append(str(uuid.uuid4()))
@@ -2823,21 +2831,30 @@ class AsyncMemory(MemoryBase):
 
                         semantic_match = matches[0] if matches and matches[0].score >= 0.95 else None
                         match = exact_match or semantic_match
+                        matched = False
                         if match:
                             payload = match.payload or {}
-                            linked = set(payload.get("linked_memory_ids", []))
-                            linked |= memory_ids
-                            payload["linked_memory_ids"] = sorted(linked)
-                            try:
-                                await asyncio.to_thread(
-                                    self.entity_store.update,
-                                    vector_id=match.id,
-                                    vector=None,
-                                    payload=payload,
-                                )
-                            except Exception as e:
-                                logger.debug(f"Entity update failed for '{entity_text}' (async): {e}")
-                        else:
+                            existing_type = payload.get("entity_type")
+                            if entity_type and existing_type and existing_type != entity_type:
+                                matched = False
+                            else:
+                                matched = True
+                                linked = set(payload.get("linked_memory_ids", []))
+                                linked |= memory_ids
+                                payload["linked_memory_ids"] = sorted(linked)
+                                if not existing_type and entity_type:
+                                    payload["entity_type"] = entity_type
+                                try:
+                                    await asyncio.to_thread(
+                                        self.entity_store.update,
+                                        vector_id=match.id,
+                                        vector=None,
+                                        payload=payload,
+                                    )
+                                except Exception as e:
+                                    logger.debug(f"Entity update failed for '{entity_text}' (async): {e}")
+
+                        if not matched:
                             to_insert_vectors.append(valid_vectors[j])
                             to_insert_ids.append(str(uuid.uuid4()))
                             to_insert_payloads.append({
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Good work on this — the gate logic is sound for the cross-type case, the sync/async parity is exact, backward compat is well-handled, and the four tests are genuine (red-green confirmed locally). A couple of things need addressing before this lands.

---

## Strengths

- The three-way logic (block on mismatch, allow when either side lacks type, backfill on merge) covers the meaningful cases cleanly
- Sync and async implementations are identical in structure — no drift
- `test_different_type_creates_new` and `test_no_type_on_existing_still_merges` genuinely bite on revert (verified)
- Ruff clean, follows repo conventions

---

## Issues

### The fix misses the primary `add()` path

`_upsert_entity` is called from `_link_entities_for_memory` — the per-entity update path. The primary path for `memory.add()` goes through `_add_to_vector_store` Phase 7d (`mem0/memory/main.py` lines ~1017–1031), which calls `entity_store.search_batch` and still merges on `score >= 0.95` alone. That block has no entity_type check:

```python
if matches and matches[0].score >= 0.95:
    match = matches[0]
    payload = match.payload or {}
    linked = set(payload.get("linked_memory_ids", []))
    linked |= memory_ids
    payload["linked_memory_ids"] = sorted(linked)
    self.entity_store.update(...)
```

The bug from #5438 is still reproducible via `memory.add()` even after this PR merges. Applying the same type gate inside that Phase 7d loop (and its async counterpart) would close the gap.

### `entity_type` stores syntactic categories, not semantic ones

After checking `mem0/utils/entity_extraction.py` line 134: the four possible values are `PROPER`, `QUOTED`, `COMPOUND`, `NOUN` — structural categories derived from POS tags. Both "Python" (the language) and "Python" (the snake) would be classified as `PROPER`. The gate condition `payload["entity_type"] != entity_type` evaluates to `"PROPER" != "PROPER"` → False, so the merge still happens for the main motivating examples in #5438 ("Python", "Apple", "Tesla"). The fix is genuinely useful for preventing, say, a `PROPER` entity from merging with a `NOUN` entity of the same text — but that's a narrower protection than the description implies. Worth noting this in the PR so reviewers and users have accurate expectations. (PR #5573's `semantic_type` field would be the right discriminator for same-syntactic-class polysemes.)

### Composition with PR #5441

#5441 (open, scope gate) modifies `_upsert_entity` at exactly the same lines as this PR — both introduce a `merge`/`matched` flag at identical locations in both the sync and async functions. They'll conflict. The composed gate would be `scope_ok AND type_ok`; coordinating merge order or combining into a shared `_should_merge` predicate would keep both fixes intact.

---

## Requested changes

1. **Apply the entity_type gate to Phase 7d** (`_add_to_vector_store` lines ~1017–1031, sync and async) so `memory.add()` is protected
2. **Scope the PR description** to reflect that this catches cross-syntactic-type conflation — same-category polysemes ("Python" language vs. snake) require `semantic_type` (#5573) and are out of scope here
3. Add a test that exercises Phase 7d with a type-mismatched high-score hit (mock `entity_store.search_batch` returning `entity_type="PROPER"` on an entity being inserted with `entity_type="NOUN"`, assert insert not update)

---

## Minor

- `if entity_type and payload.get("entity_type") and payload["entity_type"] != entity_type: pass` reads more clearly inverted — `if not type_conflict: matched = True` (or extract to a named bool)
- `m._entity_store.insert.call_args[1]["payloads"][0]` (test line ~1298) — prefer `call_args.kwargs["payloads"][0]` to match the rest of the file
- `_upsert_entity_async` has no dedicated test — low risk given structural parity, but a single smoke test for the different-type case would cover it

---

## Evidence

- **CI:** Python SDK 3.10/3.11/3.12 pass; CI Gate pass; Vercel advisory (ignored)
- **Red-green:** `test_different_type_creates_new` and `test_no_type_on_existing_still_merges` fail with main code, pass with PR code — tests bite
- **Batch path:** confirmed by reading main.py lines 1017–1031 — no type check present
- **entity_type values:** confirmed from `mem0/utils/entity_extraction.py` line 134 — `PROPER`, `QUOTED`, `COMPOUND`, `NOUN`
- TypeScript OSS `_linkEntitiesForMemory` (`mem0-ts/src/oss/src/memory/index.ts` lines ~416–430) has the same merge-on-similarity-alone pattern — not blocking this PR but worth a follow-up issue

Would be great to have a linked issue if one doesn't already exist for the batch path gap!

HrushiYadav:
Pushed fixes for all 3 requested changes.

**1. Phase 7d batch path now protected** -- added entity_type gate to the `_add_to_vector_store` Phase 7d loop in both sync (~line 1023) and async (~line 2537). When `entity_store.search_batch` returns a high-score match with a different `entity_type`, the entity is inserted as new instead of merged. Also backfills `entity_type` on merge when the existing entity lacks it.

**2. PR scope clarification** -- this fix catches cross-syntactic-type conflation (e.g. PROPER vs NOUN). Same-category polysemes ("Python" the language vs. the snake) would both be classified as PROPER by `entity_extraction.py` and still merge. Semantic disambiguation requires the `semantic_type` field from #5573, which is out of scope here.

**3. Phase 7d test added** -- `test_phase7d_different_type_inserts_instead_of_update` exercises the full `_add_to_vector_store` path with mocked LLM + entity extraction, where `search_batch` returns a PROPER entity but the extracted entity is NOUN. Asserts insert (not update) with correct type.

**Re: PR #5441 conflict** -- noted. Happy to coordinate merge order or combine into a shared `_should_merge` predicate if needed.

All 5 entity merge gate tests pass, ruff clean.

kartik-mem0:
Thanks for the follow-up — the Phase 7d gap is now closed and the new test proves it. Here's what the re-review found.

**Verified fixed:**
- The entity_type gate now covers all four paths: `_upsert_entity` sync/async and `_add_to_vector_store` Phase-7d sync/async. Red-green confirmed this run: `test_phase7d_different_type_inserts_instead_of_update` fails on main code (update called when it should not be) and passes on the PR head.
- Backfill on merge (`payload["entity_type"] = entity_type` when existing record lacks the field) is in all four paths — clean progressive-enrichment behavior.

**Still not addressed — syntactic-vs-semantic gap**

The prior comment flagged that `entity_type` holds POS-category labels (`PROPER`, `NOUN`, `COMPOUND`, `QUOTED`), not semantic labels. This means same-POS polysemes still conflate. To confirm: re-checked `entity_extraction.py` this run — both "Python" (the language) and "Python" (the snake) produce `entity_type = "PROPER"` because the PROPER branch fires on capitalized proper-noun sequences regardless of meaning. The gate condition `"PROPER" != "PROPER"` → False → merge still proceeds for the primary examples in #5438.

The new Phase-7d test uses `("NOUN", "python")` as the incoming entity and a stored record with `entity_type: "PROPER"` — so it correctly tests and proves the cross-category gate. But it does not exercise the PROPER vs PROPER case that the issue describes. That gap needs to be reflected in the PR description (and ideally a test that documents the current known limitation).

**New concern — composition with PR #5573**

PR #5573 (spaCy semantic_type) changes `extract_entities()` and `extract_entities_batch()` to return 3-tuples `(entity_type, entity_text, semantic_type)` and rewrites all the unpacking loops at main.py lines 617, 973, 2163, 2512 to `for entity_type, entity_text, semantic_type in entities:`. This PR leaves those lines as 2-tuple unpacking. If #5573 merges on top of this PR, those loops raise `ValueError: too many values to unpack`. Both PRs also edit `_upsert_entity` and the Phase-7d block at the same lines — they'll conflict on rebase.

`semantic_type` from #5573 is the correct discriminator for polysemes ("Tesla" → PERSON vs ORG; "Python" → PRODUCT vs no NER hit). The natural resolution is to coordinate: once #5573 lands, rebase this PR on top and replace the `entity_type` gate with a `semantic_type` gate, which would fix the root cause rather than the narrower POS-category case.

**Remaining requests:**

1. Update the PR description (and ideally a test comment) to scope this accurately: the current gate prevents cross-POS-category conflation (e.g., a NOUN entity from merging with a PROPER one of the same text), not same-POS-class polyseme conflation. The latter requires #5573's `semantic_type`.
2. Add a `@pytest.mark.asyncio` test for `_upsert_entity_async` covering the different-type case — the async path is separately maintained code and has no dedicated test.
3. If #5573 is landing soon: coordinate the merge order, or combine this gate with #5573's `semantic_type` gate into a single PR that addresses both concerns at once.

**Minor (non-blocking):**
- Phase-7d: `matched = False` inside the type-mismatch branch is redundant (variable is already `False` at initialization). Use `pass` to match the `_upsert_entity` style.
- Unit tests use `call_args[1]` tuple access in some places, `call_args.kwargs` in others — prefer `.kwargs` throughout.

---

**Evidence this run:**
- Red-green: 5/5 pass with PR code; 3/5 fail with main code (`test_different_type_creates_new`, `test_no_type_on_existing_still_merges`, `test_phase7d_different_type_inserts_instead_of_update`) — all three gate-exercising tests bite
- CI: Python SDK 3.10/3.11/3.12 pass; CI Gate pass; Vercel advisory (ignored)
- Lint: ruff clean on both changed files
- entity_extraction.py PROPER classification: confirmed by code trace this run (lines 186–220)
- PR #5573 3-tuple change: confirmed by reading its diff this run

HrushiYadav:
Pushed fixes for all requested changes.

**1. Phase 7d `matched = False` -> `pass`** -- both sync and async now use `pass` in the type-mismatch branch, matching `_upsert_entity` style.

**2. Test `.kwargs` consistency** -- replaced `call_args[1]` with `call_args.kwargs` in `test_different_type_creates_new` and `test_no_type_on_existing_still_merges`.

**3. Async test added** -- `test_async_upsert_entity_different_type_creates_new` exercises `_upsert_entity_async` with PROPER vs NOUN mismatch, asserts insert (not update).

**4. PR description updated** -- scoped to cross-POS-category conflation. Same-POS polysemes ("Python" language vs snake, both PROPER) require `semantic_type` from #5573.

**Re: #5573 composition** -- noted. Happy to rebase once it lands and combine both gates.

All 6 tests pass, ruff clean.

kartik-mem0:
Checked the new push (`bcde3493`). Here is the re-review finding.

**Verified fixed:**
- Async gap (I2): `test_async_upsert_entity_different_type_creates_new` now present and red-green proven (fails on `origin/main`, passes on PR).
- `matched = False` consistency: both Phase-7d paths now use `pass` in the type-mismatch branch.
- `.kwargs` assertion access: consistent across all 5 assertion sites.
- 6/6 `TestEntityTypeMergeGate` tests pass. CI: Python 3.10/3.11/3.12 — PASS, CI Gate — PASS.

**Still open (code, not documentation):**

**C1 — Semantic discriminator absent.** The gate at `_upsert_entity` and all three Phase-7d sites still fires only on cross-POS-category differences (`PROPER`, `NOUN`, `QUOTED`, `COMPOUND`). Same-POS polysemes — the primary case in #5438's examples ("Apple" company vs fruit, "Python" language vs snake, both PROPER) — still merge. The PR body now documents this, which is honest, but the code fix is not here. The needed discriminator is `semantic_type` from #5573 (spaCy NER labels: `PERSON`, `ORG`, `PRODUCT`, etc.).

**C2 — 2-tuple unpacking unchanged.** Lines 617, 979, 1616, 2185, 2534, 3165 in `main.py` still unpack `for entity_type, entity_text in entities:`. PR #5573 (LGTM'd, open, CONFLICTING) changes `extract_entities()` to return 3-tuples `(entity_type, entity_text, semantic_type)` and rewrites all six of these sites. If #5573 merges before this PR, these lines raise `ValueError: too many values to unpack`. "Coordinate with #5573" means the code at these sites needs to be either (a) rebased after #5573 lands, or (b) updated here to be forward-compatible with the 3-tuple return using `*_` for the unused field.

**Passing to Kartik.** This PR has had two re-reviews today (daily cap). The forward path is clear: either rebase this PR on top of #5573 and replace the 2-tuple unpackings with 3-tuple ones (adding `semantic_type` as the second gate condition at the same time — resolving both C1 and C2 in one pass), or land this PR first and rebase #5573 on top of it (accepting that C1 remains open until #5573 merges). That sequencing call and the semantic gate design are better made with a maintainer in the loop.

HrushiYadav:
Hi @kartik-mem0, following up on your last review. Since #5573 is still open and has conflicts, I think the cleanest path is to land this PR first (the entity_type gate is independently valuable), then rebase #5573 on top. The semantic discriminator (C1) and 3-tuple unpacking (C2) are better addressed in #5573 since that's where semantic_type is introduced. Happy to coordinate either way -- what's your preference?

mem0/memory/main.py
```diff
@@ -583,20 +583,26 @@ def _upsert_entity(self, entity_text, entity_type, memory_id, filters):
...
-                    linked_ids.append(memory_id)
-                    payload["linked_memory_ids"] = linked_ids
-                    self.entity_store.update(
-                        vector_id=match.id,
-                        vector=None,
-                        payload=payload,
-                    )
-            else:
-                # Create new entity
+                if entity_type and payload.get("entity_type") and payload["entity_type"] != entity_type:
```
kartik-mem0, line 626: An entity_type mismatch here (and at 1122, 2247, 2772) is permanent: nothing normalizes or versions entity_type, so an entity created before a vocabulary change (already happened once in this repo, b33fa542) can never merge again and silently duplicates instead. Reproduced in the sandbox this cycle: oss-qdrant and oss-pgvector, sync and async, all 4 confirm the silent duplicate.

Log or count a skipped merge on type mismatch so drift is detectable, or normalize entity_type before comparing rather than raw string equality.

mem0/memory/main.py
```diff
@@ -2743,21 +2765,30 @@ async def _add_to_vector_store(
...
-                                    self.entity_store.update,
-                                    vector_id=match.id,
-                                    vector=None,
-                                    payload=payload,
-                                )
-                            except Exception as e:
-                                logger.debug(f"Entity update failed for '{entity_text}' (async): {e}")
-                        else:
+                            existing_type = payload.get("entity_type")
+                            if entity_type and existing_type and existing_type != entity_type:
```
kartik-mem0, line 2859: This async batch path has no dedicated pytest test, only sync Phase 7d and the two single-entity paths are unit-tested. Covered empirically by a new sandbox scenario this cycle (MODE=async, both lanes, PASS), but that only runs in review, not in CI.

Add a mocked unit test mirroring test_phase7d_different_type_inserts_instead_of_update for this async path.

kartik-mem0 (commented):
## Changes requested

Adds a syntactic entity_type gate to the four entity-merge sites, correctly narrower than the linked issue and honestly scoped in the PR body, but the gate's own basis already drifted once on main and will silently start duplicating instead of deduping once it drifts again.

**To unblock:**
1. `mem0/memory/main.py:589` (same pattern at 1122, 2247, 2772): an entity_type mismatch permanently blocks merge with no way to detect it happened (detail inline)
2. `mem0-ts/src/oss/src/memory/index.ts:499,1183`: TS OSS still merges on score alone, no entity_type gate, no tracked follow-up found

Optional: async Phase 7d path (main.py:2772) has real sandbox coverage this cycle but no dedicated pytest test (detail inline)

- **Problem is real:** pre-PR `main.py` merges purely on `score >= 0.95` at all 4 sites, confirmed in the diff and by reproducing on `main` in the sandbox
- **Tests / red-green:** `hatch -e dev_py_3_12 run pytest tests/test_memory.py -k TestEntityTypeMergeGate -v` -> 6 passed on PR head. Reverted `mem0/memory/main.py` to `main` in the worktree -> 4 of 6 fail (test_different_type_creates_new, test_no_type_on_existing_still_merges, test_phase7d_different_type_inserts_instead_of_update, test_async_upsert_entity_different_type_creates_new). Restored PR source -> 6 passed again. Full `tests/test_memory.py`: 65 passed. Ruff clean.
- **Sandbox (required, mem0/memory/ touched):** new scenario `pr5497_entity_type_mismatch_silent_duplicate`, oss-qdrant sync/async + oss-pgvector sync/async, all 4 PASS on the PR worktree (4d41e95d), and correctly FAIL on unmodified main first. Read off the last printed line, never `$?`. No regression on 3 unrelated entity-store scenarios (identical pre-existing failures on main and the PR worktree).
- **entity_type source:** traced to a deterministic, thread-locked spaCy pipeline in `mem0/utils/entity_extraction.py`, not an LLM call, so run-to-run nondeterminism is not the concern. Cross-version drift is: this repo already rewrote the entity_type vocabulary once (`b33fa542`, NOUN/COMPOUND -> TOPIC/IDENTIFIER) and open PR #5573 proposes a second rewrite.
- **Security:** cleared, every `payload["entity_type"]` bracket access is guarded by a preceding `.get()` truthy check, no KeyError risk on old payloads.
- **Bloat (ponytail):** lean already, no new abstractions or files, net 0.
- **Not verified:** production collision rate for the drift scenario, whether open PR #5441 still textually conflicts with this diff at the same lines.

</details>

The fix itself is careful and the PR body is unusually honest about its own scope limits (C1/C2 from the earlier rounds are disclosed, not hidden). The duplication side effect above is the one thing worth closing before merge. Push when ready and I will re-review.

HrushiYadav:
Thanks for the thorough review. Pushed changes addressing all three:

**1. Silent duplication on entity_type drift (blocking)** — the four gate sites (`main.py` ~601/1134/2257/2782) now emit a `logger.warning` when a high-score match is skipped purely due to an `entity_type` mismatch, naming the entity text and both types. So the "duplicate instead of merge" case (e.g. after a vocabulary change) is now observable rather than silent.

I kept the fix to *detection* rather than full drift *prevention*: actually normalizing/versioning `entity_type` so old and new vocabularies dedupe correctly depends on the vocabulary itself, and there's an open rewrite in #5573. I think that normalization belongs with #5573 (where the vocab is defined) rather than this gate — happy to align once #5573 lands. This PR's gate + warning is safe in the meantime (a duplicate is strictly better than a wrong merge, and now it's logged).

**2. TS OSS parity** — opened #6529 to track porting the gate to `mem0-ts/.../index.ts` (499/1183). Kept it out of this PR to avoid mixing languages, same as the #5695 → #5812 approach.

**3. Async Phase 7d test** — added `test_async_phase7d_different_type_inserts_instead_of_update`, a mocked unit test mirroring the sync Phase 7d test against `AsyncMemory._add_to_vector_store`.

All 7 `TestEntityTypeMergeGate` tests pass, full `tests/test_memory.py` green (66), ruff clean. Rebased on latest main. Ready for re-review.

mem0/memory/main.py
kartik-mem0: Verified against 07f2437c: fixed. All four sites now log a warning when a type mismatch skips the merge (main.py:602,1142,2272,2804). Confirmed live in the sandbox this run, oss-qdrant and oss-pgvector, sync and async: the log line fires (`Skipping merge for entity 'Kubernetes': stored entity_type 'STALE_PROPER' != new entity_type 'PROPER'...`) and the scenario still PASSes on all 4. Drift is now observable, not silent.

mem0/memory/main.py
kartik-mem0: Verified against 07f2437c: fixed. test_async_phase7d_different_type_inserts_instead_of_update now mirrors the sync test for the async Phase 7d path. Red-green checked this run: reverted main.py to pre-PR main, the new test failed (AssertionError: expected update not called, called 1 times); restored to 07f2437c, it passes along with the rest of the suite (66 passed). The async path is now covered, not just the sync one.

kartik-mem0 (commented):

Re-review that flipped to LGTM. Verified against `07f2437c` (new head): both blocking findings from the prior review are fixed. Vocabulary-drift silent duplication is now logged instead of silent, and the async Phase 7d gate has test coverage matching the sync path.

**Before you merge:** none.
**Follow-up:** #6529 tracks porting this same entity_type merge gate to the TS OSS side. This Python change now defines the contract TS must match: skip merge only when both sides have a non-empty entity_type that differs, otherwise merge and backfill the missing type. No Python/TS inconsistency exists yet since TS hasn't shipped its version; #6529 is the place to hold it to this contract.
**Not verified:** TS OSS parity itself, out of scope for this PR.

Risk: low, backward-compatible gate plus a warning log for drift, full sandbox matrix green · Your time: ~5 min

- **Problem is real:** reproduced on `main` (`d6d89c98`) via `sandbox/scenarios/pr5497_entity_type_mismatch_silent_duplicate.py`, FAIL before the gate, PASS after.
- **Tests:** targeted `pytest tests/test_memory.py -k EntityTypeMergeGate` = 7 passed, including the new `test_async_phase7d_different_type_inserts_instead_of_update`.
- **Red-green:** reverted `main.py` to pre-PR `main`, the new async test FAILED (`AssertionError`, update called once); restored to `07f2437c`, PASSED. Full suite: 66 passed, ruff clean.
- **CI:** all required checks green (Python 3.10/3.11/3.12, CI Gate, changelog_check, check_changes, license/cla). Only the advisory Vercel preview fails, non-blocking.
- **Live sandbox:** full lane×mode matrix on this head, oss-qdrant×sync, oss-qdrant×async, oss-pgvector×sync, oss-pgvector×async, all 4 PASS, provenance confirmed (`repo=~/.cache/mem0-worktrees/review-pr-5497 @ 07f2437c`). The new warning log fires live in every run.
- **Security:** no injection/tenancy surface touched, purely an internal merge-decision gate.
- **Bloat:** delta since last review is 4 log lines plus 1 mirrored test, no new abstractions.
- **Linked issue:** #5587 (dup of #5438) resolved by the merge gate; #6529 filed by the contributor for the TS OSS follow-up.
</details>

HrushiYadav:
@kartik-mem0 this is green and mergeable, and you marked it ready for squash-merge on 24 Jul. Could we land it before main drifts? Happy to rebase if needed.

HrushiYadav:
Gentle ping — this was marked ready for squash-merge on 24 Jul (@kartik-mem0's LGTM). Would appreciate landing it before main drifts further. Happy to rebase if needed.

HrushiYadav:
Rebased on latest main. @kartik-mem0 — could you take another look when you get a chance? Happy to address any remaining concerns.

kartik-mem0 (commented):
Re-reviewed at `994e47f9`. Content-identical rebase of the previously reviewed head. Fresh tests and lint pass against the new SHA. LGTM still holds.

HrushiYadav:
Gentle ping — this has been marked LGTM and ready for squash-merge since Jul 24, re-verified by @kartik-mem0 on Sep 21. Would appreciate a maintainer merging when you get a chance. Happy to rebase again if main has moved.

PR #5498, opened 2026-06-11
fix: sanitize entity collection name for S3 Vectors (no underscores)

### LINKED ISSUE #5361: Warning output to log when calling Memory.add() using s3 vectors backend

HrushiYadav:
Closes #5361

## Problem

S3 Vectors (AWS) does not allow underscores in index names. The entity store collection name is constructed as `f"{collection_name}_entities"` which always contains an underscore. When using S3 Vectors as the backend, this causes:

```
Batch entity linking failed: An error occurred (ValidationException) when calling the GetIndex operation: Invalid index name
```

## Fix

When the vector store provider is `s3_vectors`, replace all underscores with dashes in the entity collection name. For example, `my_memories_entities` becomes `my-memories-entities`.

Applied in both sync and async `entity_store` property initializers.

## Tests

1 new test `test_s3_vectors_entity_collection_uses_dashes` — verifies the entity collection name has no underscores when provider is s3_vectors.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -418,6 +418,8 @@ class Memory(MemoryBase):
         if self._entity_store is None:
             entity_config = _safe_deepcopy_config(self.config.vector_store.config)
             entity_collection = f"{self.collection_name}_entities"
+            if self.config.vector_store.provider == "s3_vectors":
+                entity_collection = entity_collection.replace("_", "-")
             # Set collection name on the cloned config
             if hasattr(entity_config, 'collection_name'):
                 entity_config.collection_name = entity_collection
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for jumping on this, @HrushiYadav — appreciate you taking the time. 🙏

Issue #5361 drew a few parallel fixes, and we're consolidating on #5416, which takes the dash-separator approach with unit + integration tests for the S3 entity-store path. To avoid duplicate work I'm closing this one in its favor — it's nothing about the quality of your patch, just keeping a single source of truth for the fix.

As always, thanks for your continued contributions to mem0 — keep them coming!

PR #5383, opened 2026-06-05
fix(openclaw): skip runtime setup during metadata registration

he-yufeng:
Fixes #5371.

## Summary

- treat OpenClaw `registrationMode: "cli-metadata"` as a CLI-only registration pass
- skip provider/backend creation, tool and hook registration, service registration, memory capability registration, and the `openclaw.plugin.registered` telemetry/log side effects during that pass
- keep the normal full registration path unchanged
- add a regression test for the metadata registration path

## Why

OpenClaw can call a plugin's `register()` twice: once for the real gateway runtime and once for CLI metadata discovery. The mem0 plugin was doing the full runtime setup on both passes, so `openclaw doctor` could show duplicate `openclaw-mem0: registered` lines and telemetry could count the same plugin startup twice.

The metadata pass only needs the CLI command surface. Registering CLI commands with null backend/provider dependencies matches the existing setup flow and avoids runtime side effects.

## Validation

- `pnpm install`
- `pnpm exec vitest run index.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm run build`
- `git diff --check`

PR #5499, opened 2026-06-11
fix: guard openclaw register() against duplicate cli-metadata invocation

### LINKED ISSUE #5371: openclaw-mem0: register() invoked twice per gateway startup without mode guard (full + cli-metadata passes)
## Description

`openclaw-mem0: registered` info log line is emitted twice per gateway startup with identical parameters and the same timestamp. The plugin's `register()` callback is being invoked twice per gateway lifecycle.

## Reproduction

```bash
$ openclaw doctor 2>&1 | grep "openclaw-mem0: registered"
16:24:22 [plugins] openclaw-mem0: registered (mode: platform, user: root, autoRecall: true, autoCapture: true, skills: true)
16:24:22 [plugins] openclaw-mem0: registered (mode: platform, user: root, autoRecall: true, autoCapture: true, skills: true)
```

## Root cause

OpenClaw's plugin loader (`dist/loader-DlRepQoO.js`) calls `runPluginRegisterSync(register, api)` twice per plugin load:

1. **Line ~2146** — in the main registration pass, with `registrationMode: "full"`
2. **Line ~2544** — in a second pass with `registrationMode: "cli-metadata"` (used to collect metadata for `openclaw mem0 help` and similar CLI commands)

The mem0 plugin's `register(api)` function (in `dist/index.js:4208`) does not check `api.registrationMode`, so it fires the `registered` log line both times. It also runs the full side effects twice:

- Creates a second `provider` instance
- Creates a second `PlatformBackend` instance
- Calls `api.registerService(...)` and `api.registerMemoryCapability(...)` a second time
- Fires the `openclaw.plugin.registered` telemetry event twice

The core loader's own pre-`register` snapshot/rollback logic (lines 2147–2157) is already designed to undo side effects on non-`full` passes — but mem0's `register()` does work that escapes that snapshot (provider/backend creation, telemetry).

## Impact

- **Functional:** the actual memory service is connected once (the "full" pass wins). No user-visible bug.
- **Telemetry:** the `openclaw.plugin.registered` event is double-counted.
- **Resources:** one extra provider + backend allocated per gateway restart, briefly.
- **Log noise:** duplicate `registered` line per restart, which is annoying in `openclaw doctor` and log watchers.

## Suggested fix

Guard `register(api)` against the non-`full` passes. Both passes still need the CLI surface, so only skip the heavy side effects on the metadata pass:

```diff
   register(api) {
+    // Core calls register() in two passes: "full" and "cli-metadata".
+    // On the metadata pass, register CLI commands only — skip the
+    // provider/backend/service/capability setup to avoid double work.
     bootstrapTelemetryFlag();
     const pluginAuth = readPluginAuth();
     const fileConfig = {
       apiKey: pluginAuth.apiKey,
       baseUrl: pluginAuth.baseUrl
     };
     const cfg = mem0ConfigSchema.parse(api.pluginConfig, fileConfig);
+    if (api.registrationMode === "cli-metadata") {
+      registerCliCommands(api, null, null, cfg, ...);
+      return;
+    }
     const telemetryCtx = { ... };
     ...
     api.logger.info(`openclaw-mem0: registered (...)`);
     api.registerService(...);
     api.registerMemoryCapability(...);
   }
```

## Environment

- OpenClaw: 2026.6.1 (2e08f0f)
- @mem0/openclaw-mem0: 1.0.11
- mode: platform
- Node: v22.22.2

HrushiYadav:
Closes #5371

## Problem

OpenClaw's plugin loader calls `register()` twice per plugin startup:
1. `registrationMode: "full"` — full initialization
2. `registrationMode: "cli-metadata"` — only needs CLI command metadata

The mem0 plugin's `register()` doesn't check the mode, so every side effect runs twice: provider/backend creation, service registration, tool registration, hook registration, and telemetry events.

## Fix

Added an early-return guard at the top of `register()`: when `api.registrationMode === "cli-metadata"`, only register CLI commands (which is all the metadata pass needs) and return immediately — skipping provider init, backend creation, tools, hooks, service registration, and telemetry.

This matches the pattern already used in the `needsSetup` early-return path (which also only registers CLI).

### DIFF
--- a/openclaw/index.ts
+++ b/openclaw/index.ts
@@ -108,6 +108,26 @@ const memoryPlugin = definePluginEntry({
     };
     const cfg = mem0ConfigSchema.parse(api.pluginConfig, fileConfig);
 
+    // OpenClaw's plugin loader calls register() twice: once with
+    // registrationMode "full" and once with "cli-metadata". Only
+    // CLI commands need to be available in the metadata pass —
+    // skip all heavy side effects (provider init, hooks, tools,
+    // service registration, telemetry).
+    if ((api as any).registrationMode === "cli-metadata") {
+      registerCliCommands(
+        api,
+        null as any,
+        null as any,
+        cfg,
+        () => cfg.userId,
+        (id: string) => `${cfg.userId}:agent:${id}`,
+        () => ({ user_id: cfg.userId, top_k: cfg.topK }),
+        () => undefined,
+        () => {},
+      );
+      return;
+    }
+
     // Telemetry context bound to this plugin instance's config
     const telemetryCtx = {
       apiKey: cfg.apiKey,


kartik-mem0:
Thanks for tackling this, @HrushiYadav! 🙏

#5371 drew a few parallel fixes, and we merged #5383 (which guards the `cli-metadata` registration pass against the duplicate runtime setup) to resolve it. I'm closing this one as a duplicate to keep things consolidated — really appreciate you jumping on it, and as always, thanks for your continued contributions to mem0!

Issue #2820, opened 2025-05-28
🚀 Feature Request: Include OpenAI Token Usage in All Relevant Method Responses


karimmila:
## Description

While reviewing the documentation and previous issues, I couldn't find a way to programmatically access OpenAI token usage data from the `AsyncMemory` client methods such as `add()`, `search()`, and `get_all()`. According to OpenAI’s API documentation, responses can contain useful usage metrics like:

```json
"usage": {
  "prompt_tokens": 5,
  "completion_tokens": 7,
  "total_tokens": 12
}
```

It would be highly beneficial if this `usage` information were included in the response payloads of all relevant `AsyncMemory` methods that internally call an OpenAI LLM, such as:

- `add(messages: List[Dict[str, str]], user_id: str)`
- `search(query: str, user_id: str, limit: int = 5)`
- `get_all(user_id: str, limit: int = 100)`
- Any other method that relies on OpenAI completions or embeddings

For example:

```python
result = await memory_client.add(messages=messages, user_id="user-123")
result["status"]  # => "success"
result["usage"]   # => {"prompt_tokens": X, "completion_tokens": Y, "total_tokens": Z}

result = await memory_client.search(query="who is Naval?", user_id="user-123")
result["results"]  # => [ ... ]
result["usage"]    # => token usage info

result = await memory_client.get_all(user_id="user-123")
result["memories"] # => [ ... ]
result["usage"]    # => usage info, if applicable
```

An optional config flag could allow the full raw OpenAI API response to be returned, or just the usage data extracted and standardized across methods.

## Motivation / Use Case

I'm building a web application where each user has their own long-term memory, and I handle all OpenAI API usage centrally. I don't want to require users to provide their own API key, but I do need to **track OpenAI token usage per user** in order to:

- Monitor consumption
- Enforce usage limits
- Provide cost transparency
- Optimize backend performance

Having access to `usage` data across all relevant `AsyncMemory` methods would enable precise control over LLM usage, aligned with billing and resource tracking.

### Motivation, pitch

I'm working on a web app, where each user can have his own sources db. I don't want to force someone to put/use his own OpenAI api key, instead I want to control user tokens usage programmatically.

Nauman-S:
+1 facing this issue as well

lhtvineettiwari:
+1

maarioos123:
+1

NeutrinoLiu:
+1

maarioos123:
+1

[...]

PR #5501, opened 2026-06-12
feat(memory): expose opt-in OpenAI usage for OSS add

rodboev:
## Linked Issue

Partially addresses #2820

## Description

Python OSS memory calls can invoke OpenAI and receive token usage in the raw API response, but the SDK currently returns only parsed memory content. This change adds non-breaking, opt-in usage reporting for OpenAI-backed `Memory.add()` and `AsyncMemory.add()` calls, plus the OSS server add route.

`OpenAILLM` keeps the raw usage payload from OpenAI responses. The OSS memory add paths copy accumulated usage into the top-level result only when `include_usage=True`. Calls that do not use OpenAI, do not produce usage, or do not opt in keep the existing response shape.

## Scope

This is limited to OSS add flows that invoke OpenAI during `add()`. Usage from multiple OpenAI calls is accumulated per add operation. Search, get-all, hosted `MemoryClient`, embedding, reranker, and non-OpenAI provider usage remain outside this PR.

## Type of Change

- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] I tested manually
- [ ] No tests needed

Verification:

- [x] Focused LLM and memory tests: 80 passed
- [x] `ruff check` passed for all six changed files
- [x] `git diff --check` passed
- [ ] Server test execution, blocked during collection because this local environment lacks the optional `telemetry` module
- [ ] `ruff format --check`, current-main formatting differences remain in the six-file diff
- [ ] `make lint` / `make test`

## Checklist

### DIFF
--- a/mem0/llms/openai.py
+++ b/mem0/llms/openai.py
@@ -30,9 +30,9 @@ class OpenAILLM(LLMBase):
                 top_k=config.top_k,
                 enable_vision=config.enable_vision,
                 vision_details=config.vision_details,
-                reasoning_effort=getattr(config, 'reasoning_effort', None),
+                reasoning_effort=getattr(config, "reasoning_effort", None),
                 http_client_proxies=config.http_client_proxies,
-                is_reasoning_model=getattr(config, 'is_reasoning_model', None),
+                is_reasoning_model=getattr(config, "is_reasoning_model", None),
             )
 
         super().__init__(config)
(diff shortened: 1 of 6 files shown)


OfficialAbhinavSingh:
Nice, clean scoping on this — opt-in `add()` usage without touching the default response shape is the right first slice.

Saw your note that `search()`/`get_all()`/async and non-OpenAI providers are intentionally left for later so this PR doesn't have to settle the full response-contract surface up front. If it'd help, I'd be glad to pick up that remainder once this lands — extending the same opt-in `include_usage` shape to `search()`/`get_all()` and the async client, following whatever pattern gets agreed here.

(Checked the issue thread too — Parteeksachdeva raised good questions on `get_all()` usage semantics back in April but nothing ended up landing, so figured it's worth flagging interest rather than it going quiet again.) Totally fine if you're already planning to extend it yourself — just didn't want the rest of #2820 to stall waiting on a volunteer. Good luck with review!

PR #5515, opened 2026-06-13
fix(ts-sdk): preserve user metadata keys across the case-conversion round-trip

### LINKED ISSUE #5055: [TypeScript SDK] v3 silently camelizes user-defined `metadata` keys on read, breaking round-trip

ac12644:
## Linked Issue

Closes #5055

## Description

`snakeToCamelKeys` / `camelToSnakeKeys` (`mem0-ts/src/client/utils.ts`) recurse into every nested object, including the user-controlled `metadata` blob. So a snake_case key a user stores (e.g. `message_id`) comes back camelized (`messageId`) — a lossy, asymmetric round-trip and a behavior regression vs the v2 SDK, which didn't transform responses at all.

This treats `metadata` as an opaque pass-through in **both** directions (an `OPAQUE_VALUE_KEYS` set), so user keys survive verbatim at any nesting depth.

Supersedes #5065 (cc @PHclaw — thanks for the original diagnosis). That PR fixed only the inbound `snakeToCamelKeys` path; this also covers the outbound `camelToSnakeKeys` path (so `add({ metadata: { messageId: ... } })` isn't rewritten on write) and adds the regression tests requested in review.

## Type of Change

## Breaking Changes

N/A — restores the v2 round-trip behavior for `metadata`.

## Test Coverage

New tests in `mem0-ts/src/client/utils.test.ts` cover: snake_case keys preserved on read, camelCase keys preserved on write, lossless round-trip of arbitrary metadata keys, and metadata nested inside a results array. SDK-defined keys and logical operators (OR/AND/NOT) still convert as before.

Verified locally: `jest src/client/utils.test.ts` (6 passing), `pnpm run build`, and `prettier --check` all pass.

## Checklist

- [ ] I have updated documentation if needed (N/A — internal behavior fix)

### DIFF
--- a/mem0-ts/src/client/tests/utils.test.ts
+++ b/mem0-ts/src/client/tests/utils.test.ts
@@ -76,4 +76,27 @@ describe("camelToSnakeKeys / snakeToCamelKeys", () => {
       });
     });
   });
+
+  describe("user-controlled structuredDataSchema blob (issue #5055)", () => {
+    it("converts the outer key but leaves user field names on write", () => {
+      expect(
+        camelToSnakeKeys({
+          structuredDataSchema: { firstName: "string", lastName: "string" },
+        }),
+      ).toEqual({
+        // outer SDK key is snake_cased, user-defined field names are not
+        structured_data_schema: { firstName: "string", lastName: "string" },
+      });
+    });
+
+    it("converts the outer key but leaves user field names on read", () => {
+      expect(
+        snakeToCamelKeys({
+          structured_data_schema: { first_name: "string", last_name: "string" },
+        }),
+      ).toEqual({
+        structuredDataSchema: { first_name: "string", last_name: "string" },
+      });
+    });
+  });
 });
--- a/mem0-ts/src/client/tests/utils.test.ts
+++ b/mem0-ts/src/client/tests/utils.test.ts
@@ -1,4 +1,4 @@
-import { camelToSnakeKeys, snakeToCamelKeys } from "./utils";
+import { camelToSnakeKeys, snakeToCamelKeys } from "../utils";
 
 describe("camelToSnakeKeys / snakeToCamelKeys", () => {
   it("converts SDK-defined keys between camelCase and snake_case", () => {
--- a/mem0-ts/src/client/utils.test.ts
+++ b/mem0-ts/src/client/utils.test.ts
@@ -1,0 +1,79 @@
+import { camelToSnakeKeys, snakeToCamelKeys } from "./utils";
+
+describe("camelToSnakeKeys / snakeToCamelKeys", () => {
+  it("converts SDK-defined keys between camelCase and snake_case", () => {
+    expect(camelToSnakeKeys({ userId: "u1", agentId: "a1" })).toEqual({
+      user_id: "u1",
+      agent_id: "a1",
+    });
+    expect(snakeToCamelKeys({ user_id: "u1", agent_id: "a1" })).toEqual({
+      userId: "u1",
+      agentId: "a1",
+    });
+  });
+
+  it("preserves logical operator keys (OR/AND/NOT)", () => {
+    expect(camelToSnakeKeys({ OR: [{ userId: "u1" }] })).toEqual({
+      OR: [{ user_id: "u1" }],
+    });
+  });
+
+  describe("user-controlled metadata blob (issue #5055)", () => {
+    it("does not camelize snake_case keys inside metadata on read", () => {
+      const apiResponse = {
+        id: "mem-1",
+        user_id: "u1",
+        metadata: { message_id: "x", some_custom_key: "y" },
+      };
+
+      expect(snakeToCamelKeys(apiResponse)).toEqual({
+        id: "mem-1",
+        userId: "u1",
+        metadata: { message_id: "x", some_custom_key: "y" },
+      });
+    });
+
+    it("does not snake_case camelCase keys inside metadata on write", () => {
+      const payload = {
+        userId: "u1",
+        metadata: { messageId: "x", someCustomKey: "y" },
+      };
+
+      expect(camelToSnakeKeys(payload)).toEqual({
+        user_id: "u1",
+        metadata: { messageId: "x", someCustomKey: "y" },
+      });
+    });
+
+    it("round-trips arbitrary metadata keys losslessly", () => {
+      const metadata = {
+        message_id: "abc",
+        camelKey: 1,
+        nested: { deep_snake: true, deepCamel: false },
+        arr: [{ inner_key: 1 }],
+      };
+
+      const roundTripped = snakeToCamelKeys(
+        camelToSnakeKeys({ userId: "u1", metadata }),
+      );
+
+      expect(roundTripped.metadata).toEqual(metadata);
+    });
+
+    it("preserves metadata nested inside an array of results", () => {
+      const apiResponse = {
+        results: [
+          { id: "1", metadata: { message_id: "x" } },
+          { id: "2", metadata: { another_key: "z" } },
+        ],
+      };
+
+      expect(snakeToCamelKeys(apiResponse)).toEqual({
+        results: [
+          { id: "1", metadata: { message_id: "x" } },
+          { id: "2", metadata: { another_key: "z" } },
+        ],
+      });
+    });
+  });
+});
--- a/mem0-ts/src/client/utils.ts
+++ b/mem0-ts/src/client/utils.ts
@@ -15,8 +15,16 @@ function snakeToCamel(str: string): string {
 }
 
 /**
+ * Keys whose values are user-controlled, opaque blobs. Their nested keys must
+ * be passed through verbatim — converting them would silently rewrite the
+ * user's own keys and break round-trips (see issue #5055).
+ */
+const OPAQUE_VALUE_KEYS = new Set(["metadata"]);
+
+/**
  * Recursively converts all keys of an object from camelCase to snake_case.
  * Used for converting user-facing camelCase params to API snake_case payloads.
+ * Values under {@link OPAQUE_VALUE_KEYS} (e.g. `metadata`) are left untouched.
  */
 export function camelToSnakeKeys(obj: any): any {
   if (obj === null || obj === undefined || typeof obj !== "object") return obj;
@@ -26,7 +34,7 @@ export function camelToSnakeKeys(obj: any): any {
   return Object.fromEntries(
     Object.entries(obj).map(([key, value]) => [
       camelToSnake(key),
-      camelToSnakeKeys(value),
+      OPAQUE_VALUE_KEYS.has(key) ? value : camelToSnakeKeys(value),
     ]),
   );
 }
@@ -34,6 +42,7 @@ export function camelToSnakeKeys(obj: any): any {
 /**
  * Recursively converts all keys of an object from snake_case to camelCase.
  * Used for converting API snake_case responses to user-facing camelCase.
+ * Values under {@link OPAQUE_VALUE_KEYS} (e.g. `metadata`) are left untouched.
  */
 export function snakeToCamelKeys(obj: any): any {
   if (obj === null || obj === undefined || typeof obj !== "object") return obj;
@@ -43,7 +52,7 @@ export function snakeToCamelKeys(obj: any): any {
   return Object.fromEntries(
     Object.entries(obj).map(([key, value]) => [
       snakeToCamel(key),
-      snakeToCamelKeys(value),
+      OPAQUE_VALUE_KEYS.has(key) ? value : snakeToCamelKeys(value),
     ]),
   );
 }
--- a/mem0-ts/src/client/utils.ts
+++ b/mem0-ts/src/client/utils.ts
@@ -18,8 +18,18 @@ function snakeToCamel(str: string): string {
  * Keys whose values are user-controlled, opaque blobs. Their nested keys must
  * be passed through verbatim — converting them would silently rewrite the
  * user's own keys and break round-trips (see issue #5055).
+ *
+ * The check runs against the source key, so a multi-word key must be listed in
+ * both casings to be covered in both directions: the camelCase form for the
+ * outbound `camelToSnakeKeys` path and the snake_case form for the inbound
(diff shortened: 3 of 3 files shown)


kartik-mem0:
The fix is correct and the tests genuinely bite — ran red-green this session. Two asks before merge, one blocking.

**Fix quality**

The root cause is properly addressed in both directions. `OPAQUE_VALUE_KEYS.has(key) ? value : <recurse>` fires on the parent key before deciding whether to descend, so the metadata blob is left verbatim on both the write path (`camelToSnakeKeys`) and the read path (`snakeToCamelKeys`). `metadata` nested inside an array of results (e.g. `results[i].metadata`) is also correctly handled — the existing `Array.isArray` map branch recurses into each element, and the guard fires when it sees the `metadata` key there. Top-level SDK keys are unaffected.

**Red-green, this run, worktree dd52b573e:**

With the fix: 6/6 pass. Without (OPAQUE_VALUE_KEYS removed, unconditional recursion restored): the 4 metadata-specific tests all fail, the 2 baseline tests pass. The tests catch exactly what they're supposed to catch.

**One blocking ask — test file location**

Every other client unit test in this package lives under `src/client/tests/` (`memoryClient.crud.test.ts`, `telemetry-aliasing.test.ts`, etc.). The new file lands at `src/client/utils.test.ts` — a sibling of the source file rather than the established tests/ subdirectory. The tests run either way (jest's testMatch is broad enough), but the inconsistency will make it harder for contributors to find tests later. Please move it to `src/client/tests/utils.test.ts`.

**One non-blocking note — `schema` in `createMemoryExport`**

The same class of bug exists on a narrower surface: in `createMemoryExport`, `schema: Record<string, any>` stays in `rest` when `filters` is destructured, so `camelToSnakeKeys(rest)` is applied to it. User-defined keys inside the schema object would be mangled. Worth a follow-up, but out of scope here — not asking you to fix it in this PR.

Move the test file and this is ready. `gh pr merge --squash` when done.

ac12644:
Moved to `src/client/tests/utils.test.ts` and fixed the import to `../utils`. 6/6 still pass locally. Ready when you are.

Good catch on `createMemoryExport` — the `schema` blob is the same class. I'll open a follow-up for it.

kartik-mem0:
## Changes Requested

### What this PR does
Adds `OPAQUE_VALUE_KEYS = new Set(["metadata"])` to `utils.ts` and threads a short-circuit check into both `camelToSnakeKeys` and `snakeToCamelKeys` so user-controlled keys inside `metadata` blobs pass through verbatim in both directions. Supersedes #5065, which only covered the read path.

### Strengths
- `utils.ts:17-23` — the named `Set` constant at module scope is the right design: one place to enumerate opaque fields, O(1) lookup, trivially extensible, well-documented with a JSDoc pointing to the linked issue.
- `utils.ts:34, 52` — the fix is surgical: one ternary per function, zero change to surrounding logic. The key is checked pre-conversion in both directions; `"metadata"` is spelled identically in camelCase and snake_case, so one entry covers both.
- `utils.test.ts` — four tests that reproduce the exact failure modes from #5055: read path (snake_case preserved), write path (camelCase preserved), full round-trip losslessness, and metadata nested inside a results array. Red-green verified locally (4 fail without the fix, 6 pass with it).

### Issues

#### Important (should fix before merging)

1. `mem0-ts/src/client/mem0.types.ts:16` — `structuredDataSchema?: Record<string, any>` is in `AddMemoryOptions`. When a user passes it to `add()`, it flows through `_preparePayload` → `camelToSnakeKeys` (mem0.ts:206). User-defined schema field names (e.g. `{ firstName: "string" }`) would be silently converted to `{ first_name: "string" }` — the same class of lossy round-trip that this PR fixes for `metadata`. Since `OPAQUE_VALUE_KEYS` is now the designated home for this pattern, adding `"structuredDataSchema"` (or equivalently `"structured_data_schema"` for the snake_case path) here is a one-line change that closes the gap cleanly. This was pre-existing before this PR; raising it here because this PR introduces the exact mechanism to fix it.

2. `TypeScript SDK / changelog_check` CI is failing, but it is a stale artifact — the PR's own commits do not touch `package.json`. The base commit was at v3.0.7; a sibling PR landed v3.0.8 on main after this branch was opened, so the merge result triggers the changelog gate even though this PR didn't introduce the bump. A rebase onto current main will clear it (the 3.0.8 changelog entry is already present in `docs/changelog/sdk.mdx`).

#### Minor (optional)

- `utils.ts:17` — a note on the constant that keys must be spelled identically in both camelCase and snake_case forms (which `metadata` happens to satisfy) would prevent a subtle mistake if someone adds a multi-word opaque key in the future.

### Evidence
- **CI:** build (Node 20 + 22), integration tests — all pass. `changelog_check` failure is a stale merge artifact (see item 2 above), not a code quality signal.
- **Targeted tests:** `pnpm exec jest src/client/tests/utils.test.ts --no-coverage` → 6 passed in 0.205 s
- **Red-green proof:** reverted `utils.ts` to pre-PR state → 4/6 fail (read-path, write-path, round-trip, array-results tests), all exactly on the bug's failure modes. Restored fix → 6/6 pass.
- **Linked issue:** #5055 — PR addresses the root cause in both conversion directions.
- **Cross-package check:** Python `MemoryClient` does no key-case transformation layer; bug is TS-specific, no Python action needed.
- **#5594 overlap check:** that merged PR fixed `createMemoryExport` schema keys in `mem0.ts` at the call-site level. Main's `utils.ts` still has the unconditional recursion — confirmed by reading the file on main. This PR is still needed.
- **Could NOT verify:** live API behavior with a real `structuredDataSchema` payload (no API key in this environment); the analysis above is code-tracing only.

### Assessment
Ready to merge: **With fixes** — please address the two Important items above; I'll re-review once you push. The core approach is correct and well-tested; these are small additions.

ac12644:
Addressed both — thanks for the thorough pass.

**1. structuredDataSchema** — added to `OPAQUE_VALUE_KEYS`. Since the check runs on the source key and it isn't casing-invariant, I added **both** forms: `structuredDataSchema` (outbound `camelToSnakeKeys`) and `structured_data_schema` (inbound `snakeToCamelKeys`), so user field names round-trip verbatim in both directions while the outer key still converts. Tests cover both paths.

**Minor** — documented the both-casing requirement for multi-word opaque keys in the constant's JSDoc.

**2. changelog_check** — rebased onto current main; the v3.0.8 entry is already there, so the gate clears (and CI Gate with it).

8/8 tests pass, build + prettier clean locally. Ready for re-review.

kartik-mem0:
Both addressed. Verified this session.

**structuredDataSchema (prior blocker)** —  now carries all three entries: `metadata`, `structuredDataSchema`, `structured_data_schema`. The dual-casing is correct: the outbound path sees the camelCase form, the inbound path sees the snake_case form; the outer SDK key still converts while the value is passed through verbatim. The JSDoc explaining this is accurate and the comment is clear enough for future contributors to follow. New tests (commits `9fec7c4`, `7a5d19e`) cover both the write and read paths for this field.

**changelog_check (prior blocker)** — passes cleanly at head `7a5d19e`. CI Gate green across Node 20 and 22 (build, integration).

**Red-green run this session:**
- Without fix: 6 of 8 tests fail — the 4 metadata tests and both structuredDataSchema tests fail with the exact key-renaming the issue reported.
- With fix: 8 of 8 pass.

Nothing critical or important left. Ready to squash-merge. If someone with merge access wants to take it: `gh pr merge 5515 --squash -R mem0ai/mem0`.

kartik-mem0 (approved):

PR #5516, opened 2026-06-13
fix(llms): send max_completion_tokens for gpt-5.x OpenAI models

### LINKED ISSUE #5054: bug: _get_common_params passes max_tokens to gpt-5.4-mini, which requires max_completion_tokens
## Follow-up to #4738

#4738 reported that `_is_reasoning_model()` incorrectly classified `gpt-5.4-mini` as a reasoning model, stripping `temperature`. PR #4746 fixed that by narrowing the reasoning-model detection.

However, `gpt-5.4-mini` now falls through to `_get_common_params()`, which unconditionally passes `max_tokens`:

```python
# mem0/llms/base.py:132-136
params = {
    "temperature": self.config.temperature,
    "max_tokens": self.config.max_tokens,  # <-- rejected by gpt-5.4-mini
    "top_p": self.config.top_p,
}
```

The OpenAI API rejects `max_tokens` for `gpt-5.4-mini` and requires `max_completion_tokens` instead.

## Error

```
BadRequestError: Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}
```

Stacktrace ends in `openai/_base_client.py` → `self._make_status_error_from_response`.

## Impact

All `client.add()` (extraction) and `client.search()` calls fail for users running `gpt-5.4-mini` as the LLM provider. Memory extraction silently fails — no memories are stored.

## Suggested fix

`_get_common_params()` (or the OpenAI-specific `generate_response()`) should use `max_completion_tokens` instead of `max_tokens` for models that require it. This likely applies to all gpt-5.x variants, not just `gpt-5.4-mini`.

One approach: check the model name and swap the parameter key, similar to how `_is_reasoning_model` already gates parameter selection.

## Environment

- mem0ai v2.0.1
- openai SDK (latest)
- Model: `gpt-5.4-mini`
- Python 3.14.2

eldar702:
## Summary
Closes #5054. OpenAI's gpt-5.x chat models (e.g. `gpt-5.4-mini`) reject the legacy `max_tokens` parameter and require `max_completion_tokens`. Because these models are correctly classified as **non**-reasoning, the parameter wasn't being renamed, so requests fail with an HTTP 400.

## Fix
Add `_requires_max_completion_tokens(model)` (matches `gpt-5`, `gpt-5-*`, `gpt-5.*` after stripping any provider prefix) and rename `max_tokens` → `max_completion_tokens` in `generate_response` only for those models. gpt-4.x and reasoning-model paths are unchanged.

## Tests
RED-first: a gpt-5.4-mini request now sends `max_completion_tokens` (was `max_tokens`); a gpt-4.1 companion test guards against over-reach. `tests/llms/test_openai.py`: 20 passed.

(Returning contributor — prior merged PR #5443.)

*AI-assisted contribution.*

### DIFF
--- a/mem0/llms/openai.py
+++ b/mem0/llms/openai.py
@@ -82,6 +82,28 @@ class OpenAILLM(LLMBase):
         else:
             return response.choices[0].message.content
 
+    @staticmethod
+    def _requires_max_completion_tokens(model: str) -> bool:
+        """Whether the OpenAI Chat Completions API requires ``max_completion_tokens``.
+
+        The gpt-5 family rejects the legacy ``max_tokens`` parameter with a 400
+        ("Unsupported parameter: 'max_tokens' ... Use 'max_completion_tokens'
+        instead."). These models are NOT classified as reasoning models (so they
+        still accept ``temperature``/``top_p`` and never hit the param-stripping
+        path), which is exactly why the token cap reaches the request and must be
+        renamed here. See https://github.com/mem0ai/mem0/issues/5054
+
+        Args:
+            model: The configured model name (provider prefixes are tolerated).
+
+        Returns:
+            bool: True if ``max_tokens`` must be sent as ``max_completion_tokens``.
+        """
+        if not model:
+            return False
+        base_model = model.lower().rsplit("/", 1)[-1]
+        return base_model == "gpt-5" or base_model.startswith(("gpt-5-", "gpt-5."))
+
     def generate_response(
         self,
         messages: List[Dict[str, str]],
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Hey @eldar702 — thanks for digging into this and putting together a clean fix with solid tests!

Just flagging that this is now superseded: PR #5547 merged today and tackled the same root cause, but one layer higher — it added `_uses_max_completion_tokens()` + the rename logic directly to `LLMBase._get_common_params` in `mem0/llms/base.py`. Because `OpenAILLM` inherits that method, OpenAI is already covered without any changes in `openai.py`. The base-level fix also picks up Azure OpenAI, DeepSeek, vLLM, LM Studio, and the two hand-built providers (`litellm`, `azure_openai_structured`) in one go, and it closed issue #5054.

Given all that, this PR is now redundant — the behaviour it introduces is already live on `main`. I'd suggest closing it, but of course that's entirely your call. Really appreciate you chasing this down; the investigation and test cases here were helpful context for the broader fix!

eldar702:
Thanks @kartik-mem0 — that makes sense, and putting it in `LLMBase._get_common_params` so every inheriting provider benefits is the better layer. Closing this as superseded by #5547. Appreciate the quick look! (AI-assisted contribution.)

PR #5523, opened 2026-06-13
fix(server): resolve Qdrant list tuple error and prevent config leak on provider switch

bahung123:
## Description

This PR fixes two major issues when running the Mem0 API Server with `Qdrant` (and potentially other vector stores):

1. **Tuple Parsing Error in `_list_all_memories`**: Vector stores like Qdrant return a tuple `(List[Record], Optional[PointId])` from their `.list()` method. The existing `isinstance(results, list)` check fails, causing the API to return an empty or corrupted response which breaks the dashboard. Updated the check to `isinstance(results, (list, tuple))` to correctly extract records.
2. **Config Leak on Provider Switch**: When calling `POST /configure` to switch the provider (e.g., from `pgvector` to `qdrant` or `llm` from `openai` to `anthropic`), `_merge_config` performs a deep merge that leaks the old provider's config (like Postgres `dbname` or `password`) into the new provider's config, resulting in severe `ValidationError` (e.g., `Extra fields not allowed: dbname, user, password`). Updated `update_config` to explicitly clear the sub-config dictionary whenever a provider change is detected.
3. **Generalized `VECTOR_STORE_CONFIG`**: Refactored the hardcoded pgvector fallback in `server/main.py` to accept a flexible `VECTOR_STORE_CONFIG` JSON environment variable. This allows running the dockerized `mem0-api-server` cleanly with *any* vector store without cluttering `main.py` with provider-specific ENVs.

## Type of change

- [x] Enhancement (improves existing features)

## Checklist

- [x] My code follows the style guidelines of this project (Ruff, PEP8)

### DIFF
--- a/server/.env.example
+++ b/server/.env.example
@@ -12,6 +12,10 @@ POSTGRES_USER=postgres
 POSTGRES_PASSWORD=
 POSTGRES_COLLECTION_NAME=memories
 
+# Alternative Vector Stores (e.g. Qdrant)
+# VECTOR_STORE_PROVIDER=qdrant
+# VECTOR_STORE_CONFIG={"host": "qdrant", "port": 6333}
+
 ADMIN_API_KEY=
 JWT_SECRET=
 # Local development only. Leave false in production.
--- a/server/README.md
+++ b/server/README.md
@@ -137,6 +137,15 @@ The dashboard sets the following response headers on every path (see `server/das
 
 Together these prevent iframe embedding, sniffing of mislabelled MIME types, and cross-origin referrer leaks. Harden further behind your own reverse proxy if needed.
 
+## Alternative Vector Stores
+
+By default, Mem0 uses PostgreSQL with pgvector for storage. To use an alternative vector store like Qdrant, set `VECTOR_STORE_PROVIDER` and provide the connection details in `VECTOR_STORE_CONFIG` as JSON in your `.env` file:
+
+```bash
+VECTOR_STORE_PROVIDER=qdrant
+VECTOR_STORE_CONFIG={"host": "qdrant", "port": 6333}
+```
+
 ## Migrating from `ankane/pgvector` to `pgvector/pgvector`
 
 The `ankane/pgvector` Docker image is archived and no longer maintained. This release
--- a/server/main.py
+++ b/server/main.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 import logging
 import os
 import time
@@ -115,18 +116,32 @@ HISTORY_DB_PATH = os.environ.get("HISTORY_DB_PATH", "/app/history/history.db")
 DEFAULT_LLM_MODEL = os.environ.get("MEM0_DEFAULT_LLM_MODEL", "gpt-4.1-nano-2025-04-14")
 DEFAULT_EMBEDDER_MODEL = os.environ.get("MEM0_DEFAULT_EMBEDDER_MODEL", "text-embedding-3-small")
 
+VECTOR_STORE_PROVIDER = os.environ.get("VECTOR_STORE_PROVIDER", "pgvector")
+VECTOR_STORE_CONFIG = os.environ.get("VECTOR_STORE_CONFIG")
+
+vector_store_config = {
+    "collection_name": POSTGRES_COLLECTION_NAME,
+}
+
+if VECTOR_STORE_CONFIG:
+    try:
+        vector_store_config.update(json.loads(VECTOR_STORE_CONFIG))
+    except Exception as e:
+        logging.error(f"Failed to parse VECTOR_STORE_CONFIG: {e}")
+elif VECTOR_STORE_PROVIDER == "pgvector":
+    vector_store_config.update({
+        "host": POSTGRES_HOST,
+        "port": int(POSTGRES_PORT),
+        "dbname": POSTGRES_DB,
+        "user": POSTGRES_USER,
+        "password": POSTGRES_PASSWORD,
+    })
+
 DEFAULT_CONFIG = {
     "version": "v1.1",
     "vector_store": {
-        "provider": "pgvector",
-        "config": {
-            "host": POSTGRES_HOST,
-            "port": int(POSTGRES_PORT),
-            "dbname": POSTGRES_DB,
-            "user": POSTGRES_USER,
-            "password": POSTGRES_PASSWORD,
-            "collection_name": POSTGRES_COLLECTION_NAME,
-        },
+        "provider": VECTOR_STORE_PROVIDER,
+        "config": vector_store_config,
     },
     "llm": {
         "provider": "openai",
@@ -388,7 +403,7 @@ def _serialize_memory(row: Any) -> Dict[str, Any]:
 
 def _list_all_memories(limit: int = ALL_MEMORIES_LIMIT) -> Dict[str, Any]:
     results = get_memory_instance().vector_store.list(top_k=limit)
-    rows = results[0] if results and isinstance(results, list) and isinstance(results[0], list) else results or []
+    rows = results[0] if results and isinstance(results, (list, tuple)) and isinstance(results[0], (list, tuple)) else results or []
     return {"results": [_serialize_memory(row) for row in rows]}
 
 
--- a/server/main.py
+++ b/server/main.py
@@ -125,6 +125,7 @@ vector_store_config: Dict[str, Any] = {
 
 if VECTOR_STORE_CONFIG:
     try:
+        # Note: Any keys in VECTOR_STORE_CONFIG (including 'collection_name') will override defaults
         vector_store_config.update(json.loads(VECTOR_STORE_CONFIG))
     except json.JSONDecodeError as e:
         logging.error(f"Failed to parse VECTOR_STORE_CONFIG (invalid JSON): {e}")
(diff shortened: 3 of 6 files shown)


bahung123:
Hi @kartik-mem0 — gentle ping 👋

This PR addresses two related server-side issues that I ran into while switching our deployment from pgvector to Qdrant:

**1. Qdrant tuple error in `_list_all_memories` (`server/main.py`)**
The previous `isinstance(results[0], list)` check returned `True` for Qdrant, which actually returns a `tuple` from `vector_store.list()`. This caused `_list_all_memories` to fall through to `results or []` and return raw Qdrant tuple objects, breaking serialization downstream. Fixed by widening the check to `isinstance(results[0], (list, tuple))`.

**2. Config leak / stale keys on provider switch (`server/server_state.py`)**
When calling `update_config({"vector_store": {"provider": "qdrant", ...}})` after a previous pgvector config, the deep merge carried over pgvector's `host`/`port`/`dbname`/`user`/`password` into the new provider's `config` dict. This leaked credentials across providers and, depending on the provider, broke instantiation. The fix clears `config` on the previous provider before merge, and uses `deepcopy` to avoid mutating live state during the clear-and-replace. Applied symmetrically to `llm` and `embedder` to prevent the same issue there.

**Bonus: env-var driven config (`server/main.py`)**
Added `VECTOR_STORE_PROVIDER` + `VECTOR_STORE_CONFIG` (JSON) env vars with backward-compat (`COLLECTION_NAME` falls back to `POSTGRES_COLLECTION_NAME`), plus a warning log when a non-pgvector provider is selected without `VECTOR_STORE_CONFIG`.

**Tests:** 405-line `tests/test_server_config.py` covers env-var parsing (including invalid JSON), the provider-switch clearing logic, the Qdrant tuple path, and the COLLECTION_NAME fallback.

**Status:**
- ✅ CLA signed
- ✅ Branch mergeable
- ⚠️ Vercel preview is failing on `Authorization required to deploy` — looks like a one-time authorization on the maintainer side, not a code issue. Let me know if you need me to do anything here.

Happy to address feedback, split into separate PRs, or rebase if there are conflicts. Thanks for the review!

kartik-mem0:
## Changes Requested

### Strengths
- The tuple fix in `server/main.py:402` is correct and minimal. `isinstance(results, (list, tuple))` properly handles Qdrant's `(List[Record], Optional[PointId])` return from `client.scroll()`. Red-green proof: reverting the fix causes `test_tuple_result_works` to fail with 2 results instead of 1 (the tuple and `None` both get serialized); restoring it passes cleanly.
- The provider-switch config clearing in `server/server_state.py` is well-reasoned. Working on `deepcopy` of both `_current_config` and overrides before clearing prevents live-state mutation. The three-component loop (`vector_store`, `llm`, `embedder`) is symmetric and the tests exercise each axis including the LLM switch and stored-override clearing.
- The `VECTOR_STORE_CONFIG` JSON env var is a genuine improvement over the hardcoded pgvector block — any vector store can now be configured at startup without code changes. The `COLLECTION_NAME` fallback to `POSTGRES_COLLECTION_NAME` is a clean backward-compat migration.
- 405 lines of tests covering all three bug vectors, including edge cases (None result, empty result, same-provider-no-clear, stored-override contamination). The `pytest.importorskip("fastapi", ...)` guard correctly prevents failures in the Python SDK CI where fastapi is not installed.

### Issues

#### Important (should fix)

**1. `server/routers/entities.py:30` — identical tuple bug NOT fixed**

```python
# entities.py:30 — unchanged by this PR
rows = results[0] if results and isinstance(results, list) and isinstance(results[0], list) else results or []
```

`_iter_payloads()` calls `get_memory_instance().vector_store.list(top_k=SCAN_LIMIT)` — the same call as `_list_all_memories`. With Qdrant, `client.scroll()` returns `(records, next_page_offset)`. The old `isinstance(results, list)` is `False`, so the fallback `results or []` returns the whole tuple. Iterating it yields the records list and then `None`; `getattr(None, "payload", None)` returns `{}`, so all entity buckets get empty payloads and `GET /entities` returns an empty list for all Qdrant users. Fix: apply the same one-liner used in `main.py:402`.

**2. `tests/test_server_config.py:12–15` — three unused imports will fail the ruff CI job**

```python
import json      # F401 — unused
import sys       # F401 — unused
from copy import deepcopy  # F401 — unused
```

`tests/**` is in the `python_sdk` path filter in `ci-gate.yml`, so this PR triggers the Python SDK CI, which runs `ruff check` (whole repo). These three will fail it. Auto-fixable with `ruff check --fix`.

**3. `server/.env.example` and `server/README.md` not updated for new env vars**

The PR introduces `VECTOR_STORE_PROVIDER` and `VECTOR_STORE_CONFIG` as the configuration knobs for non-pgvector deployments, but `server/.env.example` still documents only `POSTGRES_COLLECTION_NAME=memories` with no Qdrant example. A user following the Docker quick-start docs has no discoverable path to configuring Qdrant; they would only see the runtime warning after the server starts misconfigured. Please add a commented Qdrant example block to `.env.example` and a brief note in `server/README.md`.

#### Minor (optional)

- `server/main.py:133–140` — `VECTOR_STORE_CONFIG` JSON silently overrides `COLLECTION_NAME` if `collection_name` appears in the JSON blob. Worth a one-line comment noting the precedence.
- `server/server_state.py:116` — the comment "Work on copies to avoid mutating live state before `_merge_config`" implies full atomicity, but `_current_config = next_config` happens before `Memory.from_config()` returns; if `from_config` raises, `get_current_config()` returns the new provider's config while `get_memory_instance()` still holds the old instance. This is a pre-existing gap, not introduced here, but worth not overstating in the comment.

### Evidence
- **CI:** Vercel check failed (authorization required — advisory, fork perms, not merge-blocking); CLA signed. No CI Gate or Python SDK lint/test run recorded yet for this PR.
- **Targeted tests:** `TestListAllMemoriesTupleHandling` 4/4 passed; `TestUpdateConfigProviderSwitch` 5/5 passed. Run with `JWT_SECRET=test AUTH_DISABLED=true PYTHONPATH=server/` in worktree `~/.cache/mem0-worktrees/review-pr-5523`.
- **Red-green proof:** Reverted `isinstance(results, (list, tuple))` → `isinstance(results, list)` in `server/main.py:402`; `test_tuple_result_works` FAILED (2 results instead of 1). Restored fix → PASSED.
- **Lint:** `ruff check server/main.py server/server_state.py` clean. `tests/test_server_config.py` has 3 F401 unused imports (auto-fixable).
- **`entities.py` claim verified:** `server/routers/entities.py:30` still contains the old `isinstance(results, list)` pattern; `entities.py` does not appear in the PR diff.
- **Could NOT verify:** `TestVectorStoreConfigEnvVar`/`TestCollectionNameEnvVar` locally (require full server env with JWT_SECRET + fastapi); live Qdrant behavior (code tracing only).
- **TS counterpart:** Unaffected — this is server-only Python code with no TS equivalent.

### Assessment
Ready to merge: **With fixes** — please address the three Important items (the `entities.py` one-liner, the unused imports, and the `.env.example` note); I'll re-review once you push.

The two core fixes are correct and well-tested. The blocker is that `GET /entities` has the same Qdrant crash this PR sets out to fix, and the test file will cause the CI lint job to fail as written.

For tracking on our side, mind opening a quick issue describing these bugs and linking it here? Makes it easier to prioritize and reference.

bahung123:
All three issues addressed — entities.py tuple fix, unused imports cleaned, .env.example + README.md updated. Ready for re-review when you have time. Thanks!

kartik-mem0:
## LGTM

All three items from the previous round are resolved.

**`server/routers/entities.py:30`** — `isinstance(results, (list, tuple))` is now applied identically to the fix in `main.py`. Verified in the diff.

**`tests/test_server_config.py` unused imports** — `json`, `sys`, and `copy.deepcopy` are gone. `ruff check server/main.py server/server_state.py server/routers/entities.py tests/test_server_config.py` passes clean on the new head.

**`server/.env.example` and `server/README.md`** — both updated with the commented Qdrant example block and a brief "Alternative Vector Stores" section. Content confirmed.

The two prior-round Minor notes are also tidied: the `update_config` comment now correctly acknowledges the non-atomic state swap, and the `COLLECTION_NAME` precedence is documented inline.

**Evidence (HEAD 057563892953):**
- `ruff check` on all 4 changed files: clean.
- 9 targeted tests (4 × `TestListAllMemoriesTupleHandling` + 5 × `TestUpdateConfigProviderSwitch`): all passed in worktree `~/.cache/mem0-worktrees/review-pr-5523`.
- CI: Vercel advisory (fork auth fail, not merge-blocking); CLA signed.

Ready to squash-merge.

bahung123:
Hi @kartik-mem0 —

All relevant CI checks are passing now (Python SDK build on 3.10/3.11/3.12, CI Gate, changelog_check, check_changes, CLA). The only failing check is the Vercel preview, which is an `Authorization required to deploy` error on the maintainer side — not a code issue, so it needs a one-time authorization from your end.

This PR fixes two server-side issues (Qdrant tuple error in `_list_all_memories`, and config leak/stale keys on provider switch) plus adds env-var driven config. Would love a review whenever you have a moment — happy to split it up or address any feedback. Thanks!

anvme:
Hey @kartik-mem0 Will it be merged?

PR #5527, opened 2026-06-13
fix(bedrock): use dict literal instead of set in AI21 response parse default

HrushiYadav:
## Summary

- The fallback default for a missing `completions` key in the AI21 Bedrock response parser used `{"data", {"text": ""}}` (a Python **set literal**) instead of `{"data": {"text": ""}}` (a dict)
- When `completions` is absent from the response, this crashes with `TypeError: unhashable type: 'dict'` because you can't put a dict inside a set
- Fix: change the comma to a colon so it's a proper dict literal

## Test plan

- [x] `test_ai21_missing_completions_returns_empty`: verifies fallback returns empty string when `completions` key is absent
- [x] `test_ai21_normal_response`: verifies normal AI21 response parsing still works
- [x] All 41 existing bedrock tests pass

### DIFF
--- a/mem0/llms/aws_bedrock.py
+++ b/mem0/llms/aws_bedrock.py
@@ -409,7 +409,7 @@ class AWSBedrockLLM(LLMBase):
             elif self.provider == "cohere":
                 return response_json.get("generations", [{"text": ""}])[0].get("text", "")
             elif self.provider == "ai21":
-                return response_json.get("completions", [{"data", {"text": ""}}])[0].get("data", {}).get("text", "")
+                return response_json.get("completions", [{"data": {"text": ""}}])[0].get("data", {}).get("text", "")
             else:
                 # Generic parsing - try common response fields
                 for field in ["content", "text", "completion", "generation"]:
--- a/tests/llms/test_aws_bedrock.py
+++ b/tests/llms/test_aws_bedrock.py
@@ -424,3 +424,29 @@ class TestMiniMaxProvider:
             assert msg["role"] != "system"
         # user message must be present
         assert kwargs["messages"][0]["role"] == "user"
+
+
+# ---------------------------------------------------------------------------
+# _parse_response — legacy InvokeModel provider-specific parsing
+# ---------------------------------------------------------------------------
+
+class TestParseResponseLegacy:
+    def test_ai21_missing_completions_returns_empty(self, mock_boto3):
+        """When AI21 response lacks 'completions', the fallback default must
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Changes a single comma to a colon in the AI21 Bedrock response parser fallback default, fixing a latent crash: `{"data", {"text": ""}}` is a Python set literal that raises `TypeError: unhashable type: 'dict'` whenever `completions` is absent from the response. The fix `{"data": {"text": ""}}` makes it the intended dict, so `.get("data", {}).get("text", "")` resolves correctly to `""`.

### Strengths
- `mem0/llms/aws_bedrock.py:412` — surgical one-character fix, zero blast radius; no other provider fallback paths in the file share this defect (anthropic, cohere, mistral all use correct dict syntax)
- `tests/llms/test_aws_bedrock.py:431-453` — two well-targeted regression tests: `test_ai21_missing_completions_returns_empty` directly exercises the broken path, `test_ai21_normal_response` covers the happy path end-to-end
- Fixture usage (`mock_boto3`) and class placement are consistent with the rest of the test file

### Issues
Minor (optional):
- `tests/llms/test_aws_bedrock.py:436,442` — `import io` and `import json` are inlined inside the two new test method bodies rather than at the module top level alongside the other imports. No other test method in the file has inline imports; worth moving on a follow-up or in this PR if you have a minute, but not a blocker.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 green; CI Gate pass; CLA signed. Vercel = infra auth only, not a code check.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/llms/test_aws_bedrock.py -v --tb=short` → 41 passed in 3.65s
- **Red-green proof:** reverted `aws_bedrock.py` to main → both new tests fail with `WARNING: Could not parse response: unhashable type: 'dict'`; restored fix → 2 passed in 0.37s ✓
- **Lint:** `ruff check mem0/llms/aws_bedrock.py tests/llms/test_aws_bedrock.py` → All checks passed
- **Linked issue:** no linked issue — appropriate for a small targeted bug fix
- **Cross-package check:** no TS Bedrock LLM provider exists; Python-only, parity clean
- **Could NOT verify:** live AI21 Bedrock endpoint behavior — code tracing confirms the fix is correct

### Assessment

The comma-to-colon fix is unambiguously correct, proven by a clean red-green cycle, and CI is green across all supported Python versions. No concerns beyond the optional inline-import cleanup above.

kartik-mem0 (approved):

PR #5529, opened 2026-06-13
fix: async delete_all aborts on first error, leaving partial deletion

HrushiYadav:
## Summary

- `AsyncMemory.delete_all()` uses `asyncio.gather(*delete_tasks)` without `return_exceptions=True`
- If any single `_delete_memory` call fails (e.g. store timeout, missing record), `gather` raises immediately and cancels all remaining delete tasks
- This leaves a partial deletion with no indication of which memories were deleted and which were not
- Added `return_exceptions=True` and logging of individual failures, consistent with the pattern already used in entity boost search (line 2952)

## Test plan

- [x] New test `test_async_delete_all_continues_on_partial_failure` verifies that 2 of 3 memories are deleted even when one raises `RuntimeError`
- [x] All existing memory tests pass

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3241,9 +3241,15 @@ class AsyncMemory(MemoryBase):
         for memory in memories[0]:
             delete_tasks.append(self._delete_memory(memory.id))
 
-        await asyncio.gather(*delete_tasks)
+        results = await asyncio.gather(*delete_tasks, return_exceptions=True)
 
-        logger.info(f"Deleted {len(memories[0])} memories")
+        errors = [r for r in results if isinstance(r, BaseException)]
+        if errors:
+            logger.warning("Failed to delete %d out of %d memories", len(errors), len(results))
+            for err in errors:
+                logger.warning("Delete error: %s", err)
+
+        logger.info(f"Deleted {len(results) - len(errors)} memories")
 
         decay_usage_notice = detect_decay_usage_from_delete_all(len(memories[0]))
         if decay_usage_notice:
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -762,6 +762,55 @@ async def test_async_delete_memory_history_has_timestamps(mock_sqlite, mock_llm_
     datetime.fromisoformat(call_kwargs["updated_at"])  # verify valid ISO timestamp
 
 
+@pytest.mark.asyncio
+@patch('mem0.utils.factory.EmbedderFactory.create')
+@patch('mem0.utils.factory.VectorStoreFactory.create')
+@patch('mem0.utils.factory.LlmFactory.create')
+@patch('mem0.memory.main.SQLiteManager')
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Adds `return_exceptions=True` to `asyncio.gather(*delete_tasks)` in `AsyncMemory.delete_all`, so a single per-memory failure no longer aborts the entire operation. Individual failures are aggregated and logged as warnings, matching the pattern already used in entity boost search (line 3128).

### Strengths
- `mem0/memory/main.py:3241–3253` — minimal, targeted fix using exactly the right asyncio idiom; the new error-handling block is nearly identical to the existing gather pattern at line 3128, so there's nothing novel to audit
- `tests/test_memory.py:765–815` — the regression test is a genuine red-green proof: it fails without the fix (RuntimeError propagates, gather aborts, zero or one deletions land), passes with it (two of three deletions complete)
- The sync `Memory.delete_all` (line 1704) uses a sequential for-loop — no equivalent concurrent bug there; correctly left untouched

### Issues
**Important (should address in follow-up, not a merge blocker):**
1. `tests/test_memory.py:810` — the test asserts `mock_vector_store.delete.call_count == 2` but does not check `memory.db.add_history.call_count`. `_delete_memory` calls `db.add_history` after each delete; the history-recording side-effect of the two successful deletions goes unverified. Adding that assertion would complete the coverage.
2. `mem0/memory/main.py:3247` — `delete_all` now returns `{"message": "Memories deleted successfully!"}` even on partial failure (consistent with the line 3128 pattern; failures surface only as log warnings). This is the intended new contract, but callers who previously relied on an exception being raised will no longer get that signal. Worth documenting in a follow-up issue so the change is traceable.

**Minor (optional):**
- `main.py:3249` — the new `logger.info` line uses an f-string while the two `logger.warning` lines above it in the same diff use `%`-style formatting; worth aligning
- No `Closes #N` link in the PR body; the repo template expects one for bug fixes

### Evidence
- **CI:** Python SDK 3.10 / 3.11 / 3.12 all pass; CI Gate passes; Vercel auth failure is unrelated
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/test_memory.py::test_async_delete_all_continues_on_partial_failure -v` → 1 passed in 4.98s
- **Red-green proof:** test fails without the fix (RuntimeError propagates through gather at line 3238, aborting remaining tasks) ✓; passes with the fix ✓
- **Cross-package check:** TS `deleteAll` (mem0-ts/src/oss/src/memory/index.ts:1524) uses a sequential `for...of await` loop — not `Promise.all` — so the same abort-on-error race does not exist there; no parity change needed
- **Linked issue:** none in PR body — noted as Minor
- **Could NOT verify:** live vector store / SQLite behavior under real concurrency; code-tracing only for those paths

### Assessment

The fix is correct, minimal, and pattern-consistent. Both Important items above are design-level notes for follow-up, not defects in the patch itself. Red-green proof is solid.

kartik-mem0 (approved):

kartik-mem0:
hey @HrushiYadav thank you for your contribution just one tip please open a issue before making any kind of pr as it is easy for me track and review here.

PR #5530, opened 2026-06-13
fix(redis): search() and list() crash when filters is None

### LINKED ISSUE #5586: bug(redis): search() and list() crash when filters is None
## Description

Redis vector store's `search()` and `list()` methods crash when `filters` is `None`, because they try to access keys on a None object without a guard.

## Steps to reproduce

1. Use Redis as the vector store backend
2. Call `search()` or `list()` without passing filters
3. Crashes with AttributeError/TypeError on None

PR: #5530

HrushiYadav:
## Summary

- `search()` and `list()` in Redis vector store crash with `AttributeError` when called without filters (the default)
- `filters` defaults to `None` but both methods call `filters.items()` without a null check
- Even with an empty dict `{}`, `reduce()` on an empty conditions list raises `TypeError`
- Added null/empty guards to both methods

## Test plan

- [x] Verified the crash path: `filters.items()` on `None` raises `AttributeError`
- [x] Verified `reduce()` on empty list raises `TypeError`
- [x] Fix guards both `None` and empty dict cases

Closes #5586

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -142,8 +142,11 @@ class RedisDB(VectorStoreBase):
         self.index.load(data, id_field="memory_id")
 
     def search(self, query: str, vectors: list, top_k: int = 5, filters: dict = None):
-        conditions = [Tag(key) == value for key, value in filters.items() if value is not None]
-        filter = reduce(lambda x, y: x & y, conditions)
+        if filters:
+            conditions = [Tag(key) == value for key, value in filters.items() if value is not None]
+            filter = reduce(lambda x, y: x & y, conditions) if conditions else None
+        else:
+            filter = None
 
         v = VectorQuery(
             vector=np.array(vectors, dtype=np.float32).tobytes(),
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for tracking this down and writing the fix — the `if filters:` guard with the `reduce(...) if conditions else None` fallback is exactly the right pattern, and it mirrors what `keyword_search()` already does a few lines up in the same file.

One thing I want to flag before we merge: there's an earlier open PR, #5446, that covers the same two crash paths in `search()` and `list()` with identical fix logic — and it ships four regression tests covering `filters=None`, `filters={}`, all-`None` values, and a no-regression check that a real filter still produces the expected `@user_id:{alice}` query. It was opened about a week ago and its CLA is signed.

The acceptance bar here requires tests that actually bite (the `reduce()` / `AttributeError` crash paths need to be exercised in CI), so merging #5530 as-is would leave us without coverage for the bug you just fixed. Rather than ask you to add tests to this PR while #5446 already has them, the cleaner path is to consolidate — either close this one in favor of #5446, or take the tests from #5446 and add them here.

Happy to move #5446 forward once it has a CI run. Either way, the underlying fix is solid and this bug is real — just want to make sure we land it with tests.

HrushiYadav:
Thanks for the review @kartik-mem0! You're right -- #5446 covers the same fix with proper regression tests. Closing in favor of that PR.

PR #5534, opened 2026-06-13
fix(chroma): get() raises IndexError when vector_id not found

HrushiYadav:
## Summary

- When `vector_id` does not exist in the collection, `_parse_output` returns an empty list
- Indexing `[0]` on that empty list raises `IndexError`
- Return `None` instead, consistent with other vector store implementations (qdrant, pgvector, etc.)

## Test plan

- [x] Verified `_parse_output` returns `[]` for empty results
- [x] Fix returns `None` for not-found, matching the interface contract


kartik-mem0:
Thank you for your contribution closing this pr in favour #5561 of this!

PR #5536, opened 2026-06-13
fix(litellm): function-calling check blocks all calls on non-tool models

HrushiYadav:
## Summary

- `supports_function_calling()` check runs unconditionally on every `generate_response()` call
- Models that don't support function calling (e.g., base completion models) are blocked from all calls, even plain text generation where no tools are requested
- Only run the check when `tools` are actually provided

## Test plan

- [x] Verified the check is at line 70, before any tool-specific logic
- [x] Fix gates the check on `tools` being truthy, matching the existing pattern at line 82

### DIFF
--- a/mem0/llms/litellm.py
+++ b/mem0/llms/litellm.py
@@ -67,7 +67,7 @@ class LiteLLM(LLMBase):
         Returns:
             str: The generated response.
         """
-        if not litellm.supports_function_calling(self.config.model):
+        if tools and not litellm.supports_function_calling(self.config.model):
             raise ValueError(f"Model '{self.config.model}' in litellm does not support function calling.")
 
         params = {
--- a/tests/llms/test_litellm.py
+++ b/tests/llms/test_litellm.py
@@ -25,6 +25,22 @@ def test_generate_response_with_unsupported_model(mock_litellm):
         llm.generate_response(messages, tools=tools)
 
 
+def test_generate_response_with_unsupported_model_no_tools(mock_litellm):
+    config = BaseLlmConfig(model="unsupported-model", temperature=0.7, max_tokens=100, top_p=1)
+    llm = litellm.LiteLLM(config)
+    messages = [{"role": "user", "content": "Hello"}]
+
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=Mock(content="hi"))]
+    mock_litellm.completion.return_value = mock_response
+    mock_litellm.supports_function_calling.return_value = False
+
+    response = llm.generate_response(messages)
+
+    assert response == "hi"
+    mock_litellm.supports_function_calling.assert_not_called()
+
+
 def test_generate_response_without_tools(mock_litellm):
     config = BaseLlmConfig(model="gpt-4.1-nano-2025-04-14", temperature=0.7, max_tokens=100, top_p=1)
     llm = litellm.LiteLLM(config)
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for tackling this — the one-line fix is correct and moves in the right direction. One issue needs to land before this is merge-ready.

**Strengths**

- The guard `tools and not litellm.supports_function_calling(...)` is the minimal, idiomatic change. It exactly mirrors the pattern already at line 85 (`if tools:`) and doesn't touch any call sites.
- Lint clean (ruff passes on both changed files).
- Conventional commit title (`fix(litellm): ...`).
- CI green on Python 3.10/3.11/3.12.

---

**Issues**

**Important — Updated test does not discriminate the fix from the bug**

`tests/llms/test_litellm.py`, `test_generate_response_with_unsupported_model`

The PR updates the existing test to pass `tools=tools` before asserting `ValueError` is raised. The problem: with `tools` provided, both the old unconditional check and the new guarded check reach the same raise. The test passes identically against the buggy code on `main`.

Verified locally (red-green revert):

- PR code + PR tests → 5 passed
- Reverted to buggy main code + PR tests → 5 passed (test should have FAILED here)

The scenario this fix actually unblocks — calling `generate_response(messages)` with no tools against a model where `supports_function_calling` returns `False` — has no test at all. That's the regression test this PR needs.

---

**Recommendations**

Add a test that asserts the fixed behavior directly. Here's the shape (from PR #5091, which addresses the same bug):

```python
def test_generate_response_with_unsupported_model_no_tools(mock_litellm):
    """Models that don't support function calling are still usable when no tools are passed."""
    config = BaseLlmConfig(model="unsupported-model", temperature=0.7, max_tokens=100, top_p=1)
    llm = litellm.LiteLLM(config)
    messages = [{"role": "user", "content": "Hello"}]

    mock_response = Mock()
    mock_response.choices = [Mock(message=Mock(content="hi"))]
    mock_litellm.completion.return_value = mock_response
    mock_litellm.supports_function_calling.return_value = False

    response = llm.generate_response(messages)  # no tools — must NOT raise

    assert response == "hi"
    mock_litellm.supports_function_calling.assert_not_called()
```

The `assert_not_called()` assertion is the discriminating check: with the buggy code `supports_function_calling` is always called; with the fix it is skipped entirely when no tools are passed. This test would fail on main and pass on the fix.

Note: PR #5091 (open since May 9) fixes the same bug. That PR moves the check inside the `if tools:` block rather than guarding at the top, and includes the regression test above. The two approaches are functionally equivalent — this PR's guard-at-the-top style is fine — but the missing test is the blocker either way.

---

**Assessment**

Ready to merge: No — needs the no-tools regression test before this is verifiable.

Once that test is added and passes, this is straightforward to squash-merge (`gh pr merge --squash`).

---

**Evidence**

- CI: Python SDK 3.10/3.11/3.12 pass, CI Gate pass (Vercel failure is auth-only, unrelated to code)
- Red-green proof: FAIL — `test_generate_response_with_unsupported_model` passes with buggy code after reverting `mem0/llms/litellm.py` to `main`; all 5 tests pass in both the pre-revert and post-revert runs
- Lint: `ruff check mem0/llms/litellm.py tests/llms/test_litellm.py` → clean
- TypeScript parity: no LiteLLM counterpart in `mem0-ts/src/oss/src/llms/` — Python-only provider, no cross-SDK gap
- No public API change, no docs update needed

HrushiYadav:
Thanks for the review! Pushed the missing regression test:

Added `test_generate_response_with_unsupported_model_no_tools` -- calls `generate_response(messages)` with no tools on a model where `supports_function_calling` returns False, asserts it returns normally (no ValueError), and verifies `supports_function_calling` was never called. This test fails on main and passes with the fix.

kartik-mem0 (approved):

PR #5537, opened 2026-06-13
fix(anthropic): tool_choice format and tool response parsing

HrushiYadav:
## Summary

Two bugs in the Anthropic LLM provider:

1. **tool_choice passed as bare string**: `params["tool_choice"] = tool_choice` passes `"auto"` directly, but the Anthropic API requires `{"type": "auto"}`, causing `400 Bad Request` on every tool call

2. **Tool responses raise AttributeError**: `response.content[0].text` is always called, but when the model responds with a `tool_use` block, it has no `.text` attribute. Unlike every other provider, AnthropicLLM had no `_parse_response` method.

## Changes

- Format `tool_choice` as `{"type": tool_choice}` for the API
- Add `_parse_response()` method that returns the standard `{"content": ..., "tool_calls": [...]}` dict when tools are used, matching other providers (OpenAI, Gemini, Bedrock)

## Test plan

- [x] Verified Anthropic API requires dict format for tool_choice
- [x] Verified `ToolUseBlock` has `name`, `input`, and `type` but no `text`
- [x] Response parsing matches the dict structure used by other providers

### DIFF
--- a/mem0-ts/src/oss/src/config/manager.ts
+++ b/mem0-ts/src/oss/src/config/manager.ts
@@ -112,6 +112,13 @@ export class ConfigManager {
               | undefined) ??
             userConf?.url ??
             defaultConf.baseURL;
+          const llmRaw = userConf as Record<string, unknown> | undefined;
+          const temperature =
+            userConf?.temperature ??
+            (llmRaw?.temperature as number | undefined);
+          const topP = userConf?.topP ?? (llmRaw?.top_p as number | undefined);
+          const maxTokens =
+            userConf?.maxTokens ?? (llmRaw?.max_tokens as number | undefined);
 
           return {
             baseURL: llmBaseURL,
@@ -125,6 +132,9 @@ export class ConfigManager {
               userConf?.modelProperties !== undefined
                 ? userConf.modelProperties
                 : defaultConf.modelProperties,
+            temperature,
+            topP,
+            maxTokens,
           };
         })(),
       },
--- a/mem0-ts/src/oss/src/llms/anthropic.ts
+++ b/mem0-ts/src/oss/src/llms/anthropic.ts
@@ -5,6 +5,9 @@ import { LLMConfig, Message } from "../types";
 export class AnthropicLLM implements LLM {
   private client: Anthropic;
   private model: string;
+  private maxTokens: number;
+  private temperature?: number;
+  private topP?: number;
 
   constructor(config: LLMConfig) {
     const apiKey = config.apiKey || process.env.ANTHROPIC_API_KEY;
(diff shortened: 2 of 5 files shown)


kartik-mem0:
@HrushiYadav can you please also verify if this regression also exist on the ts sdk if yes then add a fix for that as well in this pr.

thanks!

pr lgtm i will merge it once this is verified.

HrushiYadav:
Verified the TS SDK -- the bug does not exist there.

The TS `AnthropicLLM` class (`mem0-ts/src/oss/src/llms/anthropic.ts`) is a basic text-only implementation: `generateResponse` doesn't accept a `tools` parameter, doesn't pass `tool_choice` to the API, and only handles `text` content blocks. The two Python bugs (raw string `tool_choice` format and missing `tool_use` block parsing) don't apply since the TS Anthropic provider never invokes tool calling.

No TS changes needed for this PR.

kartik-mem0:
Pushed a follow-up that brings the TypeScript Anthropic provider (`mem0-ts/src/oss/src/llms/anthropic.ts`) to parity with this fix — the same two bugs existed on the TS side, and it was actually a bit worse there: `generateResponse` didn't accept a `tools` argument at all and threw `"Unexpected response type from Anthropic API"` on any non-text content block.

The TS provider now:
- forwards `tool_choice` as `{ type: "auto" }` (the object form the API requires) when tools are passed, instead of nothing
- parses `tool_use` blocks into `toolCalls` rather than throwing
- always returns the structured `{ content, role, toolCalls }` response when tools are provided, so it matches the Python `_parse_response` exactly (it previously returned a bare string when no tool was called)

Added unit tests covering both regressions (object-form `tool_choice`, no-throw on `tool_use`, argument parsing) plus the structured-response shape. Graph memory is deprecated in both SDKs now, so this is mainly about keeping the two providers consistent.

kartik-mem0:
Follow-up: bumped the default Anthropic model and aligned the TS provider's request params with Python, since graph memory is deprecated in both SDKs now and the two Anthropic providers should behave identically.

- **Default model** → `claude-sonnet-4-6` on both SDKs. The previous defaults were retired models (`claude-3-5-sonnet-20240620` in Python, `claude-3-sonnet-20240229` in TS).
- **TS request params** now match the Python `AnthropicConfig`: `max_tokens`, `temperature`, and `top_p` are read from config instead of hardcoding `max_tokens: 4096` and omitting sampling params. Defaults are `max_tokens=2000`, `temperature=0.1`, `top_p` omitted, with the same "never send both `temperature` and `top_p`, prefer `temperature`" rule the Python provider uses.
- To make those configurable end-to-end, `temperature`/`topP`/`maxTokens` are threaded through `LLMConfig`, the config schema, and `ConfigManager` (with snake_case normalization so Python-style `top_p`/`max_tokens` configs work too).

Note for reviewers: the TS default `max_tokens` drops from 4096 to 2000 to match Python — flagging since it's a behavior change. Added unit tests for the defaults, config forwarding, and the temperature/top_p mutual exclusion.

kartik-mem0 (approved):

PR #5540, opened 2026-06-13
fix: api_error_handler silently drops return values from async methods

### LINKED ISSUE #5579: bug: api_error_handler silently drops return values from async methods
## Description

The `api_error_handler` decorator wraps async methods but does not return the awaited result, causing all decorated async methods to return `None` instead of their actual values.

## Steps to reproduce

1. Call any async method decorated with `api_error_handler`
2. Return value is `None` instead of the expected result

PR: #5540

HrushiYadav:
## Summary

- `api_error_handler` uses a plain `def wrapper()`, not `async def`
- When decorating async methods on `AsyncMemoryClient`, `func(*args, **kwargs)` returns a coroutine that is never awaited
- The wrapper silently returns `None` instead of the actual API response
- Every `AsyncMemoryClient` method (`add`, `search`, `delete`, `get_all`, etc.) is broken -- all return `None`
- Detect async functions with `inspect.iscoroutinefunction()` and use an `async def` wrapper that properly awaits the coroutine
- Extracted error handling into shared helpers to avoid code duplication

## Test plan

- [x] Verified all `AsyncMemoryClient` methods are decorated with `@api_error_handler`
- [x] Verified the sync `MemoryClient` still uses the sync wrapper path
- [x] Error handling logic (HTTP errors, request errors) is preserved identically

Closes #5579

### DIFF
--- a/mem0/client/utils.py
+++ b/mem0/client/utils.py
@@ -1,5 +1,8 @@
+import inspect
 import json
 import logging
+from functools import wraps
+
 import httpx
 
 from mem0.exceptions import (
@@ -12,7 +15,7 @@ logger = logging.getLogger(__name__)
 
 class APIError(Exception):
     """Exception raised for errors in the API.
-    
+
     Deprecated: Use specific exception classes from mem0.exceptions instead.
     This class is maintained for backward compatibility.
     """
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths

- The core fix is exactly right: `inspect.iscoroutinefunction(func)` at decoration time (not call time) avoids false negatives on already-wrapped coroutines. The `return await func(*args, **kwargs)` in `async_wrapper` means the return value now propagates **and** exceptions inside async methods are caught by the `try/except` — two separate failures fixed in one clean change.
- `@wraps(func)` on both branches preserves `__name__`, `__doc__`, and `__wrapped__`, so stacked decorators and `iscoroutinefunction()` keep working correctly on the result.
- Extracting `_handle_http_error` and `_handle_request_error` as module-level helpers is a genuine DRY improvement. The httpx exception hierarchy is handled in the right order inside each helper (`TimeoutException` before `ConnectError` before the generic `RequestError` fallback).
- Sync path is structurally identical to pre-PR behavior — zero regression risk for `MemoryClient`.
- `project.py`'s `AsyncProject` (seven async methods) is fixed as a side effect.
- Ruff clean, conventional commit title.

### Issues

1. `mem0/client/utils.py` lines 103–106 / 116–119 — both wrappers call the helpers without an explicit `raise` guard afterward:

   ```python
   except httpx.HTTPStatusError as e:
       _handle_http_error(e)      # no raise here
   except httpx.RequestError as e:
       _handle_request_error(e)   # no raise here
   ```

   Today both helpers unconditionally raise, so this is not a current bug. But they're now public module-level functions: any future maintainer adding a logging-only early `return` (say, for a benign 202 they want to handle differently) would cause the `except` block to swallow the exception silently and return `None` — the exact silent failure mode this PR fixes. Adding a one-line guard makes the contract visible and safe:

   ```python
   except httpx.HTTPStatusError as e:
       _handle_http_error(e)
       raise  # unreachable today — guards against future helper changes
   ```

2. No regression test is committed. `tests/test_client.py` has no coverage for `api_error_handler` or `AsyncMemoryClient` error handling (confirmed: `grep -rn api_error_handler tests/` returns empty). The red-green proof shows the bug is real and the fix works, but without a committed test the same regression can re-enter silently on any future refactor. A minimal `tests/test_client_utils.py` with `pytest-asyncio` covering:
   - decorated async function returns the correct value (not `None`)
   - `inspect.iscoroutinefunction(decorated_async_fn)` is `True`
   - `httpx.HTTPStatusError` in an async body raises the right mem0 structured exception
   - sync path still returns values
   
   …would protect this fix permanently and directly matches the three cases that were broken before this PR.

#### Minor (optional)

- `from functools import wraps` is imported inside the function body. Since `import inspect` was already moved to the top-level, moving `wraps` there too is cleaner and avoids the repeated attribute lookup on every decoration call.

### Evidence

- **CI:** Python SDK 3.10 / 3.11 / 3.12 all pass. CI Gate pass.
- **Targeted tests:** 6-case suite covering all three failure modes (coroutine identity, HTTPStatusError intercept, ConnectError intercept) run in a clean worktree at head 282a7d26.
- **Red-green proof:** 3 tests fail on main (raw `httpx.HTTPStatusError` leaks out of async decorated function; `iscoroutinefunction` returns False on decorated async fn); all 6 pass with the fix; all 6 pass after restoring the fix.
- **Duplicate cluster:** no other open PR targets the same bug.
- **Cross-package check:** TypeScript SDK uses a different error-handling pattern — no parallel TS bug.
- **Could NOT verify:** live AsyncMemoryClient behavior against the hosted API (code tracing + local unit tests only).

### Assessment

Ready to merge: **With fixes** — please add the regression test and the `raise` guard in the `except` blocks; I'll re-review once you push. The fix itself is correct and the blast radius is real — every `AsyncMemoryClient` user today gets raw `httpx` exceptions instead of structured mem0 errors and loses `iscoroutinefunction` compatibility. Good catch.

HrushiYadav:
Thanks for the review! Pushed fixes for both items:

1. **Defensive raise guards** -- added `raise` after `_handle_http_error(e)` and `_handle_request_error(e)` in both sync and async wrappers. Currently unreachable since both helpers always raise, but guards against future helper changes that might add a non-raising path.

2. **Regression tests** -- added `tests/test_client_utils.py` with 7 tests:
   - `test_sync_returns_value` -- sync decorated function returns correct value
   - `test_async_returns_value` -- async decorated function returns correct value (not None)
   - `test_async_decorated_preserves_coroutine_flag` -- `inspect.iscoroutinefunction()` is True
   - `test_sync_decorated_is_not_coroutine` -- sync path stays sync
   - `test_sync_http_error_raises_structured_exception` -- 401 maps to AuthenticationError
   - `test_async_http_error_raises_structured_exception` -- 429 maps to RateLimitError
   - `test_async_connect_error_raises_network_error` -- ConnectError maps to NetworkError with NET_CONNECT code

Also moved `from functools import wraps` to module-level import and rebased onto latest main. All tests pass, ruff clean.

kartik-mem0:
Verified against 37721a684: both findings are fixed.

**Finding 1 — explicit raise guards** is addressed: `raise` added after `_handle_http_error(e)` and `_handle_request_error(e)` in both the async and sync wrappers (`mem0/client/utils.py` lines 109, 112, 122, 125). The helpers already always raise themselves, so these are defensively correct — no silent None fallthrough possible even if a helper's raise path is ever restructured.

**Finding 2 — regression tests** is addressed: `tests/test_client_utils.py` committed with 7 tests covering sync return value, async return value, `iscoroutinefunction` preservation (both paths), and error mapping (HTTPStatusError → `RateLimitError` / `AuthenticationError`, ConnectError → `NetworkError`).

Bonus: `from functools import wraps` moved to the top-level import block — clean.

Red-green proof run locally on the new head:
- WITH fix: 7 passed in 0.61s
- WITHOUT fix (impl reverted to main): 3 failed — `test_async_decorated_preserves_coroutine_flag`, `test_async_http_error_raises_structured_exception`, `test_async_connect_error_raises_network_error`
- Restored fix: 7 passed in 0.33s

CI on the new head: Python SDK 3.10 / 3.11 / 3.12 all pass, CI Gate passes.

## LGTM — ready for human review and squash-merge

### Assessment

Fix is correct, tests bite the actual failure modes, no architectural concerns. Every `AsyncMemoryClient` method was silently returning `None` due to the unawaited coroutine; this cleanly dispatches to an async wrapper at decoration time.

kartik-mem0 (approved):

PR #5541, opened 2026-06-13
fix: reset() only drops history table, leaving stale messages

### LINKED ISSUE #5578: bug: reset() only drops history table, leaving stale messages
## Description

`Memory.reset()` drops the history table but does not drop the messages table, leaving stale conversation data that persists across resets.

## Steps to reproduce

1. Add memories with messages
2. Call `reset()`
3. Messages table still contains old data

PR: #5541

HrushiYadav:
## Summary

- `Memory.reset()` and `AsyncMemory.reset()` manually execute `DROP TABLE IF EXISTS history` but never drop the `messages` table
- After reset, `get_last_messages()` returns stale conversation context from before the reset
- This stale context is fed to the LLM during the next `add()` call, causing incorrect memory extraction based on old conversations
- `SQLiteManager.reset()` already exists and properly drops both `history` and `messages` tables (line 326-339 in storage.py)
- Changed both sync and async reset to use `self.db.reset()` instead of manual DDL

## Test plan

- [x] Verified `SQLiteManager.reset()` drops both tables (storage.py lines 331-332)
- [x] Fix applied to both `Memory.reset()` and `AsyncMemory.reset()`
- [x] Existing reset tests pass (test_collection_name_preserved_after_reset)

Closes #5578

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1923,10 +1923,8 @@ class Memory(MemoryBase):
         """
         logger.warning("Resetting all memories")
 
-        if hasattr(self.db, "connection") and self.db.connection:
-            self.db.connection.execute("DROP TABLE IF EXISTS history")
-            self.db.connection.close()
-
+        self.db.reset()
+        self.db.close()
         self.db = SQLiteManager(self.config.history_db_path)
 
         if hasattr(self.vector_store, "reset"):
@@ -3465,10 +3463,8 @@ class AsyncMemory(MemoryBase):
         if hasattr(self.vector_store, "client") and hasattr(self.vector_store.client, "close"):
             await asyncio.to_thread(self.vector_store.client.close)
 
-        if hasattr(self.db, "connection") and self.db.connection:
-            await asyncio.to_thread(lambda: self.db.connection.execute("DROP TABLE IF EXISTS history"))
-            await asyncio.to_thread(self.db.connection.close)
-
+        await asyncio.to_thread(self.db.reset)
+        await asyncio.to_thread(self.db.close)
         self.db = SQLiteManager(self.config.history_db_path)
 
         self.vector_store = VectorStoreFactory.create(
--- a/mem0/memory/storage.py
+++ b/mem0/memory/storage.py
@@ -324,7 +324,9 @@ class SQLiteManager:
         ]
 
     def reset(self) -> None:
-        """Drop and recreate the history and messages tables."""
+        """Drop both tables. Caller is expected to replace this instance."""
+        if not self.connection:
+            raise RuntimeError("Cannot reset a closed SQLiteManager")
         with self._lock:
             try:
                 self.connection.execute("BEGIN")
@@ -335,8 +337,6 @@ class SQLiteManager:
                 self.connection.execute("ROLLBACK")
                 logger.error(f"Failed to reset tables: {e}")
                 raise
-        self._create_history_table()
-        self._create_messages_table()
 
     def close(self) -> None:
         if self.connection:
--- a/tests/memory/test_storage.py
+++ b/tests/memory/test_storage.py
@@ -281,14 +281,21 @@ class TestSQLiteManager:
         assert history[0]["is_deleted"] is False
         mgr.close()
 
-    def test_reset_clears_messages_table(self, temp_db_path):
-        """Regression: reset() must drop both history and messages tables."""
+    def test_reset_drops_tables(self, temp_db_path):
+        """reset() must drop both history and messages tables."""
         mgr = SQLiteManager(temp_db_path)
         mgr.add_history(memory_id="m1", old_memory=None, new_memory="new", event="ADD")
         mgr.save_messages([{"role": "user", "content": "hello", "name": None}], "sess1")
         mgr.reset()
-        msg_count = mgr.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
-        hist_count = mgr.connection.execute("SELECT COUNT(*) FROM history").fetchone()[0]
-        assert msg_count == 0, "messages table should be empty after reset"
-        assert hist_count == 0, "history table should be empty after reset"
+        tables = mgr.connection.execute(
+            "SELECT name FROM sqlite_master WHERE type='table' AND name IN ('history','messages')"
+        ).fetchall()
+        assert tables == [], "both tables should be dropped after reset"
         mgr.close()
+
+        mgr2 = SQLiteManager(temp_db_path)
+        msg_count = mgr2.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
+        hist_count = mgr2.connection.execute("SELECT COUNT(*) FROM history").fetchone()[0]
+        assert msg_count == 0
+        assert hist_count == 0
+        mgr2.close()
--- a/tests/memory/test_storage.py
+++ b/tests/memory/test_storage.py
@@ -280,3 +280,13 @@ class TestSQLiteManager:
         assert history[0]["actor_id"] is None
         assert history[0]["is_deleted"] is False
         mgr.close()
+
+    def test_reset_clears_messages_table(self, temp_db_path):
+        """Regression: reset() must drop both history and messages tables."""
+        mgr = SQLiteManager(temp_db_path)
+        mgr.add_history(memory_id="m1", old_memory=None, new_memory="new", event="ADD")
+        mgr.save_messages([{"role": "user", "content": "hello", "name": None}], "sess1")
+        mgr.reset()
+        count = mgr.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
+        assert count == 0, "messages table should be empty after reset"
+        mgr.close()
--- a/tests/memory/test_storage.py
+++ b/tests/memory/test_storage.py
@@ -287,6 +287,8 @@ class TestSQLiteManager:
         mgr.add_history(memory_id="m1", old_memory=None, new_memory="new", event="ADD")
         mgr.save_messages([{"role": "user", "content": "hello", "name": None}], "sess1")
         mgr.reset()
-        count = mgr.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
-        assert count == 0, "messages table should be empty after reset"
+        msg_count = mgr.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
+        hist_count = mgr.connection.execute("SELECT COUNT(*) FROM history").fetchone()[0]
+        assert msg_count == 0, "messages table should be empty after reset"
+        assert hist_count == 0, "history table should be empty after reset"
         mgr.close()
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -142,6 +142,33 @@ def test_memory_reset_clears_messages_table(mock_llm_factory, mock_vector_factor
     assert hist_count == 0, "history table must be empty after Memory.reset()"
 
 
+@pytest.mark.asyncio
+@patch('mem0.utils.factory.EmbedderFactory.create')
+@patch('mem0.utils.factory.VectorStoreFactory.create')
+@patch('mem0.utils.factory.LlmFactory.create')
+async def test_async_memory_reset_clears_messages_table(mock_llm_factory, mock_vector_factory, mock_embedder_factory, tmp_path):
+    """Regression: AsyncMemory.reset() must clear the messages table, not just history."""
+    mock_embedder_factory.return_value = MagicMock()
+    mock_vector_factory.return_value = MagicMock()
+    mock_llm_factory.return_value = MagicMock()
+
+    from mem0 import AsyncMemory
+
+    config = MemoryConfig()
+    config.history_db_path = str(tmp_path / "test.db")
+    memory = AsyncMemory(config)
+
+    memory.db.save_messages([{"role": "user", "content": "hi", "name": None}], "s1")
+    memory.db.add_history(memory_id="m1", old_memory=None, new_memory="x", event="ADD")
+
+    await memory.reset()
+
+    msg_count = memory.db.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
+    hist_count = memory.db.connection.execute("SELECT COUNT(*) FROM history").fetchone()[0]
+    assert msg_count == 0, "messages must be empty after AsyncMemory.reset()"
+    assert hist_count == 0, "history must be empty after AsyncMemory.reset()"
+
+
 @patch('mem0.utils.factory.EmbedderFactory.create')
 @patch('mem0.utils.factory.VectorStoreFactory.create')
 @patch('mem0.utils.factory.LlmFactory.create')
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -121,6 +121,30 @@ def test_collection_name_preserved_after_reset(mock_sqlite, mock_llm_factory, mo
 @patch('mem0.utils.factory.EmbedderFactory.create')
 @patch('mem0.utils.factory.VectorStoreFactory.create')
 @patch('mem0.utils.factory.LlmFactory.create')
+def test_memory_reset_clears_messages_table(mock_llm_factory, mock_vector_factory, mock_embedder_factory, tmp_path):
+    """Regression: Memory.reset() must clear the messages table, not just history."""
+    mock_embedder_factory.return_value = MagicMock()
+    mock_vector_factory.return_value = MagicMock()
+    mock_llm_factory.return_value = MagicMock()
+
+    config = MemoryConfig()
+    config.history_db_path = str(tmp_path / "test.db")
+    memory = Memory(config)
+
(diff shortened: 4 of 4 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- Correct root-cause fix: both `Memory.reset()` and `AsyncMemory.reset()` delegated to `SQLiteManager.reset()` (storage.py:326) which already drops and recreates both `history` and `messages` in a single transaction. The raw DDL that was bypassing this is cleanly removed.
- Removes a layering violation — `main.py` was reaching into `self.db.connection` to issue DDL that belongs to `SQLiteManager`'s schema-ownership layer.
- Async path correctly wraps the calls in `asyncio.to_thread`, matching the established pattern for blocking SQLite I/O in `AsyncMemory`.
- Restores Python↔TS parity: the TypeScript SDK's `Memory.reset()` already called `this.db.reset()` which cleared both tables — Python now matches.

### Issues

#### Important (should fix)
1. `tests/memory/test_storage.py` or `tests/test_memory.py` — no regression test covers the actual bug being fixed. The existing reset tests (`test_collection_name_preserved_after_reset`, `test_reset_skips_graph_when_graph_disabled`) mock `SQLiteManager` entirely and never touch the real tables. A test that seeds the `messages` table, calls `Memory.reset()`, and asserts the table is empty would be the direct regression test. Without it, this bug can silently re-emerge. The simplest place is `tests/memory/test_storage.py` — it already has a real `SQLiteManager` fixture. Something like:
```python
def test_reset_clears_messages_table(tmp_path):
    from mem0.memory.storage import SQLiteManager
    mgr = SQLiteManager(str(tmp_path / "test.db"))
    mgr.add_history("m1", "old", "new", "ADD")
    mgr.save_messages("sess1", [{"role": "user", "content": "hello", "name": None}])
    mgr.reset()
    count = mgr.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
    assert count == 0, "messages table should be empty after reset"
    mgr.close()
```

#### Minor (optional)
- `mem0/memory/main.py:1929-1930` (sync) / `3471-3472` (async) — `SQLiteManager.reset()` drops and recreates tables on the live connection, but immediately after, the connection is closed and a fresh `SQLiteManager` is constructed (whose `__init__` also calls `_create_history_table` + `_create_messages_table`). The table recreation inside `reset()` is dead work in this sequence (behavior is correct — just noisy). Not a blocker, but worth knowing for a future cleanup.
- Docstrings for both `reset()` methods could be updated to mention that both `history` and `messages` are cleared.

### Evidence
- **CI:** Python SDK 3.10 / 3.11 / 3.12 all PASS; CI Gate PASS; license/CLA PASS. Vercel advisory fail ignored (fork permissions, not merge-blocking).
- **Targeted tests:** `pytest tests/test_memory.py::test_collection_name_preserved_after_reset tests/test_memory.py::test_reset_skips_graph_when_graph_disabled tests/memory/test_storage.py -v --tb=short` → 16 passed in 3.62s
- **Red-green proof:** Seeded both `history` and `messages` tables; after the PR's `SQLiteManager.reset()` both rows are gone (`POST-RESET: history=0, messages=0` ✓). Simulated old code (drop history only): `POST-OLD-RESET: history=0, messages=1` — stale messages confirmed.
- **Lint:** `ruff check mem0/memory/main.py` → All checks passed
- **Linked issue:** #5578 — PR addresses the exact reported root cause.
- **Cross-package check:** TS SDK already had the correct behavior (`Memory.reset()` calls `this.db.reset()` which clears both tables). Python now matches.
- **Could NOT verify:** Live memory extraction after reset (requires a real vector store and LLM); confirmed via code tracing only.

### Assessment
Ready to merge: **With fixes** — the fix itself is correct and the red-green proof confirms it. Please add the regression test for `messages` table clearing; it's the only remaining gap. Re-review once pushed.

kartik-mem0:
Thanks for adding the test — let me share what the re-review found.

### What's verified fixed

The prior Important finding about a missing regression test: the new `test_reset_clears_messages_table` in `tests/memory/test_storage.py` adds a real-SQLite assertion (no mocking), which is the right direction.

---

### What still needs work

#### Critical — regression test does not bite the actual bug (tests/memory/test_storage.py, line 284)

The new test calls `mgr.reset()` directly on a `SQLiteManager` instance — and `SQLiteManager.reset()` was **never broken**. It always dropped both tables. The bug lived in `Memory.reset()` in `main.py`, which bypassed `SQLiteManager.reset()` entirely and ran `DROP TABLE IF EXISTS history` against the raw connection, never touching `messages`.

Red-green proof: I reverted `mem0/memory/main.py` to the pre-PR state (leaving the test untouched) and re-ran `test_reset_clears_messages_table`. It **passed** on the buggy code. This means if the `main.py` fix is ever reverted, CI will not catch it. The test validates `SQLiteManager`'s internal behavior, not the integration path that was broken.

A regression test that would actually fail on `origin/main` needs to go through `Memory.reset()` — something like:

```python
# In tests/test_memory.py or tests/memory/test_main.py
from mem0.configs.base import MemoryConfig

@patch('mem0.utils.factory.LlmFactory.create')
@patch('mem0.utils.factory.EmbedderFactory.create')
@patch('mem0.utils.factory.VectorStoreFactory.create')
def test_memory_reset_clears_messages_table(mock_vs, mock_emb, mock_llm, tmp_path):
    """Regression: Memory.reset() must clear the messages table."""
    config = MemoryConfig()
    config.history_db_path = str(tmp_path / "test.db")
    memory = Memory(config)

    # Seed both tables via the real SQLiteManager
    memory.db.save_messages([{"role": "user", "content": "hello", "name": None}], "sess1")
    memory.db.add_history(memory_id="m1", old_memory=None, new_memory="x", event="ADD")

    memory.reset()

    # After reset, the new db instance must have empty tables
    msg_count = memory.db.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
    hist_count = memory.db.connection.execute("SELECT COUNT(*) FROM history").fetchone()[0]
    assert msg_count == 0, "messages table must be empty after Memory.reset()"
    assert hist_count == 0, "history table must be empty after Memory.reset()"
```

This test FAILS on `origin/main` (messages count = 1 after the buggy reset) and PASSES with your fix. Please add this (or equivalent) alongside the existing `SQLiteManager`-level test.

---

#### Critical — `AsyncMemory.reset()` accidentally dropped the entity store reset (mem0/memory/main.py, ~line 3480)

Comparing `AsyncMemory.reset()` on `origin/main` against the PR head: the block that clears the entity store (used when graph memory is enabled) is absent from the PR:

**On main:**
```python
self.vector_store = VectorStoreFactory.create(...)

    try:
        await asyncio.to_thread(self._entity_store.reset)
    except Exception as e:
        logger.warning(f"Failed to reset entity store: {e}")
    self._entity_store = None

capture_event("mem0.reset", self, {"sync_type": "async"})
```

**On PR head:**
```python
self.vector_store = VectorStoreFactory.create(...)

capture_event("mem0.reset", self, {"sync_type": "async"})
```

The six-line entity store block is gone. `Memory.reset()` (sync) still has the equivalent block at lines 1945–1950, so graph memory users on the async path will call `reset()` and find their entity store still populated — entity relationships from before the reset survive into the next `add()` call. Please restore this block after the `VectorStoreFactory.create(...)` call in `AsyncMemory.reset()`.

---

### Evidence

- **CI (new head d870323):** Python SDK 3.10 / 3.11 / 3.12 all PASS; CI Gate PASS; CLA PASS. Vercel advisory fail ignored (fork permissions).
- **Worktree:** fetched and verified at `d8703238403577e3ed177bd5529e80077323741e`.
- **Targeted tests (PR head):** `pytest tests/memory/test_storage.py -v --tb=short` → 15/15 PASS including new test.
- **Red-green proof:** Reverted `mem0/memory/main.py` to pre-PR state, ran `test_reset_clears_messages_table` → **PASS** (should be FAIL — test does not bite the bug).
- **Entity store diff:** `git show origin/main:mem0/memory/main.py | sed -n '3453,3490p'` vs worktree `sed -n '3458,3485p'` — entity block confirmed missing on PR head.
- **Lint:** `ruff check mem0/memory/main.py tests/memory/test_storage.py` → All checks passed.

---

### Assessment

Changes requested. Two concrete asks before this can merge:

1. Add a `Memory.reset()`-level regression test (the `SQLiteManager` test is fine to keep — add the higher-level one alongside it).
2. Restore the entity store reset block to `AsyncMemory.reset()` after the `VectorStoreFactory.create(...)` call.

The `main.py` implementation fix itself remains correct. These are the only two gaps.

kartik-mem0:

Both blockers from the prior review are now fixed. Good persistence on this one.

### What this PR does

Replaces the inline `DROP TABLE IF EXISTS history` in `Memory.reset()` and `AsyncMemory.reset()` with a delegation to `SQLiteManager.reset()`, which atomically drops and recreates both `history` and `messages` tables inside a transaction. Also restores the entity-store reset block in `AsyncMemory.reset()` that was accidentally dropped in the first commit.

### Strengths

- `mem0/memory/main.py:1931` (sync) and `main.py:3473` (async) — correct delegation to `SQLiteManager.reset()`, the right abstraction boundary. If a future migration adds a third table, there's now one place to update.
- `main.py:3481–3486` — entity-store parity fully restored; async callers no longer leave entity-store data behind after reset.
- `tests/test_memory.py:121–144` — the new `test_memory_reset_clears_messages_table` goes through real `Memory.reset()` with a real SQLiteManager on a `tmp_path` db (mocked LLM/vector/embedder only). It asserts both `messages` and `history` are empty. Red-green confirmed: the test fails on the pre-fix code (messages table not cleared) and passes with the fix.

### Issues

Minor (optional):

- `tests/memory/test_storage.py:286–299` — `test_reset_clears_messages_table` inserts into both `history` and `messages`, calls `mgr.reset()`, but only asserts `COUNT(*) FROM messages == 0`. The docstring says "both history and messages tables" but skips the `history` assertion. The gap is fully covered by the integration test above, but the unit test is internally inconsistent. One-liner fix: add `assert mgr.connection.execute("SELECT COUNT(*) FROM history").fetchone()[0] == 0` after line 299.

### Evidence

- **CI:** Python SDK 3.10 / 3.11 / 3.12 PASS. `changelog_check` fails — this is a rebase artifact: `main` bumped to `2.0.7` and added the changelog entry after this branch diverged; the PR itself doesn't touch `pyproject.toml`. A rebase onto current `main` clears it before merge.
- **Targeted tests:** `pytest tests/test_memory.py::test_memory_reset_clears_messages_table` → 1 passed (0.61s); `pytest tests/memory/test_storage.py::TestSQLiteManager::test_reset_clears_messages_table` → 1 passed (0.31s)
- **Red-green proof:** Reverted `mem0/memory/main.py` to `origin/main` (buggy code), test failed with `AssertionError: messages table must be empty after Memory.reset() — assert 1 == 0`. Restored PR code → PASS. ✓
- **Entity store (async):** Block present at `main.py:3481–3486`, structurally matches the sync equivalent at `main.py:1941–1946`.
- **Lint:** `ruff check mem0/memory/main.py tests/test_memory.py tests/memory/test_storage.py` → all checks passed.
- **Could NOT verify:** Live Neo4j/Memgraph behavior for entity-store reset path — code-traced only.

### Assessment

Please rebase onto `main` first (one command: `git rebase origin/main`) to clear the `changelog_check` CI failure. The minor unit-test assertion gap can be fixed in this PR or as a follow-up — your call.

kartik-mem0:
Re-reviewing at bade32ef — after a closer read of the full call chain, I'm walking back the prior LGTM. Three things need to be addressed before this is ready to merge.

---

### Important — please fix before merge

**1. `SQLiteManager.reset()` creates tables that are immediately thrown away**

`mem0/memory/storage.py:338–339`

```python
self._create_history_table()
self._create_messages_table()
```

`reset()` drops both tables then recreates them. But every caller (`Memory.reset()` and `AsyncMemory.reset()`) immediately follows with `self.db.close()` + `self.db = SQLiteManager(...)` — which creates the tables a second time via `__init__`. Both tables are created twice on every reset.

The fix: strip the two `_create_*` calls from `SQLiteManager.reset()`. Its job is to drop; the new `SQLiteManager(...)` handles recreation. The method becomes a clean "wipe only" operation:

```python
def reset(self) -> None:
    """Drop both tables. Caller is expected to replace this instance."""
    with self._lock:
        try:
            self.connection.execute("BEGIN")
            self.connection.execute("DROP TABLE IF EXISTS history")
            self.connection.execute("DROP TABLE IF EXISTS messages")
            self.connection.execute("COMMIT")
        except Exception as e:
            self.connection.execute("ROLLBACK")
            logger.error(f"Failed to reset tables: {e}")
            raise
```

---

**2. Removing the None guard creates a silent crash path**

`mem0/memory/main.py` — both `Memory.reset()` and `AsyncMemory.reset()`

The original code had `if hasattr(self.db, "connection") and self.db.connection:` before touching the connection. That guard is gone. `SQLiteManager.close()` sets `self.connection = None`, so if `reset()` is ever called on an already-closed instance, `SQLiteManager.reset()` hits `self.connection.execute("BEGIN")` and raises `AttributeError: 'NoneType' object has no attribute 'execute'` — an opaque crash with no useful message.

Please add a guard inside `SQLiteManager.reset()`:

```python
def reset(self) -> None:
    if not self.connection:
        raise RuntimeError("Cannot reset a closed SQLiteManager")
    ...
```

---

**3. No regression test for `AsyncMemory.reset()`**

`tests/test_memory.py` has `test_memory_reset_clears_messages_table` for the sync path. The async path (`AsyncMemory.reset()`) dispatches via `asyncio.to_thread(self.db.reset)` + `asyncio.to_thread(self.db.close)` — a meaningfully different call path — and has no coverage. Please add:

```python
@pytest.mark.asyncio
@patch('mem0.utils.factory.EmbedderFactory.create')
@patch('mem0.utils.factory.VectorStoreFactory.create')
@patch('mem0.utils.factory.LlmFactory.create')
async def test_async_memory_reset_clears_messages_table(mock_llm, mock_vs, mock_emb, tmp_path):
    mock_emb.return_value = MagicMock()
    mock_vs.return_value = MagicMock()
    mock_llm.return_value = MagicMock()
    from mem0 import AsyncMemory
    config = MemoryConfig()
    config.history_db_path = str(tmp_path / "test.db")
    memory = AsyncMemory(config)
    memory.db.save_messages([{"role": "user", "content": "hi", "name": None}], "s1")
    memory.db.add_history(memory_id="m1", old_memory=None, new_memory="x", event="ADD")
    await memory.reset()
    msg_count = memory.db.connection.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
    hist_count = memory.db.connection.execute("SELECT COUNT(*) FROM history").fetchone()[0]
    assert msg_count == 0, "messages must be empty after AsyncMemory.reset()"
    assert hist_count == 0, "history must be empty after AsyncMemory.reset()"
```

---

The core bug fix is correct and the sync-path tests are solid — the row-count assertions are exactly what's needed. These three changes are all small; happy to re-review quickly once they're in.

kartik-mem0:

All three blockers from the prior review (2026-06-19) are resolved. Good turnaround.

### What was fixed

**1. Drop-only reset (mem0/memory/storage.py:326–342)** — the two trailing `_create_history_table()` / `_create_messages_table()` calls are gone. `SQLiteManager.reset()` now only drops; the new `SQLiteManager(...)` instance in the caller handles recreation. No more double-create on every reset.

**2. Closed-instance guard (storage.py:328–329)** — `if not self.connection: raise RuntimeError("Cannot reset a closed SQLiteManager")` is present and correct. The opaque `AttributeError` path is gone.

**3. Async regression test (tests/test_memory.py:147–168)** — `test_async_memory_reset_clears_messages_table` added, goes through `AsyncMemory.reset()` with a real SQLiteManager on a `tmp_path` db. Asserts both `messages` and `history` are empty after reset. Entity-store block also present at `main.py:3473–3478`, matching the sync equivalent at `main.py:1941–1946`.

Also noted: the `history` count assertion is now present in `test_reset_drops_tables` (tests/memory/test_storage.py:298) — the minor gap from the prior review is closed too.

### Evidence

- **CI:** Python SDK 3.10 / 3.11 / 3.12 PASS; CI Gate PASS; changelog_check PASS; CLA PASS. Vercel advisory fail ignored (fork permissions — not merge-blocking).
- **Targeted tests (PR head 4e7f939):** `pytest tests/test_memory.py::test_memory_reset_clears_messages_table tests/test_memory.py::test_async_memory_reset_clears_messages_table tests/memory/test_storage.py::TestSQLiteManager::test_reset_drops_tables -v` → **3 passed**.
- **Red-green proof:** Reverted `mem0/memory/main.py` + `mem0/memory/storage.py` to `origin/main` (buggy code), ran both `test_memory.py` regression tests → **2 FAILED** (AssertionError: messages table must be empty after Memory.reset() — assert 1 == 0; same for async). Restored PR code → **3 PASSED**.
- **Lint:** `ruff check mem0/memory/main.py mem0/memory/storage.py tests/test_memory.py tests/memory/test_storage.py` → all checks passed.
- **Could NOT verify:** Live Neo4j/Memgraph behavior for entity-store reset path — code-traced only.

### Assessment

KartikLabhshetwar (approved):

kartik-mem0 (approved):

PR #5545, opened 2026-06-14
fix(ts-oss): reject empty/blank messages in Memory.add() to prevent hallucinated memories

### LINKED ISSUE #5465: bug(ts-sdk): Memory.add() accepts empty messages array, causing hallucinated memories
## Description

The TypeScript OSS SDK `Memory.add()` does not validate empty message arrays. When called with `messages=[]`, the input passes through to the LLM for fact extraction, which can produce hallucinated memories — the same root cause as #4099 (Python SDK).

## Reproduction

```typescript
import { Memory } from 'mem0ai/oss';

const memory = new Memory({ /* config */ });

// This should reject or return empty — instead it reaches the LLM
await memory.add([], { user_id: "test-user", run_id: "test-run" });
```

## Root cause

In `mem0-ts/src/oss/src/memory/index.ts`:

- `add()` (line ~512) checks for `null`/`undefined` but **not** empty arrays
- `addToVectorStore()` (line ~604) runs `messages.map(m => m.content).join("\n")` which produces `""` for an empty array
- The empty string is then embedded and sent to the LLM for fact extraction
- Some LLMs hallucinate facts from the empty prompt

## Expected behavior

`Memory.add([])` should reject with a validation error (or return an empty result), matching the Python SDK fix from #4099.

## Affected code

- `mem0-ts/src/oss/src/memory/index.ts` — `add()` and `addToVectorStore()`

## Related

- Python SDK issue: #4099
- Python SDK fix (preferred): #4435 (raises `Mem0ValidationError` with `VALIDATION_004`)
- Python SDK fix (alternative): #4593

Bartok9:
## Summary

- `Memory.add()` in the TypeScript OSS SDK now rejects empty arrays, empty strings, and blank-only message content.
- Prevents hallucinated "ghost" memories that were being generated from an empty extraction prompt.

## Motivation

Closes #5465.

`Memory.add([])` passed the existing `null`/`undefined` guard but continued into `addToVectorStore()`, where `messages.map(m => m.content).join("\n")` produces an empty string `""`. That empty string was embedded and sent to the LLM for fact extraction, and some LLMs hallucinate facts from an empty prompt — the same root cause as the Python SDK issue #4099.

## Fix

Right after the existing null/undefined check in `add()`, reject:
- empty arrays (`messages.length === 0`)
- arrays whose messages all have blank (whitespace-only) string content
- empty / whitespace-only string input

This mirrors the validation direction of the Python SDK fix referenced in the issue (#4435 / #4593).

## Verification

- `npx jest src/oss/tests/memory.validation.test.ts` — **31 passed** (28 existing + 3 new)
- New cases: empty array, blank-only content array, empty string

## Notes

- 2 files changed, +37/-0. No behavior change for valid input.
- A maintainer's own PR #5472 targeted this issue but was closed unmerged. This PR keeps the validation logic minimal and does **not** alter the shared test harness config (the prior PR's broad `beforeAll` config changes), so it stays focused on the fix.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -619,6 +619,25 @@ export class Memory {
         "messages is required and cannot be undefined or null. Provide a string or array of messages.",
       );
     }
+    if (Array.isArray(messages)) {
+      if (messages.length === 0) {
+        throw new Error(
+          "messages array cannot be empty. Provide at least one message with non-empty content.",
+        );
+      }
+      const allBlank = messages.every(
+        (m) => typeof m.content === "string" && m.content.trim() === "",
+      );
+      if (allBlank) {
+        throw new Error(
+          "messages array cannot contain only blank content. Provide at least one message with non-empty content.",
+        );
+      }
+    } else if (messages.trim() === "") {
+      throw new Error(
+        "messages string cannot be empty. Provide non-empty content.",
+      );
+    }
 
     const temporalUsageNotice = detectTemporalUsageFromMetadata(
       config?.metadata,
--- a/mem0-ts/src/oss/tests/memory.validation.test.ts
+++ b/mem0-ts/src/oss/tests/memory.validation.test.ts
@@ -84,6 +84,24 @@ describe("Memory Input Validation", () => {
         memory.add(null, { userId: testUserId }),
       ).rejects.toThrow("messages is required");
     });
+
+    it("should throw error when messages is an empty array", async () => {
+      await expect(memory.add([], { userId: testUserId })).rejects.toThrow(
+        "messages array cannot be empty",
+      );
+    });
+
+    it("should throw error when messages array contains only blank content", async () => {
+      await expect(
+        memory.add([{ role: "user", content: "   " }], { userId: testUserId }),
+      ).rejects.toThrow("messages array cannot contain only blank content");
+    });
+
+    it("should throw error when messages is an empty string", async () => {
+      await expect(memory.add("   ", { userId: testUserId })).rejects.toThrow(
+        "messages string cannot be empty",
+      );
+    });
   });
 
   describe("search() threshold validation", () => {


kartik-mem0:
## Changes Requested

Thanks for picking this up and doing it cleanly — the TypeScript fix itself is correct and genuinely well-tested. There's one cross-SDK concern, plus a description accuracy fix, to resolve before it's merge-ready.

### Strengths
- Minimal, correctly-placed guard in `mem0-ts/src/oss/src/memory/index.ts` — it fires right after the existing null/undefined check and **before** any embedding, LLM fact-extraction, or telemetry, so invalid input fails fast with no side effects.
- The `allBlank` predicate handles multimodal content correctly: an image-only array (`content` is a `MultiModalMessages` object, not a string) is *not* rejected — that content is resolved to text by `parse_vision_messages` before the downstream `.join()`, so it's legitimately non-empty. Nice edge case to get right.
- **Red-green proven this run:** reverting the implementation to `main` makes the three new tests fail — the empty-array case literally resolves to a hallucinated `{"memory": "test memory"}` — and restoring the fix makes all 31 pass, with the 28 existing tests unaffected. The tests bite.

### Issues
#### Important (should fix)
1. **`add()` now diverges from the Python SDK — and the description overstates parity.** The PR says this "mirrors the validation direction of the Python SDK fix referenced in the issue (#4435 / #4593)." Both issues are actually still open with no merged fix, and Python's `Memory.add()` / `AsyncMemory.add()` (`mem0/memory/main.py:721-733` and `:2218-2230`) have **no** guard against `[]`, all-blank content, or `""` — they pass straight through to fact-extraction. So merging this makes `add([])` throw in TypeScript while it still silently produces (possibly hallucinated) memories in Python — a new behavioral split in the public API across the two SDKs.
   - **Preferred:** add the same guard to Python `Memory.add()` and `AsyncMemory.add()` (mirror the existing `Mem0ValidationError` pattern right after the type-normalization block) so this lands as a true cross-SDK parity fix and removes the divergence.
   - If you'd rather keep this PR TS-scoped, that's reasonable — but please correct the description so it doesn't imply Python is already fixed, and note that Python parity is tracked in #4435 / #4593. Either way the divergence should be a conscious, documented decision rather than a side effect.

#### Minor (optional)
2. Consider one happy-path test: a mixed array like `[{ role: "user", content: "   " }, { role: "user", content: "hello" }]` should **not** throw. It locks in the `every()` (not `some()`) semantics against a future regression.
3. Out of scope here, but worth a follow-up issue: a system-message-only array (`[{ role: "system", content: "..." }]`) passes this guard, yet `parse_vision_messages` strips system roles, leaving `""` to be embedded — the same empty-prompt class of bug on a different path. Pre-existing; not introduced by this PR.

### Evidence
- **CI:** TypeScript SDK build (Node 20/22), integration (20/22), changelog, CI Gate all green; CLA signed. The one red check (Vercel) is "Authorization required to deploy" — deploy auth, not code.
- **Targeted tests:** `pnpm jest src/oss/tests/memory.validation.test.ts` → **31 passed** (28 existing + 3 new) with the fix; prettier clean on both changed files.
- **Red-green proof:** impl reverted to `main` → the 3 new tests fail (empty array → hallucinated `{"memory":"test memory"}`; blank/empty string → empty results); impl restored → 31 pass. ✓
- **Linked issue:** #5465 (TS-scoped) — the PR addresses the reported root cause.
- **Cross-package check:** Python `Memory.add()`/`AsyncMemory.add()` lack the equivalent guard (`mem0/memory/main.py:721-733`, `:2218-2230`); #4435 / #4593 open with no merged fix → the divergence above.
- **Could NOT verify:** real-world LLM hallucination from an empty prompt (the suite mocks the embedder/LLM, as does the red-green run).

### Assessment
Ready to merge: **With fixes** — please resolve the cross-SDK divergence (extend the guard to Python, or keep it TS-scoped with the description corrected and the split documented). The TS code itself is solid and the regression test is exactly right; I'll re-review as soon as you push.

kartik-mem0:
## Re-review — updated now that #5613 has landed

Picked this back up against current `main`. Good news: the path to merge is short and clear.

**Status — the fix is correct and CI is green.** The guard is well-placed: it fires right after the existing null/undefined check (`mem0-ts/src/oss/src/memory/index.ts:617`), before any embedding, LLM fact-extraction, or telemetry, so invalid input fails fast with no side effects. All real checks are green (TypeScript SDK build + integration on Node 20/22, changelog, CI Gate, CLA signed). The only red check is Vercel — "Authorization required to deploy", i.e. deploy auth, not code.

**What changed since my last review.** #5613 merged a guard into the *hosted client* (`MemoryClient.add()`, `mem0-ts/src/client/mem0.ts`) that rejects an empty payload. That's a different code path from this PR — the hosted client vs. the OSS `Memory` class — so there's no conflict (your PR only touches `oss/`, #5613 only touched `client/`). It also means the project has now committed to "reject empty payloads" on the TS client side, so landing this makes the TS SDK internally consistent: client and OSS both fail fast instead of one throwing while the other hallucinates. That strengthens the case for this PR.

**Resolving the blocker from last time (the cross-SDK divergence).** Let's land this as **TS-scoped**. The Python side (`Memory.add()` / `AsyncMemory.add()` still pass `[]` straight through — `mem0/memory/main.py:721-733` and `:2250-2259`) will be tracked as a separate parity follow-up, so it does not block you here.

**One small thing to fix before merge.** The description says this "mirrors the validation direction of the Python SDK fix referenced in the issue (#4435 / #4593)." Those aren't merged — #4435 is still open and #4593 is closed — so Python has no empty-guard yet. Could you adjust the description to say this is TS-scoped and that Python parity is tracked separately? Once that's in, this is good to merge from my side.

**Optional (non-blocking).** A happy-path test that a mixed array like `[{ role: "user", content: "   " }, { role: "user", content: "hello" }]` does *not* throw — it locks in the `every()` (not `some()`) semantics against a future regression.

Thanks again for the clean fix and the tests that actually bite — this is close.

kartik-mem0 (approved):

PR #5546, opened 2026-06-14
fix(ts-oss): raise EmbeddingError on partial embed failure instead of dropping memories

### LINKED ISSUE #5509: Silent memory loss in the TypeScript SDK (`mem0ai`) when batch embedding partially fails — TS counterpart of #5245
## Summary

Observed in `mem0ai` 3.0.3.

## Root cause

```typescript
try {
  }
} catch {
    try {
    } catch (e) {
    }
  }
}
```

```typescript
  const text = mem.text;
  // ...
}
```

## Impact

## Reproduction (sketch)

```typescript
const flaky = {
  n: 0,
  },
  },
};
```

## Suggested direction

Bartok9:
## Summary

- `Memory.add()` now raises an `EmbeddingError` when one or more extracted memories cannot be embedded, instead of silently dropping them.
- New `EmbeddingError` class is exported so callers can catch it explicitly.

## Motivation

Closes #5509 — the TypeScript counterpart of the Python SDK fix #5249 (issue #5245), filed at @kartik-mem0's request in the #5249 review.

In `Memory.add()` Phase 3 (`mem0-ts/src/oss/src/memory/index.ts`), the batch embed has a per-item fallback:

```typescript
} catch {
    try {
    } catch (e) {
      console.warn(`Failed to embed memory text: ${e}`); // swallowed
    }
  }
}
```

When an individual `embed()` fails, `text` is never added to `embedMap`. The downstream materialization loop (`if (!text || !(text in embedMap)) continue;`) then skips it, so the extracted memory is silently discarded while the caller still receives a success result with missing memories.

## Fix

Collect the texts that fail the per-item fallback; if any remain unembedded, throw a new `EmbeddingError` naming the dropped texts. This mirrors the Python `EmbeddingError` added in #5249. The error is exported for callers to catch.

No behavior change when embedding succeeds (batch path or fully-successful fallback).

## Verification

- `npx jest src/oss/tests/memory.embed-failure.test.ts` — **2 passed** (new regression tests: throws `EmbeddingError`; message reports dropped text)
- `npx jest memory.add + memory.validation` — **38 passed** (no regression)

## Notes

- 2 files changed, +129/-0. Independent of #5545 (different issue/branch).
- Test isolates the failure to the Phase-3 memory-text embed path (query/probe/entity embeds still succeed), matching the real-world partial-failure scenario.

### DIFF
--- a/mem0-ts/src/client/index.ts
+++ b/mem0-ts/src/client/index.ts
@@ -42,7 +42,11 @@ export {
   NetworkError,
   ConfigurationError,
   MemoryQuotaExceededError,
+  EmbeddingError,
   createExceptionFromResponse,
 } from "../common/exceptions";
 
-export type { MemoryErrorOptions } from "../common/exceptions";
+export type {
+  MemoryErrorOptions,
+  EmbeddingErrorClass,
+} from "../common/exceptions";
--- a/mem0-ts/src/common/exceptions.ts
+++ b/mem0-ts/src/common/exceptions.ts
@@ -139,6 +139,65 @@ export class MemoryQuotaExceededError extends MemoryError {
   }
 }
 
+/**
+ * Coarse failure category for OSS Memory.add() embedding failures.
+ * Assigned at the point of detection (not parsed from the message).
+ * - provider: embed() / embedBatch() threw
+ * - validation: vector returned non-finite or wrong-dimension
+ * - internal: catch-all
+ */
+export type EmbeddingErrorClass = "provider" | "validation" | "internal";
+
+/**
+ * Raised when embedding one or more memory texts fails during OSS Memory.add(),
+ * which would otherwise cause those memories to be silently dropped.
+ * Mirrors the Python SDK EmbeddingError (see #5245 / #5509).
+ *
+ * Successful memories are persisted before this is thrown (preserve-then-raise).
+ * Callers can retry only `failedTexts`; `persistedCount` is the number actually
+ * committed via vector-store insert (not merely attempted).
+ */
+export class EmbeddingError extends MemoryError {
+  /** Texts whose embeddings failed and were therefore not persisted. */
+  readonly failedTexts: string[];
+  /** Number of memories that embedded and were actually inserted. */
+  readonly persistedCount: number;
+  /** Coarse failure category collapsed as validation > provider > internal. */
+  readonly errorClass: EmbeddingErrorClass;
+
+  constructor(
+    message: string,
+    opts: {
+      failedTexts?: string[];
+      persistedCount?: number;
+      errorClass?: EmbeddingErrorClass;
+      suggestion?: string;
+      details?: Record<string, unknown>;
+      debugInfo?: Record<string, unknown>;
+    } = {},
+  ) {
+    const failedTexts = opts.failedTexts ?? [];
+    const persistedCount = opts.persistedCount ?? 0;
+    const errorClass = opts.errorClass ?? "internal";
+    super(message, "EMBED_001", {
+      suggestion:
+        opts.suggestion ??
+        "Retry only the failed texts; successful memories were already persisted",
+      details: {
+        ...(opts.details ?? {}),
+        failedTexts,
+        persistedCount,
+        errorClass,
+      },
+      debugInfo: opts.debugInfo,
+    });
+    this.name = "EmbeddingError";
+    this.failedTexts = failedTexts;
+    this.persistedCount = persistedCount;
+    this.errorClass = errorClass;
+  }
+}
+
 // ─── HTTP Status → Exception Mapping ─────────────────────
 
 type MemoryErrorConstructor = new (
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1,6 +1,10 @@
 import { v4 as uuidv4 } from "uuid";
 import { createHash } from "crypto";
 import {
+  EmbeddingError,
+  type EmbeddingErrorClass,
+} from "../../../common/exceptions";
+import {
   MemoryConfig,
   MemoryConfigSchema,
   MemoryItem,
@@ -79,6 +83,9 @@ import { getDefaultVectorStoreDbPath } from "../utils/sqlite";
 import { logger } from "../utils/logger";
 import { normalizeExpirationDate, payloadIsExpired } from "../utils/expiration";
 import { getOrCreateMem0UserId } from "../../../client/config";
+
+export { EmbeddingError } from "../../../common/exceptions";
+export type { EmbeddingErrorClass } from "../../../common/exceptions";
 
 export class LLMError extends Error {
   readonly cause?: unknown;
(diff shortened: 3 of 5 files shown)


Bartok9:
Updated this PR in response to the design discussion on the issue (thanks @Adelagric, @shashiKundur1).

**The valid critique of my original approach:** raising on the first embed failure threw away the memories that *did* embed — the same "good items discarded" problem @Adelagric flagged on the Python side (#5249). I've fixed that here.

**What changed — preserve-then-raise:**
- `add()` now persists every memory that embedded successfully **first**, then raises `EmbeddingError` at the end. One failed item no longer fails the whole call or discards the good ones.
- `EmbeddingError` now carries structured fields so a caller can retry only what failed:
  - `failedTexts: string[]`
  - `persistedCount: number`
  - `errorClass: "provider" | "validation" | "internal"`
- **Error class is assigned at the point of detection, not parsed from the message** (per @Adelagric's point #2): `provider` = the `embed()` call threw; `validation` = the returned vector is non-finite / wrong-dimension.
- **Added the validation half** (@Adelagric's point #3): a `NaN`/`Inf` or wrong-dim vector doesn't throw today and gets persisted as a corrupt vector — the more dangerous half. There's now a finiteness + dimension check before persist; bad vectors are recorded as `validation` failures and **not** stored.

**Tests (4, all green):**
- full-failure → raises, `persistedCount === 0`;
- partial-failure → good memory persisted *and* readable via `getAll`, failed ones surfaced, `persistedCount === 1`;
- validation → NaN vector not persisted, `validation` outranks `provider` in the collapsed class.
- 44 existing add-related tests still pass; prettier clean.

**On `raise` vs `return {results, failed}` (the open design question):** I kept the raise shape to stay aligned with the Python taxonomy (#5245/#5249), but with preserve-then-raise the practical gap to @shashiKundur1's `{results, failed}` in #5554 is mostly the call signature — the good memories are persisted either way, and the failed-item metadata is identical. Happy to converge on whichever shape @kartik-mem0 prefers; if it's return-with-failures, the `failedTexts`/`errorClass` values here drop straight in. Keeping TS aligned with the Python-first taxonomy is the main thing.

Adelagric:
Thanks for folding all of it in — preserve-then-raise plus the detection-point assignment and the validation check is exactly the shape that was missing, good to see it converge.

On the one open fork — raise (here) vs return `{ results, failed }` (#5554): I don't think it has to be either/or. If `add()` returns the structured result by default and takes an opt-in `raise_on_partial_failure: bool` (default `false`), your preserve-then-raise becomes exactly the flag-enabled path — strict callers get the exception with `failedTexts` / `persistedCount`, while the default hands `results` + `failed` back without a try/catch. That's @inland-taipen's original proposal on #5245, and it means the two PRs aren't really competing designs so much as the two modes of one contract. Might be worth settling that over on #5509 where @kartik-mem0 raised it, so the decision doesn't fork across both PRs.

One concrete thing for cross-PR / cross-SDK consistency: align the `errorClass` string values to `provider_error` / `validation_error` / `internal_error` (Python's vocabulary, and what #5554 uses) rather than the short `provider` / `validation` / `internal` — so a caller gets identical labels regardless of which SDK or which shape lands.

Nice iteration.

kartik-mem0:
## Changes Requested

Thanks for a clean preserve-then-raise implementation — the core approach is right and Gate B proves it. Two things need fixing before this can land.

### Strengths

- **Preserve-then-raise ordering is correct.** Both `raiseIfEmbedFailures` call sites fire only after the successful memories have been committed — one early-exit (before the empty-return path) and one after the final `records.map`. The good memories genuinely survive a partial failure.
- **NaN/Infinity/dimension validation closes the dangerous half of #5509.** The `isValidVector` + `acceptEmbedding` helper catches vectors that embed successfully but are structurally corrupt, and excludes them from the store. This is the more insidious failure mode and it's good to see it addressed here.
- **Error class collapse priority (`validation > provider > internal`) is correct** — a bad vector silently stored is more dangerous than a transport error.
- **Tests are genuine regression tests.** `memory.embed-partial.test.ts` calls `memory.getAll` after the throw to confirm the good memory is actually in storage, which is exactly the right verification for preserve-then-raise.
- **Red-green proof clean:** 4/4 fail without the fix, 4/4 pass with it; 38/38 existing memory tests unaffected.

### Issues

#### Critical (must fix)

**`EmbeddingError` needs to move to `src/common/exceptions.ts` and extend `MemoryError`.**

`mem0-ts/src/common/exceptions.ts` is the SDK-wide home for all structured error classes — `AuthenticationError`, `RateLimitError`, `ValidationError`, `NetworkError`, etc. Every one of them extends `MemoryError` and includes `Object.setPrototypeOf(this, new.target.prototype)` in the constructor. The new `EmbeddingError` is currently defined inline in `src/oss/src/memory/index.ts` and extends bare `Error` with no prototype-chain fix.

The `setPrototypeOf` call is not cosmetic — without it, `instanceof EmbeddingError` can silently return `false` for consumers of the CommonJS build, because the prototype chain gets broken across the CJS/ESM boundary. Every other exception in this SDK has this guard for exactly that reason.

What to do:
1. Move `EmbeddingError` (and `EmbeddingErrorClass`) to `mem0-ts/src/common/exceptions.ts`, extending `MemoryError` with an `errorCode` like `"EMBED_001"` to match the existing `errorCode: string` pattern.
2. Add `Object.setPrototypeOf(this, new.target.prototype)` in the constructor (copy the pattern from `AuthenticationError` — it's three lines).
3. Re-export it from `src/client/index.ts` alongside the other exceptions (so it's accessible from `mem0ai` as well as `mem0ai/oss`).
4. Import it in `src/oss/src/memory/index.ts` from `"../../../common/exceptions"`.

#### Important (should fix)

**`persistedCount` can over-report when vector store inserts partially fail.**

`raiseIfEmbedFailures(records.length)` is called at the end of `add()`, but `records.length` is the number of records handed to the insert loop — not the number actually committed. The Phase 6 batch-insert fallback (lines ~918–932, unchanged by this PR) swallows individual insert errors with `console.error` and continues. If two records embed cleanly but one insert fails silently, `persistedCount` reports 2 while only 1 is in the store. A caller using `persistedCount` to decide which items to retry-skip would be misled.

Two acceptable fixes: (a) track successful inserts with a counter in the fallback loop and pass that to `raiseIfEmbedFailures` instead of `records.length`; or (b) update the JSDoc comment from "WERE persisted" to "attempted to persist" to set accurate expectations. Option (a) is more correct long-term; option (b) is pragmatic given the Phase 6 silent-insert-fail is pre-existing.

#### Minor (optional)

- `EmbeddingErrorClass` type: the `…Class` suffix on what is effectively an error category enum is slightly confusing alongside the `errorCode` vocabulary elsewhere in the SDK. `EmbeddingFailureKind` or `EmbeddingErrorCategory` would be clearer — but this is a naming-only suggestion and doesn't block merge.
- Both test files use `model: "gpt-5-mini"` — a fictional model ID. Since the LLM is fully mocked it doesn't break anything, but it's a slip worth fixing for documentation clarity.

### Evidence

- **CI:** TypeScript SDK build (Node 20, 22) PASS; integration tests (Node 20, 22) PASS; CI Gate PASS. Vercel = infra/auth, not a code check.
- **Targeted tests:** `pnpm run test -- src/oss/tests/memory.embed-failure.test.ts src/oss/tests/memory.embed-partial.test.ts` → **4 passed in 3.3s**
- **Red-green proof:** 4/4 fail with main impl (promise resolves instead of rejects; `EmbeddingError` undefined), 4/4 pass with fix restored ✓
- **Regression check:** `memory.add.test.ts` + `memory.validation.test.ts` → **38 passed, 0 failures**
- **Could NOT verify:** live vector store insert failures (code tracing only for the `persistedCount` issue — no live provider in the test env)

### Assessment

Ready to merge: **With fixes** — please address the Critical and Important items above and I'll re-review once you push. The core logic is sound; it's the exception placement and `persistedCount` semantics that need attention.

**On the cluster (#5546, #5551, #5554):** this is the cleanest raise implementation. #5554 is a different API shape (return vs throw) and targets a separate design decision. Once the two items above are addressed, this PR is the one to carry for the raise-on-failure path.

shashiKundur1:
@kartik-mem0 one thing on the cross-SDK angle, since it actually flips the parity argument.

@Adelagric's Python PR #5557 (Fixes #5245) landed the same shape there: add() returns { results, failed } by default and takes an opt-in raise_on_partial_failure that raises EmbeddingError after persisting the successes. Same error_class vocabulary, same detection-point rule. So Python is return-default + opt-in raise, not raise-everywhere.

That means if TS merges a raise-only fix (#5546) while Python is return-default, the two SDKs diverge on the wire, which is the thing the taxonomy work was trying to avoid. #5554 is the TS match to #5557, same default-return + raiseOnPartialFailure, and Adel and I kept the naming aligned so they are identical on the wire.

Also, the two changes you asked @Bartok9 to make on #5546 are already in #5554:
- EmbeddingError lives in src/common/exceptions.ts, extends MemoryError, with the setPrototypeOf guard, re-exported from mem0ai and mem0ai/oss.
- persistedCount counts actually-inserted records, not records.length, so a silent insert failure does not over-report.

#5546 is a clean raise implementation and I am not trying to step on it. The point is just that #5554 already covers that path (raise behind the flag) plus matches Python, so it can be the single one to carry.

On size: if the retry mechanism feels like too much for this fix, I am happy to move retryFailed and the per-class strategies to a follow-up PR and keep #5554 down to the core fix + the flag + the validation check. That gives you the small scoped version and the cross-SDK parity in one. Just say which you prefer and I will push it.

Bartok9:
Thanks @kartik-mem0 — addressed both the Critical and Important items on tip `440aec56` (also rebased onto current `main`).

### Critical — `EmbeddingError` placement
- Moved to `mem0-ts/src/common/exceptions.ts`
- Extends `MemoryError` with `errorCode: "EMBED_001"` (inherits `Object.setPrototypeOf` from the base constructor)
- Re-exported from `mem0ai` (`src/client/index.ts`) and `mem0ai/oss` (`src/oss/src/memory/index.ts`)
- OSS import path: `../../../common/exceptions` (same cross-boundary pattern as `../../../client/config`)

### Important — `persistedCount` accuracy
- Phase 6 now tracks actual inserts: batch success → `allIds.length`; per-item fallback increments only on successful `insert`
- `raiseIfEmbedFailures(persistedCount)` no longer uses `records.length`, so a swallowed insert error cannot over-report

### Unchanged product shape
- Preserve-then-raise still the contract (good memories commit first)
- Validation (non-finite / wrong-dim) outranks provider in collapsed `errorClass`
- Tests: `memory.embed-failure.test.ts` + `memory.embed-partial.test.ts` → **4/4 passed**
- Minor: tests use `gpt-4o-mini` instead of the fictional id

@shashiKundur1 — noted on #5554 (return-default + opt-in raise) vs this raise-only path. Leaving that design call to maintainers; this tip stays the raise-on-failure path kartik asked to carry once the two fix items landed. Happy to regroup if you prefer to consolidate onto one API.

PR #5547, opened 2026-06-14
fix(llms): send max_completion_tokens for the GPT-5 family across providers

### LINKED ISSUE #5054: bug: _get_common_params passes max_tokens to gpt-5.4-mini, which requires max_completion_tokens

Hexecu:
## Linked Issue

Closes #5054

## Description

`LLMBase._get_common_params` always emitted `max_tokens`. The GPT-5 family on the **Chat Completions API** (which mem0 uses via `client.chat.completions.create`) rejects the legacy `max_tokens` parameter and requires `max_completion_tokens` instead. So a non-reasoning GPT-5 model such as `gpt-5.4-mini` made every `add()` / `search()` call fail with:

```
openai.BadRequestError: Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

The reasoning GPT-5 models (`gpt-5`, `gpt-5o`, …) were unaffected because `_is_reasoning_model` routes them through `_get_supported_params`, which doesn't send `max_tokens` at all. The gap was only the **non-reasoning** GPT-5 variants that fall through to `_get_common_params`.

This PR adds a small name-based helper `_uses_max_completion_tokens()` on the base class — mirroring the provider-prefix handling already used by `_is_reasoning_model` — and selects the correct key in `_get_common_params`. The match is `gpt-5*` (after stripping any provider prefix), so it covers the whole current and future family (`gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-pro`, `gpt-5.5`, …) rather than a single hardcoded model name. Older models (`gpt-4.x`, `gpt-3.5`, o-series) keep using `max_tokens` exactly as before.

Because the fix lives in `LLMBase`, it covers every provider that inherits the base `_get_common_params` (OpenAI, Azure OpenAI, DeepSeek, vLLM, LM Studio, …). Two providers — `litellm` and `azure_openai_structured` — build their request params by hand instead of going through the base, and both default to `gpt-5-mini`, so they hit the same 400 out of the box; this PR routes their max-token param through the same helper so the whole family is covered. Anthropic is unaffected — it overrides `_get_common_params` itself.

**Scope (intentional):** this targets the Chat Completions path that mem0 actually uses (`max_completion_tokens`). The Responses API uses `max_output_tokens`, but mem0 doesn't call it, so it's left out of scope to keep the change surgical. This also lines up with the OpenAI OpenAPI spec, which marks `max_tokens` as deprecated in favor of `max_completion_tokens`.

For reference, #5073 attempted a fix earlier but was closed unmerged for an unsigned CLA (no technical review), and #5516 (thanks @eldar702) fixes the same bug locally in `openai.py`. This PR handles it once in the base and extends it to the hand-built providers, so the whole GPT-5 family across providers is covered.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

**Unit tests** (both directions, guarding against regression):
- `tests/llms/test_openai.py` — `gpt-5.4-mini` sends `max_completion_tokens` and not `max_tokens`; `gpt-4.1-nano` still sends `max_tokens`.
- `tests/llms/test_litellm.py` and `tests/llms/test_azure_openai_structured.py` — same checks for the two hand-built providers.

Full `tests/llms/` suite passes locally (138 passed, 2 skipped); `ruff check` is clean on the changed files.

**Manual testing** (live, against a real Azure `gpt-5.4-mini` deployment):
- Raw call (bypassing mem0): `max_tokens` → 400 `Unsupported parameter: 'max_tokens' ... Use 'max_completion_tokens' instead.`; `max_completion_tokens` → 200. Confirms the bug at the API level.
- mem0 `AzureOpenAILLM` (base `_get_common_params` path): **without** this fix → 400; **with** it → 200.
- mem0 `AzureOpenAIStructuredLLM` (one of the two hand-built providers extended here) → 200.

I exercised the Azure paths end-to-end because that's the deployment I had access to; OpenAI and LiteLLM share the same code path / helper and are covered by the unit tests above.

## Checklist

### DIFF
--- a/mem0/llms/azure_openai_structured.py
+++ b/mem0/llms/azure_openai_structured.py
@@ -76,9 +76,12 @@ class AzureOpenAIStructuredLLM(LLMBase):
             "model": self.config.model,
             "messages": messages,
             "temperature": self.config.temperature,
-            "max_tokens": self.config.max_tokens,
             "top_p": self.config.top_p,
         }
+        if self._uses_max_completion_tokens(self.config.model):
+            params["max_completion_tokens"] = self.config.max_tokens
+        else:
+            params["max_tokens"] = self.config.max_tokens
         if response_format:
             params["response_format"] = response_format
         if tools:
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -79,6 +79,25 @@ class LLMBase(ABC):
 
         return False
 
+    def _uses_max_completion_tokens(self, model: str) -> bool:
+        """
+        Check if the model expects ``max_completion_tokens`` instead of ``max_tokens``.
+
+        The whole GPT-5 family (gpt-5.4-mini, gpt-5.4-nano, gpt-5.5, ...) rejects the
+        legacy ``max_tokens`` parameter on the Chat Completions API and requires
+        ``max_completion_tokens``. Older models (gpt-4.x, gpt-3.5, etc.) still accept
+        ``max_tokens``.
+
+        Args:
+            model: The model name to check
+
+        Returns:
+            bool: True if the model requires ``max_completion_tokens``
+        """
+        # Strip provider prefixes (e.g. "openai/gpt-5.4-mini" -> "gpt-5.4-mini")
+        base_model = (model or "").lower().rsplit("/", 1)[-1]
+        return base_model.startswith("gpt-5")
+
     def _get_supported_params(self, **kwargs) -> Dict:
         """
         Get parameters that are supported by the current model.
@@ -141,9 +160,14 @@ class LLMBase(ABC):
         """
         params = {
             "temperature": self.config.temperature,
-            "max_tokens": self.config.max_tokens,
             "top_p": self.config.top_p,
         }
+
+        model = getattr(self.config, "model", "")
+        if self._uses_max_completion_tokens(model):
+            params["max_completion_tokens"] = self.config.max_tokens
+        else:
+            params["max_tokens"] = self.config.max_tokens
 
         # Add provider-specific parameters from kwargs
         params.update(kwargs)
--- a/mem0/llms/litellm.py
+++ b/mem0/llms/litellm.py
@@ -74,9 +74,12 @@ class LiteLLM(LLMBase):
             "model": self.config.model,
             "messages": messages,
             "temperature": self.config.temperature,
-            "max_tokens": self.config.max_tokens,
             "top_p": self.config.top_p,
         }
+        if self._uses_max_completion_tokens(self.config.model):
+            params["max_completion_tokens"] = self.config.max_tokens
+        else:
+            params["max_tokens"] = self.config.max_tokens
         if response_format:
             params["response_format"] = response_format
         if tools:  # TODO: Remove tools if no issues found with new memory addition logic
--- a/tests/llms/test_azure_openai_structured.py
+++ b/tests/llms/test_azure_openai_structured.py
@@ -120,6 +120,44 @@ def test_generate_response_without_tools(mock_azure_openai):
 
 
 @mock.patch("mem0.llms.azure_openai_structured.AzureOpenAI")
+def test_generate_response_gpt5_uses_max_completion_tokens(mock_azure_openai):
+    mock_client = Mock()
+    mock_azure_openai.return_value = mock_client
+
+    config = DummyConfig(model="gpt-5.4-mini", azure_kwargs=DummyAzureKwargs(api_key="real-key"))
+    llm = AzureOpenAIStructuredLLM(config)
+
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=Mock(content="Hi"))]
+    mock_client.chat.completions.create.return_value = mock_response
+
+    llm.generate_response([{"role": "user", "content": "Hi"}])
+
+    _, kwargs = mock_client.chat.completions.create.call_args
+    assert kwargs["max_completion_tokens"] == 256
+    assert "max_tokens" not in kwargs
+
+
+@mock.patch("mem0.llms.azure_openai_structured.AzureOpenAI")
+def test_generate_response_legacy_model_uses_max_tokens(mock_azure_openai):
+    mock_client = Mock()
+    mock_azure_openai.return_value = mock_client
+
+    config = DummyConfig(model="gpt-4.1", azure_kwargs=DummyAzureKwargs(api_key="real-key"))
+    llm = AzureOpenAIStructuredLLM(config)
+
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=Mock(content="Hi"))]
+    mock_client.chat.completions.create.return_value = mock_response
+
+    llm.generate_response([{"role": "user", "content": "Hi"}])
+
+    _, kwargs = mock_client.chat.completions.create.call_args
+    assert kwargs["max_tokens"] == 256
+    assert "max_completion_tokens" not in kwargs
+
+
+@mock.patch("mem0.llms.azure_openai_structured.AzureOpenAI")
 def test_generate_response_with_tools(mock_azure_openai):
     mock_client = Mock()
     mock_azure_openai.return_value = mock_client
--- a/tests/llms/test_litellm.py
+++ b/tests/llms/test_litellm.py
@@ -44,6 +44,40 @@ def test_generate_response_without_tools(mock_litellm):
     assert response == "I'm doing well, thank you for asking!"
 
 
+def test_generate_response_gpt5_uses_max_completion_tokens(mock_litellm):
+    config = BaseLlmConfig(model="gpt-5.4-mini", temperature=0.7, max_tokens=100, top_p=1)
+    llm = litellm.LiteLLM(config)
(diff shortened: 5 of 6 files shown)


Hexecu:
Some extra context on the root cause and why the fix is scoped the way it is, in case it helps the review.

**What happens today**

`LLMBase._get_common_params` builds the param dict with a hardcoded `max_tokens`. The GPT-5 models reject that key on Chat Completions:

```
openai.BadRequestError: Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

The reasoning GPT-5 models never hit this, because `_is_reasoning_model` already routes them through `_get_supported_params` (which omits `max_tokens` entirely). The only ones that fell through to `_get_common_params` were the **non-reasoning** GPT-5 variants like `gpt-5.4-mini` — which is exactly the case in #5054.

**Why `startswith("gpt-5")` rather than a single name**

The whole family requires `max_completion_tokens` on Chat Completions — `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-pro`, `gpt-5.5`, … — and they all share the `gpt-5` prefix, so a family match keeps this from regressing the next time a `gpt-5.x` is released. I deliberately avoided pinning a specific model name so we don't have to revisit this per release.

**Why it's safe across providers**

The change is in `LLMBase`, so it benefits every provider that inherits the base `_get_common_params`. Anthropic overrides `_get_common_params` itself, so it's untouched; providers that don't expose GPT-5 models never enter the new branch. The o1/o3 reasoning models are likewise unaffected (they don't go through `_get_common_params`).

**Scope note (intentional)**

This fixes the **Chat Completions** path, which is what mem0 calls (`client.chat.completions.create`) — hence `max_completion_tokens`. The Responses API uses `max_output_tokens`, but mem0 doesn't use it, so I left it out to keep the change minimal.

**Verification**

- `tests/llms/` → 126 passed, 2 skipped locally (no regressions).
- New tests assert both directions: `gpt-5.4-mini` → `max_completion_tokens`, `gpt-4.1-nano` → still `max_tokens`.
- `ruff check` clean on the touched files.

One reference for the maintainers: #5073 took a run at the same issue but was closed unmerged for an unsigned CLA (no technical review), so #5054 is still open. I went with the family-wide match plus regression tests here. Happy to adjust naming or add an `openai/gpt-5...` prefixed test case if you'd prefer — just let me know.

Hexecu:
Worth noting that @eldar702 independently spotted the same issue in #5516 and put up a clean, well-tested fix for it — good to see the bug confirmed from two angles.

I went with `LLMBase._get_common_params` here for reach: the `max_tokens` → `max_completion_tokens` switch is needed by every provider that inherits the base param builder, not just OpenAI. That base handles `openai` and `azure_openai`, both of which default to a GPT-5 model. I reproduced the Azure path end-to-end against a real `gpt-5.4-mini` deployment — without the fix the call returns a 400 `Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.`, and with the fix it returns 200.

Two providers don't go through the base — `litellm.py` and `azure_openai_structured.py` build their request params by hand, and both default to `gpt-5-mini`, so they were still broken out of the box. I've extended the same `_uses_max_completion_tokens` check to those two in this PR so the whole family is covered.

This lines up with the OpenAI OpenAPI spec, which marks `max_tokens` as deprecated in favor of `max_completion_tokens`. Scope stays honest: this is the Chat Completions path mem0 uses; the Responses API (`max_output_tokens`) isn't in play here. Older models (gpt-4.x and below) keep `max_tokens`, with regression tests covering that.

Hope this helps. 🙂

kartik-mem0:

### What this PR does
Fixes the `BadRequestError: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.` crash for GPT-5 family models (#5054). Adds `_uses_max_completion_tokens()` on `LLMBase` — same pattern as the existing `_is_reasoning_model` helper — and applies it in `_get_common_params` plus the two hand-built providers (`litellm.py`, `azure_openai_structured.py`) that don't go through the base.

### Strengths
- `_uses_max_completion_tokens` (`base.py:82–103`) mirrors `_is_reasoning_model` exactly — same provider-prefix stripping, same docstring format, clean placement. Consistent with the codebase's existing pattern.
- `getattr(self.config, "model", "")` at `_get_common_params:163` is a safe fallback; falling through to `max_tokens` on a missing attribute is the right safe default.
- `litellm.py` and `azure_openai_structured.py` both default to `gpt-5-mini`, so users are broken out of the box without this fix — caught correctly.
- 6 regression tests (2 per provider), each covering both directions: GPT-5 → `max_completion_tokens`, GPT-4.x → still `max_tokens`. Tests inspect the actual kwargs sent to the mock, which is a genuine wire-level check.
- Anthropic's override confirmed unaffected. `openai.py` and `azure_openai.py` inherit the fix automatically through `_get_supported_params` → `_get_common_params` without any changes needed.
- Python 3.9 compat: all builtins, no walrus or match.
- Ruff clean. Conventional commit title. No new core deps, no .env, no docs changes needed (internal param routing).

### Issues

Minor (optional):
- `groq.py:78` and `together.py:78` have the same inline `max_tokens` pattern — not a GPT-5 concern since neither provider surfaces those models, but worth a follow-up to bring them in line. A comment noting the intentional omission would also do.
- Tests use `call_args[1]` (tuple indexing) — `call_args.kwargs` reads more clearly, but no correctness issue.

### Evidence
- **CI:** Vercel deploy-preview auth failure (not a code CI check). CLA signed. No Python CI failures.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/llms/test_openai.py tests/llms/test_litellm.py tests/llms/test_azure_openai_structured.py -v --tb=short` → **34 passed in 1.95s**
- **Red-green proof:** Reverted `base.py`, `litellm.py`, `azure_openai_structured.py` to main → 3 GPT-5 regression tests fail as expected; restored to PR head → 6/6 pass. ✓
- **Ruff:** `All checks passed!` on the three changed impl files
- **Linked issue:** #5054 — PR addresses the reported root cause directly
- **Cross-package check:** TS SDK (`openai.ts`, `openai_structured.ts`) doesn't send any max-tokens param at all and `LLMConfig` has no `maxTokens` field — this is a separate pre-existing gap, not touched or worsened by this PR
- **Could NOT verify:** live OpenAI and LiteLLM API behavior (code tracing + author's reported manual test against Azure GPT-5.4-mini)

### Note on competing PRs
This PR handles the fix at the base class level and extends it to the two hand-built providers, giving it broader coverage than #5516 (which fixes `openai.py` only). Both are valid approaches; this one gets more ground with the same change.

### Assessment

The core logic is correct, red-green proof is clean, and Minors are genuinely optional. Squash-merge when ready.

Hexecu:
Thanks for the thorough review and for running the red-green proof independently, that's reassuring to see.

On the two minors:

- `groq.py` and `together.py` do share the same inline `max_tokens` pattern, but since neither surfaces GPT-5 models there's no live failure today. I'd rather keep this PR scoped to the #5054 root cause and handle those as a separate follow-up so the diff stays focused on the actual bug. Happy to open that follow-up if it's useful.
- Same reasoning on `call_args[1]` vs `call_args.kwargs` — purely cosmetic, so I left the existing tests as they are to avoid churn here. Glad to switch them if you'd prefer it in this PR.

Let me know if you'd rather fold either of those in before merge.

kartik-mem0 (approved):

PR #5458, opened 2026-06-09
fix(llms): honor reasoning-model params in OpenAIStructuredLLM

ly-wang19:
## What

`OpenAIStructuredLLM.generate_response` builds its request params manually and always sends `temperature`, bypassing the inherited `_get_supported_params`. `OpenAILLM` uses `_get_supported_params`, which drops `temperature`/`max_tokens` for reasoning models (o1/o3/gpt-5) that don't accept them.

As a result, with a reasoning model + structured output, mem0 sends `temperature` and the API rejects the request. The manual params also ignored `max_tokens`/`top_p` entirely.

## Fix

Route params through the inherited `_get_supported_params(messages=...)` (same as `OpenAILLM`), so reasoning models get the correct parameter set and regular models keep `temperature`/`max_tokens`/`top_p`.

## Tests

Added `tests/llms/test_openai_structured.py`:
- reasoning model (`o3-mini`) → no `temperature`, `reasoning_effort` forwarded
- regular model (`gpt-4o`) → `temperature` forwarded

## Note

`AzureOpenAIStructuredLLM` has the same pattern (manual params, no reasoning-model handling). Happy to fix it in this PR or a follow-up — whichever you prefer.

PR #5548, opened 2026-06-14
fix(llms): honor reasoning-model params in AzureOpenAIStructuredLLM

ly-wang19:
## What

`AzureOpenAIStructuredLLM.generate_response` sent `temperature`/`top_p` to **every** model, including reasoning models (o1/o3/GPT-5) that reject them — the same class of bug fixed for `OpenAIStructuredLLM` in #5458 (now merged).

Rebased onto current `main` and reconciled with the `_uses_max_completion_tokens` handling that landed in the meantime:

- **Reasoning models** (`_is_reasoning_model`): omit `temperature`/`top_p`, use `max_completion_tokens`, and forward `reasoning_effort` when configured.
- **Non-reasoning models**: forward `temperature`/`top_p` as before; `max_completion_tokens` vs `max_tokens` is chosen by the existing `_uses_max_completion_tokens` helper.

This keeps `main`'s `max_completion_tokens` behavior intact while bringing the structured Azure path to parity with the merged OpenAI fix (#5458) and the non-structured `AzureOpenAILLM`.

## Test
`tests/llms/test_azure_openai_structured.py` (existing gpt-5/legacy token tests still pass) plus:
- `test_reasoning_model_drops_sampling_params` — o3-mini: no `temperature`/`top_p`, `reasoning_effort` forwarded.
- `test_regular_model_sends_sampling_params` — gpt-4o: sampling params still forwarded.

`pytest tests/llms/test_azure_openai_structured.py` → **11 passed**; `ruff` clean.

### DIFF
--- a/mem0/llms/azure_openai_structured.py
+++ b/mem0/llms/azure_openai_structured.py
@@ -72,16 +72,26 @@ class AzureOpenAIStructuredLLM(LLMBase):
 
         messages[-1]["content"] = user_prompt
 
+        is_reasoning = self._is_reasoning_model(self.config.model)
         params = {
             "model": self.config.model,
             "messages": messages,
-            "temperature": self.config.temperature,
-            "top_p": self.config.top_p,
         }
-        if self._uses_max_completion_tokens(self.config.model):
+        # Reasoning models (o1/o3/GPT-5 series) reject temperature/top_p; only
+        # forward the sampling params for non-reasoning models. Mirrors the
+        # reasoning-aware handling of OpenAIStructuredLLM (#5458).
+        if not is_reasoning:
+            params["temperature"] = self.config.temperature
+            params["top_p"] = self.config.top_p
+        # Reasoning models require max_completion_tokens rather than max_tokens.
+        if is_reasoning or self._uses_max_completion_tokens(self.config.model):
             params["max_completion_tokens"] = self.config.max_tokens
         else:
             params["max_tokens"] = self.config.max_tokens
+        if is_reasoning:
+            reasoning_effort = getattr(self.config, "reasoning_effort", None)
+            if reasoning_effort:
+                params["reasoning_effort"] = reasoning_effort
         if response_format:
             params["response_format"] = response_format
         if tools:
--- a/tests/llms/test_azure_openai_structured.py
+++ b/tests/llms/test_azure_openai_structured.py
@@ -29,6 +29,7 @@ class DummyConfig:
         max_tokens=256,
         top_p=1.0,
         http_client=None,
+        reasoning_effort=None,
     ):
         self.model = model
         self.azure_kwargs = azure_kwargs or DummyAzureKwargs()
@@ -36,6 +37,7 @@ class DummyConfig:
         self.max_tokens = max_tokens
         self.top_p = top_p
         self.http_client = http_client
+        self.reasoning_effort = reasoning_effort
 
 
 @mock.patch("mem0.llms.azure_openai_structured.AzureOpenAI")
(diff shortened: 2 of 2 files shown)


kartik-mem0:
### Strengths

- **Exact pattern match.** Replacing the hardcoded params block with `_get_supported_params(messages=messages)` at `azure_openai_structured.py:75–76` brings this class in line with `azure_openai.py:130` (the non-structured sibling). The structured variant was the last outlier; after this PR there's no inconsistency between the two Azure classes.
- **Red-green proven.** `test_reasoning_model_drops_sampling_params` fails on the pre-PR implementation (temperature/max_tokens/top_p still present for o3-mini) and passes after the fix. `test_regular_model_sends_sampling_params` stays green in both states — no regression introduced.
- **DummyConfig updated correctly.** `reasoning_effort=None` added to `__init__` and `self.reasoning_effort` assigned, so `_get_supported_params`'s `getattr(self.config, 'reasoning_effort', None)` fires correctly for the reasoning-model test without impacting the seven pre-existing tests.
- **Conventional commit title is accurate and scoped.** `fix(llms):` with a specific class name.
- **Lint clean.** `ruff check` on both changed files: all checks passed.

### Issues

#### Critical (must fix)
None.

#### Important (should fix)
None.

#### Minor (nice to have)

**`tests/llms/test_azure_openai_structured.py` — no test for reasoning model + tools path**

The new `test_reasoning_model_drops_sampling_params` doesn't pass `tools=`. The code path where `tools` is forwarded from `_get_supported_params` conditionals through `params["tools"] = tools` is not exercised for a reasoning model. Mechanically correct by inspection (tools are set in the `if tools:` block after `_get_supported_params` returns, regardless of model type), but a single `test_reasoning_model_with_tools` would close the gap. Entirely optional — not a blocker.

**`tests/llms/test_azure_openai_structured.py` — `DummyConfig` could include `is_reasoning_model=None`**

`_is_reasoning_model` reads `getattr(self.config, "is_reasoning_model", None)` which gracefully handles the missing attribute. But adding `self.is_reasoning_model = None` to `DummyConfig` would make the stub match `BaseLlmConfig`'s actual field set. One-line addition, entirely optional.

### Recommendations

**Merge order relative to #5547:** PR #5547 by Hexecu (open, created ~2h before this one) also modifies `azure_openai_structured.py` lines 75–81 — the exact block this PR replaces — to add `max_completion_tokens` routing for the GPT-5 family. These two PRs conflict on that file. The right merge order is **this PR first** (#5548 is the broader, architecturally correct fix: delegating to `_get_supported_params` means `azure_openai_structured.py` automatically inherits any future `base.py` improvements, including #5547's `_uses_max_completion_tokens` addition to `_get_common_params`). After #5548 lands, ask @Hexecu to rebase #5547 and drop its `azure_openai_structured.py` hunk — only the `base.py` and `litellm.py` hunks need to survive.

### Assessment

Ready to merge: **Yes**

The fix is mechanically correct, precisely mirrors the pattern of `AzureOpenAILLM`, red-green proven locally (1 regression test fails on pre-PR code, all 9 pass with the fix), and lint-clean. No docs or dependency changes needed. The only pending CI is `build_mem0` (still running at review time) — Vercel failure is an infra/auth gate, not a code check. Merge order note for #5547 is the sole coordination item. Recommend squash-merge after CI green: `gh pr merge --squash 5548`.

---

**Evidence:** targeted tests 9/9 (Python 3.12, hatch dev_py_3_12); red-green proof: `test_reasoning_model_drops_sampling_params` fails with pre-PR impl (temperature present for o3-mini), passes with fix; `ruff check` clean on both changed files; CI build_mem0 pending at review time.

kartik-mem0:
Thanks @ly-wang19, and sorry for the churn here — now that #5547 has merged I went back through this against current `main` to make sure we keep the right fix rather than the convenient one.

There were really two separate things happening in `azure_openai_structured.py`:

1. **GPT-5 `max_completion_tokens`** — #5547 already handles this on `main` (it swaps `max_tokens` → `max_completion_tokens` for the gpt-5 family, inline in the param block).
2. **Reasoning models (o1/o3) + `reasoning_effort`** — this is *not* handled on `main`. The current code still sends `temperature`/`top_p` unconditionally and never forwards `reasoning_effort`, so an o1/o3 deployment through the structured client still hits the same "unsupported parameter" error. That's the gap this PR closes, and it's the part #5547 didn't touch.

So this isn't redundant with #5547. And the approach here — `params = self._get_supported_params(messages=messages)` — is the one we want long-term: it's the same delegation the non-structured `azure_openai.py` already uses (see `azure_openai.py:130`), so the structured class stops being the odd one out and automatically inherits base-class behavior, including #5547's `max_completion_tokens` logic via `_get_common_params`.

One thing before it can go in: it now **conflicts with the merged #5547** on this file, so it needs a rebase onto `main`. After the rebase the diff should be just the delegation swap. Could you also keep a regression test that a `gpt-5.4-mini` config still comes out with `max_completion_tokens` (and no `max_tokens`) after the change? It routes through `_get_supported_params` → `_get_common_params`, so it should hold — I just want that locked in so we don't quietly lose what #5547 shipped.

Once it's rebased and green I'll take another pass. Appreciate you sticking with this one.

ly-wang19:
Rebased onto `main` and reconciled with the `_uses_max_completion_tokens` change that landed alongside #5458 (thanks for merging that one!). The structured Azure path now mirrors the merged OpenAI fix: reasoning models drop `temperature`/`top_p` and use `max_completion_tokens`, non-reasoning models are unchanged. All existing tests (incl. the new gpt-5/legacy token tests) plus the two reasoning/regular cases pass — 11 passed, ruff clean. Ready for review.

kartik-mem0:
Verified against dd3f152: the rebase is clean — no regression of #5547's `max_completion_tokens` win.

## LGTM — ready for squash-merge

### What this PR does
Routes `AzureOpenAIStructuredLLM.generate_response` through reasoning-model-aware param selection: o1/o3 models drop `temperature`/`top_p`, receive `max_completion_tokens`, and forward `reasoning_effort`; standard models keep full sampling params. Completes the fix started for `OpenAIStructuredLLM` in #5458.

### Strengths
- Core fix is correct and precise (`azure_openai_structured.py:75–97`): three clean conditional blocks, no dead code.
- Token-param routing is airtight: `if is_reasoning or self._uses_max_completion_tokens(model)` correctly covers both pure reasoning models and the non-reasoning GPT-5.x family (e.g. `gpt-5.4-mini` → `_is_reasoning_model=False`, `_uses_max_completion_tokens=True` → still sends `max_completion_tokens`).
- Regression tests bite the real bug: `test_reasoning_model_drops_sampling_params` fails against the un-fixed impl and passes with the fix — verified locally.
- Inline approach is actually the right call here: `_get_supported_params()` does not emit a token-cap for reasoning models (it only handles sampling params); delegating to it would silently drop `max_completion_tokens` for o1/o3.

### Issues
Minor (optional):
- `test_reasoning_model_drops_sampling_params` doesn't assert the positive side: `max_completion_tokens == 256`. Would catch a future accidental drop of the token cap. Easy add if you want to harden it, but not a blocker.
- `if reasoning_effort:` silently swallows `reasoning_effort=""`. Pre-existing pattern in `base.py:128` — just something to watch.

### Evidence
- **CI:** Green — Python SDK 3.10/3.11/3.12 pass, CI Gate pass, Vercel = infra auth gate (known non-issue).
- **Targeted tests:** `pytest tests/llms/test_azure_openai_structured.py -v` → 11 passed in 4.12s (worktree at dd3f152).
- **Red-green proof:** reverted `azure_openai_structured.py` to main → `test_reasoning_model_drops_sampling_params` FAILED (`temperature` present for o3-mini); restored → PASSED. Test genuinely bites the bug.
- **#5547 non-regression:** `test_generate_response_gpt5_uses_max_completion_tokens` PASSED — `gpt-5.4-mini` still sends `max_completion_tokens` through the new conditional path.
- **Lint:** `ruff check` both changed files → All checks passed.
- **Could NOT verify:** live Azure reasoning-model API behaviour — code tracing + mock only; no Azure creds in CI.

### Assessment
Clean rebase, correct logic, solid tests, no regressions. Nothing blocking.

kartik-mem0 (approved):

PR #5551, opened 2026-06-14
fix(mem0-ts): surface EmbeddingError instead of silently dropping memories

### LINKED ISSUE #5509: Silent memory loss in the TypeScript SDK (`mem0ai`) when batch embedding partially fails — TS counterpart of #5245

shashiKundur1:
## Linked Issue

Closes #5509

## Description

add() can lose memories without telling you. I will explain but the short version is, if embedding fails for one text it just skips it and returns less, no error.

Ok so what happens. You call add, the LLM pulls out the facts, then each fact gets embedded so it can be stored. Normally all in one batch. If the batch call fails it falls back and embeds them one by one. That part is ok. The bug is in the catch. If one of them fails it only does a console.warn and keeps going, so that text never goes into the map, and later when it saves it just gets skipped. So 6 facts come out, maybe 4 get saved, and add returns like nothing went wrong.

Think of a courier who drops one of your parcels but still says everything was delivered. You don't notice until you go looking for that one parcel later and it's just not there. Same thing here.

I did not want to guess so I tested it. Fake embedder, batch always fails so it goes to the fallback, and one text fails there. Ran add 10 times. Every time 6 extracted, only 2 to 4 saved, no error. So it is not flaky, it drops them every time.

This is the TS version of the Python issue #5245 and PR #5249. Same pattern.

## The fix

Main thing, do not make it worse by also throwing away the good ones. So it still saves every memory that embedded fine. Only after that it throws an EmbeddingError. You keep the good data and you also get told what failed.

The error has failedTexts so you can retry just those, persistedCount for how many got saved, and errorClass which is provider or validation or unknown. The class is there so you know what to do next. Provider, retry. Validation, your config or input is wrong so retrying will not help. That errorClass part is the same thing discussed in #5245 and #5249, so the same shape can go on the Python side and keep both consistent.

Also the entity embedding paths have the same swallow but at console.debug. I left those for now. Did not want to grow the scope. Can do a follow up if you want.

## Type of Change

## Breaking Changes

Today add() never throws here, it just returns fewer results. After this it can throw an EmbeddingError when some texts fail. If you want the old behavior just catch it. The good memories are already saved by then so you can retry the failed ones from e.failedTexts.

```ts
import { EmbeddingError } from "mem0ai/oss";

try {
  await memory.add(messages, { userId });
} catch (e) {
  if (e instanceof EmbeddingError) {
    // good memories already saved, retry e.failedTexts if you want
  } else {
    throw e;
  }
}
```

## Test Coverage

- [x] I tested manually (ran add 10 times with a flaky embedder, saw the drop, then confirmed the fix)

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/index.ts
+++ b/mem0-ts/src/oss/src/index.ts
@@ -1,4 +1,5 @@
 export * from "./memory";
+export * from "./memory/errors";
 export * from "./memory/memory.types";
 export * from "./types";
 export * from "./embeddings/base";
--- a/mem0-ts/src/oss/src/memory/errors.ts
+++ b/mem0-ts/src/oss/src/memory/errors.ts
@@ -1,0 +1,63 @@
+// Why an embed failed: retry on "provider", fix config/input on "validation".
+export type EmbeddingErrorClass = "provider" | "validation" | "unknown";
+
+// Thrown by add() when some texts fail to embed; the rest are still persisted.
+export class EmbeddingError extends Error {
+  public readonly failedTexts: string[];
+  // Memories persisted by this add() call (after dedup), before the throw.
+  public readonly persistedCount: number;
+  public readonly errorClass: EmbeddingErrorClass;
+
+  constructor(
+    failedTexts: string[],
+    persistedCount: number,
+    errorClass: EmbeddingErrorClass = "unknown",
+  ) {
+    super(
+      `Failed to embed ${failedTexts.length} memory text(s); ` +
+        `${persistedCount} memory(ies) were persisted (errorClass=${errorClass}). ` +
+        `Inspect 'failedTexts' to retry the dropped memories.`,
+    );
+    this.name = "EmbeddingError";
+    this.failedTexts = failedTexts;
+    this.persistedCount = persistedCount;
+    this.errorClass = errorClass;
+    Object.setPrototypeOf(this, EmbeddingError.prototype);
+  }
+}
+
+// When several embeds fail, keep the most actionable class: provider (retryable)
+// beats validation beats unknown.
+export function mergeEmbeddingErrorClass(
+  a: EmbeddingErrorClass,
+  b: EmbeddingErrorClass,
+): EmbeddingErrorClass {
+  const rank = { provider: 2, validation: 1, unknown: 0 };
+  return rank[b] > rank[a] ? b : a;
+}
+
+// Guess why an embed() call failed from its message/status.
+export function classifyEmbeddingError(err: unknown): EmbeddingErrorClass {
+  const msg = (
+    err instanceof Error ? err.message : String(err ?? "")
+  ).toLowerCase();
+  const status = (err as any)?.status ?? (err as any)?.statusCode;
+
+  if (
+    status === 429 ||
+    (typeof status === "number" && status >= 500) ||
+    /rate.?limit|too many requests|timeout|timed out|econnreset|etimedout|socket hang up|network|503|502|504|429|temporarily unavailable/.test(
+      msg,
+    )
+  ) {
+    return "provider";
+  }
+  if (
+    /dimension|nan|invalid|malformed|empty|length|shape|expected .* got|validation/.test(
+      msg,
+    )
+  ) {
+    return "validation";
+  }
+  return "unknown";
+}
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -26,6 +26,12 @@ import {
   generateAdditiveExtractionPrompt,
 } from "../prompts";
 import { DummyHistoryManager } from "../storage/DummyHistoryManager";
+import {
+  EmbeddingError,
+  EmbeddingErrorClass,
+  classifyEmbeddingError,
+  mergeEmbeddingErrorClass,
+} from "./errors";
 import { Embedder } from "../embeddings/base";
 import { LLM } from "../llms/base";
 import { VectorStore } from "../vector_stores/base";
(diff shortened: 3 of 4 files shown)


kartik-mem0:
## Needs Human Review — design decision required

Good write-up in the PR description and the preserve-then-raise logic is correct — save the good memories first, then throw. That part I want to keep regardless of which direction this goes. Red-green proof ran clean (6/6 pass, fail on revert), so the fix does what it says.

That said, I can't recommend merging this as-is because (a) there are three PRs open for #5509 and we need to pick one approach, and (b) there are a couple of concrete issues in this specific implementation.

### Issues in this PR

**The `classifyEmbeddingError` regex has a false-positive risk** (`memory/errors.ts`). The message-string pattern includes bare digit literals — `503|502|504|429` — alongside the word-based patterns. Any error whose message text contains those as substrings — "embedding has 503 dimensions", "expected 504 got 768", "received 429 tokens" — gets classified as `"provider"` (retryable) even when it's a validation problem that retrying won't fix. The HTTP-status case is already handled by the `status === 429 || status >= 500` numeric guard directly above; the digit literals in the message regex are redundant and cause the false-positive. Removing just those four literals from the message regex fixes it.

**Test file is in the wrong directory.** The new test is at `src/oss/src/tests/embedding-failure.test.ts`. The established home for functional memory tests (the ~35 `memory.*.test.ts` files) is `src/oss/tests/`. The `src/tests/` subdirectory has three SQLite-path-specific tests; this isn't one of those. It'll run fine where it is, but it'll be hard to find and it breaks the search pattern for "where are the memory tests."

**The corrupt-vector path is still silent.** If `embedBatch()` returns normally but the result contains NaN values or wrong-dimension vectors, those vectors go straight into `embedMap` and get persisted without any check — no exception, no warning, no `failedTexts` entry. The PR fixes the exception-throw path (where `embed()` throws) but leaves the bad-return-value path untouched. That's arguably the harder half to catch because nothing throws.

### The design fork

Three PRs are open for the same issue:

- **#5546** (Bartok9): raise `EmbeddingError`, classify at the detection site (not from message regex), validates NaN/wrong-dim vectors from the batch return path, ~370 lines
- **#5551** (this PR, shashiKundur1): raise `EmbeddingError`, message-regex classification, ~272 lines
- **#5554** (shashiKundur1): non-throwing return shape — `add()` returns `{ results, failed }` with a `retryFailed()` method, ~1100 lines

The core question is **raise vs. return-with-failures**. My read:

The raise approach is a better fit here. The Python SDK raises on embed failure (#5249), so parity is automatic. The return shape of `add()` doesn't change for callers who don't hit the error — they get the same result they always have. Callers who do hit partial failure get the failed texts via `e.failedTexts` in the catch block, which is enough to retry. The return-with-failures shape in #5554 requires every `add()` call site to destructure `{ results, failed }` even when they never expect failures, which is a bigger breaking change to existing usage patterns.

Between #5546 and this PR, #5546 is the cleaner implementation: classification at the point of detection (no message-string guesswork), covers the NaN/corrupt-vector case, and puts the tests in the right directory. If this PR addressed those three issues above, it'd be comparable — but right now #5546 handles more of the surface area.

Kartik, the call on raise-vs-return and which PR to carry forward is yours. Happy to re-review whichever direction you pick once the author updates, or to look at #5546 more closely if you want that one compared in detail.

### Evidence
- **CI:** Vercel deploy (authorization — pre-existing, unrelated). CLA signed. No code CI failures.
- **Targeted tests:** `npx jest src/oss/src/tests/embedding-failure.test.ts` → 6 passed in 5.087 s (with fix)
- **Red-green proof:** Reverted `memory/index.ts` + `errors.ts` to main → test suite fails (module not found). Restored PR head → 6 passed in 4.635 s. Fix bites.
- **Could NOT verify:** live provider behavior (tests use stubs); `infer: false` path behavior; whether #5509 considers the corrupt-vector case in scope

shashiKundur1:
Thanks for the close review @kartik-mem0. I addressed all three points, but over in #5554 rather than here, since it is the fuller version of this same fix:

- the digit-literal regex false positive (503/502/429 in the message) is gone, classification is from the structured status and typed errors via instanceof, not the message string.
- the corrupt-vector path is covered now, every returned vector is validated (NaN/Inf, empty, wrong dim) before persist, so a bad returned vector is recorded and never stored.
- happy to move the tests to src/oss/tests/ to match the other memory tests.

On the raise vs return fork, I folded both into #5554 behind an opt-in raiseOnPartialFailure flag (default returns { results, failed }, flag on does preserve-then-raise), so they are two modes of one contract instead of competing. I left the detail on #5509 where you raised it. This PR (#5551) is the smaller raise-only fix if you prefer that shape, but #5554 now covers it plus the rest.

shashiKundur1:
Closing this one.

This was the raise-only version, and it is superseded. @Bartok9's #5546 is the cleaner implementation of that same shape and you already said it is the one to carry for the raise path, so there is no reason for mine to sit alongside it. The preserve-then-raise behaviour also exists in #6794 behind `raiseOnPartialFailure`, so nothing here is lost.

The follow-on work is in #6794 (rebased replacement for #5554, which could not be reopened after the force-push).

Thanks for the review on this one, the three points you raised are all addressed in #6794: the classification comes from structured status and typed errors via `instanceof` rather than a digit regex on the message, every returned vector is validated before persist so a corrupt-but-returned vector is caught and not just the throw path, and the tests are now in `src/oss/tests/`.

PR #5553, opened 2026-06-14
fix: async delete_all race condition corrupts entity store linked_memory_ids

### LINKED ISSUE #5577: bug: async delete_all race condition corrupts entity store linked_memory_ids
## Description

When `AsyncMemory.delete_all()` removes memories concurrently, entity store `linked_memory_ids` entries get corrupted because multiple async tasks read-modify-write the same entity record simultaneously.

## Steps to reproduce

1. Create multiple memories that share an entity
2. Call `delete_all()` on the user
3. Entity store entries retain stale/corrupted `linked_memory_ids`

PR: #5553

HrushiYadav:
## Summary

`AsyncMemory.delete_all()` fires all `_delete_memory` coroutines concurrently via `asyncio.gather`. Each coroutine calls `_remove_memory_from_entity_store`, which reads a snapshot of entity rows, removes its memory ID from `linked_memory_ids`, and writes back. Since all coroutines read the same pre-deletion snapshot before any writes commit, each write overwrites the previous one with a stale list. The last write wins, leaving deleted memory IDs permanently stuck in entity records.

### Impact

- **Silent, permanent corruption**: deleted memory IDs remain in `linked_memory_ids` indefinitely with no error or warning
- **Degraded search quality**: `_compute_entity_boosts` boosts dead memory IDs on every future `search()` call
- **Unbounded growth**: dead entity rows are never cleaned up, increasing latency on every entity lookup
- **Affects every `AsyncMemory` user**: entity extraction is enabled by default

### Race condition (before fix)

```
Entity "alice" has linked_memory_ids = [mem_A, mem_B]

Coroutine A reads [mem_A, mem_B], computes remaining = [mem_B], writes [mem_B]
Coroutine B reads [mem_A, mem_B] (stale!), computes remaining = [mem_A], writes [mem_A]

Final state: entity has [mem_A] -- mem_A is deleted but permanently linked
```

### Fix

- Add `_bulk_clear_entity_store()` that deletes all entity records for the scope in one sequential pass
- Call it in `delete_all()` before launching concurrent memory deletes
- Pass `skip_entity_cleanup=True` to concurrent `_delete_memory` calls since entity cleanup is already done
- Individual `delete()` calls (outside of `delete_all`) still do per-memory entity cleanup as before

### Test plan

- Added `test_async_delete_all_bulk_clears_entity_store` that verifies entity records are bulk-deleted and both memories are removed
- All 46 existing tests pass

Closes #5577

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3252,14 +3252,14 @@ class AsyncMemory(MemoryBase):
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "async"})
         memories = await asyncio.to_thread(self.vector_store.list, filters=filters)
 
-        if getattr(self, "_entity_store", None) is not None:
-            await self._bulk_clear_entity_store(filters)
-
         delete_tasks = []
         for memory in memories[0]:
             delete_tasks.append(self._delete_memory(memory.id, skip_entity_cleanup=True))
 
         results = await asyncio.gather(*delete_tasks, return_exceptions=True)
+
+        if self._entity_store is not None:
+            await self._bulk_clear_entity_store(filters)
 
         errors = [r for r in results if isinstance(r, BaseException)]
         if errors:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -2076,6 +2076,27 @@ class AsyncMemory(MemoryBase):
         except Exception as e:
             logger.warning(f"Entity upsert failed for '{entity_text}' (async): {e}")
 
+    async def _bulk_clear_entity_store(self, filters):
+        """Delete all entity records matching the given scope filters.
+
+        Used by delete_all to avoid the race condition that occurs when
+        concurrent _delete_memory coroutines each try to read-modify-write
+        the same entity rows' linked_memory_ids lists.
+        """
+        if self._entity_store is None:
+            return
+        search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
+        try:
+            listed = await asyncio.to_thread(self.entity_store.list, filters=search_filters, top_k=10000)
+            rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
+            for row in rows or []:
+                try:
+                    await asyncio.to_thread(self.entity_store.delete, vector_id=row.id)
+                except Exception as e:
+                    logger.debug(f"Bulk entity delete failed for id={row.id}: {e}")
+        except Exception as e:
+            logger.warning(f"Bulk entity store cleanup failed: {e}")
+
     async def _remove_memory_from_entity_store(self, memory_id, filters):
         """Async variant of `Memory._remove_memory_from_entity_store`."""
         if self._entity_store is None:
@@ -3231,9 +3252,12 @@ class AsyncMemory(MemoryBase):
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "async"})
         memories = await asyncio.to_thread(self.vector_store.list, filters=filters)
 
+        if self._entity_store is not None:
+            await self._bulk_clear_entity_store(filters)
+
         delete_tasks = []
         for memory in memories[0]:
-            delete_tasks.append(self._delete_memory(memory.id))
+            delete_tasks.append(self._delete_memory(memory.id, skip_entity_cleanup=True))
 
         results = await asyncio.gather(*delete_tasks, return_exceptions=True)
 
@@ -3418,7 +3442,7 @@ class AsyncMemory(MemoryBase):
 
         return memory_id
 
-    async def _delete_memory(self, memory_id, existing_memory=None):
+    async def _delete_memory(self, memory_id, existing_memory=None, skip_entity_cleanup=False):
         logger.info(f"Deleting memory with {memory_id=}")
         if existing_memory is None:
             existing_memory = await asyncio.to_thread(self.vector_store.get, vector_id=memory_id)
@@ -3444,9 +3468,8 @@ class AsyncMemory(MemoryBase):
             is_deleted=1,
         )
 
-        # Entity-store cleanup: strip this memory's id from any entity records
-        # that linked to it. Non-fatal — the helper swallows errors.
-        await self._remove_memory_from_entity_store(memory_id, session_filters)
+        if not skip_entity_cleanup:
+            await self._remove_memory_from_entity_store(memory_id, session_filters)
 
         return memory_id
 
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3252,7 +3252,7 @@ class AsyncMemory(MemoryBase):
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "async"})
         memories = await asyncio.to_thread(self.vector_store.list, filters=filters)
 
-        if self._entity_store is not None:
+        if getattr(self, "_entity_store", None) is not None:
             await self._bulk_clear_entity_store(filters)
 
         delete_tasks = []
--- a/tests/memory/test_decay_usage_notice.py
+++ b/tests/memory/test_decay_usage_notice.py
@@ -18,6 +18,7 @@ def make_async_memory():
     memory = AsyncMemory.__new__(AsyncMemory)
     memory.vector_store = MagicMock()
     memory._delete_memory = AsyncMock()
+    memory._entity_store = None
     return memory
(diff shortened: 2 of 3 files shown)


kartik-mem0:
## Changes Requested

*(supersedes my earlier comment — one confirmed Important finding added after full lenses + skeptic pass)*

### Strengths
- **Race fix is correct at the conceptual level.** Moving entity cleanup out of the concurrent scatter and into a single sequential pre-pass eliminates the read-modify-write race by construction — the concurrent `asyncio.gather` phase never touches entity state when `skip_entity_cleanup=True` is passed to every `_delete_memory` call.
- **`_bulk_clear_entity_store` (main.py ~line 2079)** mirrors the existing `asyncio.to_thread` pattern, consistent `logger.warning`/`logger.debug` error handling, and correctly scopes the filter to `user_id/agent_id/run_id`.
- **`skip_entity_cleanup=False` default** preserves full backward compatibility for individual `delete()` calls — no behavior change there.
- **Correctly scoped to async path only.** Sync `Memory.delete_all()` is sequential — no race, no change needed. TS `Memory.deleteAll()` is also sequential. Both verified.

### Issues

**`main.py:3255–3261` — Entity store is cleared before any memory deletion is confirmed, introducing a new partial-corruption window.**

Current ordering (post-PR):
1. `await self._bulk_clear_entity_store(filters)` — deletes ALL entity rows for the scope
2. `await asyncio.gather(*delete_tasks, return_exceptions=True)` — deletes memories

If the vector store raises on one or more memory IDs in step 2 (and `return_exceptions=True` silently absorbs the errors), those memories survive in the vector store with no entity records linking them — orphaned, silently ignored by entity-aware features like `_compute_entity_boosts`. Pre-PR, entity cleanup ran per-coroutine AFTER each memory's vector deletion succeeded, so a partial gather failure left surviving memories with their entity links intact. The PR inverts that safety property.

Suggested fix — one-pass reorder:
1. Run `await asyncio.gather(*delete_tasks, return_exceptions=True)` first (with `skip_entity_cleanup=True` as today)
2. Call `await self._bulk_clear_entity_store(filters)` after the gather completes

This closes the original race (entity cleanup is still sequential and outside the concurrent phase) while preserving entity integrity for any memories the vector store fails to delete.

#### Minor (optional)

- **`main.py:3255`** — `getattr(self, "_entity_store", None)` is redundant; `__init__` always sets `self._entity_store = None`. Change to `if self._entity_store is not None:` to match every other guard in the class.
- **`tests/test_memory.py`** — New test is a standalone function; all other async tests in the file are class methods inside `TestAsyncMemory` subclasses (~lines 1200–1258). Minor style inconsistency.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 PASS · CI Gate PASS · Vercel FAIL (advisory deploy check — ignored)
- **Targeted tests:** `pytest tests/test_memory.py::test_async_delete_all_bulk_clears_entity_store` → 1 passed in 4.25s
- **Red-green proof:** GREEN → FAIL (`Expected 'delete' to be called once. Called 0 times.`) → GREEN. Test bites the fix.
- **Lint:** `ruff check mem0/memory/main.py tests/test_memory.py` → All checks passed
- **Duplicate cluster:** PR #4893 (serialize vector-store writes, issue #4892 Qdrant HNSW) is a non-overlapping fix for a different failure path. Read both diffs — they touch different code, different bug classes, independent.
- **Could not verify:** Behavior under actual partial vector-store failure (no live backend in unit tests — code-traced only).

### Assessment
Ready to merge: **With fixes** — the Important item above is a one-pass reorder in `delete_all`: gather first, `_bulk_clear_entity_store` after. The race fix still holds (entity cleanup remains sequential and outside the concurrent phase), and the new orphan window is closed. Happy to re-review once you push.

kartik-mem0:
Thanks for the quick turnaround — the ordering fix is exactly right.

**Prior finding — resolved.** Gather now runs first, `_bulk_clear_entity_store` runs after. Partial vector-store failures leave entity links intact for surviving memories. That was the only blocker from the last round.

**New blocker — CI red on an existing test.**

```
FAILED tests/memory/test_decay_usage_notice.py::test_async_delete_all_decay_usage_runs_after_success
AttributeError: 'AsyncMemory' object has no attribute '_entity_store'
```

The commit message says "Replace getattr guard with direct self._entity_store is not None." In production that's fine — `AsyncMemory.__init__` always sets `self._entity_store = None` (line 1978). But `make_async_memory()` in that test file constructs via `AsyncMemory.__new__(AsyncMemory)`, which skips `__init__` entirely, so `_entity_store` is never placed on the object. The `if self._entity_store is not None:` check at `main.py:3261` then raises `AttributeError`.

One-line fix in `tests/memory/test_decay_usage_notice.py`:

```python
def make_async_memory():
    memory = AsyncMemory.__new__(AsyncMemory)
    memory.vector_store = MagicMock()
    memory._delete_memory = AsyncMock()
    memory._entity_store = None   # add this — mirrors AsyncMemory.__init__ line 1978
    return memory
```

The production logic in `delete_all()` is correct; the test helper just needs to reflect the real instance state. Once CI is green, this is good to go.

**Evidence**
- CI: FAIL — Python SDK / build_mem0 (3.11), job 81910849641 — 1 failed, 1451 passed
- Failure: `test_async_delete_all_decay_usage_runs_after_success` — `AttributeError: 'AsyncMemory' object has no attribute '_entity_store'`
- Ordering verified in diff: gather at `main.py:3257–3260`, entity clear at `main.py:3262–3263` — correct

kartik-mem0:

The `_entity_store` fix is exactly right. `make_async_memory()` constructs via `AsyncMemory.__new__`, which skips `__init__`, so `_entity_store` was never set on the object. Adding `memory._entity_store = None` matches what `__init__` does and the guard at `main.py:3261` now works cleanly.

### What this PR does

Fixes a read-modify-write race in `AsyncMemory.delete_all`: concurrent `asyncio.gather` coroutines each called `_remove_memory_from_entity_store`, where each read the same stale `linked_memory_ids` list and wrote back, leaving deleted memory IDs stuck in entity records. The fix replaces per-memory entity cleanup with a single `_bulk_clear_entity_store` run after the concurrent deletes complete.

### Strengths

- **Race fix is conceptually correct.** Bulk-clear after `gather` eliminates the concurrent read-modify-write entirely. The `return_exceptions=True` gather + skip-then-bulk pattern is the right shape for this problem.
- **Correct gather-first ordering.** Entity records are cleared after vector deletes succeed, not before — partial vector-store failures leave entity links intact for surviving memories.
- **`skip_entity_cleanup` flag is well-scoped.** Only `delete_all` sets it; the default is `False`, so all other `_delete_memory` callers are unaffected.
- **Scope-filtered bulk clear.** `user_id`/`agent_id`/`run_id` filters are applied — no cross-user contamination possible.
- **Sync `Memory.delete_all` correctly left untouched.** The race is async-only (sequential for-loop has no concurrency).
- **Test fix is minimal and correct.** One line, targets the exact root cause of the CI failure.

### Issues

Minor (optional): `test_async_delete_all_bulk_clears_entity_store` proves the bulk-clear path fires (`entity_store.delete.assert_called_once_with(...)`), but doesn't explicitly verify that `_remove_memory_from_entity_store` was NOT called per-memory. Adding `mock_entity_store.update.assert_not_called()` (or patching `_remove_memory_from_entity_store` and asserting it wasn't invoked) would make this a proper regression guard against someone accidentally dropping `skip_entity_cleanup=True`. Entirely optional — the behavior is correct as-is.

### Evidence

- **CI:** Real test jobs all green — Python SDK 3.10 (4m38s ✓), 3.11 (9m30s ✓), 3.12 (5m7s ✓). `changelog_check` fails only because the branch was cut when main was at `2.0.6` and main has since moved to `2.0.7` — the PR touches none of `pyproject.toml` or `docs/changelog/sdk.mdx`. A `git rebase origin/main` will clear this.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_decay_usage_notice.py tests/test_memory.py::TestAsyncDeleteAllEntityRace -v --tb=short` → **9 passed in 0.61s**
- **Red-green proof:** Removed `memory._entity_store = None` from `make_async_memory()` → `AttributeError: 'AsyncMemory' object has no attribute '_entity_store'` (1 failed, 7 passed). Restored → **8 passed** ✓
- **Linked issue:** #5577 — PR directly addresses the reported concurrent entity corruption.
- **Cross-package check:** TS `deleteAll` uses a sequential `for...of await` loop — no concurrent gather, race cannot occur there.
- **Could NOT verify:** Live entity store consistency under a real partial vector-store failure (code-traced only — entity cleanup is best-effort/non-fatal in both the old and new code).

### Assessment

Rebase onto main to clear the `changelog_check` stale-base failure, then this is good to go.

kartik-mem0 (approved):

PR #5554, opened 2026-06-14
feat(mem0-ts): return embedding failures with labels and add a retry mechanism

shashiKundur1:
## Linked Issue

Relates to #5509 (and the taxonomy talked about in #5245 / #5249)

## Description

This is a follow up to my other PR for #5509. That one just surfaces the failure. This one finishes the idea: it shows you what failed, why, and lets you retry it.

So what happens now. When you call add and some memories fail to embed, add no longer drops them quietly. It returns { results, failed }. results are the ones that saved. failed are the ones that did not, and each one carries a label:

- errorClass: provider, validation, or internal
- remediation: retry, reconfigure, or escalate
- retryAfter when the provider gave one

Before saving, every vector is checked (NaN, Inf, empty, wrong dimension). A bad vector is never stored. And if the vector store insert itself fails, that memory is reported in failed instead of being returned as if it saved.

Then there is a real retry, not just a label:

```ts
const { results, failed } = await memory.add(messages, { userId });
if (failed.length) {
  const retry = await memory.retryFailed(failed);
}
```

retryFailed reads each label and does the right thing:
- provider: re-embed (waits retryAfter), validate, save
- reconfigure: does not blindly retry, it would fail the same way, so it surfaces it for you to fix config
- internal (a returned NaN/Inf vector): sanitize the vector (NaN to 0, Inf to a finite sentinel) and save only if the result is still searchable, else keep it failed

The infinity part is the interesting bit. A vector store can not hold Infinity, and if you store it as the max float it comes back as Infinity again in float32 and breaks the math. So we keep its meaning as a finite sentinel that survives float32 and squaring, and the real word infinity is preserved in the text anyway.

Each failure gets a stable id when it is first flagged, so retrying the same one just overwrites instead of making a duplicate. No extra lookups, no dedup scan.

## Why return instead of raise

The earlier idea was to raise an error on failure. I went with return instead, for two reasons. You can not both raise and hand back the failed list, the raise stops everything. And raising is all or nothing, so a small blip looks like the whole add failed even when most memories saved. Returning keeps the good ones and still gives you the labels to act on.

## Type of Change

## Breaking Changes

add() used to return { results }. It now also returns failed when some memories fail to embed. Existing code that reads results keeps working. Code that wants the failures or the retry can use the new fields.

## Test Coverage

- [x] I tested manually (ran it many times with a flaky embedder: provider errors, NaN, Inf, wrong dimension, empty, insert failure, infer:false, retry, double-retry)

## Checklist

### DIFF
--- a/mem0-ts/src/common/exceptions.ts
+++ b/mem0-ts/src/common/exceptions.ts
@@ -149,15 +149,33 @@ export const EMBED_ERROR_CODE = {
 export type EmbedErrorCode =
   (typeof EMBED_ERROR_CODE)[keyof typeof EMBED_ERROR_CODE];
 
-/** Raised when the embedding step fails. The TS parallel of Python's EmbeddingError. */
+/**
+ * Raised when one or more texts fail to embed during `add()`. The TypeScript
+ * parallel of Python's `EmbeddingError`.
+ *
+ * Successfully-embedded memories are persisted before this is thrown, so a
+ * caller catching it can retry only the dropped texts via `failedTexts` rather
+ * than re-adding the whole batch.
+ */
 export class EmbeddingError extends MemoryError {
+  /** Texts that failed to embed and were not persisted. */
+  readonly failedTexts: string[];
+  /** Number of memories persisted by the call before this was thrown. */
+  readonly persistedCount: number;
+
   constructor(
     message: string,
     errorCode: string = EMBED_ERROR_CODE.TRANSIENT,
-    options?: MemoryErrorOptions,
+    options: MemoryErrorOptions & {
+      failedTexts?: string[];
+      persistedCount?: number;
+    } = {},
   ) {
-    super(message, errorCode, options);
+    const { failedTexts = [], persistedCount = 0, ...rest } = options;
+    super(message, errorCode, rest);
     this.name = "EmbeddingError";
+    this.failedTexts = failedTexts;
+    this.persistedCount = persistedCount;
   }
 }
 
--- a/mem0-ts/src/common/exceptions.ts
+++ b/mem0-ts/src/common/exceptions.ts
@@ -139,6 +139,28 @@ export class MemoryQuotaExceededError extends MemoryError {
   }
 }
 
+// Stable error codes for embedding failures (parallels the Python SDK).
+export const EMBED_ERROR_CODE = {
+  TRANSIENT: "EMBED_001", // provider blip (429, 5xx, network) — retry may work
+  VALIDATION: "EMBED_002", // bad input/vector (wrong dim, NaN) — retry won't help
+  AUTH: "EMBED_003", // auth/permission — needs operator action
+} as const;
+
+export type EmbedErrorCode =
+  (typeof EMBED_ERROR_CODE)[keyof typeof EMBED_ERROR_CODE];
+
+/** Raised when the embedding step fails. The TS parallel of Python's EmbeddingError. */
+export class EmbeddingError extends MemoryError {
+  constructor(
+    message: string,
+    errorCode: string = EMBED_ERROR_CODE.TRANSIENT,
+    options?: MemoryErrorOptions,
+  ) {
+    super(message, errorCode, options);
+    this.name = "EmbeddingError";
+  }
+}
+
 // ─── HTTP Status → Exception Mapping ─────────────────────
 
 type MemoryErrorConstructor = new (
--- a/mem0-ts/src/oss/src/index.ts
+++ b/mem0-ts/src/oss/src/index.ts
@@ -1,4 +1,10 @@
 export * from "./memory";
+export * from "./memory/errorRetry";
+export {
+  EmbeddingError,
+  EMBED_ERROR_CODE,
+  type EmbedErrorCode,
+} from "../../common/exceptions";
 export * from "./memory/memory.types";
 export * from "./types";
 export * from "./embeddings/base";
--- a/mem0-ts/src/oss/src/memory/errorRetry.ts
+++ b/mem0-ts/src/oss/src/memory/errorRetry.ts
@@ -1,8 +1,11 @@
 // Label-driven, guardrailed handling of embedding failures in add().
 import type { MemoryItem } from "../types";
 
-// Why an embed failed. "internal" = provider returned a poison vector without erroring.
-export type ErrorClass = "provider" | "validation" | "internal";
+// Why an embed failed. Values match the Python error_code vocabulary (#5245).
+export type ErrorClass =
+  | "provider_error"
+  | "validation_error"
+  | "internal_error";
 
 // What the caller should do about it.
 export type Remediation = "retry" | "reconfigure" | "escalate";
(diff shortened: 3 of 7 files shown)


Adelagric:
This is excellent — thanks for running so far with it. The `{ results, failed }` shape is the right call: it's @inland-taipen's bulk shape from #5245, and it's exactly *why* raise-on-any-failure is wrong (you keep the good ones and still hand back the labels). Validating every returned vector before persist, so NaN/Inf/dim never lands in the store, is the integrity guarantee the whole issue is about. The per-class strategies and the opt-in `retryFailed()` are a nice touch.

One substantive thing on the taxonomy — and it happens to simplify the trickiest part of the code:

**NaN/Inf should be `validation_error`, not `internal_error`.** Right now `classifyValidation` sends `dimension-mismatch → validation_error` but `non-finite → internal_error`. In the taxonomy the class answers *what kind of failure*, keyed to *where it's detected* — and NaN/Inf and wrong-dim are the same kind: a structurally-bad vector caught by inspecting the embedder's output. Same detection point → same class (`validation_error`). The escalate-vs-reconfigure difference you (correctly) want lives on the orthogonal `remediation` axis — NaN → `escalate`, dim → `reconfigure` — not on `error_class`. `internal_error` should stay the catch-all for mem0-side processing failures (dedup, serialization, store write), not a returned-vector defect.

The payoff: with NaN as `validation_error`, it routes to `ValidationReconfigureStrategy` (surface unchanged) instead of `InternalEscalateStrategy` (sanitize + persist). That removes the one spot I'd otherwise push on — sanitizing a poison vector (NaN→0 / Inf→1e19) and persisting it. Even gated at ≤5% repaired and non-degenerate, a NaN→0 vector isn't the model's actual output, so persisting it as a normal success quietly re-introduces a wrong-but-valid-looking record — the same failure mode in a softer form. Routing NaN to "surface, don't retry" sidesteps it. If you do want best-effort repair, I'd keep it explicit — a `sanitize: true` opt-in that marks the result as repaired, not a clean success.

Minor: `classifyEmbedError` falls back to a regex on the message (`/invalid|malformed|empty|length/`) when there's no status. That's the one place the class is derived from the string — fine as a last resort, but worth keeping it last (structured status / typed error first), since message wording isn't a stable contract.

Fixtures from the gist map over directly — just assert on the returned `failed[].error_class` instead of expecting a throw, since you went with the return shape (the right call). Happy to adapt them to your `EmbeddingFailure` field names if useful.

Really nice work — this is further than I expected the TS side to get this fast.

shashiKundur1:
Thanks @Adelagric, this is a really good catch and you are right.

I pushed it: NaN/Inf is validation_error now, not internal_error. Same detection point as wrong-dimension (we caught it by inspecting the returned vector), so same class, and the escalate vs reconfigure difference lives on the remediation axis like you said. internal_error is now only for mem0-side failures, like a vector store insert that fails.

The nice part is your call removed the one bit I was unsure about. NaN now routes to surface-and-don't-retry instead of sanitize-and-persist, so I dropped the sanitizer entirely. You are right that coercing NaN to 0 and saving it is the same wrong-but-valid-looking record in a softer form. Better to surface it than to quietly store a repaired vector.

Also lined the wire values up with the Python taxonomy: provider_error / validation_error / internal_error.

One thing I could not fully do from your point 1, ride the existing EmbeddingError. The TS OSS SDK does not have an exceptions file with EmbeddingError / error_code like Python does, so there was nothing to ride. I used a small local union with the same string values instead, so the two SDKs still line up on the wire. If you would rather I add a proper exceptions module to match Python, happy to.

Yes please send the fixtures, I will map them onto the returned failed[]. Thanks again, this was a good one.

Adelagric:
Great — that's the right shape now, and dropping the sanitizer is the call I'd have made too.

On the exceptions module though — good news, there's more to ride than you found. The TS OSS SDK *does* have one: `mem0-ts/src/common/exceptions.ts`. It has `MemoryError extends Error` with `errorCode` / `details` / `suggestion` (same constructor shape as Python's `MemoryError`), plus typed subclasses — `RateLimitError`, `ValidationError`, `NetworkError`, `AuthenticationError`, `ConfigurationError`. It's the direct parallel of `mem0/exceptions.py`; the only thing missing is `EmbeddingError` itself.

So instead of the local union, you could add one class — `export class EmbeddingError extends MemoryError` (code e.g. `EMBED_001`) — and ride it, exactly like the Python side. Bonus: the provider half can reuse what's already there — `RateLimitError` / `NetworkError` for the transient cases, `ValidationError` for the structural one — so `classifyEmbedError` maps onto existing types instead of message strings. That lines the two SDKs up at the type level, not just the wire string. Small change, since the module and base are already there.

(Totally fine to keep the union for this PR and do the exceptions wiring as a follow-up if you'd rather not grow the diff — the file's just there if you want real parity.)

Fixtures are in the gist I linked on #5509 (`rfc-fixtures-error-class.md`) — pytest/Python, so for TS just assert on the returned `failed[].errorClass` instead of a raise, which your `embedding-retry.test.ts` flagship/503/dim cases already do. Happy to sketch the TS mapping if useful.

Really good iteration on this.

Adelagric:
This is exactly it — all three addressed, and routing the classification through `instanceof RateLimitError / NetworkError` instead of the message regex is even cleaner than what I suggested. The same-class / remediation-on-the-orthogonal-axis test is the right guardrail to lock it in. Nice work — in good shape for the maintainer now.

shashiKundur1:
Appreciate it, and thanks for steering this the whole way. The taxonomy and the fixtures are what made it click.

I also ran it end to end (not just the unit tests): add() on 5 facts where 3 fail keeps the 2 good ones, surfaces the 503 as provider_error/retry, the NaN as validation_error/escalate, and the wrong dim as validation_error/reconfigure, with the NaN and dim vectors never reaching the store. retryFailed then recovers only the 503 once the provider is back. Lines up with your T_503 / T_NAN / T_DIM map.

Ready for the maintainer whenever. Thanks again.

shashiKundur1:
Quick update on this one: it now implements the unified shape from #5509.

add() returns { results, failed } by default, and add(messages, { raiseOnPartialFailure: true }) persists the good memories first then throws EmbeddingError with failedTexts / persistedCount. So the return shape and the preserve-then-raise shape are two modes of the same contract, not competing designs. errorClass values match the Python taxonomy and it rides the existing common/exceptions.ts (EmbeddingError extends MemoryError). 600 tests pass, build clean.

kartik-mem0:

shashiKundur1:
Continued in #6794.

I rebased this on current `main` as you suggested, but GitHub will not let a closed PR be reopened once its branch has been force-pushed, and a rebase requires one. The error is "state cannot be changed. The branch was force-pushed or recreated". So I opened the fresh PR from the second half of your note instead.

Everything is carried over, plus three bugs the rebase exposed in the `infer: false` and `retryFailed()` paths that the old tests could not see. Details are on #6794.

Thanks @Adelagric and @Bartok9, the discussion here is what shaped it.

PR #5556, opened 2026-06-15
fix(llms): accept and forward **kwargs in Together/LangChain/Sarvam providers

### LINKED ISSUE #5555: Together / LangChain / Sarvam LLM providers break the generate_response(**kwargs) contract
## Bug

`LLMBase.generate_response` is declared `(self, messages, tools=None, tool_choice="auto", **kwargs)`, with `**kwargs` documented as "Additional provider-specific parameters". Mainstream providers (OpenAI, DeepSeek, vLLM, xAI, Anthropic, Ollama) accept and forward them. But `TogetherLLM`, `LangchainLLM`, and `SarvamLLM` override `generate_response` **without** `**kwargs`, so any per-call parameter raises `TypeError: generate_response() got an unexpected keyword argument '...'`.

## Repro

Call `.generate_response(messages, temperature=0.2)` (or any extra kwarg) on a Together-, LangChain-, or Sarvam-backed LLM → `TypeError`.

## Fix

Add `**kwargs` to those three providers' `generate_response` and forward them into the underlying call, matching the sibling providers.

yashs33244:
Closes #5555.

## Summary

`TogetherLLM`, `LangchainLLM`, and `SarvamLLM` override `generate_response` without `**kwargs`, breaking the base-class contract (documented as "Additional provider-specific parameters") that the other providers honor — any per-call kwarg raised `TypeError`.

## Changes

- Add `**kwargs` to `generate_response` in the 3 divergent providers and **forward** them into the underlying call (`params.update(kwargs)` for Together/Sarvam; `model.invoke(..., **kwargs)` for LangChain), matching the sibling providers
- Add a forwarding regression test to each (incl. a new `tests/llms/test_sarvam.py`)

## Testing

- Red-green proven: the new `*_forwards_extra_kwargs` tests fail with the exact `TypeError` on reverted source, pass with the fix
- `pytest` on touched + sibling files → 50 passed; `ruff check .` clean tree-wide, `ruff format --check` clean on touched files

### DIFF
--- a/mem0/llms/langchain.py
+++ b/mem0/llms/langchain.py
@@ -57,6 +57,7 @@ class LangchainLLM(LLMBase):
         response_format=None,
         tools: Optional[List[Dict]] = None,
         tool_choice: str = "auto",
+        **kwargs,
     ):
         """
         Generate a response based on the given messages using langchain_community.
@@ -66,6 +67,8 @@ class LangchainLLM(LLMBase):
             response_format (str or object, optional): Format of the response. Not used in Langchain.
             tools (list, optional): List of tools that the model can call.
             tool_choice (str, optional): Tool choice method.
+            **kwargs: Additional model parameters forwarded to the underlying LangChain
+                model's ``invoke`` (matches the ``LLMBase.generate_response`` contract).
 
         Returns:
             str: The generated response.
@@ -90,5 +93,5 @@ class LangchainLLM(LLMBase):
         if tools:
             langchain_model = langchain_model.bind_tools(tools=tools, tool_choice=tool_choice)
 
-        response: AIMessage = langchain_model.invoke(langchain_messages)
+        response: AIMessage = langchain_model.invoke(langchain_messages, **kwargs)
         return self._parse_response(response, tools)
--- a/mem0/llms/sarvam.py
+++ b/mem0/llms/sarvam.py
@@ -28,7 +28,7 @@ class SarvamLLM(LLMBase):
             getattr(self.config, "sarvam_base_url", None) or os.getenv("SARVAM_API_BASE") or "https://api.sarvam.ai/v1"
         )
 
-    def generate_response(self, messages: List[Dict[str, str]], response_format=None) -> str:
+    def generate_response(self, messages: List[Dict[str, str]], response_format=None, **kwargs) -> str:
         """
         Generate a response based on the given messages using Sarvam-M.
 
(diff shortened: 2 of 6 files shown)


kartik-mem0:

### What this PR does
Adds `**kwargs` to `generate_response` in `TogetherLLM`, `LangchainLLM`, and `SarvamLLM` — three providers that violated the `LLMBase` contract — and forwards them into the underlying calls, matching the pattern already used by OpenAI, Anthropic, DeepSeek, vLLM, and xAI.

### Strengths
- `together.py:81` — `params.update(kwargs)` lands before the `response_format`/`tools` conditional blocks, so named parameters always win over any duplicate kwarg. Correct layering.
- `sarvam.py:74-76` — `params.update(kwargs)` runs after the model-dict extraction block, so call-time kwargs override config-level settings. Correct precedence order.
- `tests/llms/test_sarvam.py` (new file) — SarvamLLM had zero test coverage before; this PR adds a basic happy-path test and the regression test. Both inspect the actual JSON payload sent to `requests.post`, verifying at the right boundary.
- All three regression tests (`test_generate_response_forwards_extra_kwargs`) are the exact right shape: they pass a kwarg, assert the underlying client/request received it. Red-green confirmed below.

### Issues
Minor (optional):
- `sarvam.py:31` — `response_format` is accepted as a named param but never applied to the request payload (pre-existing, not introduced here). Worth a follow-up issue.
- `together.py:81` — `params.update(kwargs)` before the `if response_format:` block means a caller passing `response_format` via kwargs would be overwritten by the explicit conditional. Low risk in practice (callers use the named param), but the ordering is slightly fragile. Not blocking.

### Evidence
- **CI:** Green — CLA signed; Vercel deploy auth is an unrelated infra check.
- **Targeted tests:** `pytest tests/llms/test_together.py tests/llms/test_langchain.py tests/llms/test_sarvam.py -v` → **11 passed in 1.43s**
- **Red-green proof:** With implementations reverted to `main`, all three `*_forwards_extra_kwargs` tests fail with `TypeError: generate_response() got an unexpected keyword argument 'frequency_penalty'`. Restored to PR head → 3 passed. Clean red→green on all three providers. ✓
- **Lint:** `ruff check` on the three changed implementation files → **All checks passed!**
- **Linked issue:** #5555 — PR directly addresses the three providers cited in the report.
- **Cross-package check:** TS SDK (`mem0-ts/`) does not include Together/LangChain/Sarvam providers; no counterpart to update.
- **Could NOT verify:** Live API behavior against real Together/Sarvam endpoints — code tracing and mock-based tests only, which is the appropriate level for this kind of contract fix.

### Assessment

Minimal, correctly scoped fix that brings three providers into compliance with the base-class contract. Tests prove the regression with clean red-green evidence. No functional changes to any code path beyond accepting and forwarding the extra kwargs.

kartik-mem0 (approved):

PR #5557, opened 2026-06-15
fix(memory): warn on batch embedding count mismatch in add pipeline

Adelagric:
## What

Adds a count-mismatch warning in the V3 `add()` pipeline (sync + async): when `embedding_model.embed_batch()` returns fewer vectors than the extracted memory texts, `dict(zip(mem_texts, mem_embeddings_list))` truncated to the shorter side and the tail memories were dropped with no signal to operators. The mismatch is now logged, mirroring the existing entity-embedding guard.

## Scope

Detection / observability only — no public API change. Per review, the richer partial-failure contract (`{results, failed}` return, `raise_on_partial_failure`, `EmbeddingErrorClass`, NaN/Inf/dim validation, coverage of the `infer=False` / procedural / insert-fallback paths, and TS parity #5509) is deferred to a follow-up where the API shape is a maintainer call. #5245 is fully resolved there.

## Tests

`TestAddPipelineMemoryEmbeddingCountGuard` (sync + async), parallel to the entity-embedding guard's tests. Red-green holds (revert the two `main.py` blocks → fail; restore → pass). `tests/memory/` + `tests/test_main.py`: 397 passed.

Refs #5245.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -962,6 +962,13 @@ class Memory(MemoryBase):
         mem_texts = [m.get("text", "") for m in extracted_memories if m.get("text")]
         try:
             mem_embeddings_list = self.embedding_model.embed_batch(mem_texts, "add")
+            if len(mem_embeddings_list) != len(mem_texts):
+                logger.warning(
+                    "embed_batch returned %d vectors for %d memory texts — "
+                    "unmatched texts are skipped rather than silently dropped",
+                    len(mem_embeddings_list),
+                    len(mem_texts),
+                )
             embed_map = dict(zip(mem_texts, mem_embeddings_list))
         except Exception:
             # Fallback: embed individually
@@ -2596,6 +2603,13 @@ class AsyncMemory(MemoryBase):
         mem_texts = [m.get("text", "") for m in extracted_memories if m.get("text")]
         try:
             mem_embeddings_list = await asyncio.to_thread(self.embedding_model.embed_batch, mem_texts, "add")
+            if len(mem_embeddings_list) != len(mem_texts):
+                logger.warning(
+                    "embed_batch returned %d vectors for %d memory texts — "
+                    "unmatched texts are skipped rather than silently dropped",
+                    len(mem_embeddings_list),
+                    len(mem_texts),
+                )
             embed_map = dict(zip(mem_texts, mem_embeddings_list))
         except Exception:
             embed_map = {}
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -1089,3 +1089,99 @@ class TestAddPipelineEntityEmbeddingCountGuard:
         assert any("padding/truncating" in r.message for r in caplog.records), (
             "expected count-mismatch warning was not emitted"
         )
+
+
+class TestAddPipelineMemoryEmbeddingCountGuard:
+    """A misbehaving embedder returning fewer vectors than the extracted memory
+    texts must surface the loss, not drop the unmatched texts silently.
+
+    `dict(zip(mem_texts, mem_embeddings_list))` truncates to the shorter side, so a
+    short `embed_batch` return dropped the tail memories with no signal to operators.
+    The add path now logs a count-mismatch warning (mirroring the entity-embedding
+    guard above) so the loss is observable; caller-visible surfacing is tracked
+    separately.
+    """
+
+    @pytest.fixture
+    def mock_memory(self, mocker):
+        _setup_mocks(mocker)
+        memory = Memory()
+        memory.config = mocker.MagicMock()
+        memory.config.custom_instructions = None
+        memory.config.custom_update_memory_prompt = None
+        memory.custom_instructions = None
+        memory.api_version = "v1.1"
+        memory.db.get_last_messages = MagicMock(return_value=[])
+        memory.db.save_messages = MagicMock()
+        memory.db.batch_add_history = MagicMock()
+        return memory
+
+    @pytest.fixture
+    def mock_async_memory(self, mocker):
+        _setup_mocks(mocker)
+        memory = AsyncMemory()
+        memory.config = mocker.MagicMock()
+        memory.config.custom_instructions = None
+        memory.config.custom_update_memory_prompt = None
+        memory.custom_instructions = None
+        memory.api_version = "v1.1"
+        memory.db.get_last_messages = MagicMock(return_value=[])
+        memory.db.save_messages = MagicMock()
+        memory.db.batch_add_history = MagicMock()
+        return memory
+
+    @staticmethod
+    def _short_memory_embed_batch(texts, memory_action="add"):
+        # The extracted memory texts come back short (1 vector for 2); any other
+        # batch (e.g. entities) embeds normally.
+        if any(t in ("fact one", "fact two") for t in texts):
+            return [[0.1] * 10]
+        return [[0.1] * 10 for _ in texts]
+
+    def test_sync_short_memory_embeddings_warn_not_silent(self, mock_memory, mocker, caplog):
+        mock_memory.llm.generate_response.return_value = (
+            '{"memory": [{"text": "fact one"}, {"text": "fact two"}]}'
+        )
+        mock_memory.embedding_model = Mock()
+        mock_memory.embedding_model.embed_batch = Mock(side_effect=self._short_memory_embed_batch)
+        mock_memory.embedding_model.embed = Mock(return_value=[0.1] * 10)
+        mocker.patch("mem0.memory.main.extract_entities_batch", return_value=[[], []])
+        mocker.patch("mem0.memory.main.capture_event")
+
+        with caplog.at_level(logging.WARNING):
+            result = mock_memory._add_to_vector_store(
+                messages=[{"role": "user", "content": "two facts"}],
+                metadata={},
+                filters={"user_id": "u1"},
+                infer=True,
+            )
+
+        assert any("memory texts" in r.message for r in caplog.records), (
+            "expected memory-embedding count-mismatch warning was not emitted"
+        )
+        # The aligned memory persists; the unmatched text is skipped, not misaligned.
+        assert len(result) == 1
+
+    @pytest.mark.asyncio
+    async def test_async_short_memory_embeddings_warn_not_silent(self, mock_async_memory, mocker, caplog):
+        mock_async_memory.llm.generate_response.return_value = (
+            '{"memory": [{"text": "fact one"}, {"text": "fact two"}]}'
+        )
+        mock_async_memory.embedding_model = Mock()
+        mock_async_memory.embedding_model.embed_batch = Mock(side_effect=self._short_memory_embed_batch)
+        mock_async_memory.embedding_model.embed = Mock(return_value=[0.1] * 10)
+        mocker.patch("mem0.memory.main.extract_entities_batch", return_value=[[], []])
+        mocker.patch("mem0.memory.main.capture_event")
+
+        with caplog.at_level(logging.WARNING):
+            result = await mock_async_memory._add_to_vector_store(
+                messages=[{"role": "user", "content": "two facts"}],
+                metadata={},
+                effective_filters={"user_id": "u1"},
+                infer=True,
+            )
+
+        assert any("memory texts" in r.message for r in caplog.records), (
+            "expected memory-embedding count-mismatch warning was not emitted"
+        )
+        assert len(result) == 1


kartik-mem0:
## Changes Requested

### Strengths
- **Root-cause fix is correct** — the bug (#5245) traced to Phase 4's `if text not in embed_map: continue` silently skipping texts that failed to embed. The PR correctly populates `embed_map` only for verified-good vectors and collects everything else in `failed`. This is exactly the right fix.
- **Additive return shape** — `{"results": [...], "failed": []}` leaves every existing caller that reads `result["results"]` unaffected. Verified against `server/main.py` and `openmemory/api` call sites.
- **`_validate_embedding` is well-designed** — handles numpy arrays via `.tolist()`, rejects complex numbers before `math.isfinite` would crash, catches NaN/Inf, and uses intra-batch dimension consistency rather than the (potentially stale) `embedding_dims` config value.
- **18 tests, and they bite** — red-green confirmed: revert `mem0/exceptions.py` + `mem0/memory/main.py` to `main` and the new tests immediately fail with `ImportError: cannot import name 'EmbeddingErrorClass'`. Restore the PR code: 18 passed.
- **Async mirrors sync exactly** — both `_add_to_vector_store` variants receive the `failed` out-parameter; both `add()` paths apply the same `raise_on_partial_failure` logic.

### Issues

1. `mem0/exceptions.py` — **EmbeddingErrorClass docstring makes a false claim about `error_code`.**  
   The docstring reads: *"Carried on `EmbeddingError.error_code` and on each `failed[]` entry's `error_class` field."*  
   When `raise_on_partial_failure=True` raises `EmbeddingError`, the call uses `error_code="EMBED_001"` (the default) — never `"provider_error"` or `"validation_error"`. The taxonomy values appear only in `failed[i]["error_class"]` (and `EmbeddingError.details["failed"][*]["error_class"]` when raised). Fix: remove the `error_code` clause from the docstring; clarify that `error_class` lives in `failed[]` entries, not on the exception's `error_code`.

2. `tests/memory/test_embedding_error_class.py` — **Async `add()` wrapper not tested at the `add()` level.**  
   `test_async_provider_and_validation_collected` calls `m._add_to_vector_store()` directly. There is no test that calls `await m.add(...)` and asserts the `{"results": ..., "failed": [...]}` shape, nor one that verifies `await m.add(raise_on_partial_failure=True)` raises `EmbeddingError`. The async add wrapper has its own `failed = []` and raise logic (lines 2342 + 2351 in the PR) and should be exercised independently. Please add: `test_async_add_returns_failed_in_result_by_default` and `test_async_add_raise_on_partial_failure_raises`, mirroring their sync counterparts.

3. `mem0/exceptions.py` — **`EmbeddingErrorClass` should be `class EmbeddingErrorClass(str, Enum)` to match repo convention.**  
   `mem0/configs/enums.py` uses `class MemoryType(Enum)` with string values for the same purpose — a closed vocabulary. Plain-class string constants allow silent divergence (a typo like `"provifer_error"` compiles and compares wrong silently). A `str, Enum` subclass keeps JSON serialization identical (`EmbeddingErrorClass.PROVIDER == "provider_error"` stays `True`) while adding membership checks and exhaustiveness. This aligns with the established pattern.

4. `mem0/exceptions.py` — **`EmbeddingErrorClass.INTERNAL` is declared but never used.**  
   `INTERNAL = "internal_error"` is defined as a catch-all but is assigned to zero `failed.append()` calls in either sync or async `_add_to_vector_store`, and has no test case. A dead constant in a closed vocabulary misleads callers about what values can actually appear in `failed[].error_class`. Either wire it to an outer `except Exception` catch (e.g., around the per-item fallback loop) or remove it. If you keep it, add a test.

5. `mem0/memory/main.py` — **`raise_on_partial_failure` position is asymmetric between sync and async signatures.**  
   In `Memory.add()` the parameter is terminal (after `prompt`). In `AsyncMemory.add()` it sits between `prompt` and `llm`. Both are keyword-only, so call semantics are identical — but the inconsistency will surface in docs, type stubs, and IDE autocomplete. Please align: place `raise_on_partial_failure` at the same relative position in both signatures (suggest: terminal in both, after `llm` in async).

#### Minor (optional)

- `mem0/memory/main.py` — `failed` parameter in `_add_to_vector_store` (sync and async) lacks a type annotation. Suggested: `failed: Optional[List[dict]] = None`.
- `tests/memory/test_embedding_error_class.py` — `_make_memory` / `_make_async_memory` helpers are ~30 lines each and nearly identical. A shared factory with a mode flag would reduce future maintenance surface (not a merge blocker).
- Consider adding a one-paragraph note in `docs/open-source/configuration.mdx` or the relevant feature page for the new `"failed"` return key and `raise_on_partial_failure` parameter, per REVIEW-PARAMS policy on public API surface.

### Evidence
- **CI:** Vercel (advisory, always fails on forks — ignored); CLA signed. Real CI checks not yet triggered — lint verified locally.
- **Targeted tests:** `pytest tests/memory/test_embedding_error_class.py -v` → **18 passed in 0.66s** (PR code)
- **Red-green proof:** Reverted `mem0/exceptions.py` + `mem0/memory/main.py` to `main` → `ImportError: cannot import name 'EmbeddingErrorClass'` (fail). Restored PR code → 18 passed.
- **Existing notice tests:** 110 passed (test_notices.py + test_temporal_usage_notice.py) — no regressions.
- **Lint:** `ruff check mem0/exceptions.py mem0/memory/main.py` → All checks passed.
- **Competing Python PRs:** #5176 (thalesfsp) targets the same root cause but is CONFLICTING with main (stale). #5177 and #5178 fix unrelated bugs (token truncation, LLM re-raise). TS PRs #5546/#5551/#5554 are TS-only. **#5557 is the right Python fix with no live competitor.**
- **Could NOT verify:** Live embedder behavior (NaN/Inf from a real provider in production) — code tracing only.

### Assessment
Ready to merge: **With fixes** — please address items 1–5; I'll re-review once you push.

The core is solid: the bug is correctly diagnosed and fixed, tests bite, and the additive return shape is the right design for Python's existing dict return. The five items above are all bounded — no architectural rethink needed.

Adelagric:
Thanks for the thorough review — all five addressed and pushed:

1. **Docstring** — dropped the `error_code` claim. It now states that `error_class` lives on `failed[]` entries (and under `EmbeddingError.details["failed"]` when raised), not on the exception's `error_code`.
2. **Async `add()`-level tests** — added `test_async_add_returns_failed_in_result_by_default` and `test_async_add_raise_on_partial_failure_raises`, mirroring the sync ones and driving `await m.add(...)` through the wrapper (its own `failed = []` + raise logic).
3. **`str, Enum`** — `class EmbeddingErrorClass(str, Enum)` now, matching `mem0/configs/enums.py`. JSON output is unchanged — `json.dumps({"error_class": EmbeddingErrorClass.PROVIDER})` → `{"error_class": "provider_error"}` (verified).
4. **`INTERNAL` wired** — the per-item fallback now wraps the validate/record step, so an unexpected failure there (not a provider throw, not a known validation reason) is recorded as `internal_error` instead of propagating. Added `test_internal_error_on_unexpected_processing_failure`. Both sync and async.
5. **Signature symmetry** — `raise_on_partial_failure` is now terminal in `AsyncMemory.add()` as well (after `llm`).

Minor: `failed` is typed `Optional[List[dict]]`. Left the two test factories un-merged for now (low churn). On docs — happy to add the note for the `"failed"` key and `raise_on_partial_failure`; `configuration.mdx` is config-focused and `python-quickstart` doesn't show the `add()` return, so point me at whichever page fits the REVIEW-PARAMS convention and I'll put it there.

`pytest tests/memory/test_embedding_error_class.py` → 21 passed; full `tests/memory/` + `tests/test_main.py` → 389 passed; `ruff` clean.

kartik-mem0:
Verified against f06c34d7c:

**Finding A (docstring)** — fixed. The new docstring correctly describes the field as `failed[].error_class` (and `EmbeddingError.details["failed"]` when raising), with no mention of `error_code`.

**Finding B (async `add()` tests)** — fixed. `test_async_add_returns_failed_in_result_by_default` and `test_async_add_raise_on_partial_failure_raises` are both present and passing. Full `AsyncMemory` fixture (lines 235–292) mirrors the sync one.

**Finding C (`str, Enum`)** — fixed. `class EmbeddingErrorClass(str, Enum)` at `exceptions.py:347`.

**Finding D (`INTERNAL` dead)** — fixed. `INTERNAL` is now assigned in both the sync and async individual-embed fallback paths (`main.py:967` / `main.py:2530`) and covered by `test_internal_error_on_unexpected_processing_failure`.

**Finding F (async signature parity)** — fixed (equivalent). `raise_on_partial_failure` is terminal in both signatures; the async `llm=` param that precedes it is async-only, so the placement is consistent.

---

Targeted tests: `pytest tests/memory/test_embedding_error_class.py` → **21 passed** (was 18; +3 async `add()` wrapper tests). Red-green confirmed — the new async tests fail with a clean ImportError on main, pass with the PR. Ruff clean on both changed files. CI green on 3.10/3.11/3.12.

One minor item remains open: `docs/` doesn't document the new `"failed"` key or `raise_on_partial_failure` parameter — worth a short follow-up docs PR, but not a merge blocker.

**LGTM — ready for squash-merge.** Thanks for the clean set of fixes.

Adelagric:
Thanks @kartik-mem0 — appreciate the two thorough passes (the red-green especially). I'll send the docs as a separate follow-up once this lands.

To put it in the right place per the REVIEW-PARAMS convention: where would you like the `"failed"` return key and `raise_on_partial_failure` documented? `configuration.mdx` reads config-only and `python-quickstart` doesn't show the `add()` return shape — is there a canonical `add()` / API-reference page (or a features page) you'd prefer I extend?

Adelagric:
@kartik-mem0 — rebased onto current `main`, so the CONFLICTING state is cleared (MERGEABLE now) and this is ready for the review pass.

**Conflict resolution.** The base had moved ~150 commits and the V3 `add()` pipeline was refactored in the interim, so the only real conflict was in `mem0/memory/main.py`, in the Phase-3 batch-embed fallback you traced above. Resolved keeping **both** sides — upstream's refactor (incl. the new `_vector_store_list_rows` helper) and the taxonomy: `_validate_embedding`, the `failed` out-parameter threaded through `add()` → `_add_to_vector_store` (sync + async), the `provider_error` / `validation_error` / `internal_error` classification at the point of detection, and the opt-in `raise_on_partial_failure`. The contract stays additive — `add()` returns `{"results": [...], "failed": [...]}`, with `failed` empty on the happy path, non-breaking for existing callers.

**Test delta.** `test_add_stores_expiration_date` (added upstream since this branch was cut) asserts the exact `_add_to_vector_store` call signature; I updated it for the additive `failed=[]` kwarg — no production-path change.

**Verification.** Squashed to a single commit. `tests/memory/` green (372 passed), including the 21 `error_class` tests and the async `add()`-level cases. (The optional-provider suites aren't exercised locally — they need their SDKs.)

kartik-mem0:
## Changes Requested: the batch-mismatch detection is a real fix, but as shipped this doesn't resolve #5245 for any current caller

Re-reviewing at head `f5d2b14c` (this has grown a fair bit since my June pass, so this is a fresh read of the current state). The core diagnosis is right and the batch-length reconciliation is genuinely good work. Two things stop it from actually closing #5245, plus a correctness hole in the new machinery. All findings below were traced against the code at this head.

### What's genuinely good
- **Correct root cause for the batch path.** The old `dict(zip(mem_texts, mem_embeddings_list))` silently misaligned/dropped items when `embed_batch` returned fewer vectors than inputs. The new index-checked loop (`mem0/memory/main.py:1005-1018`) detects the short return and surfaces each dropped text as `error_class=PROVIDER` instead of losing it. That's the real bug, fixed.
- **Sync and async kept symmetric** throughout (same out-param, same taxonomy, tests duplicated for `AsyncMemory`).
- **Stayed on stdlib** (`math.isfinite`) rather than pulling numpy into an unconditionally-imported core module. Right call given numpy is an optional-provider dep only.
- **Red-green holds for the `infer=True` path**: the regression tests pass, revert the impl and they fail, restore and they pass.

### Blocking (all verified against `f5d2b14c`)

**1. It's opt-in and a no-op by default, so #5245 is unchanged for every current deployment.** `raise_on_partial_failure` defaults to `False` (`main.py:766` sync / `:2456` async), its only use is `if raise_on_partial_failure and failed:` (`:875` / `:2543`), and nothing in the repo reads the new `failed` key: `server/main.py` returns the dict verbatim but the OSS/Platform callers, `openmemory/api/app/routers/memories.py` and `mcp_server.py`, only read `response["results"]` (grep for `"failed"` in `server/`+`openmemory/` returns nothing). So with default settings, on every current deployment mode, memories still vanish with zero surfaced error, exactly as before. The fix exists but no caller is wired to see it. Please either make the failure observable by default (raise, or have `server/`/`openmemory` read and expose `failed` in this same PR), or state explicitly that surfacing is opt-in and out of scope.

**2. Only the `infer=True` batch path is covered; the other add() paths still drop silently and now falsely report success.** `_validate_embedding` is called only at `main.py:1012/1029` (sync) and `:2687/2704` (async), all inside the LLM-extraction block. The other paths through `add()` are untouched:
- `add(..., infer=False)` embeds raw at `main.py:910` and persists via `_create_memory` with no validation and without touching `failed` (returns at `:922`).
- Procedural-memory creation (`_create_procedural_memory` -> `_create_memory`) is the same.
- The Phase 6 per-item insert fallback (`main.py:1093-1099`) catches `vector_store.insert()` failures and only `logger.error`s them, never appends to `failed`.

On all three, a corrupt/failed embed or insert is still dropped, and because the docstring now promises `failed` enumerates embedding failures, `failed: []` actively certifies success on those paths (a false negative, worse than the original silent drop). Fix: route `_create_memory` and the insert fallback through the same `_validate_embedding` + `failed` collection, or scope the docstring/contract to "only the `infer=True` extraction pipeline is covered." No test exercises `infer=False` or procedural at all today.

**3. The batch `try/except` can double-count a text into both `results` and `failed`.** `main.py:1003-1019`: the `try` wraps `embed_batch()` **and** the whole validation loop under one `except Exception`. If `_validate_embedding` raises partway through a batch that already populated `embed_map` (e.g. an exotic `numbers.Real` whose `__float__` raises inside `math.isfinite`), the `except` falls into a full per-item re-embed (`:1021-1040`) that never clears `embed_map`/`failed`. A text already in `embed_map` then re-embedded-and-failed lands in both `failed` and `results` (persisted). Narrow trigger, but real, and the one test that forces `_validate_embedding` to raise (`test_internal_error_on_unexpected_processing_failure`) only exercises the fallback path, never this one. Fix: narrow the `try` to wrap `embed_batch()` only; validate each returned vector outside it (or in a per-item `try`, as the fallback already does).

### Worth fixing, not blocking
- **Wrong-dim check no-ops for the most common shape.** `batch_dim` is seeded from the first accepted vector (`main.py:1015-1016`), not `embedding_model.config.embedding_dims`. For a single-extracted-memory `add()` (one message, one fact, the common case) `batch_dim` is still `None` when that lone vector is validated, so the dimension check never fires and a wrong-dim-but-finite vector passes. Prefer the configured `embedding_dims` when set, fall back to first-vector otherwise.
- **Python/TS divergence.** The TS `add()` (`mem0-ts/src/oss/src/memory/index.ts`) has the same swallowed per-item embed fallback and returns only `{results}`. It's tracked as issue #5509, but that's an open issue with no PR, so this PR ships the fix on the Python side only and the two SDK contracts diverge. Worth a linked follow-up so #5245 isn't marked done while TS still drops.
- Docs for the new `failed` key / `raise_on_partial_failure` aren't updated (`docs/` has no mention; `custom-instructions.mdx` still shows `{"results": [...]}`). The manually-wrapped dict literal introduced at ~`:1008` also trips `ruff format --check` (this one is PR-introduced, unlike the pre-existing `exceptions.py` debt).

### The design question for a maintainer
Beyond the bugs: reshaping `add()`'s return, adding a public `raise_on_partial_failure` kwarg, and adding a public 3-value `EmbeddingErrorClass` is a permanent API commitment on the SDK's most-used method, and #5245 didn't ask for it. That's a product call, not a routine bug fix. Consider splitting the PR: land the batch-length-mismatch detection (the actual root-cause fix, which is ready modulo the scope gaps above) on its own, and let a maintainer decide the taxonomy/kwarg/return-shape as a separate change that also lands in TS.

### Evidence (captured this run, worktree at `f5d2b14c`)
- Red-green: PR regression tests pass; revert `mem0/memory/main.py` -> tests fail; restore -> pass.
- Findings 1-3 + the two notes: traced directly in `mem0/memory/main.py` at the line numbers above; `raise_on_partial_failure` default and the missing `failed` reads in `server/`/`openmemory/` confirmed by grep.
- CI: only license/cla checks ran (real test matrix not triggered on this head); Gate B stands in as the verification of record.

### Assessment
Ready to merge: **Not yet.** The batch-mismatch detection is correct and worth keeping, but as shipped the fix is invisible by default and leaves `infer=False`, procedural, and insert-failure paths dropping silently while `failed: []` reports success. Wire the failure surface up (or scope it explicitly), cover the other paths, and narrow the batch `try`. Est. human time to land after that, and after a maintainer rules on the API-shape split: ~15-20 min.

Adelagric:
Thanks — thorough re-review, and I'm not going to contest the three blocking findings; they're right.

- **#3 (double-count) is real**: the `try` at `main.py:1003` wraps `embed_batch()` *and* the validation loop, so a `_validate_embedding` raise mid-batch falls into the per-item re-embed without clearing `embed_map`/`failed` — a text can land in both. Narrow the `try` to `embed_batch()` only, validate per-item outside it.
- **#2 is the one I most want gone**: with the docstring promising `failed` enumerates failures, `failed: []` certifies success on `infer=False`, procedural, and the Phase-6 insert fallback — paths this PR never touches. That's a worse contract than the original silent drop.
- **#1**: opt-in + no caller reading `failed` ⇒ #5245 is unchanged on every current deployment. Agreed.

And you're right that the return reshape + a public `raise_on_partial_failure` + a public `EmbeddingErrorClass` is an API call that's yours to make, not something to land under a bug fix. So — taking the split:

- **This PR, reduced to the root cause**: the index-checked batch loop (the actual `dict(zip(...))` misalignment), with #3 fixed and the scope stated explicitly as the `infer=True` extraction path — nothing certifying success on paths it doesn't cover.
- **Follow-up, for you to shape**: the `{results, failed}` return, the `raise_on_partial_failure` kwarg, the `EmbeddingErrorClass` taxonomy, the NaN/Inf/dim integrity checks, and coverage of `infer=False` / procedural / insert-fallback — landed with the TS side (#5509) so the two SDK contracts don't diverge.

One ruling before I cut it, since it's the product call you flagged: should the reduced PR **resolve #5245 by raising `EmbeddingError`** on the detected drop (default-on, faithful to the original #5249 direction, but a behavior change for callers currently getting partial results), or land as the **detection/correctness fix only** with #5245 resolved in the follow-up? Either works for me. I'll fold the non-blocking notes (dim-check `embedding_dims` seeding, docs, `ruff format`) into whichever surface it lands on.

kartik-mem0 (commented):
## Needs a human call — unchanged since the 2026-07-20 review

Re-verified at the same head (`f5d2b14c`, no commits since). The batch-mismatch fix holds: sandbox confirms zero data loss on the `infer=True` path, sync+async, oss-qdrant+oss-pgvector.

**Decide:** raise `EmbeddingError` by default on the reduced PR, or ship detection-only and resolve #5245 in the follow-up (your call per Adelagric's question above).
**Why this needs you:** a new public kwarg + enum + return-shape, plus a default-behavior change for existing callers, is a product call, not something this pipeline should settle.
**Confirmed still open (fresh, this run):** `main.py:910` (`infer=False` embeds raw, uncovered), `main.py:1093-1099`/`:2763-2773` (insert fallback, uncovered), `main.py:1003-1019` (double-count, by inspection). No caller in `server/`/`openmemory/` reads `failed` (grep both, empty).
**Your time:** ~5 min — the three findings are already conceded, this is just the ruling above.

- Gate 0: bug real on `main` @ 14393b59, `main.py:961-987` sync / `:2595-` async (dict(zip) misalignment, silent per-item-fallback drop).
- Gate B (fresh this run): 170/170 + 411/411 tests pass, ruff clean, red-green (revert `HEAD~1` -> `ImportError`; restore `HEAD` -> 21 passed).
- Gate B-live (sandbox, mandatory for this class): new scenario `sandbox/scenarios/issue5245_embed_fallback_silent_drop.py` (kept). FAIL on `main` (sync+async, oss-qdrant). PASS on this head across sync/async x oss-qdrant/oss-pgvector (4/4).
- Re-confirmed all 3 blocking findings from 2026-07-20 still apply against unchanged code: exact file:line and grep output in the evidence file.
- Not verified live this run: #3 (double-count) is code-inspection only, same limit as the original finding.
- Duplicate scan: no other open PR fixes Python #5245. TS has its own tracking issue #5509 with 3 unmerged competing PRs, unrelated to this thread.

Full trace: `.claude/babysit-evidence/pr-5557-2026-07-22.md`.

</details>

Adelagric:
Reduced to the **detection-only** option per your call. It's now a count-mismatch warning before `dict(zip(...))` in both `Memory.add()` and `AsyncMemory.add()`, mirroring the entity-embedding guard — the silent tail-drop is now observable, no new public API.

Red-green tests (`TestAddPipelineMemoryEmbeddingCountGuard`, sync + async) parallel to the entity guard's; `tests/memory/` + `tests/test_main.py` → 397 passed; `ruff check` clean. (`ruff format --check` flags pre-existing `frozenset({...})` blocks unrelated to this change — a formatter-version delta, left untouched.)

The `{results, failed}` contract + `raise_on_partial_failure` + `EmbeddingErrorClass` + integrity checks + the uncovered paths + TS parity (#5509) are staged for the follow-up where #5245 gets fully resolved and the API call is yours.

kartik-mem0 (commented):

Verified against `2d543d1` (new head): all three blocking findings from the 2026-07-22 review are moot, not patched around. `raise_on_partial_failure`, `EmbeddingErrorClass`, and `_validate_embedding` are gone entirely (grep, zero hits), so the opt-in kwarg, the false success on `infer=False`/procedural paths, and the double-count risk can no longer occur. What remains: one `logger.warning` guard before `dict(zip(...))`, symmetric in sync and async, correctly skipping the entity-guard's pad-with-None repair (would insert a None vector here).

**Before you merge:** nothing, self-contained.
**Follow-up:** #5245 stays open; `infer=False`/procedural/insert-fallback paths remain unfixed, honestly scoped as "Refs" not "Closes".
**Not verified:** sandbox, a pure log addition it cannot observe; caplog red-green substitutes.

Risk: low, log-only. Your time: ~5 min

- **Problem is real:** `main.py:962-967` (sync), `:2603-2608` (async): `dict(zip(...))` truncates silently on a short `embed_batch()` return.
- **Tests:** `pytest tests/memory/ tests/test_main.py -q` -> 397 passed.
- **Red-green:** revert impl file, 2 new tests fail; restore, pass.
- **CI:** only `license/cla` + `label` ran; non-blocking `Vercel` red, ignored.
- **Security:** cleared, no untrusted input, no injection/SSRF, log carries two integers only.
- **Bloat:** lean already; shared helper across 4 call sites considered, out of scope.
- **Linked issue:** #5245, partially (detection-only)

</details>

Adelagric:
@kartik-mem0 quick nudge on this — the babysit pass marked it ready-for-squash-merge on Jul 31 (the three findings all moot after the reduction, `tests/memory/` green, scoped honestly as `Refs #5245`). It's the detection step the taxonomy follow-up stacks on, so landing it unblocks that. Anything still needed from me, or is it just waiting on a merge window?

PR #5558, opened 2026-06-15
fix(ts-oss): preserve custom metadata on Memory.update() (closes #5479)

### LINKED ISSUE #5479: TS SDK: memory update drops custom metadata (parity with #5160, fixed for Python in #5179)

Bartok9:
## Summary

- The TypeScript OSS SDK's update path silently dropped custom metadata (`category`, `priority`, `source`, etc.) on a text-only `Memory.update()`.
- This brings the TS SDK to parity with the Python fix in #5179.

## Motivation

Closes #5479 (follow-up to #5160, opened at the maintainer's request in the #5179 review).

`updateMemory()` in `mem0-ts/src/oss/src/memory/index.ts` rebuilt the payload from scratch and hand-copied only the session-identifier fields (`user_id`/`agent_id`/`run_id`) from the existing memory. Because the vector store replaces the **full** payload on update, every other custom metadata field stored at `add()` time was lost. The old code also dropped `textLemmatized` entirely, leaving the BM25 index pointing at the previous text.

## Fix

Mirrors the Python approach in #5179:

1. Start `newMetadata` from a copy of the **full existing payload** so no custom field is lost.
2. Overlay caller-supplied `metadata` on top (intentional metadata updates still apply).
3. Stamp the system fields last so they always reflect the latest update: `data`, `hash`, `textLemmatized` (refreshed from the new text), `createdAt` (preserved), `updatedAt` (now).

Session identifiers carry over automatically as part of the existing payload, so the explicit per-field copies are no longer needed.

## Verification

- `npx jest --config jest.config.js src/oss/tests/memory.crud.test.ts` — **23 passed**, including the new regression test `preserves custom metadata on a text-only update`.
- `npx tsup` — build succeeds (CJS + ESM + DTS).
- `npx prettier --check` on both changed files — clean.

Diff is 2 files, +30/-10.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1721,21 +1721,20 @@ export class Memory {
     const embedding =
       existingEmbeddings[data] || (await this.embedder.embed(data));
 
-    const newMetadata = {
+    // Start from the full existing payload so no custom metadata
+    // (category, priority, source, etc.) is lost when the vector store
+    // replaces the whole payload on update. Caller-supplied metadata is
+    // overlaid on top, then the system fields are stamped last so they
+    // always reflect the latest update. Session identifiers carry over
+    // automatically as part of the existing payload.
+    const newMetadata: Record<string, any> = {
+      ...existingMemory.payload,
       ...metadata,
       data,
       hash: createHash("md5").update(data).digest("hex"),
+      textLemmatized: lemmatizeForBm25(data),
       createdAt: existingMemory.payload.createdAt,
       updatedAt: new Date().toISOString(),
-      ...(existingMemory.payload.user_id && {
-        user_id: existingMemory.payload.user_id,
-      }),
-      ...(existingMemory.payload.agent_id && {
-        agent_id: existingMemory.payload.agent_id,
-      }),
-      ...(existingMemory.payload.run_id && {
-        run_id: existingMemory.payload.run_id,
-      }),
     };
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks @Bartok9 — this was a clean TS-side fix and was already in good shape. Since then #5480 has brought the TypeScript update path to parity alongside the Python fix: the same `...existingMemory.payload` spread, plus a `textLemmatized` recompute so BM25/keyword search tracks the updated text rather than the stale lemmatization.

To keep a single canonical PR covering both SDKs, we're consolidating on #5480, which also resolves #5479. Closing this in favour of #5480 — genuinely appreciate the quick turnaround, and please reopen if #5480 doesn't land. 🙏

PR #5559, opened 2026-06-15
fix(ts-oss): stop Memory.get() leaking session ids into metadata

Bartok9:
## Summary

- `Memory.get()` leaked the session identifiers (`user_id`, `agent_id`, `run_id`) into `MemoryItem.metadata`, even though they are already surfaced as top-level fields on the returned object.
- Brings `get()` in line with `search()` and `getAll()`, which already exclude them correctly.

## Motivation

`get()` builds its metadata-exclusion set with **camelCase** keys:

```ts
const excludedKeys = new Set(["userId", "agentId", "runId", ...]);
```

But the vector-store payload stores session identifiers in **snake_case** (`user_id` / `agent_id` / `run_id`) — set in both `addToVectorStore()` and `createMemory()`. The mismatch means the exclusion never matches, so `get()` copies the session ids into `metadata`.

`search()` (step 9) and `getAll()` already use the snake_case keys in their exclusion sets, so they correctly keep session ids out of `metadata`. Only `get()` was inconsistent — a caller iterating `item.metadata` to enumerate user-supplied fields would unexpectedly find the internal scope keys mixed in.

## Fix

Use the snake_case keys (`user_id` / `agent_id` / `run_id`) in `get()`'s exclusion set so all three read paths agree. Genuine custom metadata is still surfaced unchanged.

## Verification

- Added regression test `does not leak session identifiers into metadata` in `memory.crud.test.ts`.
  - Confirmed **failing before** the fix (metadata contained `user_id`/`agent_id`/`run_id`), **passing after**.
- `npx jest --config jest.config.js src/oss/tests/memory.crud.test.ts` — **23 passed**.
- `npx tsup` — build succeeds (CJS + ESM + DTS).
- `npx prettier --check` on both changed files — clean.

Diff is 2 files, +31/-4.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1273,11 +1273,15 @@ export class Memory {
       metadata: {},
     };
 
-    // Add additional metadata
+    // Add additional metadata. Session identifiers are surfaced as
+    // top-level fields below, so exclude them here. The payload stores
+    // them snake_case (user_id/agent_id/run_id) — matching search() and
+    // getAll() — so the exclusion set must use the same casing. (A prior
+    // camelCase set leaked the session ids back into `metadata`.)
     const excludedKeys = new Set([
-      "userId",
-      "agentId",
-      "runId",
+      "user_id",
+      "agent_id",
+      "run_id",
       "hash",
       "data",
       "createdAt",
--- a/mem0-ts/src/oss/tests/memory.crud.test.ts
+++ b/mem0-ts/src/oss/tests/memory.crud.test.ts
@@ -128,6 +128,36 @@ describe("Memory - get()", () => {
     expect(item!.createdAt).toBeDefined();
     expect(new Date(item!.createdAt!).toString()).not.toBe("Invalid Date");
   });
+
+  // Regression: session identifiers must NOT leak into metadata.
+  // They are surfaced as top-level fields; get() previously used a
+  // camelCase exclusion set (userId/agentId/runId) that did not match
+  // the snake_case payload keys, so they leaked into metadata — unlike
+  // search() and getAll(), which use snake_case and excluded them.
+  test("does not leak session identifiers into metadata", async () => {
+    const agentId = `agent_${Date.now()}`;
+    const runId = `run_${Date.now()}`;
+    const addResult: SearchResult = await memory.add("Scope leak test", {
+      userId,
+      agentId,
+      runId,
+      infer: false,
+      metadata: { category: "work" },
+    });
+    const id = addResult.results[0].id;
+    const item: MemoryItem | null = await memory.get(id);
+
+    expect(item!.metadata).not.toHaveProperty("user_id");
+    expect(item!.metadata).not.toHaveProperty("agent_id");
+    expect(item!.metadata).not.toHaveProperty("run_id");
+    // Genuine custom metadata is still surfaced.
+    expect(item!.metadata).toMatchObject({ category: "work" });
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- Correct root-cause identification: the vector-store payload stores session identifiers as `user_id`/`agent_id`/`run_id` (snake_case), and the prior exclusion set used camelCase, so it never matched — the fix is minimal and exactly right (`index.ts:1154–1156`).
- Perfect symmetry: after this PR all three read methods (`get()`, `search()`, `getAll()`) have identical `excludedKeys` sets, including `attributedTo`. The alignment is complete.
- The regression test uses `infer: false` for a deterministic payload, exercises all three session-id variants, and verifies that genuine custom metadata (`category: "work"`) still flows through. The inline comment in production code explaining the snake_case requirement will prevent a future contributor from "fixing" it back.
- Python counterpart is clean — `mem0/memory/main.py` already uses snake_case throughout, so no cross-language fix needed.

### Issues

#### Important (should fix)
`memory.crud.test.ts:147–151` — the regression test covers only half the contract: it proves session IDs are absent from `metadata`, but does not assert they appear as top-level fields on the returned item. Those top-level fields come from a separate `filters` spread at `index.ts:1134–1138` and `1167`. If that spread were accidentally removed, the test would still pass and callers would silently lose `item.user_id` / `item.agent_id` / `item.run_id`. Three extra lines close the gap:

```ts
expect(item).toHaveProperty("user_id", userId);
expect(item).toHaveProperty("agent_id");
expect(item).toHaveProperty("run_id");
```

#### Minor (optional)
`MemoryItem` type in `src/oss/src/types/index.ts` does not declare `user_id`, `agent_id`, `run_id` as optional fields, so callers using TypeScript get implicit `any`-access. This is pre-existing across all three read methods — not introduced here — but worth a follow-up.

### Evidence
- **CI:** CI Gate pass; TS SDK checks pass on Node 20 + 22 (build + integration); Vercel = infra auth failure (not a code check, seen on all fork PRs)
- **Targeted tests:** `npx jest --config jest.config.js src/oss/tests/memory.crud.test.ts` → 23 passed in 9.6s
- **Red-green proof:** regression test FAILS on main (metadata.user_id present, received value = user id string), PASSES with fix ✓
- **Cross-package check:** Python `Memory.get()` uses snake_case exclusion set — no parity fix needed
- **Linked issue:** none
- **Could NOT verify:** live vector-store behavior beyond in-memory SQLite mock; CI integration run covers the hosted path

### Assessment
Ready to merge: **With fixes** — please add the three `toHaveProperty` assertions; the fix itself is correct and the scope is tight. Happy to re-review once you push.

Bartok9:
Thanks @kartik-mem0 — good catch on the incomplete contract test.

Rebased onto current `main` (prior tip had lost its merge-base after upstream history moved) and addressed the Important item on tip `5541d6ab`:

1. **Top-level session fields** — the regression now asserts:
   ```ts
   expect(item).toHaveProperty("user_id", userId);
   expect(item).toHaveProperty("agent_id", agentId);
   expect(item).toHaveProperty("run_id", runId);
   ```
   so a dropped `filters` spread at `get()` would fail the test, not just the metadata-exclusion half.

2. **Product fix unchanged in shape** — `get()` `excludedKeys` uses snake_case `user_id`/`agent_id`/`run_id` to match `search()`/`getAll()` (still the real root cause on `main`).

**Verify:** `pnpm exec jest src/oss/tests/memory.crud.test.ts -t "does not leak session identifiers"` → 1 passed.

Happy to take the Minor `MemoryItem` optional session fields as a follow-up if you want them typed here.

PR #5561, opened 2026-06-15
fix(vector_stores): return None from ChromaDB.get() for missing IDs

ly-wang19:
## What

`ChromaDB.get(vector_id)` ends with `self._parse_output(result)[0]`. When the id isn't found, Chroma's `collection.get` returns empty lists, so `_parse_output` returns `[]` and `[0]` raises **`IndexError`** instead of signaling "not found".

Every other vector store returns `None` for a missing id:
- `qdrant`: `return result[0] if result else None`
- `pgvector`: `if not result: return None`
- `faiss`: `if vector_id not in self.docstore: return None`

This brings ChromaDB to parity — return `None` when the id isn't found — and tightens the annotation to `Optional[OutputData]`.

## Test
Adds `tests/vector_stores/test_chroma.py::test_get_missing_vector_returns_none`: `get()` on an unknown id returns `None` rather than raising. Reverting the fix makes it fail with `IndexError`.

`pytest tests/vector_stores/test_chroma.py` → **24 passed**; `ruff check` clean.

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -187,7 +187,7 @@ class ChromaDB(VectorStoreBase):
         """
         self.collection.update(ids=vector_id, embeddings=vector, metadatas=payload)
 
-    def get(self, vector_id: str) -> OutputData:
+    def get(self, vector_id: str) -> Optional[OutputData]:
         """
         Retrieve a vector by ID.
 
@@ -195,10 +195,11 @@ class ChromaDB(VectorStoreBase):
             vector_id (str): ID of the vector to retrieve.
 
         Returns:
-            OutputData: Retrieved vector.
+            Optional[OutputData]: Retrieved vector, or None if the ID is not found.
         """
         result = self.collection.get(ids=[vector_id])
-        return self._parse_output(result)[0]
+        parsed = self._parse_output(result)
+        return parsed[0] if parsed else None
 
     def list_cols(self) -> List[chromadb.Collection]:
         """
--- a/tests/vector_stores/test_chroma.py
+++ b/tests/vector_stores/test_chroma.py
@@ -154,6 +154,16 @@ def test_get_vector(chromadb_instance):
     assert result.payload == {"name": "vector1"}
 
 
+def test_get_missing_vector_returns_none(chromadb_instance):
+    # Chroma returns empty lists for an unknown id; get() must return None
+    # rather than raising IndexError (parity with qdrant/pgvector/faiss).
+    chromadb_instance.collection.get.return_value = {"ids": [], "metadatas": []}
+
+    result = chromadb_instance.get(vector_id="does-not-exist")
+
+    assert result is None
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
`ChromaDB.get()` was raising `IndexError` when the requested ID wasn't found — Chroma returns empty lists for a miss, and `_parse_output(result)[0]` blows up on an empty list. This PR adds a `None` guard (`return parsed[0] if parsed else None`), tightens the return annotation to `Optional[OutputData]`, and adds a regression test. Brings ChromaDB to parity with every other vector store in the repo (qdrant, pgvector) that already returns `None` on miss.

### Strengths
- `chroma.py:198–200` — minimal two-line fix, idiomatic Python, no side effects
- `chroma.py:188` — `Optional[OutputData]` is accurate and uses an already-imported symbol
- `tests/vector_stores/test_chroma.py:157–166` — genuine regression test that mocks the exact empty-list response Chroma returns for a missing ID; would have caught the original `IndexError` (verified via red-green revert below)
- All six call sites in `main.py` / `setup.py` already guard on `None`, so the new return value is safe for all callers
- Commit title is valid Conventional Commits format

### Issues
Minor (optional):
1. `tests/vector_stores/test_chroma.py:158` — existing `test_get_vector` uses `assert_called_once_with(ids=["id1"])` but the new test omits it; cosmetic, not a correctness issue
2. `base.py:39` / `pgvector.py:449` / `qdrant.py:501` have the same annotation inconsistency (untyped base, wrong `-> OutputData` or `-> dict` on siblings that also return `None`) — pre-existing, out of scope here; a follow-up to fix the whole family would be welcome

### Duplicate note
PR #5534 (open) makes the same one-line impl fix but has no regression test and no type annotation. Recommend closing #5534 in favor of this one once merged.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 PASS · CI Gate PASS · CLA signed · Vercel = infra auth-gate (not a code check)
- **Targeted tests:** `pytest tests/vector_stores/test_chroma.py -v --tb=short` → **24 passed in 5.34s**
- **Red-green proof:** reverted `chroma.py` to main → `test_get_missing_vector_returns_none` fails with `IndexError: list index out of range` at `chroma.py:201` ✓ · restored fix → 1 passed ✓
- **Lint:** `ruff check mem0/vector_stores/chroma.py tests/vector_stores/test_chroma.py` → All checks passed
- **Cross-package check:** no TypeScript ChromaDB vector store in `mem0-ts/src/vector_stores/` — N/A
- **Linked issue:** none explicitly linked; bug is self-evident from code comparison
- **Could NOT verify:** live ChromaDB collection behavior — code-tracing and mock only; consistent with Chroma's documented API contract

### Assessment

Fix is correct, minimal, and well-tested. Red-green proof is clean. No Critical or Important findings survived the skeptic pass.

kartik-mem0 (approved):

PR #5562, opened 2026-06-15
fix(vector_stores): return None from get() for missing IDs (milvus/weaviate/supabase)

ly-wang19:
## What

`get(vector_id)` should return `None` for an unknown id — as `qdrant`, `pgvector`, and `faiss` already do — but three stores mishandled it:

- **milvus**: `result[0]` raised **`IndexError`** on the empty result.
- **weaviate**: `response.properties` raised **`AttributeError`** when `fetch_object_by_id` returned `None`.
- **supabase**: returned `[]`, contradicting its own `Optional[OutputData]` hint and "None if not found" docstring.

All three now return `None`. (Companion to #5561, which fixed the same issue in ChromaDB.)

## Test
- `test_milvus`: adds `test_get_missing_returns_none`.
- `test_supabase`: adds `test_get_missing_returns_none`.
- `test_weaviate`: **repairs** `test_get_not_found` — it mocked the wrong method (`data.get_by_id` vs the `query.fetch_object_by_id` the code actually uses) and never called `get()` or asserted, so the not-found path was never exercised. It now mocks `fetch_object_by_id → None` and asserts `get()` returns `None`.

`pytest tests/vector_stores/test_{milvus,weaviate,supabase}.py` → **39 passed**; reverting each fix makes the corresponding test fail (IndexError / AttributeError / `[] != None`). `ruff` clean.

### DIFF
--- a/mem0/vector_stores/milvus.py
+++ b/mem0/vector_stores/milvus.py
@@ -290,7 +290,7 @@ class MilvusDB(VectorStoreBase):
         schema = {"id": vector_id, "vectors": vector, "metadata": payload, "text": text}
         self.client.upsert(collection_name=self.collection_name, data=schema)
 
-    def get(self, vector_id):
+    def get(self, vector_id) -> Optional[OutputData]:
         """
         Retrieve a vector by ID.
 
@@ -298,9 +298,11 @@ class MilvusDB(VectorStoreBase):
             vector_id (str): ID of the vector to retrieve.
 
         Returns:
-            OutputData: Retrieved vector.
+            Optional[OutputData]: Retrieved vector, or None if the ID is not found.
         """
         result = self.client.get(collection_name=self.collection_name, ids=vector_id)
+        if not result:
+            return None
         output = OutputData(
             id=result[0].get("id", None),
             score=None,
--- a/mem0/vector_stores/supabase.py
+++ b/mem0/vector_stores/supabase.py
@@ -176,7 +176,7 @@ class Supabase(VectorStoreBase):
         """
         result = self.collection.fetch([(vector_id,)])
         if not result:
-            return []
+            return None
 
         record = result[0]
         return OutputData(id=str(record.id), score=None, payload=record.metadata)
--- a/mem0/vector_stores/weaviate.py
+++ b/mem0/vector_stores/weaviate.py
@@ -305,7 +305,7 @@ class Weaviate(VectorStoreBase):
                 existing_payload: Mapping[str, str] = existing_data
                 collection.data.update(uuid=vector_id, properties=existing_payload, vector=vector)
 
-    def get(self, vector_id):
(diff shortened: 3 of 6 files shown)


kartik-mem0:

### What this PR does
Fixes `get(vector_id)` in three vector stores — Milvus, Weaviate, and Supabase — so they return `None` for missing IDs rather than crashing or returning the wrong type. It's the natural companion to #5561 (ChromaDB, merged yesterday) and brings all four stores to the same contract that `qdrant`, `pgvector`, and `faiss` already implement.

### Strengths
- `milvus.py:303-304`: guard before the first array access — the right place, minimal change
- `supabase.py:179`: single `return []` → `return None`, closes the type contract lie without touching anything else
- `weaviate.py:325-327`: cleans up three dead commented-out debug lines as a bonus
- `test_weaviate.py::test_get_not_found` was previously a dead test — it mocked `data.get_by_id` but the production code calls `query.fetch_object_by_id`, so it never exercised any real path. The replacement correctly mocks `query.fetch_object_by_id.return_value = None` and actually asserts the return value
- `Optional[OutputData]` return annotations added to both Milvus and Weaviate; Supabase already had the right annotation
- No TS parity gap: Milvus and Weaviate have no TS counterparts; the TS Supabase store already returns `null` for missing IDs

### Issues
Minor (optional):
1. `VectorStoreBase.get()` in `base.py` still lacks a `-> Optional[OutputData]` return annotation — pre-existing across all stores, worth a separate cleanup
2. The Weaviate `test_get_not_found` replacement doesn't call `assert_called_once_with(...)` — the return value assertion is what matters here, so this is cosmetic

### Evidence
- **CI:** Python SDK 3.10 / 3.11 / 3.12 PASS, CI Gate PASS, CLA signed
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_{milvus,weaviate,supabase}.py -v` → **39 passed in 13.67s**
- **Red-green proof:**
  - Milvus: PASS → FAIL (`IndexError: list index out of range` at `milvus.py:305`) → PASS ✓
  - Supabase: PASS → FAIL (`AssertionError: assert [] is None`) → PASS ✓
  - Weaviate: PASS → FAIL (`AttributeError: 'NoneType' object has no attribute 'properties'`) → PASS ✓
- **Overlap with #5561:** zero — #5561 touched only `chroma.py`; confirmed on main before this review
- **Cross-package check:** TS Supabase already correct; Milvus/Weaviate have no TS counterparts
- **Lint:** `ruff check` clean on all 6 changed files
- **Could NOT verify:** live Milvus / Weaviate / Supabase SDK behavior (code-tracing only — consistent with PR mocks and library contracts)

### Assessment

Three independent bugs, three independent fixes, three independent red-green proofs. Everything checks out.

kartik-mem0 (approved):

PR #5574, opened 2026-06-16
fix(memory): initialize entity store before delete/update cleanup

Bartok9:
## Summary

- Entity cleanup on memory delete/update now runs in fresh SDK processes instead of being silently skipped.
- User-facing impact: deleting (or updating, which deletes the old memory) no longer leaves stale `linked_memory_ids` pointing at memories that no longer exist.

## Motivation

Closes #4863.

`Memory._remove_memory_from_entity_store()` and its `AsyncMemory` variant returned early whenever `self._entity_store is None`. But `entity_store` is **lazy-initialized via a property** — `_entity_store` stays `None` until the first code path touches it. In a fresh process where a delete/update happens before anything initializes the entity store, cleanup was skipped entirely, leaving entity rows linked to deleted memories. This weakens delete/erasure semantics and can surface stale relationships in later search/entity-boost flows.

## Root cause

```python
def _remove_memory_from_entity_store(self, memory_id, filters):
    if self._entity_store is None:   # ← bare attribute, not the property
        return                       # ← skips cleanup in a fresh process
    ...
    listed = self.entity_store.list(...)  # property would have initialized it
```

## Fix

Route cleanup through the `entity_store` property (which initializes on demand), wrapped in `try/except` so init failure degrades to the prior no-op and never breaks the primary delete/update path:

```python
try:
    entity_store = self.entity_store
except Exception as e:
    logger.warning(f"Entity store initialization failed for memory_id={memory_id}: {e}")
    return
```

Applied symmetrically to the sync and async variants.

## Verification

- `python -m pytest tests/memory/test_main.py` — **44 passed**
- Two new tests (sync + async) build a fresh-process instance (`_entity_store = None`) and assert cleanup runs.

## Real behavior proof

- **Behavior addressed:** on a fresh-process `Memory` (`_entity_store is None`), `_delete_memory()` skipped entity cleanup — the entity row keeps the deleted memory's id in `linked_memory_ids`.
- **Real environment:** Python 3.14, editable install of this branch.
- **Repro:** fresh `Memory` instance with `_entity_store = None` and a stub entity store whose row links `["mem-1", "mem-2"]`; delete `mem-1`.

Captured output **before** fix (current `main`):
```
entity cleanup calls: []
RESULT: cleanup SKIPPED (bug)
```

Captured output **after** fix (this branch):
```
entity cleanup calls: [('update', {'vector_id': 'entity-1', 'vector': [...], 'payload': {'data': 'Paris', 'linked_memory_ids': ['mem-2']}})]
RESULT: cleanup ran (fixed)
```

- **Regression test:** `tests/memory/test_main.py::test_delete_initializes_entity_store_before_cleanup` (+ async variant) — asserts cleanup runs against a fresh-process instance; fails without the fix.
- **What was NOT tested:** real vector-store backends (Qdrant/PGVector/etc.) — cleanup is exercised against the existing mocked entity store, consistent with the rest of this test module.

## Differential value

Salvage of #4864 by @shaun0927 (open since April, currently `CONFLICTING`). This PR is the same correct fix, rebased onto current `main` and conflict-resolved against the #5214 entity-boost tests that landed since. Recommend merging this; close #4864 in favor of it (original author credited via `Co-authored-by`).

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -2100,13 +2100,15 @@ class Memory(MemoryBase):
             self.vector_store = VectorStoreFactory.create(
                 self.config.vector_store.provider, self.config.vector_store.config
             )
-        # Reset entity store if initialized
-        if self._entity_store is not None:
-            try:
-                self._entity_store.reset()
-            except Exception as e:
-                logger.warning(f"Failed to reset entity store: {e}")
-            self._entity_store = None
+        # Reset entity store. Route through the lazy `entity_store` property so a
+        # fresh process (where `_entity_store` is still None) still wipes any
+        # persisted entity collection from a prior process, rather than skipping
+        # the reset entirely. try/except keeps reset non-fatal on init failure.
+        try:
+            self.entity_store.reset()
+        except Exception as e:
+            logger.warning(f"Failed to reset entity store: {e}")
+        self._entity_store = None
 
         capture_event("mem0.reset", self, {"sync_type": "sync"})
         display_first_run_notice(self, "sync", "reset")
@@ -3772,12 +3774,16 @@ class AsyncMemory(MemoryBase):
             self.config.vector_store.provider, self.config.vector_store.config
         )
 
-        if self._entity_store is not None:
-            try:
-                await asyncio.to_thread(self._entity_store.reset)
-            except Exception as e:
-                logger.warning(f"Failed to reset entity store: {e}")
-            self._entity_store = None
+        # Reset entity store. Route through the lazy `entity_store` property so a
+        # fresh process (where `_entity_store` is still None) still wipes any
+        # persisted entity collection from a prior process. The sync reset() does
+        # this too; the async path previously skipped entity-store cleanup
+        # entirely. try/except keeps reset non-fatal on init failure.
+        try:
+            await asyncio.to_thread(self.entity_store.reset)
+        except Exception as e:
+            logger.warning(f"Failed to reset entity store (async): {e}")
+        self._entity_store = None
 
         capture_event("mem0.reset", self, {"sync_type": "async"})
         await display_first_run_notice_async(self, "async", "reset")
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -2277,15 +2277,18 @@ class AsyncMemory(MemoryBase):
         concurrent _delete_memory coroutines each try to read-modify-write
         the same entity rows' linked_memory_ids lists.
         """
-        if self._entity_store is None:
+        try:
+            entity_store = self.entity_store
+        except Exception as e:
+            logger.warning(f"Entity store initialization failed during bulk clear (async): {e}")
             return
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         try:
-            listed = await asyncio.to_thread(self.entity_store.list, filters=search_filters, top_k=10000)
+            listed = await asyncio.to_thread(entity_store.list, filters=search_filters, top_k=10000)
             rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
             for row in rows or []:
                 try:
-                    await asyncio.to_thread(self.entity_store.delete, vector_id=row.id)
+                    await asyncio.to_thread(entity_store.delete, vector_id=row.id)
                 except Exception as e:
                     logger.debug(f"Bulk entity delete failed for id={row.id}: {e}")
         except Exception as e:
@@ -3530,8 +3533,7 @@ class AsyncMemory(MemoryBase):
 
         results = await asyncio.gather(*delete_tasks, return_exceptions=True)
 
-        if self._entity_store is not None:
-            await self._bulk_clear_entity_store(filters)
+        await self._bulk_clear_entity_store(filters)
 
         errors = [r for r in results if isinstance(r, BaseException)]
         if errors:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -632,16 +632,18 @@ class Memory(MemoryBase):
           - otherwise re-embed the entity text and update the payload
             (the vector store's update() requires a vector).
 
-        No-op if the entity store has never been initialized in this process.
         Errors on individual entities are swallowed at debug level; outer
         failures are swallowed at warning level so the primary delete/update
         path is never broken by entity cleanup.
         """
-        if self._entity_store is None:
+        try:
+            entity_store = self.entity_store
+        except Exception as e:
+            logger.warning(f"Entity store initialization failed for memory_id={memory_id}: {e}")
             return
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         try:
-            listed = self.entity_store.list(filters=search_filters, top_k=10000)
+            listed = entity_store.list(filters=search_filters, top_k=10000)
             rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
             for row in rows or []:
                 try:
@@ -652,7 +654,7 @@ class Memory(MemoryBase):
                     remaining = [mid for mid in linked if mid != memory_id]
                     if not remaining:
                         try:
-                            self.entity_store.delete(vector_id=row.id)
+                            entity_store.delete(vector_id=row.id)
                         except Exception as e:
                             logger.debug(f"Entity delete failed for id={row.id}: {e}")
                     else:
@@ -667,7 +669,7 @@ class Memory(MemoryBase):
                             continue
                         new_payload = {**payload, "linked_memory_ids": remaining}
                         try:
-                            self.entity_store.update(
+                            entity_store.update(
                                 vector_id=row.id,
                                 vector=vec,
                                 payload=new_payload,
@@ -2289,11 +2291,14 @@ class AsyncMemory(MemoryBase):
 
     async def _remove_memory_from_entity_store(self, memory_id, filters):
         """Async variant of `Memory._remove_memory_from_entity_store`."""
-        if self._entity_store is None:
+        try:
+            entity_store = self.entity_store
+        except Exception as e:
+            logger.warning(f"Entity store initialization failed for memory_id={memory_id} (async): {e}")
             return
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         try:
-            listed = await asyncio.to_thread(self.entity_store.list, filters=search_filters, top_k=10000)
+            listed = await asyncio.to_thread(entity_store.list, filters=search_filters, top_k=10000)
             rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
             for row in rows or []:
                 try:
@@ -2304,7 +2309,7 @@ class AsyncMemory(MemoryBase):
                     remaining = [mid for mid in linked if mid != memory_id]
                     if not remaining:
                         try:
-                            await asyncio.to_thread(self.entity_store.delete, vector_id=row.id)
+                            await asyncio.to_thread(entity_store.delete, vector_id=row.id)
                         except Exception as e:
                             logger.debug(f"Entity delete failed for id={row.id} (async): {e}")
                     else:
@@ -2320,7 +2325,7 @@ class AsyncMemory(MemoryBase):
                         new_payload = {**payload, "linked_memory_ids": remaining}
                         try:
                             await asyncio.to_thread(
-                                self.entity_store.update,
+                                entity_store.update,
                                 vector_id=row.id,
                                 vector=vec,
                                 payload=new_payload,
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -2,7 +2,7 @@ import logging
 import time
 from datetime import datetime, timezone
 from types import SimpleNamespace
-from unittest.mock import MagicMock, Mock, patch
+from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import pytest
 
@@ -699,6 +699,63 @@ def test_reset_initializes_entity_store_before_wipe():
     assert memory.entity_store_accesses == 1
     entity_store.reset.assert_called_once()
     assert memory._entity_store is None
+
+
+@pytest.mark.asyncio
+async def test_async_reset_initializes_entity_store_before_wipe():
+    """Async reset() must also route through the lazy property so a fresh
+    process wipes a persisted entity collection from a prior process."""
+    entity_store = MagicMock()
+
+    memory = _ProbeAsyncMemory.__new__(_ProbeAsyncMemory)
+    memory._entity_store = None
+    memory._dummy_entity_store = entity_store
+    memory.entity_store_accesses = 0
+    memory.vector_store = MagicMock()
+    memory.db = MagicMock()
+    memory.config = MagicMock()
+    memory.config.vector_store.provider = "dummy"
+    memory.config.vector_store.config = {}
+    memory.config.history_db_path = ":memory:"
+
+    with patch("mem0.memory.main.VectorStoreFactory") as vf, \
+         patch("mem0.memory.main.SQLiteManager"), \
+         patch("mem0.memory.main.capture_event"), \
+         patch("mem0.memory.main.display_first_run_notice_async", new=AsyncMock()):
+        vf.create.return_value = MagicMock()
+        await memory.reset()
+
+    assert memory.entity_store_accesses == 1
+    entity_store.reset.assert_called_once()
+    assert memory._entity_store is None
+
+
+@pytest.mark.asyncio
+async def test_async_delete_all_initializes_entity_store_before_bulk_clear():
+    """async delete_all() deletes each memory with skip_entity_cleanup=True and
+    relies on _bulk_clear_entity_store. In a fresh process that must still route
+    through the lazy entity_store property (issue #4863 sibling path)."""
+    stale_row = SimpleNamespace(id="entity-1", payload={"data": "Paris", "linked_memory_ids": ["mem-1"]})
+    entity_store = MagicMock()
+    entity_store.list.return_value = [stale_row]
+
+    memory = _ProbeAsyncMemory.__new__(_ProbeAsyncMemory)
+    memory._entity_store = None
+    memory._dummy_entity_store = entity_store
+    memory.entity_store_accesses = 0
+    memory.vector_store = MagicMock()
+    memory.vector_store.list.return_value = ([SimpleNamespace(id="mem-1")], None)
+    memory.db = MagicMock()
+
+    with patch("mem0.memory.main.process_telemetry_filters", return_value=(None, None)), \
+         patch("mem0.memory.main.capture_event"), \
+         patch("mem0.memory.main.detect_decay_usage_from_delete_all", return_value=None), \
+         patch("mem0.memory.main.display_first_run_notice_async", new=AsyncMock()):
+        await memory.delete_all(user_id="user-1")
+
+    assert memory.entity_store_accesses >= 1
+    entity_store.list.assert_called_once_with(filters={"user_id": "user-1"}, top_k=10000)
+    entity_store.delete.assert_called_once_with(vector_id="entity-1")
 
 
 def test_create_then_search_and_get_all_return_same_timestamps(mocker):
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -328,6 +328,20 @@ def _build_memory_instance(mocker, memory_cls):
     return memory
 
 
+class _ProbeMemory(Memory):
+    @property
+    def entity_store(self):
+        self.entity_store_accesses += 1
+        return self._dummy_entity_store
+
+
+class _ProbeAsyncMemory(AsyncMemory):
+    @property
+    def entity_store(self):
+        self.entity_store_accesses += 1
+        return self._dummy_entity_store
+
+
 def _assert_utc_timestamp(timestamp: str):
     parsed = datetime.fromisoformat(timestamp)
     assert parsed.tzinfo == timezone.utc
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Good work on this — the fix is correct, rebased cleanly, and the CLA is signed. One thing to address before merge.

**What's right**

The root cause is real: `_remove_memory_from_entity_store` was guarding with `if self._entity_store is None` (bare attribute) instead of calling `self.entity_store` (the lazy property). A fresh process that hasn't called `add()` will always have `_entity_store = None`, so every `delete()` or `update()` silently skipped entity cleanup and left stale `linked_memory_ids` behind.

The fix is correct: binding `entity_store = self.entity_store` at the top of the method forces the lazy initializer, the `try/except` preserves the "cleanup never breaks primary delete/update" contract, and the local variable avoids any re-initialization risk within the method body. Sync and async variants are fixed identically. Lint is clean.

Red-green proof on Python 3.11 (this run): both new tests fail on `main` (`entity_store_accesses == 0`) and pass on this branch. Full `tests/memory/test_main.py` is 44/44.

**Important — one gap to close**

The new tests (`test_delete_initializes_entity_store_before_cleanup` and its async sibling) set up a fixture where `linked_memory_ids: ["mem-1"]` contains only the memory being deleted, so `remaining = []` every time and the test always exercises the `entity_store.delete()` path.

The `else` branch — where other memory IDs are still present, the entity text is re-embedded, and `entity_store.update()` is called — is never reached. It also can't be reached with the current probe setup because `embedding_model` is not set on the instances (an `AttributeError` would surface if that path ran). That branch is the logic that keeps *partial* entity relationships consistent after a deletion, and it has no coverage in this PR or in the existing suite.

A minimal fix: add a second fixture row (or a second test case) where `linked_memory_ids: ["mem-1", "mem-2"]` and assert that `entity_store.update()` is called with `linked_memory_ids: ["mem-2"]` in the new payload. You'll need to set `memory.embedding_model = MagicMock()` and configure its `embed()` return value for that fixture. Without this, a regression in the update path would go undetected.

**vs. #4864**

This is a rebase of the original fix from #4864. The core logic is identical. This PR is strictly more mergeable: rebased on current `main`, CLA signed, CI green. The same update-branch test gap exists in #4864 as well. Recommend closing #4864 in favor of this one once the test is added.

Add the update-branch test and this is ready to squash-merge.

kartik-mem0:
## Code Review

**Verdict: LGTM** — ready to squash-merge.

### Strengths

- **Correct root-cause fix.** `_remove_memory_from_entity_store` was guarding on `self._entity_store is None`, which is a bare attribute that bypasses the lazy-initializing `entity_store` property. In a fresh process where delete/update fires before any other code path touches the entity store, `_entity_store` stays `None` and cleanup silently exits. The fix routes through the property, preserving exactly the right semantics.
- **Symmetrically applied.** Both the sync (`Memory`, line 558) and async (`AsyncMemory`, line 2113) variants are fixed identically. The local `entity_store` variable is used consistently throughout the method body — no risk of mid-method re-initialization divergence.
- **Graceful degradation preserved.** The try/except wrapping means an init failure (config error, connection refused) logs a warning and returns, keeping the "primary delete path is never broken" contract that was already in the docstring.
- **Aligned with existing pattern.** `_upsert_entity` at line 1626 already calls `self.entity_store` (the property) directly — this fix brings `_remove_memory_from_entity_store` into alignment with that established pattern.
- **Correct scoping of the change.** `_bulk_clear_entity_store` (used by `delete_all`) correctly retains `if self._entity_store is None: return` — no point initializing the store just to immediately delete everything from it. That guard stays right.
- **TS SDK has no parity gap.** `mem0-ts/src/oss/src/memory/index.ts` already wraps `getEntityStore()` in try/catch at line 311 — the TS side was already correct; no TS changes needed.
- **Clean duplicate resolution.** PR #4864 (original fix) is CONFLICTING on current main due to entity-boost tests that landed since. This PR is a conflict-resolved rebase with proper Co-authored-by attribution. Recommend closing #4864 in favor of this PR.

### Issues

**Minor — async test only exercises the delete (empty-list) path**
The async regression test asserts cleanup runs when `linked_memory_ids` becomes empty after removing the target memory — it doesn't test the remaining>0 branch (entity update with re-embed). The sync test covers both paths. Given the implementation is symmetric and the fix is a single property access, this is a very minor gap, not a blocker.

### Evidence

- **CI:** Python SDK (3.10, 3.11, 3.12) all pass. Vercel: advisory/fork-perms failure, not merge-blocking.
- **Red-green proof (run this session):**
  - WITH fix: `test_delete_initializes_entity_store_before_cleanup` PASS, `test_async_delete_initializes_entity_store_before_cleanup` PASS (2 passed in 2.94s)
  - Revert to main impl: both FAIL (`entity_store_accesses == 0`, expected 1)
  - Restore PR impl: both PASS again
- **Full test file:** 44 passed, 0 failures, 0 regressions.
- **Ruff:** all checks passed on changed files.
- **What was NOT verified:** real vector-store backends — consistent with the rest of this test module using mocks.

### Assessment

Ready to squash-merge (`gh pr merge --squash`). Please close PR #4864 after merging — it is superseded by this rebase.

kartik-mem0:
## Changes Requested

Sorry for the back-and-forth — took another look before merge and found a couple of sibling-path gaps I missed on the 19th.

### Strengths
- Root cause fix is correct: routes through the lazy `entity_store` property instead of the raw `_entity_store` attribute, so a fresh process no longer skips cleanup — `mem0/memory/main.py:557-561` (sync) and `:2083-2087` (async).
- Applied symmetrically to both the delete path (`_delete_memory`, sync `:1915` / async `:3454`) and the update path (`_update_memory`, sync `:1884` / async `:3421`) — both route through the same fixed helper.
- The `try`/`except` preserves the "cleanup never breaks primary delete/update" contract, and matches the existing pattern in `_upsert_entity`.
- `delete_all()` needs no changes — sync and async both loop through `_delete_memory()` per item, so it inherits this fix automatically. (Correcting my June 19 note here: there's no separate bulk-clear helper in this codebase — I misdescribed that. Nothing to do for delete_all.)
- TS SDK already correct: `mem0-ts/src/oss/src/memory/index.ts:307-317` already wraps `getEntityStore()` in try/catch — no parity gap.

### Issues
#### Important (should fix)
1. `mem0/memory/main.py:1943-1948` (sync `reset()`) — still guards on the raw `self._entity_store is not None` attribute. If `reset()` runs before anything else touches the entity store in-process, the persisted entity collection from a prior process is never wiped, so stale entity data survives a call that's supposed to be a full reset. Same fix as this PR: route through `self.entity_store` inside a try/except.
2. `mem0/memory/main.py:3458-3484` (async `reset()`) — doesn't touch the entity store at all, so async reset never clears it, not even the guarded attempt the sync path makes.
3. `mem0/memory/main.py:557-561` / `:2083-2087` — the new init-failure `except` branch has no test coverage. `_ProbeMemory`/`_ProbeAsyncMemory` (`tests/memory/test_main.py:236-247`) override `entity_store` to always succeed, so a regression in "entity-store init errors never break delete/update" wouldn't be caught.
4. `tests/memory/test_main.py:341` and `:369` — both new tests use the same single-entry fixture (`linked_memory_ids: ["mem-1"]`), so `remaining` is always `[]` and only the delete branch runs. The `remaining > 0` branch (re-embed + `entity_store.update()`) — flagged on June 16 — still has no coverage anywhere in the suite; my June 19 comment saying the sync test covers both paths was wrong.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 green, CI Gate green — unchanged.
- **Targeted tests (this run):** `pytest tests/memory/test_main.py -k entity_store_before_cleanup -v` → 2 passed in 0.65s.
- **Red-green proof (this run):** reverted `mem0/memory/main.py` only to `origin/main`, tests untouched → both new tests FAIL (`entity_store_accesses == 0`, expected 1). Restored the PR's `main.py` → 2 passed again. Full `tests/memory/test_main.py`: 44 passed, 0 regressions.
- **Ruff:** `ruff check mem0/memory/main.py tests/memory/test_main.py` → all checks passed.
- **Could NOT verify:** real vector-store backends (mocks only, consistent with the rest of this test module).

### Assessment
Ready to merge: **With fixes** — the delete/update fix itself is solid and provable as-is. Please also patch the two `reset()` call sites and add the missing except-branch and update-branch test coverage, then I'll re-review. Recommend closing #4864 in favor of this PR either way, since it doesn't have a mergeable path.

Bartok9:
Thanks for the thorough re-review — all four points addressed in the latest push.

**1 & 2 — `reset()` entity-store gaps (both fixed)**
- Sync `reset()` (`mem0/memory/main.py`): now routes through the lazy `self.entity_store` property inside a `try/except` instead of guarding on the bare `self._entity_store is not None` attribute, so a fresh process that calls `reset()` before touching the entity store still wipes a persisted collection from a prior process. `self._entity_store = None` afterward as before.
- Async `reset()`: previously never touched the entity store at all — now mirrors the sync path (`await asyncio.to_thread(self.entity_store.reset)`, non-fatal `try/except`, then `_entity_store = None`).

**3 — init-failure except-branch coverage**
Added `test_delete_survives_entity_store_init_failure` (+ async) using a new `_FailingEntityStoreMemory`/`_FailingEntityStoreAsyncMemory` whose `entity_store` property raises. Asserts the primary delete still runs (`vector_store.delete(vector_id="mem-1")`) and no exception propagates — locks in the "entity-store init errors never break delete/update" contract.

**4 — `remaining > 0` update-branch coverage**
Added `test_delete_updates_entity_when_other_memories_remain` (+ async) with `linked_memory_ids: ["mem-1", "mem-2"]`. Sets `embedding_model = MagicMock()` with a stubbed `embed()`, then asserts `entity_store.delete` is NOT called, `embed("Paris", "update")` is called, and `entity_store.update(...)` receives `linked_memory_ids: ["mem-2"]`. This is the partial-relationship path that had no coverage anywhere in the suite.

**Verification (this run)**
- `pytest tests/memory/test_main.py` → 49 passed (44 + 5 new).
- Red-green on the reset fix: revert `main.py` reset() only → `test_reset_initializes_entity_store_before_wipe` FAILS (`entity_store_accesses == 0`); restore → passes.
- `ruff check mem0/memory/main.py tests/memory/test_main.py` → all checks passed.

Still recommend closing #4864 in favor of this one. Ready for re-review.

Bartok9:
Rebased onto current `main` to clear the conflict — PR now shows **MERGEABLE**.

One notable resolution: `main` gained its own async `reset()` entity-store cleanup since, but guarded on the bare `self._entity_store is not None` attribute (the exact fresh-process gap this PR fixes). Resolved in favor of routing through the lazy `entity_store` property inside a non-fatal try/except, consistent with the sync path and the rest of this PR.

Re-verified after rebase: `ruff check` clean, `pytest tests/memory/test_main.py` → 54 passed (including the reset/init-failure/update-branch tests added in the last push). Ready for re-review.

Bartok9:
Thanks for the thorough re-review — all four items are addressed. Two commits on this branch:

**`253543bb` — reset() sibling paths + missing test coverage**
- **Item 1 (sync `reset()`):** now routes through the `entity_store` property inside a try/except instead of guarding on the raw `self._entity_store` attribute, so a fresh in-process `reset()` wipes a persisted entity collection from a prior process. Also resets `self._entity_store = None` after.
- **Item 2 (async `reset()`):** now clears the entity store too (was previously untouched), mirroring the sync path via `asyncio.to_thread(self.entity_store.reset)`.
- **Item 3 (except-branch coverage):** added `_FailingEntityStoreMemory` / `_FailingEntityStoreAsyncMemory` probes whose `entity_store` raises on init — `test_delete_survives_entity_store_init_failure` (sync + async) proves the primary delete still completes. Added `test_reset_initializes_entity_store_before_wipe`.
- **Item 4 (update-branch coverage):** added `test_delete_updates_entity_when_other_memories_remain` (sync + async) using a two-entry fixture (`["mem-1", "mem-2"]`), so `remaining > 0` exercises the re-embed + `entity_store.update()` path and asserts `linked_memory_ids == ["mem-2"]`.

**`ca1a62f1` — fix pre-existing `test_collection_name_preserved_after_reset`**
That test was the red CI job. It inspected the *last* `VectorStoreFactory.create` call, which is now legitimately the entity store (`mem0_entities`) after this fix. Reworked it to assert the primary collection name appears among the recreated collections, rather than assuming create-call ordering.

Verified locally: `tests/memory/test_main.py` + `tests/test_memory.py` green (only pre-existing `langchain_core`-not-installed skips in my local env, which CI has). `ruff check` clean. Agreed on closing #4864 in favor of this PR.

kartik-mem0:
## Changes Requested

Verified each item from the July 2 review directly against `ca1a62f1` — three of four are genuinely fixed, one is code-complete but test-incomplete, and a new sibling-path gap surfaced that needs the same treatment before this closes #4863 in full.

### What's resolved since the last round
- **Except-branch coverage:** fixed. `test_delete_survives_entity_store_init_failure` (sync + async) uses a genuinely-raising `entity_store` property and asserts the primary delete still completes. Red-green reconfirmed: fails against pre-PR `main`, passes with this diff.
- **`remaining > 0` update-branch coverage:** fixed. `test_delete_updates_entity_when_other_memories_remain` (sync + async) asserts `linked_memory_ids == ["mem-2"]` after delete — a real payload check. Red-green reconfirmed.
- **`reset()` sibling anti-pattern, code:** fixed for both sync and async — both now route through the lazy `entity_store` property inside try/except instead of the raw `_entity_store` attribute check.
- **`reset()` sibling anti-pattern, tests:** partial. Only `test_reset_initializes_entity_store_before_wipe` (sync) exists — no async-reset regression test anywhere in the suite. `test_async_memory_reset_clears_messages_table` calls `await memory.reset()` but only asserts message/history-table clearing, never entity-store routing, so a regression on the async side would pass silently.

### Strengths
- Root-cause fix stays correct and symmetric across all four original call sites (`main.py:621`, `:2270`, `:2080`, `:3743`).
- `tests/test_memory.py::test_collection_name_preserved_after_reset` correctly reworked from a brittle "last create call" assertion to "primary name appears among all created collections" — needed since `reset()` now legitimately creates an entity collection too.
- CI green on all real checks; 112 passed locally across both test files; ruff clean.

### Issues

#### Critical (must fix)
1. `mem0/memory/main.py:3486-3487` (`AsyncMemory.delete_all`) + `:2249-2250` (`_bulk_clear_entity_store`) — both still gate entity cleanup on the raw `_entity_store` attribute, not the lazy property. Every memory in `delete_all` is deleted with `skip_entity_cleanup=True` (`:3482`), so cleanup depends entirely on `_bulk_clear_entity_store` running — in a fresh process it never does, since nothing upstream touches the `entity_store` property first. This reproduces #4863's exact failure mode in a delete-family sibling method: confirmed live, a fresh-process `AsyncMemory.delete_all()` performs zero entity-store accesses and never calls `.list()`/`.delete()` on it, while sync `delete_all()` (no `skip_entity_cleanup` on its `_delete_memory` signature) correctly inherits the fix as a control. Not a regression you introduced — `_bulk_clear_entity_store` was added by #5553 after this PR was opened (2026-06-18 vs. 2026-06-16) — but it's in this PR's current mergeable head and squarely within #4863's stated scope, so it needs the same fix before the issue is fully closed: same shape already proven at the other 4 sites (try `self.entity_store`, except → warn + no-op).

#### Important (should fix)
1. Add the missing async-`reset()` regression test — mirror `test_reset_initializes_entity_store_before_wipe` with `_ProbeAsyncMemory` + `await`. Closes the last open half of the `reset()` item from the last round.
2. `main.py:621-625, 2269-2273, 2080-2084, 3743-3747` — the "try entity_store, except warn" shape is hand-copied 4 times with no shared helper; plausibly why the 5th, structurally-identical `delete_all` site went unfixed across four review rounds. Two thin helpers (bail-out shape for delete/remove, fall-through shape for reset) would close the whack-a-mole risk for good.

#### Minor
1. `mem0-ts/src/oss/src/memory/index.ts:1668` — TS `reset()` still has the pre-fix raw `if (this._entityStore)` check (TS's own `_removeMemoryFromEntityStore` is already correct). Pre-existing, out of scope for this Python-only PR — worth its own TS issue so it isn't lost.

### Evidence
- **CI:** GREEN — Python SDK build 3.10/3.11/3.12, changelog_check, check_changes, CI Gate all pass at `ca1a62f1`. (Vercel is red but that's a fork-permission artifact, not a real check.)
- **Targeted tests:** `hatch -e dev_py_3_11 run pytest tests/memory/test_main.py tests/test_memory.py` → 112 passed.
- **Red-green proof:** confirmed for all 7 new/changed tests — fail against pre-PR `main.py`, pass against this PR's implementation.
- **Linked issue:** #4863 — root cause correctly diagnosed and fixed for `_remove_memory_from_entity_store` and `reset()` (sync+async); not yet fixed for `delete_all`'s bulk-clear path (Critical item above).
- **Duplicate cluster:** #4864 still OPEN/CONFLICTING; agreed to close in favor of this PR once merge-ready.
- **Could NOT verify:** live entity-store backend behavior (mocks only, consistent with the rest of this test module).

### Assessment
Ready to merge: **With fixes** — the `delete_all` gap needs the same one-line-shape fix already proven out at the other 4 call sites in this diff; the async-reset test is a quick add alongside it. Good progress this round — happy to take another look once pushed.

Bartok9:
Thanks — pushed `c0ec7e15` addressing both items.

**Critical — `delete_all` bulk-clear path (fixed)**
`_bulk_clear_entity_store` (async) now routes through the lazy `self.entity_store` property inside a `try/except` instead of gating on the raw `self._entity_store` attribute, and the outer `if self._entity_store is not None:` guard in `AsyncMemory.delete_all` is dropped so the call is never short-circuited in a fresh process. Same proven shape as the other 4 sites (try `self.entity_store`, except → warn + no-op). This closes the exact #4863 failure mode you reproduced: a fresh-process `delete_all()` now performs entity-store cleanup instead of zero accesses.

**Important — async-`reset()` regression test (added)**
Added `test_async_reset_initializes_entity_store_before_wipe` (mirrors the sync `_ProbeAsyncMemory` test with `await`), so an async-reset regression on the entity-store routing would now fail loudly. Also added `test_async_delete_all_initializes_entity_store_before_bulk_clear`, which asserts a fresh-process `delete_all()` routes through the lazy property and calls `entity_store.delete(vector_id=...)` — red-green: fails against the pre-fix guard, passes now.

**Verification (this run)**
- `pytest tests/memory/test_main.py` → 56 passed (2 new).
- `ruff check mem0/memory/main.py tests/memory/test_main.py` → all checks passed.

**Not addressed (agreed out of scope here):**
- The shared-helper refactor (Important #2) — happy to do it in a follow-up if you'd prefer it in this PR; kept this diff minimal/targeted to the reported gap.
- TS `reset()` raw `if (this._entityStore)` (Minor) — Python-only PR; worth a separate TS issue so it isn't lost.

Still recommend closing #4864 in favor of this one. Ready for re-review.

kartik-mem0:
Went through the new commits against the `delete_all` gap flagged last round, and it's fixed correctly — the same way as everywhere else in this file. `_bulk_clear_entity_store` (`mem0/memory/main.py:2246`) now wraps `entity_store = self.entity_store` in a try/except instead of checking the raw `self._entity_store` attribute, and the outer `if self._entity_store is not None:` guard around the call site in async `delete_all` (`main.py:3496`) is gone entirely — it just calls `await self._bulk_clear_entity_store(filters)` unconditionally now. That's the right shape: no double-guarding, consistent with the other four call sites this PR already fixed. In a freshly constructed process, `AsyncMemory.delete_all()` will now actually clean up the entity store instead of silently skipping it.

I didn't just take the diff's word for it. I reverted `main.py` to the previous head (`ca1a62f1`) in a worktree and re-ran the new test — `test_async_delete_all_initializes_entity_store_before_bulk_clear` (`tests/memory/test_main.py:661`) fails against the old code (`entity_store_accesses` stays `0`, the `list`/`delete` calls never fire), and passes once the fix is back. So this isn't just a mechanism test, it genuinely catches the bug from #4863's sibling path.

The async-reset test you added (`test_async_reset_initializes_entity_store_before_wipe`, `tests/memory/test_main.py:632`) closes the coverage gap from last time too — mirrors the sync version, checks `entity_store.reset` gets called and `_entity_store` is `None` afterward.

One thing I want to call out because it's easy to miss and you clearly didn't: fixing `delete_all` to always call `_bulk_clear_entity_store` means a second `VectorStoreFactory.create()` call now always fires during `delete_all`, which would have quietly broken the existing `test_async_delete_all_continues_on_partial_failure` test (its old single-mock setup would've handed that second `create()` call the *same* mock as the primary store, inflating `mock_vector_store.delete.call_count` and failing the `== 2` assertion). You caught that yourself and fixed it in the follow-up commit with `mock_vector_factory.side_effect = [mock_vector_store, mock_entity_store]` (`tests/test_memory.py:978`). That's exactly the kind of collateral-damage check I'd have flagged if you hadn't already done it — good catch.

What I ran this round, on `a014669e`:
- `pytest tests/memory/test_main.py tests/test_memory.py` — 114 passed
- Red-green on both new tests as described above (fails on `ca1a62f1`, passes on `a014669e`)
- `ruff check` on all three changed files — clean
- CI on the real required checks (build_mem0 3.10/3.11/3.12, changelog_check, check_changes, CI Gate) — all green; Vercel is still red but that's the usual fork-permissions noise, not a real check

Two things from last round I'm not blocking on, just flagging so they don't get lost:
- The shared "try entity_store, except warn" helper is still hand-copied across five spots now instead of extracted. You mentioned being open to doing this as a follow-up — I think that's the right call, no need to bundle it here.
- TS `reset()` (`mem0-ts/src/oss/src/memory/index.ts:1668`) still has the raw `if (this._entityStore)` check that Python's `reset()` used to have. Pre-existing, out of scope for a Python-only PR, worth its own issue whenever someone picks up TS parity.

Also, once this lands, #4864 can be closed as a duplicate — I know you'd already agreed to that.

This is ready for a squash-merge from my side.

Bartok9:
Thanks for the detailed final pass — and for catching that you re-verified the `delete_all` red-green in a worktree rather than trusting the diff. Glad the collateral `test_async_delete_all_continues_on_partial_failure` fix (`side_effect = [mock_vector_store, mock_entity_store]`) landed the way you'd have wanted.

On the two flagged non-blockers:
- **Shared helper refactor** — I'll open a focused follow-up PR that extracts the two shapes (bail-out for delete/remove, fall-through for reset) into thin helpers and rewires all five call sites, so the whack-a-mole risk is closed for good. Keeping it out of this diff as agreed.
- **TS `reset()` raw `if (this._entityStore)`** — I'll file a separate TS-parity issue referencing `mem0-ts/src/oss/src/memory/index.ts:1668` so it isn't lost.

Ready for squash-merge whenever you are; will close #4864 as duplicate on merge (original author credited via Co-authored-by).

Bartok9:
Rebased onto current `main` to clear the CONFLICTING state after the identity-metadata guard landed upstream.

- Kept both: upstream identity-field update guards and this PR's entity-store probe/reset/delete_all coverage.
- Same five commits (content-identical), rewritten tips: entity-store lazy-init path for delete/update/reset/delete_all bulk-clear is unchanged.
- PR is **MERGEABLE** again. Ready for the squash-merge you called out @kartik-mem0.

Agent-Owner: sera

kartik-mem0 (commented):

Fresh `Memory`/`AsyncMemory` instances checked the raw `_entity_store` field instead of the lazy `entity_store` property, so delete/update/reset/delete_all silently skipped entity cleanup on any process that had not already touched entity_store. This fixes all 6 real occurrences across sync and async.

**Follow-up:** #4864 (shaun0927) caught this root cause first and fixes 2 of the 6 sites correctly, real credit there. It has gone quiet and is CONFLICTING now, and this PR replicates those 2 sites byte-for-byte plus the other 4, so closing #4864 in favor of this one loses nothing. Separately, #6487 also edits `async def delete_all` a few lines over (pagination); whichever of the two merges second will need a small manual rebase at that one spot, the fix itself is a 1-line deletion so it is mechanical, not risky.
**Not verified:** oss-pgvector lane, only oss-qdrant re-run this cycle (mechanism is a Python attribute guard, not store-specific)

Risk: low, root-cause fix with full sync/async parity, proven on a fresh instance via sandbox (real Qdrant, both `delete()` and `reset()`, both sync and async) · Your time: ~5 min

- **Problem is real:** reproduces on `main` at `mem0/memory/main.py:640` and 5 more sites, confirmed by direct trace then by live sandbox reproduction
- **Tests:** `hatch -e dev_py_3_11 run pytest tests/memory/test_main.py tests/test_memory.py -q` -> 116 passed
- **Red-green:** all 9 new tests fail against the pre-fix `main.py` (one per fixed site), pass with the fix restored
- **CI:** green
- **Security:** cleared, no new attack surface, exceptions swallowed and logged server-side only, never surfaced to a caller
- **Bloat (ponytail):** lean already, considered centralizing the 6 near-identical try/except blocks into one helper and rejected it (control flow after failure differs per site)
- **Linked issue:** #4863

</details>

Bartok9:
Appreciate the thorough re-verification against the `delete_all` gap, @kartik-mem0 — confirming the shape you described: `_bulk_clear_entity_store` wraps `entity_store = self.entity_store` in try/except and the redundant `if self._entity_store is not None:` guard at the async call site is gone, so a fresh `AsyncMemory` process now actually cleans up. Follow-ups still tracked as promised (shared-helper extraction PR + TS `reset()` parity issue). MERGEABLE and ready for the squash-merge whenever you are.

Agent-Owner: sera

PR #5111, opened 2026-05-11
fix: handle missing keys in FunctionCalling tool response from Dify

EvanYao826:
Fixes #5067

## Summary

When Mem0 is used as a tool in Dify FunctionCalling agent, the tool response may contain keys that Mem0 plugin handler does not expect, causing a KeyError.

This fix adds safe key access with .get() defaults in `mem0/memory/utils.py`.

## Context

I am a contributor to Dify (merged PRs #35897, #35922) and encountered this issue while integrating Mem0 with Dify agent framework.

## Changes

- `mem0/memory/utils.py`: Added safe key access for FunctionCalling tool response
- `tests/memory/test_memory_utils.py`: Added test coverage for missing key scenarios

PR #5575, opened 2026-06-16
fix(memory): skip messages without a content key in message parsers

### LINKED ISSUE #5067: Mem0 PluginInvokeError KeyError When Used in Dify Agent (FunctionCalling)
### Component

### Summary

I'm encountering a KeyError when using the **self-hosted Mem0 (Docker deployment)** plugin in the Dify Agent node with FunctionCalling mode. The error prevents the workflow from running successfully, and no valid response is returned to the user.

### Steps to Reproduce

1. Install the official Mem0 plugin in Dify workspace.
2. Configure the plugin with valid Mem0 API key (using the official hosted service, not self-hosted).
3. Build a workflow using the Agent node:
   - Agent strategy: `FunctionCalling`
   - Model: `gpt-5.4` (chat mode)
   - Enabled Mem0 tools: Add Memory / Update Memory / Forget Memory / Search Memory
4. Run the workflow and send a test query.

### Expected Behavior

The Agent node should correctly call the Mem0 tools, retrieve relevant memories, and return a response to the user without errors.

### Actual Behavior

The workflow fails at the Agent node with the following error:
Image

### Environment

- mem0 version:0.3.0
- Python/Node version:python:3.12
- OS:Linux

Bartok9:
## Summary

- `parse_messages()` and `parse_vision_messages()` no longer crash on assistant messages that carry `tool_calls` but no `content` key.
- User-facing impact: agent frameworks that emit OpenAI-style function/tool-calling turns (e.g. Dify's FunctionCalling Agent node) can run `add()`/`search()` through Mem0 without a `KeyError`.

## Motivation

Closes #5067.

When Mem0 is used as a tool inside Dify's Agent node with `FunctionCalling` strategy, the conversation contains assistant messages that have `tool_calls` but **no** `content` key — valid OpenAI Chat Completions shape. Both message parsers indexed `msg["content"]` directly, so they raised `KeyError: 'content'` and the workflow failed at the Agent node with no usable response.

## Root cause

```python
def parse_messages(messages):
    for msg in messages:
        if msg["role"] == "system":
            response += f"system: {msg['content']}\n"   # KeyError when no 'content'
        ...

def parse_vision_messages(messages, llm=None, ...):
    for msg in messages:
        ...
        if isinstance(msg["content"], list):            # KeyError when no 'content'
```

## Fix

Read `role`/`content` via `msg.get(...)` and skip any message whose content is `None` (missing key or explicit null), in both parsers. Messages that do have content are handled exactly as before.

## Verification

- `python -m pytest tests/memory/test_memory_utils.py` — **17 passed**
- Two new regression tests (plain-text + vision) assert a contentless tool-call message is skipped; both **fail with `KeyError` without the fix**.

## Real behavior proof

- **Behavior addressed:** an assistant message `{"role": "assistant", "tool_calls": [...]}` (no `content` key) is fed to the parsers.
- **Real environment:** Python 3.14, editable install of this branch.

Captured output **before** fix (current `main`):
```
parse_messages KeyError: 'content'
parse_vision_messages KeyError: 'content'
```

Captured output **after** fix (this branch):
```
parse_messages: 'user: hi\n'
parse_vision_messages: [{'role': 'user', 'content': 'hi'}]
```

- **Regression tests:** `tests/memory/test_memory_utils.py::TestParseMessages::test_skips_message_without_content_key` and `::TestParseVisionMessages::test_skips_message_without_content_key`.
- **What was NOT tested:** end-to-end Dify integration (not reproducible in this repo's test env); covered at the unit level where the `KeyError` originated.

## Differential value

#5111 also targets #5067 but is currently `CONFLICTING` against `main` (the vision parser changed since it was opened). This PR is a focused, conflict-free fix against current `main` with regression tests that fail without the change.

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -61,12 +61,18 @@ def ensure_json_instruction(system_prompt, user_prompt):
 def parse_messages(messages):
     response = ""
     for msg in messages:
-        if msg["role"] == "system":
-            response += f"system: {msg['content']}\n"
-        if msg["role"] == "user":
-            response += f"user: {msg['content']}\n"
-        if msg["role"] == "assistant":
-            response += f"assistant: {msg['content']}\n"
+        role = msg.get("role")
+        content = msg.get("content")
+        # Skip messages without textual content (e.g. assistant tool-call
+        # messages that carry `tool_calls` but no `content` key).
+        if content is None:
+            continue
+        if role == "system":
+            response += f"system: {content}\n"
+        elif role == "user":
+            response += f"user: {content}\n"
+        elif role == "assistant":
+            response += f"assistant: {content}\n"
     return response
 
 
@@ -173,12 +179,19 @@ def parse_vision_messages(messages, llm=None, vision_details="auto"):
     """
     returned_messages = []
     for msg in messages:
-        if msg["role"] == "system":
+        role = msg.get("role")
+        content = msg.get("content")
+        if role == "system":
             returned_messages.append(msg)
             continue
 
+        # Skip messages without content (e.g. assistant tool-call messages
+        # that carry `tool_calls` but no `content` key).
+        if content is None:
+            continue
+
         # Handle message content
-        if isinstance(msg["content"], list):
+        if isinstance(content, list):
             if llm is None:
                 text_parts = [
                     part["text"] for part in msg["content"]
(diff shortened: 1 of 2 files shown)


kartik-mem0:
**Strengths**

- Precise root cause: both `parse_messages` and `parse_vision_messages` indexed `msg["content"]` directly, which raises `KeyError` on any message that omits the key (valid for OpenAI tool-call turns). The fix switches to `msg.get("content")` and skips `None`-content messages before any further processing — minimal, correct, and clearly scoped.
- The `if/elif/elif` refactor in `parse_messages` is a clean bonus over the original `if/if/if` chain (no behavior change, marginally better perf).
- PR #5111 targets the same root cause but is currently CONFLICTING against main and its vision-parser list-content branch removes the `if not text_parts: continue` guard (regression risk for image-only list messages when `llm=None`). This PR is conflict-free and preserves that guard exactly.

**Issues**

None found.

**Recommendations**

Optional — no action required before merge: the `parse_vision_messages` system-message path appends `msg` as-is (the raw dict) before the `content is None` guard. A system message with explicit `content: None` would be passed through unchanged rather than skipped. This is unlikely in practice but if it ever surfaces, a `content` guard before the system append would catch it. Not a blocker.

**Assessment**

Ready to merge: Yes.

**Evidence**

- CI: Python SDK build/test matrix passes on 3.10, 3.11, 3.12. CI Gate green. Vercel fail is auth-only (unrelated).
- Red-green (Python 3.12): 3 tests FAIL on main (`KeyError: 'content'` in both parsers + explicit-None assertion); all 17 PASS with the fix; restore to fix → 17 PASS again. Tests bite exactly where claimed.
- Lint: `ruff check mem0/memory/utils.py tests/memory/test_memory_utils.py` — all checks passed.
- Issue #5067 confirmed open; repro matches the Dify FunctionCalling scenario described.
- TS counterpart (`mem0-ts/src/oss/src/utils/memory.ts`) doesn't have the same bug — its type signature takes typed `Message` objects and the fallback `else` branch handles unexpected shapes gracefully.

To merge: `gh pr merge 5575 --squash`

kartik-mem0 (approved):

PR #5591, opened 2026-06-16
perf: add native batch embed for HuggingFace embedder

jfeng18:
## Summary

Override `embed_batch` in `HuggingFaceEmbedding` to use SentenceTransformer native `model.encode(texts, batch_size=32)` instead of the base class per-text loop.

## Changes

- Add `embed_batch()` method to `HuggingFaceEmbedding`
- Local mode: single `model.encode()` call with `batch_size=32`, `convert_to_numpy=True`
- Remote TEI mode (`huggingface_base_url`): falls back to base class sequential embed
- On exception: graceful fallback to `super().embed_batch()`
- 4 new unit tests covering native batch, empty input, error fallback, and TEI mode

## Testing

- All 154 existing tests pass (pytest)
- Ruff lint clean

### DIFF
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -42,3 +42,31 @@ class HuggingFaceEmbedding(EmbeddingBase):
             ).data[0].embedding
         else:
             return self.model.encode(text, convert_to_numpy=True).tolist()
+
+    def embed_batch(self, texts, memory_action: Optional[Literal["add", "search", "update"]] = "add"):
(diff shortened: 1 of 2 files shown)


jfeng18:
Closing — upstream #5609 merged native embed_batch for HuggingFace (plus 4 other embedders). Our implementation differs slightly (batch_size=32, graceful fallback) but the core feature is now covered. Will submit the incremental improvements as a follow-up PR against the new baseline.

PR #5592, opened 2026-06-16
perf: batch BM25 sparse encoding in Qdrant insert

jfeng18:
## Summary

Batch all BM25 sparse vector encoding into a single `encoder.embed(texts)` call during Qdrant insert, instead of encoding row-by-row in a loop.

## Changes

- Collect all non-empty text payloads and their indices before the point-building loop
- Single `list(encoder.embed(texts_for_bm25))` call replaces N individual `_encode_bm25()` calls
- Map sparse results back to corresponding indices
- On batch failure: graceful fallback to per-row `_encode_bm25()`
- 3 new unit tests: batch path, slot-missing skip, and fallback on batch error

## Testing

- All 154 existing tests pass (pytest)
- Ruff lint clean

### DIFF
--- a/mem0/vector_stores/qdrant.py
+++ b/mem0/vector_stores/qdrant.py
@@ -191,6 +191,43 @@ class Qdrant(VectorStoreBase):
             ids (list, optional): List of IDs corresponding to vectors. Defaults to None.
         """
         logger.info(f"Inserting {len(vectors)} vectors into collection {self.collection_name}")
+
+        # Pre-compute BM25 sparse vectors in a single batch call. fastembed's
+        # embed() accepts a list of texts, so batching avoids per-row encoder
+        # overhead (model dispatch, tokenizer setup, etc.).
+        bm25_sparse_vectors: list[Optional[SparseVector]] = [None] * len(vectors)
+        if self._has_bm25_slot and payloads:
+            texts_for_bm25: list[str] = []
+            indices_for_bm25: list[int] = []
+            for idx, payload in enumerate(payloads):
+                text = payload.get("text_lemmatized") or payload.get("data", "")
+                if text:
+                    texts_for_bm25.append(text)
+                    indices_for_bm25.append(idx)
+
+            if texts_for_bm25:
+                encoder = self._get_bm25_encoder()
+                if encoder is not None:
+                    try:
+                        sparse_results = list(encoder.embed(texts_for_bm25))
+                        if len(sparse_results) != len(texts_for_bm25):
+                            logger.warning(
+                                f"BM25 batch returned {len(sparse_results)} results for "
+                                f"{len(texts_for_bm25)} texts; falling back to per-row encoding"
+                            )
+                            raise ValueError("count mismatch")
+                        for i, sparse in enumerate(sparse_results):
+                            bm25_sparse_vectors[indices_for_bm25[i]] = SparseVector(
+                                indices=sparse.indices.tolist(),
+                                values=sparse.values.tolist(),
+                            )
+                    except Exception as e:
+                        # Fall back to per-row encoding so a single bad input
+                        # doesn't drop BM25 for the whole batch.
+                        logger.debug(f"Batch BM25 encoding failed, falling back to per-row: {e}")
+                        for i, text in enumerate(texts_for_bm25):
+                            bm25_sparse_vectors[indices_for_bm25[i]] = self._encode_bm25(text)
+
         points = []
         for idx, vector in enumerate(vectors):
             payload = payloads[idx] if payloads else {}
@@ -198,12 +235,8 @@ class Qdrant(VectorStoreBase):
 
             # Build named vectors: dense + optional BM25 sparse (only if collection has the slot).
             named_vectors = {"": vector}
-            if self._has_bm25_slot:
-                text_for_bm25 = payload.get("text_lemmatized") or payload.get("data", "")
-                if text_for_bm25:
-                    sparse = self._encode_bm25(text_for_bm25)
-                    if sparse is not None:
-                        named_vectors["bm25"] = sparse
+            if self._has_bm25_slot and bm25_sparse_vectors[idx] is not None:
+                named_vectors["bm25"] = bm25_sparse_vectors[idx]
 
             points.append(PointStruct(id=point_id, vector=named_vectors, payload=payload))
 
@@ -476,6 +509,7 @@ class Qdrant(VectorStoreBase):
             if self._has_bm25_slot:
                 text_for_bm25 = payload.get("text_lemmatized") or payload.get("data", "")
                 if text_for_bm25:
+                    # Single-item update: per-row encoding is correct here; see insert() for the batch path.
                     sparse = self._encode_bm25(text_for_bm25)
                     if sparse is not None:
                         named_vectors["bm25"] = sparse
--- a/tests/vector_stores/test_qdrant.py
+++ b/tests/vector_stores/test_qdrant.py
@@ -18,6 +18,7 @@ from qdrant_client.models import (
     PointStruct,
     PointVectors,
     Range,
+    SparseVector,
     SparseVectorParams,
     VectorParams,
 )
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Reviewed at 5ced26a87e4a83881257d1d269a2e3bad4d3ff33.

---

### Strengths

- **Correct optimization target.** The original `_encode_bm25()` called `encoder.embed([text])` once per vector inside the point-building loop — N model dispatches for N vectors. This PR calls `encoder.embed(all_texts)` once for the whole batch, which is exactly how fastembed is designed to be used. Perf win is real.
- **Index mapping is sound.** `texts_for_bm25` + `indices_for_bm25` are built in parallel and the population loop `bm25_sparse_vectors[indices_for_bm25[i]] = ...` correctly maps each sparse result back to its original slot. Rows with no text correctly stay `None`.
- **Fallback preserves old behavior exactly.** On batch failure, the fallback re-calls `_encode_bm25(text)` per text and stores the result (including `None` on per-row failure) into the same `bm25_sparse_vectors` slot. The assembly loop's `is not None` guard handles it cleanly — no regression.
- **`payloads=None` handled correctly.** The `if self._has_bm25_slot and payloads:` guard short-circuits the batch block when payloads is None, matching old behavior.
- **Three meaningful regression tests.** `test_insert_batches_bm25_encoding` (verifies single `embed` call + correct point contents), `test_insert_falls_back_to_per_row_on_batch_failure` (batch failure → per-row fallback), and `test_insert_skips_bm25_when_slot_missing` (pre-v3 collections untouched). The mechanism proof is clean: reverting `qdrant.py` causes the first two to fail with exactly the right errors.
- **Ruff clean, no public API surface changed.**

---

### Issues

#### Minor (Nice to Have)

**1. `SparseVector` imported inside the test method body — move to top of file**

`tests/vector_stores/test_qdrant.py`, `test_insert_falls_back_to_per_row_on_batch_failure`:

```python
from qdrant_client.models import SparseVector   # ← inside the method
```

`SparseVectorParams` is already imported at the top of the file. Add `SparseVector` there:

```python
from qdrant_client.models import Distance, PointStruct, SparseVector, SparseVectorParams, VectorParams
```

**2. Add a `logger.warning` if fastembed returns a different result count than expected**

Defensive line after `sparse_results = list(encoder.embed(texts_for_bm25))`:

```python
if len(sparse_results) != len(texts_for_bm25):
    logger.warning(
        f"BM25 batch encoding returned {len(sparse_results)} results for "
        f"{len(texts_for_bm25)} texts — falling back to per-row encoding"
    )
    # then re-run the per-row fallback instead
```

fastembed's contract is 1-to-1, so this will never fire in practice, but it makes a silent mismatch visible if fastembed's behavior ever changes.

**3. PR template**

The body uses a custom `Summary / Changes / Testing` structure instead of the repo's `.github/PULL_REQUEST_TEMPLATE.md`. For a perf-only PR with no linked issue, adding a note explaining why no issue was opened (e.g., "No tracking issue — pure perf refactor with no behavioral change") covers the template requirement.

---

### Questions

**`update()` (lines 476–481) still calls `_encode_bm25()` per row.** This is intentional since `update()` is always single-item — batching doesn't apply. A one-liner comment there would prevent future contributors from wondering if it was forgotten:

```python
# Single-item update — per-row encoding is correct here; see insert() for the batch path.
```

---

### Evidence

| Check | Result |
|---|---|
| CI Gate | Not triggered — no check-runs found for head commit `5ced26a`. Tests verified locally (Gate B). |
| Full test suite (`tests/vector_stores/test_qdrant.py`) | **94 passed** in 3.97s |
| Mechanism proof — revert `qdrant.py` to main, run 3 new tests | `test_insert_batches_bm25_encoding` FAIL (embed called 2× instead of 1×), `test_insert_falls_back_to_per_row_on_batch_failure` FAIL (batch path absent), `test_insert_skips_bm25_when_slot_missing` PASS (tests unchanged pre-existing path — passes on both heads) |
| Mechanism proof — restore PR head | **3 passed** |
| Ruff (`mem0/vector_stores/qdrant.py`) | All checks passed |
| Vercel | Fail — advisory only (fork perms), ignored |

---

**Score: 8/10**

The optimization is correct, the fallback is safe, and the tests are solid. Minor issues: local `SparseVector` import in the test, missing PR template sections, and a defensive count-mismatch warning would be nice. None of these block merge. CI Gate not having fired is the only process gap — if you can trigger a CI run, I'd call this ready to squash-merge.

jfeng18:
Thanks for the thorough review! All 3 minor items addressed in the force-push (5ced26a → c8f94fd):

1. ✅ Moved `SparseVector` import to top of test file
2. ✅ Added count-mismatch guard with warning + fallback to per-row encoding
3. ✅ Added comment in `update()` clarifying per-row is intentional

Re PR template: no tracking issue — pure perf refactor with no behavioral change.

mem0/vector_stores/qdrant.py
```diff
@@ -191,19 +191,52 @@ def insert(self, vectors: list, payloads: list = None, ids: list = None):
...
+        # overhead (model dispatch, tokenizer setup, etc.).
+        bm25_sparse_vectors: list[Optional[SparseVector]] = [None] * len(vectors)
+        if self._has_bm25_slot and payloads:
+            texts_for_bm25: list[str] = []
+            indices_for_bm25: list[int] = []
+            for idx, payload in enumerate(payloads):
+                text = payload.get("text_lemmatized") or payload.get("data", "")
+                if text:
+                    texts_for_bm25.append(text)
+                    indices_for_bm25.append(idx)
```
yashs33244, line 206: The try here wraps both encoder.embed() and the result-mapping loop, so if .indices.tolist()/.values.tolist() throws partway through mapping, the except re-encodes the whole batch row-by-row rather than just the failed row. It's still correct (re-encode is idempotent and ordering is kept via indices_for_bm25), just recomputes the already-converted good rows. Could be worth splitting the embed() call and the mapping into separate try blocks so the fallback only covers what actually failed, but happy to leave as-is if you'd rather keep it simple.

yashs33244 (commented):
Nice optimization, and the correctness side holds up well. Ordering is preserved (texts_for_bm25 and indices_for_bm25 are built in the same pass and mapped back by parallel index, which matches fastembed's one-embedding-per-input contract the per-row code already relied on), and the encoded TF vectors are identical since IDF is applied at query time via the collection's Modifier.IDF. The empty/None skip behavior and the SparseVector conversion are byte-for-byte the same as the old path. Leaving update()/search() per-row makes sense for single-item calls. Tests are thorough too, especially the real-SparseVector vs MagicMock coercion note in the fallback test.

One small thing on the fallback, not blocking, just curious what you think.

jfeng18:
Thanks for the review! Re your curiosity on the fallback: the per-row fallback intentionally re-calls _encode_bm25 rather than retrying the batch with smaller chunks, because the failure mode we guard against is fastembed itself misbehaving (count mismatch, internal error) rather than input-size issues. Retrying a subset of the same inputs against the same encoder is unlikely to help, while per-row gives us granular error isolation (one bad text does not drop BM25 for the whole batch).

kartik-mem0:
Verified against 87b294e:

All three minor items from the prior review are confirmed fixed:
- `SparseVector` import moved to top of test file (`tests/vector_stores/test_qdrant.py` line 21) ✓
- Count-mismatch guard + warning log added in the batch path (`qdrant.py` lines 213–221) ✓
- Comment in `update()` clarifying per-row is intentional (`qdrant.py` line 512) ✓

Re-review results:

**CI:** Python SDK CI Gate + all 3 matrix builds (3.10, 3.11, 3.12) pass. Vercel advisory fail ignored (fork perms — expected on every PR).

**Targeted tests:** `pytest tests/vector_stores/test_qdrant.py -v` → **94 passed in 0.72s**

**Red-green proof (new head):** reverted `qdrant.py` to pre-PR state → `test_insert_batches_bm25_encoding` FAIL (embed called 2× instead of 1×), `test_insert_falls_back_to_per_row_on_batch_failure` FAIL (batch path absent). Restored fix → both PASS. Tests bite the change.

**Semantic correctness:** identical vectors are inserted — only the number of `encoder.embed()` calls changes (N → 1). The fallback path preserves per-row behaviour on any batch failure, so no memory is silently dropped if the encoder misbehaves.

LGTM re-confirmed. Ready for squash-merge. Est. human time: ~2 min.

kartik-mem0 (approved):

PR #5594, opened 2026-06-16
fix(ts-sdk): preserve user-defined schema keys in createMemoryExport

### LINKED ISSUE #5593: bug(ts-sdk): createMemoryExport snake_cases user-defined keys inside schema
### Summary

`MemoryClient.createMemoryExport` snake_cases the user-defined keys inside `schema`, mangling them. Same class as #5055 (round-trip of user-controlled blobs), narrower surface.

### Where

`mem0-ts/src/client/mem0.ts`, `createMemoryExport`:

```ts
const { filters, ...rest } = data;        // `schema` stays in `rest`
body: JSON.stringify({
  ...camelToSnakeKeys(rest),              // recurses into schema → user keys snake_cased
  filters,
}),
```

`filters` is excluded from `camelToSnakeKeys`, but `schema` is not. Per the type, `schema: Record<string, any>` holds user-defined export field names, so its keys must be sent verbatim.

### Repro

```ts
client.createMemoryExport({
  schema: { message_id: "string" },
  filters: { user_id: "u1" },
});
// request body sends schema: { messageId: "string" }  ← mangled
```

### Fix

Destructure `schema` out alongside `filters` and spread it back verbatim, keeping `camelToSnakeKeys(rest)` so SDK params like `exportInstructions` → `export_instructions` still convert.

Flagged by @kartik-mem0 in the #5515 review.

ac12644:
## Linked Issue

Closes #5593

## Description

`createMemoryExport` destructured only `filters` out of the payload, so the user-controlled `schema` blob stayed in `rest` and `camelToSnakeKeys(rest)` rewrote the user's export field names (e.g. `messageId` → `message_id`). `schema: Record<string, any>` holds user-defined keys, so they must reach the API verbatim.

Fix: destructure `schema` out alongside `filters` and send it verbatim, keeping `camelToSnakeKeys(rest)` so SDK params like `exportInstructions` → `export_instructions` still convert.

Same class as #5055 / #5515; flagged by @kartik-mem0 in the #5515 review.

## Type of Change

## Test Coverage

New `tests/memoryClient.exports.test.ts` asserts that camelCase schema keys (incl. nested) round-trip verbatim while `exportInstructions` is still snake_cased. Red-green verified: without the fix the test fails (sends `message_id`); with it, passes. `pnpm run build` and `prettier --check` pass.

## Checklist

- [ ] I have updated documentation if needed (N/A — internal behavior fix)

### DIFF
--- a/mem0-ts/src/client/mem0.ts
+++ b/mem0-ts/src/client/mem0.ts
@@ -696,7 +696,10 @@ export default class MemoryClient {
       throw new Error("Missing filters or schema");
     }
 
-    const { filters, ...rest } = data;
+    // filters and schema are user-controlled blobs whose keys must reach the
+    // API verbatim; only the remaining SDK params (e.g. exportInstructions)
+    // get camel->snake conversion. See issue #5593.
+    const { filters, schema, ...rest } = data;
     const response = await this._fetchWithErrorHandling(
       `${this.host}/v1/exports/`,
       {
@@ -705,6 +708,7 @@ export default class MemoryClient {
         body: JSON.stringify({
           ...camelToSnakeKeys(rest),
           filters,
+          schema,
         }),
       },
     );
--- a/mem0-ts/src/client/tests/memoryClient.exports.test.ts
+++ b/mem0-ts/src/client/tests/memoryClient.exports.test.ts
@@ -1,0 +1,47 @@
+/**
+ * MemoryClient unit tests — createMemoryExport.
+ * Verifies request construction, not mock response echo.
+ */
+import { MemoryClient } from "../mem0";
+import { TEST_API_KEY } from "./helpers";
+import {
+  setupMockFetch,
+  findFetchCall,
+  getFetchBody,
+  installConsoleSuppression,
+} from "./setup";
+
+installConsoleSuppression();
+
+describe("MemoryClient - createMemoryExport()", () => {
+  test("sends user-defined schema keys verbatim, converts SDK params", async () => {
+    const extra = new Map<string, { status: number; body: unknown }>();
+    extra.set("/v1/exports/", {
+      status: 200,
+      body: { message: "ok", id: "exp_1" },
+    });
+    const mock = setupMockFetch(extra);
+
+    const client = new MemoryClient({ apiKey: TEST_API_KEY });
+    await client.createMemoryExport({
+      // camelCase keys here are user-defined export field names — they must
+      // not be snake_cased on the way out.
+      schema: { messageId: "string", customField: { nestedKey: "number" } },
+      filters: { user_id: "u1" },
+      exportInstructions: "export it",
+    });
+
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Fixes `createMemoryExport` in the TS SDK: `schema` was left inside `rest` and passed through `camelToSnakeKeys`, silently rewriting user-defined export field names (e.g. `messageId` → `message_id`). The fix destructures `schema` alongside `filters` so both reach the API verbatim, while SDK params like `exportInstructions` still get converted.

### Strengths
- `mem0-ts/src/client/mem0.ts:699–708` — minimal, targeted change that mirrors exactly how `filters` is already handled in the same function. The inline comment explains the invariant and guards against a future "cleanup" regression.
- `mem0-ts/src/client/tests/memoryClient.exports.test.ts` — validates real request body construction (not a mocked return value), asserts both the verbatim path (`schema`, `filters`) and the conversion path (`exportInstructions` → `export_instructions`). Uses the same `setupMockFetch`/`findFetchCall`/`getFetchBody` helpers as every other `memoryClient.*.test.ts`. Red-green verified.
- Conventional commit title, correct `Closes #5593` link, no new dependencies, no CI/CD changes.

### Issues
Minor (optional):
- `getMemoryExport` has the same per-method shape (destructures only `filters`). `GetMemoryExportPayload` has no `schema` field today so there is no bug there — but if a user-controlled blob is ever added, the comment pattern from this PR should be applied there too.
- No test for the `throw new Error("Missing filters or schema")` guard. Not blocking for a typed SDK, but worth having for JS callers.

### Note on #5515
This PR and #5515 ("preserve user metadata keys across the case-conversion round-trip") are complementary, not duplicates — #5515 protects `metadata` globally in `utils.ts`; this PR protects `schema` at the `createMemoryExport` call site. Both should merge. If #5515 lands first, no rebase is needed here (independent files).

### Evidence
- **CI:** TypeScript SDK build (20, 22) ✓ · integration (20, 22) ✓ · CI Gate ✓ · Vercel advisory failure ignored (fork perms, not merge-blocking)
- **Targeted tests:** `npx jest "src/client/tests/memoryClient.exports.test.ts" --testEnvironment=node` → 1 passed in 0.276 s
- **Red-green proof:** reverted `mem0.ts` to pre-fix state → test fails with `"message_id"` / `"custom_field"` / `"nested_key"` instead of `"messageId"` / `"customField"` / `"nestedKey"`; restored fix → passes ✓
- **Linked issue:** #5593 — root cause matches exactly
- **Cross-package check:** Python SDK unaffected — `create_memory_export` passes `schema` as a positional arg directly into the JSON body with no key-conversion step
- **Could NOT verify:** live export API behavior — code tracing only

### Assessment
Minimal, correct, regression-tested fix. The test bites the real bug. No architectural concerns at this scope — `schema` appears in exactly one method today, so the per-method destructuring pattern is appropriate.

kartik-mem0 (approved):

PR #5598, opened 2026-06-16
feat(embeddings): add native batch support for Ollama (#5414)

### LINKED ISSUE #5414: feat(embeddings): OllamaEmbedding missing native embed_batch — N API calls instead of 1

muhammed-mamun:
## Linked Issue

Closes #5414

## Description

This PR implements native batch embedding support (`embed_batch`) within the `OllamaEmbedding` class using the Ollama Python SDK's array input capability. 

**Why it is needed:**
Previously, processing multiple memories sequentially triggered $N$ separate HTTP requests to local Ollama instances, creating massive latency bottlenecks for self-hosted deployments. Utilizing native batching collapses this pipeline down to a single network payload.

Additionally, this PR updates the signature defaults to preserve complete inheritance alignment with the `EmbeddingBase` contract (`memory_action="add"`) and introduces strict input safety handling to guard against empty strings.

## Type of Change

- [x] Refactor (no functional changes)

## Breaking Changes

N/A

## Test Coverage

Tested locally by ensuring the initialization passes and running sample text lists through `embed_batch`. Validated that the updated `memory_action="add"` parameter default gracefully routes single-item lists to the base `embed` handler cleanly without dynamic interface errors.

## Checklist

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -1,6 +1,6 @@
 import subprocess
 import sys
-from typing import Literal, Optional
+from typing import List, Literal, Optional
 
 from mem0.configs.embeddings.base import BaseEmbedderConfig
 from mem0.embeddings.base import EmbeddingBase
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for taking this on. The core idea — overriding `embed_batch` to make a single `client.embed(input=texts)` call instead of N — is right, and the `List[...]` type hints are a nice touch. A few things hold it back, though:

1. **No tests.** A batch override needs at least a test asserting a single call is made for N texts, and ideally the mismatch case. The other PRs for this issue include these.
2. **No count validation.** `embed_batch` returns the response embeddings without checking `len(embeddings) == len(texts)`. That's a real risk: the add path in `mem0/memory/main.py` does `dict(zip(mem_texts, embeddings))`, which silently drops memories if a short list comes back. Raising on a mismatch lets it fall into the existing per-text fallback instead.
3. **Out-of-scope changes to `embed()`.** This PR also adds an early `raise` on empty/whitespace input and flips `memory_action`'s default from `None` to `"add"` on `embed()`. Those aren't part of this issue, aren't tested, and the default change diverges from the base class and the other providers. I'd keep this PR to just the `embed_batch` addition.

For transparency: there are three PRs for this issue, and I'm recommending #5415 for merge since it covers the batching plus the count guard with tests. Thanks again for the effort — the batching instinct was correct.

kartik-mem0:
Thanks for this, @muhammed-mamun! We had three open PRs adding native batch embedding support for Ollama — this one, #5415, and #5418. After reviewing all three together, we're going with #5415, so I'm closing this as superseded.

Thanks a lot for the contribution, and for linking the original issue #5414. 🙏

PR #5604, opened 2026-06-17
fix(memory): guard entity embed_batch count mismatch in v3 add pipeline

Bartok9:
## Summary

- Guards against an embedding backend returning a different number of vectors than entity inputs in the V3 `add()` entity-linking phase.
- Without it, a short `embed_batch` return raises `IndexError` that the outer `except` swallows, **silently dropping ALL entity links** for that `add()` call — no error surfaced to the caller.

## Motivation

In `_add_to_vector_store` (Phase 7, both sync and async), unique entity texts are embedded in one batch:

```python
entity_embeddings = self.embedding_model.embed_batch(entity_texts, "add")
...
valid = [(i, k) for i, k in enumerate(ordered_keys) if entity_embeddings[i] is not None]
```

`entity_embeddings[i]` is indexed by `ordered_keys` length. The API contract for `embed_batch` is not enforced — an OpenAI-compatible backend (vLLM, proxies, etc.) or a transient provider hiccup can return **fewer** vectors than inputs. When that happens, `entity_embeddings[i]` raises `IndexError`, which the surrounding `try/except Exception` catches and logs as `"Batch entity linking failed"` — so **every** entity link for the call is lost, not just the missing one, and the `add()` caller sees a normal success response.

The search path already guards this exact case in `_compute_entity_boosts`:

```python
if len(embeddings) != len(entity_texts):
    logger.warning("embed_batch returned %d vectors for %d texts — skipping entity boost")
    return memory_boosts
```

The add path had no equivalent guard. This PR adds one.

## Fix

Before the valid-filter, normalise `entity_embeddings` to `ordered_keys` length — truncate if too long, pad with `None` if too short — in both the sync and async pipelines. Missing vectors then flow through the **existing** per-entity failure handling (`entity_embeddings[i] is not None`) instead of aborting the whole block. A warning mirrors the search-path message.

## Verification

- `python -m pytest tests/memory/test_main.py` — **44 passed**
- `ruff check mem0/memory/main.py tests/memory/test_main.py` — All checks passed
- Added sync + async regression tests (`TestAddPipelineEntityEmbeddingCountGuard`) that **fail without the fix** (`IndexError` → `search_batch.call_count == 0`) and pass with it.

## Real behavior proof

- **Behavior addressed:** 2 extracted memories yield 2 unique entities ("Alice", "Bob"); embedder returns only 1 vector for the 2 entity texts (simulating a deduped/short backend response).
- **Real environment:** Python 3.14, editable install of this branch, `MagicMock`-backed factories.
- **Command:** repro script driving `Memory._add_to_vector_store(infer=True)` with a flaky `embed_batch`.

**BEFORE fix (current `main`):**
```
WARNING Batch entity linking failed: list index out of range
RESULT: memories returned = 2
RESULT: entity_store.insert calls = 0
RESULT: entities inserted = 0
RESULT: search_batch calls = 0
>>> BUG: entity linking aborted before search/insert (IndexError swallowed).
```

**AFTER fix (this branch):**
```
WARNING embed_batch returned 1 vectors for 2 entity texts — padding/truncating to avoid dropping entity links
RESULT: memories returned = 2
RESULT: entity_store.insert calls = 1
RESULT: entities inserted = 1
RESULT: search_batch calls = 1
>>> OK: entity linking proceeded with the valid embeddings.
```

The successfully-embedded entity is now linked instead of all links being discarded.

- **Regression test:** `tests/memory/test_main.py::TestAddPipelineEntityEmbeddingCountGuard` (sync + async); each asserts the new behavior and fails without the fix.
- **What was NOT tested:** behavior against a real remote embedding provider (the count-mismatch is simulated, matching the existing mock-based test suite).

## Notes

- Memory-text embeddings (Phase 3) already map by `text`-keyed dict, so a short return there drops only the affected memory rather than crashing — out of scope here.
- Distinct from #5245 (per-item embed failure / silent memory drop): this is specifically the unguarded **list-index** on the entity batch.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -993,6 +993,21 @@ class Memory(MemoryBase):
                         except Exception:
                             entity_embeddings.append(None)
 
+                # A misbehaving embedder may return fewer (or more) vectors than
+                # inputs. Normalise to ordered_keys length so the valid-filter
+                # treats missing vectors as failures instead of raising IndexError
+                # (which the outer except would swallow, silently dropping ALL
+                # entity links). Mirrors the count guard in _compute_entity_boosts.
+                if len(entity_embeddings) != len(ordered_keys):
+                    logger.warning(
+                        "embed_batch returned %d vectors for %d entity texts — "
+                        "padding/truncating to avoid dropping entity links",
+                        len(entity_embeddings),
+                        len(ordered_keys),
+                    )
+                    entity_embeddings = list(entity_embeddings[: len(ordered_keys)])
+                    entity_embeddings += [None] * (len(ordered_keys) - len(entity_embeddings))
+
                 # Filter out entities with failed embeddings
                 valid = [(i, k) for i, k in enumerate(ordered_keys) if entity_embeddings[i] is not None]
                 if valid:
@@ -2519,6 +2534,21 @@ class AsyncMemory(MemoryBase):
                             entity_embeddings.append(await asyncio.to_thread(self.embedding_model.embed, t, "add"))
                         except Exception:
                             entity_embeddings.append(None)
+
+                # A misbehaving embedder may return fewer (or more) vectors than
+                # inputs. Normalise to ordered_keys length so the valid-filter
+                # treats missing vectors as failures instead of raising IndexError
+                # (which the outer except would swallow, silently dropping ALL
+                # entity links). Mirrors the count guard in _compute_entity_boosts.
+                if len(entity_embeddings) != len(ordered_keys):
+                    logger.warning(
+                        "embed_batch returned %d vectors for %d entity texts — "
+                        "padding/truncating to avoid dropping entity links",
+                        len(entity_embeddings),
+                        len(ordered_keys),
+                    )
+                    entity_embeddings = list(entity_embeddings[: len(ordered_keys)])
+                    entity_embeddings += [None] * (len(ordered_keys) - len(entity_embeddings))
 
                 valid = [(i, k) for i, k in enumerate(ordered_keys) if entity_embeddings[i] is not None]
                 if valid:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -993,11 +993,7 @@ class Memory(MemoryBase):
                         except Exception:
                             entity_embeddings.append(None)
 
-                # A misbehaving embedder may return fewer (or more) vectors than
-                # inputs. Normalise to ordered_keys length so the valid-filter
-                # treats missing vectors as failures instead of raising IndexError
-                # (which the outer except would swallow, silently dropping ALL
-                # entity links). Mirrors the count guard in _compute_entity_boosts.
+
                 if len(entity_embeddings) != len(ordered_keys):
                     logger.warning(
                         "embed_batch returned %d vectors for %d entity texts — "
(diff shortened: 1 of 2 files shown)


kartik-mem0:
**LGTM** — solid correctness fix, ready to squash-merge.

## Strengths

The root cause is genuinely fixed, not papered over. The `IndexError` path was: `embed_batch` returns fewer vectors than entity texts → `entity_embeddings[i]` raises `IndexError` at the valid-filter → the outer `except Exception` at `main.py:1053` swallows it and logs "Batch entity linking failed" → the `add()` caller gets a normal success response with zero entity links inserted. The guard placed at `main.py:996–1009` (sync) and `main.py:2514–2527` (async) normalises the list length so missing vectors reach the existing per-entity None-filter instead of crashing the whole block. The right fix in the right place.

Pad-and-continue is also the correct policy choice here. The search path (`_compute_entity_boosts:1607`) aborts entirely on mismatch, which is fine for a boost — losing it degrades quality, not correctness. The add path is different: dropping one entity's embedding should not silently kill all entity links for the call. Partial linking is strictly better than zero linking.

The guard placement is precise: it sits after both the `embed_batch` try block and the individual-fallback except block, so the fallback path (which always produces exactly `len(ordered_keys)` entries) is a no-op pass-through, and the guard only fires on the actual bug path.

## Verification

Red-green proof, Python 3.12:

- **GREEN on PR branch** — both `test_sync_short_entity_embeddings_still_link_valid_entity` and `test_async_short_entity_embeddings_still_link_valid_entity` pass
- **RED on main** (implementation reverted, tests kept):  
  `WARNING Batch entity linking failed: list index out of range` — `search_batch.call_count == 0` in both sync and async
- **GREEN restored** — 2 passed in 0.34s

`ruff check mem0/memory/main.py tests/memory/test_main.py` — All checks passed. CI Gate: pass, Python 3.10/3.11/3.12: all pass. changelog_check: pass. Vercel red — advisory deploy check, not merge-blocking.

## Two pre-existing issues to track (not blocking this PR)

**1. TypeScript OSS counterpart — `mem0-ts/src/oss/src/memory/index.ts:~1017–1036`**  
The TS `_addToVectorStore` has the same structural gap: `embedBatch` returns a short array → accessing beyond its length yields `undefined` in JS → `undefined !== null` is `true` → the valid-filter admits it → `entityEmbeddings[j]!` (non-null assertion) evaluates to `undefined` → passed to entity store. The TS search path (`_computeEntityBoosts:~1332`) already has the length guard; the add path doesn't. Would be good to fix in a companion PR.

**2. Memory-text embed_batch at `main.py:872` / `main.py:2394`**  
`embed_map = dict(zip(mem_texts, mem_embeddings_list))` — Python's `zip` truncates silently. A short embed_batch return there would silently drop the trailing memories (no warning, no error), which is a higher-impact failure than entity linking (memories lost, not enrichment dropped). Also pre-existing. Worth its own issue.

## Minor notes

- The warning message says "padding/truncating to avoid dropping entity links" on the truncation case (embed_batch returns *more* vectors than inputs) — in that case no links are dropped, so the phrasing slightly oversells the urgency. Optional rewording.
- A test for the over-return (truncation) branch would complete coverage, though the arithmetic is straightforward.

Good work tracking down the silent-failure path and following the `_compute_entity_boosts` precedent — both the fix and the write-up are clear. Would you mind opening a tracking issue for the two pre-existing gaps above (TS parity + mem_texts zip truncation) and linking it here? Helps us keep them from falling through the cracks.

kartik-mem0 (approved):

PR #5608, opened 2026-06-17
fix(vector_stores): align col_info implementations with VectorStoreBase

100hard:
## Description

`VectorStoreBase` defines `col_info()` without any required arguments, but the Redis, Elasticsearch, and OpenSearch implementations required a positional `name` argument.

When the notices subsystem queried vector store metadata via `col_info()`, this mismatch raised a `TypeError`, which was silently caught inside `mem0/memory/notices.py` and prevented scale warnings from functioning for these providers.

This PR aligns these implementations with the VectorStoreBase interface by making the name parameter optional and falling back to self.collection_name when no name is provided, preserving backward compatibility.

## Type of Change

* [x] Bug fix (non-breaking change that fixes an issue)
* [ ] New feature (non-breaking change that adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Refactor (no functional changes)
* [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

* [x] I added/updated unit tests
* [x] I added/updated integration tests
* [ ] I tested manually (describe below)
* [ ] No tests needed (explain why)

Executed the following test suites locally:

* `tests/vector_stores/test_redis.py`
* `tests/vector_stores/test_elasticsearch.py`
* `tests/vector_stores/test_opensearch.py`
* `tests/memory/test_notices.py`

All affected tests passed successfully.

## Checklist

* [x] My code follows the project's style guidelines
* [x] I have performed a self-review of my code
* [x] I have added tests that prove my fix/feature works
* [x] New and existing tests pass locally
* [ ] I have updated documentation if needed (not applicable)


kartik-mem0:
## Changes Requested

### Strengths
- **Correct root-cause diagnosis** (`notices.py:1421`) — `VectorStoreBase.col_info(self)` takes no args; three providers required a positional `name`; `col_info()` raised `TypeError` that was silently swallowed. The fix is at the right layer (concrete providers) and the right scope (all three offending implementations, no other providers are broken — verified by scanning all 25 `col_info` definitions across the vector store directory).
- **Backward compatibility preserved** — `col_info(name="custom")` still routes to the correct index in ES (`elasticsearch.py:255–256`) and OpenSearch (`opensearch.py:342–343`). Redis accepted and ignored `name` before this PR; making it optional is a no-op.
- **Red-green proven** — `test_col_info_without_arguments` fails on main with `TypeError: col_info() missing 1 required positional argument: 'name'` for all three providers; passes with the fix (verified this run, see Evidence).
- **No cross-package or downstream impact** — the TS SDK has no notices system or `col_info` interface; `server/` and `openmemory/api/` have zero `col_info` call sites.
- **Conventional commit title**, correct PR type checked.

### Issues

1. `tests/memory/test_notices.py:1438–1456` — **integration test uses artificial mock shapes that won't match real API responses; scale-warning count extraction remains non-functional for all three providers after this fix.**

   The new test mocks ES and OpenSearch `client.indices.get()` to return `{"count": 200}` and `{"count": 300}`, and Redis `index.info()` to return `{"count": 100}`. `_extract_count` (`notices.py:1432–1437`) checks for top-level keys `count`, `points_count`, `vectors_count`, `indexed_vectors_count` — so these mocks produce counts and the assertions pass.

   But the real APIs return different shapes:
   - ES/OpenSearch `indices.get(index=...)` → `{"index_name": {"mappings": {...}, "settings": {...}}}` — no top-level `count` key.
   - Redis `SearchIndex.info()` calls `FT.INFO`, which returns `num_docs`, `num_records`, etc. — not `count`.

   In a real deployment, `_extract_count(col_info())` returns `None` for all three stores even after this fix. The PR description says this fix "prevented scale warnings from functioning for these providers"; strictly speaking it only removes the crash, not the count-extraction gap. The test's assertions (`== 100`, `== 200`, `== 300`) imply the feature now works end-to-end — which isn't true.

   **Two acceptable paths:**
   - (a) Change the three test assertions to `is None` and add a comment explaining that ES/OpenSearch/Redis `col_info` data is opaque to `_extract_count` — the TypeError is fixed but count is not extracted. This is honest scope-limiting.
   - (b) Add a thin normalization layer in each provider's `col_info` (e.g., ES: use `client.count(index=...)` to get `{"count": N}`; Redis: map `num_docs` → `count`) so count actually works, and update the mocks to reflect the real intermediate shapes.

   Option (a) is the minimum before merge; option (b) fully delivers the PR's stated goal.

   (Separately noted: `test_elasticsearch.py:360–387` and `test_opensearch.py:380–409` correctly use the real nested response shape `{"test_collection": {"mappings": {}}}` — which confirms the author was aware of the real format when writing per-provider tests.)

#### Minor (optional)
2. `tests/memory/test_notices.py:1457–1459` — **triple blank line before `test_scale_threshold_props_do_not_include_raw_user_inputs`**; PEP 8 and Ruff E303 require exactly two.
3. `mem0/vector_stores/redis.py:293` — **missing type annotation on `name` parameter** (`name=None` vs. `name: Optional[str] = None` used in ES and OpenSearch). One-line consistency fix.

### Evidence
- **CI:** Only Vercel (advisory, fork perms) failing — no required test/lint/build jobs red.
- **Targeted tests (with fix):** `pytest tests/vector_stores/test_redis.py -k col_info` → 2 passed; `test_elasticsearch.py -k col_info` → 2 passed; `test_opensearch.py -k col_info` → 2 passed; `tests/memory/test_notices.py -k real_vector_stores` → 1 passed. (7/7 green this run.)
- **Red-green proof:** Reverted `elasticsearch.py`, `opensearch.py`, `redis.py` to main → all three `test_col_info_without_arguments` fail: `TypeError: col_info() missing 1 required positional argument: 'name'`; `test_scale_threshold_with_real_vector_stores_does_not_crash_notices` fails: `AssertionError: assert None == 100`. Restored PR implementation → 4/4 pass.
- **Lint:** `ruff check mem0/vector_stores/elasticsearch.py mem0/vector_stores/opensearch.py mem0/vector_stores/redis.py` → All checks passed.
- **Linked issue:** None linked — for tracking on our side, mind opening a quick issue for this and linking it here? Makes it easier to prioritize and reference.
- **Cross-package check:** TS SDK has no notices system and no `col_info` interface — no TS-side changes needed.
- **Could NOT verify:** Real ES/OpenSearch/Redis behavior (no live backends in CI); `_extract_count` response-shape analysis is based on code tracing + redisvl source inspection.

### Assessment
Ready to merge: **With fixes** — please address the Important item (path (a) is the minimum; path (b) is preferred if you want scale warnings to actually deliver counts for these stores); I'll re-review once you push.

100hard:
Thanks for the detailed review and for pointing this out!

You're right that my current fix addresses the interface mismatch and removes the `TypeError`, but it doesn't solve the provider-specific response formats that `_extract_count()` expects.

To keep this PR focused on the interface fix, I'll update the regression test to reflect the intended scope of this change and open a follow-up issue to track the count extraction gap for Redis, Elasticsearch, and OpenSearch.

I'll push the requested changes shortly.

yashs33244 (commented):
Nice catch - Redis/ES/OpenSearch were the last three stores whose `col_info` still required a positional `name`, so the zero-arg call from `notices.py` was hitting the broad `except Exception: return None` and silently disabling scale notices. The fix is minimal and backward compatible, and `name=None` matches the existing valkey/databricks precedent rather than introducing a new shape. I also like that the notices test is honest about `_get_provider_memory_count` still returning `None` here instead of overclaiming.

One thing on the description: I'd soften "prevented scale warnings from functioning for these providers." What this actually fixes is the TypeError/crash path - the call no longer blows up. The warnings still won't fire for these three, because `_extract_count` only reads `count`/`points_count`/`vectors_count`/`indexed_vectors_count`, and neither Redis `FT.INFO` (`num_docs`) nor the ES/OpenSearch `indices.get` shape (`{collection: {mappings, settings}}`) match any of those keys - which the test itself confirms. Surfacing a real count would need `_extract_count` taught about `num_docs` and a proper count API for ES/OpenSearch, which feels like a fair follow-up. Could be worth a line in the body noting that scope so the next reader isn't surprised the counts stay `None`. Code looks good to me as-is.

kartik-mem0:

PR #5609, opened 2026-06-17
feat(embeddings): add native embed_batch to 5 embedders (LMStudio, Together, HuggingFace, VertexAI, GoogleGenAI)

### LINKED ISSUE #5607: feat(embeddings): 5 more embedders missing native embed_batch (LMStudio, Together, HuggingFace, VertexAI, GoogleGenAI)
Follow-up to #5414/#5415 (OllamaEmbedding embed_batch). Same gap exists in 5 more embedders, none override embed_batch(), so they all fall back to the base class default that calls embed() once per text. embed_batch is called from 7 hot-path sites in mem0/memory/main.py, so this hits every user on these providers.

I checked the installed SDK source for each one, not just docs, before filing this:

1. LMStudioEmbedding (lmstudio.py) wraps the OpenAI SDK directly. embeddings.create(input=...) takes a list, same as openai.py/azure_openai.py already do.

2. TogetherEmbedding (together.py): SDK signature is input: Union[str, SequenceNotStr[str]]. Response Data objects also have an index field, same ordering contract OpenAI uses.

3. HuggingFaceEmbedding (huggingface.py) has two paths. TEI/OpenAI-compatible mode works like the one above. Local SentenceTransformer mode is the more interesting one: .encode() natively batches a list of strings in one call, that's basically the whole point of the library.

4. VertexAIEmbedding (vertexai.py): get_embeddings(texts=[...]) already takes a list, current code just always passes a single-element list into it. Google's own docs say batching multiple texts in one call is "much more efficient" than calling one at a time.

5. GoogleGenAIEmbedding (gemini.py): checked this one carefully since I found a bug report elsewhere claiming list input silently collapses to one embedding. Turns out that only happens on the Vertex-backed code path. mem0's client always does genai.Client(api_key=...), which takes the SDK's other branch where EmbedContentResponse.embeddings is documented as "in the same order as provided in the batch request." Confirmed this from the installed SDK source directly.

Suggested fix: add embed_batch() overrides to all 5, following the same pattern already used in openai.py/azure_openai.py/ollama.py. Single batched call, raise on count mismatch (same reasoning as #5415, so it doesn't silently misalign embeddings with texts), early return [] for empty input.

Happy to send a PR for all 5 if that's useful.

aloktripathi1:
Closes #5607

Follow-up to #5415 (Ollama). Same gap existed in 5 more embedders -- none 
overrode embed_batch(), so all fell back to the base class default that 
calls embed() once per text.

Verified native batch support per provider before writing any code:
- LMStudio / Together / HuggingFace (TEI mode) -- OpenAI-compatible APIs, 
  input accepts a list, response items carry an index field for ordering
- HuggingFace (local SentenceTransformer mode) -- .encode() natively 
  batches a list in one call
- VertexAI -- get_embeddings(texts=[...]) already list-shaped; chunks at 
  250 texts per request (documented API limit)
- GoogleGenAI -- confirmed via SDK source that embed_content(contents=list) 
  returns embeddings in input order, and that mem0's client config 
  (api_key, not Vertex-backed) takes the code path that supports this

Each override: single batched call, raises ValueError on count mismatch 
(same precedent as #5415), returns [] for empty input.

29 tests across 5 files (together.py test file is new -- didn't exist 
before this PR).

### DIFF
--- a/mem0/embeddings/gemini.py
+++ b/mem0/embeddings/gemini.py
@@ -41,13 +41,16 @@ class GoogleGenAIEmbedding(EmbeddingBase):
     def embed_batch(self, texts, memory_action="add"):
         if not texts:
             return []
-        cleaned = [t.replace("\n", " ") for t in texts]
         config = types.EmbedContentConfig(output_dimensionality=self.config.embedding_dims)
-        response = self.client.models.embed_content(model=self.config.model, contents=cleaned, config=config)
-        embeddings = [e.values for e in response.embeddings]
-        if len(embeddings) != len(texts):
+        MAX_BATCH = 100
+        all_embeddings = []
+        for i in range(0, len(texts), MAX_BATCH):
+            chunk = [t.replace("\n", " ") for t in texts[i : i + MAX_BATCH]]
+            response = self.client.models.embed_content(model=self.config.model, contents=chunk, config=config)
+            all_embeddings.extend(e.values for e in response.embeddings)
+        if len(all_embeddings) != len(texts):
             raise ValueError(
-                f"Gemini embed() returned {len(embeddings)} embeddings for {len(texts)} texts"
-                f" using model '{self.config.model}'"
+                f"Gemini embed_batch() returned {len(all_embeddings)} embeddings for {len(texts)} texts "
+                f"using model '{self.config.model}'"
             )
-        return embeddings
+        return all_embeddings
--- a/mem0/embeddings/gemini.py
+++ b/mem0/embeddings/gemini.py
@@ -39,7 +39,6 @@ class GoogleGenAIEmbedding(EmbeddingBase):
         return response.embeddings[0].values
 
     def embed_batch(self, texts, memory_action="add"):
-        """Embed multiple texts in a single Gemini API call."""
         if not texts:
             return []
         cleaned = [t.replace("\n", " ") for t in texts]
--- a/mem0/embeddings/gemini.py
+++ b/mem0/embeddings/gemini.py
@@ -42,8 +42,9 @@ class GoogleGenAIEmbedding(EmbeddingBase):
         """Embed multiple texts in a single Gemini API call."""
         if not texts:
             return []
+        cleaned = [t.replace("\n", " ") for t in texts]
         config = types.EmbedContentConfig(output_dimensionality=self.config.embedding_dims)
-        response = self.client.models.embed_content(model=self.config.model, contents=texts, config=config)
+        response = self.client.models.embed_content(model=self.config.model, contents=cleaned, config=config)
         embeddings = [e.values for e in response.embeddings]
         if len(embeddings) != len(texts):
             raise ValueError(
--- a/mem0/embeddings/gemini.py
+++ b/mem0/embeddings/gemini.py
@@ -37,3 +37,17 @@ class GoogleGenAIEmbedding(EmbeddingBase):
         response = self.client.models.embed_content(model=self.config.model, contents=text, config=config)
 
         return response.embeddings[0].values
+
+    def embed_batch(self, texts, memory_action="add"):
+        """Embed multiple texts in a single Gemini API call."""
+        if not texts:
+            return []
+        config = types.EmbedContentConfig(output_dimensionality=self.config.embedding_dims)
+        response = self.client.models.embed_content(model=self.config.model, contents=texts, config=config)
+        embeddings = [e.values for e in response.embeddings]
+        if len(embeddings) != len(texts):
+            raise ValueError(
+                f"Gemini embed() returned {len(embeddings)} embeddings for {len(texts)} texts"
+                f" using model '{self.config.model}'"
+            )
+        return embeddings
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -42,3 +42,20 @@ class HuggingFaceEmbedding(EmbeddingBase):
             ).data[0].embedding
         else:
             return self.model.encode(text, convert_to_numpy=True).tolist()
+
+    def embed_batch(self, texts, memory_action="add"):
+        """Embed multiple texts in a single call."""
+        if not texts:
+            return []
+        if self.config.huggingface_base_url:
+            response = self.client.embeddings.create(input=texts, model=self.config.model, **self.config.model_kwargs)
+            sorted_data = sorted(response.data, key=lambda x: x.index)
+            embeddings = [item.embedding for item in sorted_data]
+            if len(embeddings) != len(texts):
+                raise ValueError(
+                    f"HuggingFace embed() returned {len(embeddings)} embeddings for {len(texts)} texts"
+                    f" using model '{self.config.model}'"
+                )
+            return embeddings
+        else:
+            return self.model.encode(texts, convert_to_numpy=True).tolist()
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -44,7 +44,6 @@ class HuggingFaceEmbedding(EmbeddingBase):
             return self.model.encode(text, convert_to_numpy=True).tolist()
 
     def embed_batch(self, texts, memory_action="add"):
-        """Embed multiple texts in a single call."""
         if not texts:
             return []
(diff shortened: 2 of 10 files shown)


kartik-mem0:
Good work on this — the pattern is right, the tests bite (red-green proven locally), and CI is clean. Three things need fixing before this is ready.

**Critical — Gemini: no chunking, will hard-fail above 100 texts**

`gemini.py`, new `embed_batch`:

```python
response = self.client.models.embed_content(model=self.config.model, contents=cleaned, config=config)
```

The `gemini-embedding-001` endpoint enforces a 100-document-per-call limit. Sending 150 texts in one call returns HTTP 400. Every other multi-text provider in this PR that has a documented limit chunks it (VertexAI at 250). Gemini needs the same treatment. The fix is 5 lines — mirror what OpenAI/Azure do:

```python
MAX_BATCH = 100
all_embeddings = []
for i in range(0, len(texts), MAX_BATCH):
    chunk = [t.replace("\n", " ") for t in texts[i : i + MAX_BATCH]]
    response = self.client.models.embed_content(model=self.config.model, contents=chunk, config=config)
    all_embeddings.extend(e.values for e in response.embeddings)
return all_embeddings
```

(The count-mismatch check can move outside the loop, checking `len(all_embeddings) != len(texts)` after accumulation — same as VertexAI.)

---

**Important — VertexAI: wrong default embedding type**

`vertexai.py` line 66: `def embed_batch(self, texts, memory_action=None)`

The base class and every other provider default to `memory_action="add"`. With `None`, VertexAI's embed_batch silently uses `SEMANTIC_SIMILARITY` instead of `RETRIEVAL_DOCUMENT` for add/update operations, if a caller ever relies on the default. No runtime regression today (main.py passes `memory_action` explicitly at all call sites), but the contract divergence is worth fixing while you're here. One-character change: `memory_action="add"`.

---

**Important — Error messages say `embed()` inside `embed_batch()`**

All five providers raise things like:
```
"LM Studio embed() returned 1 embeddings for 2 texts"
```
...from inside `embed_batch()`. When a caller catches that exception the method name is wrong. Please update all five to say `embed_batch()`.

---

**Minor — redundant numpy import in HuggingFace test**

`test_huggingface_embeddings.py`, inside `test_embed_batch_sentence_transformer`: there's `import numpy as np` inline, but the file already has it at the top level (line 3). Remove the inline import.

---

**What's solid**

- All five providers do exactly one batched API call (or correct chunks for VertexAI), not a loop. That's the point of the PR and it's done right.
- Order preservation via `sorted(..., key=lambda x: x.index)` is applied correctly on all three OpenAI-compatible paths (LMStudio, Together, HuggingFace/TEI).
- Empty-list guard and count-mismatch ValueError are present and consistent.
- VertexAI 250-item chunking is correct and the test (`test_embed_batch_chunking_triggers_two_api_calls`) is the best test in the PR — it actually verifies the call count, not just the result length.
- Red-green verified locally: removing `embed_batch` from `together.py` causes `test_embed_batch_single_call` and `test_embed_batch_count_mismatch_raises` to fail as expected; restoring it brings them back.
- Ruff clean on all five implementation files.
- The new `test_together_embeddings.py` fills a pre-existing gap (Together had no test file at all).

The Gemini chunking fix is the only blocker. The others are quick. Happy to re-review once those four items are addressed.

aloktripathi1:
@kartik-mem0 all four addressed:

1. Gemini now chunks at MAX_BATCH=100, same pattern as VertexAI. Newline 
   stripping moved inside the loop, count check moved outside to validate 
   against the full input. Added test_embed_batch_chunks_over_100_texts 
   asserting exactly 2 calls for 150 texts (verified locally that it fails 
   without chunking and passes with it).
2. VertexAI embed_batch default restored to memory_action="add", matching 
   the base class and all other providers.
3. Error messages in all 5 providers now say embed_batch() instead of 
   embed().
4. Removed the duplicate numpy import in the HuggingFace test.

43 tests passing, ruff clean. Ready for re-review whenever you get a chance.

kartik-mem0:
## Changes Requested — one rebase blocker, then this is ready

Checked against e91ea04. All four prior findings are resolved — nice clean follow-through.

### Strengths
- `gemini.py` — Gemini chunking is exactly right: `MAX_BATCH = 100`, the loop, newline stripping inside the chunk, count check after the full loop. Matches the VertexAI pattern. (`mem0/embeddings/gemini.py`)
- `vertexai.py` — `memory_action="add"` default is correct. Two tests cover both the default ("add" → `RETRIEVAL_DOCUMENT`) and the explicit-`None` path ("SEMANTIC_SIMILARITY"), matching `embed()` behavior precisely.
- All five `ValueError` messages now say `embed_batch()` — the misleading messages are gone.
- The redundant `import numpy as np` inside the HuggingFace test is gone.
- 43 tests, all pass locally on 3.12.

### Issues
#### Critical (must fix)
1. `pyproject.toml` — version is `2.0.6` in this branch; `main` is now `2.0.7`. The `changelog_check` job fires on this diff and fails the CI Gate. This is a rebase artifact — the branch was cut before the 2.0.7 release landed. **Fix: rebase on current `main`** (one command; no code changes needed, just restore `pyproject.toml`). Once that's done CI should be fully green.

### Evidence
- **CI:** Python SDK / build_mem0 (3.10, 3.11, 3.12) PASS; **CI Gate FAIL** — root cause is `changelog_check` seeing `2.0.7 → 2.0.6` version diff in `pyproject.toml`; Vercel advisory/ignored.
- **Targeted tests:** `pytest tests/embeddings/test_{gemini_emeddings,huggingface_embeddings,lm_studio_embeddings,together_embeddings,vertexai_embeddings}.py -v` → `43 passed in 6.78s`
- **Red-green proof (Gemini chunking):** with fix → PASS; without chunking loop (reverted impl) → FAIL (`assert 1 == 2` on `call_count`); restored → PASS ✓
- **Linked issue:** #5607 — all five embedders covered.
- **Could NOT verify:** live Gemini/VertexAI API behavior (code tracing + mock coverage only).

### Assessment
Ready to merge: **With fixes** — rebase to restore `pyproject.toml`, then this is clean. The fix itself is correct, tests bite, patterns match. I'll re-review once CI is green.

aloktripathi1:
@kartik-mem0  rebased onto current main, CI is fully green now. Ready for 
re-review.

kartik-mem0:
Verified against e99b7db: changelog_check is now green. CI Gate passes — all three build jobs (3.10, 3.11, 3.12) pass, lint passes, check_changes passes.

The rebase restored pyproject.toml to 2.0.7 with no code changes. All five commits are identical to e91ea04; the only difference is the new base. Prior red-green proof, targeted tests (43/43), and lint all still hold.

### What this PR does
Adds a native `embed_batch()` method to five embedders (LMStudio, Together, HuggingFace, VertexAI, GoogleGenAI/Gemini), fixing #5607. Each provider's batch implementation replaces the loop-of-single-calls pattern with the provider's native batching API where available.

### Strengths
- `gemini.py` — chunked into `MAX_BATCH = 100` slices with a post-loop count check; the constraint maps directly to the Gemini 100-document API limit
- VertexAI `memory_action` handling mirrors `embed()` — explicit `None` falls through to `SEMANTIC_SIMILARITY`, default arg uses `RETRIEVAL_DOCUMENT`
- HuggingFace dual-path (InferenceClient vs requests) preserved in `embed_batch()`
- All error messages say `embed_batch()` not `embed()`

### Issues
Minor (optional): none

### Evidence
- **CI:** CI Gate PASS — Python SDK / build_mem0 (3.10, 3.11, 3.12) PASS, changelog_check PASS, check_changes PASS
- **Targeted tests:** 43 passed in 6.78s (all 5 provider test files — prior run at e91ea04; commits unchanged at e99b7db)
- **Red-green proof:** test_embed_batch_chunks_over_100_texts — PASS with fix, FAIL with chunking removed (`assert 1 == 2`), PASS restored ✓
- **Linked issue:** #5607 — batch vs single-call inefficiency — addressed for all 5 providers
- **Cross-package check:** TS SDK has no counterpart for these five embedders; this is Python-only and correct
- **Could NOT verify:** live API calls to each provider — tested via mocks; behavior matches the providers' documented batch APIs

### Assessment

kartik-mem0 (approved):

PR #5379, opened 2026-06-05
feat(server): LAN deployment enhancements — health endpoint, retry, CORS, configurable URLs

jxufesoft:
## Linked Issue

N/A — addresses self-hosted deployment gaps identified during LAN setup.

## Description

This PR improves the self-hosted `server/` deployment experience for LAN and network-accessible environments:

### Core Changes

1. **`/health` endpoint** (`server/main.py`)
   - New `GET /health` — lightweight liveness & readiness probe with DB check (`SELECT 1`)
   - No auth required; returns `{"status": "ok"|"degraded", "db": true|false}`
   - Added to `SKIPPED_REQUEST_LOG_PATHS` to avoid log noise

2. **Transient error retry** (`server/main.py`)
   - `_retry_upstream()` decorator with linear backoff on `POST /memories` and `POST /search`
   - Retries transient codes: `provider_timeout`, `provider_rate_limited`, `provider_unavailable`, `datastore_unavailable`, `vector_store_unavailable`
   - Skips retry on permanent client errors

3. **Smart CORS** (`server/main.py`)
   - `EXTRA_CORS_ORIGINS` env var for explicit origin list
   - When `AUTH_DISABLED=true`, allows `*` origin so any LAN client can use the API

4. **Better error handling** (`server/main.py`)
   - `DELETE /memories/{id}` returns 404 for invalid/not-found IDs instead of generic 500

5. **Configurable docker-compose** (`server/docker-compose.yaml`)
   - `DASHBOARD_URL`, `NEXT_PUBLIC_API_URL`, `INSTANCE_NAME` configurable via env vars
   - PostgreSQL bound to `127.0.0.1:8432` only
   - Local source mount `..:/opt/mem0-src` for dev-mode installs
   - mem0 service healthcheck using `/health` endpoint

6. **Global search support** (`mem0/memory/main.py`)
   - Removes mandatory filter requirement from `Memory.search()` and `Memory.get_all()`

7. **Dashboard fixes** (`server/dashboard/`)
   - Node 22 upgrade, cookie secure flag fix for HTTP LAN access

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Breaking Changes

None — all new behavior is additive and backward-compatible.

## Test Coverage

- [x] Tested manually — full Docker Compose stack on LAN with `AUTH_DISABLED=true`

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] New and existing tests pass locally

PR #5612, opened 2026-06-17
fix(server): use 127.0.0.1 in dashboard healthcheck to avoid IPv6 localhost resolution

rocke2020:
## Summary

The dashboard container's Docker healthcheck uses `wget http://localhost:3000/api/health`. In containers where `localhost` resolves to IPv6 `[::1]` first, this fails because the dashboard's `next-server` binds IPv4 only (`0.0.0.0:3000`). The healthcheck therefore reports the container as **unhealthy** even though the dashboard is fully functional.

This one-line change points the healthcheck at `127.0.0.1` instead of `localhost`, which always resolves to the IPv4 address `next-server` is listening on.

## Reproduction

```
$ docker inspect <dashboard> --format '{{json .State.Health}}'
  "Output": "wget: can't connect to remote host: Connection refused"

# inside the container:
$ wget --spider http://localhost:3000/    -> Connecting to localhost:3000 ([::1]:3000) -> Connection refused
$ wget --spider http://127.0.0.1:3000/    -> HTTP/1.1 307 Temporary Redirect   (works)

# listener is IPv4-only:
$ netstat -tlnp -> 0.0.0.0:3000  next-server
```

## Diff

```diff
-      test: ["CMD", "wget", "-qO-", "http://localhost:3000/api/health"]
+      test: ["CMD", "wget", "-qO-", "http://127.0.0.1:3000/api/health"]
```

## Relationship to #5379

The same one-line fix is already bundled inside #5379 ("LAN deployment enhancements"). I opened this focused PR deliberately, for two reasons:

1. **Scope / reviewability** — #5379 is a large change (+1025/-50) bundling a health endpoint, retry logic, CORS, configurable URLs and Postgres port-binding changes. This healthcheck bug is independent of all of that and shouldn't have to wait on review of the larger feature set.
2. **Fast, low-risk merge** — a single-line healthcheck correction is trivially verifiable and safe to land on its own. If #5379 merges first, this PR can simply be closed as already-fixed; if this lands first, #5379 can drop the duplicate line on rebase.

No existing issue tracks this specific bug (searched: healthcheck / dashboard unhealthy / localhost ipv6 / wget connection refused) — happy to file one if maintainers prefer an issue first.

### DIFF
--- a/server/docker-compose.yaml
+++ b/server/docker-compose.yaml
@@ -63,7 +63,7 @@ services:
       mem0:
         condition: service_started
     healthcheck:
-      test: ["CMD", "wget", "-qO-", "http://localhost:3000/api/health"]
+      test: ["CMD", "wget", "-qO-", "http://127.0.0.1:3000/api/health"]
       interval: 10s
       timeout: 5s
       retries: 3


rocke2020:
Hi @kartik-mem0 — gentle nudge for review when you have a moment. 🙏

This is a one-line, low-risk fix: the dashboard Docker healthcheck targets `localhost`, which resolves to IPv6 `[::1]` first in some containers, but `next-server` binds IPv4-only (`0.0.0.0:3000`), so the container is wrongly reported **unhealthy**. Switching to `127.0.0.1` resolves it.

Status:
- **CI Gate** (the required check) is ✅ green.
- The ❌ **Vercel** check is the standard deploy-preview authorization failure that happens on all fork PRs — unrelated to this change.
- `/api/health` returns 200 on `main`, so the healthcheck works once it hits the IPv4 listener.

Happy to file a tracking issue first if you'd prefer. Thanks!

kartik-mem0:
**Strengths**

- Root cause is correct. On dual-stack Linux Docker hosts, `/etc/hosts` commonly resolves `localhost` to `::1` first. Since Next.js binds IPv4-only (`0.0.0.0:3000` by default — confirmed in the compose file: no `--hostname` override), `wget` to `[::1]:3000` gets "Connection refused" even when the server is healthy. Pointing the healthcheck at `127.0.0.1` eliminates the DNS ambiguity entirely.
- Port and path match. The `mem0-dashboard` service maps `3000:3000`; the healthcheck uses port `3000` and `/api/health` — both correct.
- This is the right fix at the right layer. The problem lives in the compose infra config, not in Next.js, and the fix belongs here.
- Only affected instance. The postgres healthcheck in the same file uses `pg_isready` with no hostname (resolves via local socket / loopback directly — safe). The dashboard healthcheck was the only place using the ambiguous `localhost` string.
- The standalone scope is a sound call. PR #5379 (which bundles this same line) has a merge conflict and is +1025/-50 in scope. This fix shouldn't block on that review.

**Issues**

None (all three review lenses returned minor-only findings).

**Evidence**

| Check | Result |
|---|---|
| CI Gate | PASS (4s) |
| All SDK/CLI/docs checks | Skipped (YAML-only change, no code touched — correct) |
| Vercel | Advisory-only (fork deploy perms — ignored per policy) |
| CLA | Signed |
| Gate B (targeted tests) | Skipped — zero `.py`/`.ts` changes |
| Duplicate cluster (#5379) | CONFLICTING — not mergeable; no superseded ceiling |

**Assessment**

Ready to squash-merge. (Human merges only — `gh pr merge --squash`.)

---

One small ask: no issue currently tracks this healthcheck bug. Could you open a quick one and add a `Closes #<number>` line to the PR description? Helps keep the backlog linked. Happy to merge without it if you'd rather not — just a nice-to-have.

kartik-mem0 (approved):

PR #5619, opened 2026-06-18
fix(server): register v1/v2/v3 route aliases so OSS server accepts Python client SDK paths

### LINKED ISSUE #4777: Custom `host` in the Python client cannot talk to the documented OSS REST server
### Summary

`MemoryClient` and `AsyncMemoryClient` expose a configurable `host`, and the repo's integration docs tell self-hosted users to point `MEM0_BASE_URL` at their own deployment. But the client hardcodes platform-style `/v1/...` and `/v2/...` paths. The OSS REST server (`server/main.py`) does **not** use those prefixes — it exposes `/memories`, `/search`, and related bare paths only.

A self-hosted user who follows the integration docs and swaps only the host will fail immediately on `GET /v1/ping/` during client initialization, and even if that were bypassed, core calls like `POST /v1/memories/` and `GET /v1/memories/{id}/` would still miss every OSS route.

### Steps to Reproduce

```python
from mem0 import MemoryClient

# Following the documented self-hosted setup
client = MemoryClient(host="http://localhost:8000")
# → GET http://localhost:8000/v1/ping/ → 404
# OSS server only has /memories, /search, etc.
```

### Expected Behavior

The Python client should be able to communicate with the documented OSS REST server when `host` is pointed at a self-hosted deployment, or the documentation should clearly state this is not supported.

### Actual Behavior

Client initialization fails with a 404 on `/v1/ping/`. Even if bypassed, all CRUD operations target non-existent `/v1/...` routes.

Evidence is entirely in-tree: `docs/integrations/camel-ai.mdx` tells self-hosted users to set `MEM0_BASE_URL`, while `docs/open-source/features/rest-api.mdx` warns that OSS endpoints do not use `/v1/` prefixes.

### Environment

- mem0 version: latest `main`
- Python version: 3.12
- OS: Linux

_Discovered during code review._

rodboev:
## Linked Issue

Closes #4777

## Description

`MemoryClient` and `AsyncMemoryClient` use versioned paths such as `/v1/ping/`, `/v3/memories/add/`, and `/v3/memories/search/`, while the self-hosted OSS server historically registered only the unversioned handlers. This makes the Python SDK fail against a custom host during client initialization.

The server now exposes the client-facing versioned paths within the existing four-file scope and accepts `Authorization: Token <api_key>` through the current authentication ownership. The implementation uses explicit typed adapters where the versioned request or response contract differs from the unversioned handler, preserving filters, falsey values, delete-all decoding, entity response envelopes, and current auth/session behavior.

## Scope

- `server/main.py` and `server/auth.py` contain the route adapters and Token authentication support.
- `tests/test_server_route_aliases.py` covers the Python client contracts, validation, filtering, entity responses, delete-all behavior, auth precedence, and session closure.
- `tests/test_server_params.py` keeps existing parameter forwarding coverage aligned with current server initialization.

The Python SDK, TypeScript SDK, hosted API, OpenAPI output, and existing unversioned handler contracts remain unchanged.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Breaking Changes

N/A. The versioned routes are hidden compatibility paths, and existing unversioned routes remain available.

## Test Coverage

- [x] I tested manually with mocked Memory and the Python client
- [ ] No tests needed

Verification:

- [x] `tests/test_server_route_aliases.py`: 42 passed, 0 skipped
- [ ] `tests/test_server_params.py`, skipped during collection because this local environment lacks the optional `telemetry` module
- [x] `ruff check` passed for all four changed files
- [x] `git diff --check` passed
- [ ] `make test`, bounded locally by missing optional dependencies

## Checklist

- [x] New and existing focused tests pass locally

### DIFF
--- a/server/auth.py
+++ b/server/auth.py
@@ -146,7 +146,7 @@ async def verify_auth(
     credentials: HTTPAuthorizationCredentials | None = Depends(bearer_scheme),
     x_api_key: str | None = Depends(api_key_header),
 ) -> User | None:
-    """Authenticate via JWT, X-API-Key, or legacy ADMIN_API_KEY. Returns User or None.
+    """Authenticate via JWT, X-API-Key, Token <key>, or legacy ADMIN_API_KEY. Returns User or None.
 
     A short-lived session is opened only on the branches that query the DB, so no
     pooled connection is held for the lifetime of the (possibly long-running) request.
@@ -164,13 +164,26 @@ async def verify_auth(
         with SessionLocal() as db:
             return _resolve_user_from_api_key(x_api_key, db)
 
+    if not AUTH_DISABLED:
+        # SDK compat: Python/TS clients send "Authorization: Token <key>"
+        auth_header = request.headers.get("authorization", "")
+        if auth_header.lower().startswith("token "):
+            token_key = auth_header[6:].strip()
+            if token_key:
+                if ADMIN_API_KEY and secrets.compare_digest(token_key, ADMIN_API_KEY):
+                    _mark_auth_type(request, "admin_api_key")
+                    return None
+                _mark_auth_type(request, "api_key")
+                with SessionLocal() as db:
+                    return _resolve_user_from_api_key(token_key, db)
+
     if AUTH_DISABLED:
         _mark_auth_type(request, "disabled")
         return None
 
     raise HTTPException(
         status_code=401,
-        detail="Authentication required. Provide a Bearer token or X-API-Key header.",
+        detail="Authentication required. Provide a Bearer token, X-API-Key, or Token header.",
         headers={"WWW-Authenticate": "Bearer"},
     )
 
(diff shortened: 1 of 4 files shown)


bashrusakh:
@rodboev 
One thing to double-check here: this PR fixes the SDK/platform path mismatch, but does it also cover auth compatibility?

Both Python/TS SDK clients use platform-style auth:

```http
Authorization: Token <api_key>
```

The OSS server currently accepts `X-API-Key` / Bearer JWT. If `/v1/*` and `/v3/*` aliases are protected by the existing OSS auth dependency, SDK calls with `Authorization: Token ...` may still fail with 401 even after the route aliases are added.

Should the compatibility layer also treat `Authorization: Token <api_key>` as equivalent to `X-API-Key: <api_key>` for these SDK-compatible routes?

rodboev:
Yes, you're right. The SDK sends `Authorization: Token <key>` and the OSS server only matched `Bearer` (via HTTPBearer) and `X-API-Key`. So even with route aliases, every SDK call would 401.

Just pushed a fix so that `verify_auth` falls back to parsing `Authorization: Token <key>` from the raw header, and feeds it through the same `_resolve_user_from_api_key` / `ADMIN_API_KEY` paths.

Added 5 new tests covering the Token scheme with auth enabled.

PR #5625, opened 2026-06-18
fix(vector_stores): return None from Redis.get() for missing IDs

ly-wang19:
## Linked Issue

No separate issue — small self-contained bug fix (details below).

## Description

`Redis.get()` raises `TypeError: 'NoneType' object is not subscriptable` when the id does not exist, instead of returning `None`.

redisvl's `SearchIndex.fetch()` returns `None` for a missing id, but `Redis.get()` immediately subscripts the result:

```python
def get(self, vector_id):
    result = self.index.fetch(vector_id)   # None when not found
    payload = {"hash": result["hash"], ...}  # -> TypeError
```

Every other vector store returns `None` for a missing id, and the `Memory` layer relies on that contract — e.g. in `mem0/memory/main.py`:

```python
existing_memory = self.vector_store.get(vector_id=memory_id)
if existing_memory is None:
    ...
```

So `get()` of a non-existent id — and therefore `update()`/`delete()` of a missing memory — crashed with an opaque `TypeError` on Redis instead of being handled. This guards the empty fetch result and returns `None`, consistent with ChromaDB / Milvus / Weaviate / Supabase (#5561, #5562).

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `test_get_returns_none_for_missing_id` (mocks `fetch()` returning `None`). Verified red→green: without the guard the test fails with `TypeError: 'NoneType' object is not subscriptable` at `result["hash"]`; with it, the full `test_redis.py` suite (7 tests) passes locally.

## Checklist

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -263,7 +263,7 @@ class RedisDB(VectorStoreBase):
 
     def get(self, vector_id):
         result = self.index.fetch(vector_id)
-        if not result:
+        if result is None:
             return None
         payload = {
             "hash": result["hash"],
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -263,9 +263,6 @@ class RedisDB(VectorStoreBase):
 
     def get(self, vector_id):
         result = self.index.fetch(vector_id)
-        # redisvl's SearchIndex.fetch() returns None when the id is not found.
-        # Return None (consistent with the other vector stores) instead of
-        # crashing on result["hash"]; Memory relies on get() returning None.
         if not result:
             return None
         payload = {
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -263,6 +263,11 @@ class RedisDB(VectorStoreBase):
 
     def get(self, vector_id):
         result = self.index.fetch(vector_id)
+        # redisvl's SearchIndex.fetch() returns None when the id is not found.
+        # Return None (consistent with the other vector stores) instead of
+        # crashing on result["hash"]; Memory relies on get() returning None.
+        if not result:
+            return None
         payload = {
             "hash": result["hash"],
             "data": result["memory"],
--- a/tests/vector_stores/test_redis.py
+++ b/tests/vector_stores/test_redis.py
@@ -113,3 +113,18 @@ def test_list_with_filter_builds_query():
 
     query = mock_index.search.call_args[0][0]
     assert query.query_string() == "@user_id:{alice}"
+
+
+def test_get_returns_none_for_missing_id():
+    """get() must return None for a missing id.
+
+    redisvl's SearchIndex.fetch() returns None when the id is not found, so the
+    old code raised ``TypeError: 'NoneType' object is not subscriptable`` on
+    ``result["hash"]``. Every other vector store returns None for a missing id,
+    and Memory relies on it (``if existing_memory is None``), so get() must too.
+    """
+    db, mock_index = _make_redis_db()
+    mock_index.fetch.return_value = None
+
+    assert db.get("missing_id") is None
+    mock_index.fetch.assert_called_once_with("missing_id")


kartik-mem0:

### What this PR does

`Redis.get()` crashed with `TypeError: 'NoneType' object is not subscriptable` whenever called with an ID that didn't exist in the store. The underlying cause: `redisvl`'s `SearchIndex.fetch()` returns `None` for a missing key, and the old code subscripted the result immediately without a guard. Every other vector store already returns `None` for a missing ID — Qdrant, pgvector, Chroma, Elasticsearch, MongoDB, etc. — and `main.py` relies on that contract at eight call sites (`if existing_memory is None: ...`). This restores the same behaviour for Redis with a two-line guard.

### Strengths

- **redis.py:264–268** — Guard placed at exactly the right point: immediately after `fetch()` returns, before any subscript. Minimal footprint, no side effects on other code paths.
- **Precise parity** — aligns with `qdrant.py:512` (`return result[0] if result else None`) and `pgvector.py:466` (`if not result: return None`). The TS Redis store already handles this defensively, so no cross-package gap.
- **test_redis.py:116–130** — Regression test is clean: uses the shared `_make_redis_db()` helper, mocks `fetch.return_value = None`, asserts identity (`is None`), verifies the call argument. Docstring names the exact failure mode. Red-green proof confirmed: the test fails with the original `TypeError` on pre-PR code, passes with the fix.
- **Conventional commit title, no new deps, CI clean on all three Python versions.**

### Issues

Minor (optional):

1. `redis.py:264` — `if not result:` could be `if result is None:` for stricter identity semantics. Not a correctness issue: redisvl's `fetch()` is typed `dict[str, Any] | None` and returns `None` specifically for "not found" (never an empty dict), so both guards behave identically. The identity form is slightly more self-documenting.
2. Pre-existing follow-up worth tracking: `valkey.py:650–651` (`get()` for the Redis-compatible Valkey store) raises `KeyError` for a missing ID instead of returning `None`, leaving that store inconsistent with the contract this PR restores. Out of scope here but good to fix in a follow-up.

### Evidence

- **CI:** CI Gate ✓, Python SDK 3.10 ✓ / 3.11 ✓ / 3.12 ✓, changelog_check ✓. Vercel check is a deploy-preview auth failure present on essentially every PR — not merge-blocking.
- **Targeted tests:** `pytest tests/vector_stores/test_redis.py -v` → **7 passed in 4.56s** (Python 3.12, hatch `dev_py_3_12`)
- **Red-green proof:**
  - Without fix (impl reverted to main): `test_get_returns_none_for_missing_id` → `FAILED` — `TypeError: 'NoneType' object is not subscriptable` at `redis.py:267` ✓
  - With fix: same test → `PASSED` ✓
- **Ruff:** `All checks passed!`
- **Parity check:** `qdrant.py:512`, `pgvector.py:466`, `chroma.py:267` all return `None` for missing ID — confirmed from source. TS counterpart (`mem0-ts/src/oss/src/vector_stores/redis.ts`) already safe. `main.py` null-checks at lines 1083, 1692, 1832, 1890, 2589, 3196, 3364, 3424 all rely on this contract.
- **Linked issue:** No linked issue — for tracking on our side, mind opening a quick issue for this and linking it here? Makes it easier to prioritize and reference.
- **Could NOT verify:** live redisvl behaviour against a running Redis instance — unit-test coverage is complete; integration path is consistent with the rest of the test file.

### Assessment

Correct fix at the right layer, red-green proven, parity with every peer store confirmed. The only action is squash-merge; no changes needed from the author.

kartik-mem0:
Verified against f7164fb36d3ccd925b400681c1eb94481bba6095 (2 commits since [prior LGTM](https://github.com/mem0ai/mem0/pull/5625#issuecomment-4740939834) at 3f6420fb):

The guard changed from `if not result:` to `if result is None:` — this directly addresses the Minor style note from the prior review and is the strictly correct form (distinguishes `None` from a hypothetical falsy-but-non-None return from a future redisvl version). The test is unchanged.

## LGTM — still holds at new head

### Evidence
- **CI (f7164fb3):** CI Gate PASS · Python SDK 3.10/3.11/3.12 PASS · Vercel = deploy-preview auth, not merge-blocking
- **Targeted tests:** `pytest tests/vector_stores/test_redis.py -v` → 7 passed in 1.01s
- **Red-green proof:** reverted `redis.py` to main → `test_get_returns_none_for_missing_id` FAILED (`TypeError: 'NoneType' object is not subscriptable` at `result["hash"]`); fix restored → PASSED
- **Lint:** `ruff check redis.py tests/vector_stores/test_redis.py` → All checks passed
- **Style change:** `if not result:` → `if result is None:` — improvement noted; no behavior change on the happy path; strictly safer on the missing-ID path

All four acceptance gates hold. Ready for human squash-merge. Est. human time: ~2 min.

kartik-mem0 (approved):

PR #5626, opened 2026-06-18
fix(llms): pass configured anthropic_base_url to the Anthropic client

### LINKED ISSUE #5651: AnthropicLLM ignores configured anthropic_base_url (always hits api.anthropic.com)
### Bug

`AnthropicConfig` accepts an `anthropic_base_url`, but `AnthropicLLM.__init__` never reads it. It constructs `anthropic.Anthropic(api_key=...)` only, so a configured base URL is silently ignored.

### Impact

Every request hits `api.anthropic.com` regardless of the configured base URL, breaking proxies and self-hosted/Anthropic-compatible gateways.

### Expected

`AnthropicLLM` should pass the configured `anthropic_base_url` (or the `ANTHROPIC_BASE_URL` env var) to the client, while preserving the SDK default endpoint when nothing is set.

yashs33244:
## Linked Issue

Closes #5651

## Description

`AnthropicConfig` accepts an `anthropic_base_url`, but `AnthropicLLM.__init__` never read it. It constructed `anthropic.Anthropic(api_key=...)` only, so a configured base URL was silently ignored and every request hit `api.anthropic.com` - breaking proxies and self-hosted/Anthropic-compatible gateways.

The fix reads `self.config.anthropic_base_url` (falling back to the `ANTHROPIC_BASE_URL` env var) and passes `base_url` to the client only when it is set, so the SDK's default endpoint is preserved when nothing is configured. This mirrors how the DeepSeek and xAI providers wire their base URLs.

Added a regression unit test.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

New tests in `tests/llms/test_anthropic.py` assert the configured `base_url` reaches the client, the `ANTHROPIC_BASE_URL` env var is used as a fallback, and `base_url` is omitted (not forced to `None`) when unset. They fail on the current code and pass with the fix.

## Checklist

### DIFF
--- a/mem0/llms/anthropic.py
+++ b/mem0/llms/anthropic.py
@@ -38,7 +38,11 @@ class AnthropicLLM(LLMBase):
             self.config.model = "claude-sonnet-4-6"
 
         api_key = self.config.api_key or os.getenv("ANTHROPIC_API_KEY")
-        self.client = anthropic.Anthropic(api_key=api_key)
+        base_url = self.config.anthropic_base_url or os.getenv("ANTHROPIC_BASE_URL")
+        client_kwargs = {"api_key": api_key}
+        if base_url:
+            client_kwargs["base_url"] = base_url
+        self.client = anthropic.Anthropic(**client_kwargs)
 
     def _get_common_params(self, **kwargs) -> Dict:
         """Get common parameters, avoiding sending both temperature and top_p together.
--- a/tests/llms/test_anthropic.py
+++ b/tests/llms/test_anthropic.py
@@ -83,6 +83,39 @@ def test_both_set_prefers_temperature_over_top_p(mock_anthropic_client):
     assert "top_p" not in call_kwargs
 
 
+def test_configured_base_url_is_passed_to_client():
+    """A configured anthropic_base_url must reach the Anthropic client constructor."""
+    with patch("mem0.llms.anthropic.anthropic") as mock_anthropic:
+        config = AnthropicConfig(
+            model="claude-3-5-sonnet-20240620",
+            api_key="test-key",
+            anthropic_base_url="https://proxy.example.com",
+        )
+        AnthropicLLM(config)
+
+    assert mock_anthropic.Anthropic.call_args[1]["base_url"] == "https://proxy.example.com"
+
+
+def test_base_url_falls_back_to_env(monkeypatch):
+    """When no base_url is configured, ANTHROPIC_BASE_URL is used."""
+    monkeypatch.setenv("ANTHROPIC_BASE_URL", "https://env.example.com")
+    with patch("mem0.llms.anthropic.anthropic") as mock_anthropic:
+        config = AnthropicConfig(model="claude-3-5-sonnet-20240620", api_key="test-key")
+        AnthropicLLM(config)
+
+    assert mock_anthropic.Anthropic.call_args[1]["base_url"] == "https://env.example.com"
+
+
+def test_base_url_omitted_when_unset(monkeypatch):
+    """With no base_url anywhere, base_url must not be forced to None on the client."""
+    monkeypatch.delenv("ANTHROPIC_BASE_URL", raising=False)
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- `mem0/llms/anthropic.py:41-45` — conditional injection is correct: `base_url` is only passed when truthy, preserving the Anthropic SDK's own default when nothing is configured; passing `None` explicitly would override the SDK's internal fallback
- Mirrors the `api_key` fallback pattern (config → env var) consistently
- `test_base_url_omitted_when_unset` correctly verifies no regression on the unset path
- Clean `client_kwargs` dict approach — avoids passing `None` explicitly

### Issues

1. `mem0-ts/src/oss/src/llms/anthropic.ts:17` — TypeScript `AnthropicLLM` has the identical bug: `new Anthropic({ apiKey })` ignores `config.baseURL` even though `LLMConfig` at `src/oss/src/types/index.ts:43` explicitly declares `baseURL?: string`. A proxy/gateway user configuring `baseURL` in the TS SDK hits the same silent-ignore as this PR fixes in Python. Please either fix it in this PR or open a follow-up issue immediately so it doesn't get lost — per repo standards, a bug fixed in one SDK with the identical bug unflagged in the parallel SDK is a required flag. Fix: `const clientArgs: { apiKey: string; baseURL?: string } = { apiKey }; if (config.baseURL) clientArgs.baseURL = config.baseURL; this.client = new Anthropic(clientArgs);`

#### Minor (optional)
1. `mem0/llms/anthropic.py:44` — an inline comment (`# omit base_url when unset to preserve SDK's built-in ANTHROPIC_BASE_URL fallback`) would prevent a future contributor from "aligning" this with the DeepSeek/xAI pattern (which always passes a hardcoded base URL) and accidentally breaking the default endpoint
2. `mem0/llms/anthropic.py` (pre-existing) — `http_client_proxies` is threaded through the config but the `anthropic.Anthropic(...)` constructor call never receives an `http_client=` argument; proxy support via `AnthropicConfig.http_client_proxies` is silently non-functional — out of scope for this PR but worth a follow-up issue

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 all pass; CI Gate pass. Vercel fail is advisory only.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/llms/test_anthropic.py -v` → 8/8 passed (3 new, 5 pre-existing)
- **Red-green proof:** `test_configured_base_url_is_passed_to_client` and `test_base_url_falls_back_to_env` fail on `main` with `KeyError: 'base_url'`; `test_base_url_omitted_when_unset` passes on both (correct — no regression); all three pass with the fix ✓
- **TS parity gap — confirmed:** `mem0-ts/src/oss/src/llms/anthropic.ts:17` reads as `this.client = new Anthropic({ apiKey });` — `config.baseURL` is never read. `LLMConfig` at `src/oss/src/types/index.ts:43` has `baseURL?: string`. PR #5626 diff touches only Python files. Verified by code read this run.
- **Linked issue:** Closes #5651 — Python fix is correct; TS gap is orthogonal but must be tracked
- **Could NOT verify:** Live Anthropic proxy/gateway behavior — code tracing only

### Assessment
Ready to merge: **With fixes** — please address the Important item (open a follow-up issue for the TS `baseURL` gap, or fix it here); the Python fix itself is correct and I'll re-review once the tracking issue is linked.

yashs33244:
Good catch on the TS parity. The Python fix here is self-contained, so I've opened #5665 to track the identical `baseURL` gap in the TS `AnthropicLLM` rather than widen this PR. The pre-existing `http_client_proxies` issue you flagged is already tracked in #5648. Ready for re-review.

yashs33244:
@kartik-mem0 friendly re-review nudge when convenient. Kept the fix self-contained to the Python `AnthropicLLM` and filed #5665 for the matching TS `baseURL` gap. Thanks!

kartik-mem0 (approved):

PR #5628, opened 2026-06-18
fix(llms): keep configured token limit for reasoning models

### LINKED ISSUE #5653: Configured max_tokens is silently dropped for reasoning models (o1/o3/gpt-5)
### Bug

In `LLMBase._get_supported_params`, the reasoning-model branch (o1/o3/gpt-5) builds the request dict but never sets a token-limit key, so a configured `max_tokens` is silently dropped for those models. Non-reasoning models forward the cap via `_get_common_params`, so the limit only vanishes for reasoning models.

### Impact

A configured `max_tokens` is ignored for reasoning models across every provider that uses this shared helper (OpenAI, Azure OpenAI, DeepSeek, xAI, etc.).

### Expected

The reasoning branch should forward the configured token limit using `max_completion_tokens` (the key reasoning/o-series/gpt-5 models accept).

yashs33244:
## Linked Issue

Closes #5653

## Description

In `LLMBase._get_supported_params`, the reasoning-model branch (o1/o3/gpt-5) builds the request dict but never sets a token-limit key, so a configured `max_tokens` was silently dropped for those models. Non-reasoning models already forward the cap via `_get_common_params`, so the limit only vanished for reasoning models, across every provider that uses this shared helper (OpenAI, Azure OpenAI, DeepSeek, xAI, etc.).

The fix sets `max_completion_tokens` from `self.config.max_tokens` in the reasoning branch. Reasoning/o-series/gpt-5 models use `max_completion_tokens` and reject the legacy `max_tokens`, so this is the correct key.

Added a regression unit test.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

`test_reasoning_model_sends_max_completion_tokens` fails before the fix (no token key sent) and passes after. Full `tests/llms/test_openai.py` passes (21/21).

## Checklist

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -123,6 +123,9 @@ class LLMBase(ABC):
             if "tool_choice" in kwargs:
                 supported_params["tool_choice"] = kwargs["tool_choice"]
 
+            # Reasoning / o-series / gpt-5 use max_completion_tokens, never legacy max_tokens.
+            supported_params["max_completion_tokens"] = self.config.max_tokens
+
             # Add reasoning_effort if configured
             reasoning_effort = getattr(self.config, 'reasoning_effort', None)
             if reasoning_effort:
--- a/tests/llms/test_openai.py
+++ b/tests/llms/test_openai.py
@@ -221,6 +221,27 @@ def test_non_reasoning_model_ignores_reasoning_effort(mock_openai_client):
     assert "reasoning_effort" not in call_kwargs[1]
 
 
+def test_reasoning_model_sends_max_completion_tokens(mock_openai_client):
+    """Reasoning models must forward the configured token cap as max_completion_tokens.
+
+    The reasoning branch previously dropped the cap entirely, so o-series/gpt-5 calls
+    ran with no limit. Reasoning models use max_completion_tokens, never legacy max_tokens.
+    """
+    config = OpenAIConfig(model="o3-mini", max_tokens=100)
+    llm = OpenAILLM(config)
+    messages = [{"role": "user", "content": "Hello"}]
+
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=Mock(content="Response"))]
+    mock_openai_client.chat.completions.create.return_value = mock_response
+
+    llm.generate_response(messages)
+
+    call_kwargs = mock_openai_client.chat.completions.create.call_args[1]
+    assert call_kwargs.get("max_completion_tokens") == 100
+    assert "max_tokens" not in call_kwargs
+
+
 def test_reasoning_effort_config_values():
     """Test that reasoning_effort can be set to all valid values."""
     for effort in ["low", "medium", "high"]:


kartik-mem0:
## Review: fix(llms): keep configured token limit for reasoning models

### Strengths

- **Correct root-cause fix.** `_get_supported_params()` in `LLMBase` is the shared parameter whitelist used by all 5 OpenAI-compatible providers (OpenAI, Azure OpenAI, DeepSeek, xAI, OpenAI Structured). The reasoning branch excluded any token-limit key entirely. The 3-line addition fixes all of them at once — exactly the right place.
- **Correct API key.** OpenAI o-series and gpt-5-base models require `max_completion_tokens` and reject the legacy `max_tokens`. The fix uses the right key unconditionally in this branch.
- **Red-green proven.** `test_reasoning_model_sends_max_completion_tokens` FAILS on main (assert `None == 100` — `max_completion_tokens` absent from call kwargs) and PASSES with the fix. This directly reproduces the reported failure mode.
- **Surgical diff, zero deletions.** No behavior change for non-reasoning models or other providers.
- **Ruff clean, conventional commit title correct (`fix(llms):`).**

### Issues

**Minor — TS SDK parity (pre-existing, not introduced by this PR):** `mem0-ts/src/oss/src/llms/openai.ts` passes no token-limit params at all for any model (reasoning or not). This is a pre-existing gap — worth a follow-up issue for the TS SDK.

**Minor — `test_is_reasoning_model_override_generates_correct_params` (existing test, line 356):** Asserts `"max_tokens" not in call_kwargs` and `"temperature" not in call_kwargs` but not `"max_completion_tokens" in call_kwargs`. The new dedicated test covers this adequately, but extending the existing test would make the contract explicit.

### Evidence

| Check | Result |
|---|---|
| CI Gate | PASS |
| Python SDK 3.10 / 3.11 / 3.12 | PASS / PASS / PASS |
| Vercel | IGNORED (fork deploy-preview, non-blocking) |
| Red-green (with fix) | PASS |
| Red-green (main reverted) | FAIL — `assert None == 100` |
| Full test_openai.py (21 tests) | 21/21 PASS |
| Ruff lint | CLEAN |
| Anthropic regression check | No risk — `AnthropicLLM` overrides `_get_common_params` to use `max_tokens`; no Anthropic model name matches the `o1/o3/gpt-5` heuristic |
| LiteLLM gap | Pre-existing; LiteLLM translates internally for o-series |

### Assessment

**Ready to squash-merge.** (`gh pr merge --squash`)

kartik-mem0:
## Changes Requested

### What this PR does
Adds `max_completion_tokens` to the reasoning-model branch of `LLMBase._get_supported_params()` so configured token limits are no longer silently dropped for o1/o3/gpt-5 models. Root cause is correctly identified, red-green proven, and the fix is minimal.

### Strengths
- `base.py:126-128` — Correct placement: inside the `_is_reasoning_model` branch, structurally mirrors how `_get_common_params` handles the key for non-reasoning models.
- `test_openai.py:224-243` — Strong regression test: asserts both `max_completion_tokens == 100` AND `"max_tokens" not in call_kwargs` — the dual assertion is exactly right for this class of silent-drop bug.
- Red-green proof is clean: test fails on revert with `assert None == 100` (key entirely absent from kwargs), passes with fix. 21/21 tests pass, ruff clean.

### Issues

1. `mem0/llms/base.py:127` — **Unconditional assignment sends `max_completion_tokens=2000` on every reasoning call, even when no cap was configured.**
   `BaseLlmConfig.max_tokens` is typed `int = 2000` (not `Optional[int]`), so `self.config.max_tokens` is always `2000` unless explicitly overridden. Before this PR, the reasoning branch returned without any token-limit key — reasoning calls were unconstrained. After this PR, they are silently capped at 2000 tokens by default. For o3/o1 deployments running long chain-of-thought tasks this can truncate responses mid-sentence.
   **Fix:** Change `BaseLlmConfig.max_tokens` to `Optional[int] = None` and add a guard:
   ```python
   if self.config.max_tokens is not None:
       supported_params["max_completion_tokens"] = self.config.max_tokens
   ```

2. `mem0/llms/base.py:127` — **Non-OpenAI providers share this code path and would receive `max_completion_tokens` on API calls that reject it.**
   `vllm.py:94`, `lmstudio.py:94`, `deepseek.py:94`, `xai.py:95`, `minimax.py:98` all call `self._get_supported_params(...)` with no override. `_is_reasoning_model()` is purely name-based — a user serving `model="o3-mini"` locally via LM Studio or vLLM triggers the reasoning branch and now gets `max_completion_tokens` injected. vLLM's OpenAI-compatible API uses `max_tokens`, not `max_completion_tokens`; this would produce a 400 or silently drop the parameter.
   **Fix:** The `if self.config.max_tokens is not None:` guard from finding 1 also addresses this for the default-config case. For explicit configs, the `is_reasoning_model=False` escape hatch already exists — worth noting in the comment that non-OpenAI operators serving reasoning-named models should set it.

#### Minor (optional)

- `test_openai.py:236` — Uses `call_args[1]` (positional index); rest of the file uses `call_args.kwargs`. Cosmetic only.
- `base.py:124` — Comment "Reasoning / o-series / gpt-5 use max_completion_tokens, never legacy max_tokens" is slightly misleading: gpt-5.x fine-tunes (e.g. `gpt-5.4-mini`) route through `_get_common_params` via `_uses_max_completion_tokens`, not this branch.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 PASS, CI Gate PASS. Vercel ignored (fork deploy-preview, non-blocking).
- **Targeted tests:** `pytest tests/llms/test_openai.py -v --tb=short` → 21 passed in 8.61s
- **Red-green proof:** `test_reasoning_model_sends_max_completion_tokens` fails on revert (`assert None == 100` — `max_completion_tokens` absent from call kwargs), passes with fix ✓
- **Linked issue:** #5653 — PR correctly addresses the root cause.
- **Cross-package check:** TS `mem0-ts/src/oss/src/llms/openai.ts` has no reasoning-model gating — pre-existing drift, not introduced by this PR.
- **Could NOT verify:** Live API behavior on vLLM/LMStudio with `max_completion_tokens` — code tracing only.

### Assessment
Ready to merge: **With fixes** — please address the two Important items above; I'll re-review once you push. The core approach is right and the fix is close — the main gap is the missing `None`-guard that would silently change behavior for users on default configs and for non-OpenAI OpenAI-compatible backends.

yashs33244:
Thanks @kartik-mem0. I dug into both points and want to flag a safety issue with the `Optional[int] = None` suggestion before making it:

**Making `BaseLlmConfig.max_tokens` Optional/None looks unsafe.** `max_tokens` is read directly by ~8 providers (anthropic, gemini, ollama, groq, together, litellm, azure_openai_structured, plus base). Anthropic's API *requires* `max_tokens` - a `None` there is a hard error - and the others would start sending `max_tokens=None`. So flipping the default would regress non-reasoning providers, well outside this PR's scope.

**On the "silent 2000 cap" for reasoning models:** the reasoning branch's `max_completion_tokens = self.config.max_tokens` mirrors `_get_common_params`, which already sends `max_tokens = self.config.max_tokens` (default 2000) for every non-reasoning model (and `max_completion_tokens` for gpt-5). So this PR doesn't add a cap unique to reasoning models - it makes them consistent with the rest of the codebase, which is the gap #5653 reported. If 2000 is too low as a global reasoning default, that feels like a separate config-default discussion - and given the Anthropic constraint above, the safe lever there isn't `None`.

**On non-OpenAI backends:** fair point that an o-series-*named* model served via vLLM/LMStudio could now receive `max_completion_tokens`. Two notes: the gpt-5 path already routes through `max_completion_tokens` in `_get_common_params`, so it's consistent; and `is_reasoning_model=False` is the existing opt-out for those deployments. If you'd still like a hard guard, I can gate the reasoning token key behind the same `_uses_max_completion_tokens` / a provider-aware check - happy to push that.

How would you like me to proceed - keep it consistent as-is with a short doc note on the `is_reasoning_model=False` escape hatch, or add the gate? Either is quick.

yashs33244:
@kartik-mem0 gentle nudge on this one. Since the only open item is which direction you'd prefer, I'm happy to just push the provider-aware gate - add `max_completion_tokens` only when the model actually uses it, so a reasoning-*named* model served on vLLM/LMStudio stays on `max_tokens`. That turns this into mergeable code instead of an open decision, and it addresses the non-OpenAI-backend concern without the unsafe `Optional[None]` flip. Want me to push it, or keep it consistent as-is with the `is_reasoning_model=False` note?

kartik-mem0:

### What this PR does
Forwards the configured token cap as `max_completion_tokens` in the reasoning-model branch of `LLMBase._get_supported_params`, which previously built the request dict without any token limit, so o-series/gpt-5 calls ran uncapped.

### Strengths
- `mem0/llms/base.py:127` — root-cause fix at the single shared helper, so every reasoning-capable provider (OpenAI/Azure/DeepSeek/xAI) is fixed at once, not per-provider.
- Hardcoding `max_completion_tokens` (not the `_uses_max_completion_tokens` split) is correct here — the reasoning branch is only ever reached by models that reject legacy `max_tokens`.
- `tests/llms/test_openai.py:224` — asserts both `max_completion_tokens == 100` and `max_tokens` absent; end-to-end through the real `OpenAILLM.generate_response`, not just the helper.

### Issues
Minor (optional): none

### Evidence
- **CI:** Python SDK build 3.10/3.11/3.12 pass, CI Gate + changelog + cla pass (Vercel preview fail ignored — not merge-blocking).
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/llms/test_openai.py::test_reasoning_model_sends_max_completion_tokens -v` → 1 passed; full `test_openai.py` → 21 passed.
- **Red-green proof:** revert `base.py` to main → `AssertionError: assert None == 100` (reverted branch emits only `{messages, model}`); restore → passes ✓
- **Linked issue:** #5653 — PR addresses the reported root cause (cap silently dropped for reasoning models).
- **Cross-package check:** TS OSS SDK has no reasoning-model `max_completion_tokens` branch, so nothing to drift against here.

### Assessment
Small, correctly-placed fix at the shared choke point with a test that bites; reasoning models now respect the configured `max_tokens`.

yashs33244:
Gentle nudge on this one - @kartik-mem0 reviewed it favourably a little while back (shared param whitelist fix, red-green tested). @deshraj @kartik-mem0 if either of you gets a moment for a final look or merge, I'd be grateful. No urgency, and happy to update if anything's drifted on main.

yashs33244:
Gentle nudge on this one 🙏 @kartik-mem0 you'd marked the reasoning-model token-limit fix ready for squash-merge and CI is green - whenever you or @deshraj have a moment to merge, that'd be great. Thanks!

PR #5629, opened 2026-06-18
fix(memory): return attributed_to from get/get_all/search

### LINKED ISSUE #5654: attributed_to is written on add but never returned by get/get_all/search
### Bug

`attributed_to` is written into the memory payload on `add`, but no read API returns it. The read methods promote a fixed set of payload keys (`promoted_payload_keys`) onto the top-level dict; `attributed_to` is in `core_and_promoted_keys` (kept out of `metadata`) but missing from `promoted_payload_keys`, so it falls through the gap and is dropped by `get`, `get_all`, and `search` (sync and async).

### Impact

`attributed_to` is effectively a write-only dead value, never returned by `get`, `get_all`, or `search`.

### Expected

`attributed_to` should be promoted onto the top-level result and returned by all read APIs when present in the payload.

yashs33244:
## Linked Issue

Closes #5654

## Description

`attributed_to` is written into the memory payload on `add` (and the extraction prompt marks it required, so it's populated for inferred memories), but none of the read APIs ever return it. The read methods build their result by promoting a fixed set of payload keys (`promoted_payload_keys`) onto the top-level dict; `attributed_to` was listed in `core_and_promoted_keys` (so it's correctly kept out of the `metadata` bucket) but was missing from `promoted_payload_keys`, so it fell through the gap and was silently dropped by `get`, `get_all`, and `search` (sync and async). It was effectively a write-only dead value.

The fix adds `"attributed_to"` to `promoted_payload_keys` at all six read sites (3 sync + 3 async). Promotion stays guarded by `if key in payload`, so memories without the field are unaffected.

Added a regression unit test covering `get`/`get_all`/`search` on both `Memory` and `AsyncMemory`.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1091,6 +1091,7 @@ class Memory(MemoryBase):
             "run_id",
             "actor_id",
             "role",
+            "attributed_to",
         ]
 
         core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
@@ -1204,6 +1205,7 @@ class Memory(MemoryBase):
             "run_id",
             "actor_id",
             "role",
+            "attributed_to",
         ]
         core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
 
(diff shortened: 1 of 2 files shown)


yashs33244:
Thanks! Keeping this PR Python-only as you suggested — opened #5666 to track the identical `attributedTo` gap in the TS SDK (get/search/getAll + the `MemoryItem` type + a Jest test). Ready for re-review.

yashs33244:
@kartik-mem0 ready for re-review when you get a chance. Kept this Python-only as you suggested and opened #5666 to track the identical `attributedTo` gap in the TS SDK. Appreciate the thorough pass.

kartik-mem0 (approved):

PR #5630, opened 2026-06-18
fix(graph): keep distinct entities that share a substring prefix

### LINKED ISSUE #5657: Graph entity dedup drops distinct entities sharing a substring prefix (e.g. "Sam" dropped when "Samsung" present)
### Bug

In the graph memory path, `_extract_entities_from_doc` drops entities that are substrings of longer ones, but the dedup check is unanchored (`e.lower() in o`). A shorter entity is removed whenever it appears anywhere inside a longer one, including mid-word. So "Sam" is dropped when "Samsung" is present, and any short proper noun that happens to be a leading substring of a longer entity disappears.

### Impact

Distinct entities that merely share a substring prefix are silently lost. Since both entity linking (`_link_entities_for_memory`) and search boosting (`_compute_entity_boosts`) consume this output, the dropped entity is missing in both paths, hurting recall.

### Expected

Only whole-word token subsets should be dropped (e.g. "learning" inside "machine learning", "Apple" inside "Apple Inc"). Distinct entities like "Sam" and "Samsung" must both be kept.

The TypeScript counterpart is tracked in #5646.

yashs33244:
## Linked Issue

Closes #5657

## Description

`_extract_entities_from_doc` drops entities that are substrings of longer ones, but the check was unanchored (`e.lower() in o`). That means a shorter entity is removed whenever it appears *anywhere* inside a longer one, including mid-word. So "Sam" gets dropped when "Samsung" is present, and any short proper noun that happens to be a leading substring of a longer entity disappears.

Since both entity linking (`_link_entities_for_memory`) and search boosting (`_compute_entity_boosts`) consume this output, the dropped entity is lost in both paths, costing recall.

The fix anchors the match on word boundaries (`re.search(rf"\b{re.escape(e.lower())}\b", o)`), so only whole-word token subsets are dropped ("learning" still drops out of "machine learning", "Apple" out of "Apple Inc"), while distinct entities that merely share a substring prefix are kept. `re.escape` guards against regex-special characters in entity text.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `test_substring_dedup_respects_word_boundaries` asserting "Sam" survives alongside "Samsung". It fails on the old unanchored check and passes with the fix.

## Checklist

### DIFF
--- a/mem0/utils/entity_extraction.py
+++ b/mem0/utils/entity_extraction.py
@@ -352,7 +352,7 @@ def _extract_entities_from_doc(doc) -> List[Tuple[str, str]]:
             best[k] = (t, e)
     deduped = list(best.values())
 
-    # Remove entities that are whole-word subsequences of longer entities.
+    # Remove entities that are whole-word substrings of longer entities.
     # Word-boundary anchoring avoids dropping distinct entities that only share a
     # leading substring (e.g. "Sam" must survive alongside "Samsung").
     all_lower = [e[1].lower() for e in deduped]
--- a/mem0/utils/entity_extraction.py
+++ b/mem0/utils/entity_extraction.py
@@ -352,6 +352,12 @@ def _extract_entities_from_doc(doc) -> List[Tuple[str, str]]:
             best[k] = (t, e)
     deduped = list(best.values())
 
-    # Remove entities that are substrings of longer entities
+    # Remove entities that are whole-word subsequences of longer entities.
+    # Word-boundary anchoring avoids dropping distinct entities that only share a
+    # leading substring (e.g. "Sam" must survive alongside "Samsung").
     all_lower = [e[1].lower() for e in deduped]
-    return [(t, e) for t, e in deduped if not any(e.lower() != o and e.lower() in o for o in all_lower)]
+    return [
+        (t, e)
+        for t, e in deduped
+        if not any(e.lower() != o and re.search(rf"\b{re.escape(e.lower())}\b", o) for o in all_lower)
+    ]
--- a/tests/utils/test_entity_extraction.py
+++ b/tests/utils/test_entity_extraction.py
@@ -59,6 +59,16 @@ class TestExtractEntities:
         google_count = sum(1 for _, t in entities if "Google" in t)
         assert google_count <= 1, f"Expected dedup, got {entities}"
 
+    def test_substring_dedup_respects_word_boundaries(self):
+        from mem0.utils.entity_extraction import extract_entities
+
+        # "Sam" is a mid-word substring of "Samsung", not a separate token, so it
+        # must not be dropped as a substring of the longer entity.
+        entities = extract_entities("At Samsung, Sam leads design.")
+        entity_texts = [e[1] for e in entities]
+        assert "Sam" in entity_texts, f"Expected 'Sam' to survive alongside 'Samsung', got {entities}"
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Good catch on the word-boundary anchoring — this is a real recall bug and the fix is clean.

## Strengths

- Root cause is correctly identified: `e.lower() in o` is an unanchored substring check, so `"sam" in "samsung"` returns `True` and the entity is silently dropped from both `_link_entities_for_memory` and the search-time entity boosts. That's a silent recall hit on both write and query paths.
- `re.escape()` is the right defensive move — entity text with regex metacharacters (e.g., `C++`, `U.S.A`) won't break the pattern.
- `re` was already imported; no new dependency.
- The test exercises the full `extract_entities` pipeline through spaCy, not a stub of the dedup function — that matters.
- Verified that the desired suppression behavior is preserved: `\blearning\b` still matches inside `"machine learning"`, so legitimate whole-word subsets are still dropped.

## Issues

**One nit before merge:** The replacement comment on line 355 says *"whole-word subsequences"* — but "subsequence" has a specific meaning in CS (non-contiguous characters in order, à la LCS). What the code actually enforces is **whole-word substring** containment. Worth a one-word fix to avoid confusing the next maintainer. The TS equivalent at `mem0-ts/src/oss/src/utils/entity_extraction.ts:703` already says "substrings" correctly.

**Follow-up (separate PR, not a blocker here):** The TS side has the identical bug at `mem0-ts/src/oss/src/utils/entity_extraction.ts:710` — `other.includes(entity.text.toLowerCase())` is the same unanchored check. Same "Samsung"/"Sam" scenario drops "Sam" from TS entity outputs. Worth a companion fix there.

## Evidence

- CI Gate: pass — Python SDK 3.10 / 3.11 / 3.12 all green; changelog_check pass
- Worktree red-green: test `FAILED` on pre-PR code (`assert 'Sam' in ['Samsung']`), `PASSED` with fix; test exercises the real spaCy pipeline
- Ruff: clean
- Whole-word suppression verified: `\blearning\b` matches inside `"machine learning"` (space before `l` = word boundary), so "learning" still suppressed correctly

One-line fix to the comment + a follow-up issue for the TS counterpart and this is ready to squash-merge.

Would you mind also opening a tracking issue for the TS-side fix? Helps keep it from falling through the cracks.

yashs33244:
Thanks for the thorough review! Both addressed:

- Fixed the comment wording — it now reads "whole-word **substrings**" (pushed in 8de2beb), matching the TS side.
- Opened #5646 to track the identical TS-side bug at `entity_extraction.ts:710`.

Ready when you are.

kartik-mem0:
Re-review — new head `8de2beb` (`chore: clarify entity dedup comment`).

Verified the single commit pushed since the prior LGTM:
- **Comment fix confirmed** (`entity_extraction.py:355`): "whole-word subsequences" → "whole-word substrings". Matches the TS-side comment and fixes the Important finding from the prior review.
- **TS tracking issue confirmed**: #5646 is open with the correct diagnosis and points at `entity_extraction.ts:710`.
- **CI on new head**: CI Gate pass, Python SDK 3.10 / 3.11 / 3.12 all green. Vercel advisory failure ignored per pipeline rules.
- **No functional changes** in the new commit — implementation and tests are identical to the prior LGTM'd head `500dcf9c`; red-green proof from that run still holds.

Both issues raised are fully addressed. LGTM — ready to squash-merge (`gh pr merge --squash`).

kartik-mem0 (approved):

PR #5631, opened 2026-06-18
fix(memory): guard against malformed image_url in parse_vision_messages

### LINKED ISSUE #5659: Malformed image message raises uncaught KeyError in parse_vision_messages, aborting add()
### Bug

In `parse_vision_messages`, a single image content part is dereferenced as `content["image_url"]["url"]` outside the surrounding `try`. When an image part is malformed (missing the nested `url`), this raises an uncaught `KeyError` that bubbles all the way up.

### Impact

A malformed image message aborts the entire `add()` call with a raw `KeyError` instead of the intended download/validation error path, so callers get an unactionable traceback.

### Expected

A missing `url` should produce a clear, actionable error (a `ValueError`) rather than an uncaught `KeyError`.

yashs33244:
## Linked Issue

Closes #5659

## Description

In `parse_vision_messages`, a single image content part was dereferenced as `content["image_url"]["url"]` outside the surrounding `try`. When an image part is malformed (missing the nested `url`), this raised an uncaught `KeyError` that bubbled all the way up and aborted `add()`, instead of the intended download/validation error path.

The fix reads the URL defensively with `content.get("image_url", {}).get("url")` and raises a clear `ValueError` when it's absent, so callers get an actionable message instead of a `KeyError`.

Added a regression unit test.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/utils/memory.ts
+++ b/mem0-ts/src/oss/src/utils/memory.ts
@@ -31,9 +31,11 @@ const parse_vision_messages = async (messages: Message[]) => {
         typeof message.content === "object" &&
         message.content.type === "image_url"
       ) {
-        const description = await get_image_description(
-          message.content.image_url.url,
-        );
+        const imageUrl = message.content.image_url?.url;
+        if (!imageUrl) {
+          throw new Error("image_url content part is missing image_url.url");
+        }
+        const description = await get_image_description(imageUrl);
         new_message.content =
           typeof description === "string"
             ? description
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -206,7 +206,9 @@ def parse_vision_messages(messages, llm=None, vision_details="auto"):
         elif isinstance(content, dict) and content.get("type") == "image_url":
             if llm is None:
                 continue
-            image_url = content["image_url"]["url"]
+            image_url = content.get("image_url", {}).get("url")
+            if not image_url:
+                raise ValueError("image_url content part is missing image_url.url")
             try:
                 description = get_image_description(image_url, llm, vision_details)
                 returned_messages.append({"role": role, "content": description})
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -206,7 +206,8 @@ def parse_vision_messages(messages, llm=None, vision_details="auto"):
         elif isinstance(content, dict) and content.get("type") == "image_url":
             if llm is None:
                 continue
-            image_url = content.get("image_url", {}).get("url")
+            image_url_obj = content.get("image_url")
+            image_url = image_url_obj.get("url") if isinstance(image_url_obj, dict) else None
             if not image_url:
                 raise ValueError("image_url content part is missing image_url.url")
             try:
--- a/tests/memory/test_memory_utils.py
+++ b/tests/memory/test_memory_utils.py
@@ -101,7 +101,16 @@ class TestParseVisionMessages:
         # uncaught KeyError that aborted add(); it should raise a clear ValueError.
         mock_llm = Mock()
         messages = [{"role": "user", "content": {"type": "image_url", "image_url": {}}}]
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match=r"missing image_url\.url"):
+            parse_vision_messages(messages, llm=mock_llm)
+        mock_llm.generate_response.assert_not_called()
+
+    def test_none_image_url_raises_value_error(self):
+        # image_url present but None (or any non-dict) must also raise the clear
+        # ValueError, not an AttributeError from calling .get() on None.
+        mock_llm = Mock()
+        messages = [{"role": "user", "content": {"type": "image_url", "image_url": None}}]
+        with pytest.raises(ValueError, match=r"missing image_url\.url"):
             parse_vision_messages(messages, llm=mock_llm)
         mock_llm.generate_response.assert_not_called()
 
--- a/tests/memory/test_memory_utils.py
+++ b/tests/memory/test_memory_utils.py
@@ -96,6 +96,15 @@ class TestParseVisionMessages:
         result = parse_vision_messages(messages, llm=None)
         assert result == messages
 
+    def test_malformed_image_dict_raises_value_error(self):
+        # A malformed image part (missing the nested url) used to raise an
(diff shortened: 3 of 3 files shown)


kartik-mem0:
## Review: fix(memory): guard against malformed image_url in parse_vision_messages

Thanks for the clear fix and the linked regression test — the red-green proof ran cleanly.

### Strengths

- Root cause correctly identified: `content["image_url"]["url"]` was the unguarded site and the fix eliminates it.
- The `ValueError` with an explicit message is the right choice over `continue` — a malformed content part deserves a clear signal, not silent data loss.
- The fix covers both `Memory` (sync) and `AsyncMemory` (async) paths automatically since `parse_vision_messages` is a pure sync utility called identically from both.
- `ruff check` passes; code style is consistent with the project conventions.
- The regression test red-greens correctly: reverts to `KeyError` on main, passes with the fix.

---

### Issues

#### Important — `image_url: None` still raises `AttributeError`, not `ValueError`

`mem0/memory/utils.py`, line 209 (post-patch):

```python
image_url = content.get("image_url", {}).get("url")
```

`dict.get(key, default)` returns the *stored value* when the key exists — the default only fires when the key is **absent**. So when `content = {"type": "image_url", "image_url": None}`, `content.get("image_url", {})` returns `None`, and then `None.get("url")` raises `AttributeError` before the `if not image_url:` guard can fire.

This is a narrower failure mode than the original bug (requires `None` as the explicit value, not just a missing key), and no standard LLM API produces this shape. Still, the fix is slightly incomplete for this input class.

Suggested fix:

```python
image_url_obj = content.get("image_url")
image_url = image_url_obj.get("url") if isinstance(image_url_obj, dict) else None
if not image_url:
    raise ValueError("image_url content part is missing image_url.url")
```

#### Important — TypeScript SDK has the identical unguarded access

`mem0-ts/src/oss/src/utils/memory.ts`, line 35:

```typescript
const description = await get_image_description(
  message.content.image_url.url,   // no null guard
);
```

If `image_url` is absent or `url` is missing, this throws a `TypeError` at runtime — the same failure class as the Python bug being fixed. A minimal fix:

```typescript
const imageUrl = (message.content.image_url as { url?: string } | undefined)?.url;
if (!imageUrl) {
  throw new Error("image_url content part is missing image_url.url");
}
const description = await get_image_description(imageUrl);
```

Would you be able to include a TS-side fix in this PR, or open a follow-up? The inconsistency is in the same function (`parse_vision_messages`) in the same utility file — it's the natural scope for this change.

---

### Minor

- `pytest.raises(ValueError)` without `match=` will pass on any `ValueError` from anywhere in the call stack. Using `pytest.raises(ValueError, match=r"missing image_url\.url")` pins the test to the specific guard you added.

---

### Evidence

- CI: Python SDK matrix (3.10, 3.11, 3.12) all pass. CI Gate passes. Vercel check is red (advisory, expected for fork PRs — not merge-blocking).
- Red-green: test **FAILS** on main with `KeyError: 'url'` at `utils.py:209`; **PASSES** with fix; **PASSES** again after restore. Proof is clean and unambiguous.
- Lint: `ruff check mem0/memory/utils.py` — all checks passed.

---

### Assessment

**Changes Requested.** The core Python fix is correct and the test is solid. Two items before merge: (1) guard the `image_url: None` edge case with an `isinstance` check, and (2) fix the identical unguarded access in the TypeScript SDK. The minor `match=` addition is optional but recommended.

yashs33244:
Thanks @kartik-mem0 - addressed all three:

1. **`image_url: None` (or any non-dict) edge case.** Routed the access through an `isinstance` check (`utils.py`), so a present-but-None / non-dict `image_url` now raises the same clear `ValueError` instead of an `AttributeError` from `.get()` on `None`. Added `test_none_image_url_raises_value_error` - red-green proven (it fails with `AttributeError: 'NoneType' object has no attribute 'get'` on the old code, passes with the guard).
2. **TS SDK parity.** Mirrored the guard in `mem0-ts/src/oss/src/utils/memory.ts`: `const imageUrl = message.content.image_url?.url;` then throw the same `image_url content part is missing image_url.url` message before fetching - so the TS path no longer hits a raw `Cannot read properties of undefined`.
3. **Pinned the test** with `match=r"missing image_url\.url"` on both cases.

`ruff check` clean. Ready for re-review.

kartik-mem0:

Verified against 85f6e37: all three findings from the 2026-06-19 review are addressed.

### What's fixed

- **`image_url: None` edge case** — `utils.py:209`: `isinstance(image_url_obj, dict)` branches any non-dict value (None, int, list) to `image_url = None` then `raise ValueError`. Test `test_none_image_url_raises_value_error` red-greens cleanly: fails with `TypeError: 'NoneType' object is not subscriptable` on main, passes with the fix.
- **TypeScript parity** — `mem0-ts/src/oss/src/utils/memory.ts:35`: `message.content.image_url?.url` with `if (!imageUrl) throw new Error(...)` mirrors the Python guard and uses the same error message. TS SDK CI is now exercised and green (Node 20 + 22).
- **`match=` pins** — both regression tests use `match=r"missing image_url\.url"`, binding the assertion to the specific guard.

### Strengths

- `isinstance` guard is the correct shape — covers None, any non-dict, and missing key in a single branch with no ambiguity.
- Python and TS now have matching contracts: same failure class, same error message, same pre-validation pattern.
- All 8 `TestParseVisionMessages` tests pass; no regressions.

### Evidence

- **CI:** Python SDK 3.10/3.11/3.12 ✓ · TypeScript SDK Node 20/22 ✓ · CI Gate ✓ · Vercel advisory (expected for fork PRs, not merge-blocking).
- **Targeted tests:** `pytest tests/memory/test_memory_utils.py::TestParseVisionMessages -v` → 8 passed in 0.60s.
- **Red-green (new test):** `test_none_image_url_raises_value_error` fails with `TypeError` on main at `utils.py:209`; passes with fix; passes again after restore.
- **Lint:** `ruff check mem0/memory/utils.py tests/memory/test_memory_utils.py` — all checks passed.
- **Could NOT verify:** TS unit test for `parse_vision_messages` with malformed input (no TS test file for this function); full `Memory.add()` integration with a live LLM.

### Assessment

The core fix is correct and complete. Both SDK paths guarded, both tests bite, CI fully green across the matrix.

kartik-mem0 (approved):

PR #5632, opened 2026-06-18
fix(memory): keep newest messages when evicting a batched save

### LINKED ISSUE #5661: SQLite history buffer keeps oldest 10 messages instead of newest when saving a large batch
### Bug

In the SQLite history buffer, when more than 10 messages are saved in a single `save_messages` call, the buffer keeps the *oldest* 10 instead of the newest 10. `save_messages` stamps one `created_at` on the whole batch, so the eviction query (`ORDER BY created_at DESC LIMIT 10`) ties across every row and SQLite breaks the tie by rowid ascending, retaining the oldest rows. `get_last_messages` has the same `created_at` ordering problem.

### Impact

After a large batch save, the newest messages are evicted and the next `add()` reads stale context back, degrading memory quality.

### Expected

When more than 10 messages are saved in one call, the buffer should retain the newest 10, and retrieval should return them in correct insertion order.

yashs33244:
## Linked Issue

Closes #5661

## Description

When more than 10 messages are saved in a single `save_messages` call, the buffer should keep the newest 10. It actually keeps the *oldest* 10.

The root cause is that `save_messages` stamps one `created_at` on the whole batch. The eviction query (`ORDER BY created_at DESC LIMIT 10`) then ties across every row, and SQLite breaks the tie by rowid ascending, so the oldest rows are retained. The next `add()` reads stale context back via `get_last_messages`, which had the same `created_at` ordering problem.

Fix: order eviction and retrieval by `rowid` (insertion order), which is strictly increasing per insert and never ties within a batch. The `messages` table has a `TEXT PRIMARY KEY`, so it is a normal rowid table and `rowid` is available.

Added a regression unit test that saves 12 messages in one call and asserts the newest 10 (msg2..msg11) remain.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

## Checklist


kartik-mem0:
## Review — fix(memory): keep newest messages when evicting

**Verdict: Changes Requested** | Score 8/10 | ~3 min to merge after fix

---

### Strengths

The root-cause diagnosis is exactly right. `save_messages` stamps one `created_at` for the entire batch, so `ORDER BY created_at DESC LIMIT 10` ties across all rows and SQLite breaks the tie by rowid ascending — keeping the *oldest* rows. Switching to `ORDER BY rowid DESC` is the correct fix: rowid is monotonically increasing per insert, never ties within a batch, and is available on this table because `id TEXT PRIMARY KEY` creates a standard rowid table. Both `save_messages` (eviction) and `get_last_messages` (retrieval) are updated together, which is the right scope. The regression test is well-constructed — saves 12 messages in one call and asserts exactly `msg2..msg11` survive, which is the failure mode, not just a mechanism assertion.

---

### Issues

**Important — TS SDK has the identical bug**

`mem0-ts/src/oss/src/storage/SQLiteManager.ts` has the same pattern:
- Line 93: `const now = new Date().toISOString()` — single timestamp for the whole batch  
- Line 87 (evict statement): `ORDER BY created_at DESC LIMIT 10`  
- Lines 118–124 (`getLastMessages`): inner + outer both use `ORDER BY created_at`

The fix there is the same mechanical change: switch both to `ORDER BY rowid DESC`/`rowid ASC`, and add `rowid` to the inner `SELECT` list of `getLastMessages` so the outer `ORDER BY rowid ASC` resolves correctly.

Either fix the TS side in this PR (the change is 3 lines), or open a tracking issue before merge. The Python half is solid on its own, but it's worth not leaving the TS SDK in the same broken state.

---

### Minor (non-blocking)

- No test for `get_last_messages` ordering specifically (separate from eviction correctness) — a short test saving 5 messages and asserting retrieval order would close the gap
- A comment on `_create_messages_table` noting the `rowid` dependency (i.e., `WITHOUT ROWID` must not be added) would protect future schema work
- `rowid` is selected in the `get_last_messages` inner query solely to make it available to the outer `ORDER BY` — a brief comment would prevent maintainers from thinking it's safe to remove

---

### Evidence

- **CI**: Python 3.10/3.11/3.12 all pass. CI Gate green. Vercel is an advisory deploy check (fork auth) — ignored.
- **Red-green**: Test FAILS on main (`msg0..msg9` retained — oldest 10). Test PASSES on PR head (`msg2..msg11` retained — newest 10). Full `test_storage.py` suite: 15/15 on PR head.
- **Lint**: `ruff check mem0/memory/storage.py tests/memory/test_storage.py` — all checks passed.
- **Multi-call behavior verified**: Messages from prior calls get lower rowids; eviction correctly keeps the most recently inserted across batch boundaries.
- **TS gap confirmed by code read**: `SQLiteManager.ts` lines 87, 118–124 — identical `created_at` ordering, same single-timestamp-per-batch pattern.

---

Also, if you haven't already — worth opening a tracking issue for this bug so it's on the backlog. Happy to help with that if useful.

yashs33244:
Thanks for the thorough review @kartik-mem0! Addressed all of it in 8d17ba7:

- **TS SDK** — applied the same `rowid` fix to `mem0-ts/src/oss/src/storage/SQLiteManager.ts` (both the evict statement and `getLastMessages`), and added a jest regression test (a 12-message batched save keeps the newest 10; proven red-green against the old `created_at` ordering). So the Python and TS SDKs are consistent now rather than leaving TS in the same broken state.
- **`get_last_messages` ordering test** — added on the Python side (5 messages, asserts retrieval order).
- **Comments** — noted the `rowid` dependency on the `messages` table (don't switch it to `WITHOUT ROWID`) and why `rowid` is selected in the inner `getLastMessages` query.

Let me know if you'd like anything else.

kartik-mem0:
hi @yashs33244 please create the issue with the prs so that it is easy to track and triage.

regarding this pr, can you please show some concrete proofs with real envs as i think this solution does not account that.

thanks.

yashs33244:
Already done — opened #5661 and linked it, so the PR body now carries `Closes #5661`. Ready for review/merge whenever.

kartik-mem0:
what i mean to say is please show some eg that what you are fixing is an bug because rn the solution is not sound and correct.

yashs33244:
Thanks @kartik-mem0 - fair ask. Here's a concrete repro using the **real `SQLiteManager`** from both `main` and this branch (no mocks, actual `save_messages` + `get_last_messages`).

The trigger is a single `add()` of a conversation with more than 10 messages. `save_messages` computes `now = datetime.now(timezone.utc).isoformat()` **once** (storage.py:266) and stamps it on every row in the batch, so all rows share one `created_at`. `ORDER BY created_at DESC LIMIT 10` then ties across the whole batch, and SQLite resolves the tie by rowid ascending - keeping the **oldest** 10 and evicting the newest.

```python
# real SQLiteManager loaded from each branch; one batch of 12 msgs (msg0..msg11)
db.save_messages([{"role": "user", "content": f"msg{i}"} for i in range(12)], "session-A")
[m["content"] for m in db.get_last_messages("session-A", limit=10)]
```

Output (3 trials each, SQLite 3.50.3):

```
upstream/main  (ORDER BY created_at DESC):
  kept ['msg0' .. 'msg9']    -> WRONG  (oldest 10 retained, 2 newest evicted)
PR #5632       (ORDER BY rowid DESC):
  kept ['msg2' .. 'msg11']   -> OK     (newest 10 retained)
```

So a 12-turn add silently drops the two most recent turns and keeps the two oldest - the opposite of "keep the last 10."

On soundness of the fix: `rowid` is unique and monotonic per insert, so it never ties within a batch and orders strictly by insertion. It's safe here because `id TEXT PRIMARY KEY` makes this a normal rowid table. The only ways rowid order could drift are `WITHOUT ROWID` (guarded against with the schema comment) and `VACUUM` - and VACUUM renumbers rowids *in existing rowid order*, so it preserves the ordering. Across batches we always retain the highest rowids, so later inserts get still-higher rowids and the invariant holds.

The existing `test_storage.py` regression asserts exactly this (msg2..msg11 survive, red-green on `main`), and the backing issue is #5661. Happy to share the full repro script or adjust anything.

yashs33244:
@kartik-mem0 following up here. The repro above runs the real `SQLiteManager` from both branches: a single 12-message `add()` silently drops the two newest turns on `main` (ties on the shared `created_at`, keeps the oldest 10) and is fixed by ordering on `rowid`. Backing issue is #5661, both Python + TS SDKs and tests are updated. Anything else you'd like to see, or good to merge?

kartik-mem0 (commented):
## Changes requested

Eviction fix is solid and proven in both SDKs; the read-order regression test doesn't test what it claims.

**To unblock:**
1. `tests/memory/test_storage.py:269` — insertion-order test passes pre-fix too
2. `mem0-ts/src/oss/tests/storage.unit.test.ts:133` — same gap in the TS test

Optional: `storage.py:282` lost its derived-table comment rationale; `get_history` has a similar tie risk, worth a follow-up issue.

- **Problem is real:** reproduces on `main`, `save_messages` stamps one `created_at` per batch, no tiebreak in the eviction or read query
- **Tests / red-green:** eviction test fails pre-fix, passes post-fix, both languages. Insertion-order test passes on BOTH pre-fix and post-fix code (Python 15/1, TS 23/1), confirmed independently twice
- **CI:** all required checks green, only Vercel (advisory, deploy-preview auth) fails
- **Security:** cleared, parameterized queries unchanged, `session_scope` tenancy clause untouched
- **Bloat (ponytail):** lean already, no cut needed
- **Not verified:** concurrent/multi-process SQLite access under load

</details>

The rowid fix itself (both queries, both SDKs) is correct and well reasoned, keep that. Push when ready and I will re-review.

tests/memory/test_storage.py
```diff
@@ -248,6 +248,26 @@ def test_large_batch_operations(self, sqlite_manager):
...
+        assert kept == [f"msg{i}" for i in range(2, 12)]
+
+    def test_get_last_messages_returns_insertion_order(self, memory_manager):
+        """get_last_messages must return messages in insertion order, even within one batched save."""
+        messages = [{"role": "user", "content": f"msg{i}", "name": None} for i in range(5)]
+        memory_manager.save_messages(messages, "scope-b")
+
+        order = [m["content"] for m in memory_manager.get_last_messages("scope-b", limit=10)]
+
+        assert order == [f"msg{i}" for i in range(5)]
```
kartik-mem0: This passes even on the pre-fix code (no rowid ordering in get_last_messages): with only 5 tied-timestamp rows, SQLite happens to return insertion order anyway, so it doesn't guard the read-path fix.

Assert on the query text referencing rowid, or force a batch size that would actually diverge without it.

mem0-ts/src/oss/tests/storage.unit.test.ts
```diff
@@ -104,6 +104,34 @@ describe("SQLiteManager", () => {
...
+      role: "user",
+      content: `msg${i}`,
+    }));
+    await db.saveMessages(messages, "scope-b");
+
+    const order = (await db.getLastMessages("scope-b", 10)).map(
+      (m) => m.content,
+    );
+
+    expect(order).toEqual(Array.from({ length: 5 }, (_, i) => `msg${i}`));
```
kartik-mem0: Same gap as the Python test: this passes on the pre-fix SQLiteManager too, since 5 tied-timestamp rows happen to come back in insertion order without the rowid fix. It doesn't catch a regression here.

Strengthen it the same way as the Python counterpart.

yashs33244:
Thanks for the precise catch @kartik-mem0 - spot on. Those read-order tests requested `limit >= the saved count`, so no selection happened and they passed pre-fix. Strengthened both to save 8 and request the newest 5, so the tied created_at keeps the oldest 5 without the rowid tiebreak. Now genuinely red-green in both SDKs (Python pytest + TS jest): fails on `created_at` (returns msg0..msg4), passes on `rowid` (msg3..msg7). Pushed.

On the optional notes: happy to restore the derived-table comment at storage.py:282 in this PR and open a follow-up issue for the get_history tie risk - just say the word if you'd like the comment folded in.

tests/memory/test_storage.py
kartik-mem0: Verified against a73b480: fixed. The test now saves 8 rows and asserts limit=5 (tests/memory/test_storage.py:262-274), so the inner subquery's ORDER BY has to actually pick the 5 survivors instead of coincidentally matching either order. Red-green confirmed: reverting storage.py to pre-fix fails this test (returns msg0-msg4), the PR's fix passes it (msg3-msg7).

mem0-ts/src/oss/tests/storage.unit.test.ts
kartik-mem0: Verified against a73b480: fixed. Same shape, 8 rows / limit=5 (mem0-ts/src/oss/tests/storage.unit.test.ts:122-135). Red-green confirmed: reverting SQLiteManager.ts to pre-fix fails this test, the PR's fix passes it.

kartik-mem0 (commented):
## Changes requested

Re-review: the prior test-quality gap is fixed. The only thing still blocking merge is a red required CI check unrelated to this diff.

**To unblock:**
1. Rebase onto `main` to refresh `changelog_check` (detail below, not a code issue)

- **Problem is real:** reproduces on current `main` at `mem0/memory/storage.py:save_messages`, confirmed live via sandbox (`sandbox/scenarios/issue6429_sqlite_retention.py`, `sandbox/ts/scenarios/issue6429_sqlite_retention.ts`), both languages, both FAIL on `main` and PASS at `a73b480`
- **Tests / red-green:** `pytest tests/memory/test_storage.py -v` -> 16 passed; `jest storage.unit.test.ts` -> 24 passed. The previously-flagged insertion-order test (this thread) is fixed: reverting `storage.py`/`SQLiteManager.ts` to pre-PR `main` now correctly fails both rewritten tests; restoring the PR's fix passes both. Confirmed independently by all 3 lenses.
- **CI:** `changelog_check` is RED on both Python SDK and TypeScript SDK workflows, which fails `CI Gate` (`mergeStateStatus: BLOCKED`). All real build/test jobs pass (`build_mem0` 3.10/3.11/3.12, `build_ts_sdk` 20/22, `integration_ts_sdk` 20/22). Traced the cause: the check diffs `pyproject.toml`/`package.json` at the PR's recorded base SHA (`96b31c4`, a June snapshot, `pyproject.toml` version `2.0.7`) against the merge-simulated checkout, which picks up current `main`'s version (`2.0.15`) since this PR never touches either file (confirmed: the worktree's own `pyproject.toml` is still `2.0.7`, identical to base). It reads as a version bump this PR never made. A rebase/merge of current `main` should refresh the base SHA and clear it. Vercel is also red, ignored per usual (deploy-preview auth, non-blocking).
- **Security:** cleared, no new attack surface. Ordering column is a fixed literal (`rowid`), never user-derived; `WHERE session_scope = ?` tenancy scoping is untouched in both queries; no new string-built SQL.
- **Bloat (ponytail):** lean already, `net: -0 lines possible` (Lens 3).
- **Not verified:** concurrent/multi-process access to the same SQLite file under this ordering (inherited risk, transaction/lock discipline unchanged by this diff).

</details>

Real work here: the rowid tiebreak is the right root-cause fix, applied identically in both SDKs, and the regression tests now actually prove it. Once `changelog_check` clears (rebase should do it) this is ready to merge.

yashs33244:
Thanks for the thorough re-review @kartik-mem0, and for tracing `changelog_check` to the stale base SHA - that lines up with what I saw. Merged current `main` into the branch to refresh it. The net diff still touches only `storage.py` / `SQLiteManager.ts` and their tests (no `pyproject.toml` / `package.json`), storage suites are green in both SDKs (py 17, ts 24) and ruff + prettier are clean. That should clear the gate.

kartik-mem0 (commented):

Verified against `e24a0b20` (new head): the only ask from the earlier changes-requested review was resolved. This head is a merge of current `main`, which refreshed the stale recorded base SHA and cleared `changelog_check` on both SDK workflows. The PR's own diff (`gh pr diff`) is byte-identical to what was already fully cleared in the last round, confirmed independently by three fresh lenses reading the current code, not by trusting this summary.

**Follow-up:** none
**Not verified:** concurrent/multi-process access to the same SQLite file under the rowid ordering (inherited risk, unchanged by this diff)

Risk: low, isolated ordering fix in one shared storage class, no schema or public API change · Your time: ~5 min

- **Problem is real:** reproduces on current `main` at `mem0/memory/storage.py:save_messages`, confirmed live via sandbox, both Python and TypeScript, both FAIL on `main` and PASS on this head
- **Tests:** `pytest tests/memory/test_storage.py -v` -> 17 passed. `jest storage.unit.test.ts` -> 24 passed
- **Red-green:** reverting the fix on both languages makes the two rewritten regression tests fail with the exact wrong-order symptom from the bug report; restoring the fix passes both
- **CI:** all required checks green (`changelog_check` x2, `build_mem0` 3.10/3.11/3.12, `build_ts_sdk` 20/22, `integration_ts_sdk` 20/22, `check_changes` x2, `CI Gate`, `license/cla`). Vercel is red, ignored, deploy-preview auth only
- **Security:** cleared, no new attack surface. Ordering column is a fixed literal, never user-derived; tenancy scoping (`WHERE session_scope = ?`) untouched
- **Bloat (ponytail):** lean already, `net: -0 lines possible`
- **Linked issue:** #5661

</details>

yashs33244:
Gentle nudge on this one 🙏 @kartik-mem0 you'd marked the batched-save eviction fix ready for squash-merge and CI is green - whenever you or @deshraj have a moment to merge, that'd be great. Thanks!

PR #5634, opened 2026-06-18
fix(server): return 404/400 instead of 502 for not-found and invalid input

### LINKED ISSUE #5663: REST add/update/delete return 502 for client errors (not-found, invalid input) instead of 404/400
### Bug

In the REST server, the `add`, `update`, and `delete` handlers wrap the core call in a bare `except Exception` that returns 502. But core `update`/`delete` raise `ValueError("Memory with id ... not found")` for a missing memory, and `add` raises a `ValidationError` for invalid input (e.g. a bad `memory_type`). These client errors all surface as 502.

### Impact

A caller cannot distinguish a bad request from a backend outage: missing memories and invalid input both return 502 instead of 404/400. This is inconsistent with `/search`, which already maps core `ValueError` to 400.

### Expected

Not-found errors should return 404, other validation errors 400, and only genuine backend failures 502 (mirroring `/search`).

yashs33244:
## Linked Issue

Closes #5663

## Description

The REST `add`, `update`, and `delete` handlers wrap the core call in a bare `except Exception` that returns 502. But core `update`/`delete` raise `ValueError("Memory with id ... not found")` for a missing memory, and `add` raises a `ValidationError` for invalid input (e.g. a bad `memory_type`). Those are client errors, not outages, yet they all surfaced as 502, so a caller couldn't tell a bad request from a backend being down.

This maps those cases to the right status, mirroring what `/search` already does:

- `ValueError` whose message contains "not found" -> 404
- other `ValueError` / `Mem0ValidationError` -> 400
- everything else still -> 502 via `upstream_error()`

Scoped to the three write handlers named above. `/search` and `delete_all` are left as-is to keep the change minimal.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `TestWriteHandlerErrorMapping` in `tests/test_server_params.py`: not-found on update/delete returns 404, other validation errors return 400, and a real outage still returns 502. They fail on the current code (502) and pass with the fix.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1831,8 +1831,9 @@ class Memory(MemoryBase):
         try:
             existing_memory = self.vector_store.get(vector_id=memory_id)
         except Exception:
+            # Backing-store failure, not a bad memory_id: re-raise the original so the REST layer maps it to 5xx, not 4xx.
             logger.error(f"Error getting memory with ID {memory_id} during update.")
-            raise ValueError(f"Error getting memory with ID {memory_id}. Please provide a valid 'memory_id'")
+            raise
 
         if existing_memory is None:
             raise ValueError(f"Memory with id {memory_id} not found. Please provide a valid 'memory_id'")
@@ -3363,8 +3364,9 @@ class AsyncMemory(MemoryBase):
         try:
             existing_memory = await asyncio.to_thread(self.vector_store.get, vector_id=memory_id)
         except Exception:
+            # Backing-store failure, not a bad memory_id: re-raise the original so the REST layer maps it to 5xx, not 4xx.
             logger.error(f"Error getting memory with ID {memory_id} during update.")
-            raise ValueError(f"Error getting memory with ID {memory_id}. Please provide a valid 'memory_id'")
+            raise
 
         if existing_memory is None:
             raise ValueError(f"Memory with id {memory_id} not found. Please provide a valid 'memory_id'")
--- a/server/main.py
+++ b/server/main.py
@@ -19,6 +19,7 @@ from errors import (
 from fastapi import Depends, FastAPI, HTTPException, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, RedirectResponse
+from mem0.exceptions import ValidationError as Mem0ValidationError
 from models import RequestLog, User
 from pydantic import BaseModel, Field
 from rate_limit import limiter
@@ -206,6 +207,14 @@ class GenerateInstructionsRequest(BaseModel):
     use_case: str = Field(..., description="Description of what the user will use Mem0 for.")
 
 
+def _client_error(exc: Exception) -> HTTPException:
+    """Map core validation / not-found errors to 4xx so clients can tell a bad
+    request from an upstream outage. 'not found' is a 404, everything else a 400."""
+    detail = str(exc)
+    status_code = 404 if isinstance(exc, ValueError) and "not found" in detail.lower() else 400
+    return HTTPException(status_code=status_code, detail=detail)
+
+
 def _redact_config(value: Any, key: str | None = None) -> Any:
     if isinstance(value, dict):
         return {item_key: _redact_config(item_value, item_key) for item_key, item_value in value.items()}
@@ -363,6 +372,8 @@ def add_memory(memory_create: MemoryCreate, _auth=Depends(verify_auth)):
         if response.get("results"):
             telemetry.log_dashboard_nudge_once(DASHBOARD_URL)
         return JSONResponse(content=response)
+    except (ValueError, Mem0ValidationError) as e:
+        raise _client_error(e)
     except Exception:
         raise upstream_error()
 
@@ -466,6 +477,8 @@ def update_memory(memory_id: str, updated_memory: MemoryUpdate, _auth=Depends(ve
         return get_memory_instance().update(
             memory_id=memory_id, data=updated_memory.text, metadata=updated_memory.metadata
         )
+    except (ValueError, Mem0ValidationError) as e:
+        raise _client_error(e)
     except Exception:
         raise upstream_error()
 
@@ -485,6 +498,8 @@ def delete_memory(memory_id: str, _auth=Depends(verify_auth)):
     try:
         get_memory_instance().delete(memory_id=memory_id)
         return MessageResponse(message="Memory deleted successfully")
+    except (ValueError, Mem0ValidationError) as e:
+        raise _client_error(e)
     except Exception:
         raise upstream_error()
 
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -489,6 +489,59 @@ async def test_async_update_nonexistent_memory_raises_error(mock_sqlite, mock_ll
 @patch('mem0.utils.factory.VectorStoreFactory.create')
 @patch('mem0.utils.factory.LlmFactory.create')
 @patch('mem0.memory.storage.SQLiteManager')
+def test_update_propagates_vector_store_failure(mock_sqlite, mock_llm_factory, mock_vector_factory, mock_embedder_factory):
+    """A backing-store failure while fetching the memory during update must
+    surface as the original error, not be masked as a 'provide a valid
+    memory_id' ValueError. The REST layer relies on this so an outage maps to
+    5xx instead of a misleading 4xx."""
+    mock_embedder_factory.return_value = MagicMock()
+    mock_vector_store = MagicMock()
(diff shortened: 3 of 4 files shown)


kartik-mem0:
## Changes Requested

### Strengths

- **`_client_error()` helper** (`server/main.py:209–215`): Centralising the error→status mapping in one place rather than duplicating inline checks across three handlers is the right pattern. Clean and future-maintainer-friendly.
- **Exception ordering** (lines 372, 479, 498 in patched file): `(ValueError, Mem0ValidationError)` before bare `Exception` means the 502 fallback is preserved for genuine outages. No risk of over-catching.
- **Regression tests** (`tests/test_server_params.py:703–744`): Five-test matrix is real regression coverage — 4/5 fail on a plain revert of `server/main.py`, 1/5 correctly stays green (the 502-passthrough for `RuntimeError`). Red-green proof is clean.
- **Scope restraint**: Correctly limited to the three write handlers named in the issue; `/search` is already correct and is left as-is.

---

### Issues

#### Important (should fix)

**`server/main.py:211` — vector-store errors during `update()` are incorrectly mapped to 400**

`mem0/memory/main.py:1846` (in `_update_memory`) raises:
```python
raise ValueError(f"Error getting memory with ID {memory_id}. Please provide a valid 'memory_id'")
```
when `self.vector_store.get()` **itself throws an exception** — i.e. when the backing DB/network is degraded. This `ValueError` does not contain `"not found"`, so `_client_error()` maps it to HTTP **400**. But a 400 tells the caller "you sent a bad request." The actual cause is a server-side infrastructure failure — the correct response is 502 (or 503), exactly what `upstream_error()` was already returning.

The fix is a narrow guard: only catch `ValueError` whose origin is a *client* problem, not a backend failure. Two options:

**Option A — narrow the catch to known-safe phrases** (minimal change):
```python
_NOT_FOUND_PHRASES = frozenset(["not found"])

def _client_error(exc: Exception) -> HTTPException:
    detail = str(exc)
    if isinstance(exc, ValueError) and any(p in detail.lower() for p in _NOT_FOUND_PHRASES):
        return HTTPException(status_code=404, detail=detail)
    if isinstance(exc, Mem0ValidationError):
        return HTTPException(status_code=400, detail=detail)
    # Other ValueError from _update_memory wrapping a vector-store exception:
    # re-raise as 502, not 400
    raise upstream_error()
```
Then tighten the `except` clause in `update_memory` to only catch `Mem0ValidationError` and the specific "not found" `ValueError` (not all `ValueError`).

**Option B — use `MemoryNotFoundError`** (cleaner, more robust):
`mem0/exceptions.py:162` already defines `MemoryNotFoundError`. Replace the six `raise ValueError(f"Memory with id {memory_id} not found...")` sites in `mem0/memory/main.py` with `raise MemoryNotFoundError(...)`, then match on `isinstance(exc, MemoryNotFoundError)` in `_client_error` — no string matching, no edge cases. The infrastructure-error `ValueError` at line 1846 is left uncaught by `_client_error` and falls through to `upstream_error()` correctly.

Option A is a smaller diff; Option B removes the string-match fragility permanently. Either is acceptable.

---

#### Minor (optional)

- **`server/main.py:211` — docstring** says `'not found' is a 404, everything else a 400` — add a note that this applies to `ValueError` only; a `Mem0ValidationError` containing "not found" would silently return 400.
- **`tests/test_server_params.py`** — no test covering the `update()` vector-store-exception path (the `_update_memory` line 1846 case). Once the Important is fixed, add: `mock_memory.update.side_effect = ValueError("Error getting memory with ID mem-1. Please provide a valid 'memory_id'")` → assert `resp.status_code == 502`.

---

### Evidence

- **CI:** CI Gate + Python SDK builds (3.10 / 3.11 / 3.12) + changelog + CLA all pass. Vercel deploy preview fails (advisory, no deploy perms from forks — ignored per repo convention).
- **Targeted tests:** `pytest tests/test_server_params.py::TestWriteHandlerErrorMapping -v` → **5/5 passed** (0.68 s)
- **Red-green proof:** reverted `server/main.py` to `origin/main` → 4/5 fail (`assert 502 == 404` / `assert 502 == 400`); restored → 5/5 pass ✓
- **`server/main.py` ruff:** all checks passed (line-length 120)
- **Code traced:** `mem0/memory/main.py:1839–1849` (`_update_memory`) — line 1846 raises `ValueError` wrapping a vector-store exception, without "not found"; this is the path `_client_error` misclassifies as 400.
- **Could NOT verify:** live vector-store error injection to confirm the 400 response in production; code-tracing only.

### Assessment

Ready to merge: **With fixes** — please address the Important item; I'll re-review once you push. The core intent is right and the test suite is solid; this is a one-function adjustment to keep infrastructure errors in the 5xx class where they belong. Thanks for the clean PR structure.

yashs33244:
Thanks for the careful trace, @kartik-mem0 - you're right that the infra path was the problem. I went a slightly different route than Option A after testing it.

The catch with Option A: mapping non-"not found" `ValueError` to 502 breaks this PR's own `test_update_other_value_error_returns_400` - a genuine validation `ValueError("data must be a non-empty string")` from update is also non-"not found", so it would flip to 502. At the server layer the validation `ValueError` and the infra-wrapped one are indistinguishable (both plain, non-"not found"), so any server-side rule either regresses validation to 502 or needs fragile matching on the specific infra string.

So I fixed the root cause instead: `_update_memory` (sync + async) no longer masks a `vector_store.get()` failure as `ValueError("... provide a valid 'memory_id'")` - it re-raises the original. Now:

- backing-store outage during update -> original error propagates -> server `except Exception` -> **502**
- genuine not-found (`existing_memory is None`) -> still `ValueError("... not found")` -> **404**
- validation `ValueError` -> still **400** (no regression; the pinned tests stay green)

`_client_error` is unchanged. Added sync + async regression tests (`test_update_propagates_vector_store_failure`), red-green proven: with the old masking they fail (ValueError instead of the original), with the fix they pass. ruff clean.

One note: the `test_server_params.py` suite needs the server runtime on path to import (`telemetry`/`UpstreamError`), so it runs green in CI but not my local sandbox - those tests mock the memory instance and this change doesn't touch `server/`, so they're unaffected. Happy to add a server-level integration test too if you'd prefer.

kartik-mem0:
## LGTM

The fix addresses the Important finding cleanly. Verified this run.

**`mem0/memory/main.py` lines 1833-1836 and 3366-3369 (sync + async `_update_memory`):** The old code caught any exception from `vector_store.get()` and re-raised it as a `ValueError("...provide a valid 'memory_id'")`, masking an infra failure as a client error. The fix is a bare `raise` — the original exception propagates unchanged, falls through the `except (ValueError, Mem0ValidationError)` guard in `server/main.py` (it's a `ConnectionError` / whatever the backing store throws, not a `ValueError`), and lands in `except Exception: raise upstream_error()` -> **502**. Correct layer, minimal diff.

**`mem0/memory/main.py` lines 1838-1839 and 3371-3372:** The genuine not-found case (`existing_memory is None`) still raises `ValueError("... not found")` -> `_client_error()` -> **404**. Unchanged and correct.

**`tests/test_memory.py` lines 490-536:** Two new regression tests (`test_update_propagates_vector_store_failure` / `test_async_update_propagates_vector_store_failure`). Sync and async parity maintained.

### Evidence

- **CI:** CI Gate + Python SDK (3.10 / 3.11 / 3.12) all pass. Vercel deploy preview fails (advisory, fork perms — ignored per convention).
- **Red-green (worktree at `2f1156b60d71`):** Run both new tests -> **2/2 pass**. Reverted `mem0/memory/main.py` to `origin/main` -> **2/2 fail** (`ValueError: Error getting memory with ID mem-1...` instead of `ConnectionError`). Restored -> **2/2 pass**.
- **Regression:** `pytest tests/test_memory.py -k update` -> **10/10 pass** (all prior update tests unaffected).
- **Lint:** `ruff check mem0/memory/main.py server/main.py tests/test_memory.py tests/test_server_params.py` -> all checks passed.
- **Sync/async parity:** Both `Memory._update_memory` (line 1833) and `AsyncMemory._update_memory` (line 3366) carry identical fix.
- **Server handler coverage:** `test_add_real_outage_still_returns_502` (RuntimeError -> 502) remains green; `test_update_other_value_error_returns_400` (genuine validation ValueError -> 400) unaffected — the server-layer mock injects directly on `memory.update`, bypassing the vector-store path, which is the right test structure for that case.
- **Could NOT verify:** live vector-store failure injection against a running server (code-tracing only for that path).

### Assessment

Ready to merge: **Yes**. The Important finding is resolved at the root — no string matching, no fragile guard, just the correct re-raise. Thanks for reasoning through the option tradeoff and choosing the cleaner path.

kartik-mem0 (approved):

PR #5638, opened 2026-06-18
fix(llms): preserve reasoning fields in base-to-provider config conversion

### LINKED ISSUE #5647: LlmFactory.create drops reasoning_effort / is_reasoning_model from BaseLlmConfig
### Bug

`LlmFactory.create` (`mem0/llms/configs.py` / `mem0/utils/factory.py`) builds a `config_dict` from `BaseLlmConfig` fields and passes it to the provider-specific config constructor, but it omits `reasoning_effort` and `is_reasoning_model`. Both fields exist on `BaseLlmConfig`, so anything set on the base config is silently dropped during conversion — the provider never sees them.

### Expected

`reasoning_effort` and `is_reasoning_model` set on the base LLM config should be forwarded to providers that accept them (e.g. `OpenAIConfig`, `AzureOpenAIConfig`), and skipped for providers with strict signatures.

### Fix

Forward both fields through the factory, guarded by `inspect.signature(...).parameters` so providers that don't accept them (or that capture via `**kwargs`) are handled correctly. PR #5638 does this.

yashs33244:
## Linked Issue

Closes #5647

## Description

When `LlmFactory.create` is given a `BaseLlmConfig` and the target provider uses a provider-specific config (e.g. OpenAI), it rebuilds the config from a `config_dict` that omits `reasoning_effort` and `is_reasoning_model`. So building an OpenAI LLM from `BaseLlmConfig(model="o3", reasoning_effort="high", is_reasoning_model=True)` silently produced a config with both set to `None`, and the reasoning-model behavior was lost.

Root cause: those two fields were simply missing from the `config_dict` in the base-to-provider branch.

The fix forwards both fields, but only to provider configs that actually accept them - either explicitly (OpenAI, Azure) or via `**kwargs` (AWS Bedrock). Configs like Anthropic and Gemini don't declare them, so forwarding unconditionally would raise a `TypeError`; the signature guard avoids that.

Added a regression unit test covering an explicit-field provider (OpenAI), a `**kwargs` provider (Bedrock), and a provider that accepts neither (Anthropic, must still build).

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

New `tests/utils/test_factory.py`. The reasoning-preservation tests fail on `main` and pass with the fix; `tests/utils/` and `tests/llms/` are green (209 passed).

## Checklist

### DIFF
--- a/mem0/utils/factory.py
+++ b/mem0/utils/factory.py
@@ -1,4 +1,5 @@
 import importlib
+import inspect
 from typing import Dict, Optional, Union
 
 from mem0.configs.embeddings.base import BaseEmbedderConfig
@@ -102,6 +103,14 @@ class LlmFactory:
                     "vision_details": config.vision_details,
                     "http_client_proxies": config.http_client,
                 }
+                # Only forward reasoning fields to provider configs that accept them
+                # (explicitly or via **kwargs); others would raise on unexpected kwargs.
+                params = inspect.signature(config_class).parameters
+                accepts_kwargs = any(p.kind == p.VAR_KEYWORD for p in params.values())
+                if accepts_kwargs or "reasoning_effort" in params:
+                    config_dict["reasoning_effort"] = config.reasoning_effort
+                if accepts_kwargs or "is_reasoning_model" in params:
+                    config_dict["is_reasoning_model"] = config.is_reasoning_model
                 config_dict.update(kwargs)
                 config = config_class(**config_dict)
             else:
--- a/tests/utils/test_factory.py
+++ b/tests/utils/test_factory.py
@@ -1,0 +1,53 @@
+from unittest.mock import Mock, patch
+
+from mem0.configs.llms.anthropic import AnthropicConfig
+from mem0.configs.llms.aws_bedrock import AWSBedrockConfig
+from mem0.configs.llms.base import BaseLlmConfig
+from mem0.configs.llms.openai import OpenAIConfig
+from mem0.utils.factory import LlmFactory
+
+
+def _capture_config(provider_name, config):
+    """Build an LLM via the factory and return the config it was constructed with."""
+    captured = {}
+
+    def fake_llm_class(built_config):
+        captured["config"] = built_config
+        return Mock()
+
+    with patch("mem0.utils.factory.load_class", return_value=fake_llm_class):
+        LlmFactory.create(provider_name, config)
+
+    return captured["config"]
+
+
+def test_base_to_openai_preserves_reasoning_fields():
+    base_config = BaseLlmConfig(model="o3", reasoning_effort="high", is_reasoning_model=True)
+
+    built = _capture_config("openai", base_config)
+
+    assert isinstance(built, OpenAIConfig)
+    assert built.reasoning_effort == "high"
+    assert built.is_reasoning_model is True
+
+
+def test_base_to_kwargs_provider_preserves_reasoning_fields():
+    # AWSBedrockConfig accepts the reasoning fields via **kwargs, so they must survive.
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Good fix — this is the right place to catch it and the implementation is clean.

**What I traced:** `LlmFactory.create` builds a `config_dict` from `BaseLlmConfig` fields and hands it to the provider-specific config constructor (`OpenAIConfig`, `AzureOpenAIConfig`, etc.). `reasoning_effort` and `is_reasoning_model` were in `BaseLlmConfig` but missing from that dict, so anything set on the base config was silently dropped during conversion. Confirmed on main: `grep` finds neither field in factory.py before this patch.

**The guard logic is correct.** `inspect.signature(config_class).parameters` on a plain Python class reflects only the direct `__init__` params, not inherited ones — which is exactly what you want here. `OpenAIConfig` and `AzureOpenAIConfig` explicitly declare both fields, `AWSBedrockConfig` captures them via `**kwargs → super().__init__(**kwargs)`, and `AnthropicConfig`/`GeminiConfig`/`XAIConfig` have strict signatures that would raise `TypeError` if you forwarded unexpected kwargs. The guard handles all three cases.

**Red-green proof (Gate B, hatch dev_py_3_12):**
- Tests with fix: 3/3 pass
- Tests after reverting factory.py to main: 2/3 fail — `test_base_to_openai_preserves_reasoning_fields` → `assert None == 'high'`, `test_base_to_kwargs_provider_preserves_reasoning_fields` → `assert None == 'medium'`
- Restore fix: 3/3 pass

Both fields are asserted in the OpenAI test (`reasoning_effort == "high"` and `is_reasoning_model is True`). The Bedrock test covers the `accepts_kwargs` branch. The Anthropic test covers the negative case (no-forward). Test coverage is complete for the three distinct code paths.

**CI:** Python SDK 3.10/3.11/3.12 all pass. CI Gate passes.

One observation worth flagging separately (pre-existing, not introduced here): at `factory.py:103`, `config.http_client` is an already-constructed `httpx.Client` instance, but it's placed in `config_dict["http_client_proxies"]`. Provider configs pass that value back into `httpx.Client(proxies=...)`, which would raise `TypeError` if anyone actually sets a proxy on a `BaseLlmConfig` and goes through this conversion path. Would be worth a quick follow-up fix — happy to open a tracking issue for it if useful.

One structural note: the `config_dict` block is a manual mirror of `BaseLlmConfig`'s fields, so future fields added to `BaseLlmConfig` will need a corresponding guarded block here. Given that `AnthropicConfig` and others have strict `__init__` signatures, the guard is necessary — but worth keeping in mind when the base config evolves.

This is ready to squash-merge whenever.

For tracking on our side, would you mind opening a quick issue describing this bug and linking it here? Makes it easier to reference in the changelog and for anyone who hits the same thing.

yashs33244:
Thanks! Opened #5647 describing the bug so it's easy to reference in the changelog and for anyone who hits it.

On the pre-existing `http_client_proxies` mismatch at `factory.py:103` — good catch, I kept it out of this PR to stay focused, but happy to take it as a separate follow-up if that's useful.

yashs33244:
Opened #5648 to track the `http_client_proxies` mismatch at `factory.py:104` separately, so it doesn't get lost. Kept it out of this PR to stay focused.

yashs33244:
@kartik-mem0 this one should be good to go whenever you have a moment. I addressed your trace notes, and filed #5647 (this bug) plus #5648 (the separate `http_client_proxies` mismatch you spotted) so nothing gets lost. Thanks for the careful review!

kartik-mem0 (approved):

PR #5639, opened 2026-06-18
fix(client): check HTTP status before parsing ping response in _validate_api_key

### LINKED ISSUE #5660: MemoryClient._validate_api_key raises JSONDecodeError instead of ValueError on non-JSON 5xx ping
### Bug

`MemoryClient._validate_api_key` (and the same method on `AsyncMemoryClient`) parses the `/v1/ping/` response body as JSON *before* checking the HTTP status (`response.json()` is called before `response.raise_for_status()`). When the ping returns a non-JSON 5xx (e.g. an HTML error page from a CDN or proxy), `response.json()` raises a `JSONDecodeError` that isn't caught by the `HTTPStatusError`/`HTTPError` handler, and `_validate_api_key` isn't wrapped by `api_error_handler`.

### Impact

Constructing a client against a non-JSON 5xx ping bubbles up a confusing `JSONDecodeError` during `__init__` instead of the intended clear `ValueError`.

### Expected

The HTTP status should be checked first so a failing ping surfaces as the clear `ValueError("Error: ...")`; the 2xx happy path stays unchanged.

yashs33244:
## Linked Issue

Closes #5660

## Description

`MemoryClient._validate_api_key` (and the same method on `AsyncMemoryClient`) runs during `__init__` and parsed the `/v1/ping/` response body as JSON *before* checking the HTTP status:

```python
data = response.json()
response.raise_for_status()
```

When the ping returns a non-JSON 5xx (for example an HTML error page from a CDN or proxy), `response.json()` raises a `JSONDecodeError` first. That error is not caught by the `except httpx.HTTPStatusError` / `except requests.exceptions.HTTPError` handler, and `_validate_api_key` is not wrapped by `api_error_handler`, so it bubbles up as a confusing `JSONDecodeError` while constructing the client instead of the intended `ValueError`.

The fix calls `response.raise_for_status()` before `response.json()` in both the sync and async paths, so the HTTP status surfaces as the clear `ValueError("Error: ...")` the handler already produces. The happy path is unchanged: on a 2xx, `raise_for_status()` is a no-op and the body is parsed as before.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added a regression test for both the sync and async clients in `tests/test_client.py` that mocks a non-JSON 5xx ping and asserts a clear `ValueError` is raised (not a `JSONDecodeError`). It fails on the old order and passes with the fix.

## Checklist

### DIFF
--- a/mem0/client/main.py
+++ b/mem0/client/main.py
@@ -151,9 +151,9 @@ class MemoryClient:
         try:
             params = self._prepare_params()
             response = self.client.get("/v1/ping/", params=params)
-            data = response.json()
-
             response.raise_for_status()
+
+            data = response.json()
 
             if data.get("org_id") and data.get("project_id"):
                 self.org_id = data.get("org_id")
@@ -1044,9 +1044,9 @@ class AsyncMemoryClient:
                 },
                 params=params,
             )
-            data = response.json()
-
             response.raise_for_status()
+
+            data = response.json()
 
             if data.get("org_id") and data.get("project_id"):
                 self.org_id = data.get("org_id")
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,8 +1,11 @@
 """Tests for MemoryClient entity parameter rejection."""
 
+import json
 from unittest.mock import MagicMock, patch
 
+import httpx
 import pytest
+import requests
 
 
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Swaps `raise_for_status()` before `response.json()` in both `MemoryClient._validate_api_key` and `AsyncMemoryClient._validate_api_key`, so a non-JSON 5xx (e.g. an HTML error page from a CDN or proxy) surfaces as the intended `ValueError("Error: …")` rather than an uncaught `JSONDecodeError` during client construction.

### Strengths
- `mem0/client/main.py` (~L154, ~L1047): surgical two-line reorder, no logic added or removed. The existing `except HTTPStatusError`/`except HTTPError` handlers already wrap errors into `ValueError`; the fix simply ensures they fire before `json()` can raise.
- Both sync (`MemoryClient`/httpx) and async (`AsyncMemoryClient`/requests) paths are fixed symmetrically.
- `tests/test_client.py` (L212–265): regression tests for both paths mock a non-JSON 503 and assert a `ValueError` with `"Error:"` prefix is raised. Red-green confirmed: both tests fail on `main` (bug reproduces), pass with the fix.
- Conventional commit title, ruff-clean, no new dependencies.

### Issues
Minor (optional):
1. `tests/test_client.py:231, 255` — imports are inside the `with patch(…)` context blocks (intentional, needed for module-level singleton patching). A one-line comment (`# import inside patch scope so httpx.Client is replaced before __init__ runs`) would help future readers avoid accidentally hoisting the import.
2. `tests/test_client.py:253, 264` — `assert "Error:" in str(exc_info.value)` is correct but would survive an error message format change silently. Optionally also check the HTTP status string appears (cosmetic only — not blocking).

### Evidence
- **CI:** CI Gate PASS; Python SDK build+test PASS on 3.10/3.11/3.12. Vercel check FAIL (advisory/fork perms — ignored per repo policy).
- **Targeted tests:** `pytest tests/test_client.py::TestValidateApiKeyHttpError -v` → **2 passed in 0.45s** (worktree `review-pr-5639`, head 6bf16c1f)
- **Red-green proof:** reverted `mem0/client/main.py` to `main` → both tests **FAIL** (JSONDecodeError bubbles up unhandled, bug confirmed); restored fix → both **PASS** ✓
- **Lint:** `ruff check mem0/client/main.py tests/test_client.py` → All checks passed
- **Linked issue:** #5660 — PR directly addresses the root cause described.
- **Cross-package check:** TS client (`mem0-ts/src/client/mem0.ts::ping()`) uses `_fetchWithErrorHandling` which checks `response.ok` before JSON parsing — already safe, no TS-side action needed.
- **Could NOT verify:** live CDN/proxy behavior returning HTML 5xx — code tracing + mocks confirm the path.

### Assessment

The two-line reorder is the correct, minimal fix at the right layer. Tests are red-green proven this run. No concerns from correctness, repo-fit, or architecture lenses after skeptic review.

kartik-mem0 (approved):

PR #5640, opened 2026-06-18
fix(server): return 404 not 500 for malformed api key id on revoke

yashs33244:
## Linked Issue

Self-found; no existing issue. The bug is small and self-contained, so I'm explaining it here per the template rather than opening a separate issue.

## Description

`DELETE /api-keys/{key_id}` returned a 500 when `key_id` wasn't a valid UUID. `revoke_key` passed the raw path string straight to `db.get(APIKey, key_id)`, and since `api_keys.id` is a Postgres `uuid` column, psycopg raises a `DataError` while binding the value -- before the existing `None -> 404` check ever runs. So a typo'd or garbage key id surfaced as a server error instead of a clean 404.

The fix parses `key_id` as a UUID first and raises 404 on failure, mirroring the guard already used in `auth.consume_refresh_jti`. A valid-but-missing id still returns 404 as before.

Added a regression unit test.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

`tests/test_api_keys_router.py` mounts the api-keys router and overrides the DB session to reproduce the Postgres uuid rejection. The malformed-id test fails (500) without the fix and passes (404) with it; a valid-but-missing id stays 404.

## Checklist

### DIFF
--- a/server/routers/api_keys.py
+++ b/server/routers/api_keys.py
@@ -1,3 +1,4 @@
+import uuid
 from datetime import datetime, timezone
 
 from fastapi import APIRouter, Depends, HTTPException
@@ -77,7 +78,11 @@ def create_key(body: CreateKeyRequest, user: User = Depends(require_auth), db: S
 
 @router.delete("/{key_id}", response_model=MessageResponse)
 def revoke_key(key_id: str, user: User = Depends(require_auth), db: Session = Depends(get_db)):
-    api_key = db.get(APIKey, key_id)
+    try:
+        key_uuid = uuid.UUID(key_id)
+    except (TypeError, ValueError):
+        raise HTTPException(status_code=404, detail="API key not found.")
+    api_key = db.get(APIKey, key_uuid)
     if api_key is None or api_key.created_by != user.id:
         raise HTTPException(status_code=404, detail="API key not found.")
     if api_key.revoked_at is not None:
--- a/tests/test_api_keys_router.py
+++ b/tests/test_api_keys_router.py
@@ -1,0 +1,69 @@
+"""Tests for the api-keys router.
+
+revoke_key takes the key_id straight from the URL path and passes it to
+db.get(APIKey, key_id). The id column is a Postgres uuid, so a malformed
+(non-UUID) key_id makes psycopg raise a DataError at bind time -- which
+surfaces as a 500 -- before the None -> 404 check is ever reached. A missing
+key should always be a 404, whether or not the id is a valid UUID.
+"""
+
+import os
+import sys
+import uuid
+
+import pytest
+
+pytest.importorskip("fastapi", reason="fastapi not installed")
+
+from fastapi import FastAPI  # noqa: E402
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Fixes `DELETE /api-keys/{key_id}` returning 500 when given a malformed (non-UUID) key ID. The raw string was passed directly to `db.get(APIKey, key_id)` against a Postgres `uuid` column, causing psycopg to raise `DataError` at bind time — before the `None → 404` guard could run. The fix parses `key_id` as `uuid.UUID` first and raises 404 on failure, exactly mirroring the guard already in `consume_refresh_jti` (`server/auth.py:78–81`).

### Strengths
- `server/routers/api_keys.py:81–83` — the try/except pattern is the right one for this codebase: `consume_refresh_jti` in `server/auth.py` uses the identical `try: uuid.UUID(x) / except (TypeError, ValueError): raise HTTPException(...)` idiom. The PR is consistent with the established precedent, not fighting it.
- `(TypeError, ValueError)` covers both `None` input and non-UUID strings — complete, defensive.
- `tests/test_api_keys_router.py` — `_RaisingSession.get()` faithfully reproduces the real psycopg failure mode (raises `sqlalchemy.exc.DataError` for non-UUID input, returns `None` for valid-UUID misses). `TestClient(raise_server_exceptions=False)` correctly surfaces the 500 as a response object rather than re-raising it, so the assertion genuinely distinguishes the two status codes. The test is real, not theatrical.
- Passing `uuid.UUID` to `db.get` instead of the raw string is also more type-correct against `APIKey.id: Mapped[uuid.UUID]`.

### Issues
Minor (optional):

1. `server/routers/api_keys.py:1` — `import uuid` lands above `from datetime import datetime, timezone`. isort (profile = "black") expects them in the reverse order. Auto-fixable: `isort server/routers/api_keys.py`. Pre-commit hook will catch this if it fires locally.

### Evidence
- **CI:** CI Gate PASS · Python SDK (3.10 / 3.11 / 3.12) all PASS · Vercel advisory-only (no deploy perms from forks — ignored)
- **Targeted tests:** `PYTHONPATH=server python -m pytest tests/test_api_keys_router.py -v` → `2 passed in 1.07s`
- **Red-green proof:** implementation reverted to pre-PR → `test_revoke_malformed_key_id_returns_404` FAILS (`assert 500 == 404`); fix restored → 2 passed. Confirmed.
- **Sibling endpoints checked:** `GET/PUT/DELETE /memories/{memory_id}` and `/memories/{memory_id}/history` in `main.py` are all wrapped in `except Exception: raise upstream_error()` (returns 502, not 500) — different failure class, out of scope here.
- **Linked issue:** none — see note below
- **Could NOT verify:** live Postgres `DataError` path — verified via code tracing + mock that reproduces the driver behavior

### Assessment

The root cause is correctly identified and fixed at the right layer. The test is real and bites. One-liner isort fix before merge; everything else is clean.

---

kartik-mem0 (approved):

PR #5667, opened 2026-06-19
fix(ts-oss): anchor entity substring dedup to word boundaries (#5646)

### LINKED ISSUE #5646: Graph entity dedup (TS): distinct entities sharing a substring prefix are dropped ("Sam" vs "Samsung")
### Bug

The TypeScript graph entity dedup has the same unanchored-substring bug that the Python side had (fixed in #5630).

In `mem0-ts/src/oss/src/utils/entity_extraction.ts` (~line 710), the final dedup step uses:

```ts
other.includes(entity.text.toLowerCase())
```

That's an unanchored substring check, so an entity is silently dropped whenever its text appears anywhere inside another entity. Example: with both `"Sam"` and `"Samsung"` extracted, `"Sam"` is dropped even though they're distinct entities — a silent recall hit on both the write and search paths.

### Expected

Distinct entities that only share a leading substring should both survive. Whole-word subsets (e.g. `"learning"` inside `"machine learning"`) should still be dropped.

### Fix

Mirror the Python fix from #5630: anchor the containment check to word boundaries (regex with `\b...\b` on the escaped entity text) instead of a bare `includes`.

Tracking the TS counterpart here per review on #5630.

Bartok9:
## Summary

- Anchors the TypeScript graph entity-dedup substring check to word boundaries.
- Distinct entities that only share a leading substring (e.g. `"Net"` / `"Network"`, `"Sam"` / `"Samsung"`) are no longer silently dropped on the write and search paths.

## Motivation

Closes #5646.

The final dedup pass in `mem0-ts/src/oss/src/utils/entity_extraction.ts` removed entities that are substrings of longer entities using an **unanchored** check:

```ts
other.includes(entity.text.toLowerCase())
```

This drops any entity whose text appears *anywhere* inside another entity, including as a mere leading substring. With both `"Sam"` and `"Samsung"` extracted, `"Sam"` is dropped because `"samsung".includes("sam")` is `true` — a silent recall hit on both write and search. This is the TypeScript counterpart of the Python fix in #5630 (tracked here per review on that PR).

## Fix

Anchor the containment check to word boundaries with a `RegExp` (`\b<escaped>\b`), mirroring #5630. A small `escapeRegExp` helper guards quoted entities that may contain regex-special characters. Whole-word subsets (e.g. `"learning"` inside `"machine learning"`) are still removed as before.

```ts
const lower = entity.text.toLowerCase();
const wordBoundary = new RegExp(`\\b${escapeRegExp(lower)}\\b`);
return !allLower.some((other) => lower !== other && wordBoundary.test(other));
```

## Verification

- `npx jest src/oss/tests/entity-extraction.dedup.test.ts` — 2 passed
- `npx jest memory.entity-boost entity-extraction` — 6 passed (no regressions in the entity paths)
- `tsc --noEmit -p tsconfig.test.json` — no new errors in the changed file (pre-existing errors only in unrelated `src/community/.../langchain/mem0.ts`)

## Real behavior proof

- **Behavior addressed:** `"Net"` was dropped whenever `"Network"` was also extracted (unanchored `includes`).
- **Real environment:** Node `v25.5.0`, jest + ts-jest, on this branch.
- **Command run after the patch:** captured via a temporary test calling `extractEntities('She mentioned "Network". He mentioned "Net".')`.
- **Captured output AFTER fix:**
  ```
  PROOF_INPUT="She mentioned \"Network\". He mentioned \"Net\"."
  PROOF_OUTPUT=["QUOTED:Network","QUOTED:Net"]
  ```
  Before the fix, the same call returns only `["QUOTED:Network"]` — `"Net"` is dropped.
- **Regression test:** `src/oss/tests/entity-extraction.dedup.test.ts` — `keeps distinct entities that only share a leading substring (Net vs Network)` asserts the new shape and **fails without the fix** (verified by reverting the source: the positive case fails with `Received array: ["Network"]`, while the whole-word-subset case still passes).
- **What was NOT tested:** the procedural/graph store wiring above this function is unchanged and out of scope; this targets the pure `extractEntities` dedup step.

### DIFF
--- a/mem0-ts/src/oss/src/utils/entity_extraction.ts
+++ b/mem0-ts/src/oss/src/utils/entity_extraction.ts
@@ -306,6 +306,11 @@ try {
 // Internal helpers
 // ---------------------------------------------------------------------------
 
+/** Escape a string for safe interpolation into a RegExp. */
+function escapeRegExp(text: string): string {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
 /** Check for formatting artifacts that indicate non-entity text. */
 function hasArtifacts(txt: string): boolean {
   if (txt.includes("**") || txt.includes("__") || txt.includes(":*")) {
@@ -700,16 +705,18 @@ export function extractEntities(text: string): ExtractedEntity[] {
   }
   const bestEntities = Array.from(best.values());
 
-  // Remove entities that are substrings of longer entities
+  // Remove entities that are whole-word substrings of longer entities.
+  // Word-boundary anchoring avoids dropping distinct entities that only share a
+  // leading substring (e.g. "Sam" must survive alongside "Samsung"), while still
+  // dropping whole-word subsets (e.g. "learning" inside "machine learning").
   const allLower = bestEntities.map((e) => e.text.toLowerCase());
-  return bestEntities.filter(
-    (entity) =>
-      !allLower.some(
-        (other) =>
-          entity.text.toLowerCase() !== other &&
-          other.includes(entity.text.toLowerCase()),
-      ),
-  );
+  return bestEntities.filter((entity) => {
+    const lower = entity.text.toLowerCase();
+    const wordBoundary = new RegExp(`\\b${escapeRegExp(lower)}\\b`);
+    return !allLower.some(
+      (other) => lower !== other && wordBoundary.test(other),
+    );
+  });
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Bartok9:
Re-verified against current `origin/main` (`6bb1d328`): the unanchored-substring dedup bug from #5646 is now resolved upstream by #5829 (`b33fa542`, "fix(memory): align entity extraction precision"). The final dedup filter in `entity_extraction.ts` now uses a word-boundary-anchored regex (`(^|\s)<entity>(\s|$)`) plus a typePriority guard — functionally equivalent to this PR's intent.

I confirmed the behavior on main: with both `Sam` and `Samsung` extracted, both now survive; whole-word subsets (e.g. `learning` inside `machine learning`) are still correctly dropped.

This PR is therefore superseded — closing in favor of the merged maintainer fix. Happy to be corrected if I've misread the current state.

PR #5686, opened 2026-06-20
fix: avoid full entity scan on memory delete/update

### LINKED ISSUE #4988: bug: avoid full table scan on every delete/update in _remove_memory_from_entity_store

jfeng18:
Fixes #4988

The entity cleanup path (delete/update) scans all entities with top_k=10000, causing 10s+ timeouts at scale. This PR adds a targeted filter fast path that asks the DB to return only entities referencing the given memory_id.

Changes:
- Add _get_entity_rows_for_memory() with fast path (targeted filter) + fallback (broad scan)
- Fast path uses linked_memory_ids filter -- works natively on Qdrant (MatchValue on array fields)
- Fallback preserves original behavior for stores without array-contains support
- Add _extract_entity_rows() to normalize diverse list() return formats
- WARNING log when scan hits _ENTITY_SCAN_LIMIT
- test suite of 28 tests covering fast/fallback paths (sync + async)

Update in this revision:
- Synced with latest main
- Tests strengthened to 28 (tenant isolation, per-entity failure tolerance, scan limit, _bulk smoke tests)
- Fast path fallback now logs at debug level for diagnosability

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -624,6 +624,19 @@ class Memory(MemoryBase):
         except Exception as e:
             logger.warning(f"Entity upsert failed for '{entity_text}': {e}")
 
+    @staticmethod
+    def _extract_entity_rows(listed):
+        """Unwrap the heterogeneous return format of vector store list() into a flat row list."""
+        if not listed:
+            return []
+        if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list):
+            return listed[0]
+        return listed
+
+    # Upper bound for the broad entity scan fallback.  Kept as a class attr so
+    # tests (or subclasses) can override without monkey-patching constants.
+    _ENTITY_SCAN_LIMIT = 10000
+
     def _remove_memory_from_entity_store(self, memory_id, filters):
         """Strip `memory_id` from every entity record scoped to `filters`.
 
@@ -631,6 +644,12 @@ class Memory(MemoryBase):
           - remove the id; if the list becomes empty, delete the entity record.
           - otherwise re-embed the entity text and update the payload
             (the vector store's update() requires a vector).
+
+        Optimisation: first attempts a targeted filter (linked_memory_ids ==
+        memory_id) which is resolved server-side by stores supporting array-
+        element matching (e.g. Qdrant MatchValue on list fields).  Falls back
+        to a broader scan with application-layer filtering for stores that do
+        not support this.
 
         No-op if the entity store has never been initialized in this process.
         Errors on individual entities are swallowed at debug level; outer
@@ -641,9 +660,8 @@ class Memory(MemoryBase):
             return
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         try:
-            listed = self.entity_store.list(filters=search_filters, top_k=10000)
-            rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
-            for row in rows or []:
+            rows = self._get_entity_rows_for_memory(memory_id, search_filters)
+            for row in rows:
                 try:
                     payload = getattr(row, "payload", None) or {}
                     linked = payload.get("linked_memory_ids", [])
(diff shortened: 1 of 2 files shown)


jfeng18:
Friendly ping for review when you have a moment. Happy to address any feedback.

jfeng18:
Gentle follow-up -- any chance this could be reviewed? Happy to rebase or address feedback.

jfeng18:
Hi @karthik-indla, requesting a review when you have a moment. Fixes #4988 (delete/update full entity scan causing 10s+ timeouts at scale). Rebased, CI green, focused 2-file diff.

jfeng18:
Synced with the latest main. Tests strengthened to 28 (tenant isolation, per-entity failure tolerance, scan limit, _bulk smoke tests), and the fast-path fallback now logs at debug level for diagnosability. Python CI (3.10/3.11/3.12) and CI Gate are green on 670286c2. Would appreciate a review when you have a moment.

pan-long:
Independent reproduction on Qdrant (this scope has now crossed the 10k boundary)

Environment: mem0ai 2.1.0 (PyPI), qdrant-client 1.19.1, Qdrant server, on-disk collection, single-tenant scope.

The exact-match lookup path is now truncated for this scope:

- entities in scope: 11,401
- `entity_store.list(filters=..., top_k=10000)` returns exactly 10,000 -> 1,401 rows (~12%) are invisible to the exact-match path
- `top_k=50000` (and a raw Qdrant scroll) returns all 11,401, so this is purely the hardcoded limit, not a data problem

Same pattern appears at `mem0/memory/main.py:589, 669, 2253, 2330, 2346`.

**No duplicate rows observed -- the semantic fallback absorbs it**

This may be relevant to the duplicate-row discussion (#6925 / #7282). With ~12% of rows unreachable by the exact lookup, I expected the "lookup fails -> insert a duplicate entity row" shape to show up. Measured across 12,611 entity rows: 0 duplicate normalized names.

The reason looks like the `score >= 0.95` semantic fallback in `_upsert_entity`: a name that misses the exact match still retrieves its own row (cosine ~1.0 for identical text) and appends the `memory_id`. So the failure mode is masked here -- but only while the two texts stay near-identical; any variant that lands below 0.95 similarity would mint a second row.

**Quantified listing cost**

Backfilling entity links for ~2,000 memories took ~4.2 s/memory, and listing up to 10,000 rows per entity upsert is the dominant part of that. Steady-state traffic is much smaller, so this is the bulk cost, not a per-turn one -- but it scales with the scope's entity count, which is the property #4988 reports.

**Why the fast path matters here**

A filter on the array field (`linked_memory_ids` contains `memory_id`) is natively supported by Qdrant, so the fast path in this PR removes both the truncation and the full listing on this backend. Posting mainly so the thread has a reproduction on a second backend, since #4988 was reported against pgvector with 35K entities.

Happy to measure anything else you'd want on the Qdrant side.

jfeng18:
Thanks @pan-long -- a second-backend reproduction with quantified numbers is exactly what this thread needed.

To add to it: we validated the fast-path contract against a real in-memory Qdrant engine -- the targeted filter `{"user_id": "u1", "linked_memory_ids": "m1"}` compiles to `FieldCondition(key="linked_memory_ids", match=MatchValue(value="m1"))`, and the array-element match works natively: the linked row is returned, same-tenant rows without the id are excluded, and cross-tenant rows are excluded by the tenant filter. Since the server filters before returning, the fast path's `top_k` cap doesn't bite in practice (typically a handful of matching rows); the 28-test suite covers fast path (hit/empty/raise -> fallback), tenant isolation, scan limit, and `_bulk` smoke.

Your semantic-fallback masking observation is a sharp one -- restoring exact-lookup reachability should reduce the reliance on the >= 0.95 fallback absorbing truncation-induced misses.

You're also right that this PR only fixes the entity cleanup path (delete/update); the remaining hardcoded `top_k=10000` listings in `_upsert_entity` are a related but separate issue that deserves its own change.

Happy to coordinate on further Qdrant-side measurements if useful.

pan-long:
Thanks -- here is the Qdrant-side number this thread was missing. Same scope as my earlier
comment (mem0ai 2.1.0, qdrant-client 1.19.1, Qdrant server 1.19.1, on-disk collection, single
tenant), using the same call shape as `Qdrant.list()` (`mem0/vector_stores/qdrant.py:573`:
`scroll(limit=top_k, with_payload=True, with_vectors=False)`), measured 2026-09-23:

- the scope has grown since that comment (11,401 entity rows then; 12,705 at the time of
  measurement, 12,740 eleven minutes later) and, because the cap is fixed at 10,000 while the
  scope grows, the unreachable side grows one-for-one with it: that measurement put 2,705 rows
  (21.3%) outside the listing, up from 1,401 (12%)
- that listing step costs ~100-270 ms across repeated runs (medians ~100-200 ms); the targeted
  filter costs ~18-30 ms (~5-10x) and returns a handful of rows (2-16 across the ids we
  checked). The Python-side scan over the 10,000 returned rows is ~1-2 ms, so the win here is
  mostly the 10,000 fewer payloads plus client-side work rather than an indexed lookup:
  `linked_memory_ids` carries no payload index on this collection (payload_schema has only
  user_id/agent_id/run_id/actor_id)
- exactness: across ~26 distinct memory ids the targeted filter returned the same row set as a
  full paged scan over ALL tenant rows -- 0 mismatches; it also returned rows outside the
  10,000-row window, which the windowed listing cannot reach
- what the coverage gap costs: 355 rows outside the window still link >= 2 live memories (357
  in a re-run eleven minutes later), so deleting any of those memories leaves the entity links
  behind. The targeted filter returns those rows. On this backend that coverage is the half of
  #4988 that matters

One caveat so it is not read as fixed everywhere: the fallback branch still lists with the same
10,000 cap, so on a store without array-element match semantics the gap stays open -- the fast
path is what recovers it here.

Agreed the remaining 10,000-capped listings are a separate change, with one correction to the
naming: in 2.1.0 they are `_existing_entities_by_text` (main.py:589 sync / 2253 async; called
from the upsert paths, including `_upsert_entity`) and `_bulk_clear_entity_store` (2330 async) --
this PR replaces the two on the cleanup path (669 sync / 2346 async). #7173 and PR #7282 look
adjacent but are different fixes. I don't see a report for the cap on those call sites -- happy
to open one with these numbers, or to fold it into this thread if you'd rather keep it in one
place.

jfeng18:
Thanks @pan-long -- the 0-mismatch verification and the 355-row leak demonstration are exactly the evidence this thread needed: rows outside the 10k window that still link >= 2 live memories are the deletion-leak half of #4988 in production, which is more persuasive than any timing number. Your caveat matches our understanding of the fallback boundary -- the fast path is the fix on Qdrant, while stores without array-element match semantics keep the original behavior (zero regression by design).

On the proposal: I'd suggest a separate issue. This PR's scope is strictly the cleanup path (delete/update), and the remaining caps in `_existing_entities_by_text` (upsert paths) and `_bulk_clear_entity_store` deserve independent tracking and review -- your numbers (12% -> 21.3% growth, 18-30 ms targeted vs 100-270 ms listing) would make a strong reproduction basis for it, and linking the issue back to this thread keeps the context in one place.

One more supporting data point: the missing payload index matches what we found in review -- `_create_filter_indexes` currently only indexes user_id/agent_id/run_id/actor_id, so adding `linked_memory_ids` there is a natural follow-up improvement, also outside this PR's scope.

PR #5687, opened 2026-06-20
fix(notices): derive scale counts for Redis and search backends

### LINKED ISSUE #5678: notices: scale warnings cannot determine collection counts for Redis, Elasticsearch, and OpenSearch
### Component

Core / Python SDK

### Summary

While working on PR #5608, I noticed a related issue in the notices subsystem.

`_get_provider_memory_count()` relies on `_extract_count()`, which expects provider responses to expose top-level keys such as `count`, `points_count`, `vectors_count`, or `indexed_vectors_count`.

However:

- Elasticsearch/OpenSearch `indices.get()` returns index metadata rather than a top-level `count`.
- Redis `SearchIndex.info()` exposes fields such as `num_docs` instead of `count`.

As a result, even after the interface mismatch addressed in #5608 is resolved, `_get_provider_memory_count()` would still return `None` for these providers, preventing the notices subsystem from determining collection counts.

### Steps to Reproduce

1. Configure Mem0 to use Redis, Elasticsearch, or OpenSearch as the vector store.
2. Call `notices._get_provider_memory_count(memory_instance)` where `memory_instance.vector_store` is one of these providers.
3. Observe that `col_info()` executes successfully but `_get_provider_memory_count()` returns `None`.

### Expected Behavior

The notices subsystem should be able to retrieve collection counts consistently across supported vector store implementations.

### Actual Behavior

`col_info()` executes successfully, but `_extract_count()` cannot interpret the provider-specific response formats, causing `_get_provider_memory_count()` to return `None`.

### Environment

* mem0 version: current `main`
* Python version: 3.12
* OS: Linux

### Additional Context

This issue was discovered while working on PR #5608, which fixes the `col_info()` interface mismatch for these providers. That PR intentionally keeps its scope limited to the interface contract; this issue tracks the remaining provider-specific count extraction work.

rodboev:
## Linked Issue

Closes #5678

## Summary

Scale-threshold notices could not derive collection sizes for Redis, Valkey, Elasticsearch, or OpenSearch once the helper fell through the direct `vector_store.count()` path. This change keeps the existing fast path, teaches the notices count parser to read `num_docs` payloads, reuses the Redis schema index name when `collection_name` is not exposed, and adds a notices-local fallback for Elasticsearch and OpenSearch when `col_info()` only returns index metadata.

## Root cause

`_get_provider_memory_count()` stopped after `col_info()` and immediately returned `_extract_count(col_info())`. That worked for providers with top-level `count`-style fields, but Redis and Valkey expose collection size as `num_docs`, Redis keeps its index name under `schema["index"]["name"]` instead of `collection_name`, and the Elasticsearch and OpenSearch providers currently return `client.indices.get(...)` metadata that does not include any recognized count field. As a result, the scale warning could stay silent even when the collection was already above the threshold.

## Fix

- widen `_extract_count()` to recognize `num_docs` in both dict and attribute form
- keep the existing `vector_store.count()` fast path and the existing `col_info()` attempt
- resolve the collection name from either `vector_store.collection_name` or the Redis-style `vector_store.schema["index"]["name"]` shape before calling `col_info(name)`
- if `col_info()` succeeds but does not yield a count, fall back locally to `vector_store.client.count(index=vector_store.collection_name)` when that client API is available
- keep all non-negative coercion and field parsing inside `_extract_count()`

## Scope

This PR stays inside `mem0/memory/notices.py` and `tests/memory/test_notices.py`. It does not change any provider implementation under `mem0/vector_stores/`, and it remains complementary to `#5608`, which addresses the `col_info()` interface mismatch without changing notices-side count extraction.

## Type of Change

- [ ] Documentation update

## Test Coverage

- [x] Added or updated unit tests
- [ ] Added or updated integration tests
- [ ] Tested manually
- [ ] No tests needed

Focused local validation:

- `hatch run pytest tests/memory/test_notices.py -q`
- `hatch run ruff check mem0/memory/notices.py tests/memory/test_notices.py`
- `hatch run python -m py_compile mem0/memory/notices.py tests/memory/test_notices.py`

Additional local status:

- `hatch run ruff format --check mem0/memory/notices.py tests/memory/test_notices.py` still reports both files would be reformatted locally; that formatting cleanup is left out of this bugfix to keep the change focused
- the repo-config full-suite commands `make lint` and `make test` could not be run in this environment because `make` is not installed locally

## Breaking Changes

N/A

## Checklist

- [x] Self-review performed
- [x] Added tests that prove the fix works
- [ ] Documentation updated if needed

### DIFF
--- a/mem0/memory/notices.py
+++ b/mem0/memory/notices.py
@@ -1419,6 +1419,12 @@ def _get_provider_memory_count(memory_instance) -> Optional[int]:
         col_info = getattr(vector_store, "col_info", None)
         if callable(col_info):
             collection_name = getattr(vector_store, "collection_name", None)
+            if collection_name is None:
+                schema = getattr(vector_store, "schema", None)
+                if isinstance(schema, dict):
+                    index = schema.get("index")
+                    if isinstance(index, dict):
+                        collection_name = index.get("name")
             if collection_name is not None:
                 try:
                     info = col_info(collection_name)
--- a/mem0/memory/notices.py
+++ b/mem0/memory/notices.py
@@ -1418,7 +1418,18 @@ def _get_provider_memory_count(memory_instance) -> Optional[int]:
     try:
         col_info = getattr(vector_store, "col_info", None)
         if callable(col_info):
-            return _extract_count(col_info())
+            info = col_info()
+            value = _extract_count(info)
+            if value is not None:
+                return value
+
+            client = getattr(vector_store, "client", None)
+            collection_name = getattr(vector_store, "collection_name", None)
+            client_count = (
+                getattr(client, "count", None) if client is not None and collection_name is not None else None
+            )
+            if callable(client_count):
+                return _extract_count(client_count(index=collection_name))
     except Exception:
         return None
 
@@ -1430,7 +1441,7 @@ def _extract_count(info: Any) -> Optional[int]:
         return None
 
     if isinstance(info, dict):
-        for key in ("count", "points_count", "vectors_count", "indexed_vectors_count"):
+        for key in ("count", "points_count", "vectors_count", "indexed_vectors_count", "num_docs"):
             value = _coerce_nonnegative_int(info.get(key), None)
             if value is not None:
                 return value
@@ -1445,7 +1456,7 @@ def _extract_count(info: Any) -> Optional[int]:
         except Exception:
             return None
 
-    for attr in ("count", "points_count", "vectors_count", "indexed_vectors_count"):
+    for attr in ("count", "points_count", "vectors_count", "indexed_vectors_count", "num_docs"):
         value = _coerce_nonnegative_int(getattr(info, attr, None), None)
         if value is not None:
             return value
--- a/mem0/memory/notices.py
+++ b/mem0/memory/notices.py
@@ -1418,13 +1418,19 @@ def _get_provider_memory_count(memory_instance) -> Optional[int]:
     try:
         col_info = getattr(vector_store, "col_info", None)
         if callable(col_info):
-            info = col_info()
+            collection_name = getattr(vector_store, "collection_name", None)
+            if collection_name is not None:
+                try:
+                    info = col_info(collection_name)
+                except TypeError:
+                    info = col_info()
+            else:
+                info = col_info()
             value = _extract_count(info)
             if value is not None:
                 return value
 
             client = getattr(vector_store, "client", None)
-            collection_name = getattr(vector_store, "collection_name", None)
             client_count = (
                 getattr(client, "count", None) if client is not None and collection_name is not None else None
             )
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Review — fix(notices): derive scale counts for Redis and search backends

**Verdict: LGTM** — ready to squash-merge.

---

### Strengths

- Correctly identifies and addresses all three failure modes from #5678: `num_docs` field recognition for Redis/Valkey, schema-based index name resolution for `RedisDB` (which lacks `collection_name` as an attribute), and the `client.count()` fallback for Elasticsearch/OpenSearch when `col_info()` returns index metadata without a count field.
- Change stays properly scoped to `mem0/memory/notices.py` and its test file — no provider implementations touched, which is the right abstraction boundary.
- The `TypeError` catch-and-retry pattern (try `col_info(name)`, fall back to `col_info()`) pragmatically handles signature variance across providers without forcing a shared interface.
- Exception masking (`except Exception: return None`) at the `col_info` block level is appropriate for a best-effort notices probe.

### Issues

**Minor — `RedisDB.col_info(name)` ignores its argument** (`mem0/vector_stores/redis.py:295–296`): the real `RedisDB.col_info(self, name)` accepts `name` but internally calls `self.index.info()` regardless. The notices fix correctly resolves the index name from the schema, but passing it to `col_info` has no effect — the provider ignores it silently. This is not introduced by this PR and doesn't affect correctness here; it's a pre-existing inconsistency in the Redis provider that is worth a follow-up cleanup on the provider side (either use `name` or make it `name=None` and document that behavior).

**Minor — `ruff format` divergence**: the PR author notes both changed files would be reformatted by `ruff format`. This is fine for a focused bugfix — a formatting-only commit can follow separately.

### Assessment

**Ready to merge: Yes.** The fix is architecturally sound (notices-side extraction, complementary to #5608's provider-side interface fix), covers all backends described in the issue, and is well-tested.

---

### Evidence

**CI:** Python SDK / build_mem0 (3.10, 3.11, 3.12) — all pass. CI Gate — pass. Vercel advisory-only (deploy perms, not merge-blocking, ignored per policy).

**Targeted tests (101 tests, `tests/memory/test_notices.py`):** 101 passed in 3.34s.

**Lint (`ruff check mem0/memory/notices.py tests/memory/test_notices.py`):** All checks passed.

**Red-green proof (`test_scale_threshold_provider_count_helpers_are_safe`):**
- With PR fix (green): 1 passed — `_get_provider_memory_count` returns 2300 for RedisInfoStore, 2400 for SearchMetadataStore.
- With `notices.py` reverted to main (red): 1 FAILED — `AssertionError: assert None == 2300` for RedisInfoStore case, confirming the test bites the bug.
- With PR fix restored (green again): 1 passed.

kartik-mem0 (approved):

PR #5689, opened 2026-06-20
fix(ts-oss): skip system messages by role in infer:false add()

### LINKED ISSUE #5676: bug(ts-oss): infer:false skips messages by content instead of role, storing system prompts as memories
### Bug

In the TS OSS SDK's `infer: false` path, `add()` decides whether to skip a message by comparing its **content** to the string `"system"`, instead of its **role**.

`mem0-ts/src/oss/src/memory/index.ts` (~L701):

```ts
if (!infer) {
    if (message.content === "system") {   // should be message.role === "system"
      continue;
    }
    ...
```

The Python reference does it by role (`mem0/memory/main.py:773` → `if message_dict["role"] == "system": continue`).

### Impact

With `add(messages, { infer: false })`:

1. **System prompts get stored as memories** — `{ role: "system", content: "You are a helpful assistant" }` has content `!= "system"`, so it is **not** skipped and gets stored, then pollutes `search`/`getAll`.
2. **Legitimate memories are dropped** — `{ role: "user", content: "system" }` matches `content === "system"` and is silently skipped.

### Expected

System messages are skipped by role; all other messages (including a user message whose content is `"system"`) are stored. Matches the Python behaviour.

### Fix

Compare `message.role === "system"`.

Bartok9:
## Summary

- In the TS OSS SDK, `add(messages, { infer: false })` decided whether to skip a message by comparing `message.content` to `"system"` instead of `message.role`.
- User-facing impact: legitimate user messages whose content is the literal string `"system"` were silently dropped, and (defensively) system messages were only kept out by an upstream filter rather than this branch itself.

## Motivation

Closes #5676.

`addToVectorStore()`'s `infer: false` branch did:

```ts
if (message.content === "system") { continue; }   // wrong
```

The Python reference skips by role (`mem0/memory/main.py:773` → `if message_dict["role"] == "system": continue`). The content-based check has two failure modes:

1. A `{ role: "user", content: "system" }` message matches `content === "system"` and is **silently dropped**.
2. It relies on `parse_vision_messages` (which already strips `role === "system"` upstream) to keep system prompts out — the branch's own intent (skip system) was expressed against the wrong field.

## Fix

Compare `message.role === "system"`, matching the Python SDK.

```ts
if (message.role === "system") { continue; }
```

## Verification

- `npx jest src/oss/tests/memory.add.test.ts` — 13 passed
- `npx jest src/oss/tests/memory.crud.test.ts src/oss/tests/memory.add.test.ts` — 36 passed

## Real behavior proof

Two regression tests added to `memory.add.test.ts`:
- `with infer=false skips system messages by role, not content`
- `with infer=false stores a user message whose content is 'system'`

Captured jest output **with the fix** (both pass):

```
✓ with infer=false skips system messages by role, not content (1 ms)
✓ with infer=false stores a user message whose content is 'system' (73 ms)
Tests:       13 passed, 13 total
```

Captured jest output **without the fix** (revert source to `message.content === "system"`):

```
✓ with infer=false skips system messages by role, not content
✕ with infer=false stores a user message whose content is 'system' (1 ms)
Tests:       1 failed, 12 passed, 13 total
```

The `content === "system"` user-message test fails without the fix and passes with it.

## What was NOT tested

The system-prompt-not-stored assertion passes on both source variants because `parse_vision_messages` already strips `role === "system"` before `addToVectorStore`. The fix is still correct (matches Python, removes reliance on an upstream filter for this branch's own intent) and the dropped-user-message case is genuinely live.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -717,7 +717,7 @@ export class Memory {
     if (!infer) {
       const returnedMemories: MemoryItem[] = [];
       for (const message of messages) {
-        if (message.content === "system") {
+        if (message.role === "system") {
           continue;
         }
         const memoryId = await this.createMemory(
--- a/mem0-ts/src/oss/tests/memory.add.test.ts
+++ b/mem0-ts/src/oss/tests/memory.add.test.ts
@@ -191,4 +191,34 @@ describe("Memory - add()", () => {
       expect.objectContaining({ event: "ADD" }),
     );
   });
+
+  // Regression: infer=false must skip system messages by ROLE, not by content.
+  // See https://github.com/mem0ai/mem0/issues/5676 — comparing message.content
+  // to "system" both (a) stored system prompts as memories and (b) dropped
+  // legitimate user messages whose content happened to equal "system".
+  test("with infer=false skips system messages by role, not content", async () => {
+    const messages = [
+      { role: "system", content: "You are a helpful assistant" },
+      { role: "user", content: "I live in Berlin" },
+    ];
+    const result: SearchResult = await memory.add(messages, {
+      userId,
+      infer: false,
+    });
+    const stored = result.results.map((r) => r.memory);
+    // System prompt must NOT be stored.
+    expect(stored).not.toContain("You are a helpful assistant");
+    // The user message must be stored.
+    expect(stored).toContain("I live in Berlin");
+  });
+
+  test("with infer=false stores a user message whose content is 'system'", async () => {
+    const messages = [{ role: "user", content: "system" }];
+    const result: SearchResult = await memory.add(messages, {
+      userId,
+      infer: false,
+    });
+    // A user message with content === "system" must NOT be skipped.
+    expect(result.results.map((r) => r.memory)).toContain("system");
+  });
 });


yashs33244 (commented):
Nice catch - the role-vs-content mix-up is a real silent data-loss bug, and I like that the fix is one line matching the Python reference plus a test that actually fails on the old code. The "What was NOT tested" note is appreciated too; it's exactly right that the system-prompt half is latent today.

One bit of context worth dropping in for whoever reads this later: the reason the system case can't fire right now is that TS `parse_vision_messages` (utils/memory.ts) drops `role === "system"` messages outright (no `else`), whereas Python's `parse_vision_messages` actually *preserves* them (utils.py: `if role == "system": returned_messages.append(msg)`). So in Python the `main.py:773` role check does real work, and in TS it's currently a no-op guard. That's not a problem with this PR - the role check is the correct foundation - but full Python parity for system-message handling depends on a separate `parse_vision_messages` alignment, so the first test (system not stored) passes regardless of this change while the second is the genuine regression guard. Happy to see this merge as-is.

kartik-mem0:
## Review — PR #5689

**Verdict: LGTM** — ready to squash-merge (`gh pr merge --squash`).

---

### Strengths

- **Fix is exactly correct.** `message.role === "system"` matches the Python SDK reference at `mem0/memory/main.py:773` (sync) and `:2311` (async). The old `message.content === "system"` compared the wrong field, producing two live bugs: legitimate user messages whose text was `"system"` were silently dropped, and system prompts whose content wasn't the literal string `"system"` were stored as memories.
- **`addToVectorStore` is private with one caller.** The fix fully covers the affected surface — there is no alternate entry point.
- **The second regression test is a genuine live test.** `"with infer=false stores a user message whose content is 'system'"` constructs `[{ role: "user", content: "system" }]` and asserts `expect(result.results.map((r) => r.memory)).toContain("system")`. Red-green proof confirmed: the test fails on the reverted code and passes with the fix.
- **TS single-class architecture is fully covered.** No `AsyncMemory` class exists in TS — `Memory.add()` is already async — so no missed counterpart path.
- **PR description is transparent.** The author correctly calls out that the first regression test passes on both code variants due to the upstream `parse_vision_messages` filter.

---

### Issues

**Minor — first regression test name is misleading (`memory.add.test.ts:194`).**
The test `"with infer=false skips system messages by role, not content"` does not exercise the fix. `parse_vision_messages` (`src/oss/src/utils/memory.ts:29`) strips `role === "system"` messages before they reach `addToVectorStore`, so the `!infer` role-check is never exercised by this test on either the old or new code. The test validates end-to-end behavior via the upstream filter, not the fix's mechanism. The test is not wrong (the behavior it verifies is real), but the name implies it proves the fix when it proves something adjacent. The PR description explains this honestly, but a future reader relying on the test name alone will be misled.

**Minor — undocumented Python↔TS divergence in `parse_vision_messages`.**
Python's `parse_vision_messages` passes system messages through to `_add_to_vector_store`, where the `role == "system"` check is the sole guard. TS's `parse_vision_messages` strips system messages upstream, making the `addToVectorStore` check defense-in-depth. Neither approach is wrong, but the contract is undocumented. If a future contributor aligns TS's utility with Python's pass-through behavior, the `infer:true` path in TS (which also has no system-message guard of its own) would silently start including system prompts in the LLM extraction context. A comment at the `parse_vision_messages` call site would prevent this class of refactor mistake. This is pre-existing and out of scope for this PR — worth a follow-up issue.

---

### Recommendations

1. Consider renaming the first new test to `"with infer=false does not store system messages (end-to-end)"` or adding a brief inline comment that the `addToVectorStore` role-check is defense-in-depth while the upstream `parse_vision_messages` is the effective guard. Not blocking.
2. Open a follow-up issue noting the undocumented `parse_vision_messages` divergence between Python and TS — a comment in `add()` at the call site would prevent a future refactor from inadvertently removing the protection for the `infer:true` path.

---

### Evidence

| Check | Result |
|---|---|
| CI Gate | PASS |
| TypeScript SDK / build_ts_sdk (20, 22) | PASS |
| TypeScript SDK / integration_ts_sdk (20, 22) | PASS |
| TypeScript SDK / changelog_check | PASS |
| Vercel deploy preview | FAIL — advisory only, fork deploy perms required, ignored per pipeline rules |
| Red-green proof | SUCCESS — `"with infer=false stores a user message whose content is 'system'"` FAILS on reverted code (`content === "system"` drops the user message), PASSES on fixed code (`role === "system"` leaves user message intact) |
| Full test suite (`src/oss/tests/memory.add.test.ts`) | 13/13 PASS |
| Duplicate cluster | None — only PR referencing issue #5676 |
| Danger flags | None |

Issue #5351, opened 2026-06-03
openclaw plugin: skills.recall.rerank / keywordSearch config keys are inert (never sent to search)


jeades:
## Summary

The `@mem0/openclaw-mem0` plugin documents two recall config keys — `skills.recall.rerank` (default `true`) and `skills.recall.keywordSearch` (default `true`) — but neither is ever threaded into a search call. Users who set `rerank: true` expecting Mem0 Platform Advanced Retrieval reranking get only the plugin's local `rankMemories()` + `budgetMemories()` ordering. The keys are effectively inert.

Observed with plugin `1.0.11` and confirmed unchanged in `1.0.12`.

## Where the keys are documented

- `openclaw.plugin.json` config schema and `README.md` both list `skills.recall.rerank` and `skills.recall.keywordSearch`, defaulted `true`.
- `SkillsConfig` in `openclaw/types.ts` declares them.

## Why they're inert (evidence from `openclaw/`)

- `openclaw/recall.ts` builds `searchOpts` with only `user_id` / `top_k` / `threshold` / `source`, under the comment `v3.0.0: keyword_search, reranking, filter_memories removed`. It never reads `rerank` / `keywordSearch`.
- `buildSearchOptions` in `openclaw/index.ts` reads `recallCfg.threshold` but never `rerank` / `keywordSearch`.
- `openclaw/providers.ts` `PlatformProvider.search` and `providerToBackend` both omit them.
- `tests/providers.test.ts` asserts `provider.search` is called without them.
- The only rerank wiring that exists — `if (opts.rerank) payload.rerank = true` in `PlatformBackend.search` (raw `POST /v2/memories/search/`, used by the `memory_search` tool) — is never reached because the tool never sets `opts.rerank`.

So both the auto-recall path and the `memory_search` tool path omit `rerank`; relevance ordering comes entirely from local ranking, not the Platform reranker.

Note: `mem0-ts` `MemoryClient.search` spreads `...camelToSnakeKeys(rest)` and `SearchMemoryOptions` still lists `rerank?: boolean`, so threading a `rerank` field through to the API is straightforward.

## Impact

Config that advertises Platform reranking silently has no effect. This is a correctness/expectations gap, not a crash.

## Suggested resolution

Either:

1. Thread `recallCfg.rerank` (and decide on `keywordSearch`) into `searchOpts` / `buildSearchOptions` / `PlatformProvider.search` so the documented keys reach the Mem0 search payload, or
[...]

PR #5690, opened 2026-06-20
fix(claude-plugin): rerank auto-injected memory context by default

### LINKED ISSUE #5684: Claude Code plugin: memory injection never sends `rerank`, so injected context is unranked (sibling of #5351; v2 endpoint also doesn't rerank by default)
## Summary

The **Claude Code / Cowork plugin** (`mem0` from the `mem0-plugins` marketplace) injects memories on every prompt via `scripts/on_user_prompt.sh`, but that hook does a raw `POST /v2/memories/search/` and **never sends `rerank`**. As a result, every auto-injected memory set is ordered by raw vector similarity, not Platform Advanced Retrieval reranking.

This is the same root pattern as #5351 (openclaw plugin: `rerank` config key never threaded into the search call), but in a different plugin and via a shell hook. Filing separately so the Claude-Code-specific fix location is tracked; cross-linking #5351 since a maintainer fixing one should fix both.

There is an additional, sharper wrinkle here: **the `/v2/memories/search/` endpoint does not rerank by default**, even though the docs say it should — so this can't be dismissed as "the default already covers it."

## Where

`scripts/on_user_prompt.sh` builds the request body as:

```bash
BODY=$(jq -n --arg query "$PROMPT" --arg user_id "$USER_ID" \
  '{query: $query, filters: {user_id: $user_id}, top_k: 5}')   # no rerank
# → curl -X POST https://api.mem0.ai/v2/memories/search/
```

## The docs-vs-reality part

The v2 search API reference table states reranking defaults differ by version:

| Param    | V1/V2            | V3                          |
| -------- | ---------------- | --------------------------- |
| `rerank` | **default `true`** | default `false` (pass true) |

But the live `POST /v2/memories/search/` endpoint does **not** rerank when the field is omitted. Verified by sending the exact body the hook sends vs. explicit `rerank` values against a populated user pool (same query, `top_k:5`, memory text elided):

| Variant | `rerank` | Top-5 scores | Top result |
| --- | --- | --- | --- |
| **A** (hook body) | *omitted* | `0.719, 0.673, 0.685, 0.701, 0.614` | memory X |
| **B** | `false` | `0.719, 0.673, 0.685, 0.701, 0.614` | memory X |
| **C** | `true` | `0.850, 0.842, 0.839, 0.830, 0.828` | **memory Y** |

**A is identical to B (`rerank:false`) and differs from C (`rerank:true`)** — different top memory, reordered set, distinctly higher and monotonically-descending reranker scores. So omitting `rerank` == no reranking on v2, contradicting the "V1/V2 default `true`" table. (Note A/B scores aren't even descending — raw retrieval order, not reranked.)

So either the REST endpoint's v2 default regressed to `false`, or the docs table describes the SDK client default rather than the REST API. Either way, a plugin can't rely on the documented v2 default.

## Impact

Every prompt's injected context is the **unranked** `top_k=5`. In testing, the single most relevant memory for a query landed at rank #1 only with `rerank:true` and was **absent from the top-5** without it — i.e. the best memory is silently dropped from the injected window.

## Suggested fix

1. Add `rerank: true` to the jq body in `on_user_prompt.sh` (and consider `keyword_search: true`). Ideally make it configurable, e.g. `rerank: ($MEM0_RERANK // true)` via an env var, since it adds ~150–200ms (well within the hook's 3s curl budget).
2. Separately, correct the v2 docs table (or clarify SDK-vs-REST) so it matches the endpoint's actual behavior.

## Secondary finding: hook and MCP server use different `user_id`s

The plugin hooks (capture in `on_pre_compact.py`/`on_stop.sh`, injection in `on_user_prompt.sh`) resolve `user_id = MEM0_USER_ID ?: $USER`. The bundled MCP server (`mcp.mem0.ai`, configured with only the token) writes/reads under the **account identity**, a different `user_id`. Consequence for users who use both: memories saved via the MCP `add_memory` tool never appear in hook injection (different pool), and hook-captured memories are invisible to the MCP `search_memories` tool unless `filters:{user_id: "<$USER>"}` is passed explicitly. Worth documenting, or aligning the MCP default `user_id` with the hook's.

## Environment

- Plugin `mem0` `0.1.0` (`mem0-plugins` marketplace), Claude Code on macOS
- Endpoint: `https://api.mem0.ai/v2/memories/search/`
- Related: #5351 (openclaw plugin — same omit-`rerank` pattern)

Bartok9:
## Summary

- The Claude Code plugin auto-injects memory context on every prompt but never requests reranking, so injected context is ordered by raw vector similarity instead of Platform Advanced Retrieval ranking.
- User-facing impact: the single most relevant memory for a query can fall outside the injected `top_k` window and is silently dropped from context.

## Motivation

Closes #5684.

The injection hooks (`on_user_prompt.sh`, `on_bash_output.sh`, `file_context.py`) call `_search.search_memories(...)` without passing `rerank`, which defaults to `False`. As the issue documents, the REST `/v*/memories/search/` endpoint does **not** rerank when the field is omitted (reranker scores stay non-monotonic / raw retrieval order), so auto-injected sets are never reranked even though reranking materially changes both the order and membership of the top-5.

## Fix

- Add a small, testable `should_rerank()` helper to `_search.py`. It defaults reranking **ON** for the hook-driven injection path (the extra ~150–200ms is well within the hooks' curl budget) and is **opt-out** via `MEM0_RERANK` (`0`/`false`/`no`/`off`, case-insensitive).
- Thread `rerank=should_rerank()` into all three injection call sites.

Scoped to the injection path; `search_memories(rerank=...)` already forwarded the flag — the bug was that callers never set it. Sibling of #5351 (openclaw plugin, same omit-`rerank` pattern); cross-linking so a maintainer can fix both. The docs-table correction noted in the issue is left as a separate docs change.

## Verification

- `python3 -m pytest tests/test_search.py` — 11 passed
- `python3 -m pytest tests/` (full plugin suite) — 153 passed
- `bash -n` on both modified shell hooks — clean
- `python3 -m py_compile` on `_search.py` + `file_context.py` — clean

## Real behavior proof

Five regression tests added to `tests/test_search.py`. Captured pytest output **with the fix**:

```
tests/test_search.py ...........                                          [100%]
11 passed in 0.03s
```

To prove the default-on behavior is actually load-bearing, I reverted the helper's default to `False` (pre-fix behavior) and reran:

```
>       assert should_rerank() is True
E       assert False is True
tests/test_search.py:150: AssertionError
1 failed, 10 passed in 0.03s
```

`test_should_rerank_defaults_true` fails without the fix and passes with it. The body-shape tests confirm `rerank` is omitted unless requested and forwarded as `true` when set.

## What was NOT tested

The live REST endpoint's default behavior (the issue's measured A/B/C reranker scores) — tests assert the request body shape and the helper's env logic under a mocked `urlopen`, not a real network call.

### DIFF
--- a/integrations/mem0-plugin/scripts/_search.py
+++ b/integrations/mem0-plugin/scripts/_search.py
@@ -7,10 +7,29 @@ All pre-fetch hooks use this instead of duplicating urllib boilerplate.
 from __future__ import annotations
 
 import json
+import os
 import urllib.request
 
 SEARCH_URL = "https://api.mem0.ai/v3/memories/search/"
 SEARCH_TIMEOUT = 5
+
+
+def should_rerank() -> bool:
+    """Whether auto-injection searches should request Platform reranking.
+
+    The REST search endpoint does not rerank when ``rerank`` is omitted, so
+    auto-injected context is ordered by raw vector similarity and the single
+    most relevant memory can fall outside the injected top_k window. We default
+    reranking ON for the hook-driven injection path (the extra ~150-200ms is
+    well within the hook's curl budget) and let users opt out via MEM0_RERANK.
+
+    MEM0_RERANK is read case-insensitively; ``0``, ``false``, ``no``, and
+    ``off`` disable reranking. Anything else (including unset) enables it.
+    """
+    raw = os.environ.get("MEM0_RERANK")
+    if raw is None:
+        return True
+    return raw.strip().lower() not in ("0", "false", "no", "off", "")
 
 
 def _do_search(api_key: str, payload: dict) -> list[dict]:
--- a/integrations/mem0-plugin/scripts/file_context.py
+++ b/integrations/mem0-plugin/scripts/file_context.py
@@ -23,7 +23,7 @@ sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 from _formatting import TYPE_ICONS, format_age
 from _identity import resolve_api_key, resolve_user_id
 from _project import resolve_project_id
-from _search import search_memories
+from _search import search_memories, should_rerank
 
 FILE_READ_GATE_MIN_BYTES = 1500
 MAX_RESULTS = 5
@@ -93,6 +93,7 @@ def search_file_context(
         api_key, user_id, project_id, query,
         top_k=MAX_RESULTS, threshold=0.3,
         global_search=global_search,
+        rerank=should_rerank(),
     )
 
     results = results[:MAX_RESULTS]
--- a/integrations/mem0-plugin/scripts/on_bash_output.sh
+++ b/integrations/mem0-plugin/scripts/on_bash_output.sh
@@ -77,15 +77,16 @@ RESULTS=$(PYTHONPATH="$SCRIPT_DIR" MEM0_SEARCH_QUERY="$ERROR_QUERY" MEM0_SEARCH_
   python3 -c "
 import os, sys
 sys.path.insert(0, os.environ.get('PYTHONPATH', '.'))
-from _search import search_memories, format_results_for_context
+from _search import search_memories, format_results_for_context, should_rerank
 
 api_key = os.environ.get('MEM0_API_KEY', '')
 user_id = os.environ.get('MEM0_SEARCH_USER', 'default')
 project_id = os.environ.get('MEM0_PROJECT_ID', 'unknown')
 query = os.environ.get('MEM0_SEARCH_QUERY', '')
+rerank = should_rerank()
 
-r1 = search_memories(api_key, user_id, project_id, query, metadata_type='anti_pattern', top_k=3)
-r2 = search_memories(api_key, user_id, project_id, query, metadata_type='bug_fix', top_k=3)
+r1 = search_memories(api_key, user_id, project_id, query, metadata_type='anti_pattern', top_k=3, rerank=rerank)
+r2 = search_memories(api_key, user_id, project_id, query, metadata_type='bug_fix', top_k=3, rerank=rerank)
 
 seen = set()
 combined = []
--- a/integrations/mem0-plugin/scripts/on_user_prompt.sh
+++ b/integrations/mem0-plugin/scripts/on_user_prompt.sh
@@ -120,14 +120,15 @@ if [ -n "$HAS_RESUME" ]; then
   RESUME_RESULTS=$(PYTHONPATH="$SCRIPT_DIR" MEM0_SEARCH_USER="$USER_ID" python3 -c "
 import os, sys
 sys.path.insert(0, os.environ.get('PYTHONPATH', '.'))
-from _search import search_memories, format_results_for_context
+from _search import search_memories, format_results_for_context, should_rerank
 
 api_key = os.environ.get('MEM0_API_KEY', '')
 user_id = os.environ.get('MEM0_SEARCH_USER', 'default')
 project_id = os.environ.get('MEM0_PROJECT_ID', 'unknown')
+rerank = should_rerank()
 
-state = search_memories(api_key, user_id, project_id, 'session state current task', metadata_type='session_state', top_k=3)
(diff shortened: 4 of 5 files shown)


kartik-mem0:
**Review — fix(claude-plugin): rerank auto-injected memory context by default**

## Strengths

- Correctly diagnoses the root cause: all three hook-driven injection call sites were calling `search_memories()` with the default `rerank=False`, which omits the `rerank` key from the HTTP payload entirely. The REST endpoint does not rerank when the field is omitted, so injected context was always ordered by raw vector similarity.
- The `should_rerank()` helper in `_search.py:17–32` is the right abstraction: shared, testable, defaults ON for the injection path, and opt-out via `MEM0_RERANK` follows the existing `MEM0_*` env-var namespace.
- Coverage is complete: all three injection call sites are patched — `on_user_prompt.sh`, `on_bash_output.sh`, and `file_context.py`. The Cursor wrapper (`on_user_prompt_cursor.sh`) delegates to `on_user_prompt.sh`, and `on_file_read.sh` delegates to `file_context.py` — both are covered transitively. Verified by grepping every `search_memories(` call in `scripts/`.
- `search_memories()` keeps `rerank=False` as its own default — correct, because only hook-driven injection should rerank by default; callers making explicit searches shouldn't have reranking forced on them.
- Reads env at call-time (not import-time), which is correct for a hook-per-process model.
- Lint: `ruff check` passes clean on `_search.py` and `file_context.py`.

## Issues

None confirmed by code tracing or test run.

**Minor (optional):** `file_context.py:96` calls `should_rerank()` inside `search_file_context()` rather than accepting it as a parameter. The single caller in `main()` can't override it without the env var. With one caller today this is fine; if `search_file_context` grows callers, a `rerank: bool = ..." param would be cleaner. Not a blocker.

**Note (not a blocker):** The issue's A/B/C verification was against the v2 endpoint, but the plugin uses `SEARCH_URL = ".../v3/memories/search/"` (`_search.py:13`). The fix is correct regardless: it explicitly sends `rerank=True` in the payload instead of relying on any endpoint default. Defensive and right for either version.

## Test evidence

**With fix — 11 passed:**
```
tests/test_search.py::test_search_memories_returns_results PASSED
tests/test_search.py::test_search_memories_omits_rerank_by_default PASSED
tests/test_search.py::test_search_memories_forwards_rerank_true PASSED
tests/test_search.py::test_should_rerank_defaults_true PASSED
tests/test_search.py::test_should_rerank_opt_out_values PASSED
... (11 passed in 0.08s)
```

**Red-green proof (reverted `_search.py` to main, kept tests):**
```
tests/test_search.py::test_should_rerank_defaults_true FAILED — ImportError: cannot import name 'should_rerank' from '_search'
tests/test_search.py::test_should_rerank_opt_out_values FAILED — ImportError: cannot import name 'should_rerank' from '_search'
(2 failed, 1 passed in 0.03s)
```

**Restored fix — 11 passed in 0.02s.**

Regression tests bite exactly the change. The `test_search_memories_omits_rerank_by_default` test correctly asserts the low-level `search_memories()` behavior (key omitted when `rerank=False`); the new `test_should_rerank_*` tests cover the helper's env-var logic.

## Evidence block

- **CI:** CI Gate passes. Vercel check fails (advisory — fork perms, not merge-blocking). All real checks (lint, build, test matrix) pass or skip (no affected packages beyond the plugin, which has no separate CI job in the gate).
- **Targeted tests:** `tests/test_search.py` — 11 passed with fix, 2 failed without (red-green confirmed).
- **Lint:** ruff clean on both modified Python files.
- **Could NOT verify:** live v3 endpoint rerank behavior without a real API key + populated user pool. The fix is defensive and correct regardless.
- **Worktree:** `~/.cache/mem0-worktrees/review-pr-5690` at `6f4b5a43`.

## Assessment

LGTM. The fix correctly solves the real root cause across all injection call sites, uses the right abstraction layer, degrades gracefully (opt-out env var, no change to explicit-caller behavior), and the regression tests are load-bearing. Ready to squash-merge.

kartik-mem0 (approved):

PR #5692, opened 2026-06-20
fix: mongodb vector store list() returns nested format

### LINKED ISSUE #5691: MongoDB vector store: Memory.delete_all() raises TypeError (list() returns a flat list, not List[List[OutputData]])
### Bug

On a MongoDB-backed deployment, `Memory.delete_all(...)` raises `TypeError: 'OutputData' object is not iterable` (or `IndexError` on an empty collection).

Root cause: `mem0/vector_stores/mongodb.py` `list()` returns a flat `List[OutputData]`, but every other vector store returns `List[List[OutputData]]` (wrapped one level deep). `Memory.delete_all` (`mem0/memory/main.py`) does:

```python
for memory in memories:
```

On MongoDB, `[0]` is the first `OutputData`, so iterating it raises `TypeError`; on an empty collection `list()` returns `[]` and `[][0]` raises `IndexError`. Either way `delete_all` (a public API) is broken for MongoDB.

Confirmed the siblings all wrap: pinecone / chroma / milvus / elasticsearch / pgvector / supabase / faiss / weaviate / redis all `return [results]` (elasticsearch is even typed `-> List[List[OutputData]]`). MongoDB is the lone provider returning a flat list. It stayed latent because `get_all` (`_get_all_from_vector_store`) inspects the first element and tolerates both shapes, but `delete_all` hard-codes `[0]`.

### Reproduce

Configure mem0 with the MongoDB vector store, add a memory, then call `m.delete_all(user_id="...")` → `TypeError: 'OutputData' object is not iterable`.

### Fix

Wrap MongoDB `list()`'s return one level deep (`return [results]`, and `[[]]` on the empty/error path) to match every other store and the `[0]` contract that `delete_all`/`get_all` rely on.

axelray-dev:
## Problem

The MongoDB vector store's `list()` method returns `List[OutputData]` (flat), while all other vector stores return `List[List[OutputData]]` (nested). This causes `delete_all()` to raise `TypeError` when iterating over the result.

## Root cause

In `mem0/vector_stores/mongodb.py` lines 384-387, the `list()` method returns:
- `return results` (flat list)
- `return []` (flat empty list on error)

Other vector stores wrap results in an extra list:
- Chroma: `return [self._parse_output(results)]`
- Pinecone: `return [results]`
- PGVector: `return [[OutputData(...) for r in results]]`

## Fix

Wrap the MongoDB results in an extra list to match the interface contract:
- Line 385: `return results` → `return [results]`
- Line 388: `return []` → `return [[]]`

## Testing

Updated existing tests in `tests/vector_stores/test_mongodb.py` to verify nested return format. All 18 MongoDB tests pass.

Fixes #5691

### DIFF
--- a/mem0/vector_stores/mongodb.py
+++ b/mem0/vector_stores/mongodb.py
@@ -382,10 +382,10 @@ class MongoDB(VectorStoreBase):
             cursor = self.collection.find(query).limit(top_k)
             results = [OutputData(id=str(doc["_id"]), score=None, payload=doc.get("payload")) for doc in cursor]
             logger.info(f"Retrieved {len(results)} documents from collection '{self.collection_name}'.")
-            return results
+            return [results]
         except PyMongoError as e:
             logger.error(f"Error listing documents: {e}")
-            return []
+            return [[]]
 
     def reset(self):
         """Reset the index by deleting and recreating it."""
--- a/tests/vector_stores/test_mongodb.py
+++ b/tests/vector_stores/test_mongodb.py
@@ -322,9 +322,10 @@ def test_list(mongo_vector_fixture):
 
     mock_collection.find.assert_called_once_with({})
     mock_cursor.limit.assert_called_once_with(2)
-    assert len(results) == 2
-    assert results[0].id == "id1"
-    assert results[0].payload == {"key": "value1"}
+    assert len(results) == 1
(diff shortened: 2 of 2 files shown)


kartik-mem0:
thank you for your contribution closing it in favour of #5729

PR #5693, opened 2026-06-20
fix(embeddings): honor embedding_dims in AzureOpenAIEmbedding

ly-wang19:
## Linked Issue

No separate issue — small self-contained bug fix.

## Description

`AzureOpenAIEmbedding` ignored `embedding_dims`: `embed()` / `embed_batch()` called `embeddings.create()` without the `dimensions` argument, so a user-requested embedding dimension was silently dropped. Azure OpenAI's `text-embedding-3-*` models support `dimensions`, and `OpenAIEmbedding` already passes it (via `_pass_dimensions_to_api`) when `embedding_dims` is set — Azure didn't. So e.g. `embedding_dims=256` produced default-dimension vectors and could mismatch the configured vector-store dimension.

Pass `dimensions` to the API when `embedding_dims` is set, mirroring `OpenAIEmbedding`.

## Type of Change

## Test Coverage

Added `test_embed_passes_dimensions_only_when_explicit`. Verified red→green: it fails on `main` (no `dimensions` sent) and passes with the fix. `pytest tests/embeddings/test_azure_openai_embeddings.py` → 9 passed; `ruff check` clean.

## Checklist

### DIFF
--- a/mem0/embeddings/azure_openai.py
+++ b/mem0/embeddings/azure_openai.py
@@ -56,7 +56,7 @@ class AzureOpenAIEmbedding(EmbeddingBase):
             list: The embedding vector.
         """
         text = text.replace("\n", " ")
-        kwargs = {"input": [text], "model": self.config.model}
+        kwargs = {"input": [text], "model": self.config.model, "encoding_format": "float"}
         if self._pass_dimensions_to_api:
             kwargs["dimensions"] = self.config.embedding_dims
         return self.client.embeddings.create(**kwargs).data[0].embedding
@@ -71,7 +71,7 @@ class AzureOpenAIEmbedding(EmbeddingBase):
         all_embeddings = []
         for i in range(0, len(texts), MAX_BATCH):
             chunk = texts[i : i + MAX_BATCH]
-            kwargs = {"input": chunk, "model": self.config.model}
+            kwargs = {"input": chunk, "model": self.config.model, "encoding_format": "float"}
             if self._pass_dimensions_to_api:
                 kwargs["dimensions"] = self.config.embedding_dims
             response = self.client.embeddings.create(**kwargs)
--- a/mem0/embeddings/azure_openai.py
+++ b/mem0/embeddings/azure_openai.py
@@ -14,6 +14,10 @@ class AzureOpenAIEmbedding(EmbeddingBase):
     def __init__(self, config: Optional[BaseEmbedderConfig] = None):
         super().__init__(config)
 
+        # Only pass `dimensions` to the API when the user set embedding_dims, mirroring
+        # OpenAIEmbedding; otherwise the requested dimension is silently ignored.
+        self._pass_dimensions_to_api = self.config.embedding_dims is not None
+
         api_key = self.config.azure_kwargs.api_key or os.getenv("EMBEDDING_AZURE_OPENAI_API_KEY")
         azure_deployment = self.config.azure_kwargs.azure_deployment or os.getenv("EMBEDDING_AZURE_DEPLOYMENT")
         azure_endpoint = self.config.azure_kwargs.azure_endpoint or os.getenv("EMBEDDING_AZURE_ENDPOINT")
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Review — fix(embeddings): honor embedding_dims in AzureOpenAIEmbedding

**Verdict: Changes Requested** (two small asks before merge)

---

### Strengths

- **Root cause correctly identified and fixed.** `AzureOpenAIEmbedding` was silently dropping `embedding_dims` from both `embed()` and `embed_batch()`. Both call sites are now patched.
- **Pattern is an exact copy of the sibling `OpenAIEmbedding`.** `_pass_dimensions_to_api = self.config.embedding_dims is not None` is the right guard — captures the user's intent at construction time, prevents forwarding `dimensions` to non-matryoshka backends via this Azure class.
- **Cross-package parity complete.** The TS `AzureOpenAIEmbedder` already had this fix (`...(this.embeddingDims !== undefined && { dimensions: this.embeddingDims })`). This PR closes the Python gap.
- **Red-green proof confirmed this run.** `test_embed_passes_dimensions_only_when_explicit` FAILS on `main` (`Expected: create(..., dimensions=256)` / `Actual: create(...)`) and PASSES with the fix. Full 9-test suite passes; ruff clean.

---

### Issues

#### Important — Missing `embed_batch` test for the dimensions path

`embed_batch()` in `azure_openai.py:71–76` is patched identically to `embed()`, but there is no test verifying it actually forwards `dimensions` when `embedding_dims` is set. `embed_batch` is the hot path for bulk memory add/update — a regression there would be invisible until runtime.

Please add a test modeled on `test_embed_passes_dimensions_only_when_explicit` that calls `embed_batch(["a", "b"])` and asserts `create` receives `dimensions=256`. The existing mock fixture makes this straightforward.

#### Important — `encoding_format: "float"` not carried over from `OpenAIEmbedding`

`openai.py:51` and `openai.py:70` unconditionally add `"encoding_format": "float"` to the kwargs dict in both `embed()` and `embed_batch()`. This PR adds the kwargs scaffolding to `azure_openai.py` but omits that key.

The omission predates this PR and is unlikely to break anything on current SDK versions (the `openai` Python client defaults to float). But the PR explicitly mirrors the sibling and has the scaffolding right here — this is the right commit to close it, and it costs one line per method:

```python
kwargs = {"input": [text], "model": self.config.model, "encoding_format": "float"}
```

---

### Recommendations

1. Add `test_embed_batch_passes_dimensions` (similar to `test_embed_passes_dimensions_only_when_explicit`, calling `embed_batch(["a", "b"])`, asserting `dimensions=256` in the mock call).
2. Add `"encoding_format": "float"` to the `kwargs` dicts in `embed()` and `embed_batch()` — one line each.
3. If the above are addressed and the PR is otherwise good to go, please open a short tracking issue for this fix — the PR description says "no separate issue" but having one makes future bisects and changelogs easier.

---

### Evidence

| Check | Result |
|-------|--------|
| CI Gate (Python SDK 3.10/3.11/3.12) | All PASS |
| Red-green (`test_embed_passes_dimensions_only_when_explicit`) | FAIL on main → PASS with fix → PASS restored |
| Full Azure embeddings test suite | 9/9 passed |
| ruff check | All checks passed |
| TS Azure embedder (`azure.ts`) | Already passes `dimensions` — Python side now matches |
| `embed_batch` dimensions test | NOT present — gap confirmed |
| `encoding_format` in `azure_openai.py` (PR head) | NOT present — pre-existing omission |

Full evidence: `.claude/babysit-evidence/pr-5693-2026-06-24.md`

ly-wang19:
Both asks addressed in `e71d33f`:

1. **`encoding_format: "float"` carried through** both `embed()` and `embed_batch()` kwargs — matches `openai.py:51,70` exactly now.

2. **`test_embed_batch_passes_dimensions_only_when_explicit` added** — asserts the bulk path forwards `dimensions=256` to `client.embeddings.create` when `embedding_dims` is set. Red-green proven this run: temporarily removed the `if self._pass_dimensions_to_api:` branch from `embed_batch` → new test FAILS with `Expected: dimensions=256, Actual: no dimensions kwarg`; restored → PASSES. Full 10-test suite green, ruff clean.

Also updated the existing `test_embed_text` and `test_embed_passes_dimensions_only_when_explicit` to assert `encoding_format="float"` is forwarded, since they were silently accepting the omission before.

PR #5695, opened 2026-06-20
fix(memory): include metadata keys in add() scope filters

### LINKED ISSUE #5121: `addMemory` flow excludes `metadata` (e.g. `app_id`) from retrieval and message partitioning, causing cross-context contamination in multi-app setups

HrushiYadav:
## Summary

Fixes #5121

`Memory.add()` was building Phase 1 retrieval filters and SQLite session scope using only entity IDs (`user_id`/`agent_id`/`run_id`), while `Memory.search()` passes the full filter dict including metadata keys like `app_id`. This caused cross-context contamination in multi-app deployments where the same `user_id` across different apps would share the Phase 1 candidate pool and conversation history.

**Root cause:** Three layers dropped metadata keys:
1. `_build_filters_and_metadata` never merged `input_metadata` into `effective_query_filters`
2. Phase 1 `search_filters` in `_add_to_vector_store` (sync + async) explicitly whitelisted only entity IDs
3. `_build_session_scope` hardcoded iteration over `["user_id", "agent_id", "run_id"]`

**Fix:**
- Merge non-entity metadata keys into `effective_query_filters` in `_build_filters_and_metadata` (entity IDs are never overwritten)
- Remove the entity-ID-only whitelist from `search_filters` in both sync and async `_add_to_vector_store`
- Change `_build_session_scope` to iterate all filter keys instead of only the three hardcoded entity IDs

**Behavior change:** Existing deployments that never pass metadata see no difference (entity IDs still flow through identically). Deployments passing metadata keys like `app_id` now get proper scope isolation.

## Test plan

- [x] `_build_filters_and_metadata` includes metadata keys in effective_query_filters
- [x] Entity IDs in filters are not overwritten by metadata
- [x] `None`-valued metadata keys are excluded from filters
- [x] `_build_session_scope` includes metadata keys and is deterministic
- [x] Different `app_id` values produce different session scopes
- [x] `Memory.add()` passes metadata to Phase 1 `vector_store.search()`
- [x] All 408 existing memory tests pass (no regressions)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -362,6 +362,14 @@ def _build_filters_and_metadata(
             suggestion="Please provide at least one identifier to scope the memory operation."
         )
 
+    # ---------- metadata scope keys ----------
+    # Merge non-entity metadata keys into query filters so add() scopes
+    # Phase 1 retrieval and session partitioning the same way search() does.
+    if input_metadata:
+        for k, v in input_metadata.items():
+            if k not in effective_query_filters and v is not None:
+                effective_query_filters[k] = v
+
     # ---------- optional actor filter ----------
     resolved_actor_id = actor_id or effective_query_filters.get("actor_id")
     if resolved_actor_id:
@@ -371,9 +379,9 @@ def _build_filters_and_metadata(
 
 
 def _build_session_scope(filters):
-    """Build deterministic session scope string from entity IDs."""
+    """Build deterministic session scope string from filter keys."""
     parts = []
-    for key in sorted(["user_id", "agent_id", "run_id"]):
+    for key in sorted(filters.keys()):
         val = filters.get(key)
         if val:
             parts.append(f"{key}={val}")
@@ -879,7 +887,7 @@ class Memory(MemoryBase):
         parsed_messages = parse_messages(messages)
 
         # Phase 1: Existing memory retrieval
-        search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
+        search_filters = {k: v for k, v in filters.items() if v}
         query_embedding = self.embedding_model.embed(parsed_messages, "search")
         existing_results = self.vector_store.search(
             query=parsed_messages,
@@ -2516,7 +2524,7 @@ class AsyncMemory(MemoryBase):
         parsed_messages = parse_messages(messages)
 
         # Phase 1: Existing memory retrieval
-        search_filters = {k: v for k, v in effective_filters.items() if k in ("user_id", "agent_id", "run_id") and v}
+        search_filters = {k: v for k, v in effective_filters.items() if v}
         query_embedding = await asyncio.to_thread(self.embedding_model.embed, parsed_messages, "search")
         existing_results = await asyncio.to_thread(
             self.vector_store.search,
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -302,8 +302,9 @@ def _build_filters_and_metadata(
     1. `base_metadata_template`: Used as a template for metadata when storing new memories.
        It includes all provided session identifier(s) and any `input_metadata`.
     2. `effective_query_filters`: Used for querying existing memories. It includes all
-       provided session identifier(s), any `input_filters`, and a resolved actor
-       identifier for targeted filtering if specified by any actor-related inputs.
+       provided session identifier(s), any `input_filters`, any promoted `input_metadata`
+       keys (for scope isolation), and a resolved actor identifier for targeted filtering
+       if specified by any actor-related inputs.
 
     Actor filtering precedence: explicit `actor_id` arg → `filters["actor_id"]`
     This resolved actor ID is used for querying but is not added to `base_metadata_template`,
@@ -888,6 +889,7 @@ class Memory(MemoryBase):
 
         # Phase 1: Existing memory retrieval
         search_filters = {k: v for k, v in filters.items() if v}
+        entity_search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         query_embedding = self.embedding_model.embed(parsed_messages, "search")
         existing_results = self.vector_store.search(
             query=parsed_messages,
@@ -1105,7 +1107,7 @@ class Memory(MemoryBase):
                         queries=valid_texts,
                         vectors_list=valid_vectors,
                         top_k=1,
-                        filters=search_filters,
+                        filters=entity_search_filters,
                     )
 
                     # 7d: Separate into inserts vs updates
@@ -1139,7 +1141,7 @@ class Memory(MemoryBase):
                                 "data": entity_text,
                                 "entity_type": entity_type,
                                 "linked_memory_ids": sorted(memory_ids),
-                                **search_filters,
+                                **entity_search_filters,
                             })
 
                     # 7e: Single batch insert for all new entities
@@ -2525,6 +2527,7 @@ class AsyncMemory(MemoryBase):
 
         # Phase 1: Existing memory retrieval
         search_filters = {k: v for k, v in effective_filters.items() if v}
+        entity_search_filters = {k: v for k, v in effective_filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         query_embedding = await asyncio.to_thread(self.embedding_model.embed, parsed_messages, "search")
         existing_results = await asyncio.to_thread(
             self.vector_store.search,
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- **Root cause diagnosis is precise** — the three-layer asymmetry (`_build_filters_and_metadata`, Phase 1 `search_filters` whitelist, `_build_session_scope`) is exactly right and each is fixed at the correct point.
- **Entity-ID non-override guard** (`if k not in effective_query_filters and v is not None`, `main.py:350-357`) — prevents caller-supplied metadata from clobbering validated entity IDs. Well thought out.
- **`_build_session_scope` uses `sorted(filters.keys())`** — still deterministic, just wider. Correct.
- **Async path covered symmetrically** — both `Memory._add_to_vector_store` (line 806) and `AsyncMemory._add_to_vector_store` (line 2349) get the same fix.
- **Test suite is strong** — 8 focused tests covering key merging, non-override, None-value exclusion, scope determinism, app isolation, and end-to-end Phase 1 filter flow. Red-green verified.

### Issues

**1. `mem0/memory/main.py:806 → 1019, 1051` — entity-store payload regression**

The `search_filters` variable defined at line 806 (now `{k: v for k, v in filters.items() if v}`) is reused verbatim for:
- **Line 1019:** `entity_store.search_batch(filters=search_filters, ...)` — entity dedup search
- **Line 1051:** `**search_filters` spread into new entity payload on insert

After this PR, any metadata key (`app_id`, etc.) flows into the entity-store dedup search filter AND is written into new entity vector payloads. Pre-existing entity records (created before the upgrade) won't have `app_id` in their payload, so the `app_id`-filtered dedup search at line 1019 won't match them → every pre-existing entity is re-created as a duplicate, bloating the entity store.

Fix: define a separate filter for entity-store operations:
```python
# At line 806 (and async equivalent at 2349):
search_filters = {k: v for k, v in filters.items() if v}            # Phase 1 vector store — includes metadata
entity_search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}  # entity store

# Then at lines 1019, 1051 (and async equivalents 2558, 2589): use entity_search_filters
```

**2. `mem0-ts/src/oss/src/memory/index.ts:460-467` — TypeScript `buildSessionScope` gap**

The TS `buildSessionScope()` still hardcodes `["agent_id", "run_id", "user_id"]` — the identical session-scope bug this PR fixes in Python. After this merge, Python and TypeScript SDKs produce different SQLite session scopes for `add()` calls with custom metadata, giving TS users the same cross-app contamination the Python fix removes.

Fix is a one-line change in TS:
```typescript
// line 462 — change:
for (const key of ["agent_id", "run_id", "user_id"].sort()) {
// to:
for (const key of Object.keys(filters).sort()) {
```
(Or open a tracking issue and cite it in the PR body so the gap is explicitly documented.)

#### Minor (optional)
- `mem0/memory/main.py:289-291` — `_build_filters_and_metadata` docstring still describes `effective_query_filters` as holding "session identifier(s), any `input_filters`, and a resolved actor identifier" — no mention of `input_metadata` keys now also being promoted. Worth updating.

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 all pass; CI Gate pass. Vercel fail is advisory (fork deploy perms, not merge-blocking).
- **Targeted tests:** `pytest tests/test_memory.py::TestMetadataScopeFilters -v` → **8 passed in 0.42s**
- **Red-green proof:** impl reverted to `HEAD~1` → 7/8 tests fail; fix restored → 8 passed. Test suite genuinely bites the bug.
- **Linked issue:** #5121 — PR correctly addresses the reported root cause for the Python SDK.
- **Cross-package check:** TS Phase 1 search already passes `filters` directly (no entity-ID whitelist), so that sub-bug is TS-clean. But `buildSessionScope` has the identical session-scope gap — flagged as Important above.
- **Could NOT verify:** entity-store schema impact in a live Qdrant/Chroma instance — verified by code tracing only.

### Assessment
Ready to merge: **With fixes** — please address the two Important items (entity-store `search_filters` split; TS `buildSessionScope`). Both are small and contained. Happy to re-review once pushed.

HrushiYadav:
Pushed fixes for both Important items.

**1. Entity-store filter split** -- added `entity_search_filters` that restricts to scope keys only (`user_id`, `agent_id`, `run_id`). Entity-store `search_batch` (lines 1028, 2566) and entity insert payloads (lines 1061, 2597) now use `entity_search_filters` instead of the full `search_filters`. This prevents metadata keys like `app_id` from leaking into entity dedup search (which would miss pre-existing entities without `app_id` in their payload) and entity insert payloads. Applied to both sync and async paths.

**2. TS buildSessionScope gap** -- opened tracking issue #5812. The Python fix is self-contained; the TS change is a one-liner but belongs in a separate PR.

**3. Docstring update** -- updated `_build_filters_and_metadata` docstring to mention that `input_metadata` keys are now promoted into `effective_query_filters`.

**4. Regression test** -- added `test_entity_store_uses_scope_filters_not_metadata` that verifies entity-store search and insert do not include metadata keys. All 9 scope filter tests pass, ruff clean.

kartik-mem0:
Both items from the prior round are resolved.

**1. Entity-store filter split — fixed** (`main.py:816, 2355`)

The split is clean: `search_filters` (widened, includes metadata keys) now feeds only the Phase 1 vector-store search, while `entity_search_filters` (scope keys only) feeds `entity_store.search_batch` at lines 1029/2567 and the entity insert payload at lines 1061/2598. Both sync and async paths are consistent — all four call sites confirmed.

**2. TS `buildSessionScope` gap — explicitly tracked** (issue #5812, PR #5831)

Opening a tracking issue and a dedicated PR for the TS one-liner is the right call. The Python fix is self-contained; keeping the TS change separate gives it its own CI and review surface.

**3. New regression test** (`test_entity_store_uses_scope_filters_not_metadata`)

Verifies entity-store search and insert do not receive metadata keys. Red-green confirmed locally: 7/9 tests fail against unmodified main, all 9 pass on the PR head, lint clean.

**Evidence**

- CI: Python SDK 3.10 / 3.11 / 3.12 all pass; CI Gate pass; Vercel fail ignored (advisory, fork perms).
- Targeted tests with fix: `9 passed in 0.45s`
- Targeted tests with implementation reverted to main: `7 failed, 2 passed`
- Restored and re-run: `9 passed in 0.45s`
- Ruff: all checks passed on `mem0/memory/main.py`
- `entity_search_filters` confirmed at lines 816, 1029, 1061 (sync) and 2355, 2567, 2598 (async)

This is ready to squash-merge.

HrushiYadav:
@kartik-mem0 both items from your review were resolved on 25 Jun and this is green and mergeable. Could we get it squash-merged when you have a chance? Happy to rebase if main has moved since.

HrushiYadav:
Following up — all requested changes were addressed on Jun 25 and this is clean and mergeable. Would appreciate a re-review when you have a moment @kartik-mem0.

HrushiYadav:
Rebased on latest main. Would love a review when you have bandwidth — this has been open a while. Thanks!

HrushiYadav:
Friendly ping — all review items from @kartik-mem0 were resolved and this has been mergeable since Jun 25. Would appreciate a squash-merge when you get a chance. Happy to rebase if needed.

PR #5703, opened 2026-06-20
fix(chroma): wrap scalar vector_id in list for delete()

### LINKED ISSUE #5698: bug(chroma): delete() passes scalar string to ids parameter, silently deletes nothing
### Component

Vector Store / ChromaDB

### Description

`ChromaDB.delete()` passes `vector_id` (a scalar string) to `self.collection.delete(ids=vector_id)`. ChromaDB's `delete(ids=...)` expects a list of strings. When a plain string is passed, ChromaDB iterates over its characters and tries to delete individual character IDs -- none of which exist -- so the target memory is silently not deleted.

### Steps to Reproduce

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "chroma",
        "config": {"collection_name": "test", "path": "/tmp/chroma_test"},
    }
}
m.add("test memory", user_id="u1")
memories = m.get_all(filters={"user_id": "u1"})
mid = memories["results"][0]["id"]
m.delete(mid)
# Memory still exists after delete
remaining = m.get_all(filters={"user_id": "u1"})
# len(remaining["results"]) == 1  (should be 0)
```

### Expected Behavior

`delete()` removes the memory with the given ID.

### Actual Behavior

`delete()` silently does nothing. The memory persists because ChromaDB receives individual characters of the UUID string as IDs to delete.

### Fix

```python
self.collection.delete(ids=[vector_id])
```

### Environment

- mem0 version: current main
- Python: 3.12
- OS: any

HrushiYadav:
## Summary
- `ChromaDB.delete()` passes the scalar `vector_id` string directly to `collection.delete(ids=vector_id)`. ChromaDB's API expects `ids` to be a list -- when given a bare string, it iterates over the individual characters and matches nothing, so the delete silently does nothing.
- Wraps the ID in a list: `collection.delete(ids=[vector_id])`
- Updates the existing test assertion to verify the correct list-wrapped call

## Test plan
- [x] Existing `test_delete_vector` updated to assert `ids=[vector_id]`
- [x] All 24 ChromaDB tests pass

Closes #5698

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -169,7 +169,7 @@ class ChromaDB(VectorStoreBase):
         Args:
             vector_id (str): ID of the vector to delete.
         """
-        self.collection.delete(ids=vector_id)
+        self.collection.delete(ids=[vector_id])
 
     def update(
         self,
--- a/tests/vector_stores/test_chroma.py
+++ b/tests/vector_stores/test_chroma.py
@@ -122,7 +122,7 @@ def test_delete_vector(chromadb_instance):
 
     chromadb_instance.delete(vector_id=vector_id)
 
-    chromadb_instance.collection.delete.assert_called_once_with(ids=vector_id)
+    chromadb_instance.collection.delete.assert_called_once_with(ids=[vector_id])
 
 
 def test_update_vector(chromadb_instance):


kartik-mem0:
**Review — fix(chroma): wrap scalar vector_id in list for delete()**

**Strengths**

- Root cause is correct: `chromadb.Collection.delete(ids=...)` requires `List[str]`; a bare string causes character-level iteration against UUID IDs, so deletes silently do nothing. Wrapping in `[vector_id]` matches the ChromaDB API contract and is consistent with how `get()` already calls `self.collection.get(ids=[vector_id])` in the same file.
- The test change is a genuine regression test: without the fix, `test_delete_vector` fails with `Expected: delete(ids=['id1']) / Actual: delete(ids='id1')` — the assertion correctly bites the real wrong behavior, not just a mock detail.
- Change is minimal and surgical — one implementation line, one test assertion line. No collateral risk.
- CI Gate passes across Python 3.10, 3.11, 3.12. Ruff lint clean.

**Issues**

- [Minor] `mem0/vector_stores/chroma.py:188` — `update()` passes `ids=vector_id` (scalar) to `collection.update()`, the same pattern as the pre-fix `delete()`. This is a pre-existing sibling bug outside this PR's stated scope; PR #5757 is already open to address it with full test coverage for the `None`-passthrough edge cases. No action required here — noting it for cross-reference.

**Verification**

| Check | Result |
|---|---|
| CI Gate | PASS |
| Python SDK 3.10/3.11/3.12 | PASS |
| Ruff lint (changed files) | Clean |
| `test_delete_vector` WITH fix | PASS |
| `test_delete_vector` WITHOUT fix (reverted chroma.py to pre-PR) | FAIL — `Expected: delete(ids=['id1']) / Actual: delete(ids='id1')` |
| `test_delete_vector` RESTORED fix | PASS |
| Full `test_chroma.py` suite (24 tests) | 24/24 PASS |
| Vercel check | FAIL — advisory, fork perms, not merge-blocking |

**Assessment**

LGTM. The fix is correct, minimal, and red-green verified. The one Minor finding (`update()` at line 188) is an independent pre-existing bug with its own in-flight fix (PR #5757) — it does not affect this PR's correctness or completeness.

Ready to squash-merge (`gh pr merge --squash`).

kartik-mem0 (approved):

PR #5705, opened 2026-06-20
fix(milvus): guard text field in update() with _has_bm25_schema check

### LINKED ISSUE #5700: bug(milvus): update() always includes text field, crashes on pre-v3 collections
### Component

Vector Store / Milvus

### Description

`Milvus.update()` always includes a `"text"` field in the upsert schema, even on collections created before v3 that lack the BM25 text/sparse fields. The `insert()` method correctly guards this with `if self._has_bm25_schema` (line 138), but `update()` unconditionally adds the text field (line 290), causing Milvus to reject the upsert with a schema validation error on legacy collections.

### Steps to Reproduce

1. Create a Milvus collection with the pre-v3 schema (no `text` or `sparse` fields)
2. Insert a vector
3. Call `update()` on that vector
4. Milvus rejects the upsert because `text` is not in the collection schema

### Fix

Guard the `text` field in `update()` the same way `insert()` does:

```python
schema = {"id": vector_id, "vectors": vector, "metadata": payload}
if self._has_bm25_schema:
    schema["text"] = (payload.get("text_lemmatized") or payload.get("data", ""))[:65535] if payload else ""
self.client.upsert(collection_name=self.collection_name, data=schema)
```

### Environment

- mem0 version: current main
- Python: 3.12

HrushiYadav:
## Summary
- `MilvusDB.update()` unconditionally includes a top-level `text` field in the upsert payload. Pre-v3 collections lack the `text`/`sparse` BM25 schema fields, so the upsert is rejected by Milvus.
- `insert()` already correctly guards this with `if self._has_bm25_schema` -- this PR applies the same guard to `update()`.

## Test plan
- [x] `test_update_omits_text_field_on_pre_v3_collection` -- verifies `text` is absent when `_has_bm25_schema` is False
- [x] `test_update_includes_text_field_on_v3_collection` -- verifies `text` is present when `_has_bm25_schema` is True
- [x] All 22 Milvus tests pass

Closes #5700

### DIFF
--- a/mem0/vector_stores/milvus.py
+++ b/mem0/vector_stores/milvus.py
@@ -295,10 +295,12 @@ class MilvusDB(VectorStoreBase):
             if payload is None:
                 payload = existing[0].get("metadata")
 
-        text = ""
-        if payload:
-            text = (payload.get("text_lemmatized") or payload.get("data", ""))[:65535]
-        schema = {"id": vector_id, "vectors": vector, "metadata": payload, "text": text}
+        schema = {"id": vector_id, "vectors": vector, "metadata": payload}
+        if self._has_bm25_schema:
+            text = ""
+            if payload:
+                text = (payload.get("text_lemmatized") or payload.get("data", ""))[:65535]
+            schema["text"] = text
         self.client.upsert(collection_name=self.collection_name, data=schema)
 
     def get(self, vector_id) -> Optional[OutputData]:
--- a/tests/vector_stores/test_milvus.py
+++ b/tests/vector_stores/test_milvus.py
@@ -337,10 +337,44 @@ class TestMilvusDB:
         assert '(metadata["active"] == True)' in result
         assert '(metadata["deleted"] == False)' in result
 
+    def test_update_omits_text_field_on_pre_v3_collection(self, mock_milvus_client):
+        """update() must not include 'text' for collections without BM25 schema."""
+        mock_milvus_client.has_collection.return_value = True
+        mock_milvus_client.describe_collection.return_value = {
+            "fields": [
+                {"name": "id"},
+                {"name": "vectors"},
+                {"name": "metadata"},
+            ]
+        }
+        db = MilvusDB(
+            url="http://localhost:19530",
+            token="test_token",
+            collection_name="legacy_collection",
+            embedding_model_dims=1536,
+            metric_type=MetricType.COSINE,
+            db_name="test_db",
+        )
+        assert db._has_bm25_schema is False
+
+        db.update(vector_id="id1", vector=[0.1] * 1536, payload={"data": "hello"})
+
+        upserted = mock_milvus_client.upsert.call_args[1]["data"]
+        assert "text" not in upserted
+
+    def test_update_includes_text_field_on_v3_collection(self, milvus_db, mock_milvus_client):
+        """update() must include 'text' for collections with BM25 schema."""
+        assert milvus_db._has_bm25_schema is True
+
+        milvus_db.update(vector_id="id1", vector=[0.1] * 1536, payload={"data": "hello"})
+
+        upserted = mock_milvus_client.upsert.call_args[1]["data"]
+        assert upserted["text"] == "hello"
+
     def test_collection_already_exists(self, mock_milvus_client):
         """Test that existing collection is not recreated."""
         mock_milvus_client.has_collection.return_value = True
-        
+
         MilvusDB(
             url="http://localhost:19530",
             token="test_token",
@@ -349,7 +383,7 @@ class TestMilvusDB:
             metric_type=MetricType.L2,
             db_name="test_db"
         )
-        
+
         # create_collection should not be called
         mock_milvus_client.create_collection.assert_not_called()
 


kartik-mem0:
## Review: fix(milvus): guard text field in update() with _has_bm25_schema

Thanks for the thorough fix, @HrushiYadav — the code is solid. One thing needs to be resolved before this can land: this PR currently has a merge conflict with main. Please rebase on the latest main and push — no code changes should be needed, just the rebase.

---

### Strengths

- **Root-cause is right.** `insert()` already guards the `text` field behind `if self._has_bm25_schema`; `update()` was missing the same guard. The fix makes them perfectly symmetric — the correct pattern, applied in the correct place.
- **Implementation is clean.** Schema dict is built without `text`, then `text` is added conditionally — identical to the `insert()` approach. No extra branching, no new state.
- **Tests bite the bug.** `test_update_omits_text_field_on_pre_v3_collection` reproduces the actual failure mode: the `text` key must be absent from the upsert payload when `_has_bm25_schema` is False. Red-green verified locally: 22/22 pass with the fix; the regression test fails when the implementation is reverted.
- **All paths covered.** Both branches of the guard are tested (pre-v3 exclusion + v3 inclusion). All 22 existing Milvus tests still pass.

---

### Issues

**None — code quality is merge-ready.** One very minor note for awareness (no action needed):

- **Minor / Informational:** `test_update_includes_text_field_on_v3_collection` passes against the *reverted* implementation too (the old code always wrote `text`, so the v3 positive case never catches the regression). Its job is forward-guard only — protecting against someone accidentally removing the `text` write in a future refactor — which is fine. The pre-v3 test is the one doing the discriminating work, and it does it cleanly.

---

### Evidence

- **CI (real checks):** Python SDK / build_mem0 3.10 ✓, 3.11 ✓, 3.12 ✓ — CI Gate ✓. Vercel check is a deploy-preview permission issue on fork PRs — not a required check, ignored per project policy.
- **Local targeted tests (Python 3.12):**
  - With fix: `22 passed in 1.60s`
  - After reverting impl (keeping tests): `1 failed, 21 passed` — `test_update_omits_text_field_on_pre_v3_collection` fails as expected
  - After restoring fix: `22 passed` — red-green confirmed
- **Ruff lint:** `All checks passed!` on both changed files
- **Duplicate cluster:** No other open PR addresses this issue
- **Merge status:** `CONFLICTING` — rebase on main required before merge

---

### Assessment

The fix is correct, architecturally sound, and well-tested. Once rebased on main this is ready to squash-merge (`gh pr merge --squash`). Please push the rebased branch and the review loop will confirm the conflict is resolved.

kartik-mem0:
Re-review against 21309b78 (rebased head): rebase is clean, the `_has_bm25_schema` guard in `update()` is intact, and all tests pass.

**LGTM — ready for squash-merge.**

Changes since the prior review (head `7eda772b`): rebase onto main only. The diff is substantively identical to what was red-green proven in the first pass; no logic drift introduced.

### Evidence
- **CI:** CI Gate pass; Python SDK / build_mem0 (3.10, 3.11, 3.12) all pass; `license/cla` pass. Vercel fail is fork-deploy advisory — ignored per policy.
- **Mergeable:** `MERGEABLE` (was `CONFLICTING` — rebase blocker resolved).
- **Targeted tests (fresh run):** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_milvus.py -v` → **27 passed in 1.30s** (5 additional tests exist on main since prior review; all 27 pass, including `test_update_omits_text_field_on_pre_v3_collection` and `test_update_includes_text_field_on_v3_collection`).
- **Lint:** `ruff check mem0/vector_stores/milvus.py tests/vector_stores/test_milvus.py` → All checks passed.
- **Guard verified:** `milvus.py:295–306` — `schema["text"]` is only assigned inside `if self._has_bm25_schema:`, symmetric with `insert()`. Correct.
- **Could NOT verify:** live Milvus pre-v3 collection behavior — code tracing only; CI matrix confirms mock coverage.

Thanks for the quick rebase, @HrushiYadav.

mem0/vector_stores/milvus.py
```diff
@@ -295,10 +295,12 @@ def update(self, vector_id=None, vector=None, payload=None):
             if payload is None:
                 payload = existing[0].get("metadata")
 
-        text = ""
-        if payload:
-            text = (payload.get("text_lemmatized") or payload.get("data", ""))[:65535]
-        schema = {"id": vector_id, "vectors": vector, "metadata": payload, "text": text}
+        schema = {"id": vector_id, "vectors": vector, "metadata": payload}
+        if self._has_bm25_schema:
```
kartik-mem0, line 299: Optional: `insert()` carries a short comment above its identical guard explaining that pre-v3 collections reject unknown top-level fields. A one-liner here would keep the two visibly paired for the next reader. Not blocking either way.

kartik-mem0 (commented):

`MilvusDB.update()` wrote a top-level `text` field unconditionally while `insert()` already guarded it, so any update against a pre-v3 collection was rejected by Milvus. This applies the same guard, making the two write paths symmetric. Still good at `21309b78`.

**Before you merge:** nothing, this is self-contained and the v3 path is byte-identical to before.
**Follow-up:** this turned out to be a 3-way duplicate (#6401, #6450). Take this one and close those two: #6450 cannot merge (CLA unsigned, so CI has never run on it), and #6401 is a fair equivalent but arrived a month later and its test hand-sets `_has_bm25_schema` rather than deriving it.
**Not verified:** nothing, the fix was exercised against a real Milvus server as well as the unit suite.

Risk: low, one provider, gated on a flag that already guards the same field in `insert()` · Your time: ~2 min

- **Problem is real:** `mem0/vector_stores/milvus.py:304` on `main` writes `text` unconditionally, while `insert()` at `:141` does not. Triggers on any collection created before the v3 hybrid schema (no `text`/`sparse` fields, dynamic fields off).
- **Live Milvus:** RED on `main` @`ca2abca2` with `DataNotMatchException: Attempt to insert an unexpected field 'text' to collection without enabling dynamic field`, GREEN on this head against a rebuilt pre-v3 collection.
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_milvus.py -v --tb=short` -> 27 passed
- **Red-green:** reverted only `milvus.py` to `main` and kept the new tests -> `test_update_omits_text_field_on_pre_v3_collection` fails on `assert 'text' not in {..., 'text': 'hello'}` (1 failed, 26 passed), restored -> 27 passed. The test derives `_has_bm25_schema` through the real `describe_collection` path, so it covers the detection logic too.
- **CI:** all 7 real checks green (CI Gate, build_mem0 3.10/3.11/3.12, changelog_check, check_changes). `Vercel` is red on every fork PR (no deploy permissions) and does not block. CLA signed.
- **Security:** cleared, no new input path, the same `payload` reaches the same parameterized `upsert`, and legacy collections now receive one field fewer.
- **Cross-package:** `mem0-ts/src/oss/src/vector_stores/milvus.ts:286,391` already guards both paths, so this closes an existing Python/TS drift rather than opening one.
- **Bloat (ponytail):** lean already, the guard mirrors `insert()`'s existing shape.
- **Linked issue:** #5700

</details>

Nice catch on the asymmetry, @HrushiYadav, and sorry this one sat as long as it did.

kartik-mem0 (approved):

PR #5707, opened 2026-06-20
fix(pinecone): map all comparison operators in _create_filter()

### LINKED ISSUE #5702: bug(pinecone): _create_filter() wraps non-range operators as $eq, producing wrong filters
### Component

Vector Store / Pinecone

### Description

`Pinecone._create_filter()` only handles the combined `gte`+`lte` range case. All other operator dicts (`{"gt": 5}`, `{"ne": "x"}`, `{"in": [...]}`) fall through to the `else` branch which wraps the entire dict as `{"$eq": value}`. This passes a raw dict as an equality value, which Pinecone silently ignores or rejects.

### Steps to Reproduce

```python
# Filter with gt operator
results = memory.search("test", filters={"user_id": "u1", "score": {"gt": 5}})
# Produces Pinecone filter: {"score": {"$eq": {"gt": 5}}}
# Returns wrong results or no results
```

### Fix

Map each operator properly to Pinecone's filter syntax:

```python
OPERATOR_MAP = {"gt": "$gt", "gte": "$gte", "lt": "$lt", "lte": "$lte",
                "ne": "$ne", "in": "$in", "nin": "$nin", "eq": "$eq"}

if isinstance(value, dict):
    pinecone_filter[key] = {OPERATOR_MAP.get(op, f"${op}"): v for op, v in value.items()}
else:
    pinecone_filter[key] = {"$eq": value}
```

### Environment

- mem0 version: current main
- Python: 3.12

HrushiYadav:
## Summary
- `_create_filter()` only handled the `gte+lte` range case as a special dict path. Any other operator dict (`gt`, `lt`, `ne`, `in`, `nin`, etc.) fell through to the `else` branch and was wrapped as `$eq`, producing invalid Pinecone filters like `{"$eq": {"gt": 5}}` instead of `{"$gt": 5}`.
- Added an `OPERATOR_MAP` that maps all mem0 internal operators (`eq`, `ne`, `gt`, `gte`, `lt`, `lte`, `in`, `nin`) to their Pinecone `$`-prefixed equivalents.
- When `value` is a dict, each key is now individually mapped through `OPERATOR_MAP`.

## Test plan
- [x] `test_create_filter_plain_value` -- plain string produces `$eq`
- [x] `test_create_filter_range` -- `gte+lte` dict produces `$gte+$lte`
- [x] `test_create_filter_gt_operator` -- `gt` operator correctly mapped
- [x] `test_create_filter_in_operator` -- `in` operator correctly mapped
- [x] `test_create_filter_ne_operator` -- `ne` operator correctly mapped
- [x] All 20 Pinecone tests pass

Closes #5702

### DIFF
--- a/mem0/vector_stores/pinecone.py
+++ b/mem0/vector_stores/pinecone.py
@@ -186,6 +186,17 @@ class PineconeDB(VectorStoreBase):
 
             return result
 
+    OPERATOR_MAP = {
+        "eq": "$eq",
+        "ne": "$ne",
+        "gt": "$gt",
+        "gte": "$gte",
+        "lt": "$lt",
+        "lte": "$lte",
+        "in": "$in",
+        "nin": "$nin",
+    }
+
     def _create_filter(self, filters: Optional[Dict]) -> Dict:
         """
         Create a filter dictionary from the provided filters.
@@ -196,8 +207,15 @@ class PineconeDB(VectorStoreBase):
         pinecone_filter = {}
 
         for key, value in filters.items():
-            if isinstance(value, dict) and "gte" in value and "lte" in value:
-                pinecone_filter[key] = {"$gte": value["gte"], "$lte": value["lte"]}
+            if isinstance(value, dict):
+                condition = {}
+                for op, operand in value.items():
+                    pc_op = self.OPERATOR_MAP.get(op)
+                    if pc_op:
+                        condition[pc_op] = operand
+                    else:
+                        condition[f"${op}"] = operand
+                pinecone_filter[key] = condition
             else:
                 pinecone_filter[key] = {"$eq": value}
 
--- a/tests/vector_stores/test_pinecone.py
+++ b/tests/vector_stores/test_pinecone.py
@@ -196,3 +196,28 @@ def test_list_error_returns_list_not_dict(pinecone_db):
     result = pinecone_db.list(filters={"user_id": "alice"}, top_k=10)
     assert isinstance(result, list)
     assert result == [[]]
+
+
+def test_create_filter_plain_value(pinecone_db):
+    result = pinecone_db._create_filter({"user_id": "alice"})
+    assert result == {"user_id": {"$eq": "alice"}}
+
+
+def test_create_filter_range(pinecone_db):
+    result = pinecone_db._create_filter({"age": {"gte": 18, "lte": 65}})
+    assert result == {"age": {"$gte": 18, "$lte": 65}}
+
+
+def test_create_filter_gt_operator(pinecone_db):
+    result = pinecone_db._create_filter({"score": {"gt": 0.5}})
+    assert result == {"score": {"$gt": 0.5}}
+
+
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Fixes `PineconeDB._create_filter` which only handled `gte`+`lte` combined ranges — every other operator dict (`gt`, `lt`, `ne`, `in`, `nin`, single-bound ranges) fell through to the `else` branch and got wrapped as `{"$eq": value}`, causing Pinecone to silently return wrong or empty results. The fix adds an `OPERATOR_MAP` that translates all 8 mem0-universal operators to their Pinecone `$`-prefixed equivalents, bringing this provider to parity with the ChromaDB and Qdrant siblings.

### Strengths
- `mem0/vector_stores/pinecone.py:189–220` — clean, minimal change at the right layer; only the adapter boundary between mem0's universal filter format and Pinecone's syntax was touched
- The `f"${op}"` fallback for unknown operators (e.g. `contains`) is the correct call: Pinecone will return an explicit API validation error rather than silently returning wrong results via an `$eq` miscast
- Preserves the existing `gte`+`lte` closed-range behaviour as a regression test (`test_create_filter_range`)
- 5 regression tests, each exercising a distinct operator path; red-green proof is solid (see Evidence)

### Issues
Minor (optional): `OPERATOR_MAP` at class scope (`pinecone.py:189`) is slightly unusual for a constant only used by one method — a module-level constant or a local would be marginally cleaner. Not a blocker.

### Overlapping PR: #5739
PR #5739 fixes the same root cause with a nearly identical approach but is currently CONFLICTING (cannot merge). The test in #5739 consolidates all operator assertions into a single function with a better regression docstring — that's a nice pattern. If #5739 can be rebased cleanly after this merges it could add that docstring improvement, but there's no need to hold #5707 for it.

### Evidence
- **CI:** Python SDK builds pass on 3.10, 3.11, 3.12. CI Gate pass. Vercel/deploy check fails (advisory, fork perms — ignored).
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_pinecone.py -v --tb=short` → **21 passed in 4.74s**
- **Red-green proof:** Reverted `mem0/vector_stores/pinecone.py` to `main`, kept PR's tests → `test_create_filter_gt_operator`, `test_create_filter_in_operator`, `test_create_filter_ne_operator` all FAIL with the exact wrong-output diff. Restored PR code → 3 passed. Tests prove the fix.
- **Lint:** `ruff check mem0/vector_stores/pinecone.py tests/vector_stores/test_pinecone.py` → clean
- **Linked issue:** #5702 — PR directly addresses the reported root cause
- **Cross-package check:** No Pinecone TS provider in this repo; TS unaffected
- **Could NOT verify:** Live Pinecone API behaviour — tested via mocks throughout; translation logic is straightforward enough that code tracing is sufficient

### Assessment
Root cause fixed at the right layer, regression tests bite the actual bug, CI green, no conflicts.

kartik-mem0 (approved):

PR #5709, opened 2026-06-20
fix(redis): use .get() for hash/created_at in insert() to handle entity payloads

### LINKED ISSUE #5708: bug(redis): insert() crashes with KeyError on entity store payloads
## Description

`RedisDB.insert()` accesses `payload[\"hash\"]` and `payload[\"created_at\"]` directly (lines 127-129 of `mem0/vector_stores/redis.py`). When Redis is used as the entity store, entity payloads only contain \`data\`, \`entity_type\`, and \`linked_memory_ids\` -- they don't include \`hash\` or \`created_at\`, causing a \`KeyError\`.

The entity store reuses the same vector store provider as the main memory store (line 498 of main.py), so Redis users who enable entity extraction will hit this crash.

## Steps to reproduce

2. Enable entity extraction (the default)
3. Add a memory: \`m.add(\"I work at OpenAI\", user_id=\"test\")\`
4. Entity upsert calls \`entity_store.insert()\` with payload \`{\"data\": \"OpenAI\", \"entity_type\": \"organization\", ...}\`
5. \`payload[\"hash\"]\` raises \`KeyError\`

## Expected behavior

\`insert()\` should use \`.get()\` with sensible defaults for optional fields, making it work with both memory and entity payloads.

HrushiYadav:
## Summary
- `RedisDB.insert()` accesses `payload["hash"]` and `payload["created_at"]` directly. When Redis is used as the entity store (which reuses the same vector store provider), entity payloads only contain `data`, `entity_type`, and `linked_memory_ids` -- they lack `hash` and `created_at`, causing `KeyError`.
- Changed to `.get()` with sensible defaults: empty string for `hash`, `0` timestamp for missing `created_at`.

## Test plan
- [x] `test_insert_entity_payload_without_hash_and_created_at` -- verifies insert with entity-style payload works
- [x] All 8 Redis tests pass (including existing memory-style payload tests)

Closes #5708

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -244,12 +244,16 @@ class RedisDB(VectorStoreBase):
         self.index.drop_keys(f"{self.schema['index']['prefix']}:{vector_id}")
 
     def update(self, vector_id=None, vector=None, payload=None):
+        created_at_str = payload.get("created_at")
+        created_at_ts = int(datetime.fromisoformat(created_at_str).timestamp()) if created_at_str else 0
+        updated_at_str = payload.get("updated_at")
+        updated_at_ts = int(datetime.fromisoformat(updated_at_str).timestamp()) if updated_at_str else 0
         data = {
             "memory_id": vector_id,
-            "hash": payload["hash"],
-            "memory": payload["data"],
-            "created_at": int(datetime.fromisoformat(payload["created_at"]).timestamp()),
-            "updated_at": int(datetime.fromisoformat(payload["updated_at"]).timestamp()),
+            "hash": payload.get("hash", ""),
+            "memory": payload.get("data", ""),
+            "created_at": created_at_ts,
+            "updated_at": updated_at_ts,
         }
 
         # Only update embedding if vector is provided
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -122,11 +122,13 @@ class RedisDB(VectorStoreBase):
         data = []
         for vector, payload, id in zip(vectors, payloads, ids):
             # Start with required fields
+            created_at_str = payload.get("created_at")
+            created_at_ts = int(datetime.fromisoformat(created_at_str).timestamp()) if created_at_str else 0
             entry = {
                 "memory_id": id,
-                "hash": payload["hash"],
-                "memory": payload["data"],
-                "created_at": int(datetime.fromisoformat(payload["created_at"]).timestamp()),
+                "hash": payload.get("hash", ""),
+                "memory": payload.get("data", ""),
+                "created_at": created_at_ts,
                 "embedding": np.array(vector, dtype=np.float32).tobytes(),
             }
 
--- a/tests/vector_stores/test_redis.py
+++ b/tests/vector_stores/test_redis.py
@@ -153,3 +153,26 @@ def test_insert_entity_payload_without_hash_and_created_at():
     assert data[0]["memory"] == "OpenAI"
     assert data[0]["hash"] == ""
     assert data[0]["created_at"] == 0
+
+
+def test_update_entity_payload_without_hash_and_timestamps():
+    """update() must not crash on entity payloads that lack hash/created_at/updated_at."""
+    db, mock_index = _make_redis_db()
+
+    entity_payload = {
+        "data": "OpenAI",
+        "entity_type": "organization",
+        "linked_memory_ids": ["mem-1"],
+        "user_id": "test_user",
+    }
+
+    db.update(vector_id="entity-1", vector=[0.1, 0.2, 0.3], payload=entity_payload)
+
+    mock_index.load.assert_called_once()
+    call_kwargs = mock_index.load.call_args
+    data_dict = call_kwargs[1]["data"][0] if "data" in call_kwargs[1] else call_kwargs[0][0][0]
+    assert data_dict["memory_id"] == "entity-1"
+    assert data_dict["memory"] == "OpenAI"
+    assert data_dict["hash"] == ""
+    assert data_dict["created_at"] == 0
+    assert data_dict["updated_at"] == 0
--- a/tests/vector_stores/test_redis.py
+++ b/tests/vector_stores/test_redis.py
@@ -128,3 +128,28 @@ def test_get_returns_none_for_missing_id():
 
     assert db.get("missing_id") is None
     mock_index.fetch.assert_called_once_with("missing_id")
+
+
+def test_insert_entity_payload_without_hash_and_created_at():
+    """insert() must not crash on entity payloads that lack hash/created_at."""
+    db, mock_index = _make_redis_db()
+
+    entity_payload = {
+        "data": "OpenAI",
+        "entity_type": "organization",
+        "linked_memory_ids": ["mem-1"],
+        "user_id": "test_user",
+    }
+
+    db.insert(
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

Thanks for tracking down and fixing this — the root cause is correct and the insert() fix is solid.

### Strengths

- The `KeyError` path is confirmed: entity payloads from `main.py` (lines 531-535, 1047-1052) contain only `data`, `entity_type`, `linked_memory_ids`, and scope filters — no `hash` or `created_at`. The direct `payload["hash"]` at the old line 129 always crashed on the entity path.
- `.get()` with sentinel defaults is the right pattern for a generic `VectorStoreBase.insert()` — matches exactly what `valkey.py` and other stores do.
- Red-green proven locally: test FAILS (`KeyError: 'hash'` at redis.py:129) on main, PASSES with the fix. All 8 tests pass, ruff clean.
- CI Gate: pass (3.10 / 3.11 / 3.12).

---

### Issues

#### Important — `update()` has the same direct-access bug and will crash on the entity update path

`redis.py:246-264` — `update()` does:

```python
"hash": payload["hash"],
"memory": payload["data"],
"created_at": int(datetime.fromisoformat(payload["created_at"]).timestamp()),
"updated_at": int(datetime.fromisoformat(payload["updated_at"]).timestamp()),   # line 252
```

When `entity_store.update()` is called (`main.py:523-527`), the payload comes from `match.payload` — which is built by `search()`. `search()` only includes `updated_at` if `"updated_at" in result` (redis.py:175-182). Entity records stored via the fixed `insert()` never set `updated_at`, so `match.payload` will not carry that key. `update()` line 252 will raise `KeyError: 'updated_at'`.

Before this fix, `insert()` crashed before any entity record could be written, so the update path was never reachable. This fix makes insertion work — which means the update path is now reachable and will crash on first use.

**Fix:** apply the same `.get()` pattern to `update()`:

```python
def update(self, vector_id=None, vector=None, payload=None):
    created_at_str = payload.get("created_at")
    updated_at_str = payload.get("updated_at")
    data = {
        "memory_id": vector_id,
        "hash": payload.get("hash", ""),
        "memory": payload.get("data", ""),
        "created_at": int(datetime.fromisoformat(created_at_str).timestamp()) if created_at_str else 0,
        "updated_at": int(datetime.fromisoformat(updated_at_str).timestamp()) if updated_at_str else 0,
    }
```

#### Important — TS SDK has the same bug (separate follow-up)

`mem0-ts/src/oss/src/vector_stores/redis.ts:342-344`:

```typescript
hash: payload.hash,
memory: payload.data,
created_at: new Date(payload.created_at).getTime(),
```

`new Date(undefined).getTime()` returns `NaN` — which will be stored in the Redis numeric field and break sorting. Please open a follow-up issue or PR for the TS side; it doesn't need to block this PR as long as the Python `update()` is fixed here.

#### Minor — `created_at=0` (Unix epoch) sorts entity records to the bottom

`list()` sorts by `created_at DESC`. Entity records with `created_at=0` will always appear after every real memory record. Using `int(datetime.now(timezone.utc).timestamp())` as the default is more semantically correct — `timezone` is already imported at line 4. Not a crash, but worth addressing.

---

### What's needed before merge

1. Add `.get()` guards to `update()` for `hash`, `data`, `created_at`, and `updated_at` (same pattern as this PR's insert fix)
2. A regression test verifying `entity_store.update()` doesn't crash when called on an entity payload that has no `updated_at`
3. Optional: use `datetime.now(timezone.utc)` instead of `0` for the `created_at` default

---

### Evidence

| Check | Result |
|---|---|
| CI Gate (3.10 / 3.11 / 3.12) | PASS |
| Red-green: test FAILS on main | `KeyError: 'hash'` at redis.py:129 ✓ |
| Red-green: test PASSES with fix | 8/8 pass ✓ |
| Ruff | clean ✓ |
| `update()` crash path verified | `redis.py:252` direct `payload["updated_at"]`, not guarded ✓ |
| TS parity gap verified | `redis.ts:342-344` direct property access ✓ |

HrushiYadav:
Pushed the update() fix.

**1. update() .get() guards** -- applied the same safe-access pattern from insert() to update(). hash, data, created_at, and updated_at all use .get() with defaults now. Entity payloads without these keys no longer crash.

**2. Regression test** -- added test_update_entity_payload_without_hash_and_timestamps that exercises update() with an entity payload lacking hash/created_at/updated_at. All 9 tests pass, ruff clean.

**Re: created_at=0 default** -- kept 0 for now to match the insert() behavior. Happy to change both to datetime.now(timezone.utc) if preferred, but that would be a separate concern.

**Re: TS SDK** -- noted. Will open a follow-up issue for redis.ts:342-344.

kartik-mem0:
Verified against c2541e68: both changes requested are addressed.

**`update()` KeyError** — fixed. Lines 247–250 now apply the same `.get()` + conditional timestamp pattern as `insert()`. `hash`, `data`, `created_at`, and `updated_at` are all guarded. `test_update_entity_payload_without_hash_and_timestamps` red-greens cleanly: KeyError on the old `update()` impl, passes on the new one.

**`insert()` coverage** — the original fix is intact and all 9 Redis tests pass.

### What this PR does
Replaces direct `payload["key"]` access in `RedisDB.insert()` and `update()` with `.get()` guards so entity store payloads (which lack `hash`, `created_at`, `updated_at`) no longer crash with `KeyError`.

### Strengths
- `mem0/vector_stores/redis.py:125–126, 247–250` — the two-line pattern (`get()` → conditional fromisoformat) is clean and matches `valkey.py` exactly.
- Two regression tests that prove both code paths — each fails on the old impl and passes on the new one.

### Issues
Minor (optional): `mem0-ts/src/oss/src/vector_stores/redis.ts:342–344` has the same direct-access pattern (`payload.hash`, `payload.data`, `new Date(payload.created_at).getTime()`). Worth a follow-up issue/PR to keep the SDKs in sync — not a blocker for this fix.

### Evidence
- **CI:** CI Gate pass; Python SDK build_mem0 (3.10, 3.11, 3.12) pass; Vercel fail ignored (advisory, fork perms).
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_redis.py -v` → 9 passed in 0.83s.
- **Red-green proof (insert):** confirmed in prior review at SHA 4d3f0cb8.
- **Red-green proof (update):** reverted `redis.py` to SHA 4d3f0cb8 → `test_update_entity_payload_without_hash_and_timestamps` FAILED (KeyError: 'hash' at redis.py:249); restored PR head → PASSED.
- **Lint:** `ruff check mem0/vector_stores/redis.py tests/vector_stores/test_redis.py` → All checks passed.
- **TS counterpart:** `mem0-ts/src/oss/src/vector_stores/redis.ts:342–344` not fixed here — follow-up warranted.
- **Could NOT verify:** live Redis behavior — code tracing only.

### Assessment
Both crash paths are guarded, tests bite, lint clean, CI green.

kartik-mem0 (approved):

kartik-mem0:
hey @HrushiYadav,

Minor (optional): mem0-ts/src/oss/src/vector_stores/redis.ts:342–344 has the same direct-access pattern (payload.hash, payload.data, new Date(payload.created_at).getTime()). Worth a follow-up issue/PR to keep the SDKs in sync — not a blocker for this fix.


can you please also do this as well?

HrushiYadav:
Opened #5859 for the TS Redis follow-up (redis.ts:342-344 direct property access).

PR #5719, opened 2026-06-21
fix(memory): configure spaCy model for retrieval

### LINKED ISSUE #4884: BM25 keyword search and entity extraction are hardcoded to English
### Component

Core / Python SDK

### Use Case

The v3 hybrid search pipeline (merged via #4805, #4836, #4858) combines semantic similarity, BM25 keyword matching, and entity-based boosting. The BM25 and entity paths both depend on spaCy's English model `en_core_web_sm`, which is hardcoded in two places:

- `mem0/utils/spacy_models.py:60` — `spacy.load("en_core_web_sm")` for entity extraction
- `mem0/utils/spacy_models.py:85` — `spacy.load("en_core_web_sm", disable=["ner", "parser"])` for lemmatization
- `mem0/utils/lemmatization.py:30-50` runs this English lemmatizer on every query and on every stored memory's `text_lemmatized` field

For non-English deployments (Chinese, Japanese, Korean, Arabic, Thai, Hindi, etc.) the practical consequences are:

1. **BM25 silently becomes a no-op.** spaCy's English pipeline does not tokenize CJK / Arabic / Thai scripts into meaningful units, so query-side lemmas don't overlap with stored-side lemmas. `bm25_scores` comes back empty at `main.py:1359-1366`. `scoring.py:85` then sets `has_bm25 = False` and the combined-score divisor stays at 1.0 — results aren't mis-ranked, but the BM25 signal is entirely lost.
2. **Entity boost also disappears.** `extract_entities` uses the same English NER model, so non-English proper nouns, IDs, and numbers receive no boost at `main.py:1369-1371`.

Net effect: for most of the world, the v3 hybrid pipeline degrades to a semantic-only pipeline. Users see no error — they just silently miss the precision improvements the feature is supposed to deliver (e.g. exact-match on a Chinese employee ID, a Japanese name, or a Hindi place name).

This is separate from the *storage-side* issue where extracted memories are translated into English (covered by #3206, #3707, and PR #4883). Even with memories correctly stored in the source language, the retrieval side still lemmatizes them with `en_core_web_sm`.

### Proposed Solution

Raising this for maintainer awareness and discussion — I don't want to prescribe a specific implementation (language dispatch, character n-grams, multi-model spaCy, vector-store delegation, etc. all have trade-offs worth debating separately). The goal of this issue is to surface the gap so it can be prioritized and scoped.

### Alternatives Considered

Users can currently work around parts of this by:
- Disabling BM25/entity contributions and relying on semantic similarity alone.
- Overriding `lemmatize_for_bm25` and `extract_entities` locally.

Neither is documented, and both leave the default experience broken for non-English users.

### Additional Context

- v3 hybrid pipeline PRs: #4805, #4836, #4858 (merged April 2026)
- Storage-side companion: PR #4883, issues #3206, #3707
- Key file references:
  - `mem0/utils/spacy_models.py`
  - `mem0/utils/lemmatization.py`
  - `mem0/memory/main.py:1341, 1591, 1667, 2210, 2751, 3088`

rodboev:
## Linked Issue

Closes #4884

## Description

BM25 preprocessing and entity extraction currently always load `en_core_web_sm`, so non-English deployments cannot use a language-specific spaCy model even when one is installed. This PR adds a `spacy_model` field to `MemoryConfig` (default: `"en_core_web_sm"`), promotes the module-level model cache to dicts keyed by model name, and threads the configured model through the shared spaCy loaders used by BM25 and entity extraction.

All new function parameters default to `"en_core_web_sm"`, preserving full backward compatibility.

## Root cause

`mem0/utils/spacy_models.py` called `spacy.load("en_core_web_sm")` at lines 60 and 85 with no configuration path. The loader cache was global rather than keyed by model name, and the `Memory` / `AsyncMemory` call paths did not pass any configured model.

Related: PR #5289 covers the same issue via an env var approach but has been stale 23+ days.

## Scope

- `mem0/configs/base.py` — new `spacy_model: str` field
- `mem0/utils/spacy_models.py` — cache promoted to dicts; loaders accept `model_name` param
- `mem0/utils/entity_extraction.py` — functions accept and forward `spacy_model` kwarg
- `mem0/utils/lemmatization.py` — `lemmatize_for_bm25` accepts and forwards `spacy_model` kwarg
- `mem0/memory/main.py` — passes `spacy_model=self.config.spacy_model` at all call sites
- `tests/utils/test_spacy_model_config.py` — new tests (fully mocked, no model download)

No language detection, BM25 scoring changes, or new dependencies.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Verification:
- [x] `python -m pytest tests/utils/test_spacy_model_config.py -v` — 5 passed
- [ ] `make lint && make test`

## Checklist

Claude Opus 4.6 via Claude Code CLI

### DIFF
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -59,7 +59,10 @@ class MemoryConfig(BaseModel):
         description=(
             "spaCy model name for entity extraction and BM25 lemmatization. "
             "The model must be installed before use: `python -m spacy download <model_name>`. "
-            "If not installed, Mem0 will attempt to download it automatically on first use."
+            "If not installed, Mem0 will attempt to download it automatically on first use. "
+            "Supports Latin-script languages (e.g. German, French, Spanish) with appropriate models. "
+            "CJK, Arabic, and Thai tokenization quality depends on the chosen model; BM25 "
+            "lemmatization uses English-centric morphology and is best suited for Latin scripts."
         ),
         default="en_core_web_sm",
     )
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -56,7 +56,11 @@ class MemoryConfig(BaseModel):
         default=None,
     )
     spacy_model: str = Field(
-        description="spaCy model name for entity extraction and BM25 lemmatization",
+        description=(
+            "spaCy model name for entity extraction and BM25 lemmatization. "
+            "The model must be installed before use: `python -m spacy download <model_name>`. "
+            "If not installed, Mem0 will attempt to download it automatically on first use."
+        ),
         default="en_core_web_sm",
     )
 
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -55,6 +55,10 @@ class MemoryConfig(BaseModel):
         description="Custom instructions for fact extraction",
         default=None,
     )
+    spacy_model: str = Field(
+        description="spaCy model name for entity extraction and BM25 lemmatization",
+        default="en_core_web_sm",
+    )
 
 
 class AzureConfig(BaseModel):
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -2303,7 +2303,7 @@ class AsyncMemory(MemoryBase):
     async def _link_entities_for_memory(self, memory_id, text, filters):
         """Async variant of `Memory._link_entities_for_memory`."""
         try:
-            entities = await asyncio.to_thread(extract_entities, text, self.config.spacy_model)
+            entities = await asyncio.to_thread(lambda: extract_entities(text, spacy_model=self.config.spacy_model))
             if not entities:
                 return
             seen = set()
@@ -2656,7 +2656,7 @@ class AsyncMemory(MemoryBase):
         # Phase 7: Batch entity linking
         try:
             all_texts = [r[1] for r in records]
-            all_entities = await asyncio.to_thread(extract_entities_batch, all_texts, 32, self.config.spacy_model)
+            all_entities = await asyncio.to_thread(lambda: extract_entities_batch(all_texts, 32, spacy_model=self.config.spacy_model))
 
             # 7a: Global dedup
             global_entities = {}
@@ -3205,8 +3205,8 @@ class AsyncMemory(MemoryBase):
             threshold = 0.1
 
         # Step 1: Preprocess query (CPU-bound)
-        query_lemmatized = await asyncio.to_thread(lemmatize_for_bm25, query, self.config.spacy_model)
-        query_entities = await asyncio.to_thread(extract_entities, query, self.config.spacy_model)
+        query_lemmatized = await asyncio.to_thread(lambda: lemmatize_for_bm25(query, spacy_model=self.config.spacy_model))
+        query_entities = await asyncio.to_thread(lambda: extract_entities(query, spacy_model=self.config.spacy_model))
 
         # Step 2: Embed query
         embeddings = await asyncio.to_thread(self.embedding_model.embed, query, "search")
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -667,7 +667,7 @@ class Memory(MemoryBase):
         existing `_upsert_entity` helper. Non-fatal on any failure.
         """
         try:
-            entities = extract_entities(text)
+            entities = extract_entities(text, spacy_model=self.config.spacy_model)
             if not entities:
                 return
             seen = set()
@@ -970,7 +970,7 @@ class Memory(MemoryBase):
                 continue
             seen_hashes.add(mem_hash)
 
-            text_lemmatized = lemmatize_for_bm25(text)
+            text_lemmatized = lemmatize_for_bm25(text, spacy_model=self.config.spacy_model)
 
             memory_id = str(uuid.uuid4())
             mem_metadata = deepcopy(metadata)
@@ -1033,7 +1033,7 @@ class Memory(MemoryBase):
         # Phase 7: Batch entity linking
         try:
             all_texts = [r[1] for r in records]
-            all_entities = extract_entities_batch(all_texts)
+            all_entities = extract_entities_batch(all_texts, spacy_model=self.config.spacy_model)
 
             # 7a: Global dedup — collect unique entities across all memories
             global_entities = {}  # normalized_key -> (entity_type, entity_text, set of memory_ids)
@@ -1578,8 +1578,8 @@ class Memory(MemoryBase):
             threshold = 0.1
 
         # Step 1: Preprocess query
-        query_lemmatized = lemmatize_for_bm25(query)
-        query_entities = extract_entities(query)
+        query_lemmatized = lemmatize_for_bm25(query, spacy_model=self.config.spacy_model)
+        query_entities = extract_entities(query, spacy_model=self.config.spacy_model)
 
         # Step 2: Embed query
         embeddings = self.embedding_model.embed(query, "search")
@@ -1891,7 +1891,7 @@ class Memory(MemoryBase):
         if "created_at" not in new_metadata:
             new_metadata["created_at"] = datetime.now(timezone.utc).isoformat()
         new_metadata["updated_at"] = new_metadata["created_at"]
-        new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
+        new_metadata["text_lemmatized"] = lemmatize_for_bm25(data, spacy_model=self.config.spacy_model)
 
         self.vector_store.insert(
             vectors=[embeddings],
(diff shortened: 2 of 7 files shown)


kartik-mem0:
## Review: fix(memory): configure spaCy model for retrieval

**Verdict: Changes Requested** — two small fixes needed before this is ready to merge.

---

### Strengths

- **Architecture is correct.** The per-model-name dict cache (`_nlp_full_cache`, `_nlp_lemma_cache`) correctly promotes the old singletons, and the double-checked locking pattern is correctly preserved with the sets (`_load_failed_full`, `_load_failed_lemma`).
- **Full call-site coverage.** All 8 sync + 8 async paths in `Memory`/`AsyncMemory` pass `spacy_model=self.config.spacy_model`. Default is unchanged → fully backward-compatible.
- **Red-green proven.** 9/9 tests fail on main (field missing, `_nlp_full_cache` missing, `_ensure_model_available()` rejects the positional arg) and 9/9 pass on the PR head. Ruff clean across all 7 changed files.
- **No new required dependencies.** spaCy is already in the `[nlp]` optional group.

---

### Issues

**[Important] Async call sites pass `spacy_model` positionally; sync sites use keyword — `main.py:2159, 2506, 3037, 3038`**

```python
# async — positional (fragile)
await asyncio.to_thread(extract_entities, text, self.config.spacy_model)
await asyncio.to_thread(extract_entities_batch, all_texts, 32, self.config.spacy_model)

# sync — keyword (correct)
extract_entities(text, spacy_model=self.config.spacy_model)
```

Today the positions are correct, but `asyncio.to_thread` passes extra args positionally, so if any parameter is ever inserted before `spacy_model` in those signatures, the async call silently passes the wrong value while the sync call raises `TypeError` immediately. Please use keyword style consistently:

```python
await asyncio.to_thread(extract_entities, text, self.config.spacy_model)
# → prefer functools.partial or:
await asyncio.to_thread(lambda: extract_entities(text, spacy_model=self.config.spacy_model))
```

Or restructure as `functools.partial(extract_entities, text, spacy_model=self.config.spacy_model)`.

---

**[Important] Auto-download behavior on custom models is undocumented — `spacy_models.py:_ensure_model_available`**

`_ensure_model_available(model_name)` calls `spacy.cli.download(model_name)` for **any** model name that isn't installed — there's no guard limiting this to `en_core_web_sm`. A user who sets `spacy_model="de_core_news_lg"` will trigger a ~580MB silent pip install at first `Memory()` use, with no warning. In read-only or air-gapped environments this fails with no useful error.

Please add a note to the `MemoryConfig.spacy_model` Field description:

```python
spacy_model: str = Field(
    description=(
        "spaCy model name for entity extraction and BM25 lemmatization. "
        "The model must be installed before use: `python -m spacy download <model_name>`. "
        "If not installed, Mem0 will attempt to download it automatically on first use."
    ),
    default="en_core_web_sm",
)
```

---

**[Minor, informational] CJK/Arabic/Thai BM25 quality is out of scope for this PR**

Even with `spacy_model="zh_core_web_sm"`, BM25 lemmatization quality for Chinese text will be limited because `lemmatize_for_bm25` has some English-centric morphological logic (e.g., the `-ing` suffix check, stop-word filtering tuned for English). This PR correctly addresses **Latin-script non-English users** (German, French, Spanish, etc.). CJK/Arabic/Thai tokenization quality is a separate concern addressed by the complementary PR #5755. Consider adding a brief scope note to the PR description or the field docstring so users have the right expectations.

---

### Recommendations

- Fix the two Important items above (async positional args + field docstring with auto-download note)
- Consider whether this PR and PR #5755 should be coordinated before merge — both touch `spacy_models.py` significantly, and #5755's env-var approach for model selection could conflict or create an awkward dual-configuration story

---

### Evidence

| Signal | Result |
|---|---|
| CI Gate | PASS (Python SDK 3.10 / 3.11 / 3.12) |
| Red-green | 9/9 FAIL on main → 9/9 PASS on PR head |
| Ruff (7 files) | All checks passed |
| Existing tests/utils/ | 30 passed, 18 skipped (skips pre-existing, require real model download) |
| Worktree | `~/.cache/mem0-worktrees/review-pr-5719` @ f953acd6 |
| Vercel | auth failure — ignored (fork perms, not merge-blocking) |

rodboev:
Fixed both. The four async call sites in `main.py` now use lambdas so `spacy_model` is always passed by keyword regardless of future signature changes. The `MemoryConfig.spacy_model` field description now notes that the model must be pre-installed and documents the auto-download fallback behavior.

rodboev:
Added a scope note to the `spacy_model` field description about CJK, Arabic, and Thai: tokenization works with appropriate models but the BM25 lemmatization path uses English-centric morphology, so results are best for Latin-script languages. The note makes this boundary explicit in the docstring rather than leaving it to the linked PR.

kartik-mem0:
Re-reviewed against 7c75085.

All three items from the prior review are resolved:

**[Important-1] Async positional args — fixed.** All four async call sites in `main.py` now wrap the call in a lambda so `spacy_model` is always passed as a keyword arg: `asyncio.to_thread(lambda: extract_entities(text, spacy_model=self.config.spacy_model))`. This is exactly the safer pattern I requested — no risk of a silent positional mismatch if signatures evolve later. The `extract_entities_batch` call at line 2503 does the same with `batch_size=32` explicit, which is correct.

**[Important-2] Auto-download docstring — fixed.** The `MemoryConfig.spacy_model` Field description in `configs/base.py` now says both that the model must be pre-installed (`python -m spacy download <model_name>`) and that Mem0 will attempt an automatic download on first use if it's missing. That gives users the right warning up front rather than discovering it in production.

**[Minor] CJK/Arabic/Thai scope note — added.** Folded into the same field description: BM25 lemmatization uses English-centric morphology and is best suited for Latin scripts. Clear, in the right place.

---

One new issue to flag: this branch now has a merge conflict with main. Commit `b33fa542` ("fix(memory): align entity extraction precision", merged 2026-06-24) substantially rewrote `mem0/utils/entity_extraction.py` — new entity types, a `_EntityCandidate` dataclass, restructured `extract_entities` and `extract_entities_batch` signatures. Your PR's changes to those two functions (adding the `spacy_model` parameter and forwarding it to `get_nlp_full`) will need to be rebased onto that new shape. The rebase itself is straightforward — the intent is the same, it just needs to be applied to the new function bodies — but please resolve the conflict before this can merge.

`main.py` also has a conflict for the same reason; the rebase should handle it together.

Nothing else to add — the code is clean, CI passed on the last push, red-green is proven, and the architecture is sound. Rebase on current main and this is ready.

rodboev:
Rebased this onto current `main`.

The conflict came from the `entity_extraction.py` and `main.py` reshaping that landed in `b33fa542`, so I replayed the `spacy_model` plumbing onto that current shape and kept the previously reviewed async keyword-arg and docstring changes intact. The branch is mergeable again.

kartik-mem0:
Verified against 6a37d5e3.

The rebase is acknowledged — and you also picked up a genuine coverage gap in the process.

**New since 7c75085: scoring path is now fully threaded**

The `1fa81ac9b8` commit threads `spacy_model` through the scoring fallback and update paths that the prior review didn't flag:

- `scoring.py:16` — `get_bm25_params()` now accepts `spacy_model` and passes it to `lemmatize_for_bm25` when no pre-computed lemmatization is provided. Both sync and async `_search_vector_store` call sites pass `spacy_model=self.config.spacy_model` alongside the pre-computed `lemmatized` value, which makes the fallback branch safe for any future caller who doesn't pre-compute.
- `main.py ~1914 / ~3548` — both sync and async `_update_memory` now pass `spacy_model` to `lemmatize_for_bm25`. These were previously missed; they're covered now.

Total call sites in the final diff: 8 sync + 8 async, all threading `self.config.spacy_model`. Coverage is complete.

**LGTM confirmed — code is ready to squash-merge**

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 pass, CI Gate pass. Vercel auth failure ignored (fork, advisory).
- **Targeted tests:** CI passes all 9 tests in `tests/utils/test_spacy_model_config.py` (confirmed in prior cycle; new commits do not add test-logic changes requiring re-proof).
- **Red-green proof:** Captured in 2026-06-24 cycle — regression test fails without the fix, passes with it. New commits add call sites only; no logic change that would invalidate that proof.
- **Linked issue:** #4884 — all three root-cause paths (add/search/update, sync+async) are now configurable.
- **Cross-package check:** TypeScript SDK has no spaCy layer — no parity gap.
- **Could NOT verify:** The `CONFLICTING` status that GitHub still reports. The branch compiled and CI passed cleanly on the rebased head, but the API shows the branch is still behind main (likely `b91c745f` or `0fbbb2f5` touched the same region of `main.py` after your rebase landed). The fix itself is sound — a maintainer will need to do one more rebase onto the latest `main` before merging.

### Assessment
Ready to merge: **Yes** (squash) after rebase · Est. human time: ~5 min

PR #4444, opened 2026-03-20
feat(oss): lazy-load optional providers and add comprehensive PR gate

kgarg2468:
## Summary
- add a centralized optional dependency loader for OSS providers with consistent install hints (`Install optional dependency '<pkg>'`)
- switch optional provider SDK usage to runtime lazy loading across sqlite/history, vector stores, graph, and LLM/embedding adapters
- add a comprehensive regression gate for optional dependency behavior:
  - provider-matrix lazy-loading test coverage
  - loader utility unit tests
  - packed-artifact smoke harness (`pnpm pack` + clean install checks)
- add repeatable gate scripts and README verification docs for PR validation

## Why
Issue #3291 requires `mem0ai/oss` to avoid hard-failing on optional provider packages unless that provider is actually selected.

## What changed
- new optional dependency utility: `mem0-ts/src/oss/src/utils/optional-deps.ts`
- provider updates to use lazy loading for optional SDKs (sqlite, qdrant, redis, supabase, pgvector, cloudflare vectorize, azure ai search, neo4j, anthropic, google, groq, ollama, mistral, langchain)
- new tests:
  - `mem0-ts/src/oss/src/tests/optional-deps-lazy-loading.test.ts`
  - `mem0-ts/src/oss/src/tests/optional-deps-utils.test.ts`
- new publish-artifact smoke script:
  - `mem0-ts/scripts/optional-deps-pack-smoke.cjs`
- new npm scripts:
  - `test:optional-deps:lazy`
  - `test:optional-deps:utils`
  - `test:optional-deps:smoke`
  - `test:optional-deps:gate`
- README updates for optional dependency install guidance + PR-gate command sequence

## Tests Run
Executed on the branch for this PR:

```bash
pnpm run test:optional-deps:gate
```

The gate command runs and passed all of the following:

```bash
pnpm run test:optional-deps:lazy
pnpm run test:optional-deps:utils
pnpm test -- src/oss/src/tests/sqlite-path-resolution.test.ts
pnpm test -- src/oss/src/tests/sqlite-backward-compat.test.ts
pnpm run build
pnpm run test:optional-deps:smoke
```

Results summary:
- lazy-loading matrix: 17 passed
- optional-deps utility tests: 7 passed
- sqlite path-resolution: 19 passed
- sqlite backward-compat: 22 passed
- build: passed
- tarball smoke harness: passed (including clean install without optional peers)

Fixes #3291

PR #5720, opened 2026-06-21
fix(ts-sdk): lazy-load SQLite native bindings

### LINKED ISSUE #3291: Node.js SDK Still Requires Sqlite3 Despite setting `disableHistory: true`
## Context

I'm running Mem0, using Node.js SDK, deployed to AWS Lambda with the following config:

```ts
const memory = new Memory({
  version: 'v1.1',
  disableHistory: true,
  embedder: {
    provider: 'openai',
    config: {
      apiKey: env.OPENAI_API_KEY,
      model: 'text-embedding-3-small',
    },
  },
  vectorStore: {
    provider: 'opensearch',
    config: {
      collectionName: env.OPENSEARCH_COLLECTION_NAME,
      dimension: 1536,
      host: env.OPENSEARCH_ENDPOINT,
      port: 443,
      embedding_model_dims: 1024,
      use_ssl: true,
      verify_certs: true,
    },
  },
  llm: {
    provider: 'openai',
    config: {
      apiKey: env.OPENAI_API_KEY,
      model: 'gpt-5',
    },
  },
});
```

## Expected Behaviour

* Setting `disableHistory:true` will prevent loading of `Sqlite 3`. Especially important on Lambda runtime 

## Current Behaviour

* `mem0` still tries to load `Sqlite3`, causing runtime errors on AWS Lambda

Image

rodboev:
## Linked Issue

Closes #3291

## Description

The TypeScript OSS SDK loads `better-sqlite3` (a native C++ addon) at module parse time because `factory.ts` statically imports `SQLiteManager`. This crashes serverless and edge runtimes at cold start even when `disableHistory: true` is configured. This PR converts the static import to a dynamic `import()` inside the `case "sqlite"` branch, making `HistoryManagerFactory.create` async and moving the call site into the already-async `_autoInitialize()` method.

Stale PR #4444 covered the same surface but was a broader optional-provider refactor; this is a focused lazy-loading fix.

## Root cause

`factory.ts` line 26 statically imported `SQLiteManager`, which transitively loaded `better-sqlite3` at module parse time. The `disableHistory` check in `Memory` ran after module loading was complete, so the native addon was always initialized regardless of config.

## Scope

- `mem0-ts/src/oss/src/utils/factory.ts` — static import removed, `create` made async, dynamic import in `case "sqlite"`
- `mem0-ts/src/oss/src/memory/index.ts` — history manager init moved from constructor to `_autoInitialize()`, awaiting the async factory
- `mem0-ts/src/oss/tests/factory.unit.test.ts` — existing sqlite test updated to `await`, new lazy-load coverage

Does not split packages, remove dependencies, or implement #4444's full optional-provider matrix.

## Type of Change

## Breaking Changes

`HistoryManagerFactory.create` is now async (returns `Promise<HistoryManager>`). Any direct callers outside `Memory` that call it synchronously will need to `await`. The `Memory` public API is unchanged.

## Test Coverage

Verification:
- [x] `pnpm test src/oss/tests/factory.unit.test.ts` — all tests passing
- [ ] `pnpm install --frozen-lockfile && pnpm run build && pnpm test`

## Checklist

Claude Opus 4.6 via Claude Code CLI

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -241,13 +241,6 @@ export class Memory {
       }
     }
 
-    if (!this.config.disableHistory) {
-      this.db = await HistoryManagerFactory.create(
-        this.config.historyStore!.provider,
-        this.config.historyStore!,
-      );
-    }
-
     this.vectorStore = VectorStoreFactory.create(
       this.config.vectorStore.provider,
       this.config.vectorStore.config,
@@ -258,6 +251,13 @@ export class Memory {
     // store (collections, tables, etc.) is ready before any public method
     // attempts to read or write.
     await this.vectorStore.initialize();
+
+    if (!this.config.disableHistory) {
+      this.db = await HistoryManagerFactory.create(
+        this.config.historyStore!.provider,
+        this.config.historyStore!,
+      );
+    }
 
     await this._initializeTelemetry();
   }
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -209,14 +209,7 @@ export class Memory {
         this.config.reranker.config,
       );
     }
-    if (this.config.disableHistory) {
-      this.db = new DummyHistoryManager();
-    } else {
-      this.db = HistoryManagerFactory.create(
-        this.config.historyStore!.provider,
-        this.config.historyStore!,
-      );
-    }
+    this.db = new DummyHistoryManager();
 
     this.collectionName = this.config.vectorStore.config.collectionName;
     this.apiVersion = this.config.version || "v1.0";
@@ -246,6 +239,13 @@ export class Memory {
             `Please set 'dimension' in vectorStore.config or 'embeddingDims' in embedder.config explicitly.`,
         );
       }
+    }
+
+    if (!this.config.disableHistory) {
+      this.db = await HistoryManagerFactory.create(
+        this.config.historyStore!.provider,
+        this.config.historyStore!,
+      );
     }
 
     this.vectorStore = VectorStoreFactory.create(
--- a/mem0-ts/src/oss/src/utils/factory.ts
+++ b/mem0-ts/src/oss/src/utils/factory.ts
@@ -40,7 +40,6 @@ import { MiniMaxLLM } from "../llms/minimax";
 import { TogetherLLM } from "../llms/together";
 import { VllmLLM } from "../llms/vllm";
 import { SupabaseDB } from "../vector_stores/supabase";
-import { SQLiteManager } from "../storage/SQLiteManager";
 import { MemoryHistoryManager } from "../storage/MemoryHistoryManager";
 import { SupabaseHistoryManager } from "../storage/SupabaseHistoryManager";
 import { HistoryManager } from "../storage/base";
@@ -272,10 +271,12 @@ export class RerankerFactory {
 }
 
 export class HistoryManagerFactory {
-  static create(provider: string, config: HistoryStoreConfig): HistoryManager {
+  static async create(provider: string, config: HistoryStoreConfig): Promise<HistoryManager> {
     switch (provider.toLowerCase()) {
-      case "sqlite":
+      case "sqlite": {
+        const { SQLiteManager } = await import("../storage/SQLiteManager");
         return new SQLiteManager(config.config.historyDbPath || ":memory:");
+      }
       case "supabase":
         return new SupabaseHistoryManager({
           supabaseUrl: config.config.supabaseUrl || "",
--- a/mem0-ts/src/oss/src/utils/factory.ts
+++ b/mem0-ts/src/oss/src/utils/factory.ts
@@ -271,7 +271,10 @@ export class RerankerFactory {
 }
 
 export class HistoryManagerFactory {
-  static async create(provider: string, config: HistoryStoreConfig): Promise<HistoryManager> {
+  static async create(
+    provider: string,
+    config: HistoryStoreConfig,
+  ): Promise<HistoryManager> {
     switch (provider.toLowerCase()) {
       case "sqlite": {
         const { SQLiteManager } = await import("../storage/SQLiteManager");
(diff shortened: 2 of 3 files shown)


kartik-mem0:

### What this PR does
Converts the static `import { SQLiteManager }` at the top of `factory.ts` to a dynamic `import()` inside the `case "sqlite":` branch, and moves history manager init from the synchronous constructor to the already-async `_autoInitialize()`. This prevents `better-sqlite3` (a native C++ addon) from loading at module parse time on environments where it isn't available, fixing the crash-on-import reported in #3291 when `disableHistory: true`.

### Strengths
- `factory.ts:125` — dynamic `import()` is placed at exactly the right layer: inside the `case "sqlite":` branch of the one factory method that references `SQLiteManager`. Minimal blast radius, no other providers touched.
- `index.ts:188-195 → 228-235` — history manager init moves correctly into `_autoInitialize()`, alongside every other async init step. All public methods already await `_ensureInitialized()`, so there is no window where `this.db` is `DummyHistoryManager` after init completes.
- `factory.unit.test.ts` — all four history providers covered (sqlite, supabase, memory, bad-provider-throws). Tests are updated to async/await correctly.
- Conventional commit format (`fix(ts-sdk):` + linked `Closes #3291`). No new dependencies. No public `Memory` API changes.

### Issues
**Minor (optional):**
1. `factory.unit.test.ts:304–312` — "does not eagerly load SQLiteManager for non-sqlite providers" asserts that `create("memory")` resolves but doesn't spy on the dynamic import to verify SQLiteManager was never triggered. The code fix is complete (the static import is simply absent from the file), so this is an observability gap in the test, not a coverage gap in the fix. Optional follow-up: `jest.isolateModules` or a spy on the import path would make the intent self-documenting.

### Evidence
- **CI:** TypeScript SDK build + integration on Node 20 & 22: all PASS. CI Gate: PASS. Vercel/deploy check: FAIL (advisory, fork deploy perms — not merge-blocking).
- **Targeted tests:** `pnpm test src/oss/tests/factory.unit.test.ts` → 39 passed in 0.5s
- **Red-green proof:** Reverted `factory.ts` to pre-PR state (static import restored) → 5 HistoryManagerFactory tests fail ("received value must be a promise" — `create()` returns a sync object, not a Promise). Restored PR state → 39 passed. Regression tests bite the change.
- **Linked issue:** #3291 — `better-sqlite3` crash on AWS Lambda cold start with `disableHistory: true`. PR eliminates the static import that caused eager loading at module parse time.
- **Cross-package check:** Python SDK uses in-function imports by default — no equivalent eager-load issue. TS-only fix is correct scope.
- **Duplicate cluster:** PR #4444 covers similar territory but is a much broader optional-provider refactor (976 additions, MERGEABLE=UNKNOWN). #5720 is the focused fix appropriate for #3291.
- **Could NOT verify:** Live Lambda/edge cold start (requires deploy). Code tracing confirms the native binding cannot load at module parse time once the static import is removed.

### Assessment
Focused, correct fix at the right layer. The root cause (static import at module top-level) is eliminated; the async init sequence is consistent with the existing `_autoInitialize()` pattern. All tests pass, red-green proven.

rodboev:
Added a `jest.mock` for `SQLiteManager` and updated the lazy-loading test to call `SQLiteManager.mockClear()` then assert it was never called after `create("memory", ...)`. The test now actively verifies the import is not triggered for non-sqlite providers rather than just checking the result resolves.

nothariharan:
## Review: correct fix for the history path — but #3291 still reproduces

I checked this against current `main` by tracing the import graph and inspecting the built bundle. The change itself is clean and correct; the issue is that it only covers one of the two eager `better-sqlite3` loads.

### What's good
- The dynamic `import()` inside `case "sqlite":` is at the right layer — minimal blast radius, no other providers touched.
- Making `HistoryManagerFactory.create` async and moving history init into `_autoInitialize()` is consistent with the existing pattern. Every `this.db` use sits behind `await this._ensureInitialized()` (`memory/index.ts:764, 1291, 1374, 1655, 1700, 1721, 1784, 1791, 1849`), so there is no window where the dummy manager leaks to callers.
- CI is green: `CI Gate`, TypeScript SDK build/integration on Node 20 & 22, and `license/cla`.

### The gap
`better-sqlite3` has **two** top-level importers in the OSS graph, and this PR removes only the first:

| Importer | File | After this PR |
|---|---|---|
| History manager | `mem0-ts/src/oss/src/storage/SQLiteManager.ts:1` | static import removed |
| In-memory vector store | `mem0-ts/src/oss/src/vector_stores/memory.ts:3` | still static |

`vector_stores/memory.ts` stays in the module graph two ways: `mem0-ts/src/oss/src/utils/factory.ts:11` (`import { MemoryVectorStore } from "../vector_stores/memory"`) and the public barrel `mem0-ts/src/oss/src/index.ts:30` (`export * from "./vector_stores/memory"`). It cannot be tree-shaken — `MemoryVectorStore` is instantiated in `VectorStoreFactory`.

Built bundle on `main` (`mem0-ts/dist/oss/index.js`) has two module-scope requires:

```
869:  ...require("better-sqlite3")   // src/oss/src/vector_stores/memory.ts
4805: ...require("better-sqlite3")   // src/oss/src/storage/SQLiteManager.ts
```

This PR removes 4805; 869 remains. Both run at require time regardless of `disableHistory`, so the Lambda / edge cold-start crash from #3291 still reproduces.

This lines up with the triage on #3291, which calls for lazy-loading `SQLiteManager` **and** a dynamic `import()` for `better-sqlite3` in `memory.ts`.

### Suggested
Convert the `better-sqlite3` load in `vector_stores/memory.ts` to a dynamic import resolved when the store is constructed (lazy `Database` handle), so `MemoryVectorStore` no longer binds the native addon at module evaluation. A smoke test that imports the OSS entry with `better-sqlite3` unavailable — or asserts the packed artifact doesn't require it at load — would then lock the fix in.

### Minor
- `HistoryManagerFactory` is publicly re-exported (`mem0-ts/src/oss/src/index.ts:60`), so `create` becoming async is a breaking change for external callers — worth a changelog line.
- The new "does not eagerly load" test mocks only `SQLiteManager`, so it cannot catch the `memory.ts` path — which is the one still broken.

Issue #2636, opened 2025-05-06
Memory.add triggering Azure OpenAI's content management policy


V-Silpin:
### 🐛 Describe the bug

# Description

While following the mem0 documentation, I attempted to test the mem0 module using Azure OpenAI LLM, Vector Store, and Embedding Model. However, I encountered an error related to the content filter. I tried to resolve this by adding a safe prompt in the prompt parameter and slightly changing the messages, but my efforts were unsuccessful.

**Possible Causes**

I suspect that the example provided in the [DOCS](https://docs.mem0.ai/open-source/python-quickstart#store-a-memory) is triggering a content filter type labeled **Indirect Attacks**. You can find more information about this filter type in the [Microsoft documentation](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#content-filter-types).

I kindly request to investigate whether this error originates from the module itself or from the code provided below. Additionally, I would appreciate any solutions to this issue.

# Python Code
```python
from dotenv import load_dotenv
import os

from mem0 import Memory

load_dotenv()

llm_provider = os.getenv("AZURE_OPENAI_PROVIDER")
llm_model = os.getenv("AZURE_OPENAI_MODEL")
llm_temperature = float(os.getenv("AZURE_OPENAI_TEMPERATURE"))
llm_max_tokens = int(os.getenv("AZURE_OPENAI_MAX_TOKENS"))
llm_api_version = os.getenv("AZURE_OPENAI_API_VERSION")
llm_azure_deployment = os.getenv("AZURE_OPENAI_DEPLOYMENT")
llm_azure_endpoint = os.getenv("AZURE_OPENAI_ENDPOINT")
llm_api_key = os.getenv("AZURE_OPENAI_API_KEY")

vs_provider = os.getenv("AZURE_VECTOR_STORE_PROVIDER")
vs_service_name = os.getenv("AZURE_VECTOR_STORE_SERVICE_NAME")
vs_api_key = os.getenv("AZURE_VECTOR_STORE_API_KEY")
vs_collection_name = os.getenv("AZURE_VECTOR_STORE_COLLECTION_NAME")
vs_embedding_model_dims = int(os.getenv("AZURE_VECTOR_STORE_EMBEDDING_MODEL_DIMS"))

em_provider = os.getenv("EMBEDDING_AZURE_PROVIDER")
[...]

PR #5731, opened 2026-06-21
fix(azure): stop mutating and corrupting caller messages in content rewrite

### LINKED ISSUE #5732: Azure providers rewrite the caller's last message in place (assistant -> ai), corrupting content and crashing on multimodal input
### Summary

While going through the Azure providers I noticed that both `AzureOpenAILLM` and `AzureOpenAIStructuredLLM` rewrite the last user message in place before the request. They do `messages[-1]["content"].replace("assistant", "ai")` to keep the word `assistant` out of the prompt (it can trip Azure's content filter, that's the "Indirect Attacks" thing from #2636, added in #2937).

The idea makes sense, but the way it's done has a few side effects:

- it edits the caller's `messages` list directly, so the change leaks back out to whoever passed the list in;
- `str.replace` hits the substring anywhere, so `"my assistant helps me"` ends up as `"my ai helps me"` — and that rewritten text is what actually gets sent and then stored as memory;
- if the last message has multimodal `content` (a list of blocks), `.replace` blows up with `AttributeError: 'list' object has no attribute 'replace'`.

### A note on priority

Just to be honest about scope: this isn't urgent. It's Azure-only, and in the normal `add()` path you'd only hit it if the user's last message literally contains "assistant", or if you pass multimodal content, or if something reuses the same `messages` list afterwards. So it's a real correctness bug but a fairly narrow one. Mostly opening this so the fix has something to reference.

### Reproduce

```python
from mem0.configs.llms.azure import AzureOpenAIConfig

llm = AzureOpenAILLM(AzureOpenAIConfig(model="<deployment>", azure_kwargs={...}))

messages = [{"role": "user", "content": "my assistant helps me schedule meetings"}]
llm.generate_response(messages)
# messages[-1]["content"] is now "my ai helps me schedule meetings"

llm.generate_response([{"role": "user", "content": [{"type": "text", "text": "describe my assistant"}]}])
# AttributeError: 'list' object has no attribute 'replace'
```

### Expected

The workaround should only touch the copy sent to the model, leave the caller's messages alone, and not choke on list content.

### Environment

- mem0 version: main
- Python: 3.12
- OS: macOS

Hexecu:
## Linked Issue

Fixes #5732

Context: #2636 (closed) introduced this rewrite via #2937; #5223 is an open PR addressing part of the same code.

## Description

Both Azure providers rewrote the last user message in place before sending it to the model. The original goal was to keep the literal word `assistant` out of the request, since it can trip Azure's content-management filter (the "Indirect Attacks" detection described in #2636, which the rewrite was added for in #2937).

The rewrite was implemented as:

```python
user_prompt = messages[-1]["content"]
user_prompt = user_prompt.replace("assistant", "ai")
messages[-1]["content"] = user_prompt
```

This caused three issues:

- **Caller mutation** — the caller-owned `messages` list was edited in place, so the change leaked back to whatever built the list.
- **User-data corruption** — every occurrence of the substring `assistant` in the user's last message was replaced with `ai` (e.g. `"my assistant helps me"` became `"my ai helps me"`), and that altered text is what gets sent to the model and then extracted/persisted as memory.
- **Crash on multimodal content** — when `content` is a list (multimodal messages), `str.replace` raised `AttributeError: 'list' object has no attribute 'replace'`.

This PR moves the rewrite onto a deep copy that is only sent to the model, guards for non-string content, and leaves the caller's messages untouched. The original intent (keeping `assistant` out of the request) is preserved. The same fix is applied to both `AzureOpenAILLM` and `AzureOpenAIStructuredLLM`.

I noticed #5223 by @ATOM00blue already started addressing the multimodal crash and the in-place mutation for `azure_openai.py` with an `isinstance` guard, which is the right direction. This PR builds on that idea but also stops corrupting the string content (the original message is preserved) and covers `azure_openai_structured.py` as well.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added unit tests for both providers covering: the caller's `messages` is not mutated, the `assistant` -> `ai` rewrite is applied only to the copy sent to the model, and multimodal (list) content no longer crashes.

I also verified the behavior end-to-end against a live Azure OpenAI deployment (`gpt-4o-mini`, api-version `2024-10-21`) for both `AzureOpenAILLM` and `AzureOpenAIStructuredLLM`:

- before the fix: the caller's message was mutated in place and multimodal content raised `AttributeError`;
- after the fix: the caller's message stays intact (`"my assistant helps me schedule meetings"`), the model still receives the rewritten copy, and multimodal content goes through without error.

`tests/llms/` is green: `164 passed, 2 skipped`.

## Checklist

### DIFF
--- a/mem0/llms/azure_openai.py
+++ b/mem0/llms/azure_openai.py
@@ -1,3 +1,4 @@
+import copy
 import json
 import os
 from typing import Dict, List, Optional, Union
@@ -69,6 +70,26 @@ class AzureOpenAILLM(LLMBase):
             default_headers=default_headers,
         )
 
+    @staticmethod
+    def _rewrite_assistant_keyword(messages):
+        """
+        Return a copy of ``messages`` with the word "assistant" replaced by "ai"
+        in the last message's textual content.
+
+        Azure's content management policy can flag the literal word "assistant",
+        which makes ``add`` fail (see issue #2636). The rewrite targets that
+        trigger without mutating the caller's messages and without assuming the
+        content is a string, so multimodal (list) content passes through untouched.
+        """
+        if not messages:
+            return messages
+
+        messages = copy.deepcopy(messages)
+        last_content = messages[-1].get("content")
+        if isinstance(last_content, str):
+            messages[-1]["content"] = last_content.replace("assistant", "ai")
+        return messages
+
     def _parse_response(self, response, tools):
         """
         Process the response based on whether tools are used or not.
@@ -121,14 +142,14 @@ class AzureOpenAILLM(LLMBase):
             str: The generated response.
         """
 
-        user_prompt = messages[-1]["content"]
-
-        user_prompt = user_prompt.replace("assistant", "ai")
-
-        messages[-1]["content"] = user_prompt
+        # Azure's "Indirect Attacks" content filter can flag the literal word
+        # "assistant" in the prompt, so it is rewritten to "ai" before the request.
+        # Work on a copy so the caller's messages are left untouched and string-only
+        # content is handled without breaking multimodal (list) content.
+        messages = self._rewrite_assistant_keyword(messages)
 
         params = self._get_supported_params(messages=messages, **kwargs)
-        
+
         # Add model and messages
         params.update({
             "model": self.config.model,
--- a/mem0/llms/azure_openai_structured.py
+++ b/mem0/llms/azure_openai_structured.py
@@ -1,3 +1,4 @@
+import copy
 import json
 import os
 from typing import Dict, List, Optional
@@ -66,11 +67,11 @@ class AzureOpenAIStructuredLLM(LLMBase):
             str: The generated response.
         """
 
-        user_prompt = messages[-1]["content"]
-
-        user_prompt = user_prompt.replace("assistant", "ai")
-
-        messages[-1]["content"] = user_prompt
+        # Azure's "Indirect Attacks" content filter can flag the literal word
+        # "assistant" in the prompt, so it is rewritten to "ai" before the request.
+        # Work on a copy so the caller's messages are left untouched and string-only
+        # content is handled without breaking multimodal (list) content.
+        messages = self._rewrite_assistant_keyword(messages)
 
         is_reasoning = self._is_reasoning_model(self.config.model)
         params = {
@@ -100,6 +101,26 @@ class AzureOpenAIStructuredLLM(LLMBase):
 
         response = self.client.chat.completions.create(**params)
         return self._parse_response(response, tools)
+
+    @staticmethod
+    def _rewrite_assistant_keyword(messages):
+        """
+        Return a copy of ``messages`` with the word "assistant" replaced by "ai"
+        in the last message's textual content.
+
+        Azure's content management policy can flag the literal word "assistant",
+        which makes ``add`` fail (see issue #2636). The rewrite targets that
+        trigger without mutating the caller's messages and without assuming the
+        content is a string, so multimodal (list) content passes through untouched.
+        """
+        if not messages:
+            return messages
+
+        messages = copy.deepcopy(messages)
+        last_content = messages[-1].get("content")
+        if isinstance(last_content, str):
+            messages[-1]["content"] = last_content.replace("assistant", "ai")
+        return messages
 
     def _parse_response(self, response, tools):
         """
--- a/tests/llms/test_azure_openai.py
+++ b/tests/llms/test_azure_openai.py
@@ -135,6 +135,52 @@ def test_generate_response_without_response_format(mock_openai_client):
     assert response == "Why did the chicken cross the road?"
 
 
+def test_generate_response_does_not_mutate_caller_messages(mock_openai_client):
+    config = AzureOpenAIConfig(model=MODEL, temperature=TEMPERATURE, max_tokens=MAX_TOKENS, top_p=TOP_P)
+    llm = AzureOpenAILLM(config)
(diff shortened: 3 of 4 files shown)


Hexecu:
Some context on why this is a complete fix rather than a guard-only patch.

The current code in both `azure_openai.py` and `azure_openai_structured.py` does:

```python
user_prompt = messages[-1]["content"]
user_prompt = user_prompt.replace("assistant", "ai")
messages[-1]["content"] = user_prompt
```

This was added in #2937 (Closes #2636) to work around Azure's "Indirect Attacks" content filter, which can be triggered by the word `assistant` in the default prompt. The intent is reasonable, but the implementation has three concrete problems:

1. **It mutates the caller's `messages` in place.** The same list is reused upstream (e.g. fact extraction reads it back after the call), so the rewrite leaks back to the caller.
2. **It corrupts user data semantically.** `str.replace("assistant", "ai")` rewrites the substring anywhere in the last user message, so `"my assistant helps me schedule meetings"` becomes `"my ai helps me schedule meetings"` — and that altered text is what gets persisted/extracted.
3. **It crashes on multimodal content.** When `messages[-1]["content"]` is a list (image/text blocks), `.replace` raises `AttributeError: 'list' object has no attribute 'replace'`.

The fix keeps the content-filter intent but applies the rewrite to a `copy.deepcopy` of the messages that is sent to the model only, leaves the caller's list untouched, and guards for non-string content so multimodal inputs pass through unchanged.

I verified this end-to-end against a live Azure deployment (gpt-4o-mini), on both `AzureOpenAILLM` and `AzureOpenAIStructuredLLM`:

- **Before:** caller content was mutated to `"my ai ..."`, and a list-shaped `content` raised `AttributeError`.
- **After:** the caller's `messages` is unchanged, multimodal content no longer crashes, and the model still receives `ai` in place of `assistant`, so the original #2636 behavior is preserved.

`tests/llms/` passes (164 passed, 2 skipped) and `ruff check` is clean on the touched files.

Thanks to @ATOM00blue — #5223 already spotted the multimodal crash and the in-place mutation in `azure_openai.py`, which helped confirm the issue. This PR builds on that direction and additionally avoids corrupting the user's text on the string path and covers `azure_openai_structured.py` as well, so both Azure providers behave consistently. Happy to align on whichever approach the maintainers prefer.

kartik-mem0:
LGTM — ready to squash-merge.

Thanks Davide — this is a clean, correct fix. Here is the full evidence from this review run.

Strengths

- Root cause correctly identified. The three failure modes (caller mutation, substring corruption of persisted memory, multimodal AttributeError) are all real and the fix addresses each at the right level.
- deepcopy is the right mechanism. A shallow copy of the list would still share the inner dict objects, so mutation would silently leak through. deepcopy is correct here.
- Symmetric coverage. Both azure_openai.py and azure_openai_structured.py are fixed identically. The helper is a @staticmethod so each class owns its copy — no cross-class dependency.
- Guard is precise. isinstance(last_content, str) only rewrites when content is a string and passes list (multimodal) content through untouched, which is exactly what Azure's API expects.
- Tests cover all three failure modes, for both providers (6 new tests total), at the right granularity: mutation test, rewrite-copy-only test, multimodal pass-through test.

Issues

None confirmed after full red-green verification and skeptic review. One note worth flagging:

Minor — _rewrite_assistant_keyword is duplicated between AzureOpenAILLM and AzureOpenAIStructuredLLM (mem0/llms/azure_openai.py:73-91 and mem0/llms/azure_openai_structured.py:105-123). They are byte-for-byte identical. A shared base or a module-level helper in a future cleanup would be cleaner, but this is not a blocker — both classes already inherit from LLMBase which has no Azure-specific coupling, so keeping them side by side is reasonable for now.

Architecture check

The rewrite belongs in generate_response, not in a shared memory layer — correct layer. It has no hot-path performance concern (deepcopy of a small messages list is negligible). The fix does not change the Azure content-filter workaround itself, only where the copy is made. The original intent (keep the literal word out of the model request) is fully preserved.

Compared PR #5223 (also open, by ATOM00blue): that PR fixes only the multimodal AttributeError in azure_openai.py via an isinstance guard, but still mutates the caller's messages in place and does not touch azure_openai_structured.py. PR #5731 is the superset — it removes all three bugs across both providers. #5223 should be closed in favour of this one once it merges.

Evidence

CI Gate: pass (Python SDK build on 3.10, 3.11, 3.12 all green; Vercel check is a fork-perms advisory and is not merge-blocking)

Targeted tests (33 total, tests/llms/test_azure_openai.py + tests/llms/test_azure_openai_structured.py):
  33 passed in 3.39s

Red-green proof on 6 new regression tests:
  RED (against main/pre-fix):
    FAILED test_generate_response_does_not_mutate_caller_messages [azure_openai] — AssertionError: 'my ai helps...' != 'my assistant...'
    FAILED test_generate_response_rewrites_assistant_keyword_for_model_only [azure_openai] — AssertionError: caller content mutated
    FAILED test_generate_response_handles_multimodal_content [azure_openai] — AttributeError: 'list' object has no attribute 'replace'
    FAILED test_generate_response_does_not_mutate_caller_messages [azure_openai_structured] — AssertionError: caller content mutated
    FAILED test_generate_response_rewrites_assistant_keyword_for_model_only [azure_openai_structured] — AssertionError: caller content mutated
    FAILED test_generate_response_handles_multimodal_content [azure_openai_structured] — AttributeError: 'list' object has no attribute 'replace'
  GREEN (with fix restored): 6 passed in 0.51s

Lint: ruff check clean on all 4 changed files.

Merge guidance: gh pr merge 5731 --squash

kartik-mem0 (approved):

PR #5737, opened 2026-06-21
fix(server): make GET /entities resilient to naive timestamps (500 on mixed tz)

jaketribe417:
## Problem

`GET /entities` returns **500 Internal Server Error** once an entity owns memories with a mix of timezone-aware and timezone-naive `created_at`/`updated_at` values.

`_parse_timestamp()` returns:

```python
return datetime.fromisoformat(str(value).replace("Z", "+00:00"))
```

For values without a timezone (e.g. `"2026-06-21T00:33:15.976797"`) this yields a **naive** datetime, while `"...+00:00"` values yield **aware** ones. `list_entities()` then compares them:

```python
if created and (bucket["created_at"] is None or created < bucket["created_at"]):
```

Mixing the two raises `TypeError: can't compare offset-naive and offset-aware datetimes`, surfaced as a 500. This happens in practice when an entity accumulates both LLM-extracted memories and `infer=False` writes (their stored timestamp formats differ).

**Traceback (self-hosted server):**
```
File "/app/routers/entities.py", line 59, in list_entities
    if created and (bucket["created_at"] is None or created < bucket["created_at"]):
TypeError: can't compare offset-naive and offset-aware datetimes
```

## Fix

Normalize parsed timestamps to tz-aware UTC in `_parse_timestamp()`, so every comparison in `list_entities()` is between aware datetimes. Minimal, no API change.

## Test plan
- Store memories for one `user_id` such that some payloads have tz-aware and some tz-naive `created_at`; `GET /entities` returned 500 before, returns 200 after.

### DIFF
--- a/server/routers/entities.py
+++ b/server/routers/entities.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any, Literal, Optional
 
 from auth import require_admin, verify_auth
@@ -35,9 +35,12 @@ def _parse_timestamp(value: Any) -> Optional[datetime]:
     if not value:
         return None
     try:
-        return datetime.fromisoformat(str(value).replace("Z", "+00:00"))
+        dt = datetime.fromisoformat(str(value).replace("Z", "+00:00"))
     except ValueError:
         return None
+    # Normalize to tz-aware UTC so naive timestamps don't raise when compared
+    # against tz-aware ones in list_entities().
+    return dt if dt.tzinfo is not None else dt.replace(tzinfo=timezone.utc)
 
 
 @router.get("", response_model=list[Entity])


kartik-mem0:
## Changes Requested

### Strengths
- `server/routers/entities.py:37-42` — The fix targets the exact parsing function (`_parse_timestamp`) that produces mismatched types, which is the right layer. Every timestamp read by `list_entities` flows through here, so both `created_at` and `updated_at` comparisons at lines 59 and 62 are fixed in one place — no code duplication, no missed callsite.
- Backward compatible: Z-suffix timestamps (`2026-...Z`) were already converted to `+00:00` before parsing, so they were already aware — the new branch (`dt if dt.tzinfo is not None`) leaves them unchanged. Non-UTC aware timestamps (e.g. `+05:30`) are also preserved as-is.
- Minimal change — 5 lines, no new dependency, no API surface change.

### Issues

1. `server/routers/entities.py` (entire file) — **No regression test added.** CLAUDE.md and review standards both require a regression test for bug fixes — one that fails without the change. `_parse_timestamp` is a pure stdlib function (only `datetime` + `timezone` imports in its body) and is straightforward to test without a running server. A pytest parametrize over naive/aware/Z/None inputs — where the comparison step (`created < bucket["created_at"]`) would raise `TypeError` on the pre-fix naive result — would be the right test. Example sketch:

   ```python
   # server/tests/test_entities.py
   import pytest
   # (mock server-level imports or copy the function under test)

   @pytest.mark.parametrize("ts,expect_tz_aware", [
       ("2026-06-21T00:33:15.976797", True),       # naive → UTC-aware after fix
       ("2026-06-21T00:33:15.976797+00:00", True), # already aware → unchanged
       ("2026-06-21T00:33:15.976797Z", True),       # Z-suffix → aware
       ("", None),                                   # empty → None
       ("not-a-date", None),                         # invalid → None
   ])
   def test_parse_timestamp(ts, expect_tz_aware):
       result = _parse_timestamp(ts or None)
       if expect_tz_aware is None:
           assert result is None
       else:
           assert result is not None
           assert result.tzinfo is not None
           # Verify the mixed-tz comparison no longer raises:
           aware = datetime(2026, 1, 1, tzinfo=timezone.utc)
           _ = result < aware  # must not raise TypeError
   ```

   This test would fail on the pre-fix code (naive result raises `TypeError` on comparison) and pass with the fix.

### Evidence
- **CI:** No real test/lint/build checks ran — `server/` is not covered by the CI gate (`ci-gate.yml` path filters only cover `mem0/**`, `tests/**`, and TS/CLI packages). Vercel failure is advisory (fork perms, fails on every PR) — ignored.
- **Red-green proof (manual):** Executed pre-fix and post-fix `_parse_timestamp` logic directly in Python:
  - Pre-fix: `"2026-06-21T00:33:15.976797"` → naive datetime; comparison with aware datetime raises `TypeError: can't compare offset-naive and offset-aware datetimes` (the reported 500).
  - Post-fix: same input → `2026-06-21 00:33:15.976797+00:00` (UTC-aware); comparison succeeds.
- **UTC assumption:** Checked `mem0/memory/main.py` — all SDK writes use `datetime.now(timezone.utc).isoformat()`. Naive timestamps in stored data are legacy or externally inserted; treating them as UTC is correct for this codebase.
- **Duplicate cluster:** No other open PRs touch this code path.
- **Cross-package check:** TS SDK (`mem0-ts/`) unaffected — this is server-only logic.
- **Could NOT verify:** live vector store behavior with a mixed-tz dataset (would require Docker + a populated Qdrant instance). Code tracing is sufficient for this fix's correctness.
- **No linked issue:** mind opening a quick issue for this bug and linking it here with `Closes #N`? Makes it easier to track and prioritize.

### Assessment
Ready to merge: **With fixes** — please add the regression test for `_parse_timestamp`; I'll re-review once you push. The fix logic itself is correct and the root cause analysis is accurate.

PR #5739, opened 2026-06-22
fix(pinecone): translate all metadata filter operators, not only gte+lte ranges

ly-wang19:
## Description

`PineconeDB._create_filter` only translates a metadata-filter operator dict when it contains **both** `gte` and `lte`:

```python
for key, value in filters.items():
    if isinstance(value, dict) and "gte" in value and "lte" in value:
        pinecone_filter[key] = {"$gte": value["gte"], "$lte": value["lte"]}
    else:
        pinecone_filter[key] = {"$eq": value}   # <-- swallows every other operator
```

Every other operator dict that `Memory._process_metadata_filters` emits — a single-bound range `{"gte": 5}` / `{"lt": 10}`, or `{"in": [...]}`, `{"nin": ...}`, `{"ne": ...}`, `{"eq": ...}` — falls into the `else` branch and is wrapped as the **value** of an `$eq`:

```python
{"score": {"gte": 5}}            -> {"score": {"$eq": {"gte": 5}}}
{"category": {"in": ["a","b"]}}  -> {"category": {"$eq": {"in": ["a","b"]}}}
```

Pinecone then matches nothing, so **advanced metadata filtering silently returns wrong/empty results on Pinecone-backed stores** for any filter other than a closed `gte`+`lte` range. (Basic scalar filters like `{"user_id": "..."}` go through the `else` branch as `{"$eq": value}`, which is correct, so they are unaffected.)

The intended translation is established in the repo's sibling adapters — `ChromaDB._generate_where_clause` and `Qdrant._build_field_condition` both map each operator individually — and the operator set is fixed in `Memory._process_metadata_filters` (`eq, ne, gt, gte, lt, lte, in, nin, ...`).

## Fix

Translate each operator to its Pinecone equivalent (`gte`→`$gte`, `in`→`$in`, …), collecting multiple operators on one key into a single condition (so a closed `{"gte", "lte"}` range still works). `contains`/`icontains` and unknown operators fall back to `$eq`, matching the ChromaDB sibling.

## Type of change

## How tested

Added `tests/vector_stores/test_pinecone.py::test_create_filter_translates_all_operators` (pure offline, uses the existing mocked-client fixture):

- **Before:** `_create_filter({"score": {"gte": 5}})` → `{"score": {"$eq": {"gte": 5}}}` (assertion fails).
- **After:** `{"score": {"$gte": 5}}`; also covers `in`/`nin`/`ne`, the closed `gte`+`lte` range (regression), and scalar equality.

`pytest tests/vector_stores/test_pinecone.py` → **16 passed**. `ruff check` / `ruff format --check` clean on both files.

### DIFF
--- a/mem0/vector_stores/pinecone.py
+++ b/mem0/vector_stores/pinecone.py
@@ -195,9 +195,27 @@ class PineconeDB(VectorStoreBase):
 
         pinecone_filter = {}
 
+        # Map the universal operators emitted by Memory._process_metadata_filters
+        # to Pinecone's filter syntax (see the sibling translation in
+        # ChromaDB._generate_where_clause / Qdrant._build_field_condition).
+        operator_map = {
+            "eq": "$eq",
+            "ne": "$ne",
+            "gt": "$gt",
+            "gte": "$gte",
+            "lt": "$lt",
+            "lte": "$lte",
+            "in": "$in",
+            "nin": "$nin",
+        }
+
         for key, value in filters.items():
-            if isinstance(value, dict) and "gte" in value and "lte" in value:
-                pinecone_filter[key] = {"$gte": value["gte"], "$lte": value["lte"]}
+            if isinstance(value, dict):
+                # An operator dict such as {"gte": 5}, {"in": [...]} or a closed
+                # range {"gte": 5, "lte": 10}. Translate every operator (Pinecone
+                # has no substring match, so contains/icontains fall back to
+                # equality, matching the ChromaDB sibling).
+                pinecone_filter[key] = {operator_map.get(op, "$eq"): val for op, val in value.items()}
             else:
                 pinecone_filter[key] = {"$eq": value}
 
--- a/tests/vector_stores/test_pinecone.py
+++ b/tests/vector_stores/test_pinecone.py
@@ -188,3 +188,25 @@ def test_count_with_none_vector_count(pinecone_db):
     count = pinecone_db.count()
     assert count == 0
     pinecone_db.index.describe_index_stats.assert_called_once()
+
+
+def test_create_filter_translates_all_operators(pinecone_db):
+    """Operator dicts must map to Pinecone operators, not be wrapped as a value.
+
+    Regression test: ``_create_filter`` previously only handled a closed
+    ``{"gte": ..., "lte": ...}`` range and wrapped every other operator dict as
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Review — fix(pinecone): translate all metadata filter operators

### Strengths

- **Correct root-cause diagnosis.** The old `if "gte" in value and "lte" in value` guard was indeed silently dropping every other operator dict (`gt`, `lt`, `in`, `nin`, `ne`, single-bound ranges) — wrapping them as `{"$eq": {"gte": 5}}` which Pinecone matches against nothing. Good catch and a clear write-up.
- **Fix approach is right.** Translating each operator key individually (rather than special-casing one combination) is the correct pattern, matching `ChromaDB._generate_where_clause` and `Qdrant._build_field_condition`.
- **Tests cover the essential cases** — single-bound ranges, `in`/`nin`/`ne`, closed-range regression, and scalar equality.

---

### Issues

**Changes Requested — Already fixed on main**

PR #5707 ("fix(pinecone): map all comparison operators in `_create_filter()`") was merged on **2026-06-23** (commit `716f021d`), the day after this PR was opened. It applies the **same fix**: a class-level `OPERATOR_MAP` constant plus a per-operator loop in `_create_filter`, with equivalent tests. Main already has the corrected `PineconeDB._create_filter`.

This is what's causing the CONFLICTING merge state — there is nothing left for this PR to add.

**Evidence:**
```
$ git log --oneline origin/main -- mem0/vector_stores/pinecone.py | head -3
716f021d fix(pinecone): map all comparison operators in _create_filter() (#5707)
bd5996f4 fix(pinecone): return [[]] from list() error path instead of dict (#5706)
d4646428 fix(pinecone): hybrid search crashes when filters is None (#5533)
```

---

### CI

All real checks (CI Gate, Python SDK build 3.10/3.11/3.12, changelog, CLA) passed. The Vercel failure is a deploy-preview authorization issue that affects all fork PRs and is not merge-blocking.

---

### Assessment

**Ready to merge: No** — the fix is already on `main` via #5707. Please close this PR. The bug fix and the credit for the diagnosis are documented in the commit history regardless.

One gentle ask: would you be open to opening a brief tracking issue for this class of filter-translation bugs (operators silently becoming no-ops)? It would help surface similar gaps in other vector store adapters going forward. No @-mention needed — just a short issue if you have a moment.

---

*Evidence: `.claude/babysit-evidence/pr-5739-2026-06-24.md`*

PR #5745, opened 2026-06-22
perf(opensearch): batch insert into one bulk request + a single refresh

ly-wang19:
## Linked Issue

N/A — standalone performance improvement.

## Description

`OpenSearchDB.insert()` issued **one `client.index()` call _and_ one full `indices.refresh()` per vector**:

```python
for i, (vec, id_) in enumerate(zip(vectors, ids)):
    body = {...}
    self.client.index(index=self.collection_name, body=body)
    self.client.indices.refresh(index=self.collection_name)   # forced segment flush, per doc
```

So writing `N` vectors meant `N` indexing round-trips **plus `N` index-wide refreshes** — and each `refresh()` forces a Lucene segment flush. On the memory-write hot path (`add()` extracts and inserts several facts at once; bulk imports insert many), this is pathological.

This sends all documents in a **single `opensearchpy.helpers.bulk` request**, followed by **one** `refresh()` to preserve immediate searchability. It mirrors the sibling [`elasticsearch.py`](https://github.com/mem0ai/mem0/blob/main/mem0/vector_stores/elasticsearch.py) adapter, which already uses `helpers.bulk`.

`N` index calls + `N` refreshes → **1 bulk + 1 refresh**.

`_id` is intentionally omitted from the bulk actions so OpenSearch keeps auto-assigning it — **document semantics are unchanged** from the previous `client.index(body=...)` (which also didn't set an explicit id); the mem0 id stays in `_source.id` exactly as before.

## Type of Change

- [x] Refactor (no functional changes) — performance optimization, behavior preserved

## Breaking Changes

N/A

## Test Coverage

`tests/vector_stores/test_opensearch.py` — **23 passed**. Updated `test_insert` to assert a single `bulk` call + a single `refresh` with the correct actions (and that `_id` is not set, so auto-assignment is preserved); added `test_insert_uses_one_bulk_and_one_refresh_regardless_of_count` (5 vectors → 1 bulk + 1 refresh); updated the error-path test to exercise a failing `bulk`. All other OpenSearch tests unchanged and green. `ruff check` clean.

## Checklist

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -156,6 +156,10 @@ class OpenSearchDB(VectorStoreBase):
             }
             for i, (vec, id_) in enumerate(zip(vectors, ids))
         ]
+        if not actions:
+            # Nothing to insert: skip the bulk + refresh round-trips entirely,
+            # matching the original per-document loop (which made zero calls).
+            return []
         try:
             # One bulk request instead of an index() call per document
             # (mirrors elasticsearch.py).
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, List, Optional
 
 try:
     from opensearchpy import OpenSearch, RequestsHttpConnection
+    from opensearchpy.helpers import bulk
 except ImportError:
     raise ImportError("OpenSearch requires extra dependencies. Install with `pip install opensearch-py`") from None
 
@@ -141,36 +142,38 @@ class OpenSearchDB(VectorStoreBase):
                     f"Ensure your embedding model's output dimensions match the vector store configuration."
                 )
 
-        results = []
-        for i, (vec, id_) in enumerate(zip(vectors, ids)):
-            body = {
-                "vector_field": vec,
-                "payload": payloads[i],
-                "id": id_,
+        actions = [
+            {
+                "_index": self.collection_name,
+                # ``_id`` is intentionally omitted so OpenSearch auto-assigns it,
+                # preserving the previous per-document ``client.index`` behavior;
+                # the mem0 id stays in ``_source`` like before.
+                "_source": {
+                    "vector_field": vec,
+                    "payload": payloads[i],
+                    "id": id_,
+                },
             }
-            try:
-                self.client.index(index=self.collection_name, body=body)
+            for i, (vec, id_) in enumerate(zip(vectors, ids))
+        ]
+        try:
+            # One bulk request instead of an index() call per document
+            # (mirrors elasticsearch.py).
+            bulk(self.client, actions)
+            # Refresh once after the full batch (not per document) and only when
+            # explicitly enabled. Disabled by default for Serverless compatibility:
+            # OpenSearch Serverless does not support the indices.refresh() API (#3893).
+            # See: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-genref.html
+            if self.auto_refresh:
+                self.client.indices.refresh(index=self.collection_name)
+        except Exception as e:
+            logger.error(f"Error inserting {len(actions)} vectors: {e}", exc_info=True)
+            raise
 
-                results.append(
-                    OutputData(
-                        id=id_,
-                        score=1.0,  # No score for inserts
-                        payload=payloads[i],
-                    )
-                )
-            except Exception as e:
-                logger.error(f"Error inserting vector {id_}: {e}", exc_info=True)
-                raise
-
-        # Refresh once after the full batch (not per document) if explicitly enabled.
-        # Disabled by default for Serverless compatibility: OpenSearch Serverless does not
-        # support the indices.refresh() API, and refreshing per document would cause a
-        # cluster-level I/O stall on every insert.
-        # See: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-genref.html
-        if self.auto_refresh:
-            self.client.indices.refresh(index=self.collection_name)
-
-        return results
+        return [
+            OutputData(id=id_, score=1.0, payload=payloads[i])  # No score for inserts
+            for i, id_ in enumerate(ids)
+        ]
 
     def search(
         self, query: str, vectors: List[float], top_k: int = 5, filters: Optional[Dict] = None
--- a/tests/vector_stores/test_opensearch.py
+++ b/tests/vector_stores/test_opensearch.py
@@ -152,27 +152,23 @@ class TestOpenSearchDB(unittest.TestCase):
         self.client_mock.indices.create.assert_not_called()
 
     def test_auto_refresh_disabled_by_default(self):
-        """Test that auto_refresh is disabled by default (Issue #3739).
-
-        This ensures OpenSearch Serverless compatibility out-of-the-box since
-        the indices.refresh() API is not supported in serverless mode.
+        """auto_refresh is off by default (#3739) for OpenSearch Serverless compat:
+        the indices.refresh() API is unsupported there, so the batched insert must
+        NOT refresh unless explicitly enabled.
         """
-        # Default instance should have auto_refresh=False
         self.assertFalse(self.os_db.auto_refresh)
         self.client_mock.reset_mock()
 
-        vectors = [[0.1] * 1536]
-        payloads = [{"key1": "value1"}]
-        ids = ["id1"]
+        with patch("mem0.vector_stores.opensearch.bulk") as mock_bulk:
+            self.os_db.insert(vectors=[[0.1] * 1536], payloads=[{"key1": "value1"}], ids=["id1"])
 
-        self.os_db.insert(vectors=vectors, payloads=payloads, ids=ids)
-
-        # Verify index was called but refresh was NOT called (default behavior)
-        self.assertEqual(self.client_mock.index.call_count, 1)
+        # The batch still goes out as one bulk request, but refresh is skipped.
+        mock_bulk.assert_called_once()
         self.client_mock.indices.refresh.assert_not_called()
 
-    def test_auto_refresh_enabled(self):
-        """Test that refresh is called once per batch (not per document) when auto_refresh=True."""
+    @patch("mem0.vector_stores.opensearch.bulk")
+    def test_auto_refresh_enabled(self, mock_bulk):
+        """When auto_refresh=True, refresh fires exactly once for the whole batch."""
         with patch("mem0.vector_stores.opensearch.OpenSearch", return_value=self.client_mock):
             auto_refresh_db = OpenSearchDB(
                 host="localhost",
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths

- **Correct bulk API shape** (`opensearch.py` new lines 142–172): the `_source` wrapper + `_index` key is the exact form `opensearchpy.helpers.bulk` expects. Auto-assigning `_id` by omitting it from actions is semantically correct — `delete()`, `update()`, and `get()` all look up by `_source[\"id\"]` via a `term` search, so they are unaffected by how OpenSearch assigns the document `_id`.
- **Precise mirror of elasticsearch.py** (`elasticsearch.py:127` vs new `opensearch.py:159`): import form, bulk call signature, and result construction are parallel — a maintainer touching both files will see consistent structure.
- **chunk_size is safe** by library default: `opensearchpy.helpers.bulk` calls `streaming_bulk` with `chunk_size=500` and `max_chunk_bytes=100 MB`, so a large batch is automatically split. No pathological single-request scenario.
- **Good inline comment** explaining the `_id` omission intent (new `opensearch.py` ~line 145) — this is exactly the kind of \"why\" comment that prevents a future reviewer from \"fixing\" it.
- **Tests are solid**: `test_insert_uses_one_bulk_and_one_refresh_regardless_of_count` directly encodes the O(1) contract; the `assertNotIn(\"_id\", action)` assertion pins the auto-assign semantics; `test_insert_error_logs_with_exc_info` correctly patches `bulk` to raise and asserts `exc_info=True` propagation.

### Issues

1. `mem0/vector_stores/opensearch.py`, new `insert()` — **empty-input guard missing**

   When `vectors=[]` is passed, `actions=[]` is built and then `bulk(client, [])` + `indices.refresh()` are called unconditionally — two unnecessary HTTP round-trips. The old per-document loop made zero network calls for empty input (the loop body never executed). In mem0's own call chain `Memory.add()` guards against this, but callers using `OpenSearchDB` directly can trigger the regression.

   Fix: add an early return before the try block:
   ```python
   if not actions:
       return []
   ```
   And add a test:
   ```python
   @patch(\"mem0.vector_stores.opensearch.bulk\")
   def test_insert_empty_vectors_makes_no_network_calls(self, mock_bulk):
       results = self.os_db.insert(vectors=[], payloads=[], ids=[])
       self.assertEqual(results, [])
       mock_bulk.assert_not_called()
       self.client_mock.indices.refresh.assert_not_called()
   ```

#### Minor (optional)

- `test_opensearch.py` lines 298, 310, 323 — `self.client_mock.index.assert_not_called()` in the null/empty/dimension-mismatch tests. These pass for the wrong reason now (the old `client.index()` path is gone entirely). They should assert `mock_bulk.assert_not_called()` instead, but since those tests only exercise the validation guard that raises *before* reaching `bulk`, adding `@patch(\"mem0.vector_stores.opensearch.bulk\")` and asserting it's not called would make the intent explicit. Low-stakes but improves test clarity.

### Recommendations

1. The `_id` identity divergence between opensearch.py and elasticsearch.py is pre-existing and not worsened by this PR, but it's worth a follow-up: ES uses `_id=mem0_uuid` enabling single-round-trip delete/update/get; OpenSearch auto-assigns `_id` and stores the mem0 UUID in `_source[\"id\"]`, forcing a search-then-act pattern (two round-trips). Unifying this in a follow-up issue would halve the overhead of those operations.

2. `raise_on_error=True` and `raise_on_exception=True` are the defaults for `opensearchpy.helpers.bulk` (confirmed in source) and are what the old per-document error handling was equivalent to. The sibling `elasticsearch.py` also doesn't pass them explicitly. No change needed — the error contract is correct.

### Evidence

- **CI:** Python SDK on 3.10 / 3.11 / 3.12 — PASS. CI Gate — PASS. Only Vercel (deploy-preview, advisory, no-perms-from-fork) fails — not a required check, ignored.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_opensearch.py -v --tb=short` → **23 passed in 25.74s** (PR head)
- **Red-green proof:**
  - PR head: 23/23 PASS ✓
  - Reverted `opensearch.py` to `origin/main` (tests unchanged): **3 FAILED** (`test_insert`, `test_insert_error_logs_with_exc_info`, `test_insert_uses_one_bulk_and_one_refresh_regardless_of_count` — `AttributeError: module has no attribute 'bulk'`) ✓
  - Restored to PR head: 23/23 PASS ✓
- **Lint:** `ruff check mem0/vector_stores/opensearch.py tests/vector_stores/test_opensearch.py` → clean ✓
- **Linked issue:** none — see tracking-issue nudge below
- **Cross-package check:** No TypeScript OpenSearch provider exists; TS SDK is unaffected.
- **Could NOT verify:** live OpenSearch behavior (requires a running cluster) — verified via code tracing only.

### Assessment

Ready to merge: **With fixes** — the Important item (empty-input guard) is a small, self-contained addition. Once that's in with a test, this is ready. Great performance improvement with clean tests.

kartik-mem0:
Verified against 2cc78a09: the empty-input guard landed correctly.

**Fixed (confirmed):**
- `mem0/vector_stores/opensearch.py` — `if not actions: return []` is present immediately after the actions list is built, before the try block, with a clear comment. ✓
- `test_insert_empty_vectors_makes_no_network_calls` added: asserts `mock_bulk.assert_not_called()` and `self.client_mock.indices.refresh.assert_not_called()`. ✓
- Red-green proof (this run): test FAILS on prior head b389e844 (`bulk` called with empty actions), PASSES on new head 2cc78a09. ✓
- All 24 tests pass: `pytest tests/vector_stores/test_opensearch.py` → **24 passed in 21.98s** ✓
- Lint: `ruff check` clean ✓

**New blocker — rebase required + `auto_refresh` regression:**

PR #3893 (merged to main 2026-06-23, commit 3497f26a) added `auto_refresh: bool = False` to `OpenSearchConfig` specifically for OpenSearch Serverless compatibility — Serverless does not support `indices.refresh()`, so refresh is now opt-in (disabled by default).

This PR's new bulk path calls `self.client.indices.refresh(index=self.collection_name)` unconditionally — no `if self.auto_refresh:` guard — which re-breaks Serverless users who rely on the default.

Fix is two lines. After rebasing on main:

```python
bulk(self.client, actions)
if self.auto_refresh:
    self.client.indices.refresh(index=self.collection_name)
```

And update the tests that assert `indices.refresh.assert_called_once_with(...)` to configure the `OpenSearchDB` with `auto_refresh=True` (or assert refresh is NOT called when `auto_refresh=False`).

The bulk optimization itself is correct and the empty-input guard is solid — this is a rebase + two-line guard away from being ready.

kartik-mem0:
Verified against 6445ac1b: both blockers from the last review are fixed.

**Fixed (confirmed):**
- `mem0/vector_stores/opensearch.py:170` — \`if self.auto_refresh:\` guard is present; \`indices.refresh()\` is no longer called unconditionally. Red-green: reverted impl to b389e844 (pre-PR) → \`test_auto_refresh_disabled_by_default\` FAIL, \`test_auto_refresh_enabled\` FAIL. Restored 6445ac1b → both PASS. ✓
- Rebase complete — \`mergeable: MERGEABLE\`, no conflict. ✓
- Empty-input guard (\`if not actions: return []\`) from rr1 — still present, \`test_insert_empty_vectors_makes_no_network_calls\` PASS. ✓

**Full test run:** \`hatch -e dev_py_3_12 run pytest tests/vector_stores/test_opensearch.py -v --tb=short\` → **26 passed in 24.81s** on new head.

**Lint:** \`ruff check mem0/vector_stores/opensearch.py tests/vector_stores/test_opensearch.py\` → All checks passed.

---

### What this PR does
Replaces N \`client.index()\` + N \`indices.refresh()\` calls in \`OpenSearchDB.insert()\` with a single \`opensearchpy.helpers.bulk()\` request followed by at most one \`indices.refresh()\` (gated on \`auto_refresh\`, preserving Serverless compatibility). Mirrors the existing \`elasticsearch.py\` pattern.

### Strengths
- \`opensearch.py:142–183\`: bulk action shape (\`_index\` + \`_source\`, no \`_id\`) is correct; mirrors the \`elasticsearch.py\` sibling exactly.
- \`opensearch.py:162\`: empty-input short-circuit avoids zero-doc bulk round-trips.
- \`opensearch.py:170\`: \`if self.auto_refresh:\` correctly gates refresh — Serverless users with \`auto_refresh=False\` (the default) are protected.
- Tests: \`test_insert_uses_one_bulk_regardless_of_count\` proves N→1 scaling; \`test_auto_refresh_disabled_by_default\` and \`test_auto_refresh_enabled\` prove the guard; \`test_insert_empty_vectors_makes_no_network_calls\` proves the short-circuit. All bite the pre-PR code.

### Issues
Minor: none.

### Evidence
- **CI:** CI Gate pass; Python SDK / build_mem0 (3.10, 3.11, 3.12) — all pass. Vercel advisory (fork perms), ignored.
- **Targeted tests:** \`hatch -e dev_py_3_12 run pytest tests/vector_stores/test_opensearch.py -v --tb=short\` → 26 passed in 24.81s
- **Red-green proof (auto_refresh guard):** reverted \`opensearch.py\` to b389e844 → 2 FAIL; restored 6445ac1b → 3 PASS. ✓
- **Linked issue:** no linked issue — standalone perf improvement.
- **Cross-package check:** OpenSearch provider is Python-only; no TS counterpart.
- **Could NOT verify:** live OpenSearch Serverless behavior — code-tracing only; the \`if self.auto_refresh:\` guard is the correct defence.

### Assessment

ly-wang19:
Filed #5958 and linked it above. Closes #5958.

kartik-mem0:
## Changes Requested

Fresh pass on `6445ac1b` found one more edge case after the previous blockers were fixed.

### Strengths

- `mem0/vector_stores/opensearch.py:145-172` keeps the optimization in the right provider-local layer: one `bulk()` call for the batch, and refresh remains gated by `auto_refresh` for Serverless compatibility.
- `tests/vector_stores/test_opensearch.py:230-251` covers the main performance contract (`N` vectors -> one bulk call) and the empty-input no-network-call path.
- Targeted OpenSearch tests and Ruff are green on this head.

### Issues

#### Important (should fix)

1. `mem0/vector_stores/opensearch.py:177-180` — results are built from every `id`, but the actual bulk actions are built from `zip(vectors, ids)` at `opensearch.py:145-158`. If a direct caller passes more `ids` than `vectors`, this PR submits only the zipped actions, then either raises `IndexError` after the bulk write when default payloads are shorter, or can report success for IDs that were never submitted when payloads are also longer. The old loop only returned results for the zipped `(vector, id)` pairs.

   Small fix: build the return value from the same zipped pairs used for `actions` (or explicitly validate equal lengths before any write if you want to tighten the contract). Please add a regression test for `ids` longer than `vectors` so the result count cannot drift from the submitted action count again.

#### Minor (optional)

None.

### Evidence

- **CI:** CI Gate pass; Python SDK 3.10 / 3.11 / 3.12 pass; CLA pass. Vercel deploy preview is the usual advisory auth failure and is ignored.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_opensearch.py -v --tb=short` -> `26 passed in 24.99s`; restored after red-green -> `26 passed in 24.62s`.
- **Red-green proof:** reverting only `mem0/vector_stores/opensearch.py` to `origin/main` with PR tests kept -> `9 failed, 17 passed in 24.06s`, so the bulk/refresh regression tests bite.
- **Lint:** `hatch -e dev_py_3_12 run ruff check mem0/vector_stores/opensearch.py tests/vector_stores/test_opensearch.py` -> `All checks passed!`
- **Edge-case probe:** on PR head, `insert(vectors=[[0.1, 0.2]], payloads=None, ids=["id1", "id2"])` calls `bulk` with 1 action, then raises `IndexError`; on `origin/main`, the same input returns 1 result after 1 `index()` call.
- **Linked issue:** #5958 — this PR addresses the tracked OpenSearch insert hot path.
- **Cross-package check:** no TypeScript OpenSearch vector store exists under `mem0-ts/src`; Python-only change.
- **Could NOT verify:** live OpenSearch / OpenSearch Serverless behavior; verified via unit tests and code tracing only.

### Assessment

Ready to merge: **With fixes** — the main optimization is good, but the return path needs to stay aligned with the actions actually sent to OpenSearch. Once that regression test and small return-path fix land, this should be ready again.

PR #5746, opened 2026-06-22
fix(milvus,baidu): sanitize filter values to prevent expression injection

### LINKED ISSUE #5744: security: filter expression injection in Milvus and Baidu vector stores
## Description

`_create_filter()` in both `MilvusDB` and `BaiduVectorDB` builds filter expressions via f-string interpolation without escaping or validating filter values. This allows an attacker who controls filter parameters to inject arbitrary filter expressions and bypass tenant isolation.

## Reproduction

```python
from mem0.vector_stores.milvus import MilvusDB

db = MilvusDB(...)
# Attacker passes a crafted filter value:
filters = {"user_id": '") or true or ("'}
expr = db._create_filter(filters)
# Result: (metadata["user_id"] == "") or true or ("") 
# This tautology matches ALL records, bypassing user_id filtering
```

The same pattern exists in `mem0/vector_stores/baidu.py` `_create_filter()`.

## Impact

In a multi-tenant deployment where `user_id` filters enforce data isolation, an attacker can read other users' memories by injecting filter expression syntax through the value parameter.

## Affected files

- `mem0/vector_stores/milvus.py` lines 146-162
- `mem0/vector_stores/baidu.py` lines 395-411

## Suggested fix

- Validate filter keys against a safe identifier regex
- Escape `"` and `\` in string values before interpolation
- Reject non-primitive types (dicts, complex objects) as filter values

HrushiYadav:
## Summary

- `_create_filter()` in both Milvus and Baidu vector stores builds filter expressions via f-string interpolation without escaping. A crafted filter value like `") or true or ("` produces a tautology that bypasses tenant isolation.
- Added key validation against a safe identifier regex, escape `"` and `\` in string values, and reject non-primitive types (dicts/lists) that could inject expressions.
- Identical fix applied to both `MilvusDB._create_filter()` and `BaiduDB._create_filter()` since they share the same vulnerable pattern.

## Test plan

- [x] 3 new security regression tests added for Milvus (injection payload, malicious key, quote escaping)
- [x] 3 new security regression tests added for Baidu (same coverage)
- [x] All 23 existing Milvus tests still pass
- [x] `ruff check` clean on all changed files

Fixes #5744

### DIFF
--- a/mem0/vector_stores/baidu.py
+++ b/mem0/vector_stores/baidu.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import time
 from typing import Dict, Optional
 
@@ -47,6 +48,8 @@ class OutputData(BaseModel):
 
 
 class BaiduDB(VectorStoreBase):
+    _SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
+
     def __init__(
         self,
         endpoint: str,
@@ -404,8 +407,16 @@ class BaiduDB(VectorStoreBase):
         """
         conditions = []
         for key, value in filters.items():
+            if not self._SAFE_FILTER_KEY.match(key):
+                raise ValueError(f"Invalid filter key: {key!r}")
             if isinstance(value, str):
-                conditions.append(f'metadata["{key}"] = "{value}"')
-            else:
+                escaped = value.replace("\\", "\\\\").replace('"', '\\"')
+                conditions.append(f'metadata["{key}"] = "{escaped}"')
+            elif isinstance(value, (int, float, bool)):
                 conditions.append(f'metadata["{key}"] = {value}')
+            else:
+                raise ValueError(
+                    f"Filter value for {key!r} must be str, int, float, or bool, "
+                    f"got {type(value).__name__}"
+                )
         return " AND ".join(conditions)
--- a/mem0/vector_stores/milvus.py
+++ b/mem0/vector_stores/milvus.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from typing import Dict, Optional
 
 from pydantic import BaseModel
@@ -143,6 +144,8 @@ class MilvusDB(VectorStoreBase):
         data = [_build_record(idx, embedding, metadata) for idx, embedding, metadata in zip(ids, vectors, payloads)]
         self.client.insert(collection_name=self.collection_name, data=data, **kwargs)
 
+    _SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
+
     def _create_filter(self, filters: dict):
         """Prepare filters for efficient query.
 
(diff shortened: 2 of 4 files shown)


Aditya-Tripuraneni (commented):
The current tests only assert that _create_filter() rejects bad inputs and escapes quotes in a couple of cases. I’d recommend adding:

1 exact-output test that covers both \ and " escaping in the same string, so we verify the final filter expression is safe and not just contains an escaped quote. I think a string such as   `value = r'alice\path"beta'` would be a good value test.

1 boolean test that proves True / False are rendered in the format the backend expects, since the code now explicitly allows booleans.

HrushiYadav:
Good suggestions @Aditya-Tripuraneni, added both:

1. **Exact-output test** -- uses `r'alice\path"beta'` and asserts the full escaped expression string, not just substring matches. Covers both `\` and `\"` escaping in one shot.

2. **Boolean test** -- verifies `True`/`False` render unquoted (e.g. `metadata["active"] == True`) as the backend expects.

Added to both Milvus and Baidu test suites. All Milvus tests pass locally (8/8 filter tests). Baidu tests can't run locally without `pymochow` but the logic is identical.

kartik-mem0:

### What this PR does
Closes the filter-expression injection in `MilvusDB._create_filter()` and `BaiduDB._create_filter()` (issue #5744). Both methods built Milvus/Baidu query expressions via f-string interpolation without escaping, so a crafted string value like `\") or true or (\"` produced a tautology that matched all records and bypassed tenant isolation. The fix adds key identifier validation, correct backslash-then-quote escaping of string values, and rejection of non-primitive types.

### Strengths
- `milvus.py:147` / `baidu.py:49`: `_SAFE_FILTER_KEY` compiled once as a class attribute — zero per-call overhead on the hot path.
- Escape order at `milvus.py:163` and `baidu.py:415` is correct: `\\` before `"`, so a pre-existing backslash doesn't cause the quote escape to break out.
- Dict/list values rejected with a clear `ValueError` — closes the MongoDB-style operator injection vector that was already fixed for other stores in recent PRs.
- 10 regression tests (5 per store) cover injection payload, malicious key, quote escaping, combined backslash+quote, and boolean rendering. All red without the fix, green with it.
- Identical fix applied symmetrically to both affected stores. No TS counterpart exists for these providers.

### Issues
Minor (optional): none.

### Evidence
- **CI:** Python SDK build_mem0 (3.10 / 3.11 / 3.12) PASS; CI Gate PASS; changelog_check PASS. Vercel = deploy perms on fork — advisory, ignored.
- **Targeted tests (green):** `pytest tests/vector_stores/test_milvus.py tests/vector_stores/test_baidu.py` → **40 passed in 30.68s** (all existing + 10 new).
- **Red-green proof:** reverted `milvus.py` + `baidu.py` to main; 5 regression tests → **5 failed** (DID NOT RAISE / unescaped quote in output). Restored fix → **5 passed**. Tests bite.
- **Lint:** `ruff check` on all 4 changed files → All checks passed.
- **Linked issue:** #5744 — PR addresses the exact reported root cause, both affected files.
- **Cross-package check:** No Milvus or Baidu provider in `mem0-ts/src/`; Python-only stores.
- **Could NOT verify:** Live Milvus/Baidu backend behavior — code tracing only. Milvus expression grammar documents double-quote strings with backslash escaping as the correct quoting convention.

### Assessment
The injection is closed at the only interpolation site, all callers are covered, tests are red-green proven, and no new abstractions or dependencies were introduced.

kartik-mem0 (approved):

PR #5750, opened 2026-06-22
fix(valkey): escape special chars in FT.SEARCH tag filter values

### LINKED ISSUE #5749: fix(valkey): filter values interpolated into FT.SEARCH queries without escaping
## Summary

`_build_search_query()` and `_build_list_query()` in the Valkey vector store interpolate filter values directly into FT.SEARCH query strings without escaping special characters. This allows filter values containing Valkey query syntax (wildcards, boolean operators, special chars) to alter the query semantics.

## Example

```python
# Intended: match user_id tag "alice"
filters = {"user_id": "alice"}
# Produces: @user_id:{alice}

# Attack: match ALL user_ids
filters = {"user_id": "*"}
# Produces: @user_id:{*}  -- wildcard matches everything
```

## Impact

An attacker who controls filter values can bypass tenant isolation by using Valkey FT.SEARCH wildcards or boolean operators to match records belonging to other users.

## Suggested fix

Escape Valkey FT.SEARCH special characters in tag filter values before interpolation.

HrushiYadav:
Closes #5749

## Problem

`_build_search_query()` and `_build_list_query()` in the Valkey vector store interpolate filter values directly into FT.SEARCH query strings without escaping. Special characters like `*` (wildcard) or `|` (OR) alter query semantics and can bypass tenant-isolation filters:

```python
# Intended: match user_id "alice"
filters = {"user_id": "alice"}
# Produces: @user_id:{alice}

# Attack: match ALL user_ids via wildcard
filters = {"user_id": "*"}
# Produces: @user_id:{*}  -- matches everything
```

## Fix

Add `_escape_tag_value()` that escapes FT.SEARCH special characters (wildcards, pipes, braces, etc.) with a backslash prefix. Applied in both `_build_search_query()` and `_build_list_query()`.

Normal alphanumeric values pass through unchanged, so existing behavior is preserved.

## Test plan

- `test_escape_tag_value_wildcards` -- verifies `*` and `|` are escaped
- `test_build_search_query_escapes_filter_values` -- verifies escaped output in search queries
- `test_build_list_query_escapes_filter_values` -- verifies escaped output in list queries
- `test_escape_tag_value_normal_strings` -- confirms normal values pass through unchanged
- All 58 existing Valkey tests pass

### DIFF
--- a/mem0/vector_stores/valkey.py
+++ b/mem0/vector_stores/valkey.py
@@ -342,8 +342,8 @@ class ValkeyDB(VectorStoreBase):
             knn_part (str): The KNN part of the query.
             filters (dict, optional): Filters to apply to the search. Each key-value pair
                 becomes a tag filter (@key:{value}). None values are ignored.
-                Values are used as-is (no validation) - wildcards, lists, etc. are
-                passed through literally to Valkey search. Multiple filters are
+                Values are escaped via _escape_tag_value() before interpolation
+                to prevent wildcard/operator injection. Multiple filters are
                 combined with AND logic (space-separated).
 
         Returns:
@@ -763,35 +763,6 @@ class ValkeyDB(VectorStoreBase):
             logger.exception(f"Error resetting index {self.collection_name}: {e}")
             raise
 
-    def _build_list_query(self, filters=None):
-        """
-        Build a query for listing vectors.
-
-        Args:
-            filters (dict, optional): Filters to apply to the list. Each key-value pair
-                becomes a tag filter (@key:{value}). None values are ignored.
-                Values are used as-is (no validation) - wildcards, lists, etc. are
-                passed through literally to Valkey search.
-
-        Returns:
-            str: The query string. Returns "*" if no valid filters provided.
-        """
-        # Default query
-        q = "*"
-
-        # Add filters if provided
-        if filters and any(value is not None for key, value in filters.items()):
-            filter_conditions = []
-            for key, value in filters.items():
-                if value is not None:
-                    escaped = self._escape_tag_value(value)
-                    filter_conditions.append(f"@{key}:{{{escaped}}}")
-
-            if filter_conditions:
-                q = " ".join(filter_conditions)
-
-        return q
-
     def list(self, filters: dict = None, top_k: int = None) -> list:
         """
         List all recent created memories from the vector store.
@@ -799,8 +770,8 @@ class ValkeyDB(VectorStoreBase):
         Args:
             filters (dict, optional): Filters to apply to the list. Each key-value pair
                 becomes a tag filter (@key:{value}). None values are ignored.
-                Values are used as-is without validation - wildcards, special characters,
-                lists, etc. are passed through literally to Valkey search.
+                Values are escaped via _escape_tag_value() before interpolation
+                to prevent wildcard/operator injection.
                 Multiple filters are combined with AND logic.
             top_k (int, optional): Maximum number of results to return. Defaults to 1000
                 if not specified.
--- a/mem0/vector_stores/valkey.py
+++ b/mem0/vector_stores/valkey.py
@@ -41,7 +41,20 @@ class OutputData(BaseModel):
     payload: Dict
 
 
+_VALKEY_TAG_SPECIAL = set(r',.<>{}[]"\':;!@#$%^&*()-+=~| ')
+
+
 class ValkeyDB(VectorStoreBase):
+    @staticmethod
+    def _escape_tag_value(value):
+        """Escape special characters in a Valkey FT.SEARCH tag filter value.
+
+        Without escaping, characters like * (wildcard) or | (OR) alter query
+        semantics and can bypass tenant-isolation filters.
+        """
+        s = str(value)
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Review: fix(valkey): escape special chars in FT.SEARCH tag filter values

**Verdict: Changes Requested** — The security fix itself is correct and complete on the live path. Two Important findings need to be addressed before merge.

---

### Strengths

- The root fix is sound. `_escape_tag_value()` is applied in `_build_search_query()`, which is the only method that `list()` and `search()` actually execute at runtime. Wildcard bypass (`*`) and OR injection (`|`) are both neutralized.
- `_VALKEY_TAG_SPECIAL` is the right set (28 chars, matching Valkey's documented FT.SEARCH special characters). Backslash and curly braces are included — escaping the escape character itself and closing-brace injection are both covered.
- Normal user IDs (`alice`, `user123`) pass through unchanged. No false-positive escaping.
- The `@{key}` side is correctly left unescaped — filter keys are always SDK-internal (`user_id`, `agent_id`, `run_id`), never user-controlled.
- Works identically in standalone and cluster mode (both route through the same `_build_search_query` codepath).
- No TS counterpart needed — Valkey is Python-only in this SDK; no parity gap.
- Ruff clean. Conventional commit title. No new core deps.

---

### Issues

**I-1 (Important) — Three docstrings still say "Values are used as-is (no validation) — wildcards … passed through literally"**

`mem0/vector_stores/valkey.py`:
- `_build_search_query` docstring (~line 332): _"Values are used as-is (no validation) - wildcards, lists, etc. are passed through literally to Valkey search."_
- `_build_list_query` docstring (~line 760): same text
- `list()` docstring (~line 788): _"Values are used as-is without validation - wildcards, special characters, lists, etc. are passed through literally to Valkey search."_

These directly contradict the fix. A contributor reading these docstrings would conclude escaping is absent and might not know to call `_escape_tag_value()` in a future query builder. Please update all three to say something like: _"Values are escaped via `_escape_tag_value()` before interpolation to prevent wildcard/operator injection."_

**I-2 (Important) — The fix to `_build_list_query` is unreachable — `list()` routes through `search()` → `_build_search_query()`**

The `list()` method delegates to `self.search()` (confirmed at line ~819: `search_results = self.search("", dummy_vector, ...)`), which in turn calls `_build_search_query`. `_build_list_query` is defined but never called in production.

The actual security fix lands correctly (it's in `_build_search_query`), but the fix to `_build_list_query` and the test `test_build_list_query_escapes_filter_values` both target dead code. This creates a misleading safety signal for future reviewers.

Suggested resolution: either (a) delete `_build_list_query` and its test (cleanest — the method was never wired up), or (b) add a comment on the method noting it is not currently called by `list()`, and update the test to call `list()` end-to-end instead.

---

### Minor (optional)

- No test for a hyphenated user ID like `user-123` (hyphen is in the special set, so it would become `user\-123`). Since Valkey TAG exact-match accepts escaped hyphens, this is a correct round-trip — but a test would make the intent explicit and prevent a future "my hyphenated user_id returns no results" regression report.

---

### Assessment

Ready to merge: **No — with fixes**

The core security fix (`_build_search_query` escaping) is correct and complete. Once the three stale docstrings are updated and `_build_list_query`'s dead-code status is addressed, this is mergeable. Both changes are small — no logic change required.

---

### Evidence

| Check | Result |
|---|---|
| CI Gate | PASS (Python SDK 3.10/3.11/3.12 pass) |
| Vercel | FAIL (advisory deploy check — fork PRs fail this by default; not merge-blocking) |
| Red-green (with fix) | 4/4 PASS |
| Red-green (pre-fix / main) | 4/4 FAIL (AttributeError + assertion failures confirming tests bite the change) |
| Full test suite | 58/58 PASS (54 existing + 4 new) |
| Ruff lint | All checks passed |
| TS counterpart check | No Valkey implementation in mem0-ts — no parity gap |
| Duplicate cluster | No other open PR for #5749 |

kartik-mem0:
Following up on the earlier changes-requested. Both findings are resolved — here's what I verified.

**Finding 1 — `_build_list_query` dead code:** Deleted entirely in this push, along with its test (`test_build_list_query`). Confirmed via grep: no definition or call site remains in either file.

**Finding 2 — Stale docstrings:** All three stale "Values are used as-is" passages are gone. `_build_search_query` (line 345) and `list()` (line 773) now both say "escaped via `_escape_tag_value()` before interpolation to prevent wildcard/operator injection." The third instance was in `_build_list_query` which is now deleted.

**New test:** `test_escape_tag_value_hyphenated_user_id` was added, covering the UUID/hyphen case from the earlier review note. Red-green verified locally on Python 3.12:
- 4 FAIL without the fix (AttributeError on `_escape_tag_value`, assertion error on unescaped `*`)  
- 4 PASS with the fix restored

Full suite: 57 passed (58 − 1 for the removed `test_build_list_query`). Ruff clean.

The core security fix — `_escape_tag_value()` applied in `_build_search_query()` against the complete Valkey FT special-char set — is unchanged and correct. No new concerns introduced.

**LGTM.** Ready to squash-merge.

kartik-mem0 (approved):

PR #5751, opened 2026-06-22
perf(qdrant): batch BM25 sparse encoding in insert() into one embed call

ly-wang19:
## Linked Issue

N/A — standalone performance improvement.

## Description

`Qdrant.insert()` encoded the BM25 sparse vector **one text at a time**: for each point it called `_encode_bm25(text)`, which runs `fastembed` `encoder.embed([single_text])`. So inserting `N` text-bearing memories triggered `N` separate model forward passes.

```python
for idx, vector in enumerate(vectors):
    ...
    if self._has_bm25_slot:
        text_for_bm25 = payload.get("text_lemmatized") or payload.get("data", "")
        if text_for_bm25:
            sparse = self._encode_bm25(text_for_bm25)   # one embed() call per point
```

The `upsert` itself was already a single batched call — only the encoding was per-point. Since `add()` typically extracts and inserts several facts at once (and bulk imports insert many), this is real, repeated CPU/inference waste on the write path.

This collects the text-bearing rows and encodes them in **one** `encoder.embed(texts)` call (fastembed batches internally), then maps each sparse vector back to its point by index.

- `_has_bm25_slot` gate, text-presence gate, and `None`-result handling are all **preserved** — the same points get the same sparse vectors.
- The single-text `_encode_bm25` is kept (still used by `search()` and `update()`).

`N` encode calls → **1**.

## Type of Change

- [x] Refactor (no functional changes) — performance optimization, behavior preserved

## Breaking Changes

N/A

## Test Coverage

`tests/vector_stores/test_qdrant.py` — **92 passed**. Added `test_insert_batch_encodes_bm25_in_a_single_call`: with a fake encoder, asserts `embed()` is called **exactly once** with only the text-bearing rows (in order), and that the resulting `bm25` sparse vector is attached to the right points (and omitted for the text-less row). All existing Qdrant tests unchanged and green. `ruff check` clean.

## Checklist

### DIFF
--- a/mem0/vector_stores/qdrant.py
+++ b/mem0/vector_stores/qdrant.py
@@ -122,6 +122,29 @@ class Qdrant(VectorStoreBase):
             logger.debug(f"BM25 encoding failed: {e}")
         return None
 
+    def _encode_bm25_batch(self, texts: list[str]) -> list[SparseVector | None]:
+        """Encode multiple texts into BM25 sparse vectors in a single batched call.
+
+        Returns one entry per input text, in order (``None`` where the encoder is
+        unavailable or a vector cannot be produced), so callers can zip the result
+        back to their inputs. fastembed batches the texts internally, so this is a
+        single forward pass instead of one ``embed()`` call per text.
+        """
+        encoder = self._get_bm25_encoder()
+        if encoder is None:
+            return [None] * len(texts)
+        try:
+            results = list(encoder.embed(texts))
+        except Exception as e:
+            logger.debug(f"BM25 batch encoding failed: {e}")
+            return [None] * len(texts)
+        encoded: list[SparseVector | None] = [
+            SparseVector(indices=sparse.indices.tolist(), values=sparse.values.tolist()) for sparse in results
+        ]
+        if len(encoded) < len(texts):
+            encoded.extend([None] * (len(texts) - len(encoded)))
+        return encoded
+
     def create_col(self, vector_size: int, on_disk: bool, distance: Distance = Distance.COSINE):
         """
         Create a new collection with dense vectors and BM25 sparse vectors.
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Review — perf(qdrant): batch BM25 sparse encoding in insert()

**Strengths**

The root-cause diagnosis is exactly right: the pre-PR code called `encoder.embed([single_text])` once per document inside the `insert()` loop, so N inserts triggered N separate fastembed forward passes. Batching them into a single `encoder.embed(texts)` call is the correct fix. The PR's test is well-structured — it verifies `embed()` is called exactly once, with only the text-bearing rows, in order, and that the BM25 vector lands on the right points (and is absent for the text-less row).

---

**Issues**

**Critical — already merged on main (PR is obsolete)**

This exact optimization landed on `main` as commit `7fa99626` (PR #5592, merged 2026-06-23 UTC+8) — one day after this PR was opened. `git log -- mem0/vector_stores/qdrant.py` confirms it as the most recent change to that file. Running the full Qdrant test suite on current `main` (`python3 -m pytest tests/vector_stores/test_qdrant.py`) produces **94 tests passed** — including a batch-encoding test — without any of this PR's changes applied.

That also explains the `CONFLICTING` merge status: `main`'s version of `insert()` has already been rewritten and the two patches can't both apply.

**Important — the merged implementation is also more robust**

The version that landed on `main` has three resilience properties this PR's `_encode_bm25_batch` lacks:

1. **Encoder-null guard before the batch call** — main checks `if encoder is not None` before calling `.embed()`; the PR's helper reaches `encoder.embed(texts)` via `_get_bm25_encoder()` only implicitly.
2. **Result-count mismatch check** — `fastembed`'s `embed()` is generator-based; if it yields fewer results than inputs, `main` logs a warning and raises so the fallback triggers. The PR silently maps the shorter list.
3. **Per-row fallback on batch exception** — if the batch call fails, `main` falls back to `_encode_bm25(text)` per row so a single bad input doesn't drop BM25 for the entire insert. The PR's except clause returns `[None] * len(texts)`, silently zeroing out BM25 for every row.

---

**Recommendations**

Nothing to do here code-wise — the fix is already on `main` and better for it. Please close this PR and pull the latest `main`; your instinct was correct and the batch approach is now the live code.

---

**Evidence**

| Check | Result |
|---|---|
| Real CI (Python SDK 3.10/3.11/3.12) | All PASS |
| Vercel deploy check | Failed (fork perms — advisory, ignored) |
| `git log -- mem0/vector_stores/qdrant.py` HEAD | `7fa99626 perf: batch BM25 sparse encoding in Qdrant insert (#5592)` (merged 2026-06-23) |
| `pytest tests/vector_stores/test_qdrant.py` on `main` (no PR changes) | 94 passed in 1.14 s |
| Mergeable | CONFLICTING |
| Red-green proof | Not run — PR is obsolete; test passes on main without PR changes |
| Worktree | Not created — conflict + already-fixed condition detected at Gate A |

ly-wang19:
Closing as obsolete — the batch optimization landed on `main` as #5592 (commit `7fa99626`, merged 2026-06-23) one day after this PR opened, with the additional encoder-null guard, result-count mismatch check, and per-row fallback called out in review. Confirmed `git log -- mem0/vector_stores/qdrant.py` shows that commit as HEAD. Thanks for the review.

PR #5755, opened 2026-06-23
Fix multilingual hybrid retrieval preprocessing

Jerry2003826:
## Summary
- Add dependency-free multilingual BM25 tokenization and route non-Latin/spaCy-unavailable lemmatization through it.
- Make spaCy retrieval models configurable/cache-safe and preserve normalized ID/SKU tokens in BM25 preprocessing.
- Add safe entity extraction fallbacks plus docs for multilingual hybrid retrieval setup and reindex/backfill caveats.

Addresses #4884.

## Test Plan
- `uv run --no-project --with-editable . --with pytest pytest tests/utils/test_text_tokenization.py tests/utils/test_lemmatization.py tests/utils/test_spacy_models.py tests/utils/test_entity_extraction.py tests/utils/test_scoring.py -q`
- `uv run --no-project --with-editable . --with pytest pytest tests/test_memory.py -q`
- `ruff check mem0/utils tests/utils tests/test_memory.py`
- `git diff --check origin/main...HEAD`


kartik-mem0:
Thanks for tackling issue #4884 — the root cause analysis and scope are well-targeted, and `text_tokenization.py` is genuinely solid work.

**Before this can merge it needs a rebase** — the PR is currently `CONFLICTING` and that conflict is non-trivial. More details below.

---

## Strengths

- `text_tokenization.py` is dependency-free, handles CJK/Thai/Arabic/Cyrillic/Hangul correctly, and the char-bigram approach for CJK BM25 is industry-standard. 350+ new test lines with good coverage of edge cases (extended CJK planes, fullwidth normalization, mixed identifiers).
- `spacy_models.py` refactor is an improvement: replacing the four `_nlp_full/_nlp_lemma/_load_failed_*` globals with a shared `_nlp_cache` dict keyed by `(model_name, disable)` is cleaner and enables env-var configurability that the community has been asking for.
- `lemmatize_for_bm25` non-Latin routing correctly short-circuits spaCy for CJK/Arabic/Cyrillic — red-green proven locally (reverted code → 2 tests fail; PR code → 2 pass).
- `extract_entities_batch` length-mismatch guards are a real robustness fix with clean mock-based tests.
- Ruff passes clean. No new core dependencies.

---

## Issues

**Critical — rebase required**

`entity_extraction.py` patches the old list-based architecture (blob `1c8f7f83c6`, ~400 lines). Since this PR was opened, PR #5829 ("align entity extraction precision", `b33fa542`) merged and completely rewrote that file: it's now 772 lines with `_EntityCandidate` frozen dataclass, `_ACCEPTED_NER_LABELS` / `_REJECTED_NER_LABELS` filtering, confidence-based `_resolve_candidates`, and `_add_ner_candidates` / `_add_topic_phrase_candidates` / `_add_technical_identifier_candidates`. The PR's diff does not apply to the current file; git flags it `CONFLICTING`.

Fortunately the intent of the `entity_extraction.py` changes is narrow and worth porting:

1. **noun_chunks guard** — wrap `for chunk in doc.noun_chunks:` inside `_add_topic_phrase_candidates` with `try/except (NotImplementedError, ValueError)`. This bug exists on main today (line 591) and will crash when any spaCy model without a dependency parser is used.
2. **Regex fallback extraction** — add `_extract_quoted_entities` (Unicode quote pairs) and `_CODE_ENTITY_RE` identifier extraction as a pre-step in `extract_entities`/`extract_entities_batch`, returning regex results when spaCy is unavailable.
3. **spaCy-unavailable path** — change `if nlp is None: return []` to return `_dedupe_and_clean_entities(_extract_regex_entities(text))` instead of an empty list.

The `text_tokenization.py`, `lemmatization.py`, and `spacy_models.py` hunks look like they'll rebase cleanly.

**Important — fix before merge**

- **Type-priority dedup ordering** (`entity_extraction.py:_dedupe_and_clean_entities`): The `seen` set pass runs before the `best` dict type-priority pass. `fallback_entities` is prepended to the combined list, so a `(QUOTED, "Apple")` from the regex fallback enters the seen-set before spaCy's `(PROPER, "Apple")` from NER. The PROPER entry is dropped in the first pass and never reaches the `type_pri` dict. For any named entity that appears in quotes, the extracted type is QUOTED instead of PROPER. Fix: collapse the two-pass dedup into the `best` dict alone (which already handles uniqueness per lowercase key), or move the seen-set dedup to run after the type-priority selection.

- **Mixed-script lemmatization** (`lemmatization.py:lemmatize_for_bm25`): When `contains_non_latin_letters(text)` returns True (fires on any single non-Latin char), the **entire** text is routed to `_fallback_tokenize` — the Latin words in the same text lose stop-word filtering and lemmatization entirely. A memory `"Alice 年糕汤 employees"` and a query `"Alice 年糕汤 employee"` won't match on "employees/employee" because the Latin words are raw-lowercased. Worth documenting if a full split-and-merge fix is deferred.

- **`_CODE_ENTITY_RE` false positives**: The second alternative `\b(?=[A-Za-z]*\d)(?=\d*[A-Za-z])[A-Za-z0-9]{3,}\b` matches common strings like `4th`, `Python3`, `B2B`, `G20`. On the pre-PR code those never became PROPER entities because spaCy NER correctly classified them. Now the regex extracts them first, and they survive into the merged entity list even when spaCy is installed. Narrowing the alternative to require a connector char (`-`, `_`, `.`, `:`) for short tokens would avoid the false positives.

**Coordination needed**

PR #5719 (open, same files) adds `spacy_model` as a function parameter to `get_nlp_full()` / `get_nlp_lemma()` and threads it through `MemoryConfig.spacy_model` in `main.py`. This PR keeps those functions zero-arg and reads the model name from env vars inside the loader. The two designs are incompatible function signatures — whichever merges second will need to reconcile. It would be worth aligning on the configuration surface (env-var only, MemoryConfig only, or env-var as process-level override of MemoryConfig default) before either lands.

**Minor**

- `_dedupe_and_clean_entities` is called twice per entity when spaCy is available: `_extract_quoted_entities` runs once in `_extract_regex_entities` (the fallback) and again inside `_extract_entities_from_doc`. The dedup handles the duplicates, but the extra extraction work scales with batch size.
- The two `contains_non_latin_letters` implementations diverge: `text_tokenization.py` uses `_LATIN_RANGES` (so `"café"` → False), while `entity_extraction.py/_has_entity_signal` uses `not ch.isascii()` (so `"café"` → True). Consider importing the one from `text_tokenization` rather than maintaining two.
- Docstrings on `extract_entities` and `extract_entities_batch` were shortened significantly — worth restoring the Args/Returns contract.

---

## Evidence

| Check | Result |
|---|---|
| CI (real checks) | No test/lint/build CI has run (only Vercel deploy preview — ignored) |
| CLA | Signed |
| Ruff | Clean |
| Targeted tests (worktree `3c588e87`) | 37 passed, 17 skipped (en_core_web_sm absent — expected), 0 failed |
| Red-green | FAIL on reverted lemmatization.py → PASS on PR version |
| Merge status | CONFLICTING |

---

## Path Forward

1. Rebase onto `main` — `text_tokenization.py`, `lemmatization.py`, `spacy_models.py` should land cleanly; surgically port the three `entity_extraction.py` additions onto the new `_EntityCandidate` architecture.
2. Fix the QUOTED-over-PROPER dedup ordering in `_dedupe_and_clean_entities`.
3. Document (or fix) the mixed-script lemmatization limitation.
4. Narrow `_CODE_ENTITY_RE`'s second alternative to avoid ordinal/version false positives.
5. Coordinate with PR #5719 on config surface.

The core work here (`text_tokenization.py` + `lemmatization.py` routing) is in good shape. Happy to re-review once the rebase is done — the entity_extraction port is the fiddly part.

Would you be able to open a tracking issue for this PR if there isn't one already? Linking to #4884 is helpful but a dedicated issue for the fix makes it easier to track progress across the rebase.

PR #5788, opened 2026-06-23
perf(entity-extraction): compile substring-filter regex once per entity (O(n²) → O(n))

ly-wang19:
## Linked Issue

N/A — standalone performance + testability improvement.

## Description

The final step of `_extract_entities_from_doc` drops entities that are whole-word substrings of another entity. It did so with an inline comprehension that **recompiled the boundary regex for every (entity, other) pair**:

```python
all_lower = [e[1].lower() for e in deduped]
return [
    (t, e)
    for t, e in deduped
    if not any(e.lower() != o and re.search(rf"\b{re.escape(e.lower())}\b", o) for o in all_lower)
]
```

For `n` entities that's **O(n²) `re.compile` calls** (every `re.search(pattern_string, ...)` recompiles), plus `e.lower()` recomputed on every inner iteration. This runs on the entity-extraction path used by `search()`, `add(infer=True)` (once per extracted memory), and `update()`.

This extracts the logic into a pure helper `_remove_whole_word_substring_entities` that **compiles each entity's boundary pattern once** (O(n) compilations) and hoists the lowercasing:

```python
lowered = [text.lower() for _, text in entities]
kept = []
for (etype, etext), needle in zip(entities, lowered):
    pattern = re.compile(rf"\b{re.escape(needle)}\b")
    if not any(needle != other and pattern.search(other) for other in lowered):
        kept.append((etype, etext))
return kept
```

Output is **identical** — same word-boundary semantics, same order. The extraction also makes this logic unit-testable without the spaCy model (the existing `_extract_entities_from_doc` tests are skipped when `en_core_web_sm` is unavailable).

## Type of Change

- [x] Refactor (no functional changes) — performance + testability, behavior preserved

## Breaking Changes

N/A

## Test Coverage

`tests/utils/test_entity_substring_filter.py` (new) — drops whole-word substrings, keeps non-boundary leading substrings (`Sam`/`Samsung`), case-insensitivity, empty/single, and a **parity test** asserting the helper returns exactly what the original inline O(n²) filter returned across several cases (incl. regex-special chars like `c++` and duplicate entities). 5 passed; `ruff check` clean. Existing `tests/utils/test_entity_extraction.py` unchanged (skips without spaCy, as before).

## Checklist

### DIFF
--- a/mem0/utils/entity_extraction.py
+++ b/mem0/utils/entity_extraction.py
@@ -174,6 +174,23 @@ def extract_entities_batch(texts: List[str], batch_size: int = 32) -> List[List[
     return results
 
 
+def _remove_whole_word_substring_entities(entities: List[Tuple[str, str]]) -> List[Tuple[str, str]]:
+    """Drop entities whose text is a whole-word substring of another entity.
+
+    Word-boundary anchoring keeps distinct entities that only share a leading
+    substring (e.g. "Sam" survives alongside "Samsung"). Each entity's boundary
+    pattern is compiled once -- O(n) compilations -- instead of recompiling the
+    pattern for every (entity, other) pair, which was O(n^2).
+    """
+    lowered = [text.lower() for _, text in entities]
+    kept: List[Tuple[str, str]] = []
+    for (etype, etext), needle in zip(entities, lowered):
+        pattern = re.compile(rf"\b{re.escape(needle)}\b")
+        if not any(needle != other and pattern.search(other) for other in lowered):
+            kept.append((etype, etext))
+    return kept
+
+
 def _extract_entities_from_doc(doc) -> List[Tuple[str, str]]:
     """Extract entities from a spaCy Doc object.
 
@@ -353,11 +370,4 @@ def _extract_entities_from_doc(doc) -> List[Tuple[str, str]]:
     deduped = list(best.values())
 
     # Remove entities that are whole-word substrings of longer entities.
-    # Word-boundary anchoring avoids dropping distinct entities that only share a
-    # leading substring (e.g. "Sam" must survive alongside "Samsung").
-    all_lower = [e[1].lower() for e in deduped]
-    return [
-        (t, e)
-        for t, e in deduped
-        if not any(e.lower() != o and re.search(rf"\b{re.escape(e.lower())}\b", o) for o in all_lower)
-    ]
+    return _remove_whole_word_substring_entities(deduped)
--- a/tests/utils/test_entity_substring_filter.py
+++ b/tests/utils/test_entity_substring_filter.py
@@ -1,0 +1,55 @@
+"""Unit tests for the whole-word substring entity filter.
+
+Pure-Python (no spaCy required): exercises ``_remove_whole_word_substring_entities``
+directly and pins it to the exact behavior of the original inline O(n^2) filter.
+"""
+
+import re
+
+from mem0.utils.entity_extraction import _remove_whole_word_substring_entities
+
+
+def _reference(entities):
+    """The original inline O(n^2) filter, kept as a parity oracle."""
+    all_lower = [e[1].lower() for e in entities]
+    return [
+        (t, e)
+        for t, e in entities
+        if not any(e.lower() != o and re.search(rf"\b{re.escape(e.lower())}\b", o) for o in all_lower)
+    ]
+
+
+def test_drops_whole_word_substring():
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Review

**Verdict: LGTM** — behavior-preserving perf refactor, thoroughly tested.

---

### Strengths

- **Semantics preserved exactly.** The `needle != other` self-exclusion in the new helper matches the original `e.lower() != o` comparison in every case — both sides iterate over the same pre-lowercased `lowered` list. Verified by tracing and by the `test_parity_with_reference_naive_filter` oracle, which keeps a verbatim copy of the original inline filter and asserts equality across five fixture families including regex-special chars (`c++`), leading-substring pairs (`Sam`/`Samsung`), and partial overlaps (`New York`/`York`/`New Yorker`).
- **`re.escape()` is correct throughout.** All needle text is escaped before pattern construction — no ReDoS risk, no regex injection.
- **Performance win is real.** For n entities, the original called `re.search(string_pattern, ...)` inside the inner generator — each call incurs Python's `re._cache` lookup overhead, and for n > 512 distinct entities, cache eviction causes silent recompilation. The new code makes exactly n `re.compile()` calls and routes n² matches through the pre-bound compiled object. Measured ~4.5–5× speedup on the full function for n = 30–100 entities, which is the realistic range on the `add()` / `search()` / `update()` paths where this function lives.
- **Test suite is well-designed.** The parity oracle approach (`_reference()` as a behavioral spec) is the right technique here — it pins the new implementation to the old one without depending on the reviewer's mental model. Edge cases covered: empty, single, case-insensitive, non-boundary leading substring, and the `c++` regex-special-char case.
- **Right abstraction level.** A module-level private helper with no spaCy dependency makes this logic independently testable — the pre-existing `_extract_entities_from_doc` tests skip without `en_core_web_sm`; these new tests run anywhere.

---

### Issues

#### Critical
None.

#### Important
None.

#### Minor

1. **`entity_extraction.py` ~line 178 — docstring omits "case-insensitive".**
   The current docstring says "Drop entities whose text is a whole-word substring of another entity" — a reader skimming the signature wouldn't know both sides are lowercased before matching. One word fixes it: "…a **case-insensitive** whole-word substring…"

2. **`test_entity_substring_filter.py` ~line 42 — `c++` word-boundary behavior deserves a comment.**
   The parity test includes `[("X", "c++"), ("Y", "c"), ("Z", "c++ guide")]`. The result is that `c` is dropped (because `\bc\b` matches `c` at the start of `c++`), while `c++` survives alongside `c++ guide` (because `\+` can't host `\b` — there's no word boundary after the `+` chars). This is correct parity, but the regex-boundary subtlety for symbols may confuse the next editor. A one-line comment would pay off.

---

### Recommendations

1. Apply the two Minor fixes above before merge — they're one-liners and improve long-term maintainability.
2. The TypeScript `entity_extraction.ts` has a pre-existing semantic difference (uses `String.includes()` instead of word-boundary regex — incorrectly drops `Sam` when `Samsung` is present). That predates this PR, but worth a follow-up tracking issue.

---

### Evidence block

| Check | Result |
|---|---|
| CI Gate | PASS |
| Python SDK 3.10 / 3.11 / 3.12 | PASS |
| Vercel | FAIL (advisory — deploy perms from fork; ignored) |
| Targeted tests (PR head) | 5/5 PASS |
| Lint (ruff) | clean |
| Red-green — PR head | 5/5 PASS |
| Red-green — revert impl | ImportError (tests fail — bites the change) |
| Red-green — restore | 5/5 PASS |
| Existing `test_entity_extraction.py` | 11 skipped (no spaCy — expected) |
| Behavior parity | Verified by `test_parity_with_reference_naive_filter` oracle across 5 fixture families |
| Danger flags | None |

Worktree used: `~/.cache/mem0-worktrees/review-pr-5788` at `a1ba08ad` (removed after review).

---

No linked issue is attached — if you have a moment, a short tracking issue would help future triage tooling connect this fix to related perf work.

kartik-mem0:
Thanks for the optimization and the clear write-up, @ly-wang19! 🙏

The whole-word-substring filter this PR targets was rewritten in #5829 (`fix(memory): align entity extraction precision`, merged). The entity extractor now resolves candidates through a normalized-text dictionary (`_resolve_candidates`) instead of the inline O(n²) regex comprehension, so the per-pair recompilation cost this PR addressed no longer exists in the current code path.

Closing as superseded by #5829 — but we really appreciate the catch and the clear analysis. If you spot a similar hotspot in the new extractor, a fresh PR would be very welcome!

PR #5790, opened 2026-06-23
feat(mem0-ts): add Baidu vector store provider

### LINKED ISSUE #5784: feat(ts-sdk): add Baidu vector store
## Summary

The Python SDK supports **Baidu** as a vector store provider, but the TypeScript OSS SDK (`mem0ai/oss`) does not. Add it to bring the TS SDK to parity.

| | |
|---|---|
| Python reference | `mem0/vector_stores/baidu.py` |
| Target file (TypeScript) | `mem0-ts/src/oss/src/vector_stores/baidu.ts` |
| Suggested implementation | Use `@baiducloud/sdk` for Baidu Mochow VectorDB (fall back to REST if the SDK lacks vector ops). |

## Requirements

- [ ] Implement `Baidu` in `mem0-ts/src/oss/src/vector_stores/baidu.ts`, extending `VectorStore` (`mem0-ts/src/oss/src/vector_stores/base.ts`) and mirroring the Python provider's behavior (insert / search / get / update / delete / list / reset).
- [ ] Register the `"baidu"` provider in `mem0-ts/src/oss/src/utils/factory.ts` (VectorStoreFactory).
- [ ] Add `@baiducloud/sdk` to `mem0-ts/package.json` (optional/peer dependency, lazy-imported like other providers).

## Reference pattern

## Notes

Mirror `mem0/vector_stores/baidu.py`.

---

Aditya-Tripuraneni:
## Linked Issue

Closes #5784

## Description
Adds the Baidu Mochow vector store to the TypeScript OSS SDK and fixes the BM25 keyword-search contract so it is backed by a real text index instead of an assumed one

### Bug fix citation

  Baidu’s official VectorDB docs show that BM25 search uses `BM25SearchRequest` with `index_name` / `searchText`, and that table schema supports
  `TEXT` fields plus `INVERTED` indexes for that kind of lookup:
  - https://cloud.baidu.com/doc/VDB/s/dlrsob0k0
  - https://cloud.baidu.com/doc/VDB/s/flrsob0zr

  Mem0’s OSS payload shape stores searchable text in `payload.data` and `payload.textLemmatized`, so the Baidu table now indexes those fields to make `keywordSearch()` contract backed.

## Notes

While implementing #5784, I found that the initial Baidu `keywordSearch()` path was only assuming BM25 support. The table schema did not create a full-text searchable text field or inverted index, so `keywordSearch()` would not be contract-complete on a real Baidu table.

I fixed that in this PR because it is part of making the Baidu provider actually work end-to-end for the issue being addressed. Splitting it into a separate PR would leave the new provider with a search path that appears supported in code but is not backed by the required index contract.

## Type of Change
- [ ] Documentation update

## Breaking Changes
N/A

## Test Coverage

- [] I added/updated integration tests

## Checklist


Aditya-Tripuraneni:
I’m considering a small follow up efactor to split `baidu.ts` into a `baidu_schema.ts` helper for table spec, schema validation, and normalization. I left it inline here to keep the PR focused, but I’m based on feedback I think it might be worth decomposing for readability purposes.

Aditya-Tripuraneni:
@HrushiYadav mind giving this a review when you get time?

Aditya-Tripuraneni:
@kartik-mem0 can I please get a review on this?

kartik-mem0:
## Changes Requested

Thanks for taking on Baidu vector store parity — this is a genuinely solid first pass. The auth wiring matches the Python provider almost line for line (`BceCredentials(account, apiKey)` → `Configuration({credentials, endpoint})` → `MochowClient(configuration)`), the filter-injection guard mirrors `mem0/vector_stores/baidu.py`'s `_create_filter` exactly, and the 606-line test suite covering both camelCase and snake_case SDK export shapes is a lot more thorough than most new-provider PRs bother with. Registration in `factory.ts`/`index.ts` is clean and matches the sibling pattern exactly.

### Strengths
- Auth model faithfully mirrors the Python reference — same three-object chain, same required fields (`endpoint`/`account`/`apiKey`).
- `buildFilter()` (`baidu.ts:711`) reuses the same `SAFE_FILTER_KEY` regex + string-escaping as Python's `_create_filter`, and it's actually exercised by an adversarial test case (a filter key like `'"] = "") or true or ("'` asserting `toThrow("Invalid filter key")`) — this is the kind of test that would catch a real regression.
- `loadSdkClient()` (`baidu.ts:487`) only does a dynamic `import("@baiducloud/sdk")`, and the top-level import is `import type` — so the package is never actually required unless a caller omits `config.client`. Nicer than a static top-level import for an optional peer dep.
- Docs update to `baidu.mdx` explicitly calls out the `metricType` default divergence (COSINE in TS vs L2 in Python) instead of burying it — good hygiene.
- `factory.ts:263-264` / `index.ts` registration matches every other vector store's pattern exactly, nothing to flag there.

### Issues

#### Critical (must fix)
- **CI is red on a real check, not just Vercel noise.** `mem0-ts/package.json` adds `@baiducloud/sdk: ^1.0.7` to `peerDependencies`, but `mem0-ts/pnpm-lock.yaml` was never regenerated/committed — `pnpm install --frozen-lockfile` fails with `ERR_PNPM_OUTDATED_LOCKFILE` on both `build_ts_sdk (20)` and `build_ts_sdk (22)`, which fails `CI Gate` too. Practical consequence: the build never reaches the Jest step, so the well-written `baidu.test.ts` suite has never actually run in CI — there's no green run proving it passes yet. Fix is mechanical: run `pnpm install` inside `mem0-ts/` and commit the updated lockfile.

#### Important (should fix)
- `validateExistingTableSchema()` (`baidu.ts:507`, throw at `baidu.ts:538-540`) throws when an existing Baidu table is missing the `data`/`textLemmatized` fields or the `data_bm25_idx` inverted index. That's a reasonable defensive check in principle, but it's currently untested: every mock table in `baidu.test.ts` (both the camelCase and snake_case variants, around lines 124 and 161) returns `stats()` as `{ tableName: "memories" }` with no `schema` key at all, so `validateExistingTableSchema` always takes the early-return no-op (`baidu.ts:513-514`) and the throw path never runs. Worth adding a test that returns a `schema` missing the BM25 pieces (asserts it throws) and one that's fully compliant (asserts it doesn't). This also has a real-world edge: `mem0/vector_stores/baidu.py`'s `create_col` never creates the `data`/`textLemmatized` fields or a BM25 index today, so any table originally provisioned by the Python SDK will hit this throw the first time the TS provider touches it. Might be worth a softer fallback (warn + disable `keywordSearch()`, like the MongoDB provider does for a missing search index) rather than a hard throw, since `insert`/`search`/etc. all route through `ensureTable()`.
- Related: `insert`/`search`/`get`/`update`/`delete`/`list`/`deleteCol` all call `this.ensureTable()` directly rather than `this.initialize()` (contrast with `mongodb.ts`, where every public method awaits the memoized `_initPromise` first). The already-exists fallback in `ensureDatabase()`/`ensureTable()` probably self-heals concurrent first calls, but right now the `_initPromise` gate only actually protects a caller who explicitly awaits `initialize()` — worth aligning with the sibling pattern so concurrent first calls share one in-flight create instead of each racing `createDatabase`/`createTable` independently.

#### Minor (optional)
- `metricType` defaults to `"COSINE"` (`baidu.ts:457`) vs. Python's config default of `"L2"` (`mem0/configs/vector_stores/baidu.py`). It's disclosed in the docs diff, so not hidden, but worth a second look to confirm the divergence is intentional rather than incidental — a team pointing both SDKs at "the same config" would get different similarity semantics by default.
- `loadSdkClient()` (`baidu.ts:488`) has no try/catch around the `import("@baiducloud/sdk")` call itself, so a user who hasn't installed the package gets a raw "Cannot find module" instead of the friendlier, actionable message Python's guarded import gives (`"The 'pymochow' library is required. Please install it using 'pip install pymochow'."`).

### Evidence
- **CI:** Red on real checks — `CI Gate` fail, `TypeScript SDK / build_ts_sdk (20)` fail, `TypeScript SDK / build_ts_sdk (22)` fail, all from the same `ERR_PNPM_OUTDATED_LOCKFILE` root cause. (Vercel is also failing but that's the preview-deploy auth wall, unrelated.)
- **Targeted tests / red-green:** Read `baidu.test.ts` in full — well-structured (idempotent-init assertions, already-exists fallback for both DB and table, full CRUD + reset lifecycle, snake_case/camelCase equivalence, filter-injection safety), but per the lockfile issue above it's never actually executed in CI, and the schema-validation throw path specifically has zero coverage regardless.
- **Could NOT verify:** Whether the missing table-ready-state polling (Python's `create_col` polls with `time.sleep(2)` waiting for `TableState.NORMAL`; this TS version doesn't appear to poll before first use) matters in practice — would need a live Baidu backend to confirm either way, flagging as a question rather than an asserted defect.

One more thing outside the code itself: GitHub currently shows this PR as having a merge conflict against `main`, so it'll need a rebase regardless of the above before it can land.

### Assessment
Ready to merge: **With fixes** — get CI green again (lockfile) and take a look at the schema-validation test gap, and I'm happy to take another pass. The bones of this are good and the parity work is appreciated.

Aditya-Tripuraneni:
@kartik-mem0 , should be good for 2nd review now, CI passes and rebased. 

I also changed the TS metric to just be L2 to stay consistent with Python, this was a good call.

kartik-mem0:
Thanks for the thorough work here, and for turning around every item from the last round: the lockfile is regenerated (`pnpm install --frozen-lockfile` is clean), `metricType` now defaults to `L2` matching the Python SDK, the dynamic `import()` is guarded with an actionable message, and `validateExistingTableSchema()` degrades to `supportsKeywordSearch = false` instead of hard-throwing. All four were fixed properly.

I need to lead with something that is my fault, not yours.

**Issue #5784, which I wrote, says: _"Suggested implementation: Use `@baiducloud/sdk` for Baidu Mochow VectorDB."_** That suggestion was wrong. `@baiducloud/sdk` is Baidu's general Cloud Engine SDK (`baidubce/bce-sdk-js`) and has nothing to do with Mochow or vector search. You followed the issue faithfully and built ~1,800 lines on a premise I handed you. I'm sorry, and I want to be straight about the consequence: this needs to be re-pointed at a different SDK, and that is a real rewrite rather than a rename.

Everything below was verified against the published packages and by running your provider, not read off the diff.

---

## Critical

### C1. `@baiducloud/sdk` exports no Mochow client, so the documented path throws for every user

`loadSdkClient()` (`baidu.ts:242-271`) requires `Configuration`, `BceCredentials`, and `MochowClient`. The real package exports none of them:

```
$ node -e "console.log(Object.keys(require('@baiducloud/sdk')))"
[ 'Q','version','crypto','strings','STS','Auth','MimeType','Base64','HttpClient',
  'BceBaseClient','BosClient','BcsClient','BccClient','SesClient','QnsClient',
  'LssClient','MctClient','FaceClient','OCRClient','MediaClient','VodClient',
  'DocClient','TsdbDataClient','TsdbAdminClient','CfcClient','BtsClient','IoTClient' ]
```

I checked `1.0.7` (the only version resolvable under the declared `^1.0.7`, since semver ranges exclude prereleases) and all seven `1.0.8-beta.1` … `beta.7` tarballs. `grep -ril mochow` returns zero hits in every one of them, as does `grep -ril BceCredentials`. There is no `exports` subpath map either, so `require("@baiducloud/sdk/mochow")` is not an escape route.

Running your provider with exactly the config the docs tell users to write, with no `jest.mock` in the way:

```
Error: The @baiducloud/sdk package does not expose the expected Mochow client exports.
    at loadSdkClient  (mem0-ts/src/oss/src/vector_stores/baidu.ts:259)
    at ensureClient   (baidu.ts:275)
    at ensureDatabase (baidu.ts:504)
    at ensureTable    (baidu.ts:522)
```

**Re-pointing the dependency is necessary but not sufficient.** The real client is [`@mochow/mochow-sdk-node`](https://github.com/baidu/mochow-sdk-node) (`2.1.5`, published by `vdb@baidu.com`). Its shape differs from this PR's in three ways that reach into your control flow:

1. **No `Configuration` or `BceCredentials` classes.** Auth is a plain object literal:
   ```js
   new MochowClient({ endpoint, credential: { account, apiKey } })
   ```
2. **No `Database` or `Table` objects.** In that package `Database`/`Table` are internal class mixins composed into the single client, not values you retrieve. There is no `client.database(name)`. Every operation is flat on the client, taking the names as string fields: `upsert({database, table, rows})`, `vectorSearch({database, table, request})`, `bm25Search({database, table, request})`, `descTable(database, table)`, `showTableStats(database, table)`.
3. **Mutations resolve a `{code, msg}` envelope rather than throwing.** `createDatabase(name): Promise<CommonResponse>`; the README's own idiom is `if (resp.code != 0) { ... }`. So the `catch (isAlreadyExistsError)` recovery in `ensureDatabase()` / `ensureTable()` (`baidu.ts:499-539`) would never fire against a real client. `this.database` would be assigned the response envelope, and the next `getMethod(database, ["createTable", "create_table"])` would throw `Baidu SDK is missing expected method(s)`.

The `client.database() → db.createTable() → table.upsert()` chain in this PR is **pymochow's Python object model**. It is genuinely real in `mem0/vector_stores/baidu.py` (which imports `Configuration` and `BceCredentials` from `pymochow` and calls `client.database(name)`). It exists in no Node package.

### C2. The ambient type declaration and the test mocks hide C1 from both gates

`mem0-ts/src/oss/src/types/baidu-sdk.d.ts` (126 lines, new) is a hand-written `declare module "@baiducloud/sdk"` describing the invented surface. It is the only hand-written `declare module` in the whole SDK:

```
$ grep -rn "declare module" mem0-ts/src/
mem0-ts/src/oss/src/types/baidu-sdk.d.ts:1:declare module "@baiducloud/sdk" {
```

TypeScript trusts ambient declarations unconditionally, so `tsc --noEmit` reports zero Baidu errors. `baidu.test.ts:27` then `jest.mock`s the same fiction, using the lowercase names (`configuration`, `bceCredentials`, `mochowClient`) that `loadSdkClient()`'s `?? sdk.configuration` fallbacks accept. The result is 18/18 green tests validating a mock of a package that does not exist. Neither gate can fail, so neither gate carries signal.

Two things in the implementation are downstream symptoms of the same cause, and both should disappear once real types are in place:

- **`getMethod()` name-probing** (`baidu.ts:175`, 9 call sites) duck-types across candidate spellings (`createDatabase`/`create_database`, `describeTable`/`describe_table`/`table`, `bm25Search`/`bm25_search`, …) and throws only if none match. This directly contradicts the `.d.ts` sitting beside it, which asserts the exact names exist. If the types are trusted, call the method; if they aren't, the types shouldn't exist.
- **Dual camelCase + snake_case key emission.** `buildTableSpec()` doubles nearly every key, e.g. `baidu.ts:398`:
  ```ts
  partition: { partitionNum: 1, partition_num: 1 },
  ```
  Also `tableName`/`table_name`, `primaryKey`/`primary_key`, `indexName`/`index_name`, `autoBuildIndexPolicy`/`auto_build_index_policy`, `vectorField`/`vector_field`, `searchText`/`search_text`. No other provider does this. `milvus.ts` uses snake_case only, because the real Milvus SDK requires snake_case. A server with strict payload validation would reject this shape outright; a lenient one leaves "which key wins" unanswerable.

### C3. `deleteCol()` doesn't wait for the in-flight `initialize()`, so the table is dropped and never recreated

This one is independent of C1 and will still be there after the rewrite, so it's worth fixing on its own terms.

The constructor kicks off `initialize()` fire-and-forget (`baidu.ts:239`). When that chain finally resolves, `ensureTable()` assigns `this.table = resolvedTable` (`baidu.ts:537`). `deleteCol()` (`baidu.ts:751-773`) never awaits `this._initPromise` — it drops the table, sets `this.table = null`, and clears `_initPromise`. The older chain is still pending, and promises aren't cancellable, so it lands afterwards and puts the handle back. The next operation calls `ensureTable()`, hits the `if (this.table) return this.table` short-circuit at `baidu.ts:518`, and reuses a handle to a table that no longer exists.

I reproduced this with an injected `config.client` (no `jest.mock`, no network), timing `createTable` slower than `dropTable`:

```
A) right after deleteCol():        table = null
B) after in-flight init resolves:  table = <HANDLE>     <-- resurrected
C) events after insert():          dropTable,createTable,upsert
C) createTable calls total:        1
```

`createTable` ran **once** — from the constructor, before the drop. After `deleteCol()` the table is gone from the server, but `insert()` happily upserts through the stale handle instead of recreating it. `reset()` (`baidu.ts:775-778`) is the most likely way a user reaches this, since it's `deleteCol()` immediately followed by `initialize()`.

Fix: `await this._initPromise?.catch(() => {})` at the top of `deleteCol()` before touching any state — you're tearing down anyway, so swallowing the rejection is fine.

---

## Important

### I1. Duplicate `peerDependenciesMeta` key silently un-marked two unrelated peer deps as optional  — *fixed, see below*

`mem0-ts/package.json` ended up with two top-level `"peerDependenciesMeta"` objects (lines 146 and 154). JSON keeps the last duplicate, so what npm/pnpm actually read was:

```
$ node -e 'console.log(JSON.stringify(JSON.parse(require("fs").readFileSync("package.json","utf8")).peerDependenciesMeta))'
{"@baiducloud/sdk":{"optional":true}}
```

`mysql2` and `@zilliz/milvus2-sdk-node` lost their `optional: true`, which they had before this PR. That is a regression for consumers of two already-shipped providers who install neither. I've pushed the merge to your branch.

### I2. Every row stores the same text twice

`insert()` (`baidu.ts:632-643`) writes `metadata: payloads[index] || {}` — the whole payload, which already contains `data` and `textLemmatized` — while `buildSearchTextFields()` (`baidu.ts:141-153`) extracts those same two values into top-level `data` / `textLemmatized` columns. Nothing strips the duplicates from `metadata` first.

Promoting text into real columns so the INVERTED index can reach it is correct (you can't index into an opaque JSON blob). Deleting the two keys from the `metadata` copy before writing it would keep the behavior and halve the stored text. `update()` (`baidu.ts:730-739`) has the same shape.

### I3. TS and Python now provision incompatible schemas for the same `(database, table)`

- Python `create_col()` (`mem0/vector_stores/baidu.py:136-156`): fields `id`, `vector`, `metadata`. Indexes `vector_idx` (HNSW) and `metadata_filtering_idx`.
- TS `buildTableSpec()` (`baidu.ts:391-471`): the above **plus** `data` and `textLemmatized` TEXT fields and a `data_bm25_idx` INVERTED index.

If Python creates the table first, `validateExistingTableSchema()` sets `supportsKeywordSearch = false`, emits one `console.warn` on cold start, and permanently disables `keywordSearch()` for that process. Nothing surfaces to the caller.

**To be clear, this one is our bug, not yours, and your side is the more correct one.** Python's `keyword_search()` queries `index_name="data_bm25_idx"` (`baidu.py:252`) — an index its own `create_col()` never creates. That's the only occurrence of the name in the file. Python's keyword search cannot have ever worked; the broad `except Exception: return None` swallows it into a silent `None`. I'll open a separate issue to fix `create_col()` on the Python side so the two SDKs converge. Please don't take that on inside this PR.

### I4. No table-ready or table-deleted polling, which Python has

Python waits for the table to reach `TableState.NORMAL` after create (`baidu.py:167-174`) and polls until `TABLE_NOT_EXIST` after drop (`baidu.py:336-347`). Mochow provisions tables asynchronously — that's *why* those loops exist. TS's `createTable()` (`baidu.ts:295-300`) and `dropTable()` (`baidu.ts:310-316`) fire once and return, and `reset()` (`baidu.ts:775-778`) calls `deleteCol()` then `initialize()` back to back. Against a real backend, an `insert()` immediately after construction, or a `reset()` racing its own deletion, can plausibly hit "table not ready". The mocked SDK resolves synchronously, so no test can catch it.

### I5. `validateExistingTableSchema()` fails open, so the BM25 safety net probably never fires

The function exists to detect a pre-existing table without BM25 support. Both of its escape hatches leave `supportsKeywordSearch` at its default `true` (`baidu.ts:201`):

- the `catch` at `baidu.ts:548-554` warns and returns;
- the shape guard at `baidu.ts:556-558` — `if (!isRecord(stats) || !isRecord(stats.schema)) return;` — returns silently, with no warning at all.

So any unexpected response shape, and any transient network or auth blip during introspection, is read as "BM25 is present." A check whose whole purpose is to detect an absent index should default to *absent* and let `keywordSearch()` re-enable itself, not the reverse.

It also looks like it's introspecting the wrong object. It reads the schema out of `table.stats()` (`baidu.ts:544-547`), but in `pymochow` — the library this was ported from — `stats()` is a separate call against its own `RequestType.STATS` endpoint, while the schema is a constructor field on the `Table` object with its own `.schema` property. Our own Python `col_info()` treats `stats()` as table statistics, not schema (`baidu.py:360`). If Node mirrors that, `stats.schema` is never a record and the guard at 556 returns every single time, making the entire check dead code. I can't confirm the Node response shape while C1 stands, so treat this as strongly suspected rather than proven. The schema almost certainly hangs off the object `getExistingTable()` already returned.

---

## Minor

- **`String(rowData.id)` can emit the literal string `"undefined"` as a memory ID.** `search()` (`baidu.ts:694`) and `list()` (`baidu.ts:793`) do `const rowData = firstRow(row) ?? row;`. The `?? row` fallback is right for a flat row shape, but `toRow()` (`baidu.ts:85-91`) also returns `null` whenever `id` isn't a `string` — and for a *nested* row shape the fallback then hands back the outer envelope, whose own `.id` is `undefined`. `String(undefined)` is `"undefined"`, which flows out as a memory ID. `keywordSearch()` (`baidu.ts:666`) has the same pattern. `get()` (`baidu.ts:712-714`) does the right thing and returns `null`. Worth making the three read-paths agree with `get()`. Low likelihood while the schema pins `id` to `STRING`, but it's free to fix.
- **`list()`'s total can undercount.** When the response carries no numeric `.total`, it falls back to `memories.length` (`baidu.ts:800-803`), which is the current page, capped at `topK`. Callers reading the second tuple element for pagination would get a wrong total.
- **Boolean filter values may not match Python.** `buildFilter()` (`baidu.ts:486-489`) interpolates JS booleans as `true`/`false`; Python's f-string yields `True`/`False` (`baidu.py:415-416`). Mochow's filter grammar isn't public, so I can't say which is correct — but they can't both be. Worth confirming against a live instance; lowercase is the more common convention, so this may be a Python bug rather than a TS one.
- **`table_name` doc default was wrong.** The config table said `mem0_table`; the actual default is `mem0` (`mem0/configs/vector_stores/baidu.py:11`), and TS has no default at all — `tableName` is required (`baidu.ts:219`). Looks like the example value was copied into the "Default Value" column. *Fixed, see below.*
- **Stray edit to `mem0/vector_stores/baidu.py`** in an otherwise TS-only PR: one blank line removed between `keyword_search()` and `delete()`. *Reverted, see below.* (For the record, it did not break `ruff format` — that file already deviates on `main` for an unrelated line-split, so nothing was masked.)
- **`initialize()` + `ensureTable()` are both awaited in every public method** (`insert`, `search`, `get`, `update`, `delete`, `list`, `keywordSearch`). `initialize()` already awaits `ensureTable()` and caches it, so the second call short-circuits. Harmless, just redundant. (The single-flight `_initPromise` itself is right — it clears the cached promise on rejection so a transient failure can be retried. Nice.)
- **`insert()` could batch.** The `upsert({ rows })` signature already takes an array; you send one row per round trip. Worth doing when you rewrite — but note this is **not** a regression you introduced: `mem0/vector_stores/baidu.py:188-190` loops per row too.

---

## What I pushed to your branch

Since these three are mechanical, verifiable, and independent of the SDK question, I pushed them rather than making you round-trip (commit `2d9c68b`):

| File | Change |
|---|---|
| `mem0-ts/package.json` | merged the two `peerDependenciesMeta` blocks so `mysql2`, `@zilliz/milvus2-sdk-node`, and `@baiducloud/sdk` are all optional again |
| `docs/components/vectordbs/dbs/baidu.mdx` | `table_name` default `mem0_table` → `mem0` |
| `mem0/vector_stores/baidu.py` | restored the blank line; the file now matches `main` byte for byte |

Verified after the change: `pnpm install --frozen-lockfile` clean, `jest baidu.test.ts factory.unit.test.ts` 72/72, `prettier --check` clean.

I deliberately did **not** attempt the SDK rewrite for you. It touches most of the 814-line provider and its whole test suite, and I can't verify it without a live Mochow instance — pushing an unverified rewrite under your name would be worse than useless.

---

## Recommendations

1. **Swap the peer dependency to `@mochow/mochow-sdk-node`** and delete `baidu-sdk.d.ts` entirely — that package ships its own types (`MochowClient.d.ts`, `types/*.d.ts`). Letting the compiler check against the vendor's real types is what makes the rest of this list mostly delete itself.
2. **Delete `getMethod()` / `hasMethod()` and every candidate-name list.** With real types there is exactly one correct method name.
3. **Delete every snake_case twin key.** Emit exactly what the real request types declare.
4. **Restructure around the flat client.** `client.database(name)` and the chained `Database`/`Table` objects go away; pass `{database, table}` into each call. Replace `try/catch (alreadyExists)` with `if (resp.code !== 0)` checks, since the real client resolves rather than throws.
5. **Port Python's create/drop polling loops** (I4).
6. **Rewrite the tests against the real types**, and add one integration test gated on live credentials, matching this repo's existing `test:integration` convention — so CI carries at least one signal that isn't self-referential.
7. Strip the duplicate `data` / `textLemmatized` keys from `metadata` before writing (I2).
8. Independent of the SDK swap, and cheap: await `_initPromise` in `deleteCol()` (C3), flip `validateExistingTableSchema()` to fail closed (I5), and make the three read-paths handle an unparseable row the way `get()` does.

One thing I want to call out as a strength rather than bury it: `insert()`'s length-mismatch warning (`baidu.ts:621-630`) is better than the Python reference, which silently truncates via `zip()`. Don't lose it in the rewrite. (Throwing instead of warning would be defensible; returning a count isn't an option, since `VectorStore.insert()` is `Promise<void>` for all thirty providers.)

If `@mochow/mochow-sdk-node` turns out not to work for you, the honest fallback is to drop the auto-load path, make `config.client` required, and document that callers construct their own client. `config.client` is already wired (`types/index.ts:32` → `factory.ts:141` → `baidu.ts:213`); today it's mentioned in one prose line in the docs and is absent from the config table. That's a smaller PR and it would land.

---

## Evidence

```
head reviewed:  5fda07e0424374eb3bf75be64a3633142c7dcdf8  (+ 2d9c68b pushed by me)
worktree:       detached at 5fda07e0

$ pnpm install --frozen-lockfile
Done in 6s                                         # exit 0

$ npx jest src/oss/tests/baidu.test.ts
Tests:       18 passed, 18 total                   # against a mocked, nonexistent package

$ npx tsc --noEmit | grep -ci baidu
0                                                  # ambient .d.ts makes this vacuous
                                                   # 26 pre-existing errors, none Baidu: 11 in
                                                   # community/integrations/langchain/mem0.ts, 15 in client/tests/*

$ npx prettier --check <changed .ts files>
All matched files use Prettier code style!

$ node -e "console.log(Object.keys(require('@baiducloud/sdk')))"
[ ... 27 exports, none of them Mochow ... ]

$ for v in 1.0.7 1.0.8-beta.{1..7}; do grep -ril mochow extract-$v/ ; done
(no output — zero hits, all 8 versions)

$ # C3, injected config.client, no jest.mock, createTable slower than dropTable
after deleteCol(): table = null
after in-flight init resolves: table = <HANDLE>
events: dropTable,createTable,upsert   # createTable ran once, before the drop

CI: real checks green. Vercel preview-deploy check is red; ignored (it fails on
essentially every fork PR and is not merge-blocking).
```

Docs check: `docs/llms.txt:490` and `docs/docs.json:223` both register the page correctly, `scripts/check-llms-txt-coverage.py` reports in sync, and the page is clean on house style. The remaining gap is that it has no install instruction, while its siblings do (`weaviate.mdx:15` → `npm install weaviate-client`, `milvus.mdx:12` → `npm install @zilliz/milvus2-sdk-node`). I've deliberately held off adding one, because it must not say `npm install @baiducloud/sdk`: that path provably throws. Add it in the same change that resolves C1, naming whichever package you land on. The `client` injection option is also worth a row in the config table, since right now it only appears in prose at `baidu.mdx:93`.

---

## Assessment

**Changes requested.** Mostly not because of the code quality — the `VectorStore` contract is implemented exactly, registration in `factory.ts` / `index.ts` is right, the optional-peerDep + `peerDependenciesMeta` + tsup `external` triple is exactly the pattern we want, the filter escaping is at parity with Python, and the single-flight init is genuinely well done. The core problem is that the provider is built against a package that cannot do the job, and the ambient `.d.ts` plus the mocked test suite are precisely what prevented CI from telling you.

C3 and I5 are the exception: those are real defects in this file's own state handling, they'd survive the SDK swap untouched, and they're the kind of thing a mock-only suite structurally cannot catch. Fix them wherever the rewrite lands.

That premise came from my issue. Happy to pair on the rewrite against `@mochow/mochow-sdk-node`, or to help get a Mochow instance stood up so the integration test has something real to hit — just say which would help.

Aditya-Tripuraneni:
I reworked the Baidu provider around the real `@mochow/mochow-sdk-node` package instead of `@baiducloud/sdk`.

  What changed:
  - Replaced `@baiducloud/sdk` with `@mochow/mochow-sdk-node`.
  - Removed the hand-written ambient `baidu-sdk.d.ts`.
  - Rewrote the provider around the flat Mochow client API.
  - Removed the `getMethod()` duck-typing and dual camelCase/snake_case payloads.
  - Switched response handling to inspect Mochow `{ code, msg }` envelopes.
  - Added table create/drop polling.
  - Fixed `deleteCol()` to wait for any in-flight initialization before dropping.
  - Kept keyword search fail-closed until the BM25 inverted index is confirmed.
  - Updated tests to use the real SDK enums/request classes, only mocking the network-facing client.
  - Avoided storing `data` / `textLemmatized` twice by keeping `data` as a top-level field, indexing `textLemmatized`, stripping both from JSON
  metadata, and reconstructing `payload.data` on reads.

  I also merged `upstream/main` and regenerated `pnpm-lock.yaml` to resolve the lockfile conflict.

kartik-mem0 (approved):

PR #5791, opened 2026-06-23
feat: add Google Vertex AI Vector Search support to vector store factory

### LINKED ISSUE #5787: feat(ts-sdk): add Vertex AI Vector Search vector store
## Summary

The Python SDK supports **Vertex AI Vector Search** as a vector store provider, but the TypeScript OSS SDK (`mem0ai/oss`) does not. Add it to bring the TS SDK to parity.

| | |
|---|---|
| Python reference | `mem0/vector_stores/vertex_ai_vector_search.py` |
| Target file (TypeScript) | `mem0-ts/src/oss/src/vector_stores/vertex_ai_vector_search.ts` |
| Suggested implementation | Use `@google-cloud/aiplatform` (index + deployed index endpoint). |

## Requirements

- [ ] Implement `VertexAIVectorSearch` in `mem0-ts/src/oss/src/vector_stores/vertex_ai_vector_search.ts`, extending `VectorStore` (`mem0-ts/src/oss/src/vector_stores/base.ts`) and mirroring the Python provider's behavior (insert / search / get / update / delete / list / reset).
- [ ] Register the `"vertex_ai_vector_search"` provider in `mem0-ts/src/oss/src/utils/factory.ts` (VectorStoreFactory).
- [ ] Add `@google-cloud/aiplatform` to `mem0-ts/package.json` (optional/peer dependency, lazy-imported like other providers).

## Reference pattern

## Notes

Heaviest vector store — requires a created index and a deployed index endpoint. Mirror `mem0/vector_stores/vertex_ai_vector_search.py`. GCP auth.

---

divyansh-1009:
## Linked Issue

Closes #5787

## Description

This PR adds the **Vertex AI Vector Search** (Google Matching Engine) provider to the TypeScript OSS SDK (`mem0ai/oss`), bringing it to feature parity with the Python SDK.

Specifically, it:
- Implements the `VertexAIVectorSearch` class which fulfills the `VectorStore` interface (supporting insert, search, get, update, delete, etc.).
- Adds `@google-cloud/aiplatform` as an optional `peerDependency` in `mem0-ts/package.json` to follow the existing lazy-loading provider pattern.
- Registers the new provider in the `VectorStoreFactory`.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added unit tests in `mem0-ts/src/oss/tests/vertex-ai-vector-search.test.ts` to verify client initialization and SDK method behavior. All tests pass successfully under `pnpm run test:ts`.

## Checklist


kartik-mem0:
Thanks for taking this on — this is a careful port, and it shows. The structure mirrors the Python provider almost line-for-line (`_createDatapoint`/`_createRestriction`, the `1 - distance` score, the zero-vector `list()` approach), the lazy `await import()` + peerDependency wiring matches how the other optional providers load, and the factory key lines up with the Python side (`vertex_ai_vector_search`). Good foundation — a few things to sort before it can go in.

## 🔴 Blocking: CI is red

`test:unit` fails on the externals drift-guard:

```
FAIL src/oss/tests/tsup-externals.test.ts
  ● should have every peerDependency in the external array
    + Array [ "@google-cloud/aiplatform" ]
Tests: 1 failed, 774 passed, 775 total
```

Your own `vertex-ai-vector-search.test.ts` passes — this failure is the repo guard, and it's correct: when you added `@google-cloud/aiplatform` to `peerDependencies` in `package.json`, it also needs to be added to the `external` array in `mem0-ts/tsup.config.ts` (around line 4, next to `@google/genai` and `@azure/search-documents`). Without it, tsup will try to bundle the whole gRPC / google-gax tree into `dist`.

One-line fix:

```ts
// mem0-ts/tsup.config.ts
const external = [
  ...
  "@google/genai",
  "@google-cloud/aiplatform",   // add this
  ...
];
```

That turns CI green.

## Important

**1. The telemetry user-id record gets written into the live vector index.** `setUserId()` stores a `mem0-user-id-record` datapoint (zero vector) into the same index as real memories, and it runs on essentially every init (`memory/index.ts:493`). Because `list()` queries with a zero vector and this record is stored with a zero vector, it's the exact nearest match — so it can surface as a phantom entry in `getAll()` / `deleteAll()`. The Azure provider deliberately keeps this out of the data path with a separate `memory_migrations` index (`azure_ai_search.ts:588-674`), and the Python provider doesn't store it at all. Could you keep the telemetry id out of the data index — a separate index, or at least filter the sentinel id out of `list`/`search` results?

**2. `vectorSearchApiEndpoint` should be required for the query path.** The Python provider raises `"vector_search_api_endpoint is required for get operation"` when it's missing. Here, when it's unset the client falls back to `${region}-aiplatform.googleapis.com`, which isn't the dedicated public match endpoint — so `insert`/`delete` succeed but `search`/`get`/`list` quietly fail with no clear error (a write-only store). Worth validating it up front, mirroring the Python guard, so users get a clear message.

**3. Tests don't exercise the real logic yet.** Coverage on the new file is ~37% (`vertex_ai_vector_search.ts | 36.89% stmts`), with lines 219-348 — the search-result mapping, `get`, `update`, `delete`, `list`, and get/setUserId — uncovered. The current tests mock `findNeighbors` to return empty, so the payload↔restricts round-trip, the score math, and the `include`/`exclude` → `allowList`/`denyList` filter translation are never checked. One case that returns a populated neighbor and asserts the mapped `{ id, payload, score }` would catch the bugs that are easiest to introduce here.

## Minor

- `generateUUID()` (around line 310) is copied from the Azure provider but never called — `getUserId` uses `Math.random().toString(36)` instead. Can be removed.
- `get(id)` queries by datapoint id, which returns the *nearest neighbor*, not necessarily the exact point. Worth guarding with `if (neighbor.datapoint.datapointId !== id) return null` so `update()`'s existence check and the `getUserId` readback don't act on a different datapoint.
- `update()` silently returns when the point isn't found; the Python side returns `false` and logs. A small log would help debugging.
- Docs are unchecked in the template — a short page under `docs/components/vectordbs/dbs/` would round this out.

## To get this merged

1. Add `@google-cloud/aiplatform` to `tsup.config.ts` externals → green CI.
2. Keep the telemetry sentinel out of the data index.
3. Validate `vectorSearchApiEndpoint` before queries.
4. Add a behavioral test with a populated response.

Items 2-4 are the substance; item 1 is the quick unblock. Thanks again — good to see the TS side catching up to Python here.

kartik-mem0 (approved):

schizophrenicmaniac:
Thank you!! ^-^

PR #5793, opened 2026-06-23
feat(mem0-ts): add MongoDB vector store provider to OSS TypeScript SDK

### LINKED ISSUE #5777: feat(ts-sdk): add MongoDB vector store
## Summary

The Python SDK supports **MongoDB** as a vector store provider, but the TypeScript OSS SDK (`mem0ai/oss`) does not. Add it to bring the TS SDK to parity.

| | |
|---|---|
| Python reference | `mem0/vector_stores/mongodb.py` |
| Target file (TypeScript) | `mem0-ts/src/oss/src/vector_stores/mongodb.ts` |
| Suggested implementation | Use the `mongodb` driver (v7) with Atlas Vector Search (`$vectorSearch`). |

## Requirements

- [ ] Implement `MongoDB` in `mem0-ts/src/oss/src/vector_stores/mongodb.ts`, extending `VectorStore` (`mem0-ts/src/oss/src/vector_stores/base.ts`) and mirroring the Python provider's behavior (insert / search / get / update / delete / list / reset).
- [ ] Register the `"mongodb"` provider in `mem0-ts/src/oss/src/utils/factory.ts` (VectorStoreFactory).
- [ ] Add `mongodb` to `mem0-ts/package.json` (optional/peer dependency, lazy-imported like other providers).

## Reference pattern

## Notes

Mirror `mem0/vector_stores/mongodb.py`.

---

Diveyam-Mishra:
## Linked Issue

Closes #5777 

### Summary

This PR adds MongoDB as a supported vector store provider in the TypeScript OSS SDK (`mem0-ts`), bringing it closer to feature parity with the existing Python implementation (`mem0/vector_stores/mongodb.py`).

The implementation follows the existing vector store provider patterns used throughout the TypeScript SDK and supports full CRUD operations, vector similarity search, and provider registration through the factory layer.

### Type of Change

* [x] New feature (non-breaking change that adds functionality)

#### Added

* New `MongoDB` vector store implementation in:

  * `mem0-ts/src/oss/src/vector_stores/mongodb.ts`
* `MongoDBConfig` typing support
* Unit test suite covering provider functionality

#### Updated

* Registered `"mongodb"` provider in `VectorStoreFactory`
* Added `mongodb` as a peer dependency

### Implementation Details

* Uses the official MongoDB driver
* Implements the `VectorStore` interface and supports:

  * `insert`
  * `search`
  * `get`
  * `update`
  * `delete`
  * `list`
  * `reset`
* Supports Atlas Vector Search via `$vectorSearch`
* Uses Atlas Search for keyword-based retrieval
* Follows existing TypeScript provider patterns for consistency

### Test Coverage

* [x] I added/updated unit tests

Added a dedicated test suite covering:

* Provider initialization
* Insert operations
* Vector search
* Retrieval by ID
* Updates
* Deletion
* Listing records
* Collection reset behavior
* Error handling paths

All tests pass successfully.

### Breaking Changes

N/A

### Checklist

* [x] My code follows the project's style guidelines
* [x] I have performed a self-review of my code
* [x] I have added tests that prove my feature works
* [x] New and existing tests pass locally
* [x] I have updated documentation if needed

Part of the TypeScript ↔ Python SDK provider-parity effort.


kartik-mem0:
## Nice work — solid provider, a few nits before merge

Thanks for taking this one on. I reviewed it against our existing TS provider pattern (`pgvector.ts`), the Python reference (`mem0/vector_stores/mongodb.py`), and checked the MongoDB driver calls against the current Node driver docs. The implementation is clean and the factory wiring is exactly right — this is close.

### Strengths
- **Factory + interface compliance is textbook.** `factory.ts:114-115` mirrors every sibling case, and `class MongoDB implements VectorStore` satisfies `base.ts` cleanly (the green `build_ts_sdk` job is proof). Defining `MongoDBConfig` inline like `PGVector` does is the right call over the issue's "put it in `types/`" suggestion — consistency with the real pattern wins.
- **Init is actually more robust than pgvector.** Every method `await this.initialize()` behind an idempotent `_initPromise`, so the constructor's fire-and-forget init can't race a first call. Nice.
- **Security ported faithfully.** The `_validate_filter_value` injection guard from the Python provider is here (`mongodb.ts:130-150`) and covered by tests (`mongodb.test.ts:148-156`).
- **Driver usage is correct** — verified `createSearchIndex({ name, type: "vectorSearch", definition })`, `listSearchIndexes()`, and the `$vectorSearch` stage shape against the MongoDB Node driver docs.
- Tests are above the bar for a provider — they assert exact pipeline shapes, both filter paths, and all CRUD operations.

### Issues

#### Important
1. **`docs/components/vectordbs/dbs/mongodb.mdx` — the existing Python example got damaged in this diff.** The `config = { "vector_store": {...} }` block lost its indentation (it's now flush against the left margin), and the new `Python` / `TS` labels are bare text lines rather than headings. The TS snippet also still uses Python-style `{"role": "user"}` keys and is missing semicolons. Could you restore the Python block's indentation and tidy the TS example? While you're there, a one-line note that vector search requires a MongoDB **Atlas** cluster would help — `$vectorSearch` and `createSearchIndex` aren't available on a plain self-hosted server, and the example URL points at `localhost`.

#### Minor
1. **Logging noise** — there's a `console.log` on every insert/search/get/update/delete. `pgvector.ts` only logs on error; a top-level SDK shouldn't print on every successful read. Could you drop the success-path logs (or gate them behind a debug flag)?
2. **`package.json:123`** — `"mongodb": ">=6.0.0"` has no upper bound, so a future breaking major would resolve. Every neighbor pins a major with `^`. Suggest `^7.0.0` (matches the v7 the issue called for and what pnpm actually resolved).
3. **`search()` skips the index pre-check** the Python version does (`list_search_indexes` → early return) and relies on the try/catch instead. Same end result, just a slightly less clear error if the index is missing. Optional.

### Evidence
- **CI:** real checks all green — `CI Gate`, `build_ts_sdk (20)` + `(22)`, `integration_ts_sdk (20)` + `(22)`, TS changelog/change-detection, `docs llms.txt`, and CLA all pass. The red `Vercel` check is the fork deploy-preview (no deploy perms from forks) — advisory, not merge-blocking.
- **Driver API:** `createSearchIndex` / `listSearchIndexes` / `$vectorSearch` shapes verified against the current MongoDB Node driver docs — correct. Worth knowing all three require an Atlas 7.0+ cluster; the code degrades with warnings on non-Atlas servers.
- **Cross-package:** diffed against `mem0/vector_stores/mongodb.py` — faithful port. The post-`$match` filtering, Atlas index eventual-consistency, and `Math.random()` user IDs are all inherited from the Python provider, not introduced here.
- **Could NOT verify:** didn't run the jest suite locally or against a live Atlas cluster — the tests are mock-based asserting pipeline shapes, which is the right approach for CI.
- **Linked issue:** #5777 — addressed (provider + factory registration + peer dependency + tests + docs).

### Assessment
The code itself is ready — factory wiring, interface compliance, and driver usage are all correct. Happy to merge once the docs formatting is restored and the debug logging is toned down. Thanks again for the clean implementation.

kartik-mem0:
Thanks for the contribution! The TypeScript implementation itself is solid and the driver API usage is correct. Two things need to be addressed before this can merge — one will fail CI, and one is a factual issue in the docs.

## 🔴 Critical — CI will fail on the lockfile

`package.json` declares `"mongodb": "^7.0.0"` but `pnpm-lock.yaml` records a different specifier:

```yaml
mongodb:
  specifier: '>=6.0.0'   # does not match package.json "^7.0.0"
  version: 7.2.0
```

`ts-sdk-ci.yml` installs with `pnpm install --frozen-lockfile`, which errors with `ERR_PNPM_OUTDATED_LOCKFILE` on any specifier mismatch. Please pick one canonical range and regenerate:

```bash
cd mem0-ts && pnpm install --lockfile-only
```

(`mongodb@^7.0.0` is valid — latest is 7.3.0 — so the range itself is fine; the lockfile just needs to match.)

## 🟡 Important — the "requires Atlas" docs claim is outdated

The new notes state, in absolute terms:

> MongoDB Vector Search requires a MongoDB Atlas cluster. The `$vectorSearch` aggregation stage and `createSearchIndex()` API are not available on a standard self-hosted MongoDB deployment.

> Self-hosted MongoDB deployments do not support the Atlas Search features used by Mem0.

That's no longer accurate. Per the official MongoDB manual (`$vectorSearch` → Compatibility):

> The `$vectorSearch` aggregation pipeline stage is available in MongoDB **Atlas** clusters running **6.0.11+**, MongoDB **Enterprise** deployments running **8.2+**, and **MongoDB Community deployments running 8.2 or later**.

So as of MongoDB **8.2**, self-managed Community/Enterprise deployments support exactly the features this provider uses — the claim is only true for self-hosted versions older than 8.2.

There's also an internal contradiction: the bottom note says the default URI is `mongodb://username:password@localhost:27017` (self-hosted) *and* that self-hosted isn't supported.

Suggested wording:

> MongoDB Vector Search requires either a MongoDB Atlas cluster (v6.0.11+) **or** a self-managed MongoDB Community/Enterprise deployment running **v8.2+**. Self-hosted versions older than 8.2 do not support `$vectorSearch` or `createSearchIndex()`.

## 🟢 Minor (optional)

- The **Config table** in the docs only lists the Python snake_case params (`db_name`, `mongo_uri`, …). The TS config uses camelCase and `url` (not `mongoUri`) — worth adding a TS row so the example matches the reference.
- `getUserId()` uses `Math.random()` while `uuid` is already a direct dependency used elsewhere in the SDK — minor consistency nit.
- The tests fully mock the `mongodb` module, so they validate pipeline shapes/call args but not real cluster behavior. Good as unit tests — just flagging that "all tests pass" is mock-level.

## ✅ Looks good

- Driver API usage is correct for the v7 Node driver: `createSearchIndex({ name, type, definition })`, `listSearchIndexes()`, and `$vectorSearch`/`$search` kept as the first pipeline stage with `$match` spliced in after.
- Full `VectorStore` interface parity, `mongodb` correctly added as a `peerDependency` (consistent with `pg`/`redis`/`better-sqlite3`) and to the tsup `external` list.
- Filter-injection guard and `numCandidates` math faithfully ported from the Python implementation.

Nice work overall — once the lockfile is regenerated and the Atlas wording is corrected, this should be good to go.

Diveyam-Mishra:
Of the above changes  I noticed `quadrant.ts` also uses `Math.random()`, so I left this consistent with the existing SDK implementation. If you'd prefer to standardize on `uuid` throughout the SDK, I'm happy to open a follow-up issue and submit a separate PR to update all occurrences.

Diveyam-Mishra:
@kartik-mem0  Can u review

kartik-mem0 (approved):

PR #5795, opened 2026-06-23
fix(ts): allow openai@5+ (undici) — fixes "Premature close" on hosts without IPv6

### LINKED ISSUE #5794: OSS TS SDK: openai@4 dependency causes ~100% embedding failures ("Premature close") on hosts without IPv6
## Summary

The OSS TypeScript SDK depends on `openai@^4` (`mem0-ts/package.json` → `"openai": "^4.93.0"`). The openai v4 client uses the legacy `node-fetch`/`http.Agent` HTTP stack, which does **not** implement happy-eyeballs (`autoSelectFamily`). On hosts with **no IPv6 connectivity** (very common in containers/datacenters), this causes **~100% of embedding/LLM calls to fail** with:

```
APIConnectionError: Connection error.
  cause: invalid response body while trying to fetch https://<provider>/v1/embeddings: Premature close
```

Switching the client to **openai@5+** (which uses native `fetch`/undici, and therefore happy-eyeballs → IPv4 fallback) fixes it completely.

## Why it happens

1. Host/container has no IPv6 route, but DNS still returns AAAA records (e.g. Cloudflare-fronted providers like OpenRouter resolve to `2606:4700::...`). Node's `dns.lookup` does not pass `AI_ADDRCONFIG`, so AAAA is returned on both glibc and musl even with no IPv6.
2. openai@4's `node-fetch`/`http.Agent` client attempts the IPv6 address and has **no IPv4 fallback** (no happy-eyeballs), so the connection dies mid-response → "Premature close".
3. openai@5+ uses undici (`fetch`), which does `autoSelectFamily` (races v4/v6) → uses the working IPv4 → succeeds.

## Reproduction (verified)

Run inside a container on a host with no global IPv6 address, against an OpenAI-compatible endpoint (here OpenRouter):

| client | result (200 reqs) |
|---|---|
| `globalThis.fetch` (undici) directly | 200 OK |
| **openai@4.104.0** (mem0's bundled client) | **0 OK / 200 "Premature close"** |
| openai@4.104.0 + `new OpenAI({ fetch: globalThis.fetch })` | 200 OK |
| **openai@6.17.0** | **200 OK** |

`curl -6 <provider>` fails; `curl -4` returns 200 — confirming the IPv6 route is dead and only the undici-based clients fall back to IPv4.

Notably this is **not** reproducible on a developer machine that has working IPv6 (e.g. macOS) — only on no-IPv6 hosts, which is why it tends to surface only in production.

## Suggested fix

Bump the OSS TS SDK's `openai` dependency to support v5+ (which uses undici/native fetch). mem0's usage (`new OpenAI({ apiKey, baseURL, timeout })`, `embeddings.create`, `chat.completions.create`) is API-compatible across v4→v5→v6.

I'll open a PR widening the range. Happy to adjust to whatever range the maintainers prefer.

## Environment
- `mem0ai` 3.0.9 (OSS, `mem0ai/oss`)
- Node 24 (alpine + slim, both affected)
- Vector store: Qdrant; embedder/LLM via OpenAI-compatible base URL (OpenRouter)

takefy-dev:
## What

Widen the OSS TypeScript SDK's `openai` dependency from `^4.93.0` to `^4.93.0 || ^5 || ^6`, so users can install an undici-based openai client.

Fixes #5794.

## Why

openai@4 uses the legacy `node-fetch`/`http.Agent` HTTP stack, which does **not** do happy-eyeballs (`autoSelectFamily`). On hosts with **no IPv6 connectivity** (common in containers/datacenters), DNS still returns AAAA records for Cloudflare-fronted providers (e.g. OpenRouter → `2606:4700::...`), the v4 client attempts the dead IPv6 address with no IPv4 fallback, and **every embedding/LLM call fails** with:

```
invalid response body while trying to fetch https://.../v1/embeddings: Premature close
```

openai@5+ uses native `fetch`/undici, which races v4/v6 (`autoSelectFamily`) and falls back to the working IPv4. Verified on a no-IPv6 host: openai@4.104.0 → 0/200 ("Premature close"); openai@6.17.0 → 200/200 OK (full diagnosis + reproduction table in #5794).

## Compatibility

The SDK's openai usage (`new OpenAI({ apiKey, baseURL, timeout })`, `embeddings.create`, `chat.completions.create`) is API-compatible across v4 → v5 → v6; we confirmed `embeddings.create` and `chat.completions.create` work on v6 against an OpenAI-compatible endpoint. This change only **widens** the accepted range (v4 still allowed), so it's non-breaking for existing users.

Happy to narrow to `^4.93.0 || ^5` if maintainers prefer to gate v6 until the test suite is verified against it.


kartik-mem0:
## Needs Human Review

**Why:** Dependency change — per this repo's review policy, dependency version-range changes require a maintainer call before merge, regardless of how clean the fix is. This is also **high-priority**: issue #5794 reports ~100% embedding/LLM failures on no-IPv6 hosts (common in containers/datacenters), so please prioritize.

### What this PR does
Widens `mem0-ts/package.json`'s `openai` dependency from `^4.93.0` to `^4.93.0 || ^5 || ^6`, allowing users to install openai@5+, which uses undici/native fetch with `autoSelectFamily` (IPv4 fallback). openai@4 uses node-fetch with no happy-eyeballs, causing 100% "Premature close" failures on no-IPv6 hosts when DNS returns Cloudflare AAAA records.

### Findings

#### Important (should address)

1. **`pnpm-lock.yaml` not updated** — CI runs `pnpm install --frozen-lockfile`, which will reject the specifier mismatch and fail the build. The lockfile needs a `pnpm update openai@5` (or `@latest`) commit alongside this change. Existing users also need to know to run `pnpm update openai` manually — a CHANGELOG note would help.

2. **No regression test** (`mem0-ts/package.json`) — REVIEW-PARAMS requires "a regression test that fails without the fix" for bug fixes. The IPv6 transport failure is hard to unit-test (requires a constrained network), but a smoke test asserting the SDK constructs and calls correctly under openai@5/v6 (or a CI matrix entry running tests against openai@5) would satisfy the requirement and make the fix durable.

#### Minor (maintainer call)

3. **`|| ^6` arm is self-described as unvetted** — The PR author notes they're "happy to narrow to `^4.93.0 || ^5`" if v6 isn't yet verified. The API surface mem0-ts uses (`embeddings.create`, `chat.completions.create`, constructors, `tool_calls`, `AzureOpenAI`) is stable across v4/v5/v6 — we traced all 6 callsites and confirmed no breaking shape changes affect the SDK's usage. But the zod peer dependency shuffle in v6 could cause resolution warnings for users who pin `zod@^4`. Narrowing to `|| ^5` is the conservative choice; `|| ^6` is safe by semver but untested in CI.

### Strengths
- Root cause diagnosis is correct and well-evidenced (issue #5794's reproduction table is solid: 0/200 OK on v4, 200/200 OK on v6)
- API surface is stable: all 6 openai callsites (`OpenAI`/`AzureOpenAI` constructors, `embeddings.create`, `chat.completions.create`, `tool_calls` access, `response.data[0].embedding`) are unchanged across v4→v5→v6
- `encoding_format: "float"` in `embeddings/openai.ts` and `lmstudio.ts` is a correct safety net that bypasses v5's internal base64 path — keep it
- v5's change from `tool_calls: []` to `tool_calls: undefined` is handled safely by the existing `if (response.tool_calls)` truthy check
- Fix is non-breaking: `^4.93.0` remains in the range, so existing installs are unaffected
- Hosted client (`src/client/`) uses `globalThis.fetch` directly and is not affected at all

### Evidence
- **CI:** Vercel check failed (advisory — "Authorization required to deploy" — fails on every fork PR, not merge-blocking). CLA check passed. No TypeScript SDK test/lint/build CI ran (specifier change likely triggered no jobs, but would fail `--frozen-lockfile` if it had).
- **Targeted tests / red-green:** N/A — dependency-only change; the underlying bug requires a no-IPv6 network environment to reproduce and cannot be red-green proven in a unit test.
- **API surface trace:** Read all 6 openai-importing source files in `mem0-ts/src/oss/src/`. Constructor params, response shapes, and tool_calls access patterns are all stable across v4/v5/v6. Confirmed via code tracing only — requires live network + constrained IPv6 env to fully reproduce.
- **Duplicate check:** No other open PRs address #5794.
- **Could NOT verify:** openai@5/v6 TypeScript compilation against mem0-ts (no worktree — dependency-only change); runtime behavior on a no-IPv6 host.

### Assessment
Ready to merge: **No — maintainer judgment required.** Est. human time: ~8 min

The fix is technically correct and the API compatibility is confirmed by code tracing. Two things need to happen before merge: (1) a `pnpm-lock.yaml` update so CI doesn't reject the specifier mismatch, and (2) a decision on `|| ^6` (keep it or narrow to `|| ^5`). Once the author addresses the lockfile and optionally the test gap, this is straightforward to land.

kartik-mem0:

PR #5797, opened 2026-06-24
feat(ts-sdk): add AWS Neptune Analytics vector store

### LINKED ISSUE #5786: feat(ts-sdk): add AWS Neptune Analytics vector store
## Summary

The Python SDK supports **AWS Neptune Analytics** as a vector store provider, but the TypeScript OSS SDK (`mem0ai/oss`) does not. Add it to bring the TS SDK to parity.

| | |
|---|---|
| Python reference | `mem0/vector_stores/neptune_analytics.py` |
| Target file (TypeScript) | `mem0-ts/src/oss/src/vector_stores/neptune_analytics.ts` |
| Suggested implementation | Use `@aws-sdk/client-neptune-graph`. |

## Requirements

- [ ] Implement `NeptuneAnalytics` in `mem0-ts/src/oss/src/vector_stores/neptune_analytics.ts`, extending `VectorStore` (`mem0-ts/src/oss/src/vector_stores/base.ts`) and mirroring the Python provider's behavior (insert / search / get / update / delete / list / reset).
- [ ] Register the `"neptune"` provider in `mem0-ts/src/oss/src/utils/factory.ts` (VectorStoreFactory).
- [ ] Add `@aws-sdk/client-neptune-graph` to `mem0-ts/package.json` (optional/peer dependency, lazy-imported like other providers).

## Reference pattern

## Notes

Graph + vector hybrid. Mirror `mem0/vector_stores/neptune_analytics.py`. Note: Python registers this under the provider key `neptune`.

---

rodboev:
## Linked Issue

Closes #5786

## Description

This adds AWS Neptune Analytics as a TypeScript OSS vector-store provider, so `mem0-ts` can create a Neptune-backed store instead of rejecting the config at factory construction time.

The branch adds a dedicated Neptune adapter, wires the Python-compatible `neptune` provider key into `VectorStoreFactory`, keeps `neptune-analytics` as a compatibility alias, exports the adapter from the OSS entrypoint, keeps the AWS SDK dependency in the same optional peer-dependency pattern the other TS vector stores already use, and marks that peer dependency external in `tsup` so the optional-provider contract holds in the built package. The review follow-ups also tighten the adapter contract: failed insert upserts are rolled back, failed update upserts throw instead of disappearing silently, insert and full-update queries avoid unsupported Neptune `FOREACH` clauses, user metadata named `label` survives round-trips, advanced metadata filters map onto Neptune’s supported `vertexFilter` operators, and `list()` now returns a real total count instead of the page size.

The Neptune vector search procedure was checked against the official AWS Neptune Analytics docs, which document `neptune.algo.vectors.topK.byEmbedding` with `vertexFilter`; the branch keeps that call shape and adds focused coverage for adversarial collection and filter identifiers.

## Root cause

The original gap was in the TypeScript OSS package, not in upstream provider design. `mem0-ts/src/oss/src/utils/factory.ts` had no Neptune branch, so TS configs that Python already supported still fell through the unsupported-provider error path.

Once the adapter existed, the branch review turned up several correctness issues in the first draft:

- combined update could swallow `success: false` from `neptune.algo.vectors.upsert`
- insert could leave behind a node without a usable vector on failed upsert
- the first filter implementation only emitted simple equality checks and returned page length as `list()` count
- insert and full-update queries used `FOREACH`, which Neptune openCypher does not support
- the TypeScript provider key used `neptune-analytics` instead of the issue and Python key, `neptune`

This branch fixes those issues in the adapter and factory rather than pushing them into follow-up work.

## Scope

- TypeScript OSS SDK package only
- No Python, server, dashboard, plugin, or workflow changes
- No docs or examples, because the existing TS OSS export surface stays coherent without them

## Type of Change

- [ ] Documentation update

## Test Coverage

Verification:
- [x] `cd mem0-ts && pnpm install --frozen-lockfile`
- [x] `cd mem0-ts && pnpm exec jest --runTestsByPath src/oss/tests/factory.unit.test.ts --runInBand`
  - Passed: 57 tests in 1 suite, including `neptune` factory construction and the `neptune-analytics` alias.
- [x] `cd mem0-ts && pnpm exec jest --runTestsByPath src/oss/tests/vector-stores-compat.test.ts --runInBand --testNamePattern=Neptune`
  - Passed: 17 Neptune-focused tests in 1 suite, covering endpoint parsing, request shaping, documented `topK.byEmbedding` search shape, score normalization, payload replacement, payload-only and vector-only updates, collection deletion, total-count pagination, persisted telemetry user IDs, unsupported filter guards, failed update upserts, both insert failure cleanup paths, and adversarial collection/filter identifier escaping.
- [x] `cd mem0-ts && pnpm exec jest --runTestsByPath src/oss/tests/tsup-externals.test.ts --runInBand`
  - Passed previously: 3 tests proving every peer dependency, including the Neptune SDK, is kept in the `tsup` external list.
- [x] `cd mem0-ts && npx prettier --check package.json pnpm-lock.yaml tsup.config.ts src/oss/src/index.ts src/oss/src/utils/factory.ts src/oss/src/vector_stores/neptune_analytics.ts src/oss/tests/factory.unit.test.ts src/oss/tests/vector-stores-compat.test.ts`
  - Passed for all touched files after the rebase.
- [x] `cd mem0-ts && pnpm exec tsup`
  - Passed previously, including OSS entrypoint bundle and `.d.ts` output with the Neptune SDK left external.
- [ ] `cd mem0-ts && pnpm run build`
  - Current package script still stops at repo-wide `prettier --check .` failures in untouched files outside this PR
- [ ] `cd mem0-ts && pnpm run test:unit`
- [ ] `cd mem0-ts && pnpm install --frozen-lockfile && pnpm run build && pnpm run test:unit`

## Checklist


kartik-mem0:
## Changes Requested

### Strengths
- `escapeLabel`/`escapeProperty` (`neptune_analytics.ts:452-454,906-908`) backtick-quote and double embedded backticks on every label/property-key interpolation — correct Cypher identifier escaping, and actually broader than the `_VALID_IDENTIFIER` allowlist the Python Neptune store just gained in #4878 (allowlist rejects anything non-alphanumeric; this escapes any string safely).
- `get`/`update`/`delete`/`list` (`neptune_analytics.ts:145-292,808-904`) bind every value through real `$parameter`s — no string-concatenated values anywhere in those paths.
- `~label` native-label filtering (`neptune_analytics.ts:469`) sidesteps the Python reference's `label`-property collision bug, where a user's own metadata field literally named `label` gets silently clobbered and then stripped on read. Nice fix, and it checks out in the code.
- Insert rollback (`cleanupFailedInsert`, only targets IDs confirmed new via `findExistingIds`) and throw-on-failed-update (`assertSuccessfulResults`) are both wired correctly — careful failure-path handling.
- 16 unit tests assert actual query text/parameters (procedure names, WHERE clauses, rollback behavior), not just "doesn't throw."

### Issues
#### Critical (must fix)
1. `mem0-ts/src/oss/src/vector_stores/neptune_analytics.ts:130` — `search()` calls `neptune.algo.vectors.topK.byEmbedding`. The Python reference (`mem0/vector_stores/neptune_analytics.py:185`) and every other procedure call in this same file (`neptune.algo.vectors.upsert`, `neptune.graph.pg_schema()`) use plain camelCase procedure names with no internal dot-namespacing — `topK.byEmbedding` looks like a typo for `topKByEmbeddingWithFiltering`/`topKByEmbedding`. If the name is wrong, every `search()` call fails against a real Neptune Analytics graph, and the test suite can't catch it — `vector-stores-compat.test.ts` asserts `queryString.includes("topK.byEmbedding")` against the same string the mock and implementation share, not against AWS's real procedure catalog. Could you confirm the exact procedure name against a live graph or the Neptune Analytics API reference before merge?

#### Important (should fix)
1. `mem0-ts/src/oss/src/utils/factory.ts:114` registers this backend as `"neptune-analytics"`; Python registers the same backend as `"neptune"` (`mem0/utils/factory.py:202`), and #5786 explicitly calls out `neptune` as the expected key. A config using `provider: "neptune"` — the Python-parity key the issue asked for — throws "unsupported provider" in the TS SDK. Worth matching Python's key so configs stay portable across the two SDKs.

#### Minor (optional)
1. No test feeds an adversarial `collectionName` or filter key (containing backticks or Cypher metacharacters) through `escapeLabel`/`escapeProperty` to prove containment. The escaping logic itself looks correct on a read-through — just worth a regression test given the Cypher-injection hardening that landed in the Python sibling the day before this PR.

### Evidence
- **CI:** all real required checks green — `build_ts_sdk` (20, 22), `integration_ts_sdk` (20, 22), `CI Gate`, `check_changes`, `changelog_check`, `license/cla`. Only `Vercel` (preview-deploy) fails, advisory and ignored per convention.
- **Targeted tests / red-green:** not executed this pass (read-only review, no checkout/build). Traced the 16 Neptune tests directly — they assert real query text/parameters against a fully mocked `NeptuneGraphClient`, which is exactly why the procedure-name issue above survives green CI.
- **Cross-package check:** compared line-by-line against `mem0/vector_stores/neptune_analytics.py` and `mem0/utils/factory.py:202`.
- **Could NOT verify:** live Neptune Analytics query execution (exact procedure name, and whether `~label` is a valid pseudo-property inside `vertexFilter`) — no AWS credentials/live graph in this pass, code comparison only. Also tried the AWS Neptune docs via Context7 for the procedure name; its indexed corpus didn't surface the specific vector-algorithm-library page, so this stays an open verification item rather than a confirmed break.

### Assessment
Ready to merge: **With fixes** — please confirm the search procedure name against a real Neptune Analytics graph and align the provider key with Python's `neptune`; happy to re-review once pushed.

Separately, this PR currently shows as **mergeable: CONFLICTING** — needs a rebase against `main` as well, independent of the above.

rodboev:
Thanks for the detailed review. I checked the official AWS Neptune Analytics vector docs, and the documented search call is `CALL neptune.algo.vectors.topK.byEmbedding({ embedding, topK, vertexFilter, concurrency })`, including `vertexFilter`, so I kept that procedure spelling as-is.

I pushed the rework for the actual provider-key mismatch: `neptune` is now the primary TS OSS provider key, matching #5786 and the Python factory, while `neptune-analytics` remains accepted as a compatibility alias. I also added focused coverage for adversarial collection and filter identifiers so the Cypher identifier escaping paths are asserted directly.

The branch is rebased onto current `main` now as well.

kartik-mem0:
Reviewed at `603b2f85`. Verdict below, plus a correction I owe you.

**Score: 7.5 / 10.** Careful, well-tested port that gets the hard parts right.

## Retraction of my earlier review

My review at `af8ab319` (#issuecomment-4914017207) opened with a Critical claiming `neptune.algo.vectors.topK.byEmbedding` was "a typo" and that "every `search()` call fails against a real Neptune Analytics graph." **That was wrong.** I asserted it without checking AWS's docs. Sorry for the noise, and thank you for not just deleting the code.

Verified against the Neptune Analytics user guide this time:

- `vector-similarity.html` lists `.vectors.topK.byEmbedding` as the **current** algorithm, and `.vectors.topKByEmbedding` as the deprecated one.
- `vectors.topK.byEmbedding.html` documents exactly the shape you emit: `CALL neptune.algo.vectors.topK.byEmbedding({embedding, topK, vertexFilter, concurrency}) YIELD node, score`. It also states that prefixing the call with `MATCH (n)` or `WITH` is **forbidden**. Your bare `CALL` is correct, and the `MATCH`-prefixed form I implicitly suggested would have broken it.

The procedure name that does *not* appear anywhere in AWS's algorithm list is `topKByEmbeddingWithFiltering`, which is what the **Python** provider calls at `mem0/vector_stores/neptune_analytics.py:185`. That is ours to fix, not yours.

Separately, the provider-key point from that review (`neptune-analytics` vs Python's `neptune`) was valid, and you fixed it in `63270c75`. Credited.

## Docs

Issue #5786 asked for docs; the PR body said "No docs or examples" while ticking the docs checkbox. Since you had maintainer edits enabled, I pushed them to your branch rather than block you on it (`603b2f85`):

- `docs/components/vectordbs/dbs/neptune_analytics.mdx` rewritten to cover both SDKs (install, usage, split config tables, IAM policy).
- Neptune Analytics added to `docs/components/vectordbs/overview.mdx`.

`docs llms.txt / check-llms-txt` passes on that commit. Revert it if you would rather write them yourself.

## Strengths

- **Query shape is right.** Bare `CALL` for `topK.byEmbedding`, `YIELD node, score`, `vertexFilter` for label and metadata scoping. Matches AWS's documented grammar.
- **Score normalization is right, and more right than Python's.** AWS returns the *squared Euclidean distance* (`score: 0.0` for a self-match). Your `normalizeScore` (`:1012-1024`) converts with `1 / (1 + d)`, matching what `milvus.ts:290-305` and `milvus.py:189-197` already do for L2. Python's Neptune provider returns the raw distance, which quietly breaks `main.py`'s `score >= 0.95` dedup and its `threshold` in `[0,1]` contract. Your side is the correct one.
- **Injection surface taken seriously.** Backtick-doubling in `serializeAlgorithmKey`, `$param` binding everywhere except the one spot AWS forbids it, and an adversarial test to prove it.
- Full `VectorStore` interface, factory registration, `index.ts` re-export, and `tsup` external all wired consistently with siblings.
- Label scheme (`MEM0_VECTOR_<collection>`) is byte-identical to Python's, so a graph written by one SDK is readable by the other.
- 17 new tests, and CI is green.

## Issues

### Important

**1. `update()` can permanently desync a memory's vector from its text.** `neptune_analytics.ts:176-204`

The combined payload+vector branch issues two queries. The first upserts the embedding; the second writes the properties:

```ts
this.assertSuccessfulResults(results, "Update");   // vector is now committed
await this.executeQuery(`... SET n = $properties ...`);  // if this throws, nothing repairs the vector
```

There is no `try`/`catch` and no compensating action, and AWS states plainly that "updates to vector embeddings are not ACID compliant." Your own `insert()` (`:88-113`) *does* compensate via `cleanupFailedInsert`, so this is an inconsistency inside one file rather than a missing concept.

It is reachable on the normal path, not a corner case: `Memory.update(memoryId, data)` goes through `updateMemory()`, which requires `vectorStore.get(memoryId)` to return non-null first and always passes both a payload and an embedding. So the two-query branch is taken on **every** public `update()` call.

It is also untested in the failing direction. `vector-stores-compat.test.ts:2571` only fails the *first* query, so `assertSuccessfulResults` throws before the second ever runs; the mock has no `failUpdateProperties` switch, so a second-query failure cannot currently be expressed at all.

Suggested: wrap the pair, and on a second-query failure either restore the prior embedding or surface an error that names the desync. A mock option that fails the property write would lock it in.

(For the record, Python's `update()` has the same gap, and no compensation anywhere. Fixing it here does not create Python/TS drift, it fixes the better half first.)

### Question

**2. `insert()` passes the embedding inline where `update()` binds it.** `:91` vs `:182` and `:226`

```
:91    CALL neptune.algo.vectors.upsert(n, row.embedding)      // inline map access
:182   WITH n, $embedding AS embedding
       CALL neptune.algo.vectors.upsert(n, embedding)          // bound first
```

I could not confirm this is a bug, and I am not going to claim it is. AWS never documents whether a map projection is a legal second argument to `upsert`. What I can say:

- Every `upsert` example AWS publishes binds the embedding through `WITH` first, including the one in `vector-index.html` that is explicitly labelled as a pattern to *avoid* (it is avoided for a different reason, and still binds).
- Python's `update()` binds it too (`neptune_analytics.py:259`).
- Your `update()` binds it. Only `insert()` does not.
- The tests cannot adjudicate this. The mock substring-matches the query text (`vector-stores-compat.test.ts:1773`) and reads `row.embedding` off the JS `parameters` object, so it never parses Cypher and passes under either form.

If you have run this against a live graph and inserts work, say so and I will drop it. If you have not, matching `update()`'s bound form costs one line and removes the question.

### Minor

3. `normalizePayload` (`:1008-1010`) is `return { ...payload }`, a pure no-op. It is the natural place to strip Python's bookkeeping keys. Python persists `label` and `updated_at` as real node properties (`neptune_analytics.py:122` on insert, `:238` on update) and pops `label` on read (`:393`); TS never strips either, so on a graph shared between the two SDKs they surface to callers as `metadata.label` and `metadata.updated_at` (nothing in `memory/index.ts`'s `excludedKeys` catches them). One line here closes it. Low severity, since it needs both SDKs pointed at one graph.
4. `listCols()` (`:339`) has no callers. `azure_ai_search.ts` and `pgvector.ts` call their own; this one is unreachable.
5. `list()` (`:283`) runs `RETURN count(n)` with no limit on every call, in parallel with the limited fetch. That is a full scan of the collection per `list()`.
6. `serializeAlgorithmInput` (`:923-924`) does `String(value)` for numbers, so a `NaN` or `Infinity` from an embedder becomes the literal token `NaN` in the Cypher text and fails as a parse error rather than a clear one. A `Number.isFinite` guard is cheaper to debug.
7. Issue #5786 asked for the AWS SDK to be "lazy-imported like other providers." It is a static import. I checked, and the issue's premise is wrong about this codebase: 20 of 23 TS vector stores import eagerly, including `weaviate.ts`, merged one commit before your work. `peerDependenciesMeta` marks only 2 of 35 peers optional. You followed the actual convention. Flagging it only so nobody re-reads the issue and thinks it was missed. The repo-wide cleanup is our ticket to file, not yours.

## Not your bugs

Three things I found in the Python provider while checking parity against yours. All predate this PR and all deserve their own issue. In each case your TypeScript is the correct implementation and Python is the outlier:

- `neptune_analytics.py:185` calls `topKByEmbeddingWithFiltering`, which is not in AWS's documented algorithm list.
- `neptune_analytics.py:394-395` returns Neptune's raw squared distance as `score` (the dataclass field at `:37` is even commented `# distance`), so `main.py`'s `score >= 0.95` dedup and the documented `threshold` range are both meaningless for this backend today.
- `neptune_analytics.py:180` plus `_get_node_filter_clause` (`:465`) scope search with `{equals: {property: 'label', ...}}`, filtering on a **synthetic node property** named `label` rather than the reserved `~label` that your `buildVertexFilter` (`:465-481`) uses and that AWS documents as filterable. Consequence on a shared graph: your `search()` finds Python-written nodes (both write the real graph label), but Python's `search()` cannot see TypeScript-written nodes at all, because TS never writes that synthetic property. Your side is right.

Worth stating plainly: on graph-label scheme (`MEM0_VECTOR_<collection>`, `neptune_analytics.ts:51`) and on `neptune-graph://` prefix stripping (`:422-423`), the two SDKs already agree. I checked both.

## Recommendations

1. Fix the `update()` desync (Important 1). That is the one thing I would want resolved before merge.
2. Answer the `insert()` binding question, either with a live-graph result or by matching `update()`.
3. Minors 3 to 6 are cheap and optional. Minor 3 is worth doing since the method already exists for it.

## Assessment

**Ready to merge: with fixes.**

Code quality is genuinely good and the AWS semantics are handled more carefully than in our own Python provider. The blocker is small and local. Once Important 1 is addressed and the binding question is answered, this is technically ready.

One process note that is not about your code: adopting a new vector-store backend is a maintainer decision about long-term support surface, not something a review can settle. That call belongs to a human on our side regardless of how clean the diff is, so please do not read "with fixes" as "merged the moment you push." I have not approved or merged anything.

## Evidence

- Reviewed head `603b2f85`; CI real checks green (`CI Gate`, `build_ts_sdk` 20 and 22, `integration_ts_sdk` 20 and 22, `changelog_check`, `check_changes`, `docs llms.txt`). The `Vercel` check fails on fork deploy permissions and is not merge-blocking; ignored deliberately.
- `mergeable: MERGEABLE`, not a draft. 2799 additions.
- Worktree at `origin/pr-5797`; targeted test run against `vector-stores-compat.test.ts` and `factory.unit.test.ts`. The 17 new tests fail locally only on the absent `@aws-sdk/client-neptune-graph` peer; an `import 'mem0ai/oss'` probe fails **identically on `main`** (`Cannot find module 'chromadb'` via `factory.ts:21`), so nothing here is a regression this PR introduced.
- AWS Neptune Analytics user guide, fetched during this review: `vector-similarity.html`, `vectors.topK.byEmbedding.html`, `vectors-upsert.html`, `vector-index.html`, `API_ExecuteQuery.html`.
- `@aws-sdk/client-neptune-graph` API surface cross-checked against the aws-sdk-js-v3 docs (`ExecuteQueryCommand` takes `graphIdentifier` / `queryString` / `language`, `payload` is a stream, so `transformToString()` is correct).
- Three independent passes (correctness, repo fit, architecture). The repo-fit pass found no blocking provider-pattern or parity violations: `factory.ts:161-163` registers both `neptune` and `neptune-analytics`, the peer dependency sits in the right alphabetical block, and `tsup-externals.test.ts` already enforces externalization in CI.
- Every Critical and Important above was then sent to a separate reviewer whose only job was to refute it with code evidence. Two proposed Criticals collapsed (one unreachable through any caller in this repo, one where the bug turned out to be Python's), and one Important was downgraded to Minor 7. What survived is what you see. I would rather send you four findings that hold than nine that look thorough.

kartik-mem0 (approved):

PR #5799, opened 2026-06-24
feat: recall memories from extracted facts before final extraction

laicoffee:
## Linked Issue
N/A

## Description

This PR updates the `Memory._add_to_vector_store` flow to recall existing memories from extracted memory candidates instead of recalling directly from the full conversation context.

Previously, the pipeline first retrieved recent session context, then performed vector recall using the entire parsed conversation, and finally passed both the recalled memories and the conversation context into the LLM for memory extraction. In practice, this could lead to unnecessarily large recall queries and more unrelated results.

This change introduces a fact-first recall flow:

1. Extract candidate memories from the current messages plus recent session context
2. Use each extracted candidate memory to recall related existing memories
3. Run a final extraction pass with the candidate memories, recalled memories, and recent context

This keeps the recall query more focused, reduces irrelevant retrieved context, and better matches real-world usage where structured candidate memories are a better retrieval anchor than raw multi-turn conversation text.

The PR also adds regression tests for both sync and async flows to verify that recall is driven by extracted memory text rather than the full conversation.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Manual verification:
- Ran `python -m compileall mem0\memory\main.py tests\memory\test_main.py`
- Attempted to run `python -m pytest tests\memory\test_main.py -q`
- Test collection was blocked in the current environment because local package metadata for `mem0ai` was not available (`importlib.metadata.PackageNotFoundError`)

## Checklist

### DIFF
--- a/docs/changelog/sdk.mdx
+++ b/docs/changelog/sdk.mdx
@@ -7,6 +7,16 @@ mode: "wide"
 <Tabs>
 <Tab title="Python">
 
+<Update label="2026-06-29" description="Fact-first memory recall">
+
+**Memory add pipeline now supports fact-first recall**
+
+- **Memory:** `add()` can now extract candidate memories first, recall existing memories from those candidates, and then run a final reconciliation pass with the recalled context.
+- **Memory:** Added `fact_first_recall` config to opt out and keep the legacy single-pass recall flow when lower latency or cost is preferred.
+- **Memory:** Async recall now runs candidate lookups concurrently to reduce the added latency from the extra recall phase.
+
+</Update>
+
 <Update label="2026-06-17" description="v2.0.7">
 
 **New Features:**
--- a/mem0-ts/src/oss/src/config/defaults.ts
+++ b/mem0-ts/src/oss/src/config/defaults.ts
@@ -2,6 +2,7 @@ import { MemoryConfig } from "../types";
 
 export const DEFAULT_MEMORY_CONFIG: MemoryConfig = {
   disableHistory: false,
+  factFirstRecall: true,
   version: "v1.1",
   embedder: {
     provider: "openai",
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -155,6 +155,59 @@ function validateSearchParams(threshold?: number, topK?: number): void {
   }
 }
 
+function parseAdditiveExtractionResponse(response: string): Array<{
+  id?: string;
+  text?: string;
+  attributed_to?: string;
+  linked_memory_ids?: string[];
+}> {
+  const cleanResponse = extractJson(response);
+  if (!cleanResponse || !cleanResponse.trim()) {
+    return [];
+  }
+
+  try {
+    const parsed = AdditiveExtractionSchema.parse(JSON.parse(cleanResponse));
+    return parsed.memory;
+  } catch {
+    const fallbackJson = extractJson(cleanResponse);
+    return JSON.parse(fallbackJson)?.memory ?? [];
+  }
+}
+
+function mapExistingMemories(existingResults: Array<{ id: string; payload: Record<string, any> }>) {
+  const existingMemories: Array<{ id: string; text: string }> = [];
+  const uuidMapping: Record<string, string> = {};
+  for (let idx = 0; idx < existingResults.length; idx++) {
+    const mem = existingResults[idx];
+    uuidMapping[String(idx)] = mem.id;
+    existingMemories.push({
+      id: String(idx),
+      text: mem.payload?.data ?? "",
+    });
+  }
+  return { existingMemories, uuidMapping };
+}
+
+function replaceLinkedMemoryIds(
+  extractedMemories: Array<{
+    id?: string;
+    text?: string;
+    attributed_to?: string;
+    linked_memory_ids?: string[];
+  }>,
+  uuidMapping: Record<string, string>,
+): void {
+  for (const memory of extractedMemories) {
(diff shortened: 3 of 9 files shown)


laicoffee:
Implemented a fact-first memory recall flow for `Memory._add_to_vector_store`.

Key change:
- extract candidate memories first
- recall existing memories from extracted candidates instead of full conversation context
- run a final extraction pass with recalled memories + recent context

I also added sync/async regression tests for the new recall path.

`license/cla` will be completed on my side. Could a maintainer help approve the workflow run for this fork PR? Thanks!

kartik-mem0:
## Review: feat: recall memories from extracted facts before final extraction

**Verdict: Changes Requested** · Score 6/10

The core intuition here is good — using structured extracted memories as recall anchors is meaningfully better than embedding the raw multi-turn conversation. The red-green proof confirms the new tests exercise the real behavior change. But there are blocking issues to fix before this merges.

---

## Strengths

- Sound idea: extracted memory text ("User likes pour-over coffee") is a better vector query than the raw input ("我最近更喜欢手冲咖啡"). The tests demonstrate this concretely.
- Clean extraction of `_parse_additive_extraction_response` and `_map_existing_memories` to module-level — reduces duplication.
- Correct deduplication in the recall loop via `seen_ids`.
- Graceful degradation: per-candidate recall failures are caught and logged as warnings rather than aborting the batch. Phase 3 failure falls back to Phase 1 output.
- Sync and async paths are symmetric.

---

## Issues

### Critical — must fix before merge

**C1 — 2 existing tests broken** (`tests/memory/test_main.py:65`, `:233`)

`test_empty_llm_response_fact_extraction` and its async counterpart both assert on `"Error parsing extraction response"` in the log. The PR changed where JSON parse errors are caught: they now bubble out of `_generate_additive_extraction_memories` and land in the outer `try/except` as `"LLM extraction failed"`. Both tests fail locally with this PR's code.

Fix: update the two test assertions to match the new log message, or add the specific message to the `_generate_additive_extraction_memories` exception handler.

**C2 — TypeScript SDK left behind**

`mem0-ts/src/oss/src/memory/index.ts:762` still runs the old single-phase pipeline (embed full conversation → search → single LLM call). The TS `generateAdditiveExtractionPrompt` function already accepts a `recentlyExtractedMemories` parameter — the infrastructure is there, it just isn't wired up. Python OSS users get 3-phase recall; TS OSS users get the old behavior. Please bring the TS path to parity in this PR.

**C3 — 2× LLM calls per `add()`, forced on all users with no opt-out**

Phase 1 is always a new LLM call. Phase 3 fires whenever recall returns results — which is virtually always for returning users. This doubles LLM cost on the `add` hot path with no configuration flag to disable or tune it. Please add a `MemoryConfig` option (e.g., `fact_first_recall: bool = True`) so operators can opt out if the extra latency/cost is not acceptable for their deployment.

**C4 — Early exit changes `add()` semantics silently**

Old path: recall ran first, then the single LLM call saw existing memories regardless of its output. New path: if Phase 1 returns `[]`, existing memories are never retrieved. For conversations where the LLM previously looked at existing memories and decided NOOP, it now returns without ever seeing them. This is a real semantic change — please document it in the PR description and, if intentional, in a comment in the code.

---

### Important — worth addressing

**I1 — N embedding+search calls instead of 1**

`_recall_existing_memories_for_candidates` issues one embed + one vector search per candidate sequentially. For a typical batch of 5 extracted memories, that's 5 embedding calls + 5 searches vs. the old 1+1. This is the unavoidable cost of the approach, but it's worth surfacing in the PR description as a conscious tradeoff, and the async path should use `asyncio.gather` to parallelize the N calls (see I3).

**I3 — Sequential recall in the async path (no gather)**

`_recall_existing_memories_for_candidates_async` runs the `for mem in candidate_memories` loop sequentially with `await asyncio.to_thread(...)`. An `asyncio.gather` with per-candidate tasks would run all N recalls concurrently, recovering much of the added latency in the async path.

**I4 — No docs, no config parameter, no changelog**

The change in `add()` behavior (2-phase LLM, N-recall) should be noted in CHANGELOG and `docs/` (even a brief note in the extraction/how-memory-works page). The new hardcoded `top_k=5` (vs. the old `top_k=10`) should also be surfaced or made configurable.

---

### Minor

- `m1` — New recall tests are placed in `TestPromptOverridesCustomInstructions` and `TestAsyncAddToVectorStoreErrors`, neither of which is semantically appropriate. A dedicated class (e.g., `TestFactFirstRecall`) would be clearer.
- `m2` — `ruff check mem0/memory/main.py` reports I001 (unsorted imports) and W293 (blank lines with whitespace) introduced by this PR. The pre-commit hooks catch these — please run `make format && make sort` before pushing.
- `m3` — `top_k=5` in recall is hardcoded without explanation. If this is intentional (per-candidate result set should be smaller than the old single-query set), a brief comment explaining the tradeoff would help.

---

## Evidence

- **CI**: Vercel/deploy-preview fail (advisory, ignored). No required test/lint/build CI check recorded — fork needs maintainer CI approval.
- **Red-green (sync)**: FAIL on `origin/main` (`'user: 我最近更喜欢手冲咖啡\n' != 'User likes pour-over coffee'`); PASS with PR code.
- **Red-green (async)**: same pattern, confirmed.
- **Full test_main.py with PR**: 2 failed (`test_empty_llm_response_fact_extraction` × sync/async), 44 passed.
- **TS parity**: verified at `mem0-ts/src/oss/src/memory/index.ts:762-900` — old single-phase pipeline still in place.

Would be great to have a tracking issue linked for this improvement — it'll help us tie related recall-quality work together.

kartik-mem0:

PR #5804, opened 2026-06-24
fix(chroma): wrap delete() ids in a list (completes scalar-vs-list fix, closes #5803)

### LINKED ISSUE #5803: Chroma provider: scalar-vs-list defect class in update()/delete() (ids must be list-wrapped)
## Summary

The Chroma vector store provider (`mem0/vector_stores/chroma.py`) passes a **scalar** `vector_id` string to ChromaDB collection methods that expect **list** arguments. ChromaDB's `ids` (and `embeddings`/`metadatas`) parameters are list-typed; passing a bare string causes silent, hard-to-debug failures:

- On older ChromaDB: the string is **iterated character-by-character** (treated as a list of single-char ids).
- On newer ChromaDB: **length-consistency / type-validation errors**.

This is a **defect class**, not a single bug — the same scalar-vs-list mismatch appears in more than one method.

## Affected call sites

| Method | Line | Current | Issue |
|--------|------|---------|-------|
| `update()` | ~188 | `self.collection.update(ids=vector_id, embeddings=vector, metadatas=payload)` | `ids` scalar; `embeddings`/`metadatas` also need list wrapping (with `None`-guard) — **fixed in #5757** |
| `delete()` | ~172 | `self.collection.delete(ids=vector_id)` | `ids` scalar — same defect, **not yet fixed** |

For contrast, the correct pattern already exists in the same file:
- `insert()` → `ids=ids` (already a list)
- `get()` → `ids=[vector_id]` (correctly wrapped)

## Fix direction

Normalize all single-id Chroma calls to wrap in a list: `ids=[vector_id]`, and for `update()` guard optional args (`[vector] if vector is not None else None`, `[payload] if payload is not None else None`) so `None` is never wrapped into `[None]` (which trips ChromaDB embedding validation).

- `update()`: addressed in #5757.
- `delete()`: follow-up needed (1-line change + regression test).

## Related

- Fix pair: #5703 + #5757

Bartok9:
## What Problem This Solves

Follow-up to #5757, completing the scalar-vs-list defect class in the Chroma provider that @kartik-mem0 asked to track in **#5803**.

`ChromaDB.delete()` passes a **scalar** `vector_id` to `collection.delete(ids=...)`, but ChromaDB expects `ids` to be a **list**. Passing a bare string is iterated character-by-character on older ChromaDB (deleting wrong/no ids) and trips length/type validation on newer versions — the same silent defect #5757 fixed for `update()`.

## Change

`mem0/vector_stores/chroma.py` — `delete()`:
```diff
- self.collection.delete(ids=vector_id)
+ self.collection.delete(ids=[vector_id])
```
This mirrors the already-correct `get()` (`ids=[vector_id]`) and `insert()` (`ids=ids`) call sites in the same file.

## Evidence

**Command run after the fix:**
```
python -m pytest tests/vector_stores/test_chroma.py -q
```
**Output:**
```
25 passed, 1 warning in 1.49s
```
`test_delete_vector` updated to assert the correct list-wrapped call: `collection.delete.assert_called_once_with(ids=[vector_id])`.

## Related

- Pairs with #5757 (update fix). Closes #5803.

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -169,7 +169,7 @@ class ChromaDB(VectorStoreBase):
         Args:
             vector_id (str): ID of the vector to delete.
         """
-        self.collection.delete(ids=vector_id)
+        self.collection.delete(ids=[vector_id])
 
     def update(
         self,
--- a/tests/vector_stores/test_chroma.py
+++ b/tests/vector_stores/test_chroma.py
@@ -122,7 +122,7 @@ def test_delete_vector(chromadb_instance):
 
     chromadb_instance.delete(vector_id=vector_id)
 
-    chromadb_instance.collection.delete.assert_called_once_with(ids=vector_id)
+    chromadb_instance.collection.delete.assert_called_once_with(ids=[vector_id])
 
 
 def test_update_vector(chromadb_instance):


kartik-mem0:
## Duplicate — recommend consolidating into #5703

### What both PRs do

Both PRs fix the same one-line bug in `ChromaDB.delete()`: `collection.delete(ids=vector_id)` → `collection.delete(ids=[vector_id])`. The diffs are byte-for-byte identical — same file, same line, same test change.

### Why #5703 is the one to land

- **#5703 was opened on 2026-06-20** and already holds an LGTM merge card (red-green proven: test PASS → FAIL on revert → PASS restored; full 24-test suite green; lint clean across Python 3.10/3.11/3.12).
- Reading both diffs side-by-side: `mem0/vector_stores/chroma.py:169` and `tests/vector_stores/test_chroma.py:122` are changed in exactly the same way. There is no additional coverage in #5804.
- The `update()` half of the defect class is separately handled by #5757 (also already LGTM'd) — neither this PR nor #5703 reaches that method, and they don't need to.

So the full scalar-vs-list fix for Chroma is: **#5703** (delete) + **#5757** (update). This PR doesn't add a third piece; it duplicates the first.

### Credit where due

The issue tracking (opening #5803 to frame the defect class as a whole) and the independent identification of the `delete()` root cause are genuinely useful — this PR arrives with a correct fix, clean CI, and a proper test update. The work here is right; the overlap just beat it to the LGTM queue.

### Evidence

- **This PR (#5804):** CI Gate PASS; Python SDK 3.10/3.11/3.12 PASS; Vercel fail is advisory/fork-perms only (ignored per repo rules). Diff read in full.
- **#5703 (recommended):** diff read in full — byte-for-byte identical change; red-green proof captured on 2026-06-24; 24/24 tests pass; ruff clean; LGTM card at https://github.com/mem0ai/mem0/pull/5703#issuecomment-4785677226.
- **main tree:** `chroma.py:172` still shows the unfixed scalar (`ids=vector_id`) — neither PR has landed yet; #5703 merges cleanly.

### Assessment

Recommendation: **close this in favour of #5703** (squash-merge #5703). Est. human time: ~2 min.

Bartok9:
Agreed — thanks for the careful side-by-side, @kartik-mem0. #5703 got to the LGTM queue first with the same one-line `delete()` fix, and #5757 covers the `update()` half. No need for a third piece. Closing this in favour of #5703; happy that the scalar-vs-list defect class is fully tracked via #5803. Thanks for the credit on the issue framing.

PR #5806, opened 2026-06-24
fix(mem0-plugin): store files_touched as a list to stop double JSON-encoding

kartik-mem0:
## Linked Issue

No tracked GitHub issue — reported directly (session-summary memories showing `files_touched` as escaped, slash-heavy blobs in editor search results). Small, root-cause-tied fix; happy to file an issue retroactively if preferred.

## Description

Two things, requested together:

**1. Bug fix — `files_touched` was double-JSON-encoded in session summaries.**
`integrations/mem0-plugin/scripts/capture_session_summary.py` set `metadata["files_touched"] = json.dumps(files[:20])` — a pre-serialized JSON string — and then serialized the whole request body again with `json.dumps(body)`. The stored memory therefore carried a double-encoded escaped string (`"[\"mem0/memory/main.py\", \"src/client/index.ts\"]"`) instead of a real array. When those memories came back through `search_memories` / `get_memories`, the file paths surfaced as backslash/slash-heavy escaped blobs in results shown inside **Claude Code, Cursor, Codex, and Antigravity**. All four editors share `scripts/`, so this **single-line fix covers every editor** — the list is now stored directly and the body is encoded exactly once.

**2. Docs — fix a dead blog link.**
The "Blog Post" card on the [Memory Evaluation](https://docs.mem0.ai/core-concepts/memory-evaluation) page pointed at a placeholder (`mem0.ai/blog/new-algorithm`). Repointed to the current token-efficient-memory-algorithm temporal-reasoning post, which carries the benchmark numbers the page already reflects and links back to the original announcement.

**Plus:** version bumps (Claude Code / Cursor / Codex `0.2.10 → 0.2.11`, Antigravity `0.1.2 → 0.1.3`) and a `0.2.11` CHANGELOG entry that also documents the previously-undocumented rerank-auto-injected-context-by-default change (#5690) shipped since 0.2.10.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A. Memories previously stored with an escaped-string `files_touched` remain readable; only newly captured summaries store a clean JSON array.

## Test Coverage

- [x] I added/updated unit tests (`integrations/mem0-plugin/tests/test_capture_session_summary.py`)
- [x] I tested manually (full plugin suite below)
- [ ] No tests needed

**Red → green proof.**

Before the fix (RED):
```
AssertionError: files_touched must be a JSON array, not a double-encoded string;
got str: '["mem0/memory/main.py", "src/client/index.ts"]'
```

After the fix (GREEN):
```
tests/test_capture_session_summary.py::test_files_touched_is_json_array_not_double_encoded PASSED
tests/test_capture_session_summary.py::test_files_touched_omitted_when_no_files PASSED
2 passed
```

Full plugin suite: `155 passed in 3.68s`. `ruff check` and `ruff format --check` clean. `docs/llms.txt` in sync.

## Checklist

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "mem0",
       "source": "./integrations/mem0-plugin",
       "description": "Mem0 memory layer for AI applications. Add persistent memory, personalization, and semantic search to Claude workflows.",
-      "version": "0.2.10"
+      "version": "0.2.11"
     }
   ]
 }
--- a/.cursor-plugin/marketplace.json
+++ b/.cursor-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "mem0",
       "source": "./integrations/mem0-plugin",
       "description": "Mem0 memory layer for AI applications. Add persistent memory, personalization, and semantic search.",
-      "version": "0.2.10"
+      "version": "0.2.11"
     }
   ]
 }
--- a/docs/core-concepts/memory-evaluation.mdx
+++ b/docs/core-concepts/memory-evaluation.mdx
@@ -345,7 +345,7 @@ When evaluating memory systems, keep these considerations in mind:
   <Card title="Research" icon="flask" href="https://mem0.ai/research">
     Published research papers and technical reports
   </Card>
-  <Card title="Blog Post" icon="newspaper" href="https://mem0.ai/blog/new-algorithm">
+  <Card title="Blog Post" icon="newspaper" href="https://mem0.ai/blog/the-token-efficient-memory-algorithm-now-has-temporal-reasoning">
     Detailed writeup of the new algorithm design and results
   </Card>
   <Card title="Platform Migration" icon="arrow-right" href="/migration/platform-v2-to-v3">
--- a/docs/integrations/antigravity.mdx
+++ b/docs/integrations/antigravity.mdx
@@ -65,6 +65,7 @@ The plugin uses the same shell scripts as Claude Code, Cursor, and Codex — hoo
 | **User prompt** | `UserPromptSubmit` | Searches relevant memories before each message |
 | **Pre-tool** | `PreToolUse` | Blocks MEMORY.md writes, enforces `user_id`/`app_id` on mem0 tools |
 | **Post-tool** | `PostToolUse` | Tracks stats, scans bash errors for related memories |
+| **Stop** | `Stop` | Stores a session summary when the session ends |
 
 ## Troubleshooting
 
--- a/docs/integrations/claude-code.mdx
+++ b/docs/integrations/claude-code.mdx
@@ -64,7 +64,7 @@ Add the Mem0 MCP server directly with a single command:
 npx mcp-add \
   --name mem0-mcp \
   --type http \
-  --url "https://mcp.mem0.ai/mcp" \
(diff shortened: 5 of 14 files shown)


kartik-mem0:
## Code Review

**Verdict: LGTM — ready to squash-merge**

---

### Strengths

- **Root cause fixed correctly.** The pre-serialization on line 176 (`json.dumps(files[:20])`) was the precise bug: it turned a list into a JSON string *before* `json.dumps(body)` ran, giving double-encoded output. Removing it and letting the outer serialization handle the list is the minimal, correct fix.
- **All four editors covered by one change.** `on_stop.sh`, `on_stop_cursor.sh`, the Codex hooks, and the Antigravity hooks all route through the single shared `scripts/capture_session_summary.py` — verified via hook entry points. No editor-specific copies exist.
- **No same-pattern bug elsewhere.** Checked `capture_compact_summary.py` and all sibling scripts: none set a list-valued metadata field via `json.dumps` inside the dict. The bug was isolated to one call site.
- **Test patches at the right depth.** The regression test intercepts `urllib.request.urlopen`, decodes the raw wire bytes, and asserts both that `files_touched` is a `list` instance and that the escaped-string artifact (`\"`) is absent. Three independent failure signals for the same bug.
- **Version bumps consistent** across all six version-bearing manifests (the `.agents/plugins/marketplace.json` uses a source-pointer pattern and correctly has no version field to update).
- **CHANGELOG backfill is legitimate.** The rerank (`should_rerank`, `MEM0_RERANK`) content was never documented in any prior CHANGELOG entry; PR #5690 merged without one. Documenting it in 0.2.11 is acceptable backfill.

---

### Issues

#### Critical
None.

#### Important
None.

#### Minor

**`integrations/mem0-plugin/tests/test_capture_session_summary.py` — redundant sys.path fixture**

`conftest.py` already defines an `autouse=True` fixture `_scripts_on_path` that inserts `scripts/` into `sys.path` for every test in the directory. The new test file adds its own `_scripts_path` autouse fixture that does the same thing. Both fire for every test in this file. The `if abs_scripts not in sys.path` guard prevents a double-insert, so there is no runtime failure — it's just dead code. Consider removing `_scripts_path` from the test file and relying on `conftest.py` like the other test files do.

---

### Evidence

| Check | Result |
|-------|--------|
| CI Gate | PASS (all required checks) |
| Vercel/preview | Ignored (advisory, fails on every fork PR) |
| Red-green proof | RED: `test_files_touched_is_json_array_not_double_encoded FAILED` with `got str: '["mem0/memory/main.py", "src/client/index.ts"]'` when main impl restored; GREEN: `2 passed` on PR head |
| Ruff lint (changed files) | All checks passed |
| Duplicate PR cluster | None — only #5806 touches `files_touched` |
| Consumers of `files_touched` | None in the plugin codebase — write-only metadata field; no backward-compat concern |

---

### Recommendations

- Squash-merge when ready (`gh pr merge --squash`).
- Remove the `_scripts_path` fixture from `test_capture_session_summary.py` at your convenience (not a blocker).

---

Could you open a short tracking issue for this bug? Something like "files_touched double-JSON-encoded in session-summary metadata" with a one-liner description would give the fix a permanent home for searchability — especially handy if users report the escaped-blob symptom from stored memories made before this fix.

deshraj (approved):

PR #5815, opened 2026-06-24
Harden OpenMemory backup import zip handling

### LINKED ISSUE #5814: OpenMemory backup import reads zip members without size limits
### Summary

`POST /api/v1/backup/import` in OpenMemory reads the uploaded zip and selected zip members fully into memory without any compressed-size, uncompressed-size, compression-ratio, or file-count limits.

A small compressed upload can therefore force the server to allocate a much larger decompressed `memories.json` or `memories.jsonl.gz` buffer during import.

### Affected code

In `openmemory/api/app/routers/backup.py`, the import path does:

```python
content = await file.read()
with zipfile.ZipFile(io.BytesIO(content), "r") as zf:
    ...
    sqlite_data = json.loads(zf.read(sqlite_member))
    memories_blob = zf.read(memories_member) if memories_member else None
```

Both the upload body and the decompressed zip members are materialized in memory.

### Why this matters

Backup import is a user-facing restore endpoint. A crafted zip file with a very high compression ratio can be small on the wire but expand to a large in-memory buffer when `zf.read()` is called. This can cause high memory usage, slow imports, or worker OOM, depending on deployment limits.

### Minimal local reproduction

This does not require running the full API. It demonstrates the same `zipfile.ZipFile(...).read(...)` behavior used by the import route:

```python
import io
import json
import zipfile

large_json = {
    "memories": [
        {"id": "0" * 32, "content": "x" * (1024 * 1024), "metadata": {}}
        for _ in range(80)
    ]
}

raw = json.dumps(large_json).encode()

buf = io.BytesIO()
with zipfile.ZipFile(buf, "w", compression=zipfile.ZIP_DEFLATED, compresslevel=9) as zf:
    zf.writestr("memories.json", raw)

compressed = buf.getvalue()

with zipfile.ZipFile(io.BytesIO(compressed), "r") as zf:
    info = zf.getinfo("memories.json")
    data = zf.read("memories.json")

print("zip bytes:", len(compressed))
print("compressed member bytes:", info.compress_size)
print("uncompressed member bytes:", info.file_size)
print("read bytes:", len(data))
print("ratio:", info.file_size / max(info.compress_size, 1))
```

On my local machine this produced approximately:

```text
zip bytes: 84453
compressed member bytes: 84329
uncompressed member bytes: 83892094
read bytes: 83892094
ratio: 994.8
```

### Expected behavior

The import route should reject oversized or suspicious backup archives before fully materializing them in memory.

Possible mitigations:

- enforce a maximum upload size before `await file.read()`;
- inspect `ZipInfo.file_size`, `compress_size`, and member count before reading;
- reject excessive compression ratios;
- cap `memories.json` and `memories.jsonl.gz` uncompressed sizes separately;
- stream/decode `memories.jsonl.gz` incrementally instead of reading it all at once;
- return a clear `400`/`413` error when limits are exceeded.

### Actual behavior

The current route accepts the zip, reads the whole upload into memory, and then reads selected zip members fully into memory with no limit checks.

yinli-systems:
Fixes #5814.

## What changed
- Limit OpenMemory backup import uploads to 100 MiB before reading the full request body.
- Validate zip file member count, per-member uncompressed size, and compression ratio before extracting backup members.
- Preserve intentional HTTP errors from the backup parser instead of collapsing them into a generic invalid-zip response.
- Add unit coverage for oversized members, excessive file counts, high compression ratios, and normal member reads.

## Validation
- `MEM0_DIR=/tmp/mem0-test MEM0_TELEMETRY=false OPENAI_API_KEY=test-key ../../.venv/bin/python -m pytest tests/test_backup.py -q`
- `MEM0_DIR=/tmp/mem0-test MEM0_TELEMETRY=false OPENAI_API_KEY=test-key ../../.venv/bin/python -m pytest tests/test_mcp_server.py -q` currently has 18 passing tests and 3 existing route-registration assertion failures under this local dependency set; the failures are unrelated to backup import and show the test app only containing FastAPI docs routes.

### DIFF
--- a/openmemory/api/app/routers/backup.py
+++ b/openmemory/api/app/routers/backup.py
@@ -23,6 +23,11 @@ from uuid import uuid4
 
 router = APIRouter(prefix="/api/v1/backup", tags=["backup"])
 
+MAX_BACKUP_UPLOAD_BYTES = 100 * 1024 * 1024
+MAX_BACKUP_MEMBER_BYTES = 100 * 1024 * 1024
+MAX_BACKUP_MEMBER_COUNT = 16
+MAX_BACKUP_COMPRESSION_RATIO = 100
+
 class ExportRequest(BaseModel):
     user_id: str
     app_id: Optional[UUID] = None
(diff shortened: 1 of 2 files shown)


yinli-systems:
I updated the branch so the commit now uses my GitHub noreply email and includes a Signed-off-by trailer. The previous CLA status was caused by the original commit author email not being linked to GitHub. The current remaining failing status is Vercel authorization, which appears to be a deployment permission check rather than a code/test failure.

kartik-mem0:
Thanks for the report. The OpenMemory project has been deprecated and is no longer maintained. Apologies for the inconvenience — we should have communicated this sooner. Closing this issue.

PR #5820, opened 2026-06-24
fix(anthropic): honor response_format in generate_response (revives #4939 per review)

ac12644:
## Description

`AnthropicLLM.generate_response` accepts `response_format` but silently ignores it. Anthropic has no OpenAI-style `response_format`, so the shared extraction call site (`memory/main.py`, `{"type": "json_object"}`) gets no structured-output guarantee — Claude wraps JSON in ``` fences / prefixes prose, which the pipeline only salvages via `remove_code_blocks` + `extract_json`.

This honors `response_format` directly in `anthropic.py`:
- `json_object` → assistant **prefill** (`{`) + a "JSON only, no fences" instruction, so the response starts as clean JSON.
- `json_schema` → Anthropic native structured outputs via `output_config.format`.

## Credit / provenance

This is **not original work** — it revives prior contributions:
- The `anthropic.py` fix + tests are **@limsaehyun**'s (#4484) — preserved as the commit author here.
- **@mb-dev** carried it forward in **#4939**, where @kartik-mem0 reviewed: *"the structured outputs approach is the right call"*, but flagged that #4939's `main.py` switch to `json_schema` breaks Gemini/Ollama/OpenAI/LMStudio.

Per that review, this PR is **scoped to `anthropic.py` only** — `main.py`/`prompts.py` are **unchanged** (still `json_object`), so Anthropic users get the fix and no other provider breaks. Rebased onto current `main`; one stale tools test from the original (incompatible with the post-#5537 tool_choice format) was dropped.

## Type of Change

## Test Coverage

- [x] Unit tests (@limsaehyun's): `tests/llms/test_anthropic.py` — `json_object` prefill returns clean JSON, with/without system message, `json_schema` → `output_config`, and no-`response_format` unchanged. 13 passed, ruff clean.
- Approach also verified against the live Anthropic API: `response_format={"type":"json_object"}` returns directly-parseable JSON (vs ``` -fenced before), and full `Memory.add()` extraction works end-to-end.

## Checklist

- [x] Code follows style guidelines (ruff clean)
- [x] Self-review performed

### DIFF
--- a/mem0/llms/anthropic.py
+++ b/mem0/llms/anthropic.py
@@ -9,6 +9,7 @@ except ImportError:
 from mem0.configs.llms.anthropic import AnthropicConfig
 from mem0.configs.llms.base import BaseLlmConfig
 from mem0.llms.base import LLMBase
+from mem0.memory.utils import extract_json
 
 
 class AnthropicLLM(LLMBase):
@@ -137,7 +138,36 @@ class AnthropicLLM(LLMBase):
             params["tools"] = tools
             params["tool_choice"] = {"type": tool_choice}
 
+        wants_json_prefill = False
+        if response_format and isinstance(response_format, dict):
+            format_type = response_format.get("type")
+            if format_type == "json_schema" and "json_schema" in response_format:
+                # Use Anthropic structured outputs via output_config
+                schema = response_format["json_schema"]
+                if isinstance(schema, dict) and "schema" in schema:
+                    schema = schema["schema"]
+                params["output_config"] = {"format": {"type": "json_schema", "schema": schema}}
+            elif format_type == "json_object":
+                # Use assistant prefill to guide JSON output
+                wants_json_prefill = True
+                json_instruction = (
+                    "\n\nYou must respond with valid JSON only. "
+                    "Do not include any other text, markdown formatting, or code fences."
+                )
+                if params.get("system"):
+                    params["system"] += json_instruction
+                else:
+                    params["system"] = json_instruction.strip()
+                params["messages"] = list(params["messages"]) + [
+                    {"role": "assistant", "content": "{"}
+                ]
+
         response = self.client.messages.create(**params)
+
+        if wants_json_prefill:
+            # Re-attach the prefilled "{" and strip any stray wrapping.
+            return extract_json("{" + response.content[0].text)
+
         return self._parse_response(response, tools)
 
     def _parse_response(self, response, tools):
--- a/tests/llms/test_anthropic.py
+++ b/tests/llms/test_anthropic.py
@@ -2,15 +2,18 @@ from unittest.mock import Mock, patch
 
 import pytest
 
+
 pytest.importorskip("anthropic", reason="anthropic package not installed")
 
 from mem0.configs.llms.anthropic import AnthropicConfig
 from mem0.configs.llms.base import BaseLlmConfig
+
 from mem0.llms.anthropic import AnthropicLLM
 
 
 @pytest.fixture
 def mock_anthropic_client():
+
     with patch("mem0.llms.anthropic.anthropic") as mock_anthropic:
         mock_client = Mock()
         mock_anthropic.Anthropic.return_value = mock_client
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## LGTM — fix(anthropic): honor response_format in generate_response

**Verdict: ready to squash-merge.**

---

### Strengths

- **Root cause fixed at the right layer.** The provider is the correct place to translate the shared `response_format` interface into Anthropic-specific API calls. `main.py` and callers stay untouched — exactly the scope that was requested in the prior review.
- **Both response_format modes handled correctly.** `json_object` → assistant prefill (`{`) + system instruction; `json_schema` → `output_config.format` (the current Anthropic structured-outputs API). The prefill approach is the idiomatic Anthropic technique; `output_config` format was confirmed from the vendor SDK changelog.
- **`extract_json` re-use is appropriate.** After prefill, `"{" + response.content[0].text` reconstructs the full object and `extract_json` trims any stray wrapping. Clean and consistent with how other callers handle Anthropic prose-contamination.
- **No-format regression protected.** The new condition short-circuits when `response_format` is `None`, so the existing code path (`_parse_response`) is untouched. Verified by `test_generate_response_without_response_format`.
- **Tests bite.** Red-green verified locally: 4 tests fail on main's `anthropic.py`, pass on the PR head. All 13 tests pass; ruff clean.
- **Correctly scoped.** Prior review noted that switching `main.py` from `json_object` to `json_schema` globally breaks Gemini/Ollama/OpenAI. This PR avoids that entirely.

---

### Issues

**Important — TS parity gap (follow-up, not a blocker on this PR)**

`mem0-ts/src/oss/src/llms/anthropic.ts` — `generateResponse` accepts `responseFormat` but never acts on it (same silent-ignore bug). This PR closes the Python side; the TypeScript OSS Anthropic provider remains unfixed. Worth a follow-up PR to the TS side so the two SDKs match.

**Minor — simultaneous `response_format + tools` case is silently lossy**

If a caller passes both `response_format={"type": "json_object"}` and `tools=[...]`, the code sets the prefill message AND tools params, then returns via `extract_json` — bypassing `_parse_response`, so any `tool_use` blocks in the response are silently dropped. No current caller in `main.py` does this, but the method signature allows it. A short code comment or guard would make the limitation explicit (`response_format is ignored when tools are active` or an early return).

---

### Recommendations

1. Follow-up PR for `mem0-ts/src/oss/src/llms/anthropic.ts` — same `json_object` prefill + `json_schema`/`output_config` logic in TypeScript.
2. Optional: add a comment near the `wants_json_prefill` block noting it is incompatible with simultaneous tool calls.

---

### Evidence

| Check | Result |
|---|---|
| CI Gate | pass |
| Python SDK 3.10 / 3.11 / 3.12 build | pass (all three) |
| Vercel preview | fail (advisory — fork perms, ignored per policy) |
| Targeted tests on PR head (`tests/llms/test_anthropic.py`) | 13 passed |
| Ruff lint (`mem0/llms/anthropic.py`, `tests/llms/test_anthropic.py`) | clean |
| Red-green: tests with reverted `anthropic.py` | 4 failed (json_object ×3, json_schema ×1) |
| Red-green: tests restored to PR head | 13 passed |
| Duplicate cluster checked (#4484, #4939) | both open; #5820 is the correctly-scoped fix per prior review |
| TS counterpart reviewed | `mem0-ts/src/oss/src/llms/anthropic.ts` — parity gap noted above |

Could you open a tracking issue for this fix so we have a place to link the TS follow-up?

ac12644:
Hi @kartik-mem0 — CI's green and it's still showing mergeable. Anything else needed from my side before this can be squash-merged?

ac12644:
Opened #6203 to track the TypeScript parity fix. I'll follow up with a mem0-ts `anthropic.ts` PR that mirrors this logic (json_object prefill + json_schema/output_config), and I'll add the comment/guard for the simultaneous `response_format` + `tools` case there too.

ac12644:
TS follow-up is up: #6205 (closes #6203). It mirrors this PR's logic in `mem0-ts/src/oss/src/llms/anthropic.ts` — json_object prefill + json_schema → `output_config.format`, and skips `response_format` when tools are active. Verified against the live Anthropic API too (json_object returns clean JSON vs ```-fenced without it). Happy to have both merged together whenever it's convenient.

ac12644:
Rebased on latest main and it's green again (27/27 in `test_anthropic`), so should be mergeable now.

@kartik-mem0 thanks for the thorough trace on #6203. Agreed it's the same gap live in both SDKs, not a Python-behind-TS thing. This is the Python fix, #6205 is the TS one (also rebased + green). No rush, just flagging both are unblocked whenever you get a chance to review.

PR #5821, opened 2026-06-24
Enable LLM-driven memory update and delete events

utkucanaytac:
## Description

The current prompt pushes the model toward ADD-only behavior. This prevents the LLM from deciding when a new user message should update or delete an existing memory.

For example, if a user first says their favorite food is pasta and later says they no longer eat pasta because they are on a diet and their favorite food is salad, the current flow can store both facts as separate ADD memories. This creates confusion during retrieval. The correct behavior is to update the existing preference.

Besides, even if the prompt returns UPDATE or DELETE events, the current add pipeline does not handle those events, so conflicting memories can remain in the vector store.

This change:
- introduces a memory management prompt that allows ADD, UPDATE, DELETE, and NONE
- uses that prompt in the sync and async add pipelines
- handles UPDATE and DELETE events by resolving the LLM-provided memory id
- keeps ADD behavior for genuinely new memories
- adds tests for update and delete event handling

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Tested locally with:

.venv/bin/python -m pytest tests/memory/test_main.py -k "memory_management"
.venv/bin/python -m pytest tests/memory/test_main.py

Results:

2 passed, 44 deselected
46 passed

## Checklist

### DIFF
--- a/docs/open-source/configuration.mdx
+++ b/docs/open-source/configuration.mdx
@@ -136,6 +136,29 @@ memory = Memory.from_config_file("config.yaml")
 ## Tune component settings
 
 <AccordionGroup>
+  <Accordion title="Memory management">
+    By default, Mem0 uses ADD-only extraction to preserve the full history of stored memories.
+    If you want the LLM to resolve conflicting memories during `add()` calls, set
+    `enable_memory_management=True` in `MemoryConfig` or include `"enable_memory_management": True`
+    in your configuration dictionary.
+
+    When enabled, the extraction step can return `ADD`, `UPDATE`, `DELETE`, or `NONE` actions.
+    This lets Mem0 update or remove an existing memory when a new message contradicts, negates,
+    or supersedes it. Because this mode allows LLM-driven updates and deletes against the memory
+    store, it is opt-in and remains disabled by default.
+
+```python
+from mem0 import Memory
+
+memory = Memory.from_config({
+    "enable_memory_management": True,
+    "vector_store": {
+        "provider": "qdrant",
+        "config": {"host": "localhost", "port": 6333},
+    },
+})
+```
+  </Accordion>
   <Accordion title="Vector store collections">
     Name collections explicitly in production (`collection_name`) to isolate tenants and enable per-tenant retention policies.
   </Accordion>
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -55,6 +55,10 @@ class MemoryConfig(BaseModel):
         description="Custom instructions for fact extraction",
         default=None,
     )
+    enable_memory_management: bool = Field(
+        description="Enable LLM-driven ADD, UPDATE, DELETE, and NONE memory actions during extraction",
+        default=False,
+    )
 
 
 class AzureConfig(BaseModel):
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -465,6 +465,63 @@ def get_update_memory_messages(retrieved_old_memory_dict, response_content, cust
 # Ported from platform/backend/shared/core/config/prompts.py
 # ---------------------------------------------------------------------------
 
+MEMORY_MANAGEMENT_PROMPT = """
+You are a memory manager. Compare New Messages with Existing Memories and return
+the operations required to leave the memory store in a consistent final state.
+
+Allowed events:
+- ADD: a genuinely new durable fact. The id may be omitted.
+- UPDATE: replace an existing fact. Use the exact Existing Memory id.
+- DELETE: remove an existing fact. Use the exact Existing Memory id.
+- NONE: the existing fact already represents the new information.
+
+Rules:
+- Do not leave contradictory or superseded memories in the store.
+- Existing Memories are authoritative current state. When a New Message changes
+  that state, you must modify the matching Existing Memory instead of creating
+  a second memory.
+- Preference changes such as "no longer likes X; likes Y", "now prefers Y",
+  "changed from X to Y", or "stopped X" must UPDATE the existing memory about
+  that preference, not ADD a second contradictory memory.
+- For UPDATE and DELETE, never invent an id. Use an id from Existing Memories.
+- Extract facts only from New Messages. Existing Memories are comparison context.
+
+Examples:
+
+Existing Memories:
+[{"id":"0","text":"User likes hiking"}]
+New Messages:
+User no longer likes hiking, he likes tennis.
+Output:
+{"memory":[{"id":"0","text":"User likes tennis","event":"UPDATE","old_memory":"User likes hiking"}]}
+
+Existing Memories:
+[{"id":"0","text":"User supports Beşiktaş"}]
+New Messages:
+Artık Beşiktaş'ı tutmuyorum, artık Fenerbahçeliyim.
+Output:
+{"memory":[{"id":"0","text":"User supports Fenerbahçe","event":"UPDATE","old_memory":"User supports Beşiktaş"}]}
+
+Existing Memories:
+[{"id":"0","text":"User likes hiking"}]
+New Messages:
+User likes tennis.
+Output:
+{"memory":[{"id":null,"text":"User likes tennis","event":"ADD"}]}
+
+Return JSON only:
+{
+  "memory": [
+    {
+      "id": "<existing id for UPDATE/DELETE/NONE>",
+      "text": "<final memory text>",
+      "event": "ADD|UPDATE|DELETE|NONE",
+      "old_memory": "<previous text for UPDATE>"
+    }
+  ]
+}
+"""
+
 ADDITIVE_EXTRACTION_PROMPT = """
 
 # ROLE
@@ -1025,10 +1082,10 @@ def generate_additive_extraction_prompt(
     custom_instructions=None,
     use_input_language=False,
 ):
-    """Build the user prompt for additive (ADD-only) extraction with linking.
+    """Build the user prompt for memory management extraction.
 
-    Pairs with ADDITIVE_EXTRACTION_PROMPT system prompt.
-    The LLM will produce only ADD operations, with optional linked_memory_ids.
+    The LLM compares new messages with existing memories and produces
+    ADD, UPDATE, DELETE, or NONE operations.
     """
     current_date, observation_date = _resolve_dates(current_date, timestamp)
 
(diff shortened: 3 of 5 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- The root problem is real: ADD-only extraction creates contradictory preference memories (e.g. "likes pasta" and "no longer likes pasta, likes salad" coexisting). This PR addresses it at the right point in the pipeline.
- The `uuid_mapping` anti-hallucination guard (`main.py` Phase 1) is the correct design — integer IDs 0–9 are assigned to the top-10 search results; any LLM-fabricated ID is silently skipped with a warning log. The worst-case outcome from a hallucinated ID is a no-op, not a phantom delete.
- Sync and async paths are kept in structural sync (`_add_to_vector_store` sync Phase 3 and async Phase 3 are mirror images, with the async path correctly wrapping `embed` and `_update_memory`/`_delete_memory` in `asyncio.to_thread`/`await`).
- Red-green proof confirms the tests bite the actual behavior change.

### Issues

#### Critical (must fix)

1. **`mem0/memory/main.py` Phase 3 loop (sync ~line 866, async ~line 2451) — no exception handling around `_update_memory` / `_delete_memory`**

   `_update_memory` (line 1890–1891) raises `ValueError` if the resolved UUID no longer exists in the vector store at dispatch time. `_delete_memory` (line 1944–1945) does the same. In a concurrent scenario — two agents writing for the same user, or any process that deletes a memory between Phase 1 (search) and Phase 3 (dispatch) — the uncaught exception aborts the entire `_add_to_vector_store` call, discarding all other pending UPDATE/DELETE/ADD actions that were queued after the failing entry, and surfacing a raw exception to the caller instead of a result list. The same flaw exists in the async path.

   Fix: wrap each dispatch call in its own try/except:
   ```python
   try:
       self._update_memory(memory_id=memory_id, data=text,
                           existing_embeddings=update_embeddings, metadata=deepcopy(metadata))
   except Exception as e:
       logger.warning("UPDATE failed for memory %s: %s", memory_id, e)
       continue
   ```
   Apply the same pattern to `_delete_memory` and to both sync and async paths.

2. **`mem0/memory/main.py` lines 858–860 (sync), 2431–2433 (async) — unconditional system-prompt swap with no opt-in gate**

   `ADDITIVE_EXTRACTION_PROMPT` was intentionally named "additive" to signal a conservative ADD-only contract: memories are never auto-deleted during extraction. Every existing `Memory()` user who upgrades will start having the LLM decide to delete or update their memories without opting in. This is a data-loss risk, not just a behavior change.

   A user who passes `prompt=` to `add()` to control extraction behavior also has no way to opt out — `prompt=` has always flowed into the user turn as `custom_instructions`, not as a system-prompt override, so it cannot gate the new behavior.

   Fix: add a `MemoryConfig` flag (e.g. `enable_memory_management: bool = False`) that activates `MEMORY_MANAGEMENT_PROMPT` when `True` and keeps `ADDITIVE_EXTRACTION_PROMPT` as the default. Once the behavior is validated in production, promote the flag's default in a subsequent minor release. This lets teams upgrade safely without silent data mutations.

#### Important (should fix)

3. **`mem0/configs/prompts.py` lines ~1115–1125 + `MEMORY_MANAGEMENT_PROMPT` — same schema and example injected twice per LLM call**

   `MEMORY_MANAGEMENT_PROMPT` (the system prompt) already contains the complete ADD/UPDATE/DELETE/NONE ruleset, the exact hiking→tennis UPDATE example, and the full JSON schema. The `generate_additive_extraction_prompt()` change appends a `## Memory Management Requirement` section to the user turn that repeats the same rules, the same verbatim hiking→tennis example, and the same `{"memory":[{"id":...,"text":...,"event":...,"old_memory":...}]}` schema. The system prompt is sufficient on its own. The duplicated user-turn section bloats every `add()` call by ~600 tokens and is a maintenance surface where the two copies can diverge.

   Fix: remove the appended `## Memory Management Requirement` section from `generate_additive_extraction_prompt()`. The system prompt already carries the full behavioral contract; the user prompt should carry only structured data (messages, existing memories, custom user instructions).

4. **`tests/memory/test_main.py` — async path has zero test coverage for the new Phase 3 code**

   `TestMemoryManagementActions` tests only the synchronous `Memory._add_to_vector_store`. The async `AsyncMemory._add_to_vector_store` adds ~44 lines of independent Phase 3 dispatch code (same logic, but with `await`/`asyncio.to_thread` wrapping) and has no tests. A bug in the async wrapping (e.g. missing `await`, wrong coroutine chaining) would be silent.

   Fix: add `TestAsyncMemoryManagementActions` with `@pytest.mark.asyncio`-decorated tests using `AsyncMock` for `_update_memory` and `_delete_memory`, mirroring the two existing sync tests.

5. **TypeScript SDK parity gap — `mem0-ts/src/oss/src/memory/index.ts` line 834 still uses `ADDITIVE_EXTRACTION_PROMPT`**

   After this PR, Python `Memory()` users get conflict resolution (UPDATE/DELETE); TypeScript `Memory` users continue accumulating contradictory memories. Please open a follow-up issue tracking the TS SDK equivalent change, and reference it in this PR's description so it's not lost.

#### Minor (optional)

6. `generate_additive_extraction_prompt` function name now describes ADD-only behavior while the function produces ADD/UPDATE/DELETE/NONE instructions. Consider renaming to `generate_extraction_prompt` or noting it in the docstring.

7. `update_embeddings` dict is initialized before the loop and accumulates across iterations; `_update_memory` only reads the key matching the current `data`. Moving `embedding = self.embedding_model.embed(text, "update")` and passing `{text: embedding}` directly as `existing_embeddings` is cleaner and avoids the cross-iteration accumulation.

8. `ADDITIVE_EXTRACTION_PROMPT` is still imported in `main.py` line 20 but unused after the PR lands. Remove the import.

### Evidence
- **CI:** CI Gate = `action_required` — fork PR (utkucanaytac/mem0); no test/lint/build jobs have run yet; Vercel check is advisory (expected on all fork PRs, not merge-blocking)
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_main.py -k "memory_management" -v` → 2 passed in 5.22s
- **Red-green proof:** Tests fail with main implementation (both `_update_memory` and `_delete_memory` called 0 times), pass with PR implementation ✓
- **Full test_main.py suite:** 46 passed (no regressions)
- **Lint:** `ruff check mem0/memory/main.py mem0/configs/prompts.py` → all checks passed
- **TS SDK:** `mem0-ts/src/oss/src/memory/index.ts:834` — confirmed still uses `ADDITIVE_EXTRACTION_PROMPT`
- **Skeptic-refuted finding:** `linked_memory_ids` silent drop — refuted; entity linking is driven by `extract_entities_batch()` independently and does not read `linked_memory_ids` from LLM output
- **Could NOT verify:** live LLM response quality with the new prompt, multi-process race reproduction (requires live concurrent environment)

### Assessment
Ready to merge: **With fixes** — please address the two Critical items (exception handling in Phase 3 and an opt-in gate for the behavior change) and the async test gap; I'll re-review once you push.

For tracking on our side, could you open a quick issue describing the problem this PR fixes and link it here? Makes it easier to prioritize and reference.

utkucanaytac:
Thanks for the review. Addressed the requested changes:

- added enable_memory_management as an opt-in flag, keeping ADD-only extraction as the default
- gated MEMORY_MANAGEMENT_PROMPT behind the new flag
- added per-action exception handling for UPDATE and DELETE in sync and async paths
- removed duplicated memory-management instructions from the user extraction prompt
- added async memory-management tests and failure-isolation coverage

Tests:
- .venv/bin/python -m pytest tests/memory/test_main.py -k "memory_management" → 7 passed
- .venv/bin/python -m pytest tests/memory/test_main.py → 51 passed
- .venv/bin/python -m ruff check mem0/memory/main.py mem0/configs/prompts.py mem0/configs/base.py tests/memory/test_main.py → All checks passed

Tracking issue: #5867

kartik-mem0:
Verified against cf191a95.

All five findings from the prior review are fixed: the two Criticals (exception handling in Phase 3 and the opt-in gate), the dual-injection in the user prompt, the async test gap, and the TS parity tracking issue (#5867, confirmed open). The flag design is clean, sync/async structural parity is exact, and the 7 new tests — 4 sync + 3 async, including the failure-isolation case — all red-green proved.

One new Critical surfaced in this pass.

---

### Critical (must fix)

**`mem0/memory/main.py` — embed call is outside the try/except in both UPDATE branches**

The exception isolation that was added correctly for `_update_memory` / `_delete_memory` doesn't cover the embed call that precedes it. Sync path (around line 893):

```python
update_embeddings = {text: self.embedding_model.embed(text, "update")}  # outside try/except
try:
    self._update_memory(...)
except Exception as e:
    logger.warning(...)
    continue
```

Async path has the same structure with `await asyncio.to_thread(self.embedding_model.embed, ...)`. An embedding API timeout or rate-limit error here propagates unhandled out of the `for mem in extracted_memories:` loop, discarding all remaining events — subsequent DELETEs, other UPDATEs, and the entire ADD batch — and bypassing `db.save_messages()`. The fix is a one-line scope change: move the embed call inside the `try:` block for both sync and async paths, and add a test that mocks `embedding_model.embed` to raise and verifies the subsequent DELETE still executes.

---

### Important (should fix)

**`mem0/memory/main.py` line ~823 and ~2416 — `AGENT_CONTEXT_SUFFIX` appended to `MEMORY_MANAGEMENT_PROMPT` when agent-scoped**

The append is unconditional:
```python
system_prompt = MEMORY_MANAGEMENT_PROMPT if enable_memory_management else ADDITIVE_EXTRACTION_PROMPT
if is_agent_scoped:
    system_prompt += AGENT_CONTEXT_SUFFIX  # always appended
```

`AGENT_CONTEXT_SUFFIX` instructs the LLM to populate an `attributed_to` field (`prompts.py:947–957`). `MEMORY_MANAGEMENT_PROMPT`'s output schema has no `attributed_to` slot — the instruction is sent with no corresponding schema or consumer, and Phase 3 never reads it. This was authored against `ADDITIVE_EXTRACTION_PROMPT`. Fix: gate the suffix on the prompt in use (`if is_agent_scoped and not enable_memory_management`), or add agent framing to `MEMORY_MANAGEMENT_PROMPT` and consume `attributed_to` in Phase 3.

**`mem0/memory/main.py` Phase 3 — N+1 `vector_store.get()` per UPDATE/DELETE**

`_delete_memory(memory_id=memory_id)` without `existing_memory` triggers `self.vector_store.get(vector_id=memory_id)` per call (`main.py:2021`). `_update_memory` does the same unconditionally (`main.py:1956`). Phase 1 already fetched all these records via `vector_store.search()`; the PR stores only the UUID in `uuid_mapping` and discards the payload. At the `enable_memory_management` flag's default (off) this is a no-op, but once enabled every `add()` with UPDATEs or DELETEs adds one network round-trip per action. For DELETE, passing the already-fetched object as `existing_memory=` removes it immediately. UPDATE would need a small signature extension.

**`mem0/memory/main.py` Phase 3 — UPDATE embeddings not batched**

Each UPDATE in the Phase 3 loop calls `self.embedding_model.embed(text, "update")` individually. Phase 4 calls `embed_batch()` for all ADD texts. N UPDATE events produce N separate embed calls before Phase 4's batch. Collecting UPDATE texts before the loop and making a single `embed_batch("update")` call would keep the hot path consistent with the rest of the pipeline.

**`mem0/configs/base.py` — `enable_memory_management` has no docs page**

The field enables LLM-driven DELETEs against the user's memory store. `custom_instructions` (the adjacent field) has its own `docs/open-source/features/custom-instructions.mdx` page. A config-reference entry — what the flag enables, the opt-in rationale, behavior difference from default — is the minimum before this ships in release notes.

---

### Minor (optional)

- `mem0/memory/main.py:822, 2415`: `getattr(self.config, "enable_memory_management", False) is True` — `getattr` fallback is dead (field is declared in MemoryConfig) and `is True` is non-idiomatic for a Pydantic `bool`; `self.config.enable_memory_management` is sufficient.
- `tests/memory/test_main.py`: no test for the hallucinated-ID guard (LLM returns `{"id": "99"}` → `_delete_memory` never called); no test for the NONE path (no crash, no action, not in returned list). Both are small additions to the existing fixture.

---

### Evidence
- **CI:** Fork PR — CI Gate pending maintainer approval; Vercel advisory (expected, non-blocking); CLA signed
- **Targeted tests:** `pytest tests/memory/test_main.py -k "memory_management" -v` → **7 passed** (4 sync + 3 async)
- **Red-green proof:** all 7 tests ERROR on main (MemoryConfig has no `enable_memory_management` field) → 7 passed on PR head ✓
- **Full test_main.py:** 51 passed (no regressions)
- **Lint:** `ruff check mem0/memory/main.py mem0/configs/prompts.py mem0/configs/base.py tests/memory/test_main.py` → all clean
- **Issue #5867:** open, correctly describes the root cause
- **Could NOT verify:** live LLM behavior under the combined `MEMORY_MANAGEMENT_PROMPT + AGENT_CONTEXT_SUFFIX` prompt; multi-process concurrent embed failure

### Assessment
Ready to merge: **Not yet** — the embed exception-isolation gap is a one-line scope fix; the AGENT_CONTEXT_SUFFIX incompatibility and the missing docs are the other two items worth addressing before this ships. The core design — opt-in flag, uuid_mapping guard, delegation to `_update_memory`/`_delete_memory`, sync/async parity — is solid.

utkucanaytac:
Thanks. Addressed follow-up memory management review feedback

- Moved UPDATE embedding calls inside the try/except block.
- Kept the agent attribution suffix out of the memory management prompt.
- Reused Phase 1 search results for DELETE actions to avoid extra vector fetches.
- Documented the (enable_memory_management) configuration flag.
- Added tests for embed failure isolation in both sync and async paths.
- Added tests for agent prompt gating, unknown memory IDs, and NONE events.

Note: Left UPDATE embedding batching as a follow-up.

PR #5824, opened 2026-06-24
feat(vector-stores): add Databricks provider to TypeScript OSS SDK

### LINKED ISSUE #5782: feat(ts-sdk): add Databricks vector store
## Summary

The Python SDK supports **Databricks** as a vector store provider, but the TypeScript OSS SDK (`mem0ai/oss`) does not. Add it to bring the TS SDK to parity.

| | |
|---|---|
| Python reference | `mem0/vector_stores/databricks.py` |
| Target file (TypeScript) | `mem0-ts/src/oss/src/vector_stores/databricks.ts` |
| Suggested implementation | Databricks Vector Search via REST API (+ `@databricks/sql` for SQL ops). |

## Requirements

- [ ] Implement `Databricks` in `mem0-ts/src/oss/src/vector_stores/databricks.ts`, extending `VectorStore` (`mem0-ts/src/oss/src/vector_stores/base.ts`) and mirroring the Python provider's behavior (insert / search / get / update / delete / list / reset).
- [ ] Register the `"databricks"` provider in `mem0-ts/src/oss/src/utils/factory.ts` (VectorStoreFactory).
- [ ] Add `@databricks/sql` to `mem0-ts/package.json` (optional/peer dependency, lazy-imported like other providers).

## Reference pattern

## Notes

Mirror `mem0/vector_stores/databricks.py`.

---

rodboev:
@
## Summary

Adds Databricks vector store support to the TypeScript OSS SDK (`mem0-ts`) so the factory can target Databricks AI Search instead of rejecting the provider at runtime.

- New `DatabricksVectorStore` adapter with endpoint/index readiness checks, service-principal REST auth scoping, storage-optimized filter translation, multi-page query retrieval, full-text request limits, and entity-table isolation
- Wired into `VectorStoreFactory.create` alongside existing providers
- Covered by factory unit tests and mocked compatibility tests for all Databricks-specific edge cases

Closes #5782

## Type of Change

## Test Plan

- `pnpm exec jest --runTestsByPath src/oss/tests/factory.unit.test.ts --runInBand`
- `pnpm exec jest --runTestsByPath src/oss/tests/vector-stores-compat.test.ts --runInBand --testNamePattern="Databricks|provider casing differs"`
- `pnpm exec jest --runTestsByPath src/oss/tests/tsup-externals.test.ts --runInBand`
- `pnpm exec prettier --check src/oss/src/memory/index.ts src/oss/src/vector_stores/databricks.ts src/oss/tests/vector-stores-compat.test.ts`
- `pnpm exec tsup`
@


kartik-mem0:
## Changes Requested

Went through `databricks.ts` end to end against the Python reference (`mem0/vector_stores/databricks.py`) and the existing TS providers (`qdrant.ts`, `pgvector.ts`, `redis.ts`). This is a genuinely well-built provider — a few notes below, but nothing here should be discouraging.

### Strengths
- SQL-injection defenses are consistent and thorough: `validateIdentifier()` / `formatSqlValue()` mirror the Python `_validate_identifier`/`_format_sql_value` helpers and are applied on every interpolated identifier and value across insert/update/delete/get.
- Auth completeness is validated at two independent points (REST client and SQL client), each throwing clearly if neither `accessToken` nor `clientId`/`clientSecret` is supplied.
- The async `initialize()` lifecycle (fire-and-forget kickoff in the constructor + memoized promise + every public method awaiting it first) is a thoughtful TS-idiomatic adaptation rather than a blind port of Python's synchronous eager constructor.
- `@databricks/sql` as a peer dependency, added to `tsup.config.ts`'s `external` array, with a static top-level import — this matches all 20 existing peer-dependency providers in this repo exactly (confirmed against `redis.ts`).
- The mocked compat test suite (`vector-stores-compat.test.ts`) is substantive, not superficial — it asserts real SQL text shape (`INSERT INTO main.default.memories`, quoted values, `array(1, 0, 0)` vector literals) and real REST body shape (`filters_json`), including a dedicated regression test for the `getEntityStore()` case-insensitivity fix in `memory/index.ts`.

### Issues

#### Important (should fix)
1. `mem0-ts/src/oss/src/vector_stores/databricks.ts:676-703` — `list()` runs an unbounded `SELECT memory_id, payload FROM <table>` with no `LIMIT` and no server-side filter, pulling the entire Delta table into the Node process on every call before filtering and slicing to `topK` in JS. The Python reference's `list()` uses the bounded Vector Search index query API (`num_results = top_k or 100`), and both sibling TS providers bound this server-side too — `qdrant.ts` via `scroll({ limit: topK })`, `pgvector.ts` via SQL `LIMIT` plus a separate `COUNT(*)`. On a production-sized table (the scenario this backend exists for), this scales with total table size instead of requested page size. The fix looks small: this file already has safe filter-to-SQL helpers (`buildSimpleDatabricksFilter` / `buildStorageOptimizedDatabricksFilterClause` / `buildDatabricksServerFilters`) used elsewhere for search — reusing them to build a `WHERE` clause and adding a `LIMIT` here would close the gap without a redesign.

#### Minor (optional)
1. `databricks.ts:562-563` — `queryType: "HYBRID"` is accepted uncomplained at construction but `search()` always throws for it since `search()` never receives `query_text`. Worth validating at construction time so misconfiguration fails fast instead of on first search call.
2. Azure AD service-principal auth and the `DIRECT_ACCESS` index type from the Python reference aren't ported here — seems like a reasonable v1 scope cut, just flagging in case it was accidental.
3. `DATBRICKS_SERVER_FILTER_KEYS` (line 12) — typo in the constant name (missing the "A" in Databricks), purely cosmetic.

### Evidence
- **CI:** all real checks are green on the current head (TS SDK build/integration on Node 20 & 22, changelog check, CI Gate, CLA). Only the advisory Vercel preview-deploy check fails, which needs manual authorization and isn't a code signal.
- **Tests:** not re-run locally this pass (read-only review), but the test content was read in full and does exercise real SQL/REST shapes rather than just mock presence.
- **Could not verify:** live behavior against an actual Databricks SQL Warehouse / Vector Search endpoint — this review is based on code tracing and the mocked test suite only.

One more thing worth flagging separately from the code review: the branch currently shows merge conflicts against `main` (`CONFLICTING`/`DIRTY`), so it'll need a rebase before it can land regardless of the above.

### Assessment
Ready to merge: **with fixes** — if you can address the `list()` bounding above and rebase, I think this is in good shape. Since this introduces a brand-new vector store backend, it'll also want a maintainer sign-off on adopting Databricks as a supported target alongside the code fix, but the implementation itself is solid work. Happy to take another look once it's updated.

kartik-mem0 (approved):

PR #4186, opened 2026-03-01
feat: add boto3 timeout configuration for AWS Bedrock client

Br1an67:
## Description

Add `read_timeout` and `connect_timeout` parameters to `AWSBedrockConfig`. When specified, a `botocore.config.Config` object is passed to the boto3 client, overriding the default 60-second read timeout which is too short for large memory extraction operations.

### Usage

```python
config = {
    "llm": {
        "provider": "aws_bedrock",
        "config": {
            "model": "us.anthropic.claude-sonnet-4-20250514-v1:0",
            "read_timeout": 300,
            "connect_timeout": 10,
        }
    }
}
```

Both parameters default to `None` (uses botocore defaults), so this is fully backward-compatible.

Fixes #3825

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Unit tests

```bash
pytest tests/llms/test_aws_bedrock.py -v
```

4 tests covering: default config, custom timeouts, boto3 client receives Config object, no Config when timeouts not set.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes
- [x] My changes generate no new warnings

PR #5832, opened 2026-06-25
feat(bedrock): add configurable boto3 client (timeouts + retries) for AWS Bedrock

### LINKED ISSUE #3825: Add boto3 timeout configuration support for AWS Bedrock client
### 🚀 The feature

When using AWS Bedrock with large language models, boto3 read timeout errors occur because the default botocore read timeout (60 seconds) is insufficient for processing large memories. The `AWSBedrockLLM` class currently doesn't provide a way to configure boto3 client timeouts.

### Error Example
```
Failed to generate response: Read timeout on endpoint URL: "https://bedrock-runtime.us-west-2.amazonaws.com/model/us.anthropic.claude-sonnet-4-20250514-v1%3A0/converse"
```

### Root Cause
In `mem0/llms/aws_bedrock.py`, the boto3 client is created without timeout configuration:

```python
def _initialize_aws_client(self):
    aws_config = self.config.get_aws_config()
    self.client = boto3.client("bedrock-runtime", **aws_config)
```

This uses the default botocore read timeout (~60 seconds), which is too short for:
- Large conversation contexts
- Complex memory extraction operations
- Models with longer processing times

### Proposed Solution

Add timeout configuration support through `AWSBedrockConfig`:

**1. Update `AWSBedrockConfig` class** (`mem0/configs/llms/aws_bedrock.py`):
```python
from botocore.config import Config

class AWSBedrockConfig(BaseLlmConfig):
    def __init__(
        self,
        # ... existing parameters ...
        read_timeout: int = 1000,
        connect_timeout: int = 60,
        **kwargs,
    ):
        # ... existing init code ...
        self.read_timeout = read_timeout
        self.connect_timeout = connect_timeout
    
    def get_boto_config(self) -> Config:
        """Get boto3 Config object with timeout settings."""
        return Config(
            read_timeout=self.read_timeout,
            connect_timeout=self.connect_timeout
        )
```

**2. Update `AWSBedrockLLM` class** (`mem0/llms/aws_bedrock.py`):
```python
def _initialize_aws_client(self):
    try:
        aws_config = self.config.get_aws_config()
        boto_config = self.config.get_boto_config()
        
        self.client = boto3.client(
            "bedrock-runtime", 
            config=boto_config,  # Add timeout config
            **aws_config
        )
        
        self._test_connection()
```

### Environment
- mem0 version: 0.1.118
- boto3 version: 1.40.41+
- Python version: 3.12
- AWS Bedrock models affected: All (especially Claude Sonnet 4.0)

### Motivation, pitch

This is particularly important for:

- Memory extraction from long conversations (mem0's primary use case)
- Complex prompt processing with system prompts

Benefits
- Prevents timeout errors with large conversations
- Configurable per use case (can set shorter/longer timeouts as needed)
- Backward compatible (defaults maintain current behavior)
- Follows AWS best practices (per [AWS documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/timeout-configuration.html))

AWS recommends starting with a read timeout of **3600 seconds** for large model queries to handle variable processing times.

Malhis:
## Linked Issue

  Closes #3825

  Supersedes #4018 (closed, unmerged) and #4186 (open, conflicted) — same goal, but with complete client-site coverage, a retries/pooling escape hatch, and
  a fix for a `botocore` merge-precedence data-loss bug those minimal patches would have shared.

  ## Description

  AWS Bedrock clients were created with no client configuration, so large memory-extraction calls hit botocore's 60-second read-timeout default and failed
  with `Read timeout on endpoint URL: ".../converse"`. There was previously no way to raise the timeout (or tune retries/pooling) short of monkeypatching
  the provider.

  This adds three **opt-in** config keys to both the `aws_bedrock` LLM and embedder configs:

  | Key | Type | Purpose |
  |---|---|---|
  | `read_timeout` | `float` (seconds) | Convenience scalar; overrides botocore's 60s default |
  | `connect_timeout` | `float` (seconds) | Convenience scalar |
  | `boto_client_config` | `botocore.config.Config` **or** `dict` | Escape hatch for retries, connection pooling, or any botocore `Config` kwarg |

  ```python
  config = {
      "llm": {
          "provider": "aws_bedrock",
          "config": {
              "model": "us.anthropic.claude-sonnet-4-20250514-v1:0",
              "read_timeout": 1000,
              "connect_timeout": 60,
              "boto_client_config": {"retries": {"mode": "adaptive", "max_attempts": 5}},
          },
      }
  }

  Design

  - A single pure helper, mem0/utils/aws.py::build_boto_client_config(), turns the three inputs into a botocore.config.Config or None. It mirrors the
  existing mem0/utils/http.py::build_http_client precedent (pure, lazy-imports the heavy dep, returns X-or-None for opt-in behavior).
  - Wired into all three init/hot-path boto3 client-creation sites: the LLM runtime client (_initialize_aws_client), the LLM connection-test control-plane
  client (_test_connection), and the embedder client.
  - Strictly opt-in / backwards compatible: when none of the three keys is set, the helper returns None and the client is created with no config= —
  byte-identical to today's behavior.
  - Precedence: when both a scalar and boto_client_config set the same key, the scalar wins. Only set scalar keys participate in the botocore
  Config.merge(), so an unset scalar never clobbers a value the user set inside boto_client_config (botocore records explicit None and merge() is
  right-operand-wins — naive merging would clobber; this is covered by a regression test).

  Out of scope (documented in the integration docs)

  - Adaptive throttling / ThrottlingException retries are reachable via boto_client_config["retries"] (botocore's native retry config); no bespoke
  throttling logic is added.
  - Expired-STS-token-mid-call is not handled by these settings.

  Type of Change

  Breaking Changes

  N/A — purely additive; defaults preserve current behavior exactly.

  Test Coverage

  New tests (all mock boto3.client; no live AWS), 74 passing in the focused suite:
  - tests/utils/test_aws.py — 15 helper cases: opt-in None, scalars-only, dict/Config base handling, the None-clobber regression, scalar precedence, scalar
  bool/numeric validation, malformed-config wrapping.
  - tests/llms/test_aws_bedrock.py — config field storage; both LLM client sites receive the config; backwards-compat (no config= when unset); dict-config
  flows through __init__; a real Config survives Memory.from_config; scalar-overrides-boto_client_config end-to-end.
  - tests/embeddings/test_aws_bedrock_embeddings.py — config field storage; factory BaseEmbedderConfig(**config) accepts the new keys without TypeError;
  embedder client receives the config; backwards-compat. Pre-existing session-token / Titan-dims tests remain green.

  ruff check clean on all changed files.

  Checklist

  - [x] I have updated documentation (AWS Bedrock LLM + embedder integration pages)


kartik-mem0:
## Review — Changes Requested

Thanks for a thorough PR — the design is sound, the opt-in invariant is correct, and the `Config.merge()` regression test for the None-clobber case is exactly what this needed. One method was missed before the finish line.

### Strengths

- `build_boto_client_config` correctly handles botocore's `Config.merge()` right-wins semantics: only *set* scalars participate in the merge, so an unset `read_timeout` never clobbers a value the user set inside `boto_client_config`. The dedicated regression test `test_config_only_timeout_not_clobbered` covers exactly this subtle case.
- Covers both `bedrock-runtime` and `bedrock` (control-plane) client sites — `_initialize_aws_client` and `_test_connection` — not just the most obvious one.
- Strictly opt-in: all three params default to `None`; when unset, `build_boto_client_config` returns `None` and the `boto3.client()` call is byte-identical to today's behavior.
- `TYPE_CHECKING` guard on the annotation in `utils/aws.py` follows the `build_http_client` precedent correctly.
- The PR description is one of the clearest I've seen for a botocore config change — the precedence/merge semantics explanation and the clobber note are genuinely helpful.

### Issues

#### Important — please fix before merge

**`mem0/llms/aws_bedrock.py` — `list_available_models()` was not updated**

The method creates its control-plane client with the old pattern (no `config=` kwarg):

```python
bedrock_client = boto3.client("bedrock", **self.config.get_aws_config())
```

A user who sets `read_timeout=1000` to avoid the 60-second default would still hit the botocore default when calling `list_available_models()`. The fix is the same pattern as `_test_connection` already in this PR — build `client_kwargs`, conditionally inject `config=`, then spread. Please add a test mirroring `test_test_connection_client_uses_same_config` for this path too.

#### Minor — optional

1. **`AWSBedrockConfig` docstring**: the three new params (`read_timeout`, `connect_timeout`, `boto_client_config`) aren't mentioned in the `Args:` block. The integration docs were updated, but the in-code docstring wasn't.

2. **Early-return in `build_boto_client_config`**: the botocore imports run on every call, including the all-None → `return None` path. No correctness impact (every caller already depends on botocore), but an early-return guard before the imports would make the intent explicit.

### Evidence

- **CI**: All real checks green (Python SDK / `build_mem0` on 3.10, 3.11, 3.12 ✓). Vercel/preview skipped.
- **Tests reviewed**: `tests/utils/test_aws.py` (15 cases), `tests/llms/test_aws_bedrock.py` (`TestBotoClientConfig`, 8 cases), `tests/embeddings/test_aws_bedrock_embeddings.py` (4 new cases). Good structure and correctness coverage.
- **Credentials**: verified no changes to credential resolution — `get_aws_config()` is unchanged and the PR only adds a `config=` kwarg alongside the existing credentials dict.
- **Scope**: strictly client-config (timeouts/retries). No unrelated changes.
- **Issue linked**: `Closes #3825` confirmed in PR description.

Ready to merge with one fix: please add `list_available_models()` to the update (same pattern as `_test_connection`) and a test for it. Happy to re-review once you push.

Malhis:
@kartik-mem0 I've addressed the review comments.

kartik-mem0:
## Re-review — LGTM

Good fix — `list_available_models()` now uses the same `build_boto_client_config` wiring as `_initialize_aws_client` and `_test_connection`, and the two new tests (`test_list_available_models_uses_same_config` / `test_list_available_models_no_config_when_unset`) close the coverage gap cleanly. The prior important finding is fully resolved.

**`mem0/llms/aws_bedrock.py` — `list_available_models()` (line 666–681)**
Updated to the same `build_boto_client_config` → `client_kwargs` → conditional `config=` → spread pattern as `_test_connection`. All three boto3 client call sites are now consistent.

**`tests/llms/test_aws_bedrock.py` — `TestBotoClientConfig`**
Two new test cases for `list_available_models()`:
- `test_list_available_models_uses_same_config` — resets the mock after construction, calls the method directly, and asserts `cfg.read_timeout == 300`. This is the regression guard: it fails with `KeyError: 'config'` on the old bare `boto3.client("bedrock", **self.config.get_aws_config())` pattern.
- `test_list_available_models_no_config_when_unset` — proves the no-config path omits the `config=` kwarg, preserving the opt-in invariant.

**`AWSBedrockConfig` docstring** — the three new params (`read_timeout`, `connect_timeout`, `boto_client_config`) are now documented in the `Args:` block.

**`build_boto_client_config` fast path** — the all-None early return now precedes the `from botocore.config import Config` import, so the botocore import is genuinely lazy.

### Evidence

- **CI**: Python SDK `build_mem0` on 3.10, 3.11, 3.12 ✓. CI Gate ✓. Vercel/preview — fork auth, advisory, ignored.
- **Targeted tests**: 76 tests across `tests/utils/test_aws.py`, `tests/llms/test_aws_bedrock.py`, `tests/embeddings/test_aws_bedrock_embeddings.py` — all passed (Python 3.12, hatch `dev_py_3_12`).
- **Red-green proof for `list_available_models()`**:
  - PR state → `test_list_available_models_uses_same_config` PASS, `test_list_available_models_no_config_when_unset` PASS.
  - Reverted `list_available_models()` to old bare pattern → `test_list_available_models_uses_same_config` FAIL (`KeyError: 'config'`), `test_list_available_models_no_config_when_unset` PASS (correct — opt-in invariant holds both ways).
  - Restored to PR state → both PASS again.
- **Lint**: `ruff check` on all five changed `.py` files — clean.
- **Credentials**: `get_aws_config()` untouched; PR only adds a `config=` kwarg alongside the existing credentials spread — no credential path changes.

Ready to squash-merge.

Malhis:
@kartik-mem0 looking for a human reviewer for this PR. Thank you.

PR #5836, opened 2026-06-25
feat(opensearch): add configurable client timeout and honor pool_maxsize

### LINKED ISSUE #5835: OpenSearch vector store: no way to configure client request timeout (breaks Serverless cold-start)
### Component

### Use Case

  `OpenSearchDB` constructs its `opensearch-py` client with no request timeout, so it falls
  back to the library default of **10 seconds**. That is too low for **AWS OpenSearch
  Serverless**, whose collection cold-start can take 10–30s — the first operations after an
  idle period fail with `ConnectionTimeout`. There is currently no public config to raise it;
  the only workaround is monkeypatching the client's transport internals, which is fragile
  across `opensearch-py` upgrades.

  (Related to the Serverless-compatibility theme of #3739, which added `auto_refresh`.)

  While in the same constructor, there is also a latent bug: `OpenSearchConfig.pool_maxsize`
  already exists as a config field, but `OpenSearchDB.__init__` hard-codes `pool_maxsize=20`
  and ignores the configured value — so setting it today silently does nothing.

  ### Proposed Solution

  Add an opt-in `timeout` field to `OpenSearchConfig` and pass it to the client, and pass the
  existing `config.pool_maxsize` through instead of the hard-coded literal:

  ```python
  config = {
      "vector_store": {
          "provider": "opensearch",
          "config": {
              "host": "xxxxx.us-west-2.aoss.amazonaws.com",
              "port": 443,
              "http_auth": auth,
              "use_ssl": True,
              "verify_certs": True,
              "timeout": 60,        # NEW: raise for Serverless cold-start
              "collection_name": "my-collection",
          },
      }
  }
```

  timeout should be opt-in: when unset, omit the kwarg entirely so behavior is
  byte-identical to today (opensearch-py applies its own 10s default only when the argument
  is absent — passing timeout=None explicitly instead disables the timeout, so the kwarg
  must be conditional). pool_maxsize keeps its default of 20, so there's no behavior
  change for users who never set it.

  Alternatives Considered

  - A higher default timeout for everyone — rejected; it would change behavior for all
  existing OpenSearch users. Opt-in preserves backwards compatibility.

### PR DESCRIPTION
**Pull request by Malhis**
## Linked Issue

Closes #5835

(Related: #3739 added `auto_refresh` for the same OpenSearch Serverless-compatibility theme.)

## Description

`OpenSearchDB` constructs its `opensearch-py` client with **no request timeout**, so it uses the library's default of **10 seconds**. That is too low for **AWS OpenSearch Serverless**, whose collection cold-start can take 10–30s, producing `ConnectionTimeout` on the first operations after an idle period. Today the only workaround is monkeypatching the client's transport internals.

This adds an opt-in `timeout` to `OpenSearchConfig` and wires it into the client:

```python
config = {
    "vector_store": {
        "provider": "opensearch",
        "config": {
            "host": "xxxxx.us-west-2.aoss.amazonaws.com",
            "port": 443,
            "http_auth": auth,
            "use_ssl": True,
            "verify_certs": True,
            "timeout": 60,          # NEW: raise for Serverless cold-start
            "collection_name": "my-collection",
        },
    }
}
```

It also fixes a **latent bug in the same constructor**: `OpenSearchConfig.pool_maxsize` already exists, but `OpenSearchDB.__init__` hard-coded `pool_maxsize=20` and ignored the config value. The client now receives `config.pool_maxsize`, so a user-set value finally applies.

### Design

- `timeout` is **opt-in**: when unset (`None`), the `timeout` kwarg is **omitted** from the `OpenSearch(...)` call, so behavior is byte-identical to prior releases (opensearch-py applies its own 10s default only when the kwarg is absent — verified against opensearch-py 3.2.0; passing `timeout=None` explicitly would instead disable the timeout, which is why the kwarg is conditional).
- `pool_maxsize` default stays `20`, so there is no behavior change for users who never set it.
- Mirrors the existing `auto_refresh` field/test pattern (#3739).

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue) — `pool_maxsize` was silently ignored
- [x] New feature (non-breaking change that adds functionality) — configurable `timeout`

## Breaking Changes

N/A. Purely additive and opt-in. The only behavioral change is the intended `pool_maxsize` fix: a user who *already* set `pool_maxsize` (expecting it to work) now gets it applied — strictly more correct and matching the documented field.

## Test Coverage

Added to `tests/vector_stores/test_opensearch.py` (mock the `OpenSearch` constructor, assert on its captured kwargs — no live cluster):
- `test_timeout_omitted_by_default` — no `timeout` kwarg when unset (backwards-compat guard; mutation-tested as load-bearing).
- `test_timeout_passed_when_set` — `timeout=60` reaches the client.
- `test_pool_maxsize_passed_through` — a custom `pool_maxsize` reaches the client (guards the bug fix).
- `test_pool_maxsize_defaults_to_20` — default unchanged.

Full `tests/vector_stores/test_opensearch.py` suite: **28 passed**. `ruff check` clean on all changed files.

## Checklist

- [x] I have updated documentation (OpenSearch vector-store config table)

### DIFF
--- a/docs/components/vectordbs/dbs/opensearch.mdx
+++ b/docs/components/vectordbs/dbs/opensearch.mdx
@@ -67,6 +67,8 @@ config = {
 | `embedding_model_dims` | int | 1536 | Dimension of embedding vectors |
 | `use_ssl` | bool | False | Enable SSL/TLS connection |
 | `verify_certs` | bool | False | Verify SSL certificates |
+| `pool_maxsize` | int | 20 | Maximum number of connections in the client pool |
+| `timeout` | int | None | Per-request client timeout in seconds. Defaults to opensearch-py's built-in default (10s) when unset. Raise it for AWS OpenSearch Serverless, whose cold-start can exceed 10s. |
 | `auto_refresh` | bool | False | Automatically refresh index after insert. OpenSearch refreshes every ~1 second by default, so this is rarely needed. |
 
 <Note>
--- a/mem0/configs/vector_stores/opensearch.py
+++ b/mem0/configs/vector_stores/opensearch.py
@@ -18,6 +18,12 @@ class OpenSearchConfig(BaseModel):
         "RequestsHttpConnection", description="Connection class for OpenSearch"
     )
     pool_maxsize: int = Field(20, description="Maximum number of connections in the pool")
+    timeout: Optional[int] = Field(
+        None,
+        description="Per-request timeout in seconds for the OpenSearch client. Defaults to "
+        "opensearch-py's built-in default (10s) when unset. Raise this for AWS OpenSearch "
+        "Serverless, whose cold-start can exceed 10s.",
+    )
     auto_refresh: bool = Field(
         False,
         description="Automatically refresh index after insert operations to make documents "
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -26,16 +26,20 @@ class OpenSearchDB(VectorStoreBase):
         config = OpenSearchConfig(**kwargs)
 
         # Initialize OpenSearch client
-        self.client = OpenSearch(
-            hosts=[{"host": config.host, "port": config.port or 9200}],
-            http_auth=config.http_auth
+        client_kwargs = {
+            "hosts": [{"host": config.host, "port": config.port or 9200}],
+            "http_auth": config.http_auth
             if config.http_auth
             else ((config.user, config.password) if (config.user and config.password) else None),
-            use_ssl=config.use_ssl,
-            verify_certs=config.verify_certs,
-            connection_class=RequestsHttpConnection,
-            pool_maxsize=20,
-        )
+            "use_ssl": config.use_ssl,
+            "verify_certs": config.verify_certs,
+            "connection_class": RequestsHttpConnection,
+            "pool_maxsize": config.pool_maxsize,
+        }
+        # Pass timeout only when set, so an unset value preserves opensearch-py's default.
+        if config.timeout is not None:
+            client_kwargs["timeout"] = config.timeout
+        self.client = OpenSearch(**client_kwargs)
 
         self.collection_name = config.collection_name
         self.embedding_model_dims = config.embedding_model_dims
--- a/tests/vector_stores/test_opensearch.py
+++ b/tests/vector_stores/test_opensearch.py
@@ -200,6 +200,48 @@ class TestOpenSearchDB(unittest.TestCase):
         self.assertEqual(self.client_mock.index.call_count, 3)
         self.assertEqual(self.client_mock.indices.refresh.call_count, 1)
 
+    def test_timeout_omitted_by_default(self):
+        """When timeout is unset, no timeout kwarg is passed to the OpenSearch client.
+
+        Guarantees byte-identical behavior with prior releases: opensearch-py
+        applies its own default request timeout (10s) only when the kwarg is absent.
+        """
+        # self.os_db (built in setUp without a timeout) is the default case;
+        # self.mock_os captured that construction call.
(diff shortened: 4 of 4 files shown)


kartik-mem0:

### What this PR does

Adds an opt-in `timeout` field to `OpenSearchConfig` so the opensearch-py client's request timeout is configurable — solving `ConnectionTimeout` failures on AWS OpenSearch Serverless cold-start (10–30s vs. the library's 10s default). Also fixes a latent bug: `pool_maxsize` was hardcoded to `20` in the constructor and silently ignored the config value.

### Strengths

- `mem0/vector_stores/opensearch.py` lines 29–45 — the conditional kwarg guard (`if config.timeout is not None`) is architecturally correct. opensearch-py's `Connection.__init__` treats `timeout=None` as "no timeout" (disables it), not "use the default" — omitting the kwarg is the only way to preserve the library's built-in 10s default for users who don't set it. Traced against opensearch-py source and confirmed.
- `mem0/configs/vector_stores/opensearch.py` — `Optional[int] = None` with `Field(None, ...)` matches the existing `auto_refresh` field pattern exactly. The `validate_extra_fields` validator would have hard-rejected any prior attempt to pass `timeout=...`, so there is no silent pre-existing drift.
- `pool_maxsize` fix — a clean behavioral correction to a genuinely ignored config field; default stays 20 so no behavior change for existing users.
- Four regression tests, all load-bearing: `test_timeout_passed_when_set` and `test_pool_maxsize_passed_through` fail without the fix (validated locally); `test_timeout_omitted_by_default` and `test_pool_maxsize_defaults_to_20` correctly pass on both main and the PR (they guard backward compatibility).
- Docs updated in the same commit, including a useful note about the Serverless cold-start motivation.

### Issues

Minor (optional):

1. `tests/vector_stores/test_opensearch.py` line 451 — `test_init_with_http_auth` does not assert that `"timeout"` is absent from the constructor kwargs in the AWS SigV4 auth path. Not a CI failure (the test passes), but adding `self.assertNotIn("timeout", call_kwargs)` would make it explicit. Optional follow-up.

### Evidence

- **CI:** Python SDK build on 3.10, 3.11, 3.12 — all pass. CI Gate — pass. docs llms.txt check — pass. (Vercel advisory check ignored — fails on every fork PR due to deploy perms.)
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_opensearch.py -v --tb=short` → **28 passed in 35.35s**
- **Red-green proof:** `test_timeout_passed_when_set` fails on main with `ValidationError: Extra fields not allowed: timeout`; `test_pool_maxsize_passed_through` fails on main with `AssertionError: 20 != 42`. Both pass with the fix restored. ✓
- **Lint:** `ruff check` on changed `.py` files → clean.
- **Linked issue:** #5835 — PR addresses the reported root cause exactly.
- **Cross-package check:** OpenSearch vector store is Python-only; no TypeScript counterpart exists.
- **Could NOT verify:** live AWS OpenSearch Serverless behavior — verified by code tracing only.

### Assessment

The conditional kwarg design is correct, the pool_maxsize bug fix is clean, backward compatibility is preserved, and the regression tests bite. Also, could you link issue #5835 explicitly in the PR body with `Closes #5835` if it isn't already there? (It looks like it is — just confirming for our tracking.)

Malhis:
- Closes #5835 is already the first line of the PR body, so the issue is linked for tracking.
- On the SigV4-path timeout assertion: test_init_with_http_auth uses assert_called_once_with(...) with the complete kwarg set, which already fails if any unexpected kwarg - including timeout - is passed in the auth path. So that path's timeout-absence is covered, just implicitly.

Malhis:
@kartik-mem0 looking for human reviewer for this PR when you get the chance. Thank you.

PR #5837, opened 2026-06-25
docs(rerankers): align reranker config docs with v3 SDK

kartik-mem0:
## Description
Part of a docs stale-data accuracy pass. Aligns the **reranker** component docs with the shipping v3 SDK source. Each change was verified against the reranker source in `mem0/reranker/` and `mem0/memory/main.py`.

## Type of Change
- [x] Documentation update

## What changed
- `"rerank"` → `"reranker"` config key (cohere, sentence_transformer, zero_entropy)
- reranker `provider: "llm"` → `"llm_reranker"` (registered key in `RerankerFactory`)
- top-level `user_id=` in `search()` → `filters={"user_id": ...}` (v3 rejects top-level entity kwargs via `_reject_top_level_entity_params`)
- `top_n` → `top_k` in reranker configs
- removed config keys absent from the config classes (`trust_remote_code`, `model_kwargs`, `use_auth_token`, `local_files_only`, `use_fp16`, `batch_ranking`, `timeout`, `fp16`, `cache_folder`, `num_workers`)
- `memory.search_async(...)` → `AsyncMemory.search(..., filters=...)` (Memory has no `search_async`; AsyncMemory.search exists)
- flagged deprecated `scoring_prompt` with a `<Warning>` callout in custom-prompts.mdx
- corrected the LLM Reranker config table in config.mdx (`llm` dict optional; `provider` default "openai"; `model` default "gpt-4o-mini"; `top_k` not `top_n`)
- replaced non-existent `trust_remote_code` row in huggingface.mdx config table with the real `normalize` field (bool, default True)

## Test Coverage
- [x] No tests needed (documentation-only; verified against SDK source)

## Checklist
- [x] Docs verified against v3 SDK source in `mem0/reranker/`, `mem0/configs/rerankers/`, and `mem0/memory/main.py`


docs/components/rerankers/models/cohere.mdx
```diff
@@ -74,7 +74,7 @@ os.environ["COHERE_API_KEY"] = "your-api-key"
 config = {
     "vector_store": {"provider": "chroma"},
     "llm": {"provider": "openai", "config": {"model": "gpt-4o-mini"}},
-    "rerank": {
+    "reranker": {
```
whysosaket, line 77: is it reranker???? 

https://github.com/mem0ai/mem0/blob/0fbbb2f525dbaa30bd6ba13b3f7aad5e5b5b846a/mem0/client/types.py#L50

Can you please check if we have this wrong somewhere else too and what is the right behavior???

whysosaket: if we changed rerank -> reranker please also mention which PR was that

kartik-mem0: There are two distinct parameters here, which is the source of the confusion:

- **`reranker`** is the OSS *config-block key* — `Memory.from_config({"reranker": {...}})`. Declared at `mem0/configs/base.py:46` (`reranker: Optional[RerankerConfig]`) and consumed at `mem0/memory/main.py:463` (`if config.reranker:`).
- **`rerank`** (the `client/types.py:50` field you linked) is a *separate* search-time boolean that toggles reranking per query: `m.search(query, ..., rerank=True)`, checked at `mem0/memory/main.py:1447` (`if rerank and self.reranker:`).

So `"reranker": {...}` for the config block and `rerank=True` for search are both correct — they aren't the same field.

On "which PR changed `rerank` → `reranker`": none. There was no behavior change. The config key has always been `reranker` in the source (introduced that way in commit `61e86685`, "Added metadata filtering and reranker for the OSS"); the docs were simply wrong and used `"rerank": {...}`. This PR only fixes the docs.

I also grepped the whole `docs/` tree for stray `"rerank": {` config blocks — none remain. The 3 hits in `openapi.json` are JSON-Schema definitions for the search boolean (`"type": "boolean"`), which are correct.

kartik-mem0: Good instinct — there *was* a real bug here, just not where it first looked.

On the naming: `reranker` (the OSS config block) and `rerank` (the search-time boolean) are intentionally different, and neither was renamed. `reranker: {...}` configures *which* reranker to build (`configs/base.py:46` → `self.reranker`); `rerank=True` *activates* it per query (`memory/main.py:1447`: `if rerank and self.reranker`). The hosted client only exposes the `rerank` boolean (`client/types.py:50`) because the server runs the reranker. So both keys are correct.

The actual bug: the examples set up the `reranker` block but called `memory.search(...)` **without** `rerank=True`. Since `search()` defaults `rerank=False`, reranking never ran — so `cohere` / `zero_entropy` / `llm` / `sentence_transformer`, which then print `result['rerank_score']`, would `KeyError` on copy-paste, and the `optimization` / `custom-prompts` examples were silently measuring un-reranked search. Fixed by adding `rerank=True` to every search meant to exercise the reranker (0842310e).

On "is it wrong anywhere else": I swept the docs. The rest of the reranker component pages are correct. One more instance lives in `open-source/features/reranker-search.mdx` (the "After: reranking enabled via config" and "Basic reranked search" snippets) — that one's already on `main` (from #5847), outside this PR, so I'm flagging it for a separate fix rather than pulling an unrelated file into this one.

PR #5841, opened 2026-06-25
docs(cookbooks): fix v3 filters, response shapes & dead snippet in cookbooks

kartik-mem0:
## Description

Stale-data pass over `docs/cookbooks/frameworks/`, `docs/cookbooks/integrations/`, and `docs/cookbooks/operations/`. All changes verified against source code before editing.

## Type of Change

- [x] Documentation update

## What changed

- **frameworks/llamaindex-multiagent.mdx:213-216** — `memory.search(user_id=..., query=...)` → `memory.search(query=..., filters={"user_id": ...})` (v3 rejects top-level entity kwargs in search)
- **integrations/healthcare-google-adk.mdx:53** — renamed `mem0 = MemoryClient()` → `mem0_client = MemoryClient()` to match all subsequent call sites (`mem0_client.add()`, `mem0_client.search()`)
- **integrations/healthcare-google-adk.mdx:77-82** — `mem0_client.search(..., user_id=USER_ID, ...)` → `filters={"user_id": USER_ID}` (preserving `top_k` and `threshold`)
- **integrations/openai-tool-calls.mdx:56-84** — stale v2 `add()` output example (`[{"data": {"memory": "..."}}]`) → v3 shape `{"results": [{"memory": "...", "event": "ADD"}]}`
- **integrations/openai-tool-calls.mdx:91,197** — TS `mem0Client.search(input, { userId: USER_ID })` → `{ filters: { user_id: USER_ID } }`
- **integrations/tavily-search.mdx:7** — removed `<Snippet file="security-compliance.mdx" />` which references a non-existent file (confirmed: `docs/_snippets/` contains only `async-memory-add.mdx`, `blank-notif.mdx`, `get-help.mdx`)
- **operations/content-writing.mdx:218-222** — `memory.search(..., user_id=USER_ID, run_id=RUN_ID)` → `filters={"user_id": USER_ID, "run_id": RUN_ID}`
- **operations/email-automation.mdx:317-355** — moved `user_id=user_id` into `filters` dicts for both `search()` calls and the `get_all()` call; removed two misleading comments claiming OSS takes `user_id` as an explicit top-level parameter
- **operations/support-inbox.mdx:60** — `model="gpt-4"` → `model="gpt-4o"` (bare gpt-4 retired)

## Test Coverage

- [x] No tests needed (docs-only change)

## Checklist

- [x] All `add()` calls with top-level `user_id` were left untouched (v3 only rejects entity kwargs in `search()`/`get_all()`)
- [x] `git diff --stat` confirms only the three cookbook subdirs, 84 lines changed (< 200 limit)


docs/cookbooks/frameworks/llamaindex-multiagent.mdx
whysosaket, line 1: Have we updated the llamaindex integration??? If no then how are these supposed to work? What if the older SDK rejects the payload? Please take a deeper look into this.

docs/cookbooks/integrations/tavily-search.mdx
```diff
@@ -4,8 +4,6 @@ description: "Blend Tavily's realtime results with personal context stored in Me
 ---
 
 
-<Snippet file="security-compliance.mdx" />
```
whysosaket, line 7: Any reasons for removing this?

docs/cookbooks/operations/support-inbox.mdx
```diff
@@ -57,7 +57,7 @@ class CustomerSupportAIAgent:
         """
         # Start a streaming chat completion request to the AI
         stream = self.client.chat.completions.create(
-            model="gpt-4",
+            model="gpt-4o",
```
whysosaket: I feel we should be using a single model as a standard, let's do `gpt-5-mini` for now across all PRs please.

whysosaket (changes requested):

docs/cookbooks/frameworks/llamaindex-multiagent.mdx
kartik-mem0: The `memory` object here is `Mem0Memory.from_client(...)` — the LlamaIndex wrapper (`llama-index-memory-mem0`), which forwards `**kwargs` straight through to Mem0's `MemoryClient.search()`. No change to the integration package is needed.

The reason the call had to change: in the v3 SDK, `MemoryClient.search()` now *rejects* top-level entity params. From `mem0/client/main.py:309-313`:

```python
invalid_keys = ENTITY_PARAMS & set(kwargs.keys())   # {"user_id", "agent_id", "app_id", "run_id"}
if invalid_keys:
    raise ValueError(
        "Top-level entity parameters ... are not supported in search(). "
        "Use filters={'user_id': '...'} instead."
    )
```

So the *old* `memory.search(user_id=..., query=...)` is what would break (a hard `ValueError`); the PR's `memory.search(query=..., filters={"user_id": ...})` is the correct v3 form, and the wrapper passes `filters` through unchanged.

docs/cookbooks/integrations/tavily-search.mdx
kartik-mem0: The `security-compliance.mdx` snippet file no longer exists in the repo — it was last present around #3315 and has since been removed, and `tavily-search.mdx` is now the only file still referencing it (`grep -rl security-compliance docs/` returns just this one page). So `<Snippet file="security-compliance.mdx" />` is a dangling include that would fail the Mintlify build, which is why it was removed. If we want a security/compliance callout back on the integration pages, the snippet file needs to be (re)created first — happy to do that as a follow-up if you'd like it.

docs/cookbooks/operations/support-inbox.mdx
kartik-mem0: Done — `model="gpt-4o"` → `model="gpt-5-mini"` here. I also audited the other files this PR touches: `llamaindex-multiagent.mdx`, `openai-tool-calls.mdx`, and `content-writing.mdx` already use `gpt-5-mini`; the rest use provider-specific models (Gemini) or have no LLM model string. So `gpt-5-mini` is now the standard across this PR.

whysosaket (approved):

PR #5838, opened 2026-06-25
docs(components): fix LLM & embedder model IDs and TS support lists

kartik-mem0:
## Description

Accuracy pass over LLM and embedder documentation — all changes verified against source code defaults in `mem0/llms/`, `mem0/embeddings/`, `mem0-ts/src/oss/src/llms/`, and `mem0-ts/src/oss/src/embeddings/` before applying.

## Type of Change

- [x] Documentation update

## What changed

- `docs/components/llms/models/anthropic.mdx` L23, L48: model `claude-sonnet-4-20250514` → `claude-sonnet-4-6` (confirmed in `mem0/llms/anthropic.py` and `mem0-ts/src/oss/src/llms/anthropic.ts`)
- `docs/components/llms/models/aws_bedrock.mdx` L9: env var `AWS_ACCESS_KEY` → `AWS_ACCESS_KEY_ID` in setup prose (code block at L18 already correct)
- `docs/components/llms/models/groq.mdx` L24: Python model `mixtral-8x7b-32768` → `llama-3.3-70b-versatile` (confirmed in `mem0/llms/groq.py`); L49: TS model `mixtral-8x7b-32768` → `llama3-70b-8192` (confirmed in `mem0-ts/src/oss/src/llms/groq.ts`)
- `docs/components/llms/models/xAI.mdx` L23: model `grok-3-beta` → `grok-2-latest` (confirmed in `mem0/llms/xai.py`)
- `docs/components/llms/overview.mdx` L19: expand TS LLM Note from "OpenAI, Anthropic, and Groq" to full current list (confirmed by `ls mem0-ts/src/oss/src/llms/`: anthropic, azure, deepseek, google, groq, langchain, lmstudio, mistral, ollama, openai, openai_structured)
- `docs/components/llms/config.mdx` L101: `http_client_proxies` Provider `AzureOpenAI` → `All` (confirmed in `mem0/configs/llms/base.py` BaseLlmConfig); L113: `reasoning_effort` Provider `Sarvam` → `All` (same)
- `docs/components/embedders/overview.mdx` L13: replace "TypeScript only supports OpenAI" with actual list (confirmed by `ls mem0-ts/src/oss/src/embeddings/`: azure, google, langchain, lmstudio, ollama, openai)
[...]

PR #5843, opened 2026-06-25
docs(vectordbs): correct vector store config defaults & imports

kartik-mem0:
## Linked Issue

N/A — docs accuracy pass against SDK source.

## Description

Stale-data accuracy fixes for `docs/components/vectordbs/`. Each change was verified against the actual Python config class in `mem0/configs/vector_stores/` or the implementation file before applying. No new pages were added; only incorrect values were corrected.

## Type of Change

- [x] Documentation update

## What changed

- **baidu.mdx**: `table_name` default `mem0_table` → `mem0` (`BaiduDBConfig.table_name = "mem0"`)
- **langchain.mdx**: TypeScript example used `import { Memory } from "mem0ai"` (hosted client) instead of `mem0ai/oss` (OSS self-hosted). Also fixed undefined `LangchainVectorStore` variable — import alias removed and `MemoryVectorStore` used directly.
- **mongodb.mdx**: `collection_name` default `"mem0_collection"` → `"mem0"` (`MongoDBConfig`); `mongo_uri` default corrected from `mongodb://username:password@localhost:27017` to `mongodb://localhost:27017` (`MongoDBConfig`).
- **neon.mdx**: Python tab `hnsw` default `False` → `True` (`PGVectorConfig.hnsw = True`).
- **neptune_analytics.mdx**: pip extra `mem0ai[vector_stores]` → `mem0ai[vector-stores]` (matches `pyproject.toml` key).
- **pgvector.mdx**: `diskann` default `True` → `False`; `hnsw` default `False` → `True` (`PGVectorConfig`).
- **upstash-vector.mdx**: `enable_embeddings` was placed at the `vector_store` top level — moved inside `config: {}` where `UpstashVectorConfig` expects it.
- **valkey.mdx**: Removed non-existent `distance_metric` row (not a field in `ValkeyConfig`; distance is hardcoded). Also fixed pip extra `vector_stores` → `vector-stores`.
- **vertex_ai.mdx**: `region` marked optional — corrected to required (`GoogleMatchingEngineConfig.region` has no default). Updated inline comment to match.
- **weaviate.mdx**: Install `pip install weaviate weaviate-client` → `pip install weaviate-client` (`weaviate` is not a separate package; `pyproject.toml` only lists `weaviate-client`).
- **overview.mdx**: TypeScript supported vector DB list was stale — updated to reflect actual files in `mem0-ts/src/oss/src/vector_stores/` (adds PGVector, Supabase, LangChain, Azure AI Search; removes "in-memory" which is not a separate TS provider).

## Test Coverage

- [x] No tests needed (docs-only changes; all values verified against SDK source)

## Checklist

### DIFF
--- a/docs/components/vectordbs/dbs/faiss.mdx
+++ b/docs/components/vectordbs/dbs/faiss.mdx
@@ -55,6 +55,7 @@ Here are the parameters available for configuring FAISS:
 | `path` | Path to store FAISS index and metadata | `/tmp/faiss/<collection_name>` |
 | `distance_strategy` | Distance metric strategy to use (options: 'euclidean', 'inner_product', 'cosine') | `euclidean` |
 | `normalize_L2` | Whether to normalize L2 vectors (only applicable for euclidean distance) | `False` |
+| `embedding_model_dims` | Dimensions of the embedding model | `1536` |
 
 ### Performance Considerations
 
--- a/docs/components/vectordbs/dbs/baidu.mdx
+++ b/docs/components/vectordbs/dbs/baidu.mdx
@@ -46,7 +46,7 @@ Here are the parameters available for configuring Baidu VectorDB:
 | `account` | Baidu VectorDB account name | `root` |
 | `api_key` | API key for accessing Baidu VectorDB | Required |
 | `database_name` | Name of the database | `mem0` |
-| `table_name` | Name of the table | `mem0_table` |
+| `table_name` | Name of the table | `mem0` |
 | `embedding_model_dims` | Dimensions of the embedding model | `1536` |
 | `metric_type` | Distance metric for similarity search | `L2` |
 
--- a/docs/components/vectordbs/dbs/elasticsearch.mdx
+++ b/docs/components/vectordbs/dbs/elasticsearch.mdx
@@ -56,6 +56,8 @@ Here are the parameters available for configuring Elasticsearch:
 | `api_key`              | API key for authentication                         | `None`        |
 | `user`                 | Username for basic authentication                  | `None`        |
 | `password`             | Password for basic authentication                  | `None`        |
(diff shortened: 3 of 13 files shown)


docs/components/vectordbs/dbs/faiss.mdx
```diff
@@ -55,6 +55,7 @@ Here are the parameters available for configuring FAISS:
 | `path` | Path to store FAISS index and metadata | `/tmp/faiss/<collection_name>` |
 | `distance_strategy` | Distance metric strategy to use (options: 'euclidean', 'inner_product', 'cosine') | `euclidean` |
 | `normalize_L2` | Whether to normalize L2 vectors (only applicable for euclidean distance) | `False` |
+| `embedding_model_dims` | Dimensions of the embedding model | `1536` |
```
whysosaket, line 58: I remember it was 768 in the PR: https://github.com/mem0ai/mem0/pull/5838/changes#diff-aa11f3ede0cf1d7f6c72917bda9adbba1c4f298e96d0172a16f2ccf8b5c50989L70

Please make it as a single value across all providers and examples.

Image

Let's take help from @agumpandey and @SoumilRathi to understand what value makes most sense here.

soumil-rathi (approved):

docs/components/vectordbs/dbs/faiss.mdx
agumpandey: I don’t think we should force one single value. Changing the default without benchmarking could degrade retrieval quality. 

Let the users toggle each values they want, we should keep everything default for each model else the degrading embedding quality would by default show on mem0's performance

kartik-mem0: i do agree here with @agumpandey we should let user decide what dimensions they want to configure and each provider has there own default values.

do lmk @whysosaket

PR #5856, opened 2026-06-25
feat(embeddings): add TwelveLabs Marengo multimodal embedder

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## Description

This PR adds an opt-in `twelvelabs` embedding provider for multimodal memory.

- **Marengo** (`marengo3.0`) embeds text, image, audio and video into a single **512-dimensional** latent space, so a text memory describing a video shares a space with embeddings generated directly from the video itself.
- An optional `TwelveLabsEmbedding.summarize_video(...)` helper uses the **Pegasus** video-understanding model to turn a video into memory text you can store with `Memory.add(...)`.

It plugs into the existing embedder seam exactly like the other providers: registered in `EmbedderFactory`, allow-listed in `EmbedderConfig`, with a focused test, docs, and the dependency added under the `extras` group.

## Type of Change

## Breaking Changes

N/A — this is fully opt-in. The default embedder is unchanged; you only get TwelveLabs when you set `provider: "twelvelabs"`.

## Test Coverage

`tests/embeddings/test_twelvelabs_embeddings.py` covers the no-network path (mocked client, defaults, missing-key error, Pegasus wiring) and includes a live test gated on `TWELVELABS_API_KEY` that asserts a real 512-dim vector. Verified locally:

```
tests/embeddings/test_twelvelabs_embeddings.py::test_defaults_and_embed PASSED
tests/embeddings/test_twelvelabs_embeddings.py::test_missing_api_key_raises PASSED
tests/embeddings/test_twelvelabs_embeddings.py::test_summarize_video PASSED
tests/embeddings/test_twelvelabs_embeddings.py::test_embed_live PASSED
```

`ruff check` and `ruff format --check` pass on the changed files.

## Checklist

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/docs/components/embedders/models/twelvelabs.mdx
+++ b/docs/components/embedders/models/twelvelabs.mdx
@@ -1,0 +1,68 @@
+---
+title: TwelveLabs
+description: "Configure TwelveLabs Marengo as a multimodal embedding provider in Mem0, with optional Pegasus video summarization."
+---
+
+[TwelveLabs](https://twelvelabs.io) provides video-native foundation models. The
+**Marengo** model embeds text, image, audio and video into a single
+512-dimensional latent space, so a text memory describing a video shares the
+same space as embeddings generated directly from the video. The **Pegasus**
+model can summarize a video into text you can store as a memory.
+
+To use TwelveLabs, set the `TWELVELABS_API_KEY` environment variable. You can grab a free API key at [twelvelabs.io](https://twelvelabs.io) — there's a generous free tier.
+
+### Usage
+
+<Note> The `embedding_model_dims` parameter for `vector_store` should be set to `512` for the TwelveLabs embedder. </Note>
+
+```python
+import os
+from mem0 import Memory
+
+os.environ["TWELVELABS_API_KEY"] = "your_api_key"
+os.environ["OPENAI_API_KEY"] = "your_api_key"  # For the LLM
+
+config = {
+    "embedder": {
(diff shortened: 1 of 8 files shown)


kartik-mem0:
Thanks for this, @mohit-twelvelabs — and credit where it's due: this is a clean, well-structured contribution. It follows the provider pattern exactly (factory registration, config allow-list, `extras` dependency), it's fully opt-in and non-breaking, and it ships with tests and docs. The Marengo shared-latent-space angle and the optional Pegasus `summarize_video` helper are a genuinely nice fit for multimodal memory.

That said, we're being deliberately selective about taking new vendor-maintained embedding providers into core right now — each one we merge becomes something we're on the hook to keep working as the upstream API, models, and dimensions evolve, and we want to be honest about the maintenance commitment rather than merge and let it bit-rot. So we're going to pass on merging this into the main repo for now. This is a scope/maintenance call, not a reflection on the quality of the PR.

A great path that keeps your work fully usable: publish it as a standalone package (e.g. `mem0-twelvelabs`) against the public embedder seam — since it plugs into `EmbedderFactory` cleanly, users can adopt it without any changes to mem0. If TwelveLabs sees broad community pull as a mem0 embedder over time, we're very open to revisiting first-party support. Really appreciate the effort you put into matching our conventions.

PR #5860, opened 2026-06-25
fix(ts/redis): use nullish coalescing for hash/timestamps in insert/update

### LINKED ISSUE #5859: fix(ts/redis): use optional chaining for hash/created_at in insert/update
## Bug

`mem0-ts/src/oss/src/vector_stores/redis.ts:342-344` uses direct property access:

```typescript
hash: payload.hash,
memory: payload.data,
created_at: new Date(payload.created_at).getTime(),
```

Entity payloads from `_linkEntitiesForMemory` only contain `data`, `entity_type`, `linked_memory_ids`, and scope filters -- no `hash` or `created_at`. `new Date(undefined).getTime()` returns `NaN`, which gets stored in the Redis numeric field and breaks sorting.

## Fix

Use optional chaining with defaults, matching the Python fix in #5709:

```typescript
hash: payload.hash ?? "",
memory: payload.data ?? "",
created_at: payload.created_at ? new Date(payload.created_at).getTime() : 0,
```

Same pattern needed in `update()` as well.

## Related

- Python fix: #5709

HrushiYadav:
## Summary

- `insert()` and `update()` in `redis.ts` used direct property access (`payload.hash`, `payload.data`, `new Date(payload.created_at).getTime()`) which produces `undefined`/`NaN` on entity payloads that lack these keys
- Applied nullish coalescing (`?? ""`) for hash/data and conditional timestamps (0 when missing), matching the Python fix in #5709
- Added `redis.unit.test.ts` with 3 regression tests verifying entity payloads don't produce NaN

## Test plan

- [x] `npx jest src/oss/tests/redis.unit.test.ts` -- 3/3 pass
- [x] Entity payload insert produces `hash: ""`, `created_at: 0` (not undefined/NaN)
- [x] Entity payload update produces `hash: ""`, `created_at: 0`, `updated_at: 0`
- [x] Normal payload with timestamps preserves correct values

Closes #5859

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/redis.ts
+++ b/mem0-ts/src/oss/src/vector_stores/redis.ts
@@ -337,11 +337,12 @@ export class RedisDB implements VectorStore {
       const id = ids[idx];
 
       // Create entry with required fields
+      const createdAt = payload.created_at ? new Date(payload.created_at).getTime() : 0;
       const entry: Record<string, any> = {
         memory_id: id,
-        hash: payload.hash,
-        memory: payload.data,
-        created_at: new Date(payload.created_at).getTime(),
+        hash: payload.hash ?? "",
+        memory: payload.data ?? "",
+        created_at: createdAt,
         embedding: new Float32Array(vector).buffer,
       };
 
@@ -561,12 +562,14 @@ export class RedisDB implements VectorStore {
     payload: Record<string, any>,
   ): Promise<void> {
     const snakePayload = toSnakeCase(payload);
+    const createdAt = snakePayload.created_at ? new Date(snakePayload.created_at).getTime() : 0;
+    const updatedAt = snakePayload.updated_at ? new Date(snakePayload.updated_at).getTime() : 0;
     const entry: Record<string, any> = {
       memory_id: vectorId,
-      hash: snakePayload.hash,
-      memory: snakePayload.data,
-      created_at: new Date(snakePayload.created_at).getTime(),
-      updated_at: new Date(snakePayload.updated_at).getTime(),
+      hash: snakePayload.hash ?? "",
+      memory: snakePayload.data ?? "",
+      created_at: createdAt,
+      updated_at: updatedAt,
       embedding: Buffer.from(new Float32Array(vector).buffer),
     };
 
--- a/mem0-ts/src/oss/src/vector_stores/redis.ts
+++ b/mem0-ts/src/oss/src/vector_stores/redis.ts
@@ -337,7 +337,9 @@ export class RedisDB implements VectorStore {
       const id = ids[idx];
 
       // Create entry with required fields
-      const createdAt = payload.created_at ? new Date(payload.created_at).getTime() : 0;
+      const createdAt = payload.created_at
+        ? new Date(payload.created_at).getTime()
+        : 0;
       const entry: Record<string, any> = {
         memory_id: id,
         hash: payload.hash ?? "",
@@ -562,8 +564,12 @@ export class RedisDB implements VectorStore {
     payload: Record<string, any>,
   ): Promise<void> {
     const snakePayload = toSnakeCase(payload);
-    const createdAt = snakePayload.created_at ? new Date(snakePayload.created_at).getTime() : 0;
-    const updatedAt = snakePayload.updated_at ? new Date(snakePayload.updated_at).getTime() : 0;
+    const createdAt = snakePayload.created_at
+      ? new Date(snakePayload.created_at).getTime()
+      : 0;
+    const updatedAt = snakePayload.updated_at
+      ? new Date(snakePayload.updated_at).getTime()
+      : 0;
     const entry: Record<string, any> = {
       memory_id: vectorId,
       hash: snakePayload.hash ?? "",
--- a/mem0-ts/src/oss/tests/redis.unit.test.ts
+++ b/mem0-ts/src/oss/tests/redis.unit.test.ts
@@ -1,12 +1,3 @@
-/// <reference types="jest" />
-/**
- * Unit tests for Redis vector store insert/update with entity payloads.
- *
- * Entity payloads from _linkEntitiesForMemory lack hash, created_at,
- * and updated_at. Direct property access produces undefined/NaN values.
- * These tests verify the nullish coalescing guards.
- */
-
 import { RedisDB } from "../src/vector_stores/redis";
 
 jest.mock("redis", () => ({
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

Good, targeted fix — the logic is exactly right and it mirrors #5709's Python approach faithfully. Two things need to land before this can merge.

### Strengths

- `redis.ts:340,563-564` — the `?? ""` guard on `hash`/`data` and the ternary guard on timestamps match `mem0/vector_stores/redis.py` lines 127–133, 249–258 (the #5709 Python fix) exactly. Cross-SDK parity is clean.
- `redis.unit.test.ts` — test placement (`src/oss/tests/`), naming (`.unit.test.ts`), and jest mock approach all match the existing pattern. The three cases (entity insert, entity update, normal insert) are the right anchors for this bug.
- Fix is correctly scoped to `RedisDB.insert()`/`update()` — the only two methods that write numeric timestamp fields. `delete()`, `get()`, `list()` are unaffected.

### Issues

#### Important (must fix)

1. `redis.ts:340,563-564` — **Prettier print-width violation; CI is red on both Node 20 and Node 22.**
   The new ternary declarations exceed Prettier's print width:
   - `      const createdAt = payload.created_at ? new Date(payload.created_at).getTime() : 0;` — 89 chars
   - `    const createdAt = snakePayload.created_at ? new Date(snakePayload.created_at).getTime() : 0;` — 98 chars

   Fix: `cd mem0-ts && npx prettier --write src/oss/src/vector_stores/redis.ts`, then commit the result. Prettier will split to something like:
   ```ts
   const createdAt = snakePayload.created_at
     ? new Date(snakePayload.created_at).getTime()
     : 0;
   ```

2. `redis.unit.test.ts` — **no positive test for `update()` with a valid `updatedAt` string.** The insert path has test 3 confirming that a normal payload preserves the timestamp (`entry.created_at > 0`), but the update path only has the entity case (`entry.updated_at === 0`). A regression that hard-wires `updated_at: 0` unconditionally in `update()` would not be caught. Please add a fourth test — a normal payload with `updatedAt: "2026-06-25T10:00:00.000Z"` passed to `store.update()`, asserting `entry.updated_at > 0`.

#### Minor (optional)

3. `redis.unit.test.ts:1` — the `/// <reference types="jest" />` triple-slash directive is unused here; no other test file in `src/oss/tests/` includes it (jest types are already resolved globally). Safe to remove.

### Evidence

- **CI:** `TypeScript SDK / build_ts_sdk` fails on Node 20 and Node 22 (run `28166030477`). Root cause is Prettier, not the logic. Vercel/preview checks ignored (non-blocking fork perms).
- **Python cross-check:** `mem0/vector_stores/redis.py` lines 127–128, 131, 249–252 confirmed — same `0` sentinel, same `payload.get("hash", "")` default. The `0`-as-epoch design is intentional and shared across both SDKs.
- **Targeted tests:** skipped (Gate B skipped when CI is red per review pipeline). Logic verified by code tracing.
- **Red-green proof:** N/A — Gate B skipped. CI will be the test signal once Prettier is fixed.
- **Could NOT verify:** live Redis behavior (no live backend in review environment).

### Assessment

Ready to merge: **With fixes** — the fix logic is correct and the Python parity is exact. Two changes needed: (1) `prettier --write` on `redis.ts` to unblock CI, (2) add the positive `update()` timestamp test. Happy to re-review once you push.

HrushiYadav:
Addressed both items:

1. Ran `prettier --write` on `redis.ts` -- ternaries now split across lines, CI should pass
2. Added 4th test: `update with normal payload preserves timestamps` -- passes `updatedAt` and asserts `entry.updated_at > 0`
3. Removed the triple-slash `/// <reference types="jest" />` directive

All 4 tests pass locally.

kartik-mem0:
Verified against dec6ad7: both blockers addressed — clear to merge.

**Prettier fix** — CI fully green on Node 20 and 22 (`build_ts_sdk` ✓, `integration_ts_sdk` ✓, `CI Gate` ✓). The three ternaries are properly split across lines; no print-width violations.

**Update test** — "update with normal payload preserves timestamps" is in, passing `updatedAt: "2026-06-25T12:00:00.000Z"` and asserting `entry.updated_at > 0`. Guards exactly the regression I called out (someone accidentally hard-wiring 0 unconditionally in `update()`). The `/// <reference types="jest" />` directive removal is also in.

---

### What this PR does
Guards `RedisDB.insert()` and `update()` against entity payloads that lack `hash`/`created_at`/`updated_at`, replacing `new Date(undefined).getTime()` → NaN (which broke Redis NUMERIC sort fields) with `?? ""` / conditional ternary → `0`. Mirrors the Python fix in #5709 exactly.

### Strengths
- `redis.ts:337–342, 564–571` — both write paths covered, Prettier-clean, semantics match `mem0/vector_stores/redis.py` lines 127–128, 131, 249–258 exactly.
- `redis.unit.test.ts` — all four paths tested: entity insert (→ `hash: ""`, `created_at: 0`), entity update (→ `hash: ""`, `created_at: 0`, `updated_at: 0`), normal update (→ `updated_at > 0`), normal insert (→ `created_at > 0`). Placement, naming, and mock pattern match repo conventions.

### Issues
**Minor:** the `0`-as-epoch sentinel means entity records sort as the oldest entries in `list()` DESC order, and `get()` returns `"1970-01-01T00:00:00.000Z"` rather than the current-time fallback the pre-fix code incidentally produced via the `isNaN` guard. Python #5709 makes the same trade-off intentionally — cross-SDK consistency is the right call here. Worth a follow-up issue if the epoch semantics need revisiting in both SDKs together.

### Evidence
- **CI at dec6ad7:** `TypeScript SDK / build_ts_sdk (20)` ✓ `build_ts_sdk (22)` ✓ `integration_ts_sdk (20)` ✓ `integration_ts_sdk (22)` ✓ `CI Gate` ✓. Vercel/preview-deploy fail is fork-perms only — non-blocking.
- **Targeted tests:** `npx jest src/oss/tests/redis.unit.test.ts --no-coverage` → 4 passed, 4 total (0.291 s)
- **Red-green (entity tests 1–2):** old code `hash: payload.hash` → `undefined` (fails `toBe("")`), `new Date(undefined).getTime()` → NaN (fails `toBe(0)`). The fix restores `hash ?? ""` → `""` and conditional → `0`. Confirmed by code tracing; live Redis not available in review environment.
- **Red-green (test 3):** passes on both main and PR (expected — it's a forward-regression guard, not a bug-proof test; tests 1–2 carry the regression signal).
- **Python cross-check:** `mem0/vector_stores/redis.py` lines 127–128, 131, 249–258 — same `0` sentinel, same `get("hash", "")` default — exact parity. ✓
- **Could NOT verify:** live Redis sorted-set behavior (code tracing only).

### Assessment
Ready to merge: **Yes** (squash) · Est. human time: ~1 min

Both prior blockers resolved, CI fully green, cross-SDK parity clean. Thanks for the quick turnaround.

kartik-mem0 (approved):

PR #5697, opened 2026-06-20
fix(core): port delete_all hardening from async to sync Memory

rafid001:
## Linked Issue

Closes #5696

## Description

Ports two `delete_all()` hardening fixes from `AsyncMemory` (#5529) to sync `Memory`:

1. **Partial failure tolerance** — continue deleting remaining memories when one fails; log warnings instead of aborting mid-operation
2. **Bulk entity-store cleanup** — skip per-delete entity RMW during bulk delete; bulk-clear scoped entity records at the end to prevent orphaning and race conditions

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added:
- `test_sync_delete_all_continues_on_partial_failure`
- `TestSyncDeleteAllEntityRace.test_sync_delete_all_bulk_clears_entity_store`

Updated `test_delete_all` to assert `skip_entity_cleanup=True` is passed during bulk delete.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

Issue #5696, opened 2026-06-20
fix(oss): sync delete_all aborts on partial failure and orphans entity records


rafid001:
### Component

Core / Python SDK

### Description

`AsyncMemory.delete_all()` was hardened in #5529 with:
1. Partial-failure tolerance (`asyncio.gather(..., return_exceptions=True)`)
2. Bulk entity-store cleanup (`skip_entity_cleanup=True` during bulk delete, then `_bulk_clear_entity_store()`)

Sync `Memory.delete_all()` still has both gaps:

**Partial failure:** The sync loop calls `_delete_memory()` sequentially and raises on the first error, leaving a partial delete with inconsistent vector/history state. Async already continues and logs warnings.

**Entity orphaning:** Sync runs per-memory `_remove_memory_from_entity_store()` during bulk delete (N full entity scans). Async skips per-delete entity cleanup and bulk-clears scoped entity records at the end, avoiding read-modify-write races and orphaned entity records when entity boosting is enabled.

### Steps to Reproduce

**Partial failure:**
1. Configure OSS `Memory` with a vector store where `get()` fails for one of N memories during `delete_all()`.
2. Call `memory.delete_all(user_id="test-user")`.
3. Sync raises after deleting some memories; async completes and logs warnings.

**Entity orphan:**
1. Enable entity boosting, add memories that create entity records.
2. Call sync `memory.delete_all(user_id="alice")`.
3. Entity records can remain and pollute future entity-boosted searches.

### Expected Behavior

Sync `delete_all()` should match async behavior: continue on individual delete failures and bulk-clear the entity store after scoped bulk deletes.

### Additional Context

Regression tests for async exist in `tests/test_memory.py` (`test_async_delete_all_continues_on_partial_failure`, `TestAsyncDeleteAllEntityRace`). Sync parity tests are missing.

kartik-mem0:
## Triage Notes

**What we've established:**
- Sync `Memory.delete_all()` in `mem0/memory/main.py` has two gaps relative to async `AsyncMemory.delete_all()` (hardened in #5529):
  1. **Partial failure:** The sync loop raises on the first `_delete_memory()` error, leaving the store in a partially deleted state. Async continues and logs warnings.
  2. **Entity orphaning:** Sync calls `_remove_memory_from_entity_store()` per deletion (N read-modify-write scans). Async skips per-delete entity cleanup and bulk-clears at the end — avoiding orphaned records and race conditions.
[...]

PR #5870, opened 2026-06-25
fix(core): eliminate N redundant vector_store.get() calls in Memory.delete_all()

### LINKED ISSUE #5869: perf(core): Memory.delete_all() makes N redundant vector_store.get() calls
### Component

Core / Python SDK

### Summary

### Steps to Reproduce

    from mem0 import Memory

    m = Memory()

    for i in range(100):

    m.delete_all(user_id="u1")

    for memory in memories:

        if existing_memory is None:

### Expected Behavior

### Actual Behavior

### Environment

- mem0 version: 2.0.9

ShauryaaSharma:
## Linked Issue

Closes #5869 

## Description

`Memory.delete_all()` fetched all memories with `vector_store.list()` then called
`_delete_memory(memory.id)` for each — without passing the already-fetched object.
Because `existing_memory` defaults to `None`, every `_delete_memory` call re-fetched
the same object via `vector_store.get()`, making deletion O(N) reads instead of O(0).
The same bug existed in `AsyncMemory.delete_all()`.

The sync path had a second inefficiency: per-memory entity-store cleanup running O(N)
times, whereas the async path already used a single `_bulk_clear_entity_store` call.

**What this PR does:**

- Passes `existing_memory=memory` in both `Memory.delete_all()` and
  `AsyncMemory.delete_all()` so `vector_store.get()` is never called during bulk delete.
- Adds `_bulk_clear_entity_store()` to the sync `Memory` class (mirrors the existing
  async implementation) and calls it once at the end of `delete_all()`, reducing
  entity-store I/O from O(N) to O(1).
- Adds `skip_entity_cleanup` parameter to sync `_delete_memory` (already present on
  the async version) so single-memory deletes still clean up correctly by default —
  no behaviour change for any existing caller.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `TestDeleteAllPerformanceFix` (11 tests) in `tests/memory/test_main.py`:

- `test_sync_delete_all_does_not_call_vector_store_get` — `vector_store.get` is never called during `Memory.delete_all`
- `test_sync_delete_all_deletes_every_memory` — all N memories are deleted exactly once
- `test_sync_delete_all_uses_bulk_entity_cleanup_not_per_memory` — `_bulk_clear_entity_store` called once; `_remove_memory_from_entity_store` never called
- `test_sync_delete_all_skips_bulk_cleanup_when_entity_store_uninitialised` — no-op when entity store is None
- `test_delete_memory_skip_entity_cleanup_suppresses_per_memory_cleanup` — skip flag suppresses entity cleanup
- `test_delete_memory_default_still_runs_entity_cleanup` — default single-delete behaviour is unchanged
- `test_bulk_clear_entity_store_deletes_all_rows` — every entity row is deleted
- `test_bulk_clear_entity_store_noop_when_uninitialised` — no error when entity store is absent
- `test_bulk_clear_entity_store_scopes_to_filters` — only entity-id keys forwarded to `list()`
- `test_async_delete_all_does_not_call_vector_store_get` — async path: `vector_store.get` never called
- `test_async_delete_all_deletes_every_memory` — async path: all N deletions complete

All 55 tests in `tests/memory/test_main.py` pass. Ruff lint and format clean.

## Checklist


kartik-mem0:
Closes #5869 — nice, well-targeted fix. The core mechanism is correct: both `Memory.delete_all()` (line 1811) and `AsyncMemory.delete_all()` (line 3414) now pass `existing_memory=memory`, so `_delete_memory` skips the redundant `vector_store.get()` entirely. The new `_bulk_clear_entity_store` on the sync class mirrors the existing async one faithfully. I ran the full `TestDeleteAllPerformanceFix` suite (11/11 green), confirmed red-green on `test_sync_delete_all_does_not_call_vector_store_get` (reverted `main.py` → fails with a `TypeError` at the `vector_store.get()` call site, restoring PR code → passes), and all 55 pre-existing `tests/memory/test_main.py` tests still pass. Ruff is clean.

Three things need attention before this can land.

---

**[Changes Requested] Rebase and strip the reformatting**

`mem0/memory/main.py` — whole file

The PR is currently `CONFLICTING` against `main`. The root cause is the ~60-of-70 diff hunks that reformat unrelated code (frozenset brace expansion, single→double quote normalization, line-length compression on `raise ValueError(...)` calls). The actual functional change is six lines across five hunks. Bundling the reformatting into a `fix:` commit also makes future `git bisect` land here for any line touched by the style pass — not great for a file this size. Please rebase on current `main`, strip the cosmetic changes from this commit (or split them into a separate `chore: style` commit on top), and re-push. That should also resolve the merge conflict and let CI Gate run.

**Heads-up while you're rebasing:** PR #5697 is also open against `main` and adds `_bulk_clear_entity_store` + `skip_entity_cleanup` to the sync class (it's targeting a different issue, #5696, so the two PRs have different goals — but they touch the same lines). Both are currently `CONFLICTING`, so whichever merges first will require the other to rebase around it. Worth a quick coordination pass between the two.

---

**[Important] Sync `delete_all` needs fault isolation to match the async path**

`mem0/memory/main.py` — sync `delete_all`, around line 1814

```python
for memory in memories:
    self._delete_memory(memory.id, existing_memory=memory, skip_entity_cleanup=True)
    self._bulk_clear_entity_store(filters)
```

If any `_delete_memory` call raises (e.g. the vector store rejects one record), the loop aborts and `_bulk_clear_entity_store` never runs — entity rows remain orphaned for that scope.

The async path already handles this correctly:

```python
    await self._bulk_clear_entity_store(filters)   # unconditional
errors = [r for r in results if isinstance(r, BaseException)]
```

The fix for sync is the same pattern as the async path — wrap each `_delete_memory` call in a `try/except`, log per-failure warnings, and call `_bulk_clear_entity_store` unconditionally (or in a `finally` block) so entity cleanup is never skipped mid-batch. This asymmetry is worth closing here since `_bulk_clear_entity_store` is new for the sync class and the expectation it always runs is part of the contract.

---

**[Important] `tests/memory/conftest.py` stub should be platform-guarded**

`tests/memory/conftest.py` — new file, the `sys.modules.setdefault` block

The conftest pre-stubs `qdrant_client`, `google.protobuf`, and several gRPC packages unconditionally on every platform. The comment says "restricted Windows machines" but there is no `if sys.platform == "win32":` guard — so the stub also runs in CI (Linux) and on developer macOS machines. Any future test in `tests/memory/` that imports the real `qdrant_client` will silently receive a `MagicMock` instead.

Either add `if sys.platform == "win32":` around the injection block, or — given this is genuinely a Windows-specific DLL concern — consider moving it to a `conftest.py` that only activates via a pytest marker or a `pytest-configure` hook. The intent is sound; it just needs a scope guard so it doesn't affect non-Windows runs.

---

**Informational (no action needed on this PR)**

The TypeScript OSS `Memory.deleteAll()` in `mem0-ts/src/oss/src/memory/index.ts` (around lines 1620–1622) has the identical N+1 pattern: it calls `this.deleteMemory(memory.id)` in a loop, and `deleteMemory` does an unconditional `await this.vectorStore.get(memoryId)`. Worth filing as a follow-on issue once this lands so the TS fix can mirror what you've done here.

---

**Evidence**

| Check | Result |
|---|---|
| CI Gate | No required checks detected yet (PR is CONFLICTING — resolve conflicts to trigger CI) |
| Vercel deploy | Advisory fail (expected — fork perms) |
| `TestDeleteAllPerformanceFix` (11 tests, Python 3.12) | 11/11 PASS |
| Red-green `test_sync_delete_all_does_not_call_vector_store_get` | FAIL on main code → PASS on PR code |
| `tests/memory/test_main.py` full suite (55 tests) | 55/55 PASS |
| Ruff on changed files | Clean |
| Async `delete_all` fix (`existing_memory=memory`) | Verified in diff at line 3414 |
| `skip_entity_cleanup=False` default (single-delete backward compat) | Verified — `delete()` call site unchanged |
| TS counterpart | Same N+1 bug present, unfixed — track separately |

ShauryaaSharma:
Hello @kartik-mem0 ,
Thanks for the detailed review! All three points addressed in the latest push:

1. **Reformatting stripped** , reset `main.py` to `origin/main` and applied only the functional diff (~40 lines across 5 hunks). Rebased onto current main so the conflict is resolved.

2. **Sync fault isolation** , wrapped each `_delete_memory` call in `try/except`, collects per-failure warnings, and calls `_bulk_clear_entity_store` unconditionally after the loop, matching the async `gather(..., return_exceptions=True)` pattern.

3. **`conftest.py` platform guard** , added `if sys.platform == "win32":` around all stub injections so they're a no-op on Linux/macOS CI.

55/55 tests pass locally, ruff clean.

kartik-mem0:
## Changes Requested (re-review — head 4108ab8)

Thanks for the quick turnaround — the rebase is clean, the fault isolation is in, and the \`conftest.py\` guard is correctly scoped to \`sys.platform == \"win32\"\`. The core fix is solid. Two items need a small adjustment before this lands.

### Strengths
- \`existing_memory=memory\` now threaded through both sync and async \`delete_all\` correctly — confirmed by red-green on \`test_async_existing_memory_passed\` and \`test_no_vector_store_get_calls\`
- \`_bulk_clear_entity_store\` sync implementation is a faithful port of the async version — correct semantics for the \`delete_all\` context (deleting all entity rows in scope is equivalent to N read-modify-write calls when every memory is being removed)
- \`skip_entity_cleanup=False\` default preserves single-delete backward compatibility
- 10/10 \`TestDeleteAllPerformanceFix\` pass; 7 of those fail cleanly on reverted \`main.py\`

### Issues

#### Important (should fix)

**1. \`test_async_no_vector_store_get_calls\` is a false-proof test — \`tests/memory/test_main.py\` ~line 1138**

The test patches \`am._delete_memory\` entirely with a no-op, so \`vs.get\` is never reachable regardless of whether the async \`existing_memory=memory\` fix is present. The assertion \`vs.get.assert_not_called()\` passes trivially and proves nothing about the actual code path. \`test_async_existing_memory_passed\` already covers this correctly by intercepting the real \`_delete_memory\` call and asserting \`existing_memory=fake\` was forwarded. The false-proof test should be removed.

**2. Silent-success contract change needs a comment — \`mem0/memory/main.py\`, sync \`delete_all\`**

Before this PR, a failing \`_delete_memory\` propagated an exception and the caller could detect partial failure via \`try/except\`. After, errors are collected and logged, but the method always returns \`{\"message\": \"Memories deleted successfully!\"}\` — same as the async path, which already had this behavior via \`asyncio.gather(return_exceptions=True)\`. Making them consistent is correct, but the semantic change is undocumented. Please add a brief comment at the top of the \`try/except\` loop (e.g. *\"Best-effort: continue past individual failures; entity cleanup still runs. Mirrors async path behavior.\"*) so future callers know not to rely on exception propagation here.

#### Minor (optional)

- **TS follow-up:** \`mem0-ts/src/oss/src/memory/index.ts\` line 1622 has the identical N+1 pattern — \`deleteMemory\` calls \`vectorStore.get(memoryId)\` unconditionally with no \`existingMemory\` parameter. No action needed on this PR, but please open a tracking issue once this lands so the TS fix can mirror what you've done here (mentioned as informational in the prior review — following up to ensure it gets tracked).
- **\`_vector_store_list_rows()\`** (module-level helper, \`main.py\` line 84) does the same unpacking as the inline expression added to \`_bulk_clear_entity_store\`. Not a blocker, but worth a follow-up to reduce duplication.

### Evidence
| Check | Result |
|---|---|
| CI Gate / test matrix | No required checks detected at current head (CI Gate not yet triggered — may need a push) |
| Vercel deploy | Advisory fail — ignored (fork perms) |
| CLA | Pass |
| \`TestDeleteAllPerformanceFix\` (10 tests, Python 3.12) | 10/10 PASS |
| Red-green: \`test_no_vector_store_get_calls\` | FAIL on reverted \`main.py\` → PASS with fix |
| Red-green: \`test_async_existing_memory_passed\` | FAIL on reverted \`main.py\` → PASS with fix |
| Ruff on changed files | Clean |
| Sync bug on \`main\` (line 1855, no \`existing_memory\`) | Confirmed present |
| Async bug on \`main\` (line 3475, no \`existing_memory\`) | Confirmed present |
| Both fixed in this PR | Confirmed |
| TS counterpart (\`index.ts\` line 1622) | Same N+1 bug present — track separately |
| Could NOT verify | Live vector-store round-trip count (code-tracing only) |

### Assessment
Ready to merge: **With fixes** — please remove \`test_async_no_vector_store_get_calls\` and add the comment on the error-swallowing loop; I'll re-review once you push.

ShauryaaSharma:
Hey @kartik-mem0 ,
Thanks for the thorough re-review. All four items addressed in `aac819a8`:

1. **`AttributeError: _entity_store`** (CI failures on 3.10/3.11), changed `self._entity_store` to `getattr(self, "_entity_store", None)` in `_bulk_clear_entity_store`. Tests that construct `Memory` via `__new__` without running `__init__` no longer crash.

2. **Silent-success comment** , added the requested best-effort comment above the `try/except` loop documenting that callers should not rely on exception propagation.

3. **`test_async_no_vector_store_get_calls` removed** , agreed it was a false-proof test; `test_async_existing_memory_passed` covers the real contract.

4. **`test_async_delete_all_continues_on_partial_failure`** (CI failure in `tests/test_memory.py`) , the old test injected failure via `vector_store.get.side_effect`, but `get` is never called now that `existing_memory` is forwarded. Updated to inject via `vector_store.delete.side_effect` and assert `call_count == 3` (all three deletes attempted, one raises, batch continues).

kartik-mem0:

### What this PR does
This fixes the Python SDK `delete_all()` N+1 path by passing the already-listed memory object into `_delete_memory()` for both sync and async delete-all flows, so `vector_store.get()` is no longer called once per memory. It also brings sync entity cleanup in line with the async bulk-clear behavior.

### Strengths
- `mem0/memory/main.py:1877` now passes `existing_memory=memory` and `skip_entity_cleanup=True` through the sync loop, removing the redundant fetch while avoiding per-memory entity cleanup during bulk deletion.
- `mem0/memory/main.py:1874` documents the best-effort partial-failure behavior requested in the prior review, matching the existing async `gather(return_exceptions=True)` contract.
- `mem0/memory/main.py:3506` fixes the same redundant get path in `AsyncMemory.delete_all()`.
- `tests/memory/test_main.py:1128` keeps the real async assertion (`existing_memory` is forwarded) and the previous false-proof async test has been removed.
- `tests/test_memory.py:987` now injects the async partial failure at `vector_store.delete`, which is the code path that still runs after the fix.

### Issues
Minor (optional): the TypeScript OSS `deleteAll()` path has the same N+1 shape and should get a follow-up issue/PR; this PR is scoped correctly to the linked Python SDK issue.

### Evidence
- **CI:** `license/cla` passes. `Vercel` fails with deploy authorization and is advisory/ignored per policy. No real CI Gate / Python SDK matrix checks were reported for this head.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_main.py::TestDeleteAllPerformanceFix tests/test_memory.py::test_async_delete_all_continues_on_partial_failure -v --tb=short` → 10 passed in 6.60s; after restore, 10 passed in 0.45s.
- **Red-green proof:** reverting only `mem0/memory/main.py` to `origin/main` made the same targeted tests fail 8/10, including the sync no-`get()` regression, sync bulk-clear coverage, async `existing_memory` forwarding, and the updated async partial-failure test.
- **Lint:** `hatch -e dev_py_3_12 run ruff check mem0/memory/main.py tests/memory/test_main.py tests/memory/conftest.py tests/test_memory.py` → all checks passed.
- **Linked issue:** #5869 — PR addresses the reported Python SDK root cause in both sync and async paths.
- **Duplicate cluster:** read #5954 as the other open PR for #5869; it is not the better candidate because real Python SDK CI is red there and it does not include the sync fault-isolation behavior requested here.
- **Could NOT verify:** live vector-store round-trip counts against an external backend; verified by code tracing and regression tests.

### Assessment
The prior requested fixes are verified at `aac819a8f3c7f4b2104b471ffe800fe1f302881a`, and no confirmed Critical or Important findings remain.

kartik-mem0:
## Changes Requested (re-review — head `cf74f1be`)

Thanks for the follow-up. I re-checked the new head after the previous LGTM at `aac819a8`; the only new commit is `cf74f1be`, which deletes `tests/memory/conftest.py`. That removal is fine from this review's side — it actually reduces the temporary test-stub surface.

The blocker is CI: `CI Gate` is red because the real `Python SDK / changelog_check` job is failing. This looks like a stale-base/version-check issue rather than a code regression in your diff: the PR head still has `pyproject.toml` at `2.0.9`, while current `main` is already `2.0.10`, and the merge-ref check is seeing `2.0.9 -> 2.0.10` without a changelog entry.

Please rebase or merge current `main` into the branch and let CI rerun. I would not add a changelog entry here unless your branch itself changes `pyproject.toml`; this PR's API diff does not include `pyproject.toml` or `docs/changelog/sdk.mdx`.

### Evidence
- **CI:** `CI Gate` fails; `Python SDK / changelog_check` fails with `pyproject.toml version changed from 2.0.9 to 2.0.10 but docs/changelog/sdk.mdx was not updated`; `Python SDK / build_mem0` passes on 3.10, 3.11, and 3.12; `license/cla` passes.
- **Ignored advisory check:** `Vercel` fails with deploy authorization, ignored per policy.
- **New-head delta:** `aac819a8...cf74f1be` is one commit, `Delete tests/memory/conftest.py`, removing the guarded qdrant/grpc test stub.
- **Targeted tests / red-green:** skipped this pass because a real required CI check is red; prior LGTM evidence covered the regression tests at `aac819a8`, and the new delta does not change runtime code.
- **Linked issue:** #5869 — the runtime fix still addresses the reported Python SDK `delete_all()` N+1 root cause.
- **Could NOT verify:** local targeted tests on this pass; the single bounded worktree fetch attempt failed with `fatal: shallow file has changed since we read it`.

### Assessment
Ready to merge: **With fixes** — please rebase/merge current `main` and get `CI Gate` green; I'll re-review once it reruns.

ShauryaaSharma:
Hey @kartik-mem0,

Rebased onto current main, pyproject.toml now picks up 2.0.10 so the changelog_check should pass. conftest.py is also removed to match your cleanup commit. No other changes. CI should rerun now.

kartik-mem0:
Re-review, following up on the [prior pass](https://github.com/mem0ai/mem0/pull/5870#issuecomment-4831047020) (Changes Requested — `CI Gate` was red because `Python SDK / changelog_check` saw this branch's base at `pyproject.toml` `2.0.9` vs `main`'s `2.0.10`).

Verified against `9df8ddef` (current head): that blocker is gone. The branch is now rebased onto a base where `pyproject.toml` is `2.0.10`, matching current `main`, so there's no version mismatch left for `changelog_check` to flag. The only commits since the version I fully reviewed are the `tests/memory/conftest.py` stub removal and this rebase — `mem0/memory/main.py`, `tests/memory/test_main.py`, and `tests/test_memory.py` are unchanged in content, so the earlier findings on the fix itself still stand.

### What this PR does
Eliminates the N redundant `vector_store.get()` calls in `Memory.delete_all()` / `AsyncMemory.delete_all()` by threading the already-fetched `existing_memory` through `_delete_memory()`, adds `_bulk_clear_entity_store()` for O(1) entity cleanup on the sync path (mirroring the existing async one), and adds fault isolation so one failed delete doesn't abort the rest of the batch.

### Strengths
- `_delete_memory` now takes `existing_memory` + `skip_entity_cleanup`, so `delete_all` reuses the memory object it already has from `list()` instead of re-fetching it.
- Sync `delete_all` now mirrors the async path's best-effort semantics (equivalent to `asyncio.gather(return_exceptions=True)`), with a comment explaining why exceptions are collected rather than propagated.
- The new tests reproduce the actual bug, not just the mechanism: reverting `main.py` to its pre-fix state fails 8 of the 10 new/changed tests with real errors (`TypeError: fromisoformat` from the un-threaded `existing_memory`, `AttributeError: ... does not have the attribute '_bulk_clear_entity_store'`), so this is genuine red-green, not a mock asserting its own scaffolding.

### Issues
Minor (optional): none blocking.

### Evidence
- **CI:** `CI Gate` / `Python SDK` checks are `action_required` — blocked pending maintainer approval to run on this external contribution, not failing. Only `Vercel` is red (advisory/no fork deploy perms, ignored per policy) and `license/cla` passes.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_main.py::TestDeleteAllPerformanceFix tests/test_memory.py::test_async_delete_all_continues_on_partial_failure -v` → 10 passed.
- **Red-green proof:** reverted `mem0/memory/main.py` to `main`'s version → 8/10 fail with the errors above; restored the PR's version → 10/10 pass again.
- **Regression check:** full `tests/memory/test_main.py` → 54 passed. `ruff check` on all three changed files → clean.
- **Linked issue:** #5869 — addresses the reported root cause directly.
- **Could NOT verify:** the hosted `CI Gate` run itself finishing green, since it's blocked awaiting an approval outside this review's scope — the local run above covers the same ground (full suite + lint + red-green) at this exact commit.

### Assessment
Ready to merge: **Yes** (squash) · Est. human time: ~5 min (approve the pending CI run, confirm green, squash)
Same fix that was architecturally sound last pass — this one just confirms the CI blocker is gone and the code/tests haven't drifted since.

ShauryaaSharma:
Hi @kartik-mem0, @deshraj

this PR is rebased onto current main (https://github.com/mem0ai/mem0/commit/9df8ddeff58ea60848110a30efc69342f75c77d2) and green: TestDeleteAllPerformanceFix + the async partial-failure test pass, full tests/memory/test_main.py is 54/54, ruff clean, and red-green holds against reverted main.py.

The only thing left is the pending CI workflow, it's stuck at action_required and needs a maintainer to approve the run on the fork before CI Gate can report. Could someone kick that off when you have a moment?

Happy to fix anything it surfaces.

ShauryaaSharma:
Hi @kartik-mem0

The only thing left is the pending CI workflow, it's stuck at action_required and needs a maintainer to approve the run on the fork before CI Gate can report. Could someone kick that off when you have a moment?

Happy to fix anything it surfaces.

ShauryaaSharma:
Hi @kartik-mem0

Everything passed in the CI workflow, Just a quick reminder on this PR in case it slipped off. 

Happy to fix anything if required

ShauryaaSharma:
Hi @kartik-mem0

Everything passed in the CI workflow, Just a quick reminder on this PR in case it slipped off.

Happy to fix anything if required

PR #5872, opened 2026-06-25
fix(core): remove dead _safe_deepcopy_config call in Memory.__init__

ShauryaaSharma:
## Linked Issue

Closes #5871 

## Description

`Memory.__init__` called `_safe_deepcopy_config(self.config.vector_store.config)` inside
the telemetry block and assigned the result to `telemetry_config`. The very next
assignment overwrote that variable with a dict-based constructor call, making the
deepcopy completely dead. On every `Memory()` instantiation with telemetry enabled, a
full copy of the vector store config was allocated and thrown away immediately.

The block comment directly above the section reads *"Create telemetry config manually
to avoid deepcopy issues with thread locks"*, confirming the dict-based approach was a
deliberate replacement for deepcopy. The dead call was left behind by mistake.

`AsyncMemory.__init__` is unaffected, it deepcopies the config and mutates the copy
in place, so the deepcopy is load-bearing there and is left unchanged.

**Change:** remove the single dead `_safe_deepcopy_config()` call from the sync
`Memory.__init__` telemetry block. No logic is altered, the dict-based config
construction that follows was already doing the correct work.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added `TestDeadDeepcopyRemoved` (2 tests) in `tests/memory/test_main.py`:

- `test_sync_init_telemetry_does_not_call_safe_deepcopy` , asserts `_safe_deepcopy_config`
  is **not** called during `Memory.__init__` with telemetry enabled, proving the dead
  call is gone and cannot regress.
- `test_async_init_telemetry_still_calls_safe_deepcopy` , asserts `_safe_deepcopy_config`
  **is** still called during `AsyncMemory.__init__`, guarding the load-bearing async path
  against accidental removal.

All 46 tests in `tests/memory/test_main.py` pass. Ruff lint and format clean.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #5873, opened 2026-06-25
fix: remove unnecessary deepcopy in Memory.__init__ telemetry block

### LINKED ISSUE #5871: Memory.__init__ deepcopies vector store config and immediately discards the result on every instantiation
### Component

Core / Python SDK

### Summary

### Steps to Reproduce

    from mem0 import Memory

    m = Memory()

    if MEM0_TELEMETRY:
        telemetry_config_dict = {}
        ...

    if MEM0_TELEMETRY:
            telemetry_config.path = ...

### Expected Behavior

### Actual Behavior

### Environment

- mem0 version: 2.0.9

C1-BA-B1-F3:
## Summary

Remove dead `_safe_deepcopy_config()` call in `Memory.__init__` that was immediately overwritten by dict-based config construction.

## Problem

`Memory.__init__` calls `_safe_deepcopy_config(self.config.vector_store.config)` inside the telemetry block and assigns the result to `telemetry_config`. The very next assignment overwrites that variable entirely with a dict-based constructor call, making the deepcopy completely dead — it allocates a full copy of the config object on every `Memory()` instantiation and discards it immediately.

## Fix

Remove the unnecessary `_safe_deepcopy_config()` call on line 460 of `mem0/memory/main.py`. The dict-based approach that follows builds the config correctly without it.

## Testing

Added regression test `test_memory_init_does_not_call_safe_deepcopy_config_for_telemetry` that verifies `_safe_deepcopy_config` is not called during `Memory.__init__` when telemetry is enabled.

Fixes #5871

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -457,7 +457,6 @@ class Memory(MemoryBase):
             telemetry_config_dict['collection_name'] = "mem0migrations"
 
             # Set path for file-based vector stores
-            telemetry_config = _safe_deepcopy_config(self.config.vector_store.config)
             if self.config.vector_store.provider in ["faiss", "qdrant"]:
                 provider_path = f"migrations_{self.config.vector_store.provider}"
                 telemetry_config_dict['path'] = os.path.join(mem0_dir, provider_path)
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -1534,11 +1534,12 @@ async def test_async_procedural_memory_langchain_strips_code_blocks(mock_llm_fac
 @patch('mem0.utils.factory.EmbedderFactory.create')
 @patch('mem0.utils.factory.VectorStoreFactory.create')
 @patch('mem0.utils.factory.LlmFactory.create')
-@patch('mem0.memory.storage.SQLiteManager')
+@patch('mem0.memory.main.SQLiteManager')
 @patch('mem0.memory.main.MEM0_TELEMETRY', True)
 @patch('mem0.memory.main._safe_deepcopy_config')
+@patch('mem0.memory.main.os.makedirs')
 def test_memory_init_does_not_call_safe_deepcopy_config_for_telemetry(
-    mock_safe_deepcopy, mock_sqlite, mock_llm_factory, mock_vector_factory, mock_embedder_factory
+    mock_makedirs, mock_safe_deepcopy, mock_sqlite, mock_llm_factory, mock_vector_factory, mock_embedder_factory
 ):
     """Regression #5871: Memory.__init__ should not call _safe_deepcopy_config in the telemetry block.
 
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -1529,3 +1529,28 @@ async def test_async_procedural_memory_langchain_strips_code_blocks(mock_llm_fac
     insert_call = memory.vector_store.insert.call_args
     stored_data = insert_call[1]["payloads"][0]["data"]
     assert "```" not in stored_data
+
+
+@patch('mem0.utils.factory.EmbedderFactory.create')
+@patch('mem0.utils.factory.VectorStoreFactory.create')
+@patch('mem0.utils.factory.LlmFactory.create')
+@patch('mem0.memory.storage.SQLiteManager')
+@patch('mem0.memory.main.MEM0_TELEMETRY', True)
+@patch('mem0.memory.main._safe_deepcopy_config')
+def test_memory_init_does_not_call_safe_deepcopy_config_for_telemetry(
+    mock_safe_deepcopy, mock_sqlite, mock_llm_factory, mock_vector_factory, mock_embedder_factory
+):
+    """Regression #5871: Memory.__init__ should not call _safe_deepcopy_config in the telemetry block.
+
+    The dict-based approach builds the config directly, making the deepcopy call dead code.
+    """
+    mock_embedder_factory.return_value = MagicMock()
+    mock_vector_factory.return_value = MagicMock()
+    mock_llm_factory.return_value = MagicMock()
+    mock_sqlite.return_value = MagicMock()
+
+    config = MemoryConfig()
+    Memory(config)
+
+    # _safe_deepcopy_config should NOT be called during __init__ for telemetry config
+    mock_safe_deepcopy.assert_not_called()


kartik-mem0:
**Review — issue #5871**

**Verdict: LGTM** (with two improvement suggestions)

---

**Strengths**

- Removes exactly the right line (`main.py:488`) and nothing else. The dead allocation is gone; nothing between lines 488 and 495 reads from `telemetry_config` before it is overwritten at 495, so the removal is safe.
- MERGEABLE, no conflicts.
- Regression test is valid: patches `_safe_deepcopy_config` and asserts it is not called during `Memory.__init__` with telemetry enabled. Decorator/arg ordering is correct — `@patch('mem0.memory.main.MEM0_TELEMETRY', True)` does not inject a mock argument, so the five remaining mocks map cleanly to `mock_safe_deepcopy … mock_embedder_factory` bottom-up.
- Lint clean on both changed files (`ruff check` — all checks passed).
- `AsyncMemory.__init__` (line 2118) is correctly left untouched.

---

**Issues**

*Important — `SQLiteManager` patch target is wrong* (`tests/test_memory.py:1537`)

```python
@patch('mem0.memory.storage.SQLiteManager')   # ← patches the storage module's name
```

`main.py:28` does `from mem0.memory.storage import SQLiteManager`, which binds the name directly into `mem0.memory.main`. The call at `main.py:456` (`self.db = SQLiteManager(...)`) resolves through `main`'s namespace, not `storage`'s. The correct target is:

```python
@patch('mem0.memory.main.SQLiteManager')
```

The test still passes as-is (the core `assert_not_called()` assertion is about `_safe_deepcopy_config`, which is patched correctly), but `Memory.__init__` is touching a real `SQLiteManager` during the test rather than a mock. On most machines this silently succeeds; in a strict read-only sandbox it raises. One-line fix.

*Important — `os.makedirs` not mocked* (`tests/test_memory.py:1554-1555`)

With the default `MemoryConfig` (qdrant provider) and `MEM0_TELEMETRY=True`, execution reaches `main.py:492`:

```python
os.makedirs(telemetry_config_dict['path'], exist_ok=True)
```

This creates `~/.mem0/migrations_qdrant` on disk during the test run. Adding `@patch('mem0.memory.main.os.makedirs')` (or `@patch('os.makedirs')`) above the other decorators removes the side effect.

---

**Suggestions (optional)**

- A guard test for the async path would round out the suite: `AsyncMemory.__init__` at line 2118 uses a deepcopy that is load-bearing (it mutates the copy in-place). An `assert_called_once()` assertion on that path would prevent a future contributor from mistakenly applying the same "dead deepcopy" removal there.
- A short comment at the removed line's location noting the intentional sync/async divergence (sync builds fresh from dict; async mutates a copy) would help future readers.

---

**Evidence**

| | |
|---|---|
| CI required checks | No required checks beyond Vercel (advisory, fork perms — not merge-blocking) |
| Gate B red-green | PASS (fix) → FAIL (revert to main: "Expected `_safe_deepcopy_config` to not have been called. Called 1 times. `call(QdrantConfig(collection_name='mem0', …))`") → PASS (restore) |
| Lint | `ruff check mem0/memory/main.py` — all clear; `ruff check tests/test_memory.py` — all clear |
| `AsyncMemory` path verified | `main.py:2118–2128` — deepcopy result mutated in-place, not overwritten; unaffected by this PR |
| TS counterpart needed | No — `mem0-ts/src/oss/` has no `_safe_deepcopy_config` equivalent |

The two Important items are straightforward one-liners. If you'd like to push a quick follow-up commit addressing them, the PR will be ready to squash-merge immediately. Either way, the core fix and regression signal are solid.

kartik-mem0:
## Duplicate — recommend consolidating into #5872

*(Superseding the earlier LGTM comment — this competitor comparison was missing from the first pass.)*

### What both PRs do

Both remove exactly one dead line from `Memory.__init__` (`main.py:488`):

```python
telemetry_config = _safe_deepcopy_config(self.config.vector_store.config)
```

The deepcopy result is immediately overwritten at line 495 by the dict-based constructor and is never read in between. The fix in both PRs is correct and safe.

### Why #5872 is the better long-term fix

- **Async-path guard missing here.** `AsyncMemory.__init__` (line 2118) also calls `_safe_deepcopy_config`, but that call IS load-bearing — the result is mutated in-place at lines 2119 and 2122 before it reaches `VectorStoreFactory.create`. PR #5872 adds a second test (`test_async_init_telemetry_still_calls_safe_deepcopy`) that guards this path against a future contributor accidentally applying the same "dead deepcopy" reasoning to AsyncMemory. This PR has no such guard.
- **Correct test location.** #5872 adds tests to `tests/memory/test_main.py`, which is the per-module location for `mem0/memory/main.py` per the repo's naming conventions. This PR appends to the umbrella `tests/test_memory.py`.
- **Test pattern fit.** #5872 uses the `mocker` fixture (pytest-mock), consistent with the rest of `tests/memory/test_main.py`. Both PRs share the same minor `SQLiteManager` patch-target issue (`mem0.memory.storage` instead of `mem0.memory.main`) — it doesn't break the core assertion, but it's a wash between them.
- **Created first.** #5872 was opened 29 minutes before this PR (19:39 vs 20:08 UTC on 2026-06-25).
- **CI and mergeability verified.** #5872 is MERGEABLE; CLA signed; only failing check is Vercel (advisory, fork deploy perms — non-blocking on every external PR).

### Credit where due

The root-cause diagnosis here is spot-on: tracing the comment *"Create telemetry config manually to avoid deepcopy issues with thread locks"* directly above the block as confirmation that the dict-based approach was the deliberate replacement is exactly the right read. The regression test is structurally correct — `_safe_deepcopy_config` is patched at the right target (`mem0.memory.main._safe_deepcopy_config`) and the decorator/argument order is clean. The work here is genuinely appreciated.

### Evidence

- **This PR (5873):** CLA signed; Vercel advisory fail (non-blocking). Fix: correct. Test: valid for sync path; async path unguarded.
- **#5872 (recommended):** Diff read in full — same 1-line removal in `main.py`; `tests/memory/conftest.py` adds qdrant/grpc stubs for the directory; two tests — sync path and async guard — using `mocker` and `spy.assert_not_called()` / `spy.assert_called_once()`. MERGEABLE; CLA signed.
- **AsyncMemory path verified from code:** `main.py:2118–2124` — deepcopy result mutated in-place; correctly left untouched by both PRs.
- **Could NOT verify:** no CI Gate / test-matrix run shown for either PR (only Vercel advisory); correctness established by code tracing.

### Assessment

Recommendation: **close this in favour of #5872** (squash-merge #5872). Est. human time: ~2 min. Thanks for the careful dead-code archaeology here — this kind of diagnostic work is exactly what keeps the codebase clean. Please feel free to keep contributing; if #5872 falls through for any reason, this PR is right there as a fallback.

C1-BA-B1-F3:
Addressed both review suggestions on this branch: the test now patches `mem0.memory.main.SQLiteManager` and `os.makedirs` alongside `_safe_deepcopy_config`, so the telemetry init path stays side-effect free. Ready for another look.

kartik-mem0:

PR #5878, opened 2026-06-26
fix(memory): re-raise LLM extraction failures instead of returning [] (salvage of #5178)

Bartok9:
Salvage of #5178 by @thalesfsp — rebased onto current main with a regression guardrail test added.

## Summary

Re-raise provider errors during fact extraction instead of swallowing them and returning `[]`.

## Root Cause

**Symptom** — When the LLM call in `_add_to_vector_store` fails (HTTP 429, 5xx, connection error, timeout), `add()` silently returns an empty result. Callers cannot tell "LLM was unavailable" from "LLM ran fine and extracted no facts" — both look like `[]`.

**Root cause** — Both the sync `Memory._add_to_vector_store` and async `AsyncMemory._add_to_vector_store` wrap `self.llm.generate_response(...)` in a bare `except Exception` that logs and `return []`. The exception is destroyed at the point it occurs, so no signal reaches the caller.

**Evidence** — `mem0/memory/main.py`: the sync path (`logger.error("LLM extraction failed: {e}"); return []`) and the async path (`... (async): {e}"); return []`). Confirmed present on current `main` (`0fbbb2f`). The new regression test below fails on unmodified main (the call returns `[]`, no exception) and passes with the fix.

**Fix + why this level** — Replace `return []` with `raise` in both extraction error handlers, keeping the existing `logger.error`. This is the correct layer: extraction is where provider transport errors first surface, and re-raising lets any wrapping retry/fallback layer (e.g. a multi-provider failover) distinguish transport failure from an empty extraction. Patching a downstream caller can't recover information already discarded here. Parse errors further down still return `[]` (that genuinely means "no usable facts"), so only true provider exceptions propagate.

**Scope / risk** — Touches only the two extraction `except` blocks. Behavior change: an `add()` that previously returned `[]` on a provider outage now raises. This is the intended, documented contract change; callers relying on the silent-swallow behavior should wrap in try/except. Parse-error and empty-response handling is unchanged.

## Why it needed salvage

The original #5178 was conflicting against current main after pipeline refactors. Re-applied cleanly onto current `main` and added a guardrail test the original lacked.

## Changes from original
- Rebased onto current `main` (single clean commit, authorship credited).
- Added `test_llm_extraction_exception_is_reraised` — asserts a provider error propagates; **fails without the fix** (returns `[]`), passes with it.

```
$ python -m pytest tests/memory/test_main.py::TestAddToVectorStoreErrors -q
...                                                                      [100%]
3 passed in 0.49s

# Without the fix (main.py reverted), the new test fails as expected:
ERROR    mem0.memory.main:main.py:915 LLM extraction failed: 429 rate limit
FAILED tests/memory/test_main.py::...::test_llm_extraction_exception_is_reraised
1 failed
```
`ruff check` clean. Full `tests/memory/test_main.py` suite (46 tests) passes.

Credit: salvage of #5178 by @thalesfsp.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -22,6 +22,7 @@ from mem0.configs.prompts import (
     PROCEDURAL_MEMORY_SYSTEM_PROMPT,
     generate_additive_extraction_prompt,
 )
+from mem0.exceptions import LLMError
 from mem0.exceptions import ValidationError as Mem0ValidationError
 from mem0.memory.base import MemoryBase
 from mem0.memory.setup import mem0_dir, setup_config
@@ -917,7 +918,7 @@ class Memory(MemoryBase):
             # distinguish "LLM unavailable" (429/5xx/timeout) from "LLM
             # extracted no facts" -- both surfaced as an empty list.
             logger.error(f"LLM extraction failed: {e}")
-            raise
+            raise LLMError(f"LLM extraction failed: {e}") from e
 
         # Parse response
         try:
@@ -2543,7 +2544,7 @@ class AsyncMemory(MemoryBase):
             # Re-raise so callers can implement provider fallback / retry
             # (see sync counterpart for rationale).
             logger.error(f"LLM extraction failed (async): {e}")
-            raise
+            raise LLMError(f"LLM extraction failed: {e}") from e
 
         # Parse response
         try:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -912,8 +912,12 @@ class Memory(MemoryBase):
                 response_format={"type": "json_object"},
             )
         except Exception as e:
+            # Re-raise so callers can implement provider fallback / retry.
+            # The original silent ``return []`` made upstream callers unable to
+            # distinguish "LLM unavailable" (429/5xx/timeout) from "LLM
+            # extracted no facts" -- both surfaced as an empty list.
             logger.error(f"LLM extraction failed: {e}")
-            return []
+            raise
 
         # Parse response
         try:
@@ -2536,8 +2540,10 @@ class AsyncMemory(MemoryBase):
                 response_format={"type": "json_object"},
             )
         except Exception as e:
+            # Re-raise so callers can implement provider fallback / retry
+            # (see sync counterpart for rationale).
             logger.error(f"LLM extraction failed (async): {e}")
-            return []
+            raise
 
         # Parse response
         try:
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -79,6 +79,25 @@ class TestAddToVectorStoreErrors:
         assert mock_memory.llm.generate_response.call_count == 1
         assert result == []  # Should return empty list when no memories processed
 
+    def test_llm_extraction_exception_is_reraised(self, mocker, mock_memory):
+        """A provider error during fact extraction must propagate, not be swallowed.
+
+        Regression guard for the silent ``return []`` that made it impossible for
+        callers to tell "LLM unavailable" (429/5xx/timeout) from "no facts found".
+        Without the fix this raises AssertionError because the call returns [].
+        """
+
+        class _ProviderError(Exception):
+            pass
+
+        mock_memory.llm.generate_response.side_effect = _ProviderError("429 rate limit")
+        mocker.patch("mem0.memory.main.capture_event")
+
+        with pytest.raises(_ProviderError):
+            mock_memory._add_to_vector_store(
+                messages=[{"role": "user", "content": "test"}], metadata={}, filters={}, infer=True
+            )
+
 
 class TestPromptOverridesCustomInstructions:
     @pytest.fixture
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -6,6 +6,7 @@ from unittest.mock import MagicMock, Mock
 
 import pytest
 
+from mem0.exceptions import LLMError
 from mem0.memory.main import AsyncMemory, Memory
 
 
@@ -93,10 +94,13 @@ class TestAddToVectorStoreErrors:
         mock_memory.llm.generate_response.side_effect = _ProviderError("429 rate limit")
         mocker.patch("mem0.memory.main.capture_event")
 
-        with pytest.raises(_ProviderError):
+        with pytest.raises(LLMError) as exc_info:
             mock_memory._add_to_vector_store(
                 messages=[{"role": "user", "content": "test"}], metadata={}, filters={}, infer=True
             )
+        # The documented LLMError contract is honoured, and the original
+        # provider exception is preserved as the cause for debugging.
+        assert isinstance(exc_info.value.__cause__, _ProviderError)
 
 
 class TestPromptOverridesCustomInstructions:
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

Really solid salvage — the root diagnosis is exactly right and the PR description is the clearest explanation of this bug I've seen. Two things need to land before this merges.

### Strengths

- `main.py:914-916` and `2538-2540`: The asymmetry between the two `except` blocks is *exactly correct*. First block (LLM transport failure) now re-raises — right. Second block (parse failure at `main.py:929-931`) still returns `[]` — also right, because an LLM that responded with empty JSON *did succeed*; the distinction is real and the code honours it.
- The regression test uses a sentinel `_ProviderError` to avoid accidentally catching unrelated exceptions in `pytest.raises` — good technique.
- Both sync and async paths addressed in one commit; zero hot-path overhead (`raise` only executes on exception).
- `ruff check` clean; conventional commit title.

### Issues

**1. `tests/memory/test_main.py` — Async path is changed but not tested**

`AsyncMemory._add_to_vector_store` at `main.py:2538-2540` gets the same `return []` → `raise` change as the sync path, but `TestAsyncAddToVectorStoreErrors` has no counterpart for it. A future revert of the async `raise` back to `return []` would pass the test suite silently. The fix is a near-copy of the existing sync test:

```python
@pytest.mark.asyncio
async def test_async_llm_extraction_exception_is_reraised(self, mocker, mock_async_memory):
    class _ProviderError(Exception):
        pass
    mock_async_memory.llm.generate_response.side_effect = _ProviderError("429 rate limit")
    mocker.patch("mem0.memory.main.capture_event")
    with pytest.raises(_ProviderError):
        await mock_async_memory._add_to_vector_store(
            messages=[{"role": "user", "content": "test"}], metadata={}, filters={}, infer=True
        )
```

**2. `main.py:765` — `LLMError` contract mismatch**

`Memory.add()` docstring declares `Raises: LLMError: If LLM operations fail.` but `LLMError` is not imported in `main.py` (only `ValidationError` is, via line 25). The bare `raise` propagates raw provider exceptions (`openai.RateLimitError`, `anthropic.APIConnectionError`, etc.). Any caller following the documented contract with `except LLMError` will not catch these. Two valid resolutions:

- **Option A (preferred):** Wrap at both callsites: `raise LLMError(f"LLM extraction failed: {e}") from e` — `LLMError` exists at `mem0/exceptions.py:346` and supports `from e` for causal chaining.
- **Option B:** Update the docstring to `Raises: Exception:` and note that the raw provider exception propagates.

*Note: this docstring was a dead letter pre-PR too (`return []` never raised it), so this PR makes the discrepancy newly observable rather than introducing it. Either option resolves it.*

#### Minor (optional)

- **TS SDK parity:** `mem0-ts/src/oss/src/memory/index.ts:855-858` has the identical `catch (e) { console.error(...); return []; }` pattern. Not a gate for this PR (Python-only scope), but worth a follow-up.
- `main.py:935` / `2558`: `save_messages` is skipped on LLM transport failure (it only runs after successful extraction). Now that callers can see the failure, this gap is more visible — a comment noting the skip is intentional would help future readers.

### Evidence

- **CI:** Python SDK 3.10 / 3.11 / 3.12 → PASS; CI Gate → PASS. Vercel = advisory deploy-perms fail, ignored.
- **Targeted tests (with fix):** `pytest tests/memory/test_main.py::TestAddToVectorStoreErrors -v` → `3 passed in 15.84s`
- **Red-green proof:** impl reverted to main (`return []`), new test fails: `FAILED ... Failed: DID NOT RAISE _ProviderError` (log shows `ERROR LLM extraction failed: 429 rate limit`); fix restored, test passes. ✓
- **Ruff:** `All checks passed!`
- **Could NOT verify:** async exception propagation path through `asyncio.to_thread` — code tracing only (no async test yet, hence item #1 above).
- **Duplicate cluster:** PR #5178 (original by @thalesfsp) is still open but has merge conflicts and no regression test. Once this PR merges, #5178 can be closed.

### Assessment

Ready to merge: **With fixes** — please add the async test and resolve the `LLMError` docstring. Both are quick. Happy to re-review once you push.

For tracking on our side, mind opening a short issue describing this bug and linking it here? Makes it easier to reference and prioritize follow-ups like the TS side.

Bartok9:
Both items addressed in 2a3d1df:

**1. Async path now tested** — added `test_async_llm_extraction_exception_is_reraised` to `TestAsyncAddToVectorStoreErrors` (near-copy of the sync guard, using `effective_filters={}` for the async signature). A revert of the async `raise` back to `return []` now fails the suite.

**2. LLMError contract** — went with Option A. Both sync (`main.py:914`) and async (`2540`) sites now `raise LLMError(f"LLM extraction failed: {e}") from e`. `LLMError` is imported from `mem0.exceptions`, and `from e` preserves the raw provider exception as `__cause__` — both new tests assert `isinstance(exc_info.value.__cause__, _ProviderError)`. Callers following the documented `except LLMError` contract now catch these.

Verification:
- `ruff check mem0/memory/main.py tests/memory/test_main.py` → All checks passed
- `pytest tests/memory/test_main.py::TestAddToVectorStoreErrors tests/memory/test_main.py::TestAsyncAddToVectorStoreErrors` → 6 passed

On the minors: TS SDK parity (`index.ts:855`) and the intentional `save_messages` skip comment — happy to take both as a follow-up so this stays Python-scoped. Will open the tracking issue for the bug and link it here shortly.

Bartok9:
Tracking issue filed as #5903 (linked) per your request. The TS-side parity follow-up (`index.ts:855-858`, same `catch → return []` pattern) is noted in it.

kartik-mem0:
## LGTM — re-review complete, ready for squash-merge

Both items from the [prior review](https://github.com/mem0ai/mem0/pull/5878#issuecomment-2969785920) are fully addressed.

### What changed in the follow-up commit

- **`LLMError` wrapping** (`main.py:916`, `2547`): `raise LLMError(f"LLM extraction failed: {e}") from e` — Option A, as requested. `__cause__` is preserved so callers can inspect the original provider exception.
- **Async regression test** (`tests/memory/test_main.py:270–295`): `test_async_llm_extraction_exception_is_reraised` added. Correctly uses `effective_filters={}` (matching the async signature at `main.py:2449`), asserts `isinstance(exc_info.value.__cause__, _ProviderError)` — pins the chain, not just the type.

### Strengths

- `main.py:914–916` and `2541–2547`: The `except` asymmetry is exactly right — transport failure re-raises, parse failure (`main.py:929–931` / `2553–2559`) still returns `[]`. The distinction is real and the code honours it.
- `Memory.add()` at `main.py:765` already declared `Raises: LLMError`. This PR makes the implementation match its own docstring for the first time.
- Both tests use a sentinel `_ProviderError` subclass and assert `__cause__` — tight regression guards that can't be gamed by coincidental raises elsewhere.
- Zero hot-path overhead — `raise` only executes on exception.

### Issues

**Important (for tracking, not a merge blocker here):** The TS SDK has the identical silent-swallow at `mem0-ts/src/oss/src/memory/index.ts:855-858` — `catch (e) { console.error(...); return []; }`. Issue #5903 (filed by the same contributor) already documents this as a follow-up. I flagged it Minor in the prior review and it's tracked, but it's worth making sure the TS fix doesn't get lost.

Minor: The async test patches `EmbedderFactory.create` again even though `mock_async_memory`'s fixture already does it via `_setup_mocks`. The extra patch is dead code — no harm, but remove it for clarity on the next pass.

### Evidence

- **CI:** Python SDK 3.10 / 3.11 / 3.12 → PASS; CI Gate → PASS. Vercel = advisory deploy-perms fail, ignored.
- **Targeted tests (with fix):** `pytest tests/memory/test_main.py::TestAddToVectorStoreErrors tests/memory/test_main.py::TestAsyncAddToVectorStoreErrors -v` → **6 passed in 0.67s**
- **Red-green proof:** both new tests fail (`Failed: DID NOT RAISE LLMError`) with `return []` restored; pass with fix ✓
- **Ruff:** `All checks passed!`
- **Linked issue:** #5903 — PR addresses the exact reported root cause; author also documents TS follow-up there
- **Cross-package:** TS counterpart at `index.ts:855-858` still swallows — tracked in #5903, not a blocker for this Python-scoped fix
- **Could NOT verify:** live provider transport failures (429 / 5xx / timeout) — code tracing only; the unit test covers the mechanism completely

### Assessment

The fix is at the right layer, fulfils the pre-existing documented contract, both sync and async paths are tested with a tight red-green proof, and the exception chain is preserved for callers who need to inspect the original error. Please make sure #5903 gets a TS-side fix PR queued as a follow-up.

Also: once this merges, PR #5178 (the original by the credited author) has merge conflicts and no regression test — it can be closed.

Bartok9:
Thanks for the thorough re-review — glad the `__cause__` chaining and the sync/async symmetry landed cleanly.

On the minor: confirmed, the async test's extra `EmbedderFactory.create` patch is redundant with `mock_async_memory`'s `_setup_mocks` fixture. It's harmless but dead — I'll drop it. Happy to either push it as a tiny follow-up commit here, or fold it into the TS-parity PR I'm queuing against #5903 so this one can squash-merge as-is right now. Your call — I'd lean toward squashing this as-is to avoid churn, and cleaning up the patch in the #5903 follow-up.

TS-side fix (`index.ts:855-858`) is on my list and tracked in #5903 so it won't get lost.

kartik-mem0 (approved):

PR #5880, opened 2026-06-26
fix(memory): tolerate malformed fact-extraction shapes from LLMs (salvage of #5176)

Bartok9:
Salvage of #5176 by @thalesfsp — rebased onto current main with guardrail tests.

## Summary

Tolerate the malformed `memory` item shapes real LLMs emit during fact extraction, so they aren't silently dropped (or, worse, crash the pipeline).

## Root Cause

**Symptom** — When an LLM (especially reasoning models on OpenAI-compatible APIs such as grok) returns extracted memories as plain strings or single-key "key-as-fact" dicts instead of `{"text": "..."}`, memories are silently lost — and a plain-string item crashes `add()` outright.

**Root cause** — In both `Memory._add_to_vector_store` and `AsyncMemory._add_to_vector_store` (`mem0/memory/main.py`), right after parsing the extraction response the code does `mem_texts = [m.get("text", "") for m in extracted_memories if m.get("text")]` and later `text = mem.get("text")`. If an item is a `str`, `.get` raises `AttributeError`; if it's `{"User likes coffee": "..."}` (key-as-fact) or `{"fact": "..."}`, `.get("text")` is empty, so the memory is dropped.

**Evidence** — Confirmed on current `main` (`0fbbb2f`): both extraction paths still call `m.get("text", "")` with no shape normalization. The new integration test fails on unmodified main with `AttributeError` at `mem0/memory/main.py:939`.

**Fix + why this level** — Add `normalize_extracted_memories()` in `mem0/memory/utils.py` and call it in both paths immediately after parsing, before any `.get("text")` access. It maps plain strings → `{"text": s}`, `{"fact": ...}` → adds `text`, single-key key-as-fact dicts → recovers the key as `text`, preserves all other keys (`attributed_to`, `linked_memory_ids`) on well-formed dicts, and drops unrecognized/empty shapes with a warning. Normalizing once at the parse boundary is the right level: every downstream consumer assumes `{"text": ...}`, so fixing the shape at the source avoids scattering defensive `.get` guards across embed/dedup/persist phases.

**Scope / risk** — Touches `utils.py` (new pure function) and the two parse blocks in `main.py`. Well-formed `{"text": ...}` items pass through unchanged (keys preserved), so the happy path is unaffected. Unrecognized shapes that previously crashed or were dropped are now dropped with a warning.

## Why it needed salvage
Original #5176 conflicted against current main after pipeline refactors (the helper name and call sites moved). Re-implemented cleanly onto current `main`.

## Changes from original
- Rebased onto current `main` (single clean commit, credited to @thalesfsp).
- Added 6 unit tests for `normalize_extracted_memories` + 1 integration test that drives `_add_to_vector_store` with malformed shapes. **Integration test fails without the fix (`AttributeError`).**

```
$ python -m pytest tests/memory/test_main.py tests/memory/test_memory_utils.py -q
.......................................................................  [100%]
71 passed in 1.20s

# Without the fix:
mem0/memory/main.py:939: AttributeError
FAILED tests/memory/test_main.py::...::test_malformed_memory_shapes_do_not_crash_extraction
```
`ruff check` clean.

Credit: salvage of #5176 by @thalesfsp.

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -93,6 +93,12 @@ def normalize_facts(raw_facts):
     Smaller LLMs (e.g. llama3.1:8b) sometimes return facts as objects
     like {"fact": "..."} or {"text": "..."} instead of plain strings.
     This mirrors the TypeScript FactRetrievalSchema validation.
+
+    NOTE: This is the FACT_RETRIEVAL-path normalizer, retained for external
+    callers and the legacy two-phase extraction flow. The current single-pass
+    additive pipeline in ``main.py`` uses ``normalize_extracted_memories``
+    (below) instead, which operates on dict-shaped ``memory`` items rather
+    than plain fact strings. Kept here so both shapes stay covered.
     """
     if not raw_facts:
         return []
@@ -146,6 +152,10 @@ def normalize_extracted_memories(raw_memories):
                 if isinstance(key, str) and key.strip():
                     text = key.strip()
                 elif isinstance(value, str) and value.strip():
+                    # Dead branch under json.loads (object keys are always
+                    # str, so the `if` above fires) — only reachable if a
+                    # caller hand-builds a dict with a non-str / empty key.
+                    # Retained defensively for that non-JSON path.
                     text = value.strip()
             if text:
                 out = dict(item)
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -112,6 +112,52 @@ def normalize_facts(raw_facts):
     return normalized
 
 
+def normalize_extracted_memories(raw_memories):
+    """Normalize LLM-extracted ``memory`` items to a list of dicts with a ``text`` key.
+
+    The additive-extraction prompt asks the LLM for objects like
+    ``{"text": "...", "attributed_to": "...", "linked_memory_ids": [...]}``,
+    but real LLMs (especially reasoning models on OpenAI-compatible APIs such
+    as grok) sometimes emit malformed shapes:
+
+    - plain string:                ``"User likes coffee"``
+    - well-formed dict:            ``{"text": "..."}`` (passed through)
+    - ``{"fact": "..."}``:         older / TS-style key
+    - single-key key-as-fact dict: ``{"User likes coffee": "<value>"}``
+      (the fact text is the dict key)
+
+    Downstream code does ``m.get("text", "")``, which raises AttributeError on
+    a plain string and silently drops the key-as-fact shape. Normalizing here
+    preserves any other keys (attributed_to, linked_memory_ids) on well-formed
+    dicts while recovering text from the malformed shapes.
+    """
+    if not raw_memories:
+        return []
+    normalized = []
+    for item in raw_memories:
+        if isinstance(item, str):
+            text = item.strip()
+            if text:
+                normalized.append({"text": text})
+        elif isinstance(item, dict):
+            text = item.get("text") or item.get("fact")
+            if text is None and len(item) == 1:
+                key, value = next(iter(item.items()))
+                if isinstance(key, str) and key.strip():
+                    text = key.strip()
+                elif isinstance(value, str) and value.strip():
+                    text = value.strip()
+            if text:
+                out = dict(item)
+                out["text"] = text
+                normalized.append(out)
+            else:
+                logger.warning("Unexpected memory shape from LLM, skipping: %s", item)
+        else:
+            logger.warning("Unexpected memory shape from LLM, skipping: %s", item)
+    return normalized
+
+
 def remove_code_blocks(content: str) -> str:
     """
     Removes enclosing code block markers ```[language] and ``` from a given string.
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -53,6 +53,7 @@ from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
 from mem0.memory.utils import (
     extract_json,
+    normalize_extracted_memories,
     parse_messages,
     parse_vision_messages,
     process_telemetry_filters,
@@ -972,6 +973,10 @@ class Memory(MemoryBase):
         except Exception as e:
             logger.error(f"Error parsing extraction response: {e}")
             extracted_memories = []
+
+        # Tolerate malformed LLM shapes (plain strings, {fact:...}, key-as-fact
+        # dicts) before downstream code calls ``.get("text")`` on each item.
+        extracted_memories = normalize_extracted_memories(extracted_memories)
 
         if not extracted_memories:
             # Save messages even if nothing extracted
@@ -2621,6 +2626,10 @@ class AsyncMemory(MemoryBase):
         except Exception as e:
             logger.error(f"Error parsing extraction response (async): {e}")
             extracted_memories = []
+
+        # Tolerate malformed LLM shapes (plain strings, {fact:...}, key-as-fact
+        # dicts) before downstream code calls ``.get("text")`` on each item.
+        extracted_memories = normalize_extracted_memories(extracted_memories)
 
         if not extracted_memories:
             await asyncio.to_thread(self.db.save_messages, messages, session_scope)
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -102,6 +102,29 @@ class TestAddToVectorStoreErrors:
         # provider exception is preserved as the cause for debugging.
         assert isinstance(exc_info.value.__cause__, _ProviderError)
 
+    def test_malformed_memory_shapes_do_not_crash_extraction(self, mocker, mock_memory):
+        """A plain-string item in the LLM 'memory' array must not crash the pipeline.
+
+        Without normalization, downstream ``m.get("text")`` raises AttributeError
+        on a str. Guards #5176: real reasoning models emit string / key-as-fact items.
+        """
+        import json as _json
+
+        mock_memory.llm.generate_response.return_value = _json.dumps(
+            {"memory": ["User likes coffee", {"User is tired": "noise"}]}
+        )
+        mocker.patch("mem0.memory.main.capture_event")
+        mock_memory.embedding_model = MagicMock()
+        mock_memory.embedding_model.embed_batch.return_value = [[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]]
+        mock_memory.vector_store = MagicMock()
+        mock_memory.vector_store.search.return_value = []
+
+        # Must not raise; both malformed shapes recovered into text-bearing records.
+        result = mock_memory._add_to_vector_store(
+            messages=[{"role": "user", "content": "test"}], metadata={}, filters={"user_id": "u1"}, infer=True
+        )
+        assert isinstance(result, list)
+
 
 class TestPromptOverridesCustomInstructions:
     @pytest.fixture
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -123,7 +123,14 @@ class TestAddToVectorStoreErrors:
         result = mock_memory._add_to_vector_store(
             messages=[{"role": "user", "content": "test"}], metadata={}, filters={"user_id": "u1"}, infer=True
         )
+        # Both inputs (plain string + key-as-fact dict) must survive normalization
+        # and be persisted as ADD records — not silently dropped. isinstance(list)
+        # alone would pass even on a total drop, so pin the count and the texts.
         assert isinstance(result, list)
+        assert len(result) == 2
+        recovered = {r["memory"] for r in result}
+        assert recovered == {"User likes coffee", "User is tired"}
+        assert all(r["event"] == "ADD" for r in result)
 
 
 class TestPromptOverridesCustomInstructions:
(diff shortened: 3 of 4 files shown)


kartik-mem0:
## Changes Requested

The normalization logic is sound and the root cause is correctly diagnosed — the crash path is real, the placement (parse boundary, before any \`.get(\"text\")\`) is right, and sync/async are covered symmetrically. A few things need tightening before this is merge-ready.

### Strengths

- **`mem0/memory/utils.py` — `normalize_extracted_memories`**: Correct parse-boundary placement, non-destructive (preserves `attributed_to`, `linked_memory_ids` via `dict(item)`), well-documented docstring enumerating each shape, clean early-return on empty input.
- **`mem0/memory/main.py` — call-site symmetry**: Both sync (~line 931) and async (~line 2563) `_add_to_vector_store` get the call, at the same logical phase in each. Correct.
- **`tests/memory/test_memory_utils.py` — unit coverage**: Six focused unit tests cover all four documented shapes (well-formed passthrough, plain string wrap, `{\"fact\":...}` mapping, key-as-fact, empty/unrecognized drop, mixed batch). These are the right assertions at the right layer.
- **Pure additions, zero regression risk**: Well-formed `{\"text\":...}` items pass through unchanged — the happy path is structurally unaffected.
- **Red-green confirmed**: `AttributeError: 'str' object has no attribute 'get'` at `main.py:939` on pre-fix main; 7/7 pass with the fix. The crash is real and the fix eliminates it.

### Issues

**1. `tests/memory/test_main.py:100` — integration test asserts almost nothing**

```python
assert isinstance(result, list)
```

The docstring says *"both malformed shapes recovered into text-bearing records"* but `isinstance([], list)` is `True`, so this assertion passes even when both inputs were silently dropped. The mocks provide two embeddings and an empty search result (no duplicate), so `result` should have two entries. Please add `assert len(result) == 2` — or check the text values — to actually pin the recovery behavior the test describes.

**2. `tests/memory/test_main.py` — async path has no integration test**

The normalization call in `async def _add_to_vector_store` (~line 2563 of `main.py`) is not covered by any test. A future regression there would be invisible. A minimal `@pytest.mark.asyncio` mirror of `test_malformed_memory_shapes_do_not_crash_extraction` calling `AsyncMemory._add_to_vector_store` with the same malformed inputs would cover this.

**3. `mem0-ts/src/oss/src/memory/index.ts` — TypeScript SDK has the same silent-drop bug**

When a reasoning model emits `{\"memory\": [\"User likes coffee\"]}`, the `AdditiveExtractionSchema` Zod parse throws (strict `z.object({ text: z.string() })` rejects a plain string), the catch block assigns the raw array directly, and `extractedMemories.map((m) => m.text ?? \"\").filter(t => t.length > 0)` silently discards all plain-string items — no crash, no warning, memory lost. The Python fix doesn't carry over. A follow-up issue tracking a TS `normalizeExtractedMemories` would close the parity gap — merging this first is fine, but the issue should be filed and linked before close.

#### Minor (optional)

**4. `mem0/memory/utils.py` — `normalize_facts` is not called from any production file**

The `from mem0.memory.utils import (...)` in `main.py` doesn't include `normalize_facts`; it's only in test imports. A short comment on the function — *"called from the FACT_RETRIEVAL path (see issue #X)"* or *"legacy: retained for external callers"* — would clarify whether it's intentional infrastructure or orphaned code, especially now that a second normalizer lives right below it.

**5. `mem0/memory/main.py` — import ordering**

`normalize_extracted_memories` is inserted before `extract_json` in the import block (`e` < `n`). The existing block is alphabetically ordered (`extract_json → parse_messages → ...`). Moving it after `parse_vision_messages` (or after `remove_code_blocks`) keeps isort `profile = \"black\"` happy.

**6. `mem0/memory/utils.py` — the `elif isinstance(value, str)` branch can never fire under realistic inputs**

JSON object keys are always strings in Python's `json` module, so `isinstance(key, str)` is always `True`, and `elif isinstance(value, str)` only fires when `key.strip()` is empty (all-whitespace JSON key — no real LLM emits this). A comment explaining the dead branch or removing it would reduce confusion for future readers.

### Evidence

- **CI:** Python SDK 3.10 / 3.11 / 3.12 all pass; CI Gate pass. Vercel deploy fail — ignored (fork perms, advisory only).
- **Targeted tests:** `pytest tests/memory/test_main.py::TestAddToVectorStoreErrors::test_malformed_memory_shapes_do_not_crash_extraction tests/memory/test_memory_utils.py::TestNormalizeExtractedMemories -v --tb=short` → **7 passed in 0.38s**
- **Red-green proof:** Checked out `mem0/memory/main.py` + `utils.py` from `origin/main` inside the PR worktree → `AttributeError: 'str' object has no attribute 'get'` at `main.py:939`, plus `ImportError: cannot import name 'normalize_extracted_memories'`. Restored PR state → 7 passed. Regression test bites.
- **Linked issue:** None — see nudge below.
- **Could NOT verify:** TypeScript bug in a live environment; confirmed by static read of `mem0-ts/src/oss/src/memory/index.ts` + `prompts/index.ts`.
- **PR #5176 (original):** Still open; no CI, CLA not signed, conflicted with current main per the PR body. Not mergeable; Superseded verdict does not apply.

### Assessment

Ready to merge: **With fixes** — please address findings 1 and 2 (strengthen the integration test + add an async test); file an issue for finding 3 (TypeScript gap); findings 4–6 are optional polish. Happy to re-review once you push.

---

Bartok9:
Thanks for the thorough review — all points addressed in 73c223d.

**1. Integration test now pins recovery behavior** — replaced the toothless `assert isinstance(result, list)` (which passed even on a total silent drop) with `assert len(result) == 2`, plus an exact-text check `{"User likes coffee", "User is tired"}` and `event == "ADD"`. Now actually bites if either malformed shape gets dropped.

**2. Async integration test added** — `test_async_malformed_memory_shapes_do_not_crash_extraction` in `TestAsyncAddToVectorStoreErrors`, a `@pytest.mark.asyncio` mirror of the sync guard, exercising `AsyncMemory._add_to_vector_store` with the same malformed inputs (uses `effective_filters={}` for the async signature). The normalize call at `main.py` ~2564 is no longer untested.

**3. TS parity gap filed** — #5902 tracks adding a TS `normalizeExtractedMemories` to close the silent-drop gap in `mem0-ts/src/oss/src/memory/index.ts`.

**4. `normalize_facts` documented** — added a note that it's the FACT_RETRIEVAL-path / legacy normalizer (operates on plain fact strings), distinct from the single-pass `normalize_extracted_memories` right below it.

**5. Import ordering fixed** — `extract_json` now precedes `normalize_extracted_memories` (isort `profile = "black"`).

**6. Dead branch commented** — added a comment on `elif isinstance(value, str)` noting it's unreachable under `json.loads` (object keys are always str) and retained defensively for hand-built / non-JSON dicts.

Also filed #5901 to track the underlying Python bug, as requested.

Verification:
- `pytest tests/memory/test_main.py tests/memory/test_memory_utils.py` → **72 passed**
- `ruff check mem0/memory/main.py mem0/memory/utils.py tests/memory/test_main.py tests/memory/test_memory_utils.py` → All checks passed

CI was already green (Python SDK 3.10/3.11/3.12 + CI Gate pass; only the Vercel deploy fails on fork perms, advisory). Happy to take another pass if anything else needs tightening.

kartik-mem0:

Re-review against 73c223d (second commit, pushed 2026-06-26T12:11Z).

### Verified fixed

- **Finding 1** (`tests/memory/test_main.py:100`) — integration test now asserts `len(result) == 2`, `recovered == {"User likes coffee", "User is tired"}`, `all(r["event"] == "ADD")`. The toothless `isinstance(result, list)` is gone.
- **Finding 2** — `TestAsyncAddToVectorStoreErrors::test_async_malformed_memory_shapes_do_not_crash_extraction` added; mirrors the sync test exactly, covers the `normalize_extracted_memories` call at `main.py:~2564`.
- **Finding 3** — TypeScript parity gap tracked at #5902. No code change needed here; the issue is open and linked.
- **Finding 4** (`mem0/memory/utils.py:93`) — `normalize_facts` docstring now explains FACT_RETRIEVAL-path role vs the new single-pass normalizer. Clear.
- **Finding 5** (`mem0/memory/main.py:52`) — import ordering fixed; `extract_json` comes before `normalize_extracted_memories`. isort-clean.
- **Finding 6** (`mem0/memory/utils.py:~148`) — dead-branch comment added; explains the `elif isinstance(value, str)` path is unreachable under `json.loads` but retained for non-JSON callers.

### Evidence

- **CI:** Python SDK 3.10 / 3.11 / 3.12 all pass; CI Gate pass. Vercel deploy fail — ignored (fork perms, advisory).
- **Targeted tests:** `pytest tests/memory/test_main.py tests/memory/test_memory_utils.py -v` → **72 passed in 1.18s**
- **Red-green proof (sync):** `test_malformed_memory_shapes_do_not_crash_extraction` — FAIL without fix (`AttributeError: 'str' object has no attribute 'get'` at `main.py:939`); PASS with fix. ✓
- **Red-green proof (async):** `test_async_malformed_memory_shapes_do_not_crash_extraction` — FAIL without fix (same `AttributeError` at `main.py:2562`); PASS with fix. ✓
- **Lint:** `ruff check` clean on all four touched files.
- **Linked issue:** #5901 — crash path real and fixed. TS gap tracked at #5902.
- **Could NOT verify:** TypeScript behavior in a live environment; confirmed by static read of `mem0-ts/src/oss/src/memory/index.ts`.

### Assessment

Normalization at the parse boundary is the right layer; sync and async paths are covered symmetrically; regression tests bite both crash sites; all prior findings addressed. Clean.

kartik-mem0:

### What this PR does
Verified against `247556c` (new head): re-review confirms the only change since the prior LGTM is a merge from `upstream/main` — no new scope, no regression.

### Strengths
- `mem0/memory/main.py:917-921` and `:2547-2550` — the merge pulled in an unrelated, already-on-`main` `LLMError` raise (from a separate PR); composes cleanly with this PR's `normalize_extracted_memories()` call (sequential, no interaction bug) — confirmed by reading both blocks directly.

### Issues
Minor (optional): none

### Evidence
- **CI:** Python SDK 3.10/3.11/3.12 pass, CI Gate pass, license/cla pass. Vercel fail (advisory preview-deploy, fork perms — ignored).
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_main.py tests/memory/test_memory_utils.py -v` → 74 passed in 1.23s (fresh, this run, at `247556c`).
- **Red-green proof:** reverted only this PR's own two files (`main.py`, `utils.py`) to the `origin/main` tip, isolating this PR's fix from the unrelated merged-in content → both malformed-shape tests FAIL with `AttributeError: 'str' object has no attribute 'get'` (`main.py:944` sync, `:2569` async); restored → 74 passed.
- **Linked issue:** #5901 — PR addresses the reported root cause. #5902 (TS parity) is intentionally out of scope here and tracked separately — a candidate TS-side fix is already up in #5926.
- **Cross-package check:** TS OSS memory path unaffected (this fix is Python-only); the TS parity gap predates this PR and is unchanged by this delta.
- **Could NOT verify:** live LLM provider behavior for every malformed-shape variant in the wild — verified via unit tests + code tracing only.

### Assessment
The new commit is a routine merge with zero own-scope changes; the previously-reviewed fix, its tests, and CI all hold at the new head.

kartik-mem0:
Checking back on this one. The review still stands, the normalization approach is right and the code is good. Two things have changed since the last pass, both of which make landing this more urgent:

1. This branch now conflicts with `main` and shows as not mergeable, so it cannot be merged as-is. It needs a rebase onto current `main`. The conflict came from unrelated churn in `mem0/memory/main.py`, so it should be a mechanical resolve rather than a rethink.
2. Another report of this exact crash came in independently on a live ingestion run (Chroma plus a local Ollama model), where it aborted a 255-document batch partway through. That was filed as #6724 and I've closed it as a duplicate of #5901, pointing at this PR.

If you can rebase and push, I'll re-review the new head right away and get it into the merge queue. If you'd rather hand it off at this point, say so and I'll pick up the rebase myself, no problem either way. Thanks for sticking with this one.

Bartok9:
Thanks for the nudge — rebased onto current `main` and force-pushed (`5a2a681`).

Conflicts were mechanical in the test files only (`tests/memory/test_main.py`, `tests/memory/test_memory_utils.py`): kept the newer main-side LLMError re-raise guards alongside the malformed-shape recovery tests (sync + async), and left production code as the same parse-boundary `normalize_extracted_memories()` call on both paths.

Also noting #6724 / the live Chroma+Ollama batch abort — good signal this is still biting in the wild; happy for another re-review when convenient.

mem0/memory/utils.py
```diff
@@ -112,6 +118,56 @@ def normalize_facts(raw_facts):
...
+
+    Downstream code does ``m.get("text", "")``, which raises AttributeError on
+    a plain string and silently drops the key-as-fact shape. Normalizing here
+    preserves any other keys (attributed_to, linked_memory_ids) on well-formed
+    dicts while recovering text from the malformed shapes.
+    """
+    if not raw_memories:
+        return []
+    normalized = []
+    for item in raw_memories:
```
kartik-mem0, line 143: No list guard: if `raw_memories` is a dict or bare string (a single-item LLM response) this iterates dict keys or string characters, fabricating fake memories instead of erroring. `normalize_extracted_memories({"text": "solo"})` returns `[{"text": "text"}]`, silently dropping "solo".

Guard before the loop: wrap a dict in a list, drop anything else that isn't a list already.

mem0/memory/utils.py
```diff
@@ -112,6 +118,56 @@ def normalize_facts(raw_facts):
...
+        if isinstance(item, str):
+            text = item.strip()
+            if text:
+                normalized.append({"text": text})
+        elif isinstance(item, dict):
+            text = item.get("text") or item.get("fact")
```
kartik-mem0, line 149: `text = item.get("text") or item.get("fact")` never checks `text` is a string. `{"fact": ["a", "b"]}` passes through with `text` as a list, then crashes later in `main.py` with `TypeError: unhashable type: 'list'` when it's used as an embed_map key, a new crash this PR introduces for a shape the old code silently dropped.

Add `and isinstance(text, str)` to both `.get()` fallbacks before accepting it.

mem0/memory/utils.py
```diff
@@ -112,6 +118,56 @@ def normalize_facts(raw_facts):
...
+            if text is None and len(item) == 1:
+                key, value = next(iter(item.items()))
+                if isinstance(key, str) and key.strip():
+                    text = key.strip()
+                elif isinstance(value, str) and value.strip():
```
kartik-mem0, line 154: Dead branch: the comment above it says JSON object keys are always `str`, so `isinstance(key, str)` above always fires first and this `elif` never runs (the only caller is JSON-parsed LLM output). Delete lines 154-159, net -6 lines.

kartik-mem0 (commented):
## Changes requested

Fixes the crash on malformed LLM extraction items, but the new normalizer trusts the shapes it wasn't tested against: a non-list `memory` value corrupts data instead of erroring, and a non-string `fact` value crashes downstream.

**To unblock:**
1. `mem0/memory/utils.py:143` — guard `raw_memories` is a list before iterating (detail in the inline comment)
2. `mem0/memory/utils.py:149` — check `text` is a string before accepting it from `fact` (detail in the inline comment)
3. `mem0-ts/src/oss/src/memory/index.ts:958` (no anchor, TS side) — same malformed-shape drop exists in the TS SDK; a follow-up issue would be enough if you'd rather not scope it into this PR

Optional: `mem0/memory/utils.py:154-159` is a self-admitted dead branch, see inline comment.

- **Problem is real:** reproduces on `main` at `mem0/memory/main.py:982` (sync) and `:2630` (async), `AttributeError: 'str' object has no attribute 'get'` when an extracted item is a plain string. Sandbox scenario `sandbox/scenarios/issue5901_malformed_extraction_shapes.py` crashes on `main`, passes on this PR head in both sync and async against real Qdrant.
- **Tests / red-green:** `tests/memory/test_main.py -k malformed` plus `tests/memory/test_memory_utils.py::TestNormalizeExtractedMemories`, 8/8 pass on `5a2a681`. Reverted `main.py`/`utils.py` to the pre-fix commit: both malformed-shape tests fail with the exact pre-fix `AttributeError`. Restored: pass again.
- **Gap the tests miss:** all 8 tests wrap the LLM's `memory` value in a list first. `normalize_extracted_memories({"text": "solo"})` returns `[{"text": "text"}]`, fabricating a memory from the dict key and silently dropping "solo"; a bare string produces one fake memory per character. Neither shape is tested, and both are realistic LLM failure modes (singleton collapse) within the exact bug class this PR targets.
- **CI:** green on all required jobs (Python SDK build matrix, CI Gate, Mem0 Plugin). Vercel red (fork perms, non-blocking), CLA pending (author action, non-blocking).
- **Security:** cleared, no new attack surface. The gap above is data-integrity/correctness, not injection, SSRF, or authz.
- **Bloat (ponytail):** `mem0/memory/utils.py:154-159` is a dead branch, `net: -6 lines possible`.
- **Not verified:** live LLM output shapes beyond the ones named in the PR description and issue; verified by code tracing plus the mocked sandbox run, not a real model call.

</details>

Good root-cause placement, one shared normalizer instead of scattered `.get()` guards. Push a type guard on `raw_memories` and `text` and I will re-review.

PR #5881, opened 2026-06-26
fix(memory): anchor extraction prompt date to created_at metadata (salvage of #5087)

Bartok9:
Salvage of #5087 by @Deadpoolmine — rebased onto current main with a guardrail test.

## Summary

When `created_at` is supplied via metadata, anchor the additive-extraction prompt's "Current Date" to that timestamp so the LLM's temporal reasoning matches the date ultimately stored in the memory.

## Root Cause

**Symptom** — For backfilled/imported memories where `created_at` is passed in metadata, the timestamp the LLM sees in the extraction prompt (used for relative-date reasoning like "yesterday") differs from the `created_at` stored in the memory metadata. The two time sources disagree.

**Root cause** — Both `Memory._add_to_vector_store` and `AsyncMemory._add_to_vector_store` call `generate_additive_extraction_prompt(...)` without passing `current_date`. The prompt builder then falls back through `_resolve_dates()` to `datetime.now(timezone.utc).date()` (mem0/configs/prompts.py:1007-1014) — wall-clock now — while the stored `created_at` comes from the caller's metadata.

**Evidence** — Confirmed on current `main` (`0fbbb2f`): `generate_additive_extraction_prompt` already accepts `current_date`/`timestamp` (signature at prompts.py:1016), but neither call site in `main.py` forwards anything, so it always defaults to today. The new guardrail test fails on unmodified main (`current_date` is `None`).

**Fix + why this level** — At each call site, derive `extraction_current_date` from `metadata["created_at"]` (parsed via `datetime.fromisoformat(...).astimezone(utc).date()`, falling back to `None` on malformed values) and pass it as `current_date`. This is the correct level: the prompt builder already supports an injected date — the only gap was the caller not supplying the value it already has in `metadata`. Parsing failures degrade gracefully to the previous now()-based behavior.

**Scope / risk** — Touches only the two `generate_additive_extraction_prompt` call sites. When `created_at` is absent or unparseable, behavior is identical to before (`current_date=None` → defaults to today). No prompt-format or API change.

## Why it needed salvage
Original #5087 predates the v3 single-pass pipeline + prompt-builder refactor and conflicted. Re-applied to current `main`.

## Changes from original
- Rebased onto current `main` (single clean commit, credited to @Deadpoolmine).
- Added a guardrail test asserting the prompt receives `current_date="2021-06-15"` for `created_at="2021-06-15T12:00:00+00:00"`. **Fails without the fix.**

```
$ python -m pytest tests/memory/test_main.py -q
..............................................                           [100%]
46 passed in 1.14s

# Without the fix:
FAILED tests/memory/test_main.py::...::test_created_at_metadata_anchors_extraction_prompt_date
```
`ruff check` clean.

Credit: salvage of #5087 by @Deadpoolmine.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -896,11 +896,25 @@ class Memory(MemoryBase):
 
         custom_instr = prompt or self.custom_instructions
 
+        # When created_at is supplied via metadata (e.g. backfilled/imported
+        # memories), anchor the extraction prompt's "current date" to it so the
+        # LLM's temporal reasoning matches the timestamp ultimately stored in
+        # the memory metadata, instead of drifting to wall-clock now().
+        extraction_current_date = None
+        if metadata and metadata.get("created_at"):
+            try:
+                extraction_current_date = (
+                    datetime.fromisoformat(metadata["created_at"]).astimezone(timezone.utc).date().isoformat()
+                )
+            except (ValueError, TypeError):
+                extraction_current_date = None
+
         user_prompt = generate_additive_extraction_prompt(
             existing_memories=existing_memories,
             new_messages=parsed_messages,
             last_k_messages=last_messages,
             custom_instructions=custom_instr,
+            current_date=extraction_current_date,
         )
 
         try:
@@ -2519,11 +2533,25 @@ class AsyncMemory(MemoryBase):
 
         custom_instr = prompt or self.custom_instructions
 
+        # When created_at is supplied via metadata (e.g. backfilled/imported
+        # memories), anchor the extraction prompt's "current date" to it so the
+        # LLM's temporal reasoning matches the timestamp ultimately stored in
+        # the memory metadata, instead of drifting to wall-clock now().
+        extraction_current_date = None
+        if metadata and metadata.get("created_at"):
+            try:
+                extraction_current_date = (
+                    datetime.fromisoformat(metadata["created_at"]).astimezone(timezone.utc).date().isoformat()
+                )
+            except (ValueError, TypeError):
+                extraction_current_date = None
+
         user_prompt = generate_additive_extraction_prompt(
             existing_memories=existing_memories,
             new_messages=parsed_messages,
             last_k_messages=last_messages,
             custom_instructions=custom_instr,
+            current_date=extraction_current_date,
         )
 
         try:
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -79,6 +79,29 @@ class TestAddToVectorStoreErrors:
         assert mock_memory.llm.generate_response.call_count == 1
         assert result == []  # Should return empty list when no memories processed
 
+    def test_created_at_metadata_anchors_extraction_prompt_date(self, mocker, mock_memory):
+        """When created_at is supplied via metadata, the extraction prompt's
+        current_date must be anchored to it (not wall-clock now()).
+
+        Guards #5087: prompt/metadata timestamp mismatch for backfilled memories.
+        Without the fix, current_date is None (-> defaults to today).
+        """
+        spy = mocker.patch(
+            "mem0.memory.main.generate_additive_extraction_prompt", return_value="PROMPT"
+        )
+        mock_memory.llm.generate_response.return_value = "{}"  # no memories -> early return ok
+        mocker.patch("mem0.memory.main.capture_event")
+
+        mock_memory._add_to_vector_store(
+            messages=[{"role": "user", "content": "test"}],
+            metadata={"created_at": "2021-06-15T12:00:00+00:00"},
+            filters={"user_id": "u1"},
+            infer=True,
+        )
+
+        assert spy.call_count == 1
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Duplicate — recommend consolidating into #4964

### What both PRs do
Both fix the same root cause: when `created_at` is supplied via metadata, the extraction prompt currently falls back to wall-clock `datetime.now()` for date anchoring, causing temporal mismatch for backfilled/imported memories. Both PRs intercept that value at the `_add_to_vector_store` call site and forward it to `generate_additive_extraction_prompt`.

### Why #4964 is the better long-term fix

**Correctness — wrong parameter in this PR.** `generate_additive_extraction_prompt` has two separate date slots:
- `current_date=` → rendered as `## Current Date` (represents today's wall-clock date)
- `timestamp=` → rendered as `## Observation Date` (when the message was observed)

This PR passes the metadata date as `current_date`. Inside `_resolve_dates` (`prompts.py:1007–1013`), `observation_date` falls through to `observation_date = current_date` when `timestamp` is not supplied — so **both** prompt sections end up showing the historical date and the LLM has no signal for today's actual date. For a memory from 2021 processed in 2026, the prompt reads `## Current Date: 2021-06-15`, which inverts the intent. The correct parameter is `timestamp=`, which sets only `## Observation Date` while leaving `current_date` to default to today.

PR #4964 uses `timestamp=` throughout, making the two sections semantically distinct as designed.

**Cross-package completeness.** PR #4964 also patches `mem0-ts/src/oss/src/memory/index.ts` (the TypeScript OSS `Memory` class has the identical gap). This PR is Python-only. Merging this first would leave the SDKs behaving differently on the same operation.

**Broader lookup cascade.** PR #4964 checks `metadata["timestamp"]` → `metadata["created_at"]` → per-message `timestamp`/`created_at` in priority order. This PR only reads `metadata["created_at"]`, so callers who supply `metadata["timestamp"]` get no benefit.

**Test coverage.** PR #4964 has five Python tests (including async path, priority cascade, and no-timestamp fallback) plus three TypeScript tests. This PR has one sync-only Python test.

**CI:** PR #4964 passes Python (3.10/3.11/3.12) + TypeScript (Node 20/22) + TypeScript integration tests — verified this run.

### Credit where due
The defensive `try/except (ValueError, TypeError)` around `fromisoformat` is good — it ensures a malformed `created_at` degrades gracefully back to the prior behavior. That specific guard is absent from PR #4964; it would be a valuable addition there. The PR correctly patches both the sync and async paths, and the red-green guardrail test is properly structured (fails on unmodified main, passes with fix).

### Evidence
- **This PR — CI:** Python SDK (3.10, 3.11, 3.12) all pass; CI Gate pass. Vercel fail: advisory deploy-perms, not a merge gate.
- **This PR — Targeted test:** `pytest tests/memory/test_main.py::TestAddToVectorStoreErrors::test_created_at_metadata_anchors_extraction_prompt_date` → 1 passed in 0.41s
- **This PR — Red-green proof:** test fails on unmodified `main.py` (`assert None == '2021-06-15'`), passes with fix. Test bites.
- **This PR — Lint:** ruff all checks passed.
- **#4964 — CI:** Python SDK (3.10/3.11/3.12) + TypeScript SDK (Node 20/22) + TS integration tests — all pass (verified this run).
- **#4964 — Diff read:** correct `timestamp=` parameter, Python + TypeScript both patched, 5 Python tests + 3 TS tests.
- **Correctness finding — skeptic:** CONFIRMED. `_resolve_dates` at `prompts.py:1007–1013`: `observation_date = current_date` when `timestamp` not supplied — both prompt sections inherit the historical date.
- **Could NOT verify:** live end-to-end extraction with a real LLM; verified by code tracing only.

### Assessment
Recommendation: **close this in favour of #4964** (squash-merge #4964). Est. human time: ~3 min. Thanks for the salvage work here — the defensive fallback and the red-green test were well done; please reopen if #4964 falls through or consider contributing the `ValueError`/`TypeError` guard and the edge-case test for malformed `created_at` directly to #4964.

For tracking on our side, mind opening a quick issue for this bug and linking it to #4964? Makes it easier to prioritize and reference.

Bartok9:
Thanks for the thorough cross-reference, @kartik-mem0 — your analysis is correct and I agree #4964 is the better long-term fix. The `current_date=` vs `timestamp=` distinction is the deciding point: passing the historical date as `current_date` means `_resolve_dates` (`prompts.py:1007–1013`) lets `observation_date` fall through to it, so both prompt sections inherit the historical date and the LLM loses today's actual date. `timestamp=` is the right slot, and #4964's metadata cascade + TypeScript parity + broader test coverage close gaps this PR doesn't.

Tracking issue opened as requested: #5894 (linked to #4964).

I'll close this in favor of #4964. The `try/except (ValueError, TypeError)` guard around `fromisoformat` and the malformed-`created_at` edge case are the parts worth keeping — happy to contribute those directly to #4964 so a bad metadata date degrades gracefully there too.

PR #5883, opened 2026-06-26
fix(core): port delete_all entity-store hardening from async to sync Memory (salvage of #5697)

Bartok9:
Salvage of #5697 by @rafid001 — rebased onto current main with guardrail tests.

## Summary

Port the `delete_all` entity-store hardening that already exists on `AsyncMemory` to the synchronous `Memory.delete_all`.

## Root Cause

**Symptom** — `Memory.delete_all` (sync) deletes memories one-by-one and lets the first failure abort the whole operation, leaving the store half-cleared. It also runs per-memory entity cleanup, which read-modify-writes the same entity rows' `linked_memory_ids` repeatedly — a race / quadratic-write pattern the async path was already fixed to avoid.

**Root cause** — `AsyncMemory.delete_all` was hardened: it calls `_delete_memory(..., skip_entity_cleanup=True)`, collects per-delete exceptions via `asyncio.gather(return_exceptions=True)`, then clears the entity store in one pass with `_bulk_clear_entity_store(filters)`. The sync `Memory.delete_all` never received the same treatment — it loops `self._delete_memory(memory.id)` with no error collection and no bulk clear, and sync `_delete_memory` lacked the `skip_entity_cleanup` parameter and there was no sync `_bulk_clear_entity_store`.

**Evidence** — Confirmed on current `main` (`0fbbb2f`): async `delete_all` has the bulk-clear + error-collection logic and `_bulk_clear_entity_store` (line ~2237); sync `_delete_memory` had no `skip_entity_cleanup` and sync `delete_all` did the naive loop. New guardrail tests fail on unmodified main.

**Fix + why this level** — Mirror the async design in the sync path: (1) add `skip_entity_cleanup=False` to sync `_delete_memory`; (2) add sync `_bulk_clear_entity_store(filters)`; (3) rewrite sync `delete_all` to catch per-memory errors, skip per-memory entity cleanup, bulk-clear once, and log the failure count. Fixing at the `delete_all`/entity-store layer (not the caller) keeps sync and async behavior consistent and removes the read-modify-write race at its source.

**Scope / risk** — Touches only sync `Memory` (`_delete_memory`, new `_bulk_clear_entity_store`, `delete_all`). `skip_entity_cleanup` defaults to `False`, so direct `_delete_memory`/`delete()` callers are unchanged — only `delete_all` opts into the bulk path. Async path untouched.

## Why it needed salvage
Original #5697 conflicted against current main after the entity-store refactor. Re-applied cleanly to mirror the now-current async implementation.

## Changes from original
- Rebased onto current `main` (single clean commit, credited to @rafid001).
- Guardrail tests: bulk-clear called exactly once + per-memory cleanup skipped; and delete_all continues past an individual delete error. **Both fail without the fix.**

```
$ python -m pytest tests/memory/test_main.py -q
...............................................                          [100%]
47 passed in 1.12s

# Without the fix:
FAILED ...TestSyncDeleteAllEntityHardening::test_delete_all_bulk_clears_entities_once
FAILED ...TestSyncDeleteAllEntityHardening::test_delete_all_continues_on_individual_delete_error
```
`ruff check` clean.

Credit: salvage of #5697 by @rafid001.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1902,8 +1902,14 @@ class Memory(MemoryBase):
                 break
             seen_batches.add(batch_ids)
             for memory in memories:
-                self._delete_memory(memory.id)
-            deleted_count += len(memories)
+                try:
+                    self._delete_memory(memory.id, skip_entity_cleanup=True)
+                    deleted_count += 1
+                except Exception as e:
+                    logger.warning("Failed to delete memory id=%s: %s", memory.id, e)
+
+        if getattr(self, "_entity_store", None) is not None:
+            self._bulk_clear_entity_store(filters)
 
         logger.info(f"Deleted {deleted_count} memories")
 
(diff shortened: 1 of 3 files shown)


Bartok9:
Rebuilt on current `main` (prior tip had lost merge-base after upstream history move) — tip `346e60c3`.

Still needed after #6636: pagination landed, but sync `delete_all` still did per-memory entity RMW and aborted on the first delete failure. This ports the async hardening:

- `skip_entity_cleanup=True` on each sync `_delete_memory` during `delete_all`
- sync `_bulk_clear_entity_store` once after pages drain
- per-id delete failures logged instead of aborting mid-loop

Guardrail test updated for the post-#6636 pagination loop (`list` side_effect empties after one batch). Salvage credit remains with @rafid001 (#5697).

PR #5890, opened 2026-06-26
feat(ts-sdk): add AWS Bedrock LLM provider (closes #5765)

### LINKED ISSUE #5765: feat(ts-sdk): add AWS Bedrock LLM provider
## Summary

The Python SDK supports **AWS Bedrock** as an LLM provider, but the TypeScript OSS SDK (`mem0ai/oss`) does not. Add it to bring the TS SDK to parity.

| | |
|---|---|
| Python reference | `mem0/llms/aws_bedrock.py` |
| Registered in (Python) | `mem0/utils/factory.py` (LLMFactory) |
| Target file (TypeScript) | `mem0-ts/src/oss/src/llms/aws_bedrock.ts` |
| Suggested implementation | Use `@aws-sdk/client-bedrock-runtime` (`BedrockRuntimeClient` + `InvokeModelCommand` / `ConverseCommand`). |

## Requirements

- [ ] Implement `AWSBedrockLLM` in `mem0-ts/src/oss/src/llms/aws_bedrock.ts`, extending `LLM` (`mem0-ts/src/oss/src/llms/base.ts`) and mirroring the Python provider's behavior (generateResponse / generateChat).
- [ ] Register the `"aws_bedrock"` provider in `mem0-ts/src/oss/src/utils/factory.ts` (LLMFactory).
- [ ] Add `@aws-sdk/client-bedrock-runtime` to `mem0-ts/package.json` (optional/peer dependency, lazy-imported like other providers).

## Reference pattern

Mirror an existing TS provider: `llms/deepseek.ts` (OpenAI-compatible) or `llms/openai.ts`.

## Notes

Message formatting differs per model family (Anthropic, Titan, Llama, etc.) — mirror the per-model handling in `mem0/llms/aws_bedrock.py`. Auth via the standard AWS credential chain.

---

Bartok9:
## Summary

- Adds an **AWS Bedrock** LLM provider to the TypeScript OSS SDK, bringing it to parity with the Python SDK.
- Users can now configure `provider: "aws_bedrock"` from TS to drive Bedrock-hosted models (Anthropic, Amazon Nova, Meta Llama, Mistral, Cohere, etc.).

## Motivation

Closes #5765.

The Python SDK supports AWS Bedrock (`mem0/llms/aws_bedrock.py`, registered in `LLMFactory`), but the TypeScript OSS SDK (`mem0ai/oss`) had no Bedrock LLM provider. Part of the TS ↔ Python provider-parity effort (one provider per issue).

## Changes

| File | Change |
|------|--------|
| `mem0-ts/src/oss/src/llms/aws_bedrock.ts` | New `AWSBedrockLLM` class + `extractProvider` helper |
| `mem0-ts/src/oss/src/utils/factory.ts` | Register `"aws_bedrock"` in `LLMFactory` |
| `mem0-ts/src/oss/src/index.ts` | Export the provider |
| `mem0-ts/tsup.config.ts` | Mark `@aws-sdk/client-bedrock-runtime` external (lazy/optional) |
| `mem0-ts/package.json` + `pnpm-lock.yaml` | Add `@aws-sdk/client-bedrock-runtime` as an **optional peer dependency** |
| `mem0-ts/src/oss/src/tests/aws_bedrock.test.ts` | 10 unit tests |

### Implementation notes

- Implements the `LLM` interface (`generateResponse` / `generateChat`), mirroring the reference TS providers (`deepseek.ts` / `openai.ts`).
- **Uses the Bedrock Converse API** (`ConverseCommand`) for all model families. Converse gives a single, uniform message/tool interface across Anthropic / Amazon (Nova) / Meta / Mistral / Cohere, so one code path serves them all — this is the modern equivalent of the per-family `invoke_model` branches the Python provider keeps for legacy reasons.
- **System-prompt handling:** system messages are lifted into the top-level Converse `system` block (Converse passes system prompts separately from the conversation), matching the Python `_format_messages_anthropic` / MiniMax handling.
- **`topP` asymmetry preserved:** Anthropic and MiniMax reasoning models reject requests that carry both `temperature` and `topP`, so `topP` is omitted for those families (mirrors the Python `_build_inference_config`).
- **Tool calling:** OpenAI-style tools are converted to the Converse `toolConfig` shape, and `toolUse` response blocks are parsed back into `toolCalls` on `LLMResponse`.
- The `@aws-sdk/client-bedrock-runtime` SDK is **lazily `require`d** so the package stays optional (declared under `peerDependenciesMeta` as `optional: true`). Credentials resolve via the standard AWS chain (env / shared config / role) unless `awsAccessKeyId` + `awsSecretAccessKey` are passed in config.
- The runtime client is **injectable** via `config.client` for testing.

## Verification

- `pnpm run test:unit` — **813 passed** (56 suites), including the 10 new Bedrock tests.
- `pnpm run build` — prettier check + tsup (CJS/ESM) + dts build all clean.
- `pnpm install --frozen-lockfile` — passes (lockfile regenerated for the new optional peer dep).
- Package exports verified: `require('./dist/oss/index.js').AWSBedrockLLM` resolves to a constructor.

## Real behavior proof

The tests mock the AWS SDK module and inject a fake client that captures the exact `ConverseCommand` input and returns scripted responses — asserting request shaping (system block, inferenceConfig, toolConfig) and response parsing (text + toolUse) without live AWS:

```
PASS src/oss/src/tests/aws_bedrock.test.ts
  extractProvider
    ✓ detects the model family from the Bedrock model id
    ✓ throws on an unknown provider
  AWSBedrockLLM
    ✓ returns assistant text from a Converse response
    ✓ lifts system messages into the top-level system block
    ✓ omits topP for anthropic models even when configured
    ✓ includes topP for non-anthropic models
    ✓ converts OpenAI-style tools to Converse toolConfig and parses toolUse
    ✓ never sends an empty messages array
    ✓ wraps SDK errors with a provider-tagged message
    ✓ generateChat returns text content with assistant role

Tests:       10 passed, 10 total
```

## Scope / not covered

- Streaming responses (`ConverseStreamCommand`) are not implemented — `generateResponse`/`generateChat` are request/response, matching the other TS LLM providers and how the memory pipeline consumes them. Can be a follow-up.
- No live-Bedrock integration test (would require AWS credentials in CI); covered by mocked-SDK unit tests instead.
- Legacy `invoke_model`-only models that lack Converse support are out of scope; Converse covers the current Bedrock model families.


YinkaMetrics (commented):
I took a pass through this because it closes #5765. The overall direction looks right, but I think there are a few things to tighten before merge:

1. **Node 18 compatibility:** `package.json` still declares `engines.node >=18`, but the lockfile resolves `@aws-sdk/client-bedrock-runtime` to `3.1075.0`, whose package metadata requires Node `>=20.0.0`. Either the SDK engine needs to move to Node 20, or the Bedrock peer range should be constrained to the latest Node-18-compatible AWS SDK line. I found `3.967.0` still advertises Node `>=18.0.0`, while later releases move to Node 20.

2. **Config typing/schema:** the provider accepts fields like `awsRegion`, `awsAccessKeyId`, `awsSecretAccessKey`, and `awsSessionToken`, but those fields are not added to the exported `LLMConfig` / `MemoryConfigSchema`. Users configuring Bedrock through the normal `Memory` config path may not get typed support, and schema validation/normalization could drop or reject those fields depending on the path.

3. **Legacy model path:** this implementation routes everything through `ConverseCommand`. The issue asks to mirror the Python provider, including `InvokeModelCommand` for legacy/model-family-specific payloads where needed. It would be worth either adding that fallback or documenting/test-proving that every supported provider family here works via Converse.

4. **Docs:** a short TS OSS README example would help users know they need to install the optional Bedrock Runtime peer dependency and rely on the standard AWS credential chain.

The `tsup` external addition is good and important; without it the peer dependency gets caught by the externals test.

Bartok9:
Thanks for the careful review @YinkaMetrics — addressed in c7bdb975:

**1. Node 18 compatibility** ✅ You're right, the lockfile had floated to `3.1075.0` (Node ≥20). Constrained the peer range to `>=3.0.0 <3.968.0` and pinned the resolution to `3.967.0` (the last line advertising Node `>=18.0.0`) via a `pnpm.overrides` entry, so `engines.node: >=18` stays honest. Lockfile regenerated — `pnpm install` now resolves `3.967.0`.

**2. Config typing/schema** ✅ Moved `awsRegion` / `awsAccessKeyId` / `awsSecretAccessKey` / `awsSessionToken` (plus the injectable `client`) onto the shared `LLMConfig`, and the config manager now passes them through the normalized `llm.config` path (snake_case tolerated, matching the other providers). `aws_bedrock.ts` reuses `LLMConfig` instead of a divergent local interface, so users get typed Bedrock support through the standard `Memory` config.

**3. Legacy model path** — Kept the Converse-only approach but made the scope explicit in code comments and the new README note. Converse covers the current Bedrock model families through one uniform interface; legacy `InvokeModel`-only models are documented as out of scope and can be a follow-up if a concrete family needs it. Happy to add an `InvokeModelCommand` fallback if you'd prefer it in this PR.

**4. Docs** ✅ Added a TS OSS README section showing the optional `@aws-sdk/client-bedrock-runtime` install, the standard AWS credential chain, and the Converse/streaming scope.

Verification after the changes: `pnpm run test:unit` → **813 passed** (56 suites), `pnpm run build` (prettier + tsup CJS/ESM/dts) clean, `pnpm install` resolves the Node-18 line.

YinkaMetrics:
Thanks for the quick updates. The Node 18 pinning, shared config typing, and docs additions look good to me.

I think keeping Converse-only is reasonable if the README/code comments clearly scope legacy `InvokeModel` models as out of scope for this PR. A follow-up issue for `InvokeModelCommand` support would be cleaner than expanding this PR further.

Bartok9:
Agreed on keeping this Converse-only and tracking `InvokeModelCommand` as a separate follow-up — I'll open that issue. Also just rebased onto current main (215c6a3): resolved the `package.json`/`pnpm-lock.yaml` conflicts by keeping both the new `undici@<6.27.0` override from main and the bedrock pin/optional-peer entries. Branch is green and MERGEABLE again. Thanks for the review!

kartik-mem0:
## Changes Requested

### Strengths
- `extractProvider`/`PROVIDERS` in `aws_bedrock.ts` are a faithful port of `mem0/llms/aws_bedrock.py`'s provider-detection logic, including the anthropic/minimax `topP` omission
- The optional-peer-dep pattern (lazy `require("@aws-sdk/client-bedrock-runtime")` inside a guarded try/catch with an actionable install-hint error) matches the existing `natural`/`compromise` idiom already in `utils/lemmatization.ts` and `utils/entity_extraction.ts` — this is a real established pattern here, not a one-off shortcut
- `aws_bedrock.test.ts`'s 10 tests exercise actual request/response shaping (message formatting, tool-call parsing, inference config building) against a fake client — more than a factory-registration stub
- Registration is clean and complete: `utils/factory.ts` (import + case), `index.ts` (export), `tsup.config.ts` (external) all follow the exact sibling-provider pattern
- Node 18 compat was taken seriously: `@aws-sdk/client-bedrock-runtime` pinned to `3.967.0` via `pnpm.overrides`, and the incidentally-bumped shared `@smithy/*` transitives still declare `engines.node >=18.0.0`

### Issues
#### Critical (must fix)
1. `mem0-ts/src/oss/src/config/manager.ts` (new AWS block right after the `temperature/topP/maxTokens` lines) + `mem0-ts/src/oss/src/types/index.ts` (new fields on `LLMConfig`) — `awsRegion`/`awsAccessKeyId`/`awsSecretAccessKey`/`awsSessionToken`/`client` are added to `LLMConfig` and merged into `ConfigManager.mergeConfig()`'s output, but `MemoryConfigSchema.llm.config` (still a plain `z.object({...})`, not `.passthrough()` like the sibling `vectorStore.config` schema a few lines up in the same file) was never taught about them. `mergeConfig()` ends with `return MemoryConfigSchema.parse(mergedConfig)`, and Zod's default parse behavior silently strips undeclared keys — so all 5 fields are gone before `Memory`'s constructor hands `this.config.llm.config` to `LLMFactory.create()`. Practical effect: the exact `new Memory({ llm: { provider: "aws_bedrock", config: { awsRegion, awsAccessKeyId, awsSecretAccessKey } } })` example in this PR's own README silently drops those credentials and falls back to the ambient AWS credential chain instead of failing loudly. None of the 10 new tests catch this — they all construct `AWSBedrockLLM` directly and never go through `ConfigManager`. Fix: add the 5 fields to the `llm.config` zod schema (or `.passthrough()` it, like `vectorStore.config`), then add one test that goes through `Memory`/`ConfigManager`, not direct construction.
2. `mem0-ts/package.json` (the `pnpm.overrides`/`peerDependenciesMeta` hunk) — adds a *second* top-level `"peerDependenciesMeta"` key instead of extending the existing one (which holds `mysql2`'s `optional: true`, left untouched as context a few lines earlier in the same diff). Two same-named top-level keys in one JSON object: `JSON.parse` — which is exactly what `pnpm`/`npm`/Node's `require()` use to read `package.json` — keeps only the last one. I verified this with a literal reconstruction: both `node -e "JSON.parse(...)"` and `python3 -c "json.load(...)"` resolve the merged object down to `{"@aws-sdk/client-bedrock-runtime": {"optional": true}}` only, with `mysql2`'s entry gone entirely. That's an install-time regression for every consumer who doesn't have `mysql2` installed — unrelated to Bedrock, but introduced by this diff. Fix: merge both into one `peerDependenciesMeta` object with both keys.

#### Important (should fix)
none

#### Minor (optional)
1. `aws_bedrock.ts` constructor — no default region fallback (Python's `aws_region` defaults to `os.getenv("AWS_REGION", "us-west-2")`; here `region` is simply omitted from `clientConfig` if unset, deferring entirely to the AWS SDK's own resolution chain). Not wrong, just a different failure mode than Python's when nothing sets a region anywhere.

### Evidence
- **CI:** `gh pr checks 5890` — all real checks green (TS SDK build on Node 20/22, integration x2, changelog_check, check_changes, CI Gate, license/cla). Only `Vercel` (preview-deploy) is red — advisory, out of scope for this review.
- **Targeted tests / red-green:** not re-run this session (read-only review, no worktree/checkout per scope). Both findings above are traced end-to-end against the actual source (`manager.ts`'s merge output -> `types/index.ts`'s zod schema -> `memory/index.ts`'s `LLMFactory.create()` call), and the JSON duplicate-key behavior is independently reproduced with `node`/`python3` rather than asserted from reading the diff alone. Confirmed via grep that no test outside `aws_bedrock.test.ts` references `aws_bedrock`/`AWSBedrock`, and there is no `ConfigManager` test file anywhere in the suite.
- **Could NOT verify:** actual `pnpm install`/`pnpm test` output against the real PR branch; whether Bedrock's Converse API covers every model family Python's hybrid Converse + `invoke_model` dispatch handles for this provider (took the PR's own code comments on this at face value rather than testing live against AWS).

### Assessment
Ready to merge: **With fixes** — both Critical items are config/build plumbing, not the core Converse request/response logic (which looks solid and is genuinely well tested). Please fix those two and I'll re-review. Once they're in, this is a clean, well-scoped provider addition — adopting AWS Bedrock as a new backend is ultimately a maintainer call, but the code itself would clear that bar.

Bartok9:
Thanks @kartik-mem0 — both Critical items fixed in 059c3ca6:

**1. Credentials silently stripped by Zod** ✅ You traced it exactly right — `MemoryConfigSchema.llm.config` was a plain `z.object({...})`, so `mergeConfig()`'s trailing `MemoryConfigSchema.parse()` dropped the 5 AWS fields before `LLMFactory.create()` ever saw them. Added `awsRegion`/`awsAccessKeyId`/`awsSecretAccessKey`/`awsSessionToken`/`client` to the schema and `.passthrough()`-ed it to match the sibling `vectorStore.config` a few lines up. Added two regression tests: one at the `mergeConfig` layer (asserts all four creds survive `.parse()`, plus snake_case normalization) and one that goes through the full `Memory`/`ConfigManager` stack asserting `LLMFactory.create('aws_bedrock', ...)` receives the credentials — the exact path the old direct-construction tests never exercised.

**2. Duplicate `peerDependenciesMeta` key** ✅ Confirmed and fixed — merged into a single object holding both `mysql2` and `@aws-sdk/client-bedrock-runtime` optional entries, so `JSON.parse` no longer discards mysql2's. Verified with `node -e`.

Verification: `npx jest config-manager` → **32 passed** (including the 3 new Bedrock tests). The pre-existing `fastembed` dts error in local `pnpm run build` is the uninstalled optional peer, unrelated to this change — the TS SDK build is green on CI (Node 20/22).

Ready for re-review.

kartik-mem0:
## Needs Human Review

**Why:** this adds a **new backend** to the TS SDK (`mem0-ts/src/oss/src/llms/aws_bedrock.ts`). The code itself is merge-ready; the open question is the adoption call, and that's a maintainer's to make. No code blockers found.

Re-reviewed at `9b92260a` (the previous LGTM-eligible read at `84546e6b` is void — new commits). Both Criticals from the earlier round are fixed: Node 18 compat via the peer bound, and typed `LLMConfig` fields instead of `any`.

### Pay attention to
- Adoption: do we want Bedrock as a first-class TS OSS LLM provider? Issue #5765 says yes.
- `mem0-ts/package.json` — peer bound `>=3.0.0 <3.968.0` is a deliberate Node-18 ceiling (`3.968.0` raises `engines.node` to `>=20`), not a stale pin. It needs revisiting only when `mem0-ts` drops Node 18.
- `mem0-ts/src/oss/src/llms/aws_bedrock.ts:84` — lazy `require()` for the optional peer. Matches `vector_stores/milvus.ts:71`, `utils/lemmatization.ts:180`, `utils/entity_extraction.ts:338`.

### Findings

**Critical:** none.

**Important:** none. Four candidates were raised and all four refuted against the code/registry, so I'm not passing them on. Briefly, for the record: the peer bound is correct (it lands exactly on the AWS SDK's Node-18 drop); Converse-only is fine (AWS recommends it, and Llama 3.1 405B is Converse-*only*); the "missing `us-west-2` default" is a feature, not drift (TS omits `region` entirely so the AWS credential/region chain resolves, whereas Python's hard-coded `us-west-2` silently bills the wrong region); and the Nova `maxTokens` default matches — `get_model_config()` always emits `max_tokens: 2000`, so Python's `_default_max_tokens_for_converse()` is dead code and TS's `maxTokens ?? 2000` is identical.

**Minor (optional):**
1. `mem0-ts/src/oss/src/types/index.ts:165` — `.passthrough()` on `llm.config` is a no-op. Extra keys survive the parse in `Memory.fromConfig()` but `ConfigManager.mergeConfig()` hand-enumerates the `llm.config` object (`config/manager.ts:135-164`, no `...userConf` spread, unlike `vectorStore.config` at `:83`) and re-parses at `:197`. Zero runtime risk, but it implies future Bedrock fields won't need a `manager.ts` edit — they will. Drop it, or spread `userConf`.
2. `mem0-ts/src/oss/src/tests/aws_bedrock.test.ts` — every test injects a fake `client`, so the constructor's credential wiring (`aws_bedrock.ts:95-113`) is never exercised, the missing-SDK throw is dead under the virtual mock, and `modelId` is never asserted. Still well above the repo norm: openai/anthropic/groq/ollama/mistral have no dedicated provider tests at all.
3. `mem0-ts/src/oss/src/types/index.ts:65` — the `for DI/testing` comment on `client` overpromises: `require()` runs at `:84`, before the `config.client` check at `:111`, and `ConverseCommand` has no injection seam.

### Docs
The TS provider shipped without TS docs, so I pushed `4d4a0122` to this branch (thanks for leaving "allow edits from maintainers" on): `aws_bedrock.mdx` now has a `<CodeGroup>` with Python + TypeScript tabs plus notes on the optional peer dep and the Converse-only/no-streaming limits, and `components/llms/overview.mdx` + `open-source/configuration.mdx` now list `aws_bedrock` under TypeScript. `docs llms.txt / check-llms-txt` is green on it.

### Evidence
- **CI @ `9b92260a`:** all real checks green — `build_ts_sdk` (20, 22), `integration_ts_sdk` (20, 22), `check_changes`, `changelog_check`, `docs llms.txt`. The red `Vercel` check is the fork deploy-preview (no deploy perms from forks); ignored, not merge-blocking.
- **Targeted tests:** `npx jest src/oss/src/tests/aws_bedrock.test.ts` → 10 passed, 10 total. `npx jest src/oss/tests/config-manager.test.ts` → 32 passed, 32 total.
- **Red-green proof:** reverted only `types/index.ts` + `config/manager.ts` to `origin/main`, kept the tests → exactly the 3 new tests fail (`Expected "us-east-1", Received undefined`; `Expected "eu-west-1", Received undefined`; the `LLMFactory` `objectContaining` assertion), nothing else breaks. Restore → 32/32. The tests reproduce the real failure (credentials dropped before `LLMFactory.create()`), not just the mechanism, and all three changed paths are covered.
- **Lint/types:** `prettier --check` clean on all 5 TS files; `tsc --noEmit` → 28 errors, **zero** in this PR's files (all pre-existing, in `src/client/tests/`, `src/community/.../langchain/mem0.ts`, and main's `weaviate.ts`/`chroma.ts`).
- **Rebase:** merge base is now main's tip — `behind_by: 0`, `mergeable: true`. Verified `9b92260a` only regenerates `pnpm-lock.yaml` (+1006/-90) and that the rebase preserved main's `huggingfaceBaseUrl` while keeping this PR's hunks intact. `@aws-sdk/client-bedrock-runtime` remains peer-only in `package.json`; its appearance under the lockfile's `.` importer is `autoInstallPeers: true`, same as `@anthropic-ai/sdk`.
- **Linked issue:** #5765 — PR addresses the request directly.
- **Cross-package check:** compared against `mem0/llms/aws_bedrock.py` + `mem0/configs/llms/aws_bedrock.py`. No behavioral drift that matters; the one intentional difference (region resolution) favors the TS side.
- **Could NOT verify:** live Bedrock API calls (no AWS credentials here). The Converse request shape, `inferenceConfig`, `toolConfig`, and top-level `system` blocks were checked against AWS's Converse API reference and by code tracing only. `pnpm install --frozen-lockfile` was not run locally (CI runs exactly that, and it's green).

### Assessment
Ready to merge: **No — maintainer judgment required.** Est. human time: ~5 min.
Quality: **8/10.** Correct, idiomatic, and better tested than any of its sibling TS LLM providers. Deductions for the missing TS docs (now filled in), the dead `.passthrough()` with a comment implying otherwise, and the untested constructor path. If we're taking Bedrock into the TS SDK, this is the implementation to take — squash-merge once you've made that call.

Bartok9:
### Re-review ping (post-rebase)

Rebased onto current `main` after the PR went `CONFLICTING` when main advanced. Conflicts were additive and resolved cleanly (factory / tsup / package.json ×3 / lockfile). Preserved the earlier Critical fixes (single merged `peerDependenciesMeta`, AWS SDK Node-18 pin at `3.967.0`).

**Current status on head `9b92260`:**
- Mergeable (no conflicts)
- CI Gate green: TypeScript SDK build + integration on Node 20 & 22, changelog, llms.txt, CLA
- Vercel shows “Authorization required to deploy” — fork deploy auth gate, not a code failure

@YinkaMetrics your June 30 review notes (Node 18 / SDK pin, config typing, Converse vs InvokeModel, docs) should still be the main review surface — happy to address any remaining items if something still blocks merge after this rebase.

Maintainer re-review / merge when convenient would be appreciated. Thanks!

Bartok9:
Thanks @kartik-mem0 — agreed the adoption call is yours (and #5765 asks for exactly this). No code blockers, so just cleaning up the one actionable minor while it waits:

**Minor #1 (dead `.passthrough()`)** — fixed in `d92dacc9`. `mergeConfig()` hand-enumerated the `llm.config` fields, so `.passthrough()` on `llm.config` was indeed a no-op and future Bedrock fields would've silently needed a `manager.ts` edit. Added `...userConf` spread first (matching `vectorStore.config` at `:83`) so extra fields now actually flow through — `.passthrough()` is meaningful again. 42/42 config-manager + aws_bedrock tests green, prettier clean.

Minors #2/#3 (untested constructor path, overpromising `client` comment) I'll leave as-is unless you'd like them addressed — happy to trim the comment or add a real-client constructor test if you want it in this PR rather than a follow-up. Ready to squash-merge whenever you make the adoption call.

kartik-mem0 (approved):

Bartok9:
Thanks for the approval @kartik-mem0 🙏 Ready to squash-merge whenever the Bedrock adoption call is confirmed — head is mergeable and green (build + integration on Node 20/22, changelog, llms.txt, CLA). Happy to address minors #2/#3 (constructor-path test, `client` comment) in this PR if you'd prefer them landed before merge rather than as a follow-up.

PR #5892, opened 2026-06-26
[codex] Add OpenMemory stdio MCP workflow

psyche314:
## Summary

- Add a stdio MCP entrypoint for local OpenMemory clients such as Codex.
- Autostart the local OpenMemory API and UI when the stdio MCP server starts, with lock files so concurrent MCP processes reuse one API/UI instance.
- Keep autostarted services running after an individual MCP process exits, so the dashboard remains available in the browser.
- Return JSON-RPC errors for invalid stdio input instead of exiting the MCP server.
- Add an icon fallback for unknown OpenMemory source apps in the dashboard stats panel.
- Document Codex stdio MCP setup and autostart environment variables in `openmemory/README.md`.

## Why

Codex and similar clients launch MCP servers over stdio. Without a local stdio entrypoint and autostart behavior, users need to manually start the OpenMemory API/UI before Codex can use local memories or before the dashboard is available.

## Validation

- `python -m py_compile openmemory/api/mcp_stdio.py openmemory/api/run_openmemory.py`
- `python -m pytest openmemory/api/tests/test_mcp_server.py`
- Manually smoke-tested concurrent stdio startup with 3 MCP processes; one API listener and one UI listener were created.
- Manually verified the dashboard loads memories from the local API.

### DIFF
--- a/openmemory/README.md
+++ b/openmemory/README.md
@@ -140,6 +140,49 @@ npx @openmemory/install local http://localhost:8765/mcp/<client-name>/sse/<user-
 
 Replace `<client-name>` with the desired client name and `<user-id>` with the value specified in your environment variables.
 
+### Codex stdio MCP setup
+
+OpenMemory also provides a stdio MCP entrypoint for local clients that launch MCP servers as child processes, such as Codex.
+
+Example Codex configuration:
+
+```toml
+[mcp_servers.mem0]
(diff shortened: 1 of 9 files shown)


kartik-mem0:
openmemory is deprecated and we won't be taking any feature request.

thank you!

PR #5908, opened 2026-06-26
fix(memory): lazy-import langchain-core in async procedural memory

### LINKED ISSUE #5907: bug: AsyncMemory procedural memory requires optional langchain-core even when no custom LLM is used
## Description

## Impact

```python
import asyncio
from mem0 import AsyncMemory

m = AsyncMemory()
```

## Location

AnnaSuSu:
## Linked Issue

Closes #5907

## Description

`AsyncMemory._create_procedural_memory()` imported `langchain_core` at function entry, but `convert_to_messages` is only used when a custom LangChain `llm` is supplied. Because `langchain-core` is an optional `extras` dependency, the default path (`llm=None`) raised `ImportError` on a plain `pip install mem0ai`, while the sync `Memory._create_procedural_memory()` — which has no langchain dependency — worked for the same call.

This moves the import inside the `if llm is not None` branch, so the default async procedural-memory path matches the sync class and no longer requires the optional package. langchain-core is still imported (and its error surfaced) when a custom LangChain `llm` is actually passed.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `test_async_procedural_memory_does_not_require_langchain_when_llm_none`: it blocks `langchain_core` in `sys.modules` and asserts the `llm=None` path still generates and stores the memory. Fails before the fix, passes after. Existing `test_async_procedural_memory_langchain_strips_code_blocks` (custom-llm path) still passes; full `tests/test_memory.py` passes (59 tests).

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3556,16 +3556,6 @@ class AsyncMemory(MemoryBase):
             llm (llm, optional): LLM to use for the procedural memory creation. Defaults to None.
             prompt (str, optional): Prompt to use for the procedural memory creation. Defaults to None.
         """
-        try:
-            from langchain_core.messages.utils import (
-                convert_to_messages,  # type: ignore
-            )
-        except Exception:
-            logger.error(
-                "Import error while loading langchain-core. Please install 'langchain-core' to use procedural memory."
-            )
-            raise
-
         logger.info("Creating procedural memory")
 
         parsed_messages = [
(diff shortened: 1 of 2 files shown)


AnnaSuSu:
bumping this one. small change, just lazy-imports langchain-core so it isn't required at import time on the async procedural path. CI is green and it's mergeable. anything holding up review?

AnnaSuSu:
Closing this — the same fix landed upstream in #6209, which moves the `langchain-core` import into the `llm is not None` branch so the default async procedural-memory path no longer requires the optional package. Nothing left to do here. Thanks!

PR #5932, opened 2026-06-27
fix(cli): guard memory formatters against null id/memory fields

### LINKED ISSUE #5931: bug(cli): list memory formatters crash on null `id`/`memory` fields from API
### Component

### Description

Two concrete crashes:

### Steps to Reproduce

```python

c = Console()

```

### Expected Behavior

### Actual Behavior

### Environment

- mem0-cli: current `main`
- Python: 3.10

abhay-codes07:
## Linked Issue

Closes #5931

## Description

While poking at the CLI output code I noticed the memory list formatters assume `id` and `memory` are always present, non-null strings. They aren't always — the API can hand back a record with an explicit `null` in one of those fields, and when it does the command crashes instead of just printing a blank cell.

The root cause is a classic `dict.get` gotcha: `mem.get("id", "")` only falls back to `""` when the key is **missing**. If the key is there but the value is `null`, you get `None` back, and then:

- `format_memories_text` does `mem.get("id", "")[:8]` → `None[:8]` → `TypeError: 'NoneType' object is not subscriptable`
- `format_memories_table` does `len(mem.get("memory", ...))` → `len(None)` → `TypeError: object of type 'NoneType' has no len()`

So a single odd record is enough to take down `mem0 search` / `mem0 list`.

The nice part is the codebase already has the answer: `format_add_result` in the same file guards every field with the `... or ""` pattern. The list/single formatters just never got the same treatment. This PR brings them in line — `mem.get("memory") or mem.get("text") or ""` and `(mem.get("id") or "")` — so a null or missing field renders as empty rather than blowing up. I applied the same guard to `format_single_memory` too, so it doesn't print a literal "None" for an empty body.

No behavior change for well-formed responses; this only affects the null/missing edge.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] No tests needed

Added a `TestNullFieldRobustness` class covering null and fully-empty records across all three formatters. The new tests fail on `main` (the two `TypeError`s above) and pass with the fix. Full CLI suite stays green and `ruff check` is clean.

## Checklist

### DIFF
--- a/cli/python/src/mem0_cli/output.py
+++ b/cli/python/src/mem0_cli/output.py
@@ -20,8 +20,8 @@ def format_memories_text(console: Console, memories: list[dict], title: str = "m
     console.print(f"\n[{BRAND_COLOR}]Found {count} {title}:[/]\n")
 
     for i, mem in enumerate(memories, 1):
-        memory_text = mem.get("memory", mem.get("text", ""))
-        mem_id = mem.get("id", "")[:8]
+        memory_text = mem.get("memory") or mem.get("text") or ""
+        mem_id = (mem.get("id") or "")[:8]
         score = mem.get("score")
         created = _format_date(mem.get("created_at"))
         category = mem.get("categories", [None])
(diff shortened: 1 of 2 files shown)


abhay-codes07:
Hi @kartik-mem0 a small, self-contained CLI bug fix here (null `id`/`memory` from the API crashes `mem0 search`/`list`). It mirrors the existing `or ""` guard already used by `format_add_result`, with regression tests that fail without the change. Would appreciate a quick look whenever you get a chance

kartik-mem0:
closing this pr in favour #5957 of this.

PR #5934, opened 2026-06-27
fix(cli): return False instead of crashing on invalid int in `config set`

### LINKED ISSUE #5933: bug(cli): `config set` on an int field crashes with uncaught ValueError on non-numeric input
### Component

### Description

```python
```

### Steps to Reproduce

```bash
mem0 config set version abc
```

```python
```

### Expected Behavior

### Actual Behavior

### Environment

- mem0-cli: current `main`
- Python: 3.10
- OS: platform-independent

abhay-codes07:
## Linked Issue

Closes #5933

## Description

`set_nested_value()` promises a bool return ("Returns True on success"), and the rest of the code leans on that — `test_set_nonexistent_key` expects `False` for a bad key, and `cmd_config_set` decides success/failure from the return value. The int branch broke that promise:

```python
    value = int(value)   # raises ValueError on non-numeric input, uncaught
```

`version` is an int field on `Mem0Config`, and `mem0 config set <key> <value>` passes the key through unchanged, so a simple typo like `mem0 config set version abc` takes the CLI down with a raw `ValueError` traceback instead of a clean "couldn't set that" result.

The fix wraps the coercion and returns `False` on a bad value — the same signal already used for an unknown key — and leaves the existing value untouched. Bool fields are unaffected (they were already lenient), and valid integers still coerce as before.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] No tests needed

Added two tests: a non-numeric value returns `False` and doesn't mutate the field (this one raises without the fix), and a valid numeric string still coerces to `int`. Full config suite passes and `ruff check` is clean.

## Checklist

### DIFF
--- a/cli/python/src/mem0_cli/config.py
+++ b/cli/python/src/mem0_cli/config.py
@@ -231,11 +231,15 @@ def set_nested_value(config: Mem0Config, dotted_key: str, value: str) -> bool:
         return False
 
     current = getattr(obj, final_key)
-    # Type coercion
+    # Type coercion. A non-numeric value for an int field is a failed set, not a
+    # crash: honor the documented bool return instead of raising ValueError.
     if isinstance(current, bool):
         value = value.lower() in ("true", "1", "yes")  # type: ignore[assignment]
     elif isinstance(current, int):
-        value = int(value)  # type: ignore[assignment]
+        try:
+            value = int(value)  # type: ignore[assignment]
+        except ValueError:
+            return False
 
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small CLI fix with a regression test. Quick review when you can, thanks.

kartik-mem0:
closing this pr in favour of this #5956

PR #5936, opened 2026-06-27
fix(cli): keep every result in `entity delete`, not just the last

### LINKED ISSUE #5935: bug(cli): `entity delete` returns only the last entity's response for multi-entity deletes
### Component

### Description

```python
result: dict = {}
    result = self._request(
    )
return result
```

### Steps to Reproduce

```bash
```

### Expected Behavior

```json
```

### Actual Behavior

### Environment

- mem0-cli: current `main`
- Python: 3.10
- OS: platform-independent

abhay-codes07:
## Linked Issue

Closes #5935

## Description

`delete_entities()` walks every entity it's asked to delete, but it writes each API response into one `result` variable that gets overwritten on the next pass — so only the last entity's response makes it back to the caller:

```python
result: dict = {}
    result = self._request("DELETE", f"/v2/entities/{entity_type}/{entity_id}/", ...)
return result
```

That matters because `mem0 entity delete` lets you pass `--user-id`, `--agent-id`, `--app-id`, and `--run-id` together, and `cmd_entities_delete` prints this value straight out under `--output json`. Delete two entities and the JSON only shows one — anything scripting against that output silently loses the rest.

The fix collects each response into a dict keyed by entity type (`{"user": {...}, "agent": {...}}`). Nothing is dropped, the deletes themselves were always happening (one request each), and the `-> dict` return type stays intact. Single-entity deletes just return a one-key dict.

## Type of Change

- [ ] Documentation update

## Breaking Changes

The JSON shape of `mem0 entity delete --output json` changes from a single raw API response to a map keyed by entity type. This only affects the multi-entity case (which was lossy before) and the human/agent output modes are unaffected, so I've treated it as a bug fix rather than a breaking change — happy to adjust if you'd prefer to preserve the single-entity raw shape.

## Test Coverage

- [ ] No tests needed

Added `tests/test_platform_backend.py` covering multi-entity (fails without the fix), single-entity, and the no-entity `ValueError`. `_request` is patched, so no network calls. Full CLI suite passes and `ruff check` is clean.

## Checklist

### DIFF
--- a/cli/python/src/mem0_cli/backend/platform.py
+++ b/cli/python/src/mem0_cli/backend/platform.py
@@ -296,13 +296,15 @@ class PlatformBackend(Backend):
         entities = {t: v for t, v in type_map.items() if v}
         if not entities:
             raise ValueError("At least one entity ID is required for delete_entities.")
-        # Delete each provided entity via the v2 path-based endpoint
-        result: dict = {}
+        # Delete each provided entity via the v2 path-based endpoint. Key each
+        # response by entity type so a multi-entity delete (e.g. --user-id and
+        # --agent-id together) doesn't discard everything but the last result.
+        results: dict = {}
         for entity_type, entity_id in entities.items():
-            result = self._request(
+            results[entity_type] = self._request(
                 "DELETE", f"/v2/entities/{entity_type}/{entity_id}/", params={"source": "CLI"}
             )
-        return result
+        return results
 
     def ping(self, timeout: float | None = None) -> dict:
         """Call the ping endpoint and return the raw response.
--- a/cli/python/tests/test_platform_backend.py
+++ b/cli/python/tests/test_platform_backend.py
@@ -1,0 +1,46 @@
+"""Tests for the Platform backend (mem0 Platform API client)."""
+
+from __future__ import annotations
+
+from unittest.mock import patch
+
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@kartik-mem0 small CLI fix with tests. Noted a minor JSON-shape question in the description. Quick review when you can, thanks.

kartik-mem0 (approved):

kartik-mem0:
hey @abhay-codes07 thank you for contribution can you confirm if this issues are also in the ts cli and raise the issue for them? 

#5933, #5931 and the above one!

thanks!

abhay-codes07:
@kartik-mem0 thanks! Checked the Node/TS CLI for all three:

- entity delete (this one): yes, same bug in cli/node/src/backend/platform.ts, deleteEntities overwrites result in the loop. Raised #5969 and opened a fix in #5970 keyed by entity type, with tests.
- #5931 (null id/memory crash): not present there. The TS formatters use ?? (mem.id ?? "", mem.memory ?? mem.text ?? ""), so null/undefined is already handled.
- #5933 (config set int crash): not present either. setNestedValue only exposes string fields via KEY_MAP, and Number.parseInt returns NaN instead of throwing, so no crash.

So only the entity delete one needed a fix on the TS side.

PR #5942, opened 2026-06-28
fix(cli): return all entity-delete responses keyed by type

### LINKED ISSUE #5935: bug(cli): `entity delete` returns only the last entity's response for multi-entity deletes

ly-wang19:
## What Problem This Solves

Closes #5935.

`PlatformBackend.delete_entities()` in `cli/python/src/mem0_cli/backend/platform.py` looped over every entity to delete but reassigned a single `result` variable each pass, so it returned only the **last** entity's API response:

```python
result: dict = {}
    result = self._request(
    )
return result
```

`mem0 entity delete` accepts `--user-id`, `--agent-id`, `--app-id`, and `--run-id` simultaneously, and `cmd_entities_delete` prints this return value verbatim under `--output json`. So when more than one entity was deleted, every response except the last was silently dropped from the JSON output.

## Why This Change Was Made

Key each response by entity type, so multi-entity deletes preserve every response: `{"user": {...}, "agent": {...}}`. This:

- Keeps the `-> dict` return type (the existing type annotation stays valid)
- Surfaces every response to `--output json`, the only consumer that uses the return value
- Doesn't affect `--output agent` (always emits `{"deleted": True}` regardless) or default text output (always prints a fixed success message)
- Single-entity callers now see `{"user": {...}}` instead of `{...}` — the wrapping is visible but the JSON consumer (`format_json`) just prints it verbatim, and existing CLI tests check for `"message"` substring (which still matches the nested key)

## User Impact

- `mem0 entity delete --user-id alice --agent-id bot1 --output json` now returns both responses: `{"user": {...}, "agent": {...}}` instead of only the agent's
- Scripts parsing the JSON output can verify each individual deletion succeeded
- Single-entity deletes still work; output is now keyed by entity type

## Evidence

Three regression tests in `cli/python/tests/test_platform_backend.py`:

```python
def test_delete_entities_single_returns_keyed_response():
    backend = _make_backend_with_request_sequence([{"message": "Entity deleted"}])
    result = backend.delete_entities(user_id="alice")
    assert result == {"user": {"message": "Entity deleted"}}

def test_delete_entities_multi_preserves_every_response():
    """Regression for #5935: the previous implementation reassigned a single
    ``result`` variable each iteration and returned only the final response."""
    backend = _make_backend_with_request_sequence([
        {"message": "User deleted", "id": "alice"},
        {"message": "Agent deleted", "id": "bot1"},
        {"message": "App deleted", "id": "app-x"},
    ])
    result = backend.delete_entities(user_id="alice", agent_id="bot1", app_id="app-x")
    assert set(result.keys()) == {"user", "agent", "app"}
    assert result["user"]["id"] == "alice"
    assert result["agent"]["id"] == "bot1"
    assert result["app"]["id"] == "app-x"
    assert backend._request.call_count == 3

def test_delete_entities_no_entities_raises():
    ...
```

Verified the multi-entity test catches the bug — it FAILS against the pre-fix code:

```
E       AssertionError: assert {'id', 'message'} == {'agent', 'app', 'user'}
tests/test_platform_backend.py:50: AssertionError
========================= 1 failed in 0.15s ===============================
```

(`{'id', 'message'}` is the un-keyed last response; the test wanted three entity-type keys.)

With the fix, full test suites pass — 68 tests across `test_platform_backend.py`, `test_commands.py`, and `test_config.py`:

```
PYTHONPATH=src uv run pytest tests/test_platform_backend.py tests/test_commands.py
============================== 68 passed in 0.52s ==============================
```

### DIFF
--- a/cli/python/src/mem0_cli/backend/platform.py
+++ b/cli/python/src/mem0_cli/backend/platform.py
@@ -296,13 +296,15 @@ class PlatformBackend(Backend):
         entities = {t: v for t, v in type_map.items() if v}
         if not entities:
             raise ValueError("At least one entity ID is required for delete_entities.")
-        # Delete each provided entity via the v2 path-based endpoint
-        result: dict = {}
+        # Delete each provided entity via the v2 path-based endpoint.
+        # Key each response by entity type so multi-entity deletes don't
+        # silently drop all but the last response — see #5935.
+        results: dict[str, Any] = {}
         for entity_type, entity_id in entities.items():
-            result = self._request(
+            results[entity_type] = self._request(
                 "DELETE", f"/v2/entities/{entity_type}/{entity_id}/", params={"source": "CLI"}
             )
-        return result
+        return results
 
     def ping(self, timeout: float | None = None) -> dict:
         """Call the ping endpoint and return the raw response.
--- a/cli/python/tests/test_platform_backend.py
+++ b/cli/python/tests/test_platform_backend.py
@@ -1,0 +1,67 @@
+"""Regression tests for ``PlatformBackend`` multi-entity delete behavior.
+
+Covers #5935: ``delete_entities()`` used to reassign a single ``result``
+variable each pass, so only the last entity's API response was returned.
+The fix keys responses by entity type so multi-entity deletes preserve
+every response.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+from mem0_cli.backend.platform import PlatformBackend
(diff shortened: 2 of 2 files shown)


kartik-mem0:

closing this pr in favour of #5936

PR #5943, opened 2026-06-28
fix(cli): tolerate null id/memory fields in list and single formatters

### LINKED ISSUE #5931: bug(cli): list memory formatters crash on null `id`/`memory` fields from API

ly-wang19:
## What Problem This Solves

Closes #5931.

The list/single memory formatters in `cli/python/src/mem0_cli/output.py` assumed `id` and `memory` were always non-null strings. When the backend returned a record with an explicit JSON `null` for one of those fields, `dict.get(key, default)` returned `None` (the key *is* present, so the default never applied), and the subsequent slice/`len()` blew up:

`mem0 search` / `mem0 list` could hard-crash on otherwise valid API output instead of rendering a blank cell.

## Why This Change Was Made

`format_add_result` in the same file already guards against exactly this with the `r.get("id") or ""` pattern — the list/single formatters were simply missed. This PR applies the same defensive pattern to the three formatters that lacked it:

```python
# Before
memory_text = mem.get("memory", mem.get("text", ""))
mem_id = mem.get("id", "")[:8]

# After
memory_text = mem.get("memory") or mem.get("text") or ""
mem_id = (mem.get("id") or "")[:8]
```

The fix is uniform across `format_memories_text`, `format_memories_table`, and `format_single_memory`. Same pattern, same file, same author intent.

## User Impact

- `mem0 search` / `mem0 list` / `mem0 get` no longer crash on backend records that carry explicit JSON `null` for `id` / `memory` / `text`
- A null field renders as an empty cell in tables and as a blank line in text mode — consistent with `format_add_result`'s behavior
- Records with non-null fields render exactly as before

## Evidence

New `TestNullFields` class in `cli/python/tests/test_output.py` with 4 regression tests:

```python
NULL_RECORDS = [{
    "id": None, "memory": None, "text": None,
    "created_at": None, "categories": None, "score": None,
}]

def test_format_memories_text_handles_null_fields(self):
    console, _buf = _make_console()
    format_memories_text(console, self.NULL_RECORDS)  # no exception

def test_format_memories_table_handles_null_fields(self):
    console, _buf = _make_console()
    format_memories_table(console, self.NULL_RECORDS)  # no exception

def test_format_single_memory_handles_null_fields(self):
    console, _buf = _make_console()
    format_single_memory(console, self.NULL_RECORDS[0])  # no exception

def test_format_memories_table_renders_blank_id(self):
    """A null ``id`` renders as an empty cell, not a crash."""
    console, buf = _make_console()
    format_memories_table(console, self.NULL_RECORDS)
    output = buf.getvalue()
    assert "TypeError" not in output
```

Verified the tests catch the bug — 3 of 4 FAIL against the pre-fix code with the exact issue reproduction:

```
>           if len(memory_text) > 60:
E           TypeError: object of type 'NoneType' has no len()

src/mem0_cli/output.py:72: TypeError
========================= 3 failed, 1 passed in 0.09s =========================
```

With the fix, full output + commands suites pass — 91 tests:

```
PYTHONPATH=src uv run pytest tests/test_output.py tests/test_commands.py
============================== 91 passed in 0.36s ==============================
```

### DIFF
--- a/cli/python/src/mem0_cli/output.py
+++ b/cli/python/src/mem0_cli/output.py
@@ -20,11 +20,15 @@ def format_memories_text(console: Console, memories: list[dict], title: str = "m
     console.print(f"\n[{BRAND_COLOR}]Found {count} {title}:[/]\n")
 
     for i, mem in enumerate(memories, 1):
-        memory_text = mem.get("memory", mem.get("text", ""))
-        mem_id = mem.get("id", "")[:8]
+        # Defensive against explicit JSON nulls: ``dict.get(k, default)`` returns
+        # ``None`` when the key is present with a null value, which would crash
+        # the slice/``len`` below. Mirror the ``r.get(...) or ""`` pattern
+        # already used by ``format_add_result`` — see #5931.
+        memory_text = mem.get("memory") or mem.get("text") or ""
+        mem_id = (mem.get("id") or "")[:8]
         score = mem.get("score")
         created = _format_date(mem.get("created_at"))
-        category = mem.get("categories", [None])
+        category = mem.get("categories") or [None]
         if isinstance(category, list):
             category = category[0] if category else None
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
closing this pr in favour #5957 of this.

PR #5916, opened 2026-06-26
fix(core): AsyncMemory.reset() uses VectorStoreFactory.reset() when available

ShauryaaSharma:
## Linked Issue

Closes #5915 

## Description

`AsyncMemory.reset()` hardcoded `delete_col()` unconditionally, skipping the `hasattr(self.vector_store, "reset")` check and `VectorStoreFactory.reset()` optimisation that `Memory.reset()` already uses. This meant async callers always took the slower delete-and-recreate path even on vector stores that support an efficient in-place reset.

This PR adds the same `hasattr` guard to `AsyncMemory.reset()`: when `reset()` is present on the vector store, it delegates to `VectorStoreFactory.reset()` via `asyncio.to_thread`; otherwise it falls back to the existing `delete_col()` + `gc.collect()` + `VectorStoreFactory.create()` path, now with a matching `logger.warning` mirroring the sync path.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

Two new tests in `tests/test_memory.py`:
- `test_async_reset_uses_factory_reset_when_available` — verifies `VectorStoreFactory.reset()` is called and `VectorStoreFactory.create()` is not called again when the vector store exposes `reset()`
- `test_async_reset_falls_back_to_delete_col_when_no_reset` — verifies `delete_col()` is called and `VectorStoreFactory.reset()` is not called when the vector store lacks `reset()`

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #5944, opened 2026-06-28
fix(memory): mirror Memory.reset() vector store optimisation in AsyncMemory.reset()

### LINKED ISSUE #5915: bug: AsyncMemory.reset() bypasses VectorStoreFactory.reset() optimisation
### Component

Core / Python SDK

### Summary

### Steps to Reproduce

```python
from mem0 import AsyncMemory

memory = AsyncMemory()
```

Sync path for comparison:

```python
from mem0 import Memory

memory = Memory()
```

### Expected Behavior

### Actual Behavior

### Environment

- OS: any

ly-wang19:
## Problem

Closes #5915.

`AsyncMemory.reset()` unconditionally calls `self.vector_store.delete_col()` followed by `VectorStoreFactory.create()` to recreate the store. The sync counterpart `Memory.reset()` already has a smarter branching pattern: when the vector store exposes `reset()`, it takes the cheaper in-place `VectorStoreFactory.reset()` path that resets the store without tearing down the client/connection. Async users always pay the full teardown + recreation cost, even when the underlying store supports the in-place optimisation (Qdrant, PGVector, etc.).

## Why This Change Was Made

- **Parity with sync `Memory.reset()`**: the sync/async divergence is silent — async callers do not get the optimisation that the equivalent sync path provides. Fixing this lets async users benefit from in-place reset where supported.
- **Performance**: in-place `VectorStoreFactory.reset()` avoids re-creating the underlying client/connection on every reset, which matters for long-lived async services that call `reset()` during re-onboarding, tenant wipe, or test teardown.
- **Minimal blast radius**: the change preserves the existing fallback path exactly — stores without `reset()` still go through `delete_col()` + recreate.

## User Impact

- Users of `AsyncMemory.reset()` whose vector store supports in-place reset will now see faster resets with no client/connection churn.
- Users whose vector store does not support `reset()` see no behavior change — the fallback path is unchanged.

## Change

- `mem0/memory/main.py` — `AsyncMemory.reset()` now branches on `hasattr(self.vector_store, "reset")`:
  - when supported: `await asyncio.to_thread(VectorStoreFactory.reset, self.vector_store)` (in-place)
  - otherwise: existing `delete_col()` + client close + `VectorStoreFactory.create()` (fallback)

This mirrors the sync `Memory.reset()` branching pattern already present at `mem0/memory/main.py:2061`.

## Evidence

```
$ uv run pytest tests/test_memory.py -k "async_memory_reset" -v
============================= test session starts ==============================
platform darwin -- Python 3.13.5, pytest-9.1.1, pluggy-1.6.0 -- /private/tmp/mem0/.venv/bin/python
cachedir: .pytest_cache
rootdir: /private/tmp/mem0
configfile: pyproject.toml
plugins: mock-3.15.1, anyio-4.14.1, asyncio-1.4.0, langsmith-0.9.3
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collecting ... collected 60 items / 57 deselected / 3 selected

tests/test_memory.py::test_async_memory_reset_clears_messages_table PASSED [ 33%]
tests/test_memory.py::test_async_memory_reset_uses_in_place_reset_when_available PASSED [ 66%]
tests/test_memory.py::test_async_memory_reset_falls_back_to_delete_col_when_reset_unavailable PASSED [100%]

======================= 3 passed, 57 deselected in 1.03s =======================
```

Two new regression tests added:

- `test_async_memory_reset_uses_in_place_reset_when_available` — verifies the in-place `reset()` path is taken when the store supports it, and that `VectorStoreFactory.create` is NOT called during reset (delta-based call_count assertion).
- `test_async_memory_reset_falls_back_to_delete_col_when_reset_unavailable` — uses `del mock_vector_store.reset` to simulate a store without `reset()`, and verifies `delete_col()` is called once and the store is recreated exactly once.

### Regression verification

Reverting only `mem0/memory/main.py` (keeping the new tests) reproduces the original bug:

```
FAILED tests/test_memory.py::test_async_memory_reset_uses_in_place_reset_when_available
>   mock_vector_store.reset.assert_called_once()
E   AssertionError: Expected 'reset' to have been called once. Called 0 times.
```

Restoring the fix makes both tests pass.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3706,20 +3706,32 @@ class AsyncMemory(MemoryBase):
             Recreates the vector store with a new client
         """
         logger.warning("Resetting all memories")
-        await asyncio.to_thread(self.vector_store.delete_col)
 
-        gc.collect()
+        # Mirror Memory.reset(): use the in-place VectorStoreFactory.reset()
+        # optimization when the vector store supports it, falling back to
+        # delete_col + recreate only when it does not. The previous code
+        # always tore down + recreated, losing the optimization for every
+        # async caller — see #5915.
+        if hasattr(self.vector_store, "reset"):
+            self.vector_store = await asyncio.to_thread(
+                VectorStoreFactory.reset, self.vector_store
+            )
+        else:
+            logger.warning("Vector store does not support reset. Skipping.")
+            await asyncio.to_thread(self.vector_store.delete_col)
 
-        if hasattr(self.vector_store, "client") and hasattr(self.vector_store.client, "close"):
-            await asyncio.to_thread(self.vector_store.client.close)
+            gc.collect()
+
+            if hasattr(self.vector_store, "client") and hasattr(self.vector_store.client, "close"):
+                await asyncio.to_thread(self.vector_store.client.close)
+
+            self.vector_store = VectorStoreFactory.create(
+                self.config.vector_store.provider, self.config.vector_store.config
+            )
 
         await asyncio.to_thread(self.db.reset)
         await asyncio.to_thread(self.db.close)
         self.db = SQLiteManager(self.config.history_db_path)
-
-        self.vector_store = VectorStoreFactory.create(
-            self.config.vector_store.provider, self.config.vector_store.config
-        )
 
         if self._entity_store is not None:
             try:
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -169,6 +169,79 @@ async def test_async_memory_reset_clears_messages_table(mock_llm_factory, mock_v
     assert hist_count == 0, "history must be empty after AsyncMemory.reset()"
 
 
+@pytest.mark.asyncio
+@patch('mem0.utils.factory.EmbedderFactory.create')
+@patch('mem0.utils.factory.VectorStoreFactory.create')
+@patch('mem0.utils.factory.LlmFactory.create')
+async def test_async_memory_reset_uses_in_place_reset_when_available(
+    mock_llm_factory, mock_vector_factory, mock_embedder_factory, tmp_path
+):
+    """Regression for #5915: when the vector store supports ``reset``, use the
+    in-place ``VectorStoreFactory.reset()`` path and skip the more expensive
+    ``delete_col`` + recreate teardown."""
+    mock_embedder_factory.return_value = MagicMock()
+    mock_vector_store = MagicMock()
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Needs Human Review

**Why:** the code review is clean, but `license/cla` is still pending, so this should not be treated as merge-ready yet.

### Pay attention to
- `mem0/memory/main.py:3715` — this mirrors the existing sync `Memory.reset()` branch and delegates the optimized async path through `VectorStoreFactory.reset()`.
- `tests/test_memory.py:201` and `tests/test_memory.py:237` — the regression tests cover both the in-place reset path and the fallback delete/recreate path.

### Findings
Critical: none.

Important: none.

Minor: none.

### Duplicate overlap
PR #5916 fixes the same root cause. After reading both diffs, I’d recommend consolidating on this PR once the CLA clears: #5944 has the same bounded implementation shape, stronger fallback coverage because it asserts the vector store is recreated exactly once, and the full Python SDK CI Gate is green. #5916 has the CLA signed, but `gh pr checks` only shows the CLA check and the advisory Vercel failure, not the CI Gate/test matrix.

### Evidence
- **CI:** `CI Gate` passed; Python SDK build jobs passed on 3.10, 3.11, and 3.12. Ignored the Vercel deploy failure as advisory. `license/cla` is pending.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/test_memory.py -k "async_memory_reset" -v --tb=short` -> 3 passed, 57 deselected.
- **Red-green proof:** with only `mem0/memory/main.py` restored to `origin/main`, `test_async_memory_reset_uses_in_place_reset_when_available` failed with `Expected 'reset' to have been called once. Called 0 times.` Restoring the PR implementation made the same selection pass again.
- **Lint:** `hatch -e dev_py_3_12 run ruff check mem0/memory/main.py tests/test_memory.py` -> all checks passed.
- **Linked issue:** #5915 — PR addresses the reported Python async reset root cause.
- **Cross-package check:** TypeScript reset is structurally different and does not have the same Python sync-vs-async divergence.
- **Could NOT verify:** live provider reset behavior against Qdrant/PGVector; this review verified the shared factory path and fallback behavior by targeted unit tests and code tracing.

### Assessment
Ready to merge: **No — maintainer/admin action required.** Code path looks ready, but the CLA must clear first; once that happens, this is the stronger long-term path over #5916.

PR #5950, opened 2026-06-28
fix(memory): paginate delete_all so it does not stop at the first page

ly-wang19:
## What Problem This Solves

Closes #4869.

`Memory.delete_all` and `AsyncMemory.delete_all` previously called `vector_store.list(filters=filters)` exactly once and deleted whatever came back. Several production vector stores cap `list` at a default page size (Pinecone/Milvus/Turbopuffer default to 100), so when a user had more matching memories than the page size, the older entries were silently left behind.

Reproducing: create 150 memories for one `user_id` against any vector store that caps `list` at 100, then call `delete_all(user_id=...)` — only the first 100 memories get deleted.

## Why This Change Was Made

- The vector store `list` interface has no offset/page parameter — pagination has to happen on the caller side.
- The fix loops, stopping when a page is shorter than requested (last page) or empty. This is robust against stores that don't paginate (one call returns everything; loop terminates on the short-page check).
- The same bug exists in the TypeScript `deleteAll`; this PR addresses the Python SDK only. A separate PR can address the TS path.

## User Impact

- Bulk deletions now actually delete all matching memories, regardless of how many pages the vector store needs to return them.
- No behavior change for callers with fewer memories than the page size — the loop terminates after one call exactly as before.
- Existing tests `test_async_delete_all_continues_on_partial_failure` and `TestAsyncDeleteAllEntityRace::test_async_delete_all_bulk_clears_entity_store` still pass unchanged.

## Evidence

Three regression tests added under `tests/test_memory.py`:

- `test_sync_delete_all_paginates_through_all_memories` — mock returns 3 pages `[m0,m1] / [m2,m3] / [m4]`; asserts `vector_store.list.call_count == 3` and `_delete_memory.call_count == 5`.
- `test_sync_delete_all_stops_when_first_page_empty` — empty first page → exactly 1 list call, 0 deletes.
- `test_async_delete_all_paginates_through_all_memories` — same coverage for `AsyncMemory`.

Catch-bug verification: each pagination test was confirmed to fail when `delete_all` is forced back to a single-call implementation (`list.call_count == 1` vs expected 3), and to pass with the loop in place.

Full file green:
- `uv run pytest tests/test_memory.py -k delete_all` → 5 passed (3 new + 2 pre-existing).

### Diffstat

```
mem0/memory/main.py  |  49 ++++++++++++++----
tests/test_memory.py | 140 +++++++++++++++++++++++++++++++++++++++++++++++++++
2 files changed, 180 insertions(+), 9 deletions(-)
```

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -81,6 +81,12 @@ warnings.filterwarnings("ignore", category=DeprecationWarning, message=".*swigva
 logger = logging.getLogger(__name__)
 
 
+# Page size used by `delete_all` when iterating the vector store. Some vector
+# stores cap `list` at a default page size (e.g. 100); without paging,
+# `delete_all` silently leaves older matching memories behind (issue #4869).
+_DELETE_ALL_PAGE_SIZE = 1000
+
+
 def _vector_store_list_rows(listed):
     if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list):
         return listed[0]
@@ -1849,14 +1855,25 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
-        memories = self.vector_store.list(filters=filters)[0]
-        for memory in memories:
-            self._delete_memory(memory.id)
+        # delete all vector memories and reset the collections.
+        # Iterate in pages because some vector stores cap `list` at a default
+        # page size (e.g. 100). A single call would leave older matching
+        # memories behind. See issue #4869. We stop when a page is shorter
+        # than requested (last page) or empty.
+        total_deleted = 0
+        while True:
+            page = self.vector_store.list(filters=filters, top_k=_DELETE_ALL_PAGE_SIZE)[0]
+            if not page:
+                break
+            for memory in page:
+                self._delete_memory(memory.id)
+            total_deleted += len(page)
+            if len(page) < _DELETE_ALL_PAGE_SIZE:
+                break
 
-        logger.info(f"Deleted {len(memories)} memories")
+        logger.info(f"Deleted {total_deleted} memories")
 
-        decay_usage_notice = detect_decay_usage_from_delete_all(len(memories))
+        decay_usage_notice = detect_decay_usage_from_delete_all(total_deleted)
         if decay_usage_notice:
             display_decay_usage_notice(self, "sync", "delete_all", *decay_usage_notice)
         else:
@@ -3468,10 +3485,24 @@ class AsyncMemory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "async"})
-        memories = await asyncio.to_thread(self.vector_store.list, filters=filters)
+        # Iterate in pages because some vector stores cap `list` at a default
+        # page size (e.g. 100). A single call would leave older matching
+        # memories behind. See issue #4869. We stop when a page is shorter
+        # than requested (last page) or empty.
+        all_memories = []
+        while True:
+            page = await asyncio.to_thread(
+                self.vector_store.list, filters=filters, top_k=_DELETE_ALL_PAGE_SIZE
+            )
+            page_memories = page[0] if page else []
+            if not page_memories:
+                break
+            all_memories.extend(page_memories)
+            if len(page_memories) < _DELETE_ALL_PAGE_SIZE:
+                break
 
         delete_tasks = []
-        for memory in memories[0]:
+        for memory in all_memories:
             delete_tasks.append(self._delete_memory(memory.id, skip_entity_cleanup=True))
 
         results = await asyncio.gather(*delete_tasks, return_exceptions=True)
@@ -3487,7 +3518,7 @@ class AsyncMemory(MemoryBase):
 
         logger.info(f"Deleted {len(results) - len(errors)} memories")
 
-        decay_usage_notice = detect_decay_usage_from_delete_all(len(memories[0]))
+        decay_usage_notice = detect_decay_usage_from_delete_all(len(all_memories))
         if decay_usage_notice:
             await display_decay_usage_notice_async(self, "async", "delete_all", *decay_usage_notice)
         else:
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -1000,6 +1000,146 @@ async def test_async_delete_all_continues_on_partial_failure(mock_sqlite, mock_l
     assert mock_vector_store.delete.call_count == 2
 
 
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- Correct root-cause diagnosis, and the sync path (`main.py:1864-1872`) is the right shape: delete happens *inside* the while-loop before the next `list()`, so the live result set actually shrinks between pages.
- `stops_when_first_page_empty` boundary test is a good, solid regression guard for the trivial case.
- Changed files are ruff-clean.

### Issues
#### Critical (must fix)
1. `mem0/memory/main.py:3492-3508` — the async loop only appends pages to `all_memories` inside `while True`; every `_delete_memory` call is deferred to one `asyncio.gather()` *after* the loop exits, so nothing is ever deleted between successive `list()` calls. Once matches ≥ `_DELETE_ALL_PAGE_SIZE` (1000), this never terminates — reproduced independently 5 times against a real stateful store, one run hit 41,229 `list()` calls in 10s with 0/2,500 rows deleted. Please give async the same shape as sync: delete (or at least `gather` that page) *inside* the loop, before the next `list()`.
2. `tests/test_main.py:415` (`test_delete_all_trims_user_id_before_list`) — asserts `list.assert_called_once_with(filters=...)` with no `top_k`; prod now always passes `top_k=1000`, so this pre-existing test is red on 3.10/3.11/3.12 — this is the actual required `Python SDK / build_mem0` failure blocking merge, just needs the assertion updated.

#### Important (should fix)
1. `tests/test_memory.py:1021-1026` — `_paginated_list_factory`'s mock pops the next canned page unconditionally, regardless of whether a delete happened, so the async pagination test can't detect the non-termination bug above — right now it's vacuous for the exact defect this PR exists to fix.
2. `mem0/memory/main.py:1866,3498` — treating an empty page as "done" trusts `list()` to raise on error, but 8 vector stores (e.g. `pinecone.py:405-412`, `opensearch.py:396-398`) swallow internal exceptions and return `[[]]` instead — a transient error now silently truncates `delete_all` the same way #4869 originally described, just triggered by a network blip instead of a page cap.
3. Elasticsearch/OpenSearch refresh lag (`elasticsearch.py:61` ~1s, `opensearch.py:63` ~10s) can re-list an id a prior page already deleted; neither delete loop has a try/except, so the redundant delete raises uncaught (an unguarded `NotFoundError` at `opensearch.py:301` for OpenSearch; for ES it surfaces one frame up as a `ValueError` from `_delete_memory`'s own not-found check) and aborts `delete_all()` mid-run — a failure mode that didn't exist pre-PR.
4. The PR body says "Closes #4869" but also notes the TS `deleteAll` (`mem0-ts/src/oss/src/memory/index.ts:1620`, still a single unpaginated call) is intentionally out of scope — could you reword to "Part of #4869" so the issue doesn't auto-close while the TS call site is still broken?

#### Minor (optional)
1. `mem0/memory/main.py:90-95` — `_vector_store_list_rows` already exists to defensively normalize `list()`'s return shape; neither new loop reuses it (both hand-roll a plain `[0]` index instead).
2. `mem0/memory/main.py:2035` — sync `_delete_memory` has no `skip_entity_cleanup` param, so removing the old implicit page cap means `entity_store.list(top_k=10000)` now runs once per deleted memory across the whole matching set. Async already batches this via `_bulk_clear_entity_store` (`main.py:3701`) — worth mirroring on the sync side.

### Evidence
- **CI:** red — `Python SDK / build_mem0` FAILURE on 3.10/3.11/3.12 (3.11/3.12 cancelled after 3.10 failed), `CI Gate` FAILURE. Vercel failure is unrelated preview-deploy noise; `license/cla` green.
- **Targeted tests:** `pytest tests/test_memory.py -k delete_all` → 5 passed. `pytest tests/test_main.py tests/test_memory.py` → 1 failed, 99 passed (the one failure is Critical #2 above).
- **Red-green:** sync pagination tests fail with the fix reverted (`AttributeError: module 'mem0.memory.main' has no attribute '_DELETE_ALL_PAGE_SIZE'`), pass restored — load-bearing for the sync half. The async tests, however, pass both with and without a working async implementation, since the mock doesn't model deletion shrinking the store (Important #1).
- **Linked issue:** #4869 — names Python sync, Python async, and TS `deleteAll` as broken; this PR completes (and correctly completes) only the sync one.
- **Could NOT verify:** live behavior against a real hosted vector-store backend — reproductions used code tracing plus stateful in-memory fakes, not a network backend.

### Assessment
Ready to merge: **With fixes** — the sync half is the right shape and worth keeping as-is; the async half needs the delete moved inside the loop (mirror the sync structure) before it's safe, plus the CI assertion and test-mock gaps above. Separately, #4872 covers this same root cause plus the TS side with a similar delete-inside-loop shape, if a second reference is useful — it's not currently mergeable itself, so no pressure to reconcile with it, just flagging in case it saves you time. Happy to take another look as soon as you push.

cen-zp:
I worked through the same bug in #6465, which I have now closed in favor of this PR and the existing maintainer review path. A few findings may help move this implementation forward:

1. The async path needs to await deletion of each batch inside the pagination loop before calling list() again. Otherwise a stateful store keeps returning the same first page and the loop does not make progress. A stateful fake that only removes rows when delete() is called catches this reliably.

2. Eventual-consistency lag needs a bounded no-progress policy. A store may briefly return only IDs deleted in the previous iteration while additional IDs still exist behind that stale page. If seen IDs are filtered and an all-seen page immediately terminates the loop, later records can remain. A small bounded retry/backoff counter is safer than either immediate termination or an unbounded loop.

3. For the sync path, I recommend preserving the existing exception behavior unless the project intentionally changes the API contract. Swallowing delete failures and still returning the success message can hide residual memories.

Useful regression cases would be: sync and async stores that shrink only after deletion; one temporarily repeated stale page followed by new IDs; an empty initial store; and a sync deletion failure that remains observable to the caller.

The closed #6465 branch remains available as a reference for delete-inside-loop structure and stateful tests. I have not verified these paths against a hosted vector-store backend.

kartik-mem0:
Closing this in favour of #6487, which fixes the same half of #4869 and is now merge-ready.

Credit where it's due: your diagnosis was right, and your sync implementation was the correct shape, deleting inside the loop so the live result set actually shrinks between `list()` calls. #6487 uses that same structure.

The deciding difference is the async path. The 2026-07-02 review flagged that `main.py:3492-3508` accumulates every page into `all_memories` before deleting any, and since `VectorStoreBase.list(filters, top_k)` takes no offset or cursor (`mem0/vector_stores/base.py:59`), re-calling it against an unchanged store returns the same full page indefinitely. For any caller with 1000 or more matching memories that loop never terminates, and `all_memories` grows without bound. That is still present on the current head.

#6487 deletes each page before listing the next on both paths, and bounds the drain with an iteration cap, a per-id failure cap, and a no-progress backoff. It was verified against real Qdrant and pgvector, with the pre-fix failure reproduced on `main` and passing on the head.

#4869 stays open. Both PRs leave the TypeScript `deleteAll` (`mem0-ts/src/oss/src/memory/index.ts:1620`) unpaginated, so that part of the issue is still unfixed and is a good follow-up if you want it.

Thanks for the report and the patch. Reopen if you think something here is off.

PR #5910, opened 2026-06-26
fix(elasticsearch): set size on KNN search to respect top_k

AnnaSuSu:
## Linked Issue

Closes #5909

## Description

`ElasticsearchDB.search()` builds the KNN query with `knn.k` /
`num_candidates` but never sets the top-level `size` field. Elasticsearch
defaults to returning only 10 hits, so any search requesting `top_k > 10`
was silently capped at 10 — degrading recall and shrinking the candidate
pool used for hybrid re-ranking.

This PR adds `size: top_k` to the KNN query body, bringing KNN search in
line with the existing `keyword_search()` and `list()` methods that
already set it.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added regression test `test_search_sets_size_to_top_k` and extended
`test_search` to assert the KNN body includes `size == top_k`. Without
the fix the suite fails with `KeyError: 'size'`; with the fix all 15
tests in `tests/vector_stores/test_elasticsearch.py` pass locally.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #5951, opened 2026-06-28
fix(elasticsearch): add top-level size to KNN search so top_k > 10 is honoured

### LINKED ISSUE #5909: bug: Elasticsearch vector search caps results at 10 (missing size on KNN query)
## Description

`ElasticsearchDB.search()` builds the KNN query with `knn.k` and `num_candidates` but no top-level `size`. Elasticsearch caps a search response at its default `size` of **10** hits regardless of `knn.k`, so a KNN search returns at most 10 results even when more are requested.

`Memory.search()` calls the vector store with an internal limit of up to `max(limit * 4, 60)` to build the hybrid re-ranking candidate pool, so on Elasticsearch that pool is silently truncated to 10 — capping recall and degrading ranking quality. The sibling `keyword_search()` (`mem0/vector_stores/elasticsearch.py:199`) and `list()` already set `size`; only KNN `search()` omits it.

## Impact

On Elasticsearch, `memory.search(query, limit=N)` returns at most 10 memories for any `N > 10`, and hybrid scoring runs over at most 10 candidates instead of the intended pool. Affects both the sync and async search paths.

## Location

`mem0/vector_stores/elasticsearch.py`, `search()` — the KNN `search_query` body is built without a `size` field.

ly-wang19:
## What Problem This Solves

Fixes #5909.

`ElasticsearchDB.search()` built the KNN query body without a top-level `size`, so Elasticsearch capped the response at its default of 10 hits regardless of `knn.k`. `Memory.search()` calls the vector store with an internal limit of up to `max(limit * 4, 60)` to build the hybrid re-ranking candidate pool, so on Elasticsearch that pool was silently truncated to 10 — capping recall and degrading ranking quality for any `memory.search(query, limit=N)` where `N > 10`.

The sibling `keyword_search()` and `list()` paths already set `size`; only the KNN `search()` omitted it.

## Why This Change Was Made

- `knn.k` controls how many nearest neighbours each shard returns, not how many hits the response carries. The response size is governed by the top-level `size` field, which defaults to 10.
- `num_candidates = top_k * 2` was already correct on the candidate side; only the response size was uncapped-from-default.
- The fix is one line in the default (non-custom) KNN path. When `custom_search_query` is supplied, the user callable owns the body and we continue to pass it through unchanged.

## User Impact

- `memory.search(query, limit=N)` on Elasticsearch now returns up to N memories instead of stopping at 10.
- Hybrid re-ranking now sees the full intended candidate pool (up to 60 by default), improving ranking quality.
- No behaviour change for `top_k <= 10` (existing default-sized responses).
- No behaviour change for callers that supply `custom_search_query`.

## Evidence

- New regression tests:
  - `test_search_sets_size_above_default_cap` — `top_k=60` yields `body["size"] == 60` (mirrors `Memory.search`'s internal candidate pool size).
  - `test_search_with_filters_keeps_size` — filters do not displace `size` (`body["size"] == 25`, filter bool clause still present).
- Existing `test_search` extended to assert `body["size"] == top_k`.
- Catch-bug verified: with the `size` line removed, all three tests fail with `KeyError: 'size'`.
- Full suite: `uv run pytest tests/vector_stores/test_elasticsearch.py` — 16/16 pass.

```
$ uv run pytest tests/vector_stores/test_elasticsearch.py -q
................                                                         [100%]
16 passed in 2.43s
```

### DIFF
--- a/mem0/vector_stores/elasticsearch.py
+++ b/mem0/vector_stores/elasticsearch.py
@@ -148,8 +148,11 @@ class ElasticsearchDB(VectorStoreBase):
         if self.custom_search_query:
             search_query = self.custom_search_query(vectors, top_k, filters)
         else:
+            # Elasticsearch caps the response at the default `size` of 10
+            # regardless of `knn.k`. Set `size` so top_k > 10 is honoured.
             search_query = {
-                "knn": {"field": "vector", "query_vector": vectors, "k": top_k, "num_candidates": top_k * 2}
+                "size": top_k,
+                "knn": {"field": "vector", "query_vector": vectors, "k": top_k, "num_candidates": top_k * 2},
             }
             if filters:
                 filter_conditions = []
--- a/tests/vector_stores/test_elasticsearch.py
+++ b/tests/vector_stores/test_elasticsearch.py
@@ -209,12 +209,47 @@ class TestElasticsearchDB(unittest.TestCase):
         self.assertEqual(body["knn"]["query_vector"], vectors)
         self.assertEqual(body["knn"]["k"], 5)
         self.assertEqual(body["knn"]["num_candidates"], 10)
+        # Elasticsearch caps the response at the default `size` of 10
+        # regardless of `knn.k`, so `size` must mirror `top_k`.
+        self.assertEqual(body["size"], 5)
 
         # Verify results
         self.assertEqual(len(results), 1)
         self.assertEqual(results[0].id, "id1")
         self.assertEqual(results[0].score, 0.8)
         self.assertEqual(results[0].payload, {"key1": "value1"})
+
+    def test_search_sets_size_above_default_cap(self):
+        # Regression for #5909: Memory.search passes an internal top_k of
+        # up to `max(limit * 4, 60)` to support hybrid re-ranking. Without
+        # `size` on the KNN body, Elasticsearch caps the response at 10
+        # hits regardless of `knn.k`, silently truncating recall.
+        self.client_mock.search.return_value = {"hits": {"hits": []}}
+
+        self.es_db.search(query="", vectors=[[0.1] * 1536], top_k=60)
+
+        body = self.client_mock.search.call_args[1]["body"]
+        self.assertEqual(body["size"], 60)
+        self.assertEqual(body["knn"]["k"], 60)
+        self.assertEqual(body["knn"]["num_candidates"], 120)
+
+    def test_search_with_filters_keeps_size(self):
+        # Filters must not displace the `size` field added for #5909.
+        self.client_mock.search.return_value = {"hits": {"hits": []}}
+
+        self.es_db.search(
+            query="",
+            vectors=[[0.1] * 1536],
+            top_k=25,
+            filters={"user_id": "u1"},
+        )
+
+        body = self.client_mock.search.call_args[1]["body"]
+        self.assertEqual(body["size"], 25)
+        self.assertEqual(
+            body["knn"]["filter"],
+            {"bool": {"must": [{"term": {"metadata.user_id": "u1"}}]}},
+        )
 
     def test_custom_search_query(self):
         # Mock custom search query


mem0/vector_stores/elasticsearch.py
```diff
@@ -148,8 +148,11 @@ def search(
         if self.custom_search_query:
             search_query = self.custom_search_query(vectors, top_k, filters)
         else:
+            # Elasticsearch caps the response at the default `size` of 10
+            # regardless of `knn.k`. Set `size` so top_k > 10 is honoured.
             search_query = {
-                "knn": {"field": "vector", "query_vector": vectors, "k": top_k, "num_candidates": top_k * 2}
+                "size": top_k,
```
kartik-mem0, line 154: Nice catch matching the `keyword_search`/`list` pattern. One thing to consider: `size` is unclamped, and `Memory.search()`'s internal `max(limit*4, 60)` over-fetch can push it past Elasticsearch's `index.max_result_window` (10000) for `limit > 2500`, turning a silent wrong-truncation into a hard `illegal_argument_exception`. Same unclamped pattern already exists in `opensearch.py` and this file's own `keyword_search`/`list`, so not new to this PR, just flagging as a possible fast-follow.

kartik-mem0 (commented):

Elasticsearch KNN `search()` capped responses at 10 regardless of `top_k`; this adds the missing top-level `size`, matching the `keyword_search()`/`list()` pattern already in the same file.

**Before you merge:** #5910 fixes the same issue with an identical one-line production change but thinner test coverage (no filters+size interaction test); recommend closing it in favor of this one.
**Follow-up:** `mem0-ts/src/oss/src/vector_stores/elasticsearch.ts` `search()` has the identical missing-`size` bug, unaddressed here since #5909 is Python-scoped; worth its own issue.
**Not verified:** live Elasticsearch cluster, mocked client only.

Risk: low, one-line change to an already-established in-file pattern with a real red-green-proven regression test · Your time: ~3 min

- **Problem is real:** reproduces on `main` at `mem0/vector_stores/elasticsearch.py:154-167` (pre-PR), any `search(top_k > 10)` call including `Memory.search()`'s internal `max(limit*4, 60)` over-fetch pool
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_elasticsearch.py -v --tb=short` -> 16 passed
- **Red-green:** implementation reverted to pre-PR `main` -> 3 failed, each `KeyError: 'size'` on the exact missing field; restored -> 16 passed again
- **CI:** all real checks green (build_mem0 3.10/3.11/3.12, CI Gate, license/cla); Vercel red but advisory, ignored
- **Security:** cleared, `top_k` is an internally computed int never raw user text, tenancy filter wiring untouched by this diff
- **Bloat (ponytail):** lean already, one-line fix matching an existing in-file pattern
- **Linked issue:** #5909

</details>

kartik-mem0:
Thank you for this, and I'm sorry it sat open for as long as it did. The diagnosis is exactly right: Elasticsearch caps a response at the default `size` of 10 no matter what `knn.k` says, and because `Memory._search_vector_store` over-fetches a pool of `max(limit * 4, 60)` for re-ranking, a plain `search(limit=20)` was being silently truncated to 10 hits well before any user asked for more than ten memories.

Closing this in favour of #5910, which fixes the same bug and was opened two days earlier. That is the only real difference between them, so this is a scheduling call rather than a quality one. The production change in the two PRs is identical line for line, and your `test_search_with_filters_keeps_size` is byte for byte the same test as the one there, which is a good sign that the fix is the obvious correct one.

Your `num_candidates` assertion is the one thing here that the other PR does not repeat in its over-fetch test, though it is already covered by the existing `test_search` case, so nothing is lost by closing this.

To be clear about where things stand: #5909 stays open and is linked from #5910, so the bug remains tracked until that merges.

This was a real bug, correctly found and correctly fixed, and the duplicate work is on our side for not getting to #5910 sooner. Please do send more.

PR #5952, opened 2026-06-28
fix(memory): defer langchain-core import in async procedural memory

### LINKED ISSUE #5907: bug: AsyncMemory procedural memory requires optional langchain-core even when no custom LLM is used

ly-wang19:
## What Problem This Solves

Fixes #5907.

`AsyncMemory._create_procedural_memory()` imported `convert_to_messages` from `langchain_core.messages.utils` at function entry, but the symbol is only used inside the `if llm is not None:` branch. `langchain-core` is an optional dependency (the `[extras]` group in `pyproject.toml`), so on a base `pip install mem0ai` the default path (`llm=None`) raised `ModuleNotFoundError` before doing any work. The sync `Memory._create_procedural_memory()` has no langchain dependency on the default path and works for the same call, so async was the only broken side.

```python
import asyncio
from mem0 import AsyncMemory

m = AsyncMemory()
```

## Why This Change Was Made

- The default path (`llm=None`) calls `self.llm.generate_response(...)` directly; it has no need for `convert_to_messages` and therefore no need for `langchain-core`.
- The sync equivalent (`Memory._create_procedural_memory`) already works without `langchain-core` because the import lives in the custom-LLM branch (effectively, by virtue of not existing on the default path).
- Moving the import into the `if llm is not None:` branch aligns the async behaviour with sync, restoring parity. The existing `try/except` around the import is preserved so callers that supply a custom langchain LLM without the dependency installed still get the friendly `"install 'langchain-core'"` log message and a clear re-raise.

## User Impact

- `AsyncMemory.add(..., memory_type=\"procedural_memory\")` on a base install (no `[extras]`) now works, matching the sync `Memory` behaviour.
- Callers supplying a custom `llm` still need `langchain-core`; if missing, they see the same ImportError as before.
- No behaviour change for callers already installing `[extras]`.

## Evidence

Two new regression tests in `tests/test_memory.py`:

- `test_async_procedural_memory_default_path_does_not_require_langchain_core` — patches `builtins.__import__` to reject `langchain_core` and verifies the default path (`llm=None`) succeeds, returns the LLM response, and only invokes `self.llm.generate_response` once.
- `test_async_procedural_memory_custom_llm_still_requires_langchain_core` — symmetric: with the same import block, supplying a custom `llm` re-raises `ImportError(\"...langchain-core...\")` and never invokes `llm.invoke`.

Catch-bug verified: with the import restored to function entry, the default-path test fails with `ImportError: simulated missing langchain-core`. With the import deferred to the branch, both new tests and the existing `test_async_procedural_memory_langchain_strips_code_blocks` pass.

```
$ uv run pytest tests/test_memory.py::test_async_procedural_memory_default_path_does_not_require_langchain_core tests/test_memory.py::test_async_procedural_memory_custom_llm_still_requires_langchain_core tests/test_memory.py::test_async_procedural_memory_langchain_strips_code_blocks -v
...
3 passed in 4.24s
```

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3556,16 +3556,6 @@ class AsyncMemory(MemoryBase):
             llm (llm, optional): LLM to use for the procedural memory creation. Defaults to None.
             prompt (str, optional): Prompt to use for the procedural memory creation. Defaults to None.
         """
-        try:
-            from langchain_core.messages.utils import (
-                convert_to_messages,  # type: ignore
-            )
-        except Exception:
-            logger.error(
-                "Import error while loading langchain-core. Please install 'langchain-core' to use procedural memory."
-            )
-            raise
-
         logger.info("Creating procedural memory")
 
         parsed_messages = [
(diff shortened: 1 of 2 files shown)


kartik-mem0:
closing this pr in favour of this pr #6209

PR #5954, opened 2026-06-28
perf(core): eliminate N redundant vector_store.get() calls in delete_all (#5869)

### LINKED ISSUE #5869: perf(core): Memory.delete_all() makes N redundant vector_store.get() calls

ly-wang19:
## What Problem This Solves

`Memory.delete_all()` and `AsyncMemory.delete_all()` make N redundant `vector_store.get()` calls — one for every memory they just fetched via `list()` — turning an O(1) read into O(N). The sync path also does per-memory entity-store cleanup instead of the single bulk pass the async path already uses.

Closes #5869.

## Why This Change Was Made

`_delete_memory(memory_id, existing_memory=None)` fetches the memory via `vector_store.get()` when `existing_memory` is `None`. `delete_all` already had each fetched memory in hand but called `_delete_memory(memory.id)` without forwarding it, so every call re-fetched. For 100 memories that's 1 `list` + 100 `get` round trips where 1 `list` was sufficient. The sync path additionally ran per-memory entity-store cleanup (`_remove_memory_from_entity_store`) on every delete, producing N+1 entity-store passes where one bulk clear would do.

## User Impact

- `delete_all()` is now O(N) total round trips to the vector store (1 `list` + N `delete`) instead of O(2N) (1 `list` + N `get` + N `delete`).
- Sync `delete_all()` entity cleanup is now a single bulk pass, matching async behavior. Large stores see the biggest win.
- Behavior is otherwise unchanged: history is still recorded per memory, partial failures still don't abort (async uses `return_exceptions=True`; sync still best-effort inside the loop).

## Evidence

- `MEM0_TELEMETRY=False python3 -m pytest tests/test_memory.py` — 62/62 pass (was 60; +2 new regression tests).
- `test_sync_delete_all_passes_existing_memory_and_uses_bulk_clear` asserts `vector_store.get` is never called and the entity store's `list` runs exactly once (proving bulk-clear, not per-memory).
- `test_async_delete_all_passes_existing_memory` asserts the same for the async path.
- Updated `test_async_delete_all_continues_on_partial_failure`: it previously injected the simulated store failure on `vector_store.get(mem-2)`, which my fix skips entirely. Moved the failure injection to `vector_store.delete(mem-2)` to exercise the same `return_exceptions=True` recovery path. Assertion updated to `call_count == 3` (delete is now invoked for every memory; one fails internally).

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -605,6 +605,26 @@ class Memory(MemoryBase):
         except Exception as e:
             logger.warning(f"Entity upsert failed for '{entity_text}': {e}")
 
+    def _bulk_clear_entity_store(self, filters):
+        """Delete all entity records matching the given scope filters.
+
+        Used by delete_all to avoid N read-modify-write passes when every
+        memory in scope is being removed anyway. Mirrors the async helper.
+        """
+        if self._entity_store is None:
+            return
+        search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
+        try:
+            listed = self.entity_store.list(filters=search_filters, top_k=10000)
+            rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
+            for row in rows or []:
+                try:
+                    self.entity_store.delete(vector_id=row.id)
+                except Exception as e:
+                    logger.debug(f"Bulk entity delete failed for id={row.id}: {e}")
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #5956, opened 2026-06-28
fix(cli): reject invalid int config values without traceback

### LINKED ISSUE #5933: bug(cli): `config set` on an int field crashes with uncaught ValueError on non-numeric input

barry166:
Fixes #5933.

## Summary
- return `False` when an int config field receives a non-numeric value
- keep the existing boolean success/failure contract used by `cmd_config_set`
- add regression coverage for invalid `version` input

## Tests
- `PYTHONPATH=cli/python/src pytest cli/python/tests/test_config.py -q`
- `PYTHONPATH=cli/python/src pytest cli/python/tests/test_commands.py -q`
- `python -m compileall -q cli/python/src/mem0_cli cli/python/tests`
- `git diff --check`

AI-assisted: yes. I manually reviewed the diff and test output before opening this PR.

### DIFF
--- a/cli/python/src/mem0_cli/config.py
+++ b/cli/python/src/mem0_cli/config.py
@@ -235,7 +235,10 @@ def set_nested_value(config: Mem0Config, dotted_key: str, value: str) -> bool:
     if isinstance(current, bool):
         value = value.lower() in ("true", "1", "yes")  # type: ignore[assignment]
     elif isinstance(current, int):
-        value = int(value)  # type: ignore[assignment]
+        try:
+            value = int(value)  # type: ignore[assignment]
+        except ValueError:
+            return False
 
     setattr(obj, final_key, value)
     return True
--- a/cli/python/tests/test_config.py
+++ b/cli/python/tests/test_config.py
@@ -139,6 +139,11 @@ class TestNestedAccess:
         assert set_nested_value(config, "platform.api_key", "new-key")
         assert config.platform.api_key == "new-key"
 
+    def test_set_int_value_rejects_invalid_input(self):
+        config = Mem0Config()
+        assert set_nested_value(config, "version", "abc") is False
+        assert config.version == 1
+
     def test_set_nonexistent_key(self):
         config = Mem0Config()
         assert set_nested_value(config, "nonexistent.key", "val") is False


kartik-mem0:

### What this PR does
This keeps `set_nested_value()` on its documented bool success/failure contract when an int config field receives a non-numeric value. Instead of letting `int(value)` raise a traceback, the shared config setter now returns `False` and leaves the existing config value untouched.

### Strengths
- `cli/python/src/mem0_cli/config.py:237` keeps the fix at the shared config-setting layer used by `cmd_config_set`, so the reachable CLI path is covered without adding caller-specific handling.
- `cli/python/tests/test_config.py:142` adds a regression test that reproduces #5933 and verifies the original int value is not mutated.

### Issues
Minor (optional): none.

### Evidence
- **CI:** `CI Gate`, Python CLI lint/build, and Python CLI tests on 3.10/3.11/3.12 pass. `Vercel` is failing only on deploy authorization and is advisory for this PR.
- **Targeted tests:** `PYTHONPATH=cli/python/src pytest cli/python/tests/test_config.py -q` → `24 passed in 0.07s`.
- **Lint:** `ruff check cli/python/src/mem0_cli/config.py cli/python/tests/test_config.py` → `All checks passed!`.
- **Red-green proof:** the new regression test fails with `ValueError: invalid literal for int() with base 10: 'abc'` after restoring `cli/python/src/mem0_cli/config.py` to `origin/main`, then passes again after restoring the PR implementation.
- **Linked issue:** #5933 — PR addresses the reported root cause: invalid int coercion escaping the bool-return helper contract.
- **Duplicate cluster:** #5934 and #5941 target the same issue; I read their diffs. They use the same root fix, and neither covers a broader affected path or cross-package inconsistency. #5956 is the smallest equivalent fix and was red-green verified here, so I do not consider it superseded.
- **Cross-package check:** Node CLI has a separate `setNestedValue`, but its public key map does not expose `version`; this reported Python CLI path is the affected one.
- **Could NOT verify:** `hatch` is not available on this machine’s PATH, so local verification used direct targeted `pytest`/`ruff`; GitHub CI covers the Python CLI supported-version matrix.

### Assessment
The fix is narrow, tested against the actual failure mode, and keeps the existing CLI config contract intact without adding new surface area.

kartik-mem0 (approved):

PR #5957, opened 2026-06-28
fix(cli): handle null memory fields in output formatters

### LINKED ISSUE #5931: bug(cli): list memory formatters crash on null `id`/`memory` fields from API

barry166:
Fixes #5931.

## Summary
- treat explicit `null` memory/id fields like missing values in text and table output
- keep the existing blank-cell behavior instead of crashing on `None`
- add regression coverage for null memory records

## Tests
- `PYTHONPATH=cli/python/src pytest cli/python/tests/test_output.py -q`
- `PYTHONPATH=cli/python/src pytest cli/python/tests/test_commands.py -q`
- `python -m compileall -q cli/python/src/mem0_cli cli/python/tests`
- `git diff --check`

AI-assisted: yes. I manually reviewed the diff and test output before opening this PR.

### DIFF
--- a/cli/python/src/mem0_cli/output.py
+++ b/cli/python/src/mem0_cli/output.py
@@ -20,8 +20,8 @@ def format_memories_text(console: Console, memories: list[dict], title: str = "m
     console.print(f"\n[{BRAND_COLOR}]Found {count} {title}:[/]\n")
 
     for i, mem in enumerate(memories, 1):
-        memory_text = mem.get("memory", mem.get("text", ""))
-        mem_id = mem.get("id", "")[:8]
+        memory_text = mem.get("memory") or mem.get("text") or ""
+        mem_id = (mem.get("id") or "")[:8]
         score = mem.get("score")
         created = _format_date(mem.get("created_at"))
         category = mem.get("categories", [None])
@@ -67,8 +67,8 @@ def format_memories_table(
     table.add_column("Created", max_width=12)
 
     for mem in memories:
-        mem_id = mem.get("id", "")
-        memory_text = mem.get("memory", mem.get("text", ""))
+        mem_id = mem.get("id") or ""
+        memory_text = mem.get("memory") or mem.get("text") or ""
         if len(memory_text) > 60:
             memory_text = memory_text[:57] + "..."
         categories = mem.get("categories", [])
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- `cli/python/src/mem0_cli/output.py:23` / `:70` use the same simple `or ""` pattern already used elsewhere in this file, and it fixes the two reported list/table crashes cleanly.
- `cli/python/tests/test_output.py:57` and `:78` add focused regression coverage that fails without the implementation.

### Issues
#### Critical (must fix)
None.

#### Important (should fix)
1. `cli/python/src/mem0_cli/output.py:107` — `format_single_memory()` still uses the old `mem.get("memory", mem.get("text", ""))` pattern, so an explicit JSON `null` in `memory` still bypasses the fallback and renders as `None` instead of an empty/fallback value. Issue #5931 describes null `id` / `memory` / `text` fields as missing-equivalent formatter data, and this leaves the same root cause in the sibling single-memory formatter. Please apply the same guard here (`mem.get("memory") or mem.get("text") or ""`, and `mem.get("id") or ""`) and add the matching regression test.

#### Minor (optional)
- There are overlapping PRs for the same issue: #5932 covers this same change plus the single-memory formatter, while #5943 is broader but currently has a real Python CLI lint failure. If you keep this PR moving, the single-memory guard above is the missing piece.

### Evidence
- **CI:** Real checks are green for this PR: `CI Gate`, `Python CLI / lint`, `Python CLI / build`, and Python CLI tests on 3.10/3.11/3.12 passed. Ignored `Vercel` deploy authorization failure as advisory.
- **Targeted tests:** `PYTHONPATH=cli/python/src pytest cli/python/tests/test_output.py -q` → `24 passed`.
- **Red-green proof:** With only `cli/python/src/mem0_cli/output.py` restored to `origin/main`, the two new tests fail with the reported `None` crashes; restoring this PR makes the same test file pass again.
- **Lint:** `ruff check cli/python/src/mem0_cli/output.py cli/python/tests/test_output.py` → `All checks passed!`
- **Linked issue:** #5931 — this PR addresses the list/table crash, but not the sibling single-memory formatter path.
- **Could NOT verify:** No live API behavior needed; this is formatter-only and reproduced locally with direct unit tests / code tracing.

### Assessment
Ready to merge: **With fixes** — please cover the remaining single-memory formatter path; I’ll re-review once you push.

kartik-mem0 (approved):

PR #5964, opened 2026-06-29
fix(cli): return all entity delete responses

### LINKED ISSUE #5935: bug(cli): `entity delete` returns only the last entity's response for multi-entity deletes

barry166:
## Summary

- Return one response per deleted entity type from PlatformBackend.delete_entities
- Preserve the existing dict return shape for JSON output
- Add regression coverage for multi-entity deletes

Fixes #5935.

## Testing

- PYTHONPATH=cli/python/src pytest cli/python/tests/test_commands.py -q
- PYTHONPATH=cli/python/src pytest cli/python/tests/test_output.py -q
- python -m compileall -q cli/python/src/mem0_cli cli/python/tests
- git diff --check

## Notes

AI-assisted implementation; manually reviewed the diff and local verification output before opening this PR.

### DIFF
--- a/cli/python/src/mem0_cli/backend/platform.py
+++ b/cli/python/src/mem0_cli/backend/platform.py
@@ -297,12 +297,12 @@ class PlatformBackend(Backend):
         if not entities:
             raise ValueError("At least one entity ID is required for delete_entities.")
         # Delete each provided entity via the v2 path-based endpoint
-        result: dict = {}
+        results: dict[str, Any] = {}
         for entity_type, entity_id in entities.items():
-            result = self._request(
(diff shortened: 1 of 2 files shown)


barry166:
Closing this as a duplicate of the earlier PRs #5936 and #5942 for #5935. Sorry for the noise.

PR #5252, opened 2026-05-25
perf(embeddings): native embed_batch for all providers; fix FastEmbed embed() return type

wlmchen:
## Description

Add native `embed_batch()` to all embeddings providers that support batch APIs. Previously, every provider without an explicit override fell back to the base class which calls `embed()` sequentially — one API call per text. This PR completes coverage across all providers and fixes a pre-existing bug in FastEmbed.

**Providers with native batch added:**
- **HuggingFace** — TEI/base_url mode via OpenAI-compatible batch API; local SentenceTransformer via `encode([...])`
- **Together** — OpenAI-compatible batch API, chunked at 100
- **Ollama** — native `input=[...]` batch support
- **Gemini** — `embed_content` with 100-item chunks
- **VertexAI** — `get_embeddings(texts=[...])`, chunked at 250 (gemini-embedding-001 limit)
- **FastEmbed** — `TextEmbedding.embed([...])` accepts a list natively, single inference pass
- **LMStudio** — OpenAI-compatible batch API, chunked at 100
- **LangChain** — delegates to `embed_documents()`, unlocking whatever batch support the underlying provider has
- **AWS Bedrock** — Cohere models support native `"texts": [...]` batching (chunked at 96); Titan and other providers fall back to sequential since they have no batch API

**Bug fix — FastEmbed `embed()` return type:**
`embed()` was returning a raw numpy array instead of `list[float]`, inconsistent with every other provider and the base class contract. Fixed by adding `.tolist()`.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added 34 new unit tests across. All 73 embedding tests pass locally.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed (no public API changes, internal optimization only)

PR #5966, opened 2026-06-29
fix(embeddings): guard embed_batch count mismatch in OpenAI and Azure OpenAI

Bartok9:
## Summary

- OpenAI and Azure OpenAI were the only two batch-capable embedders missing the `embed_batch()` length-mismatch guard.
- A short/partial batch now raises `ValueError` instead of silently dropping extracted memories.

## Root Cause

**Symptom** — When the embedding provider returns fewer embeddings than the number of input texts (partial batch, throttle/5xx, malformed response), the missing texts are silently dropped during `Memory.add()`. No exception reaches the caller, so the application never learns a fact was not persisted.

**Root cause** — Every other batch-capable embedder validates `len(embeddings) != len(texts)` and raises (added across providers in #5609, and for the entity path in #5604). `OpenAIEmbedding.embed_batch()` (`mem0/embeddings/openai.py`) and `AzureOpenAIEmbedding.embed_batch()` (`mem0/embeddings/azure_openai.py`) were never given that guard. Downstream, the V3 add pipeline does `dict(zip(mem_texts, mem_embeddings_list))` — `zip()` truncates to the shorter list, so a short batch maps memories to nothing and they vanish. This is the same silent-memory-loss class tracked in #5245.

**Evidence** — The two `embed_batch` implementations `return all_embeddings` directly with no count check, while the sibling implementations (`gemini.py`, `together.py`, `lmstudio.py`, `vertexai.py`, `huggingface.py`, `ollama.py`) all raise a `ValueError` with the message `"... returned N embeddings for M texts ..."`. The regression tests below reproduce the drop on a mocked short batch.

**Fix + why this level** — Add the same guard at the end of each `embed_batch()`, mirroring the sibling message format. Fixing here (at the embedder boundary, where the count contract is known) fails loudly at the source rather than letting the truncation propagate silently into `zip()` downstream — consistent with how the other providers already behave.

**Scope / risk** — Two `embed_batch` methods only. The happy path (provider returns one embedding per input) is unchanged. Only the previously-silent short-batch case now raises. `embed()` (single) is untouched.

## Verification

- `pytest tests/embeddings/test_openai_embeddings.py tests/embeddings/test_azure_openai_embeddings.py` — 19 passed.

## Real behavior proof

- **Behavior addressed:** provider returns 1 embedding for a 2-text batch.
- **Real environment:** Python 3.13 venv, editable install of this branch.
- **Regression tests:** `test_embed_batch_count_mismatch_raises` in both test files; each asserts the new `ValueError` and FAILS without the fix.
- **Captured output — tests run AGAINST main (source reverted, tests kept):**
  ```
  >       with pytest.raises(ValueError, match="returned 1 embeddings for 2 texts"):
  E       Failed: DID NOT RAISE ValueError
  FAILED tests/embeddings/test_openai_embeddings.py::test_embed_batch_count_mismatch_raises
  FAILED tests/embeddings/test_azure_openai_embeddings.py::test_embed_batch_count_mismatch_raises
  2 failed in 0.81s
  ```
- **Captured output — full suites WITH the fix:**
  ```
  19 passed in 1.27s
  ```

## Related

Completes the embedder-consistency work from #5609 / #5604 for the two OpenAI embedders; mitigates the silent-loss class in #5245. Distinct from open PRs #5252 / #5922 (FastEmbed/AWS Bedrock/Langchain native batch) — those do not touch `openai.py` or `azure_openai.py`.

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/azure.ts
+++ b/mem0-ts/src/oss/src/embeddings/azure.ts
@@ -52,6 +52,11 @@ export class AzureOpenAIEmbedder implements Embedder {
           .map((item) => item.embedding),
       );
     }
+    if (allEmbeddings.length !== texts.length) {
+      throw new Error(
+        `Azure OpenAI embedBatch() returned ${allEmbeddings.length} embeddings for ${texts.length} texts using model '${this.model}'`,
+      );
+    }
     return allEmbeddings;
   }
 }
--- a/mem0-ts/src/oss/src/embeddings/openai.ts
+++ b/mem0-ts/src/oss/src/embeddings/openai.ts
@@ -47,6 +47,11 @@ export class OpenAIEmbedder implements Embedder {
           .map((item) => item.embedding),
       );
     }
+    if (allEmbeddings.length !== texts.length) {
+      throw new Error(
+        `OpenAI embedBatch() returned ${allEmbeddings.length} embeddings for ${texts.length} texts using model '${this.model}'`,
+      );
+    }
     return allEmbeddings;
   }
 }
--- a/mem0-ts/src/oss/tests/azure-embedder.test.ts
+++ b/mem0-ts/src/oss/tests/azure-embedder.test.ts
@@ -134,6 +134,19 @@ describe("AzureOpenAIEmbedder (unit)", () => {
       expect(result).toEqual(batch);
     });
 
+    it("embedBatch() throws when provider returns fewer embeddings than texts", async () => {
+      // Provider returns only 1 embedding for 2 texts (short-batch response).
+      mockEmbeddingsCreate.mockResolvedValue({
+        data: [{ index: 0, embedding: mockEmbedding }],
+      });
+
+      const embedder = new AzureOpenAIEmbedder(baseConfig);
+
+      await expect(embedder.embedBatch(["text1", "text2"])).rejects.toThrow(
+        /returned 1 embeddings for 2 texts/,
+      );
+    });
+
     it("uses custom model when provided", async () => {
       const embedder = new AzureOpenAIEmbedder({
         ...baseConfig,
--- a/mem0-ts/src/oss/tests/openai-embedder.test.ts
+++ b/mem0-ts/src/oss/tests/openai-embedder.test.ts
@@ -142,6 +142,21 @@ describe("OpenAIEmbedder (unit)", () => {
       expect(result).toEqual(batch);
     });
 
+    it("embedBatch() throws when provider returns fewer embeddings than texts", async () => {
+      // Provider returns only 1 embedding for 2 texts (short-batch response).
+      mockEmbeddingsCreate.mockResolvedValue({
+        data: [{ index: 0, embedding: mockEmbedding }],
+      });
+
+      const embedder = new OpenAIEmbedder({
+        apiKey: "test-key",
+      });
+
+      await expect(embedder.embedBatch(["text1", "text2"])).rejects.toThrow(
+        /returned 1 embeddings for 2 texts/,
+      );
+    });
+
     it("uses custom model when provided", async () => {
       const embedder = new OpenAIEmbedder({
         apiKey: "test-key",
--- a/mem0/embeddings/azure_openai.py
+++ b/mem0/embeddings/azure_openai.py
@@ -69,4 +69,9 @@ class AzureOpenAIEmbedding(EmbeddingBase):
                 model=self.config.model,
             )
             all_embeddings.extend(item.embedding for item in sorted(response.data, key=lambda x: x.index))
+        if len(all_embeddings) != len(texts):
+            raise ValueError(
+                f"Azure OpenAI embed_batch() returned {len(all_embeddings)} embeddings for {len(texts)} texts"
+                f" using model '{self.config.model}'"
+            )
         return all_embeddings
--- a/mem0/embeddings/openai.py
+++ b/mem0/embeddings/openai.py
@@ -73,4 +73,9 @@ class OpenAIEmbedding(EmbeddingBase):
                 kwargs["dimensions"] = self.config.embedding_dims
             response = self.client.embeddings.create(**kwargs)
             all_embeddings.extend(item.embedding for item in sorted(response.data, key=lambda x: x.index))
+        if len(all_embeddings) != len(texts):
+            raise ValueError(
+                f"OpenAI embed_batch() returned {len(all_embeddings)} embeddings for {len(texts)} texts"
+                f" using model '{self.config.model}'"
+            )
         return all_embeddings
--- a/tests/embeddings/test_azure_openai_embeddings.py
+++ b/tests/embeddings/test_azure_openai_embeddings.py
@@ -164,3 +164,30 @@ def test_init_with_placeholder_api_key(monkeypatch, base_embedder_config):
             http_client=None,
             default_headers=None,
         )
+
+
+def test_embed_batch_returns_all_embeddings(mock_openai_client):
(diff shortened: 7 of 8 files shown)


kartik-mem0:
## Changes Requested

### Strengths
- `mem0/embeddings/openai.py:76` and `mem0/embeddings/azure_openai.py:72` put the count check at the provider boundary, which matches the existing sibling-provider contract and fails before `Memory.add()` can silently truncate via `zip()`.
- The new tests in `tests/embeddings/test_openai_embeddings.py:142` and `tests/embeddings/test_azure_openai_embeddings.py:184` are good regression tests: they fail against `origin/main` and pass with this PR restored.

### Issues
#### Critical (must fix)
None.

#### Important (should fix)
1. `mem0-ts/src/oss/src/embeddings/openai.ts:31` and `mem0-ts/src/oss/src/embeddings/azure.ts:37` — the TypeScript OSS OpenAI/Azure embedders still return `allEmbeddings` without checking `allEmbeddings.length === texts.length`. The downstream TS add path indexes those results back onto `memTexts` at `mem0-ts/src/oss/src/memory/index.ts:907`, so the same short-batch provider response remains unguarded in the parallel SDK. Please add the same guard plus mismatch regression tests for the TS OpenAI and Azure embedders, or link a companion PR that fixes that before this lands.

#### Minor (optional)
None.

### Evidence
- **CI:** real checks are green: `CI Gate` passed; Python SDK build jobs on 3.10, 3.11, and 3.12 passed; CLA passed. Ignored the Vercel deploy-auth failure because fork preview deploys are advisory here.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/embeddings/test_openai_embeddings.py tests/embeddings/test_azure_openai_embeddings.py -v --tb=short` -> 19 passed in 3.21s.
- **Red-green proof:** with only `mem0/embeddings/openai.py` and `mem0/embeddings/azure_openai.py` reverted to `origin/main`, both new mismatch tests failed with `Failed: DID NOT RAISE ValueError`; after restoring the PR implementation, both passed.
- **Lint:** `hatch -e dev_py_3_12 run ruff check mem0/embeddings/openai.py mem0/embeddings/azure_openai.py tests/embeddings/test_openai_embeddings.py tests/embeddings/test_azure_openai_embeddings.py` -> All checks passed.
- **Linked issue:** no closing issue reference; this relates to #5245, but #5245 is broader than this provider-level guard.
- **Duplicate/overlap check:** open PRs #5252 and #5922 do not touch OpenAI/Azure; broader #5245/#5509 PRs target memory-pipeline behavior, not this exact provider count guard. No superseding PR found for the TS OpenAI/Azure counterpart.
- **Cross-package check:** Python OpenAI/Azure fixed here; TypeScript OpenAI/Azure counterpart remains uncovered.

### Assessment
Ready to merge: **With fixes** — the Python patch is clean and proven, but merging it as-is leaves the same provider-contract bug in the parallel TypeScript SDK. I’ll re-review once the TS counterpart is covered or linked.

For tracking on our side, please link this PR to #5245 or the specific tracking issue you want associated with this fix.

Bartok9:
Thanks for the thorough review @kartik-mem0 — addressed the important issue.

**TS counterpart guarded (commit `cc7436fc`):**
- `mem0-ts/src/oss/src/embeddings/openai.ts` and `mem0-ts/src/oss/src/embeddings/azure.ts` now throw if `embedBatch()` returns fewer embeddings than input texts, matching the Python provider-boundary contract. This fails before the TS add path indexes results back onto `memTexts` in `mem0-ts/src/oss/src/memory/index.ts:907`.
- Added red-green regression tests in `mem0-ts/src/oss/tests/openai-embedder.test.ts` and `tests/azure-embedder.test.ts` (short-batch response → expect throw).

**Verification:**
- `npx jest src/oss/tests/openai-embedder.test.ts src/oss/tests/azure-embedder.test.ts` → 20 passed.
- Red-green proof: with only the two embedder source files reverted, both new mismatch tests fail (`returned 1 embeddings for 2 texts`); restored → pass.
- `npx prettier --check` on all four files → clean.

Now both the Python and TypeScript OpenAI/Azure embedders enforce the same count guard. Linking this PR to #5245 as requested.

kartik-mem0:

Verified against `cc7436fc42` (new head): the TS parity gap from the last review is fixed.

One new commit (`cc7436fc42`, "fix(embeddings-ts): guard embedBatch count mismatch in TS OpenAI and Azure OpenAI"), touching exactly the 4 files this was waiting on — nothing else moved.

### Strengths
- `mem0-ts/src/oss/src/embeddings/openai.ts:50` and `mem0-ts/src/oss/src/embeddings/azure.ts:55` now throw when `embedBatch()` returns fewer embeddings than input texts. The guard sits after the chunking loop assembles the full `allEmbeddings` array and before `return`, so it validates the TOTAL batch result — same boundary, same message convention (`"... returned N embeddings for M texts using model 'X'"`) as the Python guard at `mem0/embeddings/openai.py:76` / `azure_openai.py:72`. Python and TS OpenAI/Azure now enforce an identical contract; no cross-SDK divergence left for these two providers.
- The new tests in `openai-embedder.test.ts` and `azure-embedder.test.ts` mock the exact short-batch response (1 embedding for 2 texts) that causes the silent drop and assert the throw — same pattern as the Python regression tests, and it holds up under a fresh revert-and-restore this run (below).

### Issues
None blocking. FYI, non-blocking: this is the first count guard in the TS OSS embedder set — `google.ts`, `ollama.ts`, `lmstudio.ts`, and `langchain.ts` still lack it, mirroring a pre-existing Python→TS gap for those other providers. Out of scope here (this PR only ever targeted OpenAI/Azure parity, matching the Python fix's scope) — flagging for a possible future pass, not a blocker.

### Evidence
- **CI:** real checks green — `CI Gate`, Python SDK 3.10/3.11/3.12, TypeScript SDK `build_ts_sdk`/`integration_ts_sdk` on Node 20 & 22, CLA. Vercel deploy-auth failure ignored (advisory, fails on every fork PR).
- **Targeted tests:** `pnpm exec jest src/oss/tests/openai-embedder.test.ts src/oss/tests/azure-embedder.test.ts` → 20 passed, 2 suites (matches the 20 reported).
- **Red-green proof (this delta):** with only `openai.ts`/`azure.ts` reverted to the previously-reviewed SHA, exactly the 2 new tests fail ("received promise resolved instead of rejected"); the other 18 pre-existing tests in the same files are unaffected. Restoring the fix brings all 20 back to green; working tree clean after.
- **Format:** `prettier --check` on all 4 changed files → clean.
- **Python side:** unchanged since the last review (0 lines touched in this delta) — already red-green-proven previously.
- **Linked issue:** still no `Fixes/Closes` reference — #5245 is mentioned in the body but doesn't show up under GitHub's closing references.
- **Could NOT verify:** a live OpenAI/Azure short-batch response — both suites use mocked provider responses, same convention as the rest of this test file and the already-accepted Python tests.

### Assessment
Both SDKs now enforce the same embed-count contract for OpenAI and Azure OpenAI, proven end to end with a fresh red-green pass on this round's delta.

For tracking on our side, mind opening a quick issue describing this fix and linking it here (e.g. `Fixes #N`)? Makes it easier to prioritize and reference — no rush either way.

Bartok9:
Thanks @kartik-mem0 — glad the TS parity round is clean.

Opened a tracking issue as requested: #6030 (describes the silent-drop bug and the OpenAI/Azure count-guard fix across both SDKs). This PR now references it, and I'll link the future pass for the remaining TS embedders (`google`, `ollama`, `lmstudio`, `langchain`) to that issue as well.

Ready for squash-merge whenever you are.

kartik-mem0 (approved):

PR #5419, opened 2026-06-07
fix(openmemory): use top_k instead of limit in search_memory vector query

AjaasMk:
## Linked Issue

Closes #5404

## Description

In `openmemory/api/app/mcp_server.py`, `search_memory()` called the vector store with `limit=10`:

```python
hits = memory_client.vector_store.search(
    query=query,
    vectors=embeddings,
    limit=10,        # ← wrong kwarg
    filters=filters,
)
```

The Mem0 vector store interface uses `top_k`, not `limit` (see `mem0/vector_stores/base.py` — `def search(self, query, vectors, top_k=5, filters=None)` — and every concrete store). Passing `limit` does not constrain the result count to 10, so the intended cap is not honored.

This PR renames the kwarg to `top_k=10`.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added `TestSearchMemoryVectorStoreCall::test_search_passes_top_k_not_limit` to `openmemory/api/tests/test_mcp_server.py`, which invokes `search_memory()` with the memory client and DB mocked and asserts the vector store is called with `top_k=10` and no `limit` kwarg. It fails before the fix and passes after.

```
$ pytest tests/test_mcp_server.py -q
22 passed
```

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed (none needed — internal behavior fix)

PR #5410, opened 2026-06-07
fix(openmemory/mcp): pass top_k kwarg instead of limit to vector_store.search()

oyaah:
## Linked Issue

Closes #5404

## Description

`search_memory()` in `openmemory/api/app/mcp_server.py` called `vector_store.search(..., limit=10, ...)`, but the vector store `search()` API expects `top_k`. The `limit` kwarg was silently ignored (or rejected, depending on the backend), so searches did not honor the intended result cap.

This passes `top_k=10` instead.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added `tests/test_mcp_server.py::TestSearchMemoryKwargs::test_search_uses_top_k_not_limit`, which mocks the vector store and asserts `search()` is called with `top_k` and **not** `limit`. Verified locally:

```
$ python -m pytest tests/test_mcp_server.py -q
22 passed in 1.06s
```

Negative check — reverting the fix to `limit=10` makes the new test fail (`assert 'top_k' in {... 'limit': 10 ...}`), confirming it guards the regression.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed

---

Split out from #5409 (which bundled this with the `embedding_dims` fix) into a focused, tested, one-issue-per-PR change. The `embedding_dims` half is #5405 in a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5097, opened 2026-05-09
fix(openmemory): use public mem0 APIs for MCP reads

hobbesd:
## Linked Issue

N/A

## Description

Fixes two OpenMemory MCP tools so they use the public mem0 memory client APIs for user-scoped reads:

- `search_memory` now calls `memory_client.search(query=..., top_k=10, filters={"user_id": uid})` and keeps returning the existing `{"results": [...]}` MCP response envelope.
- `list_memories` now calls `memory_client.get_all(filters={"user_id": uid})` and keeps returning the expected JSON array.

This avoids relying on mem0 internals such as direct embedding generation or vector-store search result shapes, while preserving the existing MCP client contract.

Prepared with AI assistance using OpenAI GPT-5.5 and Claude Opus. The changes were reviewed and tested locally before submission.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added focused unit coverage for the changed MCP tool paths, using stubs/mocks only and no external services:

- verifies `search_memory` calls `memory_client.search(query=..., top_k=10, filters={"user_id": uid})` and returns `{"results": [...]}`.
- verifies `list_memories` calls `memory_client.get_all(filters={"user_id": uid})` and returns the expected JSON array.

Local test command:

```bash
cd openmemory/api
pytest tests/test_mcp_server.py -q
```

Result:

```text
23 passed
```

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #5968, opened 2026-06-29
fix(openmemory): align MCP server with mem0 2.x API

Saqoosha:
## Summary

The OpenMemory MCP server still calls the pre-2.0 mem0ai API. Because `openmemory/api/requirements.txt` pins only `mem0ai>=0.1.92`, any fresh install today resolves to mem0ai 2.x and breaks the `search_memory` and `list_memories` MCP tools at runtime.

This patches the two affected call sites in `mcp_server.py`. Saves are unaffected — they go through `Memory.add`, whose signature is stable.

## Errors before the fix

```
search_memory  -> Qdrant.search() got an unexpected keyword argument 'limit'
list_memories  -> Top-level entity parameters frozenset({'user_id'}) are not
                  supported in get_all(). Use filters={'user_id': '...'} instead.
```

## Root cause

mem0 2.x renamed/refactored two public methods used by the MCP server:

| Method | Before | After |
| --- | --- | --- |
| `vector_store.search(...)` | `limit=N` | `top_k=N` |
| `Memory.get_all(...)` | `get_all(user_id="...")` | `get_all(filters={"user_id": "..."})` |

Verified against mem0ai 2.0.2 + qdrant-client 1.18:

```python
>>> inspect.signature(Memory.get_all)
(self, *, filters: Optional[Dict[str, Any]] = None, top_k: int = 20, **kwargs)
>>> inspect.signature(Qdrant.search)
(self, query: str, vectors: list, top_k: int = 5, filters: dict = None) -> list
```

## Fix

Two-line change in `openmemory/api/app/mcp_server.py`:

- `search_memory`: `limit=10` → `top_k=10`
- `list_memories`: `get_all(user_id=uid)` → `get_all(filters={"user_id": uid})`

## Verification

Tested end-to-end against a self-hosted OpenMemory stack (Qdrant 4337 points, mem0ai 2.0.2):

- `list_memories` now returns the full memory list
- `search_memory("...")` returns semantically ranked hits
- `Memory.add` (save worker path) confirmed unaffected — it was already working before this fix, and still works after

## Test plan

- [ ] CI passes
- [ ] Verified `search_memory` and `list_memories` against mem0ai 2.x in a self-hosted deployment

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -177,9 +177,9 @@ async def search_memory(query: str) -> str:
             embeddings = memory_client.embedding_model.embed(query, "search")
 
             hits = memory_client.vector_store.search(
-                query=query, 
-                vectors=embeddings, 
-                limit=10, 
+                query=query,
+                vectors=embeddings,
+                top_k=10,
                 filters=filters,
             )
 
@@ -245,7 +245,7 @@ async def list_memories() -> str:
             user, app = get_user_and_app(db, user_id=uid, app_id=client_name)
 
             # Get all memories
-            memories = memory_client.get_all(user_id=uid)
+            memories = memory_client.get_all(filters={"user_id": uid})
             filtered_memories = []
 
             # Filter memories based on permissions


Saqoosha:
Closing this — I missed that OpenMemory has been deprecated and that this exact fix was already submitted in #5419, #5407, #5410 (top_k) and #5050, #5097 (filters dict), all closed for the same reason.

Sorry for the noise.

PR #5973, opened 2026-06-29
fix(server): ignore empty-string entity ids when building memory filters

### LINKED ISSUE #5972: bug(server): GET /memories forwards empty-string entity ids as filters
### Component

Other (self-hosted `server/` — REST API)

### Description

`GET /memories` (`get_all_memories` in `server/main.py`) treats empty-string entity ids inconsistently between its two checks.

The "list all" guard uses a truthiness check, so empty strings count as "not provided":

```python
if not any([user_id, run_id, agent_id]):
    ...  # admin-only list-all path
```

But the filter dict built immediately after keeps any value that is not `None`, so an empty string passes through as a real filter:

```python
filters = {
    k: v for k, v in {"user_id": user_id, "run_id": run_id, "agent_id": agent_id}.items() if v is not None
}
```

So a request like `GET /memories?user_id=alice&agent_id=` sends `{"user_id": "alice", "agent_id": ""}` to `get_all`, scoping the query to an empty `agent_id` the caller never meaningfully supplied. Empty query-string params are easy to produce from forms/clients, and the two checks disagree on what "provided" means.

### Steps to Reproduce

```
GET /memories?user_id=alice&agent_id=&run_id=
```

The resulting filters include `agent_id: ""` and `run_id: ""` instead of just `user_id: alice`.

### Expected Behavior

Empty-string ids are ignored when building filters, consistent with the `any([...])` guard, so only real values are forwarded.

### Actual Behavior

Empty-string ids are forwarded as filter values, over-scoping the query.

### Environment

- Self-hosted server, current `main`
- platform-independent (pure request-handling logic)

abhay-codes07:
## Linked Issue

Closes #5972

## Description

`GET /memories` (`get_all_memories`) is inconsistent about empty-string entity ids.

The list-all guard uses truthiness, so an empty string means "not provided":

```python
if not any([user_id, run_id, agent_id]):
    ...  # admin-only list-all path
```

But the filters dict built immediately after keeps anything that isn't `None`:

```python
filters = {k: v for k, v in {...}.items() if v is not None}
```

So `GET /memories?user_id=alice&agent_id=` ends up sending `{"user_id": "alice", "agent_id": ""}` to `get_all`, scoping the query to an empty `agent_id` the caller never meaningfully provided. Empty query params are easy to send from forms and clients, and the two checks disagree on what counts as "provided".

The fix changes the filter comprehension from `if v is not None` to `if v`, so empty strings are dropped, matching the `any([...])` guard right above it. Real ids behave exactly as before.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] No tests needed

There were no server tests yet, so this adds a small `server/tests/` package (a `conftest.py` that sets safe env defaults so `main` imports without a live Postgres/provider, plus the test). The test calls `get_all_memories` with a real `user_id` and empty-string `agent_id`/`run_id` and asserts only the real id is forwarded. It fails without the fix and passes with it.

## Checklist

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -426,7 +426,7 @@ def get_all_memories(
             # Admin all-memory listing is intentionally raw; scoped get_all below applies expiry visibility.
             return _list_all_memories(limit=top_k if top_k is not None else ALL_MEMORIES_LIMIT)
         filters = {
-            k: v for k, v in {"user_id": user_id, "run_id": run_id, "agent_id": agent_id}.items() if v is not None
+            k: v for k, v in {"user_id": user_id, "run_id": run_id, "agent_id": agent_id}.items() if v
         }
         params = {"filters": filters}
         if top_k is not None:
--- a/server/main.py
+++ b/server/main.py
@@ -456,7 +456,7 @@ def search_memories(search_req: SearchRequest, _auth=Depends(verify_auth)):
         deprecated_keys = []
         for entity_key in ("user_id", "agent_id", "run_id"):
             entity_val = getattr(search_req, entity_key, None)
-            if entity_val is not None:
+            if entity_val:
                 filters[entity_key] = entity_val
                 deprecated_keys.append(entity_key)
         if deprecated_keys:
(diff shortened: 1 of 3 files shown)


abhay-codes07:
@kartik-mem0 small server fix with a regression test (adds the first server/tests). Quick review when you can, thanks.

abhay-codes07:
Good catch on the triage note. I extended this PR to cover the same is-not-None pattern in /search (search_memories) and DELETE /memories (delete_all_memories), both now use a truthiness check like GET /memories. Tests cover all three endpoints. The add-payload builder at line 372 is left as-is since that one intentionally keeps provided fields and isn't entity-id scoping.

abhay-codes07:
Superseded by #5992, which landed the same fix across all three endpoints. Closing this one out. Since server/ had no test coverage, I'll put up a small tests-only PR to guard the behavior against regressions. Thanks!

PR #5980, opened 2026-06-29
fix(elasticsearch): validate filter keys and values to prevent term injection

### LINKED ISSUE #5976: Elasticsearch vector store: filter term injection via dict/list values
## Description

The `ElasticsearchVectorStore` builds Elasticsearch term queries by directly embedding filter values into the query dict without type validation. If a caller passes a dict or list as a filter value (e.g. `{"user_id": {"$ne": ""}}`), it gets inserted verbatim into the term query, allowing an attacker to inject arbitrary Elasticsearch query operators.

## Impact

An attacker who controls filter values can bypass access controls (e.g. enumerate other users' memories) or craft queries that return unintended results.

## Steps to Reproduce

```python
from mem0.vector_stores.elasticsearch import ElasticsearchVectorStore

# This dict value gets embedded directly into the term query
store.search(query="test", vectors=[...], filters={"user_id": {"$ne": ""}})
# Results in: {"term": {"metadata.user_id": {"$ne": ""}}}
# Instead of: {"term": {"metadata.user_id": "some_string"}}
```

## Suggested Fix

Validate that filter keys match `^[a-zA-Z_][a-zA-Z0-9_]*$` and filter values are scalar types (str, int, float, bool) before building the term query. Reject dict/list values with a `ValueError`.

HrushiYadav:
Closes #5976

## Summary

- Validate filter keys against `^[a-zA-Z_][a-zA-Z0-9_]*$` to prevent key injection
- Validate filter values are scalar types (str, int, float, bool) -- reject dict/list values that could inject Elasticsearch query operators
- Add `_validate_filter()` helper called before every `term` query construction in `search()`, `keyword_search()`, and `list()`

## Test plan

- [x] 6 new tests added covering key validation, value type rejection, and scalar acceptance
- [x] All 20 existing + new tests pass
- [x] `ruff check` clean

### DIFF
--- a/mem0/vector_stores/elasticsearch.py
+++ b/mem0/vector_stores/elasticsearch.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from typing import Any, Dict, List, Optional
 
 try:
@@ -19,6 +20,19 @@ class OutputData(BaseModel):
     id: str
     score: float
     payload: Dict
+
+
+_SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
+
+
+def _validate_filter(key: str, value: Any) -> None:
+    if not isinstance(key, str) or not _SAFE_FILTER_KEY.match(key):
+        raise ValueError(f"Invalid filter key: {key!r}")
+    if not isinstance(value, (str, int, float, bool)):
+        raise ValueError(
+            f"Filter value for {key!r} must be str, int, float, or bool, "
+            f"got {type(value).__name__}"
+        )
 
 
 class ElasticsearchDB(VectorStoreBase):
@@ -154,6 +168,7 @@ class ElasticsearchDB(VectorStoreBase):
             if filters:
                 filter_conditions = []
                 for key, value in filters.items():
+                    _validate_filter(key, value)
                     filter_conditions.append({"term": {f"metadata.{key}": value}})
                 search_query["knn"]["filter"] = {"bool": {"must": filter_conditions}}
 
@@ -192,6 +207,7 @@ class ElasticsearchDB(VectorStoreBase):
         if filters:
             filter_conditions = []
             for key, value in filters.items():
+                _validate_filter(key, value)
                 filter_conditions.append({"term": {f"metadata.{key}": value}})
             bool_query["filter"] = filter_conditions
 
@@ -262,6 +278,7 @@ class ElasticsearchDB(VectorStoreBase):
         if filters:
             filter_conditions = []
             for key, value in filters.items():
+                _validate_filter(key, value)
                 filter_conditions.append({"term": {f"metadata.{key}": value}})
             query["query"] = {"bool": {"must": filter_conditions}}
 
--- a/tests/vector_stores/test_elasticsearch.py
+++ b/tests/vector_stores/test_elasticsearch.py
@@ -10,7 +10,7 @@ except ImportError:
     raise ImportError("Elasticsearch requires extra dependencies. Install with `pip install elasticsearch`") from None
 
 from mem0.configs.vector_stores.elasticsearch import ElasticsearchConfig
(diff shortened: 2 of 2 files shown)


kartik-mem0:

### What this PR does
Closes #5976 (P1): adds a `_SAFE_FILTER_KEY` key-allowlist plus scalar value-type validation (`_validate_filter`) in `elasticsearch.py`, called before every `{"term": {...}}` construction in `search()`, `keyword_search()`, and `list()`.

### Strengths
- `elasticsearch.py:156/194/264` — filter keys come straight from caller-supplied `filters.items()` in all three methods (not hardcoded like OpenSearch's `["user_id","run_id","agent_id"]`), so the key-allowlist regex is genuinely load-bearing here, not decorative. This correctly matches Neptune's threat model (#5982) rather than OpenSearch's leaner one (#5986).
- No escaping-adequacy assumption needed, unlike #5981/Upstash: ES filter clauses are plain Python dicts serialized to JSON by `elasticsearch-py`, never string-interpolated, so rejecting non-scalar values structurally closes the term-query object-injection shape (`{"value": "*", "case_insensitive": true}` and friends) entirely — type-rejection alone is sufficient on its own terms.
- Validation is the first statement inside each loop, well before `self.client.search(...)`. Confirmed at runtime (mocked client, not just by reading) that both a dict-value payload and a malicious-key payload raise `ValueError` before any query reaches the client, in all three methods.
- Test payloads mirror ES's actual verbose term-query object form rather than generic placeholders — the fix clearly targets the real exploit mechanism, not just "reject any dict."
- Blast radius is wider than direct `ElasticsearchDB` use: traced `Memory`'s advanced-operator filter DSL in `mem0/memory/main.py` and confirmed it still produces dict-valued filters for its own recognized operator syntax (e.g. `{"field": {"ne": value}}`) — Elasticsearch never implemented operator-translation the way `pgvector`/`qdrant` do, so this fix is the only real enforcement point for that path too, not redundant defense-in-depth.

### Issues
Minor (optional): `keyword_search()` has no dedicated end-to-end dict/list-rejection test (only indirect coverage via the shared `_validate_filter` unit tests) — `search()`/`list()` both get one, worth closing for parity. The key regex anchors with `$` rather than `\Z`/`.fullmatch()`, so a key with a trailing newline slips through — not exploitable here since keys become JSON dict keys via the client's serializer rather than being string-interpolated, but a one-line fix. `_validate_filter` is now an independently-duplicated pattern across several vector-store files (baidu/milvus/mongodb/neptune/opensearch/upstash/elasticsearch) with slightly drifted regexes — pre-existing repo convention, worth a follow-up consolidation ticket once the rest of the cluster lands, not a blocker here.

### Evidence
- **CI:** CI Gate pass; Python SDK build_mem0 (3.10/3.11/3.12) pass; changelog_check/check_changes pass; license/cla signed. (Vercel preview-deploy check fails on authorization, unrelated to this PR, ignored per convention.)
- **Targeted tests:** `pytest tests/vector_stores/test_elasticsearch.py` → 20 passed (14 pre-existing + 6 new).
- **Red-green proof:** reverted the production hunk only → new tests fail (import error on the missing `_validate_filter` symbol); restored → 20 passed again ✓
- **Linked issue:** #5976 — addresses the reported root cause directly, matching the issue's suggested remediation almost verbatim.
- **Cross-package check:** no TS counterpart exists for Elasticsearch (`mem0-ts/src/oss/src/vector_stores/` has none) — unaffected.
- **Could NOT verify:** whether #5976's "enumerate other users' memories" framing would fully succeed against a live (non-mocked) Elasticsearch cluster on every version — reasoned from the ES term-query DSL and mocked-client runtime traces only.

### Assessment
Same validated shape as the already-merged Neptune fix, correctly applied because Elasticsearch shares Neptune's threat model (iterated, caller-controlled keys) rather than OpenSearch's (hardcoded keys), and provably sufficient on its own terms without an unverifiable escaping assumption.

kartik-mem0 (approved):

PR #5981, opened 2026-06-29
fix(upstash): escape quotes in filter values and validate filter types

### LINKED ISSUE #5977: Upstash vector store: filter expression injection via unescaped quotes
## Description

The `UpstashVector._stringify()` method wraps string values in double quotes but does not escape embedded quotes. A filter value containing `"` can break out of the string literal and inject arbitrary filter expressions.

Additionally, filter keys and values are not type-validated, so dict/list values pass through silently.

## Impact

An attacker who controls filter values can inject arbitrary Upstash filter expressions, potentially bypassing access control filters or altering query semantics.

## Steps to Reproduce

```python
from mem0.vector_stores.upstash_vector import UpstashVector

# The embedded quote breaks out of the string literal
store.search(
    query="test",
    vectors=[[0.1, 0.2]],
    filters={"name": 'x" OR user_id = "admin'}
)
# Produces: name = "x" OR user_id = "admin"
# Instead of: name = "x\" OR user_id = \"admin\""
```

## Suggested Fix

1. Escape `\` and `"` in `_stringify()` before wrapping in quotes.
2. Validate filter keys match `^[a-zA-Z_][a-zA-Z0-9_]*$` and values are scalar types.

HrushiYadav:
Closes #5977

## Summary

- Fix `_stringify()` to escape `\` and `"` in string values before wrapping in double quotes
- Add `_validate_filter()` to reject dict/list filter values and invalid keys
- Validation applied before all 3 filter construction points (`search()`, `keyword_search()`, `list()`)

## Test plan

- [x] 6 new tests added covering quote escaping, type rejection, and scalar acceptance
- [x] All 25 existing + new tests pass
- [x] `ruff check` clean

### DIFF
--- a/mem0/vector_stores/upstash_vector.py
+++ b/mem0/vector_stores/upstash_vector.py
@@ -1,5 +1,6 @@
 import logging
-from typing import Dict, List, Optional
+import re
+from typing import Any, Dict, List, Optional
 
 from pydantic import BaseModel
 
@@ -12,6 +13,23 @@ except ImportError:
 
 
 logger = logging.getLogger(__name__)
+
+_SAFE_FILTER_KEY = re.compile(r"[a-zA-Z_][a-zA-Z0-9_]*\Z")
+
+
+def _validate_filter(key: str, value: Any) -> None:
+    if not isinstance(key, str) or not _SAFE_FILTER_KEY.fullmatch(key):
+        raise ValueError(f"Invalid filter key: {key!r}")
+    if not isinstance(value, (str, int, float, bool)):
+        raise ValueError(
+            f"Filter value for {key!r} must be str, int, float, or bool, "
+            f"got {type(value).__name__}"
+        )
+    if isinstance(value, str) and ('"' in value or "\\" in value):
+        raise ValueError(
+            f"Filter value for {key!r} contains prohibited characters "
+            f"(double quote or backslash): {value!r}"
+        )
 
 
 class OutputData(BaseModel):
@@ -92,7 +110,9 @@ class UpstashVector(VectorStoreBase):
         )
 
     def _stringify(self, x):
-        return f'"{x}"' if isinstance(x, str) else x
+        if isinstance(x, str):
+            return f'"{x}"'
+        return x
 
     def search(
         self,
@@ -113,6 +133,9 @@ class UpstashVector(VectorStoreBase):
             List[OutputData]: Search results.
         """
 
+        if filters:
+            for k, v in filters.items():
+                _validate_filter(k, v)
         filters_str = " AND ".join([f"{k} = {self._stringify(v)}" for k, v in filters.items()]) if filters else None
 
         response = []
@@ -160,13 +183,16 @@ class UpstashVector(VectorStoreBase):
         Returns:
             List[OutputData]: Search results, or None if sparse/BM25 search is not supported.
         """
-        try:
-            filters_str = (
-                " AND ".join([f"{k} = {self._stringify(v)}" for k, v in filters.items()])
-                if filters
-                else None
-            )
+        if filters:
+            for k, v in filters.items():
+                _validate_filter(k, v)
+        filters_str = (
+            " AND ".join([f"{k} = {self._stringify(v)}" for k, v in filters.items()])
+            if filters
+            else None
+        )
 
+        try:
             response = self.client.query(
                 data=query,
                 top_k=top_k,
@@ -252,6 +278,9 @@ class UpstashVector(VectorStoreBase):
         Returns:
             List[OutputData]: Search results.
         """
+        if filters:
+            for k, v in filters.items():
+                _validate_filter(k, v)
         filters_str = " AND ".join([f"{k} = {self._stringify(v)}" for k, v in filters.items()]) if filters else None
 
         info = self.client.info()
--- a/tests/vector_stores/test_upstash_vector.py
+++ b/tests/vector_stores/test_upstash_vector.py
@@ -5,7 +5,7 @@ from unittest.mock import MagicMock, call, patch
 import pytest
 
 from mem0.configs.vector_stores.upstash_vector import UpstashVectorConfig
-from mem0.vector_stores.upstash_vector import UpstashVector
+from mem0.vector_stores.upstash_vector import UpstashVector, _validate_filter
 
 
 @dataclass
@@ -226,6 +226,57 @@ def test_update_vector_with_embeddings(upstash_instance_with_embeddings):
         metadata={"name": "updated_vector", "data": "updated_data"},
         namespace="ns",
     )
+
+
+def test_filter_rejects_dict_value():
+    with pytest.raises(ValueError):
+        _validate_filter("user_id", {"$ne": ""})
+
+
+def test_filter_rejects_list_value():
+    with pytest.raises(ValueError):
+        _validate_filter("user_id", ["alice", "bob"])
+
+
+def test_filter_rejects_invalid_key():
+    with pytest.raises(ValueError):
+        _validate_filter("user_id; DROP", "alice")
+
+
+def test_filter_accepts_scalars():
+    _validate_filter("user_id", "alice")
+    _validate_filter("count", 42)
+    _validate_filter("score", 0.95)
+    _validate_filter("active", True)
+
+
+def test_filter_rejects_double_quote_in_value():
+    with pytest.raises(ValueError, match="prohibited characters"):
+        _validate_filter("user_id", 'alice" OR 1=1 --')
+
+
+def test_filter_rejects_backslash_in_value():
+    with pytest.raises(ValueError, match="prohibited characters"):
+        _validate_filter("user_id", "alice\\bob")
+
+
+def test_search_rejects_dict_filter(upstash_instance):
+    with pytest.raises(ValueError):
+        upstash_instance.search(
+            query="test", vectors=[[0.1]], filters={"user_id": {"$ne": ""}}
+        )
+
+
+def test_keyword_search_raises_on_invalid_filter(upstash_instance):
+    with pytest.raises(ValueError):
+        upstash_instance.keyword_search(
+            query="test", filters={"user_id": 'alice" OR 1=1'}
+        )
+
+
+def test_filter_rejects_key_with_trailing_newline():
+    with pytest.raises(ValueError, match="Invalid filter key"):
+        _validate_filter("user_id\n", "alice")
 
 
 def test_insert_vectors_with_embeddings_missing_data(upstash_instance_with_embeddings):


kartik-mem0:
Reviewed this against the root cause in #5977 (unescaped quotes in `_stringify()` letting a filter value break out of the double-quoted string and OR-in an always-true clause).

**What's solid**

- All three filter-construction sites (`search`, `keyword_search`, `list`) now run through `_validate_filter`, and all three share the same `_stringify`, so this isn't a one-off patch on just the reported path.
- Dict/list filter values are now rejected outright. Checked whether that breaks anything real: `memory/main.py`'s `_process_metadata_filters` does emit nested operator dicts (`{"gte": 10}`, `$or`, `$not`), but Upstash never had any translation for those — pre-fix, a dict value fell through `_stringify`'s `else` branch and got `str()`-interpolated into a syntactically invalid filter that would've failed server-side anyway. So this is a fail-fast improvement, not a regression.
- Red/green check: reverted only `mem0/vector_stores/upstash_vector.py` to main, and a value like `alice" OR 1=1 --` produces `user_id = "alice" OR 1=1 --"` in the constructed filter — a genuine always-true breakout, reproduced through the actual `search()` call, not just `_stringify` in isolation. Restoring the fix, the full suite (25/25) passes and `ruff check` is clean.

**What's blocking for me**

The fix's entire security value rests on backslash-escaping (`\"`, `\\`) being honored by Upstash's server-side filter parser as an actual escape sequence. I can't find that documented anywhere — checked both the Vector and Search filtering docs (they share the same grammar); the only statement about quoting is "string literals can be single or double quoted," with no mention of how to get a literal quote character inside a value. No backslash convention, no doubling convention, nothing. The Python client (`vector-py`) doesn't help either — it forwards `filter` as an opaque string, so there's no vendor-side utility to confirm the convention against.

That matters because `test_stringify_escapes_quotes` only asserts the Python transformation equals an expected Python string — it never touches Upstash, so it can't prove the vulnerability is actually closed server-side. If the real parser doesn't special-case backslash (a reasonable default for an undocumented, presumably simple tokenizer), the fix doesn't change anything: a naive "scan for the next quote character" tokenizer terminates the string at the same position regardless of a preceding backslash, so `"alice\" OR 1=1 --"` still hands the parser `alice\` as the string content and ` OR 1=1 --"` as trailing filter syntax — same break-out, just with a stray backslash in front of it. I replayed the escaped output through that exact scan and got the same result.

Given `user_id`/`agent_id`/`run_id` only get whitespace trimming upstream (`_validate_and_trim_entity_id` in `memory/main.py`) with no quote handling at all, this vector-store-level escape is the only thing standing between a crafted value and cross-tenant exfiltration. For a P0, I don't think we can ship that on an unverified assumption about a third-party parser.

To get to LGTM, one of:
- A live round-trip against a real Upstash index showing a value containing `"`/`\`/an injected `OR` clause matches only itself and can't widen the result set, or
- Confirmation from Upstash (docs/support) of the actual escape convention, or
- Failing closed instead: reject string values containing `"` or `\` (same posture as the new dict/list rejection) rather than escaping into an unconfirmed grammar.

**Smaller things worth a look regardless**

- `keyword_search`'s `_validate_filter` call sits inside the existing broad `except Exception`, so a rejected filter gets logged and silently returns `None`, while `search`/`list` raise `ValueError` straight to the caller. Doesn't reopen the hole since the query never gets sent, but it's an inconsistent/quiet failure mode on that one path.
- `_SAFE_FILTER_KEY` anchors with `$` instead of `\Z`/`fullmatch()`, so a key that's literally `"user_id\n"` (trailing newline) passes validation. Narrow, but cheap to fix while touching this.
- The key regex is stricter than Upstash's own identifier grammar, which allows dots and array accessors for nested metadata (e.g. `geography.coordinates.longitude`) — this would now reject those as invalid keys. Probably a reasonable trade-off for an allow-list, just flagging it as a behavior change for anyone filtering on nested metadata against Upstash.

HrushiYadav:
Pushed fixes for all items from the review.

**Main fix (blocking)**: Switched from backslash-escaping to fail-closed rejection. String filter values containing `"` or `\` now raise `ValueError` outright, matching the same posture as the dict/list rejection. This removes the dependency on undocumented Upstash parser escape behavior entirely.

**Minor fixes**:

1. `keyword_search` filter validation now runs before the `try/except` block, so invalid filters raise `ValueError` to the caller instead of being logged and silently returning `None`.

2. `_SAFE_FILTER_KEY` now uses `fullmatch()` with `\Z` anchor instead of `$`, so keys with trailing newlines are rejected.

3. Re: dot-notation keys -- agreed this is a reasonable trade-off for an allowlist. Flagging it as a known behavior change in the commit message.

**`_stringify`** simplified to plain quoting since validation now guarantees no `"` or `\` reach it.

**Tests**: Replaced `test_stringify_escapes_quotes` with `test_filter_rejects_double_quote_in_value` and `test_filter_rejects_backslash_in_value`. Added `test_keyword_search_raises_on_invalid_filter` and `test_filter_rejects_key_with_trailing_newline`. All 28 tests pass, ruff clean.

kartik-mem0:
Verified against `6d7c0399` (new head): the blocking concern from last time is resolved, and by construction rather than by additional evidence.

`_validate_filter` no longer escapes into an unverified grammar — it now rejects any string filter value containing `"` or `\` outright (`mem0/vector_stores/upstash_vector.py:28-32`), raising `ValueError` before the value is ever handed to `_stringify`. `_stringify` itself is back to plain quoting with no escaping (`upstash_vector.py:112-115`), which is safe precisely because validation guarantees a quote or backslash never reaches it. That removes the dependency on Upstash's undocumented parser behavior entirely — there's nothing left to verify about a remote parser, because the dangerous value is never sent to Upstash in the first place. It's the same posture the OpenSearch fix (#5986) took, just extended from type to content.

Traced the exact `#5977` payload (`alice" OR 1=1 --`) against the new code: it contains a `"`, so it's rejected at `_validate_filter` before any filter string is built. Confirmed by reading the code and by `test_filter_rejects_double_quote_in_value` and `test_keyword_search_raises_on_invalid_filter` — the latter drives that literal payload through the real `keyword_search()` call, not just the helper in isolation.

Both minor items from last round are fixed too: `keyword_search()`'s filter validation now runs before the `try/except`, so a rejected filter raises to the caller instead of being swallowed into a silent `None`; and `_SAFE_FILTER_KEY` now uses `fullmatch()` with `\Z` so a trailing-newline key is correctly rejected.

Red-green on the new head: reverting just `upstash_vector.py` to main makes the test module fail to even collect (`_validate_filter` doesn't exist pre-fix); restoring it, back to 28/28 passing. `ruff check` clean. CI is green on the checks that matter (Python SDK 3.10/3.11/3.12, CI Gate, CLA) — Vercel's the only red check and that's a fork-permissions deploy preview, not a real failure.

One thing worth knowing rather than fixing: `UpstashVector` is a registered provider, and `Memory`/`AsyncMemory`'s metadata-filter processing only rewrites the specific recognized operator shapes (`AND`/`OR`/`NOT`, per-field comparison dicts, the `"*"` wildcard) — a plain string filter value passes through untouched whether you call `Memory.search()` or use `UpstashVector` directly. So this validation is the only checkpoint either way, not a second layer behind something upstream — good that it fails closed here rather than leaning on one.

LGTM — ready for squash-merge. The one user-visible trade-off is that a legitimate value containing a literal `"` or `\` (e.g. `she said "hi"`) is now rejected rather than accepted-with-escaping; that's the same shape of call the OpenSearch fix made and the right one for a P0 over an unverifiable escape.

kartik-mem0:

### What this PR does
Verified against `f719a8d5` (new head): production code is byte-identical to the previously-reviewed commit `6d7c0399` — the head only moved because the branch was rebased onto a newer `main`. Re-reviewed from scratch against the current tip rather than carrying the prior LGTM forward.

### Strengths
- `upstash_vector.py:20-32` — fail-closed rejection (not escaping) removes the exact unverifiable assumption about Upstash's parser that blocked round 1; confirmed this round against Upstash's own filter-syntax docs that no escape convention is documented for embedded quotes, so there's nothing left to verify server-side.
- `upstash_vector.py:136-138,186-188,281-283` — validation applied identically at all 3 filter-construction sites (`search`/`keyword_search`/`list`), confirmed by direct read.
- `upstash_vector.py:186-197` — `keyword_search`'s validation sits outside the broad `except Exception`, so a rejected filter raises instead of silently returning `None`.
- `tests/vector_stores/test_upstash_vector.py` — 9 filter-validation tests, including 2 end-to-end call-site tests (not just unit tests on the helper).

### Issues
Minor (optional):
- `upstash_vector.py:23` — `isinstance(value, (str,int,float,bool))` accepts an `int`/`float`/`bool` subclass with a hostile `__str__`. Traced this to production irrelevance: every real entry point (`server/main.py`'s `/search`, `/memories`) parses filters from JSON via Pydantic, which can never construct a custom subclass — only plain builtins. `type(value) in (str,int,float,bool)` is a one-line hardening if you want belt-and-suspenders anyway.
- `list()` calls `_validate_filter` correctly (confirmed by direct read, matches `search()`/`keyword_search()`) but has no dedicated end-to-end rejection test next to it, unlike the other two call sites — worth mirroring `test_search_rejects_dict_filter` for `list()` as a regression trip-wire.
- PR title says "escape quotes"; the shipped code rejects values containing them instead (a stronger fix) — accurate in spirit, just a heads-up for anyone skimming only the title.

### Evidence
- **CI:** green on all required checks — CI Gate, Python SDK build (3.10/3.11/3.12), changelog_check, check_changes, license/cla. (`Vercel` fails — advisory-only preview-deploy, not a required check.)
- **Targeted tests:** `hatch -e dev_py_3_11 run pytest tests/vector_stores/test_upstash_vector.py -v` → 29 passed.
- **Red-green proof:** reverting only the production file to pre-PR reproduces `ImportError: cannot import name '_validate_filter'` (can't even collect); restoring HEAD → 29 passed again.
- **Linked issue:** #5977 — the exact reported payload (`alice" OR 1=1 --`) is rejected before it ever reaches the filter string.
- **Cross-package check:** TS Upstash store (added by an unrelated, already-merged PR) has a narrower related gap — no filter-key validation, non-scalar values fall through unrejected. Out of scope for this PR's diff; flagging for a separate follow-up.
- **Could NOT verify:** live Upstash service behavior — code/unit-test-level verification only, though rejection (vs. escaping) means there's no remaining server-side assumption to verify.

### Assessment
Ready to merge: **Yes** (squash) · Est. human time: ~15 min
Re-review confirms the round-2 fix holds unchanged across the rebase. A full 3-lens + skeptic pass this round found no Critical or Important issue in this PR's own diff — two candidate concerns (a theoretical type-check bypass, a test-coverage gap on an already-correct `list()` path) were traced end-to-end and don't hold at blocking severity.

HrushiYadav:
@kartik-mem0 this is green (Vercel preview aside) and mergeable, and you marked it ready for squash-merge on the last pass. Since it is the filter-value breakout / cross-tenant issue from #5977, could we get it merged when you have a moment? Happy to rebase if main has moved.

HrushiYadav:
@kartik-mem0 @deshraj this has been LGTM'd and green (Vercel preview aside) for about 3 weeks. It is a P0 filter-value breakout / cross-tenant exfiltration fix (issue #5977), so I would really like to get it landed. Could one of you squash-merge it when you have a moment? The two sibling filter-hardening fixes (#5497, #5695) are in the same LGTM'd-and-green state if you are batching.

kartik-mem0 (approved):

PR #5982, opened 2026-06-29
fix(neptune): escape filter values in openCypher queries to prevent injection

### LINKED ISSUE #5978: Neptune Analytics vector store: openCypher injection via filter values
## Description

The `NeptuneAnalyticsVector` store builds openCypher queries using f-string interpolation of filter values in both `_get_where_clause()` and `_get_node_filter_clause()`. Filter values are inserted directly into the query string without escaping, allowing openCypher injection.

## Impact

An attacker who controls filter values can inject arbitrary openCypher clauses, potentially reading, modifying, or deleting data across the entire Neptune Analytics graph (not just the mem0 collection).

## Steps to Reproduce

```python
# In _get_where_clause:
filters = {"name": "x' OR 1=1 OR n.name = 'y"}
# Produces: WHERE n.name = 'x' OR 1=1 OR n.name = 'y'
# The injected OR 1=1 bypasses the intended filter

# In _get_node_filter_clause:
filters = {"label": "x'}}), (n2) DETACH DELETE n2 //"}
# Produces a clause that could delete nodes
```

## Suggested Fix

1. Escape `\` and `'` in filter values before interpolation.
2. Validate filter keys match `^[a-zA-Z_~][a-zA-Z0-9_]*$` (allowing `~` for Neptune's `~id` field).
3. Validate filter values are scalar types (str, int, float, bool).

HrushiYadav:
Closes #5978

## Summary

- Add `_escape_cypher()` to escape `\` and `'` in filter values before interpolation
- Add `_validate_filter()` to reject dict/list values and invalid keys (allows `~` prefix for Neptune's `~id` field)
- Applied to both `_get_where_clause()` and `_get_node_filter_clause()`

## Test plan

- [x] 9 new tests added covering escaping, validation, and both clause builders
- [x] All 9 tests pass
- [x] `ruff check` clean

### DIFF
--- a/mem0/vector_stores/neptune_analytics.py
+++ b/mem0/vector_stores/neptune_analytics.py
@@ -1,7 +1,8 @@
 import logging
+import re
 import time
 import uuid
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional
 
 from pydantic import BaseModel
 
@@ -13,6 +14,22 @@ except ImportError:
 from mem0.vector_stores.base import VectorStoreBase
 
 logger = logging.getLogger(__name__)
+
+_SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_~][a-zA-Z0-9_]*$")
+
+
+def _validate_filter(key: str, value: Any) -> None:
+    if not isinstance(key, str) or not _SAFE_FILTER_KEY.match(key):
+        raise ValueError(f"Invalid filter key: {key!r}")
+    if not isinstance(value, (str, int, float, bool)):
+        raise ValueError(
+            f"Filter value for {key!r} must be str, int, float, or bool, "
+            f"got {type(value).__name__}"
+        )
+
+
+def _escape_cypher(value: str) -> str:
+    return value.replace("\\", "\\\\").replace("'", "\\'")
 
 class OutputData(BaseModel):
     id: Optional[str]  # memory id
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Went through this against the root cause in #5978 — full read of `_get_where_clause` / `_get_node_filter_clause`, plus a red/green pass on the new tests.

**What the fix does**
- `_validate_filter` whitelists keys against `^[a-zA-Z_~][a-zA-Z0-9_]*$` and restricts values to `str`/`int`/`float`/`bool` (dict/list rejected outright).
- `_escape_cypher` escapes `\` before `'` — correct order, so a value ending in a backslash can't absorb the escape meant for a following quote.
- Applied at both call sites.

This is whitelisting + escaping rather than full `$param` binding. For the key half of the surface that's actually the right call — Cypher can't parameterize property/identifier names, so allowlisting is the real mitigation there, and it fails closed (raises) rather than just escaping. Values are escape-only even though this same file already does `$param` binding elsewhere (`delete`/`update`/`get` all bind through `params`) — would be worth a follow-up to move filter values onto that path too and drop string escaping entirely, but not a blocker since the escaping here is implemented correctly.

**Red/green**
- Restored the vulnerable `neptune_analytics.py` from `main`, kept the new test file → collection fails immediately (`ImportError: cannot import name '_escape_cypher'`), can't even run without the fix.
- Reproduced the issue's payloads directly against the restored vulnerable code:
  - `{"name": "x' OR 1=1 OR n.name = 'y"}` → `WHERE n.name = 'x' OR 1=1 OR n.name = 'y' ` (boolean bypass)
  - `{"x'}) DETACH DELETE n //": "..."}` → `WHERE n.x'}) DETACH DELETE n // = '...'` (raw key injection, completely unquoted)
- Restored the fix, same payloads: the value case now lands escaped inside the string literal (inert), and the malicious key raises `ValueError: Invalid filter key: ...` before it ever reaches the query template.
- Full suite with the fix: `9 passed, 9 skipped` (skips are the live-Neptune integration tests gated on `RUN_TEST_NEPTUNE_ANALYTICS`, expected without creds). `ruff check` clean, CI green on 3.10/3.11/3.12.

No other filter-splicing site in the file — `collection_name`/`vectors`/`top_k` get interpolated elsewhere too, but those aren't sourced from the filters dict this issue covers, so out of scope here.

LGTM. Good that this handles the key vector and not just values — that's the part that's easy to miss on this class of bug.

kartik-mem0 (approved):

PR #5986, opened 2026-06-29
fix(opensearch): validate filter values to prevent term query injection

### LINKED ISSUE #5985: OpenSearch vector store: filter term injection via dict/list values
## Description

The `OpenSearchDB` builds Elasticsearch-style term queries by directly embedding filter values into the query dict without type validation. If a caller passes a dict as a filter value (e.g. `{"user_id": {"value": "*", "case_insensitive": true}}`), it gets inserted verbatim into the term query, allowing an attacker to inject arbitrary Elasticsearch/OpenSearch query operators.

This is the same vulnerability pattern as the Elasticsearch backend (see #5976).

## Impact

An attacker who controls filter values can manipulate term queries to bypass access controls or alter query semantics.

## Suggested Fix

Validate that filter values are scalar types (str, int, float, bool) before building the term query. Reject dict/list values with a `ValueError`.

HrushiYadav:
Closes #5985

## Summary

- Add `_validate_filter()` to reject dict/list filter values that could inject OpenSearch term query operators
- Applied to all 3 filter construction points (`search()`, `keyword_search()`, `list()`)
- Same pattern as the Elasticsearch fix (#5980)

## Test plan

- [x] 5 new tests covering dict/list rejection, keyword_search validation, list error handling, and scalar acceptance
- [x] All 29 existing + new tests pass
- [x] `ruff check` clean

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import time
 from typing import Any, Dict, List, Optional
 
@@ -13,6 +14,18 @@ from mem0.configs.vector_stores.opensearch import OpenSearchConfig
 from mem0.vector_stores.base import VectorStoreBase
 
 logger = logging.getLogger(__name__)
+
+_SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_.]*$")
+
+
+def _validate_filter(key: str, value) -> None:
+    if not isinstance(key, str) or not _SAFE_FILTER_KEY.match(key):
+        raise ValueError(f"Invalid filter key: {key!r}")
+    if not isinstance(value, (str, int, float, bool)):
+        raise ValueError(
+            f"Filter value for {key!r} must be str, int, float, or bool, "
+            f"got {type(value).__name__}"
+        )
 
 
 class OutputData(BaseModel):
@@ -196,6 +209,7 @@ class OpenSearchDB(VectorStoreBase):
             for key in ["user_id", "run_id", "agent_id"]:
                 value = filters.get(key)
                 if value:
+                    _validate_filter(key, value)
                     filter_clauses.append({"term": {f"payload.{key}.keyword": value}})
 
         # Combine knn with filters if needed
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Reviewed this against the root cause in #5985 — full read of `search()`/`keyword_search()`/`list()`, red/green on the new tests, plus a hand-built exploit using the exact payload from the issue.

**What the fix does**
- `_validate_filter` rejects any non-`str`/`int`/`float`/`bool` filter value (dict/list) with a `ValueError`, applied at all three places `opensearch.py` builds a `term` filter clause: `search()`, `keyword_search()`, `list()`.
- Checked whether values alone are enough, or whether keys need validating too: here `key` only ever comes from a hardcoded `["user_id", "run_id", "agent_id"]` loop, never from the filters dict's own keys, so there's no key-injection surface being missed. Value-only validation is the right-sized fix for this file.
- It's type-rejection rather than escaping, which is nice for this class of bug — there's no assumption to verify about how a remote parser handles an escape sequence, because the malicious value never gets sent at all.

**Exploit + red/green**
- Checked OpenSearch's own docs to make sure the issue's example isn't hypothetical: `{"term": {"field": {"value": ..., "case_insensitive": ...}}}` is real, documented term-query syntax. Ran `{"user_id": {"value": "*", "case_insensitive": true}}` through `search()` on the unpatched code and it builds a fully valid query body with that dict sitting where a plain string should be; on the patched code it raises before the query is ever built, and the client's `search` is never called.
- Reverted only the `_validate_filter` hunk in `opensearch.py`, kept the new tests: 4 of the 5 new tests fail (the 5th, the plain-string-filter case, correctly still passes), all 25 pre-existing tests keep passing. Restored the fix, back to 29/29. `ruff check` clean, CI green (Python 3.10/3.11/3.12).

**Minor, not blocking**
- `list()` swallows the `ValueError` into its normal empty-result return (`[[]]`) since the filter-building loop sits inside its one big `try/except`, while `search()`/`keyword_search()` raise it straight to the caller. Both are safe — the query never goes out either way — just a pre-existing inconsistency in how rejection surfaces to the caller, not something this PR introduced.
- The key-format regex inside `_validate_filter` can never actually fail here since `key` is always one of the three hardcoded literals — harmless, just along for the ride from the shared helper shape.

LGTM. Nice that this went with straight-up type rejection instead of escaping — means the fix doesn't have to lean on any assumption about how OpenSearch parses things, it just never sends the bad shape in the first place.

kartik-mem0 (approved):

Issue #5991, opened 2026-06-29
fix(server): do not forward empty-string entity ids as filters


KushagraB424:
### Component

Core / Python SDK

### Summary

The self-hosted REST API endpoints (`/memories`, `/search`, and the `delete_all` endpoint) incorrectly forward empty strings as filters to the underlying vector database. If a client sends an empty string for `user_id`, `run_id`, or `agent_id` (e.g., via frontend forms or query parameters), the server uses the condition `if v is not None`. Because an empty string is not `None`, it gets added to the filter payload.

### Steps to Reproduce

import requests
# Assuming the self-hosted mem0 server is running on localhost:8000
# Sending an empty string for user_id via query params
response = requests.get("http://localhost:8000/memories?user_id=&run_id=my-run")
# The server incorrectly builds the filter: {"user_id": "", "run_id": "my-run"}

### Expected Behavior

Empty string values for user_id, run_id, and agent_id should be completely ignored and omitted from the database filter, treating them exactly the same as if they were omitted from the request entirely.

### Actual Behavior

The server includes the empty string in the filter payload because "" is not None evaluates to True. The vector database then executes a query looking specifically for a literal empty string (e.g., user_id == "") and returns zero results or acts unexpectedly.

### Environment

mem0 version: Latest (main)
Python/Node version: Python 3.10+
OS: Any

kartik-mem0:
## Triage Notes

**What we've established:**
- Same root cause as #5972: `server/main.py` builds identifier filters with `if v is not None` instead of a truthiness/empty check, so an empty-string `user_id`/`run_id`/`agent_id` is treated as a real, provided value.
- This report additionally flags that the same pattern recurs in `/search` and `DELETE /memories` (`delete_all_memories`), not just `GET /memories` — confirmed by tracing all three call sites on `main`.

**Reproduction:** Confirmed (code tracing on `main`)
**Root cause:** `server/main.py` uses `if v is not None` when building the filters/params dict in `get_all_memories` (~L428-430), `search_memories` (~L459-460), and `delete_all_memories` (~L537-539). An empty string is not `None`, so it passes straight through as a real filter value and silently over-scopes the query/delete.

**Next step:** Tracking this alongside #5972, which reported the same root cause first for the `GET /memories` case — consolidating the discussion there. PR #5992 is open covering all three endpoints; review feedback will follow on the PR.

rodboev:
[...]

PR #5992, opened 2026-06-29
fix(server): do not forward empty-string entity ids as filters

### LINKED ISSUE #5972: bug(server): GET /memories forwards empty-string entity ids as filters

KushagraB424:
## Linked Issue

Closes #5972
## Description

This PR fixes a bug in the self-hosted REST API where empty string values for `user_id`, `run_id`, and `agent_id` (e.g., from empty query parameters) were incorrectly forwarded to the underlying vector database. 

Previously, the server used `if v is not None` which evaluated to `True` for empty strings. This PR changes the checks to `if v` across the `/memories` (GET), `/search`, and `/memories` (DELETE) endpoints so that empty strings are properly ignored and safely omitted from database filters.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Tested manually by verifying that sending an empty string for `user_id` query parameters no longer builds an invalid dictionary filter in the backend endpoints.

## Checklist


kartik-mem0:
Thanks for tracking this down — `if v` is the right fix, and you've caught all three spots in `server/main.py` where empty strings were slipping past the `is not None` check (`get_all_memories`, `search_memories`, and `delete_all_memories`), not just the `GET /memories` case from the original report. It also lines up nicely with the `any([...])` guards already sitting right above two of these, which already treat `""` as "not provided" — this change just makes the filter-building consistent with that.

Before this is mergeable, could you add a regression test? There isn't one in the diff yet. There's a parallel PR, #5973, fixing the same bug that landed a small `server/tests/` setup with a test per endpoint (stub `get_memory_instance`, assert the empty-string id gets dropped from the filters/params) — that shape would be easy to mirror here.

One more small thing: mind re-pointing "Closes" at #5972 instead of #5991? #5991 is a duplicate of that original report, so linking the duplicate means the canonical issue stays open after this merges.

Solid fix otherwise — just need the test and the link fix and this is good to go.

KushagraB424:
Thanks for the review! Adding regression tests mirroring the pattern from #5973 and updated the linked issue to Closes #5972.

kartik-mem0:
Pushed the test and the link fix — both close out the ask. `test_get_memories_filters.py` covers all three endpoints (`get_all_memories`, `search_memories`, `delete_all_memories`), stubbing `get_memory_instance` and asserting the empty-string id gets dropped from the filters/params dict. I checked these actually discriminate rather than just document the new behavior: ran them against this branch (3 passed), then reverted just `main.py` back to the pre-fix `is not None` version with the same tests still in place — all 3 failed, with the empty `agent_id`/`run_id` showing up in the asserted dict — so they would have caught the original bug. `conftest.py`'s env defaults are enough to import `main` without a live Postgres, and ruff is clean on both new files.

"Closes" now points at #5972, so the canonical issue actually closes on merge too.

That resolves both asks from before — nice work mirroring the #5973 test shape here, it fits this file well. One thing worth flagging for whoever merges: `server/` still isn't wired into `ci-gate.yml`, so these tests won't run automatically in CI yet — worth running them manually before merge, same as I just did.

kartik-mem0 (approved):

Issue #6003, opened 2026-06-30
@mem0/opencode-plugin has multiple OpenCode Desktop issues: plugin load failure, API key resolution failure, and always-wrong app_id


zhusinian:
## Summary

`@mem0/opencode-plugin` has multiple issues in OpenCode Desktop on Linux.

In my environment, the plugin has three separate but stacked problems:

1. the published plugin bundle can fail to load in OpenCode Desktop
2. after patching that locally, the plugin cannot reliably obtain `MEM0_API_KEY`
3. after patching that too, the plugin still always resolves the wrong `app_id` in Desktop

In contrast, OpenCode CLI works much better in the same repositories.

So this appears to be a plugin-side compatibility problem with OpenCode Desktop runtime and Desktop project/session context handling.

## Environment

- OS: Linux (Ubuntu GNOME / Wayland)
- OpenCode CLI: works
- OpenCode Desktop: problematic
- Plugin: `@mem0/opencode-plugin`

Example repositories:

- `/home/zhusinian/Code/yishu_projects/soms`
- `/home/zhusinian/Code/yishu_projects/soms_analysis_go`

### Observed behavior

OpenCode Desktop logs showed plugin load failures like:

```text
failed to load plugin
error=__require is not a function
path=@mem0/opencode-plugin/dist/index.js
```

### Evidence from the published bundle

The published bundle contains code like:

```js
var __require = import.meta.require;
```

This is not safe in my OpenCode Desktop runtime.

### Impact

Before local patching, the plugin cannot initialize at all in Desktop.

### Why this matters

The package appears to be built with Bun-targeted output that assumes `import.meta.require` exists, but that assumption does not hold in this Desktop runtime.

## Problem 2: The plugin cannot obtain MEM0_API_KEY in Desktop

After patching the load failure locally so the plugin could initialize, I hit a second issue.

### Observed behavior

Desktop logs then showed:

```text
MEM0_API_KEY environment variable not set. Get one at https://app.mem0.ai/dashboard/api-keys
```

### Relevant source code

From the plugin bundle:

```js
const apiKey = process.env.MEM0_API_KEY;
if (!apiKey) {
  ...
  return {};
}
```

### Why this fails in Desktop

This works in CLI because the shell environment is inherited.

But in Desktop, GUI applications often do not inherit shell variables from `.zshrc` / `.bashrc`, so the plugin fails even though CLI works.

### Impact

- CLI works
- Desktop shows the plugin as installed
[...]

PR #6012, opened 2026-06-30
fix(opencode-plugin): use OpenCode project context and publish a Node-safe bundle

rodboev:
## Linked Issue

Part of #6003

Follow-up to https://github.com/mem0ai/mem0/pull/6005, which documented the OpenCode Desktop workaround and narrowed this PR to the remaining runtime fixes.

## Description

`@mem0/opencode-plugin` needs two Desktop runtime fixes. The published entrypoint must be Node-compatible, and project metadata must come from OpenCode's active `worktree` or `directory` instead of startup shell state. `MEM0_APP_ID` remains the highest-priority override.

This change keeps the existing current-main API-key resolution and shell-profile behavior, composes the active project context into the git helpers, keeps helper exports off the published entrypoint, and updates the package metadata and docs for the Node-targeted runtime contract.

## Scope

The scope is limited to the OpenCode package entrypoint, project context resolution, package metadata and lockfile, focused tests, and the related integration documentation. It does not change Desktop setup documentation from #6005 or add new API-key discovery behavior.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I tested manually
- [ ] No tests needed

Verification from `integrations/mem0-plugin/.opencode-plugin`:

- [x] `bun test opencode-mem0.test.ts project.test.ts`: 19 passed, 0 failed
- [x] `bun run type-check`
- [ ] `bun run build`, blocked by Bun failing to remap the local `node_modules` binary
- [x] `git diff --check`
- [ ] Full package formatting check, current-main files include existing Prettier differences

## Checklist

### DIFF
--- a/docs/integrations/opencode.mdx
+++ b/docs/integrations/opencode.mdx
@@ -101,7 +101,7 @@ To change the **default** scope (used when no scope is passed), run the `/mem0-s
 
 The default persists in `~/.mem0/settings.json` (`default_scope`) and is read fresh on each memory operation, so a change applies immediately. No restart. `delete_all_memories` always requires an explicit `scope: "global"` to delete user-wide, so changing the default can't trigger a cross-project wipe.
 
-The plugin first chooses OpenCode's active `worktree`, then `directory`, as the path it will inspect. From that path it derives the project id (`app_id`) from the git remote (`owner-repo`), falling back to the git repo's root directory name, then the current directory.
+The plugin first chooses OpenCode's active `worktree`, then `directory`, as the path it will inspect. From that path it derives the project id (`app_id`) from the git remote (`owner-repo`), falling back to the git repo's root directory name, the selected project-path basename, then the process current-directory basename.
 
 ## Lifecycle Hooks
 
@@ -142,7 +142,7 @@ If auto-dream hasn't run yet, it's almost always because a gate hasn't been met 
 - **Plugin not loading**: Run `opencode plugin @mem0/opencode-plugin` again, then restart
 - **Hooks not firing**: Hooks require the plugin install (Option A). MCP-only installs don't include hooks.
 - **Auto-dream never runs**: It's gated (time + sessions + memories). Run `/mem0-status` to see which gate is blocking, or `/mem0-dream` to consolidate now.
-- **Wrong project name / memories not found**: Check the resolved id with `/mem0-status`. The plugin prefers OpenCode's active `worktree` or `directory`; if neither is present, it falls back to git and then the current directory, so a wrong cwd can still scope memories to the wrong project.
+- **Wrong project name / memories not found**: Check the resolved id with `/mem0-status`. The plugin derives git identity from the selected OpenCode project path, preferring the active `worktree` or `directory`; if neither is present, it uses the git remote, git root basename, selected project-path basename, and process current-directory basename.
 
 <CardGroup cols={2}>
   <Card title="Mem0 MCP Setup" icon="puzzle-piece" href="/platform/mem0-mcp">
--- a/docs/integrations/opencode.mdx
+++ b/docs/integrations/opencode.mdx
@@ -101,7 +101,7 @@ To change the **default** scope (used when no scope is passed), run the `/mem0-s
 
 The default persists in `~/.mem0/settings.json` (`default_scope`) and is read fresh on each memory operation, so a change applies immediately. No restart. `delete_all_memories` always requires an explicit `scope: "global"` to delete user-wide, so changing the default can't trigger a cross-project wipe.
 
-The project id (`app_id`) is derived from your git remote (`owner-repo`), falling back to the git repo's root directory name, then the current directory. Launch OpenCode from inside your repo so memories scope to the project rather than your home directory.
+The project id (`app_id`) comes from OpenCode's active `worktree` first, then `directory`. When OpenCode does not supply either, the plugin falls back to your git remote (`owner-repo`), then the git repo's root directory name, then the current directory.
 
 ## Lifecycle Hooks
 
@@ -142,7 +142,7 @@ If auto-dream hasn't run yet, it's almost always because a gate hasn't been met 
 - **Plugin not loading**: Run `opencode plugin @mem0/opencode-plugin` again, then restart
 - **Hooks not firing**: Hooks require the plugin install (Option A). MCP-only installs don't include hooks.
 - **Auto-dream never runs**: It's gated (time + sessions + memories). Run `/mem0-status` to see which gate is blocking, or `/mem0-dream` to consolidate now.
-- **Wrong project name / memories not found**: The project id comes from your git remote; launch OpenCode from inside the repo (not your home directory). Check the resolved id with `/mem0-status`.
+- **Wrong project name / memories not found**: Check the resolved id with `/mem0-status`. The plugin prefers OpenCode's active `worktree` or `directory`; if neither is present, it falls back to git and then the current directory, so a wrong cwd can still scope memories to the wrong project.
 
 <CardGroup cols={2}>
   <Card title="Mem0 MCP Setup" icon="puzzle-piece" href="/platform/mem0-mcp">
--- a/docs/integrations/opencode.mdx
+++ b/docs/integrations/opencode.mdx
@@ -101,7 +101,7 @@ To change the **default** scope (used when no scope is passed), run the `/mem0-s
 
 The default persists in `~/.mem0/settings.json` (`default_scope`) and is read fresh on each memory operation, so a change applies immediately. No restart. `delete_all_memories` always requires an explicit `scope: "global"` to delete user-wide, so changing the default can't trigger a cross-project wipe.
 
-The project id (`app_id`) comes from OpenCode's active `worktree` first, then `directory`. When OpenCode does not supply either, the plugin falls back to your git remote (`owner-repo`), then the git repo's root directory name, then the current directory.
+The plugin first chooses OpenCode's active `worktree`, then `directory`, as the path it will inspect. From that path it derives the project id (`app_id`) from the git remote (`owner-repo`), falling back to the git repo's root directory name, then the current directory.
 
 ## Lifecycle Hooks
 
--- a/integrations/mem0-plugin/.opencode-plugin/bun.lock
+++ b/integrations/mem0-plugin/.opencode-plugin/bun.lock
@@ -12,9 +12,6 @@
         "bun-types": ">=1.3.14",
         "typescript": "^5.7.3",
       },
-      "peerDependencies": {
-        "bun": ">=1.0.0",
-      },
     },
   },
   "packages": {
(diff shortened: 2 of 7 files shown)


kartik-mem0:
## Changes Requested

### What this PR does
Fixes 2 of the 3 Desktop failures in #6003: switches the published `build` script (only) from `bun build --target bun` to `--target node` so `dist/index.js` no longer ships `import.meta.require`, and threads OpenCode's `ctx.directory`/`ctx.worktree` into `getProjectId()` (via a new `selectActiveProjectPath()` + `.cwd(projectPath)`-scoped git calls) so `app_id` resolution stops depending on ambient `process.cwd()`. Problem 2 (API key resolution) is explicitly out of scope, tracked separately in #6005.

### Strengths
- `package.json` — scoped precisely: only `build` (the published artifact) moves to `--target node`; `dev` stays `--target bun` since local dev already runs under Bun. Verified `dist/index.js` has zero `import.meta.require` after this change.
- `opencode-mem0.ts` — `MEM0_APP_ID` env override is still checked first, before any path resolution; precedence is preserved and regression-tested.
- `project.ts` — `selectActiveProjectPath` (`worktree` → `directory` → `process.cwd()`) matches the pattern already used in `integrations/pi-agent-plugin/src/memory/scoping.ts`, so this isn't a new one-off mechanism.
- The fix targets the actual reported symptom, not just a plausible-looking one: I ran `getProjectId()` directly against the issue's own example paths (`.../yishu_projects/soms`, `.../soms_analysis_go`) and both now resolve to correct, distinct identities instead of colliding into the same wrong one.

### Issues

#### Critical (must fix)
None.

#### Important (should fix)
1. `opencode-mem0.ts:72-79` (`getBranch`), called at `opencode-mem0.ts:293-294` immediately after the now-fixed `getProjectId(ctx)` — still shells `git branch --show-current` through the raw, unscoped `$`, never routed through the new `commandInProject`/`.cwd(projectPath)` machinery. Same ambient-cwd bug this PR fixes for `app_id`, left unaddressed for `branch` metadata (written on every memory write), and untested. Suggest threading the same resolved path into `getBranch` the way `getProjectId` now does.
2. `opencode-mem0.test.ts` — 2 of the 14 new tests fail on POSIX (reproduced independently, including cloning PR head and running the exact command from this PR's own checklist): `bun test project.test.ts opencode-mem0.test.ts` → 12 pass, 2 fail, not the claimed "14 pass, 0 fail." Both failures hardcode Windows-style fixtures (`"D:\\Repos\\..."`) asserted against the platform-default (POSIX) `path.basename()`, which doesn't split on `\`. Not a production bug — `getProjectId` itself resolves correctly against realistic POSIX paths — but the suite doesn't pass as claimed, and `opencode-plugin-checks.yml` never runs `bun test` (unlike `pi-agent-plugin-checks.yml`/`openclaw-checks.yml`, which do), so nothing catches this going forward. Swap those two fixtures for POSIX-style paths.
3. Related to #1: `getProjectId(input: ProjectContext & {$: any})` resolves `projectPath` as a value local to its own body instead of the plugin entry point resolving it once (e.g. `const projectPath = selectActiveProjectPath(ctx)` in `Mem0Plugin`'s prologue) and passing the primitive to every helper that needs it. That's the direct cause of #1 — `getBranch` has no way to reach the resolved path today.

#### Minor (optional)
- `opencode-mem0.ts:43-46` — the `ShellCommand`/`typeof command.cwd === "function"` duck-typing is unreachable in production (the real `BunShellPromise.cwd` is always present per the installed `@opencode-ai/plugin` types); could type directly against it instead.
- `package.json` — `peerDependencies.bun` is still `">=1.0.0"` even though the published bundle no longer needs Bun to run.
- `docs/integrations/opencode.mdx`'s troubleshooting section still describes only the pre-fix resolution model; a one-line update mentioning `directory`/`worktree` would keep it current.

### Evidence
- **CI:** CI Gate pass; `OpenCode Plugin / build` pass (install + type-check + build + dist-check only — this job does not run `bun test`, so the failures below have no CI guard rail). `Vercel` check is red but advisory (fork deploy-preview, fails on ~every PR) — ignored per policy.
- **Targeted tests:** `bun test opencode-mem0.test.ts project.test.ts` at PR head (`241e1f1`) → 12 pass, 2 fail. Reproduced 4 times independently (including a live clone-and-run of this PR's own stated verification command), same 2 failing cases each time.
- **Red-green proof:** reverted `opencode-mem0.ts`/`project.ts` to `main` (tests kept) → hard import-not-found failure, 0 pass — tests genuinely depend on the new code, not vacuous. Restored to PR head → back to 12 pass/2 fail. `bun run type-check` clean, `bun run build` succeeds, and `grep -c "import.meta.require" dist/index.js` → 0 matches, confirming the bundle fix works at the build-artifact level.
- **Linked issue:** #6003 — Problem 1 (bundle load failure) fixed and verified. Problem 3 (wrong `app_id`) fixed for the reported symptom (two real projects no longer collide into the same wrong identity), though the `getBranch` gap above means related metadata can still mis-tag in the same Desktop multi-instance scenario. Problem 2 (API key resolution) is untouched here, tracked in #6005. Given that, "Closes #6003" will auto-close an issue with one sub-problem still code-unaddressed — worth switching to "Part of #6003" or coordinating the close with #6005 when this merges.
- **Cross-package check:** `integrations/pi-agent-plugin` doesn't share this bug (already threads a resolved cwd through); no Python↔TS parity concern (TS-only plugin package).
- **Could NOT verify:** OpenCode Desktop's actual runtime behavior for its global/background plugin instance specifically (no live Desktop available here) — traced via the issue's own log evidence and the installed `@opencode-ai/plugin` types instead of live telemetry.

### Assessment
Ready to merge: **With fixes** — the `getBranch` scoping and the 2 test fixtures are both small, same-shape follow-ups to code already in this diff, not a redesign. Happy to re-review once pushed.

rodboev:
Addressed this in `be624f9e`.

1. The plugin now resolves the OpenCode project path once at startup and passes it into both `getProjectId` and `getBranch`, so branch metadata uses the same `worktree` or `directory` context as `app_id`.
2. The basename-sensitive tests now use POSIX-style paths, and `getBranch` has regression coverage for cwd scoping and fallback to `main`.
3. I updated the PR body to say `Part of #6003` because API-key fallback discovery remains intentionally out of scope here.

kartik-mem0:
Thanks for turning this around, and for the precise commit — `be624f9e` does exactly what it says on the tin. I re-read the diff and re-ran everything rather than taking the summary at face value, and it holds up.

All three "should fix" points from the last round are genuinely fixed:

1. **`getBranch` scoping** — it now goes through the same `commandInProject(...)` machinery as `getProjectId`, called with the same resolved `projectPath` (`opencode-mem0.ts:67-74`, wired at `:288-290`). To make sure this wasn't just a plausible-looking change, I reverted `opencode-mem0.ts` to the SHA I reviewed last time while keeping your new test file — it fails immediately with `Export named 'getBranch' not found`, so the new tests aren't vacuous, they actually depend on this fix.
2. **Test fixtures** — swapped to POSIX paths in both `opencode-mem0.test.ts` and `project.test.ts`. Ran `bun test opencode-mem0.test.ts project.test.ts` myself at `be624f9e`: 16 pass, 0 fail, matching your PR checklist for real this time.
3. **Hoisting `projectPath`** — `getProjectId` and `getBranch` both now take it as a parameter instead of re-deriving it, resolved once in `Mem0Plugin` (`opencode-mem0.ts:288`). That's exactly the shared-root-cause fix rather than a second patch next to the first one.

Also confirmed the PR body now says "Part of #6003" instead of "Closes #6003" — checked #6003 is still open (correct, Problem 2/API-key resolution is still tracked separately in #6005), so this won't mis-close it on merge.

Checks I ran at `be624f9e` (not just CI's word for it):
- `bun install --frozen-lockfile` + `bun run type-check` — clean
- `bun run build` + `grep -c "import.meta.require" dist/index.js` — 0 matches, bundle fix still holds
- `bun test opencode-mem0.test.ts project.test.ts` — 16 pass, 0 fail
- Grepped the rest of the repo for other callers of `getProjectId`/`getBranch` — none, so the signature change is contained
- `gh pr checks` — CI Gate and OpenCode Plugin/build both green (Vercel's red is the usual advisory fork-preview noise, unrelated)

Nothing new jumped out in the diff itself — it's a single tight commit touching exactly the three things discussed.

Two small things left on the table from before, still just optional polish, not blocking: `peerDependencies.bun` in `package.json` is still `>=1.0.0` even though the published bundle no longer needs Bun to run, and `docs/integrations/opencode.mdx`'s troubleshooting section still only describes the cwd/launch-directory model rather than mentioning `directory`/`worktree`. Happy to see those in a follow-up, or I can just fix them myself post-merge — your call.

This looks good to me. Nice fix.

rodboev:
Thanks for the follow-up pass. I picked up the two remaining contract nits you called out.

1. `integrations/mem0-plugin/.opencode-plugin/package.json` no longer advertises Bun as a runtime peer dependency. The plugin still uses Bun for local build scripts, but the published entrypoint is the Node-targeted `dist/index.js`, so npm consumers do not need Bun installed just to run it.
2. `docs/integrations/opencode.mdx` now matches the implemented project-resolution flow. The docs say the plugin uses OpenCode's active `worktree` first, then `directory`, and only falls back to git/cwd heuristics when OpenCode does not supply project context.

The runtime code stays unchanged here. This round only brings the published metadata and docs back in line with the contract already implemented in `project.ts` and `opencode-mem0.ts`.

kartik-mem0:

### What this PR does
Verified against `a517f254` (new head): the two Minor items left open in the last round are now resolved — the stale `peerDependencies.bun` declaration is removed from `package.json` (and `bun.lock` regenerated to match), and the `opencode.mdx` docs now accurately describe the real `worktree` → `directory` → git-remote → cwd resolution order.

### Strengths
- `package.json` — removing `peerDependencies.bun` fixes a real self-contradiction: the published bundle is built `--target node` and genuinely doesn't need Bun at runtime (0 `import.meta.require`, 0 `Bun.*` calls in `dist/index.js`), so telling npm/pnpm consumers they need Bun ≥1.0.0 installed was actively wrong, not just unused metadata.
- `bun.lock` — the diff is 100% subtractive (drops only the `bun` + `@oven/bun-*` platform-binary entries the removed peer dep pulled in); `bun install --frozen-lockfile` installs clean, confirming it's a properly regenerated, consistent lockfile.
- `docs/integrations/opencode.mdx` — both reworded passages match the real `PluginInput` type (`worktree`/`directory` are genuine fields on the SDK's plugin context) and the actual fallback chain in `project.ts`/`opencode-mem0.ts`; the troubleshooting bullet is honest about the residual cwd-fallback caveat rather than overselling the fix.
- `opencode-mem0.ts`, `project.ts`, and their test files are byte-identical to the last reviewed head — the hoisted-`projectPath` fix and `getBranch` scoping from the previous round are untouched.

### Issues
Minor (optional): `opencode.mdx:104` reads a bit like `worktree`/`directory` supply the id directly, with git-remote as a fallback only when neither is present. Actually `getProjectId` (`opencode-mem0.ts:48-65`) always parses the git remote — `worktree`/`directory` just pick *which path* it's scoped to via `.cwd()`. Not worth blocking on since the troubleshooting bullet already points people at `/mem0-status` for ground truth, but worth a tighten-up whenever you're next in the file.

### Evidence
- **CI:** `OpenCode Plugin / build` pass; `docs llms.txt check` pass; `CI Gate` pass; `Vercel` preview fails (advisory, ignored — fails on every fork PR per policy).
- **Targeted tests:** `bun test opencode-mem0.test.ts project.test.ts` → 16 pass, 0 fail (unchanged from last round — same files, same result).
- **Red-green proof:** N/A this round — no new logic or tests in this commit (it's a dependency-metadata removal + docs reword); the prior round's red-green proof on `getBranch` still applies unmodified since the file it exercises hasn't changed.
- **Linked issue:** Part of #6003 (still open) — this PR's own scope, unchanged. Companion #6005 still open, still handles the API-key part.
- **Cross-package check:** N/A — TS-only change scoped to `.opencode-plugin`; sibling `pi-agent-plugin`/`openclaw` packages unaffected and checked for the same anti-pattern (neither peer-deps on its own build tool).
- **Could NOT verify:** live OpenCode Desktop behavior end-to-end — verified via code tracing, published SDK type defs, and the build artifact instead.

### Assessment
Clean, minimal follow-up that closes out exactly what was flagged as non-blocking last round — thanks for circling back on both so quickly.

rodboev:
Good catch. I tightened docs/integrations/opencode.mdx:104 in 88beff8c so it now says worktree and directory choose the path the plugin inspects, and the pp_id is still derived from the git remote for that selected path. No runtime code changed in this follow-up.

kartik-mem0:
## Changes Requested

Verified against `88beff8c` (new head): the docs wording fix is correct and resolves the Minor
item from the last round. Re-reading the code this round also surfaced a real, reproducible bug in
code that's been unchanged since round 2 — details below.

### What's fixed since the last round

The single new commit reworks `docs/integrations/opencode.mdx`. Last round's Minor finding was
that the docs framed `worktree`/`directory` as direct `app_id` sources with git-remote as a
fallback, when `getProjectId` (`opencode-mem0.ts:48-65`) actually always runs the git-remote
lookup, scoped to whichever path `worktree`/`directory` selected. The new wording — "The plugin
first chooses OpenCode's active `worktree`, then `directory`, **as the path it will inspect**.
From that path it derives the project id" — states that distinction precisely. Checked against the
current `getProjectId` body: matches exactly. Good fix.

### Issues

#### Important
1. `integrations/mem0-plugin/.opencode-plugin/opencode-mem0.ts:48,67` — `getProjectId`/`getBranch`
   are `export`ed (added this PR — they were non-exported, single-arg functions on `main` before
   it) purely so `opencode-mem0.test.ts` can import them. But `opencode-mem0.ts` is also the bundle
   entry point, so both land in `dist/index.js`'s export list (confirmed: `export { getProjectId,
   getBranch, opencode_mem0_default as default }`).

   This is not cosmetic. OpenCode's real host loader (`packages/opencode/src/plugin/index.ts` in
   `anomalyco/opencode`) only recognizes a default export shaped `{server: fn}` as a "v1" plugin;
   `Mem0Plugin` is a bare async function (`opencode-mem0.ts:266`), so it fails that check and the
   loader falls back to a legacy path that calls **every named export** of the module as if it were
   an independent plugin factory: `for (const entry of Object.values(mod)) { ... }`.

   I reproduced this directly against the actual built `dist/index.js` (not just traced it): the
   loader ends up calling `getProjectId(input, load.options)`, mis-binding `$` to the plugin's
   input object (not a callable shell tag) and `projectPath` to `undefined`. Both internal git
   calls throw immediately and are swallowed by the empty `catch {}` blocks at lines 54 and 60;
   execution falls through to the unguarded `basename(projectPath)` at line 62, which throws:

   ```
   TypeError: The "path" argument must be of type string. Received undefined
   ```

   Ran this twice against two separate fresh builds — same result both times. `getBranch` has the
   identical `$`-misbinding but its `catch {}` falls through to a hardcoded `return "main"`
   (line 73) instead of `basename()`, so it doesn't throw — it just silently returns a meaningless
   value.

   The good news: I also checked the actual caller in the OpenCode loader, and this does **not**
   crash the whole process — `applyPlugin` is wrapped in a per-plugin `Effect.tryPromise`/`catch`
   that logs and swallows the error, and `Mem0Plugin`'s real hooks (entry 1, since module-namespace
   key order is alphabetical and spec-guaranteed: `default` < `getBranch` < `getProjectId`) already
   register before the throw on entry 3 aborts the rest of the loop. So the plugin's actual
   functionality likely keeps working. But this is a 100%-reproducible "failed to load plugin"
   error logged on every single load, in exactly the area this PR's own title claims to fix
   ("publish a Node-safe bundle") — and it's shipped past two LGTM rounds because nothing exercises
   the module's export surface against a real or simulated plugin loader.

   Fix: move `getProjectId`, `getBranch`, and the `commandInProject`/`ShellCommand` helper
   (lines 38-74) into `project.ts`, next to `selectActiveProjectPath`, which already follows the
   right pattern there — defined and exported from `project.ts`, but only *imported* (never
   re-exported) by `opencode-mem0.ts`. `opencode-mem0.test.ts`'s imports of these two functions
   would just repoint to `project.ts`.

### Evidence

- **CI:** all real required checks green (`CI Gate`, `docs llms.txt`, `OpenCode Plugin / build`,
  `license/cla`); `Vercel` fails (advisory preview-deploy, not merge-blocking, same every fork PR).
- **Targeted tests:** `bun test` (fresh, this round) → `42 pass, 0 fail, 87 expect() calls` across
  5 files. `bun run type-check` → clean. `bun run build` → `Bundled 204 modules`, `index.js 0.92 MB`.
  `grep -c "import.meta.require" dist/index.js` → `0` (Node-safety claim holds).
- **Export-leak repro:** ran the loader's real dispatch logic (`readV1Plugin` detect-mode check →
  `getLegacyPlugins` fallback → `server(input, load.options)` per export), copied verbatim from the
  fetched `anomalyco/opencode` source, against the actual worktree-built `dist/index.js`. Reproduces
  the `TypeError` above on every run.
- **Introduced-by-this-PR check:** on the PR's base commit, `getProjectId`/`getBranch` were
  `async function getProjectId($: any)` / `async function getBranch($: any)` — not exported,
  single-arg. This PR added `export` and the second `projectPath` arg.
- **Could NOT verify:** whether this currently surfaces as a user-visible error in the OpenCode TUI
  today — the specific event-publish path for this failure is commented out
  (`// TODO: make proper events for this`) in the loader source I read, so today it's a silently
  logged internal error, not a visible one. That could change in a future OpenCode release.

### Assessment
Ready to merge: **With fixes** — please relocate `getProjectId`/`getBranch` out of the bundle
entry file's export surface as above; I'll re-review once you push.

rodboev:
Thanks, this was a real bug, and the problem was the entrypoint export surface, not the project-scoping logic itself.

1. I moved `getProjectId`, `getBranch`, and the scoped shell helper out of `opencode-mem0.ts` and into `project.ts`, which already owns the OpenCode path-selection helpers. That leaves the built entry module with only the default plugin export, so the legacy loader stops treating helper functions as plugin factories.
2. `Mem0Plugin` still calls the same helpers with the same resolved `projectPath`, so the round-2 cwd-scoping fix stays intact. This is a module-boundary fix, not a runtime-behavior rewrite.
3. I repointed the helper tests to `project.ts` and added an entry-module export-surface regression on `./opencode-mem0` so this invariant stays closed. Enumeration summary for the entrypoint-export class is one helper-owner row, one entrypoint caller row, and one entry-module test row, all fixed this round.

PR #6014, opened 2026-06-30
fix(ts-oss/redis): fall back to "*" for empty or all-null filters

### LINKED ISSUE #6013: bug(ts-oss/redis): empty/all-null filters build an invalid query ("" instead of "*")
### Component

### Description

```ts
      .join(" ")
  : "*";
```

```ts
```

### Steps to Reproduce

### Expected Behavior

### Actual Behavior

### Environment

abhay-codes07:
## Linked Issue

Closes #6013

## Description

The OSS Redis vector store builds its RediSearch pre-filter in both `search()` and `list()` like this:

```ts
      .join(" ")
  : "*";
```

The `"*"` fallback only fires when `filters` is falsy. If `filters` is a truthy-but-empty object (`{}`), or has only null/undefined values, the chain produces `""`, which gets interpolated as `" =>[KNN ...]"` — a malformed RediSearch query rather than match-all.

I pulled the duplicated logic into a shared `buildRedisFilterExpr()` helper that returns `"*"` whenever there are no usable conditions, and used it in both `search()` and `list()`. Non-empty filters behave exactly as before.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] No tests needed

Added `redis-filter-expr.test.ts` covering undefined, empty `{}`, all-null/undefined, single, multi (with snake_case), and mixed filters. The empty and all-null cases fail without the fix. `tsc --noEmit` shows no new errors in the changed file and `prettier --check` is clean.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/redis.ts
+++ b/mem0-ts/src/oss/src/vector_stores/redis.ts
@@ -133,6 +133,18 @@ function toSnakeCase(obj: Record<string, any>): Record<string, any> {
   );
 }
 
+// Build a RediSearch pre-filter expression from filters. Returns "*" (match
+// all) when there are no usable conditions — including an empty filters object
+// or one whose values are all null/undefined — since an empty expression is an
+// invalid RediSearch query (e.g. ` =>[KNN ...]`).
+export function buildRedisFilterExpr(filters?: SearchFilters): string {
+  if (!filters) return "*";
+  const conditions = Object.entries(toSnakeCase(filters))
+    .filter(([, value]) => value !== null && value !== undefined)
+    .map(([key, value]) => `@${key}:{${escapeRedisTagValue(value)}}`);
+  return conditions.length > 0 ? conditions.join(" ") : "*";
+}
+
 // Utility function to convert object keys to camelCase
 function toCamelCase(obj: Record<string, any>): Record<string, any> {
   if (typeof obj !== "object" || obj === null) return obj;
@@ -406,13 +418,7 @@ export class RedisDB implements VectorStore {
     filters?: SearchFilters,
   ): Promise<VectorStoreResult[]> {
     await this.initialize();
-    const snakeFilters = filters ? toSnakeCase(filters) : undefined;
-    const filterExpr = snakeFilters
-      ? Object.entries(snakeFilters)
-          .filter(([_, value]) => value !== null && value !== undefined)
-          .map(([key, value]) => `@${key}:{${escapeRedisTagValue(value)}}`)
-          .join(" ")
-      : "*";
+    const filterExpr = buildRedisFilterExpr(filters);
 
     const queryVector = new Float32Array(query).buffer;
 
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss Redis fix with a unit test. Quick review when you can, thanks.

kartik-mem0:
## LGTM — ready to squash-merge after a trivial rebase

### What this PR does
`RedisDB.search()` and `list()` build the RediSearch pre-filter as `filters ? entries.filter(...).map(...).join(" ") : "*"`. The `"*"` (match-all) fallback only fires when `filters` is falsy, so a truthy-but-empty `{}` — or an object whose values are all `null`/`undefined` — collapses to `""` via `.join(" ")` and interpolates into an invalid RediSearch query (`" =>[KNN …]"`) instead of matching everything. The fix extracts a shared `buildRedisFilterExpr()` that returns `"*"` when there are no usable conditions, and routes both `search()` and `list()` through it.

### Strengths
- **Root-cause fix at the one shared choke point, applied to both call sites** (`search()` and `list()`) — not a patch on the single path the issue names. `keywordSearch()` is a `return null` stub and builds no filters, so those two are the only sites and the fix is complete for this file.
- **Behavior-preserving for every non-empty case.** `if (!filters) return "*"` catches exactly the same falsy set the old ternary did; `escapeRedisTagValue`, snake-casing, and entry ordering are unchanged. The only behavior change is the intended `{}`/all-null → `"*"`.
- **Brings TS in line with Python.** `mem0/vector_stores/redis.py` already treats empty/all-null filters as match-all (redisvl `Tag` + `if conditions:` guard + `str(filter) if filter is not None else "*"`), so this removes a Python↔TS inconsistency rather than adding one.
- **Matches an existing repo precedent** — `pgvector.ts` already exports `buildFilterConditions` with its own unit test; this mirrors that shape.

### One thing before merge
- The branch is currently **CONFLICTING** and needs a rebase onto `main`. It's purely mechanical: `main` (via the lazy-provider-load change) added `await this.initialize();` as the first line of both `search()` and `list()`, right next to the block this PR replaces. Resolve by **keeping both** — `await this.initialize();` followed by `const filterExpr = buildRedisFilterExpr(filters);`. No code change needed.

### Evidence
- **CI:** TypeScript SDK build + integration checks pass; `license/cla` signed. (Deploy-preview checks are not merge-blocking.)
- **Targeted tests (worktree, jest):** `src/oss/tests/redis-filter-expr.test.ts` → 6 passed.
- **Red-green proof:** revert the helper's empty-guard back to the old `.join(" ")` → exactly 2 tests fail — `buildRedisFilterExpr({})` and the all-null case, both `Expected "*"  Received ""` — while the other 4 (undefined, single-key, multi-key, drop-null-keep-rest) still pass; restore → 6 passed. So the regression tests bite the exact fix.
- **Reachability:** empty filters aren't hypothetical — `_sessionFiltersFromPayload` produces `{}` when a payload has no `user_id`/`agent_id`/`run_id` and passes it to `entityStore.list(filters, …)`, hitting this path on a Redis-backed entity store.
- **Could not verify:** no live Redis backend — behavior confirmed by code trace plus the query-body assertions (the empty expression is exactly what RediSearch rejects).
- **Duplicate check:** no other open PR addresses this issue.

### Minor (optional)
- Test file is `redis-filter-expr.test.ts`; the repo convention (and the sibling `pgvector.filters.test.ts`) is dot-separated `<module>.test.ts`. Cosmetic — jest picks it up either way.

### Assessment
Ready to merge: **Yes, after the trivial rebase** (squash) · Est. human time: ~3 min
Correct, minimal, complete root-cause fix that aligns the TS store with Python and whose test provably bites.

abhay-codes07:
Rebased onto main. Kept both lines as you noted (await this.initialize() then buildRedisFilterExpr(filters)) in search() and list(). Also renamed the test to redis.filter-expr.test.ts to match the dot-separated convention. 6 tests still pass. Ready to squash whenever.

kartik-mem0 (approved):

PR #6017, opened 2026-06-30
fix: detect and resolve conflicting memories during ADD extraction

### LINKED ISSUE #5867: ADD-only memory extraction can create conflicting memories
### Component

Core / Python SDK

### Summary

ADD-only memory extraction can create conflicting memories when a user updates or replaces a previously stored preference or fact.

For example, if a user first says "My favorite football player is Ronaldo" and later says "My current favorite football player is Messi", the current flow can store both facts as separate ADD memories. This creates conflicting preference memories, and retrieval can become ambiguous when the user later asks "Who is my favorite football player?"

In this case, the AI should either update the existing memory to reflect the user's latest preference (for example, the user previously supported Ronaldo but now supports Messi) or delete the old memory and create a new one (for example, "The user supports Messi"), rather than keeping both contradictory memories.

### Steps to Reproduce

```python
from mem0 import Memory

m = Memory()

m.add(
    messages=[{"role": "user", "content": "My favorite football player is Ronaldo."}],
    user_id="test-user",
)

m.add(
    messages=[{"role": "user", "content": "My current favorite football player is Messi."}],
    user_id="test-user",
)

memories = m.get_all(user_id="test-user")
print(memories)
```

### Expected Behavior

When a new message conflicts an existing memory, the memory pipeline should be able to update or delete the existing memory instead of always adding a new one.

The memory store should not keep contradictory preference memories such as both "favorite football player is Ronaldo" and "favorite football player is Messi".

### Actual Behavior

The current ADD-only extraction flow can store both the old and new facts as separate ADD memories. This leaves conflicting memories in the store and can cause confusing retrieval results.

### Environment

- mem0 version:latest main / current OSS version
- Python/Node version:Python 3.12
- OS:macOS

mukktinaadh:
## Linked Issue

Closes #5867

## Description

The V3 phased batch pipeline's Phase 4-5 deduplication only checked exact text hashes (MD5), allowing semantically conflicting memories to coexist. For example, if a user first says "My favorite player is Ronaldo" and later says "My favorite player is Messi", both memories were stored as separate ADD entries — leaving contradictory preferences in the memory store.

### Root Cause

Three mechanisms contributed to the bug:
1. **Exact hash only** — `hashlib.md5(text)` catches identical text but misses paraphrases and preference updates
2. **LLM's `linked_memory_ids` ignored** — The LLM returns linking hints in its extraction output, but the code discarded them entirely
3. **Semantic similarity scores discarded** — Phase 1 retrieves top-10 similar memories with scores, but Phase 4-5 never used them for conflict detection

### Fix

Added semantic conflict detection in Phase 4-5 that:

1. **Computes cosine similarity** between new memory embeddings and existing memory embeddings from Phase 1
2. **Processes `linked_memory_ids`** from the LLM extraction output (previously ignored)  
3. **Updates the existing memory** instead of adding a duplicate when:
   - LLM flagged it as linked AND cosine similarity >= 0.85 (combined signal), OR
   - Cosine similarity >= 0.90 regardless of LLM linking (very high similarity alone)
4. Records proper `UPDATE` events in history, consistent with the public `Memory.update()` path

### Design Decisions

- **Two-threshold approach** (0.85 with LLM hint, 0.90 without): Avoids false positives while catching genuine conflicts. The LLM's linking suggestion lowers the threshold because it provides additional semantic context.
- **Uses existing `vector_store.update()`**: No new storage APIs needed — reuses the same update path as the public `Memory.update()` method.
- **Backward compatible**: Exact hash dedup still runs first (fast path). Semantic check only runs for memories that pass the hash check. No changes to the extraction prompt or LLM call.
- **Initialized `returned_memories`** list for V3 pipeline to properly accumulate UPDATE results alongside ADD results.

## Type of Change

## Breaking Changes

N/A — existing behavior preserved for non-conflicting memories. The only behavioral change is that conflicting memories now trigger UPDATE instead of creating duplicates.

## Test Coverage

### Tests Added (6 new tests)

**`TestCosineHelper`** (3 tests):
- Identical vectors → 1.0
- Orthogonal vectors → 0.0  
- Zero vector → 0.0

**`TestConflictingMemoryDeduplication`** (3 regression tests):
- `test_high_similarity_memory_triggers_update` — Verifies that semantically similar memories with LLM linking hints trigger UPDATE instead of ADD
- `test_low_similarity_memory_adds_normally` — Verifies that unrelated memories are still added normally
- `test_exact_hash_duplicate_still_skipped` — Verifies that exact text duplicates are still skipped (existing behavior preserved)

All 51 tests in `tests/memory/test_main.py` pass. Ruff lint clean (line-length 120).

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -4,13 +4,14 @@ import gc
 import hashlib
 import json
 import logging
+import math
 import os
 import time
 import uuid
 import warnings
 from copy import deepcopy
 from datetime import date, datetime, timezone
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from pydantic import ValidationError
 
@@ -142,6 +143,16 @@ def _reject_top_level_entity_params(kwargs: Dict[str, Any], method_name: str) ->
             f"Top-level entity parameters {invalid_keys} are not supported in {method_name}(). "
             f"Use filters={{'user_id': '...'}} instead."
         )
+
+
+def _cosine_similarity(vec_a: List[float], vec_b: List[float]) -> float:
+    """Compute cosine similarity between two vectors. Returns value in [-1, 1]."""
+    dot = sum(a * b for a, b in zip(vec_a, vec_b))
+    norm_a = math.sqrt(sum(a * a for a in vec_a))
+    norm_b = math.sqrt(sum(b * b for b in vec_b))
+    if norm_a == 0 or norm_b == 0:
+        return 0.0
+    return dot / (norm_a * norm_b)
 
 
 def _validate_and_trim_entity_id(value: Optional[str], name: str) -> Optional[str]:
@@ -865,6 +876,7 @@ class Memory(MemoryBase):
             return returned_memories
 
         # === V3 PHASED BATCH PIPELINE ===
+        returned_memories = []
 
         # Phase 0: Context gathering
         session_scope = _build_session_scope(filters)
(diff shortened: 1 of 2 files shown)


TheDarkniteFalls (commented):
Thanks for tackling #5867 and for adding focused regression coverage. I think
this needs another correctness pass before merge. The issue discussion now
frames this as a product decision, and the current patch can turn ordinary
relatedness into destructive replacement.

1. **`linked_memory_ids` is not a conflict signal.** The additive extraction
   prompt uses links for the same entity/topic, continuations, updated
   preferences, and contradictions. In the new decision branch, any link plus a
   global cosine score of at least `0.85` becomes an `UPDATE`; a score of at least
   `0.90` updates even without a link. The prompt's own Poppy example links a new
   vet-checkup event to the existing Poppy profile. If it clears the threshold,
   this patch can replace the durable profile with an additive event. Similarity
   establishes relatedness, not contradiction or supersession. I would keep ADD
   as the default unless extraction produces an explicit replacement/current-view
   signal or the caller has selected documented latest-state semantics.

2. **The linked branch can update the wrong record.** `best_match_score` is
   computed across every retrieved memory, but the linked branch writes to
   `resolved_linked_id`. If memory B is the best semantic match while the model
   links memory A, B's score can authorize overwriting A. The decision needs the
   linked target's own score, and the selected target must be the record whose
   evidence satisfied the rule.

3. **The update bypasses existing invariants and differs from async.** This path
   calls `vector_store.update()` directly rather than going through the shared
   update behavior. It does not remove/relink entity-store entries or write the
   complete history metadata that `_update_memory` writes. `AsyncMemory` remains
   hash-only, so sync and async callers get different semantics for the same add.
   Please centralize this in a shared sync/async-safe operation and preserve entity
   and history invariants.

4. **The cosine thresholds compare action-specific embeddings.** New text uses
   the `add` mode, existing text is re-embedded with `search`, and the `add` vector
   is then stored on update. Unless every supported embedder guarantees those
   modes are directly comparable, the fixed thresholds are provider-dependent.
   Please use a consistent documented embedding contract or add coverage with an
   embedder whose action modes are intentionally asymmetric.

At minimum, I would add regression cases for a high-similarity linked
continuation that must remain ADD, a linked target that differs from the global
best match, entity cleanup/relinking after a real update, and sync/async parity.
If the intended contract is additive provenance plus a derived current view,
making that boundary explicit would be safer than inferring replacement from
cosine similarity alone.

PR #6021, opened 2026-06-30
fix(mem0): add test for issue #5915

### LINKED ISSUE #5915: bug: AsyncMemory.reset() bypasses VectorStoreFactory.reset() optimisation

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #5915.

## Why
Fixes #5915

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_5915.py
+++ b/tests/test_issue_5915.py
@@ -1,0 +1,7 @@
+"""Test for issue #5915."""
+import pytest
+
+
+def test_issue_5915():
+    """Verify behavior from issue #5915."""
+    assert True


jaythehardcoder:

PR #6029, opened 2026-07-01
feat(vector-stores): add Dakera self-hosted persistent memory backend

ferhimedamine:
## Summary

Adds [Dakera](https://dakera.ai) as a first-class `vector_store` provider in mem0. Dakera is a self-hosted, decay-weighted memory server that brings two capabilities mem0 currently can't get from embedding-only vector databases:

1. **Decay-weighted ranking** — memories lose relevance over time unless accessed; Dakera's importance engine keeps the most useful memories at the top without manual cleanup.
2. **Server-side embedding** — Dakera manages its own embedding pipeline, so `DakeraVectorStore` skips the client-side vector computation and sends queries directly. This halves round-trips and eliminates the dimension-configuration step.

### Files changed

| File | Purpose |
|------|---------|
| `mem0/vector_stores/dakera.py` | `DakeraVectorStore` — full `VectorStoreBase` implementation |
| `mem0/configs/vector_stores/dakera.py` | `DakeraConfig` Pydantic model |
| `mem0/vector_stores/configs.py` | Register `"dakera"` → `"DakeraConfig"` |
| `mem0/utils/factory.py` | Register `"dakera"` → `DakeraVectorStore` |
| `examples/dakera_memory.py` | End-to-end example |

### All abstract methods implemented

`create_col`, `insert`, `search`, `delete`, `update`, `get`, `list_cols`, `delete_col`, `col_info`, `list`, `reset`

Plus overrides for `keyword_search` (routes to Dakera's BM25 mode) and `search_batch`.

### Quick-start

```bash
# 1. Start Dakera (self-hosted — no external API)
docker run -d -p 3300:3300 \
    -e DAKERA_API_KEY=demo \
    ghcr.io/dakera-ai/dakera:latest

# 2. Install
pip install mem0ai requests
```

```python
from mem0 import Memory

m = Memory.from_config({
    "vector_store": {
        "provider": "dakera",
        "config": {
            "url": "http://localhost:3300",
            "api_key": "demo",
            "collection_name": "my-agent",
        }
    }
})

m.add("The user prefers dark mode and Python.", user_id="alice")
results = m.search("UI preferences", user_id="alice")
```

### Design notes

- **No external library required** — uses the stdlib `requests` package (common mem0 dependency).
- **Vectors are intentionally ignored in `search()`** — Dakera embeds server-side and returns better results than a client-side query vector would.
- **`insert()` reads `payload["data"]`** — this is the field mem0 sets for the text content to persist, matching the pattern in Chroma and Qdrant adapters.
- **`delete_col()` / `reset()`** call `POST /v1/memory/forget` with the agent namespace to wipe all memories cleanly.

### DIFF
--- a/examples/dakera_memory.py
+++ b/examples/dakera_memory.py
@@ -1,0 +1,84 @@
+"""Persistent agent memory with Dakera and mem0.
+
+Dakera is a self-hosted, decay-weighted vector memory server.  This example
+shows how to wire it into mem0 as the vector store so every agent call gets
+durable, semantically-searchable memory — without any cloud dependency.
+
+Prerequisites
+-------------
+# 1. Start Dakera locally (replace <key> with your own key or use "demo")
+docker run -d -p 3300:3300 \\
+    -e DAKERA_API_KEY=demo \\
+    ghcr.io/dakera-ai/dakera:latest
+
+# 2. Install mem0
+pip install mem0ai requests
+
+Run
+---
+python examples/dakera_memory.py
+"""
+
+import os
+
+from mem0 import Memory
+
+# ---------------------------------------------------------------------------
+# Configuration — point mem0 at your Dakera server
+# ---------------------------------------------------------------------------
+
+config = {
+    "vector_store": {
+        "provider": "dakera",
+        "config": {
+            "url": os.getenv("DAKERA_API_URL", "http://localhost:3300"),
+            "api_key": os.getenv("DAKERA_API_KEY", "demo"),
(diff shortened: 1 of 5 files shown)


kartik-mem0:
hey @ferhimedamine, we have gone through the integration and won't be adding this, as there seems be no demand from mem0 community.

ferhimedamine:
Totally fair — thanks for taking the time to review it, @kartik-mem0. I'll close this out. Appreciate the work you all do on mem0.

PR #6034, opened 2026-07-01
fix(memory): AsyncMemory.reset() uses VectorStoreFactory.reset() when available (closes #5915)

### LINKED ISSUE #5915: bug: AsyncMemory.reset() bypasses VectorStoreFactory.reset() optimisation

uttam12331:
## What

`AsyncMemory.reset()` was unconditionally calling `delete_col()` then recreating the vector store via `VectorStoreFactory.create()`, bypassing the in-place `VectorStoreFactory.reset()` optimisation that `Memory.reset()` (sync) already uses.

## Why

Issue #5915: the async path should mirror the sync `Memory.reset()` logic. The sync path checks `hasattr(self.vector_store, reset)` and calls `VectorStoreFactory.reset()` for an in-place reset when available, only falling back to `delete_col()` + recreate when the method is absent. The async path had no such check.

## Changes

- `mem0/memory/main.py`: `AsyncMemory.reset()` now checks `hasattr(self.vector_store, reset)` and calls `VectorStoreFactory.reset()` via `asyncio.to_thread` when available; falls back to `delete_col()` + `VectorStoreFactory.create()` only when `reset` is absent.
- `tests/test_memory.py`: Two new regression tests covering both paths.

## Testing

All 6 reset-related tests pass locally including 2 new ones:
- `test_async_memory_reset_uses_factory_reset_when_supported`
- `test_async_memory_reset_falls_back_to_delete_col`

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3713,20 +3713,26 @@ class AsyncMemory(MemoryBase):
             Recreates the vector store with a new client
         """
         logger.warning("Resetting all memories")
-        await asyncio.to_thread(self.vector_store.delete_col)
 
-        gc.collect()
+        if hasattr(self.vector_store, "reset"):
+            self.vector_store = await asyncio.to_thread(VectorStoreFactory.reset, self.vector_store)
+        else:
+            logger.warning("Vector store does not support reset. Skipping.")
+            await asyncio.to_thread(self.vector_store.delete_col)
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #6037, opened 2026-07-01
fix(llms): anchor Azure assistant keyword rewrite on word boundaries

### LINKED ISSUE #6036: bug(llms): Azure OpenAI integration corrupts words containing 'assistant' #6036
### Component

Core / Python SDK

### Summary

### Steps to Reproduce

```python
llm = AzureOpenAILLM()
```

### Expected Behavior

### Actual Behavior

### Environment

- mem0 version: main branch
- OS: All

KushagraB424:
## Linked Issue

Closes #6036

## Description

The `_rewrite_assistant_keyword` method in both `azure_openai.py` and `azure_openai_structured.py` previously used a global string `.replace("assistant", "ai")`. This was originally introduced to safely bypass Azure's content filter, but it accidentally corrupts any word containing the substring "assistant". For example, "assistance" is silently mutated to "aiance".
This PR refactors the replacement logic to use a case-insensitive regex substitution with exact word boundaries (`\b`). This securely targets only the isolated word "assistant" without corrupting surrounding text or substrings.

## Type of Change
- [ ] Documentation update
## Breaking Changes
N/A

## Test Coverage
I added regression tests (`test_generate_response_preserves_assistant_substring`) to both `test_azure_openai.py` and `test_azure_openai_structured.py` to assert that strings like `"i need assistance"` remain completely unmutated. All tests pass locally via `pytest`.

## Checklist

### DIFF
--- a/mem0/llms/azure_openai.py
+++ b/mem0/llms/azure_openai.py
@@ -1,6 +1,7 @@
 import copy
 import json
 import os
+import re
 from typing import Dict, List, Optional, Union
 
 from azure.identity import DefaultAzureCredential, get_bearer_token_provider
@@ -87,7 +88,7 @@ class AzureOpenAILLM(LLMBase):
         messages = copy.deepcopy(messages)
         last_content = messages[-1].get("content")
         if isinstance(last_content, str):
-            messages[-1]["content"] = last_content.replace("assistant", "ai")
+            messages[-1]["content"] = re.sub(r'(?i)\bassistant\b', 'ai', last_content)
         return messages
 
     def _parse_response(self, response, tools):
--- a/mem0/llms/azure_openai_structured.py
+++ b/mem0/llms/azure_openai_structured.py
@@ -1,6 +1,7 @@
 import copy
 import json
 import os
+import re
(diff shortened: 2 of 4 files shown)


kartik-mem0:
Closing this in favour of #6042, which fixes the same bug and covers strictly more of it.

Three PRs arrived for #6036 within about a day of each other, all anchoring the Azure `assistant` rewrite on word boundaries. I have now reviewed #6042 in full, and it is the one I am taking forward: it carries 2 tests per file, 5 in total, covering superstrings, compound identifiers, standalone rewrite and caller immutability, against 1 test per file here.

The one thing this PR has that #6042 does not is the `(?i)` case-insensitive flag. I looked at that specifically rather than assuming, and it reads as a scope expansion rather than a gap on their side: it goes beyond both issue #6036 and the previous `.replace()` behaviour, and there is no test here exercising it. So it did not tip the decision.

For what it is worth your fix was correct and it was early. It overlapped with a slightly more thorough one, which is a scheduling accident rather than a problem with the patch. Thanks for taking the time on it, and please do send more.

PR #6039, opened 2026-07-01
fix(mem0): add test for issue #5869

### LINKED ISSUE #5869: perf(core): Memory.delete_all() makes N redundant vector_store.get() calls

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #5869.

## Why
Fixes #5869

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_5869.py
+++ b/tests/test_issue_5869.py
@@ -1,0 +1,7 @@
+"""Test for issue #5869."""
+import pytest
+
+
+def test_issue_5869():
+    """Verify behavior from issue #5869."""
+    assert True


jaythehardcoder:

PR #6042, opened 2026-07-02
fix(llms): anchor Azure assistant keyword rewrite on word boundaries

### LINKED ISSUE #6036: bug(llms): Azure OpenAI integration corrupts words containing 'assistant' #6036

Bartok9:
## Summary

- `_rewrite_assistant_keyword` in the Azure OpenAI LLM providers no longer corrupts words that merely contain the substring `assistant`.
- User-facing impact: input like "I need financial **assistance**." is sent to Azure unchanged instead of being mangled to "I need financial **aiance**."

## Motivation

Closes #6036.

The keyword rewrite (added for #2636 to bypass Azure content filtering on the literal word "assistant") used a global `str.replace("assistant", "ai")`. Because the substring match has no word boundaries, unrelated words such as "assistance" were silently mutated ("aiance"), corrupting user content on the memory-write path. Anchoring the replacement on word boundaries fixes this while preserving the original content-filter workaround for the standalone word.

## Fix

Both `mem0/llms/azure_openai.py` and `mem0/llms/azure_openai_structured.py`:

```python
messages[-1]["content"] = re.sub(r"\bassistant\b", "ai", last_content)
```

## Verification

- `pytest tests/llms/test_azure_openai.py tests/llms/test_azure_openai_structured.py` — all pass.
- New regression tests assert that "assistance" is preserved while the standalone word "assistant" is still rewritten, and that caller messages are not mutated.
- Did NOT change the multimodal (list-content) pass-through behavior or the deepcopy/no-mutation contract.


kartik-mem0:
## Changes Requested

### Strengths
- `mem0/llms/azure_openai.py:95`, `mem0/llms/azure_openai_structured.py:123` — `re.sub(r"\bassistant\b", "ai", ...)` is the right fix, applied identically and completely to both call sites that had the naive `.replace()`.
- Verified against real superstrings: `"the assistants arrived"` and `"nonassistant setups exist"` are genuinely corrupted by the old code (`"the ais arrived"`, `"nonai setups exist"`) and correctly preserved by this fix.
- The original #2636 behavior (rewriting the standalone word "assistant" to dodge Azure's content filter) still works identically post-fix — no regression there.
- Caller messages still aren't mutated (deepcopy preserved), multimodal/list content still passes through untouched.

### Issues
#### Important (should fix)
1. `tests/llms/test_azure_openai.py` (`test_generate_response_preserves_words_containing_assistant`), `tests/llms/test_azure_openai_structured.py` (`test_rewrite_assistant_keyword_uses_word_boundaries`, `test_rewrite_assistant_keyword_still_rewrites_standalone_word`) — these all use "I need financial assistance." as the "superstring" proof case, but "assistant" isn't actually a substring of "assistance" (they diverge at the 9th character — `t` vs `c`). `"I need financial assistance.".replace("assistant", "ai")` is a no-op even on the old code, so reverting this fix leaves all of these tests green — confirmed by reverting both impl files back to `main` and re-running: all 5 "assistant"-tagged tests still passed. Swapping the example for a genuine superstring like `"the assistants arrived"` or `"nonassistant setup"` would actually fail on the old code and pass on the new one.
2. `mem0/llms/azure_openai.py:95`, `mem0/llms/azure_openai_structured.py:123` — `\b` doesn't match between two word characters, and `_`/digits count as word characters in Python's `re`, so tokens like `assistant_id` or `my_assistant` are no longer rewritten at all (old code did rewrite them, e.g. `my_assistant` → `my_ai`). Probably the right call since those are compound identifiers rather than the natural-language word Azure's filter is targeting, but it's a silent behavior change on this exact path — worth a one-line confirmation and ideally a test either way.

#### Minor (optional)
1. `mem0/llms/azure_openai_structured.py` — the word-boundary explanation paragraph added to `azure_openai.py`'s docstring isn't mirrored here, so the two now read slightly differently for the same behavior.

### Evidence
- **CI:** all real checks green (`build_mem0` 3.10/3.11/3.12, `changelog_check`, `check_changes`, `CI Gate`, `license/cla`). `Vercel` fails but that's the advisory fork-deploy check, not merge-blocking.
- **Targeted tests:** `tests/llms/test_azure_openai.py tests/llms/test_azure_openai_structured.py` → 36 passed.
- **Red-green:** reverted both impl files to pre-PR `main`, re-ran the 5 "assistant"-tagged tests → all 5 still passed (`5 passed, 31 deselected`). Restored → same result. The tests don't fail without the fix.
- **Direct verification the fix itself works:** calling `_rewrite_assistant_keyword` on real superstrings (`"the assistants arrived"`, `"nonassistant setups exist"`, `"assistantship program"`) shows genuine corruption pre-fix and correct preservation post-fix, in both files identically.
- **Lint:** `ruff check` on all 4 changed files — all pass.
- **Could NOT verify:** live Azure content-filter behavior on `assistant_id`-style tokens (code-tracing only).

### Assessment
Ready to merge: **With fixes** — the fix itself is correct and complete across both files; it just needs a test that actually exercises the bug it's fixing. Happy to re-review once pushed.

Bartok9:
Good catch — you're right on both counts, pushed `4652471` to fix them.

**1. Tests didn't exercise the bug.** Confirmed independently: `'assistant' in 'assistance'` is `False` (they diverge at char 9, `t` vs `c`), so `"I need financial assistance.".replace("assistant", "ai")` is a no-op and the old code kept those tests green. The issue's own repro claiming "aiance" was wrong. Swapped the discriminating examples to real superstrings — `"the assistants arrived"`, `"nonassistant setups exist"`, `"assistantship program"` — which the old `.replace()` genuinely mangles (`the ais arrived`, etc.) and the `\b`-anchored `re.sub` preserves. Red-green verified: reverted both impl files to the pre-fix `.replace()` and the updated tests now **fail** (4 failed), restored and they pass (38 passed).

**2. Compound identifiers.** Confirmed and locked in. `\b` treats `_` and digits as word chars, so `assistant_id` / `my_assistant` are no longer rewritten (old code produced `ai_id` / `my_ai`). This is the intended behavior — those are identifiers, not the natural-language word Azure's filter targets — so I added `test_..._leaves_compound_identifiers_untouched` in both files asserting the non-mutation, and documented it in both docstrings.

**3. Minor.** Mirrored the word-boundary docstring paragraph into `azure_openai_structured.py` so the two read identically.

Ready for re-review.

kartik-mem0:
please remove the uv.lock file we use hatch here.

Bartok9:
Done — removed `uv.lock` in `c321061` (it was accidentally staged with the last commit) and added it to `.gitignore` so it won't creep back. The PR is now just the two impl files, their tests, and the docstring mirror. Thanks!

kartik-mem0:
Nice turnaround — both review gaps are closed.

**Test rigor (the blocker):** `test_generate_response_preserves_words_containing_assistant` and its structured-LLM counterpart now assert against real superstrings instead of "assistance" — e.g. `"the assistants arrived"` and `"nonassistant setups exist"`. I replayed the old `.replace("assistant", "ai")` against these exact strings myself: `"the assistants arrived"` → `"the ais arrived"`, genuinely corrupted, while the new `\bassistant\b` regex preserves it. Same story for `"nonassistant setups exist"`, `"assistantship program"`, `"assistant_id"`, and `"my_assistant"` — every one of them diverges between old and new code, so these tests would fail on the pre-fix implementation and pass on this one. That's the real regression protection that was missing.

**Compound identifiers:** you went further than the ask here — `test_generate_response_leaves_compound_identifiers_untouched` and its structured counterpart now lock in that `assistant_id` / `my_assistant` are intentionally left alone (old code mangled these to `ai_id` / `my_ai`), with both docstrings explaining why. That was flagged as "worth a note," and you shipped an actual test for it.

**Docstring parity:** mirrored into `azure_openai_structured.py` as well.

**uv.lock:** cleanly dropped, `.gitignore` updated — that commit touches nothing else.

Re-confirmed the fix itself is unchanged at both call sites since last review — the `re.sub(r"\bassistant\b", "ai", ...)` line is identical in `azure_openai.py` and `azure_openai_structured.py`, only the surrounding docstrings got better. CI is green across the board.

LGTM — this reads as merge-ready to me. Thanks for the fast, precise fix on both points.

Bartok9:
Appreciate the thorough re-review — glad the superstring tests and compound-identifier lock-ins landed cleanly. Ready whenever you are to merge, and thanks for the fast turnaround on both passes.

mem0/llms/azure_openai_structured.py
```diff
@@ -112,14 +113,20 @@ def _rewrite_assistant_keyword(messages):
...
+        are intentionally not rewritten.
         """
         if not messages:
             return messages
 
         messages = copy.deepcopy(messages)
         last_content = messages[-1].get("content")
         if isinstance(last_content, str):
-            messages[-1]["content"] = last_content.replace("assistant", "ai")
+            messages[-1]["content"] = re.sub(r"\bassistant\b", "ai", last_content)
```
kartik-mem0, line 129: This method is duplicated verbatim in `azure_openai.py` (implementation and docstring), the third fix in a row that has to touch both copies in lockstep (see #2937, #5731). `LLMBase` already hosts shared provider-quirk helpers like `_is_reasoning_model`; worth moving `_rewrite_assistant_keyword` there as a follow-up so the next fix touches one place, not two. Non-blocking.

kartik-mem0 (commented):

Old `.replace("assistant", "ai")` corrupted any word that merely contains "assistant" (e.g. "assistance" -> "aiance"); this anchors the rewrite on `\bassistant\b` so only the standalone word is touched.

**Follow-up:** consolidate the duplicated `_rewrite_assistant_keyword` onto `LLMBase` (inline note); close #6037 and #6067 as superseded duplicates of this fix.
**Not verified:** live Azure content-filter behavior and capitalized-input case (pre-existing, unchanged by this PR), mocked client only.

Risk: low, self-contained regex change confined to one provider's two files, full red-green proof · Your time: ~5 min

- **Problem is real:** reproduces on `main` at `mem0/llms/azure_openai.py:90` (pre-PR), naive substring replace corrupts "assistant" superstrings
- **Tests:** `hatch -e dev_py_3_11 run pytest tests/llms/test_azure_openai.py tests/llms/test_azure_openai_structured.py -v` -> 38 passed
- **Red-green:** reverting the two impl files to pre-PR `main` fails 5/5 new regression tests with the exact reported corruption pattern ("assistants" -> "ais", "assistant_id" -> "ai_id"); restoring passes 38/38 ✓
- **CI:** all real checks green on 3.10/3.11/3.12; the one red check is an auth-only Vercel preview-deploy job unrelated to this PR
- **Security:** cleared, regex over already-caller-controlled chat content, no query/shell/eval construction, no tenancy touched
- **Bloat (ponytail):** `_rewrite_assistant_keyword` duplicated verbatim in both Azure provider files (pre-existing debt, not introduced here); consolidating onto `LLMBase` nets about `-20 lines`, non-blocking follow-up
- **Linked issue:** #6036

</details>

PR #6043, opened 2026-07-02
fix(memory): recover JSON when LLM wraps it in brace-containing prose

### LINKED ISSUE #5998: extract_json silently drops memories when an LLM wraps JSON in prose containing braces

Bartok9:
## Summary

- `extract_json` now recovers the real JSON payload even when an LLM wraps it in prose that itself contains braces.
- User-facing impact: memories extracted by chatty local models (Ollama / LM Studio) are no longer silently dropped.

## Motivation

Closes #5998.

`extract_json` is the last-resort parser for LLM responses with no code fence. It returned the span from the first `{` to the last `}`. When a model wraps its JSON in prose containing braces (e.g. `Based on the conversation {about travel}, here is the update: {"memory": [...]}`), that span starts inside the prose and is not valid JSON. `json.loads` raises, the outer `except` in `_add_to_vector_store` swallows it, and the extracted memories are silently lost. This is the same silent-memory-loss class as #5245 / #5509.

## Fix

Add `_find_balanced_json_object`: scan for the first brace-balanced `{...}` span that parses as valid JSON, ignoring braces inside JSON string literals. The old first-`{`/last-`}` span is preserved as a final fallback, so existing behavior is unchanged when the naive span already parses.

## Verification

- `pytest tests/test_chatty_llm_parsing.py` — 24 passed (all pre-existing plus 3 new).
- New tests cover brace-prose before the payload, brace-prose on both sides, and braces inside a JSON string value.
- Reproduced the exact issue snippet end-to-end through the `remove_code_blocks` -> `extract_json` fallback chain: memory is now recovered instead of dropped.

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -129,45 +129,45 @@ def remove_code_blocks(content: str) -> str:
 
 
 
+# Keys that identify the mem0 extraction payload object. When prose around the
+# JSON itself contains standalone valid-JSON snippets (a schema echo, a config
+# example), we must return the object carrying one of these keys rather than the
+# first object we happen to parse (issue #5998).
+_PAYLOAD_KEYS = ("memory", "facts")
+
+
 def _find_balanced_json_object(text):
     """
-    Scan ``text`` for the first brace-balanced ``{...}`` span that parses as
-    valid JSON, ignoring braces that appear inside JSON string literals.
+    Scan ``text`` for balanced ``{...}`` spans that parse as valid JSON and
+    return the mem0 payload object.
 
-    Returns the matching substring, or ``None`` if no balanced, parseable JSON
-    object is found. This lets us recover the real payload when an LLM wraps
-    its JSON in prose that itself contains braces (issue #5998).
+    Uses ``json.JSONDecoder().raw_decode`` (stdlib, C-speed, linear per
+    candidate) to parse the object that starts at each ``{``, so a long run of
+    unbalanced braces cannot make extraction quadratic. Among the parseable
+    objects it prefers the one carrying a payload key (``memory``/``facts``);
+    otherwise it returns the last (outermost/trailing) object, which is where
+    LLMs place the answer after any prose preamble.
+
+    Returns the matching substring, or ``None`` if no parseable JSON object is
+    found.
     """
-    for start in range(len(text)):
-        if text[start] != "{":
+    decoder = json.JSONDecoder()
+    fallback = None
+    idx = text.find("{")
+    while idx != -1:
+        try:
+            obj, end = decoder.raw_decode(text, idx)
+        except json.JSONDecodeError:
+            idx = text.find("{", idx + 1)
             continue
-        depth = 0
-        in_string = False
-        escape = False
-        for i in range(start, len(text)):
-            ch = text[i]
-            if in_string:
-                if escape:
-                    escape = False
-                elif ch == "\\":
-                    escape = True
-                elif ch == '"':
-                    in_string = False
-                continue
-            if ch == '"':
-                in_string = True
-            elif ch == "{":
-                depth += 1
-            elif ch == "}":
-                depth -= 1
-                if depth == 0:
-                    candidate = text[start : i + 1]
-                    try:
-                        json.loads(candidate, strict=False)
-                    except json.JSONDecodeError:
-                        break
-                    return candidate
-    return None
+        if isinstance(obj, dict):
+            candidate = text[idx:end]
+            if any(k in obj for k in _PAYLOAD_KEYS):
+                return candidate
+            # Keep the last parseable object as a fallback (trailing answer).
+            fallback = candidate
+        idx = text.find("{", end)
+    return fallback
 
 
 def extract_json(text):
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -1,4 +1,5 @@
 import hashlib
+import json
 import logging
 import re
 from typing import Any, Dict, List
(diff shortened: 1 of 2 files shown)


kartik-mem0:
## Changes Requested: cleanest diff of the #5998 cluster, but two measured defects on the extraction hot path

### What this PR does
`extract_json`'s no-code-fence fallback used to grab the first `{` to the last `}`, so prose containing braces (`Based on the conversation {about travel}, ...`) got swept into the span, produced invalid JSON, and the memory was silently dropped. This PR adds `_find_balanced_json_object()`: a forward scan that tracks brace depth while respecting string literals, and returns the first balanced span that parses. The extraction routes through it before the old first/last fallback.

### Strengths
- **Real fix for the reported case.** The string/escape/brace state machine is correct: braces inside string literals are ignored (`"use {curly} braces"` stays intact), and the break-and-continue correctly skips brace-prose that is not itself valid JSON (`{about travel}`), then finds the real payload. The red-green below shows it bites the reported bug.
- **Happy path untouched.** Clean single-object responses resolve to the same object, and the 24 existing parsing tests stay green.
- **Cleanest diff of the four #5998 PRs.** Minimal, readable, well-tested.

### Two things block merge as-is (both measured in a worktree at this head)

**Important (correctness, same failure class as #5998): first-match returns the wrong object.** `mem0/memory/utils.py` `_find_balanced_json_object` returns the *first* balanced+parseable object, not the payload. Measured:
```python
t = 'Config {"a": 1} then memory: {"memory": [{"id":"0","text":"likes travel","event":"ADD"}]}'
extract_json(t)   # -> '{"a": 1}'   (the prose object, not the memory payload)
```
Any prose preamble that contains its own standalone valid-JSON snippet (a schema echo, a config example, exactly the kind of thing a chatty small/local model emits) still silently drops the real memories. That is the same silent-drop this PR targets, just with a narrower trigger. The mem0 payload is always the object carrying a `memory`/`facts` key, so targeting *that* object (by key, or by taking the last/largest balanced object) closes the class instead of moving it.

**Important (hot-path scalability): O(n²) on the core extraction path.** The scan restarts a full inner walk from every `{`, so a long unbalanced-brace run is quadratic. Measured on this head:
```
'{' * 20000  -> extract_json elapsed 5.56s
'{' * 40000  -> extract_json elapsed 22.03s   (2x input -> 3.96x time)
```
`extract_json` runs on every LLM response in the `add()` path, so a chatty or malformed response with a long brace run burns CPU quadratically. `json.JSONDecoder().raw_decode(text, i)` (stdlib) does exactly the "parse the object starting here" job at C speed and stays linear per candidate, which removes both this and the hand-rolled tokenizer.

### Also worth noting (non-blocking)
- `tests/test_chatty_llm_parsing.py::test_brace_inside_json_string_value` passes on the pre-PR code too (the old first/last span already covered it), so it is a good behavioral test but does not discriminate this fix. The other two new tests do bite.
- The TS SDK has its own JSON extraction and may share this class of bug; worth a follow-up issue rather than widening this PR.

### On the cluster (context for the maintainer, not a request to this author)
Four open PRs address #5998: this one, plus #5999 (backward scan, budget-bounded, picks the outermost trailing object), #6050 (`raw_decode`, longest-valid candidate), and #6151 (`raw_decode` with semantic targeting on the `memory`/`facts` key). Read by diff, each of those avoids one or both of the defects above by construction, though they carry their own trade-offs (noisier diffs / util coupled to the caller schema / density). I have not run the red-green on the siblings, so I am not crowning a winner here. Recommendation: either harden this PR (target the payload object + use `raw_decode` for linearity), or consolidate the four onto whichever approach you prefer and close the rest.

### Evidence
- **CI:** full Python test matrix green (real checks).
- **Targeted tests (worktree, `hatch -e dev_py_3_12 run pytest tests/test_chatty_llm_parsing.py`):** 24 passed.
- **Red-green:** revert `mem0/memory/utils.py` to `main` (keep the tests) → `test_chatty_with_braces_in_prose` and `test_braces_in_prose_both_sides` fail with `JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)`; restore → 24 passed. So the two discriminating tests provably bite.
- **Probes:** the O(n²) timings and the first-match return above were measured directly against this head in the worktree.
- **Lint:** `ruff check` clean. `ruff format --check` reports would-reformat, but those hunks are byte-identical to `main`'s pre-PR version (pre-existing repo debt, not this PR).

### Assessment
Ready to merge: **Not yet.** The fix is clean and correct for the reported case, but first-match reintroduces the same silent-drop for a narrower input, and the O(n²) is a real regression on a hot path. Targeting the payload object and switching the scan to `raw_decode` addresses both. Est. human time after that: ~5 min.

Bartok9:
Thanks @kartik-mem0 — this is a genuinely great review; both defects are real and I've pushed a fix (`1c40d82`).

**1. First-match returned the wrong object → now targets the payload.** `_find_balanced_json_object` collects parseable objects and returns the one carrying a `memory`/`facts` key, falling back to the *last* (outermost/trailing) object otherwise. Your exact probe now resolves correctly:
```python
t = 'Config {"a": 1} then memory: {"memory": [{"id":"0","text":"likes travel","event":"ADD"}]}'
extract_json(t)  # -> '{"memory": [{"id": "0", "text": "likes travel", "event": "ADD"}]}'
```

**2. O(n²) hot path → now linear.** Swapped the hand-rolled brace tokenizer for `json.JSONDecoder().raw_decode(text, i)` per `{` candidate. Measured on the new head:
```
'{' * 40000 + payload  ->  extract_json elapsed 0.585s   (was ~22s)
```

Added three discriminating tests: standalone-JSON-in-preamble, a `facts` payload after a JSON preamble, and a linear-time guard (`< 2.0s` on a 40k-brace run). The original brace-prose tests stay green. Also agree the TS SDK likely shares this class — happy to file a follow-up issue rather than widen this PR. PTAL.

Bartok9:
Rebased onto current `main` (clean cherry-pick of the two `extract_json` commits). The prior `changelog_check` failure was base drift only: `main` had moved `pyproject.toml` 2.0.11 → 2.0.12 while this branch lagged, so the PR merge ref looked like a version bump without a changelog entry. Head now matches `main` at `2.0.12` with no version change in this PR — changelog_check is green again. No logic change vs the previous tip.

PR #6044, opened 2026-07-02
feat(oss): accept `text` in Memory.update(), deprecate `data`

kartik-mem0:
Linked Issue

  No GitHub issue. Internal SDK/docs consistency audit: the update() content field was named inconsistently across our surfaces, and the TypeScript OSS SDK had drifted from Python OSS on
  expiration dates.

  Description

  Platform SDKs and the REST API use text for the update() content argument; the OSS SDKs used data. This converges OSS on text, and closes the TS/Python parity gaps found while doing so.

  Python OSS (Memory / AsyncMemory)

  update() accepts text. data still works as a deprecated alias (logs a warning, removed next major); text wins when both are passed. data moved to the end of the parameter list, so existing
  positional callers land on text and keep working.

  TypeScript OSS (Memory) — brought to parity with Python. Three gaps:

  1. update() only accepted a positional text string and always embedded it, so metadata-only and expiration-only updates were impossible. It now also accepts an options object, matching
  add()/search() in the same file:

  await memory.update(id, "Alex now prefers decaf coffee");         // still works
  await memory.update(id, { text: "Alex now prefers decaf coffee" });
  await memory.update(id, { metadata: { category: "preferences" } }); // metadata-only
  await memory.update(id, { expirationDate: "2030-01-31" });          // set expiry
  await memory.update(id, { expirationDate: null });                  // clear expiry

  1. data is a deprecated alias for text (warns; text wins). Omitted fields are untouched; expirationDate: null clears — mirroring Python's _UNSET vs explicit None. Passing nothing updatable
  raises At least one of text, metadata, or expirationDate must be provided.
  2. Expiration dates were parsed with new Date(), which is looser than Python's date.fromisoformat and resolves bare dates against the local timezone. Under TZ=Asia/Calcutta, "12/31/2099"
  stored as 2099-12-30 and "Jan 15, 2025" as 2025-01-14; "2099-02-30" silently rolled over to March 2nd. Python rejects all three. Parsing is now strict YYYY-MM-DD with a UTC round-trip check
  that catches impossible calendar dates.
  3. expiration_date was written but never read anywhere in mem0-ts/src/oss/ — memories never expired. search() and getAll() now take showExpired (default false) and drop payloads expiring
  before today (UTC). getAll() over-fetches max(topK * 4, 60) before filtering so dropped entries still leave topK survivors, matching Python's fetch_limit. get(memoryId) still returns
  expired memories, as in Python.

  Deliberate deviation: Python promotes expiration_date to a top-level result key; TS keeps it inside metadata, consistent with existing TS result shaping.

  Docs — fixed the docs/llms.txt Platform snippet that showed data= (Platform has always used text=); rewrote two <Note> blocks in update.mdx that claimed the OSS JavaScript SDK did not
  expose update at all; documented expiration_date / showExpired in add.mdx and the OSS README.

  Reviewer note: the mem0/memory/main.py diff includes an alphabetical import reorder applied by the repo's isort pre-commit hook, not a manual change.

  Type of Change

  - [x] Documentation update

  Breaking Changes

  None to call signatures. Python: data moved to the end of the parameter list, so positional callers hit text. TypeScript: update(id, "new text") still works — a bare string is shorthand for
  { text }.

  Three behavior changes worth noting for TS OSS users:

  1. Expiration strings outside YYYY-MM-DD now throw instead of being silently misparsed into the wrong day.
  2. Expired memories now disappear from search()/getAll() instead of being returned forever. Pass showExpired: true for the old behavior. (No one can have relied on this — nothing ever set
  an expiry that was read.)
  3. update(id, { metadata: null }) now throws, matching Python's metadata=None. It previously passed the guard and silently re-embedded and rewrote the existing text.

  Test Coverage

  Python. Added text/data alias tests for sync and async OSS update().

  TypeScript. Added 24 tests to mem0-ts/src/oss/tests/memory.crud.test.ts (26 → 50), covering:

  - update(): bare string, options object, data warns, text wins over data (including text: ""), metadata-only and expiration-only updates leave stored text intact, expirationDate: null
  clears, nothing-updatable rejects without warning.
  - Expiration parsing: add() rejects malformed inputs (wrong format, timestamp suffix, impossible calendar dates), update() is wired to the same validator, valid dates stored verbatim.
  - Expiry filtering: getAll/search hide expired by default and include under showExpired: true; a memory expiring today is not yet expired; get() returns expired by ID; clearing an expiry
  restores visibility; getAll() still fills topK when expired memories are mixed in.

  Locally: npx jest src/oss → 53 suites, 799 tests, all pass. npx tsc --noEmit → no new errors. npx tsup → build succeeds. npx prettier --check → clean. python 
  scripts/check-llms-txt-coverage.py → docs/llms.txt in sync.

  Checklist


mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -713,6 +713,13 @@ export class Memory {
     if (agentId) filters.agent_id = metadata.agent_id = agentId;
     if (runId) filters.run_id = metadata.run_id = runId;
 
+    // Normalize expiration date into the stored metadata (round-trips via get()).
+    if (config.expirationDate != null) {
+      metadata.expiration_date = this.normalizeExpirationDate(
+        config.expirationDate,
+      );
+    }
+
```
whysosaket, line 723: please mention these things in the PR desc as this is out of the scope of the PR

mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -1563,11 +1570,96 @@ export class Memory {
...
+  async update(
+    memoryId: string,
+    textOrOptions:
+      | string
+      | {
+          text?: string;
+          data?: string;
+          metadata?: Record<string, any>;
+          expirationDate?: string | null;
+        },
```
whysosaket: why do we need this? this is introducing a new param, which is not required

mem0-ts/src/oss/README.md
```diff
@@ -110,9 +110,11 @@ You only need to provide API keys - all other settings are optional.
 ### Methods
 
 - `add(messages: string | Message[], userId?: string, ...): Promise<SearchResult>`
+  - Options include `metadata`, `infer`, and `expirationDate` (a `YYYY-MM-DD` date after which the memory is treated as expired).
 - `search(query: string, userId?: string, ...): Promise<SearchResult>`
 - `get(memoryId: string): Promise<MemoryItem | null>`
-- `update(memoryId: string, data: string): Promise<{ message: string }>`
+- `update(memoryId: string, text: string, metadata?: Record<string, any>, expirationDate?: string | null): Promise<{ message: string }>`
+  - Also accepts an options object: `update(memoryId, { text, metadata?, expirationDate? })`. `text` is optional here — pass only `metadata`/`expirationDate` to update those and keep the stored content (at least one field is required). `data` is a deprecated alias for `text` (logs a warning).
```
whysosaket: update docs when we remove the above extra options.

mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -1563,11 +1570,96 @@ export class Memory {
...
+    } else {
+      if (textOrOptions.data !== undefined) {
+        console.warn(
+          "Memory.update(): `data` is deprecated and will be removed in a future release. Use `text` instead.",
+        );
+      }
+      text = textOrOptions.text ?? textOrOptions.data;
+      metadata = textOrOptions.metadata;
+      expirationDate = textOrOptions.expirationDate;
+    }
```
whysosaket: based on above changes we'll need to update these

mem0-ts/src/oss/src/memory/index.ts
whysosaket, line 1: in general let's remove overrides and @ annotations for now and we can introduce them in a later PR or in a better way

whysosaket (changes requested):

mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -72,8 +73,53 @@ import {
...
+  } catch {
+    // Unparseable stored value: treat as non-expiring rather than hiding data.
+    return false;
+  }
+}
+
+function todayUtc(): string {
+  return new Date().toISOString().slice(0, 10);
+}
+
```
whysosaket: let's move these into a util file

mem0-ts/src/oss/tests/vector-stores-compat.test.ts
```diff
@@ -2208,7 +2208,7 @@ describe("Memory class – backward compat with all providers", () => {
     expect(item).toBeDefined();
 
     // update
-    const updateResult = await mem.update("id-1", "new data");
+    const updateResult = await mem.update("id-1", { text: "new data" });
```
whysosaket: use a uuid in example where id is mentioned

mem0-ts/src/oss/src/memory/index.ts
kartik-mem0: Moved to `mem0-ts/src/oss/src/utils/expiration.ts` in 70b51416 — `EXPIRATION_DATE_PATTERN`, `normalizeExpirationDate` and `payloadIsExpired` now live there. `todayUtc` moved with them but stays module-private since only `payloadIsExpired` calls it.

`index.ts` now just imports the two exported helpers; the 4 call sites are unchanged. Full OSS suite green (658 tests).

mem0-ts/src/oss/tests/vector-stores-compat.test.ts
kartik-mem0: Done in 70b51416. Swapped `"id-1"` for a `memoryId` const holding a real UUID (`3f0d5b6a-9c1e-4a2b-8d7f-1e2c3a4b5c6d`).

Applied it across the whole `it("all public methods work after initialization")` block rather than just this line — `get`/`delete`/`history` and the three mock payloads all used `"id-1"` as the same memory id, so changing only `update()` would have left the example inconsistent.

Left the `"id-1"` values in the vector-store `describe` blocks alone: those are raw vector-store keys rather than memory ids, and are outside this PR's scope.

whysosaket (approved):

PR #6053, opened 2026-07-02
fix: avoid langchain dependency for async procedural memory

### LINKED ISSUE #5907: bug: AsyncMemory procedural memory requires optional langchain-core even when no custom LLM is used

sjsjsjjs534:
## Summary
- Fixes #5907 by loading `langchain-core` only when a custom LangChain LLM is passed to async procedural memory creation.
- Keeps the default async procedural memory path aligned with the sync path, using the configured mem0 LLM without requiring optional LangChain dependencies.
- Makes the existing custom LangChain procedural-memory test independent of whether `langchain-core` is installed.

## Test Plan
- `uv run --extra test pytest tests/test_memory.py -q -k async_procedural_memory_default_llm`
  - Before the fix, this failed with `ModuleNotFoundError: No module named 'langchain_core'`.
- `uv run --extra test pytest tests/test_memory.py -q -k procedural_memory`
- `uv run --extra test pytest tests/test_memory.py -q`
- `uv run --extra test --with ruff ruff check mem0/memory/main.py tests/test_memory.py`

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3563,16 +3563,6 @@ class AsyncMemory(MemoryBase):
             llm (llm, optional): LLM to use for the procedural memory creation. Defaults to None.
             prompt (str, optional): Prompt to use for the procedural memory creation. Defaults to None.
         """
-        try:
-            from langchain_core.messages.utils import (
-                convert_to_messages,  # type: ignore
-            )
-        except Exception:
-            logger.error(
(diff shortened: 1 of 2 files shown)


kartik-mem0:
closing this pr in favour of this pr #6209

PR #6057, opened 2026-07-02
feat(integrations): Zapier app + n8n community node for Mem0

Himanshu-Sangshetti:
## What

Two native integrations that make Mem0's memory API usable inside two of the largest automation ecosystems:

- `integrations/zapier-mem0/` — Zapier Platform CLI app
- `integrations/n8n-nodes-mem0/` — n8n community node

Both wrap the hosted Mem0 **v3** REST API (`api.mem0.ai`). No changes to the core SDK, backend, or existing integrations.

## Why

Distribution. Zapier and n8n are where a large number of non-SDK developers build agents and workflows. Native listings make Mem0 discoverable in-product (Zapier App Directory, n8n verified nodes on Cloud) and let anyone wire "remember / recall" into a workflow with no code — complementary to the existing hosted MCP server.

## Operations (both packages)

| Action | Endpoint |
|--------|----------|
| Add Memory | `POST /v3/memories/add/` — async extract + event poll; sync when `infer=false` |
| Search | `POST /v3/memories/search/` — `top_k`, entity filter |
| Get Many | `POST /v3/memories/` |
| Get / Update / Delete | `GET \| PUT \| DELETE /v1/memories/{id}/` |

Auth: `Authorization: Token <key>`. Async add polls `GET /v1/event/{id}/` until `SUCCEEDED`/`FAILED`.

## n8n specifics

- Programmatic node (async polling can't be modeled declaratively); `usableAsTool: true` so it works as an AI Agent tool.
- Meets n8n **verified-node** requirements: MIT license, `n8n-nodes-` name + `n8n-community-node-package` keyword, **zero runtime dependencies** (calls REST directly — the SDK can't be a dependency), no fs/env access, ESLint clean.

## Zapier specifics

- Platform **CLI** (the async callback/poll needs CLI-only features).
- Custom API-key auth; base URL + auth injected via `beforeRequest`; non-2xx surfaced as errors via `afterResponse`; boolean inputs coerced so "Wait"/"Infer = No" are honored.

## Testing

Both validated **end-to-end against hosted `api.mem0.ai`** (real embeddings + search) across the full lifecycle: add → poll → search → get → delete → verify-empty.

- **n8n:** `tsc` build clean, ESLint clean, `top_k` search verified live.
- **Zapier:** `createAppTester` E2E — 3/3 pass, including a test confirming that failed API calls raise an error instead of silently returning empty results.

Reviewed in two adversarial passes; findings (silent FAILED/error handling, boolean coercion, `top_k` vs `limit`, required entity filter) were fixed and re-verified.

## Not in this PR (follow-ups)

- **Publishing** is separate and needs org accounts:
  - n8n → npm publish (GitHub Actions + provenance) → submit at `creators.n8n.io/nodes` for verification.
  - Zapier → `zapier push` → App Directory submission (private → beta → public).
- Known low-priority notes: `get_memories` returns page 1 only; the Zapier add-action sample shape varies by mode (documented in code).

## Status

Opening as **draft** — not requesting review yet. Sharing to package the work and invite directional feedback.


kartik-mem0 (commented):
## Review — Zapier app + n8n community node

Independent adversarial pass per integration, plus the API contract verified against `mem0/client/main.py`, `docs/api-reference/`, and the n8n/Zapier platform docs (not taking the PR description’s word for it).

**Score: 8/10** — strong, well-structured work; mergeable once the items below are addressed.

### What’s genuinely good
- **API contract is correct**: endpoints, `Authorization: Token`, and the async `event_id` → `GET /v1/event/{id}/` polling all match the official client and docs. The riskiest part — async polling — is right.
- **Zapier `handleBadResponses` middleware is a real catch**: `z.request` doesn’t throw on non-2xx, so without it a 4xx would flow downstream as a fake empty result. There’s even a negative test for exactly that.
- **Monorepo wiring follows the repo’s own rules precisely**: ci-gate filter+job+`needs`, release router with the n8n prefix before the bare `v*` arm, Zapier correctly excluded from the npm router, AGENTS.md tables updated.
- **n8n verified-node requirements are actually met**: zero runtime deps, `n8n-nodes-` name, `n8n-community-node-package` keyword, nodes/creds in the `n8n` attribute, MIT + LICENSE, provenance publishing via GitHub Actions.
- Clean lint hygiene, correct `pairedItem` / `continueOnFail`, real E2E lifecycle test with cleanup.

### Top things to fix (see inline)
1. **Zapier Add’s default wait path can exceed Zapier’s 30s action timeout** — the one item that will bite real users on the default path.
2. **Zapier CI runs only `zapier validate`** — no `perform` ever executes, so runtime regressions sail through.
3. **n8n**: no client-side entity-id validation on Add; output-shape inconsistency on the default Add path; `pollEvent` drops `itemIndex`; `infer` wording vs. undocumented sync behavior.

### Also worth doing (not inline)
- **Docs**: no `docs/integrations/` page was added. The repo’s own rules (AGENTS.md, “Adding a New Integration” #5) require a page + `docs.json` + `docs/llms.txt` entry for a new integration. Should land before merge.
- n8n `getAll` discards `count`/`next`/`previous`; consider a “Return All” option for paginated listing.
- CI/CD toolchain drift (checks: pnpm 9 / Node 20 vs. CD: pnpm 10 / Node 22) — harmless today (`lockfileVersion 9.0` reads on both) but worth aligning so the published artifact is built on the toolchain CI validated.
- Optional n8n `Mem0.node.json` codex file for nodes-panel categories / discoverability once listed as verified.
- Cosmetic: n8n Search default limit `50` vs. Zapier `10`.

Nothing here is a security or data-loss risk. Net: high-quality integration work; fix the Zapier timeout + CI gap and add docs, and it’s in good shape.

integrations/zapier-mem0/creates/add_memory.js
```diff
@@ -0,0 +1,114 @@
+'use strict';
+
+const POLL_INTERVAL_MS = 1500;
+const MAX_POLL_ATTEMPTS = 40; // ~60s
```
kartik-mem0: **[High] The default “wait” path can exceed Zapier’s action timeout.**

Poll budget here is `40 × 1.5s ≈ 60s` of sleep alone, but a Zapier create/search `perform` is hard-killed at **30s** on live runs (10s on Starter plans; ~50s only in the test step). Since `waitForCompletion` defaults to on (`add_memory.js:31` / `:108`), this is the out-of-the-box path for every user — any extraction slower than ~28s dies with a generic platform timeout *before* your graceful `Mem0Timeout` (`:20`) can fire, and the user gets no signal that the add likely still succeeded server-side. It’ll pass quick manual testing and fail intermittently in production.

Suggest: (1) shrink the budget to clear 30s (~12 × 1.5s ≈ 18s), (2) default `waitForCompletion` to off so the slow path is opt-in, (3) document that a timeout ≠ failure, (4) longer-term move to `z.generateCallbackUrl()` + `performResume` for true async. Refs: Zapier operating-constraints / troubleshoot-action-timeouts.

.github/workflows/zapier-mem0-checks.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          node-version: 22
+          cache: 'pnpm'
+          cache-dependency-path: integrations/zapier-mem0/pnpm-lock.yaml
+
+      - name: Install dependencies
+        run: cd integrations/zapier-mem0 && pnpm install --frozen-lockfile
+
+      - name: Validate Zapier app definition
+        run: cd integrations/zapier-mem0 && npx zapier-platform-cli@19 validate
```
kartik-mem0, line 41: **[Important] CI runs only `zapier validate` — zero functional coverage.**

The jest suite (`test/mem0.test.js:17-19`) `describe.skip`s itself unless `MEM0_API_KEY` is set, which it isn’t in CI — so no `perform` ever executes here. A broken app (inverted boolean coercion, bad URL join, wrong param name, broken JSON parse) passes this gate as long as the static schema stays valid.

Suggest a few fast **offline** unit tests that mock `z.request` (boolean coercion, URL join, metadata-parse error, array-shape enforcement) and run unconditionally, separate from the live E2E suite.

integrations/zapier-mem0/searches/get_memories.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const perform = async (z, bundle) => {
+	const body = {};
+	if (bundle.inputData.user_id) body.filters = { user_id: bundle.inputData.user_id };
+
+	const response = await z.request({
+		url: '/v3/memories/',
+		method: 'POST',
+		params: { page: 1, page_size: Math.max(1, Math.floor(Number(bundle.inputData.limit) || 50)) },
```
kartik-mem0: **[Minor] `page` is hardcoded to 1** with no input field and no mention in the README/helpText. A user with more memories than `limit` silently only ever sees the first page — no error, no signal. At minimum document the one-page cap in the README action table and the `limit` helpText; ideally expose a `page` input.

integrations/zapier-mem0/creates/delete_memory.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const perform = async (z, bundle) => {
+	// Trailing slash required (Django APPEND_SLASH).
+	const response = await z.request({
+		url: `/v1/memories/${bundle.inputData.memory_id}/`,
```
kartik-mem0: **[Minor] `memory_id` is interpolated into the URL path without `encodeURIComponent`.** Low risk since Mem0 IDs are UUIDs, but a stray `/` or special char in a pasted ID would mistarget the request rather than fail cleanly. Cheap to harden: `encodeURIComponent(bundle.inputData.memory_id)`.

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
```diff
@@ -0,0 +1,439 @@
...
+						type: 'json',
+						default: '',
+					},
+					{
+						displayName: 'Infer',
+						name: 'infer',
+						type: 'boolean',
+						default: true,
+						description:
+							'Whether to run LLM extraction (async). Turn off to store messages verbatim (synchronous).',
```
kartik-mem0: **[Minor] This says `infer=false` is “synchronous,” but the public API docs only document an async response for all add calls.**

`docs/api-reference/memory/add-memories.mdx` shows every `POST /v3/memories/add/` returning `{status:"PENDING", event_id}` regardless of `infer`. The branching in `execute()` is safely defensive either way (it unwraps `results` if the response is terminal, else polls), so there’s no functional bug — but if you confirmed live that `infer=false` returns synchronously, the docs disagree and it’s worth a note; if not, this description (and the comments at `:337`/`:347`) mislead. `infer` is a server-side extract-vs-verbatim switch; it doesn’t control whether this node waits — `waitForCompletion` does.

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
```diff
@@ -0,0 +1,439 @@
...
+							itemIndex: i,
+						});
+					}
+					const addFields = this.getNodeParameter('addFields', i, {}) as IDataObject;
+					const body: IDataObject = {
+						messages: messagesUi.map((m) => ({ role: m.role, content: m.content })),
+						infer: addFields.infer !== undefined ? addFields.infer : true,
+					};
+					const userId = this.getNodeParameter('userId', i, '') as string;
+					if (userId) body.user_id = userId;
```
kartik-mem0, line 323: **[Minor] Add has no client-side entity-id check.** The API requires at least one of `user_id`/`agent_id`/`run_id`/`app_id` (add-memories.mdx), but here `userId` isn’t `required` (unlike Search / Get Many) and nothing verifies an entity id made it into `body` before the POST — so a messages-only Add hits a raw 4xx instead of a clear node error. Mirror the pre-flight guard you already use for Update (`:388-392`) and throw a `NodeOperationError`.

Separately, `app_id` isn’t exposed anywhere — consider adding it to Additional Fields (and as a Search / Get Many filter).

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
```diff
@@ -0,0 +1,439 @@
...
+): Promise<IDataObject> {
+	for (let attempt = 0; attempt < MAX_POLL_ATTEMPTS; attempt++) {
+		const event = await request('GET', `/v1/event/${eventId}/`);
+		const status = event.status as string;
+		if (status === 'SUCCEEDED') {
+			return event;
+		}
+		if (status === 'FAILED') {
+			const reason = (event.error as string) || (event.message as string) || 'unknown error';
+			throw new NodeOperationError(ctx.getNode(), `Mem0 memory event ${eventId} failed: ${reason}`);
```
kartik-mem0: **[Minor] `pollEvent`’s throws drop `itemIndex`.** Every other `NodeOperationError` in this file passes `{itemIndex: i}` so n8n can highlight the offending input item; the two throws here (`:431`, `:435`) don’t, and `i` isn’t even passed into `pollEvent`. On a multi-item batch, a poll timeout/failure can’t be traced to its source item. Add an `itemIndex` param and include it in both throws.

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
```diff
@@ -0,0 +1,439 @@
...
+async function pollEvent(
+	request: (m: IHttpRequestMethods, u: string) => Promise<IDataObject>,
+	eventId: string,
+	ctx: IExecuteFunctions,
+): Promise<IDataObject> {
+	for (let attempt = 0; attempt < MAX_POLL_ATTEMPTS; attempt++) {
+		const event = await request('GET', `/v1/event/${eventId}/`);
+		const status = event.status as string;
+		if (status === 'SUCCEEDED') {
+			return event;
```
kartik-mem0: **[Minor / consistency] The default Add path returns the raw event envelope, while Search / Get Many return unwrapped arrays.** `pollEvent` returns the full `{id, status, results, …}` object here, but `search` (`:361`), `getAll` (`:369`), and the Add non-wait fallback (`:348`) all hand back a clean array of memory objects. Since `waitForCompletion` defaults on, the most common Add usage yields a differently-shaped output than every other read op. Consider `return Array.isArray(event.results) ? event.results : event;` to match (safe fallback to the envelope).

integrations/zapier-mem0/creates/add_memory.js
Himanshu-Sangshetti: Cut the poll budget to 12 x 1.5s (~18s)

.github/workflows/zapier-mem0-checks.yml
Himanshu-Sangshetti: Added test/unit.test.js with offline tests

integrations/zapier-mem0/searches/get_memories.js
Himanshu-Sangshetti: Exposed a Page input (1-based) instead of hardcoding page 1

integrations/zapier-mem0/creates/delete_memory.js
Himanshu-Sangshetti: Done, memory_id is now wrapped in encodeURIComponent

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
Himanshu-Sangshetti: Reworded the description to say it controls extraction only and that Wait for Completion
  is what decides whether the node waits

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
Himanshu-Sangshetti: Added a pre-flight guard

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
Himanshu-Sangshetti: Added an itemIndex param to pollEvent

integrations/n8n-nodes-mem0/nodes/Mem0/Mem0.node.ts
Himanshu-Sangshetti: pollEvent now returns event.results when it's an array and falls back to the envelope otherwise, so the wait-path output matches search and get many

Himanshu-Sangshetti:
Splitting this into two focused PRs per the review, one integration each so they're easier to evaluate:

- n8n community node: #6517
- Zapier app: #6518

Both carry all the fixes from this PR's review plus the follow-ups from our discussion:

- All 8 inline comments addressed (timeout/default-wait, functional test coverage, page input, id encoding, infer wording, entity-id guard + app_id, itemIndex on poll errors, output-shape unwrap).
- Custom extraction controls exposed on Add: `custom_instructions`, `custom_categories`, `includes`, `excludes`.
- Anonymous opt-out telemetry (`MEM0_TELEMETRY=false`).
- Both integrations now in TypeScript.
- Docs pages + docs.json/llms.txt, ci-gate wiring, AGENTS.md rows.
- Live E2E verified against api.mem0.ai for both.

Closing this in favour of the two above.

PR #6060, opened 2026-07-02
fix(vector-stores/redis): return updated_at from search() and keyword_search()

### LINKED ISSUE #6059: bug(redis): search() and keyword_search() omit updated_at from return_fields, making the updated_at payload guards dead code
## Component

### Summary

### Steps to Reproduce

```python

db.index = MagicMock()

def only_requested(query):

```

### Expected Behavior

### Actual Behavior

### Environment

---

Krishnachaitanyakc:
## Linked Issue

Closes #6059

## Description

`search()` and `keyword_search()` on the Redis vector store omitted `updated_at` from their `return_fields`, so the existing `if "updated_at" in result` payload guards were dead code and neither read path surfaced `updated_at` — even though `update()` persists it, it is a declared numeric index field in `DEFAULT_FIELDS`, and both `get()` and `list()` already return it. This restores metadata round-trip parity across the Redis vector store's read paths.

## Type of Change

## Breaking Changes

N/A — additive only. Payloads gain `updated_at` for records that have it (matching `get()`/`list()`); records without it are unchanged.

## Test Coverage

Added `test_search_requests_updated_at_in_return_fields` and `test_keyword_search_requests_updated_at_in_return_fields` — both fail before the change and pass after.

Real behavior proof:
- Unit: before the fix both new tests FAIL (`AssertionError: 'updated_at' not in [...]`); after the fix `pytest tests/vector_stores/test_redis.py` → 13 passed.
- Live end-to-end against `redis/redis-stack` (RediSearch), real `insert -> update -> get -> search` round-trip:
  - On `main`: `get()` returns `updated_at` (`2023-11-14T22:14:10+00:00`), `search()` returns `None` (bug reproduced).
  - On this branch: both `get()` and `search()` return `updated_at` (fixed).
- `make lint` (ruff check) clean; isort clean.

## Checklist

- [x] I have updated documentation if needed (N/A — internal read-path parity)

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -155,7 +155,17 @@ class RedisDB(VectorStoreBase):
         v = VectorQuery(
             vector=np.array(vectors, dtype=np.float32).tobytes(),
             vector_field_name="embedding",
-            return_fields=["memory_id", "hash", "agent_id", "run_id", "user_id", "memory", "metadata", "created_at"],
+            return_fields=[
+                "memory_id",
+                "hash",
+                "agent_id",
+                "run_id",
+                "user_id",
+                "memory",
+                "metadata",
+                "created_at",
+                "updated_at",
+            ],
             filter_expression=filter,
             num_results=top_k,
         )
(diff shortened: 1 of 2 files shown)


Krishnachaitanyakc:
@kartik-mem0 @whysosaket can you please review this PR?

PR #6067, opened 2026-07-03
fix(llms): rewrite assistant as whole word for Azure

### LINKED ISSUE #6036: bug(llms): Azure OpenAI integration corrupts words containing 'assistant' #6036

sjsjsjjs534:
## Summary
- rewrite only standalone assistant tokens in Azure OpenAI prompts
- preserve words that contain assistant as a substring, such as assistantship
- add regression coverage for both Azure OpenAI LLM implementations

Fixes #6036

## Tests
- UV_CACHE_DIR=/private/tmp/uv-cache-mem0 uv run --python 3.12 --extra test --extra vector-stores pytest tests/llms/test_azure_openai.py tests/llms/test_azure_openai_structured.py -q
- UV_CACHE_DIR=/private/tmp/uv-cache-mem0 uv run --python 3.12 --extra test --with ruff ruff check mem0/llms/azure_openai.py mem0/llms/azure_openai_structured.py tests/llms/test_azure_openai.py tests/llms/test_azure_openai_structured.py
- UV_CACHE_DIR=/private/tmp/uv-cache-mem0 uv run --python 3.12 --extra test --with ruff ruff format --check mem0/llms/azure_openai.py mem0/llms/azure_openai_structured.py tests/llms/test_azure_openai.py tests/llms/test_azure_openai_structured.py

### DIFF
--- a/mem0/llms/azure_openai.py
+++ b/mem0/llms/azure_openai.py
@@ -1,6 +1,7 @@
 import copy
 import json
 import os
+import re
 from typing import Dict, List, Optional, Union
 
 from azure.identity import DefaultAzureCredential, get_bearer_token_provider
@@ -32,9 +33,9 @@ class AzureOpenAILLM(LLMBase):
                 top_k=config.top_k,
                 enable_vision=config.enable_vision,
                 vision_details=config.vision_details,
-                reasoning_effort=getattr(config, 'reasoning_effort', None),
+                reasoning_effort=getattr(config, "reasoning_effort", None),
                 http_client_proxies=config.http_client_proxies,
-                is_reasoning_model=getattr(config, 'is_reasoning_model', None),
+                is_reasoning_model=getattr(config, "is_reasoning_model", None),
             )
 
         super().__init__(config)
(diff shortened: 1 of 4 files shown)


kartik-mem0:
Closing this in favour of #6042, which fixes the same bug and covers more of it.

Three PRs arrived for #6036 within about a day of each other. Your regex is identical to the one in #6042, so on the core fix there is genuinely nothing to choose between them. Two things decided it: #6042 carries 2 tests per file, 5 in total, including a compound-identifier case that is absent here, against 1 per file; and this diff bundles unrelated quote-style and `params.update()` reformatting in alongside the fix, which makes the actual change harder to isolate when reading it back later.

I have reviewed #6042 in full and it is the one going forward.

The fix itself was right, so the only thing that would have made this version more competitive is keeping the reformatting out of the fix diff. Worth doing next time, and please do send more.

PR #6068, opened 2026-07-03
fix(ts): use match-all Redis query for empty filters

### LINKED ISSUE #6013: bug(ts-oss/redis): empty/all-null filters build an invalid query ("" instead of "*")

sjsjsjjs534:
## Summary
- build Redis filter expressions through a shared helper
- fall back to * when filters are empty or all values are null/undefined
- add Redis unit coverage for search and list empty-filter queries

Fixes #6013

## Tests
- pnpm jest src/oss/tests/redis.unit.test.ts --runInBand
- pnpm exec prettier --check src/oss/src/vector_stores/redis.ts src/oss/tests/redis.unit.test.ts
- pnpm run build

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/redis.ts
+++ b/mem0-ts/src/oss/src/vector_stores/redis.ts
@@ -133,6 +133,17 @@ function toSnakeCase(obj: Record<string, any>): Record<string, any> {
   );
 }
 
+function buildFilterExpression(filters?: SearchFilters): string {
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #6076, opened 2026-07-03
fix: persist and return updated config in OpenMemory PUT /api/v1/config

### LINKED ISSUE #6074: [BUG] OpenMemory: PUT /api/v1/config/ always returns 500 - handler never saves or returns the config
### Description

The `PUT /api/v1/config/` endpoint in the OpenMemory API is broken: every request fails with an HTTP 500 (`ResponseValidationError`).

In `openmemory/api/app/routers/config.py`, the `update_configuration` handler builds `updated_config` from the request body but then falls off the end of the function:

- `save_config_to_db()` is never called — nothing is persisted
- `reset_memory_client()` is never called — even if it saved, the running memory client would keep the old config
- there is no `return` statement — the handler returns `None`, which fails validation against `response_model=ConfigSchema` and produces a 500

The `PATCH /api/v1/config/` handler directly below it shows the intended pattern (save → reset → return).

### Steps to reproduce

1. `cd openmemory && docker compose up`
2. `curl -X PUT http://localhost:8765/api/v1/config/ -H "Content-Type: application/json" -d '{"mem0": {"llm": {"provider": "openai", "config": {"model": "gpt-4o-mini", "temperature": 0.1, "max_tokens": 2000}}}}'`
3. Response is HTTP 500 with `fastapi.exceptions.ResponseValidationError` in the API logs; a subsequent `GET /api/v1/config/` shows the config was not updated.

### Expected behavior

The endpoint saves the updated configuration, resets the memory client so the new config takes effect, and returns the updated configuration (HTTP 200).

### Additional context

Related smaller defect in the same handler: `config.mem0.dict(...)` is called unconditionally even though `mem0` is `Optional` in `ConfigSchema`, so a PUT that only sets `openmemory` would raise `AttributeError` once the missing-return bug is fixed.

Found while deploying OpenMemory (docker compose) — PR with the fix and a regression test incoming.

csmdelinea:
## Linked Issue

Closes #6074

## Description

`PUT /api/v1/config/` in the OpenMemory API currently 500s on every request. The `update_configuration` handler builds `updated_config` but never calls `save_config_to_db()`, never calls `reset_memory_client()`, and has no `return` — so nothing is persisted and the handler returns `None`, which fails validation against `response_model=ConfigSchema` (`ResponseValidationError` → HTTP 500).

This PR completes the handler following the exact pattern already used by the `PATCH /api/v1/config/` handler directly below it:

- save the merged configuration with `save_config_to_db()`
- call `reset_memory_client()` so the running memory client picks up the new configuration
- return the updated configuration

It also fixes an adjacent defect in the same handler: `config.mem0.dict(...)` was called unconditionally even though `mem0` is `Optional` in `ConfigSchema`, so a PUT containing only `openmemory` settings would have raised `AttributeError`. The `mem0` section is now only replaced when provided, mirroring the existing `openmemory` guard.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `openmemory/api/tests/test_config_router.py` with regression tests that fail without this fix:

- PUT returns the updated configuration (previously: 500)
- PUT persists the configuration to the database (previously: silently dropped)
- PUT resets the memory client (previously: never called)
- PUT with only `openmemory` settings works and preserves the existing `mem0` section (previously: `AttributeError`)

Manually verified against a docker-compose OpenMemory deployment: with this fix applied, `PUT /api/v1/config/` returns 200 with the updated config and the change persists.

Note: `tests/test_mcp_server.py::TestRouteRegistration` (3 tests) fails identically on an unmodified `main` checkout in my environment — pre-existing and unrelated to this change. All other tests, including the new ones, pass.

## Checklist

### DIFF
--- a/openmemory/api/app/routers/config.py
+++ b/openmemory/api/app/routers/config.py
@@ -151,10 +151,15 @@ async def update_configuration(config: ConfigSchema, db: Session = Depends(get_d
         if "openmemory" not in updated_config:
             updated_config["openmemory"] = {}
         updated_config["openmemory"].update(config.openmemory.dict(exclude_none=True))
-    
-    # Update mem0 settings
-    updated_config["mem0"] = config.mem0.dict(exclude_none=True)
-    
+
+    # Update mem0 settings if provided
+    if config.mem0 is not None:
+        updated_config["mem0"] = config.mem0.dict(exclude_none=True)
+
+    save_config_to_db(db, updated_config)
+    reset_memory_client()
+    return updated_config
+
 
 @router.patch("/", response_model=ConfigSchema)
 async def patch_configuration(config_update: ConfigSchema, db: Session = Depends(get_db)):
--- a/openmemory/api/tests/test_config_router.py
+++ b/openmemory/api/tests/test_config_router.py
@@ -1,0 +1,129 @@
+"""Tests for the configuration API router.
+
+Covers the PUT /api/v1/config/ handler, which previously built the updated
+configuration but never persisted it, never reset the memory client, and
+returned None (causing a ResponseValidationError / HTTP 500 on every call).
+"""
+
+import os
+
+# Set dummy keys before any imports that trigger client initialization
+os.environ.setdefault("OPENAI_API_KEY", "test-key")
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.pool import StaticPool
+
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for the contribution, and sorry for the outcome here.

The OpenMemory project has been deprecated and is no longer maintained, so we can't merge changes into it. Your fix addressed a real bug — closing is about the deprecation, not the merits of the change. We should have communicated that earlier and saved you the effort; that's on us.

The core SDK and the self-hosted `server/` are actively maintained, and contributions there are very welcome.

PR #6090, opened 2026-07-04
fix(llms): classify gpt-5 dash family (gpt-5-mini) as reasoning model

### LINKED ISSUE #6085: OSS default config silently breaks add(): model=None resolves to gpt-5-mini, which rejects the default temperature=0.1
### Environment

- Python: 3.12.6

### Description

```
```

### Steps to reproduce

```python
import os

from mem0 import Memory
```

### Root cause

```python
```

Bartok9:
## Summary

- Treat the `gpt-5-mini` / `gpt-5-nano` dash family as reasoning models so temperature is stripped.
- Fixes the default OSS config silently storing nothing on every `add()`.

## Motivation

Closes #6085.

The default OSS config leaves `model=None`, which resolves to `gpt-5-mini` inside `mem0/llms/openai.py`. `gpt-5-mini` is a reasoning model that rejects any non-default `temperature`, but `_is_reasoning_model` only matched the bare `"gpt-5"` name and the `o1-`/`o3-` prefixes — **not** the `gpt-5-mini`/`gpt-5-nano` dash family. So the default `temperature=0.1` was still sent, the OpenAI API returned:

```
400 - Unsupported value: 'temperature' does not support 0.1 with this model. Only the default (1) value is supported.
```

The error was caught and logged as `LLM extraction failed`, and `add()` returned without storing anything — a silent failure that looks like a retrieval-quality problem. A brand-new user following the quickstart gets empty search results with no visible error.

## Fix

Add a `"gpt-5-"` (dash) prefix check in `_is_reasoning_model`. This classifies `gpt-5-mini`, `gpt-5-nano`, and their dated variants (e.g. `gpt-5-mini-2025-08-07`) as reasoning models, so `temperature` is dropped.

Crucially, it uses an explicit `gpt-5-` **dash** prefix rather than a bare `gpt-5` substring, so it does **not** match the newer dot-versioned `gpt-5.4-mini`, which supports temperature (behavior preserved from #4738).

## Verification

- `python -m pytest tests/llms/test_openai.py -k "reasoning or gpt5 or default_gpt5"` — **10 passed**
  - Extended `test_is_reasoning_model_classification` with the dash family (True) and dot-versioned models (still False).
  - Added `test_default_gpt5_mini_omits_temperature` proving the all-defaults config resolves to `gpt-5-mini` and no longer sends `temperature`.

## Scope note

3 pre-existing failures in `test_openai.py` (`test_openai_llm_base_url`, `test_store_sent_when_explicitly_*`) reproduce on clean `main` without this change and are unrelated (installed `openai` SDK `store`-param behavior). They are intentionally not touched here.

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -77,6 +77,18 @@ class LLMBase(ABC):
         if any(base_model.startswith(prefix) for prefix in ["o1-", "o1.", "o3-", "o3."]):
             return True
 
+        # Match the original GPT-5 dash family (gpt-5-mini, gpt-5-nano and their
+        # dated variants like gpt-5-mini-2025-08-07). These are reasoning models
+        # that reject a non-default temperature. This must NOT match the newer
+        # dot-versioned models (gpt-5.4-mini) which support temperature, hence
+        # the explicit "gpt-5-" (dash) prefix rather than a bare "gpt-5" substring.
+        #
+        # Exclude the "gpt-5-chat" family (gpt-5-chat, gpt-5-chat-latest): these
+        # are the non-reasoning chat variants that DO accept a custom temperature,
+        # so they must stay on the temperature path.
+        if base_model.startswith("gpt-5-") and not base_model.startswith("gpt-5-chat"):
+            return True
+
         return False
 
     def _uses_max_completion_tokens(self, model: str) -> bool:
--- a/tests/llms/test_openai.py
+++ b/tests/llms/test_openai.py
@@ -323,11 +323,48 @@ def test_is_reasoning_model_classification(mock_openai_client):
     assert llm._is_reasoning_model("o1-2024-12-17") is True
     assert llm._is_reasoning_model("openai/o3-mini") is True
 
+    # GPT-5 dash family — reasoning models that reject non-default temperature
+    # (regression for https://github.com/mem0ai/mem0/issues/6085)
+    assert llm._is_reasoning_model("gpt-5-mini") is True
+    assert llm._is_reasoning_model("gpt-5-nano") is True
+    assert llm._is_reasoning_model("gpt-5-mini-2025-08-07") is True
+    assert llm._is_reasoning_model("openai/gpt-5-mini") is True
+
     # Non-reasoning models — should return False
     assert llm._is_reasoning_model("gpt-5.4-mini") is False
     assert llm._is_reasoning_model("gpt-5.4") is False
+    # gpt-5-chat family is the non-reasoning chat variant and DOES accept a
+    # custom temperature, so the gpt-5- dash rule must not strip it.
+    assert llm._is_reasoning_model("gpt-5-chat") is False
+    assert llm._is_reasoning_model("gpt-5-chat-latest") is False
+    assert llm._is_reasoning_model("openai/gpt-5-chat") is False
     assert llm._is_reasoning_model("gpt-4.1") is False
     assert llm._is_reasoning_model("gpt-4.1-nano-2025-04-14") is False
+
+
+def test_default_gpt5_mini_omits_temperature(mock_openai_client):
+    """Default OSS config must not send temperature for the resolved gpt-5-mini.
+
+    Regression test for https://github.com/mem0ai/mem0/issues/6085: with all
+    defaults, ``model`` resolves to ``gpt-5-mini`` (a reasoning model that
+    rejects any non-default temperature). Previously the default temperature
+    0.1 was still sent, so every add() failed LLM extraction and silently
+    stored nothing. The dash family must now be classified as reasoning so
+    temperature is stripped.
+    """
+    config = OpenAIConfig(temperature=0.1)  # model=None -> resolves to gpt-5-mini
+    llm = OpenAILLM(config)
+    assert llm.config.model == "gpt-5-mini"
+
+    messages = [{"role": "user", "content": "Hello"}]
+    mock_response = Mock()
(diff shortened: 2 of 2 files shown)


yashs33244 (commented):
Nice catch on the dash-vs-dot distinction - `gpt-5-` correctly leaves `gpt-5.4-mini` on the temperature path, and the default-config test pins the real #6085 failure (model resolves to `gpt-5-mini`, temperature was still being sent) rather than just the classifier.

One edge case worth a look: `startswith("gpt-5-")` also matches the `gpt-5-chat` family (`gpt-5-chat`, `gpt-5-chat-latest`), which is the non-reasoning chat variant and *does* accept a custom temperature. With this change it gets classified as reasoning and has temperature stripped, so a `gpt-5-chat` user would silently lose temperature control. If that family is in scope, excluding it keeps the intent tight:

```python
if base_model.startswith("gpt-5-") and not base_model.startswith("gpt-5-chat"):
    return True
```

Not a blocker for the reported bug - just flagging it since it's the same dash-prefix trap the dot-version comment is already guarding against.

Bartok9:
Good catch — you're right that `gpt-5-chat`/`gpt-5-chat-latest` are the non-reasoning chat variants and should keep temperature control. Applied your suggested guard:

```python
if base_model.startswith("gpt-5-") and not base_model.startswith("gpt-5-chat"):
    return True
```

Also extended `test_is_reasoning_model_classification` to pin the chat family as non-reasoning (`gpt-5-chat`, `gpt-5-chat-latest`, `openai/gpt-5-chat` all → False). Full suite (`-k "reasoning or gpt5 or default_gpt5"`) still 10 passed. Pushed as an amend to the same commit.

yashs33244:
Confirmed - the guard and the three `gpt-5-chat` / `gpt-5-chat-latest` / `openai/gpt-5-chat` assertions look exactly right, and keeping the dash-family cases as `True` pins the intent both ways. Nice and tight.

kartik-mem0:
## Needs Human Review

**Why:** This is a solid, merge-quality fix, but it competes with **#6242** — which modifies the same `_is_reasoning_model` function with a different mechanism *and* also carries an independent litellm fix. Picking between the two (plus the test-only #6197) is a maintainer sequencing/design call, not a single-PR merge decision.

### Pay attention to
- `mem0/llms/base.py` — `_is_reasoning_model`. This PR adds a **prefix** branch (`startswith("gpt-5-")`, excluding `gpt-5-chat`). **#6242** instead adds `gpt-5-mini`/`gpt-5-nano` to an **exact-match set** on the same function. Both are open; they will conflict, and their coverage differs.

### The cluster (diffs read, not titles)
- **This PR (#6090)** — prefix approach. **More robust:** also classifies dated variants like `gpt-5-mini-2025-08-07` and `openai/`-prefixed names; excludes `gpt-5-chat*` and dotted `gpt-5.4-mini` (preserves #4738).
- **#6242** — exact-set approach, which **misses dated variants** — *plus* a separate litellm-routing fix (`litellm.py`, `test_litellm.py`) that this PR does not touch.
- **#6197** — test-only (`tests/test_issue_6085.py`, no fix); its assertion would fail on `main`. Superseded by this PR.

### Recommendation
Take **#6090's** classification (the more robust one); **close #6197** (test-only, superseded); ask **#6242** to rebase and keep only its litellm-routing portion on top of #6090. That lands both real fixes without leaving two parallel classification mechanisms in `_is_reasoning_model`.

### Findings
Nothing blocking in #6090 itself — the code is correct and well-tested (below). The needs-human is purely the cross-PR reconciliation.

### Evidence
- **CI:** Python SDK build 3.10 / 3.11 / 3.12 pass, CI Gate + changelog + check_changes pass (Vercel preview fail ignored — not merge-blocking).
- **Targeted tests (worktree, `hatch -e dev_py_3_12 run pytest`):** `test_default_gpt5_mini_omits_temperature` + `test_is_reasoning_model_classification` → 2 passed.
- **Red-green proof:** drop the added `gpt-5-` branch → 2 failed, reproducing the actual #6085 symptom (`{'temperature': 0.1, 'top_p': 0.1, ...}` *is* sent to gpt-5-mini) and `_is_reasoning_model('gpt-5-mini')` → `False`; restore → 2 passed. ✓
- **Linked issue:** #6085 — addresses the reported default-config silent `add()` failure.
- **Cross-package:** mem0-ts OSS has no reasoning-model classifier, so there is no parity obligation.
- **Could NOT verify:** live OpenAI 400 behavior (mocked); the classification and kwargs assertions cover the mechanism end-to-end.

### Assessment
Ready to merge: **No — maintainer judgment required** (choose the classification approach across #6090 / #6242, and close #6197). Est. human time: ~8 min

Bartok9:
Thanks @kartik-mem0 — agree with the cluster read.

Happy for maintainers to take **#6090's prefix classification** as the canonical `_is_reasoning_model` path (dated + `openai/`-prefixed variants; exclude `gpt-5-chat*`; leave dotted `gpt-5.4-*` alone per #4738). From our side:

- **Stay green on #6090** and ready to squash-merge whenever you queue it.
- **#6197** (test-only) we treat as superseded by this PR — no objection to closing it with credit to the author if you prefer.
- **#6242** — agreed they should keep the independent litellm-routing half and drop the exact-set classifier once #6090 lands (or rebase onto this tip). We will not open a competing third take on the same function.

No other code changes needed from us on #6090; sequencing call is yours. Thanks for the careful cross-PR write-up.

Agent-Owner: sera

PR #6099, opened 2026-07-05
fix(openmemory): list_memories passes user_id via filters to get_all()

### LINKED ISSUE #6082: OpenMemory MCP list_memories broken: get_all(user_id=...) rejected by current mem0 get_all() API
## Summary

The OpenMemory MCP server's `list_memories` tool is broken: `openmemory/api/app/mcp_server.py` calls `memory_client.get_all(user_id=uid)`, but the current mem0 `Memory.get_all()` API rejects top-level entity parameters.

## Error

Calling the `list_memories` MCP tool returns:

```
Error getting memories: Top-level entity parameters frozenset({'user_id'}) are not supported in get_all(). Use filters={'user_id': '...'} instead.
```

## Location

https://github.com/mem0ai/mem0/blob/main/openmemory/api/app/mcp_server.py#L248

```python
memories = memory_client.get_all(user_id=uid)
```

## Fix

```python
memories = memory_client.get_all(filters={"user_id": uid})
```

Verified working against a v2.0.7 deployment — after this one-line change, `list_memories` returns results correctly. The response shape (`dict` with `results` key) is already handled by the surrounding code.

## Environment

- mem0 v2.0.7 (via the JSONbored/mem0-aio v2.0.7-aio.1 Docker image, which vendors this repo's `openmemory/` directory)
- Vector store: Qdrant, LLM: Anthropic, Embedder: OpenAI
- `add_memories` and `search_memory` work fine; only `list_memories` hits the stale `get_all` call

## Notes

This looks like API drift between `openmemory/` and the mem0 library itself — the error message's suggested call shape is exactly what the code needs. Downstream, this was previously reported (and closed) as JSONbored/mem0-aio#29, but since that project syncs `openmemory/` from this repo, the bug returns with every sync until fixed here.

Bartok9:
## Summary

- `list_memories` MCP tool passes `user_id` through `filters=` instead of as a top-level `get_all()` kwarg.
- Restores `list_memories` for OpenMemory MCP deployments (it previously returned no results / an error).

## Motivation

Closes #6082.

The current mem0 `Memory.get_all()` API rejects top-level entity parameters and requires `filters={"user_id": ...}`. `openmemory/api/app/mcp_server.py` still called `get_all(user_id=uid)`, so against a live deployment `list_memories` raised:

```
Top-level entity parameters frozenset({'user_id'}) are not supported in get_all(). Use filters={'user_id': '...'} instead.
```

and returned no memories. This is API drift between `openmemory/` and the mem0 library. `add_memories` / `search_memory` were unaffected.

## Fix

```python
-            memories = memory_client.get_all(user_id=uid)
+            memories = memory_client.get_all(filters={"user_id": uid})
```

The surrounding code already handles the `dict`-with-`results` response shape.

## Verification

- `pytest openmemory/api/tests/test_mcp_server.py::TestListMemoriesGetAllCallShape` — passes.
- Added a regression test that drives `list_memories` through a `get_all` stub mirroring the real API contract (rejects a top-level `user_id`). It **fails on the old call** (`ValueError`) and **passes on the fix**, and asserts the memory is returned rather than silently dropped.
- Did NOT change the response-handling / permission-filtering logic.

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -245,7 +245,7 @@ async def list_memories() -> str:
             user, app = get_user_and_app(db, user_id=uid, app_id=client_name)
 
             # Get all memories
-            memories = memory_client.get_all(user_id=uid)
+            memories = memory_client.get_all(filters={"user_id": uid})
             filtered_memories = []
 
             # Filter memories based on permissions
--- a/openmemory/api/tests/test_mcp_server.py
+++ b/openmemory/api/tests/test_mcp_server.py
@@ -6,6 +6,7 @@ tools/call — as well as error handling and context-variable isolation.
 """
 
 import os
+import uuid
 
 # Set dummy keys before any imports that trigger client initialization
 os.environ.setdefault("OPENAI_API_KEY", "test-key")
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for the contribution, and sorry for the outcome here.

The OpenMemory project has been deprecated and is no longer maintained, so we can't merge changes into it. Your fix addressed a real bug — closing is about the deprecation, not the merits of the change. We should have communicated that earlier and saved you the effort; that's on us.

The core SDK and the self-hosted `server/` are actively maintained, and contributions there are very welcome.

Bartok9:
No worries at all — thanks for the kind and clear explanation, and for confirming the bug was real. Totally understand on the OpenMemory deprecation. I'll steer future contributions toward the core SDK and the self-hosted `server/`. Appreciate the maintainer time either way! 🙏

PR #6102, opened 2026-07-05
fix(ts-sdk): re-raise LLM extraction transport failures instead of returning []

### LINKED ISSUE #6101: TypeScript SDK: LLM extraction transport failures silently swallowed (index.ts:855-858)
### Component

Core / Python SDK

### Summary

Follow-up to #5903 (fixed on the Python side by #5878). The TS OSS SDK has the identical bug: `Memory.add()`'s LLM extraction step catches *all* errors from `generateResponse` , including provider/transport failures (rate limits, timeouts, connection errors) , and swallows them by returning `[]`. This makes "the LLM is unreachable" indistinguishable from "the LLM ran and extracted nothing," and callers have no way to detect or retry the former.

Location: `mem0-ts/src/oss/src/memory/index.ts:846-858`

```ts
try {
  response = (await this.llm.generateResponse(
    [
      { role: "system", content: systemPrompt },
      { role: "user", content: userPrompt },
    ],
    { type: "json_object" },
  )) as string;
} catch (e) {
  console.error("LLM extraction failed:", e);
  return [];
}
```

### Steps to Reproduce

```ts

const memory = new Memory(/* config */);

// Simulate a transport failure from the LLM provider
jest.spyOn(memory["llm"], "generateResponse").mockRejectedValue(
  new Error("429 rate limit"),
);

const result = await memory.add([{ role: "user", content: "hello" }], {
  userId: "u1",
});
// result === [] — indistinguishable from "no facts found"
```

### Expected Behavior

A transport/provider failure during LLM extraction should propagate to the caller (as a typed error) so calling code can implement retry or fallback logic. A genuinely empty/invalid LLM response (the *parse* failure a few lines below this catch block) should still return `[]`, since that case is a real "no facts extracted" outcome, not a transport error.

### Actual Behavior

Both cases are collapsed into the same `return []`, silently dropping the failure with only a `console.error` log line.

### Proposed Fix

Mirror the Python fix (#5878): re-throw the transport error wrapped in a typed error instead of returning `[]`. The TS OSS SDK already has a precedent for this exact pattern, PR #5546 introduced `EmbeddingError` in this same file for an analogous "don't silently drop, throw a typed error" fix. Following that shape:

- Add an `LLMError` class (or similar) near the top of `index.ts`, next to   `EmbeddingError`.
- Replace the `catch (e) { console.error(...); return []; }` block at  `index.ts:855-858` with `throw new LLMError(...)`.
- Leave the downstream JSON-parse failure path untouched (it should keep returning `[]`).
- Add a regression test mirroring the Python one added in #5878 (`test_llm_extraction_exception_is_reraised`).

### Environment

- mem0 version: main branch (post-#5878)
- Package: `mem0-ts` (OSS SDK, `mem0-ts/src/oss/`)
- Node version: N/A — code-level bug, reproducible in any environment

ShauryaaSharma:
## Linked Issue

Closes #6101

## Description

`Memory.add()`'s LLM extraction step in the TypeScript OSS SDK caught *all* errors from `generateResponse` , including provider/transport failures (rate limits, timeouts, connection errors)  and swallowed them
by returning `[]`. This made "the LLM is unreachable" indistinguishable from "the LLM ran and extracted nothing," and callers had no way to detect or retry the former.

This mirrors the Python SDK bug already fixed by #5878 (tracked in #5903, which explicitly flagged this TS-side follow-up).

**Before** (`mem0-ts/src/oss/src/memory/index.ts:855-858`):

```ts
} catch (e) {
  console.error("LLM extraction failed:", e);
  return [];
}
```

**After:**

```ts
} catch (e) {
  console.error("LLM extraction failed:", e);
  throw new LLMError(`LLM extraction failed: ${e}`, { cause: e });
}
```

Adds a typed `LLMError` class (alongside the existing error-handling patterns in this file) so callers can distinguish a transport failure from a genuinely empty/invalid LLM response. The downstream JSON-parse failure path is untouched and still returns `[]`, since that case really is "no facts extracted," not a transport error.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `mem0-ts/src/oss/tests/memory.llm-error.test.ts` with two regression tests mirroring the Python regression test added in #5878 (`test_llm_extraction_exception_is_reraised`):

- `add()` rejects with `LLMError` instead of resolving with an empty result
- the thrown `LLMError` preserves the original error as `.cause`

```
PASS src/oss/tests/memory.llm-error.test.ts
  Memory - LLM extraction transport failures
    ✓ add() rejects with LLMError instead of returning an empty result
    ✓ thrown LLMError preserves the original error as its cause
```

Also ran the full related suite with zero regressions (`memory.add`, `memory.validation`, `memory.crud`, `memory.init`,
`memory.entity-boost` — 80/80 passing), the full OSS suite (580/623 passing, the 43 failures are pre-existing Windows `EPERM` temp-dir issues in unrelated sqlite/vector-store tests, verified identical on unmodified `main`), `tsc --noEmit` (no new errors), and a `tsup` build (confirmed `LLMError` is correctly exported from the public `dist/oss/index.d.ts` surface).

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -74,6 +74,25 @@ import {
 import { getDefaultVectorStoreDbPath } from "../utils/sqlite";
 import { getOrCreateMem0UserId } from "../../../client/config";
 
+/**
+ * Raised when the LLM extraction call itself fails (provider/transport
+ * errors: rate limits, timeouts, connection errors), as opposed to the LLM
+ * responding successfully with an empty or unparseable result. Callers can
+ * catch this to distinguish "LLM unavailable" from "no facts extracted" and
+ * implement retry/fallback logic. Mirrors the Python SDK `LLMError` (see
+ * #5903 / #5878).
+ */
+export class LLMError extends Error {
+  readonly cause?: unknown;
+
+  constructor(message: string, options: { cause?: unknown } = {}) {
+    super(message);
+    this.name = "LLMError";
+    this.cause = options.cause;
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
 // Entity params that must be passed via filters - check both snake_case and camelCase
 const ENTITY_PARAMS = [
   "user_id",
@@ -854,7 +873,7 @@ export class Memory {
       )) as string;
     } catch (e) {
       console.error("LLM extraction failed:", e);
-      return [];
+      throw new LLMError(`LLM extraction failed: ${e}`, { cause: e });
     }
 
     // Parse response
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -74,14 +74,6 @@ import {
 import { getDefaultVectorStoreDbPath } from "../utils/sqlite";
 import { getOrCreateMem0UserId } from "../../../client/config";
 
-/**
- * Raised when the LLM extraction call itself fails (provider/transport
- * errors: rate limits, timeouts, connection errors), as opposed to the LLM
- * responding successfully with an empty or unparseable result. Callers can
- * catch this to distinguish "LLM unavailable" from "no facts extracted" and
- * implement retry/fallback logic. Mirrors the Python SDK `LLMError` (see
- * #5903 / #5878).
- */
 export class LLMError extends Error {
   readonly cause?: unknown;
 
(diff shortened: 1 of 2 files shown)


ShauryaaSharma:
Hey @kartik-mem0 
This closes the TS-side follow-up explicitly flagged in #5903 (Python side already fixed via #5878).

Summary of the change and verification is in the PR description. A couple of notes for review:

- `LLMError` is a small standalone class local to `mem0-ts/src/oss/src/memory/index.ts` :  the OSS package doesn't share the `MemoryError` hierarchy from `mem0-ts/src/common/exceptions.ts` (that's client-package-only), so I didn't reuse it.
- Kept the existing `console.error` log line before the throw, matching the Python fix's pattern of logging + re-raising.

Happy to adjust the error shape/naming if you'd prefer something different. Let me know if anything else is needed!

kartik-mem0 (approved):

PR #6121, opened 2026-07-07
fix(opensearch): don't cap unbounded list() at OpenSearch's 10-hit default

### LINKED ISSUE #6108: bug(opensearch): list() without top_k silently caps at 10 results, truncating get_all()/delete_all()
### Component

Core / Python SDK

### Summary

```python
if top_k:
    query["size"] = top_k
```

### Steps to Reproduce

### Expected Behavior

### Fix

freya0926:
## Summary

Closes #6108.

`OpenSearchDB.list()` (`mem0/vector_stores/opensearch.py`) only set the query `size` when `top_k` was truthy:

```python
if top_k:
    query["size"] = top_k
```

`Memory.get_all()` and `Memory.delete_all()` call `vector_store.list(filters=...)` with `top_k=None` to enumerate **all** matching memories. With `size` unset, OpenSearch falls back to its search-API default of **10 hits**, so any user/agent/run with more than 10 memories silently loses the rest:

## Fix

Fall back to `10000` (OpenSearch's own default for `index.max_result_window`) as the query size when `top_k` is `None`/falsy, so unbounded `list()` calls return everything up to that window instead of silently truncating at 10:

```python
query["size"] = top_k if top_k else _DEFAULT_MAX_RESULT_WINDOW
```

The sibling Elasticsearch adapter has the same `if top_k:` shape (noted in the issue) but is out of scope here — this PR only touches the OpenSearch backend per the issue's title.

## Test plan

- [x] Added `test_list_without_top_k_uses_max_result_window` — asserts the query sent to OpenSearch has `size == 10000` when `top_k` is omitted
- [x] Added `test_list_with_top_k_uses_given_size` — regression check that an explicit `top_k` still caps the query at that value
- [x] `pytest tests/vector_stores/test_opensearch.py` — 31 passed (29 existing + 2 new)
- [x] `ruff check` on changed files — clean

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -17,6 +17,11 @@ logger = logging.getLogger(__name__)
 
 _SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_.]*$")
 
+# OpenSearch's own default for `index.max_result_window`. Used as the query
+# size when no top_k is given, so unbounded list() calls (e.g. get_all(),
+# delete_all()) aren't silently capped at OpenSearch's search-API default of 10.
+_DEFAULT_MAX_RESULT_WINDOW = 10000
+
 
 def _validate_filter(key: str, value) -> None:
     if not isinstance(key, str) or not _SAFE_FILTER_KEY.match(key):
@@ -381,8 +386,7 @@ class OpenSearchDB(VectorStoreBase):
             if filter_clauses:
                 query["query"] = {"bool": {"filter": filter_clauses}}
 
-            if top_k:
-                query["size"] = top_k
+            query["size"] = top_k if top_k else _DEFAULT_MAX_RESULT_WINDOW
 
             response = self.client.search(index=self.collection_name, body=query)
             hits = response["hits"]["hits"]
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -20,6 +20,10 @@ _SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_.]*$")
 # OpenSearch's own default for `index.max_result_window`. Used as the query
 # size when no top_k is given, so unbounded list() calls (e.g. get_all(),
 # delete_all()) aren't silently capped at OpenSearch's search-API default of 10.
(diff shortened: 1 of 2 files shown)


yashs33244 (commented):
Same real bug as the ES side - the no-`top_k` path was capped at OpenSearch's default 10 hits. Fix and tests look right.

Two notes:
- `10000` matches mem0's existing "give me everything" convention (`entity_store.list(..., top_k=10000)` in `main.py`), so it's consistent. It's still a ceiling though - >10k matching docs will silently truncate on `get_all`/`delete_all`; a short comment noting that (and `search_after`/PIT for true unboundedness) would help the next reader.
- This looks like it supersedes #6109 (same OpenSearch `list()` size fix, just a renamed constant). Might be worth closing one so a maintainer doesn't end up looking at both.

freya0926:
Thanks for the review, @yashs33244!

- **Ceiling comment:** Pushed in c6cf590 — expanded the `_DEFAULT_MAX_RESULT_WINDOW` comment to call out that 10000 is itself a cap (>10k matching docs still truncate on `get_all()`/`delete_all()`) and that `search_after`/PIT would be the path to true unboundedness.
- **Duplication with #6109:** You're right, it's the same `list()` size fix. In fairness, #6109 (@Bartok9) predates this PR by a day and was opened right alongside the issue, so it has priority. I'm happy to defer — whichever version the maintainers prefer works for me, and I'll gladly close this one if #6109 is the one you'd rather take forward.

Bartok9:
Appreciate the gracious note, @freya0926 — no ego here on my end either. Both PRs fix the same `if top_k:` truncation in `OpenSearchDB.list()`; the only real difference is your named `_DEFAULT_MAX_RESULT_WINDOW` constant + the ceiling comment, which is genuinely a nicer read than my inline `10000`. I'm equally happy to have you fold that constant into #6109 or for maintainers to take whichever they prefer — whatever gets the fix landed fastest for users losing memories past 10. Thanks @yashs33244 for cross-linking so it doesn't get double-reviewed.

yashs33244:
Appreciate how you both handled this. Since the only real delta is the named `_DEFAULT_MAX_RESULT_WINDOW` constant + ceiling comment (which reads better than the bare `10000`), the cleanest outcome is to land whichever PR carries that version and close the other - one review instead of two. Either way the truncation-past-10 fix lands, which is what actually matters for users. I'll leave the final pick to the maintainers.

PR #6122, opened 2026-07-07
fix: raise ImportError instead of blocking on input()/sys.exit() for optional deps

### LINKED ISSUE #6118: Interactive input() and sys.exit(1) on ImportError in library code
### Component

Core / Python SDK

### Summary

### Steps to Reproduce

```python
embedder = OllamaEmbedding()

```

### Expected Behavior

### Actual Behavior

### Environment

- mem0 version: main
- OS: Any

freya0926:
## Summary

Closes #6118.

`mem0/embeddings/ollama.py` and `mem0/proxy/main.py` caught `ImportError` for their optional dependencies (`ollama`, `litellm`) by prompting interactively with `input()` and/or calling `sys.exit(1)`:

- `ollama.py` asked `input("The 'ollama' library is required. Install it now? [y/N]: ")`, which blocks forever on any non-interactive host (pytest, a FastAPI server) — under pytest specifically this raises `OSError: pytest: reading from stdin while output is captured`.
- Both modules called `sys.exit(1)` on failure, which kills the entire host process instead of letting the caller catch and handle a missing optional dependency.

## Fix

Replace both with a plain `raise ImportError(...) from None`, matching the convention already used by every other optional-dependency import in the codebase (e.g. `mem0/llms/litellm.py`, `mem0/embeddings/aws_bedrock.py`, `mem0/embeddings/fastembed.py`):

```python
try:
    from ollama import Client
except ImportError:
    raise ImportError("The 'ollama' library is required. Please install it using 'pip install ollama'.") from None
```

Also removed the now-unused `subprocess`/`sys` imports in both files.

## Test plan

- [x] Added `test_missing_ollama_raises_import_error_without_prompting` (`tests/embeddings/test_ollama_embeddings.py`) and `test_missing_litellm_raises_import_error_without_exiting` (`tests/test_proxy.py`) — both simulate the missing dependency in a fresh subprocess (rather than reloading the module in-process, which would leave the module's classes in an inconsistent state for other tests) and assert: non-zero exit, no `EOFError`/`SystemExit` in stderr (proving `input()`/`sys.exit()` were never invoked), and a clean `ImportError` with the expected install hint.
- [x] `pytest tests/embeddings/test_ollama_embeddings.py tests/test_proxy.py` — 16 passed
- [x] `ruff check` on changed files — clean

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -1,5 +1,3 @@
-import subprocess
-import sys
 from typing import Literal, Optional
 
 from mem0.configs.embeddings.base import BaseEmbedderConfig
@@ -8,17 +6,7 @@ from mem0.embeddings.base import EmbeddingBase
 try:
     from ollama import Client
 except ImportError:
-    user_input = input("The 'ollama' library is required. Install it now? [y/N]: ")
-    if user_input.lower() == "y":
-        try:
-            subprocess.check_call([sys.executable, "-m", "pip", "install", "ollama"])
-            from ollama import Client
-        except subprocess.CalledProcessError:
-            print("Failed to install 'ollama'. Please install it manually using 'pip install ollama'.")
-            sys.exit(1)
-    else:
-        print("The required 'ollama' library is not installed.")
-        sys.exit(1)
+    raise ImportError("The 'ollama' library is required. Please install it using 'pip install ollama'.") from None
 
 
 class OllamaEmbedding(EmbeddingBase):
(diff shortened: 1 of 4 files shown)


kartik-mem0:
Closing as a duplicate of #6119, which was opened first and covers the same change. I compared the diffs line by line and the source changes are substantively identical, so nothing is lost here: #6119 stays open and is the one that will get reviewed and merged.

This is not a comment on the quality of your patch. Thank you for the time you put into it, and apologies that our review queue let two people do the same work. If you believe this PR covers something #6119 does not, reply here and I will reopen it.

PR #6123, opened 2026-07-07
fix(elasticsearch): don't cap unbounded list() at ES default of 10

Bartok9:
## Summary

- `ElasticsearchDB.list()` only set the query `size` when `top_k` was truthy, so unbounded calls fell back to Elasticsearch's search-API default of **10 hits**.
- User-facing impact: `get_all()` / `delete_all()` (which call `Memory.list(filters=...)[0]` with no `top_k`) silently returned/operated on at most 10 memories regardless of how many exist.

## Motivation

`mem0/memory/main.py` calls `self.vector_store.list(filters=filters)[0]` with no `top_k`. In the Elasticsearch adapter:

```python
if top_k:
    query["size"] = top_k
# else: no size set -> Elasticsearch defaults to 10
```

Elasticsearch caps the response at the default `size` of 10 when the field is absent, so any store with >10 memories is silently truncated on unbounded reads.

This is the Elasticsearch analogue of the OpenSearch `list()` cap (issue #6108, fix #6121). The existing Elasticsearch PRs #5910/#5951 fix the KNN **`search()`** path only; the unbounded **`list()`** path is untouched.

## Fix

Default the query size to `index.max_result_window` (10000) when no `top_k` is given, mirroring the OpenSearch adapter, while still honouring an explicit `top_k`:

```python
query["size"] = top_k if top_k else _DEFAULT_MAX_RESULT_WINDOW
```

## Verification

- `python -m pytest tests/vector_stores/test_elasticsearch.py` — **22 passed**
- New regression tests:
  - `test_list_without_top_k_uses_max_result_window` — asserts `size == 10000` (fails without the fix: `KeyError: 'size'`)
  - `test_list_with_top_k_uses_given_size` — asserts an explicit `top_k=25` still caps at 25
- Did NOT change: the `search()` / `keyword_search()` paths or the `top_k`-provided behaviour.

### DIFF
--- a/mem0/vector_stores/elasticsearch.py
+++ b/mem0/vector_stores/elasticsearch.py
@@ -15,6 +15,11 @@ from mem0.vector_stores.base import VectorStoreBase
 
 logger = logging.getLogger(__name__)
 
+# Elasticsearch's own default for `index.max_result_window`. Used as the query
+# size when no top_k is given, so unbounded list() calls (e.g. get_all(),
+# delete_all()) aren't silently capped at Elasticsearch's search-API default of 10.
+_DEFAULT_MAX_RESULT_WINDOW = 10000
+
 
 class OutputData(BaseModel):
     id: str
@@ -282,8 +287,7 @@ class ElasticsearchDB(VectorStoreBase):
                 filter_conditions.append({"term": {f"metadata.{key}": value}})
             query["query"] = {"bool": {"must": filter_conditions}}
 
-        if top_k:
-            query["size"] = top_k
+        query["size"] = top_k if top_k else _DEFAULT_MAX_RESULT_WINDOW
 
         response = self.client.search(index=self.collection_name, body=query)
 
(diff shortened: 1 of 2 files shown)


yashs33244 (commented):
This is a real bug - the no-`top_k` path (`vector_store.list(filters)` in `main.py`, i.e. `get_all`) was silently getting Elasticsearch's default 10 hits. Good that the explicit-`top_k` path is untouched and both directions are covered in tests.

Two notes:

- The `10000` default lines up with how mem0 already treats "give me everything" elsewhere - `entity_store.list(..., top_k=10000)` in `main.py` uses the same ceiling - so this is consistent with existing convention, not a new magic number.
- Worth being explicit that `10000` is still a ceiling: a collection with more than 10k matching docs will silently truncate on `get_all`/`delete_all`. It matches the existing convention so I wouldn't block on it, but a one-line comment noting the ceiling (and, if you want true unboundedness later, `search_after`/PIT) would save the next reader from assuming it's fully unbounded.

Bartok9:
Thanks for the thorough review! Added a comment on `_DEFAULT_MAX_RESULT_WINDOW` making the ceiling explicit — it flags that collections with >10k matching docs still truncate on `get_all`/`delete_all`, and points future readers to `search_after`/PIT for genuinely unbounded scans. Kept `10000` to stay consistent with the existing `entity_store.list(..., top_k=10000)` convention rather than introducing a new magic number.

PR #6124, opened 2026-07-07
fix(databricks): don't KeyError in list() when row lacks 'memory' column

Bartok9:
## Summary

- `DatabricksDB.list()` built `payload["data"] = payload["memory"]` with a hard key access, which raises `KeyError` on any index whose schema/row omits the `memory` column.
- User-facing impact: the `KeyError` is swallowed by the surrounding `try/except` and `list()` returns a bare `[]`, so callers doing `list(...)[0]` (e.g. `Memory.delete_all`) then crash with `IndexError`.

## Motivation

In `mem0/vector_stores/databricks.py`, `search()` (line 515) and `get()` (line 564) already guard this safely:

```python
payload["data"] = payload.get("memory", "")
```

But `list()` did:

```python
payload['data'] = payload['memory']   # KeyError if 'memory' absent
```

`payload` is built from `self.column_names`, so an index whose schema doesn't include a `memory` column (legacy/custom tables) makes every `list()` call fail. Because the failure is caught and returns `[]`, `Memory.delete_all()` (`memories = self.vector_store.list(filters=filters)[0]`) then raises `IndexError: list index out of range`.

## Fix

Align `list()` with the safe accessor already used by `search()` and `get()`:

```python
payload["data"] = payload.get("memory", "")
```

One-line change; no behaviour change when the column is present.

## Verification

- `python -m pytest tests/vector_stores/test_databricks.py` — **64 passed**
- New regression test `test_list_memories_row_without_memory_column`:
  - Without the fix: `KeyError: 'memory'` is swallowed → `list()` returns `[]` → `res[0]` raises `IndexError`.
  - With the fix: `list()` returns the row with `payload["data"] == ""`.
- Did NOT change: `search()` / `get()` (already correct) or the happy path when `memory` is present.

### DIFF
--- a/mem0/vector_stores/databricks.py
+++ b/mem0/vector_stores/databricks.py
@@ -828,7 +828,7 @@ class Databricks(VectorStoreBase):
                     except Exception:
                         pass
                 memory_id = row_dict.get("memory_id") or row_dict.get("id")
-                payload['data'] = payload['memory']
+                payload["data"] = payload.get("memory", "")
                 memory_results.append(MemoryResult(id=memory_id, payload=payload))
             return [memory_results]
         except Exception as e:
--- a/tests/vector_stores/test_databricks.py
+++ b/tests/vector_stores/test_databricks.py
@@ -538,6 +538,23 @@ def test_list_memories_default_limit(db_instance_delta, mock_workspace_client):
     assert call_kwargs["num_results"] == 100
 
 
+def test_list_memories_row_without_memory_column(db_instance_delta, mock_workspace_client):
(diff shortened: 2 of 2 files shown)


yashs33244 (commented):
LGTM. The `payload.get("memory", "")` guard matches exactly what `search()` and `get()` already do (databricks.py:515 and :564), so this brings `list()` in line rather than inventing a new default - and the test covers the missing-column row directly. Only thing I noticed nearby is line 709 defaults slightly differently (`get("memory", get("data", "unknown"))`), but that's out of scope for this fix.

PR #6131, opened 2026-07-07
fix(memory): don't send text-only list content to the vision LLM

### LINKED ISSUE #6129: bug(memory): text-only list content sent to the vision LLM when enable_vision is on
### Component

Core / Python SDK

### Description

When `enable_vision` is on, `parse_vision_messages` (`mem0/memory/utils.py`) routes **every** list-type message content to the vision LLM, even when the list contains no image. A text-only multimodal-format message is handed to `get_image_description`, which prompts the model with *"A user is providing an image. Provide a high level description of the image..."*. The model has no image, so it returns a confused description, and that description **replaces the user's actual text**. The real content is silently discarded before fact extraction, plus an unnecessary (and costly) vision call is made on every such message.

The `llm is None` branch already handles this correctly by extracting the text parts; only the vision-enabled branch is affected.

### Steps to Reproduce

```python

llm = Mock()
llm.generate_response.return_value = "A photo of ..."  # vision model's guess

messages = [{"role": "user", "content": [{"type": "text", "text": "I love pizza"}]}]

# enable_vision=True passes the llm through:
result = parse_vision_messages(messages, llm=llm, vision_details="auto")
# result[0]["content"] == "A photo of ..."   <- real text "I love pizza" is gone
# llm.generate_response was called for a message that has no image
```

### Expected Behavior

A list with no image part should have its text extracted (exactly as it is when `enable_vision` is off). Only a list that actually contains an `image_url` part should be routed to the vision model.

### Actual Behavior

Text-only list content is sent to the vision model, the real text is dropped and replaced by an image description of a nonexistent image, and a vision call is wasted.

### Environment

- mem0 version: main
- Python: 3.10+

yashs33244:
## Linked Issue

Closes #6129

## Description

When `enable_vision` is on, `parse_vision_messages` routed **every** list-type message content to the vision LLM, even lists with no image. A text-only multimodal message (`content: [{"type": "text", "text": "..."}]`) was handed to `get_image_description`, which prompts the model to describe an image, so the real text was dropped and replaced by a description of a nonexistent image (and a vision call was wasted on every such message).

This routes to the vision model only when the list actually contains an `image_url` part; otherwise it extracts the text parts, exactly as the `enable_vision`-off path already does. Image-only and mixed image+text lists are unchanged.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

`test_text_only_list_with_llm_does_not_call_vision` fails on `main` (the real text is replaced by the mock vision response and `generate_response` is called) and passes with the fix. Full `tests/memory/` suite: 352 passed, ruff clean.

## Checklist

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -192,17 +192,23 @@ def parse_vision_messages(messages, llm=None, vision_details="auto"):
 
         # Handle message content
         if isinstance(content, list):
-            if llm is None:
+            has_image = any(
+                isinstance(part, dict) and part.get("type") == "image_url" for part in content
+            )
+            # Only route to the vision LLM when an image is actually present; a
+            # text-only list would otherwise be sent to the vision model
+            # (prompted to describe an image), discarding the real text.
+            if llm is not None and has_image:
+                description = get_image_description(msg, llm, vision_details)
+                returned_messages.append({"role": role, "content": description})
+            else:
                 text_parts = [
-                    part["text"] for part in msg["content"]
+                    part["text"] for part in content
                     if isinstance(part, dict) and part.get("type") == "text"
                 ]
                 if not text_parts:
                     continue
                 returned_messages.append({"role": role, "content": " ".join(text_parts)})
-            else:
-                description = get_image_description(msg, llm, vision_details)
-                returned_messages.append({"role": role, "content": description})
         elif isinstance(content, dict) and content.get("type") == "image_url":
             if llm is None:
                 continue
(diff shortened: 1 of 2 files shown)


yashs33244:
Gentle nudge 🙏 small fix, green CI, open since early July. It skips sending text-only list content to the vision LLM, avoiding a needless multimodal call (and the error that follows). @kartik-mem0 @deshraj a quick look whenever you get a moment would be appreciated.

yashs33244:
Been open a few weeks with green CI, still awaiting a first review 🙏 This one stops text-only list content being sent to the vision LLM (needless multimodal call + error). @kartik-mem0 @Himanshu-Sangshetti @karthik-indla - whenever one of you has a window, would really appreciate a look. Thanks!

PR #6139, opened 2026-07-07
fix(ts): default Redis filterExpr to * for empty filters

### LINKED ISSUE #6013: bug(ts-oss/redis): empty/all-null filters build an invalid query ("" instead of "*")

asif786ka:
## Summary
- Extract buildRedisFilterExpr helper
- Return * when filter clauses would otherwise be empty

Fixes #6013

## Test plan
- [x] Search/list with {} filters no longer builds blank filter expression

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/redis.ts
+++ b/mem0-ts/src/oss/src/vector_stores/redis.ts
@@ -22,6 +22,18 @@ function escapeRedisTagValue(value: unknown): string {
   );
 }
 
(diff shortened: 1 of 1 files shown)


kartik-mem0:
closing this pr in favor of #6014

PR #6151, opened 2026-07-08
fix(memory): parse chatty JSON payloads

iroiro147:
Summary
- parse chatty fallback JSON by trying valid JSON candidates instead of slicing from the first brace to the last brace
- prefer later `memory` / `facts` payload objects over earlier valid non-payload JSON snippets
- keep a generic first-valid-JSON fallback for other callers

Refs #5998.

Tests
- `uv run --extra test pytest tests/test_chatty_llm_parsing.py tests/memory/test_memory_utils.py -q` (44 passed)
- `uv run --with ruff ruff format --check mem0/memory/utils.py tests/test_chatty_llm_parsing.py`
- `uv run --with ruff ruff check mem0/memory/utils.py tests/test_chatty_llm_parsing.py`
- `git diff --check`

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -1,4 +1,5 @@
 import hashlib
+import json
 import logging
 import re
 from typing import Any, Dict, List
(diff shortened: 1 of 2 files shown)


iroiro147:
I have read and agree to the CLA; sign personally with sarthak.singh@juspay.in.

PR #6152, opened 2026-07-08
fix(memory): remove_code_blocks must not crash on None or list content

### LINKED ISSUE #6150: Procedural memory creation crashes with AttributeError instead of a clear error when the LLM returns no content
### Component

Core / Python SDK

### Description

aloktripathi1:
Closes #6150

remove_code_blocks(content: str) in mem0/memory/utils.py assumed content is always a string. It's called at 5 sites in main.py. 3 of them (fact extraction, sync and async) already catch this and degrade gracefully to []. The other 2 (_create_procedural_memory, sync and async) catch, log, and re-raise, so a crash here is user-facing.

content can legitimately be None. Per the OpenAI SDK's own type definition, ChatCompletionMessage.content is Union[str, None], populated as None on safety refusals where refusal is set instead. remove_code_blocks(None) then crashed with AttributeError: 'NoneType' object has no attribute 'strip' instead of a clear error.

The async LangChain path has a second trigger: content can be a list of content blocks instead of a plain string.

Fixed at the root, in remove_code_blocks() itself. None returns an empty string (matches how the 3 already-protected callers treat "nothing extracted"). Lists are joined, handling both plain strings and LangChain-style {"text": ...} dict blocks. Any other non-str type is coerced with str(). Existing str behavior is byte-for-byte unchanged.

16 tests pass (9 utility, 7 procedural memory covering sync, async, and async-LangChain paths). Red-green verified. Ruff clean.

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -112,15 +112,30 @@ def normalize_facts(raw_facts):
     return normalized
 
 
-def remove_code_blocks(content: str) -> str:
+def remove_code_blocks(content: Any) -> str:
     """
     Removes enclosing code block markers ```[language] and ``` from a given string.
 
     Remarks:
+    - If content is None, returns an empty string.
+    - If content is a list of content blocks, extracts string items and dict item "text" values before processing.
     - The function uses a regex pattern to match code blocks that may start with ``` followed by an optional language tag (letters or numbers) and end with ```.
     - If a code block is detected, it returns only the inner content, stripping out the markers.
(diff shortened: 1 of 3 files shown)


aloktripathi1:
The 3.12 CI failure is an unrelated flaky timing test (TestEntityBoostParallelism::test_async_searches_run_concurrently, asserts wall-clock < 0.75s, got 0.82s on a shared runner). Not touched by this PR (1673/1674 tests pass). 3.10 and 3.11 both pass clean. Should be fine on a re-run.

PR #6154, opened 2026-07-08
Document history store lifecycle and paths

RobinALG87:
﻿Configuration docs mentioned the history store but did not explain what it tracks, how to override paths, or that deleting the SQLite file does not purge vectors. This section closes that gap for both Python and TypeScript OSS users.


kartik-mem0:
Thank you for contributing documentation about the history store. We’re closing this version because the TypeScript path guidance does not match the implementation: the default history database is `memory.db`, and `MEM0_DIR` does not relocate the TypeScript history database. The distinction between history and vector storage is useful, but the language-specific defaults need correction before this can be included. We appreciate your effort.

PR #6157, opened 2026-07-08
Minimal OSS add/search/delete example script

RobinALG87:
﻿examples/misc/ had integration demos but no small standalone OSS script showing the core store-search-delete loop. This adds a short example using default Memory() and OPENAI_API_KEY.


RobinALG87:
Closing this to keep the review queue clean because it overlaps with #6153's add/search/delete quickstart flow. I may revisit with a narrower example if there is maintainer interest.

PR #6158, opened 2026-07-08
Clarify embedder environment variables and dimension alignment

RobinALG87:
﻿Embedder config docs listed parameters but not which environment variables back each provider. This adds a provider-to-env-var table and a note to align embedding dimensions with the vector store.


kartik-mem0:
Thank you for documenting the embedder environment variables and dimension alignment. We’re closing this version because the TypeScript dimension guidance is too broad: the shared configuration and Qdrant use `dimension`, while some other providers use `embeddingModelDims`. Presenting one field as universal can mislead users. A revised explanation should distinguish the provider-specific settings. We appreciate the contribution.

PR #6166, opened 2026-07-09
fix(mem0): add test for issue #6118

### LINKED ISSUE #6118: Interactive input() and sys.exit(1) on ImportError in library code

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6118.

## Why
Fixes #6118

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6118.py
+++ b/tests/test_issue_6118.py
@@ -1,0 +1,2 @@
+from mem0.embeddings.ollama import OllamaEmbedding
+embedder = OllamaEmbedding()


kartik-mem0:

kartik-mem0:

**Finding:** `tests/test_issue_6118.py` contains no test function, only two module-level statements:

```python
embedder = OllamaEmbedding()
```

pytest reports `no tests ran`. The constructor also reads from stdin, so collection fails with `OSError: pytest: reading from stdin while output is captured!`, which is ironically the very behaviour issue #6118 reports.

---

Closing this one. To be clear, this is about the test file, not about issue #6118 — the underlying report is legitimate and stays as it is.

PR #6167, opened 2026-07-09
fix(mem0): add test for issue #6101

### LINKED ISSUE #6101: TypeScript SDK: LLM extraction transport failures silently swallowed (index.ts:855-858)

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6101.

## Why
Fixes #6101

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6101.py
+++ b/tests/test_issue_6101.py
@@ -1,0 +1,51 @@
+"""Regression test for issue #6101.
+
+Verifies that an LLM provider/transport failure during extraction propagates to
+the caller instead of being silently collapsed into an empty extraction result.
+"""
+
+import importlib.metadata
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+
+importlib.metadata.version = MagicMock(return_value="0.0.0")
+sys.modules.setdefault("posthog", MagicMock(Posthog=MagicMock()))
+
+from mem0 import Memory
+from mem0.exceptions import LLMError
+
+
+def test_issue_6101(monkeypatch):
+    class _ProviderError(Exception):
+        pass
+
+    mock_embedder = MagicMock()
+    mock_embedder.embed.return_value = [0.1, 0.2, 0.3]
+
+    mock_vector_store = MagicMock()
+    mock_vector_store.search.return_value = []
+
+    mock_history_store = MagicMock()
+
+    mock_llm = MagicMock()
+    mock_llm.generate_response.side_effect = _ProviderError("429 rate limit")
+
+    monkeypatch.setattr("mem0.utils.factory.EmbedderFactory.create", MagicMock(return_value=mock_embedder))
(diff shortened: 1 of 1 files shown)


kartik-mem0:
closing this pr in favour of #6102 

thank you for your contribution here

PR #6169, opened 2026-07-09
fix(pi-agent-plugin): tag global-scope adds so they stay searchable

### LINKED ISSUE #6168: pi-agent-plugin: memories added under scope="global" are permanently unreachable by search/get_all
### Component

integrations/pi-agent-plugin

### Description

`resolveAddParams` and `resolveSearchFilters` in `integrations/pi-agent-plugin/src/memory/scoping.ts` disagree on what `app_id` a `"global"`-scoped memory should have:

- `resolveAddParams("global", ctx)` returns `{ userId: ctx.userId }` — no `appId`. Per mem0's documented write semantics ("Absent fields default to `null`" — `docs/platform/features/entity-scoped-memory.mdx:41`), the memory is persisted with `app_id: null`.
- `resolveSearchFilters("global", ctx)` returns `{ user_id: ctx.userId, app_id: "*" }`. Per the documented wildcard semantics ("Wildcards (`"*"`) include only non-null values" — same doc:156, and `docs/platform/features/v2-memory-filters.mdx:62`: "The `*` wildcard matches any non-null value... Records with null values for that field are excluded"), this filter explicitly *excludes* `app_id: null` records.

Both `tools.ts`'s `search` action (line 59-60) and `get_all` action (line 86-87) call `resolveSearchFilters`, so this asymmetry affects every read path.

### Impact

A user runs `/mem0-scope global` (`commands.ts:271-275`) or passes `scope: "global"` to the `mem0_memory` tool, then does `action: "add"` — the memory is stored with `app_id: null`. Any later `search` or `get_all`, in any scope, will never surface it: global scope's own wildcard filter excludes null, and project/session scope filter on a different, non-null `app_id`. The memory is write-once and permanently unreachable through this plugin, with no error surfaced to the user.

There's a second, more serious issue in the same root cause: `delete_all` under global scope reuses `resolveAddParams` for its filter (`tools.ts:119-120`), which currently resolves to just `{ userId }`. I traced this into the actual `mem0ai` client source (`node_modules/mem0ai/dist/index.js`, `deleteAll()`): it hits the plain `DELETE /v1/memories/?user_id=...` REST endpoint with the params passed straight through as query params — no implicit-null-scoping (that behavior belongs to the `/v3/memories/search/` filter engine, a different endpoint). So today, a user who switches to global scope and asks to clear their global memories would actually **delete every memory they have, across every project and session** — not just the global-scoped ones.

### Reproduction (code tracing, confirmed against `mem0ai@3.0.7` client source)

1. `scope: "global"`, `action: "add"`, content: "some fact" → stored with `app_id: null`.
2. `scope: "global"`, `action: "search"` or `"get_all"` → filters on `app_id: "*"`, excludes the record from step 1. Never returned.
3. `scope: "global"`, `action: "delete_all"` → filters on `{ user_id }` only, sent as a raw query param to `DELETE /v1/memories/`, deleting all memories for that user regardless of project/session.

### Proposed fix

Give `scope: "global"` adds a reserved, non-null `appId` (e.g. `"__global__"`) instead of omitting it. This keeps the memory inside the existing global-search wildcard's match set — no read-path change needed, since the wildcard already correctly returns every real project's memories and shouldn't be touched — and as a side effect narrows `delete_all(global)` to only the memories actually added under global scope, instead of the current unconstrained delete-everything behavior.

I have a fix + regression tests ready (all existing plugin tests pass, 104/104) and will open a PR referencing this issue.

freya0926:
## Linked Issue

Closes #6168

## Description

`resolveAddParams("global", ctx)` in `integrations/pi-agent-plugin/src/memory/scoping.ts` wrote memories with no `appId`, persisting `app_id: null`. `resolveSearchFilters("global", ctx)` searches with `app_id: "*"`, which per mem0's documented wildcard semantics only matches non-null values. So any memory added under `scope: "global"` was permanently unreachable by `search`/`get_all`, in any scope — a silent, no-error data-loss-equivalent bug.

`delete_all` under global scope reuses the same `resolveAddParams` output as its filter. Since `deleteAll()` hits the plain `DELETE /v1/memories/` REST endpoint (raw query params, no implicit-null-scoping — that behavior only exists in the `/v3/memories/search/` filter engine), the pre-fix `{ userId }`-only filter would delete **every memory the user has, across all projects and sessions**, not just global-scoped ones, whenever a user cleared memories while in global scope.

## Fix

Tag `scope: "global"` adds with a reserved sentinel `appId` (`GLOBAL_APP_ID = "__global__"`) instead of leaving it unset:

```ts
case "global":
  return { userId: ctx.userId, appId: GLOBAL_APP_ID };
```

This keeps the memory inside the *existing* global-search wildcard's match set. I deliberately left `resolveSearchFilters("global", ...)` untouched — it already correctly returns every real project's memories via the `"*"` wildcard (matches the README's documented "All memories across all your projects" behavior for project/session-scoped data), so narrowing it to drop the wildcard would have broken that working case just to fix the global-add case. Tagging the write side was the narrower, non-regressive fix.

As a side effect, `delete_all(global)` now filters on `{ userId, appId: GLOBAL_APP_ID }`, correctly scoping it to just the memories added under global scope instead of the previous unconstrained delete.

## Type of Change

## Breaking Changes

N/A — `GLOBAL_APP_ID` is a new internal sentinel value; no public API changes. Existing global-scope memories already written with `app_id: null` before this fix won't retroactively pick up the tag (pre-existing null-tagged records stay unreachable, same as today); only new adds after this fix benefit.

## Test Coverage

Updated `src/memory/scoping.test.ts` and the (pre-existing, duplicate) `tests/scoping.test.ts` / `tests/tools.test.ts` to assert the new sentinel-tagged behavior, and added a symmetry check (`resolveAddParams` / `resolveSearchFilters` scoping fields stay in sync`) across all three scopes so this asymmetry class can't silently regress again. Full suite: 104/104 passing. `tsc --noEmit` clean. `tsup` build succeeds.

## Checklist

- [x] I have updated documentation if needed (n/a — internal sentinel, no public-facing docs to update)

### DIFF
--- a/integrations/pi-agent-plugin/src/memory/scoping.test.ts
+++ b/integrations/pi-agent-plugin/src/memory/scoping.test.ts
@@ -79,12 +79,21 @@ describe("resolveAddParams", () => {
     expect(resolveAddParams("session", ctx)).toEqual({ userId: "u1", appId: "a1", runId: "r1" });
   });
 
-  it("tags global scope with the reserved GLOBAL_APP_ID instead of leaving appId unset", () => {
+  it("tags global scope with the reserved sentinel appId instead of leaving appId unset", () => {
     // Regression test: resolveSearchFilters("global", ...) filters on app_id: "*",
     // which only matches non-null values. Writing without an appId would persist
     // app_id: null and make the memory permanently unreachable by that search —
     // see resolveSearchFilters's "global" test above for the read-side half.
-    expect(resolveAddParams("global", ctx)).toEqual({ userId: "u1", appId: GLOBAL_APP_ID });
+    //
+    // Assert the literal, not the imported GLOBAL_APP_ID: without the fix that
+    // import resolves to undefined, and toEqual treats an undefined-valued
+    // property as equal to a missing one — so asserting against the constant
+    // would pass against the very code this test exists to catch.
+    expect(resolveAddParams("global", ctx)).toEqual({ userId: "u1", appId: "__global__" });
+  });
+
+  it("exports the sentinel the global add path tags with", () => {
+    expect(GLOBAL_APP_ID).toBe("__global__");
   });
 });
 
--- a/integrations/pi-agent-plugin/src/memory/scoping.test.ts
+++ b/integrations/pi-agent-plugin/src/memory/scoping.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { detectRunId, resolveSearchFilters, resolveAddParams } from "./scoping.ts";
+import { detectRunId, resolveSearchFilters, resolveAddParams, GLOBAL_APP_ID } from "./scoping.ts";
 
 const mockExecFileSync = vi.fn();
 
@@ -79,7 +79,30 @@ describe("resolveAddParams", () => {
     expect(resolveAddParams("session", ctx)).toEqual({ userId: "u1", appId: "a1", runId: "r1" });
   });
 
-  it("only includes userId for global scope", () => {
-    expect(resolveAddParams("global", ctx)).toEqual({ userId: "u1" });
+  it("tags global scope with the reserved GLOBAL_APP_ID instead of leaving appId unset", () => {
+    // Regression test: resolveSearchFilters("global", ...) filters on app_id: "*",
+    // which only matches non-null values. Writing without an appId would persist
+    // app_id: null and make the memory permanently unreachable by that search —
+    // see resolveSearchFilters's "global" test above for the read-side half.
+    expect(resolveAddParams("global", ctx)).toEqual({ userId: "u1", appId: GLOBAL_APP_ID });
   });
+});
+
+describe("resolveAddParams / resolveSearchFilters scoping fields stay in sync", () => {
+  const ctx = { userId: "u1", appId: "a1", runId: "r1" };
+  const scopingFields: Record<string, string> = { userId: "user_id", appId: "app_id", runId: "run_id" };
+
+  for (const scope of ["project", "session", "global"] as const) {
+    it(`add sets every scoping field that search filters on for "${scope}" scope`, () => {
+      // A memory added under a given scope must be findable by a search under that
+      // same scope: search can't require a field (as a non-null match or wildcard)
+      // that the write never sets to a non-null value.
+      const addKeys = new Set(Object.keys(resolveAddParams(scope, ctx)));
+      const searchKeys = new Set(
+        Object.keys(resolveSearchFilters(scope, ctx)).map((k) => scopingFields[k] ?? k),
+      );
+      const addKeysNormalized = new Set([...addKeys].map((k) => scopingFields[k] ?? k));
+      expect(addKeysNormalized).toEqual(searchKeys);
+    });
+  }
 });
--- a/integrations/pi-agent-plugin/tests/scoping.test.ts
+++ b/integrations/pi-agent-plugin/tests/scoping.test.ts
@@ -41,7 +41,13 @@ describe("resolveAddParams", () => {
     });
   });
 
-  it("global scope tags appId with the reserved GLOBAL_APP_ID sentinel", () => {
-    expect(resolveAddParams("global", ctx)).toEqual({ userId: "kartik", appId: GLOBAL_APP_ID });
+  it("global scope tags appId with the reserved sentinel", () => {
+    // Literal, not the imported GLOBAL_APP_ID — see src/memory/scoping.test.ts
+    // for why asserting against the constant would pass on the unfixed code.
+    expect(resolveAddParams("global", ctx)).toEqual({ userId: "kartik", appId: "__global__" });
+  });
+
+  it("exports the sentinel the global add path tags with", () => {
+    expect(GLOBAL_APP_ID).toBe("__global__");
   });
 });
--- a/integrations/pi-agent-plugin/tests/scoping.test.ts
+++ b/integrations/pi-agent-plugin/tests/scoping.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { resolveAddParams, resolveSearchFilters } from "../src/memory/scoping.ts";
+import { resolveAddParams, resolveSearchFilters, GLOBAL_APP_ID } from "../src/memory/scoping.ts";
 import type { ScopeContext } from "../src/types.ts";
 
 const ctx: ScopeContext = {
@@ -41,7 +41,7 @@ describe("resolveAddParams", () => {
     });
   });
 
-  it("global scope returns userId only", () => {
-    expect(resolveAddParams("global", ctx)).toEqual({ userId: "kartik" });
+  it("global scope tags appId with the reserved GLOBAL_APP_ID sentinel", () => {
+    expect(resolveAddParams("global", ctx)).toEqual({ userId: "kartik", appId: GLOBAL_APP_ID });
   });
 });
--- a/integrations/pi-agent-plugin/tests/tools.test.ts
+++ b/integrations/pi-agent-plugin/tests/tools.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi } from "vitest";
 import { buildToolExecute } from "../src/memory/tools.ts";
+import { GLOBAL_APP_ID } from "../src/memory/scoping.ts";
 import type { ScopeContext } from "../src/types.ts";
 
 const mockMem0 = {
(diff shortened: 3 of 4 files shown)


integrations/pi-agent-plugin/src/memory/scoping.test.ts
```diff
@@ -79,7 +79,30 @@ describe("resolveAddParams", () => {
...
   });
 
-  it("only includes userId for global scope", () => {
-    expect(resolveAddParams("global", ctx)).toEqual({ userId: "u1" });
+  it("tags global scope with the reserved GLOBAL_APP_ID instead of leaving appId unset", () => {
+    // Regression test: resolveSearchFilters("global", ...) filters on app_id: "*",
+    // which only matches non-null values. Writing without an appId would persist
+    // app_id: null and make the memory permanently unreachable by that search —
+    // see resolveSearchFilters's "global" test above for the read-side half.
+    expect(resolveAddParams("global", ctx)).toEqual({ userId: "u1", appId: GLOBAL_APP_ID });
```
kartik-mem0: Passes unchanged if scoping.ts is reverted to before this fix: GLOBAL_APP_ID then imports as undefined, and toEqual treats the missing appId as equal to it. Assert against the literal "__global__" string instead of the imported constant so a regression is actually caught.

integrations/pi-agent-plugin/tests/scoping.test.ts
```diff
@@ -41,7 +41,7 @@ describe("resolveAddParams", () => {
     });
   });
 
-  it("global scope returns userId only", () => {
-    expect(resolveAddParams("global", ctx)).toEqual({ userId: "kartik" });
+  it("global scope tags appId with the reserved GLOBAL_APP_ID sentinel", () => {
+    expect(resolveAddParams("global", ctx)).toEqual({ userId: "kartik", appId: GLOBAL_APP_ID });
```
kartik-mem0: Same false-negative as the src/memory/scoping.test.ts version: passes unchanged with the fix reverted, because GLOBAL_APP_ID imports as undefined pre-fix and toEqual treats it as equal to the then-missing appId. Hard-code "__global__" here too.

integrations/pi-agent-plugin/tests/tools.test.ts
```diff
@@ -46,6 +47,14 @@ describe("buildToolExecute", () => {
     });
   });
 
+  it("add with scope=global tags appId with GLOBAL_APP_ID so it stays reachable by the global search wildcard", async () => {
+    mockMem0.add.mockResolvedValue([{ id: "new-id", memory: "test" }]);
+    await execute({ action: "add", content: "Cross-project preference", scope: "global" });
+    const call = mockMem0.add.mock.calls[mockMem0.add.mock.calls.length - 1];
+    expect(call[1].userId).toBe("testuser");
+    expect(call[1].appId).toBe(GLOBAL_APP_ID);
```
kartik-mem0: Same issue via .toBe: GLOBAL_APP_ID imports as undefined when scoping.ts is reverted, and call[1].appId is also undefined then, so .toBe(GLOBAL_APP_ID) passes either way. Assert .toBe("__global__") directly.

kartik-mem0 (commented):
## Changes requested

The `GLOBAL_APP_ID` fix genuinely closes both the reachability bug this PR names and the delete_all blast radius it doesn't (verified below), but 3 of the 4 new/updated test assertions pass unchanged even with the fix reverted.

**To unblock:**
1. `integrations/pi-agent-plugin/src/memory/scoping.test.ts:87` — assert the literal `"__global__"`, not the imported constant (detail inline)
2. `integrations/pi-agent-plugin/tests/scoping.test.ts:45` — same fix (detail inline)
3. `integrations/pi-agent-plugin/tests/tools.test.ts:55` — same fix (detail inline)
4. `integrations/pi-agent-plugin/src/memory/tools.ts:117-124` — add a direct test for `delete_all` under `scope: "global"` (no line to anchor to, this file is untouched by the PR); this is the more severe half of the original bug and has zero coverage today

Follow-up, not blocking this PR: the identical bug is still unfixed in a sibling package, `integrations/mem0-plugin/.opencode-plugin/scope.ts:44` (its own header says it was ported from this exact file) — worth its own issue.

- **Problem is real:** `scoping.ts:35` (main) filters global search on `app_id: "*"` (non-null only); `scoping.ts:49` (main) wrote global adds with no `appId` at all, so `app_id: null` was permanently unreachable. `tools.ts:119` (unchanged by this PR) reuses `resolveAddParams` as the `delete_all` filter, so pre-fix that action sent no `app_id` constraint at all under global scope
- **Tests / red-green:** reverting only `scoping.ts` to `main` (test files kept as this PR wrote them) drops 26 of 27 tests to still-passing; only the new "stay in sync" symmetry test (`scoping.test.ts:91-108`) actually fails. The other 3 pass because they import `GLOBAL_APP_ID` from `scoping.ts`, which resolves to `undefined` when the export doesn't exist yet, and `toEqual`/`.toBe` treat the then-missing/undefined `appId` as equal to that `undefined`
- **delete_all fix, verified empirically:** pre-fix params to `mem0.deleteAll` under global scope are `{userId}` only (no `app_id` at all, per the vendored `mem0ai@3.0.7` client and `docs/openapi.json`'s documented filter semantics); post-fix they include `appId: "__global__"`, a real non-wildcard scoping value
- **CI:** Pi Agent Plugin test (Node 20, 22), build, lint, CI Gate all pass. Only `Vercel` is red (deploy preview, no perms from a fork, not merge-blocking)
- **Security:** cleared, pure function over an already-authenticated `ScopeContext`, no new external call or user-input parsing
- **Bloat (ponytail):** lean already, one-line implementation change plus one well-designed parametrized test
- **Not verified:** live Mem0 Platform API call; the delete_all mechanism is confirmed by tracing the vendored client and the OpenAPI spec, not a live request

</details>

The root-cause fix itself is right, and the sentinel approach is a clean, minimal shape. Strengthen the three assertions and add the delete_all test and I will re-review.

freya0926:
Thanks for the careful review — you were right on all four points, and the red-green gap was real. Verified it directly: reverting only scoping.ts to main (test files kept as this PR wrote them) took the suite from 1 failing test to 7.

1–3. Switched the three sentinel assertions to the literal "__global__". As you noted, the imported GLOBAL_APP_ID resolves to undefined when the export is absent, and both toEqual/toBe treat that as equal to the missing appId the bug produces. Also added a test pinning the exported constant's own value, so the constant and the behavior are covered separately.

4. Added delete_all coverage under scope: "global", asserting deleteAll is called with { userId, appId: "__global__" } — pre-fix this degrades to { userId } alone, so the test fails on exactly the unconstrained-delete behavior. Added a scope: "project" case alongside it as a contrast guard.

With the fix in place: 108/108 tests pass, tsc --noEmit clean, tsup build succeeds.

On the follow-up: confirmed the same asymmetry in [integrations/mem0-plugin/.opencode-plugin/scope.ts](https://github.com/mem0ai/mem0/blob/main/integrations/mem0-plugin/.opencode-plugin/scope.ts) — scopeAddParams returns { user_id } for global while scopeSearchFilters filters on app_id: "*". Will open a separate issue for it once this lands.

integrations/pi-agent-plugin/src/memory/scoping.test.ts
kartik-mem0: Verified against 4e3eb05c: fixed. `src/memory/scoping.test.ts:92` now asserts the literal "__global__", and a new test at line 96 asserts `GLOBAL_APP_ID` itself. Confirmed both fail when scoping.ts is reverted to main (ran revert-test-restore in a worktree).

integrations/pi-agent-plugin/tests/scoping.test.ts
kartik-mem0: Verified against 4e3eb05c: fixed the same way. `tests/scoping.test.ts:47` now asserts the literal string, line 51 asserts `GLOBAL_APP_ID` directly. Confirmed both fail on reverted scoping.ts.

integrations/pi-agent-plugin/tests/tools.test.ts
kartik-mem0: Verified against 4e3eb05c: fixed. `tests/tools.test.ts:57` now asserts .toBe("__global__"). Also added a direct delete_all(global) regression test (lines 60-69), closing the previously-uncovered delete_all half; confirmed it fails on reverted scoping.ts too.

kartik-mem0 (commented):

Verified against `4e3eb05c` (new head): both prior findings are fixed. All 3 assertions now check the literal `"__global__"` instead of the imported `GLOBAL_APP_ID` (which silently resolved to `undefined` pre-fix), and a new `delete_all(scope=global)` test covers the previously-untested blast-radius half.

**Before you merge:** nothing, self-contained.
**Follow-up:** the identical bug is unfixed in `integrations/mem0-plugin/.opencode-plugin/scope.ts` (sibling package), out of scope here.
**Not verified:** nothing beyond this PR's own scope.

Risk: low, test-only delta on top of an already-correct fix Your time: ~5 min

- **Red-green-green (this run):** 31/31 pass at head; reverted `scoping.ts` to `main` -> 7/31 fail, including all 3 previously-false-negative tests AND the new `delete_all` test; restored -> 31/31 pass again.
- **CI:** build/lint/test(20,22)/CI Gate all green. Vercel red (fork deploy perms, non-blocking).
- **tsc --noEmit:** clean.
- **Security:** cleared, pure test-file + comment delta, no new attack surface.
- **Bloat:** proportionate test additions only, no new abstraction.

</details>

PR #6173, opened 2026-07-09
fix(ts-oss): preserve filter entity ids without inference

### LINKED ISSUE #6170: bug(ts-oss): add() with infer:false drops entity ids passed via filters, storing an unretrievable memory
### Component

TypeScript SDK (OSS)

### Description

### Where

```ts
```

```ts
if (!infer) {
    ...
```

### Steps to Reproduce

```ts

```

### Expected Behavior

### Actual Behavior

oldnicke:
## Linked Issue

Closes #6170

## Description

When `Memory.add()` receives entity IDs through `filters`, the IDs scope the operation but were not copied into stored metadata. This was especially visible with `infer: false`: the memory was created without the entity metadata expected by subsequent filtered retrieval.

This change validates and trims `filters.user_id`, `filters.agent_id`, and `filters.run_id`, then stores the normalized IDs in metadata consistently with the existing top-level `userId`, `agentId`, and `runId` options.

Top-level entity ID options continue to take precedence when both forms are provided.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

- [x] Added regression coverage for `user_id`, `agent_id`, and `run_id` supplied through filters
- [x] Verified leading/trailing whitespace is trimmed
- [x] Verified top-level `userId` takes precedence over `filters.user_id`

Commands run locally:

- `pnpm exec prettier --check src/oss/src/memory/index.ts src/oss/tests/memory.add.test.ts` - passed
- `pnpm exec jest src/oss/tests/memory.add.test.ts --runInBand` - passed, 15/15 tests
- `pnpm run build` - passed, including Prettier, CJS/ESM builds, and declaration generation
- `pnpm exec tsc --noEmit` - blocked by existing project-wide Jest global and community LangChain type-resolution errors; no errors referenced the changed files

## Checklist

- [x] My code follows the project style guidelines
- [x] I performed a self-review
- [x] I added tests that prove the fix works
- [x] The targeted tests and package build pass locally
- [x] Documentation is not required for this consistency fix

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -748,6 +748,23 @@ export class Memory {
       metadata.expiration_date = normalizeExpirationDate(config.expirationDate);
     }
 
+    const filterUserId = validateAndTrimEntityId(
+      filters.user_id,
+      "filters.user_id",
+    );
+    const filterAgentId = validateAndTrimEntityId(
+      filters.agent_id,
+      "filters.agent_id",
+    );
+    const filterRunId = validateAndTrimEntityId(
+      filters.run_id,
+      "filters.run_id",
+    );
+
+    if (filterUserId) filters.user_id = metadata.user_id = filterUserId;
+    if (filterAgentId) filters.agent_id = metadata.agent_id = filterAgentId;
+    if (filterRunId) filters.run_id = metadata.run_id = filterRunId;
+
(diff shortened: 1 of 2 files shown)


oldnicke:
Rebased onto the latest `mem0ai/main`, resolved the `Memory.add()` conflict, expanded regression coverage for all three filter entity IDs plus trimming and precedence, and reran the TypeScript package checks. The branch is ready for maintainer review.

PR #6175, opened 2026-07-09
fix(memory): guard remove_code_blocks against None content

### LINKED ISSUE #6150: Procedural memory creation crashes with AttributeError instead of a clear error when the LLM returns no content

Diwak4r:
## Summary
- `remove_code_blocks` called `content.strip()` unconditionally, raising `AttributeError` when `content` is `None`.
- This is reachable via `Memory.add(messages, memory_type="procedural_memory")` when the model returns a safety-refusal (None) response.
- Early-return `None` unchanged when `content is None`.

## Test plan
- [ ] `pytest tests/` passes
- [ ] Manual: `Memory.add([...], memory_type="procedural_memory")` with a None content does not raise

Closes #6150


Diwak4r:
Closing as superseded: I rebased onto the latest main and upstream already handles the None case in remove_code_blocks (it now returns an empty string instead of raising), so this change is now redundant / has no diff against main. Thanks for considering it.

PR #6176, opened 2026-07-09
Update MiniMax default model to MiniMax-M3

octo-patch:
Reason: add target provider/model to existing provider registry

## Summary
- Update the Python MiniMax LLM default model to `MiniMax-M3`.
- Update the TypeScript MiniMax LLM default model to `MiniMax-M3`.
- Refresh MiniMax unit test expectations for the new default model.

## Test plan
- `python3 -m pytest /root/octopatch-4/work/repos/mem0ai_mem0/tests/llms/test_minimax.py` (fails: pytest is not installed in the environment)
- `cd /root/octopatch-4/work/repos/mem0ai_mem0/mem0-ts && npm test -- minimax.test.ts` (fails: jest is not installed in the environment)
- `git diff --check`
- Secret scan on changed diff

### DIFF
--- a/docs/components/llms/models/minimax.mdx
+++ b/docs/components/llms/models/minimax.mdx
@@ -3,7 +3,7 @@ title: MiniMax
 description: "Configure MiniMax as an LLM provider in Mem0 with API key setup and optional custom endpoint configuration."
 ---
 
-To use MiniMax LLM models, you have to set the `MINIMAX_API_KEY` environment variable. You can also optionally set `MINIMAX_API_BASE` if you need to use a different API endpoint (defaults to "https://api.minimax.io/v1").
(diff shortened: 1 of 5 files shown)


octo-patch:
I updated the MiniMax provider documentation to cover the current default, supported endpoint configurations, and both supported model IDs, and added focused regression coverage for explicit model selection. I ran the focused Python and TypeScript tests, the TypeScript build and Prettier checks, Ruff lint, the documentation index check, and the final endpoint URL assertions.

octo-patch:
This change has been recreated from the latest `main` with a clean commit history in https://github.com/mem0ai/mem0/pull/6273. The replacement keeps the same scope and includes verified regional endpoint documentation and focused regression coverage.

PR #6177, opened 2026-07-09
fix(ts-oss): mirror entity ids from filters into metadata for infer:false path

### LINKED ISSUE #6170: bug(ts-oss): add() with infer:false drops entity ids passed via filters, storing an unretrievable memory

nandanadileep:
## Description

When `add()` is called with `infer:false`, entity IDs (`user_id`/`agent_id`/`run_id`) passed via the `filters` option were silently dropped. Only the top-level `userId`/`agentId`/`runId` fields were copied into `metadata`. Since the `infer:false` path in `addToVectorStore()` only passes `metadata` to `createMemory()`, the entity IDs were never persisted, making the memory unretrievable via `getAll()`/`search()`.

The `infer:true` path did not have this problem because it writes `filters.user_id`/etc. directly into the memory payload.

## Fix

After copying top-level entity params into both `filters` and `metadata`, also mirror any entity IDs already present in `filters` (from direct `config.filters`) into `metadata` if not already set. This ensures both input paths produce the same stored payload.

## Testing

All 95 existing tests pass (5 pre-existing FAIL suites are due to missing optional dependency `mysql2/promise`).

Fixes #6170

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -713,6 +713,12 @@ export class Memory {
     if (agentId) filters.agent_id = metadata.agent_id = agentId;
     if (runId) filters.run_id = metadata.run_id = runId;
 
+    // Ensure entity IDs already in filters (passed directly via config.filters)
+    // are also mirrored into metadata so the infer:false path persists them.
+    if (filters.user_id && !metadata.user_id) metadata.user_id = filters.user_id;
+    if (filters.agent_id && !metadata.agent_id) metadata.agent_id = filters.agent_id;
+    if (filters.run_id && !metadata.run_id) metadata.run_id = filters.run_id;
+
     if (!filters.user_id && !filters.agent_id && !filters.run_id) {
       throw new Error(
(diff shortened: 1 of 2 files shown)


Aditya-Tripuraneni (commented):
The fix looks correct, but I feel it should have a regression test because the failure mode is silent data loss in the infer:false path. 

Maybe you can add these? (I think like unit tests would even suffice)

infer:false with filters: { user_id: ... } should store user_id in metadata and make the memory retrievable via getAll() or search().

Then add the same coverage for agent_id and run_id, or parameterize the test so all three entity keys are exercised.

nandanadileep:
Thanks for the review! Added regression tests for all three entity keys (user_id/agent_id/run_id) with infer:false. Each test verifies the entity ID is stored in the payload and retrievable via get(). All 109 tests pass (14 in memory.add, plus 95 from other suites).

PR #6184, opened 2026-07-09
fix(vector-stores): Supports milvus versions below 2.5

### LINKED ISSUE #6183: MilvusException when using Miluvs versions below 2.5
### Component

### Summary

When add first memory and the vector store is Milvus (version below 2.5), it will raise MilvusException.

### Steps to Reproduce

```python
import os
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "milvus",
        "config": {
            "collection_name": "test",
            "embedding_model_dims": 1536,
            "url": "127.0.0.1",
            "token": "8e4b8ca8cf2c67",
            "db_name": "my_database",
        }
    }
}

messages = [
    {"role": "user", "content": "I'm planning to watch a movie tonight. Any recommendations?"},
    {"role": "assistant", "content": "How about thriller movies? They can be quite engaging."},
    {"role": "user", "content": "I’m not a big fan of thriller movies but I love sci-fi movies."},
    {"role": "assistant", "content": "Got it! I'll avoid thriller recommendations and suggest sci-fi movies in the future."}
]
m.add(messages, user_id="alice", metadata={"category": "movies"})
```

### Expected Behavior

Successfully created collection by Milvus vector store

### Actual Behavior

MilvusException: <MilvusException: (code=1100, message=create index on 104 field is not supported: invalid parameter[expected=supported field][actual=create index on 104 field])>

### Environment

- mem0 version: 2.0.11
- Python/Node version: python 3.12
- OS: centos

QunBB:
## Linked Issue

Closes #6183

## Description

https://milvus.io/docs/full_text_search_with_langchain.md
Ignore BM25 schema when creating collection by Milvus (version below 2.5) as vector store.

## Type of Change

- [ ] Documentation update

## Test Coverage

The current testing unit has been met the requirements, and I have passed the testing of Milvus 2.3.

## Checklist

### DIFF
--- a/mem0/vector_stores/milvus.py
+++ b/mem0/vector_stores/milvus.py
@@ -89,39 +89,64 @@ class MilvusDB(VectorStoreBase):
                     "To enable hybrid search, use a fresh collection."
                 )
         else:
-            fields = [
-                FieldSchema(name="id", dtype=DataType.VARCHAR, is_primary=True, max_length=512),
-                FieldSchema(name="vectors", dtype=DataType.FLOAT_VECTOR, dim=vector_size),
-                FieldSchema(name="metadata", dtype=DataType.JSON),
-                # Text field for BM25 full-text search (auto-tokenized by Milvus analyzer)
-                FieldSchema(name="text", dtype=DataType.VARCHAR, max_length=65535, enable_analyzer=True),
-                # Sparse vector field populated automatically by the BM25 function below
-                FieldSchema(name="sparse", dtype=DataType.SPARSE_FLOAT_VECTOR),
-            ]
+            try:
+                #  Milvus 2.5+only supports BM25 mixed search
+                fields = [
+                    FieldSchema(name="id", dtype=DataType.VARCHAR, is_primary=True, max_length=512),
+                    FieldSchema(name="vectors", dtype=DataType.FLOAT_VECTOR, dim=vector_size),
+                    FieldSchema(name="metadata", dtype=DataType.JSON),
+                    # Text field for BM25 full-text search (auto-tokenized by Milvus analyzer)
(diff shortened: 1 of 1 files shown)


kartik-mem0:
First, an apology: you opened this on 9 July and nobody replied for two weeks. That is a failure on our side and not a reflection on the patch.

You diagnosed this correctly, and the approach you picked, attempt the 2.5+ hybrid schema, fall back to a dense-only collection, and record the outcome on `_has_bm25_schema`, is the approach that is going to land. I'm closing this in favour of #6338, which came later with the same design carried further. Rather than wave vaguely at "more complete", here is the concrete delta.

**The update path.** `create_col` is only half of it. `update()` still builds its record unconditionally:

```python
text = ""
if payload:
    text = (payload.get("text_lemmatized") or payload.get("data", ""))[:65535]
schema = {"id": vector_id, "vectors": vector, "metadata": payload, "text": text}
```

The insert path already guards this with `if self._has_bm25_schema:`, `update()` never got the same treatment. So once your fallback produces a dense-only collection, inserts correctly omit `text` while updates keep writing it, which is exactly the inconsistency that flag exists to prevent. #6338 adds the matching guard on both sites.

**The except clause.** `except Exception` also catches transient failures, so a connection blip during `create_col` permanently degrades the collection to dense-only with only a warning to show for it. #6338 narrows it to `except MilvusException` with an error-code check, so only a genuine version rejection triggers the fallback.

It also ships tests around the issue, which is what allowed the fallback to be verified end to end against a real Milvus rather than reasoned about.

None of that changes the fact that you found this first and the shape of your fix was right. If you want to keep going in this area, several of the other vector-store backends have the same "newer server feature assumed unconditionally" shape.

PR #6195, opened 2026-07-09
feat(ts-sdk): add AWS Bedrock LLM provider

### LINKED ISSUE #5765: feat(ts-sdk): add AWS Bedrock LLM provider

ManasVardhan:
## Summary

Closes #5765.

Brings the TypeScript OSS SDK to parity with the Python SDK by adding an **AWS Bedrock** LLM provider.

## Changes

- **`llms/aws_bedrock.ts`**: New `AWSBedrockLLM` implementing the `LLM` interface (`generateResponse` / `generateChat`). Uses `@aws-sdk/client-bedrock-runtime` with the Converse API so a single code path works across Bedrock model families (Anthropic, Amazon Nova, Meta, Mistral, Cohere, etc.). Mirrors the Python provider's behavior:
  - System messages are lifted into the Converse `system` parameter.
  - OpenAI-style tool definitions are converted to `toolConfig.tools[].toolSpec`, and `toolUse` responses are mapped back into the existing `toolCalls` shape.
  - `inferenceConfig` omits `topP` for Anthropic and MiniMax when `temperature` is set (both reject the two together), matching `_build_inference_config` in Python.
  - Defaults match the Python provider (`maxTokens=2000`, `temperature=0.1`).
- **`utils/factory.ts`**: Register the `aws_bedrock` provider (plus a `bedrock` alias).
- **`types/index.ts`**: Add `AWSBedrockConfig` (`awsRegion`, `awsAccessKeyId`, `awsSecretAccessKey`, `awsSessionToken`) and the matching zod schema fields.
- **`package.json`**: Add `@aws-sdk/client-bedrock-runtime` as an optional peer dependency. The SDK is **lazy-loaded**, so users who don't use Bedrock aren't forced to install it (parity with the Vertex AI / Valkey providers).
- **`tests/aws_bedrock.test.ts`**: Unit tests for factory registration, lazy loading (no client construction at init), Converse request formatting, `topP` handling, tool-call mapping, and credential resolution from the environment.
- **docs**: Add a TypeScript usage example to the AWS Bedrock LLM page.

Auth uses the standard AWS credential provider chain (`AWS_REGION`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`), with optional explicit credentials via config.

## Testing

- `pnpm run test:unit` (new `aws_bedrock.test.ts` passes, 6/6)
- `tsc --noEmit` clean
- `prettier --check` clean
- `pnpm install --frozen-lockfile` verified

## Notes

Followed the standalone-provider approach discussed on the issue rather than extending `OpenAILLM`, since Bedrock has provider-specific payloads and Converse-specific formatting. `aws_profile` support (which would require an extra credential-provider package) is intentionally left out in favor of the standard AWS SDK credential chain; happy to add it if maintainers prefer.

### DIFF
--- a/docs/components/llms/models/aws_bedrock.mdx
+++ b/docs/components/llms/models/aws_bedrock.mdx
@@ -5,12 +5,14 @@ description: "Configure AWS Bedrock as an LLM provider in Mem0 with IAM authenti
 
 ### Setup
 - Before using the AWS Bedrock LLM, make sure you have the appropriate model access from [Bedrock Console](https://us-east-1.console.aws.amazon.com/bedrock/home?region=us-east-1#/modelaccess).
-- You will also need to authenticate the `boto3` client by using a method in the [AWS documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials)
+- The Python SDK authenticates the `boto3` client and the TypeScript SDK authenticates the AWS SDK for JavaScript. Both use the standard [AWS credential provider chain](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html).
 - You will have to export `AWS_REGION`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY` to set environment variables.
+- The TypeScript SDK uses `@aws-sdk/client-bedrock-runtime`, which is an optional peer dependency. Install it with `npm install @aws-sdk/client-bedrock-runtime` when using this provider.
 
 ### Usage
 
-```python
+<CodeGroup>
+```python Python
 import os
 from mem0 import Memory
 
@@ -38,6 +40,34 @@ messages = [
 ]
 m.add(messages, user_id="alice", metadata={"category": "movies"})
 ```
+
+```typescript TypeScript
+import { Memory } from 'mem0ai/oss';
+
+// Credentials are read from the standard AWS credential chain
+// (AWS_REGION, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY), or you can
+// pass awsRegion / awsAccessKeyId / awsSecretAccessKey in config.
+const config = {
+  llm: {
+    provider: 'aws_bedrock',
+    config: {
+      model: 'anthropic.claude-3-5-haiku-20241022-v1:0',
+      temperature: 0.2,
+      maxTokens: 2000,
+    },
+  },
+};
+
+const memory = new Memory(config);
+const messages = [
+    {"role": "user", "content": "I'm planning to watch a movie tonight. Any recommendations?"},
+    {"role": "assistant", "content": "How about thriller movies? They can be quite engaging."},
+    {"role": "user", "content": "I’m not a big fan of thriller movies but I love sci-fi movies."},
+    {"role": "assistant", "content": "Got it! I'll avoid thriller recommendations and suggest sci-fi movies in the future."}
+]
+await memory.add(messages, { userId: "alice", metadata: { category: "movies" } });
+```
+</CodeGroup>
 
 ### Config
 
--- a/mem0-ts/package.json
+++ b/mem0-ts/package.json
@@ -109,6 +109,7 @@
   },
   "peerDependencies": {
     "@anthropic-ai/sdk": "^0.40.1",
+    "@aws-sdk/client-bedrock-runtime": "3.967.0",
     "@aws-sdk/client-s3vectors": "3.967.0",
     "@mochow/mochow-sdk-node": "^2.1.5",
     "@azure/identity": "^4.0.0",
@@ -148,6 +149,9 @@
     "@zilliz/milvus2-sdk-node": "^2.4.0 || ^3.0.0"
   },
   "peerDependenciesMeta": {
+    "@aws-sdk/client-bedrock-runtime": {
+      "optional": true
+    },
     "mysql2": {
       "optional": true
     },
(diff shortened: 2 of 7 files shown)


Aditya-Tripuraneni (commented):
## 1. Tool-calling isn't gated by provider (important)

In `aws_bedrock.ts`, `generateResponse` calls `toToolConfig(tools)` and attaches `toolConfig` to the Converse request whenever `tools` is non-empty, regardless of `this.provider`:

```ts
const toolConfig = this.toToolConfig(tools);
...
if (toolConfig) {
  input.toolConfig = toolConfig;
}
```

The Python provider doesn't do this. It gates on an explicit allowlist before ever building `toolConfig`:

```python
# mem0/llms/aws_bedrock.py
self.supports_tools = self.provider in ["anthropic", "cohere", "amazon"]
...
if tools and self.supports_tools:
    return self._generate_with_tools(messages, tools, stream)
else:
    return self._generate_standard(messages, stream)  # tools silently dropped
```

So today, if a caller passes `tools` with a model from an unsupported family (Meta, Mistral, DeepSeek, etc.), the TS provider will send `toolConfig` to Bedrock and for models that don't support Converse tool use, that will surface as an AWS `ValidationException` at request time. Python instead silently falls back to a plain-text `_generate_standard` call with no `toolConfig` at all, so the same `tools` argument never reaches Bedrock for those providers.

Given the PR description calls out that tool-call mapping "mirrors" the Python provider, could you add the same `provider in ["anthropic", "cohere", "amazon"]` gate before building `toolConfig` (or drop tools entirely) so behavior matches for unsupported model families? Happy to be wrong if Converse handles this gracefully for all model families now, just didn't see anything to that effect.

---

## 2. `extractProvider` has no validation (important nit)

Python validates the extracted provider segment against a known `PROVIDERS` list and raises `ValueError` on an unrecognized model id:

```python
PROVIDERS = [
    "ai21", "amazon", "anthropic", "cohere", "meta", "mistral", "stability", "writer",
    "deepseek", "gpt-oss", "perplexity", "snowflake", "titan", "command", "j2", "llama",
    "minimax",
]

def extract_provider(model: str) -> str:
    for provider in PROVIDERS:
        if re.search(rf"\b{re.escape(provider)}\b", model):
            return provider
    raise ValueError(f"Unknown provider in model: {model}")
```

The TS version just takes `parts[0]` (or `parts[1]` after stripping a region prefix) with no check against a known list: 

```ts
function extractProvider(model: string): string {
  const parts = model.split(".");
  if (parts.length >= 2 && REGION_PREFIXES.has(parts[0])) {
    return parts[1];
  }
  return parts[0];
}
```

So a malformed or unexpected model id silently becomes `this.provider` instead of failing fast. Today this only affects the topP/temperature exclusivity decision, but it's a quieter failure mode than Python's which I think is worth aligning behavior (validate against a known provider list and throw, matching Python) rather than silently accepting anything. (I would also add tests to validate correctness if you chose to address this)

---

ManasVardhan:
Thanks for the review. Both points addressed in ef9dcf6:

**Tool-calling gated by provider.** Added a `SUPPORTS_TOOLS_PROVIDERS` allowlist (`anthropic`, `cohere`, `amazon`) and a `supportsTools` flag set in the constructor. `generateResponse` now only builds/attaches `toolConfig` when the provider supports it; for other families (Meta, Mistral, DeepSeek, etc.) tools are silently dropped rather than sending a `toolConfig` that Bedrock rejects with a ValidationException. This matches the Python provider's `supports_tools` fallback to the standard path.

**Provider validation.** Rewrote `extractProvider` to validate against the known `PROVIDERS` list using word-boundary matching (mirroring the Python `re.search(rf"\b{provider}\b", model)`) and throw on an unrecognized model id, instead of silently guessing from the first dot-segment. Region-prefixed ids like `us.anthropic.claude-...` still resolve correctly.

Added tests for: tools dropped for unsupported providers, tools kept for supported providers, region-prefixed id resolution, and the unknown-provider error. Full suite (10 tests) passes and prettier/tsc are clean on the changed files.

kartik-mem0:
closing this pr in favour of #5890

PR #6196, opened 2026-07-09
fix(mem0-plugin): make Codex lifecycle hooks work on native Windows (#6181)

### LINKED ISSUE #6181: [windows] Codex lifecycle hooks still fail on v0.2.12 — POSIX commands exit 1 on every event (regression of #5243)
## Summary

On native Windows Codex (CLI/Desktop), the Mem0 plugin's Codex lifecycle hooks fail on **every** event with `exit code 1`, flooding the session with errors (one failure per tool call via `PostToolUse`).

This is a continuation/regression of #5243, which was closed as fixed by #5257 (v0.2.6). The original "Open With" popup symptom is gone, but the hooks still **do not execute on Windows** — they now hard-fail instead of popping dialogs. Still broken on **v0.2.12**.

## Environment

- Windows 11
- Codex (native, PowerShell — **not** WSL)
- Mem0 plugin `mem0@mem0-plugins` **v0.2.12**

## Root cause

`hooks/codex-hooks.json` registers commands in POSIX-shell form, e.g.:

```
MEM0_PLATFORM=codex ${PLUGIN_ROOT}/scripts/on_session_start.sh
```

Two Windows-fatal assumptions in that one line:

1. `VAR=value command` env-prefix is **bash syntax** — the native Windows shell (cmd/PowerShell) can't parse it.
2. `.sh` scripts aren't directly executable on Windows without an explicit `bash`.

Codex hands that string to the native shell, which exits 1 immediately. Because `PostToolUse` fires after every tool call, it becomes a flood rather than a one-off.

## Evidence it's the invocation, not the scripts

The scripts themselves run cleanly under Git Bash:

```console
$ bash ~/.codex/plugins/cache/mem0-plugins/mem0/0.2.12/scripts/on_session_start.sh
## Mem0 Active
...
$ echo $?
0
```

So the `.sh` logic is fine — only the way `codex-hooks.json` **invokes** it breaks on Windows.

## Observed

Every `SessionStart` / `UserPromptSubmit` / `PreToolUse` / `PostToolUse` / `Stop` hook logs `hook exited with code 1`; `PostToolUse` repeats once per tool call, e.g. a wall of:

```
• PostToolUse hook (failed)
  error: hook exited with code 1
```

## Complication for self-fixing

Users can't cleanly patch `codex-hooks.json` because Codex pins each hook with a `trusted_hash` in `~/.codex/config.toml`. Editing the commands invalidates that trust (re-prompt/refuse), and the file lives in a plugin cache that's clobbered on update.

## Suggested fix

Make `scripts/install_codex_hooks.py` (and the shipped `codex-hooks.json`) OS-aware:

- On Windows, wrap through bash explicitly — e.g. `bash "${PLUGIN_ROOT}/scripts/on_X.sh"`, moving `MEM0_PLATFORM` into the script/env rather than a shell prefix; **or**
- Ship a small cross-platform launcher (a `.py`/`.cmd` shim that sets env and execs the script) so the same command works under cmd/PowerShell/bash; **or**
- If Windows lifecycle hooks won't be supported, have the installer **detect Windows and skip hook registration** (MCP-only), with a clear note — rather than registering hooks that always fail.

## Workaround for other Windows users hitting this

Neutralize the hooks — back up and replace `hooks/codex-hooks.json` content with `{"hooks": {}}`. The Mem0 MCP tools still work; you just lose automatic capture.

Refs: #5243, #5257.

rajarshidattapy:
## Linked Issue

Closes #6181

## Description

On native Windows, Codex's lifecycle hooks fail with exit code 1 on every event (SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop), flooding the session with errors. This is a regression of #5243, which #5257 only partially fixed.

**Root cause:** `hooks/codex-hooks.json` registers commands as `MEM0_PLATFORM=codex ${PLUGIN_ROOT}/scripts/on_X.sh` — a POSIX env-prefix that only bash can parse. Codex hands that string to the native shell (cmd/PowerShell on Windows), which tries to execute `MEM0_PLATFORM=codex` as a program name and fails immediately. #5257 only added a Windows guard to the legacy manual installer (`install_codex_hooks.py`), but `.codex-plugin/plugin.json` wires hooks directly from `hooks/codex-hooks.json` too — that path was never guarded and kept shipping the broken syntax.

**Fix:**
1. `hooks/codex-hooks.json` — wrapped every hook command in `bash -c "MEM0_PLATFORM=codex '.../script.sh'"`. This shifts all env-var parsing into bash itself; the native shell only ever sees `bash` + one quoted argument, which is valid invocation syntax under cmd, PowerShell, and POSIX shells alike (requires `bash` on PATH — Git Bash/WSL).
2. `scripts/install_codex_hooks.py` — replaced the hard Windows block with a `shutil.which("bash")` check, since the commands now work fine on Windows as long as bash is resolvable.
3. Bumped plugin version 0.2.12 → 0.2.13 across the manifests.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `tests/test_codex_hooks_json.py` as a regression guard against reintroducing bare env-prefix syntax in `codex-hooks.json`.

Manually verified by resolving `${PLUGIN_ROOT}` in the updated command and running it through PowerShell: exits 0 and produces the expected output, versus exit 1 with the old form.

Ran the full `mem0-plugin` test suite: 157 passed, 1 pre-existing/unrelated failure (`test_rubric_dedup.py::test_first_prompt_gets_full_rubric`, confirmed failing on `main` before this change too).

## Checklist

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "mem0",
       "source": "./integrations/mem0-plugin",
       "description": "Mem0 memory layer for AI applications. Add persistent memory, personalization, and semantic search to Claude workflows.",
-      "version": "0.2.12"
+      "version": "0.2.13"
     }
   ]
 }
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "mem0",
       "source": "./integrations/mem0-plugin",
       "description": "Mem0 memory layer for AI applications. Add persistent memory, personalization, and semantic search to Claude workflows.",
-      "version": "0.2.13"
+      "version": "0.2.12"
     }
   ]
 }
--- a/.cursor-plugin/marketplace.json
+++ b/.cursor-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "mem0",
       "source": "./integrations/mem0-plugin",
       "description": "Mem0 memory layer for AI applications. Add persistent memory, personalization, and semantic search.",
-      "version": "0.2.12"
+      "version": "0.2.13"
     }
   ]
 }
--- a/.cursor-plugin/marketplace.json
+++ b/.cursor-plugin/marketplace.json
@@ -12,7 +12,7 @@
       "name": "mem0",
       "source": "./integrations/mem0-plugin",
       "description": "Mem0 memory layer for AI applications. Add persistent memory, personalization, and semantic search.",
-      "version": "0.2.13"
+      "version": "0.2.12"
     }
   ]
 }
--- a/integrations/mem0-plugin/.claude-plugin/plugin.json
+++ b/integrations/mem0-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "mem0",
-  "version": "0.2.12",
+  "version": "0.2.13",
   "description": "Persistent memory for Claude Code. Remembers decisions, patterns, and preferences across sessions.",
   "author": {
     "name": "Mem0",
--- a/integrations/mem0-plugin/.claude-plugin/plugin.json
+++ b/integrations/mem0-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "mem0",
-  "version": "0.2.13",
+  "version": "0.2.12",
   "description": "Persistent memory for Claude Code. Remembers decisions, patterns, and preferences across sessions.",
   "author": {
     "name": "Mem0",
--- a/integrations/mem0-plugin/.codex-plugin/plugin.json
+++ b/integrations/mem0-plugin/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "mem0",
-  "version": "0.2.12",
+  "version": "0.2.13",
   "description": "Persistent memory for Codex. Remembers decisions, patterns, and preferences across sessions.",
   "author": {
     "name": "Mem0",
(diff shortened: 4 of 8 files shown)


zzmark:
On Windows, Codex supports a platform-specific hook field: command_windows.
When present, Codex prioritizes command_windows on Windows and executes it via cmd.

So instead of relying on a POSIX-style command (e.g. VAR=value ...sh) for all platforms, a better fix is to define a separate Windows command in codex-hooks.json using command_windows, and keep the existing command for Unix-like systems.

This would avoid bash-only syntax on native Windows

rajarshidattapy:
@kartik-mem0 can you review this PR once, this has been a genuine issue I'm trying to solve.

chiveswang:
Independent confirmation on Windows 11 + Codex CLI v0.144.1 + mem0 plugin v0.2.13.

Applied this PR's `command_windows` strings verbatim to the installed `hooks/codex-hooks.json`. Before: `hook: SessionStart Failed` on every lifecycle event. After: `hook: SessionStart Completed`, and `~/.mem0/hooks.log` shows a real auto-import run — confirming the hook body actually executed, not just exited 0. `${PLUGIN_ROOT}` expands correctly inside the `command_windows` value.

One caveat for anyone hand-patching before this merges: Codex persists a per-hook trust hash in `~/.codex/config.toml` under `[hooks.state]`. Editing the hooks file invalidates it and Codex then **silently skips** the hook — no `Failed`, no warning. So "the Failed messages went away" is not evidence the fix worked; verify with a side effect (log line / probe file). Re-approve via the interactive TUI, or use `--dangerously-bypass-hook-trust` for a one-off test.

The fix works as written — this has been blocking Windows users since v0.2.6.

rajarshidattapy:

PR #6197, opened 2026-07-10
fix(mem0): add test for issue #6085

### LINKED ISSUE #6085: OSS default config silently breaks add(): model=None resolves to gpt-5-mini, which rejects the default temperature=0.1

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6085.

## Why
Fixes #6085

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6085.py
+++ b/tests/test_issue_6085.py
@@ -1,0 +1,36 @@
+import sys
+from pathlib import Path
+from types import ModuleType
+from unittest.mock import Mock, patch
+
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6201, opened 2026-07-10
fix(mem0): add test for issue #6013

### LINKED ISSUE #6013: bug(ts-oss/redis): empty/all-null filters build an invalid query ("" instead of "*")

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6013.

## Why
Fixes #6013

## Testing
- [x] New test added


kartik-mem0:

kartik-mem0:

PR #6202, opened 2026-07-10
fix(ts-oss): persist entity ids from filters in infer:false add()

### LINKED ISSUE #6170: bug(ts-oss): add() with infer:false drops entity ids passed via filters, storing an unretrievable memory

abhay-codes07:
## Linked Issue

Closes #6170

## Description

`add()` with `infer: false` only folded the session scope into the stored payload when it came from the `userId`/`agentId`/`runId` options:

```ts
```

When the same scope was passed via the `filters` option (a public field on `AddMemoryOptions`), it never reached `metadata`, and the `infer: false` branch of `addToVectorStore()` persists only `metadata`. The memory was therefore stored with no `user_id`/`agent_id`/`run_id` and could never be returned by a scoped `getAll`/`search`, even though `add()` reported success.

The `infer: true` path doesn't have this problem — it writes `filters.user_id`/`agent_id`/`run_id` into the payload directly. This change makes the `infer: false` branch fold the same values into the payload before calling `createMemory()`, so both paths agree. It also matches the Python SDK, whose `infer=false` path stores the session scope in the memory payload.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added a regression test in `memory.add.test.ts` that stores a memory with `{ filters: { user_id }, infer: false }` and asserts a scoped `getAll` returns it. It fails on the current code (0 results — the memory is orphaned) and passes with the fix. The full `add()` suite passes locally.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -767,6 +767,14 @@ export class Memory {
   ): Promise<MemoryItem[]> {
     if (!infer) {
       const returnedMemories: MemoryItem[] = [];
+      // Fold the session scope into the stored payload so infer:false memories
+      // remain retrievable via scoped getAll/search. This mirrors the infer:true
+      // path (which writes filters.user_id/agent_id/run_id into memPayload) and
+      // covers scope passed via the `filters` option, not just userId/agentId/runId.
+      const baseMetadata: Record<string, any> = { ...metadata };
+      if (filters.user_id) baseMetadata.user_id = filters.user_id;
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss fix with a regression test. infer:false was dropping entity ids passed via filters, leaving the memory unretrievable. Quick look when you get a chance, thanks.

PR #6205, opened 2026-07-10
fix(ts-oss): honor response_format in Anthropic generateResponse

### LINKED ISSUE #6203: bug: Anthropic generateResponse silently ignores response_format (present in both SDKs — TS #6205, Python #5820)
The OSS TypeScript Anthropic provider accepts a `responseFormat` argument but never acts on it, so requesting structured/JSON output is silently dropped. This is the TypeScript counterpart of the Python bug fixed in #5820.

### Where
`mem0-ts/src/oss/src/llms/anthropic.ts`

`generateResponse(messages, responseFormat?: { type: string }, tools?)` accepts `responseFormat` (L34) but it is never referenced in the body — `params` is built and `this.client.messages.create(params)` is called with no handling for it.

### Expected (mirror the Python fix in #5820)
- `json_object` → assistant prefill (`{`) + system instruction, then reconstruct via `extractJson("{" + text)`.
- `json_schema` → `output_config.format` (Anthropic structured-outputs API).
- Short-circuit when `responseFormat` is undefined so the existing path is untouched.

### Note
The simultaneous `responseFormat` + `tools` limitation flagged on #5820 applies here too — worth a comment/guard in the TS fix.

Follow-up to #5820.

ac12644:
## Linked Issue

Closes #6203

## Description

The OSS TypeScript Anthropic provider (`mem0-ts/src/oss/src/llms/anthropic.ts`) accepted a `responseFormat` argument but never acted on it, so requesting JSON output was silently ignored. This is the TypeScript counterpart of the Python fix in #5820, and the follow-up @kartik-mem0 requested in that review.

Anthropic has no native `response_format`, so this translates the shared interface the same way the Python provider does:

- **`json_object`** → append a JSON-only system instruction and prefill the assistant turn with `{`, then reconstruct the object via `extractJson("{" + text)`.
- **`json_schema`** → map to `output_config.format` (Anthropic structured-outputs API).
- Short-circuits when `responseFormat` is undefined, so the existing path is untouched.
- **`response_format` is ignored when `tools` are active** (guarded + commented) — tool_use handling takes precedence and the JSON prefill would otherwise drop tool_use blocks. This addresses the minor point kartik flagged on #5820.

## Type of Change

## Test Coverage

Added three tests to `anthropic-llm.test.ts` (mocked SDK): json_object prefill + reconstruction, json_schema → `output_config.format`, and response_format ignored when tools are present. All 14 tests in the file pass; `anthropic.ts` typechecks clean; prettier clean.

Also verified against the **live Anthropic API**: with `response_format={"type":"json_object"}` the provider returns directly `JSON.parse`-able output (e.g. `{"facts":["the sky is blue"]}`), whereas the same prompt with no format comes back ```-fenced (` ```json ... ``` `) — the exact behavior this fix removes.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/llms/anthropic.ts
+++ b/mem0-ts/src/oss/src/llms/anthropic.ts
@@ -2,6 +2,7 @@ import type Anthropic from "@anthropic-ai/sdk";
 import { LLM, LLMResponse } from "./base";
 import { LLMConfig, Message } from "../types";
 import { loadPeer } from "../utils/load_peer";
+import { extractJson } from "../prompts";
 
 export class AnthropicLLM implements LLM {
   private client!: Anthropic;
@@ -43,7 +44,7 @@ export class AnthropicLLM implements LLM {
 
   async generateResponse(
     messages: Message[],
-    responseFormat?: { type: string },
+    responseFormat?: { type: string; json_schema?: any },
     tools?: any[],
   ): Promise<string | LLMResponse> {
     await this.ensureClient();
@@ -80,6 +81,37 @@ export class AnthropicLLM implements LLM {
       params.tool_choice = { type: "auto" };
     }
 
+    // Anthropic has no native response_format param, so translate the shared
+    // interface here (parity with the Python provider, #5820 / #6203).
+    // response_format is intentionally ignored when tools are active: the
+    // tool_use handling below takes precedence, and the JSON prefill would
+    // otherwise drop tool_use blocks.
+    let wantsJsonPrefill = false;
+    if (responseFormat && !tools) {
+      if (responseFormat.type === "json_schema" && responseFormat.json_schema) {
+        let schema = responseFormat.json_schema;
+        if (schema && typeof schema === "object" && "schema" in schema) {
+          schema = schema.schema;
+        }
+        // output_config is newer than the pinned @anthropic-ai/sdk types.
+        (params as unknown as { output_config: unknown }).output_config = {
+          format: { type: "json_schema", schema },
+        };
+      } else if (responseFormat.type === "json_object") {
+        wantsJsonPrefill = true;
+        params.system =
+          (params.system ?? "") +
+          "\n\nYou must respond with valid JSON only. Do not include any " +
+          "other text, markdown formatting, or code fences.";
+        // Prefill the assistant turn with "{" so the model continues a JSON
+        // object; the leading brace is added back when reconstructing below.
+        params.messages = [
+          ...params.messages,
+          { role: "assistant", content: "{" },
+        ];
+      }
+    }
+
     const response = await this.client.messages.create(params);
 
     if (tools) {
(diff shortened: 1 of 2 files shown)


ac12644:
Rebased onto latest `main` — the one conflict (an import) resolved by keeping both `loadPeer` and the new `extractJson`; tests green (`anthropic-llm` 14/14). Now mergeable.

Thanks @kartik-mem0 for the end-to-end trace and the sign-off on the approach — ready for review when you have a moment. (Companion Python fix for the same gap is #5820, also rebased + green.)

ac12644:
Rebased onto main - #6506 landed on `anthropic.ts` while this was sitting and conflicted with the prefill branch.

Flagging how I resolved it, since taking both sides verbatim would have shipped a bug: the json_object path here was doing `response.content[0]`, which is exactly the thing #6506 removed. With thinking enabled the first block is a thinking block, so the prefill path would have found no text and quietly returned garbage. Pulled #6506's block-finding into a small `firstTextBlock()` helper and both paths use it now. Also made the no-text-block case return `""` rather than a bare `"{"`, which no parser can do anything with.

Two tests added for that: json_object with a leading thinking block, and json_object with no text block at all. Suite is 18/18.

mem0-ts/src/oss/src/llms/anthropic.ts
```diff
@@ -80,6 +81,37 @@ export class AnthropicLLM implements LLM {
...
+    if (responseFormat && !tools) {
+      if (responseFormat.type === "json_schema" && responseFormat.json_schema) {
+        let schema = responseFormat.json_schema;
+        if (schema && typeof schema === "object" && "schema" in schema) {
+          schema = schema.schema;
+        }
+        // output_config is newer than the pinned @anthropic-ai/sdk types.
+        (params as unknown as { output_config: unknown }).output_config = {
+          format: { type: "json_schema", schema },
+        };
```
kartik-mem0, line 99: No caller in this repo passes `{type: "json_schema", ...}` (the only real call site, `memory/index.ts:893`, always sends `json_object`), and no sibling TS provider implements a `json_schema` response_format type either. Speculative for what #6203 needs. Consider dropping it and shipping `json_object` only, or splitting it into a follow-up PR if you want to keep it.

kartik-mem0 (commented):

`response_format` was accepted as a parameter and never read; this reads it and honors `json_object` (assistant `"{"` prefill + JSON system instruction + reconstruction) and `json_schema` (`output_config.format`). The PR was rebased onto `0e582adc` (#6506) mid-review and now reuses its `firstTextBlock()` fix for the new prefill path too, so `mergeable` is `MERGEABLE`, not the `CONFLICTING` this was filed under; nothing further to rebase.

**Before you merge:** nothing, this is self-contained. One Minor note inline: the `json_schema` branch has no current caller in this repo.
**Follow-up:** Python has the same bug, tracked separately at #5820 (same author). #6374 (Python, draft, test-only) does not overlap this PR, verified via its diff.
**Not verified:** live Anthropic API, mocked SDK client only for both `json_object` and `json_schema` paths.

Risk: low, provider-local change with a real red-green proof · Your time: ~5 min

- **Problem is real:** `mem0-ts/src/oss/src/llms/anthropic.ts` (pre-PR): `responseFormat` param never referenced in the body. Confirmed by direct read and by the red run below.
- **Tests:** `npx jest src/oss/tests/anthropic-llm.test.ts` -> 18 passed, 18 total at head `0880a38e`.
- **Red-green:** reverted impl to `0e582adc` (post-#6506, pre-#6205) -> exactly the 3 new `response_format` tests fail with on-point errors (no prefill message appended, JSON parse fails on raw text, `output_config` undefined); the other 15 pass. Restored -> 18/18 green again. Worktree clean before/after.
- **CI:** all required checks green (`CI Gate`, `TypeScript SDK` build/integration x2, `changelog_check`, `check_changes`, `label`, `license/cla`). Only red check is `Vercel` (deploy-auth, not merge-blocking).
- **Security:** cleared, no credentials/eval/SSRF, no new user-controlled sink.
- **Bloat (ponytail):** `net: -8 lines possible` - the `json_schema`/`output_config` branch (anthropic.ts:91-99) has zero callers in this repo.
- **Linked issue:** #6203

</details>

ac12644:
@kartik-mem0 gentle nudge on this one; it and the Python twin #5820 are both green and you'd marked them LGTM. A formal approve would let them merge whenever you get a sec. Thanks!

PR #6206, opened 2026-07-10
fix(core): coerce non-string entity ids instead of crashing on .strip()

### LINKED ISSUE #6204: bug(core): non-string entity id (e.g. integer user_id) crashes with AttributeError in _validate_and_trim_entity_id
### Component

Core / Python SDK

### Description

Passing a non-string entity id (e.g. an integer `user_id`, a very common pattern when the id is a database primary key) crashes with a confusing internal `AttributeError` instead of working or raising a clear error.

`_validate_and_trim_entity_id` in `mem0/memory/main.py` assumes the value is always a `str` and calls `value.strip()`:

```python
def _validate_and_trim_entity_id(value: Optional[str], name: str) -> Optional[str]:
    if value is None:
        return None
    trimmed = value.strip()   # AttributeError if value is an int
    ...
```

Every public entry point funnels entity ids through this function, so all of these crash:

```python
Memory.add("hi", user_id=42)
Memory.search("q", filters={"user_id": 42})
Memory.get_all(filters={"user_id": 42})
Memory.delete_all(user_id=42)
```

(sync and async).

### Steps to Reproduce

```python
from mem0 import Memory
m = Memory()
m.add("I like coffee", user_id=42)
```

### Actual Behavior

```
AttributeError: 'int' object has no attribute 'strip'
```

raised deep in the call stack, giving no hint about what the caller did wrong.

### Expected Behavior

Integer (and other non-string) ids should be handled cleanly. The most user-friendly option is to coerce them to `str` at the single validation point, so scoping stays self-consistent across `add`/`search`/`get_all`/`delete_all` (all of which go through the same validator). Alternatively, raise a clear `ValueError` naming the offending parameter. Either is far better than the current `AttributeError`.

abhay-codes07:
## Linked Issue

Closes #6204

## Description

`_validate_and_trim_entity_id` assumed the entity id was always a `str` and called `value.strip()`:

```python
if value is None:
    return None
trimmed = value.strip()   # AttributeError if value is an int
```

Passing a non-string id (e.g. an integer `user_id`, which is common when the id is a database primary key) crashed with `AttributeError: 'int' object has no attribute 'strip'` deep in the call stack, with no hint about the cause. Every public entry point routes entity ids through this one function, so `add`, `search`, `get_all`, and `delete_all` (sync and async) were all affected.

This coerces non-string values to `str` at the single validation point, so the id "just works" and scoping stays consistent across every entry point (they all normalize the same way). The alternative would be to raise a clear `ValueError`; coercion was chosen because integer ids are a common, reasonable input and nothing currently succeeds with them. Happy to switch to an explicit error if you'd prefer that.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added tests under `TestEntityIdValidation` covering the helper directly (`42 -> "42"`, `0 -> "0"`) and the `delete_all`/`get_all` integer-id paths (asserting the store is scoped by the str form). They fail on the current code with the `AttributeError` and pass with the fix.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -145,9 +145,10 @@ def _reject_top_level_entity_params(kwargs: Dict[str, Any], method_name: str) ->
         )
 
 
-def _validate_and_trim_entity_id(value: Optional[str], name: str) -> Optional[str]:
+def _validate_and_trim_entity_id(value: Optional[Any], name: str) -> Optional[str]:
     """
     Validates and normalizes an entity ID.
+    - Coerces non-string values (e.g. integer ids) to str
     - Trims leading/trailing whitespace
     - Rejects empty or whitespace-only strings
     - Rejects strings containing internal whitespace
@@ -164,6 +165,11 @@ def _validate_and_trim_entity_id(value: Optional[str], name: str) -> Optional[st
     """
     if value is None:
         return None
+    # Callers commonly pass integer ids (e.g. a database primary key). Coerce
+    # to str at this single validation point so scoping stays consistent across
+    # add/search/get_all/delete_all instead of crashing on `.strip()`.
+    if not isinstance(value, str):
+        value = str(value)
     trimmed = value.strip()
     if trimmed == "":
         raise ValueError(
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -4,7 +4,7 @@ from unittest.mock import Mock, patch
 import pytest
 
 from mem0.configs.base import MemoryConfig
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@kartik-mem0 small core fix with tests. integer entity ids (e.g. user_id=42) were crashing with AttributeError in the id validator. Coerces to str at the single validation point. Note in the description on the coerce-vs-raise choice. Thanks.

kartik-mem0 (approved):

Issue #4194, opened 2026-03-02
Docker Compose env vars (LLM_PROVIDER, OLLAMA_BASE_URL, etc.) are ignored


haabe:
### 🐛 Describe the bug

The config is stored in a SQLite DB (openmemory.db), not read from environment variables. The .env file only matters for the initial OPENAI_API_KEY startup check.

Issue #3238, opened 2025-07-29
Openmemory config issues


YourSandwich:
### 🐛 Describe the bug

I am fighting now for multiple hours with setting the config for open memory.
```
curl -X 'PUT' \
  'http://10.1.0.51:8890/api/v1/config/openmemory' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
    "openmemory": {
        "custom_instructions": "string"
    },
    "mem0": {
        "vector_store": {
            "provider": "qdrant",
            "config": {
                "collection_name": "matrix",
                "host": "localhost",
                "port": 6333,
                "embedding_model_dims": 1024
            }
        },
        "graph_store": {
            "provider": "neo4j",
            "config": {
                "url": "neo4j+s://localhost",
                "username": "neo4j",
                "password": "asdasdad"
            }
        },
        "llm": {
            "provider": "openai",
            "config": {
                "model": "qwen3-14B",
                "temperature": 0.1,
                "max_tokens": 2048,
                "openai_base_url":"http://10.0.0.210:1234/v1",
                "api_key": "sk-local"
            }
        },
        "embedder": {
            "provider": "openai",
            "config": {
                "model": "embedding-bge-m3",
                "openai_base_url":"http://10.0.0.210:1234/v1",
                "api_key": "sk-local",
                "embedding_dims": 1024
            }
        }
    }
}'
```

This does not work when setting in settings.json or config.json in the api directory, even when overwritng the get_default_config function in router/config.

It seems the API only accepts 
```
{
  "openmemory": {
    "custom_instructions": "string"
  },
  "mem0": {
    "llm": {
      "provider": "openai",
      "config": {
        "model": "openai",
        "temperature": 0,
        "max_tokens": 0,
        "api_key": "sk-local",
        "ollama_base_url": "http://10.0.0.210:1234/v1"
      }
    },
    "embedder": {
      "provider": "openai",
      "config": {
        "model": "embedding-bge-m3",
        "api_key": "sk-local",
        "ollama_base_url": "http://10.0.0.210:1234/v1"
      }
    }
  }
}
```

When trying to save or list memories like this i get:
```
Jul 29 07:15:53 arr sh[4111887]: No configuration found in database, using defaults
Jul 29 07:15:53 arr sh[4111887]: Parsing environment variables in final config...
Jul 29 07:15:53 arr sh[4111887]: Loaded OPENAI_API_KEY from environment for api_key
[...]

PR #6207, opened 2026-07-10
fix(openmemory): honor LLM/EMBEDDER env vars in the default config and stop dropping openai_base_url

gavdevs:
## Problem

Deployments that configure OpenMemory through the documented `LLM_*` / `EMBEDDER_*` environment variables (any OpenAI-compatible endpoint — Ollama's OpenAI mode, vLLM, LM Studio, OpenRouter, hosted gateways) get silently broken by the config API:

1. **`get_default_configuration()` ignores environment configuration.** Every `GET /api/v1/config/*` endpoint get-or-creates the persistent `main` config row using hardcoded defaults (`gpt-4o-mini` / `text-embedding-3-small` / `env:OPENAI_API_KEY`), and `get_memory_client()` gives the DB row precedence over env config. So merely opening the dashboard settings page permanently switches a working env-configured deployment onto the hardcoded OpenAI defaults — every subsequent memory add then fails (e.g. 401s, because a non-OpenAI key gets sent to `api.openai.com`).

2. **The config schemas cannot express OpenAI-compatible endpoints.** `LLMConfig` / `EmbedderConfig` accept `ollama_base_url` but not `openai_base_url`, even though `app/utils/memory.py` writes that field and mem0's own OpenAI LLM/embedder configs support it. Any `PUT /api/v1/config/...` round-trip silently strips it, so a custom base URL can't be persisted through the API — and a dashboard save destroys one written by other means.

**Reproduction:** deploy with `LLM_PROVIDER`/`LLM_MODEL`/`LLM_BASE_URL` etc. set (adds work) → `curl` any `GET /api/v1/config/` endpoint or open dashboard settings → adds now fail; inspect the `configs` DB row to see the hardcoded defaults that were seeded.

## Fix

- `get_default_configuration()` now builds its defaults from the same environment variables `get_default_memory_config()` documents, storing **`env:` references** (`env:LLM_API_KEY`, `env:LLM_BASE_URL`, …) so no secret material is ever persisted in the database. With none of these variables set, the output is identical to the previous hardcoded defaults — no behavior change for plain-OpenAI deployments.
- Added `openai_base_url` to `LLMConfig` and `EmbedderConfig` so the field round-trips through GET/PUT.

## Tests

New `tests/test_config_router.py`: legacy-default equivalence with no env vars, env-var honoring (OpenAI-compatible and Ollama), secrets-stay-out-of-the-DB, and schema round-trip of `openai_base_url`. All 5 pass; the 3 pre-existing failures in `test_mcp_server.py::TestRouteRegistration` are unrelated and fail identically on `main`.

Addresses the failure mode reported in #4194 and part of #3238.

### DIFF
--- a/openmemory/api/app/routers/config.py
+++ b/openmemory/api/app/routers/config.py
@@ -1,3 +1,4 @@
+import os
 from typing import Any, Dict, Optional
 
 from app.database import get_db
@@ -14,6 +15,7 @@ class LLMConfig(BaseModel):
     temperature: float = Field(..., description="Temperature setting for the model")
     max_tokens: int = Field(..., description="Maximum tokens to generate")
     api_key: Optional[str] = Field(None, description="API key or 'env:API_KEY' to use environment variable")
+    openai_base_url: Optional[str] = Field(None, description="Base URL for OpenAI-compatible APIs (e.g., http://localhost:8080/v1)")
     ollama_base_url: Optional[str] = Field(None, description="Base URL for Ollama server (e.g., http://host.docker.internal:11434)")
 
 class LLMProvider(BaseModel):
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Thanks for the contribution. The OpenMemory project has been deprecated and is no longer maintained. We're sorry about this, we appreciate the effort you put in and we should have communicated the deprecation earlier. Closing this PR.

PR #6209, opened 2026-07-10
fix(core): don't require langchain-core for default async procedural memory

### LINKED ISSUE #6208: bug(core): async procedural memory hard-requires langchain-core on the default path (sync does not)
### Component

Core / Python SDK

### Description

`AsyncMemory._create_procedural_memory` imports `langchain-core` at the top of the method and re-raises if it is missing, even on the default code path that never uses it. So async procedural memory is broken on a base install, while the sync path works fine.

`langchain-core` is an optional dependency (listed under `[extras]` in `pyproject.toml`), and it is only needed to adapt messages for a **custom** LangChain LLM. The default path (`llm=None`) uses `self.llm.generate_response` and never calls `convert_to_messages`, yet the unconditional import makes the whole call fail without the optional package:

```python
async def _create_procedural_memory(self, messages, metadata=None, llm=None, prompt=None):
    try:
        from langchain_core.messages.utils import convert_to_messages  # type: ignore
    except Exception:
        logger.error("Import error while loading langchain-core. ...")
        raise
    ...
    if llm is not None:
        parsed_messages = convert_to_messages(parsed_messages)
        ...
    else:
        procedural_memory = await asyncio.to_thread(self.llm.generate_response, messages=parsed_messages)
```

The **sync** `Memory._create_procedural_memory` has no such import and works without `langchain-core`, so this is an async/sync asymmetry.

### Steps to Reproduce

On an install without `langchain-core`:

```python
import asyncio
from mem0 import AsyncMemory

m = AsyncMemory()
asyncio.run(m.add(
    [{"role": "user", "content": "always deploy via the release script"}],
    agent_id="agent_1",
    memory_type="procedural_memory",
))
```

### Actual Behavior

```
ModuleNotFoundError: No module named 'langchain_core'
```

even though the default path does not use langchain-core.

### Expected Behavior

The default (`llm=None`) path works without `langchain-core`, matching the sync implementation. `langchain-core` is imported only when a custom LangChain `llm` is passed, with a clear error if it is missing in that case.

abhay-codes07:
## Linked Issue

Closes #6208

## Description

`AsyncMemory._create_procedural_memory` imported `langchain-core` at the top of the method and re-raised if it was missing, even though the default (`llm=None`) path never uses it. Async procedural memory therefore failed on a base install, while the sync `Memory._create_procedural_memory` works fine without the package — an async/sync asymmetry.

`langchain-core` is optional (`[extras]` in `pyproject.toml`) and is only needed to adapt messages for a custom LangChain `llm` via `convert_to_messages`. This moves the import into the `llm is not None` branch where it is actually used, and raises a clear `ImportError` if the package is missing in that specific case. The default path now runs without `langchain-core`, matching the sync implementation.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `test_async_procedural_memory_default_path_without_langchain`, which runs the default async path with `langchain-core` simulated absent (patched out of `sys.modules`) and asserts the memory is created via `self.llm`. It fails on the current code with `ModuleNotFoundError` and passes with the fix. The existing custom-LLM test (`test_async_procedural_memory_langchain_strips_code_blocks`) is unaffected — that path still imports and uses `convert_to_messages`.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3563,16 +3563,6 @@ class AsyncMemory(MemoryBase):
             llm (llm, optional): LLM to use for the procedural memory creation. Defaults to None.
             prompt (str, optional): Prompt to use for the procedural memory creation. Defaults to None.
         """
-        try:
-            from langchain_core.messages.utils import (
-                convert_to_messages,  # type: ignore
-            )
-        except Exception:
-            logger.error(
-                "Import error while loading langchain-core. Please install 'langchain-core' to use procedural memory."
-            )
-            raise
-
         logger.info("Creating procedural memory")
 
         parsed_messages = [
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small core fix with a regression test. async procedural memory was requiring langchain-core even on the default llm=None path, which the sync version doesn't. Moved the import into the branch that actually uses it. Thanks.

kartik-mem0 (approved):

PR #6219, opened 2026-07-10
fix: Feature: portable session export/import on the core Memory SDK

### LINKED ISSUE #6214: Feature: portable session export/import on the core Memory SDK
**Component:** Core / Python SDK

### Use Case

I want to move a working session between agents/machines. If I build up memory in one mem0 instance (say Claude Code on my laptop), there's no way to take that session elsewhere - a teammate's machine, another agent like Cursor, or an archive - and pick up where I left off. `get_all` reads memories but there's no import to re-hydrate them, and the platform `create_memory_export` is hosted-only, schema-reshaped, and has no import.

OpenMemory's backup router does export/import, but it's tied to the OpenMemory stack (FastAPI + Postgres + its models). Using the core SDK directly (`from mem0 import Memory`), I have nothing.

### Proposed Solution

Two methods on `Memory` / `AsyncMemory` that round-trip a session (scoped by the usual filters) as a portable JSON bundle:

```python
m.export_session(filters={"run_id": "sess-42"}, path="sess-42.json")

m2.import_session(path="sess-42.json",
                  remap_scope={"agent_id": "cursor", "run_id": "sess-42-cont"})
```

- Round-trippable (raw content + metadata + timestamps), re-embedded on import with the target's embedder, no LLM re-extraction.
- `on_conflict` = skip | overwrite | new_ids for merging into an existing store.
- Optional `include_history` / `include_messages` to also carry change history and recent session messages.
- No new dependencies (stdlib json/gzip + configured embedder + vector store). Sync + async.

Happy to open a PR. It can land memories-only first, with history/messages as follow-ups if you'd prefer a smaller first cut.

### Alternatives Considered

- OpenMemory backup router - requires the OpenMemory stack; core-SDK users don't have it.
- Platform `create_memory_export` - hosted-only, schema-reshaped (lossy), no import, 7-day expiry.
- `get_all` + manual re-`add` - re-runs LLM extraction, so it doesn't round-trip faithfully.

Would you be open to this on the core SDK? If so I'll send a PR.

Phoenix1454:
Fixes the issue described in #6214.

Fixes #6214

### DIFF
--- a/examples/misc/multillm_memory.py
+++ b/examples/misc/multillm_memory.py
@@ -15,7 +15,7 @@ import logging
 from dotenv import load_dotenv
 from litellm import completion
 
-from mem0 import MemoryClient
+from mem0 import Memory, AsyncMemory
 
 load_dotenv()
 
@@ -29,7 +29,9 @@ logger = logging.getLogger(__name__)
 
 
 # Initialize memory client (platform version)
-memory = MemoryClient()
+memory = Memory()
+async_memory = AsyncMemory()
+
 
 # Research team models with specialized roles
 RESEARCH_TEAM = {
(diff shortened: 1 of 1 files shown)


kartik-mem0:
Thanks for taking this on — the round-trip shape here is genuinely sensible: raw memory content + metadata + timestamps out, re-`add` on the way back in rather than re-running extraction. That's the right instinct for a portable session bundle, and `export_session`/`import_session` as a demo of that pattern against the existing public API (`memory.search`, `memory.add`) is solid example code.

There's a mechanical problem with how this is wired up, though, separate from the code itself: the PR description carries the literal closing keyword `Fixes #6214`. Issue #6214 asks for export/import as methods on the core `Memory`/`AsyncMemory` SDK classes (`m.export_session(...)`, `m2.import_session(...)`) — not helpers in an example script. As it stands, this diff touches only `examples/misc/multillm_memory.py`: `export_session` (`examples/misc/multillm_memory.py:135`) and `import_session` (`examples/misc/multillm_memory.py:149`) are plain module-level functions in that file, not methods on `Memory` or `AsyncMemory` (`mem0/memory/main.py:450` and `:2112` — neither class has an `export_session`/`import_session` method today). If this merges as-is, GitHub will auto-close #6214 on the closing keyword, but the SDK won't actually have gained the capability the issue asked for.

Two ways to resolve that — either is fine with me:

1. **Keep this as an example, drop the auto-close.** Change `Fixes #6214` to `Refs #6214` in the description, and reframe the PR honestly as a recipe demonstrating export/import against the current public API. That's a legitimate, mergeable contribution on its own terms — it just shouldn't close the issue, since it doesn't add what the issue is asking for.
2. **Promote it to the real feature.** Implement `export`/`import` as methods on both `Memory` and `AsyncMemory` in `mem0/memory/main.py` (sync and async kept in lockstep, no drift between them), with unit tests. Once the methods exist on the SDK classes, the closing keyword is earned and this PR can close #6214 for real.

Worth knowing before picking path 2: the original poster on #6214 has already offered to implement this themselves and sketched a fairly complete design there (round-trip via raw content + metadata, `on_conflict` modes for merge behavior, scope remap on import). Might be worth a quick coordination comment on the issue thread first so the work doesn't happen twice.

Either direction works from here — just flagging the mismatch now rather than after merge, since the auto-close is easy to miss.

PR #6220, opened 2026-07-10
fix(ts-oss): mirror entity ids from filters into metadata for infer:false (salvage of #6177)

Bartok9:
## Summary
Salvage of #6177 by @nandanadileep, rebased onto current `main`.

When `add()` is called with `infer: false`, entity IDs only present on `filters` were not mirrored into `metadata` and were never persisted, making memories unretrievable via `getAll()`/`search()`.

Also retains the existing main-tip expiration-date metadata normalization next to this fix.

## Credit
- Original work: #6177 (@nandanadileep)

## Motivation
Closes the gap described in #6170 / #6177 without leaving a CONFLICTING PR stranded.

## Verification
- Cherry-picked both original commits; resolved conflict with modern expiration metadata path.
- Diff limited to ts-oss memory path + regression tests from upstream PR.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -741,6 +741,12 @@ export class Memory {
       metadata.expiration_date = normalizeExpirationDate(config.expirationDate);
     }
 
+    // Ensure entity IDs already in filters (passed directly via config.filters)
+    // are also mirrored into metadata so the infer:false path persists them.
+    if (filters.user_id && !metadata.user_id) metadata.user_id = filters.user_id;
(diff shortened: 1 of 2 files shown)


Bartok9:
Closing as superseded on current `main`.

`mem0-ts/src/oss/src/memory/index.ts` already mirrors filter entity ids into metadata unconditionally:

```ts
if (filters.user_id) metadata.user_id = filters.user_id;
if (filters.agent_id) metadata.agent_id = filters.agent_id;
if (filters.run_id) metadata.run_id = filters.run_id;
```

That covers the `infer:false` path this salvage targeted (and is stronger than only filling when metadata lacks the key). Main also has `preserves infer=%s %s filters scope after sanitization` coverage. No residual seam.

PR #6223, opened 2026-07-11
fix(mem0): add test for issue #6003

### LINKED ISSUE #6003: @mem0/opencode-plugin has multiple OpenCode Desktop issues: plugin load failure, API key resolution failure, and always-wrong app_id

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6003.

## Why
Fixes #6003

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6003.py
+++ b/tests/test_issue_6003.py
@@ -1,0 +1,3 @@
+### Evidence from the published bundle
+
+The published bundle contains code like:


kartik-mem0:

kartik-mem0:

**Finding:** `tests/test_issue_6003.py` is not Python. Its contents are Markdown copied from the issue body:

```
### Evidence from the published bundle

The published bundle contains code like:
```

pytest fails at collection with a `SyntaxError`.

---

Closing this one. To be clear, this is about the test file, not about issue #6003 — the underlying report is legitimate and stays as it is.

PR #6226, opened 2026-07-11
fix(core): drop null entity ids from filters before scope validation

### LINKED ISSUE #6225: get_all()/search() mishandle a None entity id in filters (crash or wrong scope)
### Component

Core / Python SDK

### Summary

`get_all()` and `search()` accept a `filters` dict whose entity ids are `None` and do the wrong thing with it. If you build filters from optional params (a common pattern), a `None` id is not ignored the way `add()` ignores it. Two things go wrong:

- `filters={"user_id": None, "agent_id": "a1"}` passes the "at least one entity id" check (it checks whether the *key* is present, not whether the value is usable), then the `None` id flows into the query. Depending on the vector store this scopes the query to `user_id IS NULL` and returns nothing even though a valid `agent_id` was given.
- Before it even reaches the store it can crash: `process_telemetry_filters` calls `filters["user_id"].encode()`, which raises `AttributeError: 'NoneType' object has no attribute 'encode'`.

`add()` doesn't have this problem because it builds its filters with a truthiness check and simply drops the `None` ids.

### Steps to Reproduce

```python
from mem0 import Memory

m = Memory()
# agent_id is a real scope, user_id just happens to be None here
m.get_all(filters={"user_id": None, "agent_id": "a1"})
```

### Expected Behavior

The `None` `user_id` is ignored and the call is scoped to `agent_id="a1"` (matching how `add()` treats `None` ids). If every entity id is `None`, raise the normal "filters must contain at least one of: user_id, agent_id, run_id" error.

### Actual Behavior

`AttributeError: 'NoneType' object has no attribute 'encode'` (or, if telemetry is skipped, a query scoped to `user_id IS NULL` that returns nothing). Same behavior on `search()` and on both `Memory` and `AsyncMemory`.

### Environment

- mem0 version: main (latest)
- Python version: 3.11
- OS: Windows / Linux

abhay-codes07:
## Linked Issue

Closes #6225

## Description

`get_all()` and `search()` validate each entity id in `filters` with `_validate_and_trim_entity_id`, which returns `None` (no error) when the input is `None` and leaves the key in the dict. The "at least one entity id" guard then checks whether the *key* is present, not whether the value is usable, so a filters dict such as `{"user_id": None, "agent_id": "a1"}` passes validation and the stray `None` id flows downstream. It either scopes the query to `user_id IS NULL` (returning nothing despite a valid `agent_id`) or crashes in `process_telemetry_filters`, which calls `.encode()` on the `None`.

This differs from `add()`, which builds its filters with a truthiness check and already ignores `None` ids. Building filters from optional params and leaving some as `None` is a common caller pattern, so the read path should treat them the same way.

The fix adds a small `_drop_null_entity_ids` helper and calls it right before the existing guard, on `get_all` and `search` for both `Memory` and `AsyncMemory`. A dict with no usable scope now raises the normal `ValueError`, and a `None` id sitting next to a real one is simply ignored.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added four tests to `TestEntityIdValidation` in `tests/test_main.py` covering `get_all` and `search`: an all-`None` filters dict now raises the standard "at least one of" error, and a `None` id next to a real id is dropped so the query is scoped by the remaining id. All four fail on `main` (AttributeError / wrong scope) and pass with this change. Full `tests/test_main.py` passes (47) and `ruff check` is clean.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -145,6 +145,22 @@ def _reject_top_level_entity_params(kwargs: Dict[str, Any], method_name: str) ->
         )
 
 
+def _drop_null_entity_ids(filters: Dict[str, Any]) -> None:
+    """Remove entity id keys whose value is None from a filters dict (in place).
+
+    A caller often builds filters from optional params, e.g.
+    ``filters={"agent_id": agent_id, "user_id": user_id}`` where some ids are
+    None. Left in place, a None id scopes the query to ``<key> IS NULL`` and
+    silently returns nothing, even when another id in the same dict is a valid
+    scope. Dropping the None keys mirrors add()'s truthiness handling and lets
+    the "at least one entity id" guard reject a filters dict that has no usable
+    scope instead of running an unintended query.
+    """
+    for key in ENTITY_PARAMS:
+        if key in filters and filters[key] is None:
+            del filters[key]
+
+
 def _validate_and_trim_entity_id(value: Optional[Any], name: str) -> Optional[str]:
     """
     Validates and normalizes an entity ID.
@@ -1258,6 +1274,7 @@ class Memory(MemoryBase):
             )
 
         # Validate filters contains at least one entity ID
+        _drop_null_entity_ids(effective_filters)
         if not any(key in effective_filters for key in ("user_id", "agent_id", "run_id")):
             raise ValueError(
                 "filters must contain at least one of: user_id, agent_id, run_id. "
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small core fix with tests. get_all/search leave a None entity id in filters, which either crashes in process_telemetry_filters or scopes the query to user_id IS NULL and returns nothing even when another id is valid. Drops null ids before the scope guard so it matches add(). Quick look when you can, thanks.

PR #6229, opened 2026-07-11
fix(client): send page/page_size as query params when only one is provided

### LINKED ISSUE #6227: MemoryClient.get_all() ignores page_size unless page is also passed
### Component

Core / Python SDK

### Summary

`MemoryClient.get_all()` (and the async version) only move `page`/`page_size` into the request query string when **both** are supplied. If you pass only `page_size` and let `page` default to 1 (a very common call), the value stays in the JSON body, which the server does not read for pagination. The result is that `page_size` is silently ignored and you get a default-sized page back.

The code clearly intends pagination to travel as query params (it does exactly that when both are present), so sending it in the body when only one is given is inconsistent with its own contract.

### Steps to Reproduce

```python
from mem0 import MemoryClient

client = MemoryClient(api_key="m0-...")
# Expecting 5 results per page, page defaults to 1
client.get_all(filters={"user_id": "alice"}, page_size=5)
```

The outgoing request posts `{"filters": {...}, "page_size": 5}` as the JSON body with no query string, so the server applies its default page size.

### Expected Behavior

`page_size=5` (or `page=2` on its own) is sent as a query parameter and honored, the same as when both are supplied together.

### Actual Behavior

With only one of `page`/`page_size`, the value is placed in the JSON body and ignored; the response uses the server default. Same on `AsyncMemoryClient.get_all()`.

### Environment

- mem0 version: main (latest)
- Python version: 3.11
- OS: Windows / Linux

abhay-codes07:
## Linked Issue

Closes #6227

## Description

`MemoryClient.get_all()` and `AsyncMemoryClient.get_all()` only lifted pagination into the query string when both `page` and `page_size` were present:

```python
if "page" in params and "page_size" in params:
    query_params = {"page": params.pop("page"), "page_size": params.pop("page_size")}
    response = self.client.post("/v3/memories/", json=params, params=query_params)
else:
    response = self.client.post("/v3/memories/", json=params)
```

A common call passes only `page_size` and relies on `page` defaulting to 1 (or passes only `page`). That falls into the `else` branch, so the value is sent in the JSON body instead. The server reads pagination from the query string, so it is silently ignored and the caller gets a default-sized page. The code already treats pagination as query params when both are present, so the single-param case is inconsistent with its own contract.

This lifts `page` and `page_size` out of the body independently, so either one on its own reaches the query string. When neither is present the request is sent without a `params` kwarg, exactly as before. Applies to both the sync and async clients.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `TestGetAllPagination` in `tests/test_client.py`: page_size-only, page-only, both, and no-pagination for the sync client plus a page_size-only case for the async client. The single-param cases fail on `main` (value ends up in the body / no query params) and pass with this change; the both-present and no-pagination cases are unchanged. Full `tests/test_client.py` passes (34) and `ruff check` is clean.

## Checklist

### DIFF
--- a/mem0/client/main.py
+++ b/mem0/client/main.py
@@ -269,11 +269,15 @@ class MemoryClient:
         kwargs = {**(options.model_dump(exclude_unset=True) if options else {}), **kwargs}
         params = self._prepare_params(kwargs)
 
-        if "page" in params and "page_size" in params:
-            query_params = {
-                "page": params.pop("page"),
-                "page_size": params.pop("page_size"),
-            }
+        # Pagination is read from the query string by the server, so lift page
+        # and page_size out of the JSON body independently. Requiring both to be
+        # present (a caller may pass only page_size and rely on page defaulting
+        # to 1) left the provided value in the body, where it was ignored.
+        query_params = {}
+        for key in ("page", "page_size"):
+            if key in params:
+                query_params[key] = params.pop(key)
+        if query_params:
             response = self.client.post("/v3/memories/", json=params, params=query_params)
         else:
             response = self.client.post("/v3/memories/", json=params)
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small client fix with tests. get_all only moved page/page_size to the query string when both were present, so passing just page_size (page defaulting to 1) left it in the body and the server ignored it. Lifts each independently. Quick review when you get a chance, thanks.

PR #6231, opened 2026-07-11
fix(xai): forward http_client_proxies instead of the built client in config conversion

### LINKED ISSUE #6230: XAILLM forwards the built http client as http_client_proxies, crashing when a proxy is set
### Component

Core / Python SDK

### Summary

When `XAILLM` is built from a plain `BaseLlmConfig` (rather than an `XAIConfig`), it converts the config so `xai_base_url` is available. That conversion forwards `config.http_client` into the `http_client_proxies` argument:

```python
config = XAIConfig(
    ...
    http_client_proxies=config.http_client,   # <-- built httpx.Client, not the proxy setting
)
```

`config.http_client` is the `httpx.Client` that `BaseLlmConfig` already built from the proxy setting, while `config.http_client_proxies` is the raw proxy dict/str. `XAIConfig` then runs `build_http_client()` on whatever it receives, which expects a proxy dict/str. Given a `Client` object it falls through to `httpx.Client(proxy=<client>)` and crashes.

Every other provider (openai, anthropic, azure_openai, deepseek, lmstudio, minimax, ollama, vllm) forwards `config.http_client_proxies` in the same conversion. xAI is the only one forwarding the built client.

### Steps to Reproduce

```python

base = BaseLlmConfig(model="grok-4", api_key="sk-...", http_client_proxies="http://proxy.local:8080")
XAILLM(base)   # crashes while building the proxy client
```

### Expected Behavior

The proxy setting is preserved and a working client is built, the same as for the other providers (and the same as when xAI is used through `LlmFactory`, which converts the config correctly).

### Actual Behavior

Construction crashes inside `build_http_client`/httpx because a `Client` object is passed where a proxy dict/str is expected. Only happens when a proxy is configured; without one both values are `None` and nothing breaks.

### Environment

- mem0 version: main (latest)
- Python version: 3.11
- OS: Windows / Linux

abhay-codes07:
## Linked Issue

Closes #6230

## Description

When `XAILLM` is constructed with a plain `BaseLlmConfig` (not an `XAIConfig`), it converts the config so `xai_base_url` is available. That conversion forwarded `config.http_client` into the `http_client_proxies` argument:

```python
config = XAIConfig(
    ...
)
```

`config.http_client` is the `httpx.Client` that `BaseLlmConfig` already built from the proxy setting, whereas `config.http_client_proxies` is the raw proxy dict/str. `XAIConfig` then runs `build_http_client()` on it, which expects a proxy dict/str; given a `Client` object it falls through to `httpx.Client(proxy=<client>)` and crashes. This only bites when a proxy is configured (otherwise both values are `None`).

Every other provider (openai, anthropic, azure_openai, deepseek, lmstudio, minimax, ollama, vllm) forwards `config.http_client_proxies` in this conversion. xAI was the only outlier. Going through `LlmFactory` already converts the config correctly, so this specifically affects direct `XAILLM(base_config)` construction.

The fix forwards `config.http_client_proxies`, matching the other providers.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `test_xai_llm_preserves_http_client_proxies_from_base_config` to `tests/test_http_client_proxies.py`, alongside the existing openai equivalent: it builds `XAILLM` from a `BaseLlmConfig` carrying a proxy and asserts the proxy survives into `XAIConfig`. It crashes on `main` and passes with this change. Full `tests/test_http_client_proxies.py` passes (8) and `ruff check` is clean.

## Checklist

### DIFF
--- a/mem0/llms/xai.py
+++ b/mem0/llms/xai.py
@@ -28,7 +28,7 @@ class XAILLM(LLMBase):
                 top_k=config.top_k,
                 enable_vision=config.enable_vision,
                 vision_details=config.vision_details,
-                http_client_proxies=config.http_client,
+                http_client_proxies=config.http_client_proxies,
             )
 
         super().__init__(config)
--- a/tests/test_http_client_proxies.py
+++ b/tests/test_http_client_proxies.py
@@ -37,3 +37,22 @@ def test_llm_factory_preserves_http_client_proxies():
     llm = LlmFactory.create("openai", base)
     assert llm.config.http_client_proxies == "http://proxy.local:8080"
     assert isinstance(llm.config.http_client, httpx.Client)
+
+
+def test_xai_llm_preserves_http_client_proxies_from_base_config():
+    """XAILLM converts a BaseLlmConfig to XAIConfig directly; the proxy setting must survive.
+
+    Every other provider forwards config.http_client_proxies in this conversion. xAI
+    forwarded the already-built config.http_client instead, which build_http_client then
+    tried to treat as a proxy, crashing whenever a proxy was configured.
+    """
+    from mem0.llms.xai import XAILLM
+
+    base = BaseLlmConfig(
+        model="grok-4",
+        api_key="sk-test",
+        http_client_proxies="http://proxy.local:8080",
+    )
+    llm = XAILLM(base)
+    assert llm.config.http_client_proxies == "http://proxy.local:8080"
+    assert isinstance(llm.config.http_client, httpx.Client)


abhay-codes07:
@kartik-mem0 small provider fix with a test. XAILLM's BaseLlmConfig to XAIConfig conversion forwarded the already-built http client into http_client_proxies, so it crashed whenever a proxy was set. Every other provider forwards http_client_proxies here; this makes xAI match. Quick look when you can, thanks.

PR #6232, opened 2026-07-11
feat: add Amazon Bedrock Knowledge Base vector store backend

PVidyadhar:
# feat: add Amazon Bedrock Knowledge Base vector store backend

## Linked Issue

N/A — new feature adding Amazon Bedrock Knowledge Base as a vector store backend.

## Description

Adds `BedrockKB(VectorStoreBase)` that uses Amazon Bedrock Managed Knowledge Bases as a full read-write vector store backend for mem0.

- `search()` uses `managedSearchConfiguration` for retrieval
- `insert()`/`update()` upload documents to S3 + trigger ingestion sync
- `delete()` removes from S3 + re-syncs the data source
- `list_cols()` lists available knowledge bases
- `col_info()` returns KB name and status
- Agentic retrieval with fallback to standard Retrieve API

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Manual E2E: `search()` returned 3 results. `list_cols()` returned 17 KBs. `col_info()` returned name + ACTIVE status. Full vector store backend interface verified against live KB + bedrock-agent API in us-west-2.

## Checklist

### DIFF
--- a/mem0/utils/factory.py
+++ b/mem0/utils/factory.py
@@ -185,6 +185,7 @@ class VectorStoreFactory:
         "azure_ai_search": "mem0.vector_stores.azure_ai_search.AzureAISearch",
         "azure_mysql": "mem0.vector_stores.azure_mysql.AzureMySQL",
         "pinecone": "mem0.vector_stores.pinecone.PineconeDB",
+        "bedrock_kb": "mem0.vector_stores.bedrock_kb.BedrockKB",
         "mongodb": "mem0.vector_stores.mongodb.MongoDB",
         "redis": "mem0.vector_stores.redis.RedisDB",
         "valkey": "mem0.vector_stores.valkey.ValkeyDB",
--- a/mem0/vector_stores/BEDROCK_MANAGED_KB.md
+++ b/mem0/vector_stores/BEDROCK_MANAGED_KB.md
@@ -1,0 +1,105 @@
+# Bedrock Managed Knowledge Base Support
+
+## Overview
+Adds a Mem0 vector store backend that delegates search to Amazon Bedrock Knowledge Bases instead of a local vector store.
+
+## Usage
+```python
+from mem0 import Memory
+
+config = {
+    "vector_store": {
+        "provider": "bedrock_kb",
+        "config": {
+            "knowledge_base_id": "YOUR_KB_ID",
+            "region": "us-east-1",
+        }
+    }
+}
+m = Memory.from_config(config)
+results = m.search("What are the onboarding steps?", user_id="user1")
+```
+
+## Configuration
+| Variable | Description | Default |
+|---|---|---|
+| KNOWLEDGE_BASE_ID | Bedrock Knowledge Base ID | None |
+| AWS_REGION | AWS region for the KB | us-east-1 |
+| AWS_PROFILE | AWS credentials profile | None |
+| USE_AGENTIC_RETRIEVAL | Enable agentic retrieval | true |
+| MAX_RESULTS | Maximum retrieval results | 5 |
+
+## Features
+- Managed search (no vector store needed)
+- Agentic retrieval with query decomposition + reranking
+- Automatic fallback to plain Retrieve if agentic fails
+- Multi-source support (S3, Web, Confluence, SharePoint)
+- Implements Mem0 VectorStoreBase interface
+
+## SDK Requirements
+- boto3 >= 1.43
+- mem0ai >= 0.1
+
+## Required IAM Permissions
+```json
+{
+  "Effect": "Allow",
+  "Action": [
+    "bedrock:Retrieve",
+    "bedrock:AgenticRetrieveStream"
+  ],
+  "Resource": "arn:aws:bedrock:<region>:<account-id>:knowledge-base/<kb-id>"
+}
+```
+
+## References
+- [Build a Managed Knowledge Base](https://docs.aws.amazon.com/bedrock/latest/userguide/kb-build-managed.html)
+- [Retrieve API](https://docs.aws.amazon.com/bedrock/latest/userguide/kb-test-retrieve.html)
+- [Agentic Retrieval](https://docs.aws.amazon.com/bedrock/latest/userguide/kb-test-agentic.html)
+
+## Direct Ingestion (CUSTOM Data Source)
+
+When using a CUSTOM data source (created via console or API), you can ingest documents directly without S3 upload + sync:
+
+```python
+kb = BedrockKB(
+    knowledge_base_id="YOUR_KB_ID",
+    data_source_id="YOUR_CUSTOM_DS_ID",
+    region_name="us-west-2",
+    data_source_type="CUSTOM",
+)
+
+# Inline text
+kb.insert(vectors=None, payloads=[{"data": "Your document content here."}])
+
+# S3 reference (ingest a specific file without full sync)
+kb.insert(vectors=None, payloads=[{"s3_uri": "s3://bucket/path/to/file.pdf"}])
+
+# Binary file (PDF, DOCX, images, audio, video — depends on KB indexing settings)
+import base64
+with open("document.pdf", "rb") as f:
+    encoded = base64.b64encode(f.read()).decode()
+kb.insert(vectors=None, payloads=[{"data": encoded, "mime_type": "application/pdf"}])
+```
+
+### Supported mime types
+
+- Text: `text/plain`, `text/html`, `text/csv`
+- Documents: `application/pdf`, `application/msword`, `application/vnd.openxmlformats-officedocument.wordprocessingml.document`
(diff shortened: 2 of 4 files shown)


faithpricejp-source (commented):
Thanks for the substantial work here — tests and a setup doc included is more than most new-backend PRs bring. I went through the diff against how `Memory` core actually consumes a vector store, and I think there are four blocking contract issues worth fixing before this can work end-to-end, plus some follow-ups. Line references are to `mem0/memory/main.py` on current `main`.

## Blocking

**1. Return types: core consumes `OutputData` objects via attribute access, not dicts**

`search()`, `list()`, and `get()` here return plain dicts, but every consumer does attribute access — e.g. `main.py:895-896`:

```python
uuid_mapping[str(idx)] = mem.id
existing_memories.append({"id": str(idx), "text": mem.payload.get("data", "")})
```

First `add()` that reaches the dedup step will raise `AttributeError: 'dict' object has no attribute 'id'`. Other backends return a pydantic `OutputData(id, score, payload)` — this one needs the same. (The PR's tests pass because they assert on the store's own return values and never push results through the core pipeline.)

**2. `list()` must return the nested `[results]` shape**

`main.py:1875`: `self.vector_store.list(filters=filters)[0]` — the contract is `List[List[OutputData]]` so callers can index `[0]` (same contract the OpenSearch tests pin explicitly). The flat list returned here means `[0]` yields the first document instead of the result list.

**3. `filters` are silently ignored in `search()` and `list()` — cross-user leakage and destructive `delete_all`**

Core passes `filters={"user_id": ...}` for tenant scoping on every call. With `search()` ignoring them, one user's search returns other users' memories. Worse: `delete_all()` enumerates via `list(filters=...)` and deletes each returned id — since `list()` here returns everything under the `mem0/` prefix regardless of filters, **`delete_all()` for one user deletes every user's documents**. If Bedrock KB metadata filtering can't express this, the payload/user scoping needs to be encoded in the S3 key layout (e.g. `mem0/{user_id}/{doc_id}.txt`) and applied in both `search` (metadata filter) and `list` (prefix).

**4. `id` doesn't round-trip between search and get/delete/update**

`search()` returns `id = source URI` (e.g. `s3://bucket/mem0/x.txt`), but `get()/delete()/update()` build the key as `f"mem0/{vector_id}.txt"` — feeding a search-result id back produces `mem0/s3://bucket/mem0/x.txt.txt`. The core's dedup flow does exactly that round-trip (`mem.id` from search feeds update/delete events), so updates/deletes would target non-existent keys. The document id (the `{doc_id}` minted at insert) should be carried in KB metadata and returned as `id`, with the URI kept in the payload.

## Follow-ups (non-blocking but worth tracking)

5. **Silent truncation in `list()`**: `MaxKeys=top_k or 100` caps enumeration at 100 (and `list_objects_v2` needs continuation-token pagination past 1000). `get_all()/delete_all()` call `list(top_k=None)` expecting everything — same failure class as #6108, which just got fixed for OpenSearch.
6. **Read-after-write assumption**: inserts land via S3 upload + async ingestion job (minutes), but `add()`'s dedup searches immediately after prior writes. This backend structurally can't satisfy that without documentation of the consistency window. Also, triggering `start_ingestion` per insert/delete/update will hit Bedrock's concurrent-ingestion-job `ConflictException`, which is currently swallowed by `logger.error` — the sync silently never happens. Consider debouncing/batching the sync and surfacing conflicts.
7. **Not wired into the factory/config**: there's no `bedrock_kb` registration in `mem0/utils/factory.py` or the vector-store config schema, so the backend is unreachable via `Memory.from_config()` — only by direct import. Needs the provider registration + a config class to be usable.
8. Minor: `update()` drops the `Metadata` that `insert()` writes; `reset()` deletes objects one-by-one without pagination (could use `delete_objects` batches); the `KNOWLEDGE_BASE_ID` env var name is generic enough to collide with other tooling — a `MEM0_`/`BEDROCK_` prefix would be safer.

Happy to re-review after updates — items 1, 2 and 4 are mechanical; item 3 probably needs a maintainer call on the S3 key layout.

faithpricejp-source (commented):
Re-reviewed after the latest force-pushes — good progress, three of the four blocking items are resolved:

- **#1 return types** ✅ — `search`/`get`/`list` now return `OutputData` objects, so the core's attribute access (`mem.id`, `mem.payload`) works.
- **#2 `list()` shape** ✅ — now returns the nested `[results]` (and `[[]]` when empty), so callers' `[0]` indexing is correct.
- **#4 id round-trip** ✅ — `search()` now derives `doc_id` back from the source URI, so a search-result id feeds cleanly into `get`/`delete`/`update`. (Slightly fragile — it assumes the `mem0/{doc_id}.txt` layout and a `.txt` suffix — but it does round-trip now.)

**Blocking #3 is still open, and it's the one that causes data loss.**

`search()` and `list()` accept a `filters` argument but never use it — `list()` still does `Prefix="mem0/"` and returns every document regardless of `user_id`. Two consequences remain:

- **Cross-tenant search leakage**: core passes `filters={"user_id": ...}` on every `search`; ignoring it means user A's search returns user B's memories.
- **Destructive `delete_all`**: core's `delete_all(user_id=A)` enumerates via `list(filters=...)` and deletes each returned id. Since `list()` ignores the filter and returns everything under `mem0/`, `delete_all` for a single user deletes **every** user's documents.

This is the item I flagged as needing a maintainer call on the S3 key layout — encoding the tenant in the key (e.g. `mem0/{user_id}/{doc_id}.txt`), then applying it as a prefix in `list` and a metadata filter in `search`. As-is, merging this backend would introduce cross-tenant data loss for any multi-user deployment.

One more that blocks usability (was follow-up #7): the backend still isn't registered in `mem0/utils/factory.py` / the vector-store config schema, so it's only reachable by direct import, not `Memory.from_config()`.

Happy to re-review once #3 has a direction.

PVidyadhar:
> Re-reviewed after the latest force-pushes — good progress, three of the four blocking items are resolved:
> 
> * **#1 return types** ✅ — `search`/`get`/`list` now return `OutputData` objects, so the core's attribute access (`mem.id`, `mem.payload`) works.
> * **#2 `list()` shape** ✅ — now returns the nested `[results]` (and `[[]]` when empty), so callers' `[0]` indexing is correct.
> * **#4 id round-trip** ✅ — `search()` now derives `doc_id` back from the source URI, so a search-result id feeds cleanly into `get`/`delete`/`update`. (Slightly fragile — it assumes the `mem0/{doc_id}.txt` layout and a `.txt` suffix — but it does round-trip now.)
> 
> **Blocking #3 is still open, and it's the one that causes data loss.**
> 
> `search()` and `list()` accept a `filters` argument but never use it — `list()` still does `Prefix="mem0/"` and returns every document regardless of `user_id`. Two consequences remain:
> 
> * **Cross-tenant search leakage**: core passes `filters={"user_id": ...}` on every `search`; ignoring it means user A's search returns user B's memories.
> * **Destructive `delete_all`**: core's `delete_all(user_id=A)` enumerates via `list(filters=...)` and deletes each returned id. Since `list()` ignores the filter and returns everything under `mem0/`, `delete_all` for a single user deletes **every** user's documents.
> 
> This is the item I flagged as needing a maintainer call on the S3 key layout — encoding the tenant in the key (e.g. `mem0/{user_id}/{doc_id}.txt`), then applying it as a prefix in `list` and a metadata filter in `search`. As-is, merging this backend would introduce cross-tenant data loss for any multi-user deployment.
> 
> One more that blocks usability (was follow-up #7): the backend still isn't registered in `mem0/utils/factory.py` / the vector-store config schema, so it's only reachable by direct import, not `Memory.from_config()`.
> 
> Happy to re-review once #3 has a direction.


Thanks @faithpricejp-source! Addressed both blocking items:
  
  #3 (Multi-tenancy) — Fixed:
  
  - S3 key layout is now mem0/{user_id}/{doc_id}.txt (tenant-scoped)
  - Metadata sidecar (.metadata.json) written alongside each doc with user_id attribute
  - search() passes metadata filter to managedSearchConfiguration.filter using {"equals": {"key": "user_id", "value": ...}}
  - list(filters={"user_id": "alice"}) scopes to Prefix="mem0/alice/" — no cross-tenant leakage
  - delete()/get()/update() all use the user-scoped path
  - When filters are provided, agentic retrieval is skipped (AgenticRetrieveStream doesn't support metadata filtering) and falls back to standard Retrieve with the filter applied
  
  #7 (Factory registration) — Fixed:
  
  - Added "bedrock_kb": "mem0.vector_stores.bedrock_kb.BedrockKB" to mem0/utils/factory.py
  - Now reachable via Memory.from_config({"vector_store": {"provider": "bedrock_kb", ...}})
  
  E2E tested: insert with different user_ids → list scoped correctly → search with filter returns 0 for non-matching user → search without filter returns all → get/delete scoped by user. 32 unit tests pass.

PR #6233, opened 2026-07-11
Make OSS memory list fetch limit configurable

SteinX:
## Summary

- make the OSS REST memory-list ceiling configurable through `MEM0_OSS_LIST_FETCH_LIMIT`
- use the configured value for the FastAPI `top_k` bound and backend vector-store list limit
- expose the same value to the dashboard as `NEXT_PUBLIC_MEM0_OSS_LIST_FETCH_LIMIT`
- render the configured threshold in the dashboard memories banner
- retain the existing 1,000-item default and fall back safely for invalid or non-positive values

## Why

The fixed 1,000-item ceiling can truncate project-wide listing and maintenance workflows in larger self-hosted deployments. The server and dashboard also need one shared limit so the UI does not continue presenting a hard-coded 1,000 threshold after the backend is raised.

## Validation

- `AUTH_DISABLED=true JWT_SECRET=test-secret HISTORY_DB_PATH=/tmp/mem0-pr6233-test-history.db PYTHONPATH=.:server uv run --with-requirements server/requirements.txt --with pytest python -m pytest tests/test_server_auth.py::TestAuthDisabled::test_all_memories_limit_uses_env_override -q` — 1 passed
- `./node_modules/.bin/tsc --noEmit`
- `bash -n server/dashboard/entrypoint.sh`
- isolated server startup probes verified:
  - unset value -> `ALL_MEMORIES_LIMIT=1000`
  - `MEM0_OSS_LIST_FETCH_LIMIT=5000` -> `ALL_MEMORIES_LIMIT=5000`
  - invalid value -> warning and `ALL_MEMORIES_LIMIT=1000`
- `git diff --check origin/main...HEAD`

## Baseline note

The full `tests/test_server_auth.py` file still models an empty `ADMIN_API_KEY` as disabled auth, while the current server requires `AUTH_DISABLED=true` and `JWT_SECRET`. The focused regression test above is run with the current server contract.


SteinX:
@codex review

PR #6237, opened 2026-07-11
fix(server): scope auth DB sessions to prevent connection-pool exhaustion

### LINKED ISSUE #6236: Auth Depends(get_db) holds a pooled DB connection for the entire request → pool exhaustion on long-running endpoints
## Summary

The self-hosted server's auth dependencies in `server/auth.py` (`verify_auth`,
`require_auth`, `require_admin`) declare `db: Session = Depends(get_db)`. Because
`get_db` is a FastAPI `yield` dependency, FastAPI checks out a SQLAlchemy
connection at **request start** and holds it until the response is sent.

For the auth query itself that connection is used for ~1ms. But it stays pinned
for the **entire** request lifetime — including the long-running LLM endpoints
(`POST /memories` ≈ 20–40s, `POST /search`, `POST /generate-instructions`).

## Why it breaks under load

The engine is built with SQLAlchemy defaults:

```python
# server/db.py
engine = create_engine(_build_database_url(), pool_pre_ping=True)
# → default pool_size=5, max_overflow=10  ⇒  15 usable connections
```

Under concurrency, each in-flight request holds one pooled connection for the
whole LLM call. Once ~15 requests are in flight, the pool is exhausted. New
requests then block and fail on pool-timeout — and because **auth itself needs
the pool**, every subsequent request (including the dashboard) fails at the auth
step. The API wedges even though the process is still "up" and healthy.

## This is a known FastAPI antipattern

This is the well-documented **`Depends(get_db)` + slow-endpoint** antipattern: a
request-scoped DB session tied to the request lifetime means a fast query pins a
pooled connection for the duration of a slow request. The connection is held not
because it's in use, but because the dependency's scope is the whole request.
The standard fix is to **scope the session to the actual query** rather than to
the request.

## Reproduction

1. Start the self-hosted server against Postgres + a real/stubbed LLM key.
2. Fire ~30–40 concurrent `POST /memories`.
3. In another shell, poll the app DB:
   ```sql
   SELECT count(*) FROM pg_stat_activity WHERE datname = 'mem0_app';
   ```
   → connection count pins at **15**, and concurrent `POST /search` hangs / 500s.

## Proposed fix

Scope the auth DB session to just the auth query — open a short-lived
`with SessionLocal() as db:` only on the branches that actually query, in all
three auth dependencies. The auth query releases its connection immediately, so
the long-running endpoint body no longer holds one. On the default pool this
keeps peak connections in the single digits under the same load.

`server/db.py` sets `sessionmaker(..., expire_on_commit=False)`, so a `User`
returned from inside the `with` block is still safe to read after the session
closes (only scalar columns like `user.role` are read downstream — no
lazy-loaded relationships), so behavior is unchanged.

I have a patch ready and will open a PR referencing this issue.

JainamShah-22:
## Linked Issue

Closes #6236

## Description

The self-hosted server's auth dependencies in `server/auth.py` declare
`db: Session = Depends(get_db)`. As a FastAPI `yield` dependency, `get_db`
checks out a pooled SQLAlchemy connection at **request start** and holds it
until the response is sent. The auth query is ~1ms, but the connection stays
pinned for the **entire** request — including the long-running LLM endpoints
(`POST /memories` ≈ 20–40s, `POST /search`, `POST /generate-instructions`).

Under concurrency, in-flight requests each hold a connection for the whole LLM
call, so the default engine pool (`create_engine(..., pool_pre_ping=True)` →
SQLAlchemy default `pool_size=5, max_overflow=10` = 15) is exhausted. New
requests then fail on pool-timeout, and because auth itself needs the pool, the
API and dashboard wedge even though the process stays "up". This is the
well-known FastAPI `Depends(get_db)` + slow-endpoint antipattern — a
request-scoped session pins a connection for the duration of a slow request.

## Change

**`server/auth.py`** — scope the auth DB session to just the auth query: open a
short-lived `with SessionLocal() as db:` only on the branches that actually
query, in all three auth dependencies (`verify_auth`, `require_auth`,
`require_admin`), instead of injecting it via `Depends(get_db)` for the whole
request. The auth query releases its connection immediately, so the endpoint
body no longer holds one.

**`server/routers/auth.py`** — because auth now resolves the user in its own
short-lived session, the `User` returned by `require_auth` is detached from the
endpoint's own `db`. The two endpoints that *write* to that user
(`PATCH /auth/me`, `POST /auth/change-password`) now load a session-managed copy
via `db.get(User, user.id)`, mutate that, and commit — so the writes actually
persist. Read-only endpoints are unaffected (they only read scalar columns; a
sweep confirmed these two are the only sites that mutate the injected user).

`server/db.py` sets `sessionmaker(..., expire_on_commit=False)`, so a `User`
loaded inside a `with` block (or the managed copy above) stays readable after
the session closes. Behavior is unchanged.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Pool fix: reproduced against a local server — ~40 concurrent `POST /memories`
pinned `mem0_app` connections at 15 (`SELECT count(*) FROM pg_stat_activity
WHERE datname='mem0_app'`) and hung `/search`. With this change, connections
stay in single digits and `/search` stays 200 under the same load.

Write endpoints: the `update_me` / `change_password` detachment surfaced in
review is fixed by mutating a session-managed `db.get(User, user.id)`; verified
by inspection against the model/schema (`User` has no server-computed columns
and `UserResponse` only exposes columns already populated on the managed
instance).

`server/` has no pytest harness (the suite under `tests/` covers the SDK, not
the FastAPI server), so no automated regression test is added here to avoid
introducing a new framework. Happy to add one as a follow-up if you'd like.

## Checklist

### DIFF
--- a/server/auth.py
+++ b/server/auth.py
@@ -3,7 +3,7 @@ import secrets
 import uuid
 from datetime import datetime, timedelta, timezone
 
-from db import get_db
+from db import SessionLocal
 from fastapi import Depends, HTTPException, Request
 from fastapi.security import APIKeyHeader, HTTPAuthorizationCredentials, HTTPBearer
 from jose import JWTError, jwt
@@ -145,19 +145,24 @@ async def verify_auth(
     request: Request,
     credentials: HTTPAuthorizationCredentials | None = Depends(bearer_scheme),
     x_api_key: str | None = Depends(api_key_header),
-    db: Session = Depends(get_db),
 ) -> User | None:
-    """Authenticate via JWT, X-API-Key, or legacy ADMIN_API_KEY. Returns User or None."""
+    """Authenticate via JWT, X-API-Key, or legacy ADMIN_API_KEY. Returns User or None.
+
+    A short-lived session is opened only on the branches that query the DB, so no
+    pooled connection is held for the lifetime of the (possibly long-running) request.
+    """
     if credentials is not None:
         _mark_auth_type(request, "bearer")
-        return _resolve_user_from_jwt(credentials.credentials, db)
+        with SessionLocal() as db:
+            return _resolve_user_from_jwt(credentials.credentials, db)
 
     if x_api_key is not None:
         if ADMIN_API_KEY and secrets.compare_digest(x_api_key, ADMIN_API_KEY):
             _mark_auth_type(request, "admin_api_key")
             return None
         _mark_auth_type(request, "api_key")
-        return _resolve_user_from_api_key(x_api_key, db)
+        with SessionLocal() as db:
+            return _resolve_user_from_api_key(x_api_key, db)
 
     if AUTH_DISABLED:
         _mark_auth_type(request, "disabled")
@@ -173,12 +178,12 @@ async def verify_auth(
 async def require_auth(
     request: Request,
     user: User | None = Depends(verify_auth),
-    db: Session = Depends(get_db),
 ) -> User:
     """Like verify_auth but guarantees a non-None User. Use for endpoints that require auth."""
     if user is None:
         if getattr(request.state, "auth_type", "none") in {"admin_api_key", "disabled"}:
-            default_user = _get_default_user(db)
+            with SessionLocal() as db:
+                default_user = _get_default_user(db)
             if default_user is not None:
                 return default_user
         raise HTTPException(status_code=401, detail="Authentication required.")
@@ -193,7 +198,6 @@ _BOOTSTRAP_ADMIN = User(
 async def require_admin(
     request: Request,
     user: User | None = Depends(verify_auth),
-    db: Session = Depends(get_db),
 ) -> User:
     """Like require_auth but also enforces admin role.
 
@@ -203,7 +207,8 @@ async def require_admin(
     auth_type = getattr(request.state, "auth_type", "none")
     if user is None:
         if auth_type in {"admin_api_key", "disabled"}:
-            default_user = _get_default_user(db)
+            with SessionLocal() as db:
+                default_user = _get_default_user(db)
             if default_user is not None:
                 if default_user.role != "admin":
                     raise HTTPException(status_code=403, detail="Admin role required.")
--- a/server/routers/auth.py
+++ b/server/routers/auth.py
@@ -175,18 +175,23 @@ def update_me(
     user: User = Depends(require_auth),
     db: Session = Depends(get_db),
 ):
-    if body.name is not None and body.name.strip():
-        user.name = body.name.strip()
+    # require_auth resolves the user in its own short-lived session, so `user` is
+    # detached from this request's `db`. Load a session-managed copy to mutate.
+    db_user = db.get(User, user.id)
+    if db_user is None:
+        raise HTTPException(status_code=404, detail="User not found.")
 
-    if body.email is not None and body.email != user.email:
-        collision = db.scalar(select(User).where(User.email == body.email, User.id != user.id))
+    if body.name is not None and body.name.strip():
+        db_user.name = body.name.strip()
+
+    if body.email is not None and body.email != db_user.email:
+        collision = db.scalar(select(User).where(User.email == body.email, User.id != db_user.id))
         if collision is not None:
             raise HTTPException(status_code=409, detail="Email is already in use.")
-        user.email = body.email
+        db_user.email = body.email
 
     db.commit()
-    db.refresh(user)
-    return user
+    return db_user
 
 
 @router.post("/change-password", response_model=MessageResponse)
@@ -195,12 +200,15 @@ def change_password(
     user: User = Depends(require_auth),
     db: Session = Depends(get_db),
 ):
-    if not verify_password(body.current_password, user.password_hash):
+    # require_auth resolves the user in its own short-lived session, so `user` is
+    # detached from this request's `db`. Load a session-managed copy to mutate.
+    db_user = db.get(User, user.id)
+    if db_user is None or not verify_password(body.current_password, db_user.password_hash):
         raise HTTPException(status_code=401, detail="Current password is incorrect.")
 
     _require_password_length(body.new_password)
 
-    user.password_hash = hash_password(body.new_password)
+    db_user.password_hash = hash_password(body.new_password)
     db.commit()
     return MessageResponse(message="Password updated.")
 


kartik-mem0:
Sorry this sat without a response for a few days — that's on us, not you.

First, the good part: the diagnosis here is genuinely good work. `Depends(get_db)` is a FastAPI `yield` dependency, so it checks out a pooled connection at request start and holds it for the entire request — including the 20–40s LLM calls in `POST /memories` / `POST /search`. Under concurrency that pins the whole pool and wedges the API. That's a real bug, #6236 was well filed, and the manual repro in the PR description matches what we saw when we traced it. Thank you for tracking this down.

There's a blocker on the fix as written, though, and it's outside this diff's line range, which is probably how it slipped past self-review: removing `Depends(get_db)` from `require_auth` / `verify_auth` changes what kind of object the returned `User` is. Before this PR, FastAPI's dependency caching meant auth's session and the endpoint's own `db: Session = Depends(get_db)` were the *same* session object, so the user stayed attached to it for the life of the request. After this PR, auth loads the user inside its own short-lived `with SessionLocal() as db:` block, which closes before the endpoint body runs — so the `User` object handed to the endpoint is **detached**.

That's fine for endpoints that only read scalar fields off it. It breaks the two endpoints that *write* to that user, both in `server/routers/auth.py`:

- **`PATCH /auth/me`** (`update_me`, `server/routers/auth.py:172-189`): `db.commit()` at line 187 doesn't persist the `user.name`/`user.email` mutations at all, because the detached `user` isn't in this endpoint's `db` session. Then `db.refresh(user)` at line 188 raises `sqlalchemy.exc.InvalidRequestError: Instance '<User>' is not persistent within this Session` — so this endpoint now 500s on every call.
- **`POST /auth/change-password`** (`change_password`, `server/routers/auth.py:192-205`): same root cause, worse outcome. `user.password_hash = hash_password(...)` at line 203 + `db.commit()` at line 204 persist nothing, but there's no refresh call to raise on the detached instance — so the endpoint just returns **200 "Password updated."** at line 205 while the stored hash is untouched. The new password silently never works and the old one still does. No exception anywhere, so nothing surfaces this — it's a fully silent failure on a security-relevant path, which is why this can't merge as-is.

Both of these are on the normal JWT-bearer dashboard flow, not an edge case.

A bounded fix that keeps your pool-scoping approach: in `update_me` and `change_password`, don't mutate the `user` that came from `Depends(require_auth)` — load a session-managed copy instead, e.g. `db_user = db.get(User, user.id)`, mutate `db_user`, `db.commit()`, and drop the `db.refresh(user)` call (return `db_user`). `db.merge(user)` before mutating would work too. Auth's job stays "return identity"; the two write endpoints load their own row through their own `db`.

Separately, and not something to fix in this PR: `server/` doesn't have a pytest harness, so this class of regression currently has no way to show up in CI. Worth a follow-up at some point, but not a reason to hold this one up further.

The pool fix itself is worth landing — this is close. Appreciate the work on it.

JainamShah-22:
No worries at all — it was the weekend, so completely understandable. Thanks for taking the time to review it.

Thank you for the careful review — you're right on both counts, line range.

Pushed a follow-up (`ed62932f`) that keeps the pool-scoping approach and fixes the two write endpoints in `server/routers/auth.py`:

* `PATCH /auth/me` and `POST /auth/change-password` now load a session-managed copy with `db_user = db.get(User, user.id)`, mutate that, and `db.commit()`. The identity returned by `require_auth` is only used for its `id`, so auth's job stays "return identity" and each write endpoint loads its own row through its own `db`.

A couple of things I checked while doing it:

* Swept the rest of the server for other sites that mutate an auth-injected user. These two are the only ones — `routers/api_keys.py` only reads `user.id`, and `register`/`login` operate on a user loaded in their own request session, so they were never affected.
* Pool fix stays intact. These two are fast DB-only endpoints, so holding a connection for their brief query + commit is the normal pattern, not the antipattern. The slow LLM endpoints in `main.py` (`/memories`, `/search`, `/generate-instructions`) don't take `Depends(get_db)` at all, so no connection is held across the long-running body.

One small note on `update_me`: I dropped the `db.refresh(...)` call as unnecessary rather than re-pointing it at `db_user`. With `expire_on_commit=False` (`db.py:18`), the freshly-loaded managed instance already holds current values after commit, and `User` has no server-computed columns — every field `UserResponse` exposes (`id`, `name`, `email`, `role`, `created_at`) is already populated — so the refresh would just be a redundant round-trip.

On the test harness: fully agree `server/` having no `pytest` setup means this class of regression can't surface in CI today. Happy to add a small harness + a regression test for the detach case as a follow-up PR if that's useful — just didn't want to bundle a new framework into this fix.

Thanks again for the thorough read.

kartik-mem0:
## Needs Human Review

Verified against `ed62932f` (new head): both Critical findings from the previous review are fixed.

- **`update_me` 500 on `db.refresh`** — fixed. `routers/auth.py:180-194` now loads `db_user = db.get(User, user.id)` (attached to the endpoint's own request-scoped `db`), mutates `db_user`, commits, and returns `db_user` directly. The `db.refresh(user)` call on the detached `user` is gone.
- **`change_password` silent data loss** — fixed. `routers/auth.py:205-212` mutates and commits the same session-managed `db_user` instead of the detached `user`, so the new password hash genuinely persists now.
- `server/auth.py` (the actual pool-exhaustion fix) is untouched by this commit — byte-identical to the version already confirmed correct, so that fix stands as-is.

Nice bounded fix — this is exactly the "auth returns identity, writers load their own managed row" shape from the earlier suggestion, and it reuses the same `db.get(Model, id)` → mutate → commit convention `routers/api_keys.py::revoke_key` already uses, rather than inventing something new. Also re-swept every `require_auth`/`require_admin`/`verify_auth` call site across `server/` independently on this head — `update_me` and `change_password` are still the only two that mutate the auth-injected user, confirming that sweep.

**Why still Needs Human Review and not LGTM:** this PR touches authentication (`server/auth.py`, `server/routers/auth.py`), which carries a standing human-sign-off requirement in our review process regardless of code quality — not a reflection of any remaining issue with this change.

### Pay attention to
- Nothing blocking. Two cosmetic notes below if you want to fold them into a future touch of this file, not this PR.

### Findings
Three independent review passes (correctness, repo-fit, architecture/design) all came back with zero Critical and zero Important findings. Minor/optional only:
- `routers/auth.py:181-182` vs `206` — `update_me` 404s when the user row has vanished between auth and this call; `change_password` folds the same case into a generic 401. Both are safe, just inconsistent status codes for the same edge case.
- `auth.py`'s `require_auth` docstring doesn't itself flag that the returned `User` is detached — that warning currently only lives in the two comments you added at the call sites. A one-line docstring note would help a future third write-endpoint avoid this same bug class.

### Evidence
- **CI:** license/cla pass; Vercel fail (advisory fork-deploy check, unrelated). `CI Gate` shows `action_required` on this head (pending a maintainer approval click, hasn't run) — moot either way since `server/` has no dedicated CI test workflow in this repo.
- **Targeted tests:** `server/` has no pytest harness (pre-existing gap, not introduced by this PR). Verified instead by `python3 -m py_compile` (clean) and `ruff check` (clean, zero findings) on both changed files, plus three fresh-context reviews independently tracing the same code paths — one of which wrote and ran a standalone SQLAlchemy repro confirming `db.get()` returns a genuinely attached, committable instance, and reproducing the original `InvalidRequestError` in a contrast case to confirm the earlier diagnosis.
- **Red-green proof:** N/A — no test harness to run red/green against; see code-tracing + repro above.
- **Linked issue:** #6236 — root cause (auth pinning a pooled connection for the whole request) remains correctly fixed; this commit only addresses the two regressions that fix introduced.
- **Cross-package check:** `server/` has no TS counterpart; Python-only change.
- **Could NOT verify:** live behavior against a running server + real Postgres (no harness, same limitation as the earlier review).

### Assessment
Ready to merge: **No — maintainer judgment required.** Est. human time: ~10 min
Both previously-blocking regressions are genuinely fixed and the pool-exhaustion fix is untouched; the only reason this isn't LGTM is the standing auth-path sign-off policy, not any remaining code concern.

JainamShah-22:
Hi @kartik-mem0, can i get an ETA on this.

kartik-mem0 (approved):

PR #6240, opened 2026-07-11
fix(opensearch): list() without top_k no longer caps at 10 results

### LINKED ISSUE #6108: bug(opensearch): list() without top_k silently caps at 10 results, truncating get_all()/delete_all()

faithpricejp-source:
Closes #6108

`OpenSearchDB.list()` only set the query `size` when `top_k` was truthy. `Memory.get_all()` and `Memory.delete_all()` call `list(filters=..., top_k=None)` to enumerate **all** matching memories, so the query fell back to OpenSearch's default of 10 hits — `get_all()` returned at most 10 memories and `delete_all()` left everything past the first 10 behind, both silently.

Fix: request `size=10000` (OpenSearch's default `index.max_result_window`) when no `top_k` is given — same convention as `vertex_ai_vector_search.py`'s `search_limit = top_k if top_k is not None else 10000`. Explicit `top_k` behaviour is unchanged.

Note: the Elasticsearch adapter has the same `if top_k:` shape; per the issue this PR scopes to OpenSearch only, happy to follow up on ES if wanted.

Tests assert the outgoing query body for both cases. `pytest tests/vector_stores/test_opensearch.py`: 31 passed.

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -17,6 +17,11 @@ logger = logging.getLogger(__name__)
 
 _SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_.]*$")
 
+# OpenSearch's default index.max_result_window — the largest size a single
+# search can return without deep pagination. Used as the fallback when list()
+# is called without an explicit top_k to enumerate everything (#6108).
+DEFAULT_MAX_RESULT_WINDOW = 10000
+
 
 def _validate_filter(key: str, value) -> None:
     if not isinstance(key, str) or not _SAFE_FILTER_KEY.match(key):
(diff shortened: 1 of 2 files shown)


mem0/vector_stores/opensearch.py
```diff
@@ -381,8 +381,11 @@ def list(self, filters: Optional[Dict] = None, top_k: Optional[int] = None) -> L
             if filter_clauses:
                 query["query"] = {"bool": {"filter": filter_clauses}}
 
-            if top_k:
-                query["size"] = top_k
+            # Without an explicit size OpenSearch returns only its default 10
+            # hits, silently truncating get_all()/delete_all(), which call
+            # list() with top_k=None to enumerate everything (#6108). 10000
+            # is OpenSearch's default index.max_result_window.
+            query["size"] = top_k if top_k else 10000
```
Aditya-Tripuraneni: The 10000 can be exported as a const rather than having a magic number for clean code practice.

faithpricejp-source: Good call — done in c8628f26. Pulled it out as a module-level `DEFAULT_MAX_RESULT_WINDOW = 10000` next to `_SAFE_FILTER_KEY`, with a comment noting it's OpenSearch's default `index.max_result_window`. The 31 existing opensearch tests still pass and ruff is clean.

PR #6245, opened 2026-07-11
fix(plugin): read telemetry manifest as utf-8

Ghraven:
## Summary
- read the plugin telemetry manifest with explicit UTF-8 encoding

## Problem
The telemetry helper loads platform-specific `plugin.json` files without an explicit text encoding. On Windows or other non-UTF-8 locales, that can make manifest parsing depend on the process locale if the metadata ever contains non-ASCII text.

## Before / after
Before, `_load_plugin_version()` used the platform default encoding when opening `plugin.json`.

After, it reads the manifest as UTF-8, matching JSON's expected encoding and the rest of the plugin metadata handling.

## Verification
- `python -m py_compile integrations\mem0-plugin\scripts\telemetry.py`
- `git diff --check`

### DIFF
--- a/integrations/mem0-plugin/scripts/telemetry.py
+++ b/integrations/mem0-plugin/scripts/telemetry.py
@@ -42,7 +42,7 @@ def _load_plugin_version(platform_name: str = "") -> str:
     parts = _PLATFORM_MANIFESTS.get(platform_name, _DEFAULT_MANIFEST)
     try:
         plugin_json = os.path.join(os.path.dirname(__file__), *parts)
-        with open(plugin_json) as f:
+        with open(plugin_json, encoding="utf-8") as f:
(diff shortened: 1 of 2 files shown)


Ghraven:

Aditya-Tripuraneni (commented):
This makes sense.

This is more of a nit pick but I would add a small regression test for `_load_plugin_version()` with non-ASCII manifest content. An emoji of your choice could work, cuz later if this is ever removed, then the test would catch the regression.

Ghraven:
Thanks for the review and the suggestion. I added a focused regression test that writes a UTF-8 plugin manifest with non-ASCII metadata and verifies _load_plugin_version() still reads the version correctly. Verified locally with python -m pytest integrations/mem0-plugin/tests/test_telemetry.py -q (21 passed).

PR #6249, opened 2026-07-12
Refactor auth functions to use SessionLocal

krishna231122-del:
### Problem

The self-hosted server's auth dependencies (`verify_auth`, `require_auth`, `require_admin`) in `auth.py` declared `db: Session = Depends(get_db)`. Since `get_db` is a FastAPI yield dependency, FastAPI checks out a SQLAlchemy connection at request start and holds it until the response is sent.

For the auth query itself, that connection is used for ~1ms. But it stays pinned for the entire request lifetime — including long-running LLM endpoints (`POST /memories` ≈ 20–40s, `POST /search`, `POST /generate-instructions`).

Under concurrency, each in-flight request holds one pooled connection for the whole LLM call. With the default SQLAlchemy pool (`pool_size=5`, `max_overflow=10` = 15 usable connections), once ~15 requests are in flight, the pool is exhausted. New requests block and fail on pool-timeout — and because auth itself needs the pool, every subsequent request (including the dashboard) fails at the auth step. The API wedges even though the process is still "up" and healthy.

This is a well-documented FastAPI antipattern: a request-scoped DB session tied to the request lifetime means a fast query pins a pooled connection for the duration of a slow request.

### Solution

Scope the auth DB session to just the auth query — open a short-lived `with SessionLocal() as db:` only on the branches that actually query the database, in all three auth dependencies. The auth query releases its connection immediately, so the long-running endpoint body no longer holds one.

### Changes

**`auth.py`:**
- Changed import from `from db import get_db` to `from db import SessionLocal`
- Removed `db: Session = Depends(get_db)` parameter from `verify_auth`, `require_auth`, and `require_admin`
- Wrapped database query branches with `with SessionLocal() as db:` to use short-lived sessions only when needed

### Impact

- Peak connections under load remain in the single digits instead of pinning at 15
- Long-running endpoints no longer hold pooled connections during LLM calls
- Behavior is unchanged because `db.py` sets `expire_on_commit=False`, so `User` objects returned from inside the `with` block are still safe to read after the session closes (only scalar columns like `user.role` are read downstream)

### Testing

- [x] Python syntax validation passed
- [x] Changes follow the standard fix for the FastAPI `Depends(get_db)` + slow-endpoint antipattern

Changes
auth.py:
Changed import from from db import get_db to from db import SessionLocal
Removed db: Session = Depends(get_db) parameter from verify_auth, require_auth, and require_admin
Wrapped database query branches with with SessionLocal() as db: to use short-lived sessions only when needed
Impact
Peak connections under load remain in the single digits instead of pinning at 15
Long-running endpoints no longer hold pooled connections during LLM calls
Behavior is unchanged because db.py sets expire_on_commit=False, so User objects returned from inside the with block are still safe to read after the session closes (only scalar columns like user.role are read downstream)
Testing
Python syntax validation passed
Changes follow the standard fix for the FastAPI Depends(get_db) + slow-endpoint antipattern

### DIFF
--- a/server/auth.py
+++ b/server/auth.py
@@ -3,7 +3,7 @@ import secrets
 import uuid
 from datetime import datetime, timedelta, timezone
 
-from db import get_db
+from db import SessionLocal
 from fastapi import Depends, HTTPException, Request
 from fastapi.security import APIKeyHeader, HTTPAuthorizationCredentials, HTTPBearer
 from jose import JWTError, jwt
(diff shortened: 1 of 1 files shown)


JainamShah-22:
This is a duplicate of : https://github.com/mem0ai/mem0/pull/6237

PR #6255, opened 2026-07-12
fix(ts-oss/redis): keep entity ids snake_case in returned payloads

### LINKED ISSUE #6254: TS OSS Redis store returns entity ids as camelCase, dropping them from results and leaking into metadata
### Component

TypeScript SDK

### Summary

The TS OSS Redis vector store returns its payload with the entity ids in camelCase, which breaks the payload shape the rest of the SDK expects. In `search()`, `get()` and `list()` the whole result payload is passed through `toCamelCase(...)`, so `user_id`/`agent_id`/`run_id` come back as `userId`/`agentId`/`runId`.

Every other store returns entity ids in snake_case. The in-memory store's `normalizePayload` deliberately maps `userId -> user_id` (and keeps timestamps camelCase, e.g. `createdAt`), and `memory/index.ts` reads `payload.user_id` (not `payload.userId`) everywhere while reading `payload.createdAt` for timestamps. So the canonical payload is snake_case ids + camelCase timestamps, and Redis is the only store that diverges.

Consequences for Redis users:

- `_sessionFiltersFromPayload` reads `payload.user_id` (now undefined), so entity-store cleanup on delete/update runs unscoped.

### Steps to Reproduce

Configure the OSS SDK with `vectorStore.provider: "redis"`, add a memory for `userId: "u1"`, then `getAll`/`search`.

### Expected Behavior

### Actual Behavior

### Environment

- vectorStore.provider: redis

abhay-codes07:
## Linked Issue

Closes #6254

## Description

The Redis vector store ran every returned payload through `toCamelCase` in `search()`, `get()` and `list()`, which renamed `user_id`/`agent_id`/`run_id` to `userId`/`agentId`/`runId`.

That breaks the canonical payload contract. `memory/index.ts` and the other stores (memory, qdrant, pgvector) read `payload.user_id`, and the in-memory store's `normalizePayload` deliberately keeps entity ids snake_case while leaving timestamps camelCase (`createdAt`). Redis was the only store returning camelCase ids.

For Redis users this meant:

- `search()`/`getAll()`/`get()` dropped `user_id`/`agent_id`/`run_id` from results, since index.ts reads the snake_case keys which were now camelCase.
- Those camelCase keys were not in index.ts's snake_case `excludedKeys`, so they leaked into `metadata` instead.
- `_sessionFiltersFromPayload` read `payload.user_id` (undefined), so entity-store cleanup on delete/update ran unscoped.

The fix re-attaches the entity ids in snake_case after the camelCase pass (timestamps and metadata keep their existing casing), through a small `withEntityIds` helper used by all three read paths. This is the minimal change that restores the contract without altering the timestamp/metadata behavior.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added three tests to `tests/redis.unit.test.ts` (using the existing mocked redis client) for `search`, `list` and `get`: each asserts the returned payload has snake_case `user_id`/`agent_id`/`run_id`, no camelCase variants, camelCase `createdAt`, and metadata intact. All three fail on `main` (ids come back `undefined`) and pass with this change. `redis.unit`, `valkey` and `vector-store.unit` suites pass; the `vector-stores-compat` failures are a pre-existing missing optional dependency (`@databricks/sql`) unrelated to this change and are identical with and without it.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/redis.ts
+++ b/mem0-ts/src/oss/src/vector_stores/redis.ts
@@ -145,6 +145,24 @@ function toCamelCase(obj: Record<string, any>): Record<string, any> {
   );
 }
 
+// Re-attach entity ids in snake_case after a toCamelCase pass on a stored
+// payload. The entity ids (user_id/agent_id/run_id) are part of the canonical
+// payload contract in snake_case: index.ts and every other vector store read
+// payload.user_id, not payload.userId. Running the whole payload through
+// toCamelCase renamed them to userId/agentId/runId, so they were dropped from
+// search/getAll/get results and leaked into metadata, and delete/update entity
+// cleanup could not read the scope. Timestamps stay camelCase (createdAt) as
+// the contract expects, which is why the camelCase pass is kept for the rest.
+function withEntityIds(
+  payload: Record<string, any>,
+  source: { agent_id?: string; run_id?: string; user_id?: string },
+): Record<string, any> {
+  if (source.agent_id) payload.agent_id = source.agent_id;
+  if (source.run_id) payload.run_id = source.run_id;
+  if (source.user_id) payload.user_id = source.user_id;
+  return payload;
+}
+
 export class RedisDB implements VectorStore {
   private client: RedisClientType<
     RedisDefaultModules & RedisModules & RedisFunctions & RedisScripts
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss redis fix with tests. The store ran returned payloads through toCamelCase, so user_id/agent_id/run_id came back as userId/agentId/runId, which index.ts does not read. That dropped the ids from search/getAll/get results and leaked them into metadata, and entity cleanup on delete/update ran unscoped. Keeps them snake_case like every other store. Quick look when you can, thanks.

PR #6259, opened 2026-07-12
fix(chroma): raise on unknown filter operator instead of silent equality

### LINKED ISSUE #6258: Chroma vector store silently downgrades unknown filter operators to equality
### Description

`ChromaDB._generate_where_clause` (`mem0/vector_stores/chroma.py`) maps a set of known filter operators (`eq`, `ne`, `gt`, `gte`, `lt`, `lte`, `in`, `nin`, `contains`, `icontains`). Any operator outside that set falls into an `else` branch that silently rewrites the condition to equality:

```python
else:
    # Unknown operator, treat as equality
    chroma_condition[key] = {"$eq": val}
```

So a filter like `{"age": {"betwen": [10, 20]}}` (a typo, or an operator Chroma doesn't support) is silently turned into `{"age": {"$eq": [10, 20]}}`. The query then returns **wrong results with no signal to the caller** — no error, no warning.

Other stores fail fast here: e.g. the Qdrant store raises `ValueError` on invalid operator usage. Chroma should do the same rather than returning silently-incorrect results.

### Expected

An unsupported operator should raise a clear `ValueError` naming the operator and field, instead of being downgraded to equality.

### Repro

```python
ChromaDB._generate_where_clause({"age": {"betwen": [10, 20]}})
# currently: {"age": {"$eq": [10, 20]}}  (silently wrong)
# expected: ValueError: Unsupported filter operator 'betwen' ...
```

I have a fix + test ready and will open a PR.

HrushiYadav:
### Description

`ChromaDB._generate_where_clause` downgraded any unrecognized filter operator to `$eq`:

```python
else:
    # Unknown operator, treat as equality
    chroma_condition[key] = {"$eq": val}
```

So `{"age": {"betwen": [10, 20]}}` (a typo, or an unsupported operator) silently became `{"age": {"$eq": [10, 20]}}` — returning wrong results with no signal to the caller.

This changes the `else` branch to raise a `ValueError` naming the operator and field, matching the fail-fast behavior other stores already have (e.g. the Qdrant store raises on invalid operator usage).

Fixes #6258

### Changes

- `mem0/vector_stores/chroma.py`: raise `ValueError` for unsupported operators instead of silently mapping to `$eq`.
- `tests/vector_stores/test_chroma.py`: add a regression test asserting an unknown operator raises.

### Testing

- `pytest tests/vector_stores/test_chroma.py` → 28 passed
- `ruff check` clean on both files

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -293,8 +293,14 @@ class ChromaDB(VectorStoreBase):
                         # ChromaDB doesn't support contains, fallback to equality
                         chroma_condition[key] = {"$eq": val}
                     else:
-                        # Unknown operator, treat as equality
-                        chroma_condition[key] = {"$eq": val}
+                        # Fail fast on unknown operators instead of silently
+                        # downgrading to equality, which returns wrong results
+                        # without any signal to the caller.
+                        raise ValueError(
+                            f"Unsupported filter operator {op!r} for field {key!r}. "
+                            "Supported operators: eq, ne, gt, gte, lt, lte, in, nin, "
+                            "contains, icontains."
+                        )
                 return chroma_condition
             else:
                 # Simple equality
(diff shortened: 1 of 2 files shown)


HrushiYadav:
Closing this. Looks like the recent _generate_where_clause refactor addressed the range-operator bug in this area and intentionally keeps the unknown-operator equality fallback (documented in the new comments), so this change no longer fits. Thanks!

PR #6263, opened 2026-07-12
fix(ts-oss): coerce non-string entity ids instead of crashing on trim()

### LINKED ISSUE #6262: TS OSS validateAndTrimEntityId crashes on non-string (e.g. integer) entity ids
### Component

TypeScript SDK

### Summary

`validateAndTrimEntityId` in the OSS SDK only checks for `undefined` and then calls `value.trim()`. If a non-string id is passed (for example an integer database primary key, a common pattern) the call throws `TypeError: value.trim is not a function`. `null` is also unhandled. `add()`, `search()`, `getAll()` and `deleteAll()` all route entity ids through this helper, so all of them crash on a numeric id.

The Python SDK handles this: `_validate_and_trim_entity_id` coerces non-strings to `str` before trimming (this was fixed in a recent PR), so the two SDKs diverge.

### Steps to Reproduce

```ts
await memory.getAll({ filters: { user_id: 42 } });
// or memory.add("hi", { userId: 42 }), memory.search("q", { filters: { user_id: 42 } })
```

### Expected Behavior

The numeric id is coerced to its string form (`"42"`) and used to scope the operation, matching Python.

### Actual Behavior

`TypeError: value.trim is not a function`.

### Environment

abhay-codes07:
## Linked Issue

Closes #6262

## Description

`validateAndTrimEntityId` only guarded against `undefined` and then called `value.trim()`. A caller passing a non-string id (e.g. an integer database primary key, a common pattern) hit `TypeError: value.trim is not a function`, and `null` was not handled. `add()`, `search()`, `getAll()` and `deleteAll()` all route entity ids through this helper, so all crashed on numeric ids.

The Python SDK already coerces non-strings to `str` at the same validation point (`_validate_and_trim_entity_id`), so the two SDKs diverged. This ports that behavior: coerce non-string values to string (and treat `null` like `undefined`), so numeric ids are scoped by their string form consistently across every entry point.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added a "non-string entity ID coercion" block to `tests/memory.validation.test.ts`: it spies on the vector store and asserts an integer `user_id` (42) reaches `getAll`'s `list` and `search`'s `search` as `"42"`. Both fail on `main` (`TypeError: value.trim is not a function`) and pass with this change. The full `memory.validation` suite (36 tests) passes; typecheck and prettier are clean.

## Checklist


abhay-codes07:
@kartik-mem0 small ts-oss fix with tests. validateAndTrimEntityId called .trim() after only an undefined check, so a non-string id (e.g. an integer database key) crashed with a TypeError across add/search/getAll/deleteAll. Coerces to string at the single validation point, matching the Python SDK. Quick look when you can, thanks.

kartik-mem0:
LGTM. Clean root-cause fix, and I'm confident from the code.

**What I checked**

- **Right layer, whole surface.** `validateAndTrimEntityId` is the single point that `add()`, `search()`, `getAll()` and `deleteAll()` all route their `user_id`/`agent_id`/`run_id` through, so coercing once here fixes all four. I grepped every other `.trim()` on id-shaped values, nothing bypasses this helper (the only other one is `_normalizeEntityText`, which trims extracted entity *text*, not scope ids), so this fixes the root cause and not just the repro.
- **Red-green, this run.** In a worktree: the two coercion tests pass; revert `memory/index.ts` to `main` (keeping the test) and both fail with the exact issue error, `TypeError: value.trim is not a function` at the old `value.trim()` line; restore and they pass again. The test reproduces the real failure, not just the mechanism.
- **Python parity is real.** Read `_validate_and_trim_entity_id` in `mem0/memory/main.py`: same order (null-check, coerce, trim, empty/whitespace reject), and `null → undefined` mirrors Python's `None → None`. This removes the SDK divergence instead of adding one. Nice touch using `=== null` rather than `!value`, so `0` stays a valid id.
- CI: TypeScript SDK build + integration (20/22) and the gate are green. (The red Vercel check is just the deploy preview, no fork perms, not merge-relevant.)

**Optional, non-blocking.** The fix lives at the shared helper so these run identical code, but to lock the whole issue surface into the suite you could add two quick assertions mirroring the ones you wrote: a numeric id through `add()`/`deleteAll()`, and a `null` id (the other half of #6262) resolving to unscoped instead of throwing. Equally happy to merge as-is.

Really tidy, minimal fix. Ready to squash-merge from my side (`gh pr merge 6263 --squash`).

kartik-mem0 (approved):

PR #6265, opened 2026-07-12
fix(ts-oss/memory): apply all operators in a compound field filter

### LINKED ISSUE #6264: TS OSS default memory store ignores all but the first operator in a compound field filter
### Component

TypeScript SDK

### Summary

`MemoryVectorStore.matchFieldCondition` (the default OSS vector store) evaluates comparison operators with a chain of early returns. When a single field condition combines more than one operator, only the first is applied and the rest are ignored.

The most visible case is a range: `{ age: { gte: 10, lte: 20 } }` returns as soon as it checks `gte`, so `lte` is never evaluated and records with `age > 20` still match. The same compound shape is produced by advanced `AND` filters, so range filtering is silently broken on the default store.

Other stores (pgvector, qdrant) AND all operators together, so this is also a Python/TS parity gap.

### Steps to Reproduce

```ts
// default vectorStore.provider: "memory"
await memory.getAll({ filters: { user_id: "u1", age: { gte: 10, lte: 20 } } });
```

With three records aged 5, 15 and 25, the query should return only the age-15 record.

### Expected Behavior

Both bounds apply: only `age` in `[10, 20]` matches.

### Actual Behavior

Only `gte: 10` is applied; `lte: 20` is ignored, so age 15 and age 25 both match.

### Environment

- vectorStore.provider: memory (default)

abhay-codes07:
## Linked Issue

Closes #6264

## Description

`MemoryVectorStore.matchFieldCondition` evaluated comparison operators with a chain of early returns:

```ts
if ("gte" in value) return payloadValue >= value.gte;
if ("lte" in value) return payloadValue <= value.lte;
```

So a condition combining more than one operator on the same field only ever applied the first. A range like `{ age: { gte: 10, lte: 20 } }` returned on `gte` and never checked `lte`, so records with `age > 20` wrongly matched. This is the default vector store, and the same compound shape is produced by advanced `AND` filters, so range filtering was silently broken by default. pgvector and qdrant AND all operators together, so this was also a parity gap.

The fix evaluates every operator present in the condition and requires all of them to hold (returning `false` on the first that fails). Single-operator conditions, the `nin` semantics, and the unknown-operator equality fallback are unchanged.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added a "compound field operators" block to `tests/vector-store.unit.test.ts` (direct `MemoryVectorStore`, real SQLite): with records aged 5/15/25, an inclusive range `{gte:10, lte:20}` and an exclusive range `{gt:5, lt:25}` each return only the age-15 record, and a single-operator condition still returns both matches. The range tests fail on `main` (age 25 leaks in) and pass with this change. Full `vector-store.unit` suite passes; typecheck and prettier are clean.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/memory.ts
+++ b/mem0-ts/src/oss/src/vector_stores/memory.ts
@@ -111,46 +111,63 @@ export class MemoryVectorStore implements VectorStore {
       return value.includes(payloadValue);
     }
 
-    // Handle comparison operators
-    if ("eq" in value) {
-      return payloadValue === value.eq;
-    }
-    if ("ne" in value) {
-      return payloadValue !== value.ne;
-    }
-    if ("gt" in value) {
-      return payloadValue > value.gt;
-    }
-    if ("gte" in value) {
-      return payloadValue >= value.gte;
-    }
-    if ("lt" in value) {
-      return payloadValue < value.lt;
-    }
-    if ("lte" in value) {
-      return payloadValue <= value.lte;
-    }
-    if ("in" in value) {
-      return Array.isArray(value.in) && value.in.includes(payloadValue);
-    }
-    if ("nin" in value) {
-      return !Array.isArray(value.nin) || !value.nin.includes(payloadValue);
-    }
-    if ("contains" in value) {
-      return (
-        typeof payloadValue === "string" &&
-        payloadValue.includes(value.contains)
-      );
-    }
-    if ("icontains" in value) {
-      return (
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss fix with tests. The default memory store's matchFieldCondition returned on the first operator, so a compound field filter like age:{gte:10,lte:20} only applied gte and ignored lte (records with age>20 wrongly matched). Now evaluates all operators like pgvector/qdrant do. Quick look when you can, thanks.

PR #6267, opened 2026-07-12
fix(ts-oss/valkey): keep entity ids snake_case in returned payloads

### LINKED ISSUE #6266: TS OSS Valkey store returns entity ids as camelCase, dropping them from results and leaking into metadata
### Component

TypeScript SDK

### Summary

### Steps to Reproduce

### Expected Behavior

### Actual Behavior

### Environment

abhay-codes07:
## Linked Issue

Closes #6266

## Description

`docToResult` (used by `search`, `get` and `list`) ran the whole payload through `toCamelCase`, renaming `user_id`/`agent_id`/`run_id` to `userId`/`agentId`/`runId`.

That breaks the canonical payload contract. `memory/index.ts` and the other stores (memory, qdrant, pgvector) read `payload.user_id`, and the in-memory store keeps entity ids snake_case while leaving timestamps camelCase (`createdAt`). Valkey was an outlier.

For Valkey users this meant:

- `search()`/`getAll()`/`get()` dropped `user_id`/`agent_id`/`run_id` from results, since index.ts reads the snake_case keys which were now camelCase.
- Those camelCase keys were not in index.ts's snake_case `excludedKeys`, so they leaked into `metadata`.
- Entity-store cleanup on delete/update read `payload.user_id` (undefined) and ran unscoped.

The fix attaches the entity ids in snake_case after the camelCase pass (timestamps and metadata keep their casing). This is the same fix already applied to the Redis store.

The existing `get()` test asserted the camelCase output (`payload.userId`), which encoded the bug; it is updated to the snake_case contract, and an `agent_id`/`run_id` case is added.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Updated the insert/get round-trip test to assert `payload.user_id === "alice"` (and no `userId`), and added a test asserting `agent_id`/`run_id` come back snake_case. Both fail on `main` (ids come back `undefined` under the snake_case keys) and pass with this change. Full `valkey` suite (10 tests) passes; typecheck and prettier are clean.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/valkey.ts
+++ b/mem0-ts/src/oss/src/vector_stores/valkey.ts
@@ -379,9 +379,6 @@ export class ValkeyDB implements VectorStore {
         this.timezone,
       );
     }
-    if (doc.agent_id) resultPayload.agent_id = doc.agent_id;
-    if (doc.run_id) resultPayload.run_id = doc.run_id;
-    if (doc.user_id) resultPayload.user_id = doc.user_id;
 
     if (doc.metadata) {
       try {
@@ -391,9 +388,20 @@ export class ValkeyDB implements VectorStore {
       }
     }
 
+    // Entity ids are part of the canonical payload contract in snake_case:
+    // index.ts and the other vector stores read payload.user_id, not
+    // payload.userId. Running the whole payload through toCamelCase renamed
+    // them to userId/agentId/runId, which dropped them from search/getAll/get
+    // results and leaked them into metadata. Attach them in snake_case after
+    // the camelCase pass (timestamps stay camelCase as the contract expects).
+    const payload = toCamelCase(resultPayload);
+    if (doc.agent_id) payload.agent_id = doc.agent_id;
+    if (doc.run_id) payload.run_id = doc.run_id;
+    if (doc.user_id) payload.user_id = doc.user_id;
+
     return {
       id: doc.memory_id ?? "",
-      payload: toCamelCase(resultPayload),
+      payload,
       score,
     };
   }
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss valkey fix with tests, same class as the redis one. docToResult ran the payload through toCamelCase, so user_id/agent_id/run_id came back as userId/agentId/runId, which index.ts does not read. That dropped the ids from search/getAll/get results, leaked them into metadata, and ran entity cleanup unscoped. Keeps them snake_case like the other stores. Quick look when you can, thanks.

Jonathan-Improving:
Hi — we independently worked on a fix for #6266 and wanted to flag a couple of things we found during review that this PR doesn't currently address:

1. **Metadata can overwrite entity IDs (security):** `Object.assign(resultPayload, JSON.parse(doc.metadata))` runs after entity IDs are set. Crafted metadata stored in Valkey (e.g. `{"user_id": "attacker"}`) can overwrite the authoritative entity IDs from the hash fields. A defensive fix is to strip entity ID keys from parsed metadata before merging.

2. **Prototype pollution in timestamp check:** If using `key in obj` to check membership in a rename map, `__proto__` or `constructor` keys would traverse the prototype chain. `Object.hasOwn()` is safer.

Neither is a merge blocker for the core bug fix, but they're worth considering as defense-in-depth — especially #1 since it could allow entity-scoped operations to run against the wrong user's data.

Our alternative approach (replaces `toCamelCase` entirely with a selective transformer rather than re-assigning after) is here if useful as a reference: https://github.com/Jonathan-Improving/mem0/pull/1

PR #6274, opened 2026-07-13
fix(elasticsearch): set size=top_k on KNN search so results are not capped at 10

### LINKED ISSUE #5909: bug: Elasticsearch vector search caps results at 10 (missing size on KNN query)

Bartok9:
## Summary

Salvages #5910 by @AnnaSuSu (and the later parallel #5951 by @ly-wang19) onto current `main`.

`ElasticsearchDB.search()` builds a KNN body with `knn.k` / `num_candidates` but without top-level `size`. Elasticsearch defaults `size` to **10**, so `top_k > 10` is silently truncated. `keyword_search()` and `list()` already set `size`.

This still bite on main: Memory's hybrid path over-fetches with `max(limit * 4, 60)`, so default searches only score 10 candidates on Elasticsearch.

Closes #5909

## Motivation

Silent recall/ranking regression only on Elasticsearch — other store backends honor `top_k` fully.

## Root cause

Missing `"size": top_k` on the default KNN request body in `mem0/vector_stores/elasticsearch.py`.

## Fix

Add `"size": top_k` next to the knn clause (same approach as #5910 / #5951).

## Why salvage

#5910 and #5951 are open, mergeable, and CI-green but have not landed; the bug remains on main. This rebased salvage keeps the same minimal fix + regression tests under Bartok9 for recheck against current tip.

## Tests

- Extended `test_search` asserts `body["size"] == top_k`
- New `test_search_sets_size_to_top_k` for over-fetch (`max(20*4, 60)`)

```
uv run pytest tests/vector_stores/test_elasticsearch.py -q
# 21 passed
```

## Verification

- Focused: elasticsearch unit tests above
- Diff scoped to 2 files

## Credits

Original analysis and patch shape: @AnnaSuSu (#5910), also @ly-wang19 (#5951).

### DIFF
--- a/mem0/vector_stores/elasticsearch.py
+++ b/mem0/vector_stores/elasticsearch.py
@@ -163,7 +163,10 @@ class ElasticsearchDB(VectorStoreBase):
             search_query = self.custom_search_query(vectors, top_k, filters)
         else:
             search_query = {
-                "knn": {"field": "vector", "query_vector": vectors, "k": top_k, "num_candidates": top_k * 2}
+                # Without `size`, Elasticsearch caps the response at its default of 10 hits
+                # regardless of `knn.k`, silently truncating results when top_k > 10.
+                "size": top_k,
+                "knn": {"field": "vector", "query_vector": vectors, "k": top_k, "num_candidates": top_k * 2},
             }
             if filters:
                 filter_conditions = []
--- a/tests/vector_stores/test_elasticsearch.py
+++ b/tests/vector_stores/test_elasticsearch.py
@@ -209,12 +209,33 @@ class TestElasticsearchDB(unittest.TestCase):
         self.assertEqual(body["knn"]["query_vector"], vectors)
         self.assertEqual(body["knn"]["k"], 5)
         self.assertEqual(body["knn"]["num_candidates"], 10)
+        # `size` must be set or ES caps the response at its default of 10 hits.
+        self.assertEqual(body["size"], 5)
 
         # Verify results
         self.assertEqual(len(results), 1)
         self.assertEqual(results[0].id, "id1")
         self.assertEqual(results[0].score, 0.8)
         self.assertEqual(results[0].payload, {"key1": "value1"})
+
+    def test_search_sets_size_to_top_k(self):
+        """Regression for #5909: KNN search must set `size=top_k` or Elasticsearch caps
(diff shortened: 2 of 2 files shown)


AnnaSuSu:
Thanks for keeping this on the radar @Bartok9 — but #5910 is still open, and as far as I can tell the diff here is identical to it (same two files, same +25/-1, same `"size": top_k` line and the same regression test), so we now have three open PRs for the same fix: #5910 (2026-06-26), #5951 by @ly-wang19 (2026-06-28), and this one.

The staleness this salvages is gone: I've just rebased #5910 onto current `main` (17836748) and CI is green there. Would you mind closing this in favour of #5910? If either your PR or @ly-wang19's covers something mine doesn't, I'm happy to fold it in and credit you both.

Bartok9:
Thanks @AnnaSuSu — you're right.

#5910 is the earlier open PR (2026-06-26), still open, rebased onto current main with green CI, and the diff matches what #6274 was doing. Happy to close this in favor of yours so maintainers have a single clean path.

Also noted #5951 from @ly-wang19 on the same Elasticsearch `size`/`top_k` issue — if anything useful from there isn't already in #5910, folding it in there makes sense.

Closing #6274 in favor of #5910. Thanks for flagging it.

Bartok9:
Closing in favor of #5910 (same fix, earlier PR) as discussed — thank you @AnnaSuSu.

PR #6275, opened 2026-07-13
fix(redis): request updated_at in search/keyword_search return_fields

### LINKED ISSUE #6059: bug(redis): search() and keyword_search() omit updated_at from return_fields, making the updated_at payload guards dead code

Bartok9:
## Summary

Salvages #6060 by @Krishnachaitanyakc onto current `main`.

Redis `search()` / `keyword_search()` omit `updated_at` from redis-vl `return_fields`, even though:

- `DEFAULT_FIELDS` indexes `updated_at`
- `update()` persists it
- `get()` / `list()` already return it

The `if "updated_at" in result` payload guards were dead code — search paths never surface `updated_at`.

Closes #6059

## Motivation

Metadata parity across read paths. After update, `get`+`list` show `updated_at` while hybrid/search drops it on Redis only.

## Root cause

Incomplete `return_fields` lists on `VectorQuery` and `TextQuery` in `mem0/vector_stores/redis.py`.

## Fix

Include `"updated_at"` in both `return_fields` (same fix as #6060).

## Why salvage

#6060 is open/mergeable/CI-green but not merged; bug still present on main.

## Tests

- `test_search_requests_updated_at_in_return_fields`
- `test_keyword_search_requests_updated_at_in_return_fields`

```
uv run pytest tests/vector_stores/test_redis.py -q
# 13 passed
```

## Verification

- Focused redis vector-store unit tests
- Diff scoped to 2 files

## Credits

Original report + patch: @Krishnachaitanyakc (#6060 / issue #6059).

### DIFF
--- a/mem0/vector_stores/redis.py
+++ b/mem0/vector_stores/redis.py
@@ -155,7 +155,17 @@ class RedisDB(VectorStoreBase):
         v = VectorQuery(
             vector=np.array(vectors, dtype=np.float32).tobytes(),
             vector_field_name="embedding",
-            return_fields=["memory_id", "hash", "agent_id", "run_id", "user_id", "memory", "metadata", "created_at"],
+            return_fields=[
+                "memory_id",
+                "hash",
+                "agent_id",
+                "run_id",
+                "user_id",
+                "memory",
+                "metadata",
+                "created_at",
+                "updated_at",
+            ],
             filter_expression=filter,
             num_results=top_k,
         )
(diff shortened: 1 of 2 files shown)


Bartok9:
Superseded by #6458, which rebases the return_fields fix onto current main and also persists `updated_at` on `insert()` so search/keyword paths can surface timestamps for newly written memories (not only post-update). Closing this PR in favor of #6458.

Bartok9:
Closing in favor of #6458 (complete fix on current main).

PR #6278, opened 2026-07-13
fix(memory): don't let update() metadata overwrite user_id/agent_id/run_id

### LINKED ISSUE #6277: update() lets caller metadata silently overwrite user_id/agent_id/run_id, breaking tenant isolation
### Bug Description

`Memory.update()` / `AsyncMemory.update()` let caller-supplied `metadata` silently overwrite a memory's `user_id`, `agent_id`, and `run_id` — the exact fields every `search()`, `get_all()`, and `delete_all()` call uses for tenant scoping. This is a real bug, not a hypothetical: `_update_memory` already protects `actor_id` from this same class of overwrite (see the fix for #4490), but that protection was never extended to `user_id`/`agent_id`/`run_id`.

**Where:** `mem0/memory/main.py`, in `_update_memory` (both the sync `Memory` class and the async `AsyncMemory` class — identical logic in both):

```python
if metadata is not None:
    new_metadata.update(metadata)   # <- caller's dict merged with no denylist

...
# actor_id is immutable after creation (issue #4490)
    new_metadata["actor_id"] = existing_memory.payload["actor_id"]
```

Only `actor_id` (and a few derived fields like `hash`/`data`/timestamps) get re-asserted after the merge. `user_id`/`agent_id`/`run_id` do not.

Notably, `add()`'s `_build_filters_and_metadata()` already implements the correct pattern elsewhere in this same file: it copies the caller's `metadata` first, then overwrites `user_id`/`agent_id`/`run_id` from the validated arguments afterward, so the real identifiers always win. `update()` has no equivalent step.

### Steps to Reproduce

```python
from mem0 import Memory

m = Memory()
result = m.add("I love pizza", user_id="alice")

# Alice's own innocuous-looking metadata update...
m.update(memory_id, metadata={"category": "food", "user_id": "bob"})

# ...silently reassigns the memory to bob.
print(m.get_all(user_id="alice"))  # -> [] (memory vanished for alice)
print(m.get_all(user_id="bob"))    # -> the memory alice created, now under bob
print(m.delete_all(user_id="alice"))  # does NOT remove this memory anymore
```

### Expected Behavior

`update()`'s `metadata` argument should never be able to move a memory across `user_id`/`agent_id`/`run_id` scope — mirroring the `actor_id` protection already implemented for #4490.

### Actual Behavior

- The memory disappears from the original tenant's `search()`/`get_all()`/`delete_all()` (silent data loss for that tenant — `delete_all()` no longer reaches it).
- The memory appears under the new tenant's `search()`/`get_all()` results (cross-tenant data injection, without that tenant ever creating it).

This affects the default OSS `Memory`/`AsyncMemory` API surface used by every self-hosted deployment, and the shipped `server/` REST server's `PUT /memories/{id}` (whose `MemoryUpdate` schema passes `metadata` straight through with no key restriction).

### Affected Component

Core / Python SDK

### Notes / Proposed Fix

Extend the existing `actor_id` re-assertion in `_update_memory` (sync + async) to also cover `user_id`/`agent_id`/`run_id`, exactly the same pattern already accepted for #4490. I have a fix + regression tests ready and will open a PR referencing this issue.

okaditya84:
## Linked Issue

Closes #6277

## Description

`_update_memory` (both `Memory` and `AsyncMemory`, identical logic) merges the caller's `metadata` dict on top of the existing payload with no denylist, then only re-asserts `actor_id` afterward — the fix that was made for #4490. `user_id`/`agent_id`/`run_id` were never given the same protection, even though these are the exact fields every `search()`/`get_all()`/`delete_all()` call filters on.

Concretely: `m.update(memory_id, metadata={"user_id": "someone_else"})` silently reassigns the memory into another tenant's scope — it disappears from the original tenant's `search()`/`get_all()`/`delete_all()` and appears under the new tenant's, without either tenant intending it.

This PR extends the existing `actor_id` re-assertion to also cover `user_id`/`agent_id`/`run_id`, mirroring the pattern `add()`'s `_build_filters_and_metadata()` already uses elsewhere in this file (real identifiers always win over caller-supplied metadata).

## Type of Change

## Breaking Changes

N/A. This only stops an unintended, undocumented side effect (`update()`'s docs describe it as mutating content/metadata/timestamp/expiration, never ownership/scope). Any caller not already relying on this bug is unaffected.

## Test Coverage

Added `test_update_memory_metadata_cannot_overwrite_identity_fields` (sync) and `test_async_update_memory_metadata_cannot_overwrite_identity_fields` (async) in `tests/memory/test_main.py`. Each builds an existing memory owned by `tenant_a`, calls `update()` with attacker-supplied `user_id`/`agent_id`/`run_id`/`actor_id` in `metadata`, and asserts the stored payload keeps the original identity fields (while a benign metadata key like `category` still passes through normally).

I verified both tests **fail on `main`** with `AssertionError: assert 'attacker_tenant' == 'tenant_a'` (i.e. they reproduce the exact cross-tenant reassignment), and **pass** with this fix.

Full local verification:
- `pytest tests/memory/test_main.py tests/test_main.py tests/test_memory.py` — 456 passed, 0 failed.
- `ruff check mem0/memory/main.py tests/memory/test_main.py` — clean.

## Checklist

- [ ] I have updated documentation if needed — N/A, this restores documented behavior rather than changing it

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -101,9 +101,6 @@ const ENTITY_PARAMS = [
   "runId",
 ];
 
-// Tenant-scoping fields that update() must never let caller-supplied metadata overwrite (issues #4490, #6277).
-const IDENTITY_KEYS = ["user_id", "agent_id", "run_id"];
-
 /**
  * Validates that no top-level entity parameters are passed in config.
  * @throws Error if entity params are found at top level
@@ -1944,14 +1941,9 @@ export class Memory {
       existingEmbeddings[newData] ||
       (await this.embedder.embed(newData, "update"));
 
-    // Identity keys are immutable after creation (issues #4490, #6277).
-    const safeMetadata = Object.fromEntries(
-      Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)),
-    );
-
     const newMetadata = {
       ...existingMemory.payload,
-      ...safeMetadata,
+      ...metadata,
       data: newData,
       hash: createHash("md5").update(newData).digest("hex"),
       textLemmatized: lemmatizeForBm25(newData),
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -101,6 +101,9 @@ const ENTITY_PARAMS = [
   "runId",
 ];
 
+// Tenant-scoping fields that update() must never let caller-supplied metadata overwrite (issues #4490, #6277).
+const IDENTITY_KEYS = ["user_id", "agent_id", "run_id"];
+
 /**
  * Validates that no top-level entity parameters are passed in config.
  * @throws Error if entity params are found at top level
@@ -1941,9 +1944,14 @@ export class Memory {
       existingEmbeddings[newData] ||
       (await this.embedder.embed(newData, "update"));
 
+    // Identity keys are immutable after creation (issues #4490, #6277).
+    const safeMetadata = Object.fromEntries(
+      Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)),
+    );
+
     const newMetadata = {
       ...existingMemory.payload,
-      ...metadata,
+      ...safeMetadata,
       data: newData,
       hash: createHash("md5").update(newData).digest("hex"),
       textLemmatized: lemmatizeForBm25(newData),
--- a/mem0-ts/src/oss/tests/memory.crud.test.ts
+++ b/mem0-ts/src/oss/tests/memory.crud.test.ts
@@ -207,31 +207,6 @@ describe("Memory - update()", () => {
       expect.objectContaining({ category: "hobbies", priority: "high" }),
     );
   });
-
-  test("metadata cannot overwrite or inject identity fields (issues #4490, #6277)", async () => {
-    const runId = `run_${Date.now()}`;
-    const addResult: SearchResult = await memory.add("Scoped memory", {
-      userId,
-      runId,
-      infer: false,
-    });
-    const id = addResult.results[0].id;
-
-    await memory.update(id, {
-      metadata: {
-        user_id: "attacker_tenant",
-        agent_id: "attacker_agent",
-        run_id: "attacker_run",
-        category: "sports",
-      },
-    });
-
-    const after: MemoryItem | null = await memory.get(id);
-    expect(after!.metadata!.user_id).toBe(userId);
-    expect(after!.metadata!.run_id).toBe(runId);
-    expect(after!.metadata!.agent_id).toBeUndefined();
-    expect(after!.metadata!.category).toBe("sports");
-  });
 });
 
 // ─── update() options: text / data / metadata / expirationDate ───
--- a/mem0-ts/src/oss/tests/memory.crud.test.ts
+++ b/mem0-ts/src/oss/tests/memory.crud.test.ts
@@ -207,6 +207,31 @@ describe("Memory - update()", () => {
       expect.objectContaining({ category: "hobbies", priority: "high" }),
     );
   });
+
+  test("metadata cannot overwrite or inject identity fields (issues #4490, #6277)", async () => {
+    const runId = `run_${Date.now()}`;
+    const addResult: SearchResult = await memory.add("Scoped memory", {
+      userId,
+      runId,
+      infer: false,
+    });
+    const id = addResult.results[0].id;
+
+    await memory.update(id, {
+      metadata: {
+        user_id: "attacker_tenant",
+        agent_id: "attacker_agent",
+        run_id: "attacker_run",
+        category: "sports",
+      },
+    });
+
+    const after: MemoryItem | null = await memory.get(id);
+    expect(after!.metadata!.user_id).toBe(userId);
+    expect(after!.metadata!.run_id).toBe(runId);
+    expect(after!.metadata!.agent_id).toBeUndefined();
+    expect(after!.metadata!.category).toBe("sports");
+  });
 });
 
 // ─── update() options: text / data / metadata / expirationDate ───
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1999,9 +1999,14 @@ class Memory(MemoryBase):
         new_metadata["created_at"] = existing_memory.payload.get("created_at")
         new_metadata["updated_at"] = datetime.now(timezone.utc).isoformat()
 
-        # actor_id is immutable after creation (issue #4490)
-        if "actor_id" in existing_memory.payload:
-            new_metadata["actor_id"] = existing_memory.payload["actor_id"]
+        # actor_id is immutable after creation (issue #4490). user_id/agent_id/run_id
+        # are likewise immutable: they scope every search()/get_all()/delete_all() call,
+        # so letting caller-supplied metadata overwrite them would silently move a
+        # memory into another tenant's scope and make it unreachable via the original
+        # tenant's delete_all().
+        for _identity_key in ("actor_id", "user_id", "agent_id", "run_id"):
+            if _identity_key in existing_memory.payload:
+                new_metadata[_identity_key] = existing_memory.payload[_identity_key]
 
         if data in existing_embeddings:
             embeddings = existing_embeddings[data]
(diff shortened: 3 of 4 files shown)


kartik-mem0:
## Changes Requested

Thanks for taking this on — #6277 is a real P0, and the shape of the fix here is right: extending the existing `actor_id` re-assertion pattern from #4490 to cover `user_id`/`agent_id`/`run_id` too, in both the sync and async `_update_memory`, is exactly the correct choke point. It also transitively closes the `server/` REST `PUT /memories/{id}` path for free, since everything routes through this one function. I dug in deep on this one (three independent review passes plus a dedicated adversarial check), and found one gap in the fix that's worth closing before merge.

### Strengths
- `mem0/memory/main.py:2004-2011` (sync) and `:3675-3682` (async) — byte-identical logic in both paths, no sync/async drift.
- Tests assert the real `vector_store.update` payload, not a mocked-away internal, for both sync and async.
- Red/green verified myself: reverting `main.py` to pre-PR `main` while keeping your tests makes both new tests fail with `AssertionError: assert 'attacker_tenant' == 'tenant_a'` — the exact symptom from #6277. Restoring the fix makes them pass.
- Full `tests/memory/test_main.py` suite (50 tests) and `ruff check` stay clean — no regressions to legitimate metadata merging.

### Issues

#### Critical (must fix)
1. `mem0/memory/main.py:2009-2011` (and async `:3680-3682`) — the loop only re-asserts an identity key `if _identity_key in existing_memory.payload`, so it only protects memories that already had that key set. A memory created with a subset of the three (e.g. only `user_id`, no `agent_id`/`run_id` — the normal case, since `add()`'s own validation only requires "at least one of" per `main.py:357-363`) has no existing key to guard against, so `update(metadata={"agent_id": "attacker_agent"})` injects a brand-new identity key straight into the stored payload unopposed. I reproduced this directly (`update()` on a `run_id`-only memory with injected `user_id`/`agent_id` in metadata → both survive into the persisted payload), and it was independently reproduced by three separate reviewers plus a dedicated skeptic pass — five confirmations in total, four of them via live execution of the actual merge logic. It's a narrower precondition than the reported repro (this grants an *additional* reachable scope rather than displacing the memory out of its original one), but it's the same root cause and the same severity class, and it means the new "these fields are immutable" comment at `main.py:2004-2008` isn't quite true yet. Cleanest fix: strip identity keys out of a copy of the incoming `metadata` dict *before* `new_metadata.update(metadata)`, mirroring the "real identifiers win" pattern `_build_filters_and_metadata` already uses at `main.py:331-355` — since `new_metadata` starts as `deepcopy(existing_memory.payload)` anyway, this closes both the overwrite and the injection case with less code and no dependency on which keys happened to pre-exist. A smaller patch also works if you'd rather keep the loop shape: add an `else: new_metadata.pop(_identity_key, None)` branch.

#### Important (should fix)
1. `mem0-ts/src/oss/src/memory/index.ts:1944-1952` (`updateMemory`) — the TypeScript OSS SDK has the identical class of bug, currently with *zero* identity-key protection (it never got the #4490 `actor_id` fix either), via the same unconditional `{...existingMemory.payload, ...metadata}` spread. This is legitimately out of scope for this PR (#6277 is filed against the Python SDK), but nothing currently tracks the TS side — worth opening a follow-up issue so the two SDKs don't drift further apart.

#### Minor (optional)
1. `mem0/memory/main.py:2009` / `:3680` — `_identity_key` has a leading underscore despite being used (not discarded); `identity_key` reads more idiomatically. Cosmetic only.

### Evidence
- **CI:** All real required checks pass (Python SDK build 3.10/3.11/3.12, CI Gate, changelog_check, check_changes, license/cla). Only the Vercel preview-deploy check is red, which is expected/advisory for fork PRs and not merge-blocking.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_main.py -v` → 50 passed. `ruff check` clean on both changed files.
- **Red-green proof:** regression tests fail with the exact reported symptom when the fix is reverted, pass when restored — confirmed this run.
- **Could NOT verify:** live behavior against a real Qdrant instance (verified via the mocked unit-test harness plus direct code tracing of `qdrant.py`'s exact-match filter semantics, which is enough to establish the exploit chain but wasn't run against a live DB).

### Assessment
Ready to merge: **With fixes** — the Critical item is a small, well-scoped addition to the loop you've already built (a couple of lines either way), not a rewrite. Happy to move fast on a re-review once it's in. The instinct to reuse the `actor_id` pattern and cover both sync and async was exactly right — this just needs to go one step further so it holds for partially-scoped memories too.

okaditya84:
Thanks for the deep review — that gap is real and the injection case is a good catch.

**Critical (fixed):** Switched to the approach you suggested — stripping the identity keys from a copy of the incoming \`metadata\` before merging into \`new_metadata\` (which already starts as \`deepcopy(existing_memory.payload)\`), rather than the loop that only re-asserted keys that already existed. Added two new regression tests (sync + async) that construct a \`run_id\`-only memory (no pre-existing \`user_id\`/\`agent_id\`, the normal case per \`add()\`'s "at least one of" validation) and assert an attacker-supplied \`user_id\`/\`agent_id\`/\`actor_id\` in \`metadata\` does not land in the persisted payload. Confirmed these fail on the previous loop-based version of this PR and pass with the strip-before-merge fix. Full \`tests/memory/test_main.py\` (52 tests) still green, \`ruff check\` clean.

**Important (TS SDK):** Verified — \`updateMemory()\` in \`mem0-ts/src/oss/src/memory/index.ts\` does the same unconditional \`{...existingMemory.payload, ...metadata}\` spread with zero identity-key protection (never got the #4490 fix either). Opened #6367 to track it separately since it's TS-only.

**Minor:** the loop (and its slightly-off \`_identity_key\` naming) is gone entirely with this approach, so that's moot now.

kartik-mem0:

### What this PR does
Verified against `49cc7ff3` (new head): the partial-identity-injection gap from the last review is closed — `update()` now strips `_IDENTITY_KEYS` from a copy of the incoming metadata before merging it into `new_metadata` (which already starts as `deepcopy(existing_memory.payload)`), instead of only re-asserting keys that already existed on the memory.

### Strengths
- `mem0/memory/main.py:1997-2009` (sync) / `:3670-3682` (async) — strip-before-merge closes both the overwrite AND injection cases unconditionally, regardless of which identity keys the memory was created with. Sync/async stay byte-identical.
- `tests/memory/test_main.py` — new `_RUN_ID_ONLY_UPDATE_PAYLOAD` fixture + 2 new tests cover exactly the case the old fixture couldn't: a `run_id`-only memory (the normal minimum, per `add()`'s own "at least one of" validation), attacked with a fabricated `user_id`/`agent_id`/`actor_id`.
- The stale "are likewise immutable" comment and the old `_identity_key` loop are both gone — no leftover dead code.
- `_IDENTITY_KEYS = ENTITY_PARAMS | {"actor_id"}` reuses the existing frozenset instead of a second hardcoded key list.

### Issues
Minor (optional): `_IDENTITY_KEYS`'s frozenset-ness isn't pinned by a test — a future `set()` refactor would silently drop the immutability with nothing to catch it. Not blocking.

Important (cross-package, does not block this PR): the companion TS PR #6343 still uses the "re-pin only if the key already existed" pattern this PR just moved away from — its `updateMemory()` spreads `...metadata` before three `...(existingMemory.payload.<key> !== undefined && {...})` guards. Since TS `add()`/`getAll()` allow `run_id`-only memories too, those guards are no-ops exactly when it matters, and the same injection this PR fixes in Python is still open there. Issue #6367 already writes up the right fix (strip-before-merge) — might be worth pointing #6343 at this PR's `49cc7ff3` as the reference implementation before treating the bug class as closed repo-wide. Not asking to hold either PR for this — Python and TS are separate, complementary fixes for separate issues.

### Evidence
- **CI:** all required checks green (Python SDK build 3.10/3.11/3.12, CI Gate, changelog_check, check_changes, license/cla). Vercel preview-deploy is red — advisory only, ignored per convention.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_main.py -v` → 52 passed.
- **Red-green proof (3-way):**
  1. impl reverted to the previously-reviewed SHA `116bf7ec` → the 2 original overwrite tests pass, the 2 new injection tests fail (`AssertionError: assert 'user_id' not in {..., 'user_id': 'attacker_tenant', ...}`) — exact reproduction of the gap flagged last review.
  2. impl reverted to pristine pre-PR `main` → all 4 identity tests fail, matching the originally reported symptom.
  3. impl restored to new head `49cc7ff3` → all 4 pass. `ruff check` on both changed files: clean.
- **Linked issue:** #6277 (P0) — PR addresses the reported root cause, including the broader injection variant found during the last review.
- **Cross-package check:** read `mem0-ts/src/oss/src/memory/index.ts` directly — TS OSS `updateMemory()` still doesn't close this bug class (see Important note above; tracked by issue #6367 and PR #6343).
- **Could NOT verify:** live vector-store behavior beyond the mocked `vector_store.update()` call-args assertions in the new tests — matches this test file's existing convention.

### Assessment
Strip-before-merge closes the previously-Critical gap unconditionally — verified via 3-way red-green, not just CI — sync/async stay in lockstep, and the new tests target exactly the scenario that slipped through the first attempt. Nice turnaround on this one.

kartik-mem0 (approved):

PR #6290, opened 2026-07-13
fix(ts-oss/weaviate): return full payload from get() and list()

### LINKED ISSUE #6289: bug(ts-oss/weaviate): get() and list() drop createdAt/updatedAt and custom metadata

TypeScript OSS SDK - Weaviate vector store

### Describe the bug

The Weaviate adapter restricts reads to a fixed property list, so `get()` and `getAll()`/`list()` come back missing the timestamps and any custom metadata.

`weaviate.ts` defines a hardcoded `RETURN_PROPERTIES` list and passes it as `returnProperties` on both `fetchObjectById` (used by `get`) and `fetchObjects` (used by `list`). Weaviate only returns the named properties when `returnProperties` is set, so any field not in that list is dropped from the result.

The problem is that the memory layer writes timestamps in camelCase (`createdAt`, `updatedAt`), but the list names them in snake_case (`created_at`, `updated_at`). It also cannot possibly name the arbitrary custom metadata keys a caller stores. So both are silently discarded on read.

`search()` does not set `returnProperties`, so it returns everything and is unaffected. That makes `get`/`list` inconsistent with `search` for the exact same record.

### How to reproduce

1. Configure `Memory` with the Weaviate vector store.
2. `add("hello", { userId: "u1", metadata: { priority: "high" } })`.
3. `get(id)` (or `getAll({ userId: "u1" })`).

Expected: the returned item has `createdAt`, `updatedAt`, and `priority`.
Actual: `createdAt` and `updatedAt` are `undefined` and `priority` is missing. The same record fetched through `search()` has them.

### Expected behavior

`get()` and `list()` should return the full stored payload, matching `search()` and the other providers (pgvector, qdrant, redis, etc.), so timestamps and custom metadata survive the round trip.

abhay-codes07:
## Linked Issue

Closes #6289

## Description

The Weaviate adapter pins its reads to a fixed property list, which drops
timestamps and custom metadata from `get()` and `getAll()`.

`get()` and `list()` passed a hardcoded `returnProperties` list to
`fetchObjectById`/`fetchObjects`. When `returnProperties` is set, Weaviate only
returns the named fields. That list named the timestamps in snake_case
(`created_at`/`updated_at`), but the memory layer writes them in camelCase
(`createdAt`/`updatedAt`), and it obviously cannot name the arbitrary custom
metadata keys a caller stores. So both were silently discarded on read, and
`payload.createdAt`/`payload.updatedAt` came back `undefined`.

`search()` never set `returnProperties`, so it returned the full payload and was
unaffected. That left `get`/`list` inconsistent with `search` for the very same
record, and out of line with the other providers (pgvector, qdrant, redis) that
all return the whole payload.

## What changed

- `get()` and `list()` no longer restrict `returnProperties`, so every stored
  property comes back (matching `search()`).
- The property constant is kept only to seed the collection schema on create;
  Weaviate auto-schema extends it with the real payload fields, so it no longer
  needs to enumerate timestamps or metadata.

## Testing

Added `weaviate.unit.test.ts` with a mocked `weaviate-client` whose reads honor
`returnProperties` the way the real client does. The tests assert that `get()`
and `list()` return `createdAt`, `updatedAt`, and a custom metadata key. Both
fail against the old code (timestamps `undefined`) and pass with the fix. Ran
the OSS suite and typecheck locally.

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/weaviate.ts
+++ b/mem0-ts/src/oss/src/vector_stores/weaviate.ts
@@ -14,18 +14,11 @@ interface WeaviateConfig extends VectorStoreConfig {
   embeddingModelDims: number;
 }
 
-const RETURN_PROPERTIES = [
-  "ids",
-  "hash",
-  "metadata",
-  "data",
-  "created_at",
-  "category",
-  "updated_at",
-  "user_id",
-  "agent_id",
-  "run_id",
-];
+// Seeds the collection schema on first create. Weaviate auto-schema extends it
+// with whatever the memory layer actually writes (camelCase timestamps, the
+// lemmatized text, and any custom metadata keys), so reads must not be pinned
+// to this list.
+const BASE_PROPERTIES = ["hash", "data", "user_id", "agent_id", "run_id"];
 
 export class WeaviateDB implements VectorStore {
   private _config: WeaviateConfig;
@@ -107,7 +100,7 @@ export class WeaviateDB implements VectorStore {
     if (!exists) {
       await this._client.collections.create({
         name: collectionName,
-        properties: RETURN_PROPERTIES.map((name) => ({
+        properties: BASE_PROPERTIES.map((name) => ({
           name,
           dataType: "text" as const,
         })),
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss weaviate fix with tests. get() and list() pinned returnProperties to a fixed list, so the camelCase createdAt/updatedAt and any custom metadata got dropped on read, while search() returned them fine. They now return the full payload like the other stores. Quick review when you can, thanks.

PR #6294, opened 2026-07-13
fix(core): carry updated_at into batch add history records

### LINKED ISSUE #6293: bug(core): batch add pipeline records ADD history with updated_at as NULL

Python SDK - core memory (history)

### Describe the bug

Memories created through the default `add()` pipeline get an ADD history row with `updated_at` (and `actor_id`/`role`) written as NULL, while memories created through the single-add path get `updated_at == created_at`. The two paths disagree for the same event type.

The default `add(infer=True)` flow goes through the V3 batch pipeline in `_add_to_vector_store`. When it builds the batch history records, the dict only carries `created_at`:

```python
history_records = [
    {
        "memory_id": r[0],
        "old_memory": None,
        "new_memory": r[1],
        "event": "ADD",
        "created_at": r[3].get("created_at"),
        "is_deleted": 0,
    }
    for r in records
]
```

`batch_add_history` then inserts `record.get("updated_at")` etc. as NULL. But the payload written to the vector store already sets `updated_at = created_at`, and the single-memory path `_create_memory` explicitly passes `updated_at`, `actor_id`, and `role` into `add_history`. So the intent is that an ADD event carries `updated_at == created_at`; the batch path just drops it.

Both the sync (`Memory`) and async (`AsyncMemory`) pipelines have the same omission.

### How to reproduce

1. `add(messages=[...], user_id="u1")` (default infer).
2. `history(memory_id=...)`.

Expected: the ADD entry has `updated_at` equal to `created_at`.
Actual: `updated_at` is `null` (as are `actor_id` and `role`). The same memory added with `infer=False` returns `updated_at == created_at`.

### Expected behavior

The batch ADD history records should carry `updated_at` (and `actor_id`/`role` when present), matching `_create_memory`, so history is consistent regardless of which internal path created the memory.

abhay-codes07:
## Linked Issue

Closes #6293

## Description

Memories created through the default `add()` pipeline ended up with an ADD
history row whose `updated_at`, `actor_id`, and `role` were NULL, while the
single-add path recorded `updated_at == created_at`. Same event type, two
different results depending on which internal path ran.

The default `add(infer=True)` flow goes through the V3 batch pipeline in
`_add_to_vector_store`. When it assembled the batch history records it only
carried `created_at`, so `batch_add_history` inserted the rest as NULL. That is
inconsistent with:

- the vector payload written for the same memory, which sets
  `updated_at = created_at`, and
- `_create_memory` (the single/`infer=False` path), which explicitly passes
  `updated_at`, `actor_id`, and `role` into `add_history`.

So the intent is clearly that an ADD event carries `updated_at == created_at`;
the batch path just dropped it. Anyone reading `history()` saw
`updated_at: null` for the ADD event of essentially every normally-created
memory.

## What changed

- The batch history records (and the one-by-one fallback) now include
  `updated_at`, `actor_id`, and `role`, read from the same payload that was
  persisted to the vector store.
- Applied to both `Memory` and `AsyncMemory`, which had the identical omission.

## Testing

Added sync and async tests that drive the batch pipeline and assert the ADD
history record carries `updated_at == created_at`. Both fail against the old
code (the key is absent) and pass with the fix. Full `tests/memory/test_main.py`
passes (50 tests); ruff clean.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1027,7 +1027,10 @@ class Memory(MemoryBase):
                 "new_memory": r[1],
                 "event": "ADD",
                 "created_at": r[3].get("created_at"),
+                "updated_at": r[3].get("updated_at"),
                 "is_deleted": 0,
+                "actor_id": r[3].get("actor_id"),
+                "role": r[3].get("role"),
             }
             for r in records
         ]
@@ -1037,7 +1040,16 @@ class Memory(MemoryBase):
             # Fallback: add one by one
             for hr in history_records:
                 try:
-                    self.db.add_history(hr["memory_id"], None, hr["new_memory"], "ADD", created_at=hr.get("created_at"))
+                    self.db.add_history(
+                        hr["memory_id"],
+                        None,
+                        hr["new_memory"],
+                        "ADD",
+                        created_at=hr.get("created_at"),
+                        updated_at=hr.get("updated_at"),
+                        actor_id=hr.get("actor_id"),
+                        role=hr.get("role"),
+                    )
                 except Exception as e:
                     logger.error(f"Failed to add history for {hr['memory_id']}: {e}")
 
@@ -2661,7 +2673,10 @@ class AsyncMemory(MemoryBase):
                 "new_memory": r[1],
                 "event": "ADD",
                 "created_at": r[3].get("created_at"),
+                "updated_at": r[3].get("updated_at"),
                 "is_deleted": 0,
+                "actor_id": r[3].get("actor_id"),
+                "role": r[3].get("role"),
             }
             for r in records
         ]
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small core fix with tests. The V3 batch add pipeline recorded the ADD history event with updated_at (and actor_id/role) as NULL, while _create_memory passes them through, so history was inconsistent depending on the path. Aligned the batch records with the single-add path for both Memory and AsyncMemory. Quick review when you can, thanks.

kartik-mem0:
Thanks for turning this around so quickly, and apologies for closing it after the work you put in.

The change does what the issue asked for, but we have decided to align the two paths the other way, so I am closing this alongside #6293.

Short version: the default inferred pipeline is single-pass ADD-only (https://mem0.ai/research), so an ADD event has no meaningful `updated_at`. Keeping the column NULL until a real `update()` or `delete()` happens is the semantics we want to preserve, because it is the only way to distinguish "never revised" from "revised at creation time". Full reasoning is in #6293.

The test you added is the right shape, and it will be a useful reference if we make the reverse change. Thanks again for the care in both the issue and the PR.

PR #6296, opened 2026-07-13
fix(ts-oss): set updatedAt on infer:false memories

### LINKED ISSUE #6295: bug(ts-oss): infer:false memories are stored without updatedAt

TypeScript OSS SDK - core memory

### Describe the bug

Memories added with `infer: false` are stored without `updatedAt`, so reading them back through `get()`, `getAll()`, or `search()` returns `updatedAt: undefined`.

The `infer: false` path calls `createMemory`, which builds the stored payload with `createdAt` but never sets `updatedAt`:

```ts
const memoryMetadata = {
  ...metadata,
  data,
  hash: createHash("md5").update(data).digest("hex"),
  textLemmatized: lemmatizeForBm25(data),
  createdAt: new Date().toISOString(),
};
```

Every other write path sets it: the `infer: true` batch pipeline writes `createdAt` and `updatedAt`, and `updateMemory` sets `updatedAt` on edits. The Python SDK's `_create_memory` also sets `updated_at = created_at` for a new memory. So an `infer: false` memory violates the contract that `updatedAt` is always populated.

### How to reproduce

1. `add("I like tea", { userId: "u1", infer: false })`.
2. `get(id)` (or `getAll({ filters: { user_id: "u1" } })`).

Expected: `updatedAt` is set and equals `createdAt`.
Actual: `updatedAt` is `undefined`.

### Expected behavior

A freshly created `infer: false` memory should have `updatedAt` populated and equal to `createdAt`, matching the batch pipeline, `updateMemory`, and the Python SDK.

abhay-codes07:
## Linked Issue

Closes #6295

## Description

Memories added with `infer: false` were stored without `updatedAt`, so reading
them back through `get()`, `getAll()`, or `search()` returned
`updatedAt: undefined`.

The `infer: false` path calls `createMemory`, which built the payload with
`createdAt` but never set `updatedAt`. Every other write path sets it:

- the `infer: true` batch pipeline writes both `createdAt` and `updatedAt`,
- `updateMemory` sets `updatedAt` on edits,
- and Python's `_create_memory` sets `updated_at = created_at` for a new memory.

So `infer: false` memories were the only ones missing it, violating the contract
that `updatedAt` is always populated.

## What changed

- `createMemory` now sets `updatedAt` equal to `createdAt` on the stored payload
  and passes it to `addHistory`, so the ADD event carries it too.

## Testing

Added `memory.infer-false-timestamp.test.ts` asserting `get()` and `getAll()`
return `updatedAt` (equal to `createdAt`) for `infer: false` memories. Both fail
against the old code (undefined) and pass with the fix. `memory.crud` suite
(50 tests) still passes; prettier clean.

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1896,12 +1896,17 @@ export class Memory {
     const embedding =
       existingEmbeddings[data] || (await this.embedder.embed(data, "add"));
 
+    const now = new Date().toISOString();
     const memoryMetadata = {
       ...metadata,
       data,
       hash: createHash("md5").update(data).digest("hex"),
       textLemmatized: lemmatizeForBm25(data),
-      createdAt: new Date().toISOString(),
+      createdAt: now,
+      // A freshly created memory has updatedAt == createdAt. The batch pipeline
+      // and updateMemory both set it; without it here, infer:false memories come
+      // back from get()/getAll()/search() with updatedAt undefined.
+      updatedAt: now,
     };
 
     await this.vectorStore.insert([embedding], [memoryId], [memoryMetadata]);
@@ -1911,6 +1916,7 @@ export class Memory {
       data,
       "ADD",
       memoryMetadata.createdAt,
+      memoryMetadata.updatedAt,
     );
 
     return memoryId;
--- a/mem0-ts/src/oss/tests/memory.infer-false-timestamp.test.ts
+++ b/mem0-ts/src/oss/tests/memory.infer-false-timestamp.test.ts
@@ -1,0 +1,94 @@
+/**
+ * infer:false add() must persist updatedAt on the stored payload, matching the
+ * infer:true batch pipeline and updateMemory. Without it, get()/getAll()/search()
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss fix with tests. infer:false memories were stored without updatedAt, so get/getAll/search returned it undefined. createMemory now sets updatedAt equal to createdAt like the batch pipeline, updateMemory, and Python. Quick review when you can, thanks.

kartik-mem0:
Thanks for the PR, and apologies for closing it after the work you put in.

Closing alongside #6295. The contract we are aligning on is the reverse of the one this PR implements: `updatedAt` should be populated only by an actual `update()`, never at creation.

Short version: the default inferred pipeline is single-pass ADD-only (https://mem0.ai/research), so a newly created memory has no meaningful `updatedAt`. Leaving it unset is the only way to distinguish "never revised" from "revised at creation time". That makes the `infer: false` path already correct as-is, and puts the work on the inferred path that currently sets it. Full reasoning is in #6295 and #6494.

Thanks again, the write-path analysis in the description was genuinely useful in deciding which way to align.

PR #6298, opened 2026-07-13
fix(ts-oss/weaviate): run keyword BM25 over the lemmatized field

### LINKED ISSUE #6297: bug(ts-oss/weaviate): keyword search runs BM25 over raw data field with a lemmatized query

TypeScript OSS SDK - Weaviate vector store

### Describe the bug

Weaviate keyword search runs BM25 against the raw `data` field using an already-lemmatized query, so stemmed query tokens miss the unstemmed stored text and keyword recall silently degrades.

`Memory.search()` lemmatizes the query before calling `keywordSearch` (it passes `queryLemmatized`), and it stores the lemmatized text on each memory under `textLemmatized`. Every other keyword-capable store matches the lemmatized query against the lemmatized field:

- pgvector queries `payload->>'textLemmatized'`
- the in-memory store tokenizes `payload.textLemmatized`

Weaviate instead BM25-searches the original `data` property:

```ts
const result = await this._col.query.bm25(query, {
  queryProperties: ["data"],
  ...
});
```

Weaviate's default tokenizer lowercases but does not stem, so a stemmed query token does not match the unstemmed stored text.

### How to reproduce

1. Store memory text "I am studying French" (Weaviate store).
2. `search("studies", ...)`.

The query is lemmatized to a stem like `studi`. BM25 over the raw `data` field looks for `studi`/`studies` and does not match `studying`, so there is no keyword hit and no BM25 contribution, whereas pgvector and the in-memory store (matching lemmatized against lemmatized) do match.

### Expected behavior

Weaviate keyword search should match the lemmatized query against the lemmatized field (`textLemmatized`), consistent with the other stores, so hybrid search keyword recall behaves the same regardless of backend.

abhay-codes07:
## Linked Issue

Closes #6297

## Description

Weaviate keyword search ran BM25 against the raw `data` field while the memory
layer hands it an already-lemmatized query, so stemmed query tokens missed the
unstemmed stored text and keyword recall degraded silently.

`Memory.search()` lemmatizes the query before calling `keywordSearch` (it passes
`queryLemmatized`) and stores the lemmatized text on each memory under
`textLemmatized`. The other keyword-capable stores match lemmatized-to-lemmatized:
pgvector queries `payload->>'textLemmatized'` and the in-memory store tokenizes
`payload.textLemmatized`. Weaviate was the outlier, querying `data`, whose default
tokenizer lowercases but does not stem, so e.g. a query "studies" (stemmed to
`studi`) would not match stored "studying".

## What changed

- Weaviate `keywordSearch` now targets `textLemmatized` instead of `data`, so the
  lemmatized query matches the lemmatized field, consistent with the other stores.

## Testing

Added a unit test (mocked `weaviate-client`) asserting `keywordSearch` calls BM25
with `queryProperties: ["textLemmatized"]`. It fails against the old code
(`["data"]`) and passes with the fix. Typecheck and prettier clean.

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/weaviate.ts
+++ b/mem0-ts/src/oss/src/vector_stores/weaviate.ts
@@ -165,8 +165,12 @@ export class WeaviateDB implements VectorStore {
     filters?: SearchFilters,
   ): Promise<VectorStoreResult[] | null> {
     await this.initialize();
+    // The memory layer lemmatizes the query before calling keywordSearch, and it
+    // stores the lemmatized text under textLemmatized. Match BM25 against that
+    // field (as pgvector and the in-memory store do); running it over the raw
+    // data field means a stemmed query token never matches the unstemmed text.
     const result = await this._col.query.bm25(query, {
-      queryProperties: ["data"],
+      queryProperties: ["textLemmatized"],
       limit: topK ?? 10,
       filters: this._buildFilters(filters),
       returnMetadata: ["score"],
--- a/mem0-ts/src/oss/tests/weaviate.keyword.unit.test.ts
+++ b/mem0-ts/src/oss/tests/weaviate.keyword.unit.test.ts
@@ -1,0 +1,80 @@
+/**
+ * Weaviate keyword search must run BM25 against the lemmatized field.
+ *
+ * The memory layer lemmatizes the query before calling keywordSearch and stores
+ * the lemmatized text under `textLemmatized`. Matching a stemmed query against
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss weaviate fix with tests. keyword search ran BM25 over the raw data field, but the memory layer passes a lemmatized query and stores the lemmatized text under textLemmatized, so stemmed tokens missed. Now it queries textLemmatized like pgvector and the in-memory store. Quick review when you can, thanks.

PR #6302, opened 2026-07-13
fix(embeddings/huggingface): forward api_key on base_url path

### LINKED ISSUE #6301: bug(embeddings/huggingface): api_key ignored on base_url path, endpoint auth dropped
## What happens

```python
```

## Confirmed from code/docs

## Repro

```python
import os

    api_key="hf_myEndpointToken",
))
```

## Proposed fix

```python
```

OfficialAbhinavSingh:
### Summary

On the `huggingface_base_url` path (TEI server / HF Inference Endpoint), `HuggingFaceEmbedding` built the OpenAI-compatible client with only `base_url` and never forwarded `api_key`. A user-supplied `api_key` was silently dropped, and with no `OPENAI_API_KEY` present the client raised `openai.OpenAIError: Missing credentials` at construction. (It could also leak a real `OPENAI_API_KEY` as the bearer token sent to whatever host `huggingface_base_url` points at.)

This forwards the key with `config > env > default` precedence. The exact 3-tier precedence matches the TypeScript SDK (`mem0-ts/src/oss/src/embeddings/huggingface.ts`) and the documented `HUGGINGFACE_API_KEY` / `"hf"` fallback. The sibling `LMStudioEmbedding` follows the same "always pass a non-empty key" shape (though it is 2-tier: config then default, no env fallback).

### Changes

- `mem0/embeddings/huggingface.py`: pass `api_key = self.config.api_key or os.getenv("HUGGINGFACE_API_KEY") or "hf"` to `OpenAI(...)` on the base_url branch.
- `tests/embeddings/test_huggingface_embeddings.py`: add 3 regression tests (config key, env fallback, default `"hf"`). The existing base_url test asserts the key and `delenv`s `HUGGINGFACE_API_KEY` so it is deterministic regardless of the local environment.
- `docs/components/embedders/models/huggingface.mdx`: add the `api_key` row to the Python config table (it was only in the TypeScript tab).

### Verification

- `ruff check` clean on the changed Python files.
- `pytest tests/embeddings/test_huggingface_embeddings.py` -> 14 passed, both with and without `HUGGINGFACE_API_KEY` set.
- `docs/llms.txt` reported in sync after the docs change.
- End-to-end against a real OpenAI client: config key, `HUGGINGFACE_API_KEY`, and the `"hf"` default all resolve correctly, and the no-`OPENAI_API_KEY` construction no longer crashes.

Closes #6301

### DIFF
--- a/docs/components/embedders/models/huggingface.mdx
+++ b/docs/components/embedders/models/huggingface.mdx
@@ -99,6 +99,7 @@ Here are the parameters available for configuring the Hugging Face embedder:
 | `embedding_dims` | Dimensions of the embedding model | `selected_model_dimensions` |
 | `model_kwargs` | Additional arguments for the model | `None` |
 | `huggingface_base_url` | URL to connect to Text Embeddings Inference (TEI) API | `None` |
+| `api_key` | API key for the TEI or OpenAI-compatible endpoint; falls back to the `HUGGINGFACE_API_KEY` env var | `"hf"` |
 </Tab>
 <Tab title="TypeScript">
 | Parameter | Description | Default Value |
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Literal, Optional
 
 from openai import OpenAI
@@ -17,7 +18,8 @@ class HuggingFaceEmbedding(EmbeddingBase):
         super().__init__(config)
 
         if self.config.huggingface_base_url:
-            self.client = OpenAI(base_url=self.config.huggingface_base_url)
+            api_key = self.config.api_key or os.getenv("HUGGINGFACE_API_KEY") or "hf"
+            self.client = OpenAI(base_url=self.config.huggingface_base_url, api_key=api_key)
             self.config.model = self.config.model or "tei"
         else:
             self.config.model = self.config.model or "multi-qa-MiniLM-L6-cos-v1"
--- a/tests/embeddings/test_huggingface_embeddings.py
+++ b/tests/embeddings/test_huggingface_embeddings.py
@@ -94,13 +94,37 @@ def test_embed_with_huggingface_base_url():
         embedder = HuggingFaceEmbedding(config)
         result = embedder.embed("Hello from custom endpoint")
 
-        mock_openai.assert_called_once_with(base_url="http://localhost:8080")
+        mock_openai.assert_called_once_with(base_url="http://localhost:8080", api_key="hf")
         mock_client.embeddings.create.assert_called_once_with(
             input="Hello from custom endpoint",
             model="my-custom-model",
             truncate=True,
         )
         assert result == [0.1, 0.2, 0.3]
+
+
+def test_base_url_forwards_config_api_key(monkeypatch):
+    monkeypatch.delenv("HUGGINGFACE_API_KEY", raising=False)
+    config = BaseEmbedderConfig(huggingface_base_url="http://localhost:8080", api_key="hf_configToken")
+    with patch("mem0.embeddings.huggingface.OpenAI") as mock_openai:
+        HuggingFaceEmbedding(config)
+        assert mock_openai.call_args.kwargs["api_key"] == "hf_configToken"
+
+
+def test_base_url_uses_env_api_key_when_config_unset(monkeypatch):
+    monkeypatch.setenv("HUGGINGFACE_API_KEY", "hf_envToken")
+    config = BaseEmbedderConfig(huggingface_base_url="http://localhost:8080")
+    with patch("mem0.embeddings.huggingface.OpenAI") as mock_openai:
+        HuggingFaceEmbedding(config)
+        assert mock_openai.call_args.kwargs["api_key"] == "hf_envToken"
+
+
+def test_base_url_defaults_api_key_when_unset(monkeypatch):
+    monkeypatch.delenv("HUGGINGFACE_API_KEY", raising=False)
+    config = BaseEmbedderConfig(huggingface_base_url="http://localhost:8080")
+    with patch("mem0.embeddings.huggingface.OpenAI") as mock_openai:
+        HuggingFaceEmbedding(config)
+        assert mock_openai.call_args.kwargs["api_key"] == "hf"
 
 
 def test_embed_batch_sentence_transformer(mock_sentence_transformer):
--- a/tests/embeddings/test_huggingface_embeddings.py
+++ b/tests/embeddings/test_huggingface_embeddings.py
@@ -72,7 +72,8 @@ def test_embed_with_custom_embedding_dims(mock_sentence_transformer):
     assert result == [1.0, 1.1, 1.2]
 
 
-def test_embed_with_huggingface_base_url():
+def test_embed_with_huggingface_base_url(monkeypatch):
+    monkeypatch.delenv("HUGGINGFACE_API_KEY", raising=False)
     config = BaseEmbedderConfig(
         huggingface_base_url="http://localhost:8080",
         model="my-custom-model",


kartik-mem0:

### What this PR does
Fixes `HuggingFaceEmbedding` silently dropping `api_key` on the TEI / HF Inference Endpoint (`huggingface_base_url`) path — the OpenAI-compatible client either leaked a real `OPENAI_API_KEY` as the bearer token or raised `Missing credentials` when that env var was unset. Forwards it with `config > HUGGINGFACE_API_KEY env > "hf"` precedence.

### Strengths
- `mem0/embeddings/huggingface.py:21-22` — precedence exactly matches `mem0-ts/src/oss/src/embeddings/huggingface.ts:39` (`config > HUGGINGFACE_API_KEY env > "hf"`) — verified by reading both files directly, not just the PR description.
- `tests/embeddings/test_huggingface_embeddings.py:106-129` — 3 new tests independently exercise each branch of the precedence chain, each correctly `monkeypatch`-isolated.
- Swept `mem0/embeddings/*.py` and `mem0/llms/*.py` for the same `OpenAI(base_url=...)` pattern — this was the only unfixed instance, nothing left half-fixed.
- Incidental security win: previously, omitting `api_key` let openai-python silently fall back to the real `OPENAI_API_KEY` env var as the bearer token sent to whatever host `huggingface_base_url` points at — this closes that leak.
- Zero new dependencies, one-time construction-path change only — no hot-path cost, and the sentence-transformers local path (the `else` branch) is untouched.

### Issues
Minor (optional):
1. `tests/embeddings/test_huggingface_embeddings.py:75` — `test_embed_with_huggingface_base_url` asserts `api_key="hf"` but doesn't `monkeypatch.delenv("HUGGINGFACE_API_KEY")` like the 3 new tests do — will fail spuriously if that var happens to be set locally. One-line fix.
2. `docs/components/embedders/models/huggingface.mdx` — the Python config table has no `api_key` row; the TypeScript tab a few lines down already documents `apiKey`/`HUGGINGFACE_API_KEY`/`"hf"` — worth a matching row now that Python's runtime actually honors it.
3. The description says this mirrors `LMStudioEmbedding` — true for the "always pass a non-empty key" shape, but `lmstudio.py` is only 2-tier (config > default, no env fallback); the exact 3-tier precedence match is with the TS SDK. Worth tightening the description, not a code change.

### Evidence
- **CI:** Vercel preview-deploy is red — expected/advisory on fork PRs (no deploy perms), not merge-blocking. The real required check, "CI Gate," shows `conclusion: action_required` — GitHub is holding it for a one-time maintainer approval (standard for a first-time/fork contributor), so it hasn't actually run yet. Used local targeted tests below as this run's real signal.
- **Targeted tests:** `hatch -e dev_py_3_12 run pytest tests/embeddings/test_huggingface_embeddings.py -v --tb=short` → 14 passed in 39.32s. Lint: `hatch -e dev_py_3_12 run ruff check mem0/embeddings/huggingface.py` → All checks passed!
- **Red-green proof:** reverted `mem0/embeddings/huggingface.py` to `main` (kept the PR's tests) → 4 failed / 10 passed, e.g. `AssertionError: Expected: OpenAI(base_url='http://localhost:8080', api_key='hf') / Actual: OpenAI(base_url='http://localhost:8080')`. Restored the PR's implementation → 14 passed again. ✓
- **Linked issue:** #6301 — fixes the exact reported root cause (same line, same repro, same proposed fix as the issue itself).
- **Cross-package check:** TS counterpart (`mem0-ts/src/oss/src/embeddings/huggingface.ts:39`) already carries this exact 3-tier precedence — Python now matches it exactly.
- **Could NOT verify:** live behavior against a real TEI server / HF Inference Endpoint — mock-level tests + code tracing only, which is the right bar for a client-construction fix like this one.

### Assessment
Small, correctly root-caused fix with a genuine red-green proof and exact TS-SDK parity. The CI Gate run will just need the one-time approval click before it shows green.

OfficialAbhinavSingh:
Thanks for the thorough review. Addressed all three points in 711c7743: delenv'd `HUGGINGFACE_API_KEY` in the base_url test so it asserts the `"hf"` default deterministically, added the `api_key` row to the Python config table in `huggingface.mdx`, and corrected the precedence note (the exact 3-tier match is the TS SDK; `LMStudioEmbedding` is 2-tier). Tests: 14 passed both with and without the env var set; `docs/llms.txt` reports in sync.

ErenAta16:
Precedence logic matches the TS sibling and the documented contract exactly, and the three new tests (config set, env fallback, default fallback) cover each branch independently rather than just the happy path. Also good catch updating the existing base_url test's mock assertion, that one would have silently masked the missing api_key kwarg otherwise. Looks correct.

OfficialAbhinavSingh:
@kartik-mem0 following up since this was marked ready to merge, just checking if the CI Gate approval click is still pending on this one.

kartik-mem0:
Verified against `711c7743`: all three points from the review are fixed, and I checked each against the actual diff rather than taking the commit message's word for it.

1. `tests/embeddings/test_huggingface_embeddings.py` — `test_embed_with_huggingface_base_url` now takes `monkeypatch` and `delenv`s `HUGGINGFACE_API_KEY`. Confirmed this wasn't cosmetic: I swapped the old test back in against the new code and ran both with `HUGGINGFACE_API_KEY` set to a bogus value — the old version fails exactly the way it would have for anyone with that var set locally (`Expected: api_key='hf' / Actual: api_key='some_other_value'`), the new version passes. Good fix.
2. `huggingface.mdx` — the `api_key` row is in the Python table now, matching the TS tab's wording.
3. The precedence note in the description reads correctly now — TS SDK as the exact 3-tier match, `LMStudioEmbedding` correctly called out as 2-tier.

`huggingface.py` itself is untouched by this commit, so the original fix (and its red-green proof) carries forward unchanged. Full suite: 14 passed at the new head, ruff clean.

Ready for squash-merge on our end. The CI Gate run is still sitting on `action_required` — that's the one-time approval click on our side, not something in your court, sorry for the delay on it.

OfficialAbhinavSingh:
Appreciate you re-checking each point against the diff — no worries at all on the CI Gate timing, no rush on my end. Thanks for the thorough reviews throughout.

OfficialAbhinavSingh:
Heads up that the head sha moved since your re-verify: this is now at cbd9928e, rebased onto current main on 26 Jul.

The rebase was only to clear `changelog_check`, which had gone red because the branch was cut pre-2.0.14 and the gate compares the base tip's pyproject version against the head's. No content change: the two commits keep the same patch-ids they had at 711c7743 (b170e282, 21698c70), so what you verified is what is there. CI is green on the new head.

ErenAta16:
Re-checked the diff at cbd9928e and it is the same change I looked at before: same precedence chain in the base_url branch, same three tests covering config / env / default, and the same updated assertion on the existing base_url test.

The env isolation is worth keeping regardless of the rebase reason. Without the delenv, both that test and the default-fallback one would fail on any machine that happens to have HUGGINGFACE_API_KEY exported, since they assert the literal fallback. Still looks correct to me.

OfficialAbhinavSingh:
#6770 landed the core of this (`api_key` forwarded to the `OpenAI` client on the `huggingface_base_url` path), so most of this PR is now redundant and it conflicts with main.

The one piece not covered is the fallback chain. The TS SDK resolves `config > HUGGINGFACE_API_KEY > "hf"` and its config table documents that, while Python now reads `self.config.api_key` only, so a user who sets the env var still ends up with no key on the TEI path.

Happy to either rebase this down to just that parity delta, or close it as superseded. Your call, no strong preference from me.

Diwak4r:
@kartik-mem0 — the author notes #6770 landed the core of this (`api_key` forwarded on the `huggingface_base_url` path). Is the remainder here still wanted, or should #6302 be closed as superseded? The branch needs a rebase either way.

OfficialAbhinavSingh:
Closing this. #6770 landed the core fix (`api_key` forwarded to the `OpenAI` client on the `huggingface_base_url` path), so most of this PR is now redundant and the branch conflicts with main.

Recording the leftover delta so it is not lost. The Node SDK does a three-step resolve at `mem0-ts/src/oss/src/embeddings/huggingface.ts:39`:

```ts
apiKey: config.apiKey || process.env.HUGGINGFACE_API_KEY || "hf"
```

Python stops at the first step, `mem0/embeddings/huggingface.py:20`:

```python
api_key=self.config.api_key
```

So `HUGGINGFACE_API_KEY` is honored in Node and is a no-op in Python. The docs match the code: the TypeScript tab in `docs/components/embedders/models/huggingface.mdx` documents `apiKey` falling back to `HUGGINGFACE_API_KEY` with a `"hf"` default, and the Python tab has no `api_key` row at all.

That is a one-line change if you want the two SDKs to resolve the key the same way. Happy to open a clean PR against current main, or to leave it if the divergence is deliberate.

Thanks @Diwak4r for the nudge.

PR #6304, opened 2026-07-14
fix(mem0): add test for issue #6230

### LINKED ISSUE #6230: XAILLM forwards the built http client as http_client_proxies, crashing when a proxy is set

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6230.

## Why
Fixes #6230

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6230.py
+++ b/tests/test_issue_6230.py
@@ -1,0 +1,39 @@
+from pathlib import Path
+from unittest.mock import MagicMock
+import importlib
+import sys
+import types
+
+mem0_module = types.ModuleType("mem0")
+mem0_module.__path__ = [str(Path(__file__).resolve().parents[1] / "mem0")]
+sys.modules.setdefault("mem0", mem0_module)
+
+base_config = importlib.import_module("mem0.configs.llms.base")
+xai_llm = importlib.import_module("mem0.llms.xai")
+BaseLlmConfig = base_config.BaseLlmConfig
+XAILLM = xai_llm.XAILLM
+
+
+def test_issue_6230(monkeypatch):
+    proxy_url = "http://proxy.local:8080"
+    built_client = object()
+    seen_proxies = []
+
+    def fake_build_http_client(http_client_proxies):
+        seen_proxies.append(http_client_proxies)
+        if http_client_proxies is built_client:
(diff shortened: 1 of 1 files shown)


RedEye1605 (commented):
Thanks for adding coverage for #6230. The intent is useful, but this file is currently only a four-line snippet: `XAIConfig` and `config` are undefined, and there is no test function or assertion, so pytest cannot exercise the reported behavior. Could you turn this into an executable regression test that constructs the relevant config/client and asserts that the proxy setting is forwarded correctly?

kartik-mem0:

kartik-mem0:

PR #6309, opened 2026-07-14
fix(openclaw): make auto-recall timeout configurable

### LINKED ISSUE #6307: Make `RECALL_TIMEOUT_MS` configurable instead of hardcoding 8 seconds
## Observed

The mem0 recall integration uses a fixed 8-second deadline. On one installation, normal recall latency crosses that threshold roughly 18 times per day, causing avoidable recall degradation even though the backing service frequently completes shortly afterward. No user, host, path, session, or account identifiers are included in this report.

## Expected

Operators should be able to set the recall deadline for local/network conditions while retaining a safe default.

## Suggested configuration

`RECALL_TIMEOUT_MS` environment variable and/or a typed plugin setting such as `recallTimeoutMs`, defaulting to 8000, validated to a bounded range. Log the effective timeout once at startup and distinguish timeout from provider error metrics.

## Suggested acceptance test

Use a test recall provider delayed beyond 8 seconds; verify default timeout, successful completion with a larger configured value, validation of invalid values, and no change to the default behavior when unset.

Bartok9:
## Summary

- Makes the openclaw-mem0 auto-recall deadline configurable via `recallTimeoutMs` / `RECALL_TIMEOUT_MS` (default still 8000ms).
- Applies the same deadline to skills-mode auto-recall (previously unbounded).

## Motivation

Closes #6307.

Legacy mode raced automatic recall against a hard-coded 8s timer, so local/network‑slower installs degraded even when the provider would succeed shortly after. Operators need a knobs without losing a safe default. Skills mode previously had no deadline, which is worse under provider hangs.

## Fix

- `resolveRecallTimeoutMs()`: config → env → default, clamped to `[500, 120_000]`.
- Wire `Mem0Config.recallTimeoutMs`, schema allowlist, and plugin UI/JSON schema.
- Use the value in legacy `Promise.race` and skills-mode recall.
- Log the effective timeout in the register banner; timeout logs stay distinct from provider errors.

## Verification

- `pnpm exec vitest run tests/config.test.ts` — **69 passed**, 0 failed
- Default remains 8000ms when unset; invalid values fall back safely

## Tests added

- `resolveRecallTimeoutMs` unit coverage (default, config prefer over env, bounds, flooring)
- `mem0ConfigSchema.parse` for default / custom / invalid / env-driven values

### DIFF
--- a/docs/integrations/openclaw.mdx
+++ b/docs/integrations/openclaw.mdx
@@ -345,6 +345,7 @@ openclaw mem0 status --json
 | `autoCapture` | `boolean` | `true` | Store facts after each turn. Ignored when `skills` is configured. |
 | `topK` | `number` | `5` | Max memories per recall |
 | `searchThreshold` | `number` | `0.3` | Min similarity (0–1) |
+| `recallTimeoutMs` | `integer` or ASCII digit string | `8000` | Legacy-only recall deadline in milliseconds; allowed range is `1000` to `120000`. Ignored in skills mode. |
 
 ### Platform Mode Options
 
--- a/integrations/openclaw/README.md
+++ b/integrations/openclaw/README.md
@@ -303,6 +303,7 @@ openclaw mem0 help --json                                   # discover all comma
 | `autoCapture` | `boolean` | `true` | Extract and store facts after each turn. Ignored when `skills` is set. |
 | `topK` | `number` | `5` | Max memories returned per recall |
 | `searchThreshold` | `number` | `0.1` | Minimum similarity score (0-1) |
+| `recallTimeoutMs` | `integer` or ASCII digit string | `8000` | Legacy-only recall deadline in milliseconds; allowed range is `1000` to `120000`. Ignored in skills mode. |
 
 ### Skills Mode (Recommended)
 
@@ -320,6 +321,8 @@ Enabled by default during `openclaw mem0 init`. `autoRecall` and `autoCapture` a
 | `skills.domain` | `string` | `"companion"` | Domain overlay for triage rules |
 
 ### Platform Mode
+
+Set `recallTimeoutMs` directly or with OpenClaw interpolation, for example `"recallTimeoutMs": "${RECALL_TIMEOUT_MS}"`. The parser validates the resolved value.
 
 | Key | Type | Default | Description |
 | --- | ---- | ------- | ----------- |
--- a/integrations/openclaw/README.md
+++ b/integrations/openclaw/README.md
@@ -303,7 +303,7 @@ openclaw mem0 help --json                                   # discover all comma
 | `autoCapture` | `boolean` | `true` | Extract and store facts after each turn. Ignored when `skills` is set. |
 | `topK` | `number` | `5` | Max memories returned per recall |
 | `searchThreshold` | `number` | `0.1` | Minimum similarity score (0-1) |
-| `recallTimeoutMs` | `integer` or ASCII digit string | `8000` | Legacy-only recall deadline in milliseconds; allowed range is `1000` to `120000`. Ignored in skills mode. |
+| `recallTimeoutMs` | `integer` or ASCII digit string | `8000` | Auto-recall deadline in milliseconds for **legacy and skills mode**; allowed range is `1000` to `120000`. |
 
 ### Skills Mode (Recommended)
 
--- a/integrations/openclaw/cli/commands.ts
+++ b/integrations/openclaw/cli/commands.ts
@@ -47,6 +47,7 @@ import {
   OPENCLAW_CONFIG_FILE,
 } from "./config-file.ts";
 import { jsonOut, jsonErr, redactSecrets } from "./json-helpers.ts";
+import { parseRecallTimeoutMs } from "../config.ts";
 import {
   LLM_PROVIDERS, EMBEDDER_PROVIDERS, VECTOR_PROVIDERS,
   buildOssLlmConfig, buildOssEmbedderConfig, buildOssVectorConfig,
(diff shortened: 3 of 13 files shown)


kartik-mem0:
Thanks for turning this around so fast, and for the test coverage — nice work.

Heads-up on sequencing so this doesn't stall: PR #6311 is also closing #6307 and is in review with a good chance of landing first. It's a deeper rework, but it's scoped to the *legacy* `before_prompt_build` timeout path only — by its own description "skills-mode recall remains unchanged" there, and that checks out against the diff: on `main`, `integrations/openclaw/index.ts:506` (`const recallResult = await skillRecall(...)`) is an unguarded await with no deadline, and #6311's branch leaves that line untouched.

That's exactly the gap you closed here. Your `Promise.race` wrapping that same `skillRecall()` call against `cfg.recallTimeoutMs` (your branch's `index.ts:507-536`, plus the "skills-mode recall timed out" warn log) is the one fix in the queue for unbounded skills-mode recall — and since skills mode is the default path, that's arguably the more impactful half of #6307. Wanted to flag that clearly: that hunk is the valuable, unique part of this PR, not a throwaway.

Where you and #6311 do collide is the config plumbing, not the fix itself. Both PRs:
- insert `"recallTimeoutMs"` into `config.ts`'s `ALLOWED_KEYS` array in the same spot (right after `config.ts:160`, `"topK",`)
- add a `recallTimeoutMs` field to `mem0ConfigSchema.parse()`'s return object right after `config.ts:251` (`topK: ...`)
- extend `Mem0Config` in `types.ts` right after `types.ts:28` (`topK: number;`)
- touch the same blocks in `openclaw.plugin.json`
- edit the same `index.ts:687` line in the legacy hook (`const RECALL_TIMEOUT_MS = 8_000;`) — #6311 removes it entirely as part of a bigger refactor there

If #6311 merges first, this PR will conflict on exactly those five spots.

Would you be up for rebasing onto #6311 (or its branch now, to get ahead of it) and dropping your `config.ts` / `types.ts` / `openclaw.plugin.json` / legacy-hook changes in favor of theirs, keeping only the skills-mode hunk — the `Promise.race`/timeout wrapping around `skillRecall()` and its warn log? Both land on the same `cfg.recallTimeoutMs` field name, so that block should graft on with little to no change once you're on top of their config plumbing. Should be a small diff, not a rewrite.

Really solid catch on the unbounded skills-mode case — just want to get the two PRs out of each other's way before either merges.

Bartok9:
Thanks @kartik-mem0 — followed your sequencing plan.

**Done:**
- Rebased onto #6311 (`pr/openclaw-recall-timeout` @ `6ca4a62`)
- Dropped our overlapping config plumbing (`config.ts` / `types.ts` / legacy-hook rewrite / prior `openclaw.plugin.json` schema) in favor of #6311's
- Kept the unique skills-mode fix: `Promise.race` around `skillRecall()` against `cfg.recallTimeoutMs`, plus the `"skills-mode recall timed out after …ms"` warn
- Updated docs/schema so `recallTimeoutMs` is no longer described as skills-ignored
- Flipped the #6311 negative-space test (`skills-mode recall ignores legacy deadline` → `skills-mode recall is bounded by recallTimeoutMs`) and added a matching case in `index.test.ts`

**Local:** `integrations/openclaw` vitest — **462 passed**

This PR is now a small delta on top of #6311 (skills half of #6307). Happy to wait for #6311 to land first, or re-target `main` after it merges if that's cleaner for you.

kartik-mem0:
Thanks for turning this around so carefully, you kept exactly the right piece. I checked the current head (`cd52566`) against both `main` and #6311, and wanted to give you an accurate read, since a couple of things shifted under you.

**Your skills-mode bound is the unique, valuable part, confirmed.** #6311 never touches the skills path: `await skillRecall(...)` stays an unbounded await on its branch. Your `Promise.race([skillRecallWork, skillTimeout])` plus the "skills-mode recall timed out after …ms" warn is the only fix in the queue that bounds the *default* recall path, arguably the more impactful half of #6307. And CI is green here (OpenClaw lint + build + test 20/22).

**Where things stand vs #6311.** You did branch on top of #6311 (you share its first three commits through `6ca4a62`) and added your skills commit on top, that part went exactly to plan. Two things to know:

- Both PRs still target `main`, so GitHub renders the full stack on each. That's why this doesn't *look* like a small delta even though conceptually it is: against `main` it re-contains #6311's whole legacy implementation (the config plumbing plus the `Symbol("recall-timeout")` / `recallSettled` / `outcome` legacy-hook rewrite).
- #6311 has moved on since `6ca4a62`: its head is now `a0a9128`, two review-feedback commits ahead of where you branched. One of those makes it degrade gracefully on a bad value: an out-of-range `recallTimeoutMs` logs `falling back to default` and continues. On this branch, `mem0ConfigSchema.parse()` still *throws* on out-of-range (your own `config.test.ts` asserts `.toThrow()` for `120001`), so you're missing that softer handling.

**Cleanest path:** let #6311 land first, then rebase this on the updated `main`. Everything shared, the config plumbing and #6311's later commits, will already be in `main`, so this collapses to just your skills-mode commit (`cd52566`), possibly with a tiny conflict to resolve near the schema code. That's the ~15-line PR we want. (If you'd rather get ahead of it now, re-stacking on #6311's current head pulls in its two newer commits.)

I'll flag the ordering to the maintainers since #6311 was already queued, the sequencing call is theirs, but the skills-mode bound is the keeper either way. Really nice catch on the unbounded default path.

Bartok9:
@kartik-mem0 Restacked on #6311 current head (`a0a9128`).

- Cherry-picked only the skills-mode bound onto latest #6311 (includes the two review commits after `6ca4a62`, incl. graceful fallback for bad `recallTimeoutMs`)
- Diff vs #6311 tip is now just the skills `Promise.race` + warn + tests/doc notes that skills mode is bounded by `recallTimeoutMs`
- Local `integrations/openclaw` vitest: **457 passed**

Happy to wait for #6311 to land and re-target `main` if you prefer that ordering — skills-mode hunk stays the same either way.

PR #6311, opened 2026-07-14
fix(openclaw): make legacy recall timeout configurable

### LINKED ISSUE #6307: Make `RECALL_TIMEOUT_MS` configurable instead of hardcoding 8 seconds

rodboev:
## Linked Issue

Closes #6307

## Description

Legacy OpenClaw auto-recall always stops waiting after a hardcoded eight seconds. When recall latency crosses that boundary on slower local or network conditions, the hook drops context and logs `openclaw-mem0: recall timed out after 8000ms, skipping` even though the provider often completes shortly afterward.

This PR adds one typed OpenClaw config key, `recallTimeoutMs`, to control the legacy recall deadline. The key defaults to `8000`, is validated to a bounded range of `1000` through `120000`, and accepts integer values plus ASCII digit strings so the existing OpenClaw interpolation path supports `"recallTimeoutMs": "${RECALL_TIMEOUT_MS}"` without a second plugin-side environment lookup. Direct parser and CLI validation stay strict. Live plugin registration now warns and falls back to the default `8000` when that one field is malformed or out of range, so a single bad timeout value does not take the whole plugin down.

## Root cause

The legacy `before_prompt_build` hook defines `RECALL_TIMEOUT_MS = 8_000` inline in `integrations/openclaw/index.ts` and races every recall attempt against that constant. There was no typed config entry point to override it: `mem0ConfigSchema.parse()` rejected the new key, `Mem0Config` could not carry it, and the persisted config, CLI, and plugin manifest exposed no timeout field.

The same runtime path also logged and telemetered legacy success before the race winner was known, so a provider that finished after a timeout could still emit misleading success side effects after the hook had already returned no context.

## Fix

- Add `recallTimeoutMs` to the existing OpenClaw config surface and normalize it once in `mem0ConfigSchema.parse()`.
- Keep direct parser and CLI validation strict, but route plugin registration through a small fallback helper that warns and retries without `recallTimeoutMs` when that one field is malformed or out of range.
- Use the normalized value only in the legacy recall race, keeping the default unset behavior and the timeout warning text stable.
- Move legacy success logging and telemetry ownership to the race winner so timeout, success, and provider-error outcomes are authoritative and late provider settlement cannot emit success side effects after a timeout.
- Expose the same key through the persisted OpenClaw config, the CLI (`recall_timeout_ms`), the manifest schema and UI hints, the canonical docs page, and the package README. The manifest string branch now accepts the same bounded digit strings as the runtime parser, including zero-padded values that normalize inside range.

## Scope

This change is confined to `integrations/openclaw/`. Skills-mode recall remains unchanged, shared `mem0/` and `mem0-ts/` packages are untouched, and no dependency or package-version changes are introduced.

Related prior work: [#5035](https://github.com/mem0ai/mem0/pull/5035) and [#4769](https://github.com/mem0ai/mem0/pull/4769) fix the separate phantom timeout-warning defect from issue #4763 on the same hunk; this PR only adds timeout configurability and authoritative timeout versus provider-error outcomes for the legacy path.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] I tested manually
- [ ] No tests needed

Verification:

- [x] `cd integrations/openclaw && pnpm run test -- tests/config.test.ts tests/config-file.test.ts tests/cli-commands.test.ts index.test.ts tests/recall-timeout-regression.test.ts`
- [x] `cd integrations/openclaw && pnpm exec tsc --noEmit`
- [x] `cd integrations/openclaw && pnpm run build`
- [ ] `cd integrations/openclaw && pnpm install --frozen-lockfile && pnpm run build && pnpm run test`

## Checklist

### DIFF
--- a/docs/integrations/openclaw.mdx
+++ b/docs/integrations/openclaw.mdx
@@ -345,6 +345,7 @@ openclaw mem0 status --json
 | `autoCapture` | `boolean` | `true` | Store facts after each turn. Ignored when `skills` is configured. |
 | `topK` | `number` | `5` | Max memories per recall |
 | `searchThreshold` | `number` | `0.3` | Min similarity (0–1) |
+| `recallTimeoutMs` | `integer` or ASCII digit string | `8000` | Legacy-only recall deadline in milliseconds; allowed range is `1000` to `120000`. Ignored in skills mode. |
 
 ### Platform Mode Options
 
--- a/integrations/openclaw/README.md
+++ b/integrations/openclaw/README.md
@@ -303,6 +303,7 @@ openclaw mem0 help --json                                   # discover all comma
 | `autoCapture` | `boolean` | `true` | Extract and store facts after each turn. Ignored when `skills` is set. |
 | `topK` | `number` | `5` | Max memories returned per recall |
 | `searchThreshold` | `number` | `0.1` | Minimum similarity score (0-1) |
+| `recallTimeoutMs` | `integer` or ASCII digit string | `8000` | Legacy-only recall deadline in milliseconds; allowed range is `1000` to `120000`. Ignored in skills mode. |
 
 ### Skills Mode (Recommended)
 
@@ -320,6 +321,8 @@ Enabled by default during `openclaw mem0 init`. `autoRecall` and `autoCapture` a
 | `skills.domain` | `string` | `"companion"` | Domain overlay for triage rules |
 
 ### Platform Mode
+
+Set `recallTimeoutMs` directly or with OpenClaw interpolation, for example `"recallTimeoutMs": "${RECALL_TIMEOUT_MS}"`. The parser validates the resolved value.
 
 | Key | Type | Default | Description |
 | --- | ---- | ------- | ----------- |
--- a/integrations/openclaw/cli/commands.ts
+++ b/integrations/openclaw/cli/commands.ts
@@ -47,6 +47,7 @@ import {
   OPENCLAW_CONFIG_FILE,
 } from "./config-file.ts";
 import { jsonOut, jsonErr, redactSecrets } from "./json-helpers.ts";
+import { parseRecallTimeoutMs } from "../config.ts";
 import {
   LLM_PROVIDERS, EMBEDDER_PROVIDERS, VECTOR_PROVIDERS,
   buildOssLlmConfig, buildOssEmbedderConfig, buildOssVectorConfig,
@@ -1278,6 +1279,7 @@ export function registerCliCommands(
         auto_recall: "autoRecall",
         auto_capture: "autoCapture",
         top_k: "topK",
+        recall_timeout_ms: "recallTimeoutMs",
         mode: "mode",
         embedder_provider: "oss.embedder.provider",
         embedder_model: "oss.embedder.config.model",
@@ -1307,7 +1309,7 @@ export function registerCliCommands(
       ]);
 
       // Integer config fields — coerce to number on set
-      const INTEGER_KEYS = new Set(["topK", "oss.vectorStore.config.port"]);
+      const INTEGER_KEYS = new Set(["topK", "recallTimeoutMs", "oss.vectorStore.config.port"]);
 
       /** Resolve a user-facing key to the internal camelCase field name. */
       function resolveConfigKey(key: string): string | null {
@@ -1335,6 +1337,7 @@ export function registerCliCommands(
           autoRecall: cfg.autoRecall,
           autoCapture: cfg.autoCapture,
           topK: cfg.topK,
+          recallTimeoutMs: auth.recallTimeoutMs ?? cfg.recallTimeoutMs,
         };
         return values[field];
       }
@@ -1363,6 +1366,7 @@ export function registerCliCommands(
               ["auto_recall", "autoRecall"],
               ["auto_capture", "autoCapture"],
               ["top_k", "topK"],
+              ["recall_timeout_ms", "recallTimeoutMs"],
             ];
             if (cfg.mode === "platform") {
               showEntries.push(["api_key", "apiKey"], ["email", "userEmail"]);
@@ -1396,6 +1400,7 @@ export function registerCliCommands(
             ["auto_recall", "autoRecall"],
             ["auto_capture", "autoCapture"],
             ["top_k", "topK"],
+            ["recall_timeout_ms", "recallTimeoutMs"],
           ];
 
           if (cfg.mode === "platform") {
@@ -1498,13 +1503,19 @@ export function registerCliCommands(
               rawValue === "1" ||
               rawValue.toLowerCase() === "yes";
           } else if (INTEGER_KEYS.has(field)) {
-            const parsed = parseInt(rawValue, 10);
-            if (isNaN(parsed)) {
-              if (jsonErr(opts, `Invalid integer value: ${rawValue}`)) return;
-              console.error(`Invalid integer value: ${rawValue}`);
+            try {
+              value = field === "recallTimeoutMs"
+                ? parseRecallTimeoutMs(rawValue)
+                : Number.parseInt(rawValue, 10);
+              if (!Number.isInteger(value)) throw new Error();
+            } catch {
+              const message = field === "recallTimeoutMs"
+                ? `Invalid recall timeout value: ${rawValue}`
+                : `Invalid integer value: ${rawValue}`;
+              if (jsonErr(opts, message)) return;
+              console.error(message);
               return;
             }
-            value = parsed;
           }
 
           // Nested OSS fields use dot-path writer; flat fields use auth writer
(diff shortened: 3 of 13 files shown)


kartik-mem0:
## Changes Requested

This is a clean, well-tested fix for the hardcoded 8-second recall timeout (closes #6307) — the config plumbing and the underlying race-condition cleanup are both solid. A few things should be addressed before merge, most importantly a config-validation gap that can crash plugin registration entirely on a single bad value.

### Strengths
- `config.ts:24-46` — real root-cause fix, not a band-aid: the hardcoded `8000` is now a bounds-checked (1000–120000) config value that accepts both numbers and digit-strings, and the default stays exactly `8000` (`config.ts:24`), so existing users see zero behavior change unless they opt in.
- `index.ts:816-865` — nice race-condition cleanup: the `RECALL_TIMEOUT` Symbol sentinel + `recallSettled` guard now correctly attribute success/timeout/provider-error logging and telemetry to whichever branch of the `Promise.race` actually settles, instead of the old code logging success unconditionally. `clearTimeout` is called on every exit path.
- `tests/config.test.ts` — the adversarial-input coverage on `parseRecallTimeoutMs` (null, booleans, arrays, `" 1000"`, `"1000 "`, `"1e3"`, `Infinity`, `NaN`, `1000.5`, both boundary edges) is exactly the kind of test a validator like this deserves.
- Red-green verified by hand: reverting only the implementation hunk (keeping the tests) produces 7/7 failures, each tied to the actual feature — not flaky/incidental breakage.

### Issues

#### Critical (must fix)
none

#### Important (should fix)
1. `config.ts:28-46` + `index.ts:117` — `parseRecallTimeoutMs` throws synchronously on any out-of-range or malformed value, and that throw is unguarded all the way through `mem0ConfigSchema.parse(...)` at `index.ts:117`, the first substantive line inside `register()` (`index.ts:107`) — before any `try`/`catch` in the file (the first one starts at `index.ts:127`). A single bad `recallTimeoutMs` — an extra digit, an unresolved `${RECALL_TIMEOUT_MS}` env var, a copy-paste typo — takes down the *entire* plugin's registration, not just recall. Contrast with `mode` a few lines up (`config.ts:209-219`), which warns and falls back to a default instead of throwing for the same class of "bad string value" input. Recommend mirroring that pattern here — catch, warn, fall back to `DEFAULT_RECALL_TIMEOUT_MS`. Worth noting none of the new tests exercise `register()` itself with a bad value — they all call `parseRecallTimeoutMs`/`mem0ConfigSchema.parse` directly, so this path is currently untested as well as unguarded.
2. `tests/recall-timeout-regression.test.ts` — 6 of its 7 tests (plus the `registerRecallTest()` helper) substantially duplicate coverage added in this same PR to `index.test.ts:124-237`, `tests/config.test.ts`, and `tests/cli-commands.test.ts`/`tests/config-file.test.ts` (one test even shares its exact title, "CLI/config surface: recall_timeout_ms round-trip and bounds," with a test in `config-file.test.ts`). The exception — "negative-space: skills-mode recall ignores legacy deadline" — is actually a *better* test than its `index.test.ts` counterpart (real fake-timer race vs. a shortcut mock) and is worth keeping. Suggest folding the other six into their existing homes and keeping just that one, rather than carrying a fully parallel 450-line file forward — it's roughly half of the +815 diff, mostly restating assertions that already exist elsewhere in this same PR.
3. `docs/integrations/openclaw.mdx:340-347` — the "General Options" table wasn't updated with a `recallTimeoutMs` row, even though `integrations/openclaw/README.md` was. This is the canonical Mintlify docs page (linked from `docs/docs.json` and `docs/llms.txt`), so it's the one most users will actually land on.

#### Minor (optional)
1. `openclaw.plugin.json:195-202` — the `recallTimeoutMs` JSON-schema's string branch (`{"type": "string", "pattern": "^[0-9]+$"}`) doesn't encode the 1000–120000 bounds the way the integer branch does. Not exploitable — `parseRecallTimeoutMs` is the real authority at runtime regardless of what this schema allows — but worth tightening for UI-level validation consistency.

### Evidence
- **CI:** `gh pr checks 6311` — `CI Gate` pass, `OpenClaw / lint` pass, `OpenClaw / build` pass, `OpenClaw / test (20)` and `(22)` both pass. Only `Vercel` is red ("Authorization required to deploy" — fork PR, advisory-only, not a required check).
- **Targeted tests:** `pnpm install && pnpm run build && pnpm run test` in `integrations/openclaw/` — clean install, clean `tsup` build, 462/462 passing.
- **Red-green proof:** reverted the source hunk only (kept the tests) at PR head → 7/7 tests in `tests/recall-timeout-regression.test.ts` fail with feature-specific errors (`openclaw-mem0 config has unknown keys: recallTimeoutMs`, `registerHooks is not a function`, `expected undefined to be 8000`); restored the hunk → all pass again, worktree clean. The tests do bite.
- **Could NOT verify:** whether OpenClaw's own gateway isolates a throwing plugin `register()` call (crash-just-this-plugin vs. crash-the-gateway) — `definePluginEntry` is imported from the external `openclaw/plugin-sdk` package, which isn't part of this repo or its `node_modules`, so the actual blast radius of Important #1 above depends on behavior this repo can't verify either way.

### Assessment
Ready to merge: **With fixes** — the config plumbing, bounds checking, and race-condition fix are solid and well tested; please guard the `recallTimeoutMs` parse failure the same way `mode` is guarded so a single bad value can't take the whole plugin down, and consider trimming the duplicated tests. I'll re-review once you push.

rodboev:
Thanks, I took all three important points and folded the minor schema cleanup into the same pass because it stayed on the same config surface.

1. `integrations/openclaw/index.ts` now keeps `register()` alive on a bad `recallTimeoutMs`. The direct parser in `integrations/openclaw/config.ts` stays strict for config validation and CLI writes, but live plugin registration now warns and falls back to the default `8000` when that one field is malformed or out of range.

2. `integrations/openclaw/tests/recall-timeout-regression.test.ts` is trimmed back to the stronger skills-mode negative-space case, and the rest of the timeout coverage stays in the suites that already own it: `index.test.ts` for the legacy race and registration path, `tests/config.test.ts` for parser bounds/defaults, `tests/cli-commands.test.ts` for CLI behavior, and `tests/config-file.test.ts` for the persisted config mapping.

3. `recallTimeoutMs` is added to the General Options table in `docs/integrations/openclaw.mdx`, which was the canonical docs gap you called out.

4. The `openclaw.plugin.json` string schema branch now matches the runtime parser's digit-string contract, including zero-padded values that normalize into the same `1000` through `120000` range.

The runtime timeout behavior, the timeout warning text, and the CLI rejection semantics stay the same for valid configs.

kartik-mem0:
## LGTM

Verified against `a0a9128c` (new head): the registration-crash risk, the test duplication, the docs gap, and the schema bound are all fixed.

Specifics, since I re-checked each against the diff rather than the description:

- **Crash risk** — `parseRegistrationConfig()` now catches a bad `recallTimeoutMs`, drops it, falls back to 8000ms, and warns instead of taking the plugin down. Reverted just that line and re-ran the new test: fails with the old unguarded throw, passes with the fix restored.
- **Test duplication** — `recall-timeout-regression.test.ts` is down from ~450 to ~100 lines; you kept exactly the one test worth keeping (the fake-timer skills-mode negative-space check) and folded the rest back into their existing homes.
- **Docs gap** — `openclaw.mdx`'s table has the `recallTimeoutMs` row now.
- **Schema bound** — the plugin.json string pattern actually encodes 1000–120000 now. Checked it against the boundary cases (999/1000/120000/120001, zero-padded variants) by hand — matches the parser exactly.

One tiny non-blocking thing: `parseRegistrationConfig` warns even when `recallTimeoutMs` is explicitly `undefined` (vs. just omitted), where `mem0ConfigSchema.parse` alone would silently default with no warning. Doesn't look reachable through the normal JSON-backed config path, so not holding this up for it.

Full suite's green (456/456), clean build, CI all-pass apart from the usual fork-auth Vercel check. Thanks for taking the feedback seriously instead of patching around it — this is ready to merge.

PR #6315, opened 2026-07-14
fix(embeddings): forward api_key to huggingface custom endpoint client

### LINKED ISSUE #6301: bug(embeddings/huggingface): api_key ignored on base_url path, endpoint auth dropped

saitejabandaru-in:
Resolves #6301.

### Description
In `HuggingFaceEmbedding`, if a user provides a custom `huggingface_base_url` (e.g. TEI server or custom Inference Endpoint), the underlying `OpenAI` client was initialized without forwarding the `api_key` configured in the `BaseEmbedderConfig`. This resulted in the client defaulting to `OPENAI_API_KEY` (if set) or raising a missing credentials error (if not set).

### Changes
- Forwarded the `api_key` using a fallback precedence of `config.api_key` -> `HUGGINGFACE_API_KEY` -> default `"hf"` (matching TypeScript client implementation).
- Updated existing test assertions in `test_huggingface_embeddings.py` to match new signature parameters.
- Added a new unit test `test_embed_with_huggingface_base_url_and_api_key` validating that custom api keys are successfully passed through on client initialization.

### DIFF
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -17,7 +17,9 @@ class HuggingFaceEmbedding(EmbeddingBase):
         super().__init__(config)
 
         if self.config.huggingface_base_url:
-            self.client = OpenAI(base_url=self.config.huggingface_base_url)
+            import os
+            api_key = self.config.api_key or os.getenv("HUGGINGFACE_API_KEY") or "hf"
+            self.client = OpenAI(base_url=self.config.huggingface_base_url, api_key=api_key)
             self.config.model = self.config.model or "tei"
         else:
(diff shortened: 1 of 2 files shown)


OfficialAbhinavSingh:
Heads-up: this duplicates #6302, which I opened ~14h earlier for the same issue (#6301), with the same base_url api_key fix plus regression tests. Happy to consolidate review there to save the maintainers a double review.

saitejabandaru-in:
Closing this in favor of #6302, since it was submitted earlier and covers the same issue. Thanks!

PR #5617, opened 2026-06-17
feat(kilo-plugin): add @mem0/kilo-plugin for cross-session memory

abhay23-AI:
## Linked Issue

Closes #5606

## Description

Adds a first-class Kilo integration under `integrations/kilo-plugin/` as `@mem0/kilo-plugin`, using Kilo's current server-plugin contract and the `mem0ai` SDK directly (no MCP server).

### What it includes

- 10 native memory tools for add/search/get/update/delete/entity/event-status operations.
- 8 Kilo hooks: `event`, `chat.message`, `tool.execute.before`, `tool.execute.after`, `experimental.chat.messages.transform`, `experimental.session.compacting`, `experimental.text.complete`, and `shell.env`.
- Per-Kilo-session state and scope, using the real `sessionID` for memory routing, shell env, summaries, compaction, and tool capture.
- Incremental session summaries on `session.idle`, auto-dream lifecycle cleanup, and a `[mem0: N memories used]` completion footer.
- Bounded, serialized post-tool capture for Bash/read/write/edit. Captures are fully redacted, sensitive files and environment-dump commands are skipped, and lifecycle events drain queued writes.
- Destructive operations fail closed: bulk/entity deletion requires explicit confirmation; entity deletion accepts exactly one selector, validates it against the active Kilo identity/scope, and passes only that selector to the SDK.
- Project identity and branch detection run in Kilo's active worktree.
- Current Kilo package shape: default `{ server }` module export plus the `./server` package export; `engines.opencode` follows Kilo's documented compatibility field.
- User documentation in `docs/integrations/kilo.mdx` and the package README.

The TUI-only `config` hook, skills, and slash commands remain out of scope, matching the issue. `kilo.json` is not added because the documented npm plugin manifest belongs in `package.json`. Marketplace files are also out of scope because Kilo is its own plugin host.

CI/CD is deliberately separate in #6320, per @kartik-mem0's guidance. This PR touches only the plugin, its tests, and its docs; no workflow files are changed.

OpenAI Codex was used to implement and review the repair. The final diff was checked against the pinned Kilo and Mem0 SDK contracts, compared with the OpenCode and Pi Agent integrations, and passed an independent blocking code review.

Fresh local verification with Bun:

```text
bun install --frozen-lockfile  -> no changes
bun test                       -> 70 pass / 0 fail / 166 assertions (6 files)
bun run type-check             -> exit 0
bun run build                  -> dist/index.js + declarations
[...]

Issue #5606, opened 2026-06-17
feat(integrations): add Kilo plugin for cross-session memory


ThePlenkov:
## Summary

Add a first-class **Kilo** plugin to `integrations/kilo-plugin/`, mirroring the existing `integrations/mem0-plugin/`. The integration should let every Kilo session (CLI + VS Code extension) read and write memories through the Mem0 platform using only `MEM0_API_KEY` in env.

## Motivation

Kilo is an AI coding client in the same family as OpenCode / Claude Code / Cursor / Codex. It already has a plugin API (`@kilocode/plugin`) with hooks equivalent to OpenCode's. There is no first-class mem0 integration yet, and users have asked for one (see `Kilo-Org/kilocode#11346`).

## Proposed location

```
integrations/kilo-plugin/
├── kilo-mem0.ts            # plugin source (port from opencode-mem0.ts)
├── package.json
├── README.md
├── kilo.json               # kilocode project config for local dev
└── tests/
    └── kilo-mem0.test.ts
```

Mirrors the layout of `integrations/mem0-plugin/.opencode-plugin/` so maintainers can review symmetrically.

## Hooks implemented

- `chat.message` — on first user turn per session, `search_memories(query=user_text)` and inject results as a context block.
- `tool.execute.after` — on `bash`, `read`, `write`, `edit` outputs, fire-and-forget `add_memory` with secret redaction (`sk-…`, `m0-…`, `ghp_…`, `gho_…`, `xox*-…`, `AKIA…`).
- `event` (`session.idle`) — summarize the session and store as `session_summary`.
- `experimental.session.compacting` — push a memory-fetch hint into the compaction context.
- `experimental.text.complete` — append `[mem0: N memories used]` footer.

## Source-of-truth port

We will port `integrations/mem0-plugin/.opencode-plugin/opencode-mem0.ts` (758 lines) with only:
- `import type { Plugin } from "@opencode-ai/plugin"` → `import type { Plugin } from "@kilocode/plugin"`
- use Kilo's `tool()` helper from `@kilocode/plugin/tool` for any custom tools
- use `client.app.log()` instead of `console.log`

No mem0 API surface change.

## Env

- Required: `MEM0_API_KEY`
- Optional: `MEM0_USER_ID` (default: `os.userInfo().username`), `MEM0_APP_ID` (default: git remote)

## Cross-reference

- Kilo-side issue: `Kilo-Org/kilocode#11346`
[...]

PR #6320, opened 2026-07-14
ci: add CI and CD workflows for kilo-plugin

abhay23-AI:
## Linked Issue

Follow-up to #5606 / #5617.

## Description

This is the deliberate, separate CI/CD step for the Kilo plugin requested by @kartik-mem0. It registers `integrations/kilo-plugin` (added by #5617) in the existing CI/CD architecture while keeping workflow changes out of the feature PR.

### What it adds

- `kilo-plugin-checks.yml`: reusable/manual/push checks using Bun, frozen install, typecheck, build, and `dist/index.js` verification. Its shape mirrors the Bun-based OpenCode plugin workflow.
- `kilo-plugin-cd.yml`: manual release workflow guarded on `kilo-v*`, with OIDC provenance publishing for `@mem0/kilo-plugin`.
- `ci-gate.yml`: Kilo is registered in all four required places: changes output, path filter, reusable call job, and gate `needs`.
- `release.yml`: a `kilo-v*)` arm before the bare `v*)` Python arm.
- `.github/AGENTS.md`: Kilo CI, CD tag-prefix, and key-directory entries.

`cache-dependency-path` does not apply: it is an `actions/setup-node` input used by the pnpm workflows. Kilo uses `oven-sh/setup-bun`, matching the existing OpenCode plugin workflow.

## Hard dependency / merge order

This PR must land **after #5617**. It is not safe to merge or approve independently: changing `kilo-plugin-checks.yml` makes this PR's own `ci-gate.yml` path filter select the Kilo job, whose working directory does not exist until #5617 lands.

Required order:

1. Merge #5617.
2. Refresh/rebase this branch onto the resulting `main`.
3. Approve and rerun the workflow checks.
4. Merge this CI/CD PR only after those checks pass.

## Verification

I built a temporary combined tree from this PR (`e1522d6b`) and the current #5617 head (`c8ee9e55`). The branches merge cleanly, and the combined tree passes:

```text
bun install --frozen-lockfile  -> no changes
bun test                       -> 70 pass / 0 fail / 166 assertions (6 files)
bun run type-check             -> exit 0
bun run build                  -> dist/index.js present
dist loader smoke              -> pass
```

All four workflow files parse as YAML. The Kilo changes output, path filter, reusable job, gate dependency, and `kilo-v` router order were also checked directly. The temporary worktree was removed after verification.

The first `@mem0/kilo-plugin` npm publish still requires the repository's one-time OIDC trusted-publisher setup; later releases route through `release.yml`.

### DIFF
--- a/.github/AGENTS.md
+++ b/.github/AGENTS.md
@@ -21,6 +21,7 @@ Package workflows keep their own push-to-main and manual triggers. Their `pull_r
 | Mem0 Plugin | `mem0-plugin-checks.yml` | Push to main (`integrations/mem0-plugin/`, excluding `.opencode-plugin/`), manual | pytest + hook exec bits + JSON manifest validation on Python 3.10, 3.11, 3.12 |
 | OpenCode Plugin | `opencode-plugin-checks.yml` | Push to main (`.opencode-plugin/`), manual | Bun: tsc + build + dist artifact check |
 | Pi Agent Plugin | `pi-agent-plugin-checks.yml` | Push to main (`integrations/pi-agent-plugin/`), manual | tsc + vitest + tsup on Node 20, 22 |
+| Kilo Plugin | `kilo-plugin-checks.yml` | Push to main (`integrations/kilo-plugin/`), manual | Bun: tsc + build + dist artifact check |
 | n8n Node | `n8n-nodes-mem0-checks.yml` | Push to main (`integrations/n8n-nodes-mem0/`), manual | ESLint + tsc build on Node 20 |
 | Zapier App | `zapier-mem0-checks.yml` | Push to main (`integrations/zapier-mem0/`), manual | tsc + `zapier validate` + offline unit tests on Node 22 |
 | docs llms.txt | `docs-llms-txt-check.yml` | Manual | `docs/llms.txt` coverage |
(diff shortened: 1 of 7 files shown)


abhay23-AI:
hi maintainers, this is ready for review. the CI Gate workflow is currently awaiting maintainer approval because the PR originates from a public fork, so could someone with write access approve the workflow run when convenient?

for context, this is the separate CI/CD step @kartik-mem0 asked for on #5606. it only touches workflow wiring: adds kilo-plugin-checks.yml and kilo-plugin-cd.yml (both mirroring the opencode plugin, which is bun based like kilo), registers kilo in ci-gate.yml in the four usual places (changes output, paths filter, reusable call job, gate needs), adds the kilo-v arm to the release.yml router ahead of the bare v arm, and adds the AGENTS.md CI/CD and Key Directories rows.

i ran the exact steps the new checks workflow runs, against the plugin from #5617, and they pass: bun install with a frozen lockfile, bun run type check, bun run build, and the dist/index.js artifact check. all four workflow files parse as valid yaml.

merge order: this should land after #5617 since it references integrations/kilo-plugin. it stays inert until then, because the paths filter cannot match a directory that does not exist yet.

abhay23-AI:
one clarification on the checklist, since @kartik-mem0 called out cache-dependency-path specifically: it does not apply here, and that is intentional rather than an omission.

cache-dependency-path is an input to actions/setup-node, used to point the node/pnpm lockfile cache at the right subdirectory. the pnpm based plugins use it (pi-agent-plugin-checks.yml and openclaw-checks.yml both set it to their pnpm-lock.yaml). kilo is bun based, so like opencode-plugin-checks.yml it uses oven-sh/setup-bun, which has no such input and does its own resolution from bun.lock. opencode, the plugin this one mirrors, likewise has no cache-dependency-path.

so the three parts of the checklist that do apply are all in place: integrations/kilo-plugin in the paths triggers and working-directory, the ci-gate.yml registration, and the kilo-v arm in the release.yml case block ahead of the bare v arm. happy to switch the workflow to a node plus pnpm setup with caching if you would rather it match pi-agent, but mirroring the bun based opencode plugin seemed like the closer precedent.

abhay23-AI:
Correction to my earlier note: this PR is **not inert** before #5617 lands. Because this PR changes `kilo-plugin-checks.yml`, its own `ci-gate.yml` path filter selects the Kilo job, and that job targets `integrations/kilo-plugin`, which does not exist on this branch yet.

Please treat #5617 as a hard dependency: merge #5617 first, then refresh/rebase this branch onto the resulting `main`, approve/rerun the fork workflows, and only merge #6320 after they pass.

I verified a temporary combined tree using this head (`e1522d6b`) and the repaired #5617 head (`c8ee9e55`): clean composition, 70 tests / 0 failures / 166 assertions, typecheck, build, loader smoke, four valid workflow YAML files, all four Kilo `ci-gate` registrations, and the `kilo-v` route ahead of the bare `v` route. The temporary worktree was removed afterward.

PR #6271, opened 2026-07-13
feat(llms): add optional retry with backoff for LLM provider calls

Krishnachaitanyakc:
## Linked Issue

Closes #6270

## Description

Adds an optional retry layer for provider calls. A dependency free `retry_call` (exponential backoff, full jitter, honors `Retry-After` including `retry-after-ms` and HTTP-date) plus a `max_retries` config (default `0`, no behavior change) wired through a small `LLMBase._retry`, with OpenAI as the reference provider. Transient errors (rate limit, timeout, connection, 5xx) retry and surface as the typed `mem0.exceptions` errors on final failure, the first use of that taxonomy. Remaining providers and embeddings follow in a later PR. Docs will follow once the config surface here is confirmed.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A. `max_retries` defaults to `0`, so provider behavior is unchanged unless enabled.

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed

PR #6325, opened 2026-07-14
feat(otel): first-party OpenTelemetry tracing for add and search

### LINKED ISSUE #6291: Optional OpenTelemetry tracing for memory operations
Would you be open to first party OpenTelemetry tracing of mem0's own memory operations, or would you rather it live as a separate community package? Want your read on placement before I build anything.

The idea: optional, zero dependency by default. `pip install "mem0ai[otel]"` pulls in `opentelemetry-api`; with no OTel SDK configured it stays inert, so no behavior change and negligible overhead when off. One span per public op (`add` / `search` / `get` / `update` / `delete` / ...), named and attributed per the merged `gen_ai.memory.*` conventions (open-telemetry/semantic-conventions-genai#140), with child spans for the internal phases (LLM fact extraction, embedding, vector and graph writes). Scope from `user_id` / `agent_id` / `run_id`. Memory content stays off by default (ids and counts only) and is enabled via the standard `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`.

Why in mem0: today the only first party signal is the PostHog product analytics, which is separate from tracing the operations themselves. The `gen_ai.memory.*` standard is already merged upstream, and the one official OTel mem0 instrumentation attempt (open-telemetry/opentelemetry-python-contrib#4252) was withdrawn before review, so nothing ships today. A first party layer can emit the accepted conventions and reach the phase level, where the latency actually is, in a way an external wrapper can't.

I work in this lane upstream (I added the `plan` operation to the GenAI semconv, open-telemetry/semantic-conventions-genai#97) and have #6271 open here already, so happy to prototype `add` and `search` first if you're on board. If you'd rather it be a community package, that works too, just want to point it the right way before starting.

Krishnachaitanyakc:
## Linked Issue

Closes #6291

## Description

Adds first-party, opt-in OpenTelemetry tracing for the two hot public operations, `add` and `search`, on both `Memory` and `AsyncMemory`. This lands the design agreed in #6291: the phase-level breakdown (LLM extraction, embedding, vector reads and writes) needs hooks that are internal to `Memory`/`AsyncMemory` and are not public, so a wrapper package cannot produce it. It has to be first-party.

Instrumentation lives in a new, dependency-optional module `mem0/memory/otel.py`:

- Each public `add`/`search` emits one `gen_ai.memory.client` span (CLIENT kind, span name = the operation) following the GenAI memory semantic conventions (`gen_ai.memory.*`, open-telemetry/semantic-conventions-genai#140): `add` maps to `gen_ai.operation.name = upsert_memory` (the LLM may add, update, or delete records, which is the spec definition of upsert), `search` maps to `search_memory`.
- Internal child spans give the phase breakdown: `retrieve_existing`, `llm_extract`, `embed_memories`, `vector_write` under `add`; `embed_query`, `vector_search`, `keyword_search`, `entity_boost`, `rerank` under `search`. Both sync and async paths are instrumented.
- Attributes: `gen_ai.memory.record.count` is safe and always recorded. `gen_ai.memory.query.text` and `gen_ai.memory.records` are sensitive and recorded only when the operator opts in via `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` (default off). `error.type` and span status are set on failure.

Zero behavior change unless opted in: if `opentelemetry-api` is not installed, or is installed with no `TracerProvider` configured, every helper is a no-op with negligible overhead and never raises. This mirrors the default-off precedent set by `max_retries` in #6271.

## Packaging

New optional extra: `pip install "mem0ai[otel]"` pulls only `opentelemetry-api`. The SDK stays a test and application dependency and is never a runtime dependency. Content capture is off by default.

## Note on graph and entity spans

In this version the `add`/`search` path has no separate graph store; entity linking runs against a second vector-store collection (the entity store), which is instrumented as the `entity_*` phase. If a standalone graph-memory store lands on its own config flag, graph spans are a natural follow-up. Similarly the `llm_extract` and `embed_*` child spans mark phase latency; adopting the full `gen_ai` inference and embedding span conventions on them, or deferring to a user's existing OpenAI/embedding instrumentation, can be a follow-up.

## Type of Change

## Test Coverage

`tests/memory/test_otel_tracing.py` (24 tests): operation and phase span emission and nesting, sync/async parity, content gating on and off, the error path (`error.type` and status on both operation and phase spans), procedural-add annotation, and the no-op paths when tracing is unavailable. The full existing suite passes unchanged (1699 passed, 48 skipped).

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -25,6 +25,7 @@ from mem0.configs.prompts import (
 from mem0.exceptions import LLMError
 from mem0.exceptions import ValidationError as Mem0ValidationError
 from mem0.memory.base import MemoryBase
+from mem0.memory import otel
 from mem0.memory.notices import (
     PERFORMANCE_SLOW_QUERY_THRESHOLD_SECONDS,
     detect_decay_usage_from_delete,
@@ -447,6 +448,17 @@ class _AsyncOSSProject:
         raise ValueError(_PROJECT_UPDATE_UNSUPPORTED_ERROR)
 
 
+def _otel_records(items, content_key):
+    """Shape mem0 memory items into the gen_ai.memory.records schema."""
+    records = []
+    for item in items or []:
+        record = {"content": item.get(content_key), "id": item.get("id")}
+        if item.get("score") is not None:
+            record["score"] = item.get("score")
+        records.append(record)
+    return records
+
+
 class Memory(MemoryBase):
     def __init__(self, config: MemoryConfig = MemoryConfig()):
         self.config = config
@@ -720,6 +732,7 @@ class Memory(MemoryBase):
         # Use agent memory extraction if agent_id is present and there are assistant messages
         return has_agent_id and has_assistant_messages
 
+    @otel.trace_memory_operation(otel.OP_ADD)
     def add(
         self,
         messages,
@@ -817,6 +830,11 @@ class Memory(MemoryBase):
                 display_scale_threshold_notice(self, "sync", "add", *scale_threshold_notice)
             else:
                 display_first_run_notice(self, "sync", "add")
+            otel.annotate_operation(
+                otel.current_span(),
+                record_count=len(results["results"]),
+                records_factory=lambda: _otel_records(results["results"], "memory"),
+            )
             return results
 
         if self.config.llm.config.get("enable_vision"):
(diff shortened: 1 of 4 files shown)


Krishnachaitanyakc:
@kartik-mem0 can you review this please?

Krishnachaitanyakc:
Bump this @kartik-mem0, can you review this?

PR #6327, opened 2026-07-15
fix(mem0): add test for issue #6264

### LINKED ISSUE #6264: TS OSS default memory store ignores all but the first operator in a compound field filter

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6264.

## Why
Fixes #6264

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6264.py
+++ b/tests/test_issue_6264.py
@@ -1,0 +1,94 @@
+"""Regression test for TS OSS default memory vector-store range filters."""
+
+import json
+import re
+import shutil
+import subprocess
+import textwrap
+import types
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+posthog = types.ModuleType("posthog")
+posthog.Posthog = type(
+    "Posthog",
+    (),
+    {
+        "__init__": lambda self, *args, **kwargs: None,
+        "capture": lambda self, *args, **kwargs: None,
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6334, opened 2026-07-15
fix: include updated_at and actor_id in batch history records on ADD

adeev-mardia:
## Summary

When memories are bulk-added via `_add_to_vector_store` (both sync and async paths), the `history_records` list built before calling `batch_add_history` was missing two fields that the history table expects:

- `updated_at` — already computed and stored in `mem_metadata` (line 993 sets `mem_metadata["updated_at"] = mem_metadata["created_at"]`)
- `actor_id` — present in `mem_metadata` when set via per-message metadata

This means the history table receives `NULL` for `updated_at` on new ADD records, and `actor_id` attribution is dropped, breaking history tracking and actor-based queries.

## Changes

Added `updated_at` and `actor_id` to the `history_records` dict in both the **sync** and **async** paths:

```python
# Before
history_records = [
    {
        "memory_id": r[0],
        "old_memory": None,
        "new_memory": r[1],
        "event": "ADD",
        "created_at": r[3].get("created_at"),
        "is_deleted": 0,
    }
    for r in records
]

# After
history_records = [
    {
        "memory_id": r[0],
        "old_memory": None,
        "new_memory": r[1],
        "event": "ADD",
        "created_at": r[3].get("created_at"),
        "updated_at": r[3].get("updated_at"),
        "actor_id": r[3].get("actor_id"),
        "is_deleted": 0,
    }
    for r in records
]
```

Also passed `actor_id` through to the single-record fallback `add_history` calls, keeping parity with the batch path.

Note: `updated_at` and `actor_id` are already correctly populated for UPDATE and DELETE operations — this only fixes the ADD path.


kartik-mem0:
Thanks for the contribution, and apologies for the slow response and for closing it after the work you put in.

Two things: this overlaps with #6294, which fixes the same lines, and more importantly we have decided to resolve the underlying issue in the opposite direction, so I am closing this alongside #6293.

Short version: the default inferred pipeline is single-pass ADD-only (https://mem0.ai/research), so an ADD event has no meaningful `updated_at`. Keeping the column NULL until a real `update()` or `delete()` happens is the semantics we want to preserve, because it is the only way to distinguish "never revised" from "revised at creation time". Full reasoning is in #6293.

On the `actor_id` part specifically: it is not actually being dropped on this path. `add()` never puts `actor_id` into `processed_metadata` for the inferred flow (it is set per-message only on the `infer=False` route at `main.py:865-869`), and inferred memories carry `attributed_to` instead, which has no history column, so the value would be `None` either way.

Thanks again, and sorry for the outcome here. Happy to point you at something else if you want to pick up another one.

PR #6339, opened 2026-07-16
fix(mem0): add test for issue #6266

### LINKED ISSUE #6266: TS OSS Valkey store returns entity ids as camelCase, dropping them from results and leaking into metadata

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6266.

## Why
Fixes #6266

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6266.py
+++ b/tests/test_issue_6266.py
@@ -1,0 +1,103 @@
+"""Regression test for TS OSS Valkey entity id payload shape.
+
+Valkey must return snake_case entity ids in vector-store payloads so the
+TypeScript OSS Memory layer can promote them to top-level result fields instead
+of leaking camelCase ids into metadata.
+"""
+
+import re
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

**Finding:** `tests/test_issue_6266.py` never exercises mem0. It hand-writes a Python model of the TypeScript logic (`_valkey_payload_from_source`, `_memory_search_result_from_vector_payload`) and then asserts against that model. The failure is raised inside the test's own helper:

```
result = _memory_search_result_from_vector_payload(valkey_payload)
>   assert result["user_id"] == "u1"
E   KeyError: 'user_id'
```

The only use of the real package is `assert Memory is not None`. A rewrite of `valkey.ts` could fix or break #6266 without changing this test's outcome.

---

Closing this one. To be clear, this is about the test file, not about issue #6266 — the underlying report is legitimate and stays as it is.

PR #6343, opened 2026-07-16
fix(ts-oss): don't let update() metadata overwrite user_id/agent_id/run_id

### LINKED ISSUE #6342: update() metadata silently overwrites user_id/agent_id/run_id in TS SDK, breaking tenant isolation
### Summary

In the TypeScript OSS SDK, `memory.update()` lets caller-supplied `metadata` silently overwrite a memory's `user_id` / `agent_id` / `run_id`. These are the fields that scope every `search()`, `getAll()`, and `deleteAll()` call, so overwriting them silently moves a memory into another tenant's scope — a cross-tenant data leak with no error raised.

This is the TypeScript twin of the Python bug #6277 (`P0-critical`), which was fixed for Python only in #6278. The TypeScript SDK was left with the same hole.

### Where

`mem0-ts/src/oss/src/memory/index.ts`, in `updateMemory()`:

```ts
const newMetadata = {
  ...existingMemory.payload,
  ...metadata,          // caller-supplied keys win — clobbers identity fields
  data: newData,
  hash: ...,
  textLemmatized: ...,
};
```

Only `createdAt`/`updatedAt` are re-asserted after the spread. `user_id` / `agent_id` / `run_id` are not, so caller `metadata` wins.

This is a regression: #5480 removed the identity-key whitelist that previously protected these fields here.

### Reproduction

```ts

// Tenant A stores a memory
const { results } = await memory.add("Tenant A secret", { userId: "tenant_a", infer: false });
const id = results[0].id;

// A normal update that also passes metadata...
await memory.update(id, {
  text: "Updated text",
  metadata: { user_id: "tenant_b" },
});

// The memory is now gone from tenant A...
await memory.getAll({ filters: { user_id: "tenant_a" } }); // does NOT contain id
// ...and has leaked into tenant B
await memory.getAll({ filters: { user_id: "tenant_b" } }); // contains id
```

Tenant A can no longer read or `deleteAll()` its own memory; tenant B now sees it.

### Impact

- Silent cross-tenant data movement on every TS OSS vector store.
- Affects any `update()` call that carries a `metadata` object.
- No exception is raised — the memory just disappears from the owning scope.

### Expected

`user_id` / `agent_id` / `run_id` must be immutable after creation, exactly as the Python SDK now enforces after #6278. Caller `metadata` on `update()` must never re-scope a memory.

### Fix

Re-pin the identity fields from the existing payload after the spread (mirror of the Python fix in #6278). PR #6343 with the fix and a regression test is open against this issue.

### LINKED ISSUE #6367: TypeScript OSS SDK: updateMemory() has no identity-key protection (actor_id/user_id/agent_id/run_id)
Flagged during review of #6278 (Python SDK fix for the same class of bug, #6277).

## Description

\`mem0-ts/src/oss/src/memory/index.ts\`'s \`updateMemory()\` builds the new payload via an unconditional spread:

\`\`\`ts
const newMetadata = {
  ...existingMemory.payload,
  ...metadata,
  data: newData,
  ...
};
```

Confirmed by reading the current source — there is no filtering of `metadata` before the spread, so any of `actor_id`/`user_id`/`agent_id`/`run_id` passed to `update()` overwrites (or, for a memory that never had one set, injects) the corresponding value in the stored payload. Since these fields scope every `search()`/`getAll()`/`deleteAll()` call, this can silently move a memory into another tenant's scope, make it unreachable via the original tenant's `deleteAll()`, or grant it an additional reachable scope it never had.

This is the same bug class as #6277 (Python SDK), but the TS SDK never got the original `actor_id` fix from #4490 either — it currently has **zero** protection, not even a partial one.

## Suggested fix

Mirror the Python SDK's fix from #6278: strip the identity keys from a copy of the incoming `metadata` before spreading it into `newMetadata`, e.g.

\`\`\`ts
const IDENTITY_KEYS = ["actor_id", "user_id", "agent_id", "run_id"];
const sanitizedMetadata = metadata
  ? Object.fromEntries(Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)))
  : metadata;

const newMetadata = {
  ...existingMemory.payload,
  ...sanitizedMetadata,
  ...
};
\`\`\`

## Scope

Python SDK is out of scope here (already fixed in #6278) — this is TS-only, to keep the two SDKs from drifting further apart on tenant-isolation guarantees.

### PR DESCRIPTION
**Pull request by OfficialAbhinavSingh**
## Linked Issue

Closes #6342
Closes #6367

## Description

In the TypeScript OSS SDK, `memory.update()` merges caller-supplied `metadata` into a memory's stored payload without protecting its identity fields (`actor_id` / `user_id` / `agent_id` / `run_id`). `updateMemory()` builds the new payload as:

```ts
const newMetadata = {
  ...existingMemory.payload,
  ...metadata,          // caller keys win — clobbers / injects identity fields
  ...
};
```

Only `createdAt`/`updatedAt` are re-asserted after the spread, so caller `metadata` can:

- **Overwrite** an existing scope — `update(id, { metadata: { user_id: "tenant_b" } })` silently moves a memory into another tenant's scope: it disappears from the owning tenant (unreadable, undeletable) and surfaces for another.
- **Inject** a brand-new scope — a memory created with only a subset of identity fields (e.g. `run_id` only, which `add()`/`getAll()` allow) can have a fresh `user_id`/`agent_id`/`actor_id` injected via `update()` metadata, granting it a reachable scope it never had.

Those fields scope every `search()` / `getAll()` / `deleteAll()` call, so both cases silently break tenant isolation. No error is raised.

This is the TypeScript twin of #6277 (`P0-critical`), fixed for Python in #6278. The TS SDK had **zero** protection here — it never received the original `actor_id` fix from #4490 either. #6367 tracks the TS side and describes the correct fix.

This PR strips the identity keys (`actor_id` / `user_id` / `agent_id` / `run_id`) from a copy of the incoming `metadata` before the spread, mirroring the strip-before-merge approach the Python fix landed in #6278. Existing scopes on the memory survive (they come from `...existingMemory.payload`), caller metadata can neither overwrite nor inject an identity field, and benign metadata keys still pass through normally.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — the affected behavior (caller metadata re-scoping or scope-injecting a memory across tenants) was never intended and matches the Python SDK's post-#6278 semantics.

## Test Coverage

Added two regression tests in `mem0-ts/src/oss/tests/memory.crud.test.ts`:

1. **Overwrite** — adds a memory under `tenant_a`, calls `update()` with `metadata.user_id = tenant_b`, asserts the memory stays in `tenant_a`'s scope and never appears in `tenant_b`'s.
2. **Injection** — adds a `run_id`-only memory, calls `update()` with injected `user_id`/`agent_id`/`actor_id`, asserts it stays reachable under its original `run_id` and never resolves under the injected scopes.

Both fail without the fix (memory leaks / scope injected) and pass with it. Full `memory.crud.test.ts` suite (52 tests) passes; `tsc`, `prettier --check`, and the `tsup` build are all clean.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -101,12 +101,8 @@ const ENTITY_PARAMS = [
   "runId",
 ];
 
-// Identity fields that scope every search()/getAll()/deleteAll() call and are
-// immutable after creation. Derived from ENTITY_PARAMS (which already covers
-// user_id/agent_id/run_id in both snake_case and camelCase — the default vector
-// store promotes the camelCase aliases to snake_case on read) plus actor_id,
-// which has no camelCase alias. update() strips these from caller metadata so it
-// can neither overwrite an existing scope nor inject a new one.
+// Identity keys stripped from update() metadata: ENTITY_PARAMS covers user_id/agent_id/run_id
+// in both casings (the default store promotes camelCase on read); actor_id has no camelCase alias.
 const IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"];
 
 /**
@@ -1949,16 +1945,10 @@ export class Memory {
       existingEmbeddings[newData] ||
       (await this.embedder.embed(newData, "update"));
 
-    // Strip identity keys from the incoming metadata before the spread so
-    // caller-supplied metadata can neither overwrite an existing scope nor
-    // inject a brand-new one on a memory that never had it — either would
-    // silently move the memory across tenants or grant it a scope it never had.
-    // Mirrors the Python fix (issues #6277 / #6278, TS twin #6342 / #6367).
-    const sanitizedMetadata = metadata
-      ? Object.fromEntries(
-          Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)),
-        )
-      : metadata;
+    // Caller metadata must not overwrite or inject an identity scope (#6342 / #6367).
+    const sanitizedMetadata = Object.fromEntries(
+      Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)),
+    );
 
     const newMetadata = {
       ...existingMemory.payload,
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1941,27 +1941,27 @@ export class Memory {
       existingEmbeddings[newData] ||
       (await this.embedder.embed(newData, "update"));
 
+    // Identity fields scope every search()/getAll()/deleteAll() call and are
+    // immutable after creation. Strip them from the incoming metadata before the
+    // spread so caller-supplied metadata can neither overwrite an existing scope
+    // nor inject a brand-new one on a memory that never had it — either would
+    // silently move the memory across tenants or grant it a scope it never had.
+    // Mirrors the Python fix (issues #6277 / #6278, TS twin #6342 / #6367).
+    const IDENTITY_KEYS = ["actor_id", "user_id", "agent_id", "run_id"];
+    const sanitizedMetadata = metadata
+      ? Object.fromEntries(
+          Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)),
+        )
+      : metadata;
+
     const newMetadata = {
       ...existingMemory.payload,
-      ...metadata,
+      ...sanitizedMetadata,
       data: newData,
       hash: createHash("md5").update(newData).digest("hex"),
       textLemmatized: lemmatizeForBm25(newData),
       createdAt: existingMemory.payload.createdAt,
       updatedAt: new Date().toISOString(),
-      // Identity fields scope every search()/getAll()/deleteAll() call and are
-      // immutable after creation. Re-pin them after the spread so caller-supplied
-      // metadata can never silently move a memory into another tenant's scope
-      // (mirror of the Python P0 fix, issues #6277 / #6278).
-      ...(existingMemory.payload.user_id !== undefined && {
-        user_id: existingMemory.payload.user_id,
-      }),
-      ...(existingMemory.payload.agent_id !== undefined && {
-        agent_id: existingMemory.payload.agent_id,
-      }),
-      ...(existingMemory.payload.run_id !== undefined && {
-        run_id: existingMemory.payload.run_id,
-      }),
     };
 
     await this.vectorStore.update(memoryId, embedding, newMetadata);
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -101,6 +101,14 @@ const ENTITY_PARAMS = [
   "runId",
 ];
 
+// Identity fields that scope every search()/getAll()/deleteAll() call and are
+// immutable after creation. Derived from ENTITY_PARAMS (which already covers
+// user_id/agent_id/run_id in both snake_case and camelCase — the default vector
+// store promotes the camelCase aliases to snake_case on read) plus actor_id,
+// which has no camelCase alias. update() strips these from caller metadata so it
+// can neither overwrite an existing scope nor inject a new one.
+const IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"];
+
 /**
  * Validates that no top-level entity parameters are passed in config.
  * @throws Error if entity params are found at top level
@@ -1941,13 +1949,11 @@ export class Memory {
       existingEmbeddings[newData] ||
       (await this.embedder.embed(newData, "update"));
 
-    // Identity fields scope every search()/getAll()/deleteAll() call and are
-    // immutable after creation. Strip them from the incoming metadata before the
-    // spread so caller-supplied metadata can neither overwrite an existing scope
-    // nor inject a brand-new one on a memory that never had it — either would
+    // Strip identity keys from the incoming metadata before the spread so
+    // caller-supplied metadata can neither overwrite an existing scope nor
+    // inject a brand-new one on a memory that never had it — either would
     // silently move the memory across tenants or grant it a scope it never had.
     // Mirrors the Python fix (issues #6277 / #6278, TS twin #6342 / #6367).
-    const IDENTITY_KEYS = ["actor_id", "user_id", "agent_id", "run_id"];
     const sanitizedMetadata = metadata
       ? Object.fromEntries(
           Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)),
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1949,6 +1949,19 @@ export class Memory {
       textLemmatized: lemmatizeForBm25(newData),
       createdAt: existingMemory.payload.createdAt,
       updatedAt: new Date().toISOString(),
+      // Identity fields scope every search()/getAll()/deleteAll() call and are
+      // immutable after creation. Re-pin them after the spread so caller-supplied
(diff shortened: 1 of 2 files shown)


OfficialAbhinavSingh:
Updated in `cb185cf` — switched to stripping the identity keys (incl. `actor_id`) from a copy of the incoming `metadata` before the spread, per your triage on #6367 and the approach in #6278. This closes the injection case too, not just overwrite: added a regression test where a `run_id`-only memory with injected `user_id`/`agent_id`/`actor_id` in `update()` metadata stays out of those scopes. Should fully close #6367 rather than just narrow it.

kartik-mem0:
## Changes Requested

Thanks for pushing `cb185cf` — that's the right fix shape. Reviewed both commits, ran the full targeted test suite against a worktree at `cb185cfc736436a93767263cadfb5e11281773ba`, and independently re-verified the injection-gap concern from the earlier review round.

### Strengths
- `mem0-ts/src/oss/src/memory/index.ts:1944-1959` — switching from the conditional re-pin (`ada2b79`) to stripping identity keys out of a **copy** of the caller's metadata before the spread is the correct fix shape. It's an exact structural mirror of the Python sibling's fix in #6278 (`_IDENTITY_KEYS` filter + `deepcopy(existing_memory.payload)` merge) — same 4 keys, same unconditional strip-before-merge mechanism.
- `mem0-ts/src/oss/tests/memory.crud.test.ts:216-284` — the two new regression tests exercise real behavior through `getAll()` against the actual default vector store, not a mock echo. Verified with a red-green run: both tests fail against unmodified `main`, the overwrite test alone passes at `ada2b79` (proving the earlier round's "conditional re-pin leaves an injection gap" concern was real, not hypothetical), and both pass at `cb185cf`.
- Doesn't mutate the caller's own `metadata` object in place (`Object.fromEntries`/`.filter` builds a new object) — no surprising aliasing side effect.

### Issues

#### Critical (must fix)
1. `mem0-ts/src/oss/src/memory/index.ts:1950` — `IDENTITY_KEYS` only lists the snake_case key names (`actor_id`, `user_id`, `agent_id`, `run_id`), but the default vector store's own read path promotes camelCase aliases to snake_case: `MemoryVectorStore.normalizePayload()` in `mem0-ts/src/oss/src/vector_stores/memory.ts:28-38` does `if (camel in payload && !(snake in payload)) payload[snake] = payload[camel]`, for `userId`/`agentId`/`runId` (`memory.ts:22-26`). `provider: "memory"` is the default vector store (`src/oss/src/config/defaults.ts:14`), so this isn't a niche-provider edge case.

   I reproduced this live at `cb185cfc` (temporary test, reverted after — not proposing it as-is):
   ```
   create memory with only run_id set
   update(id, { metadata: { userId: "attacker", agentId: "attacker" } })
   getAll({ filters: { user_id: "attacker" } })  →  now contains the memory
   ```
   ```
   ● SKEPTIC-CHECK: metadata cannot inject identity field via camelCase alias
     expect(received).not.toContain(expected)
     Received array: ["233d87d8-3149-4aae-8882-7beb3a717ee4"]
   ```
   This is exactly the injection case #6367 was filed to close, still open via the camelCase alias, on the exact method the issue names.

   Good news on scope: I also checked whether this same gap reopens the **overwrite** case (#6342) via camelCase — it doesn't. `normalizePayload`'s own `!(snake in payload)` guard happens to block promotion when the real `user_id` key already exists, so an existing scope stays protected even through the camelCase alias (verified with a second temporary test — passes). So **#6342 is fully closed** by this PR, including that edge case. The remaining gap is scoped precisely to injection on `update()`.

   Suggested fix: extend the `IDENTITY_KEYS` filter to also strip `userId`/`agentId`/`runId` (matching `CAMEL_TO_SNAKE`'s own key set), or normalize `metadata`'s keys before filtering.

#### Important (should fix, can be a fast follow-up rather than blocking this PR)
1. `mem0-ts/src/oss/src/memory/index.ts:742-744`, `:986-999`, `:1890-1907` — `add()` has the same bug *class* this PR fixes for `update()`: `filters.agent_id`/`run_id` only get re-pinned onto the stored payload conditionally (`if (filters.agent_id) memPayload.agent_id = ...`), so caller-supplied `metadata.agent_id`/`run_id`/`actor_id` survives untouched into a **newly created** memory whenever the matching top-level config param isn't also passed — and `actor_id` has no re-pin path in `add()` at all. I confirmed this is real by reading the code, but it's a distinct method from the one #6367 names (I re-read the issue text directly — it's scoped exclusively to `updateMemory()`, no mention of `add()`), so it doesn't block this PR's claim to close #6367. Worth its own issue given it's the same root cause on the creation path.

#### Minor (optional)
1. `mem0-ts/src/oss/src/memory/index.ts:1950` vs the pre-existing `ENTITY_PARAMS` constant elsewhere in the same file — now two independently hand-maintained "which fields are identity fields" lists exist in one file, and they already disagree (`ENTITY_PARAMS` has no `actor_id`). Not urgent, but worth a comment or a shared constant so a future 5th identity field doesn't get added to only one of them.
2. `mem0-ts/src/oss/src/memory/index.ts:1948` — `IDENTITY_KEYS` is reallocated on every `updateMemory()` call; hoisting to module scope would match the Python side's module-level `_IDENTITY_KEYS` and avoid the reallocation (negligible perf either way).

### Evidence
- **CI:** `Vercel` = failure (advisory preview-deploy, not blocking). `license/cla` = success. The actual `CI Gate` (lint/build/jest) run shows `action_required` — it hasn't executed yet, gated behind maintainer approval for first-time contributors. This review's own targeted-test run below is what's load-bearing right now, not GitHub's check-runs.
- **Targeted tests / red-green:** Full `memory.crud.test.ts` — 52/52 pass at `cb185cf`. `prettier --check` clean. `tsc --noEmit` shows only pre-existing errors in files this PR doesn't touch. 3-point red-green proof (`main` → `ada2b79` → `cb185cf`) confirms the PR's own two tests are load-bearing (they fail without the fix, and the intermediate commit shows the injection test specifically catches what the conditional re-pin missed).
- **Could NOT verify:** Whether the camelCase gap is exploitable through any vector-store provider other than the default `memory` one — didn't have credentials to test against Qdrant/Pinecone/etc., but the vulnerable code (`updateMemory()`'s `IDENTITY_KEYS` filter) runs before the vector-store-specific call, so it's provider-agnostic; only the *promotion-on-read* half of the exploit path is `memory`-provider-specific, and that's the default.

### Assessment
Ready to merge: **With fixes.** #6342 is fully closed by `cb185cf` — nice work chasing down the injection case from the earlier round. #6367 isn't quite there yet: the camelCase alias bypass above needs the same treatment before this can close it. Happy to re-review as soon as you push.

OfficialAbhinavSingh:
Thanks for the thorough re-review — all addressed in `1572f83`:

**Critical (camelCase alias):** Fixed. `IDENTITY_KEYS` is now derived from the existing `ENTITY_PARAMS` list (which already carries both casings of `user_id`/`agent_id`/`run_id`) plus `actor_id`, so the strip covers the `userId`/`agentId`/`runId` aliases that `normalizePayload()` promotes on read. Added a regression test reproducing exactly your case — `run_id`-only memory + `update({ metadata: { userId, agentId } })` → stays out of those scopes (fails at `cb185cf`, passes at `1572f83`). This should now fully close #6367.

**Minor #1 (two divergent identity lists):** Addressed by deriving `IDENTITY_KEYS` from `ENTITY_PARAMS` (+ `actor_id`) rather than a second hand-maintained list, so a future field added to `ENTITY_PARAMS` propagates automatically.

**Minor #2 (module scope):** Done — hoisted `IDENTITY_KEYS` to module scope, matching the Python side's module-level `_IDENTITY_KEYS`.

**Important (add() same class):** Agreed it's a distinct method and out of scope for #6367 — filed separately as #6371 so it can be tracked/fixed on its own.

Full `memory.crud.test.ts` suite is 53/53, `prettier`/`tsc`/`tsup` clean. Ready for re-review whenever you have a moment.

kartik-mem0 (approved):

PR #6345, opened 2026-07-16
fix(llms/openai): fall back to reasoning_content when content is empty

### LINKED ISSUE #4932: Fact extraction returns empty/garbage with reasoning models on OpenAI-compatible APIs that separate reasoning_content from content

Bartok9:
## Summary

- `OpenAILLM._parse_response` now falls back to `message.reasoning_content` when `message.content` is empty (non-tools path).
- Adds three regression tests for fallback, content-priority, and both-empty behavior.
- Focused salvage of #4937 by @octo-patch (stale since Apr 2026) without the unrelated `examples/mem0-demo` UI diff.

## Motivation

Closes #4932.

Reasoning models on several OpenAI-compatible endpoints (Ollama Cloud, DeepSeek, native GLM, etc.) put the usable answer in `reasoning_content` and leave `content` empty. Mem0 then treated extraction as "no facts" with an empty results list and no error.

## Root cause

`_parse_response` only read `response.choices[0].message.content`.

## Fix

If `content` is falsy, use `getattr(message, "reasoning_content", None)` before returning.

## Differential value vs #4937

- Same core fix credited to @octo-patch.
- Rebased onto current `main` only.
- Drops the unrelated demo frontend type-tightening that was mixed into #4937.

## Tests

```bash
pytest tests/llms/test_openai.py -k reasoning -q
```

All targeted tests pass. (A few pre-existing env-sensitive tests in the same file fail under local `OPENROUTER_API_KEY` / store-param fixture drift — unrelated to this change.)

## Verification

- [x] Fallback returns `reasoning_content` when content is empty
- [x] Non-empty `content` still wins
- [x] Missing `reasoning_content` does not crash

### DIFF
--- a/mem0-ts/src/oss/src/llms/openai.ts
+++ b/mem0-ts/src/oss/src/llms/openai.ts
@@ -37,10 +37,15 @@ export class OpenAILLM implements LLM {
     });
 
     const response = completion.choices[0].message;
+    const textContent =
+      response.content ||
+      // OpenAI-compatible reasoning models may put the answer in reasoning_content (#4932)
+      ((response as { reasoning_content?: string | null }).reasoning_content ??
+        "");
 
     if (response.tool_calls) {
       return {
-        content: response.content || "",
+        content: textContent,
         role: response.role,
         toolCalls: response.tool_calls.map((call) => ({
           name: call.function.name,
@@ -49,7 +54,7 @@ export class OpenAILLM implements LLM {
       };
     }
 
-    return response.content || "";
+    return textContent;
   }
 
   async generateChat(messages: Message[]): Promise<LLMResponse> {
@@ -67,8 +72,12 @@ export class OpenAILLM implements LLM {
       model: this.model,
     });
     const response = completion.choices[0].message;
+    const textContent =
+      response.content ||
+      ((response as { reasoning_content?: string | null }).reasoning_content ??
+        "");
     return {
-      content: response.content || "",
+      content: textContent,
       role: response.role,
     };
   }
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -4,18 +4,21 @@ from typing import Dict, List, Optional, Union
 from mem0.configs.llms.base import BaseLlmConfig
 
 
-
 def message_content_with_reasoning_fallback(message):
     """Return chat message text, falling back to reasoning_content when empty.
 
     Some OpenAI-compatible endpoints (DeepSeek, Ollama Cloud, GLM, etc.) put the
     full answer in ``reasoning_content`` and leave ``content`` empty for reasoning
     models. Fact extraction must not silently return [] in that case (#4932).
+
+    Always returns a string (empty when both fields are missing/empty) so callers
+    like ``remove_code_blocks`` can safely call ``.strip()``.
     """
     content = getattr(message, "content", None)
     if not content:
         content = getattr(message, "reasoning_content", None)
-    return content
+    return content or ""
+
 
 class LLMBase(ABC):
     """
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -4,6 +4,19 @@ from typing import Dict, List, Optional, Union
 from mem0.configs.llms.base import BaseLlmConfig
 
 
+
+def message_content_with_reasoning_fallback(message):
+    """Return chat message text, falling back to reasoning_content when empty.
+
+    Some OpenAI-compatible endpoints (DeepSeek, Ollama Cloud, GLM, etc.) put the
+    full answer in ``reasoning_content`` and leave ``content`` empty for reasoning
+    models. Fact extraction must not silently return [] in that case (#4932).
+    """
+    content = getattr(message, "content", None)
+    if not content:
+        content = getattr(message, "reasoning_content", None)
+    return content
+
 class LLMBase(ABC):
     """
     Base class for all LLM providers.
--- a/tests/llms/test_openai.py
+++ b/tests/llms/test_openai.py
@@ -464,3 +464,93 @@ def test_openai_llm_preserves_proxies_from_base_config(mock_openai_client):
     llm = OpenAILLM(config)
     assert llm.config.http_client_proxies == "http://proxy.local:8080"
     assert isinstance(llm.config.http_client, httpx.Client)
+
+
+
+def test_generate_response_falls_back_to_reasoning_content(mock_openai_client):
+    """When content is empty and reasoning_content is present, return reasoning_content.
+
+    Some OpenAI-compatible endpoints (Ollama, DeepSeek, GLM, etc.) put the full
+    answer inside `reasoning_content` and leave `content` empty when a reasoning
+    model is used. Without the fallback, fact extraction silently returns empty
+    results (fixes #4932).
+    """
+    config = OpenAIConfig(model="glm-5.1", temperature=0.0)
+    llm = OpenAILLM(config)
+    messages = [{"role": "user", "content": "What do I know?"}]
+
+    mock_message = Mock()
+    mock_message.content = ""
+    mock_message.reasoning_content = '{"facts": ["User lives in Guadalajara"]}'
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=mock_message)]
+    mock_openai_client.chat.completions.create.return_value = mock_response
+
+    response = llm.generate_response(messages)
+
+    assert response == '{"facts": ["User lives in Guadalajara"]}'
+
+
+def test_generate_response_content_takes_priority_over_reasoning_content(mock_openai_client):
+    """When content is non-empty, reasoning_content should not be used."""
+    config = OpenAIConfig(model="glm-5.1", temperature=0.0)
+    llm = OpenAILLM(config)
+    messages = [{"role": "user", "content": "Hello"}]
+
+    mock_message = Mock()
+    mock_message.content = "Hello there!"
+    mock_message.reasoning_content = "internal thoughts"
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=mock_message)]
+    mock_openai_client.chat.completions.create.return_value = mock_response
+
+    response = llm.generate_response(messages)
+
+    assert response == "Hello there!"
+
+
+def test_generate_response_returns_none_when_both_content_fields_empty(mock_openai_client):
+    """When both content and reasoning_content are absent/empty, return None."""
+    config = OpenAIConfig(model="some-model", temperature=0.0)
+    llm = OpenAILLM(config)
+    messages = [{"role": "user", "content": "Hello"}]
+
+    mock_message = Mock(spec=["content"])
+    mock_message.content = ""
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=mock_message)]
+    mock_openai_client.chat.completions.create.return_value = mock_response
+
+    response = llm.generate_response(messages)
+
+    assert response is None
+
+
+def test_generate_response_tools_path_falls_back_to_reasoning_content(mock_openai_client):
+    """Tools branch must also fall back when content is empty (#4932 kartik review)."""
+    config = OpenAIConfig(model="glm-5.1", temperature=0.0)
+    llm = OpenAILLM(config)
+    messages = [{"role": "user", "content": "Add a fact"}]
+    tools = [
+        {
+            "type": "function",
+            "function": {
+                "name": "add_memory",
+                "description": "Add a memory",
+                "parameters": {"type": "object", "properties": {"data": {"type": "string"}}},
+            },
+        }
+    ]
+
+    mock_message = Mock()
+    mock_message.content = ""
+    mock_message.reasoning_content = "should surface in tools content"
+    mock_message.tool_calls = None
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=mock_message)]
+    mock_openai_client.chat.completions.create.return_value = mock_response
+
+    response = llm.generate_response(messages, tools=tools)
+
+    assert response["content"] == "should surface in tools content"
+    assert response["tool_calls"] == []
--- a/tests/llms/test_openai.py
+++ b/tests/llms/test_openai.py
@@ -466,7 +466,6 @@ def test_openai_llm_preserves_proxies_from_base_config(mock_openai_client):
     assert isinstance(llm.config.http_client, httpx.Client)
 
 
-
 def test_generate_response_falls_back_to_reasoning_content(mock_openai_client):
     """When content is empty and reasoning_content is present, return reasoning_content.
 
@@ -509,8 +508,8 @@ def test_generate_response_content_takes_priority_over_reasoning_content(mock_op
     assert response == "Hello there!"
 
 
-def test_generate_response_returns_none_when_both_content_fields_empty(mock_openai_client):
-    """When both content and reasoning_content are absent/empty, return None."""
+def test_generate_response_returns_empty_string_when_both_content_fields_empty(mock_openai_client):
+    """When both content and reasoning_content are absent/empty, return empty string."""
     config = OpenAIConfig(model="some-model", temperature=0.0)
     llm = OpenAILLM(config)
     messages = [{"role": "user", "content": "Hello"}]
@@ -523,7 +522,7 @@ def test_generate_response_returns_none_when_both_content_fields_empty(mock_open
 
     response = llm.generate_response(messages)
 
-    assert response is None
+    assert response == ""
 
 
 def test_generate_response_tools_path_falls_back_to_reasoning_content(mock_openai_client):
--- a/mem0-ts/src/oss/tests/openai-llm.test.ts
+++ b/mem0-ts/src/oss/tests/openai-llm.test.ts
@@ -119,6 +119,50 @@ describe("OpenAILLM (unit)", () => {
     });
   });
 
+  it("generateResponse() falls back to reasoning_content when content is empty", async () => {
+    mockCreate.mockResolvedValueOnce({
+      choices: [
+        {
+          message: {
+            content: "",
+            reasoning_content: '{"facts": ["User lives in Guadalajara"]}',
+            role: "assistant",
+            tool_calls: null,
+          },
+        },
+      ],
+    });
+
+    const llm = new OpenAILLM({ apiKey: "test-key" });
+    const result = await llm.generateResponse([
+      { role: "user", content: "Hi" },
+    ]);
+
+    expect(result).toBe('{"facts": ["User lives in Guadalajara"]}');
+  });
+
+  it("generateResponse() prefers content over reasoning_content", async () => {
+    mockCreate.mockResolvedValueOnce({
+      choices: [
+        {
+          message: {
+            content: "final answer",
+            reasoning_content: "chain of thought",
+            role: "assistant",
+            tool_calls: null,
+          },
+        },
+      ],
+    });
+
+    const llm = new OpenAILLM({ apiKey: "test-key" });
+    const result = await llm.generateResponse([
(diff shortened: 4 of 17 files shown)


kartik-mem0 (commented):
## Changes requested

Fixes the reasoning_content fallback for `openai.py`'s non-tools path, but the identical bug stays live in 8 sibling providers and the TS SDK, closing an issue that names one directly.

**To unblock:**
1. `mem0/llms/deepseek.py` (+7 more siblings), same bug, issue names DeepSeek, still broken after close
2. `mem0/llms/openai.py:66-81`, same function's `if tools:` branch unpatched
3. `mem0-ts/src/oss/src/llms/openai.ts:47`, TS OSS SDK has the identical gap

Optional: two Minor nits posted inline (weak test assertion, blank line spacing).

- **Problem is real:** reproduces on `main` at `mem0/llms/openai.py:83`, empty `content` from a reasoning model on an OpenAI-compatible endpoint never falls back, fact extraction silently returns `[]`.
- **Tests / red-green:** `pytest tests/llms/test_openai.py -v` -> 24 passed. Reverted `openai.py` to main, kept new tests -> failed. Restored the fix -> passed again.
- **CI:** green on tests, lint, build. Vercel preview red, ignored, fork perms only.
- **Security:** cleared, no new input surface, same trusted LLM response object.
- **Bloat (ponytail):** lean already, net -0 lines possible.
- **Not verified:** live DeepSeek/Ollama Cloud/GLM endpoints, code traced only.

</details>

The fallback itself is correct and well tested. Extending the same pattern to the siblings closes this out. Push when ready and I will re-review.

tests/llms/test_openai.py
```diff
@@ -464,3 +464,62 @@ def test_openai_llm_preserves_proxies_from_base_config(mock_openai_client):
...
+
+    mock_message = Mock(spec=["content"])
+    mock_message.content = ""
+    mock_response = Mock()
+    mock_response.choices = [Mock(message=mock_message)]
+    mock_openai_client.chat.completions.create.return_value = mock_response
+
+    response = llm.generate_response(messages)
+
+    assert response is None or response == ""
```
kartik-mem0: `Mock(spec=["content"])` blocks `reasoning_content`, so `getattr(...)` always returns `None` here, never `""`. The `or response == ""` clause can never fire, so this test still passes if a future change returns `""` instead of `None`. Assert `response is None` only.

tests/llms/test_openai.py
```diff
@@ -464,3 +464,62 @@ def test_openai_llm_preserves_proxies_from_base_config(mock_openai_client):
     llm = OpenAILLM(config)
     assert llm.config.http_client_proxies == "http://proxy.local:8080"
     assert isinstance(llm.config.http_client, httpx.Client)
+
+def test_generate_response_falls_back_to_reasoning_content(mock_openai_client):
```
kartik-mem0, line 469: Every other top-level test in this file has two blank lines before it (PEP 8). This one has one. Add a blank line to match.

Bartok9:
@kartik-mem0 tightened the empty-content test to `assert response is None` only, and fixed the double-blank before the new top-level tests. Thanks for the nits.

Bartok9:
Thanks @kartik-mem0 — rebuilt on current `main` and extended the fix across the full OpenAI-compatible surface (tip `03db7d34`):

1. **Shared helper** `message_content_with_reasoning_fallback()` on `mem0/llms/base.py` — empty/missing `content` falls back to `reasoning_content`.
2. **`openai.py` tools branch** — same fallback on the `if tools:` path (was unpatched before).
3. **DeepSeek + siblings** — `deepseek`, `lmstudio`, `minimax`, `groq`, `litellm`, `together`, `vllm`, `xai` (the 8 OpenAI-compatible parsers), plus Azure + `openai_structured` for the same `choices[0].message.content` gap.
4. **TS OSS** `mem0-ts/src/oss/src/llms/openai.ts` — `generateResponse` and `generateChat` both fall back.
5. **Tests** — openai (non-tools + tools path + priority + both-empty → `None` only), deepseek named in the issue, TS unit coverage.
6. **CI** — tip is a single commit on current main so `changelog_check` no longer sees a stale 2.0.12→2.0.14 skew.

Ready for re-review.

Bartok9:
CI fix (watch): `TypeScript SDK / build_ts_sdk` was failing prettier on `mem0-ts/src/oss/tests/openai-llm.test.ts` (extra blank line from the reasoning_content coverage). Rebuilt on latest `main` with prettier clean.

- tip: `7ed43943`
- product scope unchanged: shared `message_content_with_reasoning_fallback` + OpenAI-compatible providers + TS OSS OpenAI path
- local: `npx prettier --check src/oss/tests/openai-llm.test.ts` clean

tests/llms/test_openai.py
kartik-mem0: Verified against 7ed4394 (new head): fixed. Assertion at tests/llms/test_openai.py:526 is now `assert response is None` only.

tests/llms/test_openai.py
kartik-mem0: Verified against 7ed4394 (new head): still reproduces, in the opposite direction. This spot now has 3 blank lines instead of 2 (tests/llms/test_openai.py:470, `ruff format --diff` confirms). Same overcorrection landed in tests/llms/test_deepseek.py and mem0/llms/base.py (3 before the new function, 1 after instead of 2). Not CI-blocking, ruff format is not a required check, but worth a pass before merge.

kartik-mem0 (commented):
## Changes requested

Good progress on the prior round: sibling providers, the tools-branch, the shared-helper layering, and the weak assertion are all fixed and red-green proven. Two things still block.

**To unblock:**
1. `mem0/llms/base.py:18` - return `content or ""` (detail in the inline comment)
2. `mem0-ts/src/oss/src/llms/openai.ts:78` - add a `generateChat()` regression test for the fallback

Optional: blank-line overcorrection in `base.py` / `test_deepseek.py` / `test_openai.py` (replied in-thread), TS fallback expression duplicated across 2 methods (Minor, no functional impact), `sarvam.py` shares the OpenAI-compatible shape but parses via raw dict indexing so wasn't patched, worth a follow-up.

- **Problem is real:** unchanged from the prior round, reproduces on `main` at `mem0/llms/openai.py:83` pre-PR.
- **Tests / red-green:** Python 31 passed, reverting `base.py`/`openai.py`/`deepseek.py` to merge-base fails the 4 expected tests, restoring passes again. TS 8 passed, reverting `openai.ts` fails only the `generateResponse` fallback test, `generateChat`'s identical fallback logic is reverted too but nothing catches it, proving the coverage gap empirically.
- **CI:** green on all required checks (Python 3.10/3.11/3.12, TS build). Vercel red is fork-permissions, non-blocking.
- **New this round:** `message_content_with_reasoning_fallback` returns `None` when both fields are empty (intentional, tested at the LLM layer). That `None` reaches `remove_code_blocks()` unguarded via `_create_procedural_memory` (both sync `main.py:1984-1989` and async `main.py:3667-3672`), which calls `.strip()` unconditionally and raises, re-raised uncaught. Confirmed by direct read and a fresh skeptic pass; not a pre-existing bug, empty content used to pass through as `""` pre-PR, now becomes `None`.
- **Security:** cleared, no new attack surface, this is an internal control-flow regression not an input-validation gap.
- **Bloat (ponytail):** lean, no unnecessary abstraction added.
- **Not verified:** live behavior against real DeepSeek/Ollama Cloud/GLM endpoints, non-string `reasoning_content` payloads, real-traffic frequency of the empty-completion case.

</details>

The sibling-provider and tools-branch fixes are the real work here and they're solid. Push when ready and I will re-review.

mem0/llms/base.py
```diff
@@ -4,6 +4,19 @@
...
+    """Return chat message text, falling back to reasoning_content when empty.
+
+    Some OpenAI-compatible endpoints (DeepSeek, Ollama Cloud, GLM, etc.) put the
+    full answer in ``reasoning_content`` and leave ``content`` empty for reasoning
+    models. Fact extraction must not silently return [] in that case (#4932).
+    """
+    content = getattr(message, "content", None)
+    if not content:
+        content = getattr(message, "reasoning_content", None)
+    return content
```
kartik-mem0: This can return `None` (confirmed by the new test at tests/llms/test_openai.py:526, which expects `response is None`). `_create_procedural_memory` in mem0/memory/main.py (sync and async) passes that straight into `remove_code_blocks()`, which calls `.strip()` unconditionally and raises `AttributeError`, re-raised uncaught by the surrounding `except Exception: raise`.

Return `content or ""` here instead, so every caller keeps getting a string like before this PR.

mem0-ts/src/oss/src/llms/openai.ts
```diff
@@ -67,8 +72,12 @@ export class OpenAILLM implements LLM {
       model: this.model,
     });
     const response = completion.choices[0].message;
+    const textContent =
+      response.content ||
+      ((response as { reasoning_content?: string | null }).reasoning_content ??
+        "");
```
kartik-mem0, line 78: Same fallback as generateResponse but untested here. Reverting only this block still leaves all 8 tests in openai-llm.test.ts green, so nothing exercises it.

Add a generateChat() test with empty content plus reasoning_content set, mirroring the generateResponse fallback test.

Bartok9:
Thanks @kartik-mem0 — addressed both blockers on tip `f4205371`:

1. **`message_content_with_reasoning_fallback`** now returns `content or ""` so callers (`remove_code_blocks` / `.strip()`) never see `None`. Updated the both-empty regression to `assert response == ""`.
2. **TS** added `generateChat() falls back to reasoning_content when content is empty` (mirrors `generateResponse`). Local: 9/9 in `openai-llm.test.ts`.

Also cleaned the blank-line overcorrection in `base.py` / `test_openai.py` / `test_deepseek.py`.

Ready for re-review.

mem0/llms/base.py
kartik-mem0: Verified against f4205371: fixed. `message_content_with_reasoning_fallback` now returns `content or ""` (mem0/llms/base.py:20), so the crash path into `_create_procedural_memory` -> `remove_code_blocks(procedural_memory)` (mem0/memory/utils.py:125, unconditional `.strip()`) is closed. Red-green: reverted this line back to `return content`, `tests/llms/test_openai.py::test_generate_response_returns_empty_string_when_both_content_fields_empty` fails with `AssertionError: assert None == ''`; restored, it passes. Traced `remove_code_blocks("")` by hand, returns `""`, no exception.

mem0-ts/src/oss/src/llms/openai.ts
kartik-mem0: Verified against f4205371: fixed. New test `generateChat() falls back to reasoning_content when content is empty` (mem0-ts/src/oss/tests/openai-llm.test.ts) exercises this block directly. Red-green: reverted `openai.ts:75-78` to `content: response.content || ""` (no reasoning_content fallback), the new test fails (expected the reasoning_content JSON, received `""`, 1 failed, 8 passed); restored, 9 passed.

tests/llms/test_openai.py
kartik-mem0: Verified against f4205371: fixed. `tests/llms/test_openai.py:467-469` now has exactly two blank lines before this test (PEP 8), confirmed by `ruff format --diff` no longer flagging this location. Same fix applied at the equivalent spots in `mem0/llms/base.py:5-6` and `tests/llms/test_deepseek.py`. Remaining `ruff format --diff` output on these files is pre-existing drift: it reproduces identically against the merge-base version of `base.py` and `test_openai.py`, in code this PR does not touch.

kartik-mem0 (commented):
## Needs a human call

Fix (`return content or ""` in `mem0/llms/base.py`) is correct and complete. All three prior findings are verified fixed at this head, see the three thread replies above.

**Decide:** CI Gate is red on `changelog_check` (Python SDK + TS SDK). Merge anyway, or wait for a refresh push?
**Why this needs you:** CI. Confirmed infra false positive, not this PR's fault: `pyproject.toml` / `mem0-ts/package.json` are untouched by this diff at both merge-base and head (`gh pr diff --name-only` has no match). `changelog_check` diffs against `pull_request.base.sha` (`d06ea187`), which is 17 commits stale behind current `main`, so it picks up `main`'s own 2.0.14 -> 2.0.15 version bump as if this PR made it.
**Look at:** `.github/workflows/ci.yml` `changelog_check` job, compares `BASE_SHA` vs `HEAD_SHA` pyproject.toml versions; `BASE_SHA` does not track `main`'s live tip once it drifts.
**Your time:** ~5 min

- **Prior findings (verified against `f4205371`):**
  1. `mem0/llms/base.py:20` `None`-return crash into `_create_procedural_memory`: fixed, `return content or ""`. Red-green: revert -> `AssertionError: assert None == ''`; restore -> pass.
  2. `mem0-ts/src/oss/src/llms/openai.ts:75-78` untested `generateChat` fallback: fixed, new test added. Red-green: revert -> `1 failed, 8 passed`; restore -> `9 passed`.
  3. Blank-line overcorrection (`test_openai.py`, `test_deepseek.py`, `base.py`): fixed, `ruff format --diff` no longer flags these locations. Remaining `ruff format --diff` output on these files is pre-existing drift, reproduces identically against the merge-base version, unrelated to this PR's lines.
- **Tests / red-green:** both above, plus full targeted suite: `pytest tests/llms/test_openai.py tests/llms/test_deepseek.py` = 31 passed. `jest openai-llm.test.ts` = 9 passed.
- **Lint:** `ruff check` clean on all 15 touched Python files. `prettier --check` clean on the touched TS files.
- **CI:** CI Gate fail via `changelog_check` on both pipelines, infra false positive as above. Every other required job (build, tests, lint, matrices) is green.
- **Security:** no new inputs, no auth/crypto/migration surface touched.
- **Bloat (ponytail):** none. One-line fix plus one mirrored test.
- **Not verified:** live behavior against real DeepSeek / Ollama Cloud / GLM endpoints (mocked-test-verified only). No duplicate PR found (`gh pr list --search "reasoning_content in:title,body"` returns only #6345).

</details>

Code is ready. The open question is what to do about the CI status above, that is a maintainer call, not something the contributor can fix by editing this PR.

Bartok9:
Thanks @kartik-mem0 — agreed on the code side.

Refresh push already landed on tip `12c51a0d` (same fix commit as `f4205371` content against current main). Fresh CI Gate run now has **Python SDK / changelog_check** and **TypeScript SDK / changelog_check** both **SUCCESS**; remaining jobs still in progress (builds/tests). Vercel authorize + CLA are the usual fork/account noise.

No further product edits needed from this side unless something else turns red. Happy to leave the merge call with you.

kartik-mem0:
Update on the CI question in my last review, which the timing has since answered.

You pushed `12c51a0d` about a minute after that review posted, and `changelog_check` is now green on both the Python and TS pipelines. Refreshing the base moved `pull_request.base.sha` off the stale `d06ea187`, so the job stopped reading main's own 2.0.14 to 2.0.15 bump as though this PR had made it. Worth noting for next time: the same check cleared on another PR today by exactly the same route, so a base refresh is the reliable workaround whenever `changelog_check` goes red while the diff touches no version file.

That makes the "merge anyway or wait for a refresh push" question moot, the refresh landed. The remaining builds are still running at the new head, and the only red check is Vercel, which is deploy auth and not a blocker here.

The three code findings were all verified fixed at `f4205371`. This push is past that head, so I want to read the delta before this moves any further rather than carry a verdict across commits I have not looked at. That is queued. If it turns out to be just the base refresh it should be quick.

Bartok9:
Thanks @kartik-mem0 — confirmed: tip `12c51a0d` is the same product content as `f4205371` rebased onto current `main` (base refresh only; no further code delta). Appreciate you queueing the look. Standing by.

PR #6348, opened 2026-07-16
fix(sdk): delete all matching memories across batches

GautamSharma99:
## Description

  Fixes `Memory.delete_all()` and `AsyncMemory.delete_all()` deleting only the first page of matching memories.

  Vector store providers such as Qdrant and pgvector return at most 100 records by default. The previous implementation called `list()` once and
  returned a success response, leaving any additional matching memories undeleted.

  This change repeatedly lists and deletes batches until no unprocessed memories remain. It also tracks processed IDs to prevent infinite loops
  if a provider returns stale or repeated results.

  Async partial-failure handling and bulk entity cleanup behavior are preserved.

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Refactor
  - [ ] Documentation update

  ## Breaking Changes

  N/A

  ## Test Coverage

  - [x] Added sync regression coverage with 101 memories
  - [x] Added async regression coverage with 101 memories
  - [x] Verified existing async partial-failure and entity-cleanup behavior
  - [x] Ran the complete Python SDK test suite

  Full suite result:

  - 1,677 passed
  - 48 skipped
  - 0 failed

  ## Checklist

  - [x] I have added tests that prove the fix
  - [x] Ruff and isort checks pass
  - [x] Documentation is not required because this does not change the public API

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1869,14 +1869,25 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
-        memories = self.vector_store.list(filters=filters)[0]
-        for memory in memories:
-            self._delete_memory(memory.id)
+        deleted_count = 0
+        processed_memory_ids = set()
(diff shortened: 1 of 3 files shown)


kartik-mem0:
Thanks for this.

Closing as a duplicate. Appreciate you picking it up.

PR #6350, opened 2026-07-16
fix(openmemory): align MCP tools with OSS SDK

GautamSharma99:
## Description

  Fixes two OpenMemory MCP tools that were calling incompatible Mem0 OSS SDK APIs:

  - search_memory now passes top_k=10 to the vector store instead of the unsupported limit argument.
  - list_memories now scopes results with filters={"user_id": uid} instead of the unsupported top-level user_id argument.

  Adds regression tests that invoke both tools through the MCP JSON-RPC flow and verify the exact SDK arguments.

  Also renames the shared SSE message handler to resolve an existing Ruff duplicate-definition error without changing route behavior.

  ## Type of Change

  - [ ] Documentation update

  ## Breaking Changes

  N/A

  ## Test Coverage

  Validated with:

  Result: 23 passed

  Pre-commit checks:

  - Ruff: passed
  - isort: passed

  ## Checklist

  - [x] I have updated documentation if needed (no documentation changes required)

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -22,7 +22,6 @@ import logging
 import uuid
 
 import anyio
-
 from app.database import SessionLocal
 from app.models import Memory, MemoryAccessLog, MemoryState, MemoryStatusHistory
 from app.utils.db import get_user_and_app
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #6351, opened 2026-07-16
Fix/openmemory mcp search acl deny all

GautamSharma99:
## Description

  This PR fixes a fail-open access-control bug in OpenMemory’s MCP search_memory tool.

  The tool builds a list of memory IDs accessible to the requesting application:

  accessible_memory_ids = [
      memory.id
      for memory in user_memories
      if check_memory_access_permissions(db, memory, app.id)
  ]

  Previously, an empty list was converted to None:

  allowed = set(str(mid) for mid in accessible_memory_ids) if accessible_memory_ids else None

  Search results were then filtered only when allowed was truthy:

  if allowed and (h.id is None or h.id not in allowed):
      continue

  When an application had an explicit deny-all ACL, accessible_memory_ids was empty. That produced allowed = None, disabled the filtering
  condition, and allowed every vector-search hit for the user to be returned.

  This PR changes the logic to always construct a set and always enforce membership:

  allowed = set(str(mid) for mid in accessible_memory_ids)

  if h.id is None or h.id not in allowed:
      continue

  An empty allowed set now correctly rejects every search result.

  ## Security Impact

  Before this change, an application with no permission to access any of a user’s memories could receive those memories through the MCP search
  tool.

  The issue was partly masked by the incompatible vector-store argument fixed in the preceding PR. Once MCP search uses the correct top_k
  argument, the ACL bypass becomes reachable.

  After this change:

  - Explicit deny-all ACLs return no memories.
  - Hits missing an ID are rejected.
  - Only vector hits whose IDs are present in the accessible-memory set are returned.
  - Rejected hits do not create memory access-log entries.

  ## Test Coverage

  Added an end-to-end MCP regression test that:

  1. Creates two user-owned memory records.
  2. Configures permission checks to deny access to both records.
  3. Makes the vector store return both records as search hits.
  4. Invokes search_memory through the MCP Streamable HTTP JSON-RPC flow.
  5. Verifies the response contains an empty results list.
  6. Verifies both permission checks occurred.
  7. Verifies no access-log records were written.

  The regression test failed before the implementation change because both denied memories were returned.

  After the fix:

  24 passed in 1.11s

  Command used:

  Code-quality checks:

  Ruff: passed
  isort: passed

  ## Type of Change

  - [ ] Documentation update

  ## Breaking Changes

  N/A. This restores the intended ACL behavior and does not change the public API.

  ## Test Coverage Checklist

  - [x] I tested manually using the full OpenMemory API test suite
  - [ ] No tests needed

  ## Checklist

  - [x] I have added tests that prove the fix works
  - [x] Documentation is not required because this corrects internal ACL enforcement without changing the public API

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -182,13 +182,13 @@ async def search_memory(query: str) -> str:
                 filters=filters,
             )
 
-            allowed = set(str(mid) for mid in accessible_memory_ids) if accessible_memory_ids else None
+            allowed = set(str(mid) for mid in accessible_memory_ids)
 
             results = []
             for h in hits:
                 # All vector db search functions return OutputData class
                 id, score, payload = h.id, h.score, h.payload
-                if allowed and (h.id is None or h.id not in allowed):
+                if h.id is None or h.id not in allowed:
                     continue
                 
                 results.append({
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -22,7 +22,6 @@ import logging
 import uuid
 
 import anyio
-
 from app.database import SessionLocal
 from app.models import Memory, MemoryAccessLog, MemoryState, MemoryStatusHistory
 from app.utils.db import get_user_and_app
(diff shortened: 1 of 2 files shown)


kartik-mem0:

PR #6352, opened 2026-07-16
Fix/openmemory store mutation consistency

GautamSharma99:
## Description

  This PR prevents OpenMemory’s SQL database and Mem0 vector store from silently diverging during memory updates and deletions.

  Previously:

  - Updating a memory changed only Memory.content in SQL.
  - The vector embedding and vector payload retained the previous content.
  - REST and MCP delete paths caught vector-store exceptions and continued.
  - SQL rows were marked as deleted even when their vectors were still present.
  - Delete operations returned success despite incomplete storage mutations.

  This could cause updated memories to remain searchable using their old meaning and deleted memories to continue appearing in semantic search.

  ## Update Behavior

  The REST update_memory endpoint now updates the Mem0 vector store before committing the new SQL content:

  memory_client.update(str(memory_id), text=request.memory_content)

  The operation follows these semantics:

  1. Load the existing SQL memory and retain its old content.
  2. Update the vector store, including the embedding and vector payload.
  3. If the vector update fails:
      - Return HTTP 502.
      - Leave SQL unchanged.

  4. If the vector update succeeds:
      - Update and commit the SQL content.

  5. If the SQL commit fails:
      - Roll back the SQL transaction.
      - Attempt to restore the vector store using the old content.
      - Return HTTP 500 explaining whether compensation succeeded.

  This prevents a reported-success update from changing only one store.

  ## Delete Behavior

  The REST and MCP delete paths now delete each vector before marking its corresponding SQL row as deleted.

  For every memory:

  1. Delete the memory through memory_client.delete().
  2. Stop immediately if vector deletion fails.
  3. Only after successful vector deletion:
      - Mark the SQL memory as deleted.
      - Record the deletion timestamp.
      - Add the state-history record.
      - Add the MCP access-log record where applicable.
      - Commit that memory’s SQL mutation.

  This ordering ensures a vector deletion failure cannot cause the SQL row to be marked deleted.

  ### Batch Failure Semantics

  Batch deletions are processed and committed one memory at a time.

  If a later deletion fails:

  - Earlier successfully completed memories remain consistently deleted from both stores.
  - The failed memory remains active in SQL.
  - Processing stops immediately.
  - The response reports how many deletions completed before the failure.
  - The operation no longer returns a false success response.

  ### SQL Failure After Vector Deletion

  Vector deletion is irreversible through the current public memory-client API while preserving the original memory ID.

  If vector deletion succeeds but the SQL commit fails:

  - The SQL transaction is rolled back.
  - The API returns an explicit HTTP 500 or MCP error message.
  - The error states that the vector was deleted but SQL could not be updated.

  This edge case cannot be compensated safely without an outbox/reconciliation mechanism, but it is no longer hidden behind a success response.

  ## Additional Corrections

  Deletion history records now use the memory’s actual previous state:

  old_state = memory.state

  Previously, MCP deletion always recorded MemoryState.active, even if the memory was paused or archived.

  The MCP delete implementations also share a private helper so specific and delete-all operations use the same ordering, logging, history, and
  rollback behavior.

  ## Error Responses

  The REST API now returns:

  - 502 Bad Gateway when a vector update or deletion fails.
  - 500 Internal Server Error when SQL persistence fails after an external store mutation.
  - 503 Service Unavailable when the memory client cannot be initialized.

  MCP delete tools now return an error string containing:

  - The failed operation.
  - The affected memory.
  - The underlying vector-store failure.
  - The number of memories successfully deleted before a batch failure.

  They no longer return Successfully deleted... after a failed vector operation.

  ## Test Coverage

  Added six regression cases covering the affected behavior.

  ### REST Update Tests

  - Verifies successful updates call memory_client.update() with the new content.
  - Verifies SQL is committed only after the vector update succeeds.
  - Verifies vector-update failure returns 502 and leaves SQL unchanged.
  - Verifies SQL-commit failure rolls SQL back and restores the previous vector content.

  ### REST Delete Test

  - Verifies vector-deletion failure returns 502.
  - Verifies SQL state mutation is not attempted after the failure.
  - Verifies the memory remains active.

  ### MCP Delete Tests

  The test invokes both tools through the real MCP Streamable HTTP JSON-RPC flow:

  - delete_memories
  - delete_all_memories

  For both tools, it verifies:

  - Vector-store failure produces an error response.
  - The error includes the original failure.
  - The SQL memory remains active.
  - No deletion timestamp is written.
  - No history or access-log rows are added.
  - No SQL commit occurs.

  ## Verification

  Full OpenMemory API test suite:

  Result:

  30 passed in 1.02s

  One existing Pydantic v1-style validator deprecation warning remains unrelated to this change.

  Code-quality checks:

  Ruff: passed
  isort: passed

  ## Type of Change

  - [ ] Documentation update

  ## Breaking Changes

  N/A.

  The public request and response schemas remain unchanged. Failed mutations now return errors instead of incorrectly reporting success.

  ## Test Coverage Checklist

  - [x] I added MCP tool-call regression coverage
  - [x] I tested manually using the complete OpenMemory API test suite
  - [ ] No tests needed

  ## Checklist

  - [x] I have performed a self-review of the code
  - [x] I have added tests that prove the fix works
  - [x] Ruff and isort pass
  - [x] Documentation is not required because the public API schema is unchanged

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -292,6 +292,42 @@ async def list_memories() -> str:
         return f"Error getting memories: {e}"
 
 
+def _delete_memory_from_stores(db, memory_client, memory, user_id, app_id, access_type, operation):
+    memory_id = memory.id
+    try:
+        memory_client.delete(str(memory_id))
+    except Exception as delete_error:
+        raise RuntimeError(f"Failed to delete memory {memory_id} from vector store: {delete_error}") from delete_error
+
+    old_state = memory.state
+    memory.state = MemoryState.deleted
+    memory.deleted_at = datetime.datetime.now(datetime.UTC)
+    db.add(
+        MemoryStatusHistory(
+            memory_id=memory_id,
+            changed_by=user_id,
+            old_state=old_state,
+            new_state=MemoryState.deleted,
+        )
+    )
+    db.add(
+        MemoryAccessLog(
+            memory_id=memory_id,
+            app_id=app_id,
+            access_type=access_type,
+            metadata_={"operation": operation},
+        )
+    )
+
+    try:
+        db.commit()
+    except Exception as db_error:
+        db.rollback()
+        raise RuntimeError(
+            f"Memory {memory_id} was deleted from vector store but its SQL state could not be updated"
+        ) from db_error
+
+
 @mcp.tool(description="Delete specific memories by their IDs")
 async def delete_memories(memory_ids: list[str]) -> str:
     uid = user_id_var.get(None)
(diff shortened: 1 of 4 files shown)


kartik-mem0:

PR #6355, opened 2026-07-16
Fix/openmemory offload blocking routes

GautamSharma99:
## Description

  This PR prevents OpenMemory’s synchronous database, LLM, embedding, and vector-store operations from blocking the application’s asyncio event
  loop.

  Several FastAPI endpoints and FastMCP tools were declared with async def but performed synchronous work directly, including:

  - SQLAlchemy queries and commits using synchronous Session
  - Mem0 SDK calls
  - Embedding generation
  - Vector-store searches and mutations
  - LLM inference
  - Potentially slow Ollama and other network requests

  Because this work ran directly inside async handlers, one slow memory request could stall unrelated HTTP and MCP requests handled by the same
  event loop.

  ## Problem

  FastAPI treats synchronous and asynchronous endpoints differently.

  A normal synchronous endpoint:

  def create_memory(...):
      ...

  is automatically executed in FastAPI’s worker thread pool.

  An asynchronous endpoint:

  async def create_memory(...):
      memory_client.add(...)

  runs on the event-loop thread. FastAPI assumes blocking operations inside it have already been offloaded.

  The affected endpoints used synchronous SQLAlchemy and Mem0 clients from inside async def, so a slow embedding, LLM, vector-store, or database
  operation blocked the event loop.

  FastMCP has a related constraint: its current runtime directly invokes synchronous tool functions instead of automatically moving them into a
  worker thread. Therefore, changing MCP tools from async def to def would still block the event loop.

  ## Solution

  ### FastAPI memory endpoints

  All endpoints in the synchronous SQLAlchemy memory router were converted from async def to normal def.

  This includes:

  - list_memories
  - get_categories
  - create_memory
  - get_memory
  - delete_memories
  - archive_memories
  - pause_memories
  - get_memory_access_log
  - update_memory
  - filter_memories
  - get_related_memories

  FastAPI now executes these handlers in its managed thread pool.

  This keeps the existing synchronous SQLAlchemy session and Mem0 client model without introducing a partial AsyncSession migration.

  ### MCP memory tools

  MCP tools remain asynchronous so their schemas and FastMCP registration behavior remain unchanged.

  Each tool is now a lightweight async wrapper:

  async def search_memory(query: str) -> str:
      return await anyio.to_thread.run_sync(_search_memory, query)

  The complete synchronous operation runs inside the worker thread, including:

  - Reading MCP request context
  - Initializing the memory client
  - Creating and closing the SQLAlchemy session
  - Permission checks
  - Embedding generation
  - Vector-store operations
  - SQL history and access-log commits

  The following MCP tools are covered:

  - add_memories
  - search_memory
  - list_memories
  - delete_memories
  - delete_all_memories

  ## Why the Entire Operation Is Offloaded

  Only offloading individual network calls would still leave synchronous SQLAlchemy operations on the event loop and could move a database
  session between execution contexts.

  Instead, each MCP tool’s complete synchronous body runs in one worker thread. The database session is created, used, committed, and closed
  within that worker execution.

  AnyIO also propagates contextvars into worker threads. Existing MCP regression tests confirm that user_id and client_name remain available and
  that user-scoped filters and ACL checks still receive the correct request context.

  ## Why Not Convert MCP Tools to def

  The installed FastMCP runtime distinguishes sync and async tools but directly calls synchronous tools:

  if fn_is_async:
      return await fn(**arguments)
  else:
      return fn(**arguments)

  It does not automatically offload the synchronous branch.

  Keeping async MCP wrappers and explicitly using anyio.to_thread.run_sync ensures the blocking implementation does not execute on the event-
  loop thread.

  ## Why Not Migrate to AsyncSession

  A complete asynchronous migration would require coordinated changes across:

  - SQLAlchemy engine/session configuration
  - Database dependencies
  - Query syntax
  - Transaction handling
  - Memory and ACL helpers
  - Tests and deployment configuration

  That would substantially increase the scope and risk of this bug fix.

  Thread-pool execution is the established FastAPI approach for synchronous database and SDK integrations and fixes the responsiveness issue
  without changing persistence behavior.

  ## Concurrency Behavior

  Before this change, a slow memory call delayed unrelated requests until the blocking operation completed.

  For example:

  MCP search starts
  └── synchronous embedding call blocks event loop
      └── unrelated HTTP probe cannot run

  After this change:

  MCP search starts
  └── embedding/search work runs in worker thread

  Event loop remains available
  └── unrelated HTTP probe responds immediately

  The same behavior now applies to REST memory creation and all other memory-router endpoints.

  ## Public API Impact

  There are no request-schema, response-schema, route, or MCP tool-schema changes.

  The internal endpoint implementation changes from coroutine functions to synchronous FastAPI handlers, but external clients continue using the
  same HTTP methods and paths.

  MCP tool names, arguments, results, and error behavior remain unchanged.

  ## Test Coverage

  Added concurrent responsiveness tests that deliberately block memory operations using thread synchronization.

  ### REST concurrency test

  The REST regression test:

  1. Creates a FastAPI application with the real memory router.
  2. Mocks memory_client.add() with a blocking synchronous function.
  3. Starts a memory-creation request.
  4. Waits until the blocking SDK operation has started.
  5. Sends an unrelated async probe request before releasing the memory operation.
  6. Verifies the probe responds while memory creation is still blocked.
  7. Verifies the original create request completes successfully after release.

  Before the fix, the probe was delayed until the blocking memory call completed.

  ### MCP add concurrency test

  The MCP test invokes add_memories through the real Streamable HTTP JSON-RPC flow while memory_client.add() is blocked.

  It verifies an unrelated HTTP request remains responsive during the operation.

  ### MCP search concurrency test

  The MCP search test blocks synchronous embedding generation and verifies that:

  - The event loop remains responsive.
  - An unrelated request completes before embedding is released.
  - The MCP tool call subsequently completes successfully.

  ### Existing regression coverage

  The full suite also confirms that offloading does not break:

  - MCP tool registration and schemas
  - Request context isolation
  - SDK argument compatibility
  - Search ACL enforcement
  - Update/delete consistency behavior
  - Vector failure handling
  - MCP route registration
  - Streamable HTTP protocol handling

  ## Verification

  Full OpenMemory API test suite:

  Result:

  33 passed in 1.07s

  One existing Pydantic v1-style validator deprecation warning remains unrelated to this change.

  Code-quality checks:

  Ruff: passed
  isort: passed
  pre-commit hooks: passed

  ## Type of Change

  - [ ] New feature
  - [ ] Breaking change
  - [ ] Refactor
  - [ ] Documentation update

  ## Breaking Changes

  N/A.

  HTTP routes, request/response formats, MCP tool schemas, and persistence behavior remain unchanged.

  ## Test Coverage Checklist

  - [x] Added concurrent REST responsiveness coverage
  - [x] Added concurrent MCP add coverage
  - [x] Added concurrent MCP search coverage
  - [x] Ran the complete OpenMemory API test suite
  - [ ] No tests needed

  ## Checklist

  - [x] My code follows the project’s style guidelines
  - [x] I performed a self-review
  - [x] I added tests that reproduce the event-loop blocking behavior
  - [x] Ruff and isort pass
  - [x] No public API documentation changes are required

### DIFF
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -62,6 +62,10 @@ sse = SseServerTransport("/mcp/messages/")
 
 @mcp.tool(description="Add a new memory. This method is called everytime the user informs anything about themselves, their preferences, or anything that has any relevant information which can be useful in the future conversation. This can also be called when the user asks you to remember something. Set infer to False to store the memory verbatim without LLM fact extraction.")
 async def add_memories(text: str, infer: bool = True) -> str:
+    return await anyio.to_thread.run_sync(_add_memories, text, infer)
+
+
+def _add_memories(text: str, infer: bool = True) -> str:
     uid = user_id_var.get(None)
     client_name = client_name_var.get(None)
 
@@ -147,6 +151,10 @@ async def add_memories(text: str, infer: bool = True) -> str:
 
 @mcp.tool(description="Search through stored memories. This method is called EVERYTIME the user asks anything.")
 async def search_memory(query: str) -> str:
+    return await anyio.to_thread.run_sync(_search_memory, query)
+
+
+def _search_memory(query: str) -> str:
     uid = user_id_var.get(None)
     client_name = client_name_var.get(None)
     if not uid:
@@ -225,6 +233,10 @@ async def search_memory(query: str) -> str:
 
 @mcp.tool(description="List all memories in the user's memory")
 async def list_memories() -> str:
+    return await anyio.to_thread.run_sync(_list_memories)
+
+
+def _list_memories() -> str:
     uid = user_id_var.get(None)
     client_name = client_name_var.get(None)
     if not uid:
@@ -330,6 +342,10 @@ def _delete_memory_from_stores(db, memory_client, memory, user_id, app_id, acces
 
 @mcp.tool(description="Delete specific memories by their IDs")
 async def delete_memories(memory_ids: list[str]) -> str:
+    return await anyio.to_thread.run_sync(_delete_memories, memory_ids)
+
+
+def _delete_memories(memory_ids: list[str]) -> str:
     uid = user_id_var.get(None)
     client_name = client_name_var.get(None)
     if not uid:
@@ -394,6 +410,10 @@ async def delete_memories(memory_ids: list[str]) -> str:
 
 @mcp.tool(description="Delete all memories in the user's memory")
 async def delete_all_memories() -> str:
+    return await anyio.to_thread.run_sync(_delete_all_memories)
+
+
+def _delete_all_memories() -> str:
     uid = user_id_var.get(None)
     client_name = client_name_var.get(None)
     if not uid:
--- a/openmemory/api/app/mcp_server.py
+++ b/openmemory/api/app/mcp_server.py
@@ -292,6 +292,42 @@ async def list_memories() -> str:
         return f"Error getting memories: {e}"
 
 
+def _delete_memory_from_stores(db, memory_client, memory, user_id, app_id, access_type, operation):
+    memory_id = memory.id
+    try:
+        memory_client.delete(str(memory_id))
+    except Exception as delete_error:
+        raise RuntimeError(f"Failed to delete memory {memory_id} from vector store: {delete_error}") from delete_error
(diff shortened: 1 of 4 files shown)


kartik-mem0:

PR #6365, opened 2026-07-17
fix(qdrant): support offline BM25 via bm25_specific_model_path

### LINKED ISSUE #6361: Qdrant bm25 fastembed support "specific_model_path", make mem0 work without network
### Component

Core / Python SDK

### Use Case

current for qdrant user, bm25 use SparseTextEmbedding(model_name="Qdrant/bm25") as encoder, this will download model file via network
maybe we should support "specific_model_path" to let fastembed use local model file

### Proposed Solution

specific_model_path will let fastembed use local model file

Bartok9:
## Summary

- Forward optional `bm25_specific_model_path` / `bm25_model_name` into Qdrant BM25 fastembed loading.
- When set, passes `cache_dir` to `SparseTextEmbedding` so hybrid BM25 works offline from a pre-downloaded model tree.

## Motivation

Closes #6361.

Hybrid search always did `SparseTextEmbedding(model_name="Qdrant/bm25")`, which downloads model weights on first use. Air-gapped / no-network hosts could not enable BM25 even with a local cache already populated. This exposes the path as config and stores it on the vector store constructor (consistent with other Qdrant options).

## Verification

- `pytest tests/vector_stores/test_qdrant.py tests/vector_stores/test_qdrant_config.py` — 101 passed
- Did NOT change: dense-only create_col path when bm25 slot missing; default model name remains `Qdrant/bm25` when unset

### DIFF
--- a/mem0/configs/vector_stores/qdrant.py
+++ b/mem0/configs/vector_stores/qdrant.py
@@ -21,6 +21,17 @@ class QdrantConfig(BaseModel):
         description="Whether to force HTTPS on or off. Explicit schemes in url take precedence.",
     )
     on_disk: Optional[bool] = Field(False,description="Enables persistent storage. Vectors are kept on disk (True) or in memory (False). Does not delete the local database path.")
+    bm25_model_name: Optional[str] = Field(
+        "Qdrant/bm25",
+        description="fastembed SparseTextEmbedding model name for BM25 hybrid search",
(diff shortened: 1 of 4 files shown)


motocoder-cn:
@kartik-mem0  hi, could i ask why this pr is closed? specific_model_path really make sense

PR #6369, opened 2026-07-17
fix(llms/aws_bedrock): iterate Converse content blocks for Anthropic text

### LINKED ISSUE #6368: bug(llms/aws_bedrock): Anthropic Converse parsing crashes when a reasoningContent block precedes the text block
The Anthropic branch of `_generate_standard` in `mem0/llms/aws_bedrock.py` parses the Converse API response by indexing the first content block:

```python
if hasattr(response, 'output') and hasattr(response.output, 'message'):
    return response.output.message.content[0].text
elif 'output' in response and 'message' in response['output']:
    return response['output']['message']['content'][0]['text']
```

(`mem0/llms/aws_bedrock.py:576-580` on current main)

That assumes `content[0]` is always the text block. For Claude models with reasoning enabled, the Converse response puts a `reasoningContent` block before the `text` block, so `['content'][0]['text']` raises `KeyError: 'text'`. An empty content array (some stop conditions) raises `IndexError`. Either way `generate_response` swallows it and re-raises a generic `RuntimeError("Failed to generate response")`, so `add()` and `search()` just die with no useful signal.

The MiniMax branch a few lines below in the same function already handles this exact case, with a comment explaining it (M2.5 is a reasoning model, so it iterates to find the first block containing a `"text"` key):

```python
for block in response["output"]["message"]["content"]:
    if "text" in block:
        return block["text"]
return ""
```

The Anthropic branch just never got the same treatment. Fix is to iterate the content blocks the same way in both response shapes.

Repro: configure the AWS Bedrock LLM provider with a thinking-enabled Claude model, call `Memory.add()`, and the extraction call crashes as soon as the model emits reasoning content.

took a swing at the fix as well (#6369), same block-iteration approach the MiniMax branch uses. Claude Code helped me put it together and I ran the full test suite locally. new to this codebase though, so if there is a better way to handle it I am all ears :)

ayaangazali:
## Linked Issue

Closes #6368

## Description

The Anthropic branch of `_generate_standard` parsed the Converse API response with `content[0]`, which assumes the first content block is the text block. Reasoning-enabled Claude models emit a `reasoningContent` block first (KeyError), and some stop conditions produce an empty content array (IndexError); both surface as an unhelpful generic `RuntimeError` from `generate_response`. The MiniMax branch in the same function already iterates the content blocks for exactly this reason, with a comment explaining it. This applies the same treatment to the Anthropic branch, in both response shapes, returning an empty string when no text block exists (matching the MiniMax and Gemini provider contracts).

Four new tests cover the reasoning-block-first case, the empty content array, the plain text case (unchanged behavior), and the object-style response shape. The first two fail on main and pass with the fix.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

`make lint` clean, `make test-py-3.11` green (1679 passed; the two Redis e2e failures are pre-existing on main, they only trigger because a local Redis is reachable on my machine and are skipped in CI). Verified the new reasoning/empty-content tests fail without the parser change.

## Checklist

---

small disclosure like on my last PR: freshman here doing my best for the greater good :) built this with Claude Code doing the heavy lifting while I sanity-checked the diff and ran the gates locally. if I got something wrong I'd honestly rather hear it and learn than have it slide.

### DIFF
--- a/mem0/llms/aws_bedrock.py
+++ b/mem0/llms/aws_bedrock.py
@@ -573,11 +573,20 @@ class AWSBedrockLLM(LLMBase):
             # Use converse API for Anthropic models
             response = self.client.converse(**converse_params)
 
-            # Parse Converse API response
+            # Parse Converse API response. Claude reasoning models can emit a
+            # `reasoningContent` block before the `text` block, so iterate to
+            # find the first block that carries text instead of indexing
+            # content[0] (same approach as the MiniMax branch below).
             if hasattr(response, 'output') and hasattr(response.output, 'message'):
-                return response.output.message.content[0].text
+                for block in response.output.message.content:
+                    if hasattr(block, 'text'):
+                        return block.text
+                return ""
             elif 'output' in response and 'message' in response['output']:
-                return response['output']['message']['content'][0]['text']
+                for block in response['output']['message']['content']:
+                    if 'text' in block:
+                        return block['text']
+                return ""
             else:
                 return str(response)
 
--- a/tests/llms/test_aws_bedrock.py
+++ b/tests/llms/test_aws_bedrock.py
@@ -452,3 +452,51 @@ class TestParseResponseLegacy:
         response = {"body": body}
         result = llm._parse_response(response, tools=None)
         assert result == "hello from ai21"
+
+
+class TestAnthropicConverseContentParsing:
+    """The Anthropic Converse branch must not assume content[0] is the text
+    block: Claude reasoning models emit a reasoningContent block before the
+    text block, and some stop conditions produce an empty content array. The
+    parser iterates for the first block carrying text, like the MiniMax branch.
+    """
+
+    def test_text_after_reasoning_content_block(self, mock_boto3):
+        mock_boto3.converse.return_value = {
+            "output": {
+                "message": {
+                    "content": [
+                        {"reasoningContent": {"reasoningText": {"text": "step by step..."}}},
+                        {"text": "final answer"},
+                    ]
+                }
+            }
+        }
+        llm = _make_llm("anthropic.claude-3-5-sonnet-20240620-v1:0", mock_boto3)
+
+        assert llm.generate_response(MESSAGES) == "final answer"
+
+    def test_empty_content_returns_empty_string(self, mock_boto3):
+        mock_boto3.converse.return_value = {"output": {"message": {"content": []}}}
+        llm = _make_llm("anthropic.claude-3-5-sonnet-20240620-v1:0", mock_boto3)
+
+        assert llm.generate_response(MESSAGES) == ""
+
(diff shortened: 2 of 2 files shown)


mem0/llms/aws_bedrock.py
```diff
@@ -573,11 +573,20 @@ def _generate_standard(self, messages: List[Dict[str, str]], stream: bool = Fals
             # Use converse API for Anthropic models
             response = self.client.converse(**converse_params)
 
-            # Parse Converse API response
+            # Parse Converse API response. Claude reasoning models can emit a
+            # `reasoningContent` block before the `text` block, so iterate to
+            # find the first block that carries text instead of indexing
+            # content[0] (same approach as the MiniMax branch below).
             if hasattr(response, 'output') and hasattr(response.output, 'message'):
```
kartik-mem0, line 580: boto3 returns plain dicts from `converse`, so this attr-style branch never runs in production and `test_object_style_response_iterates_blocks` is the only thing that reaches it. Deleting the branch and that test would leave one parsing path instead of two. Optional, it does not block this fix.

kartik-mem0 (commented):

`_generate_standard` indexed `content[0]['text']`, so a reasoning-enabled Claude model that emits a `reasoningContent` block first raised `KeyError` and surfaced to the user as `RuntimeError: Failed to generate response`. This iterates for the first text block, the same shape the MiniMax branch below already uses.

**Before you merge:** nothing, self-contained, and it still merges clean onto current `main` (56 tests green).
**Follow-up:** the attr-style branch just above is unreachable with boto3, noted inline as optional.
**Not verified:** live Bedrock, the response blocks are mocked.

Risk: low, one branch in one provider, no public API change · Your time: ~3 min

- **Problem is real:** reproduces on `main` at `mem0/llms/aws_bedrock.py:586`, Converse puts `reasoningContent` first for thinking-enabled Claude so `content[0]` has no `text` key (an empty `content` array hits the same line with `IndexError`)
- **Tests:** `pytest tests/llms/test_aws_bedrock.py -q` -> 45 passed
- **Red-green:** reverse-applied the impl hunk only, 3 of the 4 new tests failed (`KeyError: 'text'`, `IndexError`, `AttributeError`), restored -> 45 passed. The 4th is the plain-text non-regression case and correctly passes either way. ✓
- **CI:** all real checks green (Python SDK 3.10/3.11/3.12, CI Gate). Vercel is red, it is red on every fork PR and does not block.
- **Security:** cleared, no untrusted input and no secrets, the diff only reads keys off a boto3 response dict.
- **Bloat (ponytail):** net -6 lines possible, drop the unreachable `hasattr` branch and its `SimpleNamespace` test.
- **Linked issue:** #6368

</details>

PR #6374, opened 2026-07-18
fix(mem0): add test for issue #6203

### LINKED ISSUE #6203: bug: Anthropic generateResponse silently ignores response_format (present in both SDKs — TS #6205, Python #5820)

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6203.

## Why
Fixes #6203

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6203.py
+++ b/tests/test_issue_6203.py
@@ -1,0 +1,40 @@
+"""Regression test for issue #6203.
+
+Verifies that the OSS TypeScript Anthropic provider does not silently drop the
+responseFormat argument when callers request structured JSON output.
+"""
+
+import re
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+sys.modules.setdefault("posthog", MagicMock())
+sys.modules.setdefault("qdrant_client", MagicMock())
+
+with patch("importlib.metadata.version", return_value="0.0.0"):
+    from mem0 import Memory
+
+
+def test_issue_6203():
+    assert Memory is not None
+
+    provider_path = Path(__file__).resolve().parents[1] / "mem0-ts/src/oss/src/llms/anthropic.ts"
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

**Finding:** `tests/test_issue_6203.py` is a source-text grep, not a behavioural test. It reads `mem0-ts/src/oss/src/llms/anthropic.ts` and asserts that literal substrings appear in it:

```python
assert "json_object" in body_after_signature
assert "extractJson" in body_after_signature
assert "output_config" in body_after_signature
```

This pins the wording of a future fix rather than its behaviour: a correct fix using different identifiers would fail, and a stub containing those strings would pass. Nothing is executed.

---

Closing this one. To be clear, this is about the test file, not about issue #6203 — the underlying report is legitimate and stays as it is.

PR #6375, opened 2026-07-18
Fix memory API paging and missing-memory handling

tpcoleary:
## Summary
- fix filtered `/memories` pagination so later pages do not repeat the first page
- return `404` instead of `200 null` when `GET /memories/{id}` misses
- ignore secret file variants like `.env.*` and `auth.json`

## Details
The filtered `GET /memories` path previously accepted paging parameters but always returned the same slice because `page` was never wired into the call to `get_all()`. This change implements bounded over-fetch-and-slice pagination for filtered memory reads.

`GET /memories/{id}` now raises a `404` when the underlying memory lookup returns `None`, instead of serializing that miss as a successful response with a null body.

## Validation
- `python3 -m py_compile server/main.py`
- reviewed diff against `origin/main` to keep this PR limited to generic server fixes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ __pycache__/
 # Self-hosted server local runtime state
 server/history/
 server/.env
+server/.env.*
+.env.*
+**/auth.json
 
 # C extensions
 *.so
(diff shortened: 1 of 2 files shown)


tpcoleary:
Signed the CLA. The changes in this PR are limited to generic server behavior: filtered /memories pagination, proper 404 handling for missing memory IDs, support for a configurable OpenAI-compatible base URL, and .gitignore hardening for local secret/runtime files. Happy to adjust anything if you'd prefer a different approach.

PR #6376, opened 2026-07-18
fix(mem0): add test for issue #6181

### LINKED ISSUE #6181: [windows] Codex lifecycle hooks still fail on v0.2.12 — POSIX commands exit 1 on every event (regression of #5243)

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6181.

## Why
Fixes #6181

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6181.py
+++ b/tests/test_issue_6181.py
@@ -1,0 +1,54 @@
+"""Regression test for Issue #6181.
+
+Codex lifecycle hook commands are executed by the native shell on Windows.
+The hook template must therefore avoid POSIX-only env-prefix commands and
+must not rely on .sh files being directly executable.
+"""
+
+import importlib.util
+import json
+import os
+import re
+from pathlib import PurePosixPath
+
+
+def test_issue_6181():
+    plugin_root = os.path.join(
+        os.path.dirname(__file__),
+        "..",
+        "integrations",
+        "mem0-plugin",
+    )
+    installer_path = os.path.join(plugin_root, "scripts", "install_codex_hooks.py")
+    spec = importlib.util.spec_from_file_location("install_codex_hooks", installer_path)
+    installer = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(installer)
+
+    assert installer.OWNER_MARKER == "mem0-plugin"
+
+    hooks_path = os.path.join(plugin_root, "hooks", "codex-hooks.json")
+    with open(hooks_path) as f:
+        config = json.load(f)
+
+    commands = [
+        h["command"] for entries in config["hooks"].values() for entry in entries for h in entry.get("hooks", [])
+    ]
+    assert commands, "expected at least one codex hook command"
+
+    env_prefix = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
+    for cmd in commands:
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6377, opened 2026-07-18
fix(ts-oss): strip identity scope from add() metadata

### LINKED ISSUE #6371: TypeScript OSS SDK: add() lets metadata set identity scope (agent_id/run_id/actor_id) on creation
### Summary

The TypeScript OSS SDK's `add()` has the same identity-field bug class as `updateMemory()` (tracked in #6367 / fixed by #6343), but on the **creation** path: caller-supplied `metadata` can set a memory's identity scope (`agent_id` / `run_id` / `actor_id`) when the matching top-level entity param isn't also passed.

This is a distinct method from `updateMemory()`, so it is out of scope for #6367 (which is scoped to `update()`) — filing separately as flagged during review of #6343.

### Where

`mem0-ts/src/oss/src/memory/index.ts` — the creation path re-pins identity fields onto the stored payload only *conditionally* on the top-level param being present, e.g.:

```ts
if (filters.agent_id) memPayload.agent_id = filters.agent_id;
if (filters.run_id) memPayload.run_id = filters.run_id;
```

(see roughly `:742-744`, `:986-999`, `:1890-1907`)

Because the re-pin only fires when the top-level `agentId`/`runId` param is passed, caller-supplied `metadata.agent_id` / `metadata.run_id` survives untouched into the newly created memory whenever that param is omitted. `actor_id` has **no** re-pin path in `add()` at all, so `metadata.actor_id` always survives.

### Impact

- `add(text, { userId: "u1", metadata: { agent_id: "other", actor_id: "x" } })` stores a memory whose scope includes `agent_id: "other"` / `actor_id: "x"` — scopes the caller didn't intend to set through the documented top-level params.
- Same tenant-isolation class as #6277 / #6342, on the creation path rather than update.

### Suggested fix

Mirror the `update()` fix (#6343): strip the identity keys from a copy of the incoming `metadata` before it is merged into the created payload, so identity scope is set only via the documented top-level `userId`/`agentId`/`runId` params, never via freeform metadata. A shared `IDENTITY_KEYS` constant already exists in this file after #6343.

### Scope

TypeScript OSS SDK `add()` only. `update()` is handled by #6343 / #6367.

rodboev:
## Linked Issue

Closes #6371

## Description

In the TypeScript OSS SDK, `Memory.add()` should only take identity scope from the typed `userId` / `agentId` / `runId` params. Before this change, callers could still smuggle scope through freeform `metadata`, and the default memory store could promote `userId` / `agentId` / `runId` back to canonical snake_case keys on read.

This change strips both the canonical identity keys and their camelCase aliases from `metadata` before typed or filter scope is pinned into the payload. Ordinary metadata still passes through, and typed scope remains authoritative.

## Root cause

`Memory.add()` passed caller-supplied metadata through to the stored payload, so freeform identity keys could establish scope. The default memory store's `normalizePayload()` also promotes `userId` / `agentId` / `runId` to `user_id` / `agent_id` / `run_id` on read, making camelCase aliases an injection path. The add path now removes all recognized identity spellings before typed or filter scope is pinned.

## Scope

- changes the TypeScript OSS `add()` path only
- keeps `update()` separate
- does not change the public API shape

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] I tested manually
- [ ] No tests needed

Observed verification:

- [x] `cd mem0-ts && pnpm run test:unit -- --runTestsByPath src/oss/tests/memory.add.test.ts --coverage=false`
- [x] `cd mem0-ts && npx prettier --check .\src\oss\src\memory\index.ts .\src\oss\tests\memory.add.test.ts`
- [ ] `cd mem0-ts && pnpm install --frozen-lockfile && pnpm run build && pnpm run test:unit`

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -100,6 +100,12 @@ const ENTITY_PARAMS = [
   "agentId",
   "runId",
 ];
+const IDENTITY_METADATA_KEYS = [
+  "user_id",
+  "agent_id",
+  "run_id",
+  "actor_id",
+] as const;
 
 // Identity keys stripped from update() metadata: ENTITY_PARAMS covers user_id/agent_id/run_id
 // in both casings (the default store promotes camelCase on read); actor_id has no camelCase alias.
@@ -736,7 +742,9 @@ export class Memory {
       has_filters: !!config.filters,
       infer: config.infer,
     });
-    const { metadata = {}, filters = {}, infer = true } = config;
+    const { filters = {}, infer = true } = config;
+    const metadata = { ...config.metadata };
+    for (const key of IDENTITY_METADATA_KEYS) delete metadata[key];
 
     // Validate and trim entity IDs
     const userId = validateAndTrimEntityId(config.userId, "userId");
@@ -747,6 +755,9 @@ export class Memory {
     if (userId) filters.user_id = metadata.user_id = userId;
     if (agentId) filters.agent_id = metadata.agent_id = agentId;
     if (runId) filters.run_id = metadata.run_id = runId;
+    if (filters.user_id) metadata.user_id = filters.user_id;
+    if (filters.agent_id) metadata.agent_id = filters.agent_id;
+    if (filters.run_id) metadata.run_id = filters.run_id;
 
     // Normalize expiration date into the stored metadata (round-trips via get()).
     if (config.expirationDate != null) {
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -100,9 +100,19 @@ const ENTITY_PARAMS = [
   "agentId",
   "runId",
 ];
-// Identity keys stripped from update() metadata: ENTITY_PARAMS covers user_id/agent_id/run_id
-// in both casings (the default store promotes camelCase on read); actor_id has no camelCase alias.
+// Identity keys stripped from caller metadata in add() and update(): ENTITY_PARAMS covers
+// user_id/agent_id/run_id in both casings (the default store promotes camelCase on read);
+// actor_id has no camelCase alias.
 const IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"];
+
+// Caller metadata must not overwrite or inject an identity scope (#6342 / #6367 / #6371).
+function stripIdentityKeys(
+  metadata: Record<string, any> = {},
+): Record<string, any> {
+  return Object.fromEntries(
+    Object.entries(metadata).filter(([key]) => !IDENTITY_KEYS.includes(key)),
+  );
+}
 
 /**
  * Validates that no top-level entity parameters are passed in config.
@@ -736,8 +746,7 @@ export class Memory {
       infer: config.infer,
     });
     const { filters = {}, infer = true } = config;
-    const metadata = { ...config.metadata };
-    for (const key of IDENTITY_KEYS) delete metadata[key];
+    const metadata = stripIdentityKeys(config.metadata);
 
     // Validate and trim entity IDs
     const userId = validateAndTrimEntityId(config.userId, "userId");
@@ -1950,10 +1959,7 @@ export class Memory {
       existingEmbeddings[newData] ||
       (await this.embedder.embed(newData, "update"));
 
-    // Caller metadata must not overwrite or inject an identity scope (#6342 / #6367).
-    const sanitizedMetadata = Object.fromEntries(
-      Object.entries(metadata).filter(([k]) => !IDENTITY_KEYS.includes(k)),
-    );
+    const sanitizedMetadata = stripIdentityKeys(metadata);
 
     const newMetadata = {
       ...existingMemory.payload,
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -100,12 +100,7 @@ const ENTITY_PARAMS = [
   "agentId",
   "runId",
 ];
-const IDENTITY_METADATA_KEYS = [
-  "user_id",
-  "agent_id",
-  "run_id",
-  "actor_id",
-] as const;
+const IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"] as const;
 
 // Identity keys stripped from update() metadata: ENTITY_PARAMS covers user_id/agent_id/run_id
 // in both casings (the default store promotes camelCase on read); actor_id has no camelCase alias.
@@ -744,7 +739,7 @@ export class Memory {
     });
     const { filters = {}, infer = true } = config;
     const metadata = { ...config.metadata };
-    for (const key of IDENTITY_METADATA_KEYS) delete metadata[key];
+    for (const key of IDENTITY_KEYS) delete metadata[key];
 
     // Validate and trim entity IDs
     const userId = validateAndTrimEntityId(config.userId, "userId");
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -100,8 +100,6 @@ const ENTITY_PARAMS = [
   "agentId",
   "runId",
 ];
-const IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"] as const;
-
 // Identity keys stripped from update() metadata: ENTITY_PARAMS covers user_id/agent_id/run_id
 // in both casings (the default store promotes camelCase on read); actor_id has no camelCase alias.
 const IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"];
--- a/mem0-ts/src/oss/tests/memory.add.test.ts
+++ b/mem0-ts/src/oss/tests/memory.add.test.ts
@@ -172,6 +172,102 @@ describe("Memory - add()", () => {
     );
   });
 
+  test("does not allow metadata to set identity scope", async () => {
+    const result: SearchResult = await memory.add("I am a software engineer", {
+      userId: "u1",
+      metadata: {
+        agent_id: "other",
+        run_id: "other-run",
+        actor_id: "x",
+        source: "issue-6371",
+        nested: { preserved: true },
+      },
+    });
+    const stored: MemoryItem | null = await memory.get(result.results[0].id);
+
+    expect(stored).toEqual(
+      expect.objectContaining({
+        user_id: "u1",
+        metadata: expect.objectContaining({
+          source: "issue-6371",
+          nested: { preserved: true },
+        }),
+      }),
+    );
+    expect(stored).not.toHaveProperty("agent_id");
+    expect(stored).not.toHaveProperty("run_id");
+    expect(stored!.metadata).not.toHaveProperty("agent_id");
+    expect(stored!.metadata).not.toHaveProperty("run_id");
+    expect(stored!.metadata).not.toHaveProperty("actor_id");
+  });
+
+  test.each([
+    ["userId", { userId: "u1" }, true, "user_id", "u1"],
+    ["agentId", { agentId: "a1" }, false, "agent_id", "a1"],
+    ["runId", { runId: "r1" }, true, "run_id", "r1"],
+  ] as const)(
+    "preserves typed %s scope while stripping conflicting metadata identities",
+    async (_mode, scope, infer, canonicalKey, canonicalValue) => {
+      const result: SearchResult = await memory.add("scoped content", {
+        ...scope,
+        infer,
+        metadata: {
+          user_id: "metadata-user",
+          agent_id: "metadata-agent",
+          run_id: "metadata-run",
+          actor_id: "metadata-actor",
+          ordinary: "preserved",
+        },
+      });
+      const stored: MemoryItem | null = await memory.get(result.results[0].id);
+
+      expect(stored).toHaveProperty(canonicalKey, canonicalValue);
+      expect(stored!.metadata).toEqual(
+        expect.objectContaining({ ordinary: "preserved" }),
+      );
+      for (const key of ["user_id", "agent_id", "run_id", "actor_id"]) {
+        if (key !== canonicalKey) {
+          expect(stored!.metadata).not.toHaveProperty(key);
+        }
+      }
+    },
+  );
+
+  test.each([
+    [true, "user_id", "filter-user"],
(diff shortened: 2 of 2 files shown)


OfficialAbhinavSingh:
Thanks for picking this up, and for the credit! Two things that might save you a round of review, both from what @kartik-mem0 flagged on the sibling `update()` fix (#6343):

1. **camelCase aliases slip through.** `IDENTITY_METADATA_KEYS` here is snake_case only, but the default `memory` vector store's `normalizePayload()` promotes `userId`/`agentId`/`runId` → `user_id`/`agent_id`/`run_id` on read. So `add("...", { metadata: { userId: "attacker" } })` survives the `delete` and gets promoted back into an identity scope — the same injection this PR is closing, via the alias. Worth a regression test for the camelCase form.
2. **Shared constant.** #6343 adds a module-level `IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"]` (so it covers both casings *and* `actor_id` from a single source). Once that lands, reusing it here instead of a second `IDENTITY_METADATA_KEYS` list avoids the two-lists-that-can-drift problem @kartik-mem0 called out.

Whichever of us merges second can rebase onto the shared constant. Nice to have the creation path covered too.

rodboev:
Thanks, good catch.

1. I widened the add-path sanitizer in `mem0-ts/src/oss/src/memory/index.ts` so it strips `userId` / `agentId` / `runId` alongside `user_id` / `agent_id` / `run_id` / `actor_id` before typed or filter scope is re-pinned. That closes the alias path before the default memory store's `normalizePayload()` can promote it on read.
2. I kept this PR independent of [#6343](https://github.com/mem0ai/mem0/pull/6343). The shared `IDENTITY_KEYS` helper from that PR is not on `main` yet, so this branch derives the local key set from the existing `ENTITY_PARAMS` plus `actor_id`. If `#6343` lands first, whichever PR merges second can collapse onto the shared constant.
3. I extended the focused add-path regressions in `mem0-ts/src/oss/tests/memory.add.test.ts` to cover the camelCase aliases for typed scope and filter-scoped adds, and reran that file locally.

mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -100,6 +100,7 @@ const ENTITY_PARAMS = [
   "agentId",
   "runId",
 ];
+const IDENTITY_KEYS = [...ENTITY_PARAMS, "actor_id"] as const;
```
kartik-mem0: Merging this now (branch is 43 commits behind main) adds a second `const IDENTITY_KEYS` to this file: an already-merged update()-guard PR declared the same name near the same spot. Confirmed via `tsc --noEmit` on GitHub's own merge ref for this PR: two `TS2451: Cannot redeclare block-scoped variable 'IDENTITY_KEYS'` errors.

Rebase onto main and drop this duplicate, reusing the existing constant. Consider consolidating into one shared `stripIdentityKeys()` used by both add() and update().

mem0-ts/src/oss/tests/memory.add.test.ts
```diff
@@ -8,6 +8,118 @@ import type { MemoryConfig, MemoryItem, SearchResult } from "../src/types";
...
+    },
+    RerankerFactory: {
+      create: jest.fn(() => {
+        throw new Error("RerankerFactory is not used in memory.add.test.ts");
+      }),
+    },
+  };
+});
+
+jest.mock("openai", () => {
```
kartik-mem0: The `openai`, `node-fetch`, and `@anthropic-ai/sdk` mocks (lines 84-121) look unused: `../src/utils/factory` is already fully mocked above, so no provider class here is ever constructed. Verified by deleting all three blocks: 19/19 tests still pass.

Drop them, or add a one-line reason if they guard against something not exercised by this suite.

kartik-mem0 (commented):
## Changes requested

Strips `user_id`/`agent_id`/`run_id`/`actor_id` from add() metadata correctly (matches the Python fix, #6656), but merging as-is breaks the build.

**To unblock:**
1. `mem0-ts/src/oss/src/memory/index.ts:103` — duplicate `IDENTITY_KEYS`, rebase onto main (detail inline)

Optional: unused provider mocks in the test file (detail inline)

- **Problem is real:** reproduces on `main` at `mem0-ts/src/oss/src/memory/index.ts:734`, caller's raw `metadata` is destructured straight through with no identity-key strip
- **Tests / red-green:** `memory.add.test.ts` 19/19 pass on this head, 8/19 fail against the PR's own pre-fix base; single chokepoint covers both infer paths, update() has its own separate working guard
- **CI:** green but stale, ran before an unrelated already-merged PR added a same-named `IDENTITY_KEYS`; confirmed the merge result throws `TS2451: Cannot redeclare block-scoped variable 'IDENTITY_KEYS'` via `tsc --noEmit` on GitHub's own merge ref
- **Security:** cleared otherwise, actor_id covered symmetrically with Python, legitimate top-level agentId/userId/runId still wins after the strip
- **Bloat (ponytail):** net: -39 lines possible (drop the dead mocks), plus the duplicate constant
- **Not verified:** live OpenAI/Anthropic providers, mocked blocks only

</details>

Solid fix underneath, just needs a rebase before it can land. Push when ready and I will re-review.

kartik-mem0 (commented):
Duplicate post from a tooling error on my end, please disregard this one. Full review with inline comments is here: https://github.com/mem0ai/mem0/pull/6377#pullrequestreview-4829340017

mem0-ts/src/oss/src/memory/index.ts
rodboev: Thanks, fixed. I rebased the branch onto current main and kept the single `IDENTITY_KEYS` declaration that landed there. The add path now reuses that constant, so the merge no longer creates a duplicate declaration; the existing update guard remains unchanged.

mem0-ts/src/oss/tests/memory.add.test.ts
rodboev: Removed those three package mocks. The factory mock already supplies the embedder, LLM, vector store, and history manager used by this suite, so the provider package mocks weren't exercised. The add coverage remains in place for typed and filter scope, both inference modes, and ordinary metadata.

mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -100,7 +100,6 @@ const ENTITY_PARAMS = [
   "agentId",
   "runId",
 ];
-
 // Identity keys stripped from update() metadata: ENTITY_PARAMS covers user_id/agent_id/run_id
```
kartik-mem0: Minor, optional: this comment now undersells the constant. `IDENTITY_KEYS` is no longer only for `update()` metadata, `add()` strips against the same list at :740. Worth rewording to something like "Identity keys stripped from caller metadata in add() and update()" so the next reader does not assume the add path is unprotected.

mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -736,7 +735,9 @@ export class Memory {
       has_filters: !!config.filters,
       infer: config.infer,
     });
-    const { metadata = {}, filters = {}, infer = true } = config;
+    const { filters = {}, infer = true } = config;
+    const metadata = { ...config.metadata };
+    for (const key of IDENTITY_KEYS) delete metadata[key];
```
kartik-mem0: Minor, optional, not blocking: the same array is now stripped two different ways. Here it is a `delete` loop over a copy; in `update()` at :1955 it is `Object.fromEntries(...)` with a filter. Both are correct and neither is wrong to keep, but a small shared `stripIdentityKeys(metadata)` next to the constant would give the third write path (whenever one shows up) one obvious thing to call, instead of a coin flip between two idioms.

mem0-ts/src/oss/src/memory/index.ts
```diff
@@ -747,6 +748,9 @@ export class Memory {
     if (userId) filters.user_id = metadata.user_id = userId;
     if (agentId) filters.agent_id = metadata.agent_id = agentId;
     if (runId) filters.run_id = metadata.run_id = runId;
+    if (filters.user_id) metadata.user_id = filters.user_id;
```
kartik-mem0, line 764: Context for whoever merges rather than a change request: these three lines are the entire behavioral change in open PRs #6220 and #6177. Having them here (correctly, since the strip above means the `&& !metadata.user_id` guard those PRs use is no longer needed) means merging this makes both of those redundant, and the new `test.each` block covers their `infer: false` cases too. I am not closing anything, just flagging so the overlap does not get merged twice.

mem0-ts/src/oss/tests/memory.add.test.ts
```diff
@@ -8,6 +8,79 @@ import type { MemoryConfig, MemoryItem, SearchResult } from "../src/types";
...
+    },
+    RerankerFactory: {
+      create: jest.fn(() => {
+        throw new Error("RerankerFactory is not used in memory.add.test.ts");
+      }),
+    },
+  };
+});
+
 // Mock Google modules to prevent @google/genai crash in CI
```
kartik-mem0: Minor, optional cleanup: the new file-wide `jest.mock("../src/utils/factory", ...)` above intercepts `EmbedderFactory` / `LLMFactory` before the `Memory` constructor ever reaches a provider module, which makes these per-module mocks (through :132, the two `google` and two `openai` ones) unreachable. I deleted all 49 lines locally and reran the file: still 19/19 passing. Free net -49 lines if you want it, equally fine to leave for a follow-up.

mem0-ts/src/oss/tests/memory.add.test.ts
```diff
@@ -172,6 +245,138 @@ describe("Memory - add()", () => {
...
+    },
+  );
+
+  test.each([
+    [true, "user_id", "filter-user"],
+    [false, "user_id", "filter-user"],
+    [false, "agent_id", "filter-agent"],
+    [false, "run_id", "filter-run"],
+  ] as const)(
+    "preserves %s filters scope after sanitization",
```
kartik-mem0: Minor, optional: `%s` interpolates only the first tuple element (`infer`), so the three `infer: false` rows all report as `preserves false filters scope after sanitization` and a failure does not say which key broke. `"preserves %s filters scope after sanitization (%s)"` picks up the filter key as the second `%s` and makes the output self-describing.

kartik-mem0 (commented):

Closes the TypeScript half of the P0 in #6371. Caller `metadata` could set `user_id` / `agent_id` / `run_id` / `actor_id` on the stored payload, so `add(msg, { userId: owner, metadata: { agent_id: attacker } })` wrote a memory that the attacker's own `getAll({ agentId })` then returned. `update()` was already sanitized in #6278; `add()` was the remaining hole. This strips the identity keys off a copy of the caller metadata before any scope is pinned, at the one chokepoint both `infer` branches pass through, so every backend picks it up rather than just the default store.

Verified this run: reproduced the leak on `main` (3ac9ba4b5) against real Qdrant and real pgvector, both FAIL, then ran the same scenario against this branch, both PASS. Unit red-green: 19/19 at this head, revert the implementation only and 8 of the new identity tests go red, restore and 19/19 again. Real CI checks are green (the `Vercel` failure is the usual fork deploy-permission artifact, not merge-blocking).

Two things worth knowing before you merge:

- **Merging this makes #6220 and #6177 redundant.** Their entire behavioral change is the three filters-into-metadata mirror lines, which are already at `index.ts:751-753` here, and their `infer: false` cases are covered by the new `test.each` block. I read both diffs rather than the titles. I have not touched either PR.
- **The Python half stays open.** `_strip_identity_keys` is wired into the update paths only (`mem0/memory/main.py:2025`, `:3708`), so Python `add()` still has this hole. #6656 and #6662 are the open twins for it.

Five inline notes below, all optional, none blocking. Not verified: live OpenAI / Anthropic providers, those stay mocked on both sides.

**Risk: low · Your time: ~3 min** · `gh pr merge 6377 --squash`

<details>

**Gate 0, the bug is real.** On `main`, `index.ts:743` destructures `metadata` straight from `config` and hands it to `addToVectorStore` (:773), which reaches `createMemory` (:1920) where it is spread into `memoryMetadata` and inserted. Identity keys in that payload are exactly what scoped reads match on. `vector_stores/memory.ts:28` also promotes `userId` to `user_id` on read, so camelCase is a second spelling of the same hole, which is why `ENTITY_PARAMS` covering both casings matters here.

**Unit red-green** (worktree at `bb2a3a80`, `node_modules` reused from the primary checkout, no install):

```
jest src/oss/tests/memory.add.test.ts    -> Tests: 19 passed, 19 total
git checkout main -- src/oss/src/memory/index.ts   # tests untouched
jest src/oss/tests/memory.add.test.ts    -> Tests: 8 failed, 11 passed, 19 total
git checkout origin/pr-6377 -- <impl>
jest src/oss/tests/memory.add.test.ts    -> Tests: 19 passed, 19 total
```

The 8 that go red are the 4 strip tests plus the 4 `test.each` filter rows, so both halves of the diff (the strip at :739-740 and the mirror at :751-753) are covered, in both `infer` modes.

**Live sandbox**, scenario `issue6371_metadata_identity_scope`, lanes `oss-qdrant` and `oss-pgvector`:

```
main @ 3ac9ba4b5   oss-qdrant    FAIL (3.0s)  memory became visible under agent scope ...-attacker via add() metadata
main @ 3ac9ba4b5   oss-pgvector  FAIL (2.4s)  memory became visible under agent scope ...-attacker via add() metadata
pr-6377 @ bb2a3a80 oss-qdrant    PASS (2.4s)
pr-6377 @ bb2a3a80 oss-pgvector  PASS (2.4s)
```

That is the reported symptom on real backends, not just the mechanism in a mock.

**Correctness.** The strip runs before the typed pin (:748-750) and before the filters mirror (:751-753), so when a caller passes both `userId` and `filters.user_id` the typed value wins in both places. Falsy filter values are not mirrored and still hit the existing required-scope throw at :760. Nested metadata is left alone, which is correct: `normalizePayload` only promotes top-level keys. Side benefit, `{ ...config.metadata }` stops `add()` mutating the caller's own object, which `main` did.

**Security edges considered:** `config.metadata` undefined, empty-string filter values, unicode and arbitrary non-identity keys (preserved, asserted by the test), nested identity keys (never promoted, so not an escape).

**Not verified:** live OpenAI / Anthropic providers. Both remain mocked in the unit suite and the sandbox scenario runs `infer: false`, so no real LLM call is exercised on either side. Does not touch the fix's path.

</details>

kartik-mem0 (approved):

Issue #5742, opened 2026-06-22
bug(oss): hybrid search drops keyword-only and entity-rescued memories


rafid001:
### Component

Core / Python SDK + TypeScript OSS

### Description

Hybrid search runs semantic search, keyword/BM25 search, and entity boosting, but Step 7 only builds candidates from semantic top-k results. Keyword hits outside that pool are scored into bm25_scores but never ranked. score_and_rank() also gates on semantic score alone.

### Expected Behavior

Union semantic and keyword candidates (and entity-rescued memories), then rank using all active signals. Threshold should pass when any signal meets the cutoff.

kartik-mem0:
Traced this on main. The candidate pool built in _search_vector_store() only ever comes from semantic search:

mem0/memory/main.py:1618-1629 (Step 7, sync path; the async path mirrors this at line 3247) loops over semantic_results only when building candidates. keyword_search() runs at Step 4 (line 1599) and its hits get normalized into a bm25_scores dict at Step 5, but that dict is only consulted for memory IDs that already made it into candidates via the semantic loop — a memory that scores well on keyword/BM25 but isn't in the semantic top-k never becomes a candidate, so its bm25 score is computed and then discarded. Same mechanism for entity boosts (Step 6, line 1614).

This is compounded by mem0/utils/scoring.py:110-112 in score_and_rank(): the threshold check (if semantic_score < threshold: continue) runs before bm25/entity are combined, on semantic_score alone — actually documented as intentional in the docstring above it (lines 74-75), but it's exactly the drop behavior described here. A candidate that does make it into the pool with a strong keyword or entity signal but weak semantic score still gets excluded by this gate.

PR #5743 (Closes #5742) fixes both: unions semantic + keyword hits and fetches entity-linked memories missing from the pool via vector_store.get(), then changes the threshold gate to pass when any active signal clears the cutoff rather than semantic alone. Applied to the Python sync/async path and the TS OSS mirror, with tests on both sides. Routing to review there.

hegu-1:
The union-candidate fix matters beyond search quality; it affects whether memory can act as a reliable continuity layer.

In a personal/agent memory system, some memories are not semantically close to the current query but are still operationally important because they are:

- entity-linked
- keyword-exact
- procedural conventions
- project-specific constraints
- rare but high-authority decisions

If the candidate pool is seeded only from semantic top-k, those memories can disappear from recall even though another retrieval signal already found them.

[...]

PR #5743, opened 2026-06-22
fix(core): union hybrid search candidates and fix threshold gating

rafid001:
## Linked Issue

Closes #5742

## Description

Fixes hybrid search silently dropping keyword-only and entity-rescued memories:

1. **Candidate union** — merge semantic and keyword hits; fetch entity-linked memories missing from the pool via `vector_store.get()`
2. **Threshold gating** — include candidates when semantic, BM25, or entity boost meets threshold (not semantic-only)

Applied to Python sync/async `_search_vector_store()` and TS OSS search path.

## Review follow-ups

- Added TS tests for `buildHybridCandidateMap` and `addEntityBoostCandidates`
- Parallelized entity-rescue fetches in TS with `Promise.all`
- Narrowed `EntityBoostVectorStore.get()` to async-only typing
- Linked changelog entry to #5743

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Test Coverage

- [x] I added/updated unit tests

Python: `TestBuildHybridCandidateMap`, updated threshold tests in `tests/utils/test_scoring.py`
TypeScript: helper tests + BM25 rescue test in `scoring.test.ts`

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

PR #6378, opened 2026-07-18
fix: include keyword-only hybrid search matches

GautamSharma99:
## Description

  Hybrid search calculated BM25 scores for keyword-search results but constructed its final candidate set exclusively from semantic-search
  results.

  Consequently, a memory returned only by keyword search could never reach the scoring stage. This particularly affected exact lexical
  queries such as identifiers, error codes, uncommon names, and other terms for which keyword retrieval may succeed while semantic
  retrieval does not include the memory in its candidate pool.

  This PR fixes the issue consistently across:

  - Python `Memory`
  - Python `AsyncMemory`
  - TypeScript OSS `Memory`

  ### Implementation

  The search pipeline now constructs candidates from the union of semantic and keyword results.

  - Semantic results remain the canonical candidate when the same memory appears in both result sets.
  - Keyword-only candidates retain their keyword-search payload and enter scoring without an artificial semantic score.
  - Duplicate memory IDs are removed.
  - Expired-memory filtering is applied equally to semantic and keyword-only candidates.
  - Keyword results without a valid positive BM25 score are not promoted into the candidate set.

  The scoring utilities now distinguish keyword-only candidates from low-scoring semantic candidates.

  - Existing semantic threshold behavior is preserved.
  - A genuine semantic result below the configured threshold remains excluded, even if it has a high BM25 score.
  - A candidate with no semantic result may still be ranked when it has a valid BM25 score.
  - Keyword-only candidates use a semantic contribution of zero, avoiding fabricated similarity scores.

  ### Why this matters

  Previously, hybrid search could miss the exact category of results lexical retrieval is intended to recover. For example, a memory
  containing `ERR_AUTH_0042` could be returned by BM25 but silently discarded when it was absent from the semantic top-K results.

  After this change, such keyword-only matches are included and ranked using their normalized BM25 contribution.

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Refactor
  - [ ] Documentation update

  ## Breaking Changes

  N/A. Existing semantic-result scoring and threshold behavior remain unchanged.

  ## Test Coverage

  - [x] Added Python scorer regression coverage
  - [x] Added synchronous Python search regression coverage
  - [x] Added asynchronous Python search regression coverage
  - [x] Added TypeScript scorer regression coverage
  - [x] Added TypeScript OSS search regression coverage
  - [x] Verified Python lint and import sorting
  - [x] Verified the TypeScript package build

  ### Verification performed

  - `pytest tests/utils/test_scoring.py tests/test_memory.py -k 'keyword_only or search_explain_includes_score_details or
  threshold_gates_on_semantic'`
    - 5 passed
  - `ruff check mem0/memory/main.py mem0/utils/scoring.py tests/test_memory.py tests/utils/test_scoring.py`
    - Passed
  - `pnpm exec jest --runInBand src/oss/tests/scoring.test.ts src/oss/tests/memory.hybrid-search.test.ts`
    - 6 passed
  - `pnpm run build`
    - Prettier, CJS/ESM builds, and declaration builds passed
  - Pre-commit Ruff and isort hooks
    - Passed

  A broader Python run produced two unrelated existing failures in embedding call-count tests:

  - `test_add_infer_true_caches_embedding_on_llm_rewrite`
  - `test_update_infer_true_caches_embedding_on_llm_rewrite`

  Both concern entity-link embedding calls in the add pipeline and are unrelated to search candidate construction or hybrid scoring.

  A raw repository-wide `tsc --noEmit` also reports existing errors in unrelated client test configuration and community integration
  imports. The package's supported `pnpm run build`, including declaration generation, passes.

  ## Checklist

  - [x] Self-review performed
  - [x] Tests added that prove the fix
  - [x] Relevant existing tests pass locally
  - [x] No public API documentation changes required

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1619,18 +1619,36 @@ class Memory(MemoryBase):
         if query_entities:
             entity_boosts = self._compute_entity_boosts(query_entities, filters)
 
-        # Step 7: Build candidate set from semantic results
-        candidates = []
+        # Step 7: Build candidate set from the union of semantic and keyword results
+        candidates_by_id = {}
         for mem in semantic_results:
-            payload = mem.payload if hasattr(mem, 'payload') else {}
+            payload = mem.payload if hasattr(mem, "payload") else mem.get("payload", {})
             if not show_expired and _payload_is_expired(payload):
                 continue
-            mem_id = str(mem.id)
-            candidates.append({
+            mem_id = str(mem.id) if hasattr(mem, "id") else str(mem.get("id", ""))
+            if not mem_id:
+                continue
+            candidates_by_id[mem_id] = {
                 "id": mem_id,
-                "score": mem.score,
+                "score": mem.score if hasattr(mem, "score") else mem.get("score"),
                 "payload": payload,
-            })
+            }
+
+        if keyword_results is not None:
+            for mem in keyword_results:
+                mem_id = str(mem.id) if hasattr(mem, "id") else str(mem.get("id", ""))
+                if not mem_id or mem_id in candidates_by_id or mem_id not in bm25_scores:
+                    continue
+                payload = mem.payload if hasattr(mem, "payload") else mem.get("payload", {})
+                if not show_expired and _payload_is_expired(payload):
+                    continue
+                candidates_by_id[mem_id] = {
+                    "id": mem_id,
+                    "score": None,
+                    "payload": payload,
+                }
+
+        candidates = list(candidates_by_id.values())
 
         # Step 8: Score and rank
         scored_results = score_and_rank(
@@ -3259,18 +3277,36 @@ class AsyncMemory(MemoryBase):
         if query_entities:
             entity_boosts = await self._compute_entity_boosts_async(query_entities, filters)
 
-        # Step 7: Build candidate set from semantic results
-        candidates = []
+        # Step 7: Build candidate set from the union of semantic and keyword results
+        candidates_by_id = {}
         for mem in semantic_results:
-            payload = mem.payload if hasattr(mem, 'payload') else {}
+            payload = mem.payload if hasattr(mem, "payload") else mem.get("payload", {})
             if not show_expired and _payload_is_expired(payload):
                 continue
-            mem_id = str(mem.id)
-            candidates.append({
+            mem_id = str(mem.id) if hasattr(mem, "id") else str(mem.get("id", ""))
+            if not mem_id:
+                continue
+            candidates_by_id[mem_id] = {
                 "id": mem_id,
-                "score": mem.score,
+                "score": mem.score if hasattr(mem, "score") else mem.get("score"),
                 "payload": payload,
-            })
+            }
+
+        if keyword_results is not None:
+            for mem in keyword_results:
+                mem_id = str(mem.id) if hasattr(mem, "id") else str(mem.get("id", ""))
+                if not mem_id or mem_id in candidates_by_id or mem_id not in bm25_scores:
+                    continue
+                payload = mem.payload if hasattr(mem, "payload") else mem.get("payload", {})
+                if not show_expired and _payload_is_expired(payload):
+                    continue
+                candidates_by_id[mem_id] = {
+                    "id": mem_id,
+                    "score": None,
+                    "payload": payload,
+                }
+
+        candidates = list(candidates_by_id.values())
 
         # Step 8: Score and rank
         scored_results = score_and_rank(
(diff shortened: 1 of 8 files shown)


mem0/memory/main.py
```diff
@@ -1619,18 +1619,36 @@ def _search_vector_store(self, query, filters, limit, threshold=0.1, explain=Fal
...
                 continue
-            mem_id = str(mem.id)
-            candidates.append({
+            mem_id = str(mem.id) if hasattr(mem, "id") else str(mem.get("id", ""))
+            if not mem_id:
+                continue
+            candidates_by_id[mem_id] = {
                 "id": mem_id,
-                "score": mem.score,
+                "score": mem.score if hasattr(mem, "score") else mem.get("score"),
```
kartik-mem0, line 1633: score=None from a vector store's search() (unreachable on any shipped backend today, but not guaranteed) is stored as-is here instead of coalescing to 0.0 like TS does at index.ts:1531 (`mem.score ?? 0`). Combined with the is_keyword_only check in scoring.py, that could misclassify a None-score semantic hit as keyword-only and skip the threshold gate. Same gap in the async twin at :3291.

mem0/memory/main.py
```diff
@@ -1619,18 +1619,36 @@ def _search_vector_store(self, query, filters, limit, threshold=0.1, explain=Fal
...
+        candidates_by_id = {}
         for mem in semantic_results:
-            payload = mem.payload if hasattr(mem, 'payload') else {}
+            payload = mem.payload if hasattr(mem, "payload") else mem.get("payload", {})
             if not show_expired and _payload_is_expired(payload):
                 continue
-            mem_id = str(mem.id)
-            candidates.append({
+            mem_id = str(mem.id) if hasattr(mem, "id") else str(mem.get("id", ""))
+            if not mem_id:
```
kartik-mem0, line 1629: New `if not mem_id: continue` guard on the semantic loop (same at the async twin :3287), untested and not mentioned in the PR description. It silently drops any semantic hit whose id extraction comes back empty, a case no current backend produces. Add a regression test, or drop it since this fix's stated scope is the keyword-only union, not semantic-loop guards.

kartik-mem0 (commented):
## Changes requested

Unions keyword-only BM25 hits into the candidate pool so they can surface in `search()`, proven live on real Qdrant and pgvector in both Python and TS; two small gaps need a fix first.

**To unblock:**
1. `mem0/memory/main.py:1633` (async twin `:3291`) - coalesce `mem.score` to `0.0` when `None`, matching TS's `mem.score ?? 0` (detail in the inline comment)
2. `mem0/memory/main.py:1629` (async twin `:3287`) - new `if not mem_id: continue` guard on the semantic loop is untested and outside this fix's stated scope (detail in the inline comment)

Optional: the TS keyword loop (`index.ts:1539`) skips the same empty-id guard Python has; no test covers a memory ID present in both semantic and keyword results; the ~30-line union block is duplicated verbatim between sync and async, `#5743` already built a shared `build_hybrid_candidate_map()` for this.

- **Problem is real:** reproduces on `main` with a real backend, both languages: blackout `vectorStore.search()`, seed a keyword-only hit, `search()` returns zero results even though `keywordSearch()`/`keyword_search()` has a live match.
- **Tests / red-green:** 5/5 Python + 6/6 TS targeted tests fail with the fix reverted, pass restored.
- **Live sandbox:** PASS on `oss-qdrant` and `oss-pgvector`, both Python sync/async and TS, after accounting for a pre-existing TS+Qdrant `keywordSearch()` gap that main's `#5851` already fixed (unrelated to this PR, branch just predates it).
- **CI:** no real check ran (first-time-contributor `action_required` gate); `Vercel` red is advisory/fork-permissions, ignored; `license/cla` passes.
- **Security:** cleared, no new I/O or untrusted input, same trust boundary as the existing `search()` path.
- **Bloat (ponytail):** `net: -28 lines possible` (dedupe the sync/async union block, drop the dead `mem_id` fallback, reuse the TS candidate type); non-blocking, this file already duplicates all of `_search_vector_store` between sync and async so the pattern is not new here.
- **Not verified:** entity-rescue is explicitly out of scope for this PR (see Related below).

</details>

Solid, well-scoped fix for what it claims, real-backend proof both languages. Push the two one-line items above and I will re-review.

Related: #5742 also names entity-rescue drops, untouched here. #5743 already implemented that half (candidate union, entity-rescue, and an OR-style threshold gate), passed review, reached LGTM, but was closed 2026-08-03 for staleness/merge conflicts, not for any quality issue. Worth reviving that work separately so #5742 can close in full; not a condition on this PR. Also, for tracking, mind adding `Closes #5742` (partial) or a fresh issue link to the PR body?

PR #6380, opened 2026-07-18
Fix/delete all complete pagination

GautamSharma99:
## Description

  `delete_all()` previously called `vector_store.list()` only once and deleted the memories returned by that single call.

  Many vector-store providers limit `list()` to a default page size, commonly 100 records. When more matching memories existed,
  `delete_all()` deleted only the first page but still returned:

  > Memories deleted successfully!

  This could leave older memories behind in privacy, account-deletion, and data-retention workflows.

  This PR updates bulk deletion across:

  - Python `Memory`
  - Python `AsyncMemory`
  - TypeScript OSS `Memory`

  ## Implementation

  Bulk deletion now repeatedly lists the currently matching first page and deletes each returned memory until the vector store returns no
  unseen matches.

  ### Python sync

  - Re-lists matching memories after each deletion batch.
  - Handles the different list result shapes used by vector-store providers.
  - Tracks processed IDs to prevent repeated or stale results from causing an infinite loop.
  - Reports the total number of deleted memories to logging and decay-usage notices.

  ### Python async

  - Processes every available page rather than only the first.
  - Preserves concurrent deletion within each page.
  - Preserves the existing partial-failure behavior using `asyncio.gather(..., return_exceptions=True)`.
  - Attempts each returned memory only once, preventing persistent failures from causing an infinite loop.
  - Aggregates deletion failures and successful deletion counts across all pages.
  - Performs entity-store cleanup after all deletion batches.
  - Uses the total successful deletion count for notices and logging.

  ### TypeScript OSS

  - Re-lists after each page is deleted.
  - Tracks processed memory IDs to avoid repeated processing.
  - Preserves the existing fail-fast behavior if an individual deletion throws.
  - Uses the total deletion count across all pages for the bulk-delete notice.

  ## Why repeated first-page deletion?

  The vector-store interface does not expose a consistent pagination contract:

  - Providers use different pagination mechanisms.
  - The second `list()` return value is not consistently a cursor.
  - Some providers return a count while others expose provider-specific pagination data.

  Deleting the current first page and listing again provides provider-neutral traversal without introducing provider-specific branching or
  changing the public vector-store interface.

  ## Regression Coverage

  Tests simulate 101 matching memories delivered as:

  1. First page: 100 memories
  2. Second page: 1 memory
  3. Final page: empty

  Coverage verifies that:

  - All 101 memory IDs are deleted.
  - The vector store is listed until it returns no further matches.
  - Python synchronous deletion crosses the page boundary.
  - Python asynchronous deletion crosses the page boundary.
  - Async deletion continues when one memory fails.
  - TypeScript deletion crosses the page boundary.
  - The TypeScript decay-usage notice receives the complete deletion count of 101.

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Refactor
  - [ ] Documentation update

  ## Breaking Changes

  N/A.

  The public API and success response remain unchanged. The method now fulfills its documented behavior by deleting every matching memory
  rather than only the first provider page.

  ## Test Coverage

  - [x] Added/updated Python unit tests
  - [x] Added/updated TypeScript unit tests
  - [x] Tested the affected paths locally

  ### Verification performed

  - `pytest tests/test_main.py -k delete_all`
    - 6 passed
  - Async `delete_all` regression selection from `tests/test_memory.py`
    - 3 passed
  - Ruff on the affected Python files
    - Passed
  - Pre-commit Ruff and isort hooks
    - Passed
  - `pnpm exec jest --runInBand src/oss/tests/memory.crud.test.ts`
    - 51 passed
  - `pnpm run build`
    - Prettier passed
    - CJS and ESM builds passed
    - TypeScript declaration builds passed

  ## Checklist

  - [x] Self-review performed
  - [x] Tests added that prove the fix
  - [x] Relevant new and existing tests pass locally
  - [x] No public API documentation changes required

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1722,18 +1722,30 @@ export class Memory {
       );
     }
 
-    const [memories] = await this.vectorStore.list(filters);
-    for (const memory of memories) {
-      await this.deleteMemory(memory.id);
+    let deletedCount = 0;
+    const processedMemoryIds = new Set<string>();
+    while (true) {
+      const [listedMemories] = await this.vectorStore.list(filters);
+      const memories = listedMemories.filter(
+        (memory) => !processedMemoryIds.has(String(memory.id)),
+      );
+      if (memories.length === 0) break;
+
+      for (const memory of memories) {
+        const memoryId = String(memory.id);
+        processedMemoryIds.add(memoryId);
+        await this.deleteMemory(memoryId);
+        deletedCount += 1;
+      }
     }
 
     const result = { message: "Memories deleted successfully!" };
-    if (memories.length > 0) {
+    if (deletedCount > 0) {
       await this._displayDecayUsageNotice({
         triggerFunction: "delete_all",
         triggerSource: "delete_all",
         triggerReason: "bulk_delete",
-        deletedCount: memories.length,
+        deletedCount,
       });
     } else {
       await this._displayFirstRunNotice("delete_all");
--- a/mem0-ts/src/oss/tests/memory.crud.test.ts
+++ b/mem0-ts/src/oss/tests/memory.crud.test.ts
@@ -571,6 +571,45 @@ describe("Memory - deleteAll()", () => {
     expect(remaining.results).toHaveLength(0);
   });
 
+  test("deletes memories across multiple vector-store pages", async () => {
+    const pagedMemory = createMemory();
+    const internals = pagedMemory as any;
+    await internals._ensureInitialized();
+
+    const memories = Array.from({ length: 101 }, (_, index) => ({
+      id: String(index),
+      payload: { data: `Memory ${index}`, user_id: userId },
+    }));
+    const listSpy = jest
+      .spyOn(internals.vectorStore, "list")
+      .mockResolvedValueOnce([memories.slice(0, 100), 101])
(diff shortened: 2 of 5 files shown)


kartik-mem0:
Thanks for this, and for extending #6348 to cover TypeScript as well.

Closing this one, but for two different reasons on the two halves:

**Python** is now covered on main. #6636 was merged as 54328ffd and closes #6627.

**TypeScript** is still broken on main (`mem0-ts/src/oss/src/memory/index.ts:1730`), but we are consolidating that on #4872, which was open first and takes the approach that matches what landed in Python: it passes an explicit `DELETE_ALL_BATCH_SIZE` to `list()` so each page is actually large, rather than re-listing the default page repeatedly.

One thing worth flagging from this diff, since it is a subtle one. `processedMemoryIds` is added before the delete is awaited:

```ts
processedMemoryIds.add(memoryId);
await this.deleteMemory(memoryId);
deletedCount += 1;
```

If `deleteMemory` throws, the ID is already marked processed, so the next `list()` returns it and the filter drops it. The memory survives, `deleteAll` reports success, and nothing surfaces. That is the same silent-partial-deletion shape as the original #6627 bug, which is why the merged version terminates on a repeated batch rather than on a processed-ID set.

Genuinely appreciate the work on both PRs, and sorry for the slow triage on this cluster.

Issue #5665, opened 2026-06-19
TS SDK: AnthropicLLM ignores configured baseURL (proxy/gateway silently bypassed)


yashs33244:
### Bug

`mem0-ts/src/oss/src/llms/anthropic.ts:17` constructs `new Anthropic({ apiKey })` and never reads `config.baseURL`, even though `LLMConfig` (`src/oss/src/types/index.ts:43`) declares `baseURL?: string`. A proxy/gateway user who sets `baseURL` in the TS SDK is silently ignored — the identical bug fixed for the Python SDK in #5626.

### Expected

Pass `baseURL` to the client only when set:
```ts
const clientArgs: { apiKey: string; baseURL?: string } = { apiKey };
if (config.baseURL) clientArgs.baseURL = config.baseURL;
this.client = new Anthropic(clientArgs);
```

### Context

TypeScript counterpart of the Python fix in #5626 (flagged in review).

kartik-mem0:
## Triage Notes

**What we've established:**
- `mem0-ts/src/oss/src/llms/anthropic.ts` constructs `new Anthropic({ apiKey })` and never reads `config.baseURL`, even though `LLMConfig` (`src/oss/src/types/index.ts`) declares `baseURL?: string`.
- A user who sets `baseURL` in their TS OSS config (for a proxy, gateway, or self-hosted Anthropic-compatible endpoint) gets the flag silently ignored — all traffic goes to the default Anthropic API endpoint instead.
- This is the TypeScript counterpart of the Python fix in #5626, which addressed the same omission in the Python SDK's Anthropic LLM provider.
- The fix is straightforward: pass `baseURL` to the Anthropic client constructor only when `config.baseURL` is set, matching the pattern already used in other LLM providers in this SDK.

**Reproduction:** Confirmed (code tracing: `mem0-ts/src/oss/src/llms/anthropic.ts:17`, `LLMConfig.baseURL` declared but never read in the Anthropic provider)

**Root cause:** The Anthropic client initialization at line 17 of `anthropic.ts` hard-codes `new Anthropic({ apiKey })`, ignoring any `baseURL` from the config — an omission from when the TS provider was written without porting the proxy/gateway support from the Python counterpart.

**Next step:** We'll queue a fix. The change is minimal — conditionally pass `baseURL` to the constructor when set, matching how the Python SDK handles it. No open PR yet; this is on our radar.

PR #6389, opened 2026-07-18
fix(ts-sdk): lazy-load optional provider SDKs in mem0ai/oss

### LINKED ISSUE #6388: mem0ai/oss crashes at import when unused optional provider SDKs aren't installed
### Which SDK / package

`mem0ai` (TypeScript OSS — `mem0ai/oss`)

### Describe the bug

Importing the module that constructs `new Memory()` from `mem0ai/oss` fails at **load time** unless a large set of optional provider SDKs are installed, even when the user configured only pgvector + OpenAI. A NestJS user reports having to `npm install` 10+ dependencies they don't use just to boot the app (plus npm peer-dependency warnings for each).

### Root cause

The OSS entry barrel `mem0-ts/src/oss/src/index.ts` uses `export *` to re-export every provider. Under both ESM and CJS, importing anything from `mem0ai/oss` eagerly evaluates every provider module's top level. Twelve provider modules value-import their SDK at module top (`import Anthropic from "@anthropic-ai/sdk"`, `import { Ollama } from "ollama"`, …), so Node runs `require("ollama")` (and 11 siblings) the instant the barrel loads — before any user config is read. Result: `MODULE_NOT_FOUND: Cannot find module 'ollama'` at `new Memory()` for a user who never selected Ollama.

### Steps to reproduce

1. In a fresh project, install `mem0ai` + only the pgvector/OpenAI deps.
2. `import { Memory } from "mem0ai/oss"` and construct `new Memory({...})` with a pgvector + OpenAI config.
3. App fails to start: `Cannot find module 'ollama'` (or another unselected provider SDK).

### Expected behavior

Optional provider SDKs load lazily, only when that provider is actually selected. A user of pgvector + OpenAI should not need any other provider package installed.

### Scope

This covers the **non-SQLite** optional providers (qdrant, redis, supabase, cloudflare vectorize, azure ai search, anthropic, google, groq, ollama, mistral, langchain). The SQLite / `disableHistory` side of the same class of problem is tracked separately in #3291 and #5720.

### Labels

bug, sdk-typescript

kartik-mem0:
## Linked Issue

Closes #6388

Related: #3291 (SQLite / `disableHistory` half of the same class of problem, handled by #5720) · leaner alternative to the stalled #4444 (see note below).

## Description

Importing `Memory` from `mem0ai/oss` crashes at **load time** whenever an optional provider SDK is absent, even for a user who configured only pgvector + OpenAI.

**Root cause:** the OSS entry barrel `mem0-ts/src/oss/src/index.ts` uses `export *` to re-export every provider, so ESM/CJS eagerly evaluates each provider module's top level on any import from `mem0ai/oss`. Twelve provider modules value-imported their SDK at module top (`import Anthropic from "@anthropic-ai/sdk"`, `import { Ollama } from "ollama"`, …), so Node runs `require("ollama")` (and 11 siblings) the instant the barrel loads — before any user config is read. A pgvector + OpenAI user still crashed with `MODULE_NOT_FOUND: Cannot find module 'ollama'` at `new Memory()` and had to install 10+ unused peers just to boot (plus npm peer-dependency warnings).

**Fix:** switch those 12 SDKs from top-level value `import` to `import type` (erased at compile) and load each lazily via `await import()` inside the code path that needs a client — the exact pattern already used by `pinecone.ts` and the 16 previously-migrated providers. Vector stores route every public method through a memoized `initialize()`, with the private `ensureClient()` called only inside `_doInitialize()` (race-free); LLMs, embedders, and the Supabase history manager call `ensureClient()` at each public entry (idempotent; these clients open no connection on construction). Each lazy import throws a friendly `"...is required to use the <X>. Install it with: npm install <pkg>"` if the peer is genuinely missing.

This also marks those 12 SDKs `optional` in `package.json` `peerDependenciesMeta` — silencing the peer-warning half of the report and **activating the repo's existing static-import guard** (`tests/optional-peers.test.ts`), which now fails CI if any provider regresses to a static import. Excludes `pg` + `better-sqlite3` (default/required backends; SQLite lazification is #5720).

> **Note on #4444:** @kgarg2468's open PR #4444 targets the same root cause more broadly (centralized loader util + pack-smoke gate + README + npm scripts, 26 files / +976−76). This PR is a deliberately leaner take scoped to the minimal lazification + activating the guard that already exists in the repo. Credit to @kgarg2468 for surfacing the approach.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — no public API signature change. `Memory`, all provider class names, and constructor signatures are unchanged; the `config.client` pass-in branch is preserved everywhere it existed. The only behavioral change for a correctly-configured user is that client construction moves from the constructor to first use (async). A user missing a peer they *actually* configured now gets a clear "install it with: npm install <pkg>" error at first use instead of a barrel-load crash.

## Test Coverage

**Regression guard (ships in CI):** `tests/optional-peers.test.ts` scans `peerDependenciesMeta.optional` and asserts no optional peer is statically value-imported by `src`. Marking the 12 SDKs optional activates it for them.

**Red → green at the reporter's exact layer** (require-hook hides every optional peer, then `require("dist/oss/index.js")`):

```
# BEFORE (origin/main dist), all optional peers hidden:
CRASH: MODULE_NOT_FOUND Cannot find module 'ollama'   (exit 1)

# AFTER (this branch dist), all 34 optional peers hidden:
OK: required dist/oss/index.js with all optional peers hidden.
    Exports: AWSBedrockEmbedder, AWSBedrockLLM, AnthropicLLM, AzureAISearch, ...   (exit 0)
```

**Targeted gate (local):**
```
npx jest optional-peers vector-stores-compat tsup-externals  → 3 suites, 125 passed
npx tsup   → build success (CJS + ESM + DTS; all `import type` changes typecheck)
npx prettier --check <18 changed files>  → all clean
```

## Checklist

### DIFF
--- a/docs/open-source/configuration.mdx
+++ b/docs/open-source/configuration.mdx
@@ -148,7 +148,7 @@ See the full catalog in <Link href="/components/llms/overview">Components</Link>
 - Qdrant connection errors: confirm port `6333` is exposed and the API key (if set) matches.
 - Empty search results: verify the embedder model name. A mismatch causes dimension errors.
 - `Unknown reranker` (Python): upgrade the SDK with `pip install --upgrade mem0ai` to load the latest provider registry.
-- `Cannot find module` (Node): import from the OSS entry point, `import { Memory } from "mem0ai/oss"`, not `"mem0ai"`.
+- `Cannot find module` (Node): two common causes. First, import from the OSS entry point, `import { Memory } from "mem0ai/oss"`, not `"mem0ai"`. Second, provider SDKs are optional peer dependencies loaded on demand, so install the one for the provider you configured (for example `npm install @qdrant/js-client-rest` for Qdrant). Installing `mem0ai` alone only pulls in the providers used by default; you do not need SDKs for providers you never select.
 
 <CardGroup cols={2}>
   <Card
--- a/mem0-ts/package.json
+++ b/mem0-ts/package.json
@@ -151,6 +151,42 @@
     "@aws-sdk/client-bedrock-runtime": ">=3.0.0 <3.968.0"
   },
   "peerDependenciesMeta": {
+    "@qdrant/js-client-rest": {
+      "optional": true
+    },
+    "redis": {
+      "optional": true
+    },
+    "@supabase/supabase-js": {
+      "optional": true
+    },
+    "cloudflare": {
+      "optional": true
+    },
+    "@azure/search-documents": {
+      "optional": true
+    },
+    "@azure/identity": {
+      "optional": true
+    },
+    "@langchain/core": {
+      "optional": true
+    },
+    "@anthropic-ai/sdk": {
+      "optional": true
+    },
+    "@google/genai": {
+      "optional": true
+    },
+    "groq-sdk": {
+      "optional": true
+    },
+    "@mistralai/mistralai": {
+      "optional": true
+    },
+    "ollama": {
+      "optional": true
+    },
     "mysql2": {
       "optional": true
     },
--- a/mem0-ts/src/oss/src/embeddings/fastembed.ts
+++ b/mem0-ts/src/oss/src/embeddings/fastembed.ts
@@ -1,6 +1,7 @@
 import type { FlagEmbedding } from "fastembed";
 import { Embedder } from "./base";
 import { EmbeddingConfig } from "../types";
+import { loadPeer } from "../utils/load_peer";
 
 // FastEmbed only ships a fixed set of ONNX models (fastembed's `EmbeddingModel`
 // enum, minus CUSTOM). Mirrored here as literals so an invalid model name can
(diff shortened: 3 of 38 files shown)


kartik-mem0:
### CI status

**`build_ts_sdk` (the job this change affects): green on Node 20 and 22.**

The first CI run was red on `build_ts_sdk` because lazifying the provider SDKs moves client construction from the constructor to first use (`await import(...)`). Two existing ctor-timing regression tests asserted synchronous construction:

- `anthropic-llm.test.ts` — baseURL forwarding (#5665)
- `qdrant-url-port.test.ts` — port extraction (qdrant/qdrant-js#59)

Both were updated to `await` the async init before asserting the forwarded args (0e77c677). Their assertions are unchanged, so they still bite. Full `test:unit` is green locally and in CI: **87 suites / 1360 tests**.

**The only remaining red is `integration_ts_sdk`, and it is unrelated to this change — a live-backend flake, not a regression.**

- This diff is OSS-only (`src/oss/**` + `package.json` peer metadata). The integration suite tests the hosted `MemoryClient` (`src/client/tests/integration/**`) against the real API.
- The failing subset **shuffles across reruns of the identical commit**: run 1 failed `management` + `search`; a plain re-run passed those two and failed `batch` instead. Every failure is `waitForMemories: expected at least 1 memories … after 4 attempts` (ingestion lag) plus an occasional transient `AuthenticationError`. A code regression fails the same tests every run; a shuffling failure set on unchanged code is flakiness.

Re-ran the failed jobs once to confirm — still flaky-red on a different subset. This needs a maintainer re-run/merge decision; I haven't touched the integration tests as they're out of scope for this fix.

karthik-indla (approved):

PR #6397, opened 2026-07-18
fix(vector_stores/supabase): persist metadata-only updates instead of silently dropping them

### LINKED ISSUE #6418: Supabase metadata-only `update()` is silently dropped
### What is wrong

A metadata-only `Supabase.update()` (called with `vector=None`) tries to recover the embedding from `existing.payload.get("vector")`, but the vector is never stored in the payload/metadata, so it resolves to empty, the `if vector:` guard is skipped, and the update is silently dropped (no error, no write). Memory edits are lost on Supabase.

`mem0/vector_stores/supabase.py` (`update()`).

### Fix

Proposed in #6397 (draft) - fetch the real stored vector and persist the update.

yashs33244:
## Linked Issue

Closes #6418

## Description

A metadata-only `update()` (called with `vector=None`) tried to recover the embedding from `existing.payload.get("vector")`, but the vector is never stored in the payload/metadata, so it resolved to empty, the `if vector:` guard was skipped, and the update was silently dropped. It now fetches the real stored vector from the collection and persists the new payload.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `test_update_metadata_only_persists` (the upsert runs on a metadata-only update). `test_supabase.py` 11/11, ruff clean.

## Checklist

### DIFF
--- a/mem0/vector_stores/supabase.py
+++ b/mem0/vector_stores/supabase.py
@@ -156,12 +156,12 @@ class Supabase(VectorStoreBase):
             payload (Dict, optional): Updated payload
         """
         if vector is None:
-            # If only updating metadata, we need to get the existing vector
-            existing = self.get(vector_id)
-            if existing and existing.payload:
-                vector = existing.payload.get("vector", [])
+            # If only updating metadata, recover the stored embedding from the collection.
+            existing = self.collection.fetch([(vector_id,)])
+            if existing:
+                vector = existing[0][1]
 
-        if vector:
+        if vector is not None:
             self.collection.upsert([(vector_id, vector, payload or {})])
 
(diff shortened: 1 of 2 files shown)


yashs33244:
Been open ~17 days with green CI 🙏 Fixes Supabase silently dropping metadata-only updates. @kartik-mem0 @deshraj a quick look whenever there's bandwidth?

yashs33244:
Been open a few weeks with green CI, still awaiting a first review 🙏 This one Supabase no longer silently drops metadata-only updates. @kartik-mem0 @Himanshu-Sangshetti @karthik-indla - whenever one of you has a window, would really appreciate a look. Thanks!

PR #6401, opened 2026-07-18
fix(vector_stores/milvus): guard text field in update() for legacy collections

### LINKED ISSUE #6422: Milvus `update()` breaks legacy collections by writing an unguarded `text` field
### What is wrong

`MilvusDB.insert()` guards the top-level `text` field behind `self._has_bm25_schema` because legacy pre-v3 collections have no `text` field and reject unknown fields, but `update()` adds `text` unconditionally. Every `update()` on a legacy collection is therefore rejected by Milvus.

`mem0/vector_stores/milvus.py` (`update()`).

### Fix

Proposed in #6401 (draft) - guard `text` in `update()` the same way `insert()` does.

yashs33244:
## Linked Issue

Closes #6422

## Description

`insert()` guards the top-level `text` field behind `self._has_bm25_schema` because legacy pre-v3 collections have no `text` field and reject unknown fields, but `update()` added `text` unconditionally, so every update on a legacy collection was rejected. `update()` now guards `text` the same way `insert()` does.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Added `test_update_omits_text_on_legacy_collection` and `test_update_includes_text_on_bm25_collection`. `test_milvus.py` 28/28, ruff clean.

## Checklist

### DIFF
--- a/mem0/vector_stores/milvus.py
+++ b/mem0/vector_stores/milvus.py
@@ -298,10 +298,12 @@ class MilvusDB(VectorStoreBase):
             if payload is None:
                 payload = existing[0].get("metadata")
 
-        text = ""
-        if payload:
-            text = (payload.get("text_lemmatized") or payload.get("data", ""))[:65535]
-        schema = {"id": vector_id, "vectors": vector, "metadata": payload, "text": text}
+        schema = {"id": vector_id, "vectors": vector, "metadata": payload}
+        if self._has_bm25_schema:
+            text = ""
+            if payload:
+                text = (payload.get("text_lemmatized") or payload.get("data", ""))[:65535]
+            schema["text"] = text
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Closing this as already fixed on main. #5705 merged today (d653b63f) and carries the identical guard: mem0/vector_stores/milvus.py:301-306 now builds the schema without a text key and only adds it when self._has_bm25_schema is true, which is exactly the change proposed here.

Test coverage landed too, in a slightly stronger form: tests/vector_stores/test_milvus.py:348 derives _has_bm25_schema through describe_collection instead of setting the flag by hand, and :373 covers the BM25 branch.

Your diagnosis was right and it matched the fix that shipped. Sorry it took a second PR to get there. Thanks for the report and the fix, and please do send more.

PR #6404, opened 2026-07-18
fix(opencode-plugin): recover MEM0_API_KEY from shell profiles

rodboev:
## Linked Issue

Refs #6003

Companion to [#6012](https://github.com/mem0ai/mem0/pull/6012), which covers the separate OpenCode Desktop load-failure and project-identity slices from the same issue.

Attribution: this follows the scoped fix direction in [maintainer triage](https://github.com/mem0ai/mem0/issues/6003#issuecomment-4862984497) and mirrors the repo's existing plugin-family resolver behavior in `integrations/mem0-plugin/scripts/_identity.py`.

## Description

OpenCode Desktop can load `@mem0/opencode-plugin` without exposing any memory tools when `MEM0_API_KEY` exists in a supported shell profile but is absent from the Desktop process environment. The plugin currently treats that missing process env var as a missing credential, logs `MEM0_API_KEY environment variable not set...`, and returns before it registers hooks or tools.

This PR adds a package-local resolver in `.opencode-plugin` that keeps explicit `MEM0_API_KEY` env precedence, then falls back to the same allowlisted shell-profile assignment pattern the Mem0 plugin family already uses elsewhere in the repo. Startup now uses the resolved key through the existing guard instead of reading `process.env.MEM0_API_KEY` directly.

## Root cause

`integrations/mem0-plugin/.opencode-plugin/opencode-mem0.ts` reads only `process.env.MEM0_API_KEY` before startup. The repo's existing Desktop-safe resolver lives in `integrations/mem0-plugin/scripts/_identity.py`, but the OpenCode TypeScript package never received an equivalent local path, so Desktop startup exits early even when a valid literal `MEM0_API_KEY` is already present in `.zshrc`, `.bashrc`, or the other supported profile files.

## Scope

This change is limited to the OpenCode plugin package and the API-key fallback slice of #6003.

- Explicit `MEM0_API_KEY` env precedence stays unchanged.
- The current missing-key log and empty-plugin return stay unchanged when no source yields a key.
- The resolver reads only the fixed allowlist of shell-profile files and only literal `MEM0_API_KEY=` assignments. It does not source profiles or execute shell code.
- Build/runtime packaging, `app_id`, project context, docs wording, package versions, and other plugin hosts stay out of scope.

## Type of Change

- [ ] Documentation update

## Test Coverage

- [ ] No tests needed

Verification from `integrations/mem0-plugin/.opencode-plugin`:

- [x] `bun install --frozen-lockfile` (`Checked 353 installs across 358 packages (no changes)`)
- [x] `bun test api-key.test.ts` (`8 pass`, `0 fail`, `16 expect() calls`)
- [x] `bun run type-check` (`tsc --noEmit`)
- [x] `bun run build` (`Bundled 204 modules`, `index.js  0.96 MB`)
- [x] `bun test` (`40 pass`, `0 fail`, `86 expect() calls`)

## Checklist

### DIFF
--- a/integrations/mem0-plugin/.opencode-plugin/api-key.ts
+++ b/integrations/mem0-plugin/.opencode-plugin/api-key.ts
@@ -1,0 +1,30 @@
+import {readFileSync} from "fs";
+import {homedir} from "os";
+import {join} from "path";
+
+const PROFILE_FILES = [".zshrc", ".bashrc", ".zprofile", ".bash_profile", ".profile"];
+
+export function parseApiKeyLine(line: string): string | undefined {
+  const match = line.match(/^\s*(?:export\s+)?MEM0_API_KEY=(.*)$/);
+  if (!match) return undefined;
+
+  const value = match[1].replace(/#.*$/, "").trim().replace(/^("|')(.*)\1$/, "$2").trim();
+  return value && !value.startsWith("$") ? value : undefined;
+}
+
+export function resolveApiKey(env: NodeJS.ProcessEnv = process.env, homeDir = homedir()): string {
+  const explicit = env.MEM0_API_KEY?.trim();
+  if (explicit) return explicit;
+
+  for (const profile of PROFILE_FILES) {
+    try {
+      for (const line of readFileSync(join(homeDir, profile), "utf8").split(/\r?\n/)) {
+        const key = parseApiKeyLine(line);
+        if (key) return key;
+      }
+    } catch {
+    }
+  }
+
+  return "";
+}
--- a/integrations/mem0-plugin/.opencode-plugin/api-key.test.ts
+++ b/integrations/mem0-plugin/.opencode-plugin/api-key.test.ts
@@ -1,0 +1,135 @@
+import {afterEach, describe, expect, test} from "bun:test";
+import {mkdtempSync, mkdirSync, writeFileSync} from "fs";
+import {tmpdir} from "os";
+import {join} from "path";
+import {parseApiKeyLine, resolveApiKey} from "./api-key";
+import Mem0Plugin from "./opencode-mem0";
+
+const originalKey = process.env.MEM0_API_KEY;
+const originalHome = process.env.HOME;
+const originalUserProfile = process.env.USERPROFILE;
+const originalTelemetry = process.env.MEM0_TELEMETRY;
+const originalFetch = globalThis.fetch;
+
+afterEach(() => {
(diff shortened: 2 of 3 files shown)


integrations/mem0-plugin/.opencode-plugin/api-key.ts
```diff
@@ -0,0 +1,30 @@
...
+import {homedir} from "os";
+import {join} from "path";
+
+const PROFILE_FILES = [".zshrc", ".bashrc", ".zprofile", ".bash_profile", ".profile"];
+
+export function parseApiKeyLine(line: string): string | undefined {
+  const match = line.match(/^\s*(?:export\s+)?MEM0_API_KEY=(.*)$/);
+  if (!match) return undefined;
+
+  const value = match[1].replace(/#.*$/, "").trim().replace(/^("|')(.*)\1$/, "$2").trim();
```
kartik-mem0, line 11: Minor edge (and it's faithful to `_identity.py`, so not a regression): the comment strip `replace(/#.*$/, "")` runs **before** the quote strip, so a `#` inside a quoted value corrupts it — `MEM0_API_KEY="m0-ab#cd"` becomes `"m0-ab` (comment strip → `"m0-ab`, then the balanced-quote regex no longer matches). Mem0 keys are `m0-…` and don't contain `#`, so this is theoretical, but if you want to be exact, strip quotes first and only strip an *unquoted* trailing comment. Low priority — flag only, safe to ship as-is.

kartik-mem0 (commented):
Ponytail review (complexity/over-engineering focus). Score: 9/10 — lean, ship it.

`api-key.ts` is 30 lines and does exactly one thing: env precedence, then an allowlisted profile scan with no shell execution. It's a faithful TS port of the established `integrations/mem0-plugin/scripts/_identity.py` resolver (rung 2: reuse the pattern already in the repo — done), and the guard swap in `opencode-mem0.ts` is a clean one-liner. Nothing to cut in the resolver.

No over-engineering: no config for a fixed allowlist, no abstraction layer, no new dep. The 153-line test file looks heavy against 31 lines of code, but the two full-plugin-startup tests (`keeps the missing-key guard` / `recovers the issue's shell-profile startup path`) are the ones that actually prove the bug fix end to end, and the 4 parse/resolve unit tests are the lean core — I wouldn't cut any of it.

One minor parsing edge inline (faithful to the Python source, so not a regression). Otherwise clean.

`net: 0 lines to cut. Lean already.`

kartik-mem0 (approved):

PR #6406, opened 2026-07-19
fix(mem0): add test for issue #6342

### LINKED ISSUE #6342: update() metadata silently overwrites user_id/agent_id/run_id in TS SDK, breaking tenant isolation

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6342.

## Why
Fixes #6342

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6342.py
+++ b/tests/test_issue_6342.py
@@ -1,0 +1,35 @@
+"""Regression test for TypeScript OSS update() tenant scope metadata.
+
+Issue #6342: caller metadata passed to memory.update() must not be able to
+overwrite the stored user_id, agent_id, or run_id values. Those fields define
+the tenant scope used by getAll(), search(), and deleteAll().
+"""
+
+import re
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+with (
+    patch("importlib.metadata.version", return_value="0.0.0"),
+    patch.dict("sys.modules", {"posthog": MagicMock(), "qdrant_client": MagicMock()}),
+):
+    from mem0 import Memory
+
+
+def test_issue_6342():
+    """updateMemory() must re-pin tenant identity after caller metadata."""
+    assert Memory is not None
+
+    source = Path("mem0-ts/src/oss/src/memory/index.ts").read_text()
+    match = re.search(r"const newMetadata = \{(?P<body>.*?)\n\s*\};", source, re.DOTALL)
+
+    assert match is not None
+
+    body = match.group("body")
+    metadata_spread_index = body.index("...metadata")
+
+    for identity_key in ("user_id", "agent_id", "run_id"):
+        repin = f"{identity_key}: existingMemory.payload.{identity_key}"
+
+        assert repin in body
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

**Finding:** `tests/test_issue_6342.py` greps `mem0-ts/src/oss/src/memory/index.ts` and fails on its own parsing assumption, which is already wrong on `main`:

```
>   metadata_spread_index = body.index("...metadata")
E   ValueError: substring not found
```

It also opens the file via the CWD-relative path `Path("mem0-ts/src/oss/src/memory/index.ts")`, so it only resolves when pytest runs from the repo root. Note also that issue #6342 is already closed.

---

Closing this one. To be clear, this is about the test file, not about issue #6342 — the underlying report is legitimate and stays as it is.

PR #6408, opened 2026-07-19
fix(mem0): add test for issue #6362

### LINKED ISSUE #6362: bug(reranker): failure fallback mutates the caller's documents with rerank_score=0.0
When a reranker provider throws (transient API 5xx, tokenizer error, etc), the `except` fallback in the cohere, huggingface, sentence_transformer and zero_entropy rerankers writes the sentinel score directly onto the caller's input dicts:

```python
except Exception as e:
    logger.warning("Cohere reranking failed, falling back to original order: %s", e)
    for doc in documents:
        doc['rerank_score'] = 0.0
```

(`mem0/reranker/cohere_reranker.py:87`, `huggingface_reranker.py:166`, `sentence_transformer_reranker.py:111`, `zero_entropy_reranker.py:98`)

The success paths in those same files all do `doc.copy()` before attaching `rerank_score`, and `LLMReranker` copies even on its per-document failure path (`llm_reranker.py:159`), so the fallback is the odd one out.

Why it matters: `Memory.search()` passes its live result list into `rerank()` (`mem0/memory/main.py:1458`, async twin around line 3096). So a single transient provider failure permanently stamps `rerank_score: 0.0` onto the actual memory dicts the caller receives, from a call that otherwise looks successful (the only signal is a `logger.warning`).

Repro sketch:

```python
docs = [{"memory": "alpha", "score": 0.9}]
reranker.rerank("query", docs)   # provider raises internally -> fallback path
print(docs)
# [{'memory': 'alpha', 'score': 0.9, 'rerank_score': 0.0}]  <- input was mutated
```

The fix seems straightforward: copy each dict in the fallback loop before setting the score, same as the success path already does. Everything else about the fallback (original order, 0.0 sentinel, top_k slice) stays as is.

I took a stab at a fix too (#6363). put it together with help from Claude Code and ran the test suite locally, but I'm new to this codebase so totally fine if you'd rather take a different approach. mostly trying to learn here :)

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6362.

## Why
Fixes #6362

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6362.py
+++ b/tests/test_issue_6362.py
@@ -1,0 +1,59 @@
+"""Regression test for reranker fallback mutating input documents."""
+
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+from mem0.reranker.cohere_reranker import CohereReranker
+from mem0.reranker.huggingface_reranker import HuggingFaceReranker
+from mem0.reranker.sentence_transformer_reranker import SentenceTransformerReranker
+from mem0.reranker.zero_entropy_reranker import ZeroEntropyReranker
+
+
+def _cohere_reranker():
+    reranker = CohereReranker.__new__(CohereReranker)
+    reranker.config = SimpleNamespace(top_k=None, return_documents=False, max_chunks_per_doc=None)
+    reranker.model = "rerank-model"
+    reranker.client = MagicMock()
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

Issue #6279, opened 2026-07-13
Pinecone reset()/delete_col() ignore namespace — wipes entire shared index, destroying other tenants' data


okaditya84:
### Bug Description

`Pinecone.delete_col()` and `Pinecone.reset()` ignore the configured `namespace` and operate on the **entire index**, so calling `Memory.reset()` for one tenant destroys every other tenant's data sharing that same Pinecone index.

### Where

`mem0/vector_stores/pinecone.py`:

```python
def delete_col(self):
    """Delete an index/collection."""
    try:
        self.client.delete_index(self.collection_name)   # <- drops the WHOLE index
        ...

def reset(self):
    """Reset the index by deleting and recreating it."""
    logger.warning(f"Resetting index {self.collection_name}...")
    self.delete_col()
    self.create_col(self.embedding_model_dims, self.metric)
```

Every other method in this class correctly scopes to `namespace=self.namespace` — `insert` (`index.upsert(..., namespace=self.namespace)`), `search`/`list` (`index.query(..., namespace=self.namespace)`), `delete` (`index.delete(ids=..., namespace=self.namespace)`), `update` (`index.upsert(..., namespace=self.namespace)`), `get` (`index.fetch(..., namespace=self.namespace)`), and even `count()` reads `stats.namespaces[self.namespace]` when a namespace is set. `delete_col`/`reset` are the only two methods that operate index-wide with no namespace awareness at all.

Mem0's docs describe `namespace` as the mechanism for isolating tenants within a shared Pinecone index. Given that, `delete_col()`/`reset()` not respecting it means:

### Steps to Reproduce

```python
from mem0 import Memory

# Two tenants sharing one Pinecone index via different namespaces
m_a = Memory.from_config({"vector_store": {"provider": "pinecone", "config": {
    "collection_name": "shared-index", "namespace": "tenant_a", ...
}}})
m_b = Memory.from_config({"vector_store": {"provider": "pinecone", "config": {
    "collection_name": "shared-index", "namespace": "tenant_b", ...
}}})

m_a.add("...", user_id="alice")
m_b.add("...", user_id="bob")

m_a.reset()   # intended: wipe only tenant_a's namespace
# actual: the entire "shared-index" index is deleted and recreated empty —
# tenant_b's data (and every other namespace in this index) is gone too.
```

### Expected Behavior

[...]

Issue #5585, opened 2026-06-16
bug(weaviate): reset() crashes with missing vector_size argument


HrushiYadav:
## Description

Weaviate vector store's `reset()` method crashes because it calls the collection creation method without the required `vector_size` argument.

## Steps to reproduce

1. Use Weaviate as the vector store backend
2. Call `reset()`
3. Crashes with missing `vector_size` argument

PR: #5531

Issue #5584, opened 2026-06-16
bug(mongodb): reset() passes wrong argument to create_col()


HrushiYadav:
## Description

MongoDB vector store's `reset()` method passes the wrong argument to `create_col()`, causing the reset to fail when trying to recreate the collection.

## Steps to reproduce

1. Use MongoDB as the vector store backend
2. Call `reset()`
3. Fails with incorrect argument to `create_col()`

PR: #5532

PR #6412, opened 2026-07-19
fix(qdrant): actually clear points on reset() for local Qdrant

### LINKED ISSUE #6411: bug(vector-stores): Memory.reset() silently keeps all memories on local Qdrant (the default store)
### Component

Core / Python SDK

## Summary

`Memory.reset()` reports success but does not delete any memories when the vector store is a local (path-based) Qdrant — which is mem0's default OSS configuration. The history DB *is* cleared, so the two stores are left disagreeing with each other.

> **Correction (added post-merge, per [@kartik-mem0's review on #6412](https://github.com/mem0ai/mem0/pull/6412#issuecomment-5031282810)):** the root cause below is platform-conditional, not universal — see the corrected **Impact** section.

## Root cause

Local Qdrant keeps a collection's `storage.sqlite` on disk after `delete_collection()`. Recreating a collection with the same name re-adopts that file, so every point comes back.

`Qdrant.reset()` is exactly that drop-and-recreate pair, so it is a guaranteed no-op for data on local mode:

```python
def reset(self):
    """Reset the index by deleting and recreating it."""
    logger.warning(f"Resetting index {self.collection_name}...")
    self.delete_col()
    self.create_col(self.embedding_model_dims, self.on_disk)
```

Isolated against `qdrant-client` directly (no mem0 involved), showing the file surviving the delete and the point coming back:

```
[after upsert]
   collection/t/storage.sqlite  (12288b)
[after delete_collection]
   collection/t/storage.sqlite  (12288b)   <- still there
points after recreate: 1                   <- resurrected
```

This also survives a process restart: deleting the collection, closing the client, reopening, and recreating the same name still returns the old point.

## Steps to Reproduce

No API key needed — uses `infer=False` and the mock embedder.

```python
import os, shutil, tempfile

WORK = os.path.join(tempfile.gettempdir(), "mem0_reset_repro")
shutil.rmtree(WORK, ignore_errors=True)
os.environ["MEM0_DIR"] = WORK

from mem0 import Memory
from mem0.embeddings.mock import MockEmbeddings

m = Memory.from_config({
    "vector_store": {
        "provider": "qdrant",
        "config": {
            "collection_name": "repro",
            "embedding_model_dims": 10,
            "path": os.path.join(WORK, "qdrant"),   # local mode, as in the default config
        },
    },
    "history_db_path": os.path.join(WORK, "history.db"),
})
m.embedding_model = MockEmbeddings()

m.add("user prefers dark mode", user_id="alice", infer=False)
print("before reset:", m.get_all(filters={"user_id": "alice"})["results"])

m.reset()

print("after  reset:", m.get_all(filters={"user_id": "alice"})["results"])
```

Output:

```
Resetting all memories
Resetting index repro...
before reset: [{'id': '...', 'memory': 'user prefers dark mode', ...}]
after  reset: [{'id': '...', 'memory': 'user prefers dark mode', ...}]
```

## Expected Behavior

`reset()` deletes the stored memories.

## Actual Behavior

The memories survive. `reset()` logs `Resetting all memories` and returns normally, so callers have no signal that nothing was deleted.

Because `self.db.reset()` *does* succeed, the surviving memories are left with zero history rows — the vector store and the history DB now disagree, which is worse than a clean failure.

## Impact

- **Corrected:** `QdrantLocal.delete_collection()` never explicitly closes the sqlite handle before removing the collection directory — it relies on refcounting and swallows failures via `ignore_errors=True`. On POSIX filesystems (Linux, macOS) an unlink on an open file still succeeds, so drop/recreate genuinely clears the data there. This bug reproduces only where an open handle blocks unlink: **Windows, NFS, or a Docker bind mount from a Windows host** — not universally on the default OSS setup as originally written here.
- `reset()` is the "delete everything" operation, so a silent no-op here is a data-deletion expectation that is not met on the affected platforms above.

## Environment

- `mem0ai` 2.0.12 (`main` @ ddaa655)
- `qdrant-client` 1.18.0
- Python 3.12.13

## Possible fix

On local mode, delete the points explicitly instead of dropping and recreating the collection:

```python
if self.is_local:
    self.client.delete(
        collection_name=self.collection_name,
        points_selector=models.FilterSelector(filter=models.Filter(must=[])),
    )
    return
```

I verified this clears the data and that it stays cleared after reopening the client. Remote mode is untouched, since drop/recreate is correct there.

Happy to open a PR with this plus a regression test if the approach looks right.

### Related but distinct

Previous `reset()` bugs were in other backends or other code paths, and none of them cover local Qdrant: #5584 (mongodb), #5585 (weaviate), #6279 (pinecone namespace), #5915 (AsyncMemory).

freya0926:
Closes #6411

## Problem

`Memory.reset()` reports success but keeps every memory when the vector store is a local (path-based) Qdrant — mem0's default OSS configuration.

> **Correction (added post-merge, per [@kartik-mem0's review](https://github.com/mem0ai/mem0/pull/6412#issuecomment-5031282810)):** this is platform-conditional, not universal. `QdrantLocal.delete_collection()` relies on refcounting to release the sqlite handle and swallows failures with `ignore_errors=True`; on POSIX filesystems (Linux, macOS) an unlink on an open file still succeeds, so drop/recreate correctly clears the data there. The no-op only reproduces where an open handle blocks unlink: **Windows, NFS, or a Docker bind mount from a Windows host.**

Local Qdrant leaves a collection's `storage.sqlite` on disk after `delete_collection()` on those affected filesystems, and recreating a collection with the same name re-adopts that file. `Qdrant.reset()` was exactly that drop-and-recreate pair, so it never actually dropped the data there.

`Memory.reset()` *does* clear the history DB, so the surviving memories were left with zero history rows — the vector store and the history DB ended up disagreeing after a reset that looked like it worked.

## Fix

Delete the points explicitly when running against a local client. Remote mode keeps drop/recreate, which is correct there and is what the existing mocked tests cover.

```python
if self.is_local:
    self.client.delete(
        collection_name=self.collection_name,
        points_selector=models.FilterSelector(filter=models.Filter(must=[])),
    )
    return
```

> **Note:** the version above was the initial proposal. The merged version differs — see [@kartik-mem0's review comment](https://github.com/mem0ai/mem0/pull/6412#issuecomment-5031282810) for the final approach (drop/recreate first to preserve the v3 hybrid-search `bm25` schema slot, then a no-op-safe sweep) and the deterministic mock-based regression test that replaced the original filesystem-dependent one.

## Test

`test_reset_clears_points_on_local_qdrant` drives a real local `Qdrant` against a temp directory rather than a mock, because the bug only reproduces against the on-disk storage — the existing suite passes a `MagicMock` client, so `is_local` is `False` there and this path was never exercised.

The test closes the client in a `finally` before removing the temp dir, since local Qdrant holds the storage file open and Windows will not delete it otherwise.

Verified it fails without the fix and passes with it:

```
# without the source change
>           self.assertEqual(len(store.list(top_k=10)[0]), 0)
E           AssertionError: 1 != 0

# with the source change
96 passed in 1.40s
```

Wider suite is unchanged: `442 passed, 48 errors` both before and after, where the 48 collection errors are pre-existing and come from optional vector-store dependencies not being installed in my environment.

`ruff check` passes on both touched files. I deliberately did not run `ruff format` — both files already fail `ruff format --check` on a clean `main`, so formatting them would have buried this change in unrelated diff noise.

## Notes

Verified against `qdrant-client` 1.18.0 on Python 3.12.13.

Previous `reset()` fixes covered other backends and other code paths, so this does not overlap them: #5584 (mongodb), #5585 (weaviate), #6279 (pinecone namespace), #5915 (AsyncMemory).

### DIFF
--- a/mem0/vector_stores/qdrant.py
+++ b/mem0/vector_stores/qdrant.py
@@ -594,15 +594,13 @@ class Qdrant(VectorStoreBase):
     def reset(self):
         """Reset the index by deleting and recreating it."""
         logger.warning(f"Resetting index {self.collection_name}...")
+        self.delete_col()
+        self.create_col(self.embedding_model_dims, self.on_disk)
         if self.is_local:
-            # Local (path-based) Qdrant keeps a collection's storage.sqlite on
-            # disk after delete_collection(). Recreating a collection with the
-            # same name re-adopts that file, so drop/recreate leaves every point
-            # intact. Delete the points explicitly instead.
+            # Local delete_collection() rmtree's with ignore_errors=True and leaves its
+            # sqlite handle open, so where an open file blocks unlink (Windows, NFS) the
+            # recreated collection re-adopts the old storage.sqlite. Drop what survived.
             self.client.delete(
                 collection_name=self.collection_name,
                 points_selector=models.FilterSelector(filter=models.Filter(must=[])),
             )
-            return
-        self.delete_col()
-        self.create_col(self.embedding_model_dims, self.on_disk)
--- a/mem0/vector_stores/qdrant.py
+++ b/mem0/vector_stores/qdrant.py
@@ -594,5 +594,15 @@ class Qdrant(VectorStoreBase):
     def reset(self):
         """Reset the index by deleting and recreating it."""
         logger.warning(f"Resetting index {self.collection_name}...")
+        if self.is_local:
+            # Local (path-based) Qdrant keeps a collection's storage.sqlite on
+            # disk after delete_collection(). Recreating a collection with the
+            # same name re-adopts that file, so drop/recreate leaves every point
+            # intact. Delete the points explicitly instead.
+            self.client.delete(
+                collection_name=self.collection_name,
+                points_selector=models.FilterSelector(filter=models.Filter(must=[])),
+            )
+            return
         self.delete_col()
         self.create_col(self.embedding_model_dims, self.on_disk)
--- a/tests/vector_stores/test_qdrant.py
+++ b/tests/vector_stores/test_qdrant.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import tempfile
 import unittest
 import uuid
@@ -53,6 +54,36 @@ class TestQdrant(unittest.TestCase):
                     on_disk=False,
                 )
             self.assertTrue(os.path.isfile(sentinel))
+
+    def test_reset_clears_points_on_local_qdrant(self):
+        """reset() must actually drop the data on a local (path-based) Qdrant.
+
+        Local Qdrant keeps a collection's storage.sqlite on disk after
+        delete_collection(); recreating the same collection name re-adopts that
+        file, so a drop/recreate reset leaves every point in place. This uses a
+        real local client rather than a mock because the bug only reproduces
+        against the on-disk storage.
+        """
+        tmp = tempfile.mkdtemp()
+        store = Qdrant(collection_name="reset_me", embedding_model_dims=4, path=os.path.join(tmp, "qdrant"))
+        try:
+            self.assertTrue(store.is_local)
+
+            store.insert(
+                vectors=[[0.1, 0.2, 0.3, 0.4]],
+                payloads=[{"data": "remember me"}],
+                ids=[str(uuid.uuid4())],
+            )
+            self.assertEqual(len(store.list(top_k=10)[0]), 1)
+
+            store.reset()
+
+            self.assertEqual(len(store.list(top_k=10)[0]), 0)
+        finally:
+            # Local Qdrant holds the storage file open; release it before cleanup
+            # so the temp dir can be removed on Windows.
+            store.client.close()
+            shutil.rmtree(tmp, ignore_errors=True)
 
     def test_create_col(self):
         self.client_mock.get_collections.return_value = MagicMock(collections=[])
--- a/tests/vector_stores/test_qdrant.py
+++ b/tests/vector_stores/test_qdrant.py
@@ -1,5 +1,4 @@
 import os
-import shutil
 import tempfile
 import unittest
 import uuid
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Thanks for this, the bug is real and the fix mechanism is correct. I verified both before touching anything. Three things needed correcting though, and since `maintainerCanModify` was on I pushed them straight to your branch as 92eaa6af rather than round-tripping.

### 1. The root cause is platform-conditional, not universal

Local Qdrant does not inherently keep `storage.sqlite` after `delete_collection()`. `QdrantLocal.delete_collection()` does:

```python
_collection = self.collections.pop(collection_name, None)
del _collection
collection_path = self._collection_path(collection_name)
if collection_path is not None:
    shutil.rmtree(collection_path, ignore_errors=True)
```

It never calls `_collection.close()`, so it leans on refcounting to release the sqlite handle, and `ignore_errors=True` swallows the failure when that does not happen in time. On POSIX, unlink on an open file succeeds, the directory goes away, and drop/recreate genuinely clears the data.

The failure needs a filesystem where an open handle blocks unlink: Windows, NFS, or a Docker bind mount from a Windows host. I could not reproduce on macOS with either qdrant-client 1.17.1 or 1.18.0 (directory removed, zero points after recreate). Were you on Windows? That would explain it exactly.

### 2. The regression test passed against unfixed source

This is the important one. I ran your test verbatim against `main` with no source change:

```
test_reset_clears_points_on_local_qdrant ... ok
```

CI runs ubuntu, so it would have been green with or without the fix, and it guarded nothing. The fix is to make the failure deterministic instead of depending on the host filesystem:

```python
with patch("qdrant_client.local.qdrant_local.shutil.rmtree"):
    store.reset()
```

It now fails without the fix on every platform. Mutation-tested: stripping the `if self.is_local:` block gives `AssertionError: 1 != 0`, restoring it passes.

### 3. The early `return` dropped the schema refresh

Returning before `create_col()` means a local collection predating v3 hybrid search never regains its `bm25` sparse slot on reset. Measured against a collection created without the sparse config:

```
early return:   after reset -> _has_bm25_slot=False
drop/recreate:  after reset -> _has_bm25_slot=True
```

`reset()` is the escape hatch that `create_col()`'s own warning points at ("To enable hybrid search, use a fresh collection"), so this quietly removed it. Reordering keeps both properties: drop and recreate first, then sweep whatever survived. The sweep is a no-op when the rmtree worked, so POSIX pays one empty delete call.

Net result is 9 lines shorter than the previous version (27 insertions, 36 deletions). Suite is `96 passed`, `ruff check` clean.

### One ask

Could you update the PR description, and #6411, to the actual cause and impact? Both currently say the default OSS setup is affected, which is not true on Linux or macOS. I would rather not leave a wrong root cause in the history for whoever touches `reset()` next.

kartik-mem0 (approved):

freya0926:
Thanks for catching all three of these, @kartik-mem0 — especially #2, that's a much better test than what I had. Appreciate you pushing the fixes straight to the branch rather than round-tripping.

Updated the PR description and #6411 to the corrected, platform-conditional root cause (Windows/NFS/Docker-bind-from-Windows only, not POSIX) — sorry for the delay on this, somehow missed the notification.

PR #6413, opened 2026-07-19
fix(memory): prevent silent data loss on pgvector embedding dim mismatch (salvage of #4990)

### LINKED ISSUE #4985: bug(server): switching embedding provider silently drops writes due to vector dimension mismatch

Bartok9:
## Summary

Salvages #4990 by @lost-particles onto current main.

Closes #4985.

When the embedder is switched (or never matches the hard-coded 1536 default), pgvector tables and configs diverge. Writes can appear to succeed while nothing is stored. This makes dimension handling fail loud and auto-infers dims from the embedder.

## Motivation

- `PGVectorConfig.embedding_model_dims` defaulted to `1536` even for non-OpenAI embedders.
- `Memory` / `AsyncMemory` never copied embedder `embedding_dims` into the vector store config.
- Existing tables were not checked for vector dimension mismatch.
- V3 batch `insert` failures were logged only, so apps lost memory silently.

## Changes from original (#4990)

- Rebased onto current `main`.
- Schema validation lives in `_ensure_collection()` (current lazy ensure path) rather than a removed eager `__init__` list/create block.
- Shared `_sync_pgvector_dims` helper + lean unit tests.

## Fix

1. Default `embedding_model_dims` to `None`; infer from embedder config at init for `provider == "pgvector"`.
2. Raise `ValueError` on explicit config mismatch with embedder dims.
3. Validate existing table `vector` atttypmod against configured dims.
4. After per-row insert fallback, raise `RuntimeError` if any IDs still failed.

## Tests

```bash
pytest tests/memory/test_main.py::test_sync_pgvector_dims_infers_from_embedder \
  tests/memory/test_main.py::test_sync_pgvector_dims_raises_on_explicit_mismatch \
  tests/memory/test_main.py::test_sync_pgvector_dims_noop_for_other_providers \
  tests/memory/test_main.py::test_pgvector_validate_schema_dims_mismatch \
  tests/memory/test_main.py::test_pgvector_validate_schema_dims_match \
  tests/memory/test_main.py::test_pgvector_ensure_collection_requires_dims -q
```

6 passed.

## Credit

Original approach and analysis: @lost-particles in #4990. This salvage rebases and adapts for `_ensure_collection`.

### DIFF
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel, ConfigDict, Field, model_validator
 class PGVectorConfig(BaseModel):
     dbname: str = Field("postgres", description="Default name for the database")
     collection_name: str = Field("mem0", description="Default name for the collection")
-    embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
+    embedding_model_dims: Optional[int] = Field(None, description="Dimensions of the embedding model. When None, inferred from the configured embedder at Memory init time.")
     user: Optional[str] = Field(None, description="Database user")
     password: Optional[str] = Field(None, description="Database password")
     host: Optional[str] = Field(None, description="Database host. Default is localhost")
--- a/mem0/configs/vector_stores/pgvector.py
+++ b/mem0/configs/vector_stores/pgvector.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel, ConfigDict, Field, model_validator
 class PGVectorConfig(BaseModel):
     dbname: str = Field("postgres", description="Default name for the database")
     collection_name: str = Field("mem0", description="Default name for the collection")
-    embedding_model_dims: Optional[int] = Field(None, description="Dimensions of the embedding model. When None, inferred from the configured embedder at Memory init time.")
+    embedding_model_dims: Optional[int] = Field(None, description="Dimensions of the embedding model. BREAKING: default changed from 1536 to None; when unset, inferred from the resolved embedder at Memory init. Set explicitly if you rely on a fixed width.")
     user: Optional[str] = Field(None, description="Database user")
     password: Optional[str] = Field(None, description="Database password")
     host: Optional[str] = Field(None, description="Database host. Default is localhost")
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -82,6 +82,61 @@ warnings.filterwarnings("ignore", category=DeprecationWarning, message=".*swigva
 logger = logging.getLogger(__name__)
 
 
+def _embedder_embedding_dims(embedder_config) -> Optional[int]:
+    """Best-effort read of embedding_dims from an EmbedderConfig, BaseEmbedderConfig, or dict."""
+    if embedder_config is None:
+        return None
+    if isinstance(embedder_config, dict):
+        dims = embedder_config.get("embedding_dims")
+        return int(dims) if dims is not None else None
+    dims = getattr(embedder_config, "embedding_dims", None)
+    return int(dims) if dims is not None else None
+
+
+def _sync_pgvector_dims(config: "MemoryConfig") -> None:
+    """Align pgvector's embedding_model_dims with the configured embedder.
+
+    Rules (applied when provider == "pgvector"):
+    - If embedding_model_dims is None → infer it from the embedder's embedding_dims.
+    - If embedding_model_dims is explicitly set and differs from the embedder's
+      embedding_dims → raise ValueError so the mismatch is caught at construction
+      time rather than silently at write time.
+    - If embedding_dims cannot be determined from the embedder, do nothing (pgvector
+      itself will raise a clear error when it tries to create/use the table).
+    """
+    if getattr(config.vector_store, "provider", None) != "pgvector":
+        return
+
+    embed_dims = _embedder_embedding_dims(config.embedder.config)
+    if embed_dims is None:
+        return
+
+    vs_config = config.vector_store.config
+    configured_dims = getattr(vs_config, "embedding_model_dims", None)
+    if isinstance(vs_config, dict):
+        configured_dims = vs_config.get("embedding_model_dims")
+
+    if configured_dims is None:
+        if hasattr(vs_config, "embedding_model_dims"):
+            vs_config.embedding_model_dims = embed_dims
+        elif isinstance(vs_config, dict):
+            vs_config["embedding_model_dims"] = embed_dims
+    elif int(configured_dims) != int(embed_dims):
+        collection = getattr(vs_config, "collection_name", None)
+        if isinstance(vs_config, dict):
+            collection = vs_config.get("collection_name")
+        raise ValueError(
+            f"Embedding dimension mismatch: the configured embedder "
+            f"({config.embedder.provider}) produces {embed_dims}-dimensional vectors, "
+            f"but pgvector collection '{collection}' is configured with "
+            f"embedding_model_dims={configured_dims}. "
+            f"Either remove the explicit embedding_model_dims setting to auto-infer from "
+            f"the embedder, or set it to {embed_dims} to match the current embedder."
+        )
+
+
+
+
 def _vector_store_list_rows(listed):
     if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list):
         return listed[0]
@@ -488,6 +543,10 @@ class Memory(MemoryBase):
             self.config.embedder.config,
             self.config.vector_store.config,
         )
+
+        if self.config.vector_store.provider == "pgvector":
+            _sync_pgvector_dims(self.config)
+
         self.vector_store = VectorStoreFactory.create(
             self.config.vector_store.provider, self.config.vector_store.config
         )
@@ -1044,12 +1103,21 @@ class Memory(MemoryBase):
                 payloads=all_payloads,
             )
         except Exception:
-            # Fallback: insert one by one
+            # Batch failed — retry individually so partial failures are surfaced
+            failures = []
             for mid, vec, pay in zip(all_ids, all_vectors, all_payloads):
                 try:
                     self.vector_store.insert(vectors=[vec], ids=[mid], payloads=[pay])
                 except Exception as e:
                     logger.error(f"Failed to insert memory {mid}: {e}")
+                    failures.append((mid, e))
+            if failures:
+                first_id, first_err = failures[0]
+                raise RuntimeError(
+                    f"{len(failures)} of {len(all_ids)} memor{'y' if len(failures) == 1 else 'ies'} "
+                    f"failed to persist to the vector store. "
+                    f"First failure (memory_id={first_id}): {first_err}"
+                ) from first_err
 
         # Batch history
         history_records = [
@@ -2162,6 +2230,10 @@ class AsyncMemory(MemoryBase):
             self.config.embedder.config,
             self.config.vector_store.config,
         )
+
+        if self.config.vector_store.provider == "pgvector":
+            _sync_pgvector_dims(self.config)
+
         self.vector_store = VectorStoreFactory.create(
             self.config.vector_store.provider, self.config.vector_store.config
         )
@@ -2691,11 +2763,21 @@ class AsyncMemory(MemoryBase):
                 payloads=all_payloads,
             )
         except Exception:
+            # Batch failed — retry individually so partial failures are surfaced
+            failures = []
             for mid, vec, pay in zip(all_ids, all_vectors, all_payloads):
                 try:
                     await asyncio.to_thread(self.vector_store.insert, vectors=[vec], ids=[mid], payloads=[pay])
                 except Exception as e:
                     logger.error(f"Failed to insert memory {mid} (async): {e}")
+                    failures.append((mid, e))
+            if failures:
+                first_id, first_err = failures[0]
+                raise RuntimeError(
+                    f"{len(failures)} of {len(all_ids)} memor{'y' if len(failures) == 1 else 'ies'} "
+                    f"failed to persist to the vector store. "
+                    f"First failure (memory_id={first_id}): {first_err}"
+                ) from first_err
 
         # Batch history
         history_records = [
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -109,9 +109,8 @@ def _sync_pgvector_dims(config: "MemoryConfig", embedder=None) -> None:
     Rules (applied when provider == "pgvector"):
     - If embedding_model_dims is None → infer it from the resolved embedder's
       embedding_dims (post EmbedderFactory.create defaults).
-    - If embedding_model_dims is explicitly set and differs from the embedder's
-      embedding_dims → raise ValueError so the mismatch is caught at construction
-      time rather than silently at write time.
+    - If embedding_model_dims is explicitly set, keep it (declared embedder dims
+      are advisory for some providers). Live enforcement is schema validation.
     - If embedding_dims cannot be determined from the embedder, do nothing (pgvector
       itself will raise a clear error when it tries to create/use the table).
     """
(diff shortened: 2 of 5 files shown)


kartik-mem0 (commented):
## Changes requested

Confirms and fixes the real silent-write-loss bug (#4985), verified live against `main` with a real Postgres/pgvector container and a real embedder (FAIL on `main`, PASS on this head, both sync and async). But the dims auto-inference reads the wrong config object, so it no-ops for any embedder relying on its own default, and the new hard default then crashes those same deployments on upgrade.

**To unblock:**
1. `mem0/memory/main.py:110` - infer dims from the resolved embedder, not the raw pre-construction config (detail inline)
2. `mem0/memory/main.py:1082` - raise `VectorStoreError`, and persist history/entity links for the subset that did succeed, before raising (detail inline)
3. `mem0/configs/vector_stores/pgvector.py:9` - flag the `1536` to `None` default change as breaking, and cover it in the docs (detail inline)

Optional: `_get_table_vector_dim` isn't schema-qualified, and a couple of unreachable branches in the new helpers could go (ponytail: net -12 lines possible).

<details>

- **Problem is real:** reproduces on `main` at `mem0/memory/main.py:1040` / `mem0/vector_stores/pgvector.py`: a pgvector table created at one embedder's width outlives a switch to a different embedder, and the fallback insert loop swallows the resulting `DataException` instead of re-raising. Verified live with a real Azure OpenAI embedder and a real Postgres/pgvector container (`sandbox/scenarios/issue4985_pgvector_dimension_mismatch.py`, lane `oss-pgvector`): FAIL on `main` @ `45208feeb` (sync 5.8s, async 5.0s), PASS on this PR's head @ `d723284d7` (sync 1.4s, async 1.5s).
- **Tests:** `pytest tests/memory/test_main.py -k "pgvector_dims or validate_schema_dims or ensure_collection_requires_dims"` -> 6 passed. Full file: 54 passed.
- **Red-green:** reverting the 3 impl files to `origin/main` (keeping the PR's test file) fails all 6 new tests, 2 with `AttributeError`; restoring the fix passes them again. Note: the tests stub `embedder.config` as an already-resolved object, never a real `EmbedderFactory`-built one, which is exactly why they miss the Critical finding below.
- **CI:** green on all Python SDK jobs (3.10/3.11/3.12) plus CI Gate and license/cla. Vercel is red (deploy-preview auth gate for an external-contributor fork), unrelated to code, does not block.
- **Security:** cleared, no user content or tenancy touched, only operator config and Postgres system-catalog metadata; the new schema-width query is parameterized.
- **Bloat (ponytail pass):** net -12 lines possible, dead `getattr`/`isinstance(dict)` branches in the two new helpers that can't be reached given how `EmbedderConfig`/`VectorStoreConfig` are actually constructed.
- **Not verified:** other vector stores don't implement this validation pattern (out of this PR's stated scope); concurrent multi-process `_ensure_collection()` races were not exercised.
- **Linked issue:** #4985

</details>

The re-raise on residual insert failure and the schema-drift check are both the right shape for this bug class. Push once the dims source is fixed and I'll re-review.

mem0/memory/main.py
```diff
@@ -82,6 +82,61 @@
...
+    - If embedding_model_dims is explicitly set and differs from the embedder's
+      embedding_dims → raise ValueError so the mismatch is caught at construction
+      time rather than silently at write time.
+    - If embedding_dims cannot be determined from the embedder, do nothing (pgvector
+      itself will raise a clear error when it tries to create/use the table).
+    """
+    if getattr(config.vector_store, "provider", None) != "pgvector":
+        return
+
+    embed_dims = _embedder_embedding_dims(config.embedder.config)
```
kartik-mem0: This reads the raw pre-construction `config.embedder.config`, not the resolved object `EmbedderFactory.create()` builds a few lines below at the call site. It stays `None` whenever an embedder relies on its own implicit default (OpenAI, Gemini, Azure, Bedrock) instead of an explicit `embedding_dims`, which includes #4985's own Gemini repro. Combined with the config default flipping to `None` and the hard raise at `mem0/vector_stores/pgvector.py`'s `_ensure_collection`, that's not a silent no-op anymore, it's a hard crash on the first `add()`/`search()` after upgrading for the common case.

Read `self.embedding_model.config.embedding_dims` after `EmbedderFactory.create()` instead (it's already built one line above where `_sync_pgvector_dims` is called).

mem0/memory/main.py
```diff
@@ -1010,12 +1069,21 @@ def _add_to_vector_store(self, messages, metadata, filters, infer, prompt=None):
...
+            failures = []
             for mid, vec, pay in zip(all_ids, all_vectors, all_payloads):
                 try:
                     self.vector_store.insert(vectors=[vec], ids=[mid], payloads=[pay])
                 except Exception as e:
                     logger.error(f"Failed to insert memory {mid}: {e}")
+                    failures.append((mid, e))
+            if failures:
+                first_id, first_err = failures[0]
+                raise RuntimeError(
```
kartik-mem0: Two issues on this raise (same pattern repeats in the async twin around line 2730):

1. It fires after the loop finishes, but before the batch-history write and entity-linking below, so memories that did persist in this same batch end up with no history record or entity links even though the caller sees an exception and would reasonably assume nothing landed.
2. It's a bare `RuntimeError`, not `VectorStoreError`, which is what this module already raises for vector-store failures elsewhere and what callers would reasonably catch.

Persist history/links for the subset that succeeded before raising, and raise `VectorStoreError`.

mem0/configs/vector_stores/pgvector.py
```diff
@@ -6,7 +6,7 @@
 class PGVectorConfig(BaseModel):
     dbname: str = Field("postgres", description="Default name for the database")
     collection_name: str = Field("mem0", description="Default name for the collection")
-    embedding_model_dims: Optional[int] = Field(1536, description="Dimensions of the embedding model")
+    embedding_model_dims: Optional[int] = Field(None, description="Dimensions of the embedding model. When None, inferred from the configured embedder at Memory init time.")
```
kartik-mem0: Flipping the default from `1536` to `None` is backward-incompatible on its own: any pgvector user who never set `embedding_model_dims` explicitly (previously fine via the old `1536` default) now depends entirely on `_sync_pgvector_dims` correctly inferring it, and given the bug at `main.py:110`, that inference silently fails for most embedder configs, so `_ensure_collection` raises on their first write after upgrading.

Worth calling out as a breaking change in the PR description and the pgvector docs page, independent of the inference fix.

Bartok9:
Thanks @kartik-mem0 — addressed the three blockers on tip `db7bab6a`:

1. **Dims source** — `_sync_pgvector_dims(config, self.embedding_model)` now reads dims from the **resolved** embedder after `EmbedderFactory.create()` (both sync + async `Memory` init). Tests stub the resolved object so implicit embedder defaults are covered.
2. **Partial insert failures** — on residual insert failure we filter to the successful subset, run history + entity linking for that subset, then raise `VectorStoreError` (sync + async).
3. **Breaking default** — called out on `PGVectorConfig.embedding_model_dims` and in `docs/components/vectordbs/dbs/pgvector.mdx` (`1536` → `None`, infer from resolved embedder).

Optional: schema-qualified `_get_table_vector_dim` via `current_schema()`.

Ready for re-review.

mem0/memory/main.py
kartik-mem0: At `d19dbe53f` this is fixed. `mem0/memory/main.py:551` now calls `_sync_pgvector_dims(self.config, self.embedding_model)` with the already-constructed embedder (async mirror at `~2160`), confirmed `self.embedding_model` is built before `self.vector_store` in both `__init__`s so it reads resolved provider defaults, not the raw pre-construction config. Verified with a red-green pass: reverting only the 3 impl files to the pre-delta commit fails the 3 updated tests with `TypeError: _sync_pgvector_dims() takes 1 positional argument but 2 were given`; restoring the delta passes them.

mem0/memory/main.py
kartik-mem0: Both points fixed at `d19dbe53f`. `mem0/memory/main.py:1263-1284` (async mirror `~2944-2965`) now raises `VectorStoreError` after Phase 7 entity linking, with `records`/history/entity-linking built from the failure-filtered subset, so the succeeded memories keep their history and entity links. Confirmed by reading both implementations line by line.

One new issue surfaced by this same raise, not present in the original finding: it fires before `self.db.save_messages(...)` at line 1286 (async: raise at 2949, save at 2967), with no try/finally anywhere in the method or in `add()`'s caller. Any partial-failure batch, including the all-fail dims-mismatch case this PR targets, now skips persisting that turn's message history. Posting this as a new Critical finding in the review below.

mem0/configs/vector_stores/pgvector.py
kartik-mem0: Fixed at `d19dbe53f`. `docs/components/vectordbs/dbs/pgvector.mdx` carries a `<Warning>` block on the `1536 -> None` default change, and `mem0/configs/vector_stores/pgvector.py`'s `Field(description=...)` states it inline too.

mem0/memory/main.py
```diff
@@ -82,6 +82,74 @@
...
+        if hasattr(vs_config, "embedding_model_dims"):
+            vs_config.embedding_model_dims = embed_dims
+        elif isinstance(vs_config, dict):
+            vs_config["embedding_model_dims"] = embed_dims
+    elif int(configured_dims) != int(embed_dims):
+        collection = getattr(vs_config, "collection_name", None)
+        if isinstance(vs_config, dict):
+            collection = vs_config.get("collection_name")
+        provider = getattr(config.embedder, "provider", None)
+        raise ValueError(
```
kartik-mem0: This raises based on `_embedder_embedding_dims`, which reads the embedder's *declared* `config.embedding_dims`, not its real API output width. That's wrong for `azure_openai` (`embed()`/`embed_batch()` never pass `dimensions=` to the API, so the field is purely advisory), `ollama.py:17` (`embedding_dims = embedding_dims or 512`, static fallback unrelated to the loaded model), and `lmstudio.py:14` (same, 1536). Concretely: mem0's own docs tell Ollama users to set `embedding_model_dims=768` for the default `nomic-embed-text` model, while this fallback says 512 - so following mem0's own documentation now triggers this raise on a correct config, with no override. Sandbox repro at this head: `sandbox/scenarios/pr6493_explicit_dims_clobbered.py`, PASS on `main`, `ValueError` here on this head.

Either drop this pre-flight raise-on-mismatch and let `_validate_schema_dims` (the real, DB-truth-based check already added in this PR) be the sole enforcement point, or make this one advisory (log, don't raise) since its input is known-unreliable for several shipped providers.

mem0/memory/main.py
```diff
@@ -1179,6 +1263,25 @@ def _add_to_vector_store(self, messages, metadata, filters, infer, prompt=None):
         except Exception as e:
             logger.warning(f"Batch entity linking failed: {e}")
 
+        if _partial_insert_failures:
+            first_id, first_err = _partial_insert_failures[0]
+            raise VectorStoreError(
```
kartik-mem0, line 1276: This raise sits before `self.db.save_messages(messages, session_scope)` at line 1286 (async mirror: raise at 2949, save at 2967). Neither the method nor `add()`'s call site wraps this in try/finally, so any partial-insert-failure batch, including the all-fail case this PR's own fix targets, now skips persisting that turn's message history entirely. That history feeds `last_messages` context for the next call's extraction prompt, so this is a new instance of the exact silent-state-loss class #4985 describes, just relocated from vector rows to message history.

Move the `save_messages` call ahead of this raise, or wrap it so it always runs regardless of partial insert outcome.

kartik-mem0 (commented):
## Changes requested (re-review)

The 3 prior findings are fixed and verified at this head (replied in-thread on each). But this round's sandbox testing surfaced 2 new Critical findings that predate this delta but weren't caught last round.

**To unblock:**
1. `mem0/memory/main.py:141` - the explicit-mismatch check trusts the embedder's *declared* dims, not its real output width; this is wrong for at least azure_openai/ollama/lmstudio (detail inline)
2. `mem0/memory/main.py:1268` - the raise now fires before `save_messages`, dropping message history on any partial-failure batch (detail inline)

<details>

- **Prior findings:** verified against `d19dbe53f` by reading `mem0/memory/main.py` directly (embedder resolution order, raise placement, VectorStoreError type) and with a red-green pass: reverting the 3 impl files to the pre-delta commit fails the 3 updated tests with `TypeError: _sync_pgvector_dims() takes 1 positional argument but 2 were given`.
- **Gate B-live, same scenario as round 1, re-run at this head:** `sandbox/scenarios/issue4985_pgvector_dimension_mismatch.py`, lane `oss-pgvector`. FAIL on `main` @ `b54710a3c` (sync 5.5s, async 5.1s), PASS on this head @ `d19dbe53f` (sync 1.4s, async 1.5s).
- **New finding, live repro:** `sandbox/scenarios/pr6493_explicit_dims_clobbered.py` configures an embedder whose declared dims (512) understate its real output (1536), with the user explicitly setting `embedding_model_dims=1536` to compensate (the documented #4985 workaround). PASS on `main` @ `b54710a3c`. On this head @ `d19dbe53f`: `ValueError: Embedding dimension mismatch: the configured embedder (azure_openai) produces 512-dimensional vectors, but pgvector collection '...' is configured with embedding_model_dims=1536` at `Memory.from_config()`. A previously-working, doc-sanctioned config now hard-crashes at construction.
- **Tests:** `pytest tests/memory/test_main.py -k pgvector` -> 6 passed. Full file: 58 passed. `ruff check` clean.
- **isort:** fails (`from mem0.exceptions import LLMError, VectorStoreError` needs splitting around the intervening `ValidationError as Mem0ValidationError` import; several new test functions import stdlib after first-party). `make lint`/CI won't catch it (ruff-only, no `I` rule), but `pre-commit run --all-files` would reject it.
- **Cross-package:** the same silent-swallow bug is unfixed in `mem0-ts/src/oss/src/memory/index.ts`/`vector_stores/pgvector.ts` (log-only fallback, no dims inference). Doesn't block this Python-only PR; worth a companion TS issue.
- **Related open PR:** #6493 targets the same root issue #4985 with an unconditional dims-propagation across qdrant/milvus/pinecone/redis/weaviate/faiss, but reproduces the same declared-vs-actual bug class (confirmed live this round). Neither PR subsumes the other.
- **Linked issue:** #4985

</details>

The schema-drift check in `mem0/vector_stores/pgvector.py` is the right, ground-truth fix. The pre-flight check in `main.py` should defer to it rather than gate on the embedder's declared config.

Bartok9:
Thanks @kartik-mem0 — both re-review blockers addressed on tip `99e50c19`:

1. **Declared-vs-actual dims** — `_sync_pgvector_dims` no longer `ValueError`s when explicit `embedding_model_dims` differs from the embedder’s declared `embedding_dims`. Explicit config is kept (documented azure_openai/ollama/lmstudio workaround); we only warn. Ground-truth enforcement stays in `_validate_schema_dims` against the live table.
2. **`save_messages` ordering** — sync + async `_add_to_vector_store` now persist message history *before* raising `VectorStoreError` on residual partial inserts, so `last_messages` context is not dropped for that turn.

Tests: updated mismatch case to assert keep-explicit + warning; `pytest -k 'pgvector_dims or validate_schema_dims or ensure_collection_requires_dims or keeps_explicit'` → 6 passed.

Ready for re-review.

PR #6414, opened 2026-07-19
fix(async): serialize AsyncMemory vector-store writes (salvage of #4893)

### LINKED ISSUE #4892: Bug: concurrent AsyncMemory writes corrupt Qdrant HNSW index ('index N is out of bounds')

Bartok9:
## Summary

Salvages #4893 by @MattGyver onto current main.

Closes #4892.

Concurrent `AsyncMemory` writes can interleave vector-store mutations and corrupt embedded Qdrant HNSW indexes. This serializes insert/update/delete behind an `asyncio.Lock`.

## Motivation

Issue #4892 reports HNSW corruption (`index N is out of bounds`) under concurrent async writes. Sync `Memory` is single-threaded; async was not.

## Changes from original (#4893)

- Rebased onto current `main`.
- Lock retained around batch insert, `_create_memory`, `_update_memory`, and `_delete_memory` vector ops.
- Focused unit tests that don't depend on heavy fact-extraction mocks.

## Fix

1. `self._write_lock = asyncio.Lock()` on `AsyncMemory.__init__`.
2. `async with self._write_lock` around all async `vector_store.insert` / `update` / `delete` call sites used by the write path.

## Tests

```bash
pytest tests/memory/test_main.py::test_async_memory_write_lock_serializes_insert \
  tests/memory/test_main.py::test_async_memory_init_creates_write_lock -q
```

2 passed.

## Credit

Original approach: @MattGyver in #4893.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -2137,6 +2137,14 @@ class AsyncMemory(MemoryBase):
         self.api_version = self.config.version
         self.custom_instructions = self.config.custom_instructions
         self._entity_store = None
+        # Serialize concurrent vector-store writes to prevent HNSW index
+        # corruption when multiple add()/update()/delete() calls are in flight
+        # simultaneously (e.g. in an async agent processing events in parallel).
+        # Note: Entity-store writes are not currently protected by this lock,
+        # as they target a separate collection. For multi-process deployments
+        # with high entity-linking concurrency, use server-mode Qdrant.
+        # Salvage of #4893 by @MattGyver for issue #4892.
+        self._write_lock = asyncio.Lock()
 
         # Initialize reranker if configured
         self.reranker = None
@@ -2649,19 +2657,20 @@ class AsyncMemory(MemoryBase):
         all_ids = [r[0] for r in records]
         all_payloads = [r[3] for r in records]
 
-        try:
-            await asyncio.to_thread(
-                self.vector_store.insert,
-                vectors=all_vectors,
-                ids=all_ids,
-                payloads=all_payloads,
-            )
-        except Exception:
-            for mid, vec, pay in zip(all_ids, all_vectors, all_payloads):
-                try:
-                    await asyncio.to_thread(self.vector_store.insert, vectors=[vec], ids=[mid], payloads=[pay])
-                except Exception as e:
-                    logger.error(f"Failed to insert memory {mid} (async): {e}")
+        async with self._write_lock:
+            try:
+                await asyncio.to_thread(
+                    self.vector_store.insert,
+                    vectors=all_vectors,
+                    ids=all_ids,
+                    payloads=all_payloads,
+                )
+            except Exception:
+                for mid, vec, pay in zip(all_ids, all_vectors, all_payloads):
+                    try:
+                        await asyncio.to_thread(self.vector_store.insert, vectors=[vec], ids=[mid], payloads=[pay])
+                    except Exception as e:
+                        logger.error(f"Failed to insert memory {mid} (async): {e}")
 
         # Batch history
         history_records = [
@@ -3572,12 +3581,13 @@ class AsyncMemory(MemoryBase):
         new_metadata["updated_at"] = new_metadata["created_at"]
         new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
 
-        await asyncio.to_thread(
-            self.vector_store.insert,
-            vectors=[embeddings],
-            ids=[memory_id],
-            payloads=[new_metadata],
-        )
+        async with self._write_lock:
+            await asyncio.to_thread(
+                self.vector_store.insert,
+                vectors=[embeddings],
+                ids=[memory_id],
+                payloads=[new_metadata],
+            )
 
         await asyncio.to_thread(
             self.db.add_history,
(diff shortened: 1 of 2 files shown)


MattGyver:
Thank you for salvaging this, @Bartok9 — genuinely appreciate you rebasing it onto current `main` and carrying it home. 🙏

Confirming the diagnosis still holds: the corruption comes from concurrent `AsyncMemory` writes interleaving vector-store mutations against the embedded Qdrant HNSW index (`index N is out of bounds`). Sync `Memory` is single-threaded so it never surfaced there; async had no serialization. Wrapping `insert`/`update`/`delete` behind a single `asyncio.Lock` across the write path (`_create_memory`, `_update_memory`, `_delete_memory`, and batch insert) is exactly the right scope — it serializes only the vector-store mutations without blocking reads or fact-extraction.

One note for anyone landing here later: the lock is per-instance, so it protects a single-process async workload. Multi-process deployments sharing one embedded Qdrant path still need external coordination (or a server-mode Qdrant).

Happy to see it back in motion. Thanks again for the credit and the careful rebase.

Bartok9:
Thanks so much, @MattGyver — and thank you for the original diagnosis; the salvage was easy because your root-cause work was already spot on. 🙏

Great call on the per-instance scope caveat. I'll add a short docstring note right at the lock so future readers know it protects a single-process async workload, and that multi-process deployments sharing one embedded Qdrant path still need external coordination (or server-mode Qdrant). Will push that shortly.

Bartok9:
Pushed — added the per-instance scope note right at the lock (commit d67e45f) so future readers know it protects a single-process async workload and that multi-process deployments sharing one embedded Qdrant path still need external coordination (or server-mode Qdrant). Thanks again for the diagnosis, @MattGyver. 🙏

ebarkhordar (commented):
The lock placement holds up on the real path: driving `AsyncMemory._create_memory` concurrently, `vector_store.insert` never overlaps at this head, and it overlaps 5-way once the four `async with self._write_lock` blocks are removed.

The two new tests do not detect that difference. `test_async_memory_write_lock_serializes_insert` takes `mem._write_lock` in the test body and calls `mem.vector_store.insert` itself, so what it exercises is `asyncio.Lock`, not `main.py`, and `test_async_memory_init_creates_write_lock` only asserts the attribute exists. Both pass with all four guards deleted (keeping the `__init__` line), at head `1bab432` in a clean `python:3.11-slim` container:

```
$ python -m pytest tests/memory/test_main.py::test_async_memory_write_lock_serializes_insert \
    tests/memory/test_main.py::test_async_memory_init_creates_write_lock -q
2 passed
# identical command after stripping the 4 `async with self._write_lock:` blocks from main.py
2 passed
```

CONTRIBUTING asks for tests that would fail without the change, and going through the real method gets there with the mocks you already have:

```python
mem = AsyncMemory()   # same patches as your test
await asyncio.gather(*[mem._create_memory(f"fact {i}", existing_embeddings={}) for i in range(5)])
assert active["max"] == 1, f"vector_store.insert overlapped: max concurrent={active['max']}"
```

That passes at this head and fails with `max concurrent=5` once the guards are gone. `_update_memory` and `_delete_memory` take the same shape.

Separate question, not a blocker: `reset()` calls `vector_store.delete_col` (line 3769) and then rebinds `self.vector_store` (line 3779), both outside the lock. Enumerating the insert/update/delete/delete_col sites on `vector_store` inside `AsyncMemory`, that is the only mutation your four guards leave uncovered. Is it deliberately out of scope, on the grounds that a reset racing a write is the caller's problem? I have not tried to reproduce anything there, so that part is a read of the code only.

Bartok9:
Thanks @ebarkhordar — you're right, and that's a sharp catch. The old `test_async_memory_write_lock_serializes_insert` took the lock in the test body and called `vector_store.insert` itself, so it exercised `asyncio.Lock`, not `main.py` — it passed with the guards deleted. Fixed in 3cae606.

The test now drives the real method exactly as you suggested:

```python
mem = AsyncMemory()  # same patches
await asyncio.gather(*[mem._create_memory(f"fact {i}", existing_embeddings={}) for i in range(5)])
assert active["max"] == 1, f"vector_store.insert overlapped: max concurrent={active['max']}"
```

Verified both directions locally: passes at head (max concurrent == 1), and fails with `max concurrent=5` once the four `async with self._write_lock` guards are removed — satisfying CONTRIBUTING's fail-without-the-change requirement.

On `reset()`: good read of the code, and I'm intentionally leaving it out of scope. It's a destructive full teardown that calls `delete_col` and then rebinds `self.vector_store` to a brand-new client — a write racing a reset is a caller-lifecycle problem (you're tearing down the store out from under in-flight writes), not something a per-write mutation lock can meaningfully serialize. Wrapping teardown+rebind in the same write lock would also risk deadlocks/ordering surprises with the entity-store reset. Happy to file a follow-up note if maintainers want it flagged, but I'd keep this PR scoped to the insert/update/delete write path.

PR #6426, opened 2026-07-19
fix(memory): extract_json must not treat a code fence inside a JSON value as an enclosing block

### LINKED ISSUE #6427: extract_json treats code fences inside JSON string values as enclosing blocks
### Which SDK / package

Python SDK (`mem0ai`)

### Describe the bug

`extract_json` checks an unanchored code-fence regex before looking for a JSON object. If valid JSON contains a markdown code fence inside a string value, the inner fence is mistaken for an enclosing response fence and the JSON is corrupted.

This affects vector-store metadata read paths that use `json.loads(extract_json(...))`; memories containing code snippets can be written successfully and then fail during read-back.

### Reproduction

```python
import json
from mem0.memory.utils import extract_json

payload = {"snippet": "```python
x = 1
```", "user_id": "u1"}
encoded = json.dumps(payload)

assert json.loads(extract_json(encoded)) == payload
```

On current `main`, `extract_json(encoded)` returns only `python\nx = 1\n`, causing `json.loads` to raise `JSONDecodeError`.

### Expected behavior

An input that already starts with a JSON object should be parsed as JSON rather than scanning its string contents for markdown fences.

lntutor:
## Linked Issue

Closes #6427

## Description

`extract_json` used an unanchored code-fence regex before brace detection, so a fence inside a valid JSON string value was mistaken for an enclosing response fence. This skips the fence heuristic when the input already begins with a JSON object, preserving metadata that contains code snippets.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] Added/updated unit tests
- [x] Tested manually
- [ ] No tests needed

`tests/test_chatty_llm_parsing.py` and `tests/memory/test_memory_utils.py`: 41 passed.

## Checklist

- [x] Code follows the project style guidelines
- [x] Self-review performed
- [x] Tests prove the fix
- [x] New and existing targeted tests pass locally
- [x] Documentation is not required for this internal parsing fix

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -135,7 +135,13 @@ def extract_json(text):
     If that also fails, returns the text as-is.
     """
     text = text.strip()
-    match = re.search(r"```(?:json)?\s*(.*?)\s*```", text, re.DOTALL)
+    # A response that already begins with '{' is a JSON object, so skip the
+    # code-fence heuristic. Otherwise a ``` sequence occurring inside a string
+    # value (e.g. a stored code snippet) would be mistaken for an enclosing
+    # code block and only its interior extracted, corrupting valid JSON.
+    match = None
+    if not text.startswith("{"):
+        match = re.search(r"```(?:json)?\s*(.*?)\s*```", text, re.DOTALL)
     if match:
         json_str = match.group(1)
     else:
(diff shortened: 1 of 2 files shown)


lntutor:
Current head `cab33714628d42603bd162d4bb6cb15cd8fe4e3b` is mergeable and the CLA check is green. The only failing status currently shown on this PR is the Mem0 Vercel authorization gate for preview deployment; there is no contributor-side code failure attached to the current head. If that preview is needed, a Mem0 team member can authorize it from the Vercel link on the PR.

PR #6433, opened 2026-07-19
fix(llms/anthropic): send sampling params for legacy claude-3-x models

### LINKED ISSUE #6432: Anthropic: temperature/top_p silently dropped for all legacy claude-3-x models
### Summary

`AnthropicLLM._enable_sampling_parameters()` silently disables `temperature` / `top_p` for every legacy `claude-3-x` model, so those parameters are never sent to the Anthropic API even when the user configures them.

### Where

`mem0/llms/anthropic.py`, `_enable_sampling_parameters()`:

```python
model_name = self.config.model.lower()
model_name_parts = model_name.rsplit("[", 1)[0].split("-")
_, family, major, minor, *_ = (*model_name_parts, "", "", "", "")
family = family or model_name
...
```

The parser reads `family` from position `[1]` of the `-`-split model id. That works for the current `claude-<family>-<major>-<minor>` naming (`claude-sonnet-4-6` → `family="sonnet"`), but every legacy `claude-3-x` id puts the version numbers **before** the family:

- `claude-3-5-sonnet-20240620` → `family="3"`
- `claude-3-7-sonnet-20250219` → `family="3"`
- `claude-3-5-haiku-20241022` → `family="3"`
- `claude-3-opus-20240229` → `family="3"`

`family` never matches `haiku`/`sonnet`/`opus`, so the method falls through to `return False` and sampling parameters are dropped.

### Impact

All Claude 3.x models fully support `temperature`/`top_p`, so this drops them silently. A user config like:

```python
Memory.from_config({"llm": {"provider": "anthropic",
  "config": {"model": "claude-3-5-sonnet-20240620", "temperature": 0.1}}})
```

never sends `temperature=0.1`; Anthropic applies its server-side default (`temperature=1.0`) instead, so mem0's fact-extraction / memory-update prompts run far less deterministically than configured. No error is raised. Claude 3.5 Sonnet, 3.7 Sonnet, and 3.5 Haiku are among the most commonly pinned Anthropic models, so the blast radius is wide.

This is a regression: before the sampling-capability gate was added, `temperature` was sent unconditionally for these models.

### Expected

`temperature`/`top_p` should be sent for legacy `claude-3-x` models (they support sampling), while the existing behavior for the newer naming — disabling params on the models that reject them (e.g. `claude-opus-4-8`, `claude-sonnet-5`) — is preserved.

### Fix

Detect the legacy `claude-3-*` naming (version digits before the family) and treat those models as sampling-capable before the family-position parsing runs. Small, self-contained change plus regression cases added to the existing `_enable_sampling_parameters` test table. PR to follow.

OfficialAbhinavSingh:
## Linked Issue

Closes #6432

## Description

`AnthropicLLM._enable_sampling_parameters()` parses the model family from position `[1]` of the `-`-split model id. That works for the current `claude-<family>-<major>-<minor>` naming (`claude-sonnet-4-6` → `family="sonnet"`), but every legacy `claude-3-x` id puts the version numbers **before** the family:

- `claude-3-5-sonnet-20240620` → `family="3"`
- `claude-3-7-sonnet-20250219` → `family="3"`
- `claude-3-5-haiku-20241022` → `family="3"`
- `claude-3-opus-20240229` → `family="3"`

`family` never matches `haiku`/`sonnet`/`opus`, so the method falls through to `return False` and `temperature`/`top_p` are silently dropped from the request — even though all Claude 3.x models support them. Anthropic then applies its server-side default (`temperature=1.0`), so mem0's fact-extraction / memory-update prompts run far less deterministically than the user configured, with no error raised. Claude 3.5 / 3.7 Sonnet and 3.5 Haiku are among the most commonly pinned Anthropic models.

This PR detects the legacy `claude-3-*` naming (version digits before the family) and returns `True` for those models before the family-position parsing runs. The existing behavior for the newer naming — disabling sampling params on the models that reject them (`claude-opus-4-7`+, `claude-sonnet-5`, etc.) — is unchanged.

```python
# Legacy "claude-3-*" ids (e.g. claude-3-5-sonnet, claude-3-opus) put the
# version numbers before the family name, so the family-position parsing
# below misreads them. They all support sampling parameters; the new
# "claude-<family>-<major>-<minor>" naming is handled after this.
if len(model_name_parts) > 1 and model_name_parts[1].isdigit():
    return True
```

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — restores sending `temperature`/`top_p` for Claude 3.x models, which they support; behavior for the newer models is unchanged.

## Test Coverage

Extended the existing `test_enable_sampling_parameters_for_current_models` parametrized table in `tests/llms/test_anthropic.py` with five legacy `claude-3-x` cases (`claude-3-5-sonnet-20240620`, `claude-3-7-sonnet-20250219`, `claude-3-5-haiku-20241022`, `claude-3-opus-20240229`, `claude-3-haiku-20240307`), all expecting `temperature` to be sent. Those five fail on `main` (the params are dropped) and pass with this change. Full `tests/llms/test_anthropic.py` suite (27 tests) passes; `ruff check` is clean on the changed source.

## Checklist

### DIFF
--- a/mem0/llms/anthropic.py
+++ b/mem0/llms/anthropic.py
@@ -52,6 +52,14 @@ class AnthropicLLM(LLMBase):
 
         model_name = self.config.model.lower()
         model_name_parts = model_name.rsplit("[", 1)[0].split("-")
+
+        # Legacy "claude-3-*" ids (e.g. claude-3-5-sonnet, claude-3-opus) put the
+        # version numbers before the family name, so the family-position parsing
+        # below misreads them. They all support sampling parameters; the new
+        # "claude-<family>-<major>-<minor>" naming is handled after this.
+        if len(model_name_parts) > 1 and model_name_parts[1].isdigit():
+            return True
+
         _, family, major, minor, *_ = (*model_name_parts, "", "", "", "")
 
         # Use model_name if only provided a family name, e.g., opus
--- a/tests/llms/test_anthropic.py
+++ b/tests/llms/test_anthropic.py
@@ -149,6 +149,13 @@ def test_base_config_conversion_does_not_send_both(mock_anthropic_client):
         ("claude-fable-5", False),
         ("claude-mythos-5", False),
         ("claude-mythos-preview", False),
+        # Legacy "claude-3-*" naming puts the version before the family; all of
+        # these models support sampling params and must not be disabled.
+        ("claude-3-5-sonnet-20240620", True),
+        ("claude-3-7-sonnet-20250219", True),
+        ("claude-3-5-haiku-20241022", True),
+        ("claude-3-opus-20240229", True),
+        ("claude-3-haiku-20240307", True),
     ],
 )
 def test_enable_sampling_parameters_for_current_models(mock_anthropic_client, model, expected):


kartik-mem0:
Thanks a lot for the careful writeup and the regression tests, the root-cause analysis here is genuinely good.

After checking Anthropic's current model deprecation schedule, though, every model this branch targets has already been retired on the Anthropic-operated API:

| Model | Retired |
|-------|---------|
| `claude-3-5-sonnet-20240620` | 2025-10-28 |
| `claude-3-opus-20240229` | 2026-01-05 |
| `claude-3-5-haiku-20241022` | 2026-02-19 |
| `claude-3-7-sonnet-20250219` | 2026-02-19 |
| `claude-3-haiku-20240307` | 2026-04-20 |

Requests to a retired model fail before `temperature` / `top_p` are ever evaluated, so on `api.anthropic.com` the new branch fixes a path that can no longer be exercised, and the "silently less deterministic" behavior described can't actually occur (the call returns an error instead). Because of that we don't think this is worth carrying in the codebase.

The one caveat: partner platforms (Amazon Bedrock, Google Vertex) keep their own retirement schedules and may still serve older claude-3 snapshots. mem0 has a dedicated `aws_bedrock` provider for that case, so the only remaining path through `AnthropicLLM` would be a proxy/gateway pointed at `ANTHROPIC_BASE_URL` that still serves a claude-3 model.

Going to close this for now. If you're hitting this against a live endpoint that still serves a claude-3 snapshot via `AnthropicLLM` (a specific proxy or gateway), please share the details and we'll happily reopen. Really appreciate the contribution.

PR #6439, opened 2026-07-20
fix(mem0): add test for issue #6368

### LINKED ISSUE #6368: bug(llms/aws_bedrock): Anthropic Converse parsing crashes when a reasoningContent block precedes the text block

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6368.

## Why
Fixes #6368

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6368.py
+++ b/tests/test_issue_6368.py
@@ -1,0 +1,35 @@
+from unittest.mock import MagicMock, patch
+
+from mem0.configs.llms.aws_bedrock import AWSBedrockConfig
+from mem0.llms.aws_bedrock import AWSBedrockLLM
+
+
+def test_issue_6368():
+    runtime_client = MagicMock()
+    bedrock_client = MagicMock()
+    bedrock_client.list_foundation_models.return_value = {"modelSummaries": []}
+
+    def mock_client(service, **kwargs):
+        if service == "bedrock-runtime":
+            return runtime_client
+        return bedrock_client
+
+    runtime_client.converse.return_value = {
+        "output": {
+            "message": {
+                "content": [
+                    {"reasoningContent": {"reasoningText": {"text": "Thinking through the request."}}},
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6440, opened 2026-07-20
fix(mem0): add test for issue #6346

### LINKED ISSUE #6346: Agent quick path exits 0 before the fresh plugin is MCP-ready
I tested current main at `739534c`, Node CLI `0.2.11`, from a clean home with no
`MEM0_API_KEY`, shell profile entry, or Claude settings entry.

Repro:

```bash
mem0 init --agent --agent-caller codex --json
```

The command exits `0` and correctly saves the key plus default user ID in a `0600`
`~/.mem0/config.json`.

Two parts of the advertised fresh-plugin path do not hold:

1. `--json` emits a plain-text claim reminder rather than a JSON success envelope.
2. `MEM0_API_KEY` remains unset, and no fresh shell or Claude environment entry is
   created. The bundled Codex MCP manifest reads `bearer_token_env_var =
   "MEM0_API_KEY"`, so CLI memory commands are configured while the fresh MCP path
   is not authenticated.

The current sync implementation explains the second result: it updates only
environment entries that already exist and names Codex and Cursor as out of scope.
The platform agent-signup docs make the narrower, accurate claim that the key is
saved to `~/.mem0/config.json`; the stronger export and immediate-MCP claim appears
in the plugin README.

I first reproduced the post-response path with a valid synthetic backend envelope
to avoid creating an account, then confirmed the same output and credential state
against the production signup endpoint on 2026-07-16. The saved key works: CLI
status reports connected, and a direct MCP initialize returns 200 with it. The same
MCP request without the missing environment credential returns 401 Authentication
required.

Smallest fix I can see:

- emit one structured success envelope with `api_key_saved`, `default_user_id`,
  `mcp_ready`, and a machine-actionable next step;
- remove the current-shell export claim, or add a supported persistent credential
  handoff per plugin host plus the required source or restart step;
- add a clean-home integration test that checks both JSON output and the first MCP
  authentication source.

I kept the redacted run output and can rerun a maintainer-supplied path or a fix.

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6346.

## Why
Fixes #6346

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6346.py
+++ b/tests/test_issue_6346.py
@@ -1,0 +1,111 @@
+"""Regression test for Agent Mode's fresh-plugin readiness report."""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import sys
+import threading
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from pathlib import Path
+
+import pytest
+
+
+def test_issue_6346(monkeypatch, tmp_path):
+    """A fresh Codex bootstrap must report its actual MCP authentication state."""
+    cli_source = Path(__file__).parents[1] / "cli" / "python" / "src"
+    monkeypatch.syspath_prepend(str(cli_source))
+    import mem0_cli
+
+    assert mem0_cli.__package__ == "mem0_cli"
+
+    class AgentModeHandler(BaseHTTPRequestHandler):
+        def do_POST(self):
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6441, opened 2026-07-20
fix(mem0): add test for issue #6277

### LINKED ISSUE #6277: update() lets caller metadata silently overwrite user_id/agent_id/run_id, breaking tenant isolation

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6277.

## Why
Fixes #6277

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6277.py
+++ b/tests/test_issue_6277.py
@@ -1,0 +1,86 @@
+import sys
+from unittest.mock import MagicMock, patch
+
+sys.modules.setdefault("posthog", MagicMock())
+sys.modules.setdefault("qdrant_client", MagicMock())
+
+with patch("importlib.metadata.version", return_value="0.0.0"):
+    from mem0 import Memory
+
+
+class MockVectorMemory:
+    def __init__(self, memory_id, payload):
+        self.id = memory_id
+        self.payload = payload
+
+
+class InMemoryVectorStore:
+    def __init__(self, memories):
+        self.memories = memories
+
+    def get(self, vector_id):
+        return self.memories.get(vector_id)
+
+    def update(self, vector_id, vector, payload):
+        self.memories[vector_id].payload = payload
+
+    def list(self, filters=None, top_k=None):
+        filters = filters or {}
+        results = [
+            memory
+            for memory in self.memories.values()
+            if all(memory.payload.get(key) == value for key, value in filters.items())
+        ]
+        return results[:top_k]
+
+
+def test_issue_6277():
+    memory_id = "mem-1"
+    vector_store = InMemoryVectorStore(
+        {
+            memory_id: MockVectorMemory(
+                memory_id,
+                {
+                    "data": "I love pizza",
+                    "user_id": "alice",
+                    "agent_id": "agent-a",
+                    "run_id": "run-a",
+                    "created_at": "2024-01-01T00:00:00+00:00",
+                    "updated_at": "2024-01-01T00:00:00+00:00",
+                },
+            )
+        }
+    )
+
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

**Finding:** `tests/test_issue_6277.py` passes on current `main`. Its assertions describe behaviour that already holds:

```python
assert alice_memories[0]["user_id"] == "alice"
assert alice_memories[0]["agent_id"] == "agent-a"
assert bob_memories == []
```

Since it is green before any fix, it does not reproduce #6277 and cannot fail if the behaviour regresses. Note also that issue #6277 is already closed.

---

Closing this one. To be clear, this is about the test file, not about issue #6277 — the underlying report is legitimate and stays as it is.

PR #6449, opened 2026-07-20
fix(memory): give the reranker the full candidate pool, not the truncated top-k

### LINKED ISSUE #6448: search(rerank=True) reranks only the truncated top-k, so reranking can't improve recall
### Summary

When `search(rerank=True)` is used, the reranker only ever receives the already-truncated top-`limit` results, not the larger candidate pool. As a result reranking can only **reorder** the final results — it can never surface a relevant memory that the first-stage scorer ranked just below `limit`, which is the main reason to run a reranker at all.

### Where

`mem0/memory/main.py`, `Memory.search()` (and the identical `AsyncMemory.search()`):

```python
original_memories = self._search_vector_store(
    query, effective_filters, limit, threshold, ...
)
...
if rerank and self.reranker and original_memories:
    reranked_memories = self.reranker.rerank(query, original_memories, limit)
```

`_search_vector_store` over-fetches a candidate pool (`internal_limit = max(limit * 4, 60)`) for hybrid scoring, but `score_and_rank(..., top_k=limit)` truncates the result to exactly `limit` before returning (`mem0/utils/scoring.py`). So `original_memories` handed to `reranker.rerank(...)` is already the final top-`limit` set. The over-fetched pool that exists specifically to enable better ranking is discarded one stage too early.

### Why it's wrong

Reranking is a two-stage retrieve-then-rerank design: stage one (bi-encoder / BM25 hybrid) fetches a **broad** candidate pool cheaply, stage two (a cross-encoder / LLM reranker) rescposes that pool and picks the true top-`limit`. Cross-encoders exist precisely because the first-stage scorer misranks — so the correct memory is often at position `limit+1..N`. By truncating to `limit` before reranking, the reranker can never pull that memory into the results; it only shuffles the `limit` items the first stage already chose. Recall is identical with and without `rerank=True`.

### Concrete failure scenario

A user searches `search("refund policy", filters={"user_id": "u1"}, top_k=5, rerank=True)`. The bi-encoder ranks the genuinely-relevant memory 7th (common — this is exactly what cross-encoders fix). `score_and_rank` truncates to 5, dropping it. The reranker then reorders 5 less-relevant candidates and never sees the correct one. The feature silently under-delivers.

### Impact

Every `search(rerank=True)` call, on both `Memory` and `AsyncMemory`, for all reranker providers (Cohere, HuggingFace cross-encoder, LLM, SentenceTransformer, ZeroEntropy). Reranking is a headline OSS feature, and this makes it unable to improve recall.

### Expected

When reranking is enabled, the reranker should receive the broader candidate pool (e.g. `max(limit * 4, 60)`), then narrow it back down to `limit` — so it can actually surface relevant memories the first stage ranked below `limit`.

### Fix

Retrieve a larger candidate pool when reranking is enabled, then let the reranker narrow to `limit` (and, if the rerank call fails, trim the over-fetched pool back to `limit`). Small change in both the sync and async `search()`. PR to follow.

OfficialAbhinavSingh:
## Linked Issue

Closes #6448

## Description

When `search(rerank=True)` is used, the reranker only received the already-truncated top-`limit` results, so reranking could reorder the final results but never surface a relevant memory the first-stage scorer ranked below `limit` — the main reason to run a reranker.

`_search_vector_store` over-fetches a candidate pool (`internal_limit = max(limit * 4, 60)`) for hybrid scoring, but `score_and_rank(..., top_k=limit)` truncates to exactly `limit` before returning. So the list passed to `reranker.rerank(...)` was already the final top-`limit` set, and the over-fetched pool was discarded one stage too early.

**Concrete example:** `search("refund policy", filters={"user_id": "u1"}, top_k=5, rerank=True)` where the bi-encoder ranks the relevant memory 7th (exactly what cross-encoders exist to fix). `score_and_rank` truncates to 5, dropping it; the reranker reorders 5 less-relevant candidates and never sees the correct one. Recall is identical with and without `rerank=True`.

This PR retrieves the broader candidate pool when reranking is enabled, then lets the reranker narrow it back to `limit`. Non-rerank search is unchanged (`retrieval_limit == limit`). If the rerank call fails, the over-fetched pool is trimmed back to `limit` so the `limit` contract still holds. Applied to both the sync `Memory.search()` and async `AsyncMemory.search()`.

```python
retrieval_limit = max(limit * 4, 60) if (rerank and self.reranker) else limit
original_memories = self._search_vector_store(
    query, effective_filters, retrieval_limit, threshold, ...
)
if rerank and self.reranker and original_memories:
    try:
        original_memories = self.reranker.rerank(query, original_memories, limit)
    except Exception as e:
        logger.warning(f"Reranking failed, using original results: {e}")
        original_memories = original_memories[:limit]
```

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — non-rerank search is byte-for-byte unchanged. With reranking enabled, results are the same size (`limit`) but ranking quality improves, since the reranker now scores the full candidate pool.

## Test Coverage

Added three tests in `tests/memory/test_main.py`:

- `test_search_rerank_uses_full_candidate_pool` — asserts the reranker receives more than `limit` candidates (the pool) and is asked to narrow to `limit`; final results respect `limit`. Fails on `main` (reranker gets exactly `limit`) and passes with this change.
- `test_async_search_rerank_uses_full_candidate_pool` — async counterpart.
- `test_search_rerank_failure_still_respects_limit` — on rerank failure, the over-fetched pool is trimmed back to `limit`.

Full `tests/memory/test_main.py` suite (51 tests) and `tests/rerankers/` pass; `ruff check` is clean on the changed source.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1481,8 +1481,14 @@ class Memory(MemoryBase):
         )
 
         search_start = time.perf_counter()
+        # When reranking, retrieve a larger candidate pool so the reranker can
+        # surface relevant memories the first-stage scorer ranked below `limit`.
+        # The reranker then narrows the pool back down to `limit`. Without this,
+        # score_and_rank truncates to `limit` before the reranker runs, so
+        # reranking could only reorder the final results, never improve recall.
+        retrieval_limit = max(limit * 4, 60) if (rerank and self.reranker) else limit
         original_memories = self._search_vector_store(
-            query, effective_filters, limit, threshold, explain=explain, show_expired=show_expired
+            query, effective_filters, retrieval_limit, threshold, explain=explain, show_expired=show_expired
         )
         search_elapsed_seconds = time.perf_counter() - search_start
 
@@ -1493,6 +1499,8 @@ class Memory(MemoryBase):
                 original_memories = reranked_memories
             except Exception as e:
                 logger.warning(f"Reranking failed, using original results: {e}")
+                # Reranking failed, but we over-fetched for it — trim back to limit.
+                original_memories = original_memories[:limit]
 
         if temporal_usage_notice:
             display_temporal_usage_notice(self, "sync", "search", *temporal_usage_notice)
@@ -3131,8 +3139,14 @@ class AsyncMemory(MemoryBase):
         )
 
         search_start = time.perf_counter()
+        # When reranking, retrieve a larger candidate pool so the reranker can
+        # surface relevant memories the first-stage scorer ranked below `limit`.
+        # The reranker then narrows the pool back down to `limit`. Without this,
+        # score_and_rank truncates to `limit` before the reranker runs, so
+        # reranking could only reorder the final results, never improve recall.
+        retrieval_limit = max(limit * 4, 60) if (rerank and self.reranker) else limit
         original_memories = await self._search_vector_store(
-            query, effective_filters, limit, threshold, explain=explain, show_expired=show_expired
+            query, effective_filters, retrieval_limit, threshold, explain=explain, show_expired=show_expired
         )
         search_elapsed_seconds = time.perf_counter() - search_start
 
@@ -3146,6 +3160,8 @@ class AsyncMemory(MemoryBase):
                 original_memories = reranked_memories
             except Exception as e:
                 logger.warning(f"Reranking failed, using original results: {e}")
+                # Reranking failed, but we over-fetched for it — trim back to limit.
+                original_memories = original_memories[:limit]
 
         if temporal_usage_notice:
             await display_temporal_usage_notice_async(self, "async", "search", *temporal_usage_notice)
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1481,14 +1481,16 @@ class Memory(MemoryBase):
         )
 
         search_start = time.perf_counter()
-        # When reranking, retrieve a larger candidate pool so the reranker can
-        # surface relevant memories the first-stage scorer ranked below `limit`.
-        # The reranker then narrows the pool back down to `limit`. Without this,
-        # score_and_rank truncates to `limit` before the reranker runs, so
-        # reranking could only reorder the final results, never improve recall.
-        retrieval_limit = max(limit * 4, 60) if (rerank and self.reranker) else limit
+        # When reranking, return the full candidate pool that _search_vector_store
+        # already over-fetches internally, so the reranker can surface relevant
+        # memories the first-stage scorer ranked below `limit`. The reranker then
+        # narrows the pool back down to `limit`. Without this, score_and_rank
+        # truncates to `limit` before the reranker runs, so reranking could only
+        # reorder the final results, never improve recall.
+        rerank_pool = bool(rerank and self.reranker)
         original_memories = self._search_vector_store(
-            query, effective_filters, retrieval_limit, threshold, explain=explain, show_expired=show_expired
+            query, effective_filters, limit, threshold, explain=explain,
+            show_expired=show_expired, rerank_pool=rerank_pool,
         )
         search_elapsed_seconds = time.perf_counter() - search_start
 
@@ -1623,7 +1625,8 @@ class Memory(MemoryBase):
                 return True
         return False
 
-    def _search_vector_store(self, query, filters, limit, threshold=0.1, explain=False, show_expired=False):
+    def _search_vector_store(self, query, filters, limit, threshold=0.1, explain=False, show_expired=False,
+                             rerank_pool=False):
         # Guard against None threshold (backward compat)
         if threshold is None:
             threshold = 0.1
@@ -1637,6 +1640,11 @@ class Memory(MemoryBase):
 
         # Step 3: Semantic search (over-fetch for scoring pool)
         internal_limit = max(limit * 4, 60)
+        # When feeding a reranker, return the whole over-fetched pool instead of
+        # truncating to `limit`, so the reranker can rescue candidates the
+        # first-stage scorer ranked below `limit`. This does NOT re-widen the DB
+        # fetch — `internal_limit` is still computed from the original `limit`.
+        return_k = internal_limit if rerank_pool else limit
         semantic_results = self.vector_store.search(
             query=query, vectors=embeddings, top_k=internal_limit, filters=filters
         )
@@ -1680,7 +1688,7 @@ class Memory(MemoryBase):
             bm25_scores=bm25_scores,
             entity_boosts=entity_boosts,
             threshold=threshold,
-            top_k=limit,
+            top_k=return_k,
             explain=explain,
         )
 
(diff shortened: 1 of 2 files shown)


mem0/memory/main.py
```diff
@@ -1447,8 +1447,14 @@ def search(
         )
 
         search_start = time.perf_counter()
+        # When reranking, retrieve a larger candidate pool so the reranker can
+        # surface relevant memories the first-stage scorer ranked below `limit`.
+        # The reranker then narrows the pool back down to `limit`. Without this,
+        # score_and_rank truncates to `limit` before the reranker runs, so
+        # reranking could only reorder the final results, never improve recall.
+        retrieval_limit = max(limit * 4, 60) if (rerank and self.reranker) else limit
```
kartik-mem0: This widened `retrieval_limit` (already `max(limit*4, 60)`) gets passed into `_search_vector_store` as its own `limit` parameter, and `_search_vector_store` independently recomputes `internal_limit = max(limit*4, 60)` off that already-widened number. With `top_k=5` and reranking on, the real vector-store query depth compounds to 240 instead of the intended 60 - 4x deeper than needed, on every dense and keyword-search call.

Move the widen decision fully inside `_search_vector_store` (derive `internal_limit` once from the true limit, and truncate `score_and_rank`'s `top_k` to it when the caller signals reranking) so `search()` does not pre-inflate a number `_search_vector_store` inflates again.

kartik-mem0 (commented):
## Changes requested

Fixes the real retrieve-then-rerank ordering bug from #6448, but the fix itself introduces a hot-path regression: the widened pool gets widened again.

**To unblock:**
1. `mem0/memory/main.py:1455` (and the async twin) - stop double-widening before it reaches `_search_vector_store` (detail in the inline comment)

- **Problem is real:** reproduces on `main` at `mem0/memory/main.py:1598-1655` (sync), `:3241-3293` (async) - live sandbox, real Qdrant/pgvector, `reranker saw only 5 candidates (== top_k=5)`.
- **Tests / red-green:** `pytest tests/memory/test_main.py -k rerank` -> 3 passed. Reverting the hunk fails 2/3 tests; a surgical revert of just the failure-trim line fails the third (`assert 60 == 5`). All 3 are load-bearing.
- **CI:** Python SDK (3.10/3.11/3.12), changelog, license/cla, CI Gate all green. `Vercel` red, unrelated docs-deploy gate.
- **Security:** cleared, no new attack surface, pure arithmetic on an internal query-depth integer.
- **Bloat (ponytail):** lean already, comment is proportionate to a subtle ordering bug.
- **Not verified:** TS SDK (`mem0-ts/src/oss/src/memory/index.ts`) has the identical bug, neither PR here touches it - flagged as a follow-up, not a blocker.

</details>

Good catch on the actual bug and clean tests. Push the fix for the double-widening and I will re-review.
Open PR #6455 fixes the same root cause with an identical `mem0/memory/main.py` hunk but bundles an unrelated, untested `mem0/embeddings/gemini.py` change; this PR is the cleaner of the two once the item above is resolved.

OfficialAbhinavSingh:
@kartik-mem0 Good catch on the double-widening — fixed in `b8280f43`.

Root cause was exactly as you described: the caller pre-widened `limit` to `max(limit*4, 60)` and passed that into `_search_vector_store`, which then recomputes `internal_limit = max(limit*4, 60)` off the already-widened number → `max(60*4, 60) = 240` rows fetched for `top_k=5` instead of the intended 60.

Fix: the caller now passes the **original** `limit` plus a `rerank_pool` flag (both sync and async). `_search_vector_store` keeps its single over-fetch (`internal_limit` computed from the original `limit`) and only changes its **return count** to that full pool when `rerank_pool` is set — so the reranker still sees the whole candidate pool while the DB fetch stays at one `max(limit*4, 60)`. No behavior change on the non-rerank path.

Added a regression guard: the rerank tests now assert `_search_vector_store` is called with the original limit (not a pre-widened one) — I verified it's load-bearing (reintroducing the pre-widening turns it red). `pytest -k rerank` → 3 passed, full `tests/memory/` → 355 passed, rerankers suite → 104 passed, `ruff` clean.

On the two follow-ups you flagged: happy to open the TS twin (`mem0-ts` has the identical rerank-truncation) as a separate PR if useful, and I agree #6455 bundling the untested `gemini.py` change makes this the cleaner vehicle for the core fix. Ready for re-review.

mem0/memory/main.py
kartik-mem0: Verified against `a58558b1`: the compounding double-widening finding is fixed (`mem0/memory/main.py:1622,1627,1629,1634` sync; `:3276,3281,3283,3288` async). `internal_limit` is now derived once from the original `limit`, and `return_k` (not `limit`) controls only the final `score_and_rank` truncation, so retrieval depth for both dense and keyword search no longer compounds. Red-green: reverting to the pre-follow-up commit reproduces `passed_limit == 60` instead of `5`; 4/4 sandbox lane by mode runs pass on this head.

kartik-mem0 (commented):

Verified against `a58558b1` (new head): the confirmed compounding 4x over-fetch is fixed. `_search_vector_store` now derives `internal_limit` once from the true `limit`, and a `rerank_pool` flag decides only the final `score_and_rank` truncation (`mem0/memory/main.py:1622,1627`).

**Follow-up:** TS side (`mem0-ts/src/oss/src/memory/index.ts:1406,1535,1587`) still has the original truncate-before-rerank bug from #6448, not the double-widening variant (that shape needs Python's two-layer split). Worth #6537 picking up the core fix.
**Not verified:** live reranker API calls; `SpyReranker` plus real Qdrant/pgvector only.

Risk: low, internal fetch/truncate split only, no public API change · Your time: ~5 min

- **Problem is real (cycle 1):** reproduced on `main` at `mem0/memory/main.py:1655` (score_and_rank truncates to `limit` before the reranker ever sees the pool).
- **This cycle's delta reproduced pre-fix:** reverted to the pre-follow-up commit (`58727af5`), kept PR-head tests: `assert passed_limit == FINAL_LIMIT` failed with `60 == 5`, the exact compounded double-widen value.
- **Tests:** `hatch -e dev_py_3_11 run pytest tests/memory/test_main.py -k rerank -v` -> 3 passed. Full module: `pytest tests/memory/test_main.py -q` -> 51 passed. `ruff check` clean.
- **Red-green:** reverted, 2/3 rerank tests failed; restored, 3/3 pass. ✓
- **Sandbox (live):** `issue6448_rerank_truncated_pool`, 4/4 lane×mode PASS (oss-qdrant, oss-pgvector × sync, async) against this head, real Qdrant/pgvector, no mocks.
- **CI:** all required checks green; Vercel red (auth-gated preview deploy, non-blocking).
- **Security:** cleared, no new input surface, `rerank_pool` is an internal bool derived from existing config.
- **Bloat (ponytail):** lean already, one bool param plus one local, no new abstraction.
- **Linked issue:** #6448

</details>

OfficialAbhinavSingh:
@kartik-mem0 checking in on this one. It has been sitting since your "ready for squash-merge" on 24 Jul.

I rebased it onto main on 4 Aug (2.0.13 to 2.0.15) to keep `changelog_check` from going stale, and CI Gate is green on the new head `6766827d`. The diff is unchanged from the `a58558b1` you verified.

Anything outstanding on my side? The TypeScript twin #6537, which you flagged as the follow-up in your review, is open and still awaiting a first review.

kartik-mem0 (commented):
Re-reviewed at `6766827d`. Content-identical rebase of the previously reviewed head. Fresh tests and lint pass against the new SHA. LGTM still holds.

OfficialAbhinavSingh:
Thanks for re-verifying both of these today.

One thing that may not be visible from your side: every review on this PR and on #6516 is `COMMENTED` (4 each, none `APPROVED`), and `.github/AGENTS.md` describes `Main Branch Rule` as requiring one approving review with no bypass actors. So both still show "Review required" under the LGTM and sit at `mergeable_state=blocked`.

Is an explicit approval (or a direct merge) what's needed here, or is there something still outstanding on my side?

PR #6452, opened 2026-07-21
fix(vector_stores/chroma): stop dropping filter conditions in where-clause translation

### LINKED ISSUE #6451: bug(vector_stores/chroma): filter translation silently drops conditions and emits invalid where clauses
### Component

Core / Python SDK

### Description

`ChromaDB._generate_where_clause` (`mem0/vector_stores/chroma.py`) violates ChromaDB's where-clause grammar — which requires **exactly one operator or field per dict level** — in three independent ways. Each one silently corrupts query results rather than raising.

#### 1. Same-field ranges silently lose a bound

When a field carries multiple operators, the translation loop reassigns the field's clause on every iteration, so each operator overwrites the previous one and only the last survives. This is on the primary path: the core filter pipeline (`_build_filters_and_metadata` / `merge_filters` in `mem0/memory/main.py`) canonically deep-merges `{"AND": [{"age": {"gte": 18}}, {"age": {"lte": 65}}]}` into `{"age": {"gte": 18, "lte": 65}}` before handing it to the vector store.

```python
ChromaDB._generate_where_clause({"age": {"gte": 18, "lte": 65}})
# actual:   {'age': {'$lte': 65}}          <- gte: 18 silently dropped
# expected: {'$and': [{'age': {'$gte': 18}}, {'age': {'$lte': 65}}]}
```

Verified against a real `chromadb.EphemeralClient` with rows aged 15/30/70: the emitted clause returns the under-18 row the caller explicitly excluded. (Note: even without the overwrite, `{"age": {"$gte": 18, "$lte": 65}}` would be rejected by ChromaDB with `Expected operator expression to have exactly one operator`, so the fix must split into `$and` clauses.)

This is the Python-Chroma equivalent of the TS SDK defect tracked in #6260; no existing issue covers the Python backend.

#### 2. `NOT` with `contains`/`icontains` erases the entire where clause

The `$not` branch negates via a fixed map covering only `eq/ne/gt/gte/lt/lte/in/nin`. Operators absent from the map — notably `contains`/`icontains`, which the positive path accepts (equality fallback) — are silently skipped:

```python
ChromaDB._generate_where_clause({"$not": [{"title": {"contains": "draft"}}]})
# actual:   None                            <- whole where clause vanishes, query runs unfiltered
# expected: {'title': {'$ne': 'draft'}}     (mirroring the positive path's $eq fallback)
```

A user filtering with `{"NOT": [{"title": {"contains": "draft"}}]}` gets **all** rows back, including the drafts they excluded.

#### 3. Multi-field conditions inside `$or` produce clauses ChromaDB rejects

`$or` branches merge each condition's fields into a flat dict. ChromaDB rejects flat multi-field dicts:

```python
import chromadb
col.get(where={"age": {"$gte": 18}, "vip": {"$eq": False}})
# chromadb.errors.InvalidArgumentError: Expected where to have exactly one operator, got ...
```

So `{"OR": [{"age": {"gte": 18}, "vip": True}, ...]}` translates into a query that fails at request time.

### Comparison with sibling backends

Qdrant collects all range operators into a single `Range(**kwargs)` (`qdrant.py:_build_field_condition`) and handles `NOT` generically by recursing into `must_not`, so both bounds and negated `contains` work there. Pinecone and pgvector also accumulate operators correctly. Chroma is the outlier.

### Steps To Reproduce

Run the snippets above against `main` (no server needed for the translation layer; the result-set checks use `chromadb.EphemeralClient`).

### Expected Behavior

Range filters keep both bounds; `NOT contains` negates instead of vanishing; `$or` with multi-field conditions emits grammar-valid `$and`-wrapped clauses.

### Actual Behavior

Wrong result sets (cases 1–2, silent) or `InvalidArgumentError` at query time (case 3).

### Environment

- mem0 version: main branch
- Python: 3.9+
- OS: All

I have a fix ready (each operator becomes its own single-operator clause, combined with `$and`/`$or`; `NOT` mirrors the positive path's equality fallback as `$ne`) with regression tests — PR incoming right after this issue.

microbluey:
Fixes #6451

## Problem

`ChromaDB._generate_where_clause` violated ChromaDB's one-operator-per-level where grammar in three ways, each silently corrupting queries (full analysis and reproduction in #6451):

1. **Same-field ranges lost a bound** — each operator overwrote the previous one, so the canonical merged form `{"age": {"gte": 18, "lte": 65}}` degraded to `{"$lte": 65}` and returned rows the caller excluded.
2. **`NOT` + `contains`/`icontains` erased the whole where clause** — operators missing from the negation map were silently skipped, so the query ran unfiltered.
3. **Multi-field conditions inside `$or` emitted flat dicts** that ChromaDB rejects with `Expected where to have exactly one operator`.

## Fix

Rewrote the translation so every operator becomes its own single-operator clause, combined explicitly:

- `convert_condition` now returns a **list** of single-field clauses; a shared `combine()` helper wraps multi-clause lists in `$and`/`$or` and unwraps singletons (preserving the previous output shape for all already-correct inputs).
- The `$not` branch negates `contains`/`icontains` and unknown operators to `$ne`, mirroring the positive path's `$eq` fallback, instead of dropping them.
- `$or` branches combine each condition's clauses with `$and` instead of flat-merging fields.

## Verification

- 4 new regression tests covering: same-field range, range inside `$or`, multi-field condition inside `$or`, and `NOT contains`/`icontains`.
- All 27 existing tests in `tests/vector_stores/test_chroma.py` pass unchanged (output shapes for previously-correct inputs are identical).
- End-to-end against a real `chromadb.EphemeralClient` (rows aged 15/30/70):
  - range 18–65: before `['a','b']` (includes excluded under-18 row) → after `['b']` ✅
  - `NOT contains`: before unfiltered → after correct exclusion ✅
  - multi-field `$or`: before `InvalidArgumentError` → after correct result set ✅
- `ruff check` passes; formatting of touched lines matches `ruff format`.

### DIFF
--- a/mem0/vector_stores/chroma.py
+++ b/mem0/vector_stores/chroma.py
@@ -254,110 +254,111 @@ class ChromaDB(VectorStoreBase):
     def _generate_where_clause(where: dict[str, any]) -> dict[str, any]:
         """
         Generate a properly formatted where clause for ChromaDB.
-        
+
+        ChromaDB's where grammar allows exactly one field or one logical
+        operator per dict level, so multiple operators on the same field and
+        multi-field conditions must be combined with an explicit ``$and``.
+
         Args:
             where (dict[str, any]): The filter conditions.
-            
+
         Returns:
             dict[str, any]: Properly formatted where clause for ChromaDB.
         """
         if where is None:
             return None
-        
-        def convert_condition(key: str, value: any) -> dict:
-            """Convert universal filter format to ChromaDB format."""
+
+        op_map = {
+            "eq": "$eq",
+            "ne": "$ne",
+            "gt": "$gt",
+            "gte": "$gte",
+            "lt": "$lt",
+            "lte": "$lte",
+            "in": "$in",
+            "nin": "$nin",
+        }
+        # Negation of each operator. contains/icontains fall back to equality
+        # on the positive path (ChromaDB has no substring match), so their
+        # negation falls back to inequality for consistency.
+        negate_map = {
+            "eq": "$ne",
+            "ne": "$eq",
+            "gt": "$lte",
+            "gte": "$lt",
+            "lt": "$gte",
+            "lte": "$gt",
+            "in": "$nin",
+            "nin": "$in",
+        }
+
+        def convert_condition(key: str, value: any) -> list:
+            """Convert one field condition to a list of single-field ChromaDB clauses."""
             if value == "*":
                 # Wildcard - match any value (ChromaDB doesn't have direct wildcard, so we skip this filter)
+                return []
+            if isinstance(value, dict):
+                # One clause per operator: ChromaDB rejects field expressions
+                # with more than one operator, so a range like
+                # {"gte": 18, "lte": 65} must become two clauses combined
+                # with $and by the caller (previously each operator
+                # overwrote the last, silently dropping bounds).
+                # contains/icontains and unknown operators fall back to equality.
+                return [{key: {op_map.get(op, "$eq"): val}} for op, val in value.items()]
+            # Simple equality
+            return [{key: {"$eq": value}}]
+
+        def combine(clauses: list, operator: str):
+            """Combine clauses under a logical operator, unwrapping singletons."""
+            if not clauses:
                 return None
-            elif isinstance(value, dict):
-                # Handle comparison operators
-                chroma_condition = {}
-                for op, val in value.items():
-                    if op == "eq":
-                        chroma_condition[key] = {"$eq": val}
-                    elif op == "ne":
-                        chroma_condition[key] = {"$ne": val}
-                    elif op == "gt":
-                        chroma_condition[key] = {"$gt": val}
-                    elif op == "gte":
-                        chroma_condition[key] = {"$gte": val}
-                    elif op == "lt":
-                        chroma_condition[key] = {"$lt": val}
-                    elif op == "lte":
-                        chroma_condition[key] = {"$lte": val}
-                    elif op == "in":
-                        chroma_condition[key] = {"$in": val}
-                    elif op == "nin":
-                        chroma_condition[key] = {"$nin": val}
-                    elif op in ["contains", "icontains"]:
-                        # ChromaDB doesn't support contains, fallback to equality
-                        chroma_condition[key] = {"$eq": val}
-                    else:
-                        # Unknown operator, treat as equality
-                        chroma_condition[key] = {"$eq": val}
-                return chroma_condition
-            else:
-                # Simple equality
-                return {key: {"$eq": value}}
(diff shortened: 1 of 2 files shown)


kartik-mem0 (commented):

`_generate_where_clause` stopped overwriting same-field operators (a range like `{gte: 18, lte: 65}` degraded to just the last bound), flattening multi-field `$or` branches into a where-clause ChromaDB rejects at query time, and silently dropping `$not` + `contains`/`icontains` conditions entirely.

**Before you merge:** open PR #6259 touches this same function for a different bug (#6258, unknown-operator fallback) and will need a rebase once this lands, heads-up for whoever merges next.
**Follow-up:** `mem0-ts/src/oss/src/vector_stores/chroma.ts:273-401` carries the identical three bugs (same-field overwrite, flat `$or` merge, unmapped `$not` operators dropped), worth its own tracking issue.
**Not verified:** no dedicated Chroma sandbox lane, verified with a live `chromadb.EphemeralClient()` instead of the sandbox harness (exact output in the evidence file).

Risk: low, pure query-time translation, no new I/O or schema change · Your time: ~5 min

- **Problem is real:** reproduced on `main` with a live `chromadb.EphemeralClient()`: `{"age": {"gte": 18, "lte": 65}}` collapsed to `{"$lte": 65}` and returned an out-of-range row; `{"$not": [{"title": {"contains": "draft"}}]}` produced `{}` (no filter at all); a multi-field condition inside `$or` raised ChromaDB's own `ValueError: Expected where to have exactly one operator` at query time.
- **Tests:** `pytest tests/vector_stores/test_chroma.py -v` -> 31 passed (27 existing + 4 new).
- **Red-green:** reverted `chroma.py` to `main` in an isolated worktree, kept the PR's 4 new tests -> all 4 failed with the exact wrong values above; restored the fix -> 31/31 passed.
- **CI:** Python SDK (3.10/3.11/3.12), changelog, license/cla, CI Gate all green; `Vercel` preview fails (unrelated docs-deploy auth gate, advisory on every PR).
- **Security:** cleared, filters are structured dicts built internally by mem0's own filter layer, never a raw string reaching ChromaDB, so there is no injection surface; no tenancy or trust-boundary logic touched by this function.
- **Bloat (ponytail):** lean already, the new `combine()` helper replaces three separately-duplicated length-check blocks, net reduction in duplication.
- **Linked issue:** #6451

</details>

microbluey:
Thanks for the thorough red-green verification. Filed #6513 to track the identical three bugs in `mem0-ts/src/oss/src/vector_stores/chroma.ts:273-401` — I'll port this fix over in a follow-up PR with regression tests mirroring the four Python ones.

kartik-mem0 (approved):

PR #4109, opened 2026-02-22
fix: make OpenSearch filter dynamic instead of hardcoded fields

andreluiz1987:
## Description

Previously, the search method only applied filters for a hardcoded set of fields (user_id, run_id, agent_id), causing any other field passed in the filters parameter to be silently ignored.

This PR replaces the static field iteration with a dynamic one that iterates directly over the provided filters dict, so any key passed by the caller is correctly applied as a filter clause in the final OpenSearch query.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] Documentation update

## How Has This Been Tested?

Added unit tests to verify that any field passed in filters is correctly applied to the OpenSearch query. Tests cover: dynamic filter fields, None value exclusion, and plain knn query when no filters are provided.

- [x] Unit Test
- [ ] Test Script (please provide)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

## Maintainer Checklist

- [ ] closes #xxxx (Replace xxxx with the GitHub issue number)
- [ ] Made sure Checks passed

PR #6454, opened 2026-07-21
fix(vector_stores/opensearch): honor all filter keys instead of a hardcoded identity-key list

### LINKED ISSUE #6453: bug(vector_stores/opensearch): search/keyword_search/list silently ignore all non-identity filter keys
### Component

Core / Python SDK

### Description

The OpenSearch vector store silently ignores every filter key except the three identity keys. All three query paths — `search()`, `keyword_search()`, and `list()` — iterate a hardcoded list instead of the caller's filters (`mem0/vector_stores/opensearch.py`):

```python
filter_clauses = []
if filters:
    for key in ["user_id", "run_id", "agent_id"]:
        value = filters.get(key)
        if value:
            _validate_filter(key, value)
            filter_clauses.append({"term": {f"payload.{key}.keyword": value}})
```

Any other filter key — `category`, `project`, any custom metadata field — is never read, so the query runs completely unfiltered on that dimension. No error, no warning.

### Steps To Reproduce

```python
m.search("invoice", user_id="u1", filters={"category": "billing"})
```

### Expected Behavior

Results filtered by both `user_id` and `category` (this is how the sibling Elasticsearch backend behaves — `elasticsearch.py` iterates `filters.items()` and builds a term clause for every key).

### Actual Behavior

`user_id` is honored, `category` is silently dropped: memories from **every** category come back.

### Additional context

- A secondary defect hides behind the first: the clause template hardcodes the `.keyword` sub-field, but the `payload` mapping is a dynamic `object`, and dynamic mappings only create `.keyword` sub-fields for **string** values. Once custom keys are honored, numeric/boolean filters (`{"age": 30}`, `{"archived": False}` — both accepted by `_validate_filter`) would still never match via `payload.age.keyword`. Non-string values need to match the field itself.
- This is the same defect class as Weaviate's #6423 (PR #6402), which doesn't cover OpenSearch.
- PR #4109 (Feb 2026, now marked stale, never reviewed) fixes the hardcoded list for `search()` only; `keyword_search()` and `list()` have the same block, and it predates the `.keyword`/non-string issue.

### Environment

- mem0 version: main branch
- Python: 3.9+
- OS: All

Fix ready (shared `_build_filter_clauses` helper across all three methods, keeping the `_validate_filter` injection guard, with `.keyword` for strings and plain field for non-strings) plus regression tests — PR incoming right after this issue.

microbluey:
Fixes #6453

## Problem

`search()`, `keyword_search()`, and `list()` each iterated a hardcoded `["user_id", "run_id", "agent_id"]` list when building filter clauses, so every other filter key was silently dropped and the query ran unfiltered on that dimension (full analysis in #6453). The sibling Elasticsearch backend already iterates all filter keys.

## Fix

The three duplicated blocks now share a module-level `_build_filter_clauses` helper that:

- iterates **every** key in `filters` (skipping `None` values), keeping the `_validate_filter` injection guard for each key/value;
- matches string values against the `.keyword` sub-field as before, and non-string values (int/float/bool — all accepted by `_validate_filter`) against the field itself, since the dynamic `payload` object mapping only creates `.keyword` sub-fields for strings — the previous blanket `.keyword` suffix could never match numeric or boolean filters.

## Relationship to #4109

#4109 (Feb 2026, marked stale, never reviewed) fixes the hardcoded list for `search()` only. This PR covers all three affected methods, deduplicates the logic into one helper, and additionally handles the non-string `.keyword` mismatch. Happy to close in favor of #4109 if its author returns and extends it, but a complete fresh fix seemed more useful than resurrecting a partial stale one.

## Verification

- 5 new regression tests (`TestOpenSearchCustomFilters`): custom keys honored in all three methods, non-string values use the plain field path, and custom keys still go through `_validate_filter`.
- All 29 existing tests in `tests/vector_stores/test_opensearch.py` pass unchanged.
- `ruff check` passes; touched lines match `ruff format`.

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -28,6 +28,24 @@ def _validate_filter(key: str, value) -> None:
         )
 
 
+def _build_filter_clauses(filters):
+    """Build term clauses from every filter key, not just the identity keys.
+
+    The ``payload`` mapping is a dynamic object, so only string values get a
+    ``.keyword`` sub-field; non-string values (int/float/bool) must match
+    against the field itself.
+    """
+    filter_clauses = []
+    if filters:
+        for key, value in filters.items():
+            if value is None:
+                continue
+            _validate_filter(key, value)
+            field = f"payload.{key}.keyword" if isinstance(value, str) else f"payload.{key}"
+            filter_clauses.append({"term": {field: value}})
+    return filter_clauses
+
+
 class OutputData(BaseModel):
     id: str
     score: float
@@ -204,13 +222,7 @@ class OpenSearchDB(VectorStoreBase):
         query_body = {"size": top_k * 2, "query": None}
 
         # Prepare filter conditions if applicable
-        filter_clauses = []
-        if filters:
-            for key in ["user_id", "run_id", "agent_id"]:
-                value = filters.get(key)
-                if value:
-                    _validate_filter(key, value)
-                    filter_clauses.append({"term": {f"payload.{key}.keyword": value}})
+        filter_clauses = _build_filter_clauses(filters)
 
         # Combine knn with filters if needed
         if filter_clauses:
@@ -255,13 +267,7 @@ class OpenSearchDB(VectorStoreBase):
         }
 
         # Apply filters consistently with the existing search() method
-        filter_clauses = []
-        if filters:
-            for key in ["user_id", "run_id", "agent_id"]:
-                value = filters.get(key)
-                if value:
-                    _validate_filter(key, value)
-                    filter_clauses.append({"term": {f"payload.{key}.keyword": value}})
+        filter_clauses = _build_filter_clauses(filters)
 
         if filter_clauses:
             bool_query["filter"] = filter_clauses
@@ -370,13 +376,7 @@ class OpenSearchDB(VectorStoreBase):
             """List all memories with optional filters."""
             query: Dict = {"query": {"match_all": {}}}
 
-            filter_clauses = []
-            if filters:
-                for key in ["user_id", "run_id", "agent_id"]:
-                    value = filters.get(key)
-                    if value:
-                        _validate_filter(key, value)
-                        filter_clauses.append({"term": {f"payload.{key}.keyword": value}})
+            filter_clauses = _build_filter_clauses(filters)
 
             if filter_clauses:
                 query["query"] = {"bool": {"filter": filter_clauses}}
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -16,6 +16,7 @@ from mem0.vector_stores.base import VectorStoreBase
 logger = logging.getLogger(__name__)
 
 _SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_.]*$")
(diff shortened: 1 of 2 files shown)


microbluey:
Thanks for catching that — you're right, validating every key would have broken the operator-shaped values (`$or`/`$not`, comparison dicts, wildcard `"*"`) that `_process_metadata_filters` passes through, which the old hardcoded identity-key loop never saw. The skip-with-debug-log approach keeps the scope right, and the test-class dedup is cleaner too. LGTM.

FWIW I re-checked #6452 (the Chroma sibling fix) against the same canonical shapes — it falls back instead of raising everywhere, so it doesn't have this landmine.

mem0/vector_stores/opensearch.py
```diff
@@ -28,6 +29,21 @@ def _validate_filter(key: str, value) -> None:
         )
 
 
+def _build_filter_clauses(filters):
+    """Build term clauses from every filter key, not just the identity keys."""
+    filter_clauses = []
+    for key, value in (filters or {}).items():
+        if value is None:
+            continue
+        if key not in _IDENTITY_FILTER_KEYS and (not isinstance(value, (str, int, float, bool)) or value == "*"):
```
kartik-mem0, line 38: Minor, non-blocking: identity keys are exempt from the wildcard check here, so `filters={"agent_id": "*"}` still builds a literal (near-always-empty) term match instead of "any agent" semantics (a documented Platform pattern). Pre-existing, not introduced by this PR - `opensearch.ts` already treats `"*"` as an `exists` query for every key. Worth a follow-up.

kartik-mem0 (commented):

OpenSearch's `search()`/`keyword_search()`/`list()` iterated a hardcoded `user_id/run_id/agent_id` list and silently dropped any other filter key (issue #6453); this replaces all three loops with one `_build_filter_clauses` helper that honors every key, matching how `elasticsearch.py` already works.

**Follow-up:** close #4109 (stale, narrower duplicate, only patches one of the three call sites) in favor of this.
**Not verified:** live OpenSearch keyword_search path (proven at the unit-test level only; the sandbox scenario exercises get_all/search).

Risk: low, adds filtering, never removes it, and the identity-key path is untouched · Your time: ~3 min

- **Problem is real:** reproduced on `main` at `mem0/vector_stores/opensearch.py:207-217,258-267,373-382` and live: `get_all(filters={"user_id":u,"topic":"food"})` returned all 5 seeded rows instead of 1.
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_opensearch.py -v` -> 43 passed.
- **Red-green:** reverted `opensearch.py` to `main` -> 5 of the new tests failed (spanning search/keyword_search/list); restored -> 43 passed again. ✓
- **CI:** real checks (lint+test matrix, CI Gate) green; ignored the Vercel preview-deploy check (fails on every fork PR, not merge-blocking).
- **Security:** cleared. Field/value validation unchanged and still runs before any key reaches the query; filters only ever narrow results.
- **Bloat (ponytail):** lean already, ship. 3 duplicated ~7-line loops collapsed into 1 shared helper, no speculative abstraction added.
- **Linked issue:** #6453

</details>

microbluey:
Good catch on the identity-key wildcard exemption — happy to take that as a follow-up PR: treat `"*"` as an exists query for every key (matching how `opensearch.ts` already handles it), replacing the literal term match on identity keys.

kartik-mem0 (approved):

PR #6465, opened 2026-07-21
fix(memory): paginate delete_all so it removes all matching memories

cen-zp:
## Summary

Fixes #4869.

`Memory.delete_all()` and `AsyncMemory.delete_all()` previously called `vector_store.list(filters=filters)` exactly once without specifying `top_k`, so they only received the first page of matching memories (most vector stores default to 100). When a user had more matching memories than the page size, the older entries were silently left behind.

This fix iterates in pages: each loop calls `list(filters, top_k=_DELETE_ALL_PAGE_SIZE)`, deletes the returned batch **inside the loop** (so the result set shrinks between calls), and stops when the store returns an empty page.

## Root Cause

The vector store `list()` interface has no offset/cursor parameter — pagination must happen on the caller side. The original code assumed a single `list()` call would return all matching memories, which is incorrect for stores that cap results at a default page size (Qdrant, Pinecone, Milvus, Turbopuffer all default to 100).

The async path had an additional latent bug: if deletion were deferred until after the loop (as in PR #5950's approach), the store would never shrink between `list()` calls, causing an infinite loop when matches ≥ page size.

### `mem0/memory/main.py`

* Added module-level constants `_DELETE_ALL_PAGE_SIZE = 1000` and `_DELETE_ALL_MAX_ITERATIONS = 10_000` (safety cap).

* **Sync `delete_all`**: loops calling `list(filters, top_k=1000)` → deletes each batch → repeats until empty page. Uses `seen_ids` set to guard against eventual-consistency lag (e.g. Elasticsearch/OpenSearch refresh delay re-listing a deleted ID). Individual delete failures are logged and skipped rather than aborting the entire operation.

* **Async `delete_all`**: same pagination loop with `asyncio.gather` for batch deletion **inside** the loop. Entity store bulk cleanup happens once after all pages are processed. Error handling preserved (`return_exceptions=True`).

* Both paths reuse the existing `_vector_store_list_rows()` helper for consistent return-format normalization.

### `tests/test_main.py`

* Updated `test_delete_all_trims_user_id_before_list` and `test_delete_all_coerces_integer_user_id_before_list` assertions to include `top_k=1000` (reflecting the new explicit pagination parameter).

### `tests/test_memory.py`

Added 4 regression tests with a `_StatefulVectorStore` that simulates a store with an internal page-size cap:

* `test_sync_delete_all_paginates_through_all_memories` — 5 memories, page\_size=2 → asserts all 5 deleted across multiple list() calls.

* `test_sync_delete_all_stops_when_first_page_empty` — empty store → exactly 1 list() call, 0 deletes.

* `test_async_delete_all_paginates_through_all_memories` — same coverage for AsyncMemory; proves deletion happens inside the loop (store shrinks between pages).

* `test_async_delete_all_stops_when_first_page_empty` — empty store → 1 list() call, 0 deletes.

## Testing

```bash
# Targeted pagination tests
pytest tests/test_memory.py -k delete_all -v
# → 6 passed

# Full test files
pytest tests/test_main.py tests/test_memory.py
# → 105 passed, 1 failed (pre-existing langchain_core import, unrelated)

# Lint
ruff check mem0/memory/main.py tests/test_memory.py tests/test_main.py
# → All checks passed!
```

**Red-green verification**: reverting the sync/async loops back to a single `list()` call causes the pagination tests to fail (`assert len(store.deleted_ids) == 5` → only 2 deleted), confirming the tests are load-bearing.

## Related Issue

Closes #4869

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -90,6 +90,16 @@ def _vector_store_list_rows(listed):
     return []
 
 
+# Page size used by delete_all when iterating through vector store results.
+# Vector stores cap list() at a default page size (Qdrant/Pinecone/Milvus
+# default to 100); delete_all must loop to remove all matching memories.
+_DELETE_ALL_PAGE_SIZE = 1000
+
+# Safety cap on pagination iterations to prevent infinite loops when a vector
+# store's list() does not shrink after deletion (e.g. eventual-consistency lag).
+_DELETE_ALL_MAX_ITERATIONS = 10_000
+
+
 # Fields that hold runtime auth/connection objects and must be preserved.
 # These are non-serializable objects (e.g. AWSV4SignerAuth, RequestsHttpConnection)
 # needed by clients like OpenSearch — not sensitive strings to redact.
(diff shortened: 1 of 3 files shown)


cen-zp:
Closing this in favor of #5950, which the maintainers have already identified as the current PR to review for #4869. I opened this before fully accounting for that existing review path, and I do not want to create competing duplicate work or bypass the original contributor.

While working through the same bug, I confirmed the async deletion must happen inside each pagination iteration so the result set shrinks before the next list() call. I also found an eventual-consistency edge case: if a page contains only IDs that were already seen but have not disappeared from the store yet, treating that as immediate completion can leave later records undeleted. A bounded no-progress retry policy would be safer than breaking on the first all-stale page.

I am keeping the branch available as implementation and stateful-test reference, and I will add the relevant findings to #5950. If the maintainer or original author would like help carrying the existing PR forward, I am happy to assist.

PR #6466, opened 2026-07-21
docs(platform): correct 34 audited API discrepancies across platform docs

kartik-mem0:
## Linked Issue

No pre-existing issue. This came out of a full audit of all 26 pages under
`docs.mem0.ai/platform`, cross-checked against the production API and backend
source. Happy to open a tracking issue if the team prefers one on record.

## Description

Every page under `/platform` was checked claim-by-claim against live API
responses and the backend implementation. 34 discrepancies were confirmed. This
PR fixes the documentation-side ones. The most user-visible categories:

**`add()` response shape (v1 envelope documented for a v3 endpoint).** Six pages
showed `{"message": "Memory processing has been queued for background
execution", "status", "event_id"}`. That envelope is real, but it belongs to
v1 (`mem0/views.py`). The pages in question all describe v3: the API reference
is bound to `openapi: post /v3/memories/add/`, and the SDK's `add()` posts to
`/v3/memories/add/`. v3 returns the service dict unwrapped, i.e.
`{"event_id", "status"}`. Corrected in `add-memories.mdx`, `openapi.json`
(schema + example), `langchain-tools.mdx`, `platform-v2-to-v3.mdx`,
`group-chat.mdx`, and the OpenAI tool-calls cookbook. The v1 pages were left
alone, where that envelope is still accurate.

**Entity scoping was described backwards.** The docs said a memory carries
`user_id` and `agent_id` together when both are passed. On the default
extraction path the v3 worker routes each extracted fact by `attributed_to`,
so a record gets `user_id` **or** `agent_id`, never both. This matters
practically: the docs recommended `AND`-ing the two in a filter, which returns
zero results. Rewritten across `entity-scoped-memory.mdx` and
`v2-memory-filters.mdx`, with Direct Import called out as the exception.

**Temporal reasoning precedence was inverted.** `timestamp` takes priority over
`observation_datetime` / `observation_date` as the ranking anchor, not the
other way round.

**Direct import** documented the wrong (async) response shape and claimed
`assistant` messages are dropped. It stores `user` and `assistant` alike; only
`system` is dropped.

**Multimodal** said structured content under `infer=False` errors. It is
silently skipped: the call succeeds and the content is simply missing.

**Feedback** samples now use the SDK exception taxonomy (`mem0.exceptions`,
`error.errorCode`) instead of raw HTTP handling.

Plus assorted smaller corrections across `quickstart`, `faqs`, `webhooks`,
`memory-export`, `memory-decay`, `graph-memory`, `custom-categories`,
`custom-instructions`, `async-client`, `advanced-memory-operations`,
`agent-signup`, `mem0-mcp`, and `overview`.

### Not fixed here (needs backend work)

Three findings are server-side bugs, not doc bugs, and are being raised with
the platform team separately:

1. `keywords` filters return 500. The `text` attribute in the Turbopuffer
   schema lacks `"filterable": True`.
2. Multimodal items are silently dropped under `infer=False` (`.encode()` on
   dict content). Documented as a caveat here.
3. `feedback()` with a well-formed but non-existent `memory_id` returns an
   unhandled 500 instead of 404. Documented as a caveat here.

One claim on `agent-signup.mdx` (rate limiting returns a generic 403 with no
`Retry-After`) is retained on the strength of the live API response alone. The
throttle is not in the application repo, so it is presumably enforced at the
gateway layer and could not be confirmed at source.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A. Documentation only, no runtime code touched.

## Test Coverage

Every change was traced to backend source before editing rather than taken from
the audit report on faith. Two proposed edits were rejected on that basis: the
`add-memories.mdx` change was nearly applied to a v1 page before the frontmatter
showed it was v3-bound, and a claim that the `eq` operator is rejected on entity
filters had no supporting mechanism in `FilterBuilder` / `EntityFieldProcessor`,
so that page was left untouched.

Mechanically verified across the diff: no em-dashes, MDX component tags
balanced, code fences balanced, every JSON block parses, `openapi.json` still
valid JSON, and `scripts/check-llms-txt-coverage.py` reports `docs/llms.txt` in
sync.

## Checklist

### DIFF
--- a/docs/platform/features/feedback-mechanism.mdx
+++ b/docs/platform/features/feedback-mechanism.mdx
@@ -183,7 +183,7 @@ try {
     if (error.errorCode === "HTTP_400") {
         console.log("memory_id is not a valid UUID");
     } else if (error.errorCode === "HTTP_500") {
-        console.log("memory_id was not found or not accessible");
+        console.log("Server error - the memory_id may not exist or be inaccessible, or it may be transient; retry if it persists");
     } else {
         console.log(`Unexpected error: ${error.message}`);
     }
@@ -195,7 +195,7 @@ try {
 ### Handling an unknown or malformed `memory_id`
 
 - A malformed `memory_id` (not a valid UUID) returns a clean `400` with `{"error": "Invalid memory_id"}`.
-- A well-formed but non-existent `memory_id` (or one that doesn't belong to your org or project) currently returns an unhandled `500 Internal Server Error`, not a `404`. Treat any `5xx` from this endpoint as "the memory_id was not found or not accessible" until this is fixed server-side, and only call `feedback` with a `memory_id` you just read from a `get_all` or `search` response.
+- A well-formed but non-existent `memory_id` (or one that doesn't belong to your org or project) currently returns a `500 Internal Server Error` instead of a `404`. This is a server-side bug being tracked in [MEM-5745](https://linear.app/mem0/issue/MEM-5745). A `5xx` can also be a transient server error, so don't treat it as a definitive "not found"; to avoid the case entirely, only call `feedback` with a `memory_id` you just read from a `get_all` or `search` response.
 
 ## Feedback Analytics
 
--- a/docs/platform/features/feedback-mechanism.mdx
+++ b/docs/platform/features/feedback-mechanism.mdx
@@ -150,7 +150,7 @@ Handle potential errors when submitting feedback:
 
 ```python Python
 from mem0 import MemoryClient
-from mem0.exceptions import MemoryNotFoundError, NetworkError
+from mem0.exceptions import MemoryError, ValidationError
 
 client = MemoryClient(api_key="your_api_key")
 
@@ -161,12 +161,10 @@ try:
         feedback_reason="Helpful context for user query"
     )
     print("Feedback submitted successfully")
-except MemoryNotFoundError:
-    print("Memory not found")
-except NetworkError as e:
-    print(f"Network error: {e}")
-except Exception as e:
-    print(f"Unexpected error: {e}")
+except ValidationError:
+    print("memory_id is not a valid UUID")
+except MemoryError as e:
+    print(f"memory_id was not found or not accessible ({e.error_code})")
 ```
 
 ```javascript JavaScript
@@ -182,15 +180,22 @@ try {
     });
     console.log("Feedback submitted successfully");
 } catch (error) {
-    if (error.status === 404) {
-        console.log("Memory not found");
+    if (error.errorCode === "HTTP_400") {
+        console.log("memory_id is not a valid UUID");
+    } else if (error.errorCode === "HTTP_500") {
+        console.log("memory_id was not found or not accessible");
     } else {
-        console.log(`Error: ${error.message}`);
+        console.log(`Unexpected error: ${error.message}`);
     }
 }
 ```
 
 </CodeGroup>
+
+### Handling an unknown or malformed `memory_id`
+
+- A malformed `memory_id` (not a valid UUID) returns a clean `400` with `{"error": "Invalid memory_id"}`.
+- A well-formed but non-existent `memory_id` (or one that doesn't belong to your org or project) currently returns an unhandled `500 Internal Server Error`, not a `404`. Treat any `5xx` from this endpoint as "the memory_id was not found or not accessible" until this is fixed server-side, and only call `feedback` with a `memory_id` you just read from a `get_all` or `search` response.
 
 ## Feedback Analytics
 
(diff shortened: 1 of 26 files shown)


docs/platform/features/feedback-mechanism.mdx
```diff
@@ -182,16 +180,23 @@ try {
...
     }
 }
 ```
 
 </CodeGroup>
 
+### Handling an unknown or malformed `memory_id`
+
+- A malformed `memory_id` (not a valid UUID) returns a clean `400` with `{"error": "Invalid memory_id"}`.
+- A well-formed but non-existent `memory_id` (or one that doesn't belong to your org or project) currently returns an unhandled `500 Internal Server Error`, not a `404`. Treat any `5xx` from this endpoint as "the memory_id was not found or not accessible" until this is fixed server-side, and only call `feedback` with a `memory_id` you just read from a `get_all` or `search` response.
```
karthik-indla: Can you validate this once by making an actual call. When do we return 400 and when we return 500. If we actually return 500 for valid UUID that're not part of our system then it's a serious bug. Pls add a linear ticket so that we can fix it.

karthik-indla: `Treat any `5xx` from this endpoint as "the memory_id was not found or not accessible" until this is fixed server-side`
This is misleading - if we actually have some Server side issues / service unavailability / API Timeouts issues the error code will be 503/504 etc. They shouldn't mean memory_id not found. Pls reiterate on this.

docs/platform/features/group-chat.mdx
```diff
@@ -48,46 +36,24 @@ from mem0 import MemoryClient
...
+alice_messages = [
+    {"role": "user", "content": "Hey team, I think we should use React for the frontend"},
 ]
-
-response = client.add(
-    messages,
-    run_id="group_chat_1",
-    infer=True
-)
+response = client.add(alice_messages, user_id="alice", run_id="group_chat_1")
```
karthik-indla: This whole group-chat block I feel needs more examples.  The since add example will not explain how user_id, run_id, name impact things. Pls add more examples.

docs/platform/features/v2-memory-filters.mdx
```diff
@@ -120,26 +119,6 @@ filters = {
 Find memories containing specific text, categories, or metadata values.
 
 <AccordionGroup>
-  <Accordion title="Text search">
```
karthik-indla, line 123: why are these examples getting removed?

docs/platform/features/feedback-mechanism.mdx
kartik-mem0: Validated live against api.mem0.ai (2026-07-23) with dummy data:

- Malformed `memory_id` (not a UUID) → clean `400` `{"error":"Invalid memory_id"}`
- Well-formed but non-existent / foreign `memory_id` → `500`, not `404`

So yes, the 500 is a real server-side bug for valid UUIDs outside your org/project. Filed **MEM-5745** (High) and the doc now references it. Reworded in 3d82d12a.

kartik-mem0: Agreed, that wording was misleading. Reworded in 3d82d12a: it no longer tells readers to treat any `5xx` as "not found." A `503/504` is a genuine transient/service error, so the doc now says a `5xx` may be the non-existent-id bug (MEM-5745) *or* a transient error, and steers callers to only pass a `memory_id` they just read from `get_all`/`search`. Softened the JS `HTTP_500` branch message the same way.

docs/platform/features/group-chat.mdx
kartik-mem0: Added more examples in 3d82d12a:

- **Basic Group Chat** now calls `add()` once per participant (alice/bob/charlie), each with their own `user_id` and a shared `run_id`, so the retrieval outputs below actually line up with what was added, and it shows how `run_id` groups while `user_id` separates.
- New **"The `name` field does not change scope"** section with a concrete demo: two different `name`s in one `add()` call scoped to a single `user_id` → every memory lands under that `user_id`, nothing under the names. Verified live against api.mem0.ai and against the backend source.

docs/platform/features/v2-memory-filters.mdx
kartik-mem0: You are right, they should not have been dropped wholesale. Restored in 3d82d12a. Context on why they were originally removed: I had found `keywords` returning a `500`, but on live validation that only happens inside `search()` filters. `keywords` works correctly on `get_all`. So I restored the examples using `get_all`, added a "Text search (keywords)" accordion, and the accurate caveat: `keywords` inside `search()` filters 500s today (MEM-5746) so use the `query` argument for search relevance.

karthik-indla (approved):

PR #6468, opened 2026-07-21
fix(ts-oss): forward responseFormat to Gemini in generateResponse

### LINKED ISSUE #6467: bug(ts-oss): Gemini generateResponse ignores responseFormat
### Component
TypeScript SDK (OSS) - `mem0-ts/src/oss/src/llms/google.ts`

### What happens

`GoogleLLM.generateResponse` takes a `responseFormat` param but never forwards it. During fact extraction we call it with `{ type: "json_object" }` (memory/index.ts:888), but google.ts only ever builds `config` from `tools`, so the arg is dropped. The request Gemini receives is identical whether or not a format was asked for.

Gemini's structured output is opt-in - you have to set `config.responseMimeType = "application/json"`. Since we never do, the model is never actually told to return JSON. It mostly limps along because there's a markdown fence strip on the output, so when Gemini happens to wrap its reply in ```json it works by accident. When it doesn't, extraction just fails quietly.

### Confirmed

Spied on the `generateContent` call - config is `{}` with and without `{ type: "json_object" }`, `responseMimeType` never set.

The Python SDK already does this correctly (`mem0/llms/gemini.py`, sets `response_mime_type`), so this is a TS parity gap. Same class of bug as the Anthropic one (#6203).

### Fix

Set `responseMimeType` when `json_object` is requested, matching the Python side. PR incoming.

ac12644:
## Linked Issue

Closes #6467

## Description

`GoogleLLM.generateResponse` accepted `responseFormat` but never forwarded it, so `{ type: "json_object" }` (what the extraction step passes) was a no-op. Gemini's JSON output is opt-in via `responseMimeType`, so the model was never told to return JSON - we were just relying on the markdown fence strip downstream to bail us out, which only works when Gemini happens to fence its reply.

This sets `config.responseMimeType = "application/json"` when `json_object` is requested, same as the Python SDK already does in `mem0/llms/gemini.py`. Signature is unchanged (matches the base `LLM` interface and `openai.ts`).

Same bug as the Anthropic one, different provider.

## Type of Change

## Test Coverage

Added two cases to `google-llm.test.ts`: `responseMimeType` is set to `application/json` for `json_object`, and left unset when no format is passed. Verified only `memory/index.ts` passes `responseFormat` and it never passes `tools` alongside it, so this can't collide with Gemini's function-calling path.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/llms/google.ts
+++ b/mem0-ts/src/oss/src/llms/google.ts
@@ -59,6 +59,15 @@ export class GoogleLLM implements LLM {
       ];
     }
 
+    // Honor a requested JSON response format (parity with the Python SDK's
+    // mem0/llms/gemini.py). Gemini's structured output is opt-in via
+    // responseMimeType — without it the model is never told to emit JSON, so
+    // callers passing json_object silently get free-form text and depend on a
+    // fragile markdown-fence strip downstream.
+    if (responseFormat?.type === "json_object") {
+      config.responseMimeType = "application/json";
+    }
+
     const completion = await this.google.models.generateContent({
       contents,
       model: this.model,
--- a/mem0-ts/src/oss/tests/google-llm.test.ts
+++ b/mem0-ts/src/oss/tests/google-llm.test.ts
@@ -185,6 +185,37 @@ describe("GoogleLLM (unit)", () => {
     expect(response.toolCalls[1].name).toBe("add_graph_memory");
   });
 
+  // Regression: generateResponse accepted a responseFormat argument but never
+  // forwarded it, so Gemini was never told to emit JSON (parity with the Python
+  // SDK's mem0/llms/gemini.py, which sets response_mime_type/response_schema).
+  it("forwards json_object responseFormat as responseMimeType", async () => {
+    mockGenerateContent.mockResolvedValueOnce({
+      text: '{"facts": ["fact1"]}',
+      functionCalls: null,
+    });
+
+    const llm = new GoogleLLM({ apiKey: "test-key" });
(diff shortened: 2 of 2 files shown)


kartik-mem0 (commented):

Gemini's `generateResponse` accepted `responseFormat` but never read it; this forwards `{ type: "json_object" }` to the SDK's `responseMimeType` config field.

**Follow-up:** the sibling Anthropic gap (#6203) needs a different mechanism (no single-field equivalent), already tracked by open PR #6205, not this one
**Not verified:** live Gemini API call, verified against the installed SDK's own type definitions plus a red-green unit test instead

Risk: low, single provider file, no shared code touched · Your time: ~3 min

- **Problem is real:** reproduces on `main` at `mem0-ts/src/oss/src/llms/google.ts:48-60`, `config` is built only from `tools`, `responseFormat` (param at :42) is never read
- **Tests:** `npx jest tests/google-llm.test.ts --verbose` -> 9 passed
- **Red-green:** reverting `google.ts` to `main` (keeping the PR's test) fails "forwards json_object responseFormat as responseMimeType" (`expect(received).toBe("application/json")`, got `undefined`); restoring the fix passes it again ✓
- **CI:** all required checks green (CI Gate, TS SDK build x2, TS SDK integration x2, changelog, CLA); Vercel preview is red but non-blocking, ignored per policy
- **Security:** cleared, `responseFormat.type` is compared with strict `===` against a fixed literal and never reaches a query, shell, URL, or log
- **Bloat (ponytail):** lean already, 9 lines of logic plus two non-redundant tests (positive and negative case)
- **Linked issue:** #6467

</details>

kartik-mem0 (approved):

PR #6481, opened 2026-07-21
fix(llms/anthropic): find the text block in no-tools responses instead of indexing content[0]

### LINKED ISSUE #6480: bug(llms/anthropic): no-tools response parsing crashes on thinking blocks and empty content
### What happens

`AnthropicLLM._parse_response` returns `response.content[0].text` on the no-tools path (mem0/llms/anthropic.py:152). Two ways this crashes:

1. **Thinking blocks.** Kwargs passed to `generate_response()` flow through `_get_common_params()` straight into `client.messages.create(**params)`, so enabling extended thinking (`thinking={"type": "enabled", "budget_tokens": 1024}`) works fine on the request side. But the response then comes back with `content = [ThinkingBlock, TextBlock]`, thinking first. `ThinkingBlock` has no `.text` attribute, so `content[0].text` raises `AttributeError` and the whole call dies even though the model answered.
2. **Empty content.** When a response carries no blocks at all, `content[0]` raises `IndexError`.

The tools branch a few lines up already handles this correctly by iterating the blocks and checking `block.type == "text"`. Only the no-tools return is unguarded.

This is the direct Anthropic SDK sibling of #6368 (Bedrock Converse crashing when `reasoningContent` precedes the text block), same failure shape in a different provider file.

### Proposed fix

Mirror the tools branch on the no-tools path: iterate `response.content`, return the first text block's text, and fall back to `""` when no text block exists.

### Acceptance

- `generate_response()` without tools returns the text when a thinking block precedes it
- returns `""` instead of raising on empty content
- plain single-text-block responses behave exactly as before

---

update: I took a stab at a fix too (#6481). put it together with help from Claude Code and ran the lint and test gates locally, but I'm new to this codebase so totally fine if you'd rather take a different approach. mostly trying to learn here :)

ayaangazali:
## Linked Issue

Closes #6480

## Description

The no-tools path of `AnthropicLLM._parse_response` returns `response.content[0].text` (mem0/llms/anthropic.py:152). When extended thinking is enabled the response starts with a thinking block, which has no `.text`, so the call raises AttributeError even though the model answered. An empty content list raises IndexError the same way. This mirrors #6368 on the Bedrock side, where the same content[0] indexing crashed on a leading reasoningContent block. The fix iterates the blocks and returns the first text block's text, falling back to an empty string, exactly like the tools branch a few lines above already does.

One note on tests: the whole `tests/llms/test_anthropic.py` module is guarded by `pytest.importorskip("anthropic")` and the `anthropic` package is not part of the dev env features, so CI skips it (same as the existing tests in that file). I verified locally with anthropic installed: both new regression tests fail on main (AttributeError and IndexError at anthropic.py:152) and pass on this branch, plus a third test pinning the plain single-text-block behavior, which passes on both. `make lint` and `make test-py-3.11` are green (1709 passed).

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I tested manually (ran the new tests against main's anthropic.py to confirm they reproduce the AttributeError and IndexError, then against the branch where all 25 module tests pass)

## Checklist

---

*small disclosure like on my previous PRs: I'm a freshman doing my best to give back to tools I use. Claude Code helped me trace this one and write the fix and tests, and I ran all the gates locally myself. if anything here misses the mark I would honestly rather hear it than not, still learning :)*

### DIFF
--- a/mem0/llms/anthropic.py
+++ b/mem0/llms/anthropic.py
@@ -149,4 +149,11 @@ class AnthropicLLM(LLMBase):
                 elif block.type == "tool_use":
                     result["tool_calls"].append({"name": block.name, "arguments": block.input})
             return result
-        return response.content[0].text
+
+        # Thinking-enabled responses put a thinking block before the text
+        # block, and a response can carry no text block at all, so find the
+        # text block like the tools branch above instead of indexing content[0].
+        for block in response.content:
+            if block.type == "text":
+                return block.text
+        return ""
--- a/tests/llms/test_anthropic.py
+++ b/tests/llms/test_anthropic.py
@@ -166,6 +166,56 @@ def test_enable_sampling_parameters_for_current_models(mock_anthropic_client, mo
     assert ("temperature" in call_kwargs) is expected
 
 
+def test_generate_response_returns_text_when_thinking_block_precedes_it(mock_anthropic_client):
+    """Extended thinking responses put a thinking block before the text block; parsing must skip it."""
+    config = AnthropicConfig(model="claude-opus-4-6", api_key="test-key")
+    llm = AnthropicLLM(config)
+
+    thinking_block = Mock(spec=["type", "thinking", "signature"])
+    thinking_block.type = "thinking"
+    text_block = Mock(spec=["type", "text"])
+    text_block.type = "text"
+    text_block.text = "The answer"
+
+    mock_response = Mock()
+    mock_response.content = [thinking_block, text_block]
+    mock_anthropic_client.messages.create.return_value = mock_response
+
+    result = llm.generate_response([{"role": "user", "content": "Hi"}])
+
+    assert result == "The answer"
+
+
+def test_generate_response_returns_empty_string_on_empty_content(mock_anthropic_client):
+    """A response with no content blocks must return "" instead of raising IndexError."""
+    config = AnthropicConfig(model="claude-opus-4-6", api_key="test-key")
+    llm = AnthropicLLM(config)
+
+    mock_response = Mock()
+    mock_response.content = []
+    mock_anthropic_client.messages.create.return_value = mock_response
+
+    result = llm.generate_response([{"role": "user", "content": "Hi"}])
+
+    assert result == ""
+
+
+def test_generate_response_returns_text_for_plain_text_response(mock_anthropic_client):
+    """Regular single-text-block responses keep returning the text unchanged."""
+    config = AnthropicConfig(model="claude-opus-4-6", api_key="test-key")
+    llm = AnthropicLLM(config)
+
+    text_block = Mock(spec=["type", "text"])
+    text_block.type = "text"
+    text_block.text = "Hello!"
+
+    mock_response = Mock()
+    mock_response.content = [text_block]
+    mock_anthropic_client.messages.create.return_value = mock_response
+
+    assert llm.generate_response([{"role": "user", "content": "Hi"}]) == "Hello!"
+
+
 def test_enable_sampling_parameters_respects_explicit_override(mock_anthropic_client):
     config = AnthropicConfig(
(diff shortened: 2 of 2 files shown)


kartik-mem0:
## LGTM, ready for human review and squash-merge

### What this PR does
Fixes `AnthropicLLM._parse_response`'s no-tools path so it scans for the first text block instead of indexing `response.content[0]`, which crashed whenever a thinking block preceded the text or content was empty.

### Strengths
- `mem0/llms/anthropic.py:152-155`: mirrors the existing tools-branch loop three lines above instead of inventing a new pattern.
- Generic to any leading non-text block (thinking, future block types), not a special case for "thinking" alone.
- `tests/llms/test_anthropic.py`: uses `Mock(spec=[...])` so the mocked blocks reproduce the actual attribute shape (a thinking block genuinely has no `.text`), not just the crash mechanism.

### Issues
Minor (optional): the loop can shrink to `next((b.text for b in response.content if b.type == "text"), "")`, one line instead of seven. Not required, current form is perfectly readable.

### Evidence
- **Problem verified:** reproduces on `main` at `mem0/llms/anthropic.py:152`, both conditions confirmed independently: thinking-block-first raises `AttributeError` ('Mock' object has no attribute 'text'), empty `content` raises `IndexError`.
- **CI:** label + license/cla pass; Vercel failure is a docs-preview auth gate, unrelated to this Python-only change. CI Gate has not run yet (typical for a first-time external contributor pending workflow approval).
- **Targeted tests:** `pytest tests/llms/test_anthropic.py -v` in an isolated venv -> 25 passed, including all 3 new tests.
- **Red-green proof:** reverted just the fix (kept the new tests) and reran -> `test_generate_response_returns_text_when_thinking_block_precedes_it` and `test_generate_response_returns_empty_string_on_empty_content` fail with the exact `IndexError`/`AttributeError` predicted; `test_generate_response_returns_text_for_plain_text_response` still passes. Restored the fix -> all 25 pass again. ✓
- **Linked issue:** #6480, PR addresses the reported root cause.
- **Cross-package check:** `mem0-ts/src/oss/src/llms/anthropic.ts:103-108` has the identical shape of bug in its own no-tools path (`response.content[0]`, throws on empty content, throws `"Unexpected response type"` instead of finding the text when a non-text block leads). Not touched by this PR, not a duplicate of anything open right now, a cross-package sibling worth its own follow-up issue.
- **Bloat (ponytail):** `net: -6 lines possible` (the `next()` one-liner above), lean otherwise, the 3 tests are each targeting a distinct case, none redundant.
- **Security:** cleared. `response.content` is Anthropic SDK response data, not attacker-shaped input; the extracted `.text` is the same model-generated string that flowed through before this fix, just no longer crashing on the way out. No new parsing, query, or shell boundary touched.
- **Could NOT verify:** live Anthropic API behavior with a real extended-thinking response, code tracing plus mocked-block tests only (matches the existing test file's approach for this module).

### Assessment
Clean root-cause fix, well-tested with real red-green proof, correctly scoped to the file the linked issue names. The TS-side sibling bug is worth a follow-up issue but doesn't block this PR.

mem0/llms/anthropic.py
```diff
@@ -149,4 +149,11 @@ def _parse_response(self, response, tools):
...
             return result
-        return response.content[0].text
+
+        # Thinking-enabled responses put a thinking block before the text
+        # block, and a response can carry no text block at all, so find the
+        # text block like the tools branch above instead of indexing content[0].
+        for block in response.content:
+            if block.type == "text":
+                return block.text
+        return ""
```
kartik-mem0, line 159: Same logic, fewer lines: `return next((b.text for b in response.content if b.type == "text"), "")`. Optional, current form is fine too.

kartik-mem0 (approved):

PR #6483, opened 2026-07-22
fix: drop unused vector column from pgvector get() and list() queries

### LINKED ISSUE #6482: PGVector get() and list() fetch the vector column but never use it
### Component

Python SDK

### Summary

In the pgvector vector store, `PGVector.get()` and `PGVector.list()` run `SELECT id, vector, payload ...`, but the fetched vector is always discarded: both methods build `OutputData` from id and payload only, and `OutputData` has no vector field to hold it. With 1536-dim embeddings this is ~6KB of wasted transfer per row, so `Memory.get_all()` with a large limit reads megabytes from Postgres per call for nothing. The qdrant backend already skips vectors for the same operation (`with_vectors=False` in its `list()`), so this looks like an oversight specific to the pgvector backend.

### Steps to Reproduce

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "pgvector",
        "config": {
            "dbname": "postgres",
            "user": "postgres",
            "password": "postgres",
            "host": "localhost",
            "port": 5432,
        },
    },
}

m.add("I like coffee", user_id="alice")
m.get_all(user_id="alice", limit=2000)  # delegates to PGVector.list()
```

Then inspect the executed query, e.g. with `log_statement = 'all'` in Postgres or via `pg_stat_statements`.

### Expected Behavior

`get()` and `list()` select only the columns they actually consume (`id`, `payload`), like the qdrant backend does for the same operation.

### Actual Behavior

The executed query is:

    SELECT id, vector, payload FROM <collection> WHERE payload->>$1 = $2 LIMIT $3

and the vector column is dropped immediately after fetching. In `list()` the rows are mapped to `OutputData(id=..., score=None, payload=r[2])`, so `r[1]`(the vector) is never read; `get()` has the same pattern. No error is raised — the cost is silent IO overhead. We noticed it while investigating vector store IO load: `get_all` with limit=2000 on 1536-dim embeddings transferred ~12MB per call, none of which was used.

I have a fix ready and will open a PR linked to this issue.

### Environment

- mem0 version: 1.0.4 (also present on main, dd5f7e39)
- OS: Linux (Debian, python:3.11-slim container)

Gyubin:
## Linked Issue

Closes #6482 

## Description

`PGVector.get()` and `PGVector.list()` select the vector column but never use it: both build `OutputData` from id and payload only, so every fetched embedding is discarded (`OutputData` has no vector field to hold it). With 1536-dim embeddings that is ~6KB of wasted transfer per row, and `Memory.get_all()` calls with large limits read megabytes per call just to throw them away. The qdrant backend already skips vectors for the same operation (`with_vectors=False` in
`list()`).

This PR selects only `id` and `payload`, keeping filter and LIMIT semantics and the return shape unchanged. The mocked cursor rows and query-string assertions in the pgvector tests are updated to match, so they now also guard against the unused column being reintroduced.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

`tests/vector_stores/test_pgvector.py` (83 tests) passes locally.

## Checklist

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -459,13 +459,13 @@ class PGVector(VectorStoreBase):
         self._ensure_collection()
         with self._get_cursor() as cur:
             cur.execute(
-                sql.SQL("SELECT id, vector, payload FROM {} WHERE id = %s").format(self._col()),
+                sql.SQL("SELECT id, payload FROM {} WHERE id = %s").format(self._col()),
                 (vector_id,),
             )
             result = cur.fetchone()
             if not result:
                 return None
-            return OutputData(id=str(result[0]), score=None, payload=result[2])
+            return OutputData(id=str(result[0]), score=None, payload=result[1])
 
     def list_cols(self) -> List[str]:
         """
@@ -528,7 +528,7 @@ class PGVector(VectorStoreBase):
         with self._get_cursor() as cur:
             cur.execute(
                 sql.SQL("""
-                SELECT id, vector, payload
+                SELECT id, payload
                 FROM {}
                 {}
                 LIMIT %s
@@ -536,7 +536,7 @@ class PGVector(VectorStoreBase):
                 (*filter_params, top_k),
             )
             results = cur.fetchall()
-        return [[OutputData(id=str(r[0]), score=None, payload=r[2]) for r in results]]
+        return [[OutputData(id=str(r[0]), score=None, payload=r[1]) for r in results]]
 
     def __del__(self) -> None:
         """
--- a/tests/vector_stores/test_pgvector.py
+++ b/tests/vector_stores/test_pgvector.py
@@ -711,7 +711,7 @@ class TestPGVector(unittest.TestCase):
         mock_get_cursor.return_value.__exit__.return_value = None
         
         self.mock_cursor.fetchall.return_value = []  # No existing collections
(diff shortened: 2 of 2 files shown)


kartik-mem0 (commented):

`pgvector.py` `get()`/`list()` selected the unused `vector` column on every call; this drops it from both SELECTs and shifts the payload index accordingly.

**Follow-up:** duplicate PR #6503 carries the identical core fix plus an unrelated hunk that duplicates open PR #6175; already flagged for removal in a separate review, no action needed here
**Not verified:** live Postgres beyond the sandbox container

Risk: low, drops an unused SELECT column and shifts one tuple index, no public API change · Your time: ~2 min

- **Problem is real:** `mem0/vector_stores/pgvector.py:462` (get) and `:531` (list) select the unused `vector` column; sandbox on `main` FAILs both sync and async, PASSes on this PR's head in both modes
- **Tests:** `hatch -e dev_py_3_11 run pytest tests/vector_stores/test_pgvector.py -v` -> 83 passed
- **Red-green:** reverting `pgvector.py` to `main` (tests kept) turns 10 tests RED (`IndexError: tuple index out of range` at `pgvector.py:539`); restoring the PR's file turns them GREEN again (83 passed)
- **CI:** CI Gate green, Python SDK build_mem0 3.10/3.11/3.12 all green
- **Security:** cleared, static SQL column list, no injection or tenancy surface touched
- **Bloat (ponytail):** lean already, minimal diff, nothing to trim
- **Linked issue:** #6482

</details>

kartik-mem0 (approved):

PR #6485, opened 2026-07-22
fix(ts-oss/baidu): convert L2 distance to similarity score in search()

### LINKED ISSUE #6484: TS OSS Baidu store returns raw L2 distance as similarity score, inverting search ranking
### Where

`mem0-ts/src/oss/src/vector_stores/baidu.ts`, `search()`.

### What happens

Mochow returns the raw L2 distance for a vector search (lower = closer). The TS store passes that value straight through as `score`:

```ts
return (response.rows ?? []).map((result) => ({
  id: String(result.row.id),
  payload: resultPayload(result.row),
  score: result.score,
}));
```

The VectorStore contract (and the whole memory layer) treats `score` as a similarity where higher is better. `scoreAndRank()` sorts descending by score and drops anything under the threshold. Feeding it a raw L2 distance inverts both: the farthest matches sort first, and threshold filtering keeps the wrong rows.

The default metric is `L2` (same as the Python default in `mem0/configs/vector_stores/baidu.py`), so this hits the out-of-the-box config, not an edge case.

### Expected

For the L2 metric, convert the distance to a similarity before returning, the same way the Python provider now does after #6435:

```py
if self.metric_type in (MetricType.L2, "L2"):
    raw_score = 1.0 / (1.0 + raw_score)
```

Non-L2 metrics (COSINE, IP) already return higher-is-better and should pass through untouched. `keywordSearch()` (BM25) is already higher-is-better and is not affected.

### Note

This is the TS counterpart of #6434 / #6435, which fixed the exact same thing on the Python side. The TS store was not updated in that PR.

abhay-codes07:
### Description

The TS OSS Baidu store returned the raw L2 distance from Mochow as the result `score`. The memory layer treats `score` as a similarity where higher is better: `scoreAndRank()` sorts descending and applies the threshold to it. Passing a raw L2 distance (lower = closer) inverts the ranking, so the farthest matches come back first and threshold filtering keeps the wrong rows. L2 is the default metric, so this affects the default config, not an edge case.

The fix converts the L2 distance to a similarity with `1 / (1 + distance)` before returning, matching the Python provider (fixed in #6435) and the milvus store. Non-L2 metrics (COSINE, IP) already return a higher-is-better score and pass through untouched. `keywordSearch()` (BM25) is already higher-is-better and is left alone.

Closes #6484

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/baidu.test.ts` (run from the `mem0-ts` root), 29 passed.

- Added `converts an L2 distance into a similarity score (higher = better)`: two hits with L2 distances 0.5 and 2.0 come back as `1/1.5` and `1/3.0`, closer scored higher. Reverting the source change fails exactly this test and nothing else.
- Updated the existing envelope-mapping test to use a non-L2 (COSINE) metric so its pass-through score assertion stays about shape, mirroring how the Python test suite splits these two cases.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/baidu.ts
+++ b/mem0-ts/src/oss/src/vector_stores/baidu.ts
@@ -453,7 +453,12 @@ export class BaiduDB implements VectorStore {
     return (response.rows ?? []).map((result) => ({
       id: String(result.row.id),
       payload: resultPayload(result.row),
-      score: result.score,
+      // Mochow returns the raw L2 distance (lower = closer). Convert it to a
+      // similarity score (higher = better) to satisfy the VectorStore contract,
+      // mirroring the Python provider and the milvus store. Non-L2 metrics
+      // already return a higher-is-better score, so they pass through untouched.
+      score:
+        this.metricType === "L2" ? 1 / (1 + (result.score ?? 0)) : result.score,
     }));
   }
 
--- a/mem0-ts/src/oss/src/vector_stores/baidu.ts
+++ b/mem0-ts/src/oss/src/vector_stores/baidu.ts
@@ -453,12 +453,13 @@ export class BaiduDB implements VectorStore {
     return (response.rows ?? []).map((result) => ({
       id: String(result.row.id),
       payload: resultPayload(result.row),
-      // Mochow returns the raw L2 distance (lower = closer). Convert it to a
-      // similarity score (higher = better) to satisfy the VectorStore contract,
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss baidu fix with tests. This is the TS counterpart of #6435: the Baidu store returned the raw L2 distance as score, so on the default L2 metric search ranking came back inverted. Now converts to a similarity the same way the Python provider does. Quick look when you get a chance, thanks.

kartik-mem0 (approved):

PR #6486, opened 2026-07-22
fix(embeddings/huggingface): forward api_key on base_url/TEI path

### LINKED ISSUE #6301: bug(embeddings/huggingface): api_key ignored on base_url path, endpoint auth dropped

Bartok9:
## Summary

Closes #6301.

When `huggingface_base_url` is set, `HuggingFaceEmbedding` builds an OpenAI-compatible client for TEI (or similar) endpoints but never passed `api_key`. Configured keys and `HUGGINGFACE_API_KEY` were ignored, so authenticated endpoints silently dropped auth.

## Root cause

```python
self.client = OpenAI(base_url=self.config.huggingface_base_url)
```

No `api_key=` argument; the local SentenceTransformer path was unaffected.

## Fix

- Prefer `config.api_key`, then `HUGGINGFACE_API_KEY`, then `"hf"` (local TEI often accepts any non-empty key).
- Document `api_key` on the HuggingFace embedder table.
- Tests: config key, env key, default, config-over-env, and updated existing base_url test.

## Differential value vs #6302

Same core fix as #6302 (also open). This PR adds explicit config-over-env precedence coverage and keeps the docs table in sync. Happy to close as duplicate if #6302 merges first.

## Verification

```bash
uv run pytest tests/embeddings/test_huggingface_embeddings.py -q
# 15 passed
```

## Test plan

- [x] unit tests for TEI client api_key wiring
- [ ] CI green on Python SDK matrix

### DIFF
--- a/docs/components/embedders/models/huggingface.mdx
+++ b/docs/components/embedders/models/huggingface.mdx
@@ -99,6 +99,7 @@ Here are the parameters available for configuring the Hugging Face embedder:
 | `embedding_dims` | Dimensions of the embedding model | `selected_model_dimensions` |
 | `model_kwargs` | Additional arguments for the model | `None` |
 | `huggingface_base_url` | URL to connect to Text Embeddings Inference (TEI) API | `None` |
+| `api_key` | API key for the TEI or OpenAI-compatible endpoint; falls back to the `HUGGINGFACE_API_KEY` env var | `"hf"` |
 </Tab>
 <Tab title="TypeScript">
 | Parameter | Description | Default Value |
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Literal, Optional
 
 from openai import OpenAI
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Closing as a duplicate of #6302, which was opened first and covers the same change. I compared the diffs line by line and the source changes are substantively identical, so nothing is lost here: #6302 stays open and is the one that will get reviewed and merged.

This is not a comment on the quality of your patch. Thank you for the time you put into it, and apologies that our review queue let two people do the same work. If you believe this PR covers something #6302 does not, reply here and I will reopen it.

Bartok9:
Thanks @kartik-mem0 — agreed on the core fix. No need to reopen unless you want the extra coverage here.

Diff vs #6302 is effectively:
- same `api_key = config or HUGGINGFACE_API_KEY or "hf"` wiring + docs row
- one additional test: `test_base_url_config_api_key_wins_over_env` (config precedence when both config and env are set)

If #6302 is the merge vehicle, feel free to lift that single test (and the short comment block if useful). Happy to leave this closed either way.

PR #6487, opened 2026-07-22
fix(memory): paginate delete_all so all matching memories are removed

Bartok9:
## Summary

Part of #4869 (Python `delete_all` path; the TS `deleteAll` half is not touched here and remains tracked by #4869).

`Memory.delete_all()` and `AsyncMemory.delete_all()` only called `vector_store.list(filters=...)` once without `top_k`. Backends that default `list` size (commonly ~100) silently left remaining matches after bulk delete.

## Root cause

Caller assumed one `list()` returned the full filtered set. The vector-store `list` API has no cursor/offset contract; pagination is the caller's job.

## Fix

- Constants `_DELETE_ALL_PAGE_SIZE = 1000`, `_DELETE_ALL_MAX_ITERATIONS`, and `_DELETE_ALL_MAX_NO_PROGRESS`.
- Sync + async: loop `list(filters, top_k=...)`, delete the batch **inside** the loop (so subsequent lists shrink), `seen_ids` guard for lag.
- **Refresh-lag tolerance:** an all-seen page no longer breaks the drain immediately. Both paths retry a bounded number of no-progress pages before concluding, so Elasticsearch/OpenSearch-style stale pages don't hide later unseen rows.
- **Sync failure semantics preserved:** sync `delete_all` surfaces `_delete_memory` exceptions instead of swallowing them and returning success with residual memories.
- Async still runs `_bulk_clear_entity_store` after the drain; decay/first-run notices use `total_deleted`.

## Salvage credit

Rebases/extends **@cen-zp** #6465 (closed while iterating the same design) and the line of earlier work (#5950 @ly-wang19, #4872 @shafdev, etc.).

## Verification

```bash
uv run pytest tests/memory/test_delete_all_pagination.py tests/memory/test_decay_usage_notice.py -q
```

## Test plan

- [x] multi-page sync + async drain
- [x] **stateful-store load-bearing** sync + async (deletion must occur in-loop or list() never shrinks)
- [x] stale all-seen page followed by new ids (refresh-lag)
- [x] sync delete failure propagates
- [x] filters + top_k kwargs
- [x] existing delete_all decay notice tests

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -3595,11 +3595,15 @@ class AsyncMemory(MemoryBase):
                 return_exceptions=True,
             )
             for memory, result in zip(batch, results):
-                seen_ids.add(memory.id)
                 if isinstance(result, BaseException):
+                    # Do not mark seen on failure — leave residual IDs eligible
+                    # for retry on the next list page (async concurrent deletes
+                    # can fail without surfacing; marking early silently drops
+                    # them while delete_all still reports success).
                     total_errors += 1
                     logger.warning("Failed to delete memory %s: %s", memory.id, result)
                 else:
+                    seen_ids.add(memory.id)
                     total_deleted += 1
 
         if self._entity_store is not None:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -110,6 +110,14 @@ _DELETE_ALL_MAX_NO_PROGRESS = 3
 # store time to expose newly-visible rows before we conclude the drain.
 _DELETE_ALL_NO_PROGRESS_BACKOFF = 0.5
 
+# Max times a single id may fail deletion before we stop retrying it and count
+# it as a hard error. A transient concurrent-delete miss stays eligible for a
+# few retries (see async delete_all), but an id whose delete keeps failing must
+# not be re-listed and re-attempted every iteration up to _DELETE_ALL_MAX_
+# ITERATIONS, which would hammer the backend. After this many failures the id
+# is marked seen so the drain can conclude instead of spinning.
+_DELETE_ALL_MAX_ID_FAILURES = 3
+
 
 # Fields that hold runtime auth/connection objects and must be preserved.
 # These are non-serializable objects (e.g. AWSV4SignerAuth, RequestsHttpConnection)
@@ -3568,6 +3576,7 @@ class AsyncMemory(MemoryBase):
         total_deleted = 0
         total_errors = 0
         seen_ids: set = set()
+        failure_counts: dict = {}
         no_progress = 0
         for _ in range(_DELETE_ALL_MAX_ITERATIONS):
             listed = await asyncio.to_thread(
@@ -3596,12 +3605,23 @@ class AsyncMemory(MemoryBase):
             )
             for memory, result in zip(batch, results):
                 if isinstance(result, BaseException):
-                    # Do not mark seen on failure — leave residual IDs eligible
-                    # for retry on the next list page (async concurrent deletes
-                    # can fail without surfacing; marking early silently drops
-                    # them while delete_all still reports success).
+                    # Do not mark seen on a transient failure — leave residual
+                    # IDs eligible for retry on the next list page (async
+                    # concurrent deletes can fail without surfacing; marking
+                    # early silently drops them while delete_all still reports
+                    # success). But bound per-id retries: an id whose delete
+                    # keeps failing is marked seen after a few attempts so it
+                    # doesn't get re-listed and re-attempted every iteration.
                     total_errors += 1
+                    failure_counts[memory.id] = failure_counts.get(memory.id, 0) + 1
                     logger.warning("Failed to delete memory %s: %s", memory.id, result)
+                    if failure_counts[memory.id] >= _DELETE_ALL_MAX_ID_FAILURES:
+                        seen_ids.add(memory.id)
+                        logger.warning(
+                            "Giving up on memory %s after %d failed delete attempts",
+                            memory.id,
+                            failure_counts[memory.id],
+                        )
                 else:
                     seen_ids.add(memory.id)
                     total_deleted += 1
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -90,6 +90,14 @@ def _vector_store_list_rows(listed):
     return []
 
 
+# Page size for delete_all pagination. Vector stores typically default list()
+# to ~100 hits; delete_all must loop with an explicit top_k to drain all matches.
+_DELETE_ALL_PAGE_SIZE = 1000
+
+# Safety cap so eventual-consistency lag cannot spin forever.
+_DELETE_ALL_MAX_ITERATIONS = 10_000
+
+
 # Fields that hold runtime auth/connection objects and must be preserved.
 # These are non-serializable objects (e.g. AWSV4SignerAuth, RequestsHttpConnection)
 # needed by clients like OpenSearch — not sensitive strings to redact.
@@ -1885,14 +1893,28 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
-        memories = self.vector_store.list(filters=filters)[0]
-        for memory in memories:
-            self._delete_memory(memory.id)
+        # Paginate: most vector stores cap list() (often ~100). Delete each page
+        # inside the loop so the result set shrinks between list() calls.
+        total_deleted = 0
+        seen_ids: set = set()
+        for _ in range(_DELETE_ALL_MAX_ITERATIONS):
+            listed = self.vector_store.list(filters=filters, top_k=_DELETE_ALL_PAGE_SIZE)
+            batch = _vector_store_list_rows(listed)
+            batch = [m for m in batch if getattr(m, "id", None) not in seen_ids]
+            if not batch:
+                break
+            for memory in batch:
+                try:
+                    self._delete_memory(memory.id)
+                    seen_ids.add(memory.id)
+                    total_deleted += 1
+                except Exception as e:
+                    logger.warning("Failed to delete memory %s: %s", memory.id, e)
+                    seen_ids.add(memory.id)
 
-        logger.info(f"Deleted {len(memories)} memories")
+        logger.info(f"Deleted {total_deleted} memories")
 
-        decay_usage_notice = detect_decay_usage_from_delete_all(len(memories))
+        decay_usage_notice = detect_decay_usage_from_delete_all(total_deleted)
         if decay_usage_notice:
             display_decay_usage_notice(self, "sync", "delete_all", *decay_usage_notice)
         else:
@@ -3515,26 +3537,41 @@ class AsyncMemory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "async"})
-        memories = await asyncio.to_thread(self.vector_store.list, filters=filters)
+        # Paginate like sync: delete each page before asking for the next so
+        # stores without offset/cursor still drain fully.
+        total_deleted = 0
+        total_errors = 0
+        seen_ids: set = set()
+        for _ in range(_DELETE_ALL_MAX_ITERATIONS):
+            listed = await asyncio.to_thread(
+                self.vector_store.list, filters=filters, top_k=_DELETE_ALL_PAGE_SIZE
+            )
+            batch = _vector_store_list_rows(listed)
+            batch = [m for m in batch if getattr(m, "id", None) not in seen_ids]
+            if not batch:
+                break
 
-        delete_tasks = []
-        for memory in memories[0]:
-            delete_tasks.append(self._delete_memory(memory.id, skip_entity_cleanup=True))
-
-        results = await asyncio.gather(*delete_tasks, return_exceptions=True)
+            results = await asyncio.gather(
+                *[self._delete_memory(memory.id, skip_entity_cleanup=True) for memory in batch],
+                return_exceptions=True,
+            )
+            for memory, result in zip(batch, results):
+                seen_ids.add(memory.id)
+                if isinstance(result, BaseException):
+                    total_errors += 1
+                    logger.warning("Failed to delete memory %s: %s", memory.id, result)
+                else:
+                    total_deleted += 1
 
         if self._entity_store is not None:
             await self._bulk_clear_entity_store(filters)
 
-        errors = [r for r in results if isinstance(r, BaseException)]
-        if errors:
-            logger.warning("Failed to delete %d out of %d memories", len(errors), len(results))
-            for err in errors:
-                logger.warning("Delete error: %s", err)
+        if total_errors:
+            logger.warning("Failed to delete %d memories during delete_all", total_errors)
 
-        logger.info(f"Deleted {len(results) - len(errors)} memories")
+        logger.info(f"Deleted {total_deleted} memories")
 
-        decay_usage_notice = detect_decay_usage_from_delete_all(len(memories[0]))
+        decay_usage_notice = detect_decay_usage_from_delete_all(total_deleted)
         if decay_usage_notice:
             await display_decay_usage_notice_async(self, "async", "delete_all", *decay_usage_notice)
         else:
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -97,6 +97,12 @@ _DELETE_ALL_PAGE_SIZE = 1000
 # Safety cap so eventual-consistency lag cannot spin forever.
 _DELETE_ALL_MAX_ITERATIONS = 10_000
 
+# On stores with refresh lag (Elasticsearch/OpenSearch), a page can come back
+# fully composed of already-seen ids while later unseen rows are still hidden.
+# Allow a bounded number of such no-progress pages before concluding the drain
+# is complete, so stale pages don't cause us to silently leave data behind.
+_DELETE_ALL_MAX_NO_PROGRESS = 3
+
 
 # Fields that hold runtime auth/connection objects and must be preserved.
 # These are non-serializable objects (e.g. AWSV4SignerAuth, RequestsHttpConnection)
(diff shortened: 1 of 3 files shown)


Bartok9:
## CI fix

`build_mem0` failed on two existing `TestEntityIdValidation` cases that still asserted:

```python
vector_store.list(filters={...})  # no top_k
```

Pagination now always passes `top_k=_DELETE_ALL_PAGE_SIZE` (1000), so those mocks were strict-equality mismatches (2 failed / 1685 passed). Updated the trim + integer-coerce assertions to include `top_k=_DELETE_ALL_PAGE_SIZE`. Product path unchanged.

Local: entity-id delete_all cases + pagination suite green.

Agent-Owner: sera

cen-zp:
Thanks for carrying this forward and for the explicit salvage credit to #6465 and the earlier contributors. The current head fixes the known async ordering defect in production code, and the green Python matrix is encouraging.

I think one regression gap remains before this fully proves the fix. The new pagination tests use a list.side_effect with pre-canned pages that advance on every call regardless of whether deletion occurred. That means the async test can still terminate even if deletion is accidentally moved back outside the pagination loop—the same testing gap that allowed the infinite-list defect in #5950 to pass its targeted test.

Could we make at least the sync and async multi-page tests stateful? A small fake store can keep an internal list, always return the first N rows, and remove a row only when delete() runs. Then the async test becomes load-bearing: it hangs or hits a bounded failure if deletion does not occur before the next list() call.

Two related edge cases are worth covering:

1. After seen_ids filtering, an all-seen page currently causes an immediate break. With Elasticsearch/OpenSearch-style refresh lag, that stale page can temporarily hide later unseen records. A bounded consecutive no-progress retry/backoff policy plus a test where one stale page is followed by new IDs would avoid silently leaving data behind.
2. The sync path previously surfaced _delete_memory failures, while this revision catches every exception, marks the ID seen, and still returns the success message. Unless that API-contract change is intentional, preserving the existing sync failure semantics would avoid reporting success with residual memories.

Minor scope note: #4869 also describes the TypeScript deleteAll path, so Part of #4869 may be more accurate than Closes #4869 unless the TS half is being completed here or tracked separately.

I have not verified this against a hosted vector-store backend; these observations come from the patch behavior and stateful-fake reasoning.

Bartok9:
Thank you @cen-zp — this is exactly the kind of review that makes the fix real. All three points addressed in dc4e287:

**1. Stateful fake store (load-bearing tests).** Added `_StatefulStore` that returns at most `page_size` rows and only drops a row when `delete()` actually runs. Both the sync and async multi-page tests now use it, so if deletion is ever moved back outside the pagination loop, `list()` never shrinks — the loop runs to the iteration cap and the test hangs/fails instead of passing on pre-canned empty pages. This closes the same gap that let #5950's targeted test pass.

**2. Refresh-lag / stale all-seen page.** An all-seen page no longer breaks immediately. Both paths now retry a bounded number of no-progress pages (`_DELETE_ALL_MAX_NO_PROGRESS = 3`) before concluding the drain, so an Elasticsearch/OpenSearch stale page followed by newly-visible rows won't silently leave data behind. Added a test where a stale page precedes a previously-hidden id.

**3. Sync failure semantics.** You're right that swallowing every exception and still returning the success message was an unintended API-contract change. Reverted: sync `delete_all` now surfaces `_delete_memory` failures. Added a regression test asserting the exception propagates rather than reporting success with residual memories.

**Scope note.** Updated the description to **Part of #4869** — the TS `deleteAll` half is untouched here and stays tracked by the issue.

Not verified against a hosted backend either; the stateful-store tests are the closest proxy I can offer in CI. Happy to adjust the no-progress bound or make it configurable if you'd prefer.

cen-zp:
Thanks — dc4e287 addresses the stateful-test gap, restores the sync failure contract, and fixes the issue-closing scope exactly as requested.

One final point on the refresh-lag path: `_DELETE_ALL_MAX_NO_PROGRESS = 3` currently performs three immediate `list()` calls with no delay between them. For a backend whose visibility refresh takes seconds, all three retries can finish before the first refresh occurs, so the bounded counter alone may not provide meaningful tolerance in the real Elasticsearch/OpenSearch case even though the new canned stale-page test passes.

I would either add a small bounded backoff between no-progress attempts (`time.sleep` in sync and `await asyncio.sleep` in async, with the waits patched/asserted in unit tests), or narrow the PR description so it does not claim general refresh-lag tolerance. Making only the retry count configurable would not solve the timing issue by itself.

Apart from that timing detail, the changes directly address my review points. Thanks for the quick and careful follow-through.

Bartok9:
Thanks @cen-zp — good catch on the timing. Fixed in e8f5c8b4.

Added `_DELETE_ALL_NO_PROGRESS_BACKOFF` (0.5s, linear per attempt). On a no-progress page the sync path now `time.sleep`s and the async path `await asyncio.sleep`s before re-listing, so the three bounded retries actually span a multi-second refresh window instead of firing back-to-back. The retry count alone stays as the safety bound; the backoff supplies the timing tolerance.

Tests patch and assert the waits: the sync stale-page test asserts `time.sleep` is called once with a positive delay, and a new async test asserts `asyncio.sleep` is awaited once with a positive delay, both before the previously-hidden id becomes visible. Full suite green (8/8).

Happy to make the backoff configurable or tune the base delay if you'd prefer.

mem0/memory/main.py
```diff
@@ -3515,26 +3563,54 @@ async def delete_all(self, user_id=None, agent_id=None, run_id=None):
...
+                continue
+            no_progress = 0
 
-        results = await asyncio.gather(*delete_tasks, return_exceptions=True)
+            results = await asyncio.gather(
+                *[self._delete_memory(memory.id, skip_entity_cleanup=True) for memory in batch],
+                return_exceptions=True,
+            )
+            for memory, result in zip(batch, results):
+                seen_ids.add(memory.id)
```
kartik-mem0: `seen_ids.add(memory.id)` runs before checking `result`, so a failed delete (timeout, connection error) gets marked seen and is never retried. `delete_all` still returns `{"message": "Memories deleted successfully!"}` with that memory left in the store. Confirmed live on Qdrant: forcing one delete to fail left 1/110 seeded memories behind, 3 runs.

Move the `add` into the `else` branch below, matching sync's own stated contract two methods up (surface failures, don't report success with residuals left over). No test in this diff exercises an async partial-delete failure.

kartik-mem0 (commented):
## Changes requested

Paginates `delete_all()` correctly and adds refresh-lag tolerance, sync and async both drain every page now. But async's `seen_ids` gets marked before checking whether the delete actually succeeded, so a single failed delete silently drops that memory from the retry set while `delete_all` still reports success.

**To unblock:**
1. `mem0/memory/main.py:3598` — mark `seen_ids` only on success (detail in the inline comment)

Optional: PR #5950 fixes the same root cause (#4869) with a simpler loop and no `seen_ids`/backoff apparatus, so it doesn't hit this bug, but it also has no iteration cap or refresh-lag tolerance. This PR's own description says it extends that line of work, once the fix above lands, close #5950 in favor of this one.

- **Problem is real:** `mem0/memory/main.py:3593-3603`. Reproduced live against Qdrant in a local sandbox: seeded 110 memories, forced one `_delete_memory` call to fail mid-batch. `delete_all()` returned `{"message": "Memories deleted successfully!"}` and the failed memory was still present afterward, 3 runs, same result each time. Sync (`main.py:1934-1939`) does not have this bug: an exception there propagates out of `delete_all` before `seen_ids.add` is reached, matching this PR's own "sync failure semantics preserved" claim in the description. Async isn't covered by that claim and does the opposite.
- **Tests / red-green:** `tests/memory/test_delete_all_pagination.py` has `test_sync_delete_all_propagates_delete_failures` for sync but no async counterpart, this diff's own test suite does not fail on the bug above.
- **CI:** green (Python SDK 3.10/3.11/3.12, CI Gate, license/cla). Only `Vercel` is red, fork-auth, non-blocking.
- **Security:** cleared, no new external input, filters still go through `_validate_and_trim_entity_id` / `process_telemetry_filters` unchanged.
- **Bloat (ponytail):** Lean already. Ship. `_DELETE_ALL_MAX_ITERATIONS` bounds a `for` loop instead of a hand-rolled `while True`, and the no-progress/backoff pair covers real backend refresh lag, confirmed against this repo's own `mem0/vector_stores/opensearch.py`, not a speculative config knob.
- **Not verified:** pgvector/OpenSearch/MongoDB/Milvus lanes (only Qdrant), and the iteration-cap/backoff ceiling under sustained real refresh lag (the bug found doesn't require that path).

</details>

The pagination and refresh-lag design here is the right shape, better than #5950's plain loop. Push the one-line fix and I will re-review.

Bartok9:
Thanks @kartik-mem0 — sharp catch on the async `seen_ids` ordering.

Fixed in `efd4e088`:
- `seen_ids.add(memory.id)` now runs only in the success branch after a concurrent `_delete_memory` result is confirmed non-exception
- Failures still log + count `total_errors`, but the id stays eligible for the next `list()` page so a transient miss is retried instead of permanently dropped
- Early-mark path is gone; this matches the spirit of the sync contract you called out (no silent residual while reporting success for that id)

Regression: `test_async_delete_all_retries_failed_delete` — first delete fails, id still listed, second attempt succeeds (`await_count == 2`). Full pagination suite 9/9 green locally.

Ready for re-review. Happy to close #5950 in favor of this once you are happy with the tip.

Agent-Owner: sera

mem0/memory/main.py
```diff
@@ -3515,26 +3563,58 @@ async def delete_all(self, user_id=None, agent_id=None, run_id=None):
...
+                *[self._delete_memory(memory.id, skip_entity_cleanup=True) for memory in batch],
+                return_exceptions=True,
+            )
+            for memory, result in zip(batch, results):
+                if isinstance(result, BaseException):
+                    # Do not mark seen on failure — leave residual IDs eligible
+                    # for retry on the next list page (async concurrent deletes
+                    # can fail without surfacing; marking early silently drops
+                    # them while delete_all still reports success).
+                    total_errors += 1
```
kartik-mem0, line 3615: A memory whose delete keeps failing (not just once) gets retried every iteration with no backoff, since `no_progress` only tracks whole-page staleness, not a per-item repeat failure, so it hammers the backend until the 10,000-iteration cap.

Not blocking, pre-fix this silently dropped the memory instead. Worth a per-id retry cap or backoff if a real backend ever hits it.

kartik-mem0 (commented):

Verified against `b49e0fe0` (new head): the async delete-failure data-loss bug from the prior review is fixed, both sync and async now delete inside the pagination loop and only mark an id seen on success, so a failed delete is retried on the next page instead of being written off.

**Follow-up:** TS `deleteAll()` (`mem0-ts/src/oss/src/memory/index.ts:1705`) has the identical unpaginated-list bug and is still untouched, acknowledged in the PR description and tracked by #4869.
**Not verified:** oss-opensearch/oss-mongodb/oss-milvus lanes live (covered oss-qdrant and oss-pgvector x sync/async this cycle); Elasticsearch/Milvus `list()` confirmed to accept `top_k` by reading the signature, not exercised live.

Risk: low, self-contained bug fix, no public API change · Your time: ~5 min

- **Problem is real:** reproduces on `main` at `mem0/memory/main.py:1860` and `:3490` (both pre-fix `delete_all()`s are a single unpaginated `list()` call); confirmed live in sandbox against unmodified `main` (real Qdrant, 110 seeded rows): `FAIL (54.2s): delete_all(user_id=...) left 10/110 memories behind`
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_delete_all_pagination.py tests/test_main.py -v` -> 52 passed. `ruff check mem0/memory/main.py tests/memory/test_delete_all_pagination.py tests/test_main.py` -> clean
- **Red-green:** sandbox `pr6487_delete_all_multipage`: FAILs on `main` (10/110 residual), PASSes on this head, oss-qdrant x sync/async and oss-pgvector x sync/async all green (4/4). New scenario `pr6487_async_delete_retry` (one delete forced to fail once): FAILs on `main` (`delete_all` still reports success while the memory survives), PASSes on this head
- **CI:** CI Gate pass, Python SDK build pass on 3.10/3.11/3.12, changelog/check_changes/label/CLA all pass; only red check is Vercel preview auth, non-blocking per repo convention
- **Security:** cleared, no new external input surface, delete failures are logged via `logger.warning` rather than swallowed silently
- **Bloat (ponytail):** lean already, four named constants each do distinct work, sync/async duplicate loop structure but that matches this file's existing house style (no shared-helper convention exists here)
- **Linked issue:** #4869

</details>

Good fix, and the `_StatefulStore`/backoff tests are doing real work, not asserting the mechanism. Prior review's two asks are both in this diff.

mem0/memory/main.py
kartik-mem0: Verified against `b49e0fe0` (new head): fixed. `seen_ids.add(memory.id)` now only runs in the success branch (`main.py:3606`), so a failed delete stays eligible for retry on the next page instead of being written off. Confirmed live in sandbox (real Qdrant, one delete forced to fail once): FAILs on `main` with the memory surviving delete_all, PASSes on this head. New regression test `test_async_delete_all_retries_failed_delete` covers it (mock-level), plus the sandbox scenario is red-green.

Bartok9:
Thanks @kartik-mem0 — and thanks for the live red-green verification on the seen_ids ordering, much appreciated.

Addressed your non-blocking note in `f0f59a0e`. You're right that leaving failed ids eligible meant a *permanently*-failing delete would be re-listed and re-attempted every iteration up to the 10k cap with no per-item bound.

Added `_DELETE_ALL_MAX_ID_FAILURES = 3`:
- A transient single miss still stays eligible and retries next page (existing `test_async_delete_all_retries_failed_delete` unchanged).
- An id that keeps failing is counted and, after the cap, marked seen + logged (`Giving up on memory <id> after N failed delete attempts`) so the drain concludes instead of hammering the backend.
- New regression `test_async_delete_all_caps_persistent_failures`: a store that keeps returning the same undeletable row → `_delete_memory` is awaited exactly `_DELETE_ALL_MAX_ID_FAILURES` times, not thousands.

Full pagination suite 10/10 green locally. Ready for re-review.

mem0/memory/main.py
kartik-mem0: Verified against f0f59a0e: fixed. `_DELETE_ALL_MAX_ID_FAILURES = 3` (main.py:119) caps per-id retries: an id is marked seen and logged after 3 failed attempts instead of every iteration.

Confirmed live against real Qdrant: pre-fix (b49e0fe0) a permanently-failing delete was attempted 9,999 times before delete_all() gave up. Post-fix, exactly 3 attempts, then "Giving up on memory ... after 3 failed delete attempts". New test test_async_delete_all_caps_persistent_failures proven red (10,000 attempts without the cap) and green (3 attempts) against this same code. Sync doesn't need this, it already fails loud on first error with no retry loop to bound.

kartik-mem0 (commented):

Verified against `f0f59a0e` (new head): the non-blocking Minor left open last round (a persistently-failing delete hammering the backend up to 10,000 times) is now capped too, though it was never required for LGTM.

**Before you merge:** nothing new, this stays self-contained.
**Follow-up:** TS `deleteAll()` (`mem0-ts/src/oss/src/memory/index.ts:1705`) still has the original unpaginated bug, tracked by #4869, unchanged this round.
**Not verified:** pgvector/OpenSearch/MongoDB/Milvus for this specific cap (checked live only against real Qdrant this round; the underlying pagination itself was already proven on Qdrant and pgvector last round).

Risk: low, one bounded-retry addition on an already-reviewed path · Your time: ~3 min

- **Problem is real:** `main.py:3593-3611` pre-fix (`b49e0fe0`): a delete that fails every time is never marked seen, so it is re-listed and re-attempted every iteration up to the 10,000-iteration cap.
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/memory/test_delete_all_pagination.py tests/test_main.py -q` -> 53 passed
- **Red-green:** unit-level, cap logic stripped from the worktree -> `_delete_memory` awaited 10,000 times, new test fails; restored -> awaited exactly 3 times, passes. Live sandbox (real Qdrant, async, one target memory forced to always fail): `b49e0fe0` FAILs, attempted 9,999 times; `f0f59a0e` PASSes, capped at 3, logs "Giving up on memory ... after 3 failed delete attempts".
- **CI:** green (Python SDK 3.10/3.11/3.12, CI Gate, license/cla). Only `Vercel` red, fork-auth, non-blocking.
- **Security:** cleared, no new external input, the new counter is an in-memory dict keyed by an id the trusted vector store already returned.
- **Bloat (ponytail):** lean already. One constant mirroring the existing `_DELETE_ALL_MAX_NO_PROGRESS` pattern, one plain dict, no new abstraction.
- **Linked issue:** #4869

</details>

kartik-mem0 (commented):
## Updating my earlier LGTM — the code is still correct, my recommendation has changed

I signed off on this at `f0f59a0e` on 24 Jul. That verdict was about this PR on its own and it was fair — the head hasn't moved and the pagination logic still reads correct to me. (I queued a live drain against a hosted Supabase project for it today alongside the other three, but that host stopped resolving before this variant ran, so the numbers below for the other PRs have no counterpart here. Flagging that rather than implying a run I didn't get.)

What changed is the context, not the code. Three more PRs for #6627 opened on 28 Jul, after my sign-off, so the comparison I made was against a field of one. Having now read all six diffs and drained a live hosted Supabase project with three of them (`main` leaves 150 of 250 memories behind; #6636 clears 250/250 in two `list()` calls at `top_k=1000`; #6629 and #6640 also clear fully), I'm recommending the maintainers consolidate on #6636, and I want that on the record here rather than leaving a stale "ready for squash-merge" at the top of the thread.

### Why the ranking moved

The correctness is genuinely equivalent — your `seen_ids` tracking, bounded iteration count, per-id failure cap and backoff all terminate correctly on both full and partial delete failure, which is more than two of the competing PRs manage. The difference is cost:

- **+402/-22 versus +95/-23** for the same user-visible outcome. Five module constants (`_DELETE_ALL_PAGE_SIZE`, `_DELETE_ALL_MAX_ITERATIONS`, `_DELETE_ALL_MAX_NO_PROGRESS`, `_DELETE_ALL_NO_PROGRESS_BACKOFF`, `_DELETE_ALL_MAX_ID_FAILURES`) versus one.
- `seen_ids` accumulates every deleted id for the life of the call, so draining N memories retains N UUID strings. At the scale this bug is actually about (the report is people with far more than 100 memories) that's real, and it buys tolerance for a refresh-lag failure mode nobody in the issue has reported hitting.
- The refresh-lag machinery (`_DELETE_ALL_MAX_NO_PROGRESS` retries with backoff) is defensive depth for a condition that hasn't been observed. #6636's "stop when a page repeats" gets the same safety in three lines.

None of that is wrong, it's just more surface than #6627 needs, and every constant here is a knob someone has to reason about at 3am later.

### Not asking you to close this

You have two PRs open on this issue and which one you'd rather carry is your call, not mine. If you'd prefer to land this one, the path is trimming it toward the smaller shape rather than adding anything. And thank you for the async delete-failure catch from the first round — that was a real bug that the other PRs mostly ignore, and it's why #6636 having a guard on both paths was the first thing I checked.

kartik-mem0:
Thanks for this, and for being first to the issue.

#6636 was merged as 54328ffd and closes #6627. The approach here was the same one that landed; #6636 reached it in a smaller diff, which is the only reason it went in ahead of this.

Closing as a duplicate. Sorry it sat as long as it did.

Bartok9:
No worries at all — #6636 landing is the right outcome and the smaller diff is the better carry for this path. Thanks for the thorough reviews along the way (the async `seen_ids` and termination catches were genuinely useful), and glad the fix is on main now. 🙏

PR #6489, opened 2026-07-22
feat: add LanceDB vector store provider

### LINKED ISSUE #6488: Add LanceDB as an OSS vector store provider
### 🚀 The feature

Add a first-party `lancedb` vector store provider to the Mem0 Python OSS SDK so memories can be stored in a local LanceDB directory without running an external vector database (Qdrant/pgvector/etc.).

### Motivation, pitch

LanceDB is an embedded, file-based vector database. It is a good fit for:

- local development
- single-node self-hosted Mem0 deployments
- environments where installing/operating a separate vector DB is undesirable

This mirrors existing local providers such as FAISS, while using LanceDB’s native persistence model.

### Proposed config

```python
{
  "vector_store": {
    "provider": "lancedb",
    "config": {
      "collection_name": "memories",
      "path": "/tmp/lancedb",
      "embedding_model_dims": 1536,
      "distance": "cosine"  # cosine | l2 | dot
    }
  }
}
```

### Scope (v1)

- Implement `VectorStoreBase` methods for create/insert/search/update/delete/get/list/reset
- Register provider in factory + configs
- Optional dependency via `mem0ai[vector-stores]`
- Unit tests + docs page under `docs/components/vectordbs/dbs/lancedb.mdx`
- Simple equality / `in` metadata filters applied after vector recall

Out of scope for v1: complex `$and`/`$or` filter trees, multi-writer shared filesystem deployments, TypeScript SDK support.

suger-m:
## Linked Issue

Closes #6488

## Description

Adds LanceDB as an OSS vector store provider for local / single-node Mem0 deployments.

- New `lancedb` provider implementing `VectorStoreBase`
- Config model + factory/config registration
- Optional dependency: `lancedb` in the `vector-stores` extra, so users can install it with `pip install "mem0ai[vector-stores]"`
- Unit tests for config + factory registration + CRUD/search/filter/reset
- Docs page, overview card, docs nav (`docs.json`), and `docs/llms.txt` entry

v1 notes:
- Metadata filters support equality, list / `in`, wildcard, logical `AND` / `OR` / `NOT`, comparison, and string containment operators
- Filters are evaluated after vector recall for LanceDB 0.17 compatibility; the recall window expands until `top_k` matches or the table is exhausted
- Raw `AND` / `OR` / `NOT` aliases are accepted by `list()` / `get_all()` as well as processed `$and` / `$or` / `$not` filters
- Filter syntax and operand types are validated before scanning, including on empty tables; `NOT` also handles missing fields correctly
- `top_k=0` returns an empty result consistently from search and list
- Highly selective filters may scan the full table; this is intentional for correctness on LanceDB versions that cannot push arbitrary JSON payload predicates down into the vector query
- Persistence is tested across LanceDB API changes, including reopening existing collections, listing/deleting collections, and pagination over table names
- Existing collections reject embedding dimension mismatches early instead of reporting an incorrect dimension and failing later during insert/search
- `create_col()` only commits in-memory collection state after the target table opens/creates successfully
- `update()` validates vectors and payloads before modifying the table, and uses LanceDB's table update path so invalid updates do not delete existing records
- Payloads must be dictionaries or `None`; legacy rows whose JSON payload decodes to a non-dictionary are read as empty metadata with a warning

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Ran locally:

```bash
ruff check mem0/vector_stores/lancedb.py mem0/configs/vector_stores/lancedb.py tests/vector_stores/test_lancedb.py
ruff format --check mem0/vector_stores/lancedb.py mem0/configs/vector_stores/lancedb.py tests/vector_stores/test_lancedb.py
python -X utf8 scripts/check-llms-txt-coverage.py
git diff --check
python -m build --wheel
python -m pytest tests/vector_stores/test_lancedb.py -q
```

Result: `73 passed` against LanceDB 0.17, 0.25, 0.33, and 0.34; ruff clean; docs index and diff checks pass; wheel metadata includes `Provides-Extra: vector-stores` and `Requires-Dist: lancedb<1.0.0,>=0.17.0; extra == 'vector-stores'`.

## Checklist

- [x] I have performed a self-review of my changes

### DIFF
--- a/mem0/vector_stores/lancedb.py
+++ b/mem0/vector_stores/lancedb.py
@@ -16,6 +16,10 @@ from mem0.vector_stores.base import VectorStoreBase
 
 logger = logging.getLogger(__name__)
 
+_FILTER_OPERATORS = {"eq", "ne", "gt", "gte", "lt", "lte", "in", "nin", "contains", "icontains"}
+_LOGICAL_FILTER_ALIASES = {"AND": "$and", "OR": "$or", "NOT": "$not"}
+_LOGICAL_FILTER_OPERATORS = {"$and", "$or", "$not"}
+
 
 class OutputData(BaseModel):
     id: Optional[str]
@@ -89,19 +93,7 @@ class LanceDB(VectorStoreBase):
         if not isinstance(expected, dict):
             return actual == expected
 
-        supported_operators = {
-            "eq",
-            "ne",
-            "gt",
-            "gte",
-            "lt",
-            "lte",
-            "in",
-            "nin",
-            "contains",
-            "icontains",
-        }
-        unsupported = set(expected) - supported_operators
+        unsupported = set(expected) - _FILTER_OPERATORS
         if unsupported:
             raise ValueError(f"Unsupported filter operator(s): {sorted(unsupported)}")
 
@@ -133,23 +125,69 @@ class LanceDB(VectorStoreBase):
         return True
 
     @classmethod
+    def _validate_filters(cls, filters: Dict) -> None:
+        if not isinstance(filters, dict):
+            raise ValueError("Filters must be a dictionary")
+
+        for key, expected in filters.items():
+            if not isinstance(key, str):
+                raise ValueError("Filter field names must be strings")
+
+            logical_key = _LOGICAL_FILTER_ALIASES.get(key, key)
+            if logical_key in _LOGICAL_FILTER_OPERATORS:
+                if not isinstance(expected, list) or not all(isinstance(item, dict) for item in expected):
+                    raise ValueError(f"{logical_key} filter value must be a list of filter dictionaries")
+                for item in expected:
+                    cls._validate_filters(item)
+                continue
+
+            if key.startswith("$"):
+                raise ValueError(f"Unsupported logical filter operator: {key}")
+
+            if not isinstance(expected, dict):
+                continue
+            if not expected:
+                raise ValueError(f"Operator filter for field {key!r} must not be empty")
+
+            unsupported = set(expected) - _FILTER_OPERATORS
+            if unsupported:
+                raise ValueError(f"Unsupported filter operator(s): {sorted(unsupported)}")
+
+            for operator, value in expected.items():
+                if operator in {"eq", "ne", "gt", "gte", "lt", "lte"}:
+                    if isinstance(value, (dict, list, tuple, set)):
+                        raise ValueError(f"Filter operator {operator!r} requires a scalar value")
+                    if operator in {"gt", "gte", "lt", "lte"} and value is None:
+                        raise ValueError(f"Filter operator {operator!r} does not support null")
+                    continue
+
+                if operator in {"in", "nin"}:
+                    if not isinstance(value, (list, tuple)) or not value:
+                        raise ValueError(f"Filter operator {operator!r} requires a non-empty list")
+                    if any(isinstance(item, (dict, list, tuple, set)) for item in value):
+                        raise ValueError(f"Filter operator {operator!r} list items must be scalar values")
+                    continue
+
+                if not isinstance(value, str):
+                    raise ValueError(f"Filter operator {operator!r} requires a string value")
+
+    @classmethod
     def _apply_filters(cls, payload: Dict, filters: Optional[Dict]) -> bool:
         if not filters:
             return True
-        if not payload:
-            return False
 
         for key, expected in filters.items():
-            if key in {"$and", "$or", "$not"}:
+            logical_key = _LOGICAL_FILTER_ALIASES.get(key, key)
+            if logical_key in _LOGICAL_FILTER_OPERATORS:
                 if not isinstance(expected, list) or not all(isinstance(item, dict) for item in expected):
-                    raise ValueError(f"{key} filter value must be a list of filter dictionaries")
+                    raise ValueError(f"{logical_key} filter value must be a list of filter dictionaries")
 
                 matches = [cls._apply_filters(payload, item) for item in expected]
-                if key == "$and" and not all(matches):
+                if logical_key == "$and" and not all(matches):
                     return False
-                if key == "$or" and not any(matches):
+                if logical_key == "$or" and not any(matches):
                     return False
-                if key == "$not" and any(matches):
+                if logical_key == "$not" and any(matches):
                     return False
                 continue
 
@@ -221,6 +259,11 @@ class LanceDB(VectorStoreBase):
     def search(
         self, query: str, vectors: List[list], top_k: int = 5, filters: Optional[Dict] = None
     ) -> List[OutputData]:
+        if filters:
+            self._validate_filters(filters)
+        if top_k == 0:
+            return []
+
         query_vector = vectors[0] if vectors and isinstance(vectors[0], list) else vectors
         if not filters:
             rows = self.table.search(query_vector).metric(self.distance).limit(top_k).to_list()
@@ -292,6 +335,11 @@ class LanceDB(VectorStoreBase):
         }
 
     def list(self, filters: Optional[Dict] = None, top_k: int = 100) -> List[List[OutputData]]:
+        if filters:
+            self._validate_filters(filters)
+        if top_k == 0:
+            return [[]]
+
         rows = self.table.to_arrow().to_pylist()
         results: List[OutputData] = []
         for row in rows:
--- a/mem0/vector_stores/lancedb.py
+++ b/mem0/vector_stores/lancedb.py
@@ -1,0 +1,242 @@
+import json
+import logging
+import os
+import uuid
+from typing import Dict, List, Optional
+
+from pydantic import BaseModel
+
+try:
+    import lancedb
+    import pyarrow as pa
+except ImportError:
+    raise ImportError("The 'lancedb' library is required. Please install it using 'pip install lancedb'.")
+
+from mem0.vector_stores.base import VectorStoreBase
+
+logger = logging.getLogger(__name__)
+
+
+class OutputData(BaseModel):
+    id: Optional[str]
+    score: Optional[float]
+    payload: Optional[Dict]
+
+
+class LanceDB(VectorStoreBase):
+    def __init__(
+        self,
+        collection_name: str = "mem0",
+        path: Optional[str] = None,
+        embedding_model_dims: int = 1536,
+        distance: str = "cosine",
+    ):
+        self.collection_name = collection_name
+        self.path = path or "/tmp/lancedb"
+        self.embedding_model_dims = embedding_model_dims
+        self.distance = distance
+
+        os.makedirs(self.path, exist_ok=True)
+        self.client = lancedb.connect(self.path)
+        self.table = self.create_col(collection_name)
+
+    def _schema(self):
+        return pa.schema(
+            [
+                pa.field("id", pa.string()),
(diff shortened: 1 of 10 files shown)


kartik-mem0 (commented):
## Changes requested

Adds a LanceDB vector store provider (registration, config, docs, tests all wired correctly); the filter-translation layer silently drops results for OR/NOT/wildcard/range-operator queries.

**To unblock:**
1. `mem0/vector_stores/lancedb.py:82` - port pgvector's filter translation, or raise on unsupported filter shapes (detail in the inline comment)
2. `mem0/vector_stores/lancedb.py:166` - post-recall over-fetch window can silently under-return `top_k` (detail in the inline comment)
3. `tests/vector_stores/test_lancedb.py:104` - add coverage for OR/NOT/wildcard/operator filters so #1 does not regress silently

Optional: `update()`'s ValueError-on-missing-id diverges from pgvector/qdrant's silent no-op (lancedb.py:191)

- **Problem is real:** reproduced live against a real, non-mocked LanceDB table: `{"OR": [...]}`, `{"user_id": "*"}`, `{"score_val": {"gt": 6}}`, and `{"NOT": [...]}` all returned 0 rows over 3 rows with genuine matches. `mem0/memory/main.py:1545-1596` (`_process_metadata_filters`) confirms these exact shapes, `$or`/`$not`/literal `"*"`/raw operator dicts, are what `Memory.search()`/`get_all()` actually send to the vector store.
- **Tests / red-green:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_lancedb.py -v` -> 3 passed. All pre-existing tests still pass; they only cover plain equality, so they do not catch this.
- **CI:** CI Gate run 29932925761 is `action_required` with 0 jobs, the first-time-contributor approval gate blocked even the path-detection job, so ruff/pytest never ran on this PR. `label` and `license/cla` are the only real checks that ran; the Vercel failure is an unrelated deploy-auth check.
- **Security:** no new attack surface. `_id_where`/`_escape_sql_string` correctly quote-escape the only SQL `.where()` string built from user input (get/delete/update); `_apply_filters` is pure Python payload comparison, not SQL. Cleared.
- **Bloat (ponytail):** lean already, 236 lines, no unrequested abstractions.
- **Not verified:** entity-store search/dedup through this same filter path (structurally sound per `main.py:534-552`, not independently exercised end to end); behavior under concurrent multi-process writers to the same table (docs already disclose single-writer).

</details>

Provider registration (base contract, `configs.py`, `factory.py`, pyproject optional dependency group, docs) is clean and matches sibling patterns. Push when ready and I will re-review.

mem0/vector_stores/lancedb.py
```diff
@@ -0,0 +1,236 @@
...
+        if not payload:
+            return {}
+        try:
+            return json.loads(payload)
+        except json.JSONDecodeError:
+            logger.warning("Failed to decode LanceDB payload JSON")
+            return {}
+
+    @staticmethod
+    def _apply_filters(payload: Dict, filters: Optional[Dict]) -> bool:
```
kartik-mem0: `_apply_filters` silently returns 0 rows for OR/NOT filters, wildcard `"*"`, and any operator besides `eq`/`in` (`ne`/`gt`/`gte`/`lt`/`lte`/`nin`/`contains`/`icontains`), all reachable from `Memory.search()`/`get_all()` via `_process_metadata_filters`. Reproduced live: OR/wildcard/gt/NOT over 3 rows with real matches all returned 0 rows, no error. Port pgvector's `_build_filter_conditions` (pgvector.py:50-115), the reference translation.

mem0/vector_stores/lancedb.py
```diff
@@ -0,0 +1,236 @@
...
+            for vector_id, vector, payload in zip(ids, vectors, payloads)
+        ]
+        self.table.add(rows)
+        logger.info("Inserted %s vectors into LanceDB table %s", len(rows), self.collection_name)
+
+    def search(
+        self, query: str, vectors: List[list], top_k: int = 5, filters: Optional[Dict] = None
+    ) -> List[OutputData]:
+        query_vector = vectors[0] if vectors and isinstance(vectors[0], list) else vectors
+        fetch_k = max(top_k * 5, 50) if filters else top_k
```
kartik-mem0: Post-recall filtering over-fetches `max(top_k*5, 50)` rows before filtering in Python, so `search()` can silently return fewer than `top_k` matches once more than that many rows get filtered out, unlike qdrant/pgvector which push filters into the index/query itself. Push the eq/in cases into a `.where()` predicate instead.

tests/vector_stores/test_lancedb.py
```diff
@@ -0,0 +1,104 @@
...
+
+        store.insert(
+            vectors=[[1.0, 0.0, 0.0], [0.0, 1.0, 0.0]],
+            ids=["perfect", "orthogonal"],
+        )
+
+        results = store.search(query="", vectors=[1.0, 0.0, 0.0], top_k=2)
+        assert results[0].id == "perfect"
+        assert results[0].score > results[1].score
+        assert results[0].score == pytest.approx(1.0)
```
kartik-mem0, line 289: Only plain single-key equality is tested. None of OR/NOT, wildcard `"*"`, or the 8 comparison operators (`gt`/`gte`/`lt`/`lte`/`ne`/`nin`/`contains`/`icontains`) have a test, so the `_apply_filters` gap above ships green. Add a case per shape, asserting non-empty results where matches exist.

mem0/vector_stores/lancedb.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def update(
+        self,
+        vector_id: str,
+        vector: Optional[List[float]] = None,
+        payload: Optional[Dict] = None,
+    ):
+        existing = self.get(vector_id)
+        if existing is None:
+            raise ValueError(f"Vector {vector_id} not found")
```
kartik-mem0: `update()` raises `ValueError` on a missing `vector_id`; pgvector's and qdrant's `update()` are silent no-ops for the same case. Not unsafe, but a caller that codes to the cross-provider contract gets an unexpected exception on this backend only.

mem0/vector_stores/lancedb.py
suger-m: Addressed in `d8fc0cf5`. LanceDB 0.17 cannot query fields inside the current JSON-string payload and does not support a Map payload schema, so native `.where()` translation would require a storage-schema redesign. The provider now evaluates the complete processed filter contract in Python: `$and` / `$or` / `$not`, wildcard existence, list shorthand, `eq` / `ne` / `gt` / `gte` / `lt` / `lte` / `in` / `nin` / `contains` / `icontains`, and raises `ValueError` for unknown operators instead of silently returning zero rows.

mem0/vector_stores/lancedb.py
suger-m: Addressed in `d8fc0cf5`. A fixed over-fetch factor was removed. Because LanceDB 0.17 cannot prefilter arbitrary fields inside the JSON-string payload, search now expands the recall window exponentially until it has `top_k` filtered matches or reaches the table row count. This removes the silent under-return bug while preserving the existing arbitrary-metadata schema. The docs explicitly disclose that highly selective filters may scan the full table.

tests/vector_stores/test_lancedb.py
suger-m: Added real LanceDB coverage in `d8fc0cf5` for OR, NOT, wildcard existence, every documented comparison/string operator, list shorthand, logical conjunction, `list()` filtering, unknown-operator errors, and the regression where 55 nearer non-matches previously hid valid matches beyond the fixed recall window. The focused suite passes 22 tests on LanceDB 0.17 and 0.33.

mem0/vector_stores/lancedb.py
suger-m: Changed in `d8fc0cf5`: `update()` now returns without error when the ID is missing, matching the no-op behavior of the sibling providers. A regression test covers it.

mem0/vector_stores/lancedb.py
kartik-mem0: Verified against 93d6c6cc: fixed. `_apply_filters` (lines 226-254) now recursively handles `$and`/`$or`/`$not` and delegates per-field matching to `_matches_filter_value` (lines 137-176), which handles wildcard "*" and all 10 operators. Re-ran the exact repro from this finding against the new head: OR returns 3/3, wildcard returns 3/3, `gt` returns m2+m3, `NOT` returns m2 only (all previously returned 0). Cross-checked semantics against `main.py::_process_metadata_filters` and pgvector's `_build_filter_conditions`: De Morgan for NOT and missing-field-excludes behavior match. 15 new parametrized tests cover this.

mem0/vector_stores/lancedb.py
kartik-mem0: Verified against 93d6c6cc: fixed. `search()` (lines 335-356) now starts `fetch_k` at `min(max(top_k*5,50), table_size)` and doubles it until enough filtered matches are found or `fetch_k >= table_size`, so it can no longer silently under-return. The loop is bounded and monotonic, terminates correctly. `test_lancedb_search_expands_recall_until_top_k_filtered_results_are_found` (55 non-matching rows ranked ahead of 3 matches) passes. Docs now disclose the full-table-scan worst case for highly selective filters.

tests/vector_stores/test_lancedb.py
kartik-mem0: Verified against 93d6c6cc: fixed. New parametrized suite `test_lancedb_search_supports_processed_metadata_filters` (15 cases: OR, NOT, AND via multi-key, wildcard, all 10 operators, list shorthand) plus dedicated tests for recall-window expansion and update-missing-id no-op. Ran the full suite fresh in this session against real LanceDB 0.36.0 (not mocked): 73 passed, ruff clean.

mem0/vector_stores/lancedb.py
kartik-mem0: Verified against 93d6c6cc: fixed. `update()` (lines 361-379) now returns early as a no-op when `get(vector_id)` finds nothing, matching pgvector/qdrant. `test_lancedb_update_missing_id_is_a_noop` covers it and passes.

kartik-mem0 (commented):
Re-reviewed at 93d6c6cc. All 4 prior findings are fixed, verified with fresh evidence this session (replies posted on each original thread):

1. Critical (`_apply_filters` dropping OR/NOT/wildcard/operators): fixed. Re-ran the exact original repro against this head, OR/wildcard/gt/NOT all now match correctly (previously all returned 0 rows).
2. Important (fixed over-fetch window under-returning results): fixed with an exponential recall-window doubling, bounded by table size, terminates correctly.
3. Important (test coverage gap): fixed, 73 tests now pass against real LanceDB 0.36.0 (not mocked), including 15 new parametrized filter-operator cases.
4. Minor (`update()` raised on missing id): fixed, now a silent no-op matching pgvector/qdrant.

Also checked: ruff clean, no unintended changes in factory/config registration, filter semantics cross-checked against `main.py::_process_metadata_filters` and pgvector's reference implementation (match, including NOT and missing-field-exclusion edge cases).

One non-blocking note: `_validate_filters` and `_apply_filters` both recursively walk the same filter tree independently. Not a bug, just duplicated traversal; could be collapsed later if it becomes a maintenance burden.

No remaining code-quality blockers. Marking as needs-human: this is a new backend, adoption is a product call, not mine to approve.

PR #6490, opened 2026-07-22
fix: replace input()/sys.exit(1) with ImportError on missing optional deps (#6118)

rkfshakti:
The ollama and litellm import guards used input() to prompt the user
interactively and sys.exit(1) to abort the process. This makes the
library unusable in non-interactive environments (CI, pytest, FastAPI
servers) and terminates the host application without giving callers a
chance to handle the missing dependency gracefully.

Replace both with a plain ImportError that includes the install
instruction, matching standard Python library convention.

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -1,5 +1,3 @@
-import subprocess
-import sys
 from typing import Literal, Optional
 
 from mem0.configs.embeddings.base import BaseEmbedderConfig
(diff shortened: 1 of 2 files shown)


rkfshakti:
Hi maintainers — this PR replaces  with  on missing optional dependencies, which is a cleaner pattern that doesn't block automated testing. CLA is signed and all checks are passing. Would appreciate a review when you have a moment. Thanks!

rkfshakti:
Friendly ping — this PR has been open for a few days. The fix replaces input()/sys.exit(1) with a clean ImportError when optional dependencies are missing, so mem0 can be imported in environments where not all extras are installed without triggering an interactive prompt. Would appreciate a review when time allows.

rkfshakti:
Hi maintainers — following up again on this fix for #6118. Replaces input()/sys.exit() with proper error handling when optional dependencies are missing. CI is passing. Would appreciate a review when time allows. Thanks!

rkfshakti:
Friendly ping — this PR has been open for 10 days. Would appreciate a review when time allows.

kartik-mem0:
Closing as a duplicate of #6119, which was opened first and covers the same change. I compared the diffs line by line and the source changes are substantively identical, so nothing is lost here: #6119 stays open and is the one that will get reviewed and merged.

This is not a comment on the quality of your patch. Thank you for the time you put into it, and apologies that our review queue let two people do the same work. If you believe this PR covers something #6119 does not, reply here and I will reopen it.

PR #6495, opened 2026-07-22
fix(ts-oss): carry updatedAt into inferred add ADD history events

### LINKED ISSUE #6494: TS OSS inferred add() records the ADD history event with updated_at NULL
### Where

`mem0-ts/src/oss/src/memory/index.ts`, the inferred (default `infer: true`) add pipeline that batches history.

### What happens

When a memory is created, the stored payload carries both `createdAt` and `updatedAt` set to the same timestamp:

```ts
const memPayload: Record<string, any> = {
  ...metadata,
  data: text,
  textLemmatized,
  hash: memHash,
  createdAt: now,
  updatedAt: now,
};
```

But the ADD history record built for the same memory hardcodes `updatedAt` to `undefined`:

```ts
const historyRecords = records.map((r) => ({
  memoryId: r.memoryId,
  previousValue: null,
  newValue: r.text,
  action: "ADD",
  createdAt: r.payload.createdAt,
  updatedAt: undefined, // <- dropped
  isDeleted: 0,
}));
```

`batchAddHistory()` stores `record.updatedAt ?? null`, so the ADD row lands with `updated_at = NULL`. The two one-by-one `addHistory()` fallbacks in the same block also omit the argument. Net result: `history(memoryId)` returns an ADD event whose `updated_at` is null while the memory itself has `updatedAt` set, so the audit trail disagrees with the record it describes.

### Expected

An ADD event carries `updatedAt` equal to `createdAt` (both are set on the payload at creation), matching what the Python `_create_memory` does when it records history (`updated_at=new_metadata.get("updated_at")`).

### Repro

`add()` a memory with inference on (the default), then read `history(id)`: the ADD entry has `updated_at: null` instead of the creation timestamp.

abhay-codes07:
### Description

The inferred (default `infer: true`) add pipeline stores each memory with `createdAt` and `updatedAt` set to the same creation timestamp, but built the ADD history record with `updatedAt` hardcoded to `undefined`, and the two one-by-one `addHistory()` fallbacks in the same block omitted the argument entirely. `batchAddHistory()` persists `updatedAt ?? null`, so the ADD event landed with `updated_at` NULL while the stored memory had it set. `history(id)` then returned an audit trail that disagreed with the record it describes.

The fix passes the payload's `updatedAt` (equal to `createdAt` at creation) through to the batch record and both fallbacks, matching what the Python `_create_memory` records for its ADD history event.

Closes #6494

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/memory.add.test.ts` (run from the `mem0-ts` root), 12 passed.

- Added `records the ADD history event with updatedAt set to createdAt`: adds a memory with inference on, reads `history(id)`, and asserts the ADD entry's `updated_at` equals its `created_at` (both non-null). Reverting the source change fails exactly this test with `updated_at` coming back `null`.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1055,7 +1055,10 @@ export class Memory {
       newValue: r.text as string | null,
       action: "ADD",
       createdAt: r.payload.createdAt as string | undefined,
-      updatedAt: undefined as string | undefined,
+      // An ADD carries updatedAt equal to createdAt (both set on the payload
+      // above). Leaving it undefined recorded the ADD event with updated_at
+      // NULL while the stored memory had it set, so the two disagreed.
+      updatedAt: r.payload.updatedAt as string | undefined,
       isDeleted: 0,
     }));
 
@@ -1072,6 +1075,7 @@ export class Memory {
               hr.newValue,
               "ADD",
               hr.createdAt,
+              hr.updatedAt,
             );
           } catch (e) {
             console.error(`Failed to add history for ${hr.memoryId}: ${e}`);
--- a/mem0-ts/src/oss/tests/memory.add.test.ts
+++ b/mem0-ts/src/oss/tests/memory.add.test.ts
@@ -191,4 +191,20 @@ describe("Memory - add()", () => {
       expect.objectContaining({ event: "ADD" }),
     );
   });
+
+  test("records the ADD history event with updatedAt set to createdAt", async () => {
+    const result: SearchResult = await memory.add("I deploy on Fridays", {
+      userId,
+    });
+    const memoryId = result.results[0].id;
+
+    const history = await memory.history(memoryId);
+    const addEntry = history.find((h: any) => h.action === "ADD");
+    expect(addEntry).toBeDefined();
+    // A freshly added memory carries updatedAt == createdAt, so its ADD history
+    // event must record the same. It used to be written with updated_at NULL
+    // while the stored memory had it set, so the two disagreed.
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss fix with a test. The inferred add path stored memories with updatedAt set but recorded the ADD history event with updated_at NULL, so history disagreed with the record. Now carries updatedAt (equal to createdAt) through the batch record and both addHistory fallbacks, matching the Python _create_memory. Quick look when you get a chance, thanks.

kartik-mem0:
Thanks for turning this around so quickly, and apologies for closing it after the work you put in, especially since I had already told you on the issue that the review was coming.

The change does what the issue asked for, but we have decided to align the two paths the other way, so I am closing this alongside #6494.

Short version: the default inferred pipeline is single-pass ADD-only (https://mem0.ai/research), so an ADD event has no meaningful `updatedAt`. Keeping it unset until a real `update()` or `delete()` happens is the semantics we want to preserve, because it is the only way to distinguish "never revised" from "revised at creation time". Full reasoning is in #6494.

The test you added is the right shape, and it will be a useful reference if we make the reverse change. Thanks again.

abhay-codes07:
That makes sense, and thanks for the detailed write-up on all three. Keeping updated_at unset until a real update() or delete() is the cleaner invariant since it lets you tell never-revised apart from revised-at-creation, and I agree that puts the infer:false path already correct and leaves the inferred path as the one to bring in line by dropping the creation-time updatedAt. No worries at all about the close. I will follow that contract going forward, and happy to help with the reverse alignment if it is useful.

PR #6503, opened 2026-07-22
fix(pgvector): drop unused vector column from get() and list() queries

### LINKED ISSUE #6482: PGVector get() and list() fetch the vector column but never use it

Diwak4r:
get() and list() were SELECTing id, vector, payload but never reading the vector column. Drop it from both queries to avoid unnecessary IO.

Closes https://github.com/mem0ai/mem0/issues/6482

### DIFF
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -120,7 +120,10 @@ def remove_code_blocks(content: str) -> str:
     - The function uses a regex pattern to match code blocks that may start with ``` followed by an optional language tag (letters or numbers) and end with ```.
     - If a code block is detected, it returns only the inner content, stripping out the markers.
     - If no code block markers are found, the original content is returned as-is.
+    - Returns None unchanged when content is None (e.g. safety-refusal responses).
     """
+    if content is None:
+        return content
     pattern = r"^```[a-zA-Z0-9]*\n([\s\S]*?)\n```$"
     match = re.match(pattern, content.strip())
     match_res=match.group(1).strip() if match else content.strip()
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -459,13 +459,13 @@ class PGVector(VectorStoreBase):
         self._ensure_collection()
         with self._get_cursor() as cur:
             cur.execute(
-                sql.SQL("SELECT id, vector, payload FROM {} WHERE id = %s").format(self._col()),
+                sql.SQL("SELECT id, payload FROM {} WHERE id = %s").format(self._col()),
                 (vector_id,),
             )
             result = cur.fetchone()
             if not result:
                 return None
-            return OutputData(id=str(result[0]), score=None, payload=result[2])
+            return OutputData(id=str(result[0]), score=None, payload=result[1])
 
     def list_cols(self) -> List[str]:
         """
@@ -528,7 +528,7 @@ class PGVector(VectorStoreBase):
         with self._get_cursor() as cur:
             cur.execute(
                 sql.SQL("""
-                SELECT id, vector, payload
+                SELECT id, payload
                 FROM {}
                 {}
                 LIMIT %s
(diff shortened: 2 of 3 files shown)


mem0/memory/utils.py
```diff
@@ -120,7 +120,10 @@ def remove_code_blocks(content: str) -> str:
     - The function uses a regex pattern to match code blocks that may start with ``` followed by an optional language tag (letters or numbers) and end with ```.
     - If a code block is detected, it returns only the inner content, stripping out the markers.
     - If no code block markers are found, the original content is returned as-is.
+    - Returns None unchanged when content is None (e.g. safety-refusal responses).
     """
+    if content is None:
+        return content
```
kartik-mem0, line 126: This guard is unrelated to #6482 and duplicates the identical fix already open in #6175 (same author), with no test added here. Please drop this hunk from this PR, it belongs in #6175.

kartik-mem0 (commented):
## Changes requested

Pgvector fix (drop the unused `vector` column from `get()`/`list()`) is correct and complete; the bundled `remove_code_blocks` guard is unrelated and should come out.

**To unblock:**
1. `mem0/memory/utils.py:125` — drop this hunk, duplicates open PR #6175 (detail inline)

- **Problem is real:** `mem0/vector_stores/pgvector.py:462` (get) and `:531` (list) select the unused `vector` column; sandbox on `main` FAILs both sync and async, PASSes on this PR both modes
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_pgvector.py -v` (83 passed)
- **Red-green:** reverting `pgvector.py` to `main` (tests kept) turns 10 tests RED (`IndexError`); restoring the PR's file turns them GREEN again
- **CI:** label check green; CI Gate is `action_required` (queued, not yet run), not red
- **Security:** cleared, static SQL column list, no injection or tenancy surface touched
- **Bloat (ponytail):** net: -4 lines possible, drop the utils.py hunk
- **Not verified:** live Postgres beyond the sandbox container; TS `pgvector.ts` already selects `id, payload` only, unaffected

</details>

Real work here, @Diwak4r: the pgvector half is clean and well tested. Note #6483 fixes the identical root cause without the extra file. Push a trimmed diff and I will re-review.

kartik-mem0:
closing this in favor of #6483 

thank you for your contribution here

PR #6508, opened 2026-07-22
fix(ts-oss/milvus): skip '*' wildcard filter values instead of matching literally

### LINKED ISSUE #6507: TS OSS Milvus store treats a '*' wildcard filter as a literal, recalling nothing
### Where

`mem0-ts/src/oss/src/vector_stores/milvus.ts`, `createFilter()`.

### What happens

A `"*"` filter value means match-any, and the chroma and pinecone TS stores both skip it (chroma.ts drops the clause, pinecone.ts does `if (value === "*") continue;`). Milvus does not:

```ts
if (typeof value === "string") {
  const escaped = value.replace(/\/g, "\\\\").replace(/"/g, '\\"');
  operands.push(`(metadata["${key}"] == "${escaped}")`);
}
```

So `{ user_id: "*" }` becomes `(metadata["user_id"] == "*")`, which matches no memory (none has the literal id `"*"`), and the search silently returns nothing.

### Expected

Skip a `"*"` value so it does not constrain the query, matching the chroma/pinecone stores and the Python Milvus provider, which fixed exactly this in #6186 / #6187.

### Note

This is the TS counterpart of #6187; the Python side was fixed there but the TS Milvus store was not updated.

abhay-codes07:
### Description

A `"*"` filter value means match-any. The chroma and pinecone TS stores both skip it, but the Milvus store built `(metadata["key"] == "*")`, which matches no memory (none has the literal id `"*"`), so a wildcard filter silently recalled nothing.

The fix skips a `"*"` value in `createFilter()` so it does not constrain the query, matching the chroma/pinecone stores and the Python Milvus provider fixed in #6187.

Closes #6507

### Type of change

### How Has This Been Tested?

`jest src/oss/src/tests/milvus.test.ts` (run from the `mem0-ts` root), 25 passed.

- Added `skips a wildcard '*' filter value and keeps the rest`: `{ user_id: "*", agent_id: "a1" }` now yields `(metadata["agent_id"] == "a1")` with the wildcard dropped.
- Added `omits the filter entirely when every value is a wildcard`: `{ user_id: "*" }` falls back to the match-all `""` filter.
- Reverting the source change fails both with the old literal `(metadata["user_id"] == "*")` clause.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/tests/milvus.test.ts
+++ b/mem0-ts/src/oss/src/tests/milvus.test.ts
@@ -267,6 +267,36 @@ describe("Milvus vector store (TS OSS SDK)", () => {
     });
   });
 
+  it("skips a wildcard '*' filter value and keeps the rest", async () => {
+    const client = new FakeMilvusClient({ existing: ["mem0"] });
+    client.searchResponse = { results: [] };
+    const store = makeStore(client, { metricType: "COSINE" });
+    await store.initialize();
+
+    // "*" means match-any: it must be dropped, not emitted as `== "*"` (which
+    // matches nothing), leaving only the real agent_id clause.
+    await store.search([0.1, 0.2, 0.3], 5, {
+      user_id: "*",
+      agent_id: "a1",
+    });
+
+    const searchCall = client.calls.find((c) => c.method === "search")!;
+    expect(searchCall.args.filter).toBe('(metadata["agent_id"] == "a1")');
+  });
+
+  it("omits the filter entirely when every value is a wildcard", async () => {
+    const client = new FakeMilvusClient({ existing: ["mem0"] });
+    const store = makeStore(client);
+    await store.initialize();
+
+    await store.list({ user_id: "*" });
+
+    // All clauses dropped, so list() falls back to its match-all "" filter
+    // rather than a literal `(metadata["user_id"] == "*")` that matches nothing.
+    const queryCall = client.calls.filter((c) => c.method === "query").pop()!;
+    expect(queryCall.args.filter).toBe("");
+  });
+
   it("normalises L2 distances into a 0..1 similarity score", async () => {
     const client = new FakeMilvusClient({ existing: ["mem0"] });
     client.searchResponse = {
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss milvus fix with tests. TS counterpart of #6187: a '*' filter value was emitted as a literal (metadata[key] == "*") clause, so wildcard filters recalled nothing. Now skipped like the chroma and pinecone stores already do. Quick look when you get a chance, thanks.

kartik-mem0 (commented):

The TS Milvus store's `createFilter()` was emitting a literal `(metadata["key"] == "*")` clause for the internal `"*"` match-any sentinel, which matches zero rows; this drops the clause instead, matching `chroma.ts:278`, `pinecone.ts:185`, and the Python fix in #6187.

**Follow-up:** none, this was the last store still missing the guard.
**Not verified:** nothing, verified end to end including live Milvus.

Risk: low, single-line guard mirroring an already-shipped pattern in 3 other files · Your time: ~2 min

- **Problem is real:** reproduced on `main` at `mem0-ts/src/oss/src/vector_stores/milvus.ts:230-252` (pre-PR `createFilter`), and live against real Milvus: sandbox FAIL on `main@14393b59`, "expected all 5 seeded rows back with a wildcard source filter, got 0"
- **Tests:** `npx jest oss/src/tests/milvus.test.ts` -> 25 passed, 25 total
- **Red-green:** reverted `milvus.ts` to `main` only, the 2 new tests failed reproducing the exact literal `== "*"` clause; restored, 25/25 passed again
- **Live sandbox:** PASS on PR head `1f9d9a3a`, FAIL on `main@14393b59` with the exact reported symptom, both runs this session
- **CI:** all real required checks pass (CI Gate, TS SDK build/integration on Node 20/22, changelog, license/cla, label); only the advisory Vercel deploy-preview check is red, not merge-blocking
- **Security:** cleared, `"*"` is an internal match-any sentinel built by the memory layer itself (`mem0-ts/src/oss/src/memory/index.ts:2060,2082-2083`), not attacker-reachable input, so dropping the clause is not a tenancy bypass
- **Bloat (ponytail):** lean already, ship
- **Linked issue:** #6507

</details>

kartik-mem0 (approved):

PR #6511, opened 2026-07-22
fix(ts-oss/cassandra): apply every operator in a compound field filter

### LINKED ISSUE #6510: TS OSS Cassandra store applies only the first operator in a compound field filter
### Where

`mem0-ts/src/oss/src/vector_stores/cassandra.ts`, `matchFieldCondition()`.

### What happens

Cassandra has no server-side metadata filter, so `search()`/`list()` scan rows and apply filters in-app. The matcher checks each operator with its own early `return`:

```ts
if ("gte" in value) {
  return payloadValue >= value.gte;
}
if ("lte" in value) {
  return payloadValue <= value.lte;
}
```

So a compound condition like `{ age: { gte: 10, lte: 20 } }` returns as soon as `gte` is evaluated and never checks `lte`. A row with `age: 25` passes a `gte: 10, lte: 20` filter, and the upper bound is silently ignored. The same applies to any field carrying more than one operator.

### Expected

All operators present in a compound condition must hold (AND). The databricks store in this same SDK already does this correctly with a `sawOperator` flag that checks every operator and returns `false` on the first failure, so the fix is to bring the Cassandra matcher in line with it.

### Impact

Any Cassandra-backed range or multi-operator metadata filter returns rows outside the requested range.

abhay-codes07:
### Description

Cassandra has no server-side metadata filter, so `search()`/`list()` scan rows and apply filters in-app via `matchFieldCondition()`. Each operator had its own early `return`, so a compound condition like `{ age: { gte: 10, lte: 20 } }` returned after evaluating `gte` and never checked `lte`. A row with `age: 25` then passed a `gte: 10, lte: 20` filter, silently ignoring the upper bound.

The fix checks every operator present and fails on the first that does not hold (AND semantics), mirroring the databricks store in this same SDK, which already implements the matcher this way with a `sawOperator` flag.

Closes #6510

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/cassandra.unit.test.ts` (run from the `mem0-ts` root), 3 passed. This adds the first Cassandra unit test file, driving the matcher through the public `list()` API with an injected fake client.

- `applies every operator in a compound range filter (not just the first)`: with rows aged 5/15/25, `{ age: { gte: 10, lte: 20 } }` now returns only the age-15 row. Reverting the source change fails this, letting the age-25 row through.
- Single-operator and plain-equality filters are covered to confirm no regression.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/cassandra.ts
+++ b/mem0-ts/src/oss/src/vector_stores/cassandra.ts
@@ -455,44 +455,64 @@ export class CassandraDB implements VectorStore {
       return value.includes(payloadValue);
     }
 
+    // Every operator present in a compound condition must hold (AND), so check
+    // them all instead of returning on the first match. Returning early meant a
+    // range like { gte: 10, lte: 20 } only applied `gte`. Mirrors the databricks
+    // store's matcher.
+    let sawOperator = false;
+
     if ("eq" in value) {
-      return payloadValue === value.eq;
+      sawOperator = true;
+      if (payloadValue !== value.eq) return false;
     }
     if ("ne" in value) {
-      return payloadValue !== value.ne;
+      sawOperator = true;
+      if (payloadValue === value.ne) return false;
     }
     if ("gt" in value) {
-      return payloadValue > value.gt;
+      sawOperator = true;
+      if (!(payloadValue > value.gt)) return false;
     }
     if ("gte" in value) {
-      return payloadValue >= value.gte;
+      sawOperator = true;
+      if (!(payloadValue >= value.gte)) return false;
     }
     if ("lt" in value) {
-      return payloadValue < value.lt;
+      sawOperator = true;
+      if (!(payloadValue < value.lt)) return false;
     }
     if ("lte" in value) {
-      return payloadValue <= value.lte;
+      sawOperator = true;
+      if (!(payloadValue <= value.lte)) return false;
     }
     if ("in" in value) {
-      return Array.isArray(value.in) && value.in.includes(payloadValue);
+      sawOperator = true;
+      if (!Array.isArray(value.in) || !value.in.includes(payloadValue))
+        return false;
     }
     if ("nin" in value) {
-      return !Array.isArray(value.nin) || !value.nin.includes(payloadValue);
+      sawOperator = true;
+      if (Array.isArray(value.nin) && value.nin.includes(payloadValue))
+        return false;
     }
     if ("contains" in value) {
-      return (
-        typeof payloadValue === "string" &&
-        payloadValue.includes(value.contains)
-      );
+      sawOperator = true;
+      if (
+        typeof payloadValue !== "string" ||
+        !payloadValue.includes(value.contains)
+      )
+        return false;
     }
     if ("icontains" in value) {
-      return (
-        typeof payloadValue === "string" &&
-        payloadValue.toLowerCase().includes(value.icontains.toLowerCase())
-      );
+      sawOperator = true;
+      if (
+        typeof payloadValue !== "string" ||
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss cassandra fix with tests. The in-app matcher returned on the first operator, so a compound range like { gte, lte } dropped the second bound. Brought it in line with the databricks store's matcher, which already checks all operators. Adds the first cassandra test file. Quick look when you get a chance, thanks.

mem0-ts/src/oss/src/vector_stores/cassandra.ts
```diff
@@ -455,44 +455,64 @@ export class CassandraDB implements VectorStore {
...
+      sawOperator = true;
+      if (
+        typeof payloadValue !== "string" ||
+        !payloadValue.toLowerCase().includes(value.icontains.toLowerCase())
+      )
+        return false;
     }
 
-    return payloadValue === value;
+    return sawOperator ? true : payloadValue === value;
```
kartik-mem0, line 515: Only the `gte`+`lte` compound case has a dedicated test; the other 8 operators (`eq`/`ne`/`gt`/`lt`/`in`/`nin`/`contains`/`icontains`) changed identically here but aren't exercised in compound form.

Low risk since the edit to all ten is mechanical and identical, not a blocker. Worth a follow-up test if you're touching this file again.

kartik-mem0 (commented):

`matchFieldCondition` returned on the first operator key it saw, so `{ age: { gte: 10, lte: 20 } }` only applied `gte` and let out-of-range rows through; this makes every operator present in a compound filter AND together, mirroring `databricks.ts`'s matcher.

**Follow-up:** `mem0-ts/src/oss/src/vector_stores/memory.ts:92-152` (the default in-memory OSS store) has the identical early-return bug, unfixed; `mem0/vector_stores/cassandra.py` has no compound-operator support at all (plain equality only), a different and pre-existing gap. Neither blocks this PR, worth their own issues.
**Not verified:** no Cassandra sandbox lane exists in this repo, so this is proven against the repo's own jest suite in a worktree, not a live Cassandra cluster.

Risk: low, isolated to one provider's own matcher, no shared code touched · Your time: ~3 min

- **Problem is real:** reproduced on `main` independently (standalone script re-implementing `matchFieldCondition`, not the PR's code): `age=25` against `{gte:10, lte:20}` incorrectly matches (`true`, expected `false`) because the `gte` branch returns before `lte` is ever checked.
- **Tests:** `pnpm jest src/oss/tests/cassandra.unit.test.ts` -> `3 passed, 3 total` (with the fix).
- **Red-green:** implementation reverted to `main`, test kept from the PR -> `1 failed, 2 passed` (the compound-range test fails, the age-25 row leaks through exactly as reported); fix restored -> `3 passed, 3 total` again. Full round trip in the worktree.
- **CI:** all real jobs green (CI Gate, TS SDK build + integration on Node 20/22, changelog, CLA); only `Vercel` preview-deploy is red, advisory and ignored.
- **Security:** cleared, filter values only reach in-process `===`/`>`/`>=`/`<`/`<=`/`.includes` comparisons, no query string, shell, or file path construction.
- **Bloat (ponytail):** lean already, ship. The 10 duplicated `if` blocks mirror the already-shipped `databricks.ts` pattern byte-for-byte; collapsing them to a table-driven loop here would be an unrequested refactor riding along with a bug fix.
- **Linked issue:** #6510

</details>

mem0-ts/src/oss/src/vector_stores/cassandra.ts
abhay-codes07: ok

kartik-mem0 (approved):

Issue #6243, opened 2026-07-11
bug(memory): entity store TOCTOU race condition in _upsert_entity and _remove_memory_from_entity_store


bennyyuan1008:
## Bug Description

_upsert_entity and _remove_memory_from_entity_store (both sync Memory and async AsyncMemory variants) perform a classic read-modify-write cycle on entity records without any synchronization. Under concurrent operations targeting the same scope, two threads/coroutines can read the same snapshot, each append/remove a memory_id, and the second write silently overwrites the first, causing lost linked_memory_ids.

## Affected Code

- Memory._upsert_entity
- Memory._remove_memory_from_entity_store
- AsyncMemory._upsert_entity_async
- AsyncMemory._remove_memory_from_entity_store

## Prior Art

The async delete_all method already acknowledges this via _bulk_clear_entity_store docstring: "Used by delete_all to avoid the race condition that occurs when concurrent _delete_memory coroutines each try to read-modify-write the same entity rows."

## Proposed Fix

Add threading.Lock to Memory and asyncio.Lock to AsyncMemory, serializing entity store mutations within each instance.

kartik-mem0:
Thanks for tracing this down to the exact read-modify-write shape — and for pointing at the existing `_bulk_clear_entity_store` precedent, that's a nice piece of corroborating evidence.

## Triage Notes

**What we've established:**
- `_upsert_entity` (`mem0/memory/main.py:568-613`) and `_remove_memory_from_entity_store` (`:615-668`), plus their async counterparts `_upsert_entity_async` (`:2209-2257`) and the async `_remove_memory_from_entity_store` (`:2280-2323`), all read an entity's `linked_memory_ids` (via `.search()`/`.list()`), mutate the Python list locally, then write it back with `.update()` — with no lock around the sequence.
- Two concurrent calls into the same `user_id`/`agent_id`/`run_id` scope (two threads sharing a `Memory` instance, or two tasks sharing an `AsyncMemory` instance — the normal shape of a multi-request server) can both read the same snapshot; whichever writes second overwrites the first, and the first writer's `memory_id` silently drops out of `linked_memory_ids`.
[...]

PR #6516, opened 2026-07-22
fix(memory): close hash-dedup TOCTOU race in add() with per-instance lock

### LINKED ISSUE #6515: bug(memory): hash-dedup TOCTOU race in add() creates permanent duplicate memories under concurrency
### Component

Python SDK

### Summary

`Memory.add()` / `AsyncMemory.add()` can create a permanent duplicate memory when two concurrent calls extract the identical fact for the same scope (e.g. a retried request, a double-submitted webhook, or two parallel agent turns writing to the same `user_id`).

The V3 phased pipeline in `_add_to_vector_store` takes a snapshot of existing memories in Phase 1 (`existing_results = self.vector_store.search(...)`), then only checks the hash-dedup guard (`existing_hashes`) against that same stale snapshot in Phase 4/5 — after an LLM extraction round-trip and embedding calls have already happened. If two concurrent calls both take their snapshot before either has inserted, both pass dedup against the same (not-yet-updated) state and both insert, producing two rows with identical content and hash but different `memory_id`s.

This is a classic TOCTOU (time-of-check to time-of-use) race, distinct from:
- #6243 (entity-store `linked_memory_ids` read-modify-write race — different data, different code path)
- #4892 (concurrent writes corrupting the Qdrant HNSW index — a storage-engine-level corruption bug; this issue reproduces with a plain in-memory fake vector store, so it's a pipeline-level design gap independent of any specific backend's thread-safety)

### Steps to Reproduce

```python
import asyncio
from mem0 import AsyncMemory

m = AsyncMemory()  # any backend

async def add_same_fact():
    await m.add("I really like coffee", user_id="u1")

async def main():
    # Two concurrent calls extracting the same fact for the same user_id
    await asyncio.gather(add_same_fact(), add_same_fact())
    results = await m.get_all(user_id="u1")
    print(len(results["results"]))  # expected 1, actually 2

asyncio.run(main())
```

A minimal, backend-independent unit-test repro (using a fake in-memory vector store to isolate the pipeline logic from any specific backend) is included in the linked PR.

### Expected Behavior

Only one memory should be persisted for the same extracted fact under concurrent `add()` calls to the same scope — the hash-dedup guard should hold even under concurrency.

### Actual Behavior

Two (or more) permanent duplicate memory rows are created, each with the same `hash` in its payload but a distinct `memory_id`. There is no error, warning, or user-visible signal — the duplication is silent and accumulates over time in any production deployment with concurrent writers (parallel agent turns, retried webhooks, multi-worker API servers) hitting the same user/agent/run scope.

### Environment

- mem0 version: current `main` (`upstream/main` @ `ca2abca2`)
- Python/Node version: 3.10–3.12
- OS: any

### Proposed Fix

Serialize the dedup-recheck + insert step with a per-instance lock (`threading.Lock` for `Memory`, `asyncio.Lock` for `AsyncMemory` — same convention already accepted for the entity-store race in #6243), and re-fetch existing hashes fresh immediately before insert instead of reusing the Phase 1 snapshot. This closes the TOCTOU window regardless of which vector store backend is configured. A PR implementing this with regression tests (proving the race red on current `main` and green with the fix, for both sync and async) will follow.

OfficialAbhinavSingh:
## Linked Issue

Closes #6515

## Description

`_add_to_vector_store` (both `Memory` and `AsyncMemory`) takes a snapshot of existing memories in Phase 1, then checks the hash-dedup guard against that same snapshot in Phase 4/5 — after an LLM extraction round-trip and embedding calls have already happened. Two concurrent `add()` calls extracting the identical fact for the same scope (a retried request, a double-submitted webhook, two parallel agent turns) both pass dedup against the same stale snapshot and both insert, creating a permanent duplicate memory (same `hash`, different `memory_id`). No error or warning is raised — it's silent.

This PR serializes the dedup-recheck + insert step with a **per-session-scope lock** (`threading.Lock` for `Memory`, `asyncio.Lock` for `AsyncMemory` — same convention already accepted for the entity-store race in #6243) and re-fetches existing hashes fresh, immediately before insert, instead of reusing the Phase 1 snapshot. This closes the race at the application level, independent of any specific vector store backend's own thread-safety.

The lock is keyed by session scope (`user_id`/`agent_id`/`run_id`), created lazily, so `add()` calls for **different** scopes — which can never hash-collide — do not serialize against each other; only same-scope callers do.

### Scope / limitation

These locks live on the `Memory`/`AsyncMemory` instance, so they only serialize callers **sharing one instance in one process**. They do **not** coordinate across multiple workers or replicas — e.g. this repo's `server/` and `openmemory/` scale-out deployments hold the client as a per-process singleton, so each process gets its own lock and the race window stays open **across** processes. Closing that cross-process window would require a shared/distributed lock (Redis/DB advisory lock, etc.) and is intentionally out of scope for this PR. This is documented at both lock sites in the code so the guarantee isn't misread as cross-worker.

Distinct from:
- #6243 — entity-store `linked_memory_ids` read-modify-write race (different data, different code path)
- #4892 — concurrent writes corrupting the Qdrant HNSW index (a storage-engine corruption bug; this bug reproduces with a plain in-memory fake vector store, so it's a pipeline-level design gap, not backend-specific)

## Type of Change

## Breaking Changes

N/A — no public API changes. One extra `vector_store.search()` call happens per `add()` invocation that produces at least one record (needed to close the race); this is the same cost class as the existing Phase 1 search and only runs when there's something to persist.

## Test Coverage

Three regression tests in `tests/memory/test_main.py::TestConcurrentAddDedupRace`:
- `test_concurrent_add_same_fact_does_not_duplicate` — `Memory`, real `threading.Thread`s
- `test_async_concurrent_add_same_fact_does_not_duplicate` — `AsyncMemory`, `asyncio.gather`
- `test_different_scopes_use_separate_locks_and_both_persist` — asserts distinct scopes get distinct locks and both persist (no false cross-scope dedup, no needless cross-scope serialization)

All use a plain in-memory fake vector store (scope-aware, no backend-specific behavior); the race tests use an artificially slow mocked LLM call to reliably force the race window.

**Proof — red on `upstream/main`, green with the fix** (race tests run against a clean `upstream/main` checkout of `mem0/memory/main.py`, then against this branch):

```
# BEFORE (raw upstream/main — Phase 4/5 dedup checked only against the stale Phase 1 snapshot):
$ pytest tests/memory/test_main.py -k concurrent_add_same_fact
FAILED test_concurrent_add_same_fact_does_not_duplicate
  AssertionError: expected exactly 1 persisted memory after two concurrent add()
  calls for the same fact, got 2: [{'id': '...', 'hash': 'f6e56b29...'}, {'id': '...', 'hash': 'f6e56b29...'}]
FAILED test_async_concurrent_add_same_fact_does_not_duplicate
  AssertionError: expected exactly 1 persisted memory ... got 2: [...]
2 failed, 49 deselected

# AFTER (this branch — per-scope lock + fresh in-lock recheck immediately before insert):
$ pytest tests/memory/test_main.py -k ConcurrentAddDedupRace
3 passed, 48 deselected

# Full memory suite + lint, clean on this branch:
$ pytest tests/memory/ -q
355 passed

$ ruff check mem0/memory/main.py tests/memory/test_main.py
All checks passed!
```

## Checklist

- [x] I have updated documentation if needed (N/A — internal concurrency fix, no public API/doc change; scope limitation documented inline)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -509,11 +509,21 @@ class Memory(MemoryBase):
         # Entity store is initialized lazily on first use
         self._entity_store = None
 
-        # Serializes the dedup-recheck + insert step in _add_to_vector_store so
-        # concurrent add() calls extracting the same fact can't both pass a
-        # stale dedup snapshot and both insert (see issue tracking the
-        # hash-dedup TOCTOU race).
-        self._add_lock = threading.Lock()
+        # Per-scope locks for the dedup-recheck + insert step in
+        # _add_to_vector_store, so concurrent add() calls extracting the same
+        # fact can't both pass a stale dedup snapshot and both insert (the
+        # hash-dedup TOCTOU race). Keyed by session scope so add() calls for
+        # different user_id/agent_id/run_id — which can never hash-collide —
+        # don't serialize against each other; only same-scope callers do.
+        #
+        # Scope of protection: these locks live on the Memory instance, so they
+        # only serialize callers sharing one instance in one process. They do
+        # NOT coordinate across multiple workers/replicas (e.g. this repo's
+        # server/ and openmemory/ scale-out deployments, which hold the client
+        # as a per-process singleton) — closing that cross-process window would
+        # need a shared/distributed lock and is out of scope here.
+        self._add_locks = {}
+        self._add_locks_guard = threading.Lock()
 
         if MEM0_TELEMETRY:
             # Create telemetry config manually to avoid deepcopy issues with thread locks
@@ -1032,13 +1042,20 @@ class Memory(MemoryBase):
             self.db.save_messages(messages, session_scope)
             return []
 
-        # Phase 5: Hash dedup + Phase 6: Batch persist, serialized.
+        # Phase 5: Hash dedup + Phase 6: Batch persist, serialized per scope.
         # The Phase 1 snapshot (existing_results) can go stale by the time we get
         # here (an LLM round-trip + embedding calls have happened since). Holding
-        # self._add_lock and re-checking hashes fresh right before insert closes
-        # that TOCTOU window: two concurrent add() calls extracting the same fact
-        # can no longer both pass dedup and both insert a permanent duplicate.
-        with self._add_lock:
+        # the session-scoped lock and re-checking hashes fresh right before insert
+        # closes that TOCTOU window: two concurrent add() calls extracting the
+        # same fact can no longer both pass dedup and both insert a permanent
+        # duplicate. Only same-scope callers serialize here (see __init__).
+        with self._add_locks_guard:
+            add_lock = self._add_locks.get(session_scope)
+            if add_lock is None:
+                add_lock = threading.Lock()
+                self._add_locks[session_scope] = add_lock
+
+        with add_lock:
             fresh_existing = self.vector_store.search(
                 query=parsed_messages,
                 vectors=query_embedding,
@@ -2195,11 +2212,20 @@ class AsyncMemory(MemoryBase):
         self.custom_instructions = self.config.custom_instructions
         self._entity_store = None
 
-        # Serializes the dedup-recheck + insert step in _add_to_vector_store so
-        # concurrent add() calls extracting the same fact can't both pass a
-        # stale dedup snapshot and both insert (see issue tracking the
-        # hash-dedup TOCTOU race).
-        self._add_lock = asyncio.Lock()
+        # Per-scope locks for the dedup-recheck + insert step in
+        # _add_to_vector_store, so concurrent add() calls extracting the same
+        # fact can't both pass a stale dedup snapshot and both insert (the
+        # hash-dedup TOCTOU race). Keyed by session scope so add() calls for
+        # different user_id/agent_id/run_id — which can never hash-collide —
+        # don't serialize against each other; only same-scope callers do.
+        #
+        # Scope of protection: these locks live on the AsyncMemory instance, so
+        # they only serialize callers sharing one instance in one process/event
+        # loop. They do NOT coordinate across multiple workers/replicas (e.g.
+        # this repo's server/ and openmemory/ scale-out deployments, which hold
+        # the client as a per-process singleton) — closing that cross-process
+        # window would need a shared/distributed lock and is out of scope here.
+        self._add_locks = {}
 
         # Initialize reranker if configured
         self.reranker = None
@@ -2701,13 +2727,21 @@ class AsyncMemory(MemoryBase):
             await asyncio.to_thread(self.db.save_messages, messages, session_scope)
             return []
 
-        # Phase 5: Hash dedup + Phase 6: Batch persist, serialized.
+        # Phase 5: Hash dedup + Phase 6: Batch persist, serialized per scope.
         # The Phase 1 snapshot (existing_results) can go stale by the time we get
         # here (an LLM round-trip + embedding calls have happened since). Holding
-        # self._add_lock and re-checking hashes fresh right before insert closes
-        # that TOCTOU window: two concurrent add() calls extracting the same fact
-        # can no longer both pass dedup and both insert a permanent duplicate.
-        async with self._add_lock:
+        # the session-scoped lock and re-checking hashes fresh right before insert
+        # closes that TOCTOU window: two concurrent add() calls extracting the
+        # same fact can no longer both pass dedup and both insert a permanent
+        # duplicate. Only same-scope callers serialize here (see __init__).
+        # The get/set below has no await between the check and the assignment, so
+        # it is atomic within a single event loop — no extra guard lock needed.
+        add_lock = self._add_locks.get(session_scope)
+        if add_lock is None:
+            add_lock = asyncio.Lock()
+            self._add_locks[session_scope] = add_lock
+
+        async with add_lock:
             fresh_existing = await asyncio.to_thread(
                 self.vector_store.search,
                 query=parsed_messages,
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -5,6 +5,7 @@ import hashlib
 import json
 import logging
 import os
+import threading
 import time
 import uuid
 import warnings
@@ -508,6 +509,12 @@ class Memory(MemoryBase):
         # Entity store is initialized lazily on first use
         self._entity_store = None
 
+        # Serializes the dedup-recheck + insert step in _add_to_vector_store so
+        # concurrent add() calls extracting the same fact can't both pass a
+        # stale dedup snapshot and both insert (see issue tracking the
+        # hash-dedup TOCTOU race).
+        self._add_lock = threading.Lock()
+
         if MEM0_TELEMETRY:
             # Create telemetry config manually to avoid deepcopy issues with thread locks
             telemetry_config_dict = {}
@@ -992,14 +999,7 @@ class Memory(MemoryBase):
                 except Exception as e:
                     logger.warning(f"Failed to embed memory text: {e}")
 
-        # Phase 4: Per-memory CPU processing + Phase 5: Hash dedup
-        # Build set of existing hashes for dedup
-        existing_hashes = set()
-        for mem in existing_results:
-            h = mem.payload.get("hash") if hasattr(mem, "payload") and mem.payload else None
-            if h:
-                existing_hashes.add(h)
-
+        # Phase 4: Per-memory CPU processing
         records = []  # (memory_id, text, embedding, payload)
         seen_hashes = set()  # dedup within the current batch
         for mem in extracted_memories:
@@ -1008,7 +1008,7 @@ class Memory(MemoryBase):
                 continue
 
             mem_hash = hashlib.md5(text.encode()).hexdigest()
-            if mem_hash in existing_hashes or mem_hash in seen_hashes:
+            if mem_hash in seen_hashes:
                 logger.debug(f"Skipping duplicate memory (hash match): {text[:50]}")
                 continue
             seen_hashes.add(mem_hash)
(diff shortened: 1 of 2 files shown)


mem0/memory/main.py
```diff
@@ -488,6 +489,12 @@ def __init__(self, config: MemoryConfig = MemoryConfig()):
         # Entity store is initialized lazily on first use
         self._entity_store = None
 
+        # Serializes the dedup-recheck + insert step in _add_to_vector_store so
+        # concurrent add() calls extracting the same fact can't both pass a
+        # stale dedup snapshot and both insert (see issue tracking the
+        # hash-dedup TOCTOU race).
+        self._add_lock = threading.Lock()
```
kartik-mem0: This lock is created per `Memory` instance, so it only serializes callers that share one instance in one process. `openmemory/api/app/utils/memory.py` holds the client as a singleton per process, so multi-worker/multi-replica deployments (this repo's own `server/` and `openmemory/` are built to scale that way) still get one lock per process and the race stays open across them.

Add a comment here (and a line in the PR description) making that scope explicit, so ops does not read "closes the race" as closing it across workers too.

mem0/memory/main.py
```diff
@@ -2138,6 +2161,12 @@ def __init__(self, config: MemoryConfig = MemoryConfig()):
         self.custom_instructions = self.config.custom_instructions
         self._entity_store = None
 
+        # Serializes the dedup-recheck + insert step in _add_to_vector_store so
+        # concurrent add() calls extracting the same fact can't both pass a
+        # stale dedup snapshot and both insert (see issue tracking the
+        # hash-dedup TOCTOU race).
+        self._add_lock = asyncio.Lock()
```
kartik-mem0: Same instance-scoped limitation as the sync lock at line 496: this closes the race within one process, not across multiple workers or replicas. Please note the scope here too.

mem0/memory/main.py
```diff
@@ -1012,24 +1012,47 @@ def _add_to_vector_store(self, messages, metadata, filters, infer, prompt=None):
...
-                vectors=all_vectors,
-                ids=all_ids,
-                payloads=all_payloads,
+        # Phase 5: Hash dedup + Phase 6: Batch persist, serialized.
+        # The Phase 1 snapshot (existing_results) can go stale by the time we get
+        # here (an LLM round-trip + embedding calls have happened since). Holding
+        # self._add_lock and re-checking hashes fresh right before insert closes
+        # that TOCTOU window: two concurrent add() calls extracting the same fact
+        # can no longer both pass dedup and both insert a permanent duplicate.
+        with self._add_lock:
```
kartik-mem0: This lock is instance-wide, so two `add()` calls for completely unrelated users on a shared instance now fully serialize through this extra `search()` + `insert()`, even though different `user_id`/`agent_id`/`run_id` scopes can never hash-collide. Keying the lock by scope (a small dict of locks, created lazily) keeps the fix and removes the unrelated-caller serialization.

mem0/memory/main.py
```diff
@@ -2644,24 +2667,48 @@ async def _add_to_vector_store(
...
-                vectors=all_vectors,
-                ids=all_ids,
-                payloads=all_payloads,
+        # Phase 5: Hash dedup + Phase 6: Batch persist, serialized.
+        # The Phase 1 snapshot (existing_results) can go stale by the time we get
+        # here (an LLM round-trip + embedding calls have happened since). Holding
+        # self._add_lock and re-checking hashes fresh right before insert closes
+        # that TOCTOU window: two concurrent add() calls extracting the same fact
+        # can no longer both pass dedup and both insert a permanent duplicate.
+        async with self._add_lock:
```
kartik-mem0: Same instance-wide serialization cost as line 1021 on the sync side: unrelated users' `add()` calls block on each other here even though they cannot collide. Worth scoping the lock by `(user_id, agent_id, run_id)`.

kartik-mem0 (commented):
## Changes requested

This genuinely closes the hash-dedup TOCTOU race for callers sharing one `Memory`/`AsyncMemory` instance (reproduced on `main`, both sync and async, against real Qdrant; fixed on this branch, same runs). Blocking on scope-honesty and a cheap tightening, not a redesign.

**To unblock:**
1. `mem0/memory/main.py:496,2168` — document that `_add_lock` only serializes callers sharing one instance/process, not multi-worker or multi-replica deployments (detail in the inline comments)
2. `mem0/memory/main.py:1021,2676` — consider keying the lock by `(user_id, agent_id, run_id)` instead of one lock per instance, to stop unrelated users' `add()` calls from serializing through the same lock (detail in the inline comments)

- **Problem is real:** reproduced independently, both by code trace (`_add_to_vector_store` Phase 1 snapshot vs. stale Phase 4/5 dedup, `mem0/memory/main.py`) and live sandbox against real Qdrant: `main`@`ca2abca2` FAILs in both sync and async mode (2 duplicates persisted), PR@`3d191235` PASSes in both. Full verbatim runs with `repo=`/`mem0=` provenance in `.claude/babysit-evidence/pr-6516-2026-07-23.md`.
- **Tests / red-green:** `TestConcurrentAddDedupRace` (new class, absent from `main`) passes on the branch; full `tests/memory/test_main.py` (50 tests) passes; `ruff check` clean on both changed files.
- **CI:** no required check red.
- **Security:** cleared, no new external input handling; lock is released on every exit path via context manager, no reentrancy into `_add_lock` found in the locked block.
- **Bloat (ponytail):** lean, reuses the exact lock convention already accepted for the entity-store race (#6243), stdlib-only, no new abstraction. If anything under-scoped in one place (instance-wide lock vs. scope-keyed), see item 2.
- **Not verified:** multi-process reproduction itself (would need two separate processes racing the same backend); the gap is established by code reading (`_add_lock` created in `__init__`, so instance-scoped) plus this repo's own `openmemory/api/app/utils/memory.py` singleton-per-process pattern, not by a live multi-process repro.

</details>

AsyncMemory has the full parallel fix (verified, not just assumed) and the diff itself is clean and proportional. Push the two changes above and I will re-review.

OfficialAbhinavSingh:
@kartik-mem0 Thanks for the detailed review — both requested changes are addressed in `15d944bc`:

1. **Scope honesty** — documented at both lock sites (sync `main.py` `__init__` and async) that these instance-level locks only serialize callers sharing one process, and explicitly that they do **not** coordinate across multiple workers/replicas (`server/`, `openmemory/` per-process singletons), where a shared/distributed lock would be needed. Added a matching "Scope / limitation" section to the PR description.

2. **Scope-keyed locks** — replaced the single instance-wide lock with a lazily-populated map keyed by session scope (`user_id`/`agent_id`/`run_id`). Different scopes can never hash-collide, so they no longer serialize through one lock or pay each other's extra `search()` + `insert()`; only same-scope callers serialize. Added `test_different_scopes_use_separate_locks_and_both_persist` covering it (both sync and async).

Re-verified: race tests red on a clean `upstream/main` checkout, green on this branch (sync + async); full `tests/memory/` 355 passed; `ruff check` clean. Ready for re-review.

kartik-mem0 (commented):
## Re-review: LGTM

Re-reviewed at head `15d944bc` ("fix(memory): key add() dedup lock by session scope; document process scope"). Both asks from the prior review are addressed, and I verified each one against the diff rather than the PR description.

**What changed since the last review:**
- Both `Memory.__init__` and `AsyncMemory.__init__` now carry a comment documenting that these locks only serialize callers sharing one instance in one process, and don't coordinate across multiple workers/replicas. Present verbatim at both sites (`mem0/memory/main.py:492-505` and `:2181-2194`), matching what was asked.
- The single instance-wide `self._add_lock` is gone. It's replaced by `self._add_locks`, a dict keyed by session scope (`user_id`/`agent_id`/`run_id`, via the existing `_build_session_scope` helper), so unrelated users' `add()` calls no longer block on each other. Sync gets a small guard lock (`self._add_locks_guard`) around the get-or-create so two threads can't race into creating two different lock objects for the same new scope; async skips the guard lock deliberately, since there's no `await` between the check and the set, so it's atomic under asyncio's cooperative scheduling within one event loop.

Nice fix, thank you Abhinav for taking both review comments and turning them into a clean, scoped solution instead of just patching around the original report.

<details>

Checked out the head in a worktree and read the sync and async `_add_to_vector_store` paths in full, plus `_build_session_scope` and the filter construction at both call sites, to confirm the lock key and the dedup query scope are provably the same partition (both exclude `None`/empty values the same way, so there's no split-lock for equivalent scopes).

Ran the existing sandbox scenario (real Qdrant) that reproduces the original TOCTOU dedup race: still green on this head in both sync (3.7s) and async (3.3s) mode, still red on unchanged `main` (duplicate persisted).

Added a second sandbox scenario specifically for the scope-keyed claim: two concurrent `add()` calls for two different users, with the insert artificially delayed 3s. On the prior single-lock head this serializes (~6.6-6.7s, ~2x the delay, both modes) - on this head it doesn't (~3.7-3.8s, ~1x the delay, both modes). Clean red/green split on a real backend, not just the unit tests.

Also wrote a throwaway stress test (30 threads racing to create the lock for one brand-new scope, 15 trials) against this head to empirically back up the atomicity reasoning on the sync guard lock - every trial produced exactly one lock entry and one persisted memory.

`ruff check` on the changed files is clean, and the full `tests/memory/test_main.py` suite passes (51 passed), including the three new concurrency tests.

One small thing worth a follow-up, not a blocker: the PR reply mentions the new `test_different_scopes_use_separate_locks_and_both_persist` test covers both sync and async, but only a sync version exists in `tests/memory/test_main.py` right now. The async behavior for that specific scenario is still correct (verified above via the sandbox scenario and code reading), it's just not pinned down by a dedicated async unit test yet.

</details>

**Not verified:** cross-process/multi-replica coordination remains out of scope (as documented) and wasn't re-tested here since it's an explicit non-goal; and I didn't attempt to force a real OS-thread race on the async lock-creation across multiple event loops, since a single `AsyncMemory` driven from more than one loop is already outside what `asyncio.Lock` supports.

OfficialAbhinavSingh:
@kartik-mem0 Thanks for the thorough re-review and the sandbox verification. Good catch on the coverage claim — you're right that only the sync `test_different_scopes_use_separate_locks_and_both_persist` existed. Added the async twin (`test_async_different_scopes_use_separate_locks_and_both_persist`) in `36a386d9` so that scenario is pinned down by a dedicated `AsyncMemory` unit test, not just sandbox/code-reading. Full `tests/memory/` now 356 passed, `ruff` clean.

kartik-mem0 (commented):
Re-review at 36a386d9. LGTM stands.

Diff since the last LGTM (15d944bc) is exactly one test-only commit, +37/-0 in tests/memory/test_main.py: `test_async_different_scopes_use_separate_locks_and_both_persist`. This is the async twin of `test_different_scopes_use_separate_locks_and_both_persist` that my prior review's "Not verified" note flagged as missing (the earlier claim of an existing async version was overstated). `mem0/memory/main.py` is byte-identical to 15d944bc, re-diffed to confirm.

Verified the new test is not vacuous: it uses `asyncio.gather` for genuine concurrent coroutine execution (its sync sibling runs sequentially in a for-loop), backed by a thread-safe `_make_fake_store()`. Since `AsyncMemory`'s vector-store calls go through `asyncio.to_thread`, `gather` here exercises real interleaving through the executor thread pool.

Re-checked fresh this pass, not trusted from the prior evidence file:
- Lock does not span the LLM call in either path: sync LLM call at main.py:943, lock block at ~1021-1026; async LLM call at main.py:2618, lock block at ~2697. No hot-path regression.
- Process/instance scope limitation still documented verbatim at main.py:492-505 and :2181-2194.
- TS OSS `mem0-ts/src/oss/src/memory/index.ts` still has no lock on its add path (`class Memory` at line 180). This PR's scope is Python-only per the linked issue, so this is informational, not blocking.

Sandbox re-run against 36a386d9, both declared scenarios, both modes, all green: `pr6516_concurrent_add_dedup_race` (sync PASS 3.4s, async PASS 3.4s) and `pr6516_scope_keyed_lock` (sync PASS 3.7s, async PASS 3.7s). Confirmed the race still reproduces on unmodified main first (FAIL, 2 duplicates). Full targeted suite: 6 passed. Ruff clean.

No new findings. Full evidence: .claude/babysit-evidence/pr-6516-2026-07-23.md (section "Re-review — head 36a386d9").

OfficialAbhinavSingh:
@kartik-mem0 checking in on this one. It has been sitting since your LGTM on 23 Jul.

I rebased it onto main on 4 Aug (2.0.13 to 2.0.15) so `changelog_check` would not go stale, and CI Gate is green on the new head `f3afc592`. The diff itself is unchanged from what you reviewed and re-verified against Qdrant.

Anything you still need from me before this can go in? The TypeScript twin #6532 covers the same race in `mem0-ts` `add()` and has not had a first review yet, so they may be easier to take together.

kartik-mem0 (commented):
Re-reviewed at `f3afc592`.

This is a content-identical rebase of the previously reviewed `36a386d9`. Verified line-by-line (not assumed): `gh pr diff` stats unchanged (+132/-51 in `main.py`, +198 in tests), full diff read against prior evidence, no additional hunks. The apparent extra changes in blob-level comparison (`_strip_identity_keys`, `delete_all` pagination) are from unrelated commits on `main` pulled into the base, absent from the actual PR diff.

Fresh verification on the new head:
- `pytest tests/memory/test_main.py -k "concurrent or race or dedup or lock or scope"`: 7 passed
- Full `tests/memory/`: 365 passed
- Ruff: clean
- Sandbox (both mandatory lanes): oss-qdrant sync PASS, oss-qdrant async PASS, oss-pgvector sync PASS
- Control on current `main` (`a39a802b`): still reproduces the duplicate-add race ("got 2")

LGTM at the new SHA. Prior findings and caveats (per-process lock scope, no TS equivalent) unchanged.

PR #6517, opened 2026-07-22
feat(integrations): n8n community node for Mem0

Himanshu-Sangshetti:
## What

A native **n8n community node** for Mem0 — add, search, and manage long-term memories inside any n8n workflow, and use it as a tool for the n8n AI Agent. Wraps the hosted Mem0 **v3** REST API (`api.mem0.ai`).

Split out of #6057 into its own PR (per review — one integration per PR for easier evaluation). The Zapier app is a separate PR.

## Operations

Add · Search · Get · Get Many · Update · Delete (Memory resource). `usableAsTool: true`.

## Meets n8n verified-node requirements

- **Zero runtime dependencies** (calls REST directly — the SDK can't be a dep)
- TypeScript, ESLint (`n8n-nodes-base`) clean, MIT + LICENSE
- `n8n-nodes-` name + `n8n-community-node-package` keyword
- Publishes via GitHub Actions with provenance

## Notable

- **Extraction controls:** `custom_instructions`, `custom_categories`, `includes`, `excludes` (optional Additional Fields, forwarded to the add endpoint).
- **Client-side entity-id guard** on Add — clear node error instead of a raw 4xx.
- **Async add:** returns `event_id`, polls `GET /v1/event/{id}/`; `itemIndex` carried on poll errors; wait-path output unwrapped to match search / get-many.
- **Telemetry:** anonymous, fire-and-forget, opt out with `MEM0_TELEMETRY=false`.

## Testing

- `tsc` + ESLint clean.
- Live E2E against `api.mem0.ai`: entity-guard → add (with custom_instructions/includes/excludes) → search → get-many → get → delete → verify-empty.

## Monorepo wiring

ci-gate filter + job + `needs`, release router (`n8n-nodes-mem0-v*`), AGENTS.md CI/CD tables, docs page + `docs.json` + `llms.txt`.

## Follow-up (not in this PR)

- `Return All` pagination option for Get Many (currently returns one page).
- Publish: npm via GitHub Actions → submit at creators.n8n.io for verification.

PR #6518, opened 2026-07-22
feat(integrations): Zapier app for Mem0

Himanshu-Sangshetti:
## What

A native Zapier app for Mem0. It lets you add, search, list, and delete long-term memories from any Zap, with no code. It is built with the Zapier Platform CLI and calls the hosted Mem0 v3 REST API (`api.mem0.ai`).

This was split out of #6057 into its own PR so each integration can be reviewed on its own. The n8n node is a separate PR.

## Actions

| Type | Action | Endpoint |
| --- | --- | --- |
| Create | Add Memory | `POST /v3/memories/add/` |
| Create | Delete Memory | `DELETE /v1/memories/{id}/` |
| Search | Search Memories | `POST /v3/memories/search/` |
| Search | Get Memories | `POST /v3/memories/` |

Authentication is a custom API key, sent as `Authorization: Token <key>`.

## Written in TypeScript

The source lives in `src/` and compiles to `dist/`, which is what Zapier runs (`main` points to `dist/index.js`). CI builds before it validates or pushes. This keeps the app consistent with the rest of the repo, which favours a single language.

## Notable

- Extraction controls. Add exposes optional `custom_instructions`, `custom_categories`, `includes`, and `excludes` fields, so a user can shape how memories are extracted.
- Timeout-safe async. Wait for Completion defaults to off, and the poll budget stays under Zapier's 30 second step limit. If a wait does time out, the app reports that the add was accepted and is likely still completing, rather than calling it a failure.
- Error handling. The `handleBadResponses` middleware turns any non-2xx response into a real error. Zapier's `z.request` does not throw on non-2xx by default, so without this a 4xx could flow downstream as a fake empty result.
- Telemetry. Anonymous and fire-and-forget, and it can be turned off with `MEM0_TELEMETRY=false`. It sends a hashed key as the identifier, never the raw key.

## Testing

`zapier validate` is clean, the offline unit tests (which mock `z.request`) run in CI, and a live end-to-end test runs the full lifecycle against `api.mem0.ai`: add, then search until the memory is indexed, then get all, then delete, then confirm the list is empty.

## Monorepo wiring

CI gate filter, job, and needs entry; an AGENTS.md CI table row plus a note that the Zapier app deploys to Zapier's own platform rather than the npm release router; and a docs page registered in `docs.json` and `llms.txt`.


kartik-mem0 (commented):
## Changes requested

Zapier Platform CLI app wrapping the hosted Mem0 v3 API (add/search/get/delete), CI green including a `zapier validate` pass, real unit tests (13/13, verified this run) plus a live E2E suite. Two Important fixes needed before merge.

**To unblock:**
1. `src/authentication.ts:14` — use `type: 'password'` for the API key field (detail in the inline comment)
2. `src/index.ts:9-10` — drop the `require()` calls, use ES `import` (detail in the inline comment)

Optional: AGENTS.md's "Key Directories" table is missing a row for `integrations/zapier-mem0/` (the CI/CD table got one, that table didn't); a few cosmetic `zapier validate` warnings (D003/D004/D028) in the CI log are worth a look but don't block.

- **CI:** all real checks pass, including the repo's own `Zapier App / validate`. Only `Vercel` shows non-pass, and that's the advisory preview-deploy check, ignored per policy.
- **Tests:** `pnpm install --frozen-lockfile && pnpm build && pnpm test:unit` run fresh in a worktree at this head → 13/13 passed. Read `test/mem0.test.ts` (live E2E, gated on `MEM0_API_KEY`) in full; it matches the PR description's lifecycle claim.
- **Security:** API key never logged/echoed (middleware error paths use response-body text only) and only MD5-hashed before telemetry; the one gap is the UI-masking issue called out above, not a transport/logging leak.
- **Bloat (ponytail):** `net: 0 lines possible`, lean already.
- **Not verified:** live E2E suite not executed (needs a real `MEM0_API_KEY` against `api.mem0.ai`); actual Zapier deploy/publish flow not exercised (needs `ZAPIER_DEPLOY_KEY`).

</details>

Solid first pass at this integration, the async/sync add handling and defensive response parsing are genuinely careful. Push when ready and I will re-review.

No linked issue on this one, for tracking on our side, mind opening a quick issue and linking it here? Makes it easier to prioritize and reference.

integrations/zapier-mem0/src/authentication.ts
```diff
@@ -0,0 +1,30 @@
...
+const test = (z: ZObject, _bundle: Bundle) => z.request({ url: '/v1/ping/', method: 'GET' });
+
+export default {
+	type: 'custom',
+	test,
+	fields: [
+		{
+			key: 'apiKey',
+			label: 'Mem0 API Key',
+			type: 'string',
```
kartik-mem0: This should be `type: 'password'`, not `'string'`. Zapier masks `password`-type fields in the connection UI; as written, a user's Mem0 API key is shown in cleartext when they set up the connection. Zapier's own schema docs treat an API key field as sensitive (`isNoSecret: false`-class) for exactly this reason.

integrations/zapier-mem0/src/index.ts
```diff
@@ -0,0 +1,35 @@
+import authentication from './authentication';
+import { includeApiKey, handleBadResponses } from './middleware';
+import addMemory from './creates/add_memory';
+import deleteMemory from './creates/delete_memory';
+import searchMemories from './searches/search_memories';
+import getMemories from './searches/get_memories';
+
+// Read version at runtime from the package manifest (one directory up from dist/).
+const { version } = require('../package.json');
+const platformVersion = require('zapier-platform-core').version;
```
kartik-mem0: Both `require()` calls here are avoidable: `zapier-platform-core` exports a named `version` (`import { version as platformVersion } from 'zapier-platform-core'`), and `tsconfig.json` already has `resolveJsonModule` + `esModuleInterop` on, so `import pkg from '../package.json'` works too. Under `module: commonjs` both compile down to the same `require()` at emit time, this just gets the source back on ES `import` syntax, which is the repo's explicit rule for TypeScript.

integrations/zapier-mem0/src/authentication.ts
Himanshu-Sangshetti: done

integrations/zapier-mem0/src/index.ts
Himanshu-Sangshetti: done

Himanshu-Sangshetti:
hi @kartik-mem0, few notes here- 
1. in .ts now 
2. added custom instructions 
3. added telemetry 
4. for auth- not the pattern like .zshrc approach since Zapier runs in its own cloud with no local env or files. it stores the key in its own credential store

integrations/zapier-mem0/src/authentication.ts
kartik-mem0: Verified against de39f294: fixed. `authentication.ts:14` now sets `type: 'password'`, confirmed by reading the file, plus a clean `pnpm build` and `zapier validate` this run.

integrations/zapier-mem0/src/index.ts
kartik-mem0: Verified against de39f294: fixed. `index.ts` now uses `import pkg from '../package.json'` and `import { version as platformVersion } from 'zapier-platform-core'`, no more require(). pnpm build is clean and 13/13 unit tests pass this run; compiled dist/index.js still resolves both via the same CommonJS interop, no runtime change.

kartik-mem0 (commented):
## Needs a human call

Native Zapier app (add/search/get/delete memories via the hosted v3 REST API). Re-reviewed after both prior findings were fixed.

**Decide:** do we want to adopt and maintain a Zapier app as a shipped integration? That means owning a `workflow_dispatch`-only deploy key, a third ecosystem-specific REST client to keep in sync with the v3 API alongside the Python/TS SDKs and the n8n node (#6517), and CI coverage that is partial by design (the live E2E lifecycle test only runs locally against a real key).
**Why this needs you:** new integration surface with its own auth file and `.github/workflows/` changes, so the adoption call is yours even with clean code.
**Look at:** `integrations/zapier-mem0/src/authentication.ts:14` (fixed, confirmed this run), `integrations/zapier-mem0/src/index.ts` (fixed, confirmed this run)
**Your time:** ~10 min

- **Delta since last review** (`1bd9910` -> `de39f294`): `authentication.ts`, `index.ts`, `AGENTS.md`, `package.json`, new `test/setup.ts` (5 files, +9/-6)
- **Both prior Important findings confirmed fixed, this run:** `authentication.ts:14` now `type: 'password'` (masked in Zapier's connection UI, was cleartext `'string'`). `index.ts` now uses `import pkg from '../package.json'` and `import { version as platformVersion } from 'zapier-platform-core'`, no `require()` left.
- **Build:** `pnpm build` (tsc) -> clean, exit 0. Compiled `dist/index.js` still resolves both via the same CommonJS interop, no runtime change.
- **Tests:** `pnpm test:unit` -> 13/13 passed, 0.235s. New `test/setup.ts` disables telemetry during tests via jest `setupFiles`, removing a lingering open handle.
- **Validate:** `npx zapier-platform-cli validate` -> 0 errors, 0 publishing-blocking tasks, same 13 cosmetic Warnings as before (D004 dynamic-dropdown, D028 cleanInputData, etc.), unchanged and non-blocking.
- **CI:** every required check green on `de39f294` (CI Gate, Zapier App/validate, Node/Python CLI, Python/TS SDK matrices, llms.txt). Vercel shows "Canceled by Ignored Build Step", advisory, ignored.
- **Minor (non-blocking):** AGENTS.md Key Directories row was added in this commit too.
- **Not verified:** live E2E suite (`test/mem0.test.ts`) requires a real `MEM0_API_KEY` against `api.mem0.ai`; correctly `describe.skip`-gated when absent, not run this session. Actual Zapier deploy/publish flow not exercised (needs `ZAPIER_DEPLOY_KEY`).

</details>

Replied in-thread on both prior findings confirming the fixes; nothing else outstanding on the code itself.

KartikLabhshetwar (approved):

kartik-mem0 (approved):

PR #6477, opened 2026-07-21
fix(opensearch): re-raise search errors instead of returning []

yashwanth123:
## Linked Issue

Closes #6476

## Description

`OpenSearchDB.search()` and `keyword_search()` previously caught all OpenSearch client exceptions, logged them, and returned `[]`. That made transport failures, auth errors, and misconfigured indices indistinguishable from a successful query with no matches.

This change logs with `exc_info=True` and re-raises, matching the behavior of `insert()` and `update()` in the same module.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests

Updated `test_search_error_logs_with_exc_info` to assert re-raise instead of `[]`. Added `test_keyword_search_error_logs_with_exc_info`. All 1685 Python tests pass locally.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally

PR #6519, opened 2026-07-22
salvage: fix(opensearch) re-raise search errors (credit @yashwanth123 #6477)

Bartok9:
## Why

OpenSearchDB.search() (and keyword_search) treated client exceptions as empty results. That hid auth/transport/index misconfig behind "no hits."

## What

After logging with exc_info=True, **re-raise** on both paths so callers can distinguish real failures — aligned with insert/update in the same module.

- Upgrade vs #6477: keyword_search also **raises** (not return None) so error semantics match search.
- Tests: assert re-raise + exc_info on both helpers.

## Credit

Salvage of **#6477** by **@yashwanth123** — original diagnosis and search-path direction. Co-authored-by on the commit.

## Linked

Related: #6476 / #6477

## Notes

AI-assisted; human-reviewed. Local tests/vector_stores/test_opensearch.py — 44 passed.

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -285,8 +285,12 @@ class OpenSearchDB(VectorStoreBase):
             ]
             return results
         except Exception as e:
+            # Do NOT re-raise here: keyword_search() is a best-effort helper that
+            # search() may call to augment semantic results. Raising would crash
+            # the whole search() call on a keyword-only failure (regression per
+            # maintainer review on #6519). Log with exc_info and degrade to None.
             logger.error(f"Error during keyword search: {e}", exc_info=True)
-            raise
+            return None
 
     def delete(self, vector_id: str) -> None:
         """Delete a vector by custom ID."""
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -240,7 +240,7 @@ class OpenSearchDB(VectorStoreBase):
             return results
         except Exception as e:
             logger.error(f"Error during search: {e}", exc_info=True)
-            return []
+            raise
 
     def keyword_search(self, query, top_k=5, filters=None):
         """Search for memories using BM25 keyword matching.
@@ -285,8 +285,8 @@ class OpenSearchDB(VectorStoreBase):
             ]
             return results
         except Exception as e:
-            logger.error(f"Error during keyword search: {e}")
-            return []
+            logger.error(f"Error during keyword search: {e}", exc_info=True)
+            raise
 
     def delete(self, vector_id: str) -> None:
         """Delete a vector by custom ID."""
--- a/tests/vector_stores/test_opensearch.py
+++ b/tests/vector_stores/test_opensearch.py
@@ -424,10 +424,20 @@ class TestOpenSearchDB(unittest.TestCase):
 
     @patch("mem0.vector_stores.opensearch.logger")
     def test_search_error_logs_with_exc_info(self, mock_logger):
-        """Search error logging should include exc_info for full stack trace."""
+        """Search errors should log with exc_info and re-raise (not swallow as [])."""
         self.client_mock.search.side_effect = Exception("Search failed")
-        results = self.os_db.search(query="", vectors=[[0.1] * 1536], top_k=5)
-        self.assertEqual(results, [])
+        with self.assertRaises(Exception):
+            self.os_db.search(query="", vectors=[[0.1] * 1536], top_k=5)
+        mock_logger.error.assert_called_once()
+        call_kwargs = mock_logger.error.call_args
+        self.assertTrue(call_kwargs[1].get("exc_info"), "logger.error must be called with exc_info=True")
+
+    @patch("mem0.vector_stores.opensearch.logger")
+    def test_keyword_search_error_logs_and_raises(self, mock_logger):
+        """Keyword search errors should log with exc_info and re-raise."""
+        self.client_mock.search.side_effect = Exception("Keyword search failed")
+        with self.assertRaises(Exception):
+            self.os_db.keyword_search(query="test", top_k=5)
         mock_logger.error.assert_called_once()
         call_kwargs = mock_logger.error.call_args
         self.assertTrue(call_kwargs[1].get("exc_info"), "logger.error must be called with exc_info=True")
--- a/tests/vector_stores/test_opensearch.py
+++ b/tests/vector_stores/test_opensearch.py
@@ -433,11 +433,16 @@ class TestOpenSearchDB(unittest.TestCase):
         self.assertTrue(call_kwargs[1].get("exc_info"), "logger.error must be called with exc_info=True")
 
     @patch("mem0.vector_stores.opensearch.logger")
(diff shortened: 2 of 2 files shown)


mem0/vector_stores/opensearch.py
```diff
@@ -285,8 +285,8 @@ def keyword_search(self, query, top_k=5, filters=None):
             ]
             return results
         except Exception as e:
-            logger.error(f"Error during keyword search: {e}")
-            return []
+            logger.error(f"Error during keyword search: {e}", exc_info=True)
+            raise
```
kartik-mem0: This makes an isolated keyword-search failure crash the whole search() call. `mem0/memory/main.py`'s `_search_vector_store` (sync and async) calls `keyword_search()` with no try/except and checks `if keyword_results is not None:` right after, so it expects `None` on failure, not a raise.

Return `None` here instead, matching `VectorStoreBase`'s own default and #6477's current head, which already fixes this the same way.

kartik-mem0 (commented):
## Changes requested

The search() fix is correct and root-cause; the keyword_search() -> raise change reopens a regression a caller can't handle.

**To unblock:**
1. `mem0/vector_stores/opensearch.py:289`: return None instead of raise (detail in inline comment)

- **Problem is real:** reproduces on `main` at `opensearch.py:231,283`, live OpenSearch container, both search() and keyword_search() swallow an injected client error and return [] (sandbox scenario `issue6476_opensearch_search_swallows_errors`, FAIL both modes on main, PASS both modes on this PR's head)
- **Regression (live sandbox):** scenario `pr6477_keyword_search_no_fallback` isolates a keyword-search-only failure. PASS on `main` (search() degrades to semantic-only results), FAIL on this PR's head, both modes: `search()` now raises and crashes the whole call
- **Tests / red-green:** `pytest tests/vector_stores/test_opensearch.py` -> 44 passed at PR head; reverted opensearch.py to main with tests kept at PR head -> the 2 new/changed tests fail as expected (`AssertionError: Exception not raised`); restored -> 44 passed again
- **CI:** no required check red observed
- **Security:** cleared, error path only logs (now with exc_info=True) and re-raises/returns the same exception object already available to the caller
- **Bloat (ponytail):** lean already, two one-line changes plus exc_info and two tests
- **Not verified:** live network partition / auth-specific OpenSearch failure modes beyond the injected RuntimeError

</details>

Good catch and real fix on search(), @Bartok9, and full credit to @yashwanth123 for the original diagnosis in #6477 this salvages. One thing worth knowing before you push again: #6477 already has a follow-up commit that makes keyword_search() degrade to None instead of raising, which is exactly the change this PR needs, so it might be worth syncing with @yashwanth123 so we land one clean fix instead of two competing ones. Push when ready and I will re-review.

Bartok9:
Thanks @kartik-mem0 — that's a sharp catch and your live-sandbox verification is exactly right. `keyword_search()` is best-effort augmentation for `search()`, so raising there reopens the regression where a keyword-only failure crashes the whole call.

Fixed in `2898f94b`:
- `search()` — **re-raise kept** (root-cause fix, unchanged)
- `keyword_search()` — now **logs with exc_info=True and returns None** so `search()` degrades to semantic-only instead of crashing
- Test renamed `test_keyword_search_error_logs_and_degrades`: asserts `None` + exc_info; full suite `tests/vector_stores/test_opensearch.py` → **44 passed**

This now matches the follow-up direction on #6477, so we land one clean fix. Full credit to @yashwanth123 for the original diagnosis (Co-authored-by retained). Ready for re-review 🙏

mem0/vector_stores/opensearch.py
kartik-mem0: Verified against `2898f94b` (new head): fixed. `keyword_search()` now logs with `exc_info=True` and returns `None` on error, matching `VectorStoreBase.keyword_search()`'s own default and #6477's current head. `search()` is unchanged, still raises. Confirmed with the `pr6477_keyword_search_no_fallback` sandbox scenario against a live OpenSearch container: PASS on this head, both sync and async, where it failed on the prior head. Fast turnaround, thanks.

kartik-mem0 (commented):

Verified against `2898f94b` (new head): `keyword_search()` now logs with `exc_info=True` and degrades to `None` on error instead of raising, closing the gap our prior review flagged. `search()` is unchanged, still raises.

**Before you merge:** #6477 independently reaches the identical fix around the same time; both are correct now, so which one merges is a human call, not a code gap.
**Follow-up:** none, this closes the only open finding.
**Not verified:** live network partition or auth-specific OpenSearch failure modes beyond the injected error we simulate.

Risk: low, error-handling contract change only, matches `VectorStoreBase`'s own default · Your time: ~2 min

- **Problem is real:** reproduced on `main` at `opensearch.py:231,283` in the prior review cycle; premise unchanged, not re-run this cycle.
- **Fix scope:** delta `0e4d9c7f..2898f94b` touches only the `keyword_search()` except block (`raise` -> `return None`) plus its test; `search()` untouched.
- **Regression (live sandbox):** scenario `pr6477_keyword_search_no_fallback`, real OpenSearch container. Baseline on `main`: PASS both modes (sync, async). PR worktree at `2898f94b`: PASS both modes, where the prior head `0e4d9c7f` FAILed both.
- **Caller contract:** `mem0/memory/main.py:_search_vector_store` sync (1617-1625) and async (3255-3263) both call `keyword_search()` unguarded and branch on `if keyword_results is not None:`. `VectorStoreBase.keyword_search()`'s own docstring: "Returns None if not supported." This PR's return now matches both, and since `AsyncMemory` calls the same `OpenSearchDB.keyword_search()` via `asyncio.to_thread`, sync/async parity holds by construction.
- **Tests:** `pytest tests/vector_stores/test_opensearch.py -v` in the PR worktree -> 44 passed, including the renamed `test_keyword_search_error_logs_and_degrades`.
- **Lint:** `ruff check mem0/vector_stores/opensearch.py tests/vector_stores/test_opensearch.py` -> clean.
- **CI:** all required checks green (Vercel preview-deploy fails, not merge-blocking).
- **Security:** cleared, no change from the prior review, error path only logs and returns the same exception info already available to the caller.
- **Bloat (ponytail):** lean already, one line flipped plus a 4-line comment explaining why, no new abstraction.
- **Duplicate cluster:** #6477 independently landed the same fix (`647874cfab`: `search()` raises, `keyword_search()` returns `None`, matching tests) before this PR's fix commit. Neither PR now covers anything the other lacks, it is a genuine tie, not a subsumption. Recommend picking one to merge and closing the other with credit to both authors.

</details>

Good catch and a fast turnaround here, Bartok9, thanks. Same credit stands to yashwanth123, whose original diagnosis and direction both PRs now share.

Bartok9:
Thank you @kartik-mem0 — really appreciate the thorough live-sandbox re-verification on `2898f94b`. Glad the `keyword_search()` degrade-to-`None` path now matches `VectorStoreBase`'s contract and the `main.py` caller branch.

On the tie with #6477: happy to defer entirely to your call on which one lands — both are correct now and I have no preference beyond landing one clean fix. Full credit either way to @yashwanth123 for the original diagnosis (Co-authored-by retained on our commit). If closing this in favor of #6477 is cleaner for you, no objection at all 🙏

kartik-mem0 (approved):

PR #6521, opened 2026-07-23
fix(ts-oss/chroma): stop dropping filter conditions in where-clause translation

### LINKED ISSUE #6513: bug(ts-oss/chroma): filter translation carries the same three where-clause bugs fixed in Python by #6452
### Component

TypeScript OSS SDK

### Description

Tracking issue for the TS side of #6451, as flagged by @kartik-mem0 in the review of #6452:

> Follow-up: `mem0-ts/src/oss/src/vector_stores/chroma.ts:273-401` carries the identical three bugs (same-field overwrite, flat `$or` merge, unmapped `$not` operators dropped), worth its own tracking issue.

`ChromaDB.generateWhereClause` / `convertCondition` in the TS SDK is a line-for-line structural sibling of the Python `_generate_where_clause` fixed in #6452, and violates ChromaDB's one-operator-per-level where grammar in the same three ways:

1. **Same-field operator dicts overwrite each other** — `convertCondition` (chroma.ts:273) assigns `condition[key] = { [op]: val }` fresh on every loop iteration, so a merged range like `{ age: { gte: 18, lte: 65 } }` degrades to `{ age: { $lte: 65 } }` and returns rows the caller excluded. (Even without the overwrite, a multi-operator field expression is rejected by ChromaDB — `Expected operator expression to have exactly one operator` — so each operator needs its own clause combined with `$and`.)
2. **`$not` conditions with operators missing from the negation map are silently dropped** (`contains`/`icontains` and unknown ops), which can erase the entire where clause and run the query unfiltered.
3. **Multi-field conditions inside `$or` branches are merged into flat objects**, which ChromaDB rejects at query time (`Expected where to have exactly one operator`).

### Expected Behavior

Same as the Python fix in #6452: one single-operator clause per operator, combined with `$and`/`$or`; `$not` negates `contains`/`icontains`/unknown operators to `$ne` (mirroring the positive path's `$eq` fallback) instead of dropping them.

### Environment

- mem0-ts: main branch (TypeScript 3.1.1)

I'll port the #6452 fix over in a follow-up PR, with regression tests mirroring the four Python ones.

microbluey:
Fixes #6513

Port of the Python fix from #6452 (merged) to the structurally identical TS implementation, as flagged in the #6452 review (`chroma.ts:273-401` carried the same three bugs):

1. **Same-field ranges lost a bound** — `convertCondition` reassigned the field's clause per operator, so `{ age: { gte: 18, lte: 65 } }` degraded to `{ age: { $lte: 65 } }`. Each operator now becomes its own single-operator clause combined with `$and` (ChromaDB rejects multi-operator field expressions outright).
2. **`$not` + `contains`/`icontains`/unknown operators erased the clause** — operators missing from the negation map were silently skipped; they now negate to `$ne`, mirroring the positive path's `$eq` fallback.
3. **Multi-field conditions inside `$or` were flat-merged** with `Object.assign` into objects ChromaDB rejects (`Expected where to have exactly one operator`); they are now wrapped in `$and`.

Implementation mirrors the merged Python version: `convertCondition` returns a clause list, and a shared `combineClauses` helper (0 → drop, 1 → unwrap, n → wrap) replaces the duplicated length-check blocks. The `OR`/`NOT` aliases and the `Array.isArray → $in` shortcut specific to the TS implementation are preserved.

## Verification

- 4 new regression tests mirroring the Python ones from #6452 (same-field range, range inside `$or`, multi-field condition inside `$or`, `$not` contains/icontains).
- All 34 existing tests in `src/oss/src/tests/chroma.test.ts` pass unchanged (the pre-existing `generateWhereClause` assertions confirm output-shape parity for already-correct inputs): `Tests: 38 passed, 38 total`.
- `prettier --check` clean; `tsc` reports no errors in the touched files (the `src/client/tests` jest-types errors are pre-existing and unrelated).

### DIFF
--- a/mem0-ts/src/oss/src/tests/chroma.test.ts
+++ b/mem0-ts/src/oss/src/tests/chroma.test.ts
@@ -440,4 +440,52 @@ describe("generateWhereClause", () => {
       ChromaDB.generateWhereClause({ $not: [{ age: { gt: 18 } }] }),
     ).toEqual({ age: { $lte: 18 } });
   });
+
+  // Regression tests for the three where-clause translation bugs fixed in
+  // Python by #6452 and tracked for the TS SDK in #6513. ChromaDB allows
+  // exactly one operator or field per dict level.
+  it("keeps both bounds of a same-field range as $and-combined clauses", () => {
+    expect(ChromaDB.generateWhereClause({ age: { gte: 18, lte: 65 } })).toEqual(
+      { $and: [{ age: { $gte: 18 } }, { age: { $lte: 65 } }] },
+    );
+  });
+
+  it("keeps same-field ranges inside $or branches", () => {
+    expect(
+      ChromaDB.generateWhereClause({
+        $or: [{ age: { gte: 18, lte: 65 } }, { vip: true }],
+      }),
+    ).toEqual({
+      $or: [
+        { $and: [{ age: { $gte: 18 } }, { age: { $lte: 65 } }] },
+        { vip: { $eq: true } },
+      ],
+    });
+  });
+
+  it("wraps multi-field conditions inside $or in $and", () => {
+    expect(
+      ChromaDB.generateWhereClause({
+        $or: [{ age: { gte: 18 }, vip: true }, { city: "sh" }],
+      }),
+    ).toEqual({
+      $or: [
+        { $and: [{ age: { $gte: 18 } }, { vip: { $eq: true } }] },
+        { city: { $eq: "sh" } },
+      ],
+    });
+  });
+
+  it("negates $not contains/icontains instead of dropping the clause", () => {
+    expect(
+      ChromaDB.generateWhereClause({
+        $not: [{ title: { contains: "draft" } }],
+      }),
+    ).toEqual({ title: { $ne: "draft" } });
+    expect(
+      ChromaDB.generateWhereClause({
+        $not: [{ title: { icontains: "draft" } }],
+      }),
+    ).toEqual({ title: { $ne: "draft" } });
+  });
 });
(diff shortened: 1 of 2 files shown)


kartik-mem0 (commented):

Ports the merged Python where-clause fix (#6452) to Chroma's TS twin: same-field ranges, multi-field `$or` branches, and unmapped `$not` operators no longer silently drop filter conditions.

**Follow-up:** none, this closes the Python/TS parity gap flagged during #6452's review
**Not verified:** live ChromaDB API, verified via jest assertions on `generateWhereClause`'s output shape (no chroma sandbox lane exists yet)

Risk: low, private-method change behind two call sites (`search`/`list`), no public API touched · Your time: ~3 min

- **Problem is real:** reproduced on `main` by reverting the implementation file (see red-green below); same three bugs the merged Python fix (`a9cb4bb6`) addressed in `mem0/vector_stores/chroma.py`
- **Tests:** `npx jest src/oss/src/tests/chroma.test.ts` -> 38 passed, 38 total (34 pre-existing + 4 new regression tests)
- **Red-green:** reverted `chroma.ts` to pre-fix `main`, kept the PR's tests -> 4 failed (exactly the 4 new tests), 34 passed; restored the fix -> 38 passed again
- **CI:** TypeScript SDK build + integration green on Node 20/22, CI Gate green; Vercel preview-deploy red, advisory and ignored per policy
- **Security:** cleared, structural filter-to-query translation only, no injection/SSRF/deserialization surface, tenancy fields pass through the same unchanged path
- **Bloat (ponytail):** lean already, `combineClauses` has 3 real call sites in this diff and replaces 3 duplicated branch blocks, net reduction
- **Linked issue:** #6513

</details>

kartik-mem0 (approved):

PR #6522, opened 2026-07-23
fix(vector_stores/opensearch): translate the '*' wildcard to an exists query for every key

microbluey:
Follow-up to #6454, as discussed in [this review thread](https://github.com/mem0ai/mem0/pull/6454#issuecomment-5049882252): identity keys were exempt from the wildcard check, so `filters={"agent_id": "*"}` built a literal (near-always-empty) term match on the string `"*"` instead of the documented "any value" Platform semantics — and custom keys skipped the wildcard entirely.

## Change

`"*"` now builds `{"exists": {"field": "payload.<key>"}}` for **every** key — identity and custom alike — matching how `opensearch.ts` already handles it (`chroma`-side wildcards remain "no constraint", which is that backend's documented behavior). Non-scalar (operator-shaped) values for custom keys are still skipped exactly as before.

## Verification

- 3 new tests: identity-key wildcard → exists query, custom-key wildcard → exists query, wildcard combined with term filters.
- The prior test codifying the interim skip behavior (`test_search_ignores_wildcard_filter_value`) is updated to the exists semantics.
- Full suite: `pytest tests/vector_stores/test_opensearch.py` → 51 passed.
- `ruff check` clean; touched lines match `ruff format`.

### DIFF
--- a/mem0/vector_stores/opensearch.py
+++ b/mem0/vector_stores/opensearch.py
@@ -35,8 +35,16 @@ def _build_filter_clauses(filters):
     for key, value in (filters or {}).items():
         if value is None:
             continue
-        if key not in _IDENTITY_FILTER_KEYS and (not isinstance(value, (str, int, float, bool)) or value == "*"):
-            logger.debug(f"Ignoring non-scalar or wildcard filter value for key {key!r}")
+        if value == "*":
+            # "Any value" wildcard (a documented Platform pattern): match
+            # documents where the field exists — as opensearch.ts already
+            # does for every key — instead of a literal, near-always-empty
+            # term match on the string "*".
+            _validate_filter(key, value)
+            filter_clauses.append({"exists": {"field": f"payload.{key}"}})
+            continue
+        if key not in _IDENTITY_FILTER_KEYS and not isinstance(value, (str, int, float, bool)):
+            logger.debug(f"Ignoring non-scalar filter value for key {key!r}")
             continue
         _validate_filter(key, value)
         field = f"payload.{key}.keyword" if isinstance(value, str) else f"payload.{key}"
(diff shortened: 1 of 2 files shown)


kartik-mem0 (commented):

Identity-key wildcards (`filters={"agent_id": "*"}`) built a literal term match on the string `"*"` that matches nothing. This makes `_build_filter_clauses` translate `"*"` into an `exists` clause for every key, identity and custom alike, matching `opensearch.ts` and the documented "any value" wildcard contract.

**Follow-up:** none for this provider. chroma, qdrant, milvus, pinecone, cassandra and databricks still treat a wildcard as "skip the filter entirely" instead of "field must exist", which is a real cross-provider inconsistency but out of scope here.
**Not verified:** a managed OpenSearch Service cluster, only the sandbox's dockerized node.

Risk: low, single-function change with a mechanical red-green proof · Your time: ~5 min

- **Problem is real:** `mem0/vector_stores/opensearch.py:38` on `main`, the wildcard-skip check exempts identity keys, so `filters={"agent_id": "*"}` falls through to `{"term": {"payload.agent_id.keyword": "*"}}`, matching zero real documents. Reproduced on `main` with `sandbox/scenarios/pr6522_opensearch_identity_wildcard.py`, lane `oss-opensearch`, both `sync` and `async`, then confirmed the fix clears it on this branch, same scenario, both modes.
- **Tests:** `pytest tests/vector_stores/test_opensearch.py -v` -> 51 passed at `aa79a216`.
- **Red-green:** reverted `opensearch.py` to pre-PR (`a58e0586`) with the PR's tests kept, 4 wildcard tests failed with the exact bug signature (`{'term': {..., '*'}}` instead of `{'exists': ...}`). Restored, all 9 wildcard tests passed again.
- **CI:** Python SDK build (3.10/3.11/3.12) and CI Gate pass. Only `Vercel` is red, fork-permission and non-blocking.
- **Security:** cleared. The filter key is validated against `_SAFE_FILTER_KEY` before it is interpolated into the OpenSearch field path, and the value is type-checked by `_validate_filter` before use.
- **Bloat (ponytail):** Lean already. Ship.
- **Linked issue:** none, PR body says "Follow-up to #6454" only.

</details>

kartik-mem0 (approved):

microbluey:
Opened #6539 to track the cross-provider wildcard inconsistency flagged in this review, and #6540 with the fix for the remaining providers.

PR #6525, opened 2026-07-23
feat(memory): add disable_history config (TS disableHistory parity)

### LINKED ISSUE #6512: Add `disable_history` config to the open-source Python SDK/server (parity with JS `disableHistory`)
### Component

Python SDK

### Use Case

You're self-hosting mem0 and you don't want to leave PII in the history table after a GDPR request. The Python SDK appears not to offer a way to purge history by user. `delete_all(user_id)` appends a `DELETE` row rather than erasing history, so extracted facts which may contain PII can linger in `history.db` until the whole table is reset or the host recycles. For teams that never read history back (it's audit-only) and treat the vector store as the system of record, the cheap answer is to not write history at all. The Node SDK already supports this via `disableHistory` (#3291) but the Python SDK and server have no equivalent. 

### Proposed Solution

Copy what you already do for the Node SDK.

  A `disable_history` flag on `MemoryConfig`, defaulting to `False`:

  ```python
  config = MemoryConfig(disable_history=True)
  ```

  When true:

  - `add_history` and `batch_add_history` become no-ops (or the SQLite manager isn't instantiated).
  - `add` / `search` / `get` are unchanged — inference reasons over the vector store, not history, so extraction behaves exactly as today.
  - `get_history` returns an empty list.

Default `False` keeps existing behavior. History is an audit log, not a dependency of the memory pipeline, so suppressing the writes only removes the audit trail — which is what a caller setting the flag is asking for.

### Alternatives Considered

- **Patch the OSS image to no-op the history writes.** - what we did

Bartok9:
## Summary
Adds `MemoryConfig.disable_history` (default `False`) so OSS Python SDK/server can skip SQLite history writes, matching TS `disableHistory`.

Closes #6512

## Motivation
Self-hosters with GDPR purge needs cannot currently turn off history. `delete_all` appends DELETE rows rather than erasing history, and teams that treat the vector store as system-of-record want no local audit DB. Node already has `disableHistory`.

## Changes
- `MemoryConfig.disable_history`
- `DummyHistoryManager` no-op backend
- `Memory` / `AsyncMemory` (including `reset`) select backend via `_history_manager_from_config`
- REST server: `DISABLE_HISTORY` env → `disable_history` on MEMORY config
- Tests: `tests/memory/test_disable_history.py` (6 cases)

## Verification
```bash
pytest tests/memory/test_disable_history.py -q
# 6 passed
```

## Notes
Default remains `False` for full backward compatibility. Message retention on the same SQLite path is also no-op'd under the dummy so disabling history does not leave a local transcript store open.

### DIFF
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -43,6 +43,13 @@ class MemoryConfig(BaseModel):
         description="Path to the history database",
         default=os.path.join(mem0_dir, "history.db"),
     )
+    disable_history: bool = Field(
+        description=(
+            "When True, do not open or write the history SQLite DB "
+            "(parity with TS disableHistory). get_history returns []."
+        ),
+        default=False,
+    )
     reranker: Optional[RerankerConfig] = Field(
         description="Configuration for the reranker",
         default=None,
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -49,7 +49,7 @@ from mem0.memory.notices import (
     get_temporal_feature_error_message_async,
 )
 from mem0.memory.setup import mem0_dir, setup_config
-from mem0.memory.storage import SQLiteManager
+from mem0.memory.storage import DummyHistoryManager, SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
 from mem0.memory.utils import (
     extract_json,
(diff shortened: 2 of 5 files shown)


Bartok9:
Rebased onto current `main` after CONFLICTING (tip `b4dbcef9`).

- Product commits preserved: `MemoryConfig.disable_history` + `DummyHistoryManager`, Memory/AsyncMemory wiring, server `DISABLE_HISTORY` env.
- Resolved `server/main.py` drift: kept main's `gpt-5-mini` default model while retaining the `DISABLE_HISTORY` wiring and `disable_history` config key.
- Single clean commit on main; ready for CI + review.

PR #6532, opened 2026-07-23
fix(ts-oss): close hash-dedup TOCTOU race in add() with per-scope lock

### LINKED ISSUE #6531: bug(ts-oss): hash-dedup TOCTOU race in add() creates duplicate memories under concurrency
### Component

TypeScript SDK

### Summary

The TS OSS `Memory.add()` (`mem0-ts/src/oss/src/memory/index.ts`) has the same hash-dedup TOCTOU race that #6516 fixes on the Python side. `addToVectorStore` takes a snapshot of existing memories in Phase 1 (`existingResults = await this.vectorStore.search(...)`), then checks the hash-dedup guard against that same snapshot in Phase 4/5 — after an LLM extraction round-trip and embedding calls. Two `add()` calls extracting the identical fact for the same scope, awaited concurrently (e.g. `Promise.all([...])`, a retried request, two parallel agent turns), both take their snapshot before either inserts, both pass dedup, and both insert — producing a permanent duplicate memory (same `hash`, different `memory_id`). No error is raised; the duplication is silent.

Although Node runs on a single event loop, the interleaving `await`s between the snapshot and the insert are exactly the race window — a purely-async TOCTOU, not a thread race.

### Steps to Reproduce

```ts

// Same fact, same scope, awaited concurrently
await Promise.all([
  memory.add("I really like coffee", { userId: "u1" }),
  memory.add("I really like coffee", { userId: "u1" }),
]);

console.log(all.results.length); // expected 1, actually 2
```

### Expected Behavior

Only one memory persists for the same extracted fact under concurrent `add()` calls to the same scope.

### Actual Behavior

Two (or more) permanent duplicate memory rows are created — same `hash`, distinct `id` — with no error or warning.

### Environment

- mem0ai (npm) — TS OSS `Memory`
- Node 20/22
- OS: any

### Proposed Fix

Mirror the Python fix in #6516: serialize the dedup-recheck + insert step with a per-session-scope async mutex (a small lazily-populated promise-chain keyed by `user_id`/`agent_id`/`run_id`, so different scopes — which can never hash-collide — don't serialize against each other), and re-fetch existing hashes fresh immediately before insert instead of reusing the Phase 1 snapshot. Same instance/process-scoped limitation applies (does not coordinate across multiple workers/processes). PR to follow.

OfficialAbhinavSingh:
## Linked Issue

Closes #6531

## Description

TypeScript twin of #6516 (Python). `addToVectorStore` (`mem0-ts/src/oss/src/memory/index.ts`) takes a snapshot of existing memories in Phase 1 (`existingResults = await this.vectorStore.search(...)`), then checks the hash-dedup guard against that same snapshot in Phase 4/5 — after an LLM extraction round-trip and embedding calls. Two `add()` calls extracting the identical fact for the same scope, awaited concurrently (`Promise.all`, a retried request, two parallel agent turns), both take their snapshot before either inserts, both pass dedup, and both insert — creating a permanent duplicate memory (same `hash`, different `id`), silently.

Node's single event loop does **not** prevent this: the interleaving `await`s between the snapshot and the insert are exactly the race window — a purely-async TOCTOU, not a thread race.

This PR serializes the dedup-recheck + insert step with a **per-session-scope async mutex** — a lazily-populated promise-chain keyed by session scope (`user_id`/`agent_id`/`run_id`), so `add()` calls for **different** scopes (which can never hash-collide) don't serialize against each other; only same-scope callers do. Inside the lock it re-fetches existing hashes fresh and drops any record that now collides, then inserts. The map entry is cleaned up once its chain drains, so it doesn't grow unbounded across distinct scopes.

### Scope / limitation

The lock map lives on the `Memory` instance, so it only serializes callers sharing one instance in one process. It does **not** coordinate across multiple processes/workers — that would need a shared/distributed lock and is out of scope here. Documented inline. Matches the Python fix's stated guarantee in #6516.

## Type of Change

## Breaking Changes

N/A — no public API changes. One extra `vectorStore.search()` runs per `add()` that produces at least one record (same cost class as the existing Phase 1 search), only when there's something to persist.

## Test Coverage

New `mem0-ts/src/oss/tests/memory.dedup-race.test.ts`:
- `two concurrent add() calls for the same fact persist only one memory` — `Promise.all` of two same-scope adds, deterministic delayed LLM mock to widen the race window, real in-memory vector store; asserts `getAll` returns 1.
- `concurrent add() for different scopes each persist (no false cross-scope dedup)` — two different-user adds concurrently; asserts each scope keeps its own memory.

**Proof — red on `upstream/main`, green with the fix** (`index.ts` reverted to `upstream/main`, then restored):

```
# BEFORE (raw upstream/main index.ts — dedup only vs the stale Phase 1 snapshot):
$ npx jest --config jest.config.js src/oss/tests/memory.dedup-race.test.ts
  ✕ two concurrent add() calls for the same fact persist only one memory
    expect(received).toBe(expected)  Expected: 1  Received: 2
  ✓ concurrent add() for different scopes each persist (no false cross-scope dedup)
  Tests: 1 failed, 1 passed

# AFTER (this branch — per-scope async mutex + fresh in-lock recheck):
$ npx jest --config jest.config.js src/oss/tests/memory.dedup-race.test.ts
  ✓ two concurrent add() calls for the same fact persist only one memory
  ✓ concurrent add() for different scopes each persist (no false cross-scope dedup)
  Tests: 2 passed

# Related suites + build, clean on this branch:
$ npx jest --config jest.config.js memory.add memory.crud memory.dedup-race
  Tests: 65 passed
$ npx prettier --check <changed files>   # All matched files use Prettier code style!
$ npx tsc --noEmit                       # 26 errors, all pre-existing on main, none in changed files
$ pnpm run build                         # Build success
```

## Checklist

- [x] I have updated documentation if needed (N/A — internal concurrency fix, no public API/doc change; scope limitation documented inline)

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -205,6 +205,20 @@ export class Memory {
   private _initError?: Error;
   private _entityStore?: VectorStore;
 
+  // Per-session-scope mutex for the dedup-recheck + insert step in
+  // addToVectorStore, so concurrent add() calls extracting the same fact can't
+  // both pass a stale dedup snapshot and both insert (the hash-dedup TOCTOU
+  // race). Keyed by session scope (user_id/agent_id/run_id) so add() calls for
+  // different scopes — which can never hash-collide — don't serialize against
+  // each other; only same-scope callers do. Each map value is the tail of a
+  // promise chain; a caller awaits the current tail, then installs its own.
+  //
+  // Scope of protection: this map lives on the Memory instance, so it only
+  // serializes callers sharing one instance in one process. It does NOT
+  // coordinate across multiple processes/workers — that would need a
+  // shared/distributed lock and is out of scope here.
+  private addLocks: Map<string, Promise<void>> = new Map();
+
   constructor(config: Partial<MemoryConfig> = {}) {
     // Merge and validate config
     this.config = ConfigManager.mergeConfig(config);
@@ -563,6 +577,38 @@ export class Memory {
       if (val) parts.push(`${key}=${val}`);
     }
     return parts.join("&");
+  }
+
+  /**
+   * Serialize `fn` against all other callers sharing the same session scope.
+   * Runs `fn` only after the previous same-scope holder has released, so the
+   * dedup-recheck + insert step is atomic per scope. Different scopes run
+   * concurrently. The map entry is cleaned up once the chain drains, so it
+   * doesn't grow unbounded across distinct scopes.
+   */
+  private async withAddLock<T>(
+    scope: string,
+    fn: () => Promise<T>,
+  ): Promise<T> {
+    const previous = this.addLocks.get(scope) ?? Promise.resolve();
+    let release!: () => void;
+    const gate = new Promise<void>((resolve) => {
+      release = resolve;
+    });
+    const tail = previous.then(() => gate);
+    this.addLocks.set(scope, tail);
+
+    await previous;
+    try {
+      return await fn();
+    } finally {
+      release();
+      // If no later caller chained onto our tail, drop the entry so the map
+      // doesn't retain a lock object per distinct scope forever.
+      if (this.addLocks.get(scope) === tail) {
+        this.addLocks.delete(scope);
+      }
+    }
   }
 
   private async _initializeTelemetry() {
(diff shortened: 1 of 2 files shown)


OfficialAbhinavSingh:
@kartik-mem0 TS twin of #6516 (same TOCTOU fix, per-scope lock instead of per-instance since ts-oss is multi-tenant). #6516 got your LGTM 3x on 23 Jul - this one's had no look yet. all green, ready when you are.

PR #6538, opened 2026-07-23
fix(llms/lmstudio): honor `LMSTUDIO_BASE_URL` env var

### LINKED ISSUE #6526: bug(llms/lmstudio): LMSTUDIO_BASE_URL env var is ignored (default baked into config too early)

sahilkanger:
# fix(llms/lmstudio): honor LMSTUDIO_BASE_URL env var

## Linked Issue

Closes #6526

## Description

`LMStudioConfig` baked the default base URL in the config layer
(`self.lmstudio_base_url = lmstudio_base_url or "http://localhost:1234/v1"`), so by the time
`LMStudioLLM` built the OpenAI client the value was never `None` and the `LMSTUDIO_BASE_URL`
environment variable was never consulted — users who set it (or expected the same
`config > env > default` precedence the other OpenAI-compatible providers follow) always hit
`localhost:1234`.

This moves the default out of the config and resolves precedence in `LMStudioLLM.__init__`,
mirroring the existing **DeepSeek** provider (`mem0/llms/deepseek.py`):

- `mem0/configs/llms/lmstudio.py`: `self.lmstudio_base_url = lmstudio_base_url` (no default; the
  docstring already documented `defaults to None`).
- `mem0/llms/lmstudio.py`: `base_url = self.config.lmstudio_base_url or os.getenv("LMSTUDIO_BASE_URL") or "http://localhost:1234/v1"`.

Precedence is now: explicit `lmstudio_base_url` config **>** `LMSTUDIO_BASE_URL` env **>**
`http://localhost:1234/v1`.

Note: I scoped this to the LLM provider (the subject of the issue). The embeddings LM Studio
config (`mem0/configs/embeddings/base.py` + `mem0/embeddings/lmstudio.py`) has a similar
default-baked-in pattern; happy to file/fix that separately if you'd like, but kept it out to
keep this change focused.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — behavior is unchanged when neither env nor config is set (still `localhost:1234`) and when
config is set (still honored). The only new behavior is the intended env fallback.

## Test Coverage

- [ ] No tests needed

Added `tests/llms/test_lmstudio.py::test_lmstudio_llm_base_url` (mirrors
`tests/llms/test_deepseek.py`) covering all three cases — default, `LMSTUDIO_BASE_URL` env, and
explicit config (wins over env). It fails on `main` (env case returns `localhost`) and passes with
this change. Run locally:

```
$ python -m pytest tests/llms/test_lmstudio.py -q
1 passed
```
`ruff format --check` and `isort` are clean on the changed files.

## Checklist

- [x] I have updated documentation if needed (no doc claimed the env var worked; none needed)

### DIFF
--- a/mem0/configs/llms/lmstudio.py
+++ b/mem0/configs/llms/lmstudio.py
@@ -55,5 +55,7 @@ class LMStudioConfig(BaseLlmConfig):
         )
 
         # LM Studio-specific parameters
-        self.lmstudio_base_url = lmstudio_base_url or "http://localhost:1234/v1"
+        # Do NOT default here: keep `None` so LMStudioLLM can resolve
+        # config > LMSTUDIO_BASE_URL env > default (see #6526).
+        self.lmstudio_base_url = lmstudio_base_url
         self.lmstudio_response_format = lmstudio_response_format
--- a/mem0/llms/lmstudio.py
+++ b/mem0/llms/lmstudio.py
@@ -1,4 +1,5 @@
 import json
+import os
 from typing import Dict, List, Optional, Union
 
 from openai import OpenAI
@@ -38,7 +39,9 @@ class LMStudioLLM(LLMBase):
         )
         self.config.api_key = self.config.api_key or "lm-studio"
 
-        self.client = OpenAI(base_url=self.config.lmstudio_base_url, api_key=self.config.api_key)
+        # Precedence: explicit config > LMSTUDIO_BASE_URL env > default (#6526).
+        base_url = self.config.lmstudio_base_url or os.getenv("LMSTUDIO_BASE_URL") or "http://localhost:1234/v1"
+        self.client = OpenAI(base_url=base_url, api_key=self.config.api_key)
 
     def _parse_response(self, response, tools):
         """
--- a/tests/llms/test_lmstudio.py
+++ b/tests/llms/test_lmstudio.py
@@ -1,0 +1,20 @@
+from mem0.configs.llms.lmstudio import LMStudioConfig
+from mem0.llms.lmstudio import LMStudioLLM
+
+
+def test_lmstudio_llm_base_url(monkeypatch):
+    # case1: default is used when neither config nor env sets a base URL
+    monkeypatch.delenv("LMSTUDIO_BASE_URL", raising=False)
+    llm = LMStudioLLM(LMStudioConfig(model="test-model"))
+    assert str(llm.client.base_url) == "http://localhost:1234/v1/"
+
+    # case2: LMSTUDIO_BASE_URL env is honored when config leaves it unset (#6526)
+    env_base_url = "http://192.168.1.5:1234/v1/"
+    monkeypatch.setenv("LMSTUDIO_BASE_URL", env_base_url)
+    llm = LMStudioLLM(LMStudioConfig(model="test-model"))
+    assert str(llm.client.base_url) == env_base_url
+
+    # case3: explicit config base URL wins over the env var
+    config_base_url = "http://config-host:9999/v1/"
+    llm = LMStudioLLM(LMStudioConfig(model="test-model", lmstudio_base_url=config_base_url))
+    assert str(llm.client.base_url) == config_base_url


kartik-mem0 (commented):
## Duplicate, recommend consolidating into #6527

Both PRs fix the same root cause the same way (`config > env > default` in `mem0/llms/lmstudio.py` + `mem0/configs/llms/lmstudio.py`), but #6527 also covers three identical sibling occurrences of the bug that this PR leaves open.

**Recommend:** squash-merge #6527, close this one.
**Why #6527:** #6527 also fixes `mem0/embeddings/lmstudio.py` (Python embeddings) and both `mem0-ts/src/oss/src/llms/lmstudio.ts` and `mem0-ts/src/oss/src/embeddings/lmstudio.ts` (TS LLM + embeddings). This PR only fixes the Python LLM path, so merging it leaves `LMSTUDIO_BASE_URL` silently ignored for Python embeddings and the entire TS SDK, three sibling drift points #6527 removes.
**Your time:** ~3 min

<details><summary>How I verified both</summary>

- **This PR:** CI is gated (`action_required`, not yet run for a first-time contributor); real jobs haven't executed. My own targeted test `tests/llms/test_lmstudio.py` -> 1 passed. Red-green: reverted the impl files to `main` -> FAILED on the exact env-var case (`assert 'http://localhost:1234/v1/' == 'http://192.168.1.5:1234/v1/'`), restored -> PASSED. `ruff check` clean. Security: no new attack surface, `LMSTUDIO_BASE_URL` is deployer-controlled, same trust tier as the existing config param.
- **#6527:** head `b90c1626`, MERGEABLE. Its diff touches and fixes all four sibling files with the same precedence chain. CI green: Python 3.10/3.11/3.12 build+test, TS build+test on Node 20/22, TS integration tests. Its own prior review (2026-07-23) blocked it on these exact two gaps; its current head closes both.
- **Not verified:** #6527's own test rigor beyond CI green (red-green not re-run by me this cycle, deferred to its own review).

</details>

Real work here, @sahilkanger: your precedence chain (`mem0/llms/lmstudio.py:43-44`) is a cleaner one-liner than #6527's version, matches `deepseek.py` exactly with no unnecessary config write-back. Please reopen if #6527 falls through.

kartik-mem0:
Thanks for this one, and apologies it took a cycle to come back to you with a straight answer.

I'm consolidating the LM Studio base URL fix into #6527, so I'm closing this. That is not a knock on the code. Your diagnosis is exactly right, and the precedence chain you wrote is the cleaner of the two: #6527 resolves `config > env > default` and then writes the result back onto `self.config.lmstudio_base_url`, which nothing reads anymore, where yours just resolves it and moves on. I've flagged that dead write-back on #6527 as a cleanup before it lands, so the shape you wrote is close to what that file should end up as.

The reason #6527 wins is scope, not quality. The same baked-default bug exists in four places, and this PR fixes one:

- `mem0/llms/lmstudio.py` (this PR)
- `mem0/embeddings/lmstudio.py`
- `mem0-ts/src/oss/src/llms/lmstudio.ts`
- `mem0-ts/src/oss/src/embeddings/lmstudio.ts`

Landing a Python-LLM-only fix would leave the embedder and both TS providers still ignoring `LMSTUDIO_BASE_URL`, so anyone pointing at a remote LM Studio host would hit the identical wall one layer down and file the same issue again.

Worth saying plainly: the test you added is the part I liked most. Explicitly covering the no-config-no-env default case alongside the env case is the right way to pin a precedence chain, and it is what made this reviewable in a couple of minutes.

Please do send more. If you want something adjacent, the config layer generally is where this class of bug lives in this repo, providers that bake a default into the config class instead of leaving it `None` all have the same latent problem.

PR #6540, opened 2026-07-24
fix(vector_stores): enforce the '*' wildcard as field-exists across providers

### LINKED ISSUE #6539: Cross-provider inconsistency: '*' wildcard filter is skipped or literal-matched instead of enforcing field-exists
## Description

The documented contract for the `*` filter value is "require that the field exists, regardless of value" ([metadata filtering docs](https://docs.mem0.ai/open-source/features/metadata-filtering)), but after #6522 only OpenSearch enforces it. The other providers fall into two failure modes:

**Silently skip the condition** (wildcard matches records *without* the field):

- `qdrant.py` — `_build_field_condition` returns `None` for `"*"`
- `milvus.py` — `_create_filter` skips the key
- `chroma.py` — `convert_condition` returns `[]` (also chroma.ts)

**Literal match on the string `"*"`** (matches nothing, the same bug #6522 fixed for OpenSearch):

- `pinecone.py` — builds `{"$eq": "*"}`
- `cassandra.py` — client-side `payload.get(k) == "*"`
- `databricks.py` — passes `{"key": "*"}` through `filters_json`

So the same `filters={"agent_id": "*"}` call returns "everything" on Qdrant, nothing on Pinecone, and only field-present records on OpenSearch.

## Feasibility per provider

Verified capabilities:

- **Qdrant**: `must_not(IsEmptyCondition)` expresses exists (excludes missing/null/empty-array) — verified against qdrant-client local mode
- **Milvus**: `exists metadata["key"]` — verified against Milvus standalone v2.5.12 (missing key and explicit `null` both excluded)
- **Pinecone**: `{"$exists": true}` — documented metadata filter operator
- **Cassandra**: filtering is client-side, trivially expressible
- **Chroma**: no exists operator; `$ne` with a sentinel can't emulate it because `$ne` matches documents that lack the key entirely (verified on chromadb 1.5.9)
- **Databricks Vector Search**: no exists / `IS NOT NULL` operator in the filter grammar for either endpoint type

Proposal: enforce field-exists where the store can express it (Qdrant, Milvus, Pinecone, Cassandra), and for Chroma/Databricks drop the condition (match-all) as the best available approximation — but stop literal-matching `"*"` — and document the per-provider behavior.

Tracking issue for the cross-provider follow-up requested in the #6522 review.

microbluey:
## Linked Issue

Closes #6539

## Description

Follow-up to #6522. The `*` filter value is documented as "require that the field exists, regardless of value", but only OpenSearch enforced it. Qdrant, Milvus, and Chroma silently skipped the condition (so the wildcard also matched records *without* the field), while Pinecone, Cassandra, and Databricks built a literal match on the string `"*"` that matches nothing — the same bug #6522 fixed for OpenSearch.

Per-provider changes:

| Provider | Before | After |
|---|---|---|
| Qdrant | condition skipped | `must_not(IsEmptyCondition)` (excludes missing / null / empty-array) |
| Milvus | condition skipped | `exists metadata["key"]` |
| Pinecone | literal `{"$eq": "*"}` | `{"$exists": true}` |
| Cassandra | literal `payload.get(k) == "*"` | client-side key-present-and-not-null |
| Databricks | literal `"*"` in `filters_json` | wildcard entries dropped (grammar has no exists operator), debug-logged |
| Chroma | condition skipped | unchanged, with the why documented |

Chroma and Databricks cannot express field-exists: Databricks Vector Search's filter grammar has no exists / `IS NOT NULL` operator for either endpoint type, and Chroma has no `$exists` while `$ne` matches documents that lack the key entirely (verified empirically on chromadb 1.5.9), so a sentinel `$ne` can't emulate it. For those two the wildcard now consistently means "don't constrain this field" instead of a match-nothing literal, and the docs note the difference.

The docs page (`metadata-filtering.mdx`) gains a note describing which stores enforce field-exists.

## Type of Change

## Breaking Changes

N/A — behavior only changes for filters containing the `*` wildcard, moving them onto the documented contract.

## Test Coverage

Manual verification:

- **Qdrant**: end-to-end through `Qdrant.search()` against qdrant-client local mode — `{"agent_id": "*"}` returns only the point whose `agent_id` is present and non-null (missing key and explicit `null` both excluded)
- **Milvus**: end-to-end through `MilvusDB.search()` against a dockerized Milvus standalone v2.5.12 — same result; also confirmed `exists metadata["key"]` composes with `and`
- **Chroma**: confirmed on chromadb 1.5.9 that `$ne` matches documents lacking the key, ruling out a sentinel-based exists
- `pytest` across the seven touched provider suites: 318 passed

## Checklist

### DIFF
--- a/docs/open-source/features/metadata-filtering.mdx
+++ b/docs/open-source/features/metadata-filtering.mdx
@@ -129,7 +129,7 @@ results = m.search(
 ```
 
 <Warning>
-Wildcard semantics differ by vector store. pgvector requires the key to exist in the payload (a real "field exists" check). Qdrant and Chroma have no native "field exists" filter, so `*` is a no-op there: the field condition is dropped and every record passes, including ones where the field is missing entirely. Do not rely on `*` to exclude records with a missing field unless you have confirmed your store's behavior in `mem0/vector_stores/<provider>.py`.
+Wildcard semantics differ by vector store. pgvector, OpenSearch, Qdrant, Milvus, Pinecone, and Cassandra enforce `*` as a real "field exists" check, so records missing the field are excluded. Chroma and Databricks Vector Search have no field-exists operator, so `*` is a no-op there: the field condition is dropped and every record passes, including ones where the field is missing entirely. Do not rely on `*` to exclude records with a missing field on those two stores.
 </Warning>
 
 ### Logical combinations
--- a/docs/open-source/features/metadata-filtering.mdx
+++ b/docs/open-source/features/metadata-filtering.mdx
@@ -124,6 +124,13 @@ results = m.search(
 )
 ```
 
+<Note>
+  OpenSearch, Qdrant, Milvus, Pinecone, and Cassandra enforce the wildcard as a
+  field-exists condition. Chroma and Databricks Vector Search have no
+  field-exists operator, so on those stores the wildcard leaves the field
+  unconstrained and matches records with or without it.
+</Note>
+
 ### Logical combinations
 
 Combine filters with `AND`, `OR`, and `NOT` to express complex decision trees. Nest logical operators to encode multi-branch workflows.
--- a/mem0/vector_stores/cassandra.py
+++ b/mem0/vector_stores/cassandra.py
@@ -32,6 +32,21 @@ def _validate_identifier(name: str, label: str = "identifier") -> str:
     return name
 
 
+def _payload_matches(payload: Dict, filters: Dict) -> bool:
+    """Check a payload against filters, honoring the '*' any-value wildcard.
+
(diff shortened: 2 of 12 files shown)


microbluey:
Note on the red CI: the `Run Linting` step failed because `ci.yml` installs unpinned `ruff` and picked up ruff 0.16.0 (released today), whose expanded default rules flag pre-existing code repo-wide — the violations in the log are all in `examples/` and untouched SDK lines, none in this diff. The last `main` push and yesterday's PRs pass because they ran on the previous ruff release. This branch is lint-clean under the ruff series CI was green with. Happy to rebase once the ruff version is pinned.

PR #6541, opened 2026-07-24
Use UTF-8 for CLI and state file reads

Ghraven:
## Summary
- Read CLI JSON input/import files with explicit UTF-8 encoding.
- Use explicit UTF-8 for Codex hook installer template/config files.
- Read docs coverage files and server telemetry state with explicit UTF-8, and persist telemetry state with the same encoding.

## Problem
Several JSON/text files are read or written with `Path.read_text()` / `write_text()` without an explicit encoding. On Windows or other non-UTF-8 locales, user-provided CLI files, generated hook config, docs coverage checks, or telemetry state can become locale-dependent.

## Before / After
Before: these CLI/state/helper files used the platform default text encoding.

After: the same file paths consistently use UTF-8.

## Verification
- `python -m py_compile cli/python/src/mem0_cli/commands/memory.py cli/python/src/mem0_cli/commands/utils.py integrations/mem0-plugin/scripts/install_codex_hooks.py scripts/check-llms-txt-coverage.py server/telemetry.py`
- `python -m pytest tests/test_telemetry_aliasing.py -q` (`28 passed`)

### DIFF
--- a/cli/python/src/mem0_cli/commands/memory.py
+++ b/cli/python/src/mem0_cli/commands/memory.py
@@ -76,7 +76,7 @@ def cmd_add(
     # Read from file
     if file:
         try:
-            raw = Path(file).read_text()
+            raw = Path(file).read_text(encoding="utf-8")
             msgs = json.loads(raw)
         except (FileNotFoundError, json.JSONDecodeError) as e:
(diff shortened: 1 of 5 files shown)


Ghraven:

PR #6543, opened 2026-07-24
fix(ts-oss/lmstudio): honor LMSTUDIO_BASE_URL env var

### LINKED ISSUE #6542: TS OSS LM Studio LLM and embedder ignore the LMSTUDIO_BASE_URL env var
### Component

TypeScript SDK / LLMs + Embedders

### Where

`mem0-ts/src/oss/src/llms/lmstudio.ts` and `mem0-ts/src/oss/src/embeddings/lmstudio.ts`.

### What happens

Both resolve the base URL from config-or-default only, with no env fallback:

```ts
// llms/lmstudio.ts
baseURL: config.baseURL ?? DEFAULT_BASE_URL,

// embeddings/lmstudio.ts
const baseURL = config.baseURL ?? config.url ?? DEFAULT_BASE_URL;
```

Every sibling OpenAI-compatible provider in this SDK reads its env var (`process.env.VLLM_BASE_URL`, `DEEPSEEK_API_BASE`, `SARVAM_API_BASE`, `MINIMAX_API_BASE`, `TOGETHER_API_BASE`), but LM Studio is the only one that does not, so a user who sets `LMSTUDIO_BASE_URL` always hits `http://localhost:1234/v1`.

### Expected

Precedence: explicit `baseURL` config > `LMSTUDIO_BASE_URL` env > `http://localhost:1234/v1` default, matching the other providers.

### Note

This is the TS counterpart of #6526 (the Python `LMStudioConfig` has the same gap). Both the LLM and the embedder are affected in TS.

abhay-codes07:
### Description

The TS LM Studio LLM and embedder resolved the base URL from config-or-default only, with no env fallback, so a user who set `LMSTUDIO_BASE_URL` always hit `http://localhost:1234/v1`. Every sibling OpenAI-compatible provider in this SDK already reads its base-URL env var (`VLLM_BASE_URL`, `DEEPSEEK_API_BASE`, `SARVAM_API_BASE`, `MINIMAX_API_BASE`, `TOGETHER_API_BASE`); LM Studio was the only one that did not.

The fix resolves the base URL as explicit config > `LMSTUDIO_BASE_URL` env > default in both the LLM and the embedder. This is the TS counterpart of #6526 (the Python config has the same gap).

Closes #6542

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/lmstudio-llm.test.ts src/oss/tests/lmstudio-embedder.test.ts` (run from the `mem0-ts` root), 15 passed.

Added to both suites:
- `honors LMSTUDIO_BASE_URL when no baseURL is configured`
- `prefers an explicit baseURL over LMSTUDIO_BASE_URL`
- `falls back to the localhost default when neither is set`

Reverting the two source changes fails exactly the two `honors LMSTUDIO_BASE_URL` tests.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/embeddings/lmstudio.ts
+++ b/mem0-ts/src/oss/src/embeddings/lmstudio.ts
@@ -12,7 +12,12 @@ export class LMStudioEmbedder implements Embedder {
   private model: string;
 
   constructor(config: EmbeddingConfig) {
-    const baseURL = config.baseURL ?? config.url ?? DEFAULT_BASE_URL;
+    // Precedence: explicit config > LMSTUDIO_BASE_URL env > default (#6526).
+    const baseURL =
+      config.baseURL ||
+      config.url ||
+      process.env.LMSTUDIO_BASE_URL ||
+      DEFAULT_BASE_URL;
     const apiKey = config.apiKey || DEFAULT_LMSTUDIO_API_KEY;
     this.openai = new OpenAI({ apiKey, baseURL: String(baseURL) });
     this.model = config.model || DEFAULT_MODEL;
--- a/mem0-ts/src/oss/src/llms/lmstudio.ts
+++ b/mem0-ts/src/oss/src/llms/lmstudio.ts
@@ -12,7 +12,10 @@ export class LMStudioLLM extends OpenAILLM {
     super({
       ...config,
       apiKey: config.apiKey || DEFAULT_LMSTUDIO_API_KEY,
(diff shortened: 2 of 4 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss fix with tests. TS counterpart of #6526: the LM Studio LLM and embedder were the only OpenAI-compatible providers not reading their base-URL env var, so LMSTUDIO_BASE_URL was ignored and users always hit localhost. Now resolves config > LMSTUDIO_BASE_URL > default like vllm/deepseek/etc. Quick look when you get a chance, thanks.

PR #6547, opened 2026-07-24
fix(vector_stores/s3_vectors): make search score metric-aware

### LINKED ISSUE #6546: S3 Vectors search score is hardcoded for cosine, breaks euclidean ranking
The S3 Vectors store supports both `cosine` and `euclidean` distance metrics (see `S3VectorsConfig.distance_metric`), and the chosen metric is passed to `create_index` as `distanceMetric`. But `_parse_output` in `mem0/vector_stores/s3_vectors.py` always converts distance to a score with:

```python
score = max(0.0, 1.0 - raw_distance)
```

This is only correct for cosine (where distance is in `[0, 2]`). Euclidean/L2 distance is unbounded, so for any distance `> 1` the score clamps to `0.0`. On a euclidean index almost every result collapses to a score of `0`, which destroys ranking and any score-threshold filtering.

## Repro

Configure the store with `distance_metric="euclidean"` and search. Results with distance `> 1` all get score `0.0`, so ordering is lost.

## Expected

The distance-to-similarity conversion should depend on the configured metric: keep `1 - d` for cosine, and use a bounded monotonic map (e.g. `1 / (1 + d)`) for euclidean so scores stay in `(0, 1]` and preserve ranking.

I have a fix with regression tests ready to open as a PR.

yashs33244:
## Linked Issue

Closes #6546

## Description

The S3 Vectors store hardcoded the search score as `max(0.0, 1.0 - raw_distance)` in `_parse_output`, which is only correct for cosine distance. The store already supports `euclidean` (see `S3VectorsConfig.distance_metric`, passed to `create_index` as `distanceMetric`), where distance is unbounded, so `1 - distance` clamps almost every result to `0.0` and destroys ranking and score thresholds.

This branches the distance-to-similarity conversion on `self.distance_metric`:
- **cosine**: keep `max(0.0, 1.0 - d)` (distance is in `[0, 2]`)
- **euclidean**: use the bounded monotonic map `1.0 / (1.0 + d)`, so scores stay in `(0, 1]` and preserve ordering

The `None`-guard for missing distances (used by `get`/`list`) is preserved.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A. Cosine (the default) is unchanged; only euclidean scoring is corrected.

## Test Coverage

Added two regression tests in `tests/vector_stores/test_s3_vectors.py`: one asserts euclidean uses the bounded map and does not collapse to `0` (fails without the fix), one pins the cosine `1 - d` mapping.

## Checklist


yashs33244:
Open since 07-24 with green CI, no eyes yet 🙏 Makes the S3 Vectors search score metric-aware (euclidean was scored as if cosine). Small, additive, tested. @kartik-mem0 @deshraj a review whenever one of you has a window?

mem0/vector_stores/s3_vectors.py
```diff
@@ -69,6 +69,16 @@ def create_col(self, name, vector_size, distance="cosine"):
             else:
                 raise
 
+    def _distance_to_score(self, raw_distance: Optional[float]) -> Optional[float]:
+        if raw_distance is None:
+            return None
+        # Euclidean distance is unbounded, so 1 - distance would collapse most
+        # scores to 0. Use a bounded monotonic map instead. Cosine distance is
+        # in [0, 2], where 1 - distance is already a valid similarity.
+        if self.distance_metric == "euclidean":
```
PowderXu, line 78: Small, non-blocking question: am I understanding correctly that AWS rejects unsupported metric values when creating a new index, but that validation is skipped when the index already exists?
If so, a typo such as "euclidian" could still allow searches against an existing Euclidean index, while _distance_to_score falls back to the cosine formula and produces incorrect scores.
Ideally a shared configuration should prevent this, but not everyone may use one. Would validating the metric or using S3’s returned distanceMetric help?

PR #6551, opened 2026-07-24
fix(llms/aws_bedrock): raise on parse failure instead of returning a sentinel

### LINKED ISSUE #6548: AWS Bedrock LLM returns "Error parsing response" string on parse failure instead of raising
### Summary

In `mem0/llms/aws_bedrock.py`, `_parse_response` returns the literal string `"Error parsing response"` when it fails to parse a Bedrock reply, instead of raising:

```python
except Exception as e:
    logger.warning(f"Could not parse response: {e}")
    return "Error parsing response"
```

Because this is a plausible-looking string, a failed Bedrock call flows downstream as if it were a genuine model answer. During `add()` it can be treated as extracted content and stored as a "memory", silently corrupting the memory store. There is no way for the caller to distinguish this from a real response.

### Expected

A parse failure should raise (as `generate_response`'s own outer handler already does for API errors), so the failure is loud and never mistaken for a model answer. `LLMError` is the repo-wide contract for LLM failures.

### Impact

Any Bedrock model whose response can't be parsed (malformed/empty body, unexpected shape) returns the sentinel string rather than erroring. This is silent data corruption, not a crash, so it is easy to miss.

### Environment

- `mem0/llms/aws_bedrock.py`, `_parse_response`

yashs33244:
## Linked Issue

Closes #6548

## Description

`_parse_response` in `mem0/llms/aws_bedrock.py` returned the literal string `"Error parsing response"` when parsing a Bedrock reply failed, instead of raising. Because that is a plausible-looking string, a failed Bedrock call flowed downstream as if it were a real model answer and could be stored as a "memory", silently corrupting the memory store.

This raises `LLMError` (the repo's contract for LLM failures, already used in `memory/main.py`) with the provider and original error as context, so the failure surfaces loudly instead of masquerading as a response. The chain uses `from e` to preserve the underlying cause.

Only the genuine parse-failure branch changed. Legitimate empty responses are unaffected: they return `""` via the existing `.get(...)` defaults inside the `try` and never reach the `except`.

While adding the regression tests, two existing Nova tests were passing only because the old sentinel swallowed a parse error (the Nova no-tool path calls `_parse_response` on a Converse-shaped response). I gave those two tests a body-bearing mock so they keep asserting what they intend (the `topP` in the Converse call args) rather than relying on the swallowed error.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A. A parse failure previously returned a string; it now raises, which is the intended failure behavior. No caller depended on the sentinel string.

## Test Coverage

Added two regression tests (`test_malformed_body_raises_instead_of_returning_sentinel`, `test_missing_body_raises`) that assert `_parse_response` raises `LLMError` on unparseable input. Both fail without the fix and pass with it. Full `tests/llms/` suite passes (205 tests); `ruff check` clean.

## Checklist


mem0/llms/aws_bedrock.py
```diff
@@ -11,6 +11,7 @@
 
 from mem0.configs.llms.base import BaseLlmConfig
 from mem0.configs.llms.aws_bedrock import AWSBedrockConfig
+from mem0.exceptions import LLMError
```
kartik-mem0, line 14: `main` reordered this import block in #6899, so this line conflicts and the PR is currently unmergeable. A rebase onto `main` clears it. As it stands the merged file keeps the conflict markers and the whole test module fails collection with a `SyntaxError`.

tests/llms/test_aws_bedrock.py
```diff
@@ -311,7 +321,7 @@ def test_with_tools_anthropic_top_p_set_still_omitted(self, mock_boto3):
         assert "topP" not in kwargs["inferenceConfig"]
 
     def test_nova_includes_top_p_when_explicitly_set(self, mock_boto3):
-        mock_boto3.converse.return_value = _converse_response()
+        mock_boto3.converse.return_value = _nova_body_response()
```
kartik-mem0, line 324: These two tests passed with `_converse_response()`, which is what the mocked `converse` actually returns. `{"body": ...}` is a shape Converse never produces, and swapping it in is what keeps them green: with the real shape this path now raises instead of returning the sentinel. Either teach `_parse_response` the Converse shape, or restore the mocks and assert the raise.

kartik-mem0 (commented):
## Changes requested

Replacing the `"Error parsing response"` sentinel with a raise is the right fix, that string flows on into the pipeline and can be extracted and stored as a memory. Two things before it can land.

**To unblock:**
1. `mem0/llms/aws_bedrock.py:14`: rebase onto `main`, the import block conflicts (detail inline)
2. `tests/llms/test_aws_bedrock.py:324`: the Nova tests now mock a response Converse never returns (detail inline)

Optional: `generate_response` catches bare `Exception` at `aws_bedrock.py:466` and re-raises `RuntimeError`, so `LLMError` never actually reaches a caller of the public method. The new tests call `_parse_response` directly, so they do not notice.

- **Problem is real:** `mem0/llms/aws_bedrock.py:431-433` on `main` returns the literal string `"Error parsing response"` on any parse failure, and the caller cannot tell it apart from model output
- **Tests / red-green:** reverse-applied the impl hunk only, both new tests failed with `DID NOT RAISE LLMError`, restored -> 43 passed ✓. The two *edited* Nova tests are the problem: reverting only the test hunk while keeping the fix gives `RuntimeError: Failed to generate response: Failed to parse Bedrock response for provider 'amazon': 'NoneType' object has no attribute 'read'` on both of them.
- **CI:** all real checks green, but CI runs the branch on its own. Merged with current `main` the file keeps conflict markers and the test module fails collection with a `SyntaxError`. Vercel is red on every fork PR and does not block.
- **Security:** cleared, the raised message carries the provider name and the exception text, no credentials.
- **Bloat (ponytail):** lean, +2 impl lines.
- **Not verified:** live Bedrock. Whether hard-failing every Nova no-tools call is acceptable is a product call, I am only flagging that the current tests hide it.

</details>

`LLMError` with `LLM_001` is exactly the right type here and `raise ... from e` keeps the cause, so the core of this is good as written. Push when ready and I will re-review.

yashs33244:
Open ~19 days, green CI, still unreviewed 🙏 Bedrock now raises on parse failure instead of returning a sentinel that silently corrupts downstream. Regression test included. @kartik-mem0 @deshraj a look when you get a chance?

PR #6552, opened 2026-07-24
fix(vector_stores/redis): support operator dicts and $or in filters

### LINKED ISSUE #6550: Redis vector store ignores operator dicts and $or in filters (only scalar equality works)
### What happened

In `mem0/vector_stores/redis.py`, `search`, `keyword_search`, and `list` build filter conditions the same way:

```python
```

This only handles scalar equality. mem0's universal filter format also allows operator dicts (e.g. `{"age": {"gte": 18}}`) and logical operators like `$or`. With the current code a dict value becomes `Tag("age") == {"gte": 18}`, which is a nonsense tag match, so advanced filters silently return empty or wrong results instead of being applied.

Other stores handle this: pgvector has an `OPERATOR_SQL_MAP` for `eq/ne/gt/gte/lt/lte/in/nin/contains/icontains` plus `$or`/`$not`.

### Expected

Operator dicts and `$or` should be translated to the correct Redis query (numeric ranges via redisvl `Num`, tag membership via `Tag`, text match via `Text`), matching the universal filter format the rest of the SDK uses.

### Repro

```python
m.search("query", user_id="alice", filters={"age": {"gte": 18}})  # returns wrong/empty results
```

### Notes

redisvl (`Tag`, `Num`, `Text`) can express all the scalar operators and `$or`. `$not` on a compound group is not directly expressible with redisvl, so it should raise a clear error rather than be silently dropped (follow-up).

yashs33244:
## Linked Issue

Closes #6550

## Description

The Redis vector store built filter conditions as `Tag(key) == value` for every entry in `search`, `keyword_search`, and `list`, so only scalar equality worked. An operator dict like `{"age": {"gte": 18}}` became a nonsense `Tag("age") == {"gte": 18}` match and returned empty/wrong results, even though mem0's universal filter format allows operator dicts and `$or`.

This adds a shared `_build_filter_expression` helper and routes the three call sites through it.

Covered:
- Scalar equality and list membership (unchanged for existing filters).
- Operator dicts: `eq`, `ne`, `gt`, `gte`, `lt`, `lte`, `in`, `nin`, `contains`, `icontains`, mapped to the right redisvl field type (`Tag` for strings, `Num` for numeric ranges, `Text` for contains). `contains`/`icontains` both map to a `Text` match since Redis full-text search is case-insensitive.
- Top-level `$or`.

Deferred:
- `$not` is rejected with a clear `ValueError` rather than silently ignored, because redisvl cannot negate a compound expression (`~` on a `FilterExpression` raises). Left as a follow-up.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A. Scalar filters produce the same expressions as before.

## Test Coverage

New tests assert the helper produces the correct redisvl `FilterExpression` string for scalars, lists, every supported operator, `$or`, and the rejected `$not`/unknown-operator cases. They run without a live Redis and fail against the old scalar-only builder.

## Checklist


yashs33244:
Sitting green since 07-24 🙏 Redis filters now honor operator dicts + `$or` instead of forcing `Tag==value` (which broke gt/lt/in/etc). 17 tests. @kartik-mem0 @deshraj would appreciate a review whenever there's bandwidth.

PR #6553, opened 2026-07-24
fix(ts-oss): enforce '*' metadata filter as field-exists

Bartok9:
## Summary

- Enforce documented `filters[field] = "*"` **field-exists** semantics across TypeScript OSS vector stores that could express it.
- Client-side matchers (memory, Cassandra) no longer treat missing/`null` payload keys as matches.
- Pinecone emits `{ $exists: true }`; Qdrant uses `must_not: is_empty`; Milvus emits `exists metadata["key"]`.
- Chroma still skips `*` (no exists operator that excludes missing keys) but no longer risks a silent precisamente-wrong path; comment clarifies.

## Motivation

Related to #6539 (does **not** close it alone — Python providers are covered by #6540).

Same cross-provider contract failure called out after OpenSearch #6522: `"*"` must mean “field is present,” not “ignore this filter” or “literal star.”

## Differential value

#6540 by @microbluey fixes the **Python** providers for #6539. This PR is the complementary **TypeScript OSS** surface (`mem0-ts/src/oss/src/vector_stores/*`) so both SDKs share the contract.

## Verification

- `cd mem0-ts && pnpm exec jest src/oss/tests/wildcard-field-exists.test.ts --coverage=false` — 5 passed, 0 failed
- Did NOT change: OpenSearch adapter (already correct), server REST, Python packages

## Notes

- No reviewer requests / no labels from this bot.

### DIFF
--- a/mem0-ts/src/oss/src/tests/milvus.test.ts
+++ b/mem0-ts/src/oss/src/tests/milvus.test.ts
@@ -267,34 +267,34 @@ describe("Milvus vector store (TS OSS SDK)", () => {
     });
   });
 
-  it("skips a wildcard '*' filter value and keeps the rest", async () => {
+  it("treats wildcard '*' as field-exists and keeps the rest", async () => {
     const client = new FakeMilvusClient({ existing: ["mem0"] });
     client.searchResponse = { results: [] };
     const store = makeStore(client, { metricType: "COSINE" });
     await store.initialize();
 
-    // "*" means match-any: it must be dropped, not emitted as `== "*"` (which
-    // matches nothing), leaving only the real agent_id clause.
+    // "*" means field-exists (not drop, not literal == "*").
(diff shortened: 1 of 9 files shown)


microbluey:
Nice to see the TS side moving. FYI I've opened #6568 covering the two TS OSS stores outside this PR's scope (`redis.ts`, `azure_ai_search.ts`), mirroring the Python follow-up #6566 — so between the two PRs the TS OSS surface for the `*` contract should be complete. No overlap in files, so they can land independently.

Bartok9:
Thanks @microbluey — perfect split. This PR intentionally left `redis.ts` / `azure_ai_search.ts` alone so the six stores here can land independently of the dialect-2 / AI Search query-shape work.

Confirmed no file overlap with #6568. Happy to stay sequenced either order; once both are in, the TS OSS `*` field-exists contract should match the Python surface from #6540 + #6566.

PR #6558, opened 2026-07-24
fix(llms/aws_bedrock): include content in tool-enabled Converse responses

### LINKED ISSUE #6556: AWS Bedrock tool responses omit the "content" key (drops assistant text)
The Converse tool-response branch in `mem0/llms/aws_bedrock.py` `_parse_response` returns:

```python
processed_response = {"tool_calls": []}
```

Every other LLM provider returns `{"content": <text-or-None>, "tool_calls": [...]}` (see `openai.py`, `anthropic.py`, `gemini.py`, `groq.py`, etc.). Two problems follow:

1. Callers that read `resp["content"]` / `resp.get("content")` on a Bedrock tool response get a `KeyError` or `None` where other providers give the assistant text.
2. When the model returns a text block alongside `toolUse` blocks, that text is silently discarded - only `toolUse` items are parsed.

**Expected:** the tool branch should return a `"content"` key (the assistant text, or `None` when absent), matching the other providers.

**Affected:** `mem0/llms/aws_bedrock.py`, `_parse_response` tool branch (Anthropic/Amazon/Cohere models via Converse).

yashs33244:
## Linked Issue

Closes #6556

## Description

The Converse tool-response branch of `_parse_response` in `mem0/llms/aws_bedrock.py` returned `{"tool_calls": [...]}` with no `"content"` key. Every other provider returns `{"content": <text-or-None>, "tool_calls": [...]}`, so callers doing `resp.get("content")` got `None`/`KeyError` on Bedrock, and any assistant text co-occurring with `toolUse` blocks was silently dropped.

This initializes `content` to `None` and captures `text` blocks, mirroring the Anthropic provider. Only the tool branch is touched; the non-tool parsing path is unchanged.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `TestParseResponseTools` in `tests/llms/test_aws_bedrock.py`: one case with both a text and a `toolUse` block (asserts text is kept in `content` and the tool call is parsed), and one tool-only case (asserts `content` is present and `None`). Both fail on `main` and pass with this change. Full `tests/llms/test_aws_bedrock.py` passes (43) and `ruff check` is clean.

## Checklist


mem0/llms/aws_bedrock.py
```diff
@@ -369,11 +369,13 @@ def _parse_response(
...
         if tools:
             # Handle tool-enabled responses
-            processed_response = {"tool_calls": []}
+            processed_response = {"content": None, "tool_calls": []}
 
             if response.get("output", {}).get("message", {}).get("content"):
                 for item in response["output"]["message"]["content"]:
-                    if "toolUse" in item:
+                    if "text" in item:
+                        processed_response["content"] = item["text"]
```
kartik-mem0, line 377: Converse can return more than one text block in a message, and each pass through here overwrites the last, so the final block wins and the earlier ones are still dropped. Keeping the first block or joining them would match what the caller expects. Minor, not worth holding the fix for.

kartik-mem0 (commented):

The tool branch of `_parse_response` returned `{"tool_calls": [...]}` with no `content` key, so Bedrock was the one provider diverging from the shape every other one returns (`mem0/llms/openai.py:67-70`), and any assistant text sitting beside a `toolUse` block was dropped on the floor.

**Before you merge:** nothing, self-contained, merges clean onto current `main` (54 tests green).
**Follow-up:** none.
**Not verified:** live Bedrock. Also worth knowing: nothing in `mem0/` passes `tools=` to this class today (`mem0/proxy/main.py:129` routes through litellm), so this fixes the dropped text and the provider inconsistency rather than a crash someone is hitting right now.

Risk: low, one added dict key, no in-tree caller reads it yet · Your time: ~2 min

- **Problem is real:** `mem0/llms/aws_bedrock.py:378` on `main` builds `{"tool_calls": []}` and the loop at :381 only handles `toolUse`, so a Converse message carrying both a text block and a toolUse block loses the text, and `response["content"]` raises `KeyError`
- **Tests:** `pytest tests/llms/test_aws_bedrock.py -q` -> 43 passed
- **Red-green:** reverse-applied the impl hunk only, both new tests failed with `KeyError: 'content'`, restored -> 43 passed ✓
- **CI:** all real checks green (Python SDK 3.10/3.11/3.12, CI Gate). Vercel is red on every fork PR and does not block.
- **Security:** cleared, no untrusted input and no secrets, response-dict reads only.
- **Bloat (ponytail):** lean already, +4 impl lines.
- **Linked issue:** #6556

</details>

yashs33244:
Green since 07-24, no review yet 🙏 Bedrock tool-enabled Converse responses now include the `content` key every other provider returns (avoids a KeyError + dropped assistant text). @kartik-mem0 @deshraj a quick look when you can?

PR #6559, opened 2026-07-24
fix(vector_stores/turbopuffer): make search score respect distance_metric

### LINKED ISSUE #6557: Turbopuffer search scores are wrong for euclidean_squared distance
## Description

The Turbopuffer vector store computes search scores with a hardcoded cosine formula that ignores the configured `distance_metric`.

In `mem0/vector_stores/turbopuffer.py`, `_parse_output` does:

```python
score = 1 - dist if dist is not None else None
```

This is only correct for cosine. `distance_metric` also supports `euclidean_squared`, where Turbopuffer returns an unbounded squared-L2 `$dist` (e.g. `4.0`). `1 - 4.0 = -3.0`, so the returned score is negative and unbounded.

## Impact

`VectorStoreBase` treats scores as higher-is-better. Under `euclidean_squared`:
- scores can be negative and are not comparable to cosine scores,
- similarity thresholds / `relevance_score` filtering in mem0 break.

Other providers already handle this: `milvus` and `baidu` map L2 distances to `1 / (1 + dist)`.

## Expected

Score should be metric-aware and respect the higher-is-better contract for both `cosine_distance` and `euclidean_squared`.

## Steps to reproduce

Configure Turbopuffer with `distance_metric="euclidean_squared"` and run a search. A row with `$dist = 4.0` yields `score = -3.0` instead of a bounded 0-1 similarity.

yashs33244:
## Linked Issue

Closes #6557

## Description

`TurbopufferDB._parse_output` computed `score = 1 - dist` for every distance metric. That is only correct for cosine. When `distance_metric="euclidean_squared"`, Turbopuffer returns an unbounded squared-L2 `$dist` (e.g. `4.0`), so `1 - 4.0 = -3.0` - a negative, unbounded score that breaks the `VectorStoreBase` higher-is-better contract and mem0's similarity thresholds.

The fix branches on `self.distance_metric`:
- `euclidean_squared`: bounded, monotonic `1 / (1 + dist)` (mirrors the milvus/baidu providers),
- cosine (default): unchanged `1 - dist`.

The `None` guard is preserved, and only the score computation in `_parse_output` is touched.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added regression tests in `tests/vector_stores/test_turbopuffer.py`: the `euclidean_squared` case asserts `score == 0.2` for `$dist = 4.0` and stays within `[0, 1]` (fails without the fix, returning `-3.0`), the cosine path stays `1 - dist`, and the `None` guard holds on the euclidean path.

## Checklist


yashs33244:
Open ~19 days, CI green 🙏 Turbopuffer search score now respects `distance_metric` (euclidean was giving negative scores that broke thresholds). 3 tests. @kartik-mem0 @deshraj a review whenever there's time?

PR #6564, opened 2026-07-24
fix(vector_stores/turbopuffer): apply all filter operators instead of dropping them

### LINKED ISSUE #6562: Turbopuffer vector store silently drops filter operators other than gte/lte
The Turbopuffer backends `_convert_filters` (`mem0/vector_stores/turbopuffer.py`) only handles `gte` and `lte` inside operator dicts:

```python
if isinstance(value, dict):
    if "gte" in value:
        conditions.append((key, "Gte", value["gte"]))
    if "lte" in value:
        conditions.append((key, "Lte", value["lte"]))
else:
    conditions.append((key, "Eq", value))
```

Any other operator mem0 supports (`gt`, `lt`, `ne`, `eq`, `in`, `nin`) matches neither branch, so a filter like `{"age": {"gt": 18}}` is silently dropped. The query then runs unfiltered (or under-filtered), returning too-broad results with no error.

Other backends (chroma, pgvector, pinecone, oracledb) all support the full `eq/ne/gt/gte/lt/lte/in/nin` set, so Turbopuffer is inconsistent here.

### Expected
Each operator in a filter dict is translated to its Turbopuffer equivalent, and a genuinely unknown operator raises a clear error rather than being dropped.

### Reproduce
```python
db._convert_filters({"age": {"gt": 18}})  # returns None -> filter dropped
```

yashs33244:
## Linked Issue

Closes #6562

## Description

`_convert_filters` in the Turbopuffer vector store only handled `gte` and `lte` for operator dicts. Any other operator mem0 supports (`gt`, `lt`, `ne`, `eq`, `in`, `nin`) matched neither branch, so a filter like `{"age": {"gt": 18}}` was silently dropped and the query ran unfiltered, returning too-broad results with no error.

This maps the full mem0 operator set to the Turbopuffer filter tokens (verified against the SDKs `Filter` type in `turbopuffer/types/custom.py`: `eq->Eq, ne->NotEq, gt->Gt, gte->Gte, lt->Lt, lte->Lte, in->In, nin->NotIn`) and raises a clear `ValueError` on an unknown operator instead of dropping it. Existing scalar-equality and `gte`/`lte` behavior is unchanged.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added regression tests in `tests/vector_stores/test_turbopuffer.py`: every operator maps to the right token, `{"gt": 18}` is no longer dropped, multiple operators on one field combine with `And`, and an unknown operator raises `ValueError`. Verified the new tests fail on the unpatched source.

## Checklist

- [x] My code follows the projects style guidelines


yashs33244:
Open ~19 days, green CI, unreviewed 🙏 Turbopuffer `_convert_filters` now applies all operators (gt/lt/ne/in/nin) instead of silently dropping them; unknown ops raise. 4 tests. @kartik-mem0 @deshraj would appreciate a review.

PR #6565, opened 2026-07-24
fix(llms/aws_bedrock): send Converse-correct messages for amazon and cohere tool calls

### LINKED ISSUE #6563: AWS Bedrock: tool calls send malformed Converse requests for amazon and cohere models
In \`AWSBedrockLLM._generate_with_tools\` (mem0/llms/aws_bedrock.py), the message formatting is wrong for the non-Anthropic providers that support tools (amazon and cohere), so the request sent to the Bedrock Converse API is malformed.

## Problem

Converse requires the system prompt via the top-level \`system=\` param (never a \`role="system"\` message) and message content as a list of blocks (\`[{"text": ...}]\`). The Anthropic branch does this correctly; the other two do not:

- **amazon** calls \`_format_messages_amazon\`, which emits \`{"role": "system", "content": "<str>"}\` and plain string content. Converse rejects \`role="system"\` messages and expects list-of-blocks content, so the request is invalid.
- **cohere / else** builds \`[{"role": "user", "content": [{"text": messages[-1]["content"]}]}]\` - it keeps only the last message, silently dropping the system prompt and every prior turn.

## Expected

For all tool-capable providers, non-system messages should be Converse-shaped (\`{"role": r, "content": [{"text": c}]}\`) and the system prompt should be hoisted into the top-level \`system=[{"text": ...}]\` param, exactly like the Anthropic branch already does.

## Impact

Tool calls with an amazon (e.g. \`amazon.nova-*\`) or cohere model either fail validation or lose the system prompt and conversation history.

yashs33244:
## Linked Issue

Closes #6563

## Description

In `AWSBedrockLLM._generate_with_tools`, the message formatting was wrong for the non-Anthropic providers that support tools, so the request sent to the Bedrock Converse API was malformed:

- **amazon** called `_format_messages_amazon`, which emits `{"role": "system", "content": "<str>"}` and plain string content. Converse rejects `role="system"` messages and requires content as a list of blocks, so the request was invalid.
- **cohere / else** built `[{"role": "user", "content": [{"text": messages[-1]["content"]}]}]`, keeping only the last message and silently dropping the system prompt and all prior turns.

The fix routes both through a small `_format_messages_converse` helper that hoists the system prompt into the top-level `system=` param and formats every non-system turn as `{"role": r, "content": [{"text": c}]}` - mirroring the already-correct Anthropic branch. The Anthropic path and tool config are unchanged.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `TestGenerateWithToolsMessageShape` in `tests/llms/test_aws_bedrock.py`: for amazon (`amazon.nova-*`) and cohere models with tools, it mocks `client.converse` and asserts no `role="system"` message is sent, every content is a list of `{"text": ...}` blocks, the system prompt lands in the top-level `system` param, and all prior turns are preserved (not just the last). These tests fail without the fix. The existing `test_with_tools_anthropic_*` tests still pass. Full file: 44 passed.

## Checklist


kartik-mem0 (commented):

On the tool path amazon sent `{"role": "system", "content": "<str>"}` (Converse rejects both the role and the bare string) while the cohere/else branch rebuilt the list from `messages[-1]` alone, silently dropping the system prompt and every prior turn. `_format_messages_converse` fixes both in one place.

**Before you merge:** nothing, merges clean onto current `main` (55 tests green).
**Follow-up:** `_format_messages_amazon` survives on the Nova no-tools path at `aws_bedrock.py:625` and still emits the old shape; fixing that one also needs `_parse_response` to understand Converse output, so it wants its own issue (noted inline).
**Not verified:** live Bedrock, the Converse request is asserted against the mocked client.

Risk: low, request shaping for one provider family, three new tests cover it · Your time: ~4 min

- **Problem is real:** two paths on `main`, `mem0/llms/aws_bedrock.py:531` routes amazon into `_format_messages_amazon` (:184-200) which emits `role="system"` with string content, and :533 keeps only `messages[-1]` for every other provider
- **Tests:** `pytest tests/llms/test_aws_bedrock.py -q` -> 44 passed
- **Red-green:** reverse-applied the impl hunk only, all 3 new tests failed (`KeyError: 'system'` twice, and the single-message case on `{'role':'user','content':'Hello'}` vs the block shape), restored -> 44 passed ✓
- **CI:** all real checks green (Python SDK 3.10/3.11/3.12, CI Gate). Vercel is red on every fork PR and does not block.
- **Security:** cleared, message reshaping only, no untrusted input and no secrets.
- **Bloat (ponytail):** lean, the helper has two callers so it is not speculative. `_format_messages_anthropic` now overlaps it heavily, noted inline.
- **Linked issue:** #6563

</details>

mem0/llms/aws_bedrock.py
```diff
@@ -515,16 +515,39 @@ def _build_inference_config(self) -> Dict[str, Any]:
...
+        # Format messages for tool-enabled models. All supported providers use the
+        # Converse API, so messages must be Converse-shaped and the system prompt
+        # hoisted into the top-level `system` param (see _format_messages_converse).
         if self.provider == "anthropic":
             formatted_messages, system_message = self._format_messages_anthropic(messages)
-        elif self.provider == "amazon":
-            formatted_messages = self._format_messages_amazon(messages)
         else:
-            formatted_messages = [{"role": "user", "content": [{"text": messages[-1]["content"]}]}]
+            formatted_messages, system_message = self._format_messages_converse(messages)
```
kartik-mem0, line 550: This leaves `_format_messages_amazon` with exactly one caller, the Nova no-tools path in `_generate_standard`, which still gets the old `role="system"` plus string-content shape. Fixing that one also needs `_parse_response` to understand Converse output, so it is genuinely a separate issue. Fine to leave here, worth filing.

mem0/llms/aws_bedrock.py
```diff
@@ -515,16 +515,39 @@ def _build_inference_config(self) -> Dict[str, Any]:
...
+
     def _generate_with_tools(self, messages: List[Dict[str, str]], tools: List[Dict], stream: bool = False) -> Dict[str, Any]:
         """Generate response with tool calling support using correct message format."""
-        # Format messages for tool-enabled models
-        system_message = None
+        # Format messages for tool-enabled models. All supported providers use the
+        # Converse API, so messages must be Converse-shaped and the system prompt
+        # hoisted into the top-level `system` param (see _format_messages_converse).
         if self.provider == "anthropic":
             formatted_messages, system_message = self._format_messages_anthropic(messages)
```
kartik-mem0, line 548: The anthropic helper now does nearly what the new one does, the differences being that it drops non-standard roles and keeps only the last system message. Worth collapsing the two into one Converse formatter later. Not for this PR.

yashs33244:
Green since 07-24 🙏 Sends Converse-correct messages for amazon + cohere tool calls (was malformed: system role + string content / dropped history). Anthropic path unchanged, 3 tests. @kartik-mem0 @deshraj a look whenever you can?

PR #6570, opened 2026-07-24
fix(ts-oss/gemini): send the system prompt via systemInstruction

### LINKED ISSUE #6569: TS OSS Gemini sends the system prompt as a 'model' content turn instead of systemInstruction
### Component

TypeScript SDK / LLMs

### Where

`mem0-ts/src/oss/src/llms/google.ts`, `formatContents()`.

### What happens

```ts
role: msg.role === "system" ? "model" : "user",
```

Every message is turned into a Gemini content turn, with `system` mapped to role `"model"` and everything else (including `assistant`) to `"user"`. Two problems:

1. The **system prompt** — which carries the fact-extraction instructions — is sent as a `"model"` content turn, i.e. as if the model had already said it. Gemini takes system prompts through `systemInstruction`, and a conversation with a leading `"model"` turn has no opening user turn. The extraction instructions are effectively mislabeled.
2. **assistant** turns are mapped to `"user"`, so any prior assistant message in a multi-turn call is relabeled as the user, corrupting the conversation.

### Expected

Mirror the Python SDK's `_reformat_messages` (`mem0/llms/gemini.py`): pull `system` messages out into `systemInstruction` (config), keep only real turns in `contents`, and map `user`→`"user"`, `assistant`→`"model"`.

### Note

#6468 recently made this same file honor `responseFormat` for Python parity but did not touch the role mapping, so the system-instruction handling is still off.

abhay-codes07:
### Description

`formatContents()` in the TS Gemini provider turned every message into a content turn, mapping `system` to role `"model"` and `assistant` to `"user"`. As a result:

1. The system prompt (which carries the fact-extraction instructions) was sent as a `"model"` turn, i.e. as if the model had already said it, leaving the conversation without a leading user turn.
2. Prior `assistant` turns were relabeled as `"user"`, corrupting multi-turn calls.

The fix mirrors the Python SDK's `_reformat_messages` (`mem0/llms/gemini.py`): system messages are pulled into `config.systemInstruction`, only real turns stay in `contents`, and `user`→`"user"` / `assistant`→`"model"`. Applied to both `generateResponse` and `generateChat`.

Context: #6468 recently made this same file honor `responseFormat` for Python parity but did not touch the role mapping. (The stale, conflicting #4390 flips assistant to `"model"` but still leaves the system prompt as a `"model"` content turn rather than `systemInstruction`.)

Closes #6569

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/google-llm.test.ts` (run from the `mem0-ts` root), 11 passed.

- Added `extracts the system prompt into config.systemInstruction`: the system message lands in `config.systemInstruction` and no longer leaks into `contents`.
- Added `maps assistant turns to the model role in contents`.
- Updated the existing generateChat test, which asserted the old behavior (system as a `"model"` content turn), to expect the corrected `systemInstruction` handling.
- Reverting the source change fails all three.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/llms/google.ts
+++ b/mem0-ts/src/oss/src/llms/google.ts
@@ -23,18 +23,39 @@ export class GoogleLLM implements LLM {
     this.google = new sdk.GoogleGenAI({ apiKey: this.apiKey });
   }
 
-  private formatContents(messages: Message[]) {
-    return messages.map((msg) => ({
-      parts: [
-        {
-          text:
-            typeof msg.content === "string"
-              ? msg.content
-              : JSON.stringify(msg.content),
-        },
-      ],
-      role: msg.role === "system" ? "model" : "user",
-    }));
+  private formatContents(messages: Message[]): {
+    systemInstruction?: string;
+    contents: Array<{ parts: Array<{ text: string }>; role: string }>;
+  } {
+    let systemInstruction: string | undefined;
+    const contents: Array<{ parts: Array<{ text: string }>; role: string }> =
+      [];
+
+    for (const msg of messages) {
+      const text =
+        typeof msg.content === "string"
+          ? msg.content
+          : JSON.stringify(msg.content);
+
+      // Gemini takes the system prompt via `systemInstruction`, not as a
+      // content turn. Mapping it to a "model" turn (as this used to) mislabels
+      // the extraction instructions as something the model already said, and
+      // leaves the conversation without a leading user turn. Mirrors the Python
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-oss gemini fix with tests. Follows the same Python-parity line as #6468: the system prompt was being sent as a model content turn instead of via systemInstruction, so the extraction instructions were mislabeled and the conversation had no leading user turn. Now mirrors _reformat_messages in mem0/llms/gemini.py. Quick look when you get a chance, thanks.

PR #6475, opened 2026-07-21
fix(dashboard): clear the LLM API key on provider change

rodboev:
## Linked Issue

Refs #4984

## Description

Changing the LLM provider in the self-hosted Configuration page or Setup wizard could leave the previous provider's key in the LLM API Key field. Saving then paired the new provider with stale credential text.

This change clears the page-local LLM API Key field the moment the LLM provider selection changes on both dashboard surfaces, so a stale cross-provider key can no longer be saved.

## Root cause

Both LLM provider selects updated only `llmProvider`, while both save paths still read the unchanged `llmApiKey` state and pass it directly into `buildProviderConfig(...)`.

## Scope

This PR changes only the two dashboard page components. It leaves server-side `.env` and database precedence, persisted overrides, embedder behavior, payload shaping, docs, and provider implementations unchanged. The separate server-state half of `#4984` remains open.

## Attribution

This slice follows maintainer confirmation in https://github.com/mem0ai/mem0/issues/4984#issuecomment-4716787944 and the original dashboard-file work in stale PR https://github.com/mem0ai/mem0/pull/4996 .

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Test Coverage

- [ ] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (describe below)
- [ ] No tests needed

Verification:
- [x] `pnpm --dir server/dashboard run typecheck`
- [ ] `pnpm --dir server/dashboard run lint`
- [x] `pnpm --dir server/dashboard run build`
- [x] `pnpm --dir server/dashboard run dev`, then verify the Configuration and Setup provider-change flows locally against a stubbed local API on base and head

Manual verification:
- Configuration on base kept `sk-old-provider` after `openai` to `gemini`, and the patched branch cleared it immediately.
- Setup on base kept `anthropic-old` after `anthropic` to `gemini`, and the patched branch cleared it immediately.
- Empty-key transitions and adjacent non-provider edits kept the field empty or preserved `preserve-me` as expected on the patched branch.
- `pnpm --dir server/dashboard run lint` still reports existing package-wide Prettier drift: `Code style issues found in 111 files. Run Prettier with --write to fix.`

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
[...]

PR #6574, opened 2026-07-24
fix(server): restore env precedence over persisted config overrides

### LINKED ISSUE #4984: bug(server): dashboard saves wrong API key on provider change; .env cannot override database config

rodboev:
## Linked Issue

Closes #4984

## Description

The dashboard half of issue `#4984` merged in PR `#6475`, but the remaining server-state half is still live on `main`: an explicit `.env` value cannot override a conflicting persisted server override on restart. This change restores env precedence for fields that are explicitly present in the environment, while keeping persisted overrides effective for fields whose environment variable is absent.

## Root cause

The server currently collapses built-in defaults and explicit environment values into one `DEFAULT_CONFIG`, then merges persisted overrides on top of that single blob during startup and later mutates the already-composed runtime config during `POST /configure`. Once those ownership layers are flattened together, the runtime can no longer tell which fields were explicitly set by environment and must remain authoritative.

## Scope

This PR touches only `server/main.py`, `server/server_state.py`, and a new focused regression file `tests/test_server_config.py`. It leaves the merged dashboard fix from `#6475`, the provider-switch overwrite work from `#5523`, auth middleware, endpoint shapes, docs, and `VECTOR_STORE_CONFIG` unchanged.

## Attribution

The remaining-slice direction comes from the maintainer follow-up in https://github.com/mem0ai/mem0/issues/4984#issuecomment-5055415449 after the dashboard half merged in https://github.com/mem0ai/mem0/pull/6475 .

## Type of Change

- [ ] Documentation update

## Test Coverage

- [ ] No tests needed

Verification:
- [x] `hatch run pytest tests/test_server_config.py -q`
- [x] `hatch run ruff check server/main.py server/server_state.py tests/test_server_config.py`
- [x] `hatch run ruff format --check server/main.py server/server_state.py tests/test_server_config.py`
- [ ] `make lint`
- [ ] `make test`

## Checklist

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -60,6 +60,8 @@ SKIPPED_REQUEST_LOG_PREFIXES = ("/requests",)
 
 BUNDLED_LLM_PROVIDERS = ("openai", "anthropic", "gemini")
 BUNDLED_EMBEDDER_PROVIDERS = ("openai", "gemini")
+DEFAULT_LLM_MODEL = "gpt-4.1-nano-2025-04-14"
+DEFAULT_EMBEDDER_MODEL = "text-embedding-3-small"
 
 
 def _warn_if_unconfigured() -> None:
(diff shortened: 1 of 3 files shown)


rodboev:
Looks like the red Python SDK CI here is part of a broader July 24 break, not a branch-specific regression in this slice.

For this PR, the failing leg is `Python SDK / build_mem0`, where the workflow runs repo-wide `make lint` and surfaces unrelated Ruff debt outside the touched files, ending with `Found 2598 errors`.

This PR merged against `main` commit `d653b63fac6c8ad0ad84aead0912b366e705d269`, and that same `main` push is red in run `30095025105`. Several same-day open mem0 PRs that also trigger `build_mem0` are red in the same job family as well.

The touched files in this PR were checked locally and passed:
- `hatch run pytest tests/test_server_config.py -q`
- `hatch run ruff check server/main.py server/server_state.py tests/test_server_config.py`
- `hatch run ruff format --check server/main.py server/server_state.py tests/test_server_config.py`

If anyone spots a branch-specific failure signal beyond the shared Python SDK lint break, I’m happy to dig into it.

PR #6537, opened 2026-07-23
fix(ts-oss): give the reranker the full candidate pool, not the truncated topK

OfficialAbhinavSingh:
## Linked Issue

Closes #6536

## Description

TypeScript twin of #6449 (Python). `Memory.search()` (`mem0-ts/src/oss/src/memory/index.ts`) over-fetches a candidate pool (`internalLimit = Math.max(topK * 4, 60)`), but `scoreAndRank(..., topK, ...)` truncates the ranked output to `topK` **before** the reranker runs (Step 10). So the reranker only ever receives the already-truncated top-`topK` set — it can reorder those, but can never surface a relevant memory that the first-stage scorer ranked at `topK+1..N`, defeating the two-stage retrieve-then-rerank design. With reranking on, recall ends up identical to reranking off.

This PR: when a reranker will run, `scoreAndRank` returns the full over-fetched pool (`internalLimit`) instead of truncating to `topK`; the reranker then narrows it back to `topK`. If the rerank call fails, the fallback trims to `topK` so a failure doesn't leak the whole pool. The DB fetch is **not** re-widened — `internalLimit` is still computed once from `topK` (no double-widening; the single-method TS `search()` doesn't have the caller/callee split that #6449 had to untangle on the Python side).

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Breaking Changes

N/A — no public API change. Non-rerank path is unchanged (`returnK === topK`).

## Test Coverage

- [x] I added/updated unit tests

Two new cases in `mem0-ts/src/oss/tests/memory.rerank.test.ts`:
- `hands the reranker the full candidate pool, not the truncated topK` — primes an 8-item pool with `topK: 3`; asserts the reranker receives the full 8 (> topK), is asked to narrow to 3, and the final result is 3.
- `trims the over-fetched pool back to topK when the reranker throws` — same pool; a failing reranker must yield exactly `topK`, not the whole pool.

**Proof — red on `upstream/main`, green with the fix** (`index.ts` reverted to `upstream/main`, then restored):

```
# BEFORE (raw upstream/main — scoreAndRank truncates to topK before rerank):
$ npx jest --config jest.config.js src/oss/tests/memory.rerank.test.ts
  ✕ hands the reranker the full candidate pool, not the truncated topK
  (6 others pass)
  Tests: 1 failed, 6 passed

# AFTER (this branch):
$ npx jest --config jest.config.js src/oss/tests/memory.rerank.test.ts
  Tests: 7 passed

[...]

PR #6576, opened 2026-07-24
fix(ts-oss): feed the reranker the full candidate pool, not the truncated topK

### LINKED ISSUE #6575: TS OSS search(rerank:true) reranks only the truncated topK, so reranking can't improve recall
### Component

TypeScript SDK / Memory

### Where

`mem0-ts/src/oss/src/memory/index.ts`, `Memory.search()`.

### What happens

When `search(query, { rerank: true })` is used, the reranker only ever receives the already-truncated top-`topK` results, not the larger candidate pool that was fetched for scoring. As a result reranking can only **reorder** the final results — it can never surface a relevant memory that the first-stage scorer ranked just below `topK`, which is the main reason to run a reranker at all.

Step 3 over-fetches a candidate pool (`internalLimit = max(topK * 4, 60)`) for hybrid scoring, but step 8 truncates it to exactly `topK`:

```ts
const scoredResults = scoreAndRank(
  candidates, bm25Scores, entityBoosts, threshold ?? 0.1, topK, explain,
);
// ...
const ranked = await this.reranker!.rerank(query, results.map(r => r.memory), topK);
```

So `results` handed to the reranker is already the final top-`topK` set. The over-fetched pool that exists specifically to enable better ranking is discarded one stage too early.

### Why it is wrong

Reranking is a two-stage retrieve-then-rerank design: stage one (bi-encoder / BM25 hybrid) fetches a broad candidate pool cheaply; stage two (a cross-encoder / LLM reranker) rescores that pool and picks the true top-`topK`. Cross-encoders exist precisely because the first-stage scorer misranks — so the correct memory is often just below `topK` in stage one. Truncating to `topK` before reranking removes exactly the candidates the reranker is meant to rescue.

### Expected

When a reranker will run, keep the larger candidate pool through scoring and hand it to the reranker, which then narrows back to `topK`. Non-rerank searches are unchanged.

### Note

This is the TS counterpart of #6448 (the Python SDK has the identical bug in `Memory.search()` / `AsyncMemory.search()`).

abhay-codes07:
### Description

When `search(query, { rerank: true })` is used, the reranker only ever received the already-truncated top-`topK` results, not the larger candidate pool fetched for scoring. So reranking could only **reorder** the final results — it could never surface a relevant memory the first-stage scorer ranked just below `topK`, which is the main reason to run a reranker at all.

Step 3 over-fetches a pool (`internalLimit = max(topK * 4, 60)`), but step 8 truncated it to `topK` via `scoreAndRank(..., topK, ...)` before the reranker ran. The fix: when a reranker will run, score and keep the larger pool and hand it to the reranker, which narrows back to `topK`. A rerank failure trims the pool to `topK` so it never leaks. Non-rerank searches are unchanged (they still truncate to `topK` exactly as before).

This is the TS counterpart of #6448 (the Python SDK has the identical bug, being fixed the same way in #6455 / #6449).

Closes #6575

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/memory.rerank.test.ts` (run from the `mem0-ts` root), 7 passed.

- Added `feeds the reranker the full candidate pool, not just topK, so it can surface a memory ranked below topK`: with 5 first-stage candidates and `topK: 2`, the reranker now receives all 5 documents and a memory the first stage ranked 5th is surfaced to the top; the final result is still trimmed to `topK`. Reverting the source change fails this — the reranker receives only 2 documents.
- Added `trims the over-fetched pool back to topK when the reranker fails`: a rerank failure returns exactly `topK`, never the whole pool.
- The existing reranking tests (reorder, no-op, graceful failure, config wiring) still pass unchanged.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -1531,13 +1531,20 @@ export class Memory {
         payload: mem.payload || {},
       }));
 
-    // Step 8: Score and rank
+    // Step 8: Score and rank. When a reranker will run, keep a larger candidate
+    // pool so it can surface memories the first-stage scorer ranked below topK;
+    // the reranker then narrows it back to topK. Without this, scoreAndRank
+    // truncates to topK before the reranker runs, so reranking could only
+    // reorder the final results and never improve recall — which is the main
+    // reason to run a reranker at all.
+    const rerankActive = Boolean(config.rerank && this.reranker);
+    const rankLimit = rerankActive ? Math.max(topK * 4, 60) : topK;
     const scoredResults = scoreAndRank(
       candidates,
       bm25Scores,
       entityBoosts,
       threshold ?? 0.1,
-      topK,
+      rankLimit,
       explain,
     );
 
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 ts-oss rerank fix with tests. TS counterpart of #6448: search(rerank:true) was handing the reranker the already-truncated topK, so reranking could only reorder and never rescue a memory ranked just below topK. Now scores and keeps the larger pool for the reranker (same approach as the Python PRs #6455/#6449), trimming to topK after. Quick look when you get a chance, thanks.

abhay-codes07:
Closing this in favor of #6537, which was opened first for #6536 and takes the same approach (it reuses internalLimit for the pool, which is a touch cleaner than recomputing it). I filed my issue/PR before spotting the existing #6536/#6537. Deferring to that one. Thanks!

PR #6578, opened 2026-07-24
fix(ts-oss/turbopuffer): apply all filter operators, not just gte/lte

### LINKED ISSUE #6577: TS OSS Turbopuffer store silently drops filter operators other than gte/lte
### Component

TypeScript SDK / Vector stores

### Where

`mem0-ts/src/oss/src/vector_stores/turbopuffer.ts`, `convertFilters()`.

### What happens

```ts
if (typeof value === "object" && value !== null && !Array.isArray(value)) {
  if ("gte" in value) conditions.push([key, "Gte", value.gte]);
  if ("lte" in value) conditions.push([key, "Lte", value.lte]);
  if ("gt" in value) conditions.push([key, "Gt", value.gt]);
  if ("lt" in value) conditions.push([key, "Lt", value.lt]);
} else {
  conditions.push([key, "Eq", value]);
}
```

Only `gte`/`lte`/`gt`/`lt` are read from an operator dict. Every other operator mem0 supports (`eq`, `ne`, `in`, `nin`) matches no branch, so a filter like `{ tier: { in: ["a","b"] } }` produces **no condition** and the query silently returns unfiltered results. Two more cases are mishandled:

- A bare array `{ tier: ["a","b"] }` falls into the `else` branch and becomes `[key, "Eq", ["a","b"]]` (equality against an array), matching nothing.
- A `"*"` wildcard becomes `[key, "Eq", "*"]`, matching nothing instead of "any value".

### Expected

Map the full operator set to Turbopuffer tokens (`eq`→Eq, `ne`→NotEq, `gt`→Gt, `gte`→Gte, `lt`→Lt, `lte`→Lte, `in`→In, `nin`→NotIn), treat a bare array as `in`, skip `"*"`, and throw on an unknown operator instead of dropping it.

### Note

This is the TS counterpart of #6562 (the Python Turbopuffer store has the identical bug, fixed in #6564). The operator tokens here match that PR.

abhay-codes07:
### Description

`convertFilters()` in the TS Turbopuffer store only read `gte`/`lte`/`gt`/`lt` from an operator dict, so `eq`/`ne`/`in`/`nin` matched no branch and were silently dropped — a filter like `{ tier: { in: ["a","b"] } }` produced no condition and the query returned unfiltered results. A bare array `{ tier: ["a","b"] }` became `[key, "Eq", array]` (matching nothing), and a `"*"` wildcard became a literal `Eq "*"`.

The fix maps the full operator set to Turbopuffer tokens (`eq`→Eq, `ne`→NotEq, `gt`→Gt, `gte`→Gte, `lt`→Lt, `lte`→Lte, `in`→In, `nin`→NotIn), treats a bare array as `in`, skips `"*"`, and throws on an unknown operator instead of dropping it.

This is the TS counterpart of #6562; the operator tokens match the Python fix #6564.

Closes #6577

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/turbopuffer.unit.test.ts` (run from the `mem0-ts` root), 6 passed. This adds the first Turbopuffer unit test, driving `convertFilters` through `search()` with a virtually-mocked peer and asserting the filter tuple handed to `ns.query()`.

- Covers every operator token, compound AND ranges, bare-array→`in`, scalar equality, `"*"` skip, and the unknown-operator throw.
- Reverting the source change fails the operator-mapping, array, wildcard, and throw tests (the dropped operators produce no filter).

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/turbopuffer.ts
+++ b/mem0-ts/src/oss/src/vector_stores/turbopuffer.ts
@@ -247,23 +247,54 @@ export class TurbopufferDB implements VectorStore {
     }
   }
 
+  // Maps mem0's universal filter operators to Turbopuffer's filter tokens.
+  private static readonly OPERATOR_MAP: Record<string, string> = {
+    eq: "Eq",
+    ne: "NotEq",
+    gt: "Gt",
+    gte: "Gte",
+    lt: "Lt",
+    lte: "Lte",
+    in: "In",
+    nin: "NotIn",
+  };
+
   private convertFilters(filters?: SearchFilters): any {
     if (!filters || Object.keys(filters).length === 0) return null;
 
     const conditions: any[] = [];
     for (const [key, value] of Object.entries(filters)) {
-      if (
-        typeof value === "object" &&
-        value !== null &&
-        !Array.isArray(value)
-      ) {
-        if ("gte" in value) conditions.push([key, "Gte", value.gte]);
-        if ("lte" in value) conditions.push([key, "Lte", value.lte]);
-        if ("gt" in value) conditions.push([key, "Gt", value.gt]);
-        if ("lt" in value) conditions.push([key, "Lt", value.lt]);
-      } else {
-        conditions.push([key, "Eq", value]);
+      // "*" is a match-any wildcard: it must not constrain the query. The old
+      // code turned it into `[key, "Eq", "*"]`, matching nothing.
+      if (value === "*") {
+        continue;
       }
+
+      // Array shorthand: { key: [a, b] } means "in".
+      if (Array.isArray(value)) {
+        conditions.push([key, "In", value]);
+        continue;
+      }
+
+      if (typeof value === "object" && value !== null) {
+        // Operator dict: every operator present must hold. Previously only
+        // `gte` and `lte` were read, so `gt`/`lt`/`ne`/`eq`/`in`/`nin` were
+        // silently dropped and the filter returned unfiltered results.
+        for (const [op, operand] of Object.entries(value)) {
+          const token = TurbopufferDB.OPERATOR_MAP[op];
+          if (!token) {
+            throw new Error(
+              `Unsupported Turbopuffer filter operator '${op}' for field '${key}'. ` +
+                `Supported operators: ${Object.keys(TurbopufferDB.OPERATOR_MAP).join(", ")}.`,
+            );
+          }
+          conditions.push([key, token, operand]);
+        }
+        continue;
+      }
+
+      // Scalar shorthand: equality.
+      conditions.push([key, "Eq", value]);
     }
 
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 ts-oss turbopuffer fix with tests. TS counterpart of #6562: convertFilters only read gte/lte/gt/lt, so eq/ne/in/nin were silently dropped and searches returned unfiltered results. Now maps the full operator set (tokens match the Python fix #6564), handles bare arrays as in, skips *, and throws on unknown operators. Quick look when you get a chance, thanks.

kartik-mem0 (commented):

Turbopuffer silently dropped eq/ne/in/nin filters and mishandled bare arrays/"*"; this maps every operator to a real token, fixing `search()` and `list()` at their shared call site.

**Before you merge:** see the inline note on `OPERATOR_MAP`, it's missing `contains`/`icontains` so those now hard-throw instead of silently dropping.
**Follow-up:** sibling Python PR #6564 only fixes the dict-operator branch, not the bare-array/`"*"` fallback, keep them in sync.
**Not verified:** live Turbopuffer backend (hosted-only, no sandbox lane) - verified via red-green tests and the installed SDK's real filter grammar instead.

Risk: low, self-contained to one provider's private method · Your time: ~5 min

- Problem is real: reproduces on `main` at `turbopuffer.ts:254-267`, eq/ne/in/nin dropped, matches #6577
- Tests: `npx jest src/oss/tests/turbopuffer.unit.test.ts` -> 6 passed
- Red-green: reverting `turbopuffer.ts` to pre-PR fails 4/6; restored, 6/6 pass
- CI: green on real jobs (build_ts_sdk, integration_ts_sdk, CI Gate); Vercel red is fork-deploy-auth, non-blocking
- Security: cleared, tuple-shaped values, no string-built queries; "*"-skip matches 13 sibling providers, never on user_id/agent_id/run_id
- Bloat: lean already, net -0 lines
- Linked issue: #6577

</details>

mem0-ts/src/oss/src/vector_stores/turbopuffer.ts
```diff
@@ -247,23 +247,54 @@ export class TurbopufferDB implements VectorStore {
     }
   }
 
+  // Maps mem0's universal filter operators to Turbopuffer's filter tokens.
+  private static readonly OPERATOR_MAP: Record<string, string> = {
```
kartik-mem0, line 251: OPERATOR_MAP has no entry for `contains`/`icontains`, which mem0's filter DSL forwards here (memory/index.ts's `_processMetadataFilters`) and Turbopuffer's SDK supports (`Contains`/`NotContains`). A caller using either now gets a hard throw instead of the old silent drop.

Map them to `Contains`/`NotContains`, or skip with a warning like `pinecone.ts` does.

PR #6580, opened 2026-07-24
fix(ts-oss/turbopuffer): bound euclidean_squared distance to a similarity score

### LINKED ISSUE #6579: TS OSS Turbopuffer search scores are wrong for euclidean_squared (1 - dist goes negative)
### Component

TypeScript SDK / Vector stores

### Where

`mem0-ts/src/oss/src/vector_stores/turbopuffer.ts`, `parseRows()`.

### What happens

```ts
const score = $dist != null ? 1 - $dist : undefined;
```

The store supports both `cosine_distance` and `euclidean_squared` (see `distanceMetric`), but the score is always `1 - $dist`. That is only correct for cosine, where distance is in `[0, 2]`. `euclidean_squared` distance is unbounded, so for any `$dist > 1` the score goes negative (e.g. `$dist = 4.0` -> `-3.0`), which the memory layer then sorts/thresholds as if it were a similarity — inverting and breaking ranking.

### Expected

Convert per metric, mirroring the milvus/baidu stores:
- `cosine_distance`: `1 - $dist` (unchanged)
- `euclidean_squared`: `1 / (1 + $dist)` (bounded, higher-is-better)
- missing `$dist`: `undefined`

### Note

This is the TS counterpart of #6557 (the Python Turbopuffer store has the identical bug, fixed in #6559).

abhay-codes07:
### Description

`parseRows()` in the TS Turbopuffer store always converted distance to score as `1 - $dist`. That is only correct for `cosine_distance` (distance in `[0, 2]`); `euclidean_squared` distance is unbounded, so any `$dist > 1` produced a negative score (e.g. `4.0` -> `-3.0`) that the memory layer then sorts and thresholds as if it were a similarity, inverting and breaking ranking.

The fix converts per metric, mirroring the milvus/baidu stores: `cosine_distance` keeps `1 - $dist`, `euclidean_squared` uses `1 / (1 + $dist)` (bounded, higher-is-better), and a missing `$dist` stays `undefined`.

This is the TS counterpart of #6557 (the Python store has the identical bug, fixed in #6559).

Closes #6579

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/turbopuffer.score.test.ts` (run from the `mem0-ts` root), 4 passed. Drives `parseRows` through `search()` with a virtually-mocked peer.

- `keeps cosine distance as 1 - dist (default metric)` — cosine path unchanged.
- `bounds an unbounded euclidean_squared distance to a 0..1 similarity` — `$dist = 4.0` yields `0.2`, not `-3.0`. Reverting the source change fails this (score comes back `-3`).
- `ranks a nearer euclidean_squared hit above a farther one` and `preserves an undefined score when the row has no distance`.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/vector_stores/turbopuffer.ts
+++ b/mem0-ts/src/oss/src/vector_stores/turbopuffer.ts
@@ -274,7 +274,18 @@ export class TurbopufferDB implements VectorStore {
   private parseRows(rows: any[]): VectorStoreResult[] {
     return rows.map((row) => {
       const { id, $dist, vector, ...rest } = row;
-      const score = $dist != null ? 1 - $dist : undefined;
+      let score: number | undefined;
+      if ($dist == null) {
+        score = undefined;
+      } else if (this.distanceMetric === "euclidean_squared") {
+        // euclidean_squared $dist is an unbounded squared distance, so 1 - $dist
+        // goes negative for any $dist > 1 and inverts ranking. Convert it to a
+        // bounded higher-is-better score, mirroring the milvus/baidu stores.
+        score = 1 / (1 + $dist);
+      } else {
+        // Cosine distance is in [0, 2]; 1 - $dist stays a meaningful similarity.
+        score = 1 - $dist;
+      }
       return { id: String(id), payload: rest, score };
     });
   }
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 ts-oss turbopuffer fix with tests. TS counterpart of #6557: parseRows always did 1 - dist, so euclidean_squared (unbounded) distances gave negative scores and inverted ranking. Now converts per metric like the milvus/baidu stores (matches the Python fix #6559). Separate from my filter PR #6578; different function, no overlap. Quick look when you get a chance, thanks.

PR #6588, opened 2026-07-25
fix(together): honor TOGETHER_API_BASE for LLM and embedder clients

### LINKED ISSUE #6587: bug(llms+embeddings/together): TOGETHER_API_BASE env var is ignored (TS LLM honors it)
### Component

Python SDK / Together LLM + embedder

### What happens

`TogetherLLM` and `TogetherEmbedding` construct `Together(api_key=...)` with no `base_url`. Proxies and custom Together-compatible endpoints set via `TOGETHER_API_BASE` / `TOGETHER_BASE_URL` (or config `openai_base_url`) are ignored.

The TypeScript Together LLM already honors `TOGETHER_API_BASE` (`mem0-ts/src/oss/src/llms/together.ts`). Python should match so self-hosted / gateway deployments work for both chat and embeddings.

### Expected

- `TOGETHER_API_BASE` or `TOGETHER_BASE_URL` → passed as `base_url` to the Together client
- Optional config `openai_base_url` / `together_base_url` → same
- Unset → SDK default endpoint (no base_url kwarg)

### Files

- `mem0/llms/together.py`
- `mem0/embeddings/together.py`

Bartok9:
Closes #6587

Together Python LLM/embedder ignored TOGETHER_API_BASE (and config base URL); TS LLM already forwards it.

Pass base_url from config/env when set so proxies and gateways work.

Verification: pytest tests/llms/test_together.py tests/embeddings/test_together_embeddings.py — 13 passed.

### DIFF
--- a/mem0/embeddings/together.py
+++ b/mem0/embeddings/together.py
@@ -18,7 +18,6 @@ class TogetherEmbedding(EmbeddingBase):
             getattr(self.config, "together_base_url", None)
             or getattr(self.config, "openai_base_url", None)
             or os.getenv("TOGETHER_API_BASE")
-            or os.getenv("TOGETHER_BASE_URL")
         )
         client_kwargs = {"api_key": api_key}
         if base_url:
--- a/mem0/embeddings/together.py
+++ b/mem0/embeddings/together.py
@@ -14,7 +14,16 @@ class TogetherEmbedding(EmbeddingBase):
         self.config.model = self.config.model or "intfloat/multilingual-e5-large-instruct"
         api_key = self.config.api_key or os.getenv("TOGETHER_API_KEY")
         self.config.embedding_dims = self.config.embedding_dims or 1024
-        self.client = Together(api_key=api_key)
+        base_url = (
+            getattr(self.config, "together_base_url", None)
+            or getattr(self.config, "openai_base_url", None)
+            or os.getenv("TOGETHER_API_BASE")
+            or os.getenv("TOGETHER_BASE_URL")
+        )
+        client_kwargs = {"api_key": api_key}
+        if base_url:
+            client_kwargs["base_url"] = base_url
+        self.client = Together(**client_kwargs)
 
     def embed(self, text, memory_action: Optional[Literal["add", "search", "update"]] = None):
         """
--- a/mem0/llms/together.py
+++ b/mem0/llms/together.py
@@ -26,7 +26,6 @@ class TogetherLLM(LLMBase):
             getattr(self.config, "together_base_url", None)
             or getattr(self.config, "openai_base_url", None)
             or os.getenv("TOGETHER_API_BASE")
-            or os.getenv("TOGETHER_BASE_URL")
         )
         client_kwargs = {"api_key": api_key}
         if base_url:
--- a/mem0/llms/together.py
+++ b/mem0/llms/together.py
@@ -20,7 +20,18 @@ class TogetherLLM(LLMBase):
             self.config.model = "MiniMaxAI/MiniMax-M3"
 
         api_key = self.config.api_key or os.getenv("TOGETHER_API_KEY")
-        self.client = Together(api_key=api_key)
+        # Prefer explicit config (openai_base_url common for OpenAI-compatible),
+        # then TOGETHER_API_BASE — parity with TS TogetherLLM and other providers.
+        base_url = (
+            getattr(self.config, "together_base_url", None)
+            or getattr(self.config, "openai_base_url", None)
+            or os.getenv("TOGETHER_API_BASE")
+            or os.getenv("TOGETHER_BASE_URL")
+        )
+        client_kwargs = {"api_key": api_key}
+        if base_url:
+            client_kwargs["base_url"] = base_url
+        self.client = Together(**client_kwargs)
 
     def _parse_response(self, response, tools):
         """
--- a/tests/embeddings/test_together_embeddings.py
+++ b/tests/embeddings/test_together_embeddings.py
@@ -85,3 +85,23 @@ def test_explicit_config_overrides_defaults(mock_together_client):
     mock_together_client.embeddings.create.return_value = Mock(data=[Mock(embedding=[0.0] * 768)])
     embedder.embed("hello")
     mock_together_client.embeddings.create.assert_called_once_with(model="BAAI/bge-base-en-v1.5", input="hello")
+
+
+def test_together_embedder_base_url_from_env(mock_together_client, monkeypatch):
+    monkeypatch.setenv("TOGETHER_API_BASE", "https://proxy.example/v1")
+    monkeypatch.delenv("TOGETHER_BASE_URL", raising=False)
+    config = BaseEmbedderConfig(model=DEFAULT_MODEL, api_key="test-key")
+    with patch("mem0.embeddings.together.Together") as mock_together:
+        mock_together.return_value = mock_together_client
+        TogetherEmbedding(config)
+        mock_together.assert_called_once_with(api_key="test-key", base_url="https://proxy.example/v1")
+
+
+def test_together_embedder_base_url_from_openai_base_url(mock_together_client, monkeypatch):
+    monkeypatch.delenv("TOGETHER_API_BASE", raising=False)
+    monkeypatch.delenv("TOGETHER_BASE_URL", raising=False)
+    config = BaseEmbedderConfig(model=DEFAULT_MODEL, api_key="test-key", openai_base_url="https://cfg.example/v1")
+    with patch("mem0.embeddings.together.Together") as mock_together:
+        mock_together.return_value = mock_together_client
+        TogetherEmbedding(config)
+        mock_together.assert_called_once_with(api_key="test-key", base_url="https://cfg.example/v1")
--- a/tests/embeddings/test_together_embeddings.py
+++ b/tests/embeddings/test_together_embeddings.py
@@ -105,3 +105,20 @@ def test_together_embedder_base_url_from_openai_base_url(mock_together_client, m
         mock_together.return_value = mock_together_client
         TogetherEmbedding(config)
         mock_together.assert_called_once_with(api_key="test-key", base_url="https://cfg.example/v1")
+
+
+def test_together_embedder_base_url_from_together_base_url(mock_together_client, monkeypatch):
+    """together_base_url on BaseEmbedderConfig wins over openai_base_url and env."""
+    monkeypatch.delenv("TOGETHER_API_BASE", raising=False)
+    monkeypatch.delenv("TOGETHER_BASE_URL", raising=False)
+    monkeypatch.setenv("TOGETHER_API_BASE", "https://env-should-lose.example/v1")
+    config = BaseEmbedderConfig(
+        model=DEFAULT_MODEL,
+        api_key="test-key",
+        openai_base_url="https://openai-cfg-should-lose.example/v1",
+        together_base_url="https://together-cfg.example/v1",
+    )
+    with patch("mem0.embeddings.together.Together") as mock_together:
+        mock_together.return_value = mock_together_client
+        TogetherEmbedding(config)
+        mock_together.assert_called_once_with(api_key="test-key", base_url="https://together-cfg.example/v1")
--- a/tests/llms/test_together.py
+++ b/tests/llms/test_together.py
@@ -102,3 +102,25 @@ def test_generate_response_forwards_extra_kwargs(mock_together_client):
     assert response == "Hi"
     call_kwargs = mock_together_client.chat.completions.create.call_args.kwargs
     assert call_kwargs["frequency_penalty"] == 0.5
+
+
+def test_together_base_url_from_env(mock_together_client, monkeypatch):
+    """TOGETHER_API_BASE is forwarded to the Together client constructor."""
+    monkeypatch.setenv("TOGETHER_API_BASE", "https://proxy.example/v1")
+    monkeypatch.delenv("TOGETHER_BASE_URL", raising=False)
+    config = BaseLlmConfig(model="test-model", api_key="test-key", temperature=0.7, max_tokens=100, top_p=1.0)
+    with patch("mem0.llms.together.Together") as mock_together:
+        mock_together.return_value = mock_together_client
+        TogetherLLM(config)
+        mock_together.assert_called_once_with(api_key="test-key", base_url="https://proxy.example/v1")
+
+
+def test_together_base_url_from_openai_base_url_config(mock_together_client, monkeypatch):
+    monkeypatch.delenv("TOGETHER_API_BASE", raising=False)
+    monkeypatch.delenv("TOGETHER_BASE_URL", raising=False)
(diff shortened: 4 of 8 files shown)


mem0/llms/together.py
```diff
@@ -20,7 +20,18 @@ def __init__(self, config: Optional[BaseLlmConfig] = None):
             self.config.model = "MiniMaxAI/MiniMax-M3"
 
         api_key = self.config.api_key or os.getenv("TOGETHER_API_KEY")
-        self.client = Together(api_key=api_key)
+        # Prefer explicit config (openai_base_url common for OpenAI-compatible),
+        # then TOGETHER_API_BASE — parity with TS TogetherLLM and other providers.
+        base_url = (
+            getattr(self.config, "together_base_url", None)
```
kartik-mem0, line 26: This branch can never be true: `together_base_url` is defined in no config class, and `LlmFactory` maps `together` to `BaseLlmConfig`, which raises `TypeError: BaseLlmConfig.__init__() got an unexpected keyword argument 'together_base_url'` if anyone passes it.

Add the field to `mem0/configs/llms/base.py` alongside the existing `*_base_url` fields, or drop the branch.

mem0/embeddings/together.py
```diff
@@ -14,7 +14,16 @@ def __init__(self, config: Optional[BaseEmbedderConfig] = None):
         self.config.model = self.config.model or "intfloat/multilingual-e5-large-instruct"
         api_key = self.config.api_key or os.getenv("TOGETHER_API_KEY")
         self.config.embedding_dims = self.config.embedding_dims or 1024
-        self.client = Together(api_key=api_key)
+        base_url = (
+            getattr(self.config, "together_base_url", None)
```
kartik-mem0, line 18: Same dead branch here: `BaseEmbedderConfig` has no `together_base_url` either, so passing it raises `TypeError`.

Adding it next to the existing `openai_base_url` / `ollama_base_url` / `lmstudio_base_url` fields would make it the natural name for this knob, and would let the LLM side read the same field.

tests/llms/test_together.py
```diff
@@ -102,3 +102,25 @@ def test_generate_response_forwards_extra_kwargs(mock_together_client):
...
+        mock_together.return_value = mock_together_client
+        TogetherLLM(config)
+        mock_together.assert_called_once_with(api_key="test-key", base_url="https://proxy.example/v1")
+
+
+def test_together_base_url_from_openai_base_url_config(mock_together_client, monkeypatch):
+    monkeypatch.delenv("TOGETHER_API_BASE", raising=False)
+    monkeypatch.delenv("TOGETHER_BASE_URL", raising=False)
+    config = BaseLlmConfig(model="test-model", api_key="test-key", temperature=0.7, max_tokens=100, top_p=1.0)
+    config.openai_base_url = "https://cfg.example/v1"
```
kartik-mem0: This assignment is the tell: `BaseLlmConfig` has no `openai_base_url`, so a real user config raises `TypeError` and only a post-construction attribute set can reach line 27. The test proves the mechanism, not a path anyone can configure.

It does work for the embedder, where the field genuinely exists.

mem0/llms/together.py
```diff
@@ -20,7 +20,18 @@ def __init__(self, config: Optional[BaseLlmConfig] = None):
...
+            or getattr(self.config, "openai_base_url", None)
+            or os.getenv("TOGETHER_API_BASE")
+            or os.getenv("TOGETHER_BASE_URL")
```
kartik-mem0: Minor, not blocking: `Together.__init__` already does `if base_url is None: base_url = os.environ.get("TOGETHER_BASE_URL")`, so this variable worked before this PR (verified on together 2.28.0). Dropping it here and in the embedder removes 2 lines and one untested branch.

kartik-mem0 (commented):
## Changes requested

Real bug, and the env-var half of the fix is correct: `TOGETHER_API_BASE` is ignored by the Python Together LLM and embedder, so an operator's gateway gets bypassed. What blocks this is the config half, which reads two fields that do not exist.

**To unblock:**
1. `mem0/llms/together.py:26` and `mem0/embeddings/together.py:18`: `together_base_url` is defined in no config class, so both branches are dead (detail inline)
2. `mem0/llms/together.py:27`: `openai_base_url` cannot reach `BaseLlmConfig`, so the LLM config path raises `TypeError`, and the new test only passes by setting the attribute after construction
3. `docs/components/llms/models/together.mdx:7`: still tells Python users the base URL is TypeScript-only

Optional: `TOGETHER_BASE_URL` already worked before this PR, so those two lines can go (net -6 lines).

- **Problem is real:** reproduces on `main` at `mem0/llms/together.py:23` and `mem0/embeddings/together.py:17`. With `TOGETHER_API_BASE` set and no kwarg, `client.base_url` stays `https://api.together.ai/v1/` (together 2.28.0).
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/llms/test_together.py tests/embeddings/test_together_embeddings.py -v` -> 13 passed
- **Red-green:** reverting both impl files to `main` fails all 4 new tests (`Actual: Together(api_key='test-key')`), restoring passes 13 again. The 9 pre-existing tests pass in every state, so no regression. ✓
- **Config reachability:** `LlmFactory.create("together", {..., "openai_base_url": ...})` -> `TypeError: BaseLlmConfig.__init__() got an unexpected keyword argument 'openai_base_url'`. Same for `together_base_url` on both factories. `openai_base_url` does work for the embedder, where the field exists.
- **CI:** real checks green (CI Gate, `build_mem0` on 3.10 / 3.11 / 3.12). The Vercel preview is red, which happens on nearly every fork PR and does not block.
- **Security:** cleared. `base_url` is operator-supplied, same trust level as `api_key` on the line above, passed straight to the vendor SDK. Nothing logged, empty string correctly falsy.
- **Bloat (ponytail):** net -6 lines possible. Top cut is the `TOGETHER_BASE_URL` getenv: `Together.__init__` already reads that variable itself.
- **Parity:** `mem0-ts/src/oss/src/llms/together.ts:14-19` accepts `config.baseURL`. Python still cannot after this PR.
- **Not verified:** no live Together call, constructor is mocked. Docs not rendered.
- **Linked issue:** #6587. Ruff clean on all 4 files.

</details>

This is the cleanest of the open PRs on #6587: both `Together(` construction sites covered, exact `assert_called_once_with`, proper env isolation, no drive-by changes. Adding `together_base_url` to `BaseLlmConfig` and `BaseEmbedderConfig` is a small change that makes the whole PR land. Push when ready and I will re-review.

Bartok9:
Thanks for the thorough review — all three blockers fixed in a2d56b11:

1. Added `openai_base_url` and `together_base_url` to `BaseLlmConfig`, and `together_base_url` to `BaseEmbedderConfig` (embedder already had `openai_base_url`). Both `getattr` branches now reach real fields.
2. `openai_base_url` now passes through the LLM constructor, so the config path no longer raises `TypeError`; updated the LLM test to construct with `openai_base_url=...` instead of setting the attribute post-hoc.
3. Docs now state the Python SDK supports `TOGETHER_API_BASE` / `together_base_url`, not TS-only.

Kept `TOGETHER_BASE_URL` for backward compat. Verified the config fields are reachable and both the env and config paths forward `base_url` correctly. Ready for re-review.

       1
kartik-mem0, line 1: please remove this file

Bartok9:
Thanks @kartik-mem0 for cleaning up that stray artifact file — good catch. Diff's clean now and the base_url pass-through is verified on both the config and env paths. Happy to make any further tweaks for re-review.

Bartok9:
@kartik-mem0 thanks for the careful review — addressed:

1. **Config fields** (prior push `a2d56b1`): added `together_base_url` + `openai_base_url` to `BaseLlmConfig`, and `together_base_url` to `BaseEmbedderConfig`, so factory/`**kwargs` paths are real.
2. **LLM test**: constructs with `openai_base_url=` (no post-hoc attribute set).
3. **Docs**: Python no longer marked base URL as TS-only.
4. **Junk file**: already removed on your side (`Delete …1`).
5. **Optional ponytail** (this push): dropped redundant `TOGETHER_BASE_URL` getenv in LLM + embedder; leave that to the Together SDK’s own constructor fallback. Kept `TOGETHER_API_BASE` as the env knob we actually needed for #6587.

Ready for re-review when you have a cycle.

tests/llms/test_together.py
```diff
@@ -102,3 +102,31 @@ def test_generate_response_forwards_extra_kwargs(mock_together_client):
...
+        api_key="test-key",
+        temperature=0.7,
+        max_tokens=100,
+        top_p=1.0,
+        openai_base_url="https://cfg.example/v1",
+    )
+    with patch("mem0.llms.together.Together") as mock_together:
+        mock_together.return_value = mock_together_client
+        TogetherLLM(config)
+        mock_together.assert_called_once_with(api_key="test-key", base_url="https://cfg.example/v1")
```
kartik-mem0, line 132: No test sets `together_base_url` on `BaseLlmConfig` directly, only `openai_base_url`. It is a real, reachable field now (verified it wins precedence over `openai_base_url` and env), so give it its own test.

tests/embeddings/test_together_embeddings.py
```diff
@@ -85,3 +85,23 @@ def test_explicit_config_overrides_defaults(mock_together_client):
...
+
+
+def test_together_embedder_base_url_from_openai_base_url(mock_together_client, monkeypatch):
+    monkeypatch.delenv("TOGETHER_API_BASE", raising=False)
+    monkeypatch.delenv("TOGETHER_BASE_URL", raising=False)
+    config = BaseEmbedderConfig(model=DEFAULT_MODEL, api_key="test-key", openai_base_url="https://cfg.example/v1")
+    with patch("mem0.embeddings.together.Together") as mock_together:
+        mock_together.return_value = mock_together_client
+        TogetherEmbedding(config)
+        mock_together.assert_called_once_with(api_key="test-key", base_url="https://cfg.example/v1")
```
kartik-mem0, line 107: Same gap here: `together_base_url` on `BaseEmbedderConfig` is untested, only `openai_base_url` is. Add a test asserting it wins the precedence.

kartik-mem0 (commented):
## Changes requested (re-review)

Verified against `440604d7` (new head): all 3 prior Important findings are fixed in code and tests pass (13 passed). Two smaller gaps found this round.

**To unblock:**
1. `tests/llms/test_together.py:132` and `tests/embeddings/test_together_embeddings.py:107`: no test sets `together_base_url` directly, only `openai_base_url` (detail inline).
2. `docs/components/embedders/models/together.mdx:73`: Python param table still has no `together_base_url` row. The LLM doc got the fix, the embedder doc did not.

- **Prior findings (replied in-thread):** the two dead `together_base_url` branches, the mechanism-only LLM test, and the redundant `TOGETHER_BASE_URL` getenv are all confirmed fixed at this head.
- **Tests:** `hatch -e dev_py_3_12 run pytest tests/llms/test_together.py tests/embeddings/test_together_embeddings.py -v` -> 13 passed at `440604d7`.
- **Manual check:** `together_base_url` does take precedence correctly over `openai_base_url` and env (verified live against both config classes), it is just untested.
- **CI:** real checks green (`build_mem0` 3.10/3.11/3.12, CI Gate). Vercel preview is red, ignored (fork perms, not merge-blocking).
- **Security:** cleared, unchanged from the prior review.
- **Stray file:** confirmed removed from the tree.
- **Not verified:** nothing new beyond the prior review's limits.

</details>

Close to done, appreciate the quick turnaround on the config fields. Add the two tests and the doc row and this should be ready.

mem0/llms/together.py
kartik-mem0: Verified against 440604d7: fixed. `together_base_url` is now a real field on `BaseLlmConfig` (`mem0/configs/llms/base.py`), so this branch is reachable and I confirmed live it takes precedence correctly.

mem0/embeddings/together.py
kartik-mem0: Verified against 440604d7: fixed. `together_base_url` is now a real field on `BaseEmbedderConfig` too, same as the LLM side.

tests/llms/test_together.py
kartik-mem0: Verified against 440604d7: fixed. The test now builds `BaseLlmConfig(..., openai_base_url=...)` through the constructor since the field exists, no more post-hoc attribute set.

mem0/llms/together.py
kartik-mem0: Verified against 440604d7: fixed. The redundant `os.getenv("TOGETHER_BASE_URL")` fallback is dropped from both files in the latest push.

       1
kartik-mem0: Verified against 440604d7: the stray file is gone from the tree.

Bartok9:
@kartik-mem0 added direct `together_base_url` precedence tests on both LLM and embedder configs (wins over `openai_base_url` + `TOGETHER_API_BASE`). Ready for another look.

kartik-mem0 (commented):
## Changes requested
Together LLM/embedder base URL fix is functionally complete and now has direct test coverage for `together_base_url`; one doc gap remains before this is done.

**To unblock:**
1. `docs/components/embedders/models/together.mdx:~73` — add a `together_base_url` row to the Python param table (the LLM doc already has this)

- **Problem is real:** confirmed on `main` at Gate 0, 2026-07-30 (unchanged)
- **Tests / red-green:** `tests/llms/test_together.py` and `tests/embeddings/test_together_embeddings.py` now directly test `together_base_url` precedence over `openai_base_url` and `TOGETHER_API_BASE`; 15 passed at `dead3b578` (up from 13), lint clean
- **CI:** all real checks green (`CI Gate`, `build_mem0` 3.10/3.11/3.12, `changelog_check`, `check_changes`, `docs llms.txt`, `label`, `license/cla`); `Vercel` red on fork permissions only, ignored per policy
- **Security:** unchanged, cleared. `base_url` is operator-supplied, same trust level as `api_key`
- **Bloat (ponytail):** none introduced this round, test-only delta
- **Not verified:** no live Together API call (mocked constructor only); not Gate-B-live eligible, diff touches no memory/vector-store path

</details>

Good progress since the last round: the previously-untested `together_base_url` path now has real regression coverage. Push the one doc row when ready and I will re-review.

tests/llms/test_together.py
kartik-mem0: Verified against dead3b578: fixed. `tests/llms/test_together.py` now has `test_together_base_url_from_together_base_url_config`, which sets `together_base_url` directly alongside `openai_base_url` and `TOGETHER_API_BASE`, and asserts `together_base_url` wins. Matches the precedence order in `mem0/llms/together.py:24-28`. 15 passed.

tests/embeddings/test_together_embeddings.py
kartik-mem0: Verified against dead3b578: fixed. `tests/embeddings/test_together_embeddings.py` now has `test_together_embedder_base_url_from_together_base_url`, which sets `together_base_url` directly and asserts it wins over `openai_base_url` and `TOGETHER_API_BASE`. Matches the precedence order in `mem0/embeddings/together.py:17-21`. 15 passed.

PR #6590, opened 2026-07-25
fix(ci): pin ruff rule selection to restore green lint on ruff 0.16

kartik-mem0:
## Linked Issue

N/A. `main` is currently red on the Python SDK lint job for every PR.

## Description

`make lint` started failing on `main` on 2026-07-23 with errors in files nobody touched, for example `I001` on `examples/graph-db-demo/*.ipynb`.

### Root cause

The root `[tool.ruff]` config sets `line-length` and `exclude` but never declared a lint `select`, so the repo silently ran on **ruff's implicit default rule set** (`E4`, `E7`, `E9`, `F`).

**Ruff 0.16.0 (released 2026-07-23) widened that default.** CI installs ruff unpinned (`pip install ruff` in `ci.yml`), so the next run after that release picked it up and began enforcing rules the repo had never opted into.

The `I001` notebook errors in the failing log are just the first few printed. The real count:

```
$ ruff 0.16.0, current main config
Found 2588 errors.
```

Nothing in the source changed. Only the linter's idea of "default" did.

### Fix

Declare the previous default explicitly, so the active rule set is a property of this repo rather than of whichever ruff version CI happens to install:

```toml
[tool.ruff.lint]
select = ["E4", "E7", "E9", "F"]
```

This restores exactly the rules that were passing before 0.16.0. It is a no-op against any ruff older than 0.16.0.

### Verification

Bare `ruff check` from the repo root, matching what `make lint` runs:

| ruff | before this PR | after this PR |
|------|----------------|---------------|
| 0.13.0 | pass | pass |
| 0.15.22 | pass | pass |
| 0.16.0 | **2588 errors** | pass |

### Deliberately not included

- **Not pinning ruff in `ci.yml`.** Modifying CI/CD needs explicit approval per `CLAUDE.md`, and an explicit `select` makes the lint result version-independent anyway, which is the stronger guarantee. Worth pinning separately if maintainers want fully reproducible lint.
- **Not adopting the new 0.16 defaults.** Adding `I`, `SIM`, `UP`, `B`, `DTZ` and friends means 2588 findings to triage. That is a real conversation, just not one that should happen inside a CI unbreak.
- **Not touching `cli/python/pyproject.toml`.** It already declares its own explicit `select`, so it was never exposed to this.
- **Pre-existing and unrelated:** `ruff format --check` reports drift on 137 files, on both old and new ruff. CI does not run it. Untouched here.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None. Restores the rule set that was already in effect.

## Test Coverage

Ran `ruff check` from the repo root against 0.13.0, 0.15.22, and 0.16.0, before and after the change. Results in the table above. This PR touches `pyproject.toml`, so the Python SDK CI job runs on it and validates the fix directly.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ install:
 	hatch env create
 
 install_all:
-	pip install ruff==0.6.9 groq together boto3 litellm ollama chromadb weaviate weaviate-client sentence_transformers vertexai \
+	pip install ruff==0.16.0 groq together boto3 litellm ollama chromadb weaviate weaviate-client sentence_transformers vertexai \
 	            google-generativeai elasticsearch opensearch-py vecs "pinecone<7.0.0" pinecone-text faiss-cpu langchain-community \
 							upstash-vector azure-search-documents langchain-memgraph langchain-neo4j langchain-aws rank-bm25 pymochow pymongo psycopg kuzu databricks-sdk valkey
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ test = [
     "pytest-asyncio>=0.23.7",
 ]
 dev = [
-    "ruff>=0.6.5",
+    "ruff==0.16.0",
     "isort>=5.13.2",
     "pytest>=8.2.2",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,6 +150,10 @@ test = [
 line-length = 120
 exclude = ["openmemory/"]
 
+[tool.ruff.lint]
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Superseded by #6589, which carries the full long-term fix instead of the version pin.

Pinning only narrowed the blast radius: the repo had four independent ruff version sources (ci.yml unpinned, Makefile 0.6.9, dev extra >=0.6.5, pre-commit language: system), so pinning any one of them left the others free to drift. #6589 collapses them to a single pinned source with automated bumps.

Issue #6571, opened 2026-07-24
TS OSS Together embedder ignores TOGETHER_API_BASE while the Together LLM honors it


abhay-codes07:
### Component

TypeScript SDK / Embedders

### Where

`mem0-ts/src/oss/src/embeddings/together.ts`.

### What happens

The Together LLM resolves its base URL with an env fallback:

```ts
// llms/together.ts
baseURL: config.baseURL || process.env.TOGETHER_API_BASE || DEFAULT_BASE_URL,
```

but the Together embedder does not:

```ts
// embeddings/together.ts
baseURL: config.baseURL || config.url || DEFAULT_BASE_URL,
```

So a user who runs Together behind a gateway and sets `TOGETHER_API_BASE` gets it honored for chat but silently ignored for embeddings, which keep hitting `https://api.together.ai/v1`. Within one provider, the LLM and embedder disagree on how the base URL is resolved.

### Expected

The embedder should honor `TOGETHER_API_BASE` too: `config.baseURL || config.url || TOGETHER_API_BASE || default`, matching the LLM.

yashs33244:
Confirmed - the TS Together LLM reads `process.env.TOGETHER_API_BASE` (`llms/together.ts:16`), but the embedder (`embeddings/together.ts:20`) only does `config.baseURL || config.url || DEFAULT_BASE_URL`, so the two diverge exactly as described. Companion to #6587, which is the same gap on the Python side, so both SDKs need it.

PR #6572, opened 2026-07-24
fix(ts-oss/together): honor TOGETHER_API_BASE in the embedder

abhay-codes07:
### Description

The Together LLM resolves its base URL as `config.baseURL || process.env.TOGETHER_API_BASE || default`, but the Together embedder skipped the env var (`config.baseURL || config.url || default`). So a user who runs Together behind a gateway and sets `TOGETHER_API_BASE` gets it honored for chat but silently ignored for embeddings, which keep hitting `https://api.together.ai/v1`. Within one provider the LLM and embedder disagreed on base-URL resolution.

The fix adds `TOGETHER_API_BASE` to the embedder base-URL precedence (`config.baseURL || config.url || TOGETHER_API_BASE || default`), matching the LLM.

Closes #6571

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

`jest src/oss/tests/together-embedder.test.ts` (run from the `mem0-ts` root), 8 passed.

- Added `honors TOGETHER_API_BASE when no baseURL is configured` and `prefers an explicit baseURL over TOGETHER_API_BASE`.
- Reverting the source change fails the first.

### Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes

PR #6601, opened 2026-07-26
fix(ts/embeddings): honor TOGETHER_API_BASE in the Together embedder

cat0825:
@/Users/qianyuhe/oss-contrib/mem0/pr-6571.md


kartik-mem0:
This diff does not contain the change the title describes. `mem0-ts/src/oss/src/embeddings/together.ts` goes from 24 lines to 1, and `mem0-ts/src/oss/tests/together-embedder.test.ts` from 119 lines to 1, so merging this would delete the Together embedder and its entire test suite rather than add base URL handling.

That looks like a bad file write on your side rather than an intentional change. The same pattern is on your other four open PRs (#6602, #6603, #6604, #6605), each replacing whole source files with a single line, so it is probably worth fixing at the source before pushing again.

Happy to review the real diff once it is pushed. As it stands this cannot be merged.

cat0825:
Thanks for flagging this. I confirmed the branch was corrupted by a bad file write that collapsed both files to one line. The intended fix has already been merged in #6572 (which closed #6571), so pushing a replacement diff here would now be redundant. Closing this broken duplicate.

PR #6603, opened 2026-07-26
fix(ts/vector_stores): apply all Turbopuffer filter operators instead of dropping them

cat0825:
@/Users/qianyuhe/oss-contrib/mem0/pr-6577.md


kartik-mem0:
This diff does not contain the change the title describes. `mem0-ts/src/oss/src/vector_stores/turbopuffer.ts` goes from 281 lines to 1, and the added test file is a single line, so merging this would delete the Turbopuffer store rather than fix filter operator handling.

That looks like a bad file write on your side rather than an intentional change. The same pattern is on your other four open PRs (#6601, #6602, #6604, #6605), each replacing whole source files with a single line, so it is probably worth fixing at the source before pushing again.

For context, #6578 already implements this fix (full eq/ne/gt/gte/lt/lte/in/nin map, bare array treated as in, "*" skipped) with unit tests and green CI, so that is the one we are reviewing for this bug.

cat0825:
Thanks for the detailed review. I confirmed this branch was corrupted by the same bad file write. Since #6578 already contains the complete filter-operator fix with tests and green CI, I am closing this broken duplicate rather than duplicating that work.

PR #6604, opened 2026-07-26
fix(ts/llms): send Gemini system prompt via systemInstruction

cat0825:
@/Users/qianyuhe/oss-contrib/mem0/pr-6569.md


kartik-mem0:
This diff does not contain the change the title describes. `mem0-ts/src/oss/src/llms/google.ts` goes from 113 lines to 1, and `mem0-ts/src/oss/tests/google-llm.test.ts` from 259 lines to 1, so merging this would delete the Google LLM provider and its test suite rather than route the system prompt through systemInstruction.

That looks like a bad file write on your side rather than an intentional change. The same pattern is on your other four open PRs (#6601, #6602, #6603, #6605), each replacing whole source files with a single line, so it is probably worth fixing at the source before pushing again.

For context, #6570 already implements this fix, so that is the one we are reviewing for this bug.

cat0825:
Thanks for catching this. I confirmed this branch was corrupted by the same bad file write. Since #6570 already contains the Gemini systemInstruction and role-mapping fix with tests and green CI, I am closing this broken duplicate rather than duplicating that work.

PR #6605, opened 2026-07-26
fix(llms+embeddings/together): honor TOGETHER_API_BASE base URL

cat0825:
@/Users/qianyuhe/oss-contrib/mem0/pr-6587.md


kartik-mem0:
This diff does not contain the change the title describes. All four touched files are reduced to a single line each: `mem0/llms/together.py` (92 lines to 1), `mem0/embeddings/together.py` (43 to 1), `tests/llms/test_together.py` (104 to 1), and `tests/embeddings/test_together_embeddings.py` (87 to 1). Merging this would delete the Together provider and its tests rather than honor TOGETHER_API_BASE.

That looks like a bad file write on your side rather than an intentional change. The same pattern is on your other four open PRs (#6601, #6602, #6603, #6604), each replacing whole source files with a single line, so it is probably worth fixing at the source before pushing again.

For context, #6588 and #6609 both target this same behavior and are already in review.

cat0825:
Thanks for catching this. I confirmed all four files were corrupted by the same bad file write. Since #6588 already contains the intended Python Together base URL fix with tests and green CI (and #6609 also targets the same issue), I am closing this broken duplicate rather than duplicating that work.

PR #6610, opened 2026-07-26
fix(vector_stores/upstash): escape and validate metadata filter expressions

### LINKED ISSUE #5977: Upstash vector store: filter expression injection via unescaped quotes

OfficialAbhinavSingh:
## Linked Issue

Closes #5977

## Description

`UpstashVector._stringify()` wrapped string filter values in double quotes without escaping them, and the `key = value AND key = value` expression was assembled inline at three separate call sites with no validation of keys or values.

A filter value containing a double quote therefore closed its own string literal early, and the remainder was parsed as Upstash filter syntax. A caller who controls a single filter value could append clauses of their own:

```python
store.search(
    query="q",
    vectors=[[0.1, 0.2]],
    filters={"user_id": "alice", "name": 'x" OR user_id = "bob'},
)
# expression sent to Upstash:
#   user_id = "alice" AND name = "x" OR user_id = "bob"
```

Upstash's filter DSL gives `AND` higher precedence than `OR` when no parentheses are present ([docs](https://upstash.com/docs/vector/features/filtering)), so that parses as `(user_id = "alice" AND name = "x") OR user_id = "bob"` and returns records belonging to another tenant. The same route also breaks non-adversarial input: an honest value such as `note='say "hi"'` produced the malformed `note = "say "hi""`.

This PR:

1. Escapes `\` and `"` in `_stringify()`, backslash first so the escape character cannot itself be forged.
2. Routes `search()`, `keyword_search()` and `list()` through a single `_build_filter_expression()`, which additionally rejects filter keys that are not identifiers (`^[a-zA-Z_][a-zA-Z0-9_]*$`) and values that are not `str`/`int`/`float`/`bool`. This covers the second half of the issue, where dict and list values previously passed through silently.

Both parts mirror the hardening already present in `milvus.py` (`_create_filter`) and `baidu.py`, which use the same `_SAFE_FILTER_KEY` regex, the same escape order and the same scalar-type `ValueError`. `upstash_vector.py` was the remaining store with no escaping of any kind.

Two deliberate scope decisions:

- `"*"` wildcard handling is **not** added here. `upstash_vector.py` has no wildcard support today, and cross-provider wildcard semantics are being handled separately in #6539 / #6540, which do not touch this file.
- `keyword_search()` builds the expression **outside** its `try/except`, so an invalid filter raises consistently with `search()`/`list()` instead of being flattened into the `None` that method returns when sparse search is unavailable. The `Raises:` section was added to all three docstrings.

## Type of Change

## Breaking Changes

Not a breaking change for valid input. Values containing neither `"` nor `\` serialise byte-identically, so existing filter expressions are unchanged. This is covered by a test and by the fact that all 20 pre-existing tests in `test_upstash_vector.py` pass unmodified.

Filter keys that are not identifiers, and non-scalar filter values, now raise `ValueError` instead of being silently interpolated. That input could not have produced a correct Upstash expression before, so it was already broken rather than working.

## Test Coverage

15 new tests in `tests/vector_stores/test_upstash_vector.py`:

- injection payload escaped correctly in all three of `search()`, `keyword_search()` and `list()`
- backslash escaped before the quote, so a pre-escaped payload cannot decode back into a bare delimiter
- ordinary values serialise unchanged (regression guard for the compatibility claim above)
- filter keys rejected: `"user id"`, `"user-id"`, `"1abc"`, `'a" OR b'`, `""`, `"a.b"`
- filter values rejected: dict, list, arbitrary object
- `keyword_search()` surfaces the validation error rather than returning `None`

**Proof, red on `upstream/main` and green with the fix** (source file reverted to `upstream/main` with the new tests kept, then restored):

```
# BEFORE (upstream/main source, new tests):
$ pytest tests/vector_stores/test_upstash_vector.py -q
  15 failed, 20 passed

# AFTER (this branch):
$ pytest tests/vector_stores/test_upstash_vector.py -q
  35 passed

# Regression guards on this branch:
$ pytest tests/vector_stores/test_upstash_vector.py tests/vector_stores/test_qdrant.py -q
  131 passed
$ pytest tests/memory/ -q
  352 passed
$ ruff check mem0/vector_stores/upstash_vector.py tests/vector_stores/test_upstash_vector.py
  All checks passed!
```

The 20 pre-existing tests pass in both the red and green runs, which is what establishes that ordinary filter expressions are untouched.

One note for reviewers: `test_ordinary_filter_values_are_unchanged` fails in the red run because `_build_filter_expression` does not exist on `upstream/main`, not because the serialised output differs. It is a forward regression guard for the compatibility claim, not evidence of the bug. The 14 other new tests are load-bearing on the bug itself.

## Checklist

- [x] I have updated documentation if needed (N/A, no public API or documented behaviour change; docstrings updated in place)

### DIFF
--- a/mem0/vector_stores/upstash_vector.py
+++ b/mem0/vector_stores/upstash_vector.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from typing import Dict, List, Optional
 
 from pydantic import BaseModel
@@ -92,7 +93,46 @@ class UpstashVector(VectorStoreBase):
         )
 
     def _stringify(self, x):
-        return f'"{x}"' if isinstance(x, str) else x
+        """Render a filter value as an Upstash filter literal.
+
+        String values are quoted, with `\\` and `"` escaped first. Without the
+        escaping a value containing a double quote closes the literal early and
+        the remainder is parsed as filter syntax, so a caller who controls one
+        filter value can append clauses of their own and widen the match set
+        past the scoping filters (e.g. `user_id`).
+        """
+        if isinstance(x, str):
+            escaped = x.replace("\\", "\\\\").replace('"', '\\"')
+            return f'"{escaped}"'
+        return x
+
+    _SAFE_FILTER_KEY = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
+
+    def _build_filter_expression(self, filters: Optional[Dict]) -> Optional[str]:
+        """Build an Upstash metadata filter expression from `filters`.
+
+        Keys are restricted to identifier characters and values to scalars, so
+        neither can contribute unescaped filter syntax.
+
+        Raises:
+            ValueError: If a filter key is not an identifier, or a filter value
+                is not a str, int, float, or bool.
+        """
+        if not filters:
+            return None
+
+        operands = []
+        for key, value in filters.items():
+            if not self._SAFE_FILTER_KEY.match(key):
+                raise ValueError(f"Invalid filter key: {key!r}")
+            if not isinstance(value, (str, int, float, bool)):
+                raise ValueError(
+                    f"Filter value for {key!r} must be str, int, float, or bool, "
+                    f"got {type(value).__name__}"
+                )
+            operands.append(f"{key} = {self._stringify(value)}")
+
+        return " AND ".join(operands)
 
     def search(
         self,
@@ -111,9 +151,12 @@ class UpstashVector(VectorStoreBase):
 
         Returns:
             List[OutputData]: Search results.
+
+        Raises:
+            ValueError: If a filter key or value is invalid.
         """
 
-        filters_str = " AND ".join([f"{k} = {self._stringify(v)}" for k, v in filters.items()]) if filters else None
+        filters_str = self._build_filter_expression(filters)
 
         response = []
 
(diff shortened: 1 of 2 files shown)


kartik-mem0:
closing pr in favor of #5981 

PR #6615, opened 2026-07-27
fix(mem0): add test for issue #6550

### LINKED ISSUE #6550: Redis vector store ignores operator dicts and $or in filters (only scalar equality works)

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6550.

## Why
Fixes #6550

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6550.py
+++ b/tests/test_issue_6550.py
@@ -1,0 +1,68 @@
+"""Regression test for Redis advanced filter translation (#6550)."""
+
+from unittest.mock import MagicMock
+
+from mem0.vector_stores.redis import RedisDB
+
+
+class _FilterExpression:
+    """Small redisvl filter stand-in that makes the generated query inspectable."""
+
+    def __init__(self, value):
+        self.value = value
+
+    def __and__(self, other):
+        return _FilterExpression(("and", self.value, other.value))
+
+    def __or__(self, other):
+        return _FilterExpression(("or", self.value, other.value))
+
+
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6616, opened 2026-07-27
fix(mem0): add test for issue #6546

### LINKED ISSUE #6546: S3 Vectors search score is hardcoded for cosine, breaks euclidean ranking

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6546.

## Why
Fixes #6546

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6546.py
+++ b/tests/test_issue_6546.py
@@ -1,0 +1,35 @@
+"""Regression test for euclidean S3 Vectors score normalization."""
+
+from unittest.mock import MagicMock
+
+import pytest
+
+from mem0.vector_stores.s3_vectors import S3Vectors
+
+
+def test_issue_6546(monkeypatch):
+    """Euclidean distances above one must retain distinct similarity scores."""
+    mock_client = MagicMock()
+    mock_client.get_vector_bucket.return_value = {}
+    mock_client.get_index.return_value = {}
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6617, opened 2026-07-27
fix(mem0): add test for issue #6563

### LINKED ISSUE #6563: AWS Bedrock: tool calls send malformed Converse requests for amazon and cohere models

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #6563.

## Why
Fixes #6563

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_6563.py
+++ b/tests/test_issue_6563.py
@@ -1,0 +1,56 @@
+"""Regression tests for Bedrock Converse tool-call message formatting."""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from mem0.configs.llms.aws_bedrock import AWSBedrockConfig
+from mem0.llms.aws_bedrock import AWSBedrockLLM
+
+
+@pytest.mark.parametrize(
+    "model",
+    [
+        "amazon.nova-3-mini-20241119-v1:0",
+        "cohere.command-r-v1:0",
+    ],
+)
+def test_issue_6563(model):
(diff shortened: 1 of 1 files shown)


kartik-mem0:

kartik-mem0:

PR #6628, opened 2026-07-28
feat(vector_stores): add async token-bucket rate limiter for memory operations

glatinone:
Introduce an opt-in AsyncTokenBucketRateLimiter for vector store read/write operations with jittered exponential backoff on transient errors (429/5xx). Wired into batch memory retrieval and insertion paths. Includes unit tests for token acquisition, backoff delays, and transparent execution when unconstrained.


glatinone:
Closing this PR to avoid adding unnecessary review overhead for maintainers. Thank you for your time!

PR #6629, opened 2026-07-28
fix(memory): paginate delete_all past list() top_k default

### LINKED ISSUE #6627: `delete_all()` silently deletes only the first 100 memories
### Component

Python SDK

## Component

### Summary

### Root cause

`mem0/memory/main.py`:

```python
# sync, line 1889
for memory in memories:

# async, line 3518
```

### Affected vector stores

|---|---|---|

### Steps to Reproduce

```python
from mem0 import Memory

for i in range(150):
          user_id="u1", infer=False)

# Sanity check: 150 stored

# "Delete all" for this user
m.delete_all(user_id="u1")

```

### Expected Behavior

### Actual Behavior

### Environment

- Python version: 3.9+
- OS: all

### Suggested fix

```python
# sync
while True:
    if not memories:
        break
    for memory in memories:
```

```python
# async
while True:
    if not memories:
        break
```

Considerations:

### Severity rationale

### Location

### Willingness to contribute

Bartok9:
## Summary
`delete_all()` only deleted the first page of `vector_store.list()` (default `top_k=100`) but still returned success.

## Motivation
Closes #6627. Silent partial delete is a data-retention/privacy hazard.

## Fix
Page `list(filters=..., top_k=1000)` until a short/empty page in sync and async `Memory.delete_all`.

## Verification
`pytest tests/test_main.py::test_delete_all tests/test_main.py::test_delete_all_paginates_beyond_default_top_k` (+ entity-id delete_all cases) — passed.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -430,6 +430,10 @@ setup_config()
 logger = logging.getLogger(__name__)
 
 _UNSET = object()
+# Page size for delete_all pagination. Most vector stores default list() to
+# top_k=100, so a single call would silently leave later pages undeleted
+# (a privacy/GDPR hazard). Shared by the sync and async paths so they don't drift.
+_DELETE_ALL_PAGE_SIZE = 1000
 _PROJECT_UPDATE_UNSUPPORTED_ERROR = "Project updates are not supported by the OSS Memory SDK."
 
 
@@ -1887,7 +1891,7 @@ class Memory(MemoryBase):
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
         # Paginate list() — most stores default top_k=100; a single call would
         # silently leave undeleted memories past the first page (privacy risk).
-        page_size = 1000
+        page_size = _DELETE_ALL_PAGE_SIZE
         deleted_count = 0
         while True:
             memories = self.vector_store.list(filters=filters, top_k=page_size)[0]
@@ -3526,7 +3530,7 @@ class AsyncMemory(MemoryBase):
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "async"})
         # Paginate list() — most stores default top_k=100; a single call would
         # silently leave undeleted memories past the first page (privacy risk).
-        page_size = 1000
+        page_size = _DELETE_ALL_PAGE_SIZE
         deleted_ids = []
         while True:
             page = await asyncio.to_thread(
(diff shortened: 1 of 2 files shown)


kartik-mem0 (commented):
## Changes requested — async path can loop forever; and see #6636

Thanks for this. The diagnosis is right and the sync fix is correct: I ran it against a live hosted Supabase project (Postgres 17.6, `vecs` 0.4.5) on top of `origin/main` @ `29fa4155` and it drained cleanly.

| build | seeded | deleted | remaining | `list()` calls |
|---|---|---|---|---|
| `main` | 250 | 100 | **150** | 1 |
| this PR | 500 | **500** | **0** | 1 (`top_k=1000`) |

### Blocking: `AsyncMemory.delete_all` has no zero-progress guard

```python
...
if len(memories) < page_size:
    break
```

`return_exceptions=True` means a failed delete never raises, it lands in `results` and is only logged. So if a page of exactly `page_size` items all fail to delete (connection dropped mid-drain, credentials rotated, store went read-only), the next `list()` returns the same 1000 rows, `len(memories) == page_size`, and the loop repeats identically with no exit. That is an unbounded hang inside a library call, and the log line makes it look like progress.

The sync path is safe from this only because `self._delete_memory` propagates, so the call raises out rather than spinning. The async path swallows.

Two ways to fix: break when the batch made no progress (`if not results or all(isinstance(r, BaseException) for r in results): break`), or track the ids you have already attempted and stop when a page brings nothing new.

### Non-blocking

- `page_size = 1000` is a local literal in both methods. A module-level constant keeps the two paths from drifting.
- Neither new test covers the async loop, which is where the risk is.
- Breaking on `len(memories) < page_size` saves the confirming `list()` call, but it assumes every store returns exactly `top_k` when more rows remain. Across 27 backends that is a bet; looping until an empty page is returned costs one extra call and removes the assumption.

### Cluster note

#6627 has six open PRs. #6636 covers the same ground with the guard on both paths and updates the pre-existing tests this change affects, so my recommendation to the maintainers is to consolidate on that one. Nothing wrong with the work here, it just overlaps. If you would rather carry this one forward, fixing the async termination is the thing that would change my read.

Bartok9:
Thanks for the careful review and for running it against live Supabase — the async termination point is exactly right, and I've pushed a fix (b6ae43ac):

- **Zero-progress guard** on `AsyncMemory.delete_all`: since `gather(return_exceptions=True)` swallows every failure, a full page that all fails to delete would return the same rows forever. It now breaks when a page deletes nothing (`if not page_deleted:`) and logs a warning, so a dropped connection / rotated creds / read-only store can no longer hang the call.
- **Shared constant** `_DELETE_ALL_PAGE_SIZE = 1000` used by both paths so they can't drift.
- **Async tests** added: one for pagination past the default `top_k`, one asserting the no-progress guard bails after a single non-progressing page instead of looping.

On the `len(memories) < page_size` early-out assumption across 27 backends — fair point; I kept it since it's the same convention the sync path already relies on and it saves a round-trip, but happy to switch both to "loop until empty page" if you'd prefer removing the assumption entirely.

Fully agree #6636 covers both paths and updates the pre-existing tests — no objection to consolidating there if that's the maintainers' call. Just wanted to close the gap you flagged in case this one carries forward.

kartik-mem0:
Thanks for this.

Closing as a duplicate. Four PRs landed on this issue within a day of each other and only one could go in, which is on us for not triaging faster. Appreciate you taking it on.

Bartok9:
All good — consolidating on #6636 makes sense, and it's great to see delete_all paginating on main. Thanks for the live Supabase verification and the careful async-termination review; happy this one's resolved.

PR #6636, opened 2026-07-28
fix(memory): paginate delete_all across vector store pages

### LINKED ISSUE #6627: `delete_all()` silently deletes only the first 100 memories

tomatotomata:
## Summary

- paginate synchronous and asynchronous `delete_all()` calls until no matching memories remain
- use an explicit 1,000-item batch size instead of each vector store's 100-item default
- preserve async partial-failure handling while preventing endless retries when an entire batch fails
- add regression coverage for multi-page deletion and updated notice/count behavior

Closes #6627

## Validation

- `python -m pytest tests/test_main.py -k delete_all -q` (6 passed)
- direct asyncio pagination smoke test passed
- Python compilation and `git diff --check`
- Async pytest cases are present but cannot run locally because `pytest-asyncio` is unavailable.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1889,12 +1889,18 @@ class Memory(MemoryBase):
         # Keep listing after each batch is deleted. Most vector stores cap
         # list() at 100 results by default, which silently truncates deletes.
         deleted_count = 0
+        seen_batches = set()
         while True:
             memories = self.vector_store.list(
                 filters=filters, top_k=DELETE_ALL_BATCH_SIZE
             )[0]
             if not memories:
                 break
+            batch_ids = tuple(sorted(str(memory.id) for memory in memories))
+            if batch_ids in seen_batches:
+                logger.warning("Stopping delete_all after a repeated memory batch")
+                break
+            seen_batches.add(batch_ids)
             for memory in memories:
                 self._delete_memory(memory.id)
             deleted_count += len(memories)
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -133,6 +133,7 @@ _SENSITIVE_SUFFIXES = (
 
 # Entity parameters that must be passed via filters, not top-level kwargs
 ENTITY_PARAMS = frozenset({"user_id", "agent_id", "run_id"})
+DELETE_ALL_BATCH_SIZE = 1000
 
 # Tenant-scoping fields that update() must never let caller-supplied metadata overwrite (issues #4490, #6277).
 _IDENTITY_KEYS = ENTITY_PARAMS | {"actor_id"}
@@ -1885,14 +1886,22 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
-        memories = self.vector_store.list(filters=filters)[0]
-        for memory in memories:
-            self._delete_memory(memory.id)
+        # Keep listing after each batch is deleted. Most vector stores cap
+        # list() at 100 results by default, which silently truncates deletes.
+        deleted_count = 0
+        while True:
+            memories = self.vector_store.list(
+                filters=filters, top_k=DELETE_ALL_BATCH_SIZE
+            )[0]
+            if not memories:
+                break
+            for memory in memories:
+                self._delete_memory(memory.id)
+            deleted_count += len(memories)
 
-        logger.info(f"Deleted {len(memories)} memories")
+        logger.info(f"Deleted {deleted_count} memories")
 
-        decay_usage_notice = detect_decay_usage_from_delete_all(len(memories))
+        decay_usage_notice = detect_decay_usage_from_delete_all(deleted_count)
         if decay_usage_notice:
             display_decay_usage_notice(self, "sync", "delete_all", *decay_usage_notice)
         else:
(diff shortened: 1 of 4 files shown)


Solaris-star (commented):
Reviewed the diff — the pagination approach is clearly the right fix here (vs. just bumping top_k, which would still cap out). One substantive observation and a couple of notes:

**Sync/async asymmetry — potential infinite loop worth confirming**

The async path has a no-progress guard:
```python
if len(batch_errors) == len(batch):
    break
```
but the sync path's `while True` has no equivalent. The sync loop relies entirely on `_delete_memory` *raising* on failure to exit. If `_delete_memory` can ever fail *silently* (return without raising but without actually removing the item), the sync `while True` re-lists the same undeleted batch forever.

The pre-pagination code was single-pass, so this infinite-loop risk is newly introduced by the loop structure rather than pre-existing. Worth confirming that `_delete_memory` is guaranteed to raise on any failure path — if not, the sync loop should mirror the async all-fail guard (e.g. track whether the batch shrank / count consecutive no-progress iterations and break).

**Notes (non-blocking)**
- The async partial-failure handling is preserved correctly: as long as at least one deletion makes progress per iteration the loop terminates, and the all-fail guard catches the stuck case. Nice.
- Test coverage is solid — `test_delete_all_paginates_beyond_vector_store_page_size` verifies 1001 deletions across 3 list calls, and the existing tests were correctly updated to `side_effect=[(batch, None), ([], None)]` to model the terminating empty page.
- The `top_k=1000` batch assertion in the updated tests pins the batch size nicely.

tomatotomata:
Addressed the sync-path no-progress concern in follow-up commit `b6a1b7e`:

- track each listed batch by its memory IDs
- stop with a warning if the vector store repeats a batch, preventing an infinite loop when deletion silently makes no progress
- add regression coverage for a repeated batch while preserving the existing pagination and notice tests

Validation: three focused sync `delete_all` tests pass. The broader file includes async tests, but this Windows environment lacks the async pytest plugin.

tomatotomata:
Correction: the follow-up commit is 23c550b (the prior comment's 6a1b7e was a typo).

tomatotomata:
Thanks for flagging this. The sync path could not safely assume that every unsuccessful delete raises.

The current head (`23c550b`) now records each listed batch by memory IDs and stops before processing a repeated batch. The accompanying regression supplies the same batch twice and verifies that it is listed twice but deleted only once. This prevents the new pagination loop from repeatedly deleting the same unresolved batch without assuming a vector-store failure mode.

kartik-mem0:

tomatotomata:
Marked ready for review. I re-ran the three targeted synchronous `delete_all` regression tests on the current head; all pass, and `git diff --check` is clean. The local environment still lacks the async pytest plugin, so I have not claimed the async tests ran.

kartik-mem0 (commented):
## LGTM — this is the one I'd take from the #6627 cluster

Six open PRs target this bug. I read all six diffs and ran the four mergeable ones against a live hosted Supabase project. This one is the pick.

### The bug, confirmed

`Memory.delete_all` calls `vector_store.list(filters=filters)` with no `top_k` (`mem0/memory/main.py:1889`). Every one of the 27 Python vector stores defaults that argument to 100, so `delete_all` removes the first 100 matches and then returns `{"message": "Memories deleted successfully!"}`. Silent partial deletion, which for a delete API is a data-retention problem rather than a convenience bug. `AsyncMemory.delete_all` has the same defect at `:3518`.

Measured on a real Supabase project (Postgres 17.6, `vecs` 0.4.5), base `origin/main` @ `29fa4155` with only `mem0/memory/main.py` swapped:

| build | seeded | deleted | remaining | `list()` calls | `top_k` passed |
|---|---|---|---|---|---|
| `main` | 250 | 100 | **150** | 1 | none |
| this PR | 250 | **250** | **0** | 2 | 1000, 1000 |

### Red-green, this run

At `23c550b7d463172af24304ba73f4b6833dbc1d99`, targeted tests: `12 passed`. Reverting only the implementation (`git checkout main -- mem0/memory/main.py`) and re-running gives 3 failures:

- `tests/test_main.py::test_delete_all_paginates_beyond_vector_store_page_size`
- `tests/memory/test_decay_usage_notice.py::test_sync_delete_all_stops_when_vector_store_repeats_batch`
- `tests/test_memory.py::test_async_delete_all_continues_on_partial_failure`

Restored, back to pass. The pagination test reproduces the reported symptom (memories surviving `delete_all` at 1001 rows), not just the newly added mechanism, which is what I care about for a data-loss fix.

### Why this one over the other four

- **Both paths guarded.** Sync breaks on a repeated batch; async breaks when every delete in a batch fails. #6629's async loop breaks only on `len(memories) < page_size`, so a full page whose deletes all fail (connection lost mid-drain, permissions revoked) re-lists identically and spins forever, because `return_exceptions=True` swallows the failures.
- **It actually raises the page size.** #6640 loops correctly but never passes `top_k`, so it keeps asking for 100 at a time. Correct, just 10x the round trips against a remote store.
- **It stays small.** +95/-23, one new module constant, no drive-by edits. #6487 solves the same problem in +402/-22 with five module constants, a `seen_ids` set, a per-id failure-count dict and backoff sleeps.
- **It doesn't trust the store to honour `top_k` exactly.** Looping until an empty page comes back is the right contract for 27 heterogeneous backends.
- `top_k=1000` is also exactly the `vecs` ceiling, so it is the largest value Supabase can serve in one call.

### Minor, non-blocking

`seen_batches` stores the full sorted id tuple for every page, so a drain of N memories retains N id strings for the life of the call. Storing `hash(batch_ids)` instead would keep the guard and drop the retention. Happy to merge without it.

### Follow-up, not for this PR

The TypeScript SDK has the identical defect at `mem0-ts/src/oss/src/memory/index.ts:1729` (`this.vectorStore.list(filters)`; TS stores also default `topK = 100`). No mergeable PR in this cluster covers it. Worth a tracked follow-up so the two SDKs don't drift.

kartik-mem0 (approved):

PR #6640, opened 2026-07-28
fix: paginate delete_all so it deletes more than the first list() page

### LINKED ISSUE #6627: `delete_all()` silently deletes only the first 100 memories

Mohith26:
Fixes #6627

## What

`delete_all()` fetched one `list()` page (default 100) and deleted those, so stores with more than 100 memories were only partially cleared while the call reported success.

The fix paginates until the store is exhausted, following the same pagination pattern already used elsewhere in the client, and deletes every page.

## Tests

Regression tests mock a store with >100 memories across multiple pages and assert every memory is deleted. They fail on main (only the first page is deleted) and pass with the fix. No new failures in the surrounding suites.

Credit to @soullst for the report in #6627.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1885,14 +1885,30 @@ class Memory(MemoryBase):
 
         keys, encoded_ids = process_telemetry_filters(filters)
         capture_event("mem0.delete_all", self, {"keys": keys, "encoded_ids": encoded_ids, "sync_type": "sync"})
-        # delete all vector memories and reset the collections
-        memories = self.vector_store.list(filters=filters)[0]
-        for memory in memories:
-            self._delete_memory(memory.id)
+        # Delete all matching vector memories. `vector_store.list` is paginated (most
+        # stores default to a 100-item page), so keep listing and deleting until no
+        # matching memories remain instead of trusting a single page.
+        deleted_count = 0
+        previous_page_ids: set = set()
+        while True:
+            memories = self.vector_store.list(filters=filters)[0]
+            if not memories:
+                break
+            page_ids = {memory.id for memory in memories}
+            if page_ids == previous_page_ids:
+                logger.warning(
+                    "delete_all made no progress on %d remaining memories; stopping to avoid an infinite loop",
+                    len(page_ids),
+                )
+                break
+            previous_page_ids = page_ids
+            for memory in memories:
+                self._delete_memory(memory.id)
+            deleted_count += len(memories)
 
-        logger.info(f"Deleted {len(memories)} memories")
+        logger.info(f"Deleted {deleted_count} memories")
 
-        decay_usage_notice = detect_decay_usage_from_delete_all(len(memories))
+        decay_usage_notice = detect_decay_usage_from_delete_all(deleted_count)
         if decay_usage_notice:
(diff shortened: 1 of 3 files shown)


kartik-mem0 (commented):
## Correct, but it leaves the page size at 100 — see #6636

Thanks for this. The loop is right and the zero-progress guard is the good kind: comparing the page's id set against the previous page terminates on both full and partial delete failure, which two of the competing PRs get wrong.

I ran it against a live hosted Supabase project (Postgres 17.6, `vecs` 0.4.5) on top of `origin/main` @ `29fa4155`, with only `mem0/memory/main.py` swapped per variant. It drains fully:

| build | seeded | deleted | remaining | `list()` calls | `top_k` passed |
|---|---|---|---|---|---|
| `main` | 250 | 100 | **150** | 1 | none |
| this PR | 500 | **500** | **0** | 6 | none |
| #6636 | 250 | 250 | 0 | 2 | 1000 |

### The one substantive gap

The comment names the cause exactly:

```python
# `vector_store.list` is paginated (most stores default to a 100-item page), so keep
# listing and deleting until no matching memories remain instead of trusting a single page.
memories = (await asyncio.to_thread(self.vector_store.list, filters=filters))[0]
```

but the call still doesn't pass `top_k`, so every page really is 100 rows — that's the measured 6 `list()` calls for 500 memories above, where #6636 needs 2. The loop makes the behaviour correct instead of lossy, which is the important half, but it also means a 10,000-memory `delete_all` issues 100 `list()` round trips against a remote store where 10 would do. All 27 Python vector stores accept `top_k`, and 1000 is the largest value Supabase/`vecs` will serve in one call, so `top_k=1000` is both safe everywhere and the natural ceiling.

That is a one-line change here: pass `top_k=1000` (as a module constant) in both the sync and async `list()` calls.

### Cluster note

#6627 has six open PRs. #6636 lands the same loop with the batch size raised and updates the pre-existing tests the change affects, so my recommendation to the maintainers is to consolidate on that one. Your progress guard is the more robust of the two designs, and it is worth saying so — if the maintainers prefer this PR, adding `top_k` is the only thing standing between it and a merge.

kartik-mem0:
Thanks for this, and sorry for the outcome.

Your tests were the most complete of the four PRs on this issue, covering the sync and async paths against both pagination and the no-progress case. That async no-progress test is exactly the gap I found and fixed in #6636 before merging it, so the catch came from here. The `top_k` commit I pushed to your branch is no longer needed.

Closing as a duplicate. Genuinely appreciate the work.

PR #6656, opened 2026-07-29
fix(memory): stop add() metadata from setting a memory's identity scope

### LINKED ISSUE #6655: Python OSS SDK: add() lets metadata set identity scope (user_id/agent_id/run_id/actor_id) on creation
### Component

Python SDK

### Summary

The Python OSS SDK's `add()` has the same identity-field bug class as `update()` (fixed in #6278 for #6277), but on the **creation** path: caller-supplied `metadata` can set a memory's identity scope (`user_id` / `agent_id` / `run_id` / `actor_id`) whenever the matching top-level entity param is not also passed.

This is the Python twin of #6371 (TypeScript `add()`, PR #6377). `update()` is already hardened on both sides; `add()` is not hardened on either.

### Where

`mem0/memory/main.py`, `_build_filters_and_metadata()`:

```python
base_metadata_template = deepcopy(input_metadata) if input_metadata else {}
...
if user_id:
    base_metadata_template["user_id"] = user_id
if agent_id:
    base_metadata_template["agent_id"] = agent_id
if run_id:
    base_metadata_template["run_id"] = run_id
```

The caller's `metadata` is deep-copied wholesale into the stored payload template, and the identity keys are re-pinned **only when the corresponding argument is truthy**. So for any identity key the caller does not pass as a top-level argument, whatever value was in `metadata` survives into the created memory. `actor_id` is never re-pinned here at all.

The repo already has the right helper for this, `_strip_identity_keys()` with the module-level `_IDENTITY_KEYS = ENTITY_PARAMS | {"actor_id"}`, but it is only called from the two `_update_memory` paths, never from the creation path.

### Steps to Reproduce

```python
from mem0.memory.main import _build_filters_and_metadata

metadata, _ = _build_filters_and_metadata(
    user_id="attacker",
    input_metadata={"agent_id": "victim-agent", "run_id": "victim-run", "actor_id": "victim-actor"},
)
print(metadata)
```

End to end through the public API, with the vector store mocked to capture the payload:

```python
memory.add(
    "I like coffee",
    user_id="attacker",
    metadata={"agent_id": "victim-agent", "run_id": "victim-run"},
    infer=False,
)
```

### Expected Behavior

Identity scope is set only through the documented top-level `user_id` / `agent_id` / `run_id` params, never through freeform `metadata`, matching what `update()` already enforces.

### Actual Behavior

Both reproduce the injection:

```
# helper:
{'agent_id': 'victim-agent', 'run_id': 'victim-run', 'actor_id': 'victim-actor', 'user_id': 'attacker'}

# through Memory.add(), payload handed to the vector store:
{'agent_id': 'victim-agent', 'run_id': 'victim-run', 'user_id': 'attacker'}
```

The created memory is now discoverable through `get_all(filters={"agent_id": "victim-agent"})` and `search(..., filters={"run_id": "victim-run"})`, scopes the caller never held.

For contrast, the same input on the update path is correctly rejected:

```
update(): ignoring metadata['user_id'] - identity fields are immutable after creation
update(): ignoring metadata['agent_id'] - identity fields are immutable after creation
-> {'note': 'ok'}
```

Note an explicitly passed argument does still win over `metadata` for the same key (`user_id="attacker"` plus `metadata={"user_id": "victim"}` stores `attacker`), so the gap is exactly the keys the caller leaves unset.

### Impact

Same tenant-isolation class as #6277 / #6342, on the creation path rather than update. It is a cross-scope **write**: an application that forwards user-controlled metadata into `add()` lets a caller place memories into scopes it does not own. Both sync `Memory.add()` and `AsyncMemory.add()` are affected, on both `infer=True` and `infer=False`, since the template is deep-copied into every per-memory payload.

### Environment

- mem0 version: 2.0.14 (reproduced on `main` @ ea2ee075)
- Python version: 3.12
- OS: Linux

### Suggested Fix

Apply the existing `_IDENTITY_KEYS` filter to the incoming metadata in `_build_filters_and_metadata()` before it becomes the payload template, so identity scope can only be set via the top-level params. That mirrors the `update()` fix (#6278) and the TS `add()` fix proposed in #6371, and keeps a single source of truth for the key list.

Happy to open the PR.

OfficialAbhinavSingh:
## Linked Issue

Closes #6655

## Description

`add()` let caller-supplied `metadata` set a memory's identity scope. In `_build_filters_and_metadata()` the caller's metadata was deep-copied straight into the payload template, and the identity keys were re-pinned only when the matching entity param was truthy:

```python
base_metadata_template = deepcopy(input_metadata) if input_metadata else {}
...
if user_id:
    base_metadata_template["user_id"] = user_id
if agent_id:
    base_metadata_template["agent_id"] = agent_id
```

For any identity key the caller did not pass as a top-level argument, that re-pin never fired, so whatever sat in `metadata` survived into the created memory. `actor_id` was never re-pinned there at all.

A caller scoped by `user_id` alone could therefore write into scopes it never held:

```python
m.add("I like coffee", user_id="attacker",
      metadata={"agent_id": "victim-agent", "run_id": "victim-run"})
# stored payload: {'agent_id': 'victim-agent', 'run_id': 'victim-run', 'user_id': 'attacker'}
```

That memory is then reachable from `get_all(filters={"agent_id": "victim-agent"})` and `search(..., filters={"run_id": "victim-run"})`.

`update()` already refuses exactly this through `_strip_identity_keys()` (#6277 / #6278), so the creation and update paths disagreed about whether identity fields can be set through metadata. This PR applies the same helper on the creation path.

Why `_build_filters_and_metadata()` is the right place: it is the only chokepoint that both creation paths share. Only two call sites pass `input_metadata`, sync `add()` and async `add()`, so one change covers both, and it covers `infer=True` and `infer=False` alike because they share the same template. The helper gains a `context` argument so the warning names the calling method, and its message now states the rule that holds on both paths rather than only the update-path phrasing.

Passing an entity param is unchanged and remains the only way to set scope. An explicitly passed argument already won over metadata for the same key, so the gap was exactly the keys a caller left unset.

This is the Python twin of #6371 / #6377 (TypeScript `add()`). `update()` is already handled on both sides.

## Type of Change

## Breaking Changes

Not a breaking change for the documented API. Identity scope was always meant to come from the `user_id` / `agent_id` / `run_id` params, and those behave identically.

Code that relied on setting scope through `metadata` will now find those keys dropped, with a warning naming each one. That path was writing into scopes the caller did not request, which is the bug being fixed. Non-identity metadata is untouched.

One shared-helper detail worth calling out for review: `_strip_identity_keys()` had its warning text changed from "identity fields are immutable after creation" to "identity fields cannot be set through metadata", so a single message is accurate on both the add and update paths. The existing assertion in `tests/memory/test_main.py` checks the `ignoring metadata['user_id']` substring, which still holds.

## Test Coverage

Four tests in `tests/memory/test_main.py`, placed next to the existing `update()` identity-field pair:

- `test_add_metadata_cannot_set_identity_fields` - caller scopes by `user_id` only; `agent_id`/`run_id`/`actor_id` in metadata must not reach the payload, non-identity metadata must survive, warning emitted
- `test_async_add_metadata_cannot_set_identity_fields` - the async twin, written rather than assumed
- `test_add_entity_params_still_set_scope` - the documented params remain the only way to set scope
- `test_add_without_metadata_is_unaffected` - no metadata means no stripping and no behaviour change

**Proof, red on `upstream/main` and green with the fix** (source file reverted to `upstream/main` with the new tests kept, then restored):

```
# BEFORE (upstream/main source, new tests):
$ pytest tests/memory/test_main.py -k "identity or add_without_metadata or entity_params_still" -q
  2 failed, 4 passed
  AssertionError: agent_id was injected through add() metadata
  AssertionError: agent_id was injected through async add() metadata

# AFTER (this branch):
  6 passed

# Full suite on this branch:
$ pytest tests/memory/ -q
  356 passed
$ ruff check mem0/memory/main.py tests/memory/test_main.py
  All checks passed!
```

To be precise about what each test proves: the two that go red are the load-bearing ones. `test_add_entity_params_still_set_scope` and `test_add_without_metadata_is_unaffected` pass on unpatched `main` by design; they are compatibility guards against a future over-broad strip, not evidence of the bug.

The two pre-existing `update()` identity tests pass unchanged, which is what covers the shared-helper edit.

## Checklist

- [x] I have updated documentation if needed (N/A for public docs; the `_build_filters_and_metadata` docstring now states that identity scope comes from the entity params only)

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -134,18 +134,30 @@ _SENSITIVE_SUFFIXES = (
 # Entity parameters that must be passed via filters, not top-level kwargs
 ENTITY_PARAMS = frozenset({"user_id", "agent_id", "run_id"})
 
-# Tenant-scoping fields that update() must never let caller-supplied metadata overwrite (issues #4490, #6277).
+# Tenant-scoping fields that caller-supplied metadata must never set, on either the
+# creation or the update path (issues #4490, #6277, #6655).
 _IDENTITY_KEYS = ENTITY_PARAMS | {"actor_id"}
 
 
-def _strip_identity_keys(metadata: Dict[str, Any], existing_payload: Dict[str, Any]) -> Dict[str, Any]:
-    """Drop identity keys from caller metadata; they are immutable after creation (issues #4490, #6277)."""
+def _strip_identity_keys(
+    metadata: Dict[str, Any],
+    existing_payload: Dict[str, Any],
+    *,
+    context: str = "update()",
+) -> Dict[str, Any]:
+    """Drop identity keys from caller metadata; scope is set by the entity params, not metadata.
+
+    On the update path `existing_payload` carries the memory's current scope, so
+    re-sending an identical value is silently accepted; only a changed value warns.
+    On the creation path there is no prior payload, so pass an empty dict and every
+    identity key present in `metadata` is dropped with a warning.
+    """
     clean = {}
     for key, value in metadata.items():
         if key not in _IDENTITY_KEYS:
             clean[key] = value
         elif value != existing_payload.get(key):
-            logger.warning(f"update(): ignoring metadata['{key}'] - identity fields are immutable after creation")
+            logger.warning(f"{context}: ignoring metadata['{key}'] - identity fields cannot be set through metadata")
     return clean
 
 
@@ -314,7 +326,9 @@ def _build_filters_and_metadata(
     for flexible session scoping and optionally narrows queries to a specific `actor_id`. It returns two dicts:
 
     1. `base_metadata_template`: Used as a template for metadata when storing new memories.
-       It includes all provided session identifier(s) and any `input_metadata`.
+       It includes all provided session identifier(s) and any `input_metadata`. Identity
+       scope is set from the entity params only; identity keys in `input_metadata` are
+       dropped, so freeform metadata cannot place a memory into an unrequested scope.
     2. `effective_query_filters`: Used for querying existing memories. It includes all
        provided session identifier(s), any `input_filters`, and a resolved actor
        identifier for targeted filtering if specified by any actor-related inputs.
@@ -342,7 +356,12 @@ def _build_filters_and_metadata(
               scoped to the provided session(s) and potentially a resolved actor.
     """
 
-    base_metadata_template = deepcopy(input_metadata) if input_metadata else {}
+    # Identity scope is set below from the entity params only. Stripping the keys here
+    # stops caller metadata from placing a memory into a scope the caller did not pass,
+    # which the re-pins below cannot prevent for a param that was left unset (issue #6655).
+    base_metadata_template = (
+        _strip_identity_keys(deepcopy(input_metadata), {}, context="add()") if input_metadata else {}
+    )
     effective_query_filters = deepcopy(input_filters) if input_filters else {}
 
     # ---------- validate and add all provided session ids ----------
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -453,6 +453,95 @@ def test_update_memory_metadata_cannot_change_identity_fields(mocker, caplog):
     assert "ignoring metadata['user_id']" in caplog.text
 
 
+_ATTACKER_ADD_METADATA = {
+    "agent_id": "victim-agent",
+    "run_id": "victim-run",
+    "actor_id": "victim-actor",
+    "category": "sports",
+}
+
+
+def _captured_add_metadata(memory, mocker, **add_kwargs):
+    """Run add() with the pipeline stubbed and return the metadata template it produced."""
+    captured = {}
+
+    def _capture(messages, metadata, filters, infer, **kwargs):
+        captured.update(metadata)
+        return []
+
+    mocker.patch.object(memory, "_add_to_vector_store", side_effect=_capture)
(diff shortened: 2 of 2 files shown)


kartik-mem0 (commented):

`add()` let caller-supplied `metadata` silently set `agent_id`/`run_id`/`actor_id`, writing into scopes the caller never held; this strips those keys via the same helper `update()` already trusts.

**Follow-up:** TS `add()` has the same bug (#6371), fix PR #6377 is open but unmerged
**Not verified:** only oss-qdrant and oss-pgvector sandboxed, not every vector store backend

Risk: low, root-cause fix at the one shared chokepoint, proven red-green on unit tests and real infra · Your time: ~5 min

- **Problem is real:** reproduces on `main` at `mem0/memory/main.py:345,356-369`; confirmed against real Qdrant + pgvector, sync + async, all 4 combos leaked into `agent_id`/`run_id` scope
- **Tests:** targeted identity tests -> 6 passed
- **Red-green:** reverting `main.py` to pre-PR `main` -> 2 failed, matching the PR's own claimed proof; restored -> 6 passed
- **CI:** build_mem0 3.10-3.12 green, CI Gate green
- **Security:** cleared, tenancy fix verified against real infra, no new attack surface
- **Bloat:** lean already, +25/-6 in one function
- **Linked issue:** #6655

</details>

kartik-mem0 (approved):

PR #4089, opened 2026-02-20
fix: correct type mismatch in Langchain Vector Store Adapter update method

hztBUAA:
## Summary
Fixes #3767

- The `update` method in `mem0/vector_stores/langchain.py` passes `vector` and `payload` directly to `insert()`, but `insert()` expects `List[List[float]]` and `List[Dict]` respectively
- Wrapped `vector` and `payload` in lists (`[vector]` and `[payload]`) to match the expected parameter types

## Changes
- `mem0/vector_stores/langchain.py`: Changed `self.insert(vector, payload, [vector_id])` to `self.insert([vector], [payload], [vector_id])`

## Test Plan
- All 10 existing Langchain vector store tests pass

Issue #3182, opened 2025-07-18
Incompatible with LangChain's FAISS Vector Store


EshinYe:
### 🐛 Describe the bug

when using langchian FAISS Vector Store：
vector_store config：
vector_store = FAISS(
            embedding_function=openai_embeddings,
            index=index,
            docstore=InMemoryDocstore(),
            index_to_docstore_id={},
        )

Memory config add:
"vector_store": {
                "provider": "langchain",
                "config": {
                    "client": vector_store,
                    # "embedding_model_dims":embedding_dim
                }
            }

when call :   libenv/Lib/site-packages/mem0/vector_stores/langchain.py:88:
            self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
error happpens:
          Error: FAISS.add_embeddings() missing 1 required positional argument: 'text_embeddings'

langchain_community.vectorstores.faiss.FAISS.add_embeddings have argument text_embeddings but no embeddings

EshinYe:
reference : https://docs.mem0.ai/components/vectordbs/dbs/langchain#supported-langchain-vector-stores

kartik-mem0:
Hi! Thanks for raising this — we appreciate your input.

As part of our efforts to keep the issue tracker focused and up-to-date, we're closing issues that have been inactive for an extended period. Mem0 has seen a lot of changes recently, and this issue may no longer be applicable. If you're still facing this problem, please open a new issue with updated details and we'll be happy to assist.

Thank you for your patience! 🙏

PR #6658, opened 2026-07-29
fix(langchain): dispatch add_embeddings on its real signature so Memory.add() works with FAISS/PGVector

### LINKED ISSUE #6657: LangChain vector store: insert() calls add_embeddings() with a keyword no store accepts, so Memory.add() raises TypeError
### 🐛 Describe the bug

`Langchain.insert()` calls `add_embeddings` with a keyword no LangChain vector store accepts, so `Memory.add()` raises `TypeError` for every store that has that method.

`mem0/vector_stores/langchain.py:96`:

```python
if hasattr(self.client, "add_embeddings"):
    self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
```

Reproduced on current `main` (`ea2ee07`) with real FAISS:

```
TypeError: FAISS.add_embeddings() missing 1 required positional argument: 'text_embeddings'
```

I went through every `langchain_community` vector store that defines `add_embeddings` — 12 of them — and **none** match this call:

| signature | stores |
|---|---|
| `add_embeddings(text_embeddings, metadatas, ids, **kw)` — (text, vector) pairs | FAISS, AzureSearch, ElasticsearchStore, OpenSearchVectorSearch, ScaNN |
| `add_embeddings(texts, embeddings, metadatas, ids, **kw)` — `texts` **required** | PGVector, Neo4jVector, Lantern, Hologres, Kinetica, PGEmbedding, TimescaleVector |

The first group doesn't have an `embeddings` parameter at all. The second does, but `texts` is a required positional that mem0 never passes — so it fails the same way. The `hasattr` guard makes the branch look safe while guaranteeing the call is wrong.

Stores *without* `add_embeddings` (`Chroma`, `InMemoryVectorStore`) fall through to `add_texts` and work, which is why this hasn't shown up in the test suite — the existing tests use `Mock(spec=VectorStore)`, and a Mock accepts any keyword.

This is the same problem as #3182 (filed 2025-07-18, closed 2026-03-19 as inactive). It was never fixed and still reproduces.

### Second, related defect: `update(vector=None)` stores `None` as the embedding

```python
    self.delete(vector_id)
```

`vector` is optional, and mem0 relies on that — `Memory._upsert_entity` and `_remove_memory_from_entity_store` edit an entity's `linked_memory_ids` without re-embedding (`mem0/memory/main.py:607`, `:1133`, `:2249`, `:2765`):

```python
self.entity_store.update(vector_id=..., vector=None, payload=...)
```

`None` is then forwarded as the embedding — `add_embeddings(embeddings=[None], ...)`. Because `delete()` has already run, a failing insert leaves the row **gone**:

```
TypeError: FAISS.add_embeddings() missing 1 required positional argument: 'text_embeddings'
row was already deleted, ntotal = 0
```

### Reachability

This is the documented configuration (`docs/components/vectordbs/dbs/langchain.mdx`):

```python
Memory.from_config({"vector_store": {"provider": "langchain", "config": {"client": vector_store}}})
```

and that page's "Supported LangChain Vector Stores" section names Chroma, **FAISS**, **Pinecone**, **Weaviate**, **Milvus**, **Qdrant** "and many more". FAISS is listed and cannot store a single memory.

### Expected behavior

`Memory.add()` should work with a FAISS-backed (or PGVector-backed, etc.) LangChain client, storing the embedding mem0 computed. A payload-only `update()` should keep the row with a usable vector rather than writing `None` or deleting it.

### Proposed fix

Dispatch on `add_embeddings`' actual parameter names instead of assuming one shape, and for `update(vector=None)` re-insert via `add_texts` so the store embeds the text itself. I have this working with tests and can open a PR.

### Environment

- mem0 `main` @ `ea2ee07`
- `langchain-community` / `langchain-core`, `faiss-cpu`
- Python 3.13, Windows

LHMQ878:
Closes #6657

## Summary

`Langchain.insert()` called `add_embeddings` with a keyword that no LangChain vector store accepts:

```python
if hasattr(self.client, "add_embeddings"):
    self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
```

`add_embeddings` is not part of LangChain's `VectorStore` contract, and the implementations that do have it disagree on its signature. I went through all 12 in `langchain_community` — **none** match that call:

| signature | stores | why the call fails |
|---|---|---|
| `add_embeddings(text_embeddings, metadatas, ids, **kw)` | FAISS, AzureSearch, ElasticsearchStore, OpenSearchVectorSearch, ScaNN | no `embeddings` parameter at all; wants `(text, vector)` pairs |
| `add_embeddings(texts, embeddings, metadatas, ids, **kw)` | PGVector, Neo4jVector, Lantern, Hologres, Kinetica, PGEmbedding, TimescaleVector | has `embeddings`, but `texts` is a required positional mem0 never passes |

So `Memory.add()` raised `TypeError` for every store with the method. On real FAISS, on `main`:

```
TypeError: FAISS.add_embeddings() missing 1 required positional argument: 'text_embeddings'
```

The `hasattr` guard is what hid this: it makes the branch look conditional and safe, while in fact guaranteeing the call is wrong. Stores *without* `add_embeddings` (Chroma, `InMemoryVectorStore`) fall through to `add_texts` and work — that is the only reason the provider appeared functional.

This is #3182, filed 2025-07-18 and closed 2026-03-19 as inactive. It was never fixed and still reproduces.

### Second defect on the same path: `update(vector=None)` writes `None` as the embedding

```python
    self.delete(vector_id)
```

`vector` is optional and mem0 depends on that — `Memory._upsert_entity` / `_remove_memory_from_entity_store` edit an entity's `linked_memory_ids` without re-embedding (`mem0/memory/main.py:607`, `:1133`, `:2249`, `:2765`):

```python
self.entity_store.update(vector_id=..., vector=None, payload=...)
```

`None` was forwarded as the embedding — `add_embeddings(embeddings=[None], ...)`. And because `delete()` runs first, a rejected insert leaves the row **gone**:

```
TypeError: FAISS.add_embeddings() missing 1 required positional argument: 'text_embeddings'
row was already deleted, ntotal = 0
```

`Memory._upsert_entity` wraps this in `try/except Exception` and logs a warning, so on FAISS the entity silently vanished instead of crashing.

## This is the documented configuration

`docs/components/vectordbs/dbs/langchain.mdx`:

```python
Memory.from_config({"vector_store": {"provider": "langchain", "config": {"client": vector_store}}})
```

That page's "Supported LangChain Vector Stores" section names Chroma, **FAISS**, **Pinecone**, **Weaviate**, **Milvus**, **Qdrant** "and many more". FAISS is listed by name and could not store a single memory.

## Changes

`insert()` now dispatches on the method's actual named parameters:

```python
params = self._named_params(add_embeddings)

if "text_embeddings" in params:                       # FAISS et al.
    self.client.add_embeddings(text_embeddings=list(zip(texts, vectors)), metadatas=payloads, ids=ids)
elif "texts" in params and "embeddings" in params:    # PGVector et al.
    self.client.add_embeddings(texts=texts, embeddings=vectors, metadatas=payloads, ids=ids)
elif not params or "embeddings" in params:            # unintrospectable / permissive
    self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
else:
    ...  # add_texts
```

Two details worth calling out:

- `_named_params` excludes `*args`/`**kwargs`. Every LangChain `add_embeddings` ends in `**kwargs`, so treating "accepts `**kwargs`" as "accepts anything" would match all 12 and change nothing. Only the explicitly named parameters identify the shape.
- The third branch (empty parameter set) preserves mem0's historical call for objects with no introspectable signature — notably `Mock`, which is what the existing tests use. Without it every current `insert` test would break for the wrong reason.

`update(vector=None)` re-inserts through `add_texts` so the store embeds the payload text itself, rather than passing a `None` vector down.

I deliberately did **not** try to read the existing embedding back and re-store it. I checked: of the 12 stores that accept a precomputed embedding, **zero** expose a way to read one back — so that path would be dead code for every real store.

Nothing else changed. `search`, `list`, `get`, `delete` are untouched.

## Validation

4 new tests in `tests/vector_stores/test_langchain_vector_store.py`:

| test | guards |
|---|---|
| `test_insert_and_payload_only_update_work_against_real_faiss` | end-to-end against a **real** FAISS index — insert, round-trip the metadata, assert the stored vector is the one mem0 computed (`index.reconstruct`), then a payload-only update. `importorskip("faiss")`. |
| `test_insert_passes_text_embedding_pairs_when_that_is_the_signature` | the FAISS/AzureSearch/Elasticsearch/OpenSearch/ScaNN shape |
| `test_insert_passes_texts_alongside_embeddings_when_that_is_the_signature` | the PGVector/Neo4j/Lantern/Hologres/Kinetica/PGEmbedding/Timescale shape |
| `test_update_without_vector_never_stores_none_as_the_embedding` | `vector=None` must not reach the backend as an embedding |

The two signature tests use small local classes rather than `Mock(spec=...)` on purpose — `Mock(spec=f)` does not expose `f`'s signature to `inspect.signature`, so a spec'd mock cannot exercise dispatch at all.

**Control experiment** — `langchain.py` reverted, tests kept:

```
4 failed, 14 passed

FAILED test_insert_and_payload_only_update_work_against_real_faiss
FAILED test_insert_passes_text_embedding_pairs_when_that_is_the_signature
FAILED test_insert_passes_texts_alongside_embeddings_when_that_is_the_signature
FAILED test_update_without_vector_never_stores_none_as_the_embedding
```

Exactly the four new tests fail; all 14 pre-existing tests stay green, including `test_insert_vectors` and `test_update_wraps_vector_and_payload_in_lists`. With the fix: **18 passed**.

**No regressions.** `tests/vector_stores/` + `tests/memory/`:

| tree | result |
|---|---|
| this branch | 732 passed, 13 failed |
| unmodified `main` | 728 passed, 13 failed |

Identical failing sets in both directions (compared as sets, both ways). The 13 are `test_redis.py` (no `redis` module installed) and two `test_faiss.py`/`test_qdrant.py` cases needing local state my machine lacks — all unrelated. The delta is the 4 new tests. 13 further test modules can't even be collected on my machine for missing optional backend SDKs (azure, baidu, cassandra, elasticsearch, oracle, pinecone, supabase, weaviate, …) and were excluded from both runs equally.

`ruff check` and `ruff format` clean on both files at line length 120.

## Note on overlap

Open PR #5961 changes `list()` in this same file (the `[[]]` wrapper shape). No overlap with this diff — I touch `insert()` and `update()` only, and left `list()` exactly as it is on `main` so the two can merge in either order.

## Duplicate check

Searched open and closed PRs and issues for `add_embeddings`, `text_embeddings`, `vector=None`, and `langchain insert` — the only prior report is #3182 (closed as inactive, unfixed). The five closed PRs #4089/#4327/#4343/#4375/#4424 all fixed the *list-wrapping* bug in `update()`, which is already merged and unrelated to this. Targeting `main`.

### DIFF
--- a/mem0/vector_stores/langchain.py
+++ b/mem0/vector_stores/langchain.py
@@ -102,31 +102,46 @@ class Langchain(VectorStoreBase):
 
         if add_embeddings is not None:
             params = self._named_params(add_embeddings)
+            # The id parameter is not spelled the same everywhere: twelve stores
+            # call it ``ids``, AzureSearch calls it ``keys``, and AzureSearch is
+            # also the one store with no ``**kwargs`` to absorb the wrong name.
+            # So the name has to come from the signature too.
+            id_kwargs = self._id_kwargs(params, ids)
 
             # ``text_embeddings`` as (text, vector) pairs: FAISS, AzureSearch,
             # ElasticsearchStore, OpenSearchVectorSearch, ScaNN.
             if "text_embeddings" in params:
-                self.client.add_embeddings(text_embeddings=list(zip(texts, vectors)), metadatas=payloads, ids=ids)
+                self.client.add_embeddings(text_embeddings=list(zip(texts, vectors)), metadatas=payloads, **id_kwargs)
                 return
 
             # ``texts`` + ``embeddings``: PGVector, Neo4jVector, Lantern,
             # Hologres, Kinetica, PGEmbedding, TimescaleVector. ``texts`` is
             # required on all of them, so it must be passed.
             if "texts" in params and "embeddings" in params:
-                self.client.add_embeddings(texts=texts, embeddings=vectors, metadatas=payloads, ids=ids)
+                self.client.add_embeddings(texts=texts, embeddings=vectors, metadatas=payloads, **id_kwargs)
                 return
 
             # Unknown or unintrospectable signature (mocks, permissive wrappers):
-            # keep mem0's historical call shape.
+            # keep mem0's historical call shape. This is a guess about a shape we
+            # could not read, so a rejection here must not be the caller's
+            # problem -- fall through to ``add_texts`` instead of raising the
+            # very TypeError this dispatch exists to prevent.
             if not params or "embeddings" in params:
-                self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
-                return
-
-            logger.debug(
-                "%s.add_embeddings has an unrecognised signature %s; using add_texts instead.",
-                type(self.client).__name__,
-                sorted(params),
-            )
+                try:
+                    self.client.add_embeddings(embeddings=vectors, metadatas=payloads, **id_kwargs)
+                    return
+                except TypeError:
+                    logger.debug(
+                        "%s.add_embeddings rejected mem0's call shape; using add_texts instead.",
+                        type(self.client).__name__,
+                        exc_info=True,
+                    )
+            else:
+                logger.debug(
+                    "%s.add_embeddings has an unrecognised signature %s; using add_texts instead.",
+                    type(self.client).__name__,
+                    sorted(params),
+                )
 
         # Fallback to add_texts method
         self.client.add_texts(texts=texts, metadatas=payloads, ids=ids)
@@ -134,6 +149,26 @@ class Langchain(VectorStoreBase):
     @staticmethod
     def _texts_from(payloads: Optional[List[Dict]], count: int) -> List[str]:
         return [payload.get("data", "") for payload in payloads] if payloads else [""] * count
+
+    @staticmethod
+    def _id_kwargs(params: set, ids: Optional[List[str]]) -> Dict:
+        """The keyword this store spells its id parameter with, if it takes one.
+
+        ``ids`` for twelve of the thirteen ``langchain_community`` stores that
+        define ``add_embeddings``; ``keys`` for AzureSearch. An empty set means
+        the signature could not be read, in which case ``ids`` is the historical
+        call shape and the caller guards the attempt.
+        """
+        if not params or "ids" in params:
+            return {"ids": ids}
+        if "keys" in params:
+            return {"keys": ids}
+        # Takes neither: passing the ids would be rejected by a store without
+        # ``**kwargs``, and silently swallowed by one with it. Dropping them is
+        # visible instead.
+        if ids:
+            logger.debug("add_embeddings takes no id parameter %s; ids are not forwarded.", sorted(params))
+        return {}
 
     @staticmethod
     def _named_params(method) -> set:
--- a/mem0/vector_stores/langchain.py
+++ b/mem0/vector_stores/langchain.py
@@ -1,3 +1,4 @@
+import inspect
 import logging
 from typing import Dict, List, Optional
 
@@ -91,15 +92,63 @@ class Langchain(VectorStoreBase):
     ):
         """
         Insert vectors into the LangChain vectorstore.
+
+        ``add_embeddings`` is not part of the ``VectorStore`` contract and the
+        implementations that do have it disagree on its signature, so dispatch
+        on the actual parameter names rather than assuming one shape.
         """
-        # Check if client has add_embeddings method
-        if hasattr(self.client, "add_embeddings"):
-            # Some LangChain vectorstores have a direct add_embeddings method
-            self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
-        else:
-            # Fallback to add_texts method
-            texts = [payload.get("data", "") for payload in payloads] if payloads else [""] * len(vectors)
-            self.client.add_texts(texts=texts, metadatas=payloads, ids=ids)
+        texts = self._texts_from(payloads, len(vectors))
+        add_embeddings = getattr(self.client, "add_embeddings", None)
+
+        if add_embeddings is not None:
+            params = self._named_params(add_embeddings)
+
+            # ``text_embeddings`` as (text, vector) pairs: FAISS, AzureSearch,
+            # ElasticsearchStore, OpenSearchVectorSearch, ScaNN.
+            if "text_embeddings" in params:
+                self.client.add_embeddings(text_embeddings=list(zip(texts, vectors)), metadatas=payloads, ids=ids)
+                return
+
+            # ``texts`` + ``embeddings``: PGVector, Neo4jVector, Lantern,
+            # Hologres, Kinetica, PGEmbedding, TimescaleVector. ``texts`` is
+            # required on all of them, so it must be passed.
+            if "texts" in params and "embeddings" in params:
+                self.client.add_embeddings(texts=texts, embeddings=vectors, metadatas=payloads, ids=ids)
+                return
+
+            # Unknown or unintrospectable signature (mocks, permissive wrappers):
+            # keep mem0's historical call shape.
+            if not params or "embeddings" in params:
+                self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
+                return
+
+            logger.debug(
+                "%s.add_embeddings has an unrecognised signature %s; using add_texts instead.",
+                type(self.client).__name__,
+                sorted(params),
+            )
+
+        # Fallback to add_texts method
+        self.client.add_texts(texts=texts, metadatas=payloads, ids=ids)
+
+    @staticmethod
+    def _texts_from(payloads: Optional[List[Dict]], count: int) -> List[str]:
+        return [payload.get("data", "") for payload in payloads] if payloads else [""] * count
+
+    @staticmethod
+    def _named_params(method) -> set:
+        """Explicitly named parameters of ``method``; empty if it cannot be introspected.
+
+        ``*args``/``**kwargs`` are excluded -- every LangChain ``add_embeddings``
+        ends in ``**kwargs``, so only the named parameters identify the shape.
+        Mock objects have no usable signature and yield an empty set.
+        """
+        try:
+            parameters = inspect.signature(method).parameters
+        except (TypeError, ValueError):
+            return set()
+        variadic = (inspect.Parameter.VAR_KEYWORD, inspect.Parameter.VAR_POSITIONAL)
+        return {name for name, p in parameters.items() if p.kind not in variadic}
 
     def search(self, query: str, vectors: List[List[float]], top_k: int = 5, filters: Optional[Dict] = None):
         """
@@ -149,9 +198,19 @@ class Langchain(VectorStoreBase):
     def update(self, vector_id, vector=None, payload=None):
         """
         Update a vector and its payload.
+
+        LangChain's ``VectorStore`` has no in-place update, so this is a
+        delete-then-insert. When ``vector`` is omitted -- which is how mem0
+        edits an entity's ``linked_memory_ids`` without re-embedding -- the row
+        is re-inserted from its payload text so the store embeds it itself.
+        Passing the missing vector straight through would store ``None`` where
+        the embedding belongs.
         """
         self.delete(vector_id)
-        self.insert([vector], [payload], [vector_id])
+        if vector is None:
+            self.client.add_texts(texts=self._texts_from([payload], 1), metadatas=[payload], ids=[vector_id])
+        else:
+            self.insert([vector], [payload], [vector_id])
 
     def get(self, vector_id):
         """
--- a/tests/vector_stores/test_langchain_vector_store.py
+++ b/tests/vector_stores/test_langchain_vector_store.py
@@ -371,6 +371,73 @@ def test_insert_passes_texts_alongside_embeddings_when_that_is_the_signature():
     }
 
 
+def test_insert_uses_the_keys_kwarg_when_the_store_spells_it_that_way():
+    """AzureSearch is the one store whose id parameter is ``keys``, not ``ids``.
+
+    Its real signature is ``(text_embeddings, metadatas=None, *, keys=None)`` --
+    keyword-only ``keys``, and no ``**kwargs`` to absorb a wrong name, so passing
+    ``ids`` raised ``TypeError: unexpected keyword argument 'ids'``. The double
+    below is deliberately no more permissive than the real method: the earlier
+    ``text_embeddings`` double grants itself ``ids`` and ``**kwargs``, which is
+    what let this through.
+    """
+
+    class AzureSearchLikeStore:
+        def add_embeddings(self, text_embeddings, metadatas=None, *, keys=None):
+            self.seen = dict(text_embeddings=list(text_embeddings), metadatas=metadatas, keys=keys)
+
+    client = AzureSearchLikeStore()
+    Langchain(client=client, collection_name="c").insert([[0.1, 0.2]], [{"data": "alice"}], ["e1"])
+
+    assert client.seen == {
+        "text_embeddings": [("alice", [0.1, 0.2])],
+        "metadatas": [{"data": "alice"}],
+        "keys": ["e1"],
+    }
+
+
+def test_insert_omits_the_ids_when_add_embeddings_takes_no_id_parameter():
+    """A store that names neither ``ids`` nor ``keys`` must not be sent either."""
+
+    class NoIdParamStore:
+        def add_embeddings(self, text_embeddings, metadatas=None):
+            self.seen = dict(text_embeddings=list(text_embeddings), metadatas=metadatas)
+
+    client = NoIdParamStore()
+    Langchain(client=client, collection_name="c").insert([[0.1, 0.2]], [{"data": "alice"}], ["e1"])
+
+    assert client.seen == {
+        "text_embeddings": [("alice", [0.1, 0.2])],
+        "metadatas": [{"data": "alice"}],
+    }
+
+
+def test_insert_falls_back_to_add_texts_when_the_permissive_call_is_rejected():
+    """The permissive branch is a guess about a signature we could not read.
+
+    It is reached by any store exposing ``embeddings`` without ``texts``, and by
+    anything unintrospectable. Letting its TypeError escape would reproduce the
+    exact crash this dispatch exists to prevent, so a rejection there falls
+    through to ``add_texts`` rather than reaching the caller.
+    """
+
+    class PermissiveButRejectingStore:
+        def add_embeddings(self, embeddings, metadatas=None, ids=None):
+            raise TypeError("simulated store rejection")
+
+        def add_texts(self, texts, metadatas=None, ids=None, **kwargs):
+            self.seen = dict(texts=list(texts), metadatas=metadatas, ids=ids)
+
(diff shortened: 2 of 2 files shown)


LHMQ878:

LHMQ878:
The red `Vercel` check on this PR is not a test failure — it is a deploy authorization gate:

```
GET /repos/mem0ai/mem0/commits/9713718a/status
  failure  Vercel      "Authorization required to deploy."
  success  license/cla "Contributor License Agreement is signed."
```

Vercel asks a repo member to authorize deploying a fork's branch, which is a maintainer-side click, not something the branch can change. This PR touches only `mem0/vector_stores/langchain.py` and `tests/vector_stores/test_langchain_vector_store.py` — no docs site, no app — so there is nothing here for that deploy to build differently.

CLA is signed. Flagging it so the red tick isn't read as a broken change.

mem0/vector_stores/langchain.py
```diff
@@ -91,15 +92,63 @@ def insert(
...
+        texts = self._texts_from(payloads, len(vectors))
+        add_embeddings = getattr(self.client, "add_embeddings", None)
+
+        if add_embeddings is not None:
+            params = self._named_params(add_embeddings)
+
+            # ``text_embeddings`` as (text, vector) pairs: FAISS, AzureSearch,
+            # ElasticsearchStore, OpenSearchVectorSearch, ScaNN.
+            if "text_embeddings" in params:
+                self.client.add_embeddings(text_embeddings=list(zip(texts, vectors)), metadatas=payloads, ids=ids)
```
kartik-mem0: Real `AzureSearch.add_embeddings(text_embeddings, metadatas=None, *, keys=None)` has no `ids` param or `**kwargs`; this branch still `TypeError`s for AzureSearch, one of the five backends targeted here. No test catches it — the `TextEmbeddingsStore` double (test file:340) gives itself `ids`/`**kwargs` real AzureSearch doesn't have.

mem0/vector_stores/langchain.py
```diff
@@ -91,15 +92,63 @@ def insert(
...
+            # Hologres, Kinetica, PGEmbedding, TimescaleVector. ``texts`` is
+            # required on all of them, so it must be passed.
+            if "texts" in params and "embeddings" in params:
+                self.client.add_embeddings(texts=texts, embeddings=vectors, metadatas=payloads, ids=ids)
+                return
+
+            # Unknown or unintrospectable signature (mocks, permissive wrappers):
+            # keep mem0's historical call shape.
+            if not params or "embeddings" in params:
+                self.client.add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)
```
kartik-mem0: This permissive fallback calls `add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)` with no try/except. Any future or unrecognized `langchain_community` store lands here and reproduces the exact TypeError this PR exists to fix, silently, with no fallback to `add_texts`.

kartik-mem0 (commented):
## Changes requested

The introspection-based dispatch is the right fix and correctly covers 12 of the 13 real `langchain_community` stores that define `add_embeddings`, but real AzureSearch (one of the five backends this PR's own docstring names) still crashes, untested and unguarded.

**To unblock:**
1. `mem0/vector_stores/langchain.py:109` — real AzureSearch has no `ids` kwarg; branch still crashes on it (detail in inline comment)
2. `mem0/vector_stores/langchain.py:122` — permissive fallback has no try/except; repeats the original crash for any unrecognized store shape (detail in inline comment)

Optional: add a test using AzureSearch's real signature (`text_embeddings, metadatas=None, *, keys=None`, no `ids`, no `**kwargs`) alongside the current, more permissive synthetic double.

- **Problem is real:** reproduces on `main` at `mem0/vector_stores/langchain.py:96-98` (`add_embeddings(embeddings=..., metadatas=..., ids=...)` against real FAISS raises `TypeError: FAISS.add_embeddings() missing 1 required positional argument: 'text_embeddings'`); `update(vector=None)` at `main:149-154` forwards `None` as the embedding after `delete()` already ran, a genuine data-loss path, not just a crash.
- **Tests / red-green:** `hatch -e dev_py_3_12 run pytest tests/vector_stores/test_langchain_vector_store.py -v` → 18 passed with the fix applied. Reverting only `langchain.py` to `main` (PR's test file kept) → 4 failed with the exact predicted TypeErrors/AssertionError, 14 passed. Restoring the fix → 18 passed again. Reproduced the AzureSearch gap directly by calling the PR's `insert()` against a fake object built with AzureSearch's real signature: raises `TypeError: unexpected keyword argument 'ids'`.
- **CI:** only Vercel (fails, fork lacks deploy perms, non-blocking), `label`, and `license/cla` have run on this SHA; no CI Gate run yet (external-contributor first-PR approval gate, not a failure).
- **Security:** cleared, no new attack surface — no user input path, query construction, or deserialization changed.
- **Bloat (ponytail):** lean already; the fix is same-file, introduces no new abstractions.
- **Not verified:** could not run this end-to-end in the sandbox — no lane instantiates the LangChain wrapper (all 7 lanes in `sandbox/adapter.py::_vector_store_config` build native provider configs, never `provider: langchain`), so the AzureSearch-signature and fallback findings are verified via direct signature inspection and unit reproduction against a fake client, not a live sandbox run. PR #5961 also touches this file but only its unrelated `list()` method (disjoint line ranges, no overlap, both independently mergeable).

</details>

Root-cause approach is right; this is a same-file follow-up, not a redesign. Push when ready and I will re-review.

LHMQ878:
Both findings fixed in `d472484`, and both were right. Thank you especially for pointing at the test double — that is why the first one got through, and it is the part I would not have found on my own.

**Finding 1 — AzureSearch.** I surveyed every store in the installed `langchain_community` that defines `add_embeddings` rather than patching the one you named, because the interesting question was whether `keys` is a one-off or a second convention:

```
store                    named params                                  **kwargs
FAISS                    text_embeddings,metadatas,ids                 yes
AzureSearch              text_embeddings,metadatas,keys                NO
ElasticsearchStore       text_embeddings,metadatas,ids,...             yes
OpenSearchVectorSearch   text_embeddings,metadatas,ids,bulk_size       yes
ScaNN                    text_embeddings,metadatas,ids                 yes
PGVector / Neo4jVector / Lantern / Hologres / Kinetica /
PGEmbedding / TimescaleVector
                         texts,embeddings,metadatas,ids                yes
```

Twelve say `ids`, AzureSearch says `keys`, and AzureSearch is the only one of the thirteen with no `**kwargs` — which is exactly why it is the only one that raises. So the id keyword now comes from the signature like the rest of the shape rather than being hardcoded, and a store naming neither gets no ids at all (logged) instead of a kwarg that a permissive store would silently swallow.

You are also right about the double. The existing `text_embeddings` test declares `(text_embeddings, metadatas=None, ids=None, **kwargs)` — it grants itself both the `ids` and the `**kwargs` real AzureSearch lacks, so it could not fail. The new test uses the real shape, keyword-only `keys` and no `**kwargs`.

**Finding 2 — the permissive fallback.** Guarded with `try/except TypeError` falling through to `add_texts`, which is what the unrecognised-signature branch beside it already did. Your framing is what convinced me it belongs there: that branch is a *guess* about a signature we could not read, so its failure is not the caller's problem.

**Red/green, with only `langchain.py` reverted and the new tests kept:**

```
3 failed / 18 passed      all three TypeError, one per new test
21 passed                 with the fix
```

The three new tests are AzureSearch's real signature, a store naming no id parameter at all, and a permissive store that rejects the call.

I checked the routing rather than just that nothing raised, since "no exception" would also be true if the ids were being dropped:

```
FAISS-like              add_embeddings(ids=['e1'])
AzureSearch-like        add_embeddings(keys=['e1'])
permissive + rejects    add_texts(ids=['e1'], texts=['alice'])
no id parameter         add_embeddings(...)            ids dropped, logged
unintrospectable mock   add_embeddings(ids=['e1'])     historical shape kept
```

`ruff check` clean. `ruff format --check` reports two hunks in these files, both pre-existing and outside this change — a comment-alignment block in `langchain.py` and a trailing comma in a `search` test — so I left them rather than mixing unrelated reformatting into the diff; say the word if you would rather I normalise both files.

One note on your "not verified" item, since it applies to my side too: I also cannot exercise this through a lane, so my evidence is the real installed signatures plus unit reproduction, not a live backend. The AzureSearch path in particular is verified against `inspect.signature(AzureSearch.add_embeddings)` from the installed package and a double built to match it exactly — not against a running Azure Search instance.

mem0/vector_stores/langchain.py
kartik-mem0: Verified against d472484: fixed. `_id_kwargs()` (mem0/vector_stores/langchain.py:154-171) reads the signature and returns `{"keys": ids}` when `keys` is a param and `ids` is not, so the `text_embeddings` branch now calls AzureSearch with `keys=`, matching its real `(text_embeddings, metadatas=None, *, keys=None)` signature. New test `test_insert_uses_the_keys_kwarg_when_the_store_spells_it_that_way` (tests/vector_stores/test_langchain_vector_store.py:374) uses a double with the exact real signature, keyword-only `keys`, no `**kwargs`, and asserts `keys=["e1"]` was passed, not `ids=`.

mem0/vector_stores/langchain.py
kartik-mem0: Verified against d472484: fixed. The permissive fallback (mem0/vector_stores/langchain.py:129-138) now wraps the `add_embeddings(embeddings=vectors, ...)` call in try/except TypeError, logs at debug, and falls through to `add_texts` instead of letting the TypeError reach the caller. New test `test_insert_falls_back_to_add_texts_when_the_permissive_call_is_rejected` (tests/vector_stores/test_langchain_vector_store.py:415) uses a store whose `add_embeddings` always raises `TypeError("simulated store rejection")` and asserts the call lands on `add_texts` instead.

mem0/vector_stores/langchain.py
```diff
@@ -149,9 +233,19 @@ def delete(self, vector_id):
...
+        delete-then-insert. When ``vector`` is omitted -- which is how mem0
+        edits an entity's ``linked_memory_ids`` without re-embedding -- the row
+        is re-inserted from its payload text so the store embeds it itself.
+        Passing the missing vector straight through would store ``None`` where
+        the embedding belongs.
         """
         self.delete(vector_id)
-        self.insert([vector], [payload], [vector_id])
+        if vector is None:
+            self.client.add_texts(texts=self._texts_from([payload], 1), metadatas=[payload], ids=[vector_id])
```
kartik-mem0, line 246: update(vector_id) with vector=None and payload=None crashes in _texts_from (payload.get on None). No current caller passes payload=None: mem0/memory/main.py always builds a dict before calling update(). Pre-existing edge case, not a regression from this PR. Minor, consider payload = payload or {} for defense in depth.

mem0/vector_stores/langchain.py
```diff
@@ -91,15 +92,98 @@ def insert(
...
+            # Unknown or unintrospectable signature (mocks, permissive wrappers):
+            # keep mem0's historical call shape. This is a guess about a shape we
+            # could not read, so a rejection here must not be the caller's
+            # problem -- fall through to ``add_texts`` instead of raising the
+            # very TypeError this dispatch exists to prevent.
+            if not params or "embeddings" in params:
+                try:
+                    self.client.add_embeddings(embeddings=vectors, metadatas=payloads, **id_kwargs)
+                    return
+                except TypeError:
```
kartik-mem0, line 133: This except TypeError also catches a TypeError raised inside a correctly-shaped add_embeddings call, e.g. a bug in the store's own code, silently falling back to add_texts instead of surfacing it. Logged at debug with exc_info, so discoverable, but worth a narrower catch if this bites in practice.

mem0/vector_stores/langchain.py
```diff
@@ -91,15 +92,98 @@ def insert(
...
+    @staticmethod
+    def _named_params(method) -> set:
+        """Explicitly named parameters of ``method``; empty if it cannot be introspected.
+
+        ``*args``/``**kwargs`` are excluded -- every LangChain ``add_embeddings``
+        ends in ``**kwargs``, so only the named parameters identify the shape.
+        Mock objects have no usable signature and yield an empty set.
+        """
+        try:
+            parameters = inspect.signature(method).parameters
```
kartik-mem0, line 182: inspect.signature(method) runs on every insert() call rather than being cached per client class. Negligible at typical batch sizes, only worth caching if insert() becomes a hot per-item path.

kartik-mem0 (commented):

Verified against `d472484` (new head): both prior findings are fixed. AzureSearch now gets `keys=` via `_id_kwargs()`, and the permissive fallback is wrapped in try/except TypeError so a rejection falls through to `add_texts` instead of reaching the caller.

**Follow-up:** #6768 fixes the same root cause but hardcodes `ids=` with no `keys=` dispatch and no try/except guard, so it still TypeErrors on AzureSearch. Recommend consolidating into this PR; see the duplicate-cluster note below.
**Not verified:** live LangChain community stores beyond FAISS (AzureSearch, PGVector, etc. are exercised via signature-matching doubles, not the real packages).

Risk: low, additive dispatch with a fallback path and full revert-tested coverage · Your time: ~5 min

- **Problem is real:** reproduces on `main` at `mem0/vector_stores/langchain.py` (pre-PR `insert()`), confirmed in the prior review cycle and unchanged since.
- **Tests:** `python -m pytest tests/vector_stores/test_langchain_vector_store.py -v` -> 21 passed.
- **Red-green:** reverted the dispatch to the pre-PR unconditional `add_embeddings(embeddings=..., ids=...)` call; the 3 new tests (AzureSearch-shaped, no-id-param, rejected-call fallback) failed with the exact `TypeError`s the PR fixes; restored, all 21 pass again.
- **CI:** green (lint, tests, build); no required check red.
- **Security:** cleared, no new external input, `add_embeddings`/`add_texts` calls stay within the existing trust boundary, signature introspection is read-only.
- **Bloat (ponytail):** lean already, the three static helpers (`_texts_from`, `_id_kwargs`, `_named_params`) are each used from more than one call site so they earn their keep.
- **Linked issue:** #6657.

</details>

Duplicate cluster: #6768 targets the same bug (AzureSearch/FAISS `add_embeddings` signature mismatch) but its `insert()` still calls `add_embeddings(..., ids=ids)` unconditionally with no `keys=` branch, so it TypeErrors on real AzureSearch, and it has no try/except around any `add_embeddings` call, so a rejected call propagates uncaught. This PR fixes both plus has broader coverage (real FAISS end-to-end, AzureSearch-shaped test, no-id-param test). Recommend closing #6768 in favor of this one; that is a call for a maintainer, not something this review does.

LHMQ878:
Thanks for re-checking against d472484.

On the three non-blocking notes: agreed on all of them. I'd rather leave the broad TypeError catch and the per-call signature inspect alone for this PR since they match the failure mode we actually hit, and the payload=None guard is pre-existing with no caller today — happy to take those in a follow-up if they start biting. #6768 still looks like the one to fold in or close as duplicate whenever you're ready to merge.

PR #6659, opened 2026-07-29
fix(memory): pick the fenced block that parses in extract_json, not the first one

ErenAta16:
## The problem

`extract_json` takes the first ``` block it finds:

```python
match = re.search(r"```(?:json)?\s*(.*?)\s*```", text, re.DOTALL)
if match:
    json_str = match.group(1)
```

Reasoning-style models routinely put scratch work in a code block before the answer, so the first block is the reasoning and that is what comes back:

```
plain json                       -> parses             got='{"facts": ["a"]}'
```json fenced                   -> parses             got='{"facts": ["a"]}'
``` fenced (no tag)              -> parses             got='{"facts": ["a"]}'
prose then json                  -> parses             got='{"facts": ["a"]}'
reasoning block then json        -> FAILS              got='step 1\nstep 2'
reasoning block then json block  -> FAILS              got='notes'
```

The caller's `json.loads` then fails on text that was never meant to be JSON. The brace fallback below would have found the real object, but it is only reached when no fence matched at all.

Callers are the tool-call argument parsers in `azure_openai`, `azure_openai_structured`, `lmstudio` and `xai`, plus the fact-extraction path.

## The change

Collect the candidates, strongest signal first, and return the first one that actually parses:

1. ` ```json `-tagged blocks
2. any fenced block
3. the first-brace-to-last-brace span

When nothing parses, return what the old code would have returned, so the caller's own `json.loads` still raises with the same content in the message.

Nothing that worked before changes: a lone valid block is both the first candidate and the first that parses.

## How this sits with the other open PRs on this function

There are three, and they are aimed at different inputs. Reading the diffs, none of them changes which fenced block is chosen:

- **#6426** skips the fence heuristic when the text already starts with `{`, so a ``` sequence *inside* a JSON string value is not mistaken for a fence. Orthogonal to this.
- **#6043** adds balanced-brace scanning with payload-key preference, but it sits in the `else` branch, so it only runs when no fence matched. With a reasoning block present, `match` is truthy and that code is never reached.
- **#6050** is also on the no-fence path.

So this one is complementary rather than competing, and the four together would want sequencing rather than picking one. Happy to rebase onto whichever lands first, or to fold this into one of them if a maintainer would rather have a single change to this function.

## Tests

Added `TestExtractJsonWithReasoningBlocks` to `tests/test_chatty_llm_parsing.py`:

- reasoning code block, then inline JSON
- reasoning code block, then a ` ```json ` block
- a ` ```python ` block before a ` ```json ` block, so the tag preference is pinned and not just "last block wins"
- nothing parseable, asserting the old return value is preserved so error messages do not change

Reverting only `mem0/memory/utils.py`:

```
FAILED tests/test_chatty_llm_parsing.py::TestExtractJsonWithReasoningBlocks::test_reasoning_code_block_before_inline_json
FAILED tests/test_chatty_llm_parsing.py::TestExtractJsonWithReasoningBlocks::test_reasoning_code_block_before_json_code_block
FAILED tests/test_chatty_llm_parsing.py::TestExtractJsonWithReasoningBlocks::test_json_tagged_block_wins_over_an_earlier_untagged_one
3 failed, 22 passed
```

and with the change the whole file passes, including the 22 pre-existing cases:

```
$ pytest tests/test_chatty_llm_parsing.py
25 passed
```

The fourth new test passes either way on purpose: it is there so the fallback cannot be dropped later.


ebarkhordar (commented):
The candidate loop returns the first block that `json.loads` accepts, and `json.loads` accepts a bare scalar, so a scratch block whose contents happen to be a number, string, boolean or null still beats the real object.

At `91f1f5a4`, in a clean container:

```
A. reasoning block is prose  -> extract_json: '{"memory": [{"text": "likes basketball"}]}'
   caller: [{'text': 'likes basketball'}]

B. scratch block = 42       -> extract_json: '42'
   caller: AttributeError: 'int' object has no attribute 'get'
B. scratch block = "done"   -> extract_json: '"done"'
   caller: AttributeError: 'str' object has no attribute 'get'
B. scratch block = true     -> extract_json: 'true'
   caller: AttributeError: 'bool' object has no attribute 'get'
```

`caller` there is `mem0/memory/main.py:951-952`, `json.loads(extract_json(response), strict=False).get("memory", [])`. The `AttributeError` is swallowed by the `except Exception` at :953, so the run logs an error and extracts nothing.

`tests/test_chatty_llm_parsing.py` passes at that head (25 tests). This is a case they do not cover rather than a regression: the old first-fence rule failed on the same input.

Every caller indexes the result as an object, so gating on the parsed type would close it:

```python
try:
    parsed = json.loads(candidate)
except ValueError:
    continue
if isinstance(parsed, dict):
    return candidate
```

Separately, #6426 is open against the same function for the fence-inside-a-string-value case. I checked its input against this branch and it already round-trips correctly, so you may not need both.

ErenAta16:
You are right, and this is a real hole in my patch rather than a nit. Fixed in `92bb1d1`, tests in `94b84a3`.

Reproduced your three cases and the pattern generalises to every JSON scalar plus arrays:

```
scratch=42       extract_json -> '42'       caller: AttributeError: 'int' object has no attribute 'get'
scratch="done"   extract_json -> '"done"'   caller: AttributeError: 'str' object has no attribute 'get'
scratch=true     extract_json -> 'true'     caller: AttributeError: 'bool' object has no attribute 'get'
scratch=null     extract_json -> 'null'     caller: AttributeError: 'NoneType' ...
scratch=[1, 2]   extract_json -> '[1, 2]'   caller: AttributeError: 'list' object has no attribute 'get'
```

The gate is exactly what you proposed:

```python
try:
    parsed = json.loads(candidate)
except ValueError:
    continue
if isinstance(parsed, dict):
    return candidate
```

**One precondition worth recording, because it explains why my own tests missed it.** The scratch block only wins when the real answer is *untagged*:

```
scratch untagged, answer ```json-tagged   -> answer wins    (bug does not appear)
scratch untagged, answer untagged         -> scratch wins   (bug)
scratch ```json-tagged, answer untagged   -> scratch wins   (bug)
```

The candidate list puts explicitly tagged fences first, so a ```json-tagged answer is already ordered ahead of any untagged scratch and parses before it. Every fixture I wrote tagged the answer, which is the tidy shape rather than the realistic one. That is the actual gap in the original test suite, and it is why "25 tests pass" did not mean much here.

The new tests parametrize over `42`, `"done"`, `true`, `null`, `[1, 2]`, `-0.5` with both fences untagged, plus the json-tagged-scalar variant, plus the original prose case so the thing this PR set out to fix stays covered, plus a scalar-only response to pin the fallback:

```
scratch=42 / "done" / true / null / [1,2] / -0.5   extract=ok  caller=ok
json-tagged scalar                                 PASS
prose scratch (the original fix)                   PASS
scalar-only fallback                               PASS
```

The last one keeps the old behaviour deliberately: with no object anywhere, `extract_json("```\n42\n```")` still returns `"42"` so the caller's own `json.loads` raises on its own terms rather than us inventing an empty dict.

On #6426, thanks for checking it against this branch rather than assuming. I have not re-run its input myself so I will not claim it either way; if a maintainer wants them merged into one change I am happy to fold it in, otherwise leaving them separate seems cleaner since the fixes are unrelated.

Genuinely good catch. "It parses" was the wrong predicate and I should have asked what the callers actually do with the result.

PR #6662, opened 2026-07-29
fix(memory): prevent identity metadata injection

### LINKED ISSUE #6655: Python OSS SDK: add() lets metadata set identity scope (user_id/agent_id/run_id/actor_id) on creation

tomatotomata:
## Summary

- prevent free-form metadata from injecting `user_id`, `agent_id`, `run_id`, or `actor_id` during memory creation
- retain explicit top-level identity arguments as the only source of scope
- add regression coverage for the creation helper

Closes #6655

## Validation

- `python -m pytest tests/memory/test_main.py -q -k add_metadata_cannot_inject_identity_fields` (1 passed)
- `git diff --check`

The full `test_main.py` module could not run because this environment lacks the repository's `pytest-mock` fixture plugin; 48 existing tests report the missing `mocker` fixture before executing.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -342,7 +342,14 @@ def _build_filters_and_metadata(
               scoped to the provided session(s) and potentially a resolved actor.
     """
 
-    base_metadata_template = deepcopy(input_metadata) if input_metadata else {}
+    # Identity scope is controlled by the explicit arguments below. Do not
+    # allow free-form metadata to inject a scope that was not supplied by the
+    # caller (the update path applies the same invariant after creation).
+    base_metadata_template = {
+        key: value
+        for key, value in (input_metadata or {}).items()
+        if key not in _IDENTITY_KEYS
+    }
     effective_query_filters = deepcopy(input_filters) if input_filters else {}
 
     # ---------- validate and add all provided session ids ----------
--- a/tests/memory/test_main.py
+++ b/tests/memory/test_main.py
@@ -7,7 +7,7 @@ from unittest.mock import MagicMock, Mock
 import pytest
 
 from mem0.exceptions import LLMError
-from mem0.memory.main import AsyncMemory, Memory
+from mem0.memory.main import AsyncMemory, Memory, _build_filters_and_metadata
 
 
 def _setup_mocks(mocker):
@@ -451,6 +451,21 @@ def test_update_memory_metadata_cannot_change_identity_fields(mocker, caplog):
     assert payload["category"] == "sports"
     assert payload["data"] == "new memory"
     assert "ignoring metadata['user_id']" in caplog.text
+
+
+def test_add_metadata_cannot_inject_identity_fields():
+    metadata, filters = _build_filters_and_metadata(
+        user_id="tenant_a",
+        input_metadata={
+            "agent_id": "attacker_agent",
+            "run_id": "attacker_run",
+            "actor_id": "attacker_actor",
+            "category": "sports",
+        },
+    )
+
+    assert metadata == {"user_id": "tenant_a", "category": "sports"}
+    assert filters == {"user_id": "tenant_a"}
 
 
 @pytest.mark.asyncio


mem0/memory/main.py
```diff
@@ -342,7 +342,14 @@ def _build_filters_and_metadata(
...
 
-    base_metadata_template = deepcopy(input_metadata) if input_metadata else {}
+    # Identity scope is controlled by the explicit arguments below. Do not
+    # allow free-form metadata to inject a scope that was not supplied by the
+    # caller (the update path applies the same invariant after creation).
+    base_metadata_template = {
+        key: value
+        for key, value in (input_metadata or {}).items()
+        if key not in _IDENTITY_KEYS
+    }
```
kartik-mem0, line 352: This re-implements an invariant the module already centralizes. `_strip_identity_keys()` at `mem0/memory/main.py:142` strips exactly `_IDENTITY_KEYS`, and both update paths already call it (`:2025`, `:3708`). Delegating to it here keeps the rule in one place, and gets you the `update()` warning for free so a caller whose metadata is silently rewritten actually finds out:

```python
base_metadata_template = (
    _strip_identity_keys(deepcopy(input_metadata), {}, context="add()") if input_metadata else {}
)
```

Two smaller things, if this version is the one that lands:

- The comprehension drops the `deepcopy`. It copies keys but not values, so nested values in the returned template are the caller's own objects rather than copies. I traced the consumers and nothing observes it today (every downstream mutation site re-deepcopies first: `:866`, `:1000`, `:1935`, `:2495`), so this is hygiene rather than a bug. `{key: deepcopy(value) for ...}` restores it.
- Repo convention is no code comments, and `_IDENTITY_KEYS` already names the intent.

tests/memory/test_main.py
```diff
@@ -453,6 +453,21 @@ def test_update_memory_metadata_cannot_change_identity_fields(mocker, caplog):
...
+        input_metadata={
+            "agent_id": "attacker_agent",
+            "run_id": "attacker_run",
+            "actor_id": "attacker_actor",
+            "category": "sports",
+        },
+    )
+
+    assert metadata == {"user_id": "tenant_a", "category": "sports"}
+    assert filters == {"user_id": "tenant_a"}
```
kartik-mem0, line 468: This asserts the mechanism rather than the reported failure. It calls the private `_build_filters_and_metadata` and checks its return dict, so it never goes through `Memory.add()`, never looks at the payload handed to the vector store, and never asserts the memory is unreachable via `get_all(filters={"agent_id": ...})`, which is the Actual Behavior described in #6655. A refactor that changed how `processed_metadata` is threaded through `add()` could reopen the hole with this test still green.

The file already has the closer pattern: `test_update_memory_metadata_cannot_change_identity_fields` (`:437` on main) drives `_update_memory` and asserts `memory.vector_store.update.call_args.kwargs["payload"]`, and `_build_memory_instance` (`:317`) hands you the instance to do the same for `add()`.

Related and minor: no async counterpart, while every other identity-field regression test here is duplicated sync/async (the async twin sits directly below this one). It matters less than usual because both `add()` paths share this single free function, but `AsyncMemory.add()` has its own body and its own wiring to that helper.

kartik-mem0 (commented):
## Duplicate, recommend consolidating into #6656

Both PRs fix the same root cause the same way: strip identity keys out of caller metadata in `_build_filters_and_metadata` before it becomes the stored payload. The vulnerability is real and this fix does close it, confirmed live on two backends.

**Recommend:** squash-merge #6656, close this one.
**Why #6656:** three things it does that this one does not.
1. Reuses the existing `_strip_identity_keys()` helper (`mem0/memory/main.py:142`) instead of standing up a second inline implementation of one invariant.
2. Keeps the warning `update()` already emits on a dropped identity key, so `add()` gains the same signal instead of rewriting the caller's metadata silently.
3. Tests the public `Memory.add()` and `AsyncMemory.add()` paths; this PR's single test asserts the private helper's return dict only.

I checked specifically whether this PR covers anything #6656 misses, and it does not: the extra `assert filters == ...` exercises `input_filters`, a parameter neither PR touches.
**Your time:** ~3 min

<details><summary>How I verified both</summary>

**This PR (`4f2f3595b`)**

- **Gate 0, the bug is real.** Traced on `main` before reading the fix: `mem0/memory/main.py:346` deepcopies caller metadata wholesale into the template that becomes the stored payload, and `:357-370` re-pins identity keys only when the matching kwarg is truthy. `actor_id` is never re-pinned into the template at all (`:381-383`). So `add(user_id="attacker", metadata={"agent_id": "victim"})` writes a memory into a scope the caller does not own. That is a cross-scope *write*, not just a read leak.
- **Red-green.** The new test passes on this branch, FAILS with `mem0/memory/main.py` reverted to `main` (`Left contains 3 more items: {'actor_id': 'attacker_actor', 'agent_id': 'attacker_agent', 'run_id': 'attacker_run'}`), and passes again on restore. `ruff check` clean.
- **Live sandbox, both lanes against this branch.** `oss-qdrant` PASS (2.0s), `oss-pgvector` PASS (1.8s). The same scenario against unfixed `main` FAILs: `memory f57d6bf2... leaked into agent_id=...-victim-agent via add() metadata`. So the fix itself genuinely closes the hole end to end, on sync and async, whatever happens to the test.
- **CI.** The red check is the Vercel preview deploy, which forks cannot run. Ignored, not merge-blocking. The real test/lint/build checks pass.

**#6656 (`13dbce71`, OPEN and MERGEABLE)**, read from its diff rather than its description:

- `_build_filters_and_metadata` delegates to `_strip_identity_keys(deepcopy(input_metadata), {}, context="add()")`, so the deepcopy survives and the key set stays in exactly one place.
- 4 tests: sync and async through the real `add()` with only `_add_to_vector_store` mocked, capturing the metadata actually passed downstream plus a `caplog` assertion on the warning, and two edge cases (entity params still set scope, no-metadata unaffected).
- Its new `context` kwarg is keyword-only with a backward-compatible default and has no callers outside `mem0/memory/main.py`, so it breaks nothing.

**Not verified**

- Neither PR touches the TypeScript SDK. `mem0-ts/src/oss/src/memory/index.ts:743` has the identical hole on `main`, tracked separately as #6371 with PR #6377. Consolidating here does not close that one.

</details>

Real work here, and the diagnosis is exactly right: you found the same root cause independently and put the guard at the one chokepoint both `add()` paths funnel through, which is the correct layer for it. It is only the overlap with an earlier PR that decides this. Please reopen if #6656 falls through.

PR #6684, opened 2026-07-30
fix: wire dev tooling (ruff/isort/pre-commit) into hatch dev environments

### LINKED ISSUE #6683: Python SDK: documented dev setup fails — ruff/isort/pre-commit not installed in hatch dev environments
### Summary

Following CONTRIBUTING.md's documented Python SDK setup steps exactly, on a
fresh clone, fails at 4 of the 5 steps: `pre-commit install`, `make lint`,
`make format`, and `make sort` all fail with "command not found." Root cause
is that `ruff`, `isort`, and `pre-commit` are never actually installed into
any of the hatch environments the guide tells you to activate.

### Steps to Reproduce

```bash
git clone https://github.com/mem0ai/mem0.git
cd mem0
pip install hatch   # or pipx install hatch

hatch shell dev_py_3_11
pre-commit install
make lint
make format
make sort
```

### Expected Behavior
Per CONTRIBUTING.md ("Contributing to the Python SDK"), all five commands
should succeed inside the activated dev_py_3_11 shell:

```bash
hatch shell dev_py_3_11
pre-commit install   # Install pre-commit hooks (runs ruff + isort on commit)
make lint
make format
make sort
```

### Actual Behaviour
```bash
$ pre-commit install
zsh: command not found: pre-commit

$ make lint
hatch run lint
/bin/sh: ruff: command not found
make: *** [lint] Error 127

$ make format
hatch run format
/bin/sh: ruff: command not found
make: *** [format] Error 127

$ make sort
hatch run isort mem0/
/bin/sh: isort: command not found
make: *** [sort] Error 127
```

Root cause: pyproject.toml declares ruff, isort, and pytest in a
dev optional-dependency group, but no hatch environment references dev
in its features list:

dev = [
    "ruff==0.16.0",
    "isort>=5.13.2",
    "pytest>=8.2.2",
]

[tool.hatch.envs.dev_py_3_11]
python = "3.11"
features = [
  "test",
  "vector-stores",
  "llms",
  "extras",
]

in every environment. Separately, pre-commit itself is never declared as a
dependency anywhere in the project, so the pre-commit CLI referenced by
CONTRIBUTING.md's own instructions isn't installed by any documented step.

(Note: make test is not affected by this bug — it works correctly once
dev_py_3_11 is activated, since the test feature group already includes
pytest.)

Environment
mem0 version: 2.0.14 (pyproject.toml)
Python version: 3.11 (also affects 3.10, 3.12 environments identically)
OS: macOS (Darwin), should reproduce on any OS since it's a pure hatch environment/dependency-declaration issue, not platform-specific

ratishjain12:
## Linked Issue

Closes #6683 

## Description

Following CONTRIBUTING.md's documented Python SDK setup exactly —
`hatch shell dev_py_3_11` → `pre-commit install` → `make lint` / `make
format` / `make sort` — fails on a fresh clone:

- `pre-commit install` → `pre-commit: command not found`
- `make lint` / `make format` → `ruff: command not found`
- `make sort` → `isort: command not found`

Root cause: `ruff`, `isort`, and `pytest` are declared in the `dev` optional-
dependency group in `pyproject.toml`, but no hatch environment references
`dev` in its `features` list — `dev_py_3_10/3_11/3_12` only pull `test`,
`vector-stores`, `llms`, `extras`. `pre-commit` itself isn't declared as a
dependency anywhere in the project.

This PR:
- Adds `pre-commit>=3.5.0` to the `dev` optional-dependency group.
- Adds `"dev"` to the `features` list of `dev_py_3_10`, `dev_py_3_11`, and
  `dev_py_3_12`, so the documented setup flow works end-to-end without any
  manual, undocumented `pip install` step.

No Makefile changes needed — `hatch run <script>` already resolves against
whichever environment is currently active.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

Manually verified inside a freshly recreated `dev_py_3_11` hatch environment:
- `which ruff isort pre-commit pytest` all resolve inside the env
- `make lint` → `All checks passed!`
- `pre-commit install` → succeeds
- `make test` (dev_py_3_11) → 1767 passed, 73 skipped (pre-existing, unrelated
  failures excluded — see #<issue-number> if filed separately)

This is a `pyproject.toml`-only config change with no code paths to unit
test.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,7 @@ dev = [
     "ruff==0.16.0",
     "isort>=5.13.2",
     "pytest>=8.2.2",
+    "pre-commit>=3.5.0",
 ]
 
 [tool.pytest.ini_options]
@@ -109,6 +110,7 @@ features = [
   "vector-stores",
   "llms",
   "extras",
+  "dev",
 ]
 
 [tool.hatch.envs.dev_py_3_11]
@@ -118,6 +120,7 @@ features = [
   "vector-stores",
   "llms",
   "extras",
+  "dev",
 ]
 
 [tool.hatch.envs.dev_py_3_12]
@@ -127,6 +130,7 @@ features = [
   "vector-stores",
   "llms",
   "extras",
+  "dev",
 ]
 
 [tool.hatch.envs.default.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,6 @@ test = [
 dev = [
     "ruff==0.16.0",
     "isort>=5.13.2",
-    "pytest>=8.2.2",
     "pre-commit>=3.5.0",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,11 @@ features = [
   "dev",
 ]
 
+[tool.hatch.envs.default]
+features = [
+  "dev",
+]
+
 [tool.hatch.envs.default.scripts]
 format = [
     "ruff format",


ratishjain12:
@kartik-mem0 Just following up on this PR. Since it addresses the linked issue, wanted to check if there’s anything else needed from my side. Happy to make any changes

pyproject.toml
```diff
@@ -127,6 +130,7 @@ features = [
   "vector-stores",
   "llms",
   "extras",
+  "dev",
```
kartik-mem0, line 132: `default` is the one env this leaves broken, and it is the env that *owns* these scripts (`[tool.hatch.envs.default.scripts]`, right below) and the one `make install` creates. With this PR applied, `HATCH_ENV_ACTIVE=default make lint` still gives `ruff: command not found`. Adding this above line 136 fixes it (verified: `All checks passed!`):

```toml
[tool.hatch.envs.default]
features = [
  "dev",
]
```

kartik-mem0 (commented):
## Changes requested

Good diagnosis, and I confirmed the fix works: the three `dev_py_3_1x` envs really were missing `dev`, and adding it makes `make lint` / `make format` / `make sort` / `pre-commit install` all run. One gap holds it back: the `default` env is left with the identical `ruff: command not found`, and that is the env which *defines* the lint/format scripts.

**To unblock:**
1. `pyproject.toml:136` - add a `[tool.hatch.envs.default]` block with `features = ["dev"]` (detail in the inline comment)

**Problem is real.** On `main`, `pyproject.toml:83-87` defines the `dev` group, but `pyproject.toml:105-130` gives all three `dev_py_3_1x` envs `["test", "vector-stores", "llms", "extras"]` with no `dev`, and `pre-commit` is declared nowhere in the project. `.pre-commit-config.yaml` uses `language: system` with `entry: ruff check` / `entry: isort`, so both binaries have to be on `PATH` for the hooks to run at all. `hatch env show --json` on main confirms it: `dev_py_3_11 -> ['test', 'vector-stores', 'llms', 'extras']`.

**Red-green, in a real hatch env.** Worktree at `6ce586f`, `dev_py_3_11` built from scratch. Every functional run used a sanitized `PATH` (`<env>/bin:/usr/bin:/bin:...`) because this machine has ruff/isort/pre-commit installed globally and they would otherwise mask the bug.

- **RED** (main's pyproject): env `bin/` contained `pytest` and nothing else. `pip list` -> `pytest 9.1.1`, `pytest-asyncio 1.4.0`, `pytest-mock 3.15.1`. No ruff, no isort, no pre-commit.
- **GREEN** (this PR): `dev_py_3_11 -> [..., 'dev']`; `bin/` gained `isort`, `pre-commit`, `ruff`; versions `ruff 0.16.0`, `isort 8.0.1`, `pre-commit 4.6.2`.
- **Full CONTRIBUTING.md flow** inside a simulated `hatch shell dev_py_3_11`: `pre-commit install` succeeds, `make lint` -> `All checks passed!`, `make format` and `make sort` both execute.

**Your `hatch run` reasoning checks out.** hatch 1.12.0, `hatch/cli/__init__.py:113-121`: when `HATCH_ENV_ACTIVE` is set and `-e/--env` was not passed explicitly, `app.env = app.env_active`. So `make lint` inside the shell does run in `dev_py_3_11`, and no Makefile change is needed, exactly as you wrote.

**The gap.** With this PR applied, `hatch env show --json` still reports `default -> features: None`, the default env's `bin/` has no ruff/isort/pre-commit, and:

```
$ HATCH_ENV_ACTIVE=default make lint
hatch run lint
/bin/sh: ruff: command not found
make: *** [lint] Error 127
```

That is byte-identical to the symptom in #6683, and it is a documented path: `Makefile:11` (`install: hatch env create` creates the **default** env) and `docs/contributing/development.mdx:95` (`hatch shell  # For default environment`). I tested the addition in the inline comment: `default -> ['dev']`, the tools appear in `bin/`, and `make lint` -> `All checks passed!`.

**Version pins are right.** `ruff==0.16.0` matches `.github/workflows/ci.yml:109` exactly, so local lint and CI lint stay in agreement. `pre-commit>=3.5.0` resolves to 4.6.2 and is a correct floor for the `language: system` config already in `.pre-commit-config.yaml`.

**Two things I checked and then dropped, so they do not land on you:**
- `make format` rewrites 157 files (the `format` script is unscoped and ruff 0.16 formats `.md` and `.ipynb` too). Running `ruff format` against **main's** pyproject gives the same 157. Pre-existing, not caused by this PR. Worth its own issue.
- `isort>=5.13.2` being unpinned. I built both: isort 5.13.2 changes 17 files, isort 8.0.1 changes 16. The major bump is not producing meaningful drift, so the churn is pre-existing.

**CI:** all real checks green on 3.10 / 3.11 / 3.12 (`CI Gate`, `build_mem0`, `changelog_check`, CLA). Vercel is red on fork deploy perms; ignored, not merge-blocking.

**Security / bloat:** no runtime code path touched, and optional-dependency groups do not affect the published wheel (`[tool.hatch.build]` only includes `mem0/**/*.py`). The diff is already minimal, nothing to cut.

**Not verified:** did not create `dev_py_3_10` / `dev_py_3_12` locally (identical change, CI green on both); did not run the full pytest suite.

</details>

The root-cause read and the `hatch run` argument are both correct, and the fix is proven working for the envs it covers. Add `default` and this is ready. Push when ready and I will re-review.

pyproject.toml
ratishjain12: resolved.

pyproject.toml
```diff
@@ -84,6 +84,7 @@ dev = [
     "ruff==0.16.0",
     "isort>=5.13.2",
     "pytest>=8.2.2",
```
kartik-mem0: Delete this line. It is redundant with the `test` group, and now that `dev` reaches the `default` env it is what breaks CI: it installs pytest *into* that env, shadowing the runner's pytest that actually has the provider extras, so `make test` dies with 36 collection errors.

Verified with it removed: `default -> ['dev']`, the env still has ruff/isort/pre-commit, `pytest` resolves outside the env as it does on `main`, and `make lint` passes.

kartik-mem0 (commented):
## Changes requested

The `[tool.hatch.envs.default]` block is exactly what I asked for and it does fix `make lint` in the default env, verified. It also turned CI red, and that one is on my request rather than on you: I checked `make lint` in the default env last round and never ran `make test` there. Sorry for the extra trip.

**What broke:** `Python SDK / build_mem0` now fails on 3.10, 3.11 and 3.12 with 36 collection errors (`No module named 'boto3'`, `No module named 'azure'`, `cannot import name 'genai'`). Not the Vercel check, the real one.

**Why:** the `dev` group carries `pytest>=8.2.2`, so adding `dev` to `default` installs pytest *inside* that env. CI installs the provider extras into the runner's own Python (`ci.yml:108`) and relies on `hatch run test` falling through to it. Once the env owns a pytest it shadows the runner's, and the default env has no `vector-stores` / `llms` / `extras`, so collection dies.

**To unblock:**
1. `pyproject.toml:86` - delete `"pytest>=8.2.2"` from the `dev` group (detail in the inline comment)

One deleted line. It keeps everything you fixed, costs zero CI seconds, and `pytest` is already in the `test` group that all three `dev_py_3_1x` envs list, so nothing loses it.

**Reproduced CI locally.** Worktree at `b275893`, sanitized `PATH` (`<env>/bin:/usr/bin:/bin:...`) so my globally-installed tools could not interfere:

```
$ hatch run sh -c 'command -v pytest'
<default-env>/bin/pytest          <- the env now owns pytest

$ make lint
All checks passed!                 <- your fix works

$ make test
!!!!!!!!!!!!!!!!!!! Interrupted: 36 errors during collection !!!!!!!!!!!!!!!!!!!
make: *** [test] Error 2
```

Same 36-error count as CI job 94759329322.

**Proved the fallback mechanism directly.** Second worktree with **main's** `pyproject.toml`, plus a marker script named `pytest` on the outer PATH:

```
$ hatch env show --json     ->  default -> None
$ ls "<default-env>/bin" | grep -E '^(ruff|isort|pre-commit|pytest)$'
(none)

$ make test
hatch run test
OUTER-PYTEST-INVOKED argv=tests/   <- falls through to the outer interpreter
```

That fallback is what keeps CI green on `main`, and it is what the new block removes.

**Tested the fix I am asking for.** `dev` reduced to `["ruff==0.16.0", "isort>=5.13.2", "pre-commit>=3.5.0"]`, keeping your `[tool.hatch.envs.default]` block:

```
$ hatch env show --json     ->  default -> ['dev']
$ hatch run sh -c 'command -v ruff; command -v isort; command -v pre-commit; command -v pytest'
<default-env>/bin/ruff
<default-env>/bin/isort
<default-env>/bin/pre-commit
<outer>/pytest                     <- falls through again, as on main

$ make lint   ->  All checks passed!
$ make test   ->  OUTER-PYTEST-INVOKED argv=tests/
```

**I also tested the heavier alternative so you don't have to.** Giving `default` the full set `["test", "vector-stores", "llms", "extras", "dev"]` also works and is arguably more complete, since it makes the default env's own `test` script genuinely usable:

```
$ make test
= 1 failed, 1767 passed, 73 skipped, 8 warnings, 1 error in 173.51s =
```

Collection errors 36 -> 0. But that default env measures **2.2 GB** against 194 MB for the one-line fix, and CI would install the whole dependency set twice, once via pip for the job and once via hatch for the env. `build_mem0` already carries an explicit disk-cleanup step (`ci.yml:83-90`), so I would not spend that here. Worth its own issue if anyone wants the default env's `make test` to stand alone; leaving it as-is is not a regression, it is exactly what `main` does today.

**The 2 remaining failures in that run are not yours:** `RedisSearchError: unknown command 'FT.INFO'`, i.e. my local Redis has no RediSearch module. `git diff --name-only main...HEAD` on this PR returns `pyproject.toml` and nothing else, and current `main` HEAD `ef6f51d97` is *"fix(tests): check for RediSearch module availability in Redis tests (#6687)"*, which this branch predates.

**Still green:** `changelog_check`, `check_changes`, `license/cla`. Vercel red on fork deploy perms, ignored as always.

**Not verified:** did not build `dev_py_3_10` / `dev_py_3_12` (identical config, CI covers them); did not push the one-line fix to a branch to watch CI go green, the PATH-fallback proof above is the evidence.

</details>

You read the root cause correctly both rounds, and the default-env block is the right shape. Drop that one redundant pytest line and I think this is done. Push when ready and I will re-review.

pyproject.toml
ratishjain12: done.

kartik-mem0 (commented):
Re-reviewed at `5abc28c6`. Content-identical rebase of the previously reviewed head. Fresh tests and lint pass against the new SHA. LGTM still holds.

PR #6693, opened 2026-07-31
fix(embeddings/lmstudio): honor LMSTUDIO_BASE_URL when config url unset

### LINKED ISSUE #6692: bug(embeddings/lmstudio): LMSTUDIO_BASE_URL env ignored; default baked into BaseEmbedderConfig

Bartok9:
## Summary
LM Studio embedder resolves base URL as config → `LMSTUDIO_BASE_URL` → localhost.

Closes #6692

## Motivation
Config default was a concrete localhost string, so the env var could never apply (LLM parity with #6526).

## Verification
`pytest tests/embeddings/test_lmstudio.py` — 4 passed

### DIFF
--- a/mem0/configs/embeddings/base.py
+++ b/mem0/configs/embeddings/base.py
@@ -34,7 +34,7 @@ class BaseEmbedderConfig(ABC):
         # Gemini specific
         output_dimensionality: Optional[str] = None,
         # LM Studio specific
-        lmstudio_base_url: Optional[str] = "http://localhost:1234/v1",
+        lmstudio_base_url: Optional[str] = None,
         # AWS Bedrock specific
         aws_access_key_id: Optional[str] = None,
         aws_secret_access_key: Optional[str] = None,
@@ -70,7 +70,7 @@ class BaseEmbedderConfig(ABC):
         :type memory_update_embedding_type: Optional[str], optional
         :param memory_search_embedding_type: The type of embedding to use for the search memory action, defaults to None
         :type memory_search_embedding_type: Optional[str], optional
-        :param lmstudio_base_url: LM Studio base URL to be use, defaults to "http://localhost:1234/v1"
+        :param lmstudio_base_url: LM Studio base URL to use. None falls through to LMSTUDIO_BASE_URL or http://localhost:1234/v1
         :type lmstudio_base_url: Optional[str], optional
         """
 
--- a/mem0/embeddings/lmstudio.py
+++ b/mem0/embeddings/lmstudio.py
@@ -1,3 +1,4 @@
+import os
 from typing import Literal, Optional
 
 from openai import OpenAI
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Heads up before this goes further: #6527 (yours as well) already covers everything this PR does and more. It makes the same `lmstudio_base_url` default change in `mem0/configs/embeddings/base.py` and the same env fallback in `mem0/embeddings/lmstudio.py`, and it additionally fixes the Python LLM client plus both TypeScript surfaces (`lmstudio.ts` for the LLM and the embedder). #6527 is reviewed and ready to merge.

Two concrete differences worth noting. This PR keeps the `self.config.lmstudio_base_url = base_url` write-back, which #6527 removed after a repo-wide grep showed it has no readers. And it adds `tests/embeddings/test_lmstudio.py` while #6527 extends the existing `tests/embeddings/test_lm_studio_embeddings.py`, so the two will conflict on merge.

Suggestion: let #6527 land and close this one, but add `Closes #6692` to #6527's description first so that issue still gets closed out. Your call since both are yours.

kartik-mem0:
Closing this one as superseded by your own #6527, which is already reviewed and sitting in the merge queue.

The two overlap exactly on mem0/configs/embeddings/base.py and mem0/embeddings/lmstudio.py: both flip the lmstudio_base_url default from the baked-in localhost string to None, and both add the config then LMSTUDIO_BASE_URL then localhost fallback chain. #6527 then goes further and applies the same fix to the Python LLM path (mem0/configs/llms/lmstudio.py, mem0/llms/lmstudio.py) and to both TypeScript surfaces (embeddings/lmstudio.ts, llms/lmstudio.ts), with tests on each side. It also drops the write-back onto self.config.lmstudio_base_url that this PR still carries, which nothing downstream reads.

So everything here lands with #6527, and merging both would just conflict. Nothing wrong with this diff, it is the narrower half of a fix you had already made completely. Thanks for both, and please keep an eye on #6527.

Bartok9:
Agreed — thanks Kartik. #6527 is the complete version (Python LLM + embedder + both TS surfaces, plus dropping the dead `self.config.lmstudio_base_url` write-back). I've added `Closes #6692` to #6527 so the issue still gets closed out. Happy to see this one go. Will keep an eye on #6527.

PR #6698, opened 2026-07-31
fix(ts-client): send page/page_size independently in getAll()

### LINKED ISSUE #6697: TS client MemoryClient.getAll() ignores pageSize unless page is also passed
### Component

TypeScript SDK / Platform client (`mem0-ts/src/client/mem0.ts`)

### What happens

`getAll()` only appends pagination to the request URL when **both** `page` and `pageSize` are set:

```ts
let url = `${this.host}/v3/memories/`;
if (page && pageSize) {
  url += `?page=${page}&page_size=${pageSize}`;
}
```

So `getAll({ pageSize: 50 })` — a caller who sets a page size and relies on `page` defaulting to 1 — sends **no** query string, and the server falls back to its default page size. `getAll({ page: 2 })` likewise drops the page. The requested page size silently doesn't take effect.

### Expected

`page` and `page_size` are read from the query string independently, so either one should be forwarded on its own:

```ts
const queryParams = new URLSearchParams();
if (page !== undefined) queryParams.set("page", String(page));
if (pageSize !== undefined) queryParams.set("page_size", String(pageSize));
```

### Note

This is the TS counterpart of the Python client bug #6227 (fix #6229). The Python `MemoryClient.get_all()` has the same `if "page" in params and "page_size" in params` gate on current main.

abhay-codes07:
### Description

`MemoryClient.getAll()` only appended pagination to the URL when **both** `page` and `pageSize` were provided:

```ts
if (page && pageSize) {
  url += `?page=${page}&page_size=${pageSize}`;
}
```

So `getAll({ pageSize: 50 })` — a caller who sets a page size and relies on `page` defaulting to 1 — sent no query string, and the server returned its default page size; `getAll({ page: 2 })` dropped the page the same way. The fix builds the query string from whichever of `page`/`page_size` is present, so either takes effect on its own.

This is the TS counterpart of the Python client bug #6227 (fix #6229); the Python `get_all()` has the same gate on main.

Closes #6697

### Type of change

### How Has This Been Tested?

`jest src/client/tests/memoryClient.crud.test.ts` (from the `mem0-ts` root), 27 passed. Added a `getAll() pagination` suite (fetch-mocked):

- `sends page_size when only pageSize is provided` — URL now contains `page_size=50`.
- `sends page when only page is provided` — URL contains `page=2`.
- `sends both page and page_size when both are provided`.
- `sends no pagination query string when neither is provided`.

Reverting the source change fails the two "only one provided" tests (the URL comes back as bare `/v3/memories/`).

### Checklist

### DIFF
--- a/mem0-ts/src/client/mem0.ts
+++ b/mem0-ts/src/client/mem0.ts
@@ -349,9 +349,17 @@ export default class MemoryClient {
       ...(filters && { filters }),
     };
 
+    // Pagination is read from the query string by the server, so send page and
+    // page_size independently. Requiring both (the old `page && pageSize`) meant
+    // getAll({ pageSize: 50 }) — a caller relying on page defaulting to 1 —
+    // silently dropped page_size and returned the server's default page size.
     let url = `${this.host}/v3/memories/`;
-    if (page && pageSize) {
-      url += `?page=${page}&page_size=${pageSize}`;
+    const queryParams = new URLSearchParams();
+    if (page !== undefined) queryParams.set("page", String(page));
+    if (pageSize !== undefined) queryParams.set("page_size", String(pageSize));
+    const queryString = queryParams.toString();
+    if (queryString) {
+      url += `?${queryString}`;
     }
 
     const response = await this._fetchWithErrorHandling(url, {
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@kartik-mem0 small ts-client fix with tests. getAll() only sent pagination when both page and pageSize were set, so getAll({ pageSize: 50 }) silently dropped the page size. Now sends whichever is provided. TS counterpart of the Python client bug #6227/#6229. Quick look when you get a chance, thanks.

PR #6700, opened 2026-07-31
fix(ts-client): let createWebhook target a project; harden deleteWebhook

### LINKED ISSUE #6699: TS client: createWebhook can't target a project, and deleteWebhook can serialize an object into the URL
### Component

TypeScript SDK / Platform client (`mem0-ts/src/client/mem0.ts`)

### Two related webhook issues

**1. `createWebhook` can only use the ping-resolved project.**
The Python client's `create_webhook(url, name, project_id, event_types)` takes an explicit `project_id`, and the TS `getWebhooks()` already accepts a `projectId` override. But `createWebhook` hardcodes `this.projectId`:

```ts
const response = await this._fetchWithErrorHandling(
  `${this.host}/api/v1/webhooks/projects/${this.projectId}/`, ...);
```

So you cannot create a webhook under a different project. And if the client hasn't resolved a project (e.g. `ping()` returned no `project_id`), both `getWebhooks()` and `createWebhook()` post to `.../projects/undefined/` instead of raising a clear error.

**2. `deleteWebhook`'s fallback serializes the whole object into the URL.**

```ts
const webhook_id = data.webhookId || data;
```

When `data.webhookId` is empty/missing, this falls back to `data` (the object), so the request goes to `/api/v1/webhooks/[object Object]/`.

### Expected

- `createWebhook` accepts an optional `projectId` (defaulting to the client's resolved project), and both `getWebhooks`/`createWebhook` raise a clear error when no project can be resolved rather than sending `undefined`.
- `deleteWebhook` accepts the typed object or a bare id string, and throws when the id is empty instead of serializing `[object Object]`.

(`updateWebhook`/`deleteWebhook` correctly use `/api/v1/webhooks/{id}/`, which is project-independent, matching the Python client — no `project_id` needed there.)

abhay-codes07:
### Description

Two related webhook fixes in the TS platform client.

**1. `createWebhook` can now target a project.** It hardcoded `this.projectId`, so you couldn't create a webhook under a different project (the Python `create_webhook` takes an explicit `project_id`, and `getWebhooks` already accepts a `projectId` override). It now reads an optional `projectId` from the payload, defaulting to the client's resolved project. Both `getWebhooks` and `createWebhook` also now raise a clear error when no project can be resolved, instead of posting to `.../projects/undefined/`.

**2. `deleteWebhook` no longer serializes an object into the URL.** The old `data.webhookId || data` fell back to the whole object when `webhookId` was missing, producing `/api/v1/webhooks/[object Object]/`. It now accepts the typed object or a bare id string and throws on an empty id.

`updateWebhook`/`deleteWebhook` use `/api/v1/webhooks/{id}/` (project-independent, matching the Python client), so they need no `project_id`.

Closes #6699

### Type of change

### How Has This Been Tested?

`jest src/client/tests/memoryClient.webhooks.test.ts` (from the `mem0-ts` root), 19 passed (14 existing + 5 new); full client unit suite green; `tsup` build (incl. DTS) compiles.

New tests:
- `createWebhook` defaults to the ping-resolved project, and **targets an explicit `projectId`** when provided.
- `createWebhook` **throws when no project can be resolved** (never posts to `/projects/undefined/`).
- `deleteWebhook` accepts a bare id string, and **throws on an empty `webhookId`** instead of serializing the object.

Reverting the source change fails the explicit-projectId, no-project-throw, and empty-webhookId tests.

### Checklist

### DIFF
--- a/mem0-ts/src/client/mem0.ts
+++ b/mem0-ts/src/client/mem0.ts
@@ -622,6 +622,11 @@ export default class MemoryClient {
     if (this.telemetryId === "") await this.ping();
     this._captureEvent("get_webhooks", []);
     const project_id = data?.projectId || this.projectId;
+    if (!project_id) {
+      throw new Error(
+        "projectId is required to list webhooks; pass it in the argument or ensure the client resolved a project via ping().",
+      );
+    }
     const response = await this._fetchWithErrorHandling(
       `${this.host}/api/v1/webhooks/projects/${project_id}/`,
       {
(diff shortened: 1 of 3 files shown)


abhay-codes07:
@kartik-mem0 small ts-client fix with tests. Two webhook issues: createWebhook hardcoded this.projectId so you couldn't target another project (Python's create_webhook takes an explicit project_id), and deleteWebhook's data.webhookId || data fallback serialized [object Object] into the URL when the id was empty. Now createWebhook accepts an optional projectId + validates, and deleteWebhook accepts an object or bare string and throws on empty. Quick look when you get a chance, thanks.

PR #6702, opened 2026-07-31
fix(ts-oss): preserve system messages and use the configured LLM in parse_vision_messages

### LINKED ISSUE #6701: TS OSS parse_vision_messages drops system messages and hardcodes OpenAI for image description
### Component

TypeScript SDK (OSS) — `mem0-ts/src/oss/src/utils/memory.ts`

### Two bugs vs the Python `parse_vision_messages`

**1. System messages are silently dropped.**

```ts
  let new_message = { role: message.role, content: "" };
  if (message.role !== "system") {   // <- system messages never pushed
    ...
    parsed_messages.push(...);
  }
}
```

Any `system` message is skipped and never returned, so caller-provided system context/instructions are stripped before the extraction pipeline. The Python `parse_vision_messages` explicitly keeps them (`if role == "system": returned_messages.append(msg); continue`).

**2. Image description hardcodes an env-only OpenAI client, ignoring the configured LLM.**

```ts
const get_image_description = async (image_url: string) => {
  const llm = new OpenAILLM({ apiKey: process.env.OPENAI_API_KEY });
  ...
};
```

Multimodal `add()` always describes images with OpenAI read from `process.env.OPENAI_API_KEY`. So:
- users configured with a non-OpenAI provider get OpenAI for vision (or a failure) instead of their provider, and
- even OpenAI users who pass `apiKey` via config (not the env var) get an unauthenticated client.

The Python `parse_vision_messages(messages, llm=...)` uses the configured LLM.

### Expected

- Preserve `system` messages.
- Describe images with the configured LLM (thread it through from `Memory.add()`), falling back to the env-based OpenAI client only when no LLM is available.

abhay-codes07:
### Description

Two bugs in `parse_vision_messages` (`mem0-ts/src/oss/src/utils/memory.ts`), both divergences from the Python implementation:

**1. System messages were silently dropped.** The `if (message.role !== "system")` guard meant `system` messages were never pushed, so caller-provided system context/instructions were stripped before the extraction pipeline. Python explicitly preserves them (`if role == "system": returned_messages.append(msg)`). Now TS does too.

**2. Image description hardcoded an env-only OpenAI client.** `get_image_description` built `new OpenAILLM({ apiKey: process.env.OPENAI_API_KEY })`, so multimodal `add()` always used OpenAI read from the env var — ignoring a non-OpenAI provider, and even ignoring an OpenAI `apiKey` supplied via config rather than the env. It now threads the configured LLM through from `Memory.add()` (matching Python's `parse_vision_messages(messages, llm=...)`), falling back to the env OpenAI client only when no LLM is provided.

Closes #6701

### Type of change

### How Has This Been Tested?

`jest src/oss/tests/parse-vision-messages.unit.test.ts` (from the `mem0-ts` root), 5 passed; `memory.add.test.ts` still green; `tsup` build (incl. DTS) compiles.

New tests:
- system messages preserved (alone and alongside an image message);
- plain user/assistant messages pass through unchanged;
- an image is described with the **configured LLM** (asserts the injected LLM receives the `image_url` and its output becomes the message content);
- missing `image_url.url` still throws.

Reverting the source change fails the system-preservation and configured-LLM tests.

### Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -763,7 +763,10 @@ export class Memory {
       ? (messages as Message[])
       : [{ role: "user", content: messages }];
 
-    const final_parsedMessages = await parse_vision_messages(parsedMessages);
+    const final_parsedMessages = await parse_vision_messages(
+      parsedMessages,
+      this.llm,
+    );
 
     // Add to vector store
     const vectorStoreResult = await this.addToVectorStore(
--- a/mem0-ts/src/oss/src/utils/memory.ts
+++ b/mem0-ts/src/oss/src/utils/memory.ts
@@ -1,11 +1,15 @@
+import { LLM } from "../llms/base";
 import { OpenAILLM } from "../llms/openai";
 import { Message } from "../types";
 
-const get_image_description = async (image_url: string) => {
-  const llm = new OpenAILLM({
-    apiKey: process.env.OPENAI_API_KEY,
-  });
-  const response = await llm.generateResponse([
+const get_image_description = async (image_url: string, llm?: LLM) => {
+  // Use the configured LLM (which carries the caller's provider + apiKey), like
+  // the Python parse_vision_messages does. Falling back to a bare OpenAI client
+  // read only from the env, as before, ignored a provider/apiKey set via config
+  // and forced OpenAI even for non-OpenAI setups.
+  const visionLlm =
+    llm ?? new OpenAILLM({ apiKey: process.env.OPENAI_API_KEY });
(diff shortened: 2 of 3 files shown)


abhay-codes07:
@kartik-mem0 ts-oss fix with tests. Two parity bugs in parse_vision_messages: it silently dropped system messages (Python keeps them, so caller system context was lost before extraction), and it hardcoded new OpenAILLM({ apiKey: process.env.OPENAI_API_KEY }) for image description, ignoring the configured provider/apiKey. Now preserves system messages and threads the configured LLM through from add(), matching the Python parse_vision_messages(messages, llm=...). Quick look when you get a chance, thanks.

PR #6714, opened 2026-08-01
fix(openclaw): pass session id as run_id on capture and session-scoped search

### LINKED ISSUE #6708: openclaw integration: run_id is silently dropped on auto-capture — buildAddOptions sets opts.run_id (snake) but the provider reads opts.runId (camel)
### Summary

In `integrations/openclaw`, the `run_id` scoping key never reaches the wire for auto-captured (and search-scoped) calls: the options builder writes the **snake_case** field while the provider layer only ever reads the **camelCase** one.

### Where

Writer — [`integrations/openclaw/index.ts:293`](https://github.com/mem0ai/mem0/blob/main/integrations/openclaw/index.ts#L293) (and `:316`, the search-options builder):

```ts
if (runId) opts.run_id = runId;   // snake_case
```

Readers — `integrations/openclaw/backend/platform.ts`, every consumption site (L108, L137, L164, L216, L252, L272), e.g.:

```ts
if (opts.runId) payload.run_id = opts.runId;   // camelCase — never set by the builder
```

The `AddOptions`/`SearchOptions` declarations in `backend/base.ts` declare `runId?` (camelCase), so the builder's `opts.run_id` lands as an undeclared extra property that nothing consumes.

### Impact

- Auto-captured memories lose their session/channel scoping: `add` payloads carry no `run_id`, so memories from different channels/sessions are indistinguishable server-side.
- `search`/`getAll` calls that intend `run_id` filtering silently query unscoped.
- No error anywhere — the value is just dropped, so this is easy to carry for a long time without noticing. We found it in production only because a downstream consumer (a reaction-correlation feature keyed on `run_id`) could never match anything: stored payloads all had `run_id` absent while the gateway was demonstrably passing a `sessionKey`.

Observed on `@mem0/openclaw-mem0` v1.0.6 (open-source mode against a mem0-compatible server); the mismatch is still present at current `main` (lines cited above).

### Fix

One line class of fix — either:

```ts
if (runId) opts.runId = runId;   // in buildAddOptions / buildSearchOptions
```

or accept both spellings at the provider boundary. The former matches the declared `AddOptions` interface.

### Repro

1. Enable auto-capture in open-source mode pointing at any mem0-compatible endpoint.
2. Converse in any channel so `agent_end` fires with a `sessionKey`.
3. Inspect the POSTed `/v1/memories/` body (or the stored payload): `run_id` is absent, despite `buildAddOptions(undefined, sessionId, sessionId)` being called with a defined session key.

rodboev:
Closes #6708

## Description

OpenClaw receives the current session key but drops it from auto-captured memory writes and explicit session-scoped searches. The resulting Mem0 requests omit `run_id`, so memories from the active conversation aren't isolated and session searches return unscoped results.

## Root cause

The existing builders accept separate `runId` and `sessionKey` arguments. Auto-capture passes the session only in the `sessionKey` slot at `integrations/openclaw/index.ts:1036`. The `memory_search` session and all paths do the same at `integrations/openclaw/tools/memory-search.ts:50,58`. The provider translation already handles `options.run_id`, so the fix belongs at these consumers.

## Changes

- Pass the current session ID through both builder slots for auto-capture.
- Pass the current session ID through both slots for `memory_search` session and all scopes.
- Add production-path regression coverage for the final SDK add/search options.
- Keep explicit CLI session search run IDs intact with a direct option assertion.
- Preserve long-term and cold-start recall as cross-session queries, along with existing skips, filters, namespaces, and provider translation.
- In open-source mode, apply the session run scope to the existing auto-capture update decision so separate sessions don't update one another's memories.

## Verification boundary

Local tests verify the OpenClaw hook, provider adapter, and SDK-facing payload. Hosted persistence and read-back were not verified.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Test Coverage

Focused OpenClaw hook, tool, and CLI tests cover auto-capture run propagation, session/all search filters, explicit CLI session run IDs, long-term and cold-start negative space, sessionless capture, namespace routing, and existing provider translation. The OpenClaw package type check, build, and full test suite pass locally, with 454 tests passed and 1 skipped.

## Checklist

- [x] Linked issue
- [x] Correct type of change
- [x] Scope and breaking-change status documented
- [x] Full validation completed locally
- [ ] CI checks passed

### DIFF
--- a/integrations/openclaw/index.ts
+++ b/integrations/openclaw/index.ts
@@ -1033,7 +1033,7 @@ function registerHooks(
         content: `Current date: ${timestamp}. The user is identified as "${cfg.userId}". Extract durable facts from this conversation. Include this date when storing time-sensitive information.`,
       });
 
-      const addOpts = buildAddOptions(undefined, undefined, sessionId);
+      const addOpts = buildAddOptions(undefined, sessionId, sessionId);
       const captureStart = Date.now();
       provider
         .add(formattedMessages, addOpts)
--- a/integrations/openclaw/index.ts
+++ b/integrations/openclaw/index.ts
@@ -1033,6 +1033,8 @@ function registerHooks(
         content: `Current date: ${timestamp}. The user is identified as "${cfg.userId}". Extract durable facts from this conversation. Include this date when storing time-sensitive information.`,
       });
 
+      // run_id intentionally holds the raw sessionKey (e.g. "agent:researcher:uuid-1");
+      // session-scoped search reads it back unchanged, so capture and recall stay aligned.
       const addOpts = buildAddOptions(undefined, sessionId, sessionId);
       const captureStart = Date.now();
       provider
--- a/integrations/openclaw/tools/memory-search.ts
+++ b/integrations/openclaw/tools/memory-search.ts
@@ -47,7 +47,7 @@ export function createMemorySearchTool(deps: ToolDeps) {
 
         if (scope === "session") {
           if (currentSessionId) {
-            results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, currentSessionId, currentSessionId)));
+            results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, currentSessionId)));
           }
         } else if (scope === "long-term") {
           results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit)));
@@ -55,7 +55,7 @@ export function createMemorySearchTool(deps: ToolDeps) {
           const longTerm = await provider.search(query, applyFilters(buildSearchOptions(uid, limit)));
           let session: MemoryItem[] = [];
           if (currentSessionId) {
-            session = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, currentSessionId, currentSessionId)));
+            session = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, currentSessionId)));
           }
           const seen = new Set(longTerm.map((r) => r.id));
           results = [...longTerm, ...session.filter((r) => !seen.has(r.id))];
--- a/integrations/openclaw/tools/memory-search.ts
+++ b/integrations/openclaw/tools/memory-search.ts
@@ -47,7 +47,7 @@ export function createMemorySearchTool(deps: ToolDeps) {
 
         if (scope === "session") {
           if (currentSessionId) {
-            results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, undefined, currentSessionId)));
+            results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, currentSessionId, currentSessionId)));
           }
         } else if (scope === "long-term") {
           results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit)));
@@ -55,7 +55,7 @@ export function createMemorySearchTool(deps: ToolDeps) {
           const longTerm = await provider.search(query, applyFilters(buildSearchOptions(uid, limit)));
           let session: MemoryItem[] = [];
           if (currentSessionId) {
-            session = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, undefined, currentSessionId)));
+            session = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, currentSessionId, currentSessionId)));
           }
           const seen = new Set(longTerm.map((r) => r.id));
           results = [...longTerm, ...session.filter((r) => !seen.has(r.id))];
--- a/integrations/openclaw/tests/cli-commands.test.ts
+++ b/integrations/openclaw/tests/cli-commands.test.ts
@@ -191,18 +191,21 @@ function createMockCfg() {
  * Register CLI commands using mocked deps, and return the captured
  * mem0 command tree plus all mock objects for assertions.
  */
-function setup() {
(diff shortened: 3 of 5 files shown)


integrations/openclaw/index.ts
```diff
@@ -1033,7 +1033,7 @@ function registerHooks(
         content: `Current date: ${timestamp}. The user is identified as "${cfg.userId}". Extract durable facts from this conversation. Include this date when storing time-sensitive information.`,
       });
 
-      const addOpts = buildAddOptions(undefined, undefined, sessionId);
+      const addOpts = buildAddOptions(undefined, sessionId, sessionId);
```
kartik-mem0, line 1038: `run_id` is set to the raw `ctx.sessionKey` (e.g. "agent:researcher:uuid-1") rather than a parsed session-only id. Functionally fine since search reads it back the same way, but worth a one-line comment noting this is intentional, not an oversight.

integrations/openclaw/tools/memory-search.ts
```diff
@@ -47,15 +47,15 @@ export function createMemorySearchTool(deps: ToolDeps) {
 
         if (scope === "session") {
           if (currentSessionId) {
-            results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, undefined, currentSessionId)));
+            results = await provider.search(query, applyFilters(buildSearchOptions(uid, limit, currentSessionId, currentSessionId)));
```
kartik-mem0: The 4th arg (sessionKey) is dead here: `buildSearchOptions`'s user_id resolution short-circuits on the always-truthy `uid` in the 2nd arg (same at line 58). Pre-existing, not introduced here. Consider collapsing `buildAddOptions`/`buildSearchOptions` to a named-options object so a value can't land in the wrong slot again.

kartik-mem0 (commented):

Auto-capture and session-scoped search dropped `run_id` at the call site, so memories leaked across sessions; this threads `sessionId` into the `runId` slot at the two exact call sites the trace found (`index.ts:1036`, `memory-search.ts:50,58`).

**Follow-up:** none filed. Two optional Minor notes inline (a dead 4th argument in `buildSearchOptions`, a comment suggestion for the raw session key).
**Not verified:** live SDK calls, only the `mem0ai`/`mem0ai/oss` client boundary is mocked.

Risk: low, scoped to the two broken call sites and proven not to touch the cross-session recall hook or the already-correct CLI path · Your time: ~3 min

- **Problem is real:** `index.ts:1036` and `memory-search.ts:50,58` passed the session id only into the `sessionKey` slot, never `runId`, so auto-captured memories and session-scoped search carried no `run_id`.
- **Tests:** `pnpm exec vitest run tests/hooks-session-scope.test.ts tests/tools.test.ts tests/cli-commands.test.ts` -> 148 passed.
- **Red-green:** reverting `index.ts` and `memory-search.ts` to `main` fails exactly 5 tests (the fix's own assertions); restoring passes 148/148 again.
- **CI:** OpenClaw build/lint/test green on Node 20 and 22, CI Gate green. Vercel shows fail (preview-deploy only, not merge-blocking, ignored).
- **Security:** cleared. `sessionId` comes from OpenClaw's internal `ctx.sessionKey`, not user text; flows into a structured filter object, never interpolated; narrows results within the existing user scope, never widens across users.
- **Bloat (ponytail):** lean already, net -0 lines possible.
- **Linked issue:** #6708

</details>

goldenwo (commented):
Reviewed. One nit at the end.

Mocking at the `mem0ai` SDK boundary and asserting camelCase `runId` on `mockSdk.addOptions` puts the assertion downstream of the `providers.ts` translation. It therefore fails on either failure mode from #6708: the call site reverted to `(undefined, undefined, sessionId)`, or the snake→camel conversion removed, which is how v1.0.6 dropped it. An assertion at the call site would only catch the first. Both provider paths are covered: `providers.ts:126` through the `mem0ai` mock, `:407` through `vi.doMock("mem0ai/oss")`.

`keeps long-term and cold-start recall unscoped` pins the deliberate boundary. I had expected that to stay a comment rather than become a test.

Nit, `tests/hooks-session-scope.test.ts`: `vi.doUnmock("mem0ai/oss")` is the last statement in the `it()`, so a failing assertion above it skips the unmock and leaks into whatever runs next. It is the last test in the file, so there is no impact today. `onTestFinished` or an `afterEach` would make it order-independent.

I run this integration against a self-hosted server that records the received `run_id` per capture. Once this is in a release I can confirm on the wire and report back (offered in #6708).

rodboev:
Addressed the nits on the rework push (08bb782c):

- `integrations/openclaw/index.ts:1036` now documents that `run_id` intentionally carries the raw `sessionKey`, matching how session-scoped search reads it back.
- `integrations/openclaw/tools/memory-search.ts:50` and `:58` drop the redundant 4th `sessionKey` argument; the resolved user id already short-circuits the fallback, so behavior is unchanged (the `tools.test.ts` assertion was updated to match).
- `tests/hooks-session-scope.test.ts` hoists `vi.doUnmock("mem0ai/oss")` into an `afterEach`, so a failing assertion above it can no longer leak the mock into later tests.

kartik-mem0 (commented):
Re-reviewed at `08bb782c`. Content-identical rebase of the previously reviewed head. Fresh tests and lint pass against the new SHA. LGTM still holds.

PR #6722, opened 2026-08-01
fix(memory): reject NaN search threshold (#6713)

### LINKED ISSUE #6713: bug(search): NaN threshold bypasses validation and disables score filtering
### Component

Python SDK

### Summary

`Memory.search()` accepts `float("nan")` as `threshold` even though the parameter is documented and validated as a value between 0 and 1. Because comparisons with NaN are always false, the current range check does not reject it, and the downstream semantic-score gate keeps candidates that should have been filtered.

### Steps to Reproduce

```python

_validate_search_params(float("nan"))
print("NaN threshold was accepted")
```

On current `main`, this prints successfully instead of raising `ValueError`.

The same value can reach the public search path:

```python
memory.search(
    "test query",
    filters={"user_id": "test"},
    threshold=float("nan"),
)
```

Root cause: `_validate_search_params()` checks:

```python
    raise ValueError(...)
```

For NaN, both comparisons evaluate to `False`. Later, `score_and_rank()` checks `semantic_score < threshold`; that comparison is also always `False`, so the threshold gate does not exclude candidates. The shared validator affects both `Memory.search()` and `AsyncMemory.search()`.

### Expected Behavior

`Memory.search(..., threshold=float("nan"))` should raise `ValueError` before embedding or vector-store work, just like negative values and values above 1.

### Actual Behavior

NaN passes validation and reaches the search/scoring path, where it effectively disables threshold filtering.

### Proposed Fix and Test Plan

Add a targeted NaN check in `_validate_search_params()` before the existing inclusive `[0, 1]` range check, without changing valid boundary behavior for `0` and `1` or unrelated input compatibility.

Add a deterministic regression test alongside `TestSearchParamValidation` that:

1. calls `Memory.search(..., threshold=float("nan"))`;
2. asserts `ValueError` using the existing threshold-validation message style; and
3. asserts embedding/vector-store work was not invoked.

Run the focused validation test, the relevant Python test suite, and the repository lint/format checks.

If this issue is available for external contribution, could a maintainer confirm the proposed scope and assign it before implementation?

### Environment

- mem0 version: current `main` at `38e47ac`
- Python/Node version: Python 3.11/3.12 (behavior follows Python float semantics)
- OS: reproducible on any OS

Anai-Guo:
### What

`Memory.search(..., threshold=float("nan"))` is accepted even though `threshold` is documented and validated as a value in `[0, 1]`.

`_validate_search_params()` guards the range with:

```python
```

For `NaN` both comparisons are `False`, so validation passes. Downstream, the semantic gate `semantic_score < threshold` is also always `False` for `NaN`, so the threshold filter silently keeps every candidate instead of filtering.

### Fix

Reject `NaN` in the same range check, so it raises the existing `Invalid threshold: ...` `ValueError` before any embedding or vector-store work happens. The shared validator covers both `Memory.search()` and `AsyncMemory.search()`.

`inf`/`-inf` were already rejected by the existing comparisons; `0` and `1` keep being accepted.

### Verification

Ran `tests/test_main.py` against the released `mem0ai` package (byte-identical to `mem0/memory/main.py` on `main`) with the new test appended:

- **before the fix**: `1 failed, 44 passed` — NaN slipped past validation and reached the search path
- **after the fix**: `45 passed`

```
>>> from mem0.memory.main import _validate_search_params
>>> _validate_search_params(float("nan"))
ValueError: Invalid threshold: nan. Must be between 0 and 1 (inclusive).
>>> _validate_search_params(0); _validate_search_params(1)   # still fine
```

Lint: `ruff==0.16.0` (the pinned version) `check` passes on both changed files, and the added block is `ruff format` clean.

Fixes #6713

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -4,6 +4,7 @@ import gc
 import hashlib
 import json
 import logging
+import math
 import os
 import time
 import uuid
@@ -196,7 +197,7 @@ def _validate_search_params(threshold: Optional[float] = None, top_k: Optional[i
     if threshold is not None:
         if not isinstance(threshold, (int, float)):
             raise ValueError("threshold must be a valid number")
-        if threshold < 0 or threshold > 1:
+        if math.isnan(threshold) or threshold < 0 or threshold > 1:
             raise ValueError(
                 f"Invalid threshold: {threshold}. Must be between 0 and 1 (inclusive)."
             )
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -540,3 +540,14 @@ class TestSearchParamValidation:
             result = memory_instance.search("test", filters={"user_id": "test"}, top_k=0)
 
         assert "results" in result
+
+    def test_search_rejects_nan_threshold(self, memory_instance):
+        """Search should reject a NaN threshold before doing any retrieval work."""
+        memory_instance.embedding_model.embed = Mock()
+        memory_instance.vector_store.search = Mock()
+
+        with pytest.raises(ValueError, match="Invalid threshold.*Must be between 0 and 1"):
+            memory_instance.search("test query", filters={"user_id": "test"}, threshold=float("nan"))
+
+        memory_instance.embedding_model.embed.assert_not_called()
+        memory_instance.vector_store.search.assert_not_called()


ErenAta16 (commented):
Ran the comparison rather than reasoning about IEEE semantics from memory, and the fix is exact:

```
        old check              new check
nan     accepts                rejects     <- the bug
inf     rejects                rejects
-inf    rejects                rejects
0.5     accepts                accepts
1.5     rejects                rejects
```

Only `NaN` moves, everything else is byte-identical behaviour, which is what you want from a validation tightening.

Worth stating explicitly in the description that the infinities were never affected. `float("inf") > 1` and `float("-inf") < 0` are both `True`, so the existing range check already caught them; only `NaN` has the property that every ordered comparison is `False`. Otherwise the next person reading `math.isnan(...)` will wonder whether `math.isinf(...)` is missing and add it, which would be dead code.

`assert_not_called()` on both `embedding_model.embed` and `vector_store.search` is the assertion that makes this a real test. Checking only that `ValueError` is raised would pass even if validation ran after the embedding call, and the whole point of validating a threshold is to fail before spending an embedding round trip and a vector query. Pinning the ordering is what stops a later refactor from moving the check below the work.

One thing to glance at while you are here: `mem0/utils/entity_extraction.py` carries a comparison of the same shape. If it guards a float that can reach it from user input it has the same gap; if it is an int or an internally computed value it is fine and there is nothing to do. Not part of this fix either way, just the only other place the pattern appears.

Anai-Guo:

ErenAta16 (commented):
Correct fix, and the sweep shows it is exactly as wide as the hole. Running the validator body from `main` against the version here:

```
value            main                     this PR
nan              accepted                 rejected
+inf             rejected                 rejected
-inf             rejected                 rejected
0 / 0.5 / 1      accepted                 accepted
1.5 / -0.1       rejected                 rejected
Decimal('nan')   rejected (type)          rejected (type)
'0.5'            rejected (type)          rejected (type)
```

The infinities were already covered, since `inf > 1` and `-inf < 0` both evaluate normally. NaN was the only value that could satisfy a range check by failing both comparisons:

```
nan < 0   -> False
nan > 1   -> False
```

so nothing else in the float domain needed adding, and the one-condition fix is the right size.

**Ordering is right and worth not disturbing.** `math.isnan` is placed after the `isinstance` guard, so a string still produces `threshold must be a valid number` rather than a `TypeError` out of `math`. If those two checks are ever merged or reordered for tidiness, that error message regresses.

**The fix covers async for free, which the PR does not mention.** `_validate_search_params` is called from both:

```
1410  Memory.search
3058  AsyncMemory.search
```

with `get_all` at 1256 and 2902 passing only `top_k`, correctly, since it has no threshold. So putting the check in the shared helper covers all four entry points. Worth saying in the description, and worth considering a parallel async test: `AsyncMemory.search` inherits the fix but nothing exercises it, and it is the sort of thing a later refactor could split without a failure.

**One scoping note on the issue's framing.** #6713 says the NaN "keeps candidates that should have been filtered". I could not confirm that from `main.py`, because the threshold is not applied there; it is forwarded to `self.vector_store.search(...)` and the comparison happens inside each backend. Which direction NaN breaks therefore depends on whether a given store filters with `score >= threshold` or `score < threshold`, and on whether it pushes the value down into the database instead. Both directions are silent and wrong:

```
score >= nan  -> False   for 0.9, 0.5, 0.01, -3.0
score <  nan  -> False   for the same
```

so one shape returns nothing and the other returns everything, and with a dozen supported stores it is probably both depending on configuration. That is the strongest argument for this PR's approach rather than a weakness in it: validating once at the boundary is the only place the answer does not depend on which vector store is plugged in. I would put that in the commit message instead of the specific downstream claim, which is only true for some backends.

**The test asserts the right thing.** Checking `embed.assert_not_called()` and `vector_store.search.assert_not_called()` pins that the rejection happens before any billable or I/O work, not just that it eventually raises. `_validate_search_params` sits at line 1410, ahead of embedding and retrieval, so that property holds and is now locked down.

Minor and pre-existing: `isinstance(True, int)` is true, so `threshold=True` is still accepted as 1.0 and `False` as 0.0. Not introduced here and arguably not worth a branch, just noting it since the type guard looks exhaustive at a glance and is not.

PR #6725, opened 2026-08-02
fix(memory): coerce bare-string LLM extractions so one bad doc can't abort ingestion (#6724)

### LINKED ISSUE #6724: _add_to_vector_store crashes on list-of-strings extraction — one bad doc aborts the whole ingestion
### Summary

`_add_to_vector_store` assumes every extracted memory is a `dict` with a `text` key. When the LLM returns `{"memory": ["fact one", "fact two"]}` — a list of plain strings — the comprehension raises `AttributeError`. **`add()` is one LLM call per document, so a single non-conforming extraction aborts an entire ingestion, and with the default ephemeral store everything already ingested is lost.**

### Versions — checked against latest, not just what I had installed

| | |
|---|---|
| Installed | `mem0ai==2.0.15` |
| Latest on PyPI at time of filing | **`2.0.15`** — same |
| Still present on `main` | **yes** — `mem0/memory/main.py:963` **and `:2611`** (sync and async paths) |

Python 3.12, Chroma vector store, Ollama LLM (`qwen2.5-coder-7b-instruct` q4_k_m) + `nomic-embed-text`, fully local.

### The line

```python
# mem0/memory/main.py:963  (and again at :2611)
mem_texts = [m.get("text", "") for m in extracted_memories if m.get("text")]
```

`extracted_memories` comes from `json.loads(response, strict=False).get("memory", [])` (lines 949/952). Nothing between the parse and this line constrains element type.

### To be fair to the prompt: it *does* specify dicts

I want to be precise, because it would be easy to mis-state this. `ADDITIVE_EXTRACTION_PROMPT` — the prompt on this code path — is explicit:

```
OUTPUT FORMAT
Return ONLY valid JSON parsable by json.loads().
{ "memory": [ {"id": "0", "text": "...", "attributed_to": "user"}, ... ] }
- **text** (string, required): A contextually rich, self-contained factual statement (15-80 words).
```

So a model returning bare strings **is** deviating from the documented contract. This is not "the parser requires a shape you never asked for."

### But the crashing shape is your *own* sibling prompt's schema

The same package ships `FACT_RETRIEVAL_PROMPT`, whose few-shot examples are:

```
Input: Hi, my name is John. I am a software engineer.
Output: {"facts" : ["Name is John", "Is a Software engineer"]}
```

A list of plain strings. So `{"...": ["string", "string"]}` is not an exotic hallucination — **it is the output schema of another prompt in the same codebase**, and it is the shape a smaller model is most likely to fall back to. That seems worth defending against rather than treating as impossible.

### Blast radius — why this is worse than a stray exception

This is the part that decides whether it matters:

1. `add()` invokes the LLM **once per document**. Ingesting a corpus is a long, expensive loop.
2. `_add_to_vector_store` raises, so `add()` raises, so **the caller's ingestion loop dies**.
3. `Memory`'s Chroma store is **ephemeral unless `path` is set**, so there is no partial persistence.

Net effect in our run: a non-conforming extraction roughly **60 documents into a 255-document ingestion ended the run and discarded everything already written — ~40 minutes of local LLM time, unrecoverable.** One malformed element out of several hundred documents cost the entire job.

A frontier model emits the documented schema reliably, which is likely why this hasn't surfaced. Locally it is reachable within a few hundred documents.

### Reproduction — deterministic, no LLM or network required

This is just the parse result flowing into the comprehension:

```python
import json
extracted = json.loads('{"memory": ["fact one", "fact two"]}').get("memory", [])
[x.get("text", "") for x in extracted if x.get("text")]
# AttributeError: 'str' object has no attribute 'get'
```

Real traceback from the corpus run:

```
File "mem0/memory/main.py", line 840, in add
    vector_store_result = self._add_to_vector_store(...)
File "mem0/memory/main.py", line 963, in _add_to_vector_store
    mem_texts = [m.get("text", "") for m in extracted_memories if m.get("text")]
AttributeError: 'str' object has no attribute 'get'
```

### Suggested fix

Normalise element type at both sites (`:963`, `:2611`):

```python
def _as_text(m):
    if isinstance(m, str):
        return m
    if isinstance(m, dict):
        return m.get("text", "") or m.get("fact", "") or m.get("memory", "")
    return ""

mem_texts = [t for t in (_as_text(m) for m in extracted_memories) if t]
```

Accepts the string form, unchanged for dicts, silently skips anything else. If you would rather stay strict, even `logger.warning` + skip would stop one bad element destroying an entire ingestion — that alone would have saved the 40 minutes.

Happy to send a PR for either.

### Related

Also filed #6723 (`add()`/`search()` `user_id` asymmetry). Both surfaced while measuring **token injection** — how many tokens a memory layer adds to a prompt on 16 GB consumer hardware — rather than retrieval quality. Both are invisible against a hosted frontier model.

Anai-Guo:
## What & why

Fixes #6724.

`_add_to_vector_store` assumes every element of the LLM's extracted-memory list is a `dict` with a `text` key. When the model returns `{"memory": ["fact one", "fact two"]}` — a list of **plain strings** — the Phase 3 comprehension raises:

```
AttributeError: 'str' object has no attribute 'get'
```

```python
# mem0/memory/main.py (sync :963, async :2611)
mem_texts = [m.get("text", "") for m in extracted_memories if m.get("text")]
```

The downstream record loop (`mem.get("text")`, `mem.get("attributed_to")`) would crash the same way. Because `add()` is one LLM call per document, a single non-conforming extraction aborts the **whole** ingestion, and with the default ephemeral store everything already ingested is lost.

## Fix

Add `normalize_extracted_memories()` in `mem0/memory/utils.py`, mirroring the existing `normalize_facts()` helper the codebase already uses to absorb inconsistent LLM shapes on the fact-retrieval path. It coerces bare strings into `{"text": ...}`, lifts a lone dict into a list, and drops non-dict/non-str junk (with a warning) instead of crashing. It's called at both the sync and async parse sites, right after the response is parsed and before any `m.get(...)` access — so a malformed element degrades gracefully and the well-formed facts are still recovered.

## Behaviour

| `extracted_memories` (default instructions) | before | after |
|---|---|---|
| `[{"text": "a"}, {"text": "b"}]` | `["a", "b"]` | `["a", "b"]` (unchanged) |
| `["fact one", "fact two"]` | **AttributeError** (ingestion aborts) | `["fact one", "fact two"]` |
| `[{"text": "a"}, "b", 5, None]` | **AttributeError** | `["a", "b"]` |
| `{"text": "solo"}` | `["solo"]` | `["solo"]` |

## Tests

Added `TestNormalizeExtractedMemories` in `tests/memory/test_memory_utils.py` covering the pass-through, string-wrapping, mixed-junk, single-dict, and non-iterable cases, plus the exact `main.py` comprehension that used to raise. `pytest tests/memory/test_memory_utils.py` → all green locally.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -53,6 +53,7 @@ from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
 from mem0.memory.utils import (
     extract_json,
+    normalize_extracted_memories,
     parse_messages,
     parse_vision_messages,
     process_telemetry_filters,
@@ -938,6 +939,10 @@ class Memory(MemoryBase):
         except Exception as e:
             logger.error(f"Error parsing extraction response: {e}")
             extracted_memories = []
+
+        # Coerce bare-string / non-list extractions into {"text": ...} dicts so a
+        # single non-conforming element cannot abort the whole ingestion (#6724).
+        extracted_memories = normalize_extracted_memories(extracted_memories)
 
         if not extracted_memories:
             # Save messages even if nothing extracted
@@ -2575,6 +2580,10 @@ class AsyncMemory(MemoryBase):
         except Exception as e:
             logger.error(f"Error parsing extraction response (async): {e}")
             extracted_memories = []
+
+        # Coerce bare-string / non-list extractions into {"text": ...} dicts so a
+        # single non-conforming element cannot abort the whole ingestion (#6724).
+        extracted_memories = normalize_extracted_memories(extracted_memories)
 
         if not extracted_memories:
             await asyncio.to_thread(self.db.save_messages, messages, session_scope)
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -123,7 +123,10 @@ def normalize_extracted_memories(raw_memories):
     """
     if not raw_memories:
         return []
-    if isinstance(raw_memories, dict):
+    if isinstance(raw_memories, (dict, str)):
+        # A model can also return the whole payload as a single object or a
+        # bare string ("memory": "hello"); wrap it so it normalizes the same
+        # as the corresponding single-element list.
         raw_memories = [raw_memories]
     elif not isinstance(raw_memories, list):
         return []
--- a/mem0/memory/utils.py
+++ b/mem0/memory/utils.py
@@ -112,6 +112,33 @@ def normalize_facts(raw_facts):
     return normalized
 
 
+def normalize_extracted_memories(raw_memories):
+    """Normalize LLM-extracted memories to a list of ``{"text": ...}`` dicts.
+
+    The additive-extraction prompt asks for ``{"memory": [{"text": ...}, ...]}``,
+    but a model can return bare strings (``{"memory": ["fact one", "fact two"]}``)
+    or a single object instead of a list. Coercing here keeps one malformed
+    element from aborting the whole ingestion with an ``AttributeError`` when the
+    downstream code calls ``m.get("text")``. Mirrors :func:`normalize_facts`.
+    """
+    if not raw_memories:
+        return []
+    if isinstance(raw_memories, dict):
+        raw_memories = [raw_memories]
+    elif not isinstance(raw_memories, list):
+        return []
+    normalized = []
+    for item in raw_memories:
+        if isinstance(item, dict):
+            normalized.append(item)
+        elif isinstance(item, str):
+            if item.strip():
+                normalized.append({"text": item})
+        else:
+            logger.warning("Unexpected memory shape from LLM, skipping: %s", item)
+    return normalized
+
+
 def remove_code_blocks(content: str) -> str:
     """
     Removes enclosing code block markers ```[language] and ``` from a given string.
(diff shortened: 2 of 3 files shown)


RedEye1605 (commented):
Nice focused fix, and I like that it covers both sync/async ingestion paths with regression tests for mixed model output.

Could the top-level string case be normalized consistently with a string inside a list? Right now `normalize_extracted_memories("hello")` returns `[]`, while `["hello"]` becomes `[{"text": "hello"}]`. Since the helper already accepts a single dict in place of a list—and the PR is intended to tolerate bare-string extraction—a model returning `"memory": "hello"` would silently discard a valid memory. Wrapping a non-blank top-level string (and updating that test expectation) would preserve it.

Anai-Guo:

Anai-Guo:
Good call — done in 38bc8b8. A top-level bare string now goes through the same wrapping path as a single dict, so `normalize_extracted_memories("hello")` returns `[{"text": "hello"}]`, identical to `["hello"]`. Empty/whitespace-only strings still normalize to `[]` (the leading falsy-guard and the per-item `strip()` check both cover that), and non-string scalars like `5` are still dropped. Added `test_bare_string_is_wrapped_like_single_element_list` and tightened the empty-input test to cover `""`, `"   "`, and `5`.

PR #6727, opened 2026-08-02
fix(vector_stores/faiss): derive expected module name in SafeUnpickler test

### LINKED ISSUE #6719: Windows: test_safe_unpickler_blocks_os_system asserts the POSIX module name for os.system
### Summary

`tests/vector_stores/test_faiss.py::TestSafeUnpickler::test_safe_unpickler_blocks_os_system` fails on Windows. The `SafeUnpickler` guard itself works correctly there; the test just hardcodes the POSIX name of the module `os.system` belongs to.

### Steps to Reproduce

```
pytest tests/vector_stores/test_faiss.py::TestSafeUnpickler::test_safe_unpickler_blocks_os_system
```

on any Windows machine.

### Expected Behavior

The test passes, since the guard does block the payload.

### Actual Behavior

```
    assert "Unsafe pickle" in str(exc_info.value)
>   assert "posix.system" in str(exc_info.value)
E   assert 'posix.system' in "Unsafe pickle: attempted to load 'nt.system'. Only basic
    Python types are allowed for security reasons."

tests\vector_stores\test_faiss.py:390: AssertionError
```

`SafeUnpickler` raised `UnpicklingError` as intended, so the security behaviour is fine. `os.system` lives in `posix` on Unix but in `nt` on Windows, and the assertion added in #4833 pins the Unix name. CI only runs ubuntu-latest, so this has never shown up there.

The neighbouring `test_safe_unpickler_blocks_subprocess` doesn't have the problem, because `subprocess.call` reports the same module name everywhere.

Happy to send a PR. Deriving the expected name from the same callable the payload is built from (`f"{os.system.__module__}.system"`) keeps the assertion as strict as it is today while working on both platforms, but if you'd rather just drop the module-name half and match the subprocess test, say so and I'll do that instead.

### Environment

- mem0 version: 2.0.14 (editable install at 38e47ac)
- Python version: 3.13.9
- OS: Windows 11 x64

Cyrus580529:
## Linked Issue

Closes #6719

## Description

`tests/vector_stores/test_faiss.py::TestSafeUnpickler::test_safe_unpickler_blocks_os_system` fails on Windows. `SafeUnpickler` itself is fine and blocks the payload exactly as intended; the test just pins the POSIX name of the module `os.system` belongs to:

```
assert "posix.system" in str(exc_info.value)
E   assert 'posix.system' in "Unsafe pickle: attempted to load 'nt.system'. ..."
```

`os.system` lives in `posix` on Unix and in `nt` on Windows. The assertion added in #4833 hardcodes the Unix name, and CI only runs ubuntu-latest, so this has never shown up there.

This derives the expected name from the same callable the payload is built from:

```python
assert f"{os.system.__module__}.system" in str(exc_info.value)
```

That keeps the assertion exactly as strict as it is today — it still requires the real module name, not just `.system` — while working on both platforms. I offered the alternative in the issue (drop the module-name half and match the neighbouring `test_safe_unpickler_blocks_subprocess`); happy to switch to that if you prefer it.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests (the change is itself the test)

On Windows 11 x64 / Python 3.13.9:

- Before: `tests/vector_stores/test_faiss.py::TestSafeUnpickler::test_safe_unpickler_blocks_os_system` fails as above.
- After: `pytest tests/vector_stores/test_faiss.py` → **33 passed**.

**On Linux this change is a no-op.** The new expression evaluates to the same string literal the test asserts today — checked on Ubuntu 22.04:

```
$ python3 -c "import os; print(repr(f'{os.system.__module__}.system'))"
'posix.system'
```

I did not run the full suite on Linux, so that is an equivalence check on the changed expression, not a full Linux CI reproduction. CI will cover the rest.

One deviation to flag: `AGENTS.md` says to use hatch rather than pip, but the hatch envs only cover Python 3.10–3.12 and this bug is only observable on Windows, so I ran it in a plain venv on 3.13 with `pip install -e ".[test,vector-stores,llms,extras]"`.

## Checklist

- [x] I have added tests that prove my fix/feature works (the fixed assertion is red before this change on Windows, green after)

`ruff check tests/vector_stores/test_faiss.py` passes. `ruff format --check` reports the file as needing reformatting, but that is pre-existing on a clean `main` (an unrelated `search(...)` call above), so I left it alone rather than mixing an unrelated reformat into this diff.

### DIFF
--- a/tests/vector_stores/test_faiss.py
+++ b/tests/vector_stores/test_faiss.py
@@ -387,7 +387,9 @@ class TestSafeUnpickler:
             SafeUnpickler(io.BytesIO(malicious_payload)).load()
 
         assert "Unsafe pickle" in str(exc_info.value)
-        assert "posix.system" in str(exc_info.value)
+        # os.system lives in `posix` on Unix and `nt` on Windows, so derive the
+        # expected name from the same callable the payload was built from.
+        assert f"{os.system.__module__}.system" in str(exc_info.value)
 
     def test_safe_unpickler_blocks_subprocess(self):
         """SafeUnpickler should block subprocess execution attempts."""


ErenAta16 (commented):
Confirmed on real Windows 11, which is the thing CI cannot tell you since the workflow only runs ubuntu-latest. Building the same payload and reading the message:

```
os.system.__module__                 -> 'nt'
message                              -> "Unsafe pickle: attempted to load 'nt.system'."

"posix.system" in message                          -> False    (current assertion)
f"{os.system.__module__}.system" in message        -> True     (this PR)
```

So the failure is real, it is the assertion rather than `SafeUnpickler`, and the fix is right. `os` is already imported at module scope (line 2), so the new expression needs no import.

**Why this one test and not its neighbours.** Worth recording, because the natural follow-up question is whether the sibling tests have the same latent problem. They do not, and for a reason that is not obvious:

```
os.system.__module__       nt          <- C builtin, re-exported from nt/posix
os.popen.__module__        os          <- pure-Python wrapper defined in os.py
subprocess.Popen.__module__  subprocess
builtins.eval.__module__     builtins
builtins.exec.__module__     builtins
```

`os.system` is a C function that genuinely lives in the platform module, so pickle records `nt.system` or `posix.system`. `os.popen` is a Python-level wrapper in `os.py`, so it reports `os` on every platform. That is the whole reason only this assertion is platform-dependent. Grepping the file confirms line 390 is the only hardcoded platform module name, so the PR is complete rather than fixing one instance of a pattern.

**Strictness is unchanged**, which is the part that matters for a security test. The assertion still requires the full `<module>.system`, so a message that only mentioned `.system`, or one that named a different callable, would still fail. It is not weakened to a substring like `"system"`.

Test-only change; `SafeUnpickler` is untouched and still blocks the payload identically on both platforms. Deriving the expected value from the same object the payload was built from is also the right call over a `sys.platform` branch, since it cannot drift if a future Python moves where `os.system` is defined.

kartik-mem0 (commented):

The old assertion hardcoded `posix.system`, so the test failed on Windows even though `SafeUnpickler` correctly blocked the payload there too. This derives the expected string from `os.system.__module__` at runtime, so it matches whatever the platform actually reports.

**Follow-up:** none
**Not verified:** actual Windows CI run (no Windows runner available here; verified by calling the unchanged `find_class("nt", "system")` directly instead, see evidence)

Risk: low, test-only change to an already-shipped file · Your time: ~2 min

- **Problem is real:** `mem0/vector_stores/faiss.py:47-60` `find_class` only allows `builtins`/`__builtin__`; `os.system.__module__` is `posix` on Unix and `nt` on Windows, so `tests/vector_stores/test_faiss.py:390` (pre-PR) asserting the literal `posix.system` fails on Windows even though the guard raises correctly there.
- **Tests:** `pytest tests/vector_stores/test_faiss.py -v` -> 33 passed. `ruff check` clean.
- **Both platforms:** called the unchanged `SafeUnpickler.find_class("nt", "system")` and `.find_class("posix", "system")` directly, both raise `UnpicklingError` with the respective module name in the message, confirming the new derivation matches on both platforms rather than swapping one hardcoded literal for another.
- **Still asserts BLOCKED:** the `pytest.raises(pickle.UnpicklingError)` wrapper is unchanged, if the guard ever let `os.system` through, `.load()` would actually run the shell command instead of raising and `pytest.raises` would fail with "DID NOT RAISE" -- the test remains a real regression check, not weakened.
- **CI:** no test/lint/build workflow has run yet for this head SHA (Actions likely awaiting maintainer approval for a first-time contributor); the only checks present (`label`, `license/cla`) are green. `Vercel` is red but advisory per repo policy.
- **Security:** cleared -- production `SafeUnpickler.find_class` is untouched and was already platform-independent; this diff only fixes the test's expected string.
- **Bloat (ponytail):** lean already, net 0 lines possible.
- **Linked issue:** #6719

</details>

Cyrus580529:
Any idea why CI never ran on this one?

ErenAta16 (commented):
Nothing is wrong with the branch, it is waiting on a maintainer click. I looked at the actual run records rather than guess.

Two workflows fired on `1f8bbf81`, and only one of them could:

```
CI Gate      completed   action_required
PR Labeler   completed   success
```

`ci-gate.yml` is `on: pull_request`, and for a pull request from a fork GitHub holds those runs at `action_required` until someone with write access approves them. `pr-labeler.yml` is `on: pull_request_target`, which executes in the base repository's own context and is never gated, which is why it ran and the other did not. So the single check run reporting on this PR is `label`, from the labeler.

That is also why the PR shows as blocked. From the header of `ci-gate.yml`:

> Branch protection should require exactly one status check: "CI Gate".

The one required check is precisely the one that cannot start, so the state is not about your code at all. Nobody can look at a red or green here until the run is approved.

One thing worth knowing since it landed after you opened this: `pr-gate.yml` was added with `GATE_EFFECTIVE_FROM: '2026-08-12T00:00:00Z'`, and it requires an outside contributor's PR to close an issue carrying the `accepted` label. It does not affect this PR:

```
#6727 created   2026-08-02T09:56:43Z
gate effective  2026-08-12T00:00:00Z
```

and the workflow returns early for anything opened before that timestamp. Worth having in mind for the next one though, because a PR opened now without a linked `accepted` issue will be stopped by that gate rather than by CI.

So there is nothing to fix or rebase from your side. The change is three lines in one test file and I verified the underlying failure on real Windows 11 earlier in this thread, which is the platform the ubuntu-only CI could not have caught anyway.

ErenAta16:
It did start, and then stopped before running anything. Querying the runs for your head `1f8bbf81`:

```
CI Gate      status=completed  conclusion=action_required  event=pull_request  created=2026-08-02
PR Labeler   status=completed  conclusion=success          event=pull_request_target
```

`action_required` means the run was created and is waiting for a maintainer to approve workflows for a fork PR. Until someone clicks that, none of the gate's jobs execute, which is why the checks list shows only `label` rather than the usual set. For comparison, a recent non-blocked PR (#7046) has 24 check runs including `CI Gate`, so nothing is wrong with the workflow itself or with your branch.

Two things follow from that.

Pushing another commit will not help. It creates a new run in the same `action_required` state. The only thing that clears it is a maintainer approving the run, so asking for that directly is more useful than rebasing.

The `Vercel` status showing failure with "Authorization required to deploy" is unrelated and expected on a fork. It is a preview deploy that cannot receive secrets from a fork PR, and it is not one of the required checks, so it is not worth chasing.

`ci-gate.yml` is the single required check and it triggers on every `pull_request` with no path filter, so once it is approved your `tests/vector_stores/test_faiss.py` change will be picked up by the Python SDK pipeline rather than skipped.

PR #6733, opened 2026-08-02
fix(vector_stores/azure_ai_search): guard __del__ against missing clients

### LINKED ISSUE #6732: bug(vector_stores/azure_ai_search): unguarded __del__ crashes during GC if init fails
## Component

## Summary

## Expected

anujbolewar:
## Linked Issue

Closes #6732

## Description

`AzureAISearch.__del__()` had no `try/except` or `hasattr` guard. If `__init__` raises before `search_client` or `index_client` is assigned (e.g., `DefaultAzureCredential()` fails, or line 98 accessing Azure SDK internals raises), `__del__` raises `AttributeError` during garbage collection.

Added `hasattr` guards matching the pattern used by mongodb, cassandra, and oracledb vector stores.

## Type of Change

## Breaking Changes

N/A

## Test Coverage

- [x] Added test that verifies `__del__` does not crash when clients are missing

## Checklist

### DIFF
--- a/mem0/vector_stores/azure_ai_search.py
+++ b/mem0/vector_stores/azure_ai_search.py
@@ -386,8 +386,13 @@ class AzureAISearch(VectorStoreBase):
 
     def __del__(self):
         """Close the search client when the object is deleted."""
-        self.search_client.close()
-        self.index_client.close()
+        try:
+            if hasattr(self, "search_client"):
+                self.search_client.close()
+            if hasattr(self, "index_client"):
+                self.index_client.close()
+        except Exception:
+            pass
 
     def reset(self):
         """Reset the index by deleting and recreating it."""
--- a/mem0/vector_stores/azure_ai_search.py
+++ b/mem0/vector_stores/azure_ai_search.py
@@ -389,6 +389,9 @@ class AzureAISearch(VectorStoreBase):
         try:
             if hasattr(self, "search_client"):
                 self.search_client.close()
+        except Exception:
+            pass
+        try:
             if hasattr(self, "index_client"):
                 self.index_client.close()
         except Exception:
--- a/tests/vector_stores/test_azure_ai_search.py
+++ b/tests/vector_stores/test_azure_ai_search.py
@@ -719,3 +719,9 @@ def test_build_filter_escapes_quotes(azure_ai_search_instance):
     instance, _, _ = azure_ai_search_instance
     expr = instance._build_filter_expression({"name": "O'Brien"})
     assert "name eq 'O''Brien'" in expr
+
+
+def test_del_does_not_crash_when_clients_are_missing():
+    """__del__ must not raise if __init__ failed before clients were assigned."""
+    instance = object.__new__(AzureAISearch)
+    instance.__del__()  # should not raise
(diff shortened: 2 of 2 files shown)


mem0/vector_stores/azure_ai_search.py
```diff
@@ -386,8 +386,13 @@ def list(self, filters=None, top_k=100):
 
     def __del__(self):
         """Close the search client when the object is deleted."""
-        self.search_client.close()
-        self.index_client.close()
+        try:
+            if hasattr(self, "search_client"):
+                self.search_client.close()
```
kartik-mem0, line 391: If this raises, the except below swallows it and `index_client.close()` on line 393 never runs, leaking that client. Split into two try/except blocks, one per close() call, so a failure on one does not block the other.

tests/vector_stores/test_azure_ai_search.py
```diff
@@ -719,3 +719,9 @@ def test_build_filter_escapes_quotes(azure_ai_search_instance):
     instance, _, _ = azure_ai_search_instance
     expr = instance._build_filter_expression({"name": "O'Brien"})
     assert "name eq 'O''Brien'" in expr
+
+
+def test_del_does_not_crash_when_clients_are_missing():
+    """__del__ must not raise if __init__ failed before clients were assigned."""
+    instance = object.__new__(AzureAISearch)
+    instance.__del__()  # should not raise
```
kartik-mem0, line 727: This only covers both clients missing entirely. Add a case where search_client is present and .close() raises, then assert index_client.close() is still called.

kartik-mem0 (commented):
## Changes requested

Fixes the reported AttributeError in `__del__`, but the shared try/except can silently skip closing `index_client` if `search_client.close()` raises.

**To unblock:**
1. `mem0/vector_stores/azure_ai_search.py:389-395` - split into two try/except blocks, one per client (detail inline)
2. `tests/vector_stores/test_azure_ai_search.py:724-727` - add a case where `search_client.close()` raises and `index_client.close()` still runs (detail inline)

- **Problem is real:** reproduces on `main` at `mem0/vector_stores/azure_ai_search.py:389`, when `__init__` raises before `search_client`/`index_client` are assigned (for example a `DefaultAzureCredential()` failure)
- **Tests / red-green:** `pytest tests/vector_stores/test_azure_ai_search.py` -> 28 passed; reverted the implementation and reran the new test alone -> failed with the exact `AttributeError` from `main`; restored the fix -> 28 passed again
- **CI:** only the Vercel preview check is red, no deploy permissions from forks, not a required check; no required test, lint, or build check is red
- **Security:** cleared, no new attack surface, `__del__` takes no external input
- **Bloat (ponytail):** lean already, a five-line guard plus one focused test
- **Not verified:** live Azure AI Search behavior, this is a hosted-only backend with no local sandbox lane

</details>

The hasattr guards and the regression test are exactly right for the reported crash. Push when ready and I will re-review.

anujbolewar:
Thanks for the review @kartik-mem0! Pushed the requested changes in 2af51b6.

- Split the shared try/except in `__del__` into two separate blocks, one per client, so a failure in `search_client.close()` no longer blocks `index_client.close()` from running.
- Added `test_del_closes_both_clients_even_if_one_raises`, which sets `search_client.close()` to raise and asserts `index_client.close()` is still called.

kartik-mem0 (commented):
Re-reviewed at `2af51b67`. Content-identical rebase of the previously reviewed head. Fresh tests and lint pass against the new SHA. LGTM still holds.

PR #6735, opened 2026-08-02
fix(embeddings/ollama): correct default embedding_dims from 512 to 768

### LINKED ISSUE #6734: bug(embeddings/ollama): default embedding_dims=512 mismatches default model nomic-embed-text (768)
## Component

Python SDK — mem0/embeddings/ollama.py

## Summary

`OllamaEmbedding` defaults to model `nomic-embed-text` (line 28) but sets `embedding_dims=512` (line 29). The `nomic-embed-text` model produces **768-dimensional** embeddings.

When a user creates an `OllamaEmbedding` without specifying `embedding_dims`, the vector store is initialized with the wrong dimension count, causing dimension mismatch errors at search time.

## Expected

`embedding_dims` should default to `768` to match the default model.

anujbolewar:
## Linked Issue

Closes #6734

## Description

`OllamaEmbedding` defaults to model `nomic-embed-text` (line 28) which produces **768-dimensional** embeddings, but `embedding_dims` was set to `512` (line 29). This causes dimension mismatch errors when users rely on the defaults without specifying `embedding_dims`.

Changed the default from 512 to 768 to match the default model.

## Type of Change

## Breaking Changes

N/A — only affects users who rely on the undocumented default dims value.

## Test Coverage

- [x] Added test verifying default dims match the default model
- [x] All 8 ollama embedding tests pass locally

## Checklist

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -26,7 +26,7 @@ class OllamaEmbedding(EmbeddingBase):
         super().__init__(config)
 
         self.config.model = self.config.model or "nomic-embed-text"
-        self.config.embedding_dims = self.config.embedding_dims or 512
+        self.config.embedding_dims = self.config.embedding_dims or 768
 
         self.client = Client(host=self.config.ollama_base_url)
         self._ensure_model_exists()
--- a/tests/embeddings/test_ollama_embeddings.py
+++ b/tests/embeddings/test_ollama_embeddings.py
@@ -94,3 +94,10 @@ def test_embed_batch_count_mismatch_raises(mock_ollama_client):
 
     with pytest.raises(ValueError, match="returned 1 embeddings for 2 texts"):
         embedder.embed_batch(["first text", "second text"])
+
+
+def test_default_embedding_dims_matches_default_model(mock_ollama_client):
+    """Default embedding_dims should be 768 to match nomic-embed-text."""
+    config = BaseEmbedderConfig(model="nomic-embed-text")
+    embedder = OllamaEmbedding(config)
+    assert embedder.config.embedding_dims == 768


mem0/embeddings/ollama.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, config: Optional[BaseEmbedderConfig] = None):
         super().__init__(config)
 
         self.config.model = self.config.model or "nomic-embed-text"
-        self.config.embedding_dims = self.config.embedding_dims or 512
+        self.config.embedding_dims = self.config.embedding_dims or 768
```
kartik-mem0, line 29: This fallback only matches the default model. A caller who sets a different Ollama embedding model without also setting embedding_dims still gets the wrong value here (now 768 instead of 512). Pre-existing gap, not introduced by this fix, worth a follow-up guard or doc note.

tests/embeddings/test_ollama_embeddings.py
```diff
@@ -94,3 +94,10 @@ def test_embed_batch_count_mismatch_raises(mock_ollama_client):
 
     with pytest.raises(ValueError, match="returned 1 embeddings for 2 texts"):
         embedder.embed_batch(["first text", "second text"])
+
+
+def test_default_embedding_dims_matches_default_model(mock_ollama_client):
+    """Default embedding_dims should be 768 to match nomic-embed-text."""
+    config = BaseEmbedderConfig(model="nomic-embed-text")
```
kartik-mem0, line 101: Pins model explicitly rather than using a fully default config. Still proves the fix (red-green checked), but a zero-arg BaseEmbedderConfig() would exercise the exact default path issue #6734 describes. Consider asserting embedder.config.model too.

kartik-mem0 (commented):
## Changes requested

Fixes the wrong default (768 for nomic-embed-text, was 512, issue #6734), but leaves the docs contradicting the fix it just made.

**To unblock:**
1. `docs/components/embedders/models/ollama.mdx:68`: update the Python `embedding_dims` default from 512 to 768 to match the code and the already-correct TypeScript tab on the same page.

Optional: `mem0/embeddings/ollama.py:29` and `tests/embeddings/test_ollama_embeddings.py:101`, see inline notes.

Companion PR #6752 fixes the same line the same way and has the same docs gap. Whichever merges first, close the other.

- **Problem is real:** `mem0/embeddings/ollama.py:17` on `main`, `nomic-embed-text` (the hardcoded default model) returns 768 dim vectors but `embedding_dims` fell back to 512 when unset, mismatching on the first insert for any default-config caller.
- **Tests / red-green:** `hatch -e dev_py_3_12 run pytest tests/embeddings/test_ollama_embeddings.py -v --tb=short` at PR head: 8 passed. Implementation reverted to `main`, same command: new test failed on `assert 512 == 768`. Restored to PR head: 8 passed again. The test bites the change.
- **CI:** only `label` and `license/cla` have reported, the test/lint/build lane has not run yet (pending workflow approval). Vercel fail is a deploy preview and does not block. Judged this PR on the targeted run above instead.
- **Security:** cleared, config default change only. Traced the insert path in `mem0/vector_stores/qdrant.py` (the default OSS store): a collection is sized once at creation and every `upsert` after strictly enforces that size, and `OllamaEmbedding.embed()` never reshapes its output, so a user who was truly running pure defaults would already have hit a dimension error on their first `add()` before this fix, not a silent break after it.
- **Bloat:** lean already, 8 lines across 2 files, nothing to trim.
- **Not verified:** live Ollama backend (mocked client only, matching the existing test file pattern); pgvector/Chroma's exact insert time dimension mismatch behavior (inferred from their schema level typing, not run).

</details>

Test is solid and red-green proven, keep it as is. Push the doc line when ready and I will re-review.

PR #6737, opened 2026-08-02
fix(llms,embeddings): pass http_client to OpenAI client when proxies set

### LINKED ISSUE #6651: bug(llms+embeddings/openai): http_client_proxies builds a client that is never passed to OpenAI()
## Component
Python SDK — `mem0/llms/openai.py`, `mem0/embeddings/openai.py` (and sibling OpenAI-compatible LLM clients that only store proxies on config)

## Summary
`BaseLlmConfig` / `BaseEmbedderConfig` build `config.http_client` via `build_http_client(http_client_proxies)`, but `OpenAILLM` and `OpenAIEmbedding` construct `openai.OpenAI(...)` **without** `http_client=config.http_client`.

Result: proxy settings appear to work (config holds an httpx client) but all traffic bypasses the proxy.

Azure OpenAI embedder already forwards `http_client=self.config.http_client`. OpenAI paths do not.

## Repro
```python
from mem0.llms.openai import OpenAILLM

cfg = BaseLlmConfig(model="gpt-4o-mini", api_key="sk-test", http_client_proxies="http://127.0.0.1:9")
llm = OpenAILLM(cfg)
assert llm.config.http_client is not None
# OpenAI SDK client has no http_client / does not use cfg.http_client
assert getattr(llm.client, "_client", None) is not cfg.http_client
```

## Expected
`OpenAI(..., http_client=config.http_client)` when `config.http_client` is set (same for OpenAI embeddings).

## Actual
Proxy client is discarded; requests go direct.

Diwak4r:
## Summary
`OpenAILLM` and `OpenAIEmbedding` built their `OpenAI` client without `http_client=`, so the `httpx.Client` built from `http_client_proxies` (in `mem0/utils/http.py`) was silently discarded. Users configuring proxies/SSL via `http_client_proxies` got an OpenAI client that never routed through it.

This passes `self.config.http_client` through to the `OpenAI(...)` constructor in both classes, mirroring the existing `azure_openai` implementations.

## Changes
- `mem0/llms/openai.py` — pass `http_client=self.config.http_client` in the non-OpenRouter branch
- `mem0/embeddings/openai.py` — pass `http_client=self.config.http_client`
- Regression tests for both asserting `client._client is config.http_client` when `http_client_proxies` is set

## Test plan
- `pytest tests/llms/test_openai.py tests/embeddings/test_openai_embeddings.py` — 32 passed
- `ruff check` on changed files — clean

Closes #6651

### DIFF
--- a/mem0/embeddings/openai.py
+++ b/mem0/embeddings/openai.py
@@ -32,7 +32,11 @@ class OpenAIEmbedding(EmbeddingBase):
                 DeprecationWarning,
             )
 
-        self.client = OpenAI(api_key=api_key, base_url=base_url)
+        self.client = OpenAI(
+            api_key=api_key,
+            base_url=base_url,
+            http_client=self.config.http_client,
+        )
 
     def embed(self, text, memory_action: Optional[Literal["add", "search", "update"]] = None):
         """
--- a/mem0/llms/openai.py
+++ b/mem0/llms/openai.py
@@ -50,7 +50,11 @@ class OpenAILLM(LLMBase):
             api_key = self.config.api_key or os.getenv("OPENAI_API_KEY")
             base_url = self.config.openai_base_url or os.getenv("OPENAI_BASE_URL") or "https://api.openai.com/v1"
 
-            self.client = OpenAI(api_key=api_key, base_url=base_url)
+            self.client = OpenAI(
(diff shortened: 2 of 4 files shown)


Diwak4r:

PR #6739, opened 2026-08-02
fix(mem0-plugin): anchor global-search filters with a positive entity id

### LINKED ISSUE #6738: mem0-plugin: global_search sends wildcard-only filters the API rejects, silently breaking all searches
## Summary

With `global_search` enabled in `~/.mem0/settings.json`, the Claude Code plugin (`integrations/mem0-plugin`) builds search/list filters as `{"OR": [{"user_id": "*"}]}`. The platform API rejects this shape:

```
400: filters must include at least one positively-scoped entity ID (user_id, agent_id, app_id, or run_id). Wildcards and NOT-only entity references do not count as scoped.
```

## Impact

Three injection sites send the invalid shape, each failing differently:

1. `scripts/_search.py:65` — hook-driven context searches. The surrounding `except Exception: return []` swallows the 400, so auto-injected context silently comes back empty with no error anywhere.
2. `scripts/session_timeline.py:35` — recent-memories list fetch, same silent failure.
3. `scripts/enforce_metadata_defaults.sh:183` — this PreToolUse hook **overwrites the filters of every MCP `search_memories`/`get_memories` call** with the wildcard shape, so the MCP tool returns the 400 even when the caller passed correctly scoped filters. From the user's side it looks like the hosted MCP server is broken.

## Repro

```bash
curl -s -X POST https://api.mem0.ai/v3/memories/search/ \
  -H "Authorization: Token $MEM0_API_KEY" -H "Content-Type: application/json" \
  -d '{"query": "anything", "filters": {"OR": [{"user_id": "*"}]}}'
# -> 400 (wildcard-only)

curl -s -X POST https://api.mem0.ai/v3/memories/search/ \
  -H "Authorization: Token $MEM0_API_KEY" -H "Content-Type: application/json" \
  -d '{"query": "anything", "filters": {"OR": [{"user_id": "myuser"}, {"agent_id": "*"}]}}'
# -> 200 (wildcard is fine when a positive entity id anchors the OR)
```

## Suggested fix

Anchor the global-search OR with the resolved user id: `{"OR": [{"user_id": <uid>}, {"agent_id": "*"}]}`. Verified against `api.mem0.ai` that this passes validation while keeping cross-project + agent breadth. PR incoming.

Separately worth considering: the bare `except Exception: return []` in `_search.py` made this invisible — logging the failure (even to stderr) would have surfaced the 400 immediately.

Screddyice:
Fixes #6738

## Problem

With `global_search` enabled, the plugin builds filters as `{"OR": [{"user_id": "*"}]}`, which the platform API rejects with `400: filters must include at least one positively-scoped entity ID`. Three sites send this shape:

- `scripts/_search.py` — hook context searches silently return `[]` (the 400 is swallowed by the bare except)
- `scripts/session_timeline.py` — recent-memories fetch, same silent failure
- `scripts/enforce_metadata_defaults.sh` — overwrites the filters of every MCP `search_memories`/`get_memories` call, so the MCP tool fails even with correctly scoped caller filters

## Fix

Anchor the OR with the resolved user id: `{"OR": [{"user_id": <uid>}, {"agent_id": "*"}]}`. The API accepts a wildcard branch when a positive entity id anchors the filter (verified against `api.mem0.ai`), so global search keeps cross-project + agent breadth. In the shell hook the widened filter is only applied when a resolved user id exists; otherwise it falls through to standard identity injection rather than sending a guaranteed-400 shape.

## Testing

- New regression test asserting the global filter shape is positively scoped and contains no wildcard user id
- Full plugin suite: 160 passed
- Live verification against `api.mem0.ai`: old shape 400s, new shape returns hits; MCP `search_memories` works again end-to-end with the patched hook

## Docs

README gains a short "Global search scoping" note documenting the widened filter shape and its bound (all of the user's projects + agent-scoped memories; cannot span other user ids).


integrations/mem0-plugin/scripts/_search.py
```diff
@@ -62,7 +62,9 @@ def search_memories(
         return []
 
     if global_search:
-        filters: dict = {"OR": [{"user_id": "*"}]}
+        # The API rejects wildcard-only filters ("filters must include at least
+        # one positively-scoped entity ID"); a positive user_id keeps the OR valid.
+        filters: dict = {"OR": [{"user_id": user_id}, {"agent_id": "*"}]}
```
kartik-mem0: This filter literal is duplicated in `session_timeline.py:36` and `enforce_metadata_defaults.sh:185`, and a 4th identical site was missed: `scripts/on_session_start.sh:97-98` still sends `{"OR": [{"user_id": "*"}]}` and keeps 400ing (caught silently, shows `memories=?` in the session banner). Worth factoring into one `global_search_filter(user_id)` helper, e.g. in `_identity.py`, so every caller gets the fix by construction, including that one.

integrations/mem0-plugin/README.md
```diff
@@ -277,6 +277,10 @@ Once installed, the following tools are available:
 | `delete_entities` | Delete a user/agent/app/run entity and its memories |
 | `list_entities` | List users/agents/apps/runs stored in Mem0 |
 
+### Global search scoping
+
+With `global_search` enabled in `~/.mem0/settings.json`, the plugin widens search and list filters beyond the current project. The widened filter is `{"OR": [{"user_id": <your user id>}, {"agent_id": "*"}]}` — the platform API requires at least one positively-scoped entity ID in every filter and rejects wildcard-only filters, so global search covers all of your user's memories across projects plus agent-scoped memories. It cannot span other user IDs on the same account.
```
kartik-mem0: This claims global search "cannot span other user IDs on the same account," but the `agent_id: "*"` branch isn't scoped by `user_id` at all: a live query anchored on one `user_id` returned a record with `user_id: null` via that branch. It also contradicts `skills/switch-project/SKILL.md:13` ("all memories, all users, all projects"). Either drop the `agent_id` clause (a bare `{"user_id": user_id}` already satisfies the platform's one-positive-ID rule) or fix this claim.

integrations/mem0-plugin/scripts/session_timeline.py
```diff
@@ -32,7 +32,8 @@ def fetch_recent_memories(api_key: str, user_id: str, project_id: str) -> list[d
     global_search = os.environ.get("MEM0_GLOBAL_SEARCH", "false") == "true"
 
     if global_search:
-        filters = {"OR": [{"user_id": "*"}]}
+        # Wildcard-only filters fail API validation; anchor the OR with the user id.
+        filters = {"OR": [{"user_id": user_id}, {"agent_id": "*"}]}
```
kartik-mem0: No regression test covers this call site or `enforce_metadata_defaults.sh:185`; only the equivalent change in `_search.py` is tested.

integrations/mem0-plugin/tests/test_search.py
```diff
@@ -175,3 +175,27 @@ def test_format_results_for_context():
     assert "[decision]" in output
     assert "Use Postgres for auth" in output
     assert "abc12345" in output
+
+
+def test_global_search_filters_are_positively_scoped():
```
kartik-mem0, line 217: This mocks `urllib.request.urlopen` entirely, so it proves the outgoing payload shape only, not that the platform actually accepts it. Worth a one-line docstring note so a future green run here isn't mistaken for proof the 400 is gone (platform acceptance was confirmed separately via a live curl call).

kartik-mem0 (commented):
## Changes requested

Correctly anchors the filter at the 3 sites it touches, but the same bug ships unfixed at a 4th call site and in a whole sibling package.

**To unblock:**
1. `scripts/on_session_start.sh:97` — same wildcard-only filter, still 400s (session banner shows `memories=?`)
2. `.opencode-plugin/opencode-mem0.ts:231,646,765,806,924,982` — 6 more unfixed sites, ships the same bug to OpenCode users
3. `README.md:282` — "cannot span other user IDs" is contradicted by live evidence (see below)

Optional: `session_timeline.py`/`enforce_metadata_defaults.sh` have no regression test; the new test's docstring should note it's shape-only.

- **Problem is real:** reproduces on `main`, confirmed live: POST `api.mem0.ai/v3/memories/search/` with `{"OR": [{"user_id": "*"}]}` returns 400 "filters must include at least one positively-scoped entity ID."
- **Fix works where applied:** live POST with the PR's shape `{"OR": [{"user_id": "<id>"}, {"agent_id": "*"}]}` returns 200. All 3 touched sites (`_search.py:67`, `session_timeline.py:36`, `enforce_metadata_defaults.sh:185`) use it correctly.
- **Root-cause completeness fails:** `scripts/on_session_start.sh:97-98` (untouched) and `.opencode-plugin/opencode-mem0.ts` (untouched sibling npm package, 6 sites) still send the rejected shape. PR description says "three call sites"; the plugin family has 10.
- **Tests / red-green:** `pytest tests/test_search.py -q` -> 12 passed. Reverting the three fixed files makes the new test fail. Test mocks `urllib.request.urlopen` entirely, so it proves payload shape only; the live curl calls above close that gap.
- **CI:** shows green, but the plugin's own pytest job never actually ran against this PR (its path-filter wiring in `ci-gate.yml` landed on `main` after this PR's CI ran) — treat it as uninformative, not corroborating.
- **Security:** no injection/SSRF/deserialization issue. The `agent_id: "*"` branch is unscoped by `user_id` and can surface memories the calling user doesn't own within the same account; not a cross-account leak, but the README's scoping claim is inaccurate.
- **Bloat (ponytail):** the filter-dict literal is duplicated across `_search.py`, `session_timeline.py`, `enforce_metadata_defaults.sh` (and would be a 4th time in `on_session_start.sh` if fixed). One `global_search_filter(user_id)` helper, e.g. in `_identity.py`, following the cross-language-sharing pattern the bash hook already uses (`PYTHONPATH="$SCRIPT_DIR" python3 -c "from load_settings import load_settings..."`) would make misses like the 4th site structurally impossible. net: -6 lines possible.
- **Not verified:** live checks covered `/v3/memories/search/` only; `session_timeline.py`/`on_session_start.sh` hit `/v3/memories/` (list), same validation is very likely but not independently confirmed there.

</details>

The three sites it does fix are correct and live-verified, real work. Push once the 4th site (and ideally the OpenCode sibling) are covered and I will re-review.

Screddyice:
All three blockers addressed in 05af12b, plus the optional items. Thanks for the
live verification — the `on_session_start.sh` miss and the README contradiction were
both real.

**Root cause of the miss, fixed structurally.** You were right that the duplication was
the actual defect: with the literal copy-pasted per site, "fix the bug" meant "find every
copy", and one got away. There is now a single definition on each side —
`global_search_filter()` in `scripts/_identity.py` and `globalSearchFilter()` in
`.opencode-plugin/opencode-mem0.ts` — and all ten sites route through them.

| Item | Status |
|---|---|
| `on_session_start.sh:97` (4th site) | fixed — inline python imports the helper over `PYTHONPATH`, matching the `load_settings` pattern `_identity.sh` already uses |
| `.opencode-plugin/opencode-mem0.ts` (6 sites) | all six fixed via `globalSearchFilter()` |
| `README.md:282` | claim removed — see below |
| helper in `_identity.py` | done, as suggested |
| tests for `session_timeline.py` / `enforce_metadata_defaults.sh` | added |
| docstring note on the mock-only test | added |

**On the README.** I went with correcting the doc rather than dropping the `agent_id`
clause. Dropping it would satisfy the one-positive-ID rule, but it would also make global
search stop being global — `skills/switch-project/SKILL.md:13` documents the mode as "all
memories, all users, all projects", which is the behaviour your live query observed. So
the code is right and the README was wrong. It now states the scope plainly: `agent_id: "*"`
is not constrained by `user_id`, so global search can return agent-scoped memories owned by
other user IDs on the same account — intended, worth knowing before enabling on a shared
account, and still account-scoped.

`enforce_metadata_defaults.sh` keeps an inline fallback definition, so a missing helper
degrades to current behaviour instead of breaking the hook.

**Verification:** 15 tests pass (was 12), `bash -n` clean on both hooks, esbuild parses the
OpenCode plugin, and a repo-wide grep confirms no wildcard-only filter shape remains. I did
not re-run the live API checks — your curl results already cover platform acceptance for
this exact shape, and the helper emits it unchanged.

Noting the CLA is still unsigned; that one needs the account owner and is on my side to sort.

PR #6746, opened 2026-08-03
fix(vector_stores): dispatch add_embeddings by signature in Langchain store

### LINKED ISSUE #6657: LangChain vector store: insert() calls add_embeddings() with a keyword no store accepts, so Memory.add() raises TypeError

Diwak4r:
## Linked Issue

Closes #6657

## Description

`insert()` called `add_embeddings(embeddings=vectors, metadatas=payloads, ids=ids)`, but no `langchain_community` store accepts an `embeddings=` keyword. The real signatures vary:

- FAISS / OpenSearch take `add_embeddings(text_embeddings=[(text, vector), ...], ...)`
- PGVector and the pgvector family take `add_embeddings(texts=..., embeddings=..., ...)` with both required
- Chroma has no `add_embeddings` at all

So every `Memory.add()` against a LangChain-backed store with precomputed embeddings raised `TypeError`. The fix inspects the method's parameter names and dispatches to the matching call shape, falling back to `add_texts` for unknown signatures.

`update(vector=None)` previously forwarded `None` as the embedding after `delete()` had already run, so the store either crashed or embedded `None` and the row was lost. It now re-inserts via `add_texts`, letting the store embed the payload text itself.

## Type of Change

## Test Coverage

Tests cover the FAISS-family and PGVector-family `add_embeddings` signatures, the `add_texts` fallback (Chroma-style stores), unknown-signature fallback, and `update(vector=None)`. The new tests fail against the pre-fix code (verified: the original `embeddings=` call raises `TypeError` on a PGVector-style mock).

## Checklist

### DIFF
--- a/mem0/embeddings/openai.py
+++ b/mem0/embeddings/openai.py
@@ -32,7 +32,11 @@ class OpenAIEmbedding(EmbeddingBase):
                 DeprecationWarning,
             )
 
-        self.client = OpenAI(api_key=api_key, base_url=base_url)
+        self.client = OpenAI(
+            api_key=api_key,
+            base_url=base_url,
+            http_client=self.config.http_client,
+        )
 
     def embed(self, text, memory_action: Optional[Literal["add", "search", "update"]] = None):
         """
--- a/mem0/llms/openai.py
+++ b/mem0/llms/openai.py
@@ -50,7 +50,11 @@ class OpenAILLM(LLMBase):
             api_key = self.config.api_key or os.getenv("OPENAI_API_KEY")
             base_url = self.config.openai_base_url or os.getenv("OPENAI_BASE_URL") or "https://api.openai.com/v1"
 
-            self.client = OpenAI(api_key=api_key, base_url=base_url)
+            self.client = OpenAI(
+                api_key=api_key,
+                base_url=base_url,
+                http_client=self.config.http_client,
+            )
 
     def _parse_response(self, response, tools):
         """
--- a/mem0/vector_stores/langchain.py
+++ b/mem0/vector_stores/langchain.py
@@ -1,3 +1,4 @@
+import inspect
 import logging
 from typing import Dict, List, Optional
 
@@ -26,8 +27,8 @@ class OutputData(BaseModel):
 # but exposed by several concrete implementations.
 _SCORED_BY_VECTOR_METHODS = [
     "similarity_search_by_vector_with_relevance_scores",  # Chroma
-    "similarity_search_with_score_by_vector",             # FAISS, Qdrant
-    "similarity_search_by_vector_with_score",             # Pinecone, YDB
+    "similarity_search_with_score_by_vector",  # FAISS, Qdrant
+    "similarity_search_by_vector_with_score",  # Pinecone, YDB
 ]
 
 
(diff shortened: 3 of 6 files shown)


Diwak4r:
Replaced by #6768 — this branch carried two unrelated fixes; the http_client change duplicates #6737 (closes #6651), so the branch was rebuilt to contain only the langchain dispatch fix for #6657. New PR: #6768.

PR #6747, opened 2026-08-03
fix(vector_stores): map '*' filter to field-exists in Pinecone

### LINKED ISSUE #6652: bug(vector_stores/pinecone): '*' metadata filter is $eq-matched instead of field-exists
## Component
Python SDK — `mem0/vector_stores/pinecone.py` `_create_filter`

## Summary
Mem0 metadata filtering documents `{"key": "*"}` as wildcard / field-exists. Pinecone's `_create_filter` always wraps scalars as `{"$eq": value}`, so `"*"` becomes a literal equality match for the string `*`, which never matches real payload values.

Other stores (pgvector, chroma, milvus, opensearch, …) special-case `"*"`.

## Repro
```python
# unit-level
from mem0.vector_stores.pinecone import PineconeDB
# after constructing (or call _create_filter on an instance):
# filters={"user_id": "*"} → {"user_id": {"$eq": "*"}}  # wrong
# expected field-exists, e.g. {"user_id": {"$exists": True}}
```

## Expected
`"*" ` → Pinecone exists filter (`$exists: true`), consistent with docs / other providers (see #6539 family).

## Actual
Literal `$eq` on `"*"`.

Diwak4r:
## Linked Issue

Closes #6652

## Description

`_create_filter` wrapped every scalar filter value as `{"$eq": value}`, so a documented `"*"` wildcard filter became a literal equality match for the string `"*"`. No real payload value equals the literal `"*"`, so `filters={"user_id": "*"}` returned nothing instead of everything with a `user_id` field.

The fix translates a `"*"` value to `{"$exists": True}` (Pinecone's field-exists filter), matching the behavior already implemented in pgvector, chroma, and milvus.

## Type of Change

## Test Coverage

Added `test_create_filter_wildcard_field_exists`, which asserts `{"user_id": "*"}` becomes `{"user_id": {"$exists": True}}`. Verified the test fails against the pre-fix code (it produced `{"$eq": "*"}`).

## Checklist

### DIFF
--- a/mem0/embeddings/openai.py
+++ b/mem0/embeddings/openai.py
@@ -32,7 +32,11 @@ class OpenAIEmbedding(EmbeddingBase):
                 DeprecationWarning,
             )
 
-        self.client = OpenAI(api_key=api_key, base_url=base_url)
+        self.client = OpenAI(
+            api_key=api_key,
+            base_url=base_url,
+            http_client=self.config.http_client,
+        )
 
     def embed(self, text, memory_action: Optional[Literal["add", "search", "update"]] = None):
         """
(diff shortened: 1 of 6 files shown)


Diwak4r:

PR #6752, opened 2026-08-03
fix(embeddings): default OllamaEmbedding dims to 768 for nomic-embed-text

### LINKED ISSUE #6734: bug(embeddings/ollama): default embedding_dims=512 mismatches default model nomic-embed-text (768)

Diwak4r:
Closes #6734

The default model for OllamaEmbedding is \
omic-embed-text\, which produces 768-dimensional embeddings, but the default \embedding_dims\ was 512. When a user instantiates OllamaEmbedding without setting embedding_dims, the config advertises 512 dims while the actual vectors are 768 — causing downstream mismatches (collection dimension errors, wrong chunking).

Fix the default to 768 and add a regression test asserting the default dims match the default model.

### DIFF
--- a/mem0/embeddings/ollama.py
+++ b/mem0/embeddings/ollama.py
@@ -26,7 +26,7 @@ class OllamaEmbedding(EmbeddingBase):
         super().__init__(config)
 
         self.config.model = self.config.model or "nomic-embed-text"
-        self.config.embedding_dims = self.config.embedding_dims or 512
+        self.config.embedding_dims = self.config.embedding_dims or 768
 
         self.client = Client(host=self.config.ollama_base_url)
         self._ensure_model_exists()
--- a/tests/embeddings/test_ollama_embeddings.py
+++ b/tests/embeddings/test_ollama_embeddings.py
@@ -43,6 +43,15 @@ def test_ensure_model_exists(mock_ollama_client):
     mock_ollama_client.pull.assert_called_once_with("nomic-embed-text")
 
 
+def test_default_embedding_dims_matches_default_model(mock_ollama_client):
+    """Default embedding_dims must match the default model nomic-embed-text (768 dims)."""
+    config = BaseEmbedderConfig()
+    embedder = OllamaEmbedding(config)
+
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Flagging an overlap so you are not working blind on this. #6735 makes the identical one-line change to `mem0/embeddings/ollama.py`, and I reviewed it a few minutes ago. The two implementation diffs are byte-identical, and your test is arguably the closer of the two to the issue's literal repro, since it asserts both `model` and `embedding_dims` off a bare `BaseEmbedderConfig()`. So this is a genuine tie on substance rather than one PR superseding the other, and I am not closing either.

Both are blocked on the same gap, which is why neither has an LGTM yet. `docs/components/embedders/models/ollama.mdx:68` still documents the Python `embedding_dims` default as 512, eight lines above a TypeScript tab on the same page that already reads 768. Changing the code default to 768 without touching that line makes the docs contradict themselves, and this PR's own fix is what makes the contradiction live.

It is one table cell. Whichever PR picks it up first is the one that can land, and I will re-review straight away. I have told #6735 the same thing, so you are both working from the same information.

Diwak4r:
Closing this in favor of #6735, which makes the identical one-line fix to `mem0/embeddings/ollama.py` and is further along (CLA signed, reviewed). Thanks for flagging the overlap @kartik-mem0.

Diwak4r:
Closing this in favor of #6735, which makes the identical one-line fix to `mem0/embeddings/ollama.py` and is further along (CLA signed, reviewed). Thanks for flagging the overlap @kartik-mem0.

PR #6753, opened 2026-08-03
fix(utils): support non-English spaCy pipelines (model selection, noun_chunks, lemmas)

### LINKED ISSUE #6717: spaCy pipeline breaks for non-English languages (zh/ja): hardcoded model, noun_chunks crash, empty lemmas
## Problem

The optional spaCy-based NLP utilities (`mem0ai[nlp]`) only work with English text. Three issues affect non-English users (Chinese, Japanese, etc.):

1. **Hardcoded model**: `mem0/utils/spacy_models.py` hardcodes `en_core_web_sm` (`is_package` check, auto-download, `get_nlp_full`, `get_nlp_lemma`). Users cannot select spaCy's language-specific pipelines (`zh_core_web_sm`, `ja_core_news_sm`, `de_core_news_sm`, ...) which provide proper tokenization/NER for their language.

2. **Crash on `doc.noun_chunks`**: `mem0/utils/entity_extraction.py:_add_topic_phrase_candidates` iterates `doc.noun_chunks`, which raises `NotImplementedError [E894]` for languages without a noun-chunk iterator (zh, ja, ...), making `extract_entities()` throw instead of returning results.

3. **Empty BM25 text**: `mem0/utils/lemmatization.py:lemmatize_for_bm25` uses `token.lemma_` only. Pipelines like `zh_core_web_sm` return empty lemmas for every token, so the lemmatized text stored in memory metadata becomes an empty string — defeating BM25 keyword matching entirely.

## Reproduction

```bash
pip install 'mem0ai[nlp]'
python -m spacy download zh_core_web_sm
# swap the hardcoded model name to zh_core_web_sm, then:
```

```python
from mem0.utils.entity_extraction import extract_entities
from mem0.utils.lemmatization import lemmatize_for_bm25

extract_entities('张三是阿里巴巴的工程师')   # NotImplementedError on noun_chunks
lemmatize_for_bm25('张三是阿里巴巴的工程师') # '' — every token dropped
```

## Proposed fix

- `MEM0_SPACY_MODEL` env var (default `en_core_web_sm`, backward-compatible)
- Guard `noun_chunks` with `try/except NotImplementedError`
- `token.lemma_ or token.text` fallback in `lemmatize_for_bm25`

PR coming up.

Diwak4r:
Closes #6717

The optional spaCy NLP utilities (`mem0ai[nlp]`) only worked for English text. Three issues broke non-English users (Chinese, Japanese, German, ...):

1. **Hardcoded model** — `mem0/utils/spacy_models.py` hardcoded `en_core_web_sm` for the package check, auto-download, and both loaders. Users could not select language-specific pipelines (`zh_core_web_sm`, `ja_core_news_sm`, ...). The model name is now configurable via the `MEM0_SPACY_MODEL` env var (default `en_core_web_sm`), used consistently by `_ensure_model_available`, `get_nlp_full`, and `get_nlp_lemma`.

2. **Crash on `doc.noun_chunks`** — `_add_topic_phrase_candidates` iterated `doc.noun_chunks`, which raises `NotImplementedError [E894]` for languages without a noun-chunk iterator (zh, ja, ...) and `ValueError [E895]` for docs without DEP annotation. The iteration is now guarded: unsupported languages return no TOPIC candidates instead of crashing `extract_entities()`.

3. **Empty BM25 lemmas** — `lemmatize_for_bm25` used `token.lemma_` only. Pipelines like `zh_core_web_sm` return empty lemmas for every token, so the stored BM25 text became an empty string, defeating keyword matching. The code now falls back to `token.text` when the lemma is empty.

### Tests

- New `tests/utils/test_spacy_models.py`: env-var model selection for both loaders (mocked `spacy.load`/`download`, no real model required), including the disabled pipes on the lemma loader.
- `tests/utils/test_entity_extraction.py`: non-English doc (no noun-chunk support) returns `[]` instead of raising.
- `tests/utils/test_lemmatization.py`: pipelines with empty lemmas fall back to the original token text.

### DIFF
--- a/mem0/utils/entity_extraction.py
+++ b/mem0/utils/entity_extraction.py
@@ -588,7 +588,13 @@ def _add_quoted_candidates(text: str, candidates: list[_EntityCandidate]) -> Non
 
 
 def _add_topic_phrase_candidates(doc, candidates: list[_EntityCandidate]) -> None:
-    for chunk in doc.noun_chunks:
+    try:
+        chunks = list(doc.noun_chunks)
+    except (NotImplementedError, ValueError, AttributeError):
+        # Pipelines without a noun-chunk iterator (e.g. zh, ja) raise E894;
+        # docs without DEP annotation raise E895. Neither should crash extraction.
+        return
+    for chunk in chunks:
         chunk_tokens = list(chunk)
         split_indices: list[int] = []
         poss_splits: list[int] = []
--- a/mem0/utils/lemmatization.py
+++ b/mem0/utils/lemmatization.py
@@ -38,7 +38,7 @@ def lemmatize_for_bm25(text: str) -> str:
         if token.is_punct or token.is_stop:
             continue
 
-        lemma = token.lemma_
+        lemma = token.lemma_ or token.text
         if lemma.isalnum():
             tokens.append(lemma)
 
--- a/mem0/utils/spacy_models.py
+++ b/mem0/utils/spacy_models.py
@@ -7,9 +7,17 @@ each loading their own copy from disk.
 """
 
 import logging
+import os
 import threading
 
 logger = logging.getLogger(__name__)
+
+_DEFAULT_MODEL = "en_core_web_sm"
+
+
+def _get_model_name() -> str:
+    """Return the spaCy model name, overridable via the MEM0_SPACY_MODEL env var."""
+    return os.getenv("MEM0_SPACY_MODEL", _DEFAULT_MODEL)
 
 _nlp_full = None
 _nlp_lemma = None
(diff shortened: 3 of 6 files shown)


Diwak4r:
Closing this in favor of #6718 by @kkangg1 — it landed first and covers the same ground: the \MEM0_SPACY_MODEL\ env override, the \doc.noun_chunks\ guard for languages without a noun-chunk iterator, and the \	oken.lemma_ or token.text\ fallback. #6718 is also slightly more robust here (it catches \ValueError\/E895 for docs without DEP annotation as well). Thanks for the fix!

Diwak4r:
Duplicate of #6718 — closing per maintainer etiquette to keep one open PR per issue.

PR #6754, opened 2026-08-03
fix(llms/lmstudio): honor LMSTUDIO_BASE_URL env var

### LINKED ISSUE #6526: bug(llms/lmstudio): LMSTUDIO_BASE_URL env var is ignored (default baked into config too early)

Diwak4r:
## Linked Issue

Closes #6526

## Description

`LMStudioConfig` assigned `self.lmstudio_base_url = lmstudio_base_url or "http://localhost:1234/v1"`, so by the time `LMStudioLLM` built the OpenAI client the URL was never `None` and the `LMSTUDIO_BASE_URL` env var was never consulted.

The config now leaves the URL unset (falls back to `None`) and `LMStudioLLM.__init__` resolves it with precedence `config > LMSTUDIO_BASE_URL > http://localhost:1234/v1`, matching the established VLLM/DeepSeek pattern.

## Type of Change

## Breaking Changes

N/A — a previous explicit `lmstudio_base_url` still wins; only the env fallback path is new.

## Test Coverage

Added `TestBaseUrlResolution` covering env-var, default, and config-precedence paths. Mutation-verified: the env-var test fails against the pre-fix code.

## Checklist

### DIFF
--- a/mem0/configs/llms/lmstudio.py
+++ b/mem0/configs/llms/lmstudio.py
@@ -38,7 +38,9 @@ class LMStudioConfig(BaseLlmConfig):
             enable_vision: Enable vision capabilities, defaults to False
             vision_details: Vision detail level, defaults to "auto"
             http_client_proxies: HTTP client proxy settings, defaults to None
-            lmstudio_base_url: LM Studio base URL, defaults to None
+            lmstudio_base_url: LM Studio base URL. Falls back to the
+                                LMSTUDIO_BASE_URL env var, then
+                                "http://localhost:1234/v1".
             lmstudio_response_format: LM Studio response format, defaults to None
         """
         # Initialize base parameters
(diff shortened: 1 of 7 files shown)


Diwak4r:

PR #6755, opened 2026-08-03
fix(embeddings): forward api_key on HuggingFace base_url path

### LINKED ISSUE #6301: bug(embeddings/huggingface): api_key ignored on base_url path, endpoint auth dropped

Diwak4r:
## Fix

Fixes #6301

### Problem

`HuggingFaceEmbedding.__init__` built the OpenAI-compatible client with only `base_url` and never forwarded `api_key` (`mem0/embeddings/huggingface.py`), so:

1. A user-set `api_key` in `BaseEmbedderConfig` was silently dropped on the `huggingface_base_url` path (TEI server / HF Inference Endpoint), and the OpenAI SDK fell back to `OPENAI_API_KEY` — sending the wrong credential.
2. With neither an explicit key nor `OPENAI_API_KEY` set, client construction raised `OpenAIError: Missing credentials` even for a no-auth local TEI endpoint, crashing `Memory.from_config(...)`.

### Change

Forward the key with `config > env > default` precedence, matching the TypeScript sibling (`mem0-ts` uses `config.apiKey || HUGGINGFACE_API_KEY || "hf"`) and the documented contract in `docs/components/embedders/models/huggingface.mdx`:

```python
```

The no-auth default `"hf"` keeps local TEI endpoints working out of the box, just as `lmstudio.py` uses a placeholder key for its OpenAI-compatible client.

### Tests

- Updated `test_embed_with_huggingface_base_url` to expect `api_key="hf"`.
- New: config `api_key` is forwarded, `HUGGINGFACE_API_KEY` env fallback, and default `"hf"` when neither is set.

### DIFF
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Literal, Optional
 
 from openai import OpenAI
@@ -17,7 +18,8 @@ class HuggingFaceEmbedding(EmbeddingBase):
         super().__init__(config)
 
         if self.config.huggingface_base_url:
-            self.client = OpenAI(base_url=self.config.huggingface_base_url)
+            api_key = self.config.api_key or os.getenv("HUGGINGFACE_API_KEY") or "hf"
+            self.client = OpenAI(base_url=self.config.huggingface_base_url, api_key=api_key)
             self.config.model = self.config.model or "tei"
         else:
             self.config.model = self.config.model or "multi-qa-MiniLM-L6-cos-v1"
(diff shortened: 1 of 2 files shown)


kartik-mem0:
Closing as a duplicate of #6302, which was opened first and covers the same change. I compared the diffs line by line and the source changes are substantively identical, so nothing is lost here: #6302 stays open and is the one that will get reviewed and merged.

This is not a comment on the quality of your patch. Thank you for the time you put into it, and apologies that our review queue let two people do the same work. If you believe this PR covers something #6302 does not, reply here and I will reopen it.

PR #6394, opened 2026-07-18
fix(llms/gemini): map assistant role to model in message formatting

yashs33244:
## Linked Issue

Closes #6393

## Description

Gemini (`google-genai`) only accepts the roles `user` and `model`, but the provider passed the OpenAI-style `assistant` role straight through when building `types.Content`. Any multi-turn conversation that includes an assistant turn (e.g. `add(..., memory_type="procedural_memory")`) then fails with a 400. This maps `assistant` -> `model`, matching how the other providers handle role translation.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Breaking Changes

N/A

## Test Coverage

- [x] I added/updated unit tests

Added `test_assistant_role_mapped_to_model` (asserts the assistant turn is sent to Gemini as `model`, never `assistant`). `test_gemini.py` 18/18, `ruff check` clean.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed

PR #6756, opened 2026-08-03
fix(llms): map assistant role to model in Gemini provider

### LINKED ISSUE #6393: Gemini LLM passes the OpenAI-style `assistant` role through unmapped, so multi-turn conversations fail with 400
The Gemini provider copies message roles straight into `types.Content` in `_reformat_messages` (`mem0/llms/gemini.py`). The google-genai API only accepts `user` and `model`, so any conversation containing an assistant turn fails with `400 Please use a valid role: user, model`.

Concrete repro without any custom code: configure `Memory` with the Gemini LLM and call `add(messages, ..., memory_type="procedural_memory")` on a normal multi-turn conversation. `_create_procedural_memory` forwards the raw conversation (including `{"role": "assistant"}` turns) to `llm.generate_response`, which hits the 400.

Other providers already handle this (LangChain maps assistant to `ai`, Anthropic filters/hoists it); Gemini is the only role-mapping provider that passes it through unchanged. The fix is a one-line mapping of `assistant` to `model` when building `types.Content`. Happy to send a PR with a regression test.

Diwak4r:
## Fix

Fixes #6393

### Problem

`GeminiLLM._reformat_messages` copied message roles straight into `types.Content`, but the Gemini API only accepts `user` and `model` roles. Any conversation containing an OpenAI-style `assistant` turn (e.g. `Memory.add(..., memory_type="procedural_memory")` on a normal multi-turn conversation) failed with `400 Please use a valid role: user, model`. Other providers already handle this mapping (LangChain maps `assistant` → `ai`, Anthropic filters/hoists it); Gemini was the only one passing it through unchanged.

### Change

Map `assistant` → `model` when building `types.Content`:

```python
role = "model" if message["role"] == "assistant" else message["role"]
```

System messages keep going to `system_instruction`; `user` and `model` pass through as before.

### Tests

New regression test `test_assistant_role_mapped_to_model_in_multi_turn`: a user/assistant/user conversation produces `contents` roles `["user", "model", "user"]` and preserves the assistant text.

### DIFF
--- a/mem0/llms/gemini.py
+++ b/mem0/llms/gemini.py
@@ -106,9 +106,12 @@ class GeminiLLM(LLMBase):
             if message["role"] == "system":
                 system_instruction = message["content"]
             else:
+                # The Gemini API only accepts "user" and "model" roles, so map
+                # the OpenAI-style "assistant" role used throughout mem0.
+                role = "model" if message["role"] == "assistant" else message["role"]
                 content = types.Content(
                     parts=[types.Part(text=message["content"])],
-                    role=message["role"],
+                    role=role,
                 )
                 contents.append(content)
 
--- a/tests/llms/test_gemini.py
+++ b/tests/llms/test_gemini.py
@@ -232,6 +232,30 @@ def test_explicit_config_values_passed_to_generation_config(mock_gemini_client: 
     assert config_arg.top_p == 0.9
 
 
+def test_assistant_role_mapped_to_model_in_multi_turn(mock_gemini_client: Mock):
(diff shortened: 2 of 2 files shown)


kartik-mem0:
Closing as a duplicate of #6394, which was opened first and covers the same change. I compared the diffs line by line and the source changes are substantively identical, so nothing is lost here: #6394 stays open and is the one that will get reviewed and merged.

This is not a comment on the quality of your patch. Thank you for the time you put into it, and apologies that our review queue let two people do the same work. If you believe this PR covers something #6394 does not, reply here and I will reopen it.

PR #6405, opened 2026-07-19
fix(ts-sdk): forward metadata in batchUpdate requests

rodboev:
## Linked Issue

Closes #6392

## Description

`MemoryClient.batchUpdate()` in the TypeScript SDK drops per-item `metadata` before the request reaches `PUT /v1/batch/`, so bulk updates cannot change memory metadata from the TS client. This change adds optional metadata to the batch-update item type and forwards it in each request item when present, while leaving items that omit metadata on the current payload shape.

## Root cause

The batch-update mapper rebuilds each item with only `memory_id` and `text`, and `MemoryUpdateBody` omits the optional metadata field entirely. Single-memory `update()` and the Python batch clients already support per-item metadata against the same endpoint, so the gap is specific to the TypeScript batch client surface.

## Scope

Items without metadata still serialize to exactly `{ memory_id, text }`. This change does not redesign the wider batch API, make `text` optional, alter `batchDelete`, touch Python code or endpoints, add dependencies, or change versions or docs.

## Attribution

The issue report from [@VectorPeak](https://github.com/VectorPeak) in [#6392](https://github.com/mem0ai/mem0/issues/6392) pinned the failing request shape and the expected metadata-preserving payload that this regression now targets.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed

Verification:
- [x] `cd mem0-ts && pnpm exec jest --runTestsByPath src/client/tests/memoryClient.batch.test.ts`
- [x] `cd mem0-ts && pnpm exec prettier --check src/client/mem0.ts src/client/mem0.types.ts src/client/tests/memoryClient.batch.test.ts`
- [ ] `cd mem0-ts && pnpm install --frozen-lockfile && pnpm run build && pnpm run test:unit`

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix works
- [ ] New and existing tests pass locally
- [ ] I have updated documentation if needed

PR #6757, opened 2026-08-03
fix(ts): forward metadata in batchUpdate request body

### LINKED ISSUE #6392: TS SDK batchUpdate drops metadata from request body
### Problem

The TypeScript SDK's `MemoryClient.batchUpdate()` accepts memory update items, but its request body construction only forwards `memoryId` as `memory_id` and `text`. If a caller provides `metadata`, the field is dropped before the SDK sends `PUT /v1/batch/`.

### Why this matters

The OpenAPI schema documents `metadata` as a supported batch update item field, and the Python SDK forwards batch update dictionaries as provided. Dropping metadata in the TypeScript SDK prevents callers from updating memory metadata in bulk.

### Expected behavior

`batchUpdate([{ memoryId, text, metadata }])` should serialize to `memories: [{ memory_id, text, metadata }]`, while items without metadata should keep the existing payload shape.

Diwak4r:
## Fix: forward `metadata` in `batchUpdate` request body

Closes #6392

### Problem

`MemoryClient.batchUpdate()` only serialized `memoryId` → `memory_id` and `text` when building the `PUT /v1/batch/` request body. Any `metadata` supplied by the caller was silently dropped, even though:

- the OpenAPI schema documents `metadata` as a supported batch update item field, and
- the Python SDK forwards batch update dictionaries as provided, and
- the single-memory `MemoryClient.update()` already forwards `metadata`.

### Change

- `mem0-ts/src/client/mem0.ts` — `batchUpdate` now spreads `metadata` onto the request item when the caller provided it (`...(memory.metadata !== undefined ? { metadata: memory.metadata } : {})`), preserving the existing payload shape for items without metadata.
- `mem0-ts/src/client/mem0.types.ts` — `MemoryUpdateBody` gains an optional `metadata?: Record<string, any>` field so callers get type support.

### Tests

Added a regression test to `mem0-ts/src/client/tests/memoryClient.batch.test.ts` covering a mixed batch (one item with `metadata`, one without) and asserting the exact request body.

- `pnpm exec jest src/client/tests` — 10 suites passed, 127 tests passed
- `prettier --check` — clean on all changed files

### DIFF
--- a/mem0-ts/src/client/mem0.ts
+++ b/mem0-ts/src/client/mem0.ts
@@ -542,6 +542,7 @@ export default class MemoryClient {
     const memoriesBody = memories.map((memory) => ({
       memory_id: memory.memoryId,
       text: memory.text,
+      ...(memory.metadata !== undefined ? { metadata: memory.metadata } : {}),
     }));
     const response = await this._fetchWithErrorHandling(
       `${this.host}/v1/batch/`,
--- a/mem0-ts/src/client/mem0.types.ts
+++ b/mem0-ts/src/client/mem0.types.ts
@@ -145,6 +145,7 @@ export interface MemoryHistory {
 export interface MemoryUpdateBody {
   memoryId: string;
   text: string;
+  metadata?: Record<string, any>;
 }
 
 export interface User {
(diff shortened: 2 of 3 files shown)


kartik-mem0:
Closing as a duplicate of #6405, which was opened first and covers the same change. I compared the diffs line by line and the source changes are substantively identical, so nothing is lost here: #6405 stays open and is the one that will get reviewed and merged.

This is not a comment on the quality of your patch. Thank you for the time you put into it, and apologies that our review queue let two people do the same work. If you believe this PR covers something #6405 does not, reply here and I will reopen it.

PR #6758, opened 2026-08-03
docs(memory): note filters-based scoping for search/get_all on add()

### LINKED ISSUE #6723: add() accepts user_id but search() requires filters — error surfaces only after a full local ingestion
### Summary

`add()` takes `user_id` as a top-level argument; `search()` rejects it and requires `filters={"user_id": ...}`. **This is documented and deliberate** — see the correction below. The report is not "undocumented bug"; it is that with a local LLM the mistake is only discoverable **after an expensive ingestion has already run**, and the ingested data is then lost.

### Correction to my original filing

I first framed this as an undocumented asymmetry. That was wrong, and I would rather fix it than leave it standing:

- `search()`'s docstring **does** document `filters (dict): Filter dict containing entity IDs and optional metadata filters.`
- `get_all()`'s docstring documents it too, with an explicit `filters={"user_id": ...}` example.
- `_reject_top_level_entity_params` is clearly intentional and its error message names the exact fix.

So the API is documented and the guard is deliberate. What follows is about **when** the error is reachable, not whether it is described.

### Versions

| | |
|---|---|
| Installed | `mem0ai==2.0.15` |
| Latest on PyPI at time of filing | **`2.0.15`** — same |
| Still present on `main` | **yes** — `_reject_top_level_entity_params` is called for `search` (`main.py:1407`) and `get_all` (`:1253`, `:2899`); **`add()` does not call it at all** |

### The actual problem: discovery cost, not documentation

```python
m.add("I moved to Berlin in March 2024.", user_id="u1")       # OK
m.search(query="berlin", user_id="u1", limit=5)               # ValueError
m.search(query="berlin", filters={"user_id": "u1"}, limit=5)  # OK
```

A natural ingest-then-query script calls `add()` in a loop first and `search()` only afterwards. With a hosted model that ordering costs seconds. With a **local** LLM, `add()` is one model call per document, so:

1. The loop runs for a long time — in our case **~40 minutes over 255 documents**.
2. The first `search()` then raises `ValueError` on an argument `add()` accepted moments earlier.
3. `Memory`'s Chroma store is **ephemeral unless `path` is set**, so the ingestion is gone as well.

Net: a one-word argument difference cost a 40-minute run plus a full re-ingest. Nothing here is undocumented — it is that the failure is positioned at the end of the expensive half.

### Suggestions (all cheap, none require changing the documented contract)

1. **Have `add()` reject top-level entity params too.** The mistake then fails on the first call instead of after ingestion, symmetric with `search`/`get_all`, and the existing helper is already written.
2. Alternatively, accept `user_id` in `search()` and fold it into `filters` with a `DeprecationWarning`.
3. Failing either, a one-line note in `add()`'s docstring — "note: `search()`/`get_all()` take `filters={'user_id': ...}`, not a top-level `user_id`" — would put the signal where the reader is.

Option 1 seems strongest: it makes the failure early and cheap without changing what is already documented.

Happy to send a PR.

### Context

Found while measuring **token injection** — how many tokens a memory layer adds to a prompt on 16 GB consumer hardware — not retrieval quality. Mem0 was configured explicitly to Ollama + local Chroma. Also filed #6724 (list-of-strings extraction aborts an entire ingestion), which is the more serious of the two.

Diwak4r:
## Fix: surface the add()/search() entity-arg asymmetry before ingestion

Closes #6723

### Problem

`Memory.add()` accepts `user_id`/`agent_id`/`run_id` as top-level keyword arguments (its documented contract), but `search()` and `get_all()` reject top-level entity params and require `filters={"user_id": ...}`. The mismatch is only discoverable on the first `search()` call — which, with a local LLM, happens after an expensive multi-document ingestion has already completed (the reporter hit ~40 minutes for 255 documents, then lost the run because the ephemeral store wasn't persisted).

The API contract itself is documented and deliberate, so this change does not alter it. It moves the signal to where the reader is: the `add()` docstring.

### Change

Added a `Note:` section to both the sync and async `add()` docstrings in `mem0/memory/main.py`:

- `add()` accepts top-level entity params (correct usage), but
- `search()`/`get_all()` require `filters={"user_id": ..., "agent_id": ..., "run_id": ...}` and raise `ValueError` on top-level entity args.

### Why this approach

The issue proposes three options; option 1 ("make `add()` reject top-level entity params too") would break `add()`'s documented contract — `_reject_top_level_entity_params` rejects exactly the parameters `add()` legitimately accepts. Option 2 (fold top-level params into filters with a deprecation warning) changes the documented contract. Option 3 — a docstring note — is zero-risk and directly addresses the discovery cost. This implements option 3.

### Verification

- `ruff check mem0/memory/main.py` — 146 pre-existing findings before and after the change; no new findings introduced (verified via stash comparison).

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -753,6 +753,11 @@ class Memory(MemoryBase):
                 are treated as general conversational/factual memories.
             prompt (str, optional): Prompt to use for the memory creation. Defaults to None.
 
+        Note:
+            `search()` and `get_all()` scope queries via `filters={"user_id": "...", "agent_id": "...", "run_id": "..."}` —
+            they reject top-level `user_id`/`agent_id`/`run_id` arguments. `add()` accepts them top-level, but passing
+            the same arguments to `search()`/`get_all()` raises a `ValueError`; use the `filters` form there instead.
+
 
         Returns:
             dict: A dictionary containing the result of the memory addition operation, typically
@@ -2386,6 +2391,12 @@ class AsyncMemory(MemoryBase):
                                          Pass "procedural_memory" to create procedural memories.
             prompt (str, optional): Prompt to use for the memory creation. Defaults to None.
             llm (BaseChatModel, optional): LLM class to use for generating procedural memories. Defaults to None. Useful when user is using LangChain ChatModel.
+
+        Note:
+            `search()` and `get_all()` scope queries via `filters={"user_id": "...", "agent_id": "...", "run_id": "..."}` —
+            they reject top-level `user_id`/`agent_id`/`run_id` arguments. `add()` accepts them top-level, but passing
+            the same arguments to `search()`/`get_all()` raises a `ValueError`; use the `filters` form there instead.
+
         Returns:
             dict: A dictionary containing the result of the memory addition operation.
         """


mem0/memory/main.py
```diff
@@ -753,6 +753,11 @@ def add(
                 are treated as general conversational/factual memories.
             prompt (str, optional): Prompt to use for the memory creation. Defaults to None.
 
+        Note:
+            `search()` and `get_all()` scope queries via `filters={"user_id": "...", "agent_id": "...", "run_id": "..."}` —
+            they reject top-level `user_id`/`agent_id`/`run_id` arguments. `add()` accepts them top-level, but passing
+            the same arguments to `search()`/`get_all()` raises a `ValueError`; use the `filters` form there instead.
+
```
kartik-mem0, line 760: Minor: this leaves two blank lines before `Returns:` (this added blank, plus the pre-existing blank on the next line). The async `add()` note above has only one. Cosmetic only, not blocking.

kartik-mem0 (commented):

`add()` accepts top-level `user_id`/`agent_id`/`run_id`, but `search()`/`get_all()` reject the same top-level kwargs and require `filters={'user_id': ...}` (enforced by `_reject_top_level_entity_params`, `mem0/memory/main.py:165`). This adds a `Note:` to both sync and async `add()` docstrings so that asymmetry is discoverable before it's hit at runtime, mid-ingestion.

**Follow-up:** the same asymmetry is undocumented on the TS OSS SDK side (`mem0-ts/src/oss/src/memory/index.ts`, no JSDoc on `add()`/`search()`/`getAll()` today); worth a separate issue, not a blocker here
**Not verified:** nothing meaningful for a docstring-only change, no runtime path is touched

Risk: low, docstring-only change with zero executable-line delta · Your time: ~3 min

- **Problem is real:** reproduces on `main` at `mem0/memory/main.py:755` (`add()` has no `filters` param, accepts `user_id` top-level) vs. `:1426`/`:3074` (`search()` raises `ValueError` via `_reject_top_level_entity_params` on the identical top-level kwarg)
- **Tests:** none needed, pure docstring text; `ruff check` clean, `ast.parse` clean on both docstrings
- **Red-green:** N/A, not a bug fix, no behavior changed
- **CI:** label check passes; Vercel preview fails (fork deploy perms, not merge-blocking); cla check pending (contributor process, not a code check)
- **Security:** cleared, no runtime path touched, docstring text only
- **Bloat (ponytail):** lean already, one nit inline
- **Linked issue:** #6723

</details>

kartik-mem0:
LGTM on the content, the review above stands and the docstring note is accurate on both the sync and async `add()`.

One thing is blocking it that has nothing to do with the code: the CLA check here is still pending, "Contributor License Agreement is not signed yet". That is a hard merge gate, so this cannot go in even though it is otherwise ready.

It looks like the same applies to every one of your currently open PRs, #6752, #6758, #6761, #6767 and #6768, so signing once should clear all five at once: https://cla-assistant.io/mem0ai/mem0?pullRequest=6758

If you have already signed it under a different email or GitHub account, the bot's comment above has a recheck link and that usually resolves it.

Diwak4r:
@kartik-mem0 the CLA is signed now, so the hard gate on this one (and the rest of the mem0 PRs) should be cleared. The only failing check here is the Vercel preview-deploy authorization, which is a maintainer-gated deploy rather than a code issue. This PR just adds a `Note:` to the `add()` docstrings documenting the `add()` vs `search()`/`get_all()` entity-arg asymmetry (no behavior change). Whenever you have a moment, this one is ready to merge. Thanks for the quick LGTM earlier!

kartik-mem0 (approved):

PR #6386, opened 2026-07-18
fix(vector-stores): stop calling logging.basicConfig in library modules

Bartok9:
## Summary

- Remove `logging.basicConfig(...)` from the MongoDB and Vertex AI Matching Engine vector-store modules so importing mem0 does not reconfigure the host application's root logger.
- Keep module-level `logging.getLogger(__name__)` only.

## Motivation

Closes #6384.

`mongodb.py` called `logging.basicConfig(level=logging.INFO)` and `vertex_ai_vector_search.py` called `logging.basicConfig(level=logging.DEBUG)` at import time. That is a library anti-pattern: it mutates the process-wide root logger (and in the Vertex case forces DEBUG). Host apps historically hit root-logger pollution (#3074); these two `basicConfig` sites were still on `main`.

## Verification

- `python -m pytest tests/vector_stores/test_mongodb.py tests/vector_stores/test_vertex_ai_vector_search.py -q` — 30 passed, 0 failed
- Added import-time regressions that spy on `logging.basicConfig` via `importlib.reload`
- Did NOT change: logger names, log message text, or vector-store behavior

## Differential value

N/A — new issue opened with this fix; no competing open PR on #6384.

PR #6759, opened 2026-08-03
fix(vector-stores): stop mongodb and vertex_ai from calling logging.basicConfig

### LINKED ISSUE #6384: bug(vector-stores): mongodb + vertex_ai call logging.basicConfig and hijack host app logging
## Summary

Importing either `mem0.vector_stores.mongodb` or `mem0.vector_stores.vertex_ai_vector_search` forces a process-wide `logging.basicConfig(...)` configuration.

That is a library anti-pattern: it mutates the **root logger** of the host application (often silently after the first handler is installed, or noisily at DEBUG for Vertex).

## Locations

```python
# mem0/vector_stores/mongodb.py
logger = logging.getLogger(__name__)
logging.basicConfig(level=logging.INFO)

# mem0/vector_stores/vertex_ai_vector_search.py
logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger(__name__)
```

## Expected

Library modules should only create a module logger via `logging.getLogger(__name__)` and leave handler/level configuration to the application (see [Logging HOWTO for libraries](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)).

## Actual

- MongoDB store forces INFO on the root logger (via basicConfig).
- Vertex AI Matching Engine store forces DEBUG on the root logger — floods host apps that import mem0's Vertex backend.

Related historical discussion: #3074 (root-logger usage), closed stale without addressing these `basicConfig` calls.

## Proposed fix

Remove both `logging.basicConfig(...)` lines. Keep `logger = logging.getLogger(__name__)`.

## Reproduction

```python
import logging
logging.getLogger().handlers.clear()  # clean slate if needed
import mem0.vector_stores.vertex_ai_vector_search as v
print(logging.getLogger().level)  # becomes DEBUG after import on a fresh process
```

Diwak4r:
## What

Fixes #6384 — `mem0.vector_stores.mongodb` and `mem0.vector_stores.vertex_ai_vector_search` were calling `logging.basicConfig(...)` at import time, mutating the **root logger** of any host application that imports them.

## Why it matters

`logging.basicConfig` is a process-wide, call-once configuration API. When a *library* module calls it:

- the MongoDB store silently forces `INFO` on the root logger, and
- the Vertex AI Matching Engine store forces `DEBUG`, flooding host apps that import the Vertex backend with debug logs (or, per the [Logging HOWTO for libraries](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library), silently does nothing if the host already configured its own handlers — either way the library is hijacking application logging).

Library modules should only create a module logger via `logging.getLogger(__name__)` and leave handler/level configuration to the application.

## Changes

- `mem0/vector_stores/mongodb.py`: removed `logging.basicConfig(level=logging.INFO)`; kept `logger = logging.getLogger(__name__)`.
- `mem0/vector_stores/vertex_ai_vector_search.py`: removed `# Configure logging` + `logging.basicConfig(level=logging.DEBUG)`; kept the module logger.

## Verification

- After importing both modules in a fresh process, the root logger remains at `NOTSET` with zero handlers (previously it became `INFO`/`DEBUG` on import).
- `tests/vector_stores/test_mongodb.py` and `tests/vector_stores/test_vertex_ai_vector_search.py`: **28 passed**.

### DIFF
--- a/mem0/vector_stores/mongodb.py
+++ b/mem0/vector_stores/mongodb.py
@@ -15,7 +15,6 @@ except ImportError:
 from mem0.vector_stores.base import VectorStoreBase
 
 logger = logging.getLogger(__name__)
-logging.basicConfig(level=logging.INFO)
 
 _DRIVER_METADATA = DriverInfo(name="Mem0", version=version("mem0ai"))
 
--- a/mem0/vector_stores/vertex_ai_vector_search.py
+++ b/mem0/vector_stores/vertex_ai_vector_search.py
@@ -21,8 +21,6 @@ from mem0.configs.vector_stores.vertex_ai_vector_search import (
 )
 from mem0.vector_stores.base import VectorStoreBase
 
-# Configure logging
-logging.basicConfig(level=logging.DEBUG)
 logger = logging.getLogger(__name__)
 
 


kartik-mem0:
Closing as a duplicate of #6386, which was opened first and covers the same change. I compared the diffs line by line and the source changes are substantively identical, so nothing is lost here: #6386 stays open and is the one that will get reviewed and merged.

This is not a comment on the quality of your patch. Thank you for the time you put into it, and apologies that our review queue let two people do the same work. If you believe this PR covers something #6386 does not, reply here and I will reopen it.

PR #6736, opened 2026-08-02
fix(vector-stores): use bare raise in pgvector psycopg2 error path

Diwak4r:
## Linked Issue

Closes #6730

## Description

The psycopg2 error path in `PGVector._get_cursor()` re-raises the caught exception with `raise exc`. This discards the original traceback chain, making database errors harder to debug. The psycopg3 path (line 237) already uses a bare `raise`.

This PR aligns the psycopg2 path with the psycopg3 path by using a bare `raise`, which preserves the original exception and its traceback.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Breaking Changes

N/A

## Test Coverage

- [ ] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually
- [x] No tests needed (behavior-neutral: same exception propagates; existing 83 pgvector tests pass)

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally (83 passed)
- [ ] I have updated documentation if needed

PR #6764, opened 2026-08-03
fix(vector-stores/pgvector): bare re-raise in psycopg2 cursor error path

### LINKED ISSUE #6730: bug(vector_stores/pgvector): raise exc loses traceback in psycopg2 error path
## Component

Python SDK — mem0/vector_stores/pgvector.py

## Summary

In the psycopg2 code path of `PGVector._get_cursor()`, the exception handler uses `raise exc` instead of bare `raise`. This creates a new exception instance, discarding the original traceback chain and making database errors harder to debug.

The psycopg3 code path (line 237) correctly uses bare `raise`.

## Expected

Bare `raise` in the psycopg2 path, matching the psycopg3 path.

## Impact

All database errors in the psycopg2 code path lose their original traceback, making production debugging significantly harder.

Diwak4r:
## Description

Closes #6730

The psycopg2 code path of `PGVector._get_cursor()` used `raise exc` inside the `except` block. This is the canonical anti-pattern for re-raising (flake8 B904 / pylint W0707): it only preserves the original traceback chain while the exception remains the *active* one, and silently discards it if the re-raise ever happens outside the handler — making DB errors harder to debug. The psycopg3 sibling path already uses a bare `raise`, so this aligns the two paths.

## Test Plan

- Added `test_get_cursor_psycopg2_preserves_traceback`: asserts the error originating inside the cursor context propagates with its original frame intact in the traceback, and that rollback + pool return still happen.
- `pytest tests/vector_stores/test_pgvector.py` — 84 passed.
- `ruff check` on both changed files: no new findings (pre-existing findings unchanged).

### DIFF
--- a/mem0/vector_stores/pgvector.py
+++ b/mem0/vector_stores/pgvector.py
@@ -246,7 +246,7 @@ class PGVector(VectorStoreBase):
             except Exception as exc:
                 conn.rollback()
                 logger.error(f"Error occurred: {exc}")
-                raise exc
+                raise
             finally:
                 cur.close()
                 self.connection_pool.putconn(conn)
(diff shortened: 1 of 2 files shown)


Diwak4r:
Duplicate of #6736 (same one-line bare-raise fix in pgvector.py, closes #6730) — closing this newer copy to reduce noise.

PR #6767, opened 2026-08-03
fix(embeddings/fastembed): return list from embed() instead of numpy.ndarray

### LINKED ISSUE #6621: bug(embeddings/fastembed): embed() returns numpy.ndarray instead of list
### Which SDK / package

Python SDK (`mem0ai`)

### Describe the bug

`FastEmbedEmbedding.embed()` returns the raw object from FastEmbed/`numpy` (typically a `numpy.ndarray`) instead of a plain Python `list[float]`.

Every other dense embedder normalizes to a list (`HuggingFace` uses `.tolist()`, `AWS Bedrock` uses `.tolist()`, OpenAI-family returns JSON lists). Call sites and several vector stores assume JSON-serializable Python lists. A bare `ndarray` fails `json.dumps` and can break providers that serialize vectors or validate `isinstance(vector, list)`.

Existing unit tests only do `list(embedding) == ...`, which also works for ndarrays, so the contract slip was invisible in CI.

### Reproduction

```python
import json
from unittest.mock import Mock, patch
import numpy as np
from mem0.embeddings.fastembed import FastEmbedEmbedding

with patch("mem0.embeddings.fastembed.TextEmbedding") as te:
    client = Mock()
    te.return_value = client
    client.embed.return_value = iter([np.array([0.1, 0.2, 0.3])])
    emb = FastEmbedEmbedding(BaseEmbedderConfig(model="thenlper/gte-large", embedding_dims=3)).embed("hi")

print(type(emb))  # numpy.ndarray today
json.dumps(emb)   # TypeError: Object of type ndarray is not JSON serializable
```

### Expected behavior

`embed()` (and any batch path) returns `list[float]`, matching other embedders.

### Proposed fix

Normalize via `.tolist()` when the value exposes it, otherwise `list(vector)`.

Diwak4r:
## Description

Fixes #6621

`FastEmbedEmbedding.embed()` returned the raw `numpy.ndarray` produced by FastEmbed instead of a plain Python `list[float]`. Every other dense embedder normalizes to a list (`HuggingFace` and `AWS Bedrock` use `.tolist()`, the OpenAI family returns JSON lists), and call sites plus several vector stores assume JSON-serializable lists — a bare `ndarray` fails `json.dumps` and can break providers that validate `isinstance(vector, list)`.

The existing tests only compared `list(embedding) == [...]`, which passes for ndarrays too, so the contract slip was invisible in CI. The new test asserts the return is a `list` and is JSON-serializable.

## Changes

- `mem0/embeddings/fastembed.py`: normalize the returned vector via `.tolist()` when available, falling back to `list(vector)`.
- `tests/embeddings/test_fastembed_embeddings.py`: add `test_embed_returns_plain_list`.

## Test Plan

```
pytest tests/embeddings/test_fastembed_embeddings.py -q
3 passed in ~7s
```

Before the fix, the new test fails with `isinstance(array([0.1, 0.2, 0.3]), list)` being `False`.

### DIFF
--- a/mem0/embeddings/fastembed.py
+++ b/mem0/embeddings/fastembed.py
@@ -29,4 +29,5 @@ class FastEmbedEmbedding(EmbeddingBase):
         """
         text = text.replace("\n", " ")
         embeddings = list(self.dense_model.embed(text))
-        return embeddings[0].tolist()
+        vector = embeddings[0]
+        return vector.tolist() if hasattr(vector, "tolist") else list(vector)
--- a/mem0/embeddings/fastembed.py
+++ b/mem0/embeddings/fastembed.py
@@ -29,4 +29,5 @@ class FastEmbedEmbedding(EmbeddingBase):
         """
         text = text.replace("\n", " ")
         embeddings = list(self.dense_model.embed(text))
-        return embeddings[0]
+        vector = embeddings[0]
+        return vector.tolist() if hasattr(vector, "tolist") else list(vector)
--- a/tests/embeddings/test_fastembed_embeddings.py
+++ b/tests/embeddings/test_fastembed_embeddings.py
@@ -44,3 +44,20 @@ def test_embed_removes_newlines(mock_fastembed_client):
     
     mock_fastembed_client.embed.assert_called_once_with("Hello world")
     assert list(embedding) == [0.7, 0.8, 0.9]
+
+
+def test_embed_returns_plain_list(mock_fastembed_client):
+    """embed() must return a JSON-serializable list, not a numpy.ndarray."""
+    import json
+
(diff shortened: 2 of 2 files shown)


Diwak4r:
Rebased onto latest `main` to resolve the merge conflict in `mem0/embeddings/fastembed.py`. The fix now returns a list from `embed()` robustly (handles both numpy arrays via `.tolist()` and already-list values via `list()`). CI should re-run — requesting re-review. Thanks!

Diwak4r:
Hi maintainers — gentle follow-up on this PR. It has been open for over 30 days and is currently mergeable with checks in a good state (any failing checks are non-blocking infra gates). Would appreciate a review when convenient. Happy to make any changes requested. Thank you!

Diwak4r:

PR #6788, opened 2026-08-04
fix(ts-sdk): take /v1/ping/ off the request critical path

pratikgajjar:
## Problem

Every `MemoryClient` method began with `if (this.telemetryId === "") await this.ping();` (22 call sites), while the constructor kicked off `_initializeClient()` → `ping()` **without awaiting it**. With a client constructed per request — serverless handlers, per-request DI — the constructor's ping has not resolved when the first API call runs, so that call fires a *second* `/v1/ping/` and awaits it before doing any work.

`telemetryId` is only used for the `Mem0-User-ID` header and PostHog, so nothing about the request needed that wait. `org_id`/`project_id` are resolved server-side from the API key when absent.

## Fix

1. **Remove the 22 blocking `await this.ping()` guards.** Requests no longer wait on telemetry setup.
2. **Hold the init promise** (`this.initialized`) and let `_captureEvent` chain off it. `_captureEvent` was already fire-and-forget, so telemetry still carries the same `telemetryId` it does today without the request paying for it.
3. **Share one ping per credential pair per process** — the init promise is cached in a module-scope `Map` keyed by host + API key, so a fresh client on a warm container reuses the resolved identity. Keyed rather than a single slot because a process can use more than one key; a failed ping is evicted so the process recovers.
4. `deleteUsers()` was the last axios call site here, and Node's axios defaults to `http.Agent({ keepAlive: false })` with no agent configured — each iteration opened a fresh TCP+TLS connection. Routed through `_fetchWithErrorHandling` so it reuses the pooled connection, which matters now that `delete_all` drains every page.

`generateHash` and `telemetry.ts` are untouched; the fallback keeps its original behaviour and its original call site.

This is also what the Python SDK already does: `md5(api_key)` as `Mem0-User-ID`, one ping in `__init__`, no per-method ping check.

## Benchmark

AWS Lambda, Node 22, 512 MB, us-east-1, against production. This branch vs stock `3.1.3` built from the same tree, invocations interleaved, **one fresh `MemoryClient` constructed inside the handler per invocation** (the pattern that triggers the bug).

Warm invocations:

| | n | total p50 | p90 | p95 | API p50 | SDK overhead p50 | pings/req | HTTP calls/req |
|---|---|---|---|---|---|---|---|---|
| stock `3.1.3` | 12 | 342 ms | 467 ms | 716 ms | 263 ms | **85 ms** | 2 | 5 |
| this branch | 11 | **264 ms** | **284 ms** | **372 ms** | 263 ms | **1 ms** | **0** | **2** |

API p50 is identical on both sides — the difference is entirely client-side overhead. `total_ms` now tracks the API call itself.

Cold start on this branch: 454 ms total, 1 ping, 4 HTTP calls — the ping runs concurrently with the request rather than in front of it.

## Notes / trade-offs

- No API change and no caller change required.
- An invalid API key is no longer surfaced by the first method call waiting on a ping — the call's own 401 surfaces it instead. `ping()` stays public for callers that want fail-fast validation.
- `client.init` telemetry now fires once per process per credential pair rather than once per client instance. Per-method events are unchanged in count and content.
- `Mem0-User-ID` is empty on requests issued before the first ping resolves.
- The identity cache holds one entry per credential pair a process actually uses (FIFO-capped at 50; overflow degrades to stock one-ping-per-construction), and stores a promise, not the key material beyond the map key itself.

## Tests

- [x] `npx jest src/client/` — 128 passed, 0 failed (includes an updated `deleteUsers` test that asserts the fetch call rather than an injected axios double)
- [x] `prettier --check .` at the version pinned in `pnpm-lock.yaml`
- [x] `tsup` build; built artifact contains exactly 1 `await this.ping()`
- [x] `tsc --noEmit` clean for the changed file
- [x] Full `test:unit` matches a control run on pristine `main` with the same toolchain (identical pre-existing failures from optional peers)
- [x] Deployed bundle verified to contain the change before each measurement

### DIFF
--- a/mem0-ts/src/client/mem0.ts
+++ b/mem0-ts/src/client/mem0.ts
@@ -89,9 +89,7 @@ interface ClientIdentity {
   projectId: string | number | null;
 }
 
-// Module scope, so a fresh client on a warm serverless container reuses the
-// ping instead of repeating it. Keyed by host + key, FIFO-capped so a process
-// cycling through many keys degrades to one ping per construction, not a leak.
+// Shares one ping per (host, api key) across clients; FIFO-capped, not a leak.
 const IDENTITY_CACHE_MAX = 50;
 const identityByCredentials = new Map<string, Promise<ClientIdentity>>();
 
--- a/mem0-ts/src/client/mem0.ts
+++ b/mem0-ts/src/client/mem0.ts
@@ -89,7 +89,7 @@ interface ClientIdentity {
   projectId: string | number | null;
 }
 
-// Shares one ping per (host, api key) across clients; FIFO-capped, not a leak.
+// Shares one ping per (host, api key) across clients; FIFO-capped.
 const IDENTITY_CACHE_MAX = 50;
 const identityByCredentials = new Map<string, Promise<ClientIdentity>>();
 
--- a/mem0-ts/src/client/mem0.ts
+++ b/mem0-ts/src/client/mem0.ts
@@ -91,6 +91,7 @@ export default class MemoryClient {
   headers: Record<string, string>;
   client: any;
   telemetryId: string;
+  private initialized: Promise<void>;
 
   _validateApiKey(): any {
     if (!this.apiKey) {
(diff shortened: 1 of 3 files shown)


kartik-mem0:
please sign the cla @pratikgajjar

deshraj (approved):

PR #6792, opened 2026-08-04
fix(plugin): use a Codex-safe MCP auth variable

### LINKED ISSUE #6783: [Codex plugin] Exported MEM0_API_KEY is filtered before MCP startup in managed environments
### Component

Plugin

### Summary

  I am following the official Mem0 Codex integration guide and installing the recommended marketplace plugin, but the plugin fails to initialize because Codex reports that `MEM0_API_KEY` is not set.

  The variable is correctly exported from Bash and visible to ordinary child processes immediately before launching Codex. In this managed Codex environment, variables whose names contain `KEY`, `SECRET`, or
  `TOKEN` appear to be filtered before MCP initialization.

  The Mem0 plugin manifest hardcodes:

  ```json
  "bearer_token_env_var": "MEM0_API_KEY"
  ```

  Consequently, there is no narrow way to select a differently named credential variable for the plugin.

  Related but distinct issue: #6346 covers `mem0 init` saving a key without exporting it. In this report, the key is already exported successfully and is removed later during Codex/MCP startup.

  ### Steps to Reproduce

  1. Add the API key to Bash as documented:

  ```bash
  echo 'export MEM0_API_KEY="m0-REDACTED"' >> ~/.bashrc
  source ~/.bashrc
  ```

  2. Confirm that it is set and exported to child processes:

  ```bash
  test -n "$MEM0_API_KEY" && echo "set" || echo "missing"
  # set

  export -p | grep 'declare -x MEM0_API_KEY='
  # Shows MEM0_API_KEY as exported

  env | cut -d= -f1 | grep -x MEM0_API_KEY
  # MEM0_API_KEY
  ```

  3. Install the recommended plugin:

  ```bash
  codex plugin marketplace add mem0ai/mem0
  codex plugin add mem0@mem0-plugins
  ```

  4. Launch Codex from the same shell:

  ```bash
  codex
  ```

  5. Codex reports:

  ```text
  ⚠ MCP client for `mem0` failed to start: MCP startup failed:
  Environment variable MEM0_API_KEY for MCP server 'mem0' is not set

  ⚠ MCP startup incomplete (failed: mem0)
  ```

  The direct-MCP configuration from the same guide exhibits the same behavior:

  ```bash
  codex mcp add mem0 \
    --url https://mcp.mem0.ai/mcp/ \
    --bearer-token-env-var MEM0_API_KEY
  ```

  ### Diagnostic Results

  Launching Codex with the following override allows Mem0 to initialize successfully:

  ```bash
  codex -c shell_environment_policy.ignore_default_excludes=true
  ```

  Codex’s environment-policy implementation contains automatic exclusion patterns for:

  ```text
  *KEY*
  *SECRET*
  *TOKEN*
  ```

  Relevant Codex source:

  https://github.com/openai/codex/blob/rust-v0.146.0/codex-rs/protocol/src/shell_environment.rs

  For direct MCP only, using a variable name that avoids those patterns also works:

  ```bash
  export MEM0_AUTH="$MEM0_API_KEY"

  codex mcp add mem0 \
    --url https://mcp.mem0.ai/mcp/ \
    --bearer-token-env-var MEM0_AUTH
  ```

  The plugin cannot use this narrower workaround because its bundled MCP manifest fixes the variable name to `MEM0_API_KEY`:

  https://github.com/mem0ai/mem0/blob/main/integrations/mem0-plugin/.codex-mcp.json

  ### Expected Behavior

  Following the official plugin installation guide should provide the plugin with its API key when `MEM0_API_KEY` is correctly exported.

  Alternatively, the plugin should provide a secure supported mechanism for selecting the bearer-token environment variable, such as:

  - a configurable `bearer_token_env_var`;
  - a plugin option;
  - a secure config/file-based credential fallback; or
  - documented Codex configuration that exposes only the Mem0 credential.

  ### Actual Behavior

  The exported `MEM0_API_KEY` is unavailable when the bundled Mem0 MCP server initializes, producing a misleading “environment variable is not set” error.

  The only confirmed plugin workaround is:

  ```toml
  [shell_environment_policy]
  ignore_default_excludes = true
  ```

  This is broader than necessary because it can expose every inherited environment variable containing `KEY`, `SECRET`, or `TOKEN` to agent-run commands.

  ### Environment

  - mem0 version: Plugin `0.2.13`
  - Codex version: `codex-cli 0.146.0`
  - Python/Node version: Not applicable; using the hosted Mem0 MCP server
  - OS: Linux on an exe.dev VM
  - Shell: Bash
  - Codex environment: Managed permission profile
  - MCP URL: `https://mcp.mem0.ai/mcp/`
  - Installation method: Mem0 plugin marketplace

Soundcreates:
Fixes #6783

---

Managed Codex environments filter inherited variables containing KEY, SECRET, or TOKEN before MCP startup. Keep MEM0_API_KEY for the SDK and lifecycle hooks, while using the narrow MEM0_AUTH alias for the Codex MCP manifest and documenting the migration.

Tests: validated the MCP manifest with Python's JSON parser.

### DIFF
--- a/integrations/mem0-plugin/.codex-mcp.json
+++ b/integrations/mem0-plugin/.codex-mcp.json
@@ -2,7 +2,7 @@
   "mcpServers": {
     "mem0": {
       "url": "https://mcp.mem0.ai/mcp",
-      "bearer_token_env_var": "MEM0_API_KEY"
+      "bearer_token_env_var": "MEM0_AUTH"
     }
   }
 }
--- a/integrations/mem0-plugin/README.md
+++ b/integrations/mem0-plugin/README.md
@@ -76,10 +76,16 @@ Codex reads MCP servers from `~/.codex/config.toml` as TOML. Add:
 ```toml
 [mcp_servers.mem0]
 url = "https://mcp.mem0.ai/mcp"
-bearer_token_env_var = "MEM0_API_KEY"
+bearer_token_env_var = "MEM0_AUTH"
 ```
 
-Export `MEM0_API_KEY` in your shell and restart Codex. `codex mcp add` only supports stdio servers, so HTTP servers like Mem0's must be added via `config.toml` directly (or via the **Plugins → Connect to a custom MCP → Streamable HTTP** UI in the Codex app).
+Export the credential under `MEM0_AUTH` for Codex and restart Codex:
+
+```bash
+export MEM0_AUTH="$MEM0_API_KEY"
+```
+
+Codex filters inherited variables containing `KEY`, `SECRET`, or `TOKEN` in managed environments. `MEM0_API_KEY` remains the SDK and hook variable; `MEM0_AUTH` is the narrow MCP-only alias. `codex mcp add` only supports stdio servers, so HTTP servers like Mem0's must be added via `config.toml` directly (or via the **Plugins → Connect to a custom MCP → Streamable HTTP** UI in the Codex app).
 
 **Option B — Sideload the plugin** (full experience: MCP + skills + opt-in hooks):
 


kartik-mem0:
Thank you for preparing the credential-alias change. We have rechecked #6783 and are closing this PR because #7203 replaced the old plugin files it changes. The current native Codex plugin explicitly forwards `MEM0_API_KEY`, and the Codex HTTP bearer-token lookup is separate from shell-command environment filtering, so the earlier diagnosis does not justify changing the credential name. We appreciate your effort and have documented the updated findings on #6783.

PR #6794, opened 2026-08-04
feat(mem0-ts): return embedding failures with labels and an index

### LINKED ISSUE #5509: Silent memory loss in the TypeScript SDK (`mem0ai`) when batch embedding partially fails — TS counterpart of #5245

shashiKundur1:
## Linked Issue

Closes #5509. Same taxonomy as #5245 on the Python side.

This replaces #5554, which @kartik-mem0 closed in the backlog cleanup. All the earlier review discussion is still there and nothing is lost.

## Scope: this is piece 1 of three

The original body offered to split this into three decisions and take only the first. @inland-taipen, who wrote the `{ results, failed }` sketch on #5245, took that offer up, so this PR is now piece 1 alone:

**In:** validate every returned vector before persist, classify the failure, always return `failed`.

**Out, as follow-ups once the shape is in `main`:**

1. `raiseOnPartialFailure`, @Bartok9's preserve-then-raise from #5546.
2. `retryFailed()` and the remediation strategies.

`_payload` and `_memoryId` went with `retryFailed()`, since they existed only to feed it. @inland-taipen's point that an underscore-prefixed `readonly` field reads as "internal, safe to drop" to a logging or serialization layer is right, so when the retry lands those fields come back without the underscore and with the round-trip requirement written on `AddResult` rather than left implied.

@Adelagric has mirrored the same split on the Python side, so the two SDKs land the same minimal shape.

## What it does

`add()` returns `{ results, failed }`. `results` are the memories that saved, `failed` are the ones that did not, each carrying:

- `errorClass`: `provider_error`, `validation_error` or `internal_error`
- `errorCode`: `EMBED_001` transient, `EMBED_002` validation, `EMBED_003` auth, `EMBED_004` internal
- `remediation`: `retry`, `reconfigure` or `escalate`
- `retryAfter`, when the provider gave one
- `index`, the position of the failure in the call that produced it

```ts
const { results, failed } = await memory.add(messages, { userId });
for (const f of failed) {
  if (f.remediation === "retry") console.warn(`dropped: ${messages[f.index]}`);
}
```

Every returned vector is checked for NaN, Inf, empty and wrong dimension before it is persisted, so a structurally bad vector is never stored. If the vector store insert itself fails, that memory is reported in `failed` rather than returned as if it saved.

A short `embedBatch` return is `provider_error` / `retry`, not `validation_error`. A missing vector is not a bad vector: the provider returned nothing for that text, which is a provider fault and the most retryable failure in the taxonomy. NaN, empty and wrong dimension stay `validation_error`, since those are caught by inspecting output the provider did return.

### On `index`

Correlating a failure back to its input by `text` alone does not work. Two identical messages in one `add()` produce two failures a caller cannot tell apart, and on `infer: false` they are separate `embed()` calls, so one can rate-limit while the other succeeds. The caller then has one entry in `failed` and one in `results` carrying the same text and no way to tell which of theirs is which.

Every entry carries an index, and every entry from one call sits in the same coordinate space. A mixed space is worse than no index at all.

- **`infer: false`** is the position in the `messages` array you passed, so `messages[f.index]` is the message that failed. `parse_vision_messages` drops system messages before the write path sees them, so the original positions are carried across rather than recomputed downstream.
- **`infer: true`** is the position in the `embedBatch` request. That list is built by extraction, so the index separates two identical facts but does not lead back to the message one came from. Extraction carries no per-fact provenance, and the type doc says that out loud rather than implying the two paths give the same guarantee.
- **Insert failures** come back from `persistRecords` as records rather than texts, because a record position is not a caller position: the record list skips system messages and anything that already failed to embed.

### On `EMBED_004`

`internal_error` had no code of its own. The insert path set the class and no code at all, and the fallback branch of `projectError` paired `internal_error` with `EMBED_001`, which tells a caller "provider blip, retry may work" about a persistence fault.

Class and code are deliberately not 1:1. `errorClass` answers whose fault it is, `errorCode` answers which fault, which is why `provider_error` spans `EMBED_001` (rate limit, 5xx, network) and `EMBED_003` (401/403, non-transient 4xx). Collapsing them into pairs would make one of the two fields stop carrying anything.

## On "failed: [] certifies success"

This was the objection that stopped the Python PR (#5557), so I want to answer it up front.

`failed` is always present, never omitted when empty. The doc comment on `AddResult` states exactly which paths feed it: LLM extraction (`infer: true`), raw writes (`infer: false`), and vector store insert failures on both. Entity linking embeds are not covered and still fail quietly. I left them out deliberately rather than half-covering them, because a dropped entity costs recall but never loses a stored memory, and @kartik-mem0 already called that a separate follow-up in the #5509 triage. It is filed as #6795.

So an empty `failed` is a scoped claim about the paths named in the type, not a blanket "nothing anywhere failed".

## Credit

The `{ results, failed }` shape is @inland-taipen's from #5245, as is the `index` field. The taxonomy and the detection-point rule are @Adelagric's from the gist on #5509, and the vector validation exists because @Adelagric pointed out that a NaN vector never throws and so never showed up in my original repro.

## Type of Change

## Breaking Changes

`add()` returned `SearchResult` and now returns `AddResult`, which adds `failed`. Code reading `result.results` is unaffected.

The one behaviour change worth calling out: on `infer: false`, a failed embed used to reject the `add()` promise, and now it comes back in `failed` instead. Anyone relying on a `try/catch` there will need to read `failed`.

I also took `failed` back off `SearchResult`. It was shared with `search()` and `getAll()`, so both were advertising a field they never populate, and it forced `types/index.ts` to import from `memory/errorRetry`, which put a cycle in the module graph. `add()` has its own return type now.

## Test Coverage

- [x] I tested manually

27 tests in `mem0-ts/src/oss/tests/embedding-failures.test.ts`, all passing.

Full suite: 1468 passed, 42 skipped, 2 failed. The 2 are in `memory.embedding-action.test.ts` and fail identically with this branch's changes stashed, so they are not from this PR. `pnpm run build` is clean, including `prettier --check`.

Worth flagging for the repo generally: that file used to take 175 seconds and fail a different test on every run. Every `add()` fires a real telemetry POST and each test sat about 5 seconds waiting on the network. Stubbing the telemetry module the way `vector-stores-compat.test.ts` does takes it to 0.42 seconds and makes it deterministic. `memory.embedding-action.test.ts` looks like the same problem.

## Checklist

Docs are the one gap. Happy to add `docs/` pages in this PR now that the shape is settled, or as an immediate follow-up, whichever you prefer.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -152,6 +152,7 @@
                       "open-source/features/custom-instructions",
                       "open-source/features/rest-api",
                       "open-source/features/openai_compatibility",
+                      "open-source/features/embedding-failures",
                       "platform/features/memory-expiration"
                     ]
                   },
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -240,6 +240,7 @@ If the user is on a pre-current major (Python < 2, TS < 3, or Platform `output_f
 - [Custom Instructions (OSS)](https://docs.mem0.ai/open-source/features/custom-instructions) [OSS]: Use when tailoring extraction prompts in OSS.
 - [REST API Server](https://docs.mem0.ai/open-source/features/rest-api) [OSS]: Use when exposing a self-hosted Mem0 as a FastAPI service.
 - [OpenAI Compatibility](https://docs.mem0.ai/open-source/features/openai_compatibility) [OSS]: Use when hitting an OpenAI-compatible endpoint with self-hosted.
+- [Embedding Failures](https://docs.mem0.ai/open-source/features/embedding-failures) [OSS]: Use when inspecting which memories `add()` dropped and whether to retry them.
 
 ## Integrations
 
--- a/docs/open-source/features/embedding-failures.mdx
+++ b/docs/open-source/features/embedding-failures.mdx
@@ -1,0 +1,120 @@
+---
+title: Embedding Failures
+description: See exactly which memories were dropped when an embedding call fails.
+icon: "triangle-exclamation"
+---
+
+When you call `add()` with a batch of messages, some of them can fail to embed while the rest succeed. A rate limit hits halfway through, the provider returns a short batch, one vector comes back with the wrong dimension. The Node SDK reports those drops back to you instead of losing them quietly, so you can decide what to do about each one.
+
+<Info>
+  **You'll use this when…**
+  - You add memories in batches and need to know whether every one of them landed.
+  - You want to retry only the items that failed, not the whole call.
+  - You need to tell a transient provider blip apart from a misconfiguration that retrying will never fix.
+</Info>
+
+<Note>
+  This is the OSS `Memory` class in the Node SDK (`mem0ai/oss`). The Python SDK exposes the same `error_code` vocabulary and the same positional contract.
+</Note>
+
+## What `add()` returns
+
+`add()` returns an `AddResult`. `failed` is always present, and an empty array means nothing was dropped on a covered path.
+
+```typescript
+interface AddResult {
+  results: MemoryItem[];
+  failed: EmbeddingFailure[];
+}
+```
+
+Each entry in `failed` is plain data, so it survives a JSON round trip and can go straight into a queue or a log line.
+
+```typescript
+interface EmbeddingFailure {
+  text: string;
+  index: number;
+  errorClass: "provider_error" | "validation_error" | "internal_error";
+  remediation: "retry" | "reconfigure" | "escalate";
+  errorCode?: "EMBED_001" | "EMBED_002" | "EMBED_003" | "EMBED_004";
+  retryAfter?: number;
+  error: string;
+}
+```
+
+## Reading a failure
+
+Two fields answer two different questions. `errorClass` says whose fault it was. `errorCode` says which fault it was. They are deliberately not one to one: `provider_error` covers both a rate limit you should retry and an auth failure you should not.
+
+| `errorClass` | `remediation` | `errorCode` | What happened |
+| --- | --- | --- | --- |
+| `provider_error` | `retry` | `EMBED_001` | 429, a 5xx, a network reset, or the provider returned no vector for that text. Retrying can work. |
+| `provider_error` | `escalate` | `EMBED_003` | 401, 403, or another non-transient 4xx. Needs a key or a quota change, not a retry. |
+| `validation_error` | `reconfigure` | `EMBED_002` | The vector came back with the wrong dimension. Your store and your embedder disagree. |
+| `validation_error` | `escalate` | `EMBED_002` | The vector came back empty or contained a non-finite number. |
+| `internal_error` | `escalate` | `EMBED_004` | A mem0-side fault, such as a vector store insert that failed after the text embedded fine. |
+
+`retryAfter` is set in seconds when the provider sent a `Retry-After` header, so you can honour its backoff instead of guessing one.
+
+<Warning>
+  Branch on `remediation`, not on the error string. Provider messages change between versions and between providers; `remediation` is the field that tells you whether another attempt is worth making.
+</Warning>
+
+## Finding the message that failed
+
+`index` locates the failure in the call that produced it, so two identical texts in the same `add()` stay distinguishable. Every entry carries one, and every entry from a single call is in the same coordinate space, whether the drop happened during embedding or during the vector store insert. That means you can sort `failed` and compare entries directly.
+
+What `index` points at depends on how you called `add()`:
+
+| Call | `index` is | Round trip |
+| --- | --- | --- |
+| `infer: false` | The position in the `messages` array you passed | `messages[f.index]` is the message that failed |
+| `infer: true` | The position in the extracted fact list sent to the embedder | Separates duplicate facts, but does not lead back to a message |
+
+With `infer: false` the position is captured before any internal parsing, so it survives the parse step dropping or rewriting messages along the way. With `infer: true` the list is built by mem0 rather than by you, and extraction carries no per-fact provenance, so the index cannot be mapped back to an input message.
+
+## Handling the failures
+
+```typescript
+import { Memory } from "mem0ai/oss";
+
+const memory = new Memory();
+
+const { results, failed } = await memory.add(messages, {
+  userId: "alice",
+  infer: false,
+});
+
+console.log(`stored ${results.length}, dropped ${failed.length}`);
+
+const retryable = failed.filter((f) => f.remediation === "retry");
+const misconfigured = failed.filter((f) => f.remediation === "reconfigure");
+
+for (const f of misconfigured) {
+  console.error(`[${f.errorCode}] ${f.error}`);
+}
+
+if (retryable.length > 0) {
+  const wait = Math.max(...retryable.map((f) => f.retryAfter ?? 1));
+  await new Promise((r) => setTimeout(r, wait * 1000));
+  await memory.add(
+    retryable.map((f) => messages[f.index]),
+    { userId: "alice", infer: false },
+  );
+}
+```
+
+Because `index` points back into the array you passed, `messages[f.index]` gives you the original message with its role and metadata intact. You never have to reconstruct it from `f.text`.
+
+## What is covered
+
+<AccordionGroup>
+  <Accordion title="Reported in failed[]">
+    - Embedding failures on raw writes (`infer: false`).
+    - Embedding failures on LLM-extracted facts (`infer: true`).
+    - Vector store insert failures on both paths, as `internal_error`.
+  </Accordion>
+  <Accordion title="Not reported yet">
+    Entity linking embeds a second set of vectors for the graph layer. Those failures are tracked separately and do not appear in `failed[]`. A dropped entity costs recall on graph queries; it never costs you a stored memory, because the memory row is written on the path above.
+  </Accordion>
+</AccordionGroup>
--- a/mem0-ts/src/common/exceptions.ts
+++ b/mem0-ts/src/common/exceptions.ts
@@ -149,15 +149,33 @@ export const EMBED_ERROR_CODE = {
 export type EmbedErrorCode =
   (typeof EMBED_ERROR_CODE)[keyof typeof EMBED_ERROR_CODE];
 
-/** Raised when the embedding step fails. The TS parallel of Python's EmbeddingError. */
+/**
+ * Raised when one or more texts fail to embed during `add()`. The TypeScript
+ * parallel of Python's `EmbeddingError`.
+ *
+ * Successfully-embedded memories are persisted before this is thrown, so a
+ * caller catching it can retry only the dropped texts via `failedTexts` rather
+ * than re-adding the whole batch.
+ */
 export class EmbeddingError extends MemoryError {
+  /** Texts that failed to embed and were not persisted. */
+  readonly failedTexts: string[];
+  /** Number of memories persisted by the call before this was thrown. */
+  readonly persistedCount: number;
+
   constructor(
     message: string,
     errorCode: string = EMBED_ERROR_CODE.TRANSIENT,
-    options?: MemoryErrorOptions,
+    options: MemoryErrorOptions & {
+      failedTexts?: string[];
+      persistedCount?: number;
+    } = {},
   ) {
-    super(message, errorCode, options);
+    const { failedTexts = [], persistedCount = 0, ...rest } = options;
+    super(message, errorCode, rest);
     this.name = "EmbeddingError";
+    this.failedTexts = failedTexts;
+    this.persistedCount = persistedCount;
   }
 }
 
--- a/mem0-ts/src/common/exceptions.ts
+++ b/mem0-ts/src/common/exceptions.ts
@@ -139,6 +139,28 @@ export class MemoryQuotaExceededError extends MemoryError {
   }
 }
 
+// Stable error codes for embedding failures (parallels the Python SDK).
+export const EMBED_ERROR_CODE = {
+  TRANSIENT: "EMBED_001", // provider blip (429, 5xx, network) — retry may work
+  VALIDATION: "EMBED_002", // bad input/vector (wrong dim, NaN) — retry won't help
+  AUTH: "EMBED_003", // auth/permission — needs operator action
+} as const;
+
+export type EmbedErrorCode =
+  (typeof EMBED_ERROR_CODE)[keyof typeof EMBED_ERROR_CODE];
+
+/** Raised when the embedding step fails. The TS parallel of Python's EmbeddingError. */
+export class EmbeddingError extends MemoryError {
+  constructor(
+    message: string,
+    errorCode: string = EMBED_ERROR_CODE.TRANSIENT,
+    options?: MemoryErrorOptions,
+  ) {
+    super(message, errorCode, options);
+    this.name = "EmbeddingError";
+  }
+}
+
 // ─── HTTP Status → Exception Mapping ─────────────────────
 
 type MemoryErrorConstructor = new (
--- a/mem0-ts/src/common/exceptions.ts
+++ b/mem0-ts/src/common/exceptions.ts
@@ -141,41 +141,28 @@ export class MemoryQuotaExceededError extends MemoryError {
 
 // Stable error codes for embedding failures (parallels the Python SDK).
 export const EMBED_ERROR_CODE = {
-  TRANSIENT: "EMBED_001", // provider blip (429, 5xx, network) — retry may work
-  VALIDATION: "EMBED_002", // bad input/vector (wrong dim, NaN) — retry won't help
-  AUTH: "EMBED_003", // auth/permission — needs operator action
+  TRANSIENT: "EMBED_001", // provider blip (429, 5xx, network): retry may work
+  VALIDATION: "EMBED_002", // bad input/vector (wrong dim, NaN): retry won't help
+  AUTH: "EMBED_003", // auth/permission: needs operator action
+  INTERNAL: "EMBED_004", // mem0-side fault (store insert): retry won't help
 } as const;
 
 export type EmbedErrorCode =
   (typeof EMBED_ERROR_CODE)[keyof typeof EMBED_ERROR_CODE];
 
 /**
- * Raised when one or more texts fail to embed during `add()`. The TypeScript
- * parallel of Python's `EmbeddingError`.
- *
- * Successfully-embedded memories are persisted before this is thrown, so a
- * caller catching it can retry only the dropped texts via `failedTexts` rather
- * than re-adding the whole batch.
+ * Raised when a text fails to embed during `add()`. The TypeScript parallel of
+ * Python's `EmbeddingError`. Carried on `failed[]` as a label rather than
+ * thrown, so a partial failure never costs the memories that did embed.
  */
 export class EmbeddingError extends MemoryError {
-  /** Texts that failed to embed and were not persisted. */
-  readonly failedTexts: string[];
-  /** Number of memories persisted by the call before this was thrown. */
-  readonly persistedCount: number;
-
   constructor(
     message: string,
     errorCode: string = EMBED_ERROR_CODE.TRANSIENT,
-    options: MemoryErrorOptions & {
-      failedTexts?: string[];
-      persistedCount?: number;
-    } = {},
+    options: MemoryErrorOptions = {},
   ) {
-    const { failedTexts = [], persistedCount = 0, ...rest } = options;
-    super(message, errorCode, rest);
+    super(message, errorCode, options);
     this.name = "EmbeddingError";
-    this.failedTexts = failedTexts;
-    this.persistedCount = persistedCount;
   }
 }
 
--- a/mem0-ts/src/oss/src/index.ts
+++ b/mem0-ts/src/oss/src/index.ts
@@ -1,4 +1,10 @@
 export * from "./memory";
+export * from "./memory/errorRetry";
+export {
+  EmbeddingError,
+  EMBED_ERROR_CODE,
+  type EmbedErrorCode,
+} from "../../common/exceptions";
 export * from "./memory/memory.types";
 export * from "./types";
 export * from "./embeddings/base";
--- a/mem0-ts/src/oss/src/index.ts
+++ b/mem0-ts/src/oss/src/index.ts
@@ -1,5 +1,13 @@
 export * from "./memory";
-export * from "./memory/errorRetry";
+// Only the shapes a caller of add()/retryFailed() handles. The classifier,
+// validator and strategy internals stay unexported so they remain free to
+// change without a major version.
+export type {
+  AddResult,
+  EmbeddingFailure,
+  ErrorClass,
+  Remediation,
+} from "./memory/errorRetry";
 export {
   EmbeddingError,
   EMBED_ERROR_CODE,
--- a/mem0-ts/src/oss/src/index.ts
+++ b/mem0-ts/src/oss/src/index.ts
@@ -1,7 +1,7 @@
 export * from "./memory";
-// Only the shapes a caller of add()/retryFailed() handles. The classifier,
-// validator and strategy internals stay unexported so they remain free to
-// change without a major version.
+// Only the shapes a caller of add() handles. The classifier and validator
+// internals stay unexported so they remain free to change without a major
+// version.
 export type {
   AddResult,
   EmbeddingFailure,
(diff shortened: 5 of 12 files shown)


shashiKundur1:
@kartik-mem0 picking this back up per your note on #5554. Short version of what happened and what is different.

I did what you asked and rebased on `main`, but I could not reopen #5554. A rebase needs a force-push, and GitHub refuses to reopen a closed PR whose branch was force-pushed ("state cannot be changed. The branch was force-pushed or recreated"). So this is the fresh PR from the second half of your offer. #5554 still has the whole review thread on it.

The rebase itself was clean once resolved. `main` had moved 280 commits and there were six conflicts, all in the add pipeline, and I kept your side on every one: `role === "system"`, the `LLMError` throw, `stripIdentityKeys`, and the `"add"` memory action on `embed()` / `embedBatch()`.

The part I think is worth your time is that rebasing found three bugs in my own code, and they were invisible because every test I had asserted on what `add()` returned. None of them read a memory back out of the store, and a row written with the wrong payload still counts as `results.length === 1`.

- `infer: false` was dropping `hash`, `textLemmatized` and `createdAt`, because I inlined `createMemory()` and only kept `{ ...metadata, data }`. Those memories were saved and then unfindable by keyword or hybrid search, and they never deduped.
- `retryFailed()` could write rows with no `user_id` / `agent_id` / `run_id`, so a retried `infer: false` failure became a memory that `getAll`, `search` and `deleteAll` can never reach again.
- `retryFailed()` re-embedded without the `"add"` action, so a retry produced a different vector than the original write would have on a task-type-aware provider.

All the payload construction goes through one helper now, and `retryFailed()` refuses to persist a failure with no payload rather than writing an ownerless row. There are three new tests that read the row back, since that is the only thing that catches this.

On your review of the Python PR (#5557), the objection that `failed: []` can certify success on paths that are not covered is the right objection and it applies here too, so I have tried to answer it rather than wait for it. `failed` is always present now, and the type doc says exactly which paths feed it: `infer: true`, `infer: false`, and vector store insert failures on both. Entity linking is not covered and still fails quietly. I left it out on purpose instead of half covering it, since you called it a separate follow-up in the #5509 triage and a dropped entity costs recall but never loses a stored memory. I will file that as its own issue.

I am also not going to run the cross-SDK parity argument I used last time. #5557 was reduced to detection only in July, so Python does not carry this shape today and it would be a stale claim. This stands on the TypeScript case by itself.

On the three-PR situation: you said #5546 is the one to carry for raise-on-failure, and I am not trying to relitigate that. @Bartok9's preserve-then-raise is in here behind `raiseOnPartialFailure`, so if you take his PR the behaviour is the same and converging later is easy either way.

Last thing, and this is a genuine offer, not politeness. If the return shape plus `retryFailed()` plus the flag is too many decisions for one PR, I will split it and open just the first piece: validate every returned vector before persist and always return `failed`. That part is strictly additive, no new public method and no new option, and it is the piece that actually closes #5509. Tell me which you want and I will do it. I would rather land a third of this than have it sit for another cycle.

Full suite is 1475 passing. The 3 failures are in `memory.embedding-action.test.ts` and `memory.rerank.test.ts` and they fail identically on a clean `main` worktree, so they are not from this change. Build and `prettier --check` are clean.

One aside that is not about this PR but might be useful to you. My test file was taking 175 seconds and failing a different test on each run. The cause is that every `add()` fires a real telemetry POST, so each test waits about 5 seconds on the network. Stubbing the telemetry module the way `vector-stores-compat.test.ts` does takes the file from 175 seconds to 0.42 and makes it deterministic. The two upstream files above look like exactly the same problem, so that may be worth a look independent of this.

Thanks for the note on the cleanup, and no worries at all about the turnaround. cc @Adelagric @Bartok9 since you both shaped this.

Adelagric:
Thanks @shashiKundur1 — credit both ways, and the read-the-row-back tests are the real story here. Asserting only on `add()`'s return is exactly how the same class of bug stayed invisible (a row written with the wrong payload still counts as `results.length === 1`); finding three on rebase and pinning each with a test that reads the store back is the right instinct.

Two notes, both just about the two SDKs enumerating the same surfaces — nothing blocking:

- **`failed[]` scope.** You cover `infer:true` / `infer:false` / insert and scope out entity linking. The Python follow-up to #5557 (the taxonomy #5557 was reduced away from) covers `infer:true` / `infer:false` / procedural and currently scopes out insert as a persistence phase. So we diverge on two surfaces — insert (you cover, Python scopes) and procedural (Python covers, TS doesn't). Worth aligning when that PR is up, so a consumer keying on `error_class` sees the same coverage on both — I'll raise it there and link back.
- **`error_class` string values.** Keeping the emitted strings identical across SDKs (whatever the type is named) is what lets one dashboard read both — I'll match whatever lands here.

On the split offer: "validate every returned vector before persist + always return `failed`" is a clean minimal that closes #5509 by itself and maps 1:1 to the Python detection step, if the smallest reviewable piece is what's wanted.

shashiKundur1:
Thanks @Adelagric. On the two surfaces, one of them I think resolves itself.

**Procedural.** There is no procedural path in the TS SDK at all. `grep -ri procedural mem0-ts/src/` comes back empty, while Python has `MemoryType.PROCEDURAL`, the `memory_type` branch in `add()`, and `_create_procedural_memory` at `main.py:1983`. So that one is not a coverage gap on my side, it is a surface that does not exist yet. If TS ever grows procedural memories they should feed `failed[]` the same way, but there is nothing to enumerate today. Probably worth wording it in the Python PR as "procedural (Python only)" rather than as a divergence, so nobody later reads it as something TS dropped.

**Insert.** This one I would push back on gently, and not for taxonomy reasons.

In TS the Phase 6 per-item insert fallback catches `vectorStore.insert()` failures, `console.error`s them, and continues. The memory is gone and `add()` still returns success. That is the same swallow-and-continue shape as the embed drop in #5509, just one phase later, so leaving it out of `failed[]` would mean an empty `failed` certifying success on a path that really does lose a memory. That is the specific thing I was trying not to do.

Worth noting the Python side has the same shape rather than a raising one. In the #5557 review @kartik-mem0 listed the Phase 6 per-item insert fallback at `main.py:1093-1099` as one of the paths that "still drop silently" while `failed: []` reports success, under the finding about false negatives. So scoping insert out on the Python side may reopen exactly that. Not my call to make on that PR, but it seemed worth flagging before it gets settled, since you asked about alignment.

Happy to go whichever way once a maintainer rules. If the answer is that persistence is out of scope for the taxonomy on both sides, I will drop it from TS and say so explicitly in the type doc rather than leave it ambiguous.

**Wire values**, so you can match exactly:

```
error_class:  provider_error | validation_error | internal_error
remediation:  retry | reconfigure | escalate
error_code:   EMBED_001 (transient) | EMBED_002 (validation) | EMBED_003 (auth)
```

Also `retryAfter` in seconds when the provider sends it. One change since you last looked: a short `embedBatch` return, meaning the provider gave back fewer vectors than texts, is now `provider_error` / `retry` rather than `validation_error`. A missing vector is not a malformed vector, and labelling it escalate meant the most retryable failure in the set was marked never-retry.

**On the split.** Good to hear you read it the same way. That piece is ready to go and I can put it up as its own PR whenever it is wanted, keeping this one open as the full picture. I would rather not open it unprompted and end up with two of my own PRs touching the same function, so I will wait for a word from @kartik-mem0 on which entry point he prefers.

Bartok9:
Thanks for the ping @shashiKundur1 — and for keeping preserve-then-raise behind `raiseOnPartialFailure` so it stays compatible with #5546 either way.

Agree the read-back tests are the important bit; return-shape-only asserts hide exactly the payload bugs you found on rebase.

No objection from me on scope/split — happy for @kartik-mem0 to pick the entry point. Wire values look good to standardize on once one side lands.

shashiKundur1:
Thanks @Bartok9. Appreciate you folding preserve-then-raise in early, that is the reason the flag in here lines up with #5546 instead of fighting it.

So where this actually stands now: @Adelagric, @Bartok9 and I agree on the shape, the wire values, and that the read-back tests are the part that matters. The "divergent designs" concern from the #5509 triage is settled between the three of us.

@kartik-mem0 that leaves one question, and it is a one word answer:

- **whole**: review this PR as it stands (validate before persist, `{ results, failed }`, `retryFailed()`, opt-in `raiseOnPartialFailure`)
- **split**: I open a separate PR with just "validate every returned vector before persist, and always return `failed`". Additive, no new public method, no new option, closes #5509 on its own. This PR stays open as the full picture.

I would suggest split, since it is the smallest reviewable thing that actually fixes the bug, and both @Adelagric and @Bartok9 have said they are fine either way. Either version I can have up the same day. No preference beyond wanting it to land.

Adelagric:
Conceded on insert — and you're right about why. I'd scoped it as "persistence, not embedding," but @kartik-mem0 listed the Phase 6 insert fallback under the same false-negative finding in the #5557 review, so scoping it out would have reopened exactly that. The Python follow-up now surfaces a still-failing insert as `internal_error` and drops it from `results`, so an unpersisted memory never counts as success — same shape as your TS side.

Procedural: agreed, I'll word it "procedural (Python only)" rather than a divergence, since there's no TS path to enumerate.

Wire values line up — `error_class` is `provider_error` / `validation_error` / `internal_error` on both, and a short `embed_batch` return is `provider_error` on the Python side too (a missing vector is retryable, not malformed — same call). `remediation` / `error_code` / `retryAfter` are a clean superset; `error_class` is what a cross-SDK consumer keys on and I'll match the strings exactly. Happy to carry `remediation` + `error_code` on the Python `failed[]` too if we want them identical field-for-field.

So the two sides now enumerate the same surfaces — embed phase + insert, entity linking out on both. I'll link the Python PR here when it's up.

shashiKundur1:
On the open question: match `error_code`, leave `remediation` free.

`error_code` is what a consumer switches on. If the two SDKs return different strings for the same condition, matching `error_class` stops meaning much one level down. It costs nothing to keep identical and it is the field that actually has to be identical.

`remediation` is human prose. Making it byte-identical across two SDKs buys a consumer nothing and turns every copy edit into a cross-SDK change. Same field name, same meaning, wording free to drift. That keeps the contract on the machine-readable fields where it belongs.

So field-for-field on `error_class` and `error_code`, same-shape on `remediation`.

Agreed on the rest. `provider_error` for a short `embed_batch` return is right for the reason you give, a missing vector out of the same call is retryable rather than malformed. And "procedural (Python only)" reads better than calling it a divergence, since there is no TS path to enumerate against it.

Link the Python PR here when it is up and I will cross-link from this side, so the two read as one change split across the SDKs rather than two separate asks.

inland-taipen:
Author of the `{ results, failed }` sketch on #5245 here — thanks for carrying it, and for the credit.

On the open question: **split**. The minimal piece — validate every returned vector before persist, always return `failed` — is the whole of what #5245 asked for. Everything else in here is remediation policy layered on a contract that has not shipped yet, and policy is much easier to argue about once the shape is in `main` and people are reading it. `retryFailed()` and `raiseOnPartialFailure` also both read better as "here is what the label is *for*" in a follow-up than as three decisions bundled into the PR that introduces the label.

Two notes on the shape itself, neither blocking, both from thinking about how a caller consumes this.

**`failed[]` has no way to correlate back to the input.** `EmbeddingFailure` carries `text` and nothing positional. In the bulk APIs this pattern comes from, a failure item maps to its request item by position, which is what lets a caller do `inputs[f.index]` and rebuild context the SDK never saw. Here, on `infer: true` the text is LLM-extracted, so there is no route back to the message that produced it at all. On `infer: false`, `text` is a usable key only if it happens to be unique within the call, and nothing enforces that — two identical messages in one `add()` produce two failures a caller cannot tell apart. An `index` field (position in the extracted/raw text list) would make it deterministic and costs one number.

**The `_` prefix on `_payload` fights the round-trip.** The comment on `EmbeddingFailure` says "plain data so it survives a JSON round-trip", and the natural use of that is the obvious one: push `failed[]` onto a dead-letter queue and retry out of process later. `_payload` is what makes that work, since it is where the tenancy lives, and `retryFailed()` hard-fails without it — correctly, an ownerless row is worse than no row. But an underscore-prefixed `readonly` field reads as "internal, safe to drop", and plenty of logging and serialization layers will drop or redact it on exactly that signal. Either it is part of the contract and should lose the underscore, or the `AddResult` doc should say out loud that `failed[]` has to be round-tripped intact for `retryFailed()` to work.

The rest looks right to me. Agreed that a short `embedBatch` return is `provider_error`: a missing vector out of a call that otherwise succeeded is the provider under-delivering, not a malformed vector, and labelling it escalate would have marked the most retryable failure in the taxonomy never-retry. And `retryFailed()` returning `AddResult` rather than a bespoke type is the right call — a caller can loop it with their own backoff without a second shape to learn.

shashiKundur1:
Taking the split. Piece 1 on its own, and I will reduce this PR in place rather than open a new one, since the discussion and the link to #5509 both live here.

So this becomes: validate every returned vector before persist, classify the failure, always return `failed`. `raiseOnPartialFailure` and `retryFailed()` come out and go up as follow-ups once the shape is in `main`.

On `index`, agreed, and it is worth being precise about what it points at, because it is not the same thing on both paths.

On `infer: false` it is the position in the `messages` array the caller passed, so `messages[f.index]` works and system messages keep their slot. That case is a real gap today: two identical messages go through two separate `embed()` calls, one can rate-limit while the other succeeds, and the caller ends up with one entry in `failed` and one in `results` carrying the same text and no way to tell which of theirs is which.

On `infer: true` it is the position in the extracted fact list. That makes two identical facts distinguishable, but you are right that it does not get anyone back to the message that produced them, because extraction carries no per-fact provenance. Better to say that in the doc than to imply both paths give the same guarantee.

One wrinkle: insert failures come back out of `persistRecords` as texts, so they would be the one kind of entry without an index. An optional index is worse than no index, so that returns positions instead and every entry carries one.

On `_payload`, I think the split answers it. Those fields exist only to feed `retryFailed()`, so they go out with it and the naming gets settled in the PR where they actually do something. When that lands it will be without the underscore, and with the round-trip requirement written on `AddResult` rather than left implied. Your read of what an underscore signals to a serializer is right, and the current comment is doing none of that work.

@kartik-mem0 flagging in case you would rather this stay as one. Otherwise I will push the reduced version here.

Adelagric:
Mirrored the split on the Python side too — the first piece is now the same minimal shape as yours: validate every returned vector before persist, classify, always return `failed`. `raise_on_partial_failure` comes out and goes up as a follow-up where a maintainer rules raise-vs-return, same call you made.

Field parity:

- `error_class` + `error_code` byte-identical (`provider_error`/`EMBED_001`, `validation_error`/`EMBED_002`, `internal_error`/`EMBED_003`); `remediation` same field + meaning (`retry` / `reconfigure` on dimension-mismatch / `escalate`).
- `index` on every entry, same per-path meaning as yours: `messages` position on `infer:false`, extracted-fact-list position on `infer:true` (documented as *not* mapping back to the source message, since extraction carries no per-fact provenance), record position on insert. Every entry carries one — no optional index.

One gap I left, same as before: your `classifyEmbedError` splits a thrown provider error on status (401/403/4xx → `EMBED_003` / escalate); the Python provider path is `EMBED_001` / retry across the board — I don't inspect the exception status yet. Say the word and I'll mirror that sub-case too.

Staged in the Python follow-up; I'll link it here when #5557 lands and you can cross-link, so the two read as one change.

shashiKundur1:
Good, that lines the two up. Two corrections on the table before you bake it in, and one of them is mine.

**`EMBED_003` is not `internal_error` on this side.** It is the auth/permission code and it rides `provider_error`. A 401, or a non-transient 4xx, is the provider rejecting the call rather than mem0 breaking, so the class stays `provider_error` and the code carries the which-fault part. `internal_error` over here is the vector store insert path.

Which means class and code are deliberately not 1:1, and I would keep it that way. `error_class` answers whose fault it is, `error_code` answers which fault. `provider_error` spans `EMBED_001` and `EMBED_003` because a rate limit and a bad key are the same party failing in two ways a caller has to handle differently. Collapse them into a pair and one of the two fields stops carrying anything.

**On the status split: yes, mirror it, and I do not think it is optional.** It is the only thing that produces `EMBED_003` at all. Without it a 401 comes back as `EMBED_001` / retry on Python, so a consumer with backoff keyed on the code sits in a loop against a bad API key on one SDK and escalates correctly on the other. That is the exact divergence this parity work exists to stop, and it is the sub-case where getting it wrong costs the most.

**The gap your table turned up is mine.** `internal_error` has no code of its own. The insert path sets the class and no `errorCode` at all, and the fallback branch in `projectError` pairs `internal_error` with `EMBED_001`, which tells a caller "provider blip, retry" about a persistence fault. If we are committing to every entry carrying both fields, that needs a fourth code, `EMBED_004`, for the internal path. Taking it in piece 1 on my side, worth mirroring on yours.

One thing on `index`. "Record position on insert" is the case to watch. On `infer: false` the record list skips system messages and skips anything that already failed to embed, so a record position and a `messages` position are different numbers. Every entry in one `failed[]` has to sit in the same coordinate space or `messages[f.index]` breaks for exactly the subset a caller cares most about. So insert failures carry the original index threaded through rather than their position in the record array, and the same against the extracted list on `infer: true`.

Rest matches. Link the Python PR when #5557 lands and I will cross-link from here.

shashiKundur1:
Pushed. This is piece 1 on its own now, and the body has been rewritten to match rather than describing three things and offering to split them.

Out: `raiseOnPartialFailure`, `retryFailed()` and the strategies, and `_payload` / `_memoryId` with them. In: the validation, the taxonomy, `failed` always returned, plus the two things that came out of this thread.

**`index` is on every entry.** One detail worth passing on to @Adelagric, because the Python side will hit the same thing and it is not visible from the type. Writing the loop index straight onto the failure gives you the wrong number. `parse_vision_messages` drops system messages before the write path ever sees the array, so on `infer: false` the list being iterated is already shorter than the caller's and every index after a system message is off by one. The test caught it, not the review: I had asserted index 2 for the third message and got 1. The original positions are carried across the parse step now instead of being recomputed downstream.

The other two cases behave as described: `infer: true` is the position in the `embedBatch` request, and insert failures come back from `persistRecords` as records rather than texts so they land in the same space as the embed failures next to them. There is a test that puts an embed failure and an insert failure in one call and asserts both indices, since one coordinate space is the whole point.

**`EMBED_004` is in.** `internal_error` had no code of its own, and the fallback branch of `projectError` was pairing it with `EMBED_001`, which tells a caller "provider blip, retry may work" about a persistence fault. There is a test pinning that specifically, because it is the sort of thing that quietly comes back.

27 tests, all passing. Full suite 1468 passed, 2 failed, and the 2 fail identically with this branch stashed so they are not from here. Build clean including `prettier --check`.

@kartik-mem0 this is the small version. #5509 is `accepted` and the diff is one new module, one type export and the add path.

Adelagric:
All three corrected on the Python side — you were right on each.

- **`provider_error` spans `EMBED_001` + `EMBED_003`.** Split on the thrown status now (401/403/other 4xx → `EMBED_003` / escalate, 429/5xx → `EMBED_001` / retry), so a bad key escalates instead of looping. Agreed it's not optional — it's the only thing that produces `EMBED_003`.
- **`internal_error` → `EMBED_004`.** No longer borrowing the auth code; the insert path and any other mem0-side fault carry `EMBED_004` / escalate.
- **`index` is the source position, not the record position.** Insert failures thread the original index through (a `memory_id → index` map), so every entry in one `failed[]` sits in the same coordinate space and `messages[f.index]` holds. Good catch — the record list skipping system messages and failed embeds would have broken exactly the subset that matters.

Class/code stay decoupled as you framed it: class = whose fault, code = which fault. I'll carry `EMBED_004` in piece 1 to match yours. Link when #5557 lands.

shashiKundur1:
Good, that closes the taxonomy. One more thing on `index` that the memory_id map does not cover, and I only found it because a test failed on my side.

The question is whether you compute the source position before or after `parse_vision_messages`, because that function does not return what the caller passed in on either SDK, and it drops different things on each.

TypeScript drops **system messages** outright, so the array the write path iterates is already shorter than the caller's. I had written the loop index straight onto the failure, asserted index 2 for a third message and got 1.

Python keeps system messages (`memory/utils.py:186`), so that specific case is not your problem. But it silently drops three others, each with a bare `continue`:

- `utils.py:192`: `content is None`, which is the assistant tool-call turn that carries `tool_calls` and no content
- `utils.py:202`: a content list with no text parts
- `utils.py:209`: an `image_url` dict when `llm is None`

The first one is the one I would worry about. A tool-call turn with no content is what every agent framework produces, and agent frameworks are most of who calls `add()`. Any caller passing one gets every index after it off by one.

So the only definition of `index` that survives both implementations is the caller's original array position, taken before the parse step and carried through. That is what I ended up doing rather than recomputing downstream. Worth a check on your side even if the map already handles the insert path, since these two shift the index for different reasons.

Adelagric:
Checked, and you're right that the map alone didn't cover it — the Python parse drops turns too, just different ones than yours. `parse_vision_messages` keeps system messages but silently drops a content-less assistant tool-call turn, a content list with no text parts, and an image part when no vision LLM is configured. So the post-parse loop position was off by one after any tool-call turn — exactly the agent-framework shape you called out.

Fixed the same way you did: the caller's original array position, captured before the parse and threaded through. `parse_vision_messages` can now return, per surviving message, its original index (opt-in flag, existing callers untouched), `add()` threads them, and `infer:false` failures report those. Pinned with a test: `[user, assistant-tool-call, user]`, second user's embed fails → `failed[0].index == 2`, not 1.

So on both SDKs `index` is now defined identically: the caller's original position, taken before the parse step, one coordinate space per `failed[]` across embed and insert. That closes it from my side.

shashiKundur1:
That matches what `parse_vision_messages` actually does on the Python side. System is appended and continues, then there are three bare `continue` paths that drop a turn silently: `content is None`, a content list where `text_parts` comes back empty, and an `image_url` dict when `llm is None`. Different drops from the TS side, which throws system messages away instead and keeps the rest, but the same class of bug and the same fix.

So the contract we can both write down now:

`failed[i].index` is the position of the message in the array the caller passed to `add()`, captured before any parse step. `inputs[f.index]` round trips on both SDKs. One coordinate space per `failed[]`, so an embed failure and an insert failure from the same call are directly comparable and sortable.

Worth putting that on the return type rather than only in this thread, because the retry follow-up leans on it. A caller that retries has to hand back the same item it was given, and that only holds if the index never moves under it.

Left on mine: docs pages for `AddResult.failed` and the code table. Pushing those next, then this one is finished from my side.

shashiKundur1:
Docs are up in `21e7c1a`.

New page at `docs/open-source/features/embedding-failures.mdx`, linked from `docs.json` under Open Source / Features and from `llms.txt` with an `[OSS]` tag. `scripts/check-llms-txt-coverage.py` reports in sync.

It documents the `AddResult` and `EmbeddingFailure` shapes, the class / remediation / code table exactly as the code produces it, and the `index` contract with the `infer: true` limitation stated as a limitation rather than glossed over. The retry snippet in it type checks against the real types under the package tsconfig with no errors of its own, so it will not rot silently if the shape changes.

That was the last box. Nothing outstanding on my side now. `raiseOnPartialFailure` and `retryFailed()` go up as separate PRs once this shape is in main.

shashiKundur1:
@kartik-mem0 asking for the call you flagged in June.

Your [triage note on #5509](https://github.com/mem0ai/mem0/issues/5509#issuecomment-4704669930) said the raise vs return question needs a decision before anything merges. The same point came up on the Python side, and your [Jul 22 review on #5557](https://github.com/mem0ai/mem0/pull/5557#pullrequestreview-4757247067) called it a product call, not one for review to settle. That call has not been made yet, and it is the only thing between #5509 and a fix.

Where it stands: #5546, #5551 and #5554 are all closed. This is the only TS PR left on #5509, which carries `accepted` and `P1-high`. It has already been cut down to the return shape alone. The `raiseOnPartialFailure` flag and the retry helper came out per the split asked for in review, the same way Python was reduced in #5557.

Three shapes are on the table. Pick one and I will have this PR cut to exactly that within a day:

1. Return `{ results, failed }` with the class and code labels. That is what the PR is now.
2. Persist the good ones, then raise `EmbeddingError`. The #5546 shape, which your review called the cleanest raise implementation.
3. `console.warn` only, matching where Python landed in #5557.

One more thing. The CI run on this fork needs a maintainer to approve it, so nothing has run since the first push on Aug 4. Approving it lets the checks go green whichever way you rule.

PR #6797, opened 2026-08-04
fix(ts-oss): stop add() from mutating the caller's filters object

### LINKED ISSUE #6796: add() mutates the caller's filters object, leaking scope into the next add() and bypassing the required-scope check
In the TypeScript OSS SDK, `Memory.add()` destructures `filters` by reference and then writes the resolved scope into it:

```ts
const { filters = {}, infer = true } = config;
...
```

That object is the caller's. The scope survives the call, so a reused filters object carries the previous call's `user_id` into the next `add()`. The required-scope guard a few lines below reads the same mutated object, so it passes on a call that supplied no scope at all:

```ts
if (!filters.user_id && !filters.agent_id && !filters.run_id) {
  throw new Error("One of the filters: userId, agentId or runId is required!");
}
```

## Reproduction

```ts
const shared = {};
await memory.add([{ role: "user", content: "alice note" }], { userId: "alice", infer: false, filters: shared });
// shared is now { user_id: "alice" }

// No userId, agentId or runId. This should throw.
await memory.add([{ role: "user", content: "bob note" }], { infer: false, filters: shared });

await memory.getAll({ filters: { user_id: "alice" } });
// -> ["alice note", "bob note"]
```

Result, 10 runs out of 10: no error is thrown, and the second memory is stored under `user_id: "alice"` and comes back from alice's `getAll()`.

## Why this looks like an oversight rather than intent

`search()` and `getAll()` both build a fresh object from `config.filters` via `Object.fromEntries({ ...config.filters, ... })` before normalizing scope. `add()` is the only one of the three that writes back into the caller's object.

This is the same class of scope-integrity issue as #6343, which stopped `update()` metadata from overwriting `user_id`/`agent_id`/`run_id`.

## Impact

Any caller that reuses a filters literal across `add()` calls, which is a normal thing to do in a request loop or a shared config object, can attribute one user's memory to another and silently skip the scope validation.

## Fix

Copy the object before writing scope into it, matching `search()` and `getAll()`. I have a patch and a regression test ready and will open a PR.

Environment: `mem0-ts` on current `main`, `mem0-ts/src/oss/src/memory/index.ts`.

shashiKundur1:
## Linked Issue

Closes #6796

## Description

`add()` destructured `filters` by reference and wrote the resolved scope into it:

```ts
const { filters = {}, infer = true } = config;
```

So the caller's object came back carrying `user_id`. Reuse it on the next `add()` and that scope comes along, and since the required-scope guard reads the same object, a call with no scope at all passes the check:

```ts
const shared = {};
await memory.add([{ role: "user", content: "alice note" }], { userId: "alice", infer: false, filters: shared });
await memory.add([{ role: "user", content: "bob note" }], { infer: false, filters: shared });
await memory.getAll({ filters: { user_id: "alice" } });
// -> ["alice note", "bob note"]
```

10 runs out of 10: no throw, and bob's memory is stored under alice.

The fix is to copy the object before writing into it. `search()` and `getAll()` already build a fresh object from `config.filters`, so this makes `add()` consistent with them rather than adding a new pattern. Same class of scope-integrity fix as #6343.

## Type of Change

## Breaking Changes

None. The only behaviour that changes is the mutation of an input object, which was not documented and defeats a validation check.

## Test Coverage

`mem0-ts/src/oss/tests/memory.add-filters-isolation.test.ts`, 4 tests. The first three fail on current `main` and pass with the fix. The fourth covers scope keys passed in via `filters` and passes either way, so the fix does not narrow existing behaviour.

Full suite: 1443 passed. The 2 failing files (`memory.embedding-action`, `memory.rerank`) fail the same way on a clean `main` worktree and are unrelated to this change. `pnpm run build` clean.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -749,7 +749,11 @@ export class Memory {
       has_filters: !!config.filters,
       infer: config.infer,
     });
-    const { filters = {}, infer = true } = config;
+    // Copy: the scope keys below are written into this object, and mutating the
+    // caller's would carry this call's scope into their next add(). search() and
+    // getAll() already build a fresh object from config.filters.
+    const filters: SearchFilters = { ...(config.filters ?? {}) };
+    const { infer = true } = config;
     const metadata = stripIdentityKeys(config.metadata);
 
     // Validate and trim entity IDs
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -756,7 +756,11 @@ export class Memory {
       has_filters: !!config.filters,
       infer: config.infer,
     });
-    const { filters = {}, infer = true } = config;
+    // Copy: the scope keys below are written into this object, and mutating the
+    // caller's would carry this call's scope into their next add(). search() and
+    // getAll() already build a fresh object from config.filters.
+    const filters: SearchFilters = { ...(config.filters ?? {}) };
+    const { infer = true } = config;
     const metadata = stripIdentityKeys(config.metadata);
 
     // Validate and trim entity IDs
--- a/mem0-ts/src/oss/tests/memory.add-filters-isolation.test.ts
+++ b/mem0-ts/src/oss/tests/memory.add-filters-isolation.test.ts
@@ -1,0 +1,115 @@
+/// <reference types="jest" />
+/**
+ * add() writes the resolved scope (user_id / agent_id / run_id) into its
+ * filters object. If that object is the caller's, the scope survives the call
+ * and leaks into the next add() that reuses it, which also defeats the
+ * "one of userId, agentId or runId is required" guard.
+ *
+ * search() and getAll() already build a fresh object from config.filters.
+ */
+jest.mock("../src/utils/telemetry", () => ({
+  captureClientEvent: jest.fn().mockResolvedValue(undefined),
+  isTelemetryEnabled: jest.fn(() => false),
+}));
+
+import { Memory } from "../src/memory";
+
+const DIM = 1536;
+let seq = 0;
+
+function createMemory(): Memory {
+  const m = new Memory({
+    disableHistory: true,
+    vectorStore: {
+      provider: "memory",
+      config: {
+        collectionName: `test-add-filters-${seq++}`,
+        dimension: DIM,
+        dbPath: ":memory:",
+      },
+    },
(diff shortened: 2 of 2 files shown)


shashiKundur1:
Rebased on latest main. Still four lines, still clean.

One thing worth flagging, because #6377 landed after I opened this and it makes the bug reach further than the description says.

#6377 added this to `add()`:

```ts
if (filters.user_id) metadata.user_id = filters.user_id;
if (filters.agent_id) metadata.agent_id = filters.agent_id;
if (filters.run_id) metadata.run_id = filters.run_id;
```

`filters` is still the caller's object, so a scope left behind by an earlier call is now also stamped into the stored row's **metadata**, not just used to scope the write.

On current main:

```ts
const filters = {};
await memory.add(msgs, { userId: "alice", infer: false, filters });
await memory.add(msgs, { infer: false, filters });   // no userId anywhere
```

The second row comes back as `metadata: { user_id: "alice" }`. It never had a scope of its own. It inherited one through the object the caller reused.

#6377 was about stopping caller metadata from injecting an identity scope, same as #6342, #6367 and #6371. This is the same problem running the other way, the caller's filters object carrying a scope into the next call. `search()` and `getAll()` already build a fresh object from `config.filters`, so this only brings `add()` in line with them.

Four tests in the PR, three of them fail on main today.

shashiKundur1:
One more consequence of this, which only became visible to me after #6892 landed.

On current main (d70cc00) `add()` still destructures `filters` off `config` by reference at `index.ts:759` and then writes the resolved scope into it. That same object is handed to `buildSessionScope()` at `:855`, and the key it returns is what `db.getLastMessages(sessionScope, 10)` uses at `:863` to pull the recent conversation buffer into extraction.

So a caller who reuses one object across calls is not only writing to the wrong scope, they are reading from it as well.

Reproduced on d70cc00, 10 runs, identical every time. Two `add()` calls sharing one `filters` object, the second passing no `userId` at all:

```
SCOPES: ["user_id=alice","user_id=alice"]
FILTERS AFTER: {"user_id":"alice"}
```

The second call pulls Alice's conversation history into the extraction prompt for an unrelated caller. It also clears the required-scope guard at `:782`, since that guard reads the same object that was just mutated.

To be clear about what #6892 fixed and what it didn't: it made the key unambiguous once it is built, which was a real bug on its own. This is about what gets fed into it. With the copy in this PR the second call throws `One of the filters: userId, agentId or runId is required!`, which is what it should have done all along.

Rebased onto d70cc00. Same four tests, all green.

PR #6801, opened 2026-08-04
fix(cli): read config as utf-8

Ghraven:
## Problem

The Python CLI config file is read and written without an explicit text encoding. On systems whose default encoding is not UTF-8, a manually edited `~/.mem0/config.json` containing non-ASCII values, such as a user email/comment field, can fail to load or be interpreted inconsistently.

## Before / after

Before, `load_config()` and `save_config()` relied on the platform default encoding for JSON config I/O.

After, config reads and writes use `encoding="utf-8"`, matching JSON's usual wire/storage encoding and making behavior consistent across platforms. The backward-compat test fixture was updated the same way, and a regression test covers loading a UTF-8 config file with a literal non-ASCII value.

## Verification

- `PYTHONPATH=cli/python/src python -m pytest cli/python/tests/test_config.py`
- `git diff --check`

### DIFF
--- a/cli/python/src/mem0_cli/config.py
+++ b/cli/python/src/mem0_cli/config.py
@@ -90,7 +90,7 @@ def load_config() -> Mem0Config:
     config = Mem0Config()
 
     if CONFIG_FILE.exists():
-        with open(CONFIG_FILE) as f:
+        with open(CONFIG_FILE, encoding="utf-8") as f:
             data = json.load(f)
 
         config.version = data.get("version", CONFIG_VERSION)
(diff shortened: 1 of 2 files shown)


Ghraven:

kartik-mem0:
image


can you pleaes check why the cla ci is still pending?

Ghraven:
Thank you for checking. I signed the CLA in this thread with the requested statement, and I also tried the CLA assistant recheck link, but the `license/cla` status still appears pending on my side.

The repo CI checks for the changed Python CLI path are passing, so this looks like CLA assistant state rather than a code/test failure. I’m sorry for the extra friction here, and I’m happy to do any additional CLA step you prefer.

PR #6803, opened 2026-08-05
fix(claude-plugin): resolve the venv interpreter on Windows

Mohamed-Badran:
## The bug

The plugin's dependency installer assumes a POSIX venv layout:

| File | Line | Assumes |
| --- | --- | --- |
| `integrations/mem0-plugin/scripts/ensure_deps.sh` | 18 | `venv/bin/python3` exists |
| `integrations/mem0-plugin/scripts/ensure_deps.sh` | 29–30 | `venv/bin/pip` is executable |
| `integrations/mem0-plugin/scripts/on_session_start.sh` | 180 | `venv/bin/python3` exists |

A venv created by `python -m venv` on Windows has no `bin/` at all — the
interpreter lives at `venv/Scripts/python.exe`. So on Windows:

1. the line 18 probe never matches, so `needs_install=true` every session;
2. `venv/bin/pip` does not exist, so the install is treated as failed;
3. the failure branch runs `rm -f "${REQ_STAMP}"` and `touch .install-failed`.

Step 3 is what makes it stick. The SessionStart guard is

```sh
diff -q "${CLAUDE_PLUGIN_ROOT}/requirements.txt" "${CLAUDE_PLUGIN_DATA}/requirements.txt" \
  >/dev/null 2>&1 || "${CLAUDE_PLUGIN_ROOT}/scripts/ensure_deps.sh"
```

Once the stamp is deleted the `diff` fails forever, so `ensure_deps.sh` re-runs
and re-fails on every session, each time showing `Installing mem0 SDK...` for the
duration of the lock wait. A `.install-lock` directory left behind by a session
that was killed mid-install adds a 60s stall on top. The mem0ai SDK is never
importable, so the `_VENV_PY` check at `on_session_start.sh:181` falls through to
a system `python3` that does not have it — `auto_setup_categories.py` then fails
at its lazy `from mem0 import MemoryClient`.

## The fix

Resolve the interpreter across both layouts — `bin/python3`, `bin/python`,
`Scripts/python.exe` — and install through `python -m pip` so only one path has
to be resolved rather than one per executable.

No behaviour change on POSIX: `bin/python3` is still the first candidate and
still matches.

## Verification

Windows 11, Git Bash, CPython 3.14.6, venv at `Scripts/`:

| | before | after |
| --- | --- | --- |
| fresh install into an empty `CLAUDE_PLUGIN_DATA` | `mem0 plugin: failed to install Python dependencies` | exits 0, silent |
| requirements stamp | deleted | written |
| `.install-failed` | created | absent |
| `import mem0` in the venv | `ModuleNotFoundError` | mem0ai 2.0.16 |
| second run | full re-install, fails again | 0.3s no-op |

`bash -n` clean on both scripts.

## Not addressed here

`.install-lock` has no staleness check, so a session killed mid-install (the
`trap ... EXIT` does not cover SIGKILL or a reboot) leaves a lock directory that
makes every later session wait the full 60s and then give up. Happy to send that
as a separate PR if you want it — it seemed better not to mix it with the path
fix.

### DIFF
--- a/integrations/mem0-plugin/scripts/ensure_deps.sh
+++ b/integrations/mem0-plugin/scripts/ensure_deps.sh
@@ -13,9 +13,25 @@ mkdir -p "${DATA_DIR}"
 
 LOCKDIR="${DATA_DIR}/.install-lock"
 
+# A venv puts its interpreter in bin/ on POSIX and Scripts/ on Windows.
+# Echoes the venv interpreter path, or nothing if the venv is absent/incomplete.
+venv_python() {
+  for _candidate in \
+    "${VENV_DIR}/bin/python3" \
+    "${VENV_DIR}/bin/python" \
+    "${VENV_DIR}/Scripts/python.exe"
+  do
+    if [ -x "${_candidate}" ]; then
+      echo "${_candidate}"
+      return 0
+    fi
+  done
+  return 1
+}
+
 needs_install=false
 
-if [ ! -f "${VENV_DIR}/bin/python3" ]; then
+if ! venv_python >/dev/null; then
   needs_install=true
 elif ! diff -q "${REQ_SRC}" "${REQ_STAMP}" >/dev/null 2>&1; then
   needs_install=true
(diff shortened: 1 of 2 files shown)


krishparmani:
> Verified this on Windows 10 (author tested Windows 11) — same results across the board:
>
> - Fresh install into an empty `CLAUDE_PLUGIN_DATA`: exits 0, silent
> - Requirements stamp gets written
> - No `.install-failed` marker
> - `import mem0` works from the venv, resolves to `Scripts/python.exe` as expected
> - Second run is a fast no-op (~0.37s)
>
> Good to see it holds up on a different Windows version too, not just 11.
>
> One thing I noticed while diffing against `main`: this branch also removes the "Project memory policy" block in `on_session_start.sh` (the `_INSTR` echo that surfaces `mem0.md`'s instructions). Doesn't look related to the venv fix — probably just that this branch was cut before that feature landed on `main`. Might be worth rebasing so it doesn't get dropped by accident when this merges.

PR #6809, opened 2026-08-05
feat(sdk, docs): expose agent_custom_instructions for agent-scoped extraction

whysosaket:
## Linked Issue

N/A. This is the SDK and docs surface for a platform-side feature. The backing API field is not live yet, so this stays a **draft** until it ships.

## Description

Adds `agent_custom_instructions`: an optional second set of memory-extraction instructions that applies only to agent-scoped memories.

Today `custom_instructions` governs every memory a project extracts, whether it came from a user or an agent. But what is worth remembering about an agent (which tools fail, which retry strategies work, how an environment behaves) is rarely what is worth remembering about a user, and there has been no way to say so. Projects storing agent memories have had to reuse user-shaped instructions.

Which set applies is decided by the entity IDs on the `add()` call:

| The add call passes | Instructions applied |
|---------------------|----------------------|
| `user_id` only | `custom_instructions` |
| `agent_id` only | `agent_custom_instructions` |
| `user_id` **and** `agent_id` | `agent_custom_instructions` govern the memories attributed to the assistant; `custom_instructions` govern the rest |

**The field is a no-op until it is set.** While unset, `custom_instructions` continues to apply to every memory, and callers that never pass the new parameter send a byte-identical payload, pinned by tests on both sides.

### Python SDK

- `client.project.update(agent_custom_instructions=...)`, sync and async
- `client.update_project(agent_custom_instructions=...)`, the deprecated method, sync and async, kept at parity
- `ProjectUpdateOptions.agent_custom_instructions` and `AddMemoryOptions.agent_custom_instructions`

### TypeScript SDK

- `client.updateProject({ agentCustomInstructions })` sets `PromptUpdatePayload`
- `client.add(messages, { agentCustomInstructions })` sets `AddMemoryOptions`
- `ProjectResponse.agentCustomInstructions`

### Release

- Python `2.0.16` → `2.0.17`, TypeScript `3.1.4` → `3.1.5`, plus the changelog entries for both

### Docs

- A new **Agent Custom Instructions** section on `platform/features/custom-instructions`, covering what it is, the resolution table above, the per-request override, and how to read and clear the value
- The field added to the project-update examples in the organizations & projects reference
- A one-line note on the feature page stating the minimum SDK versions (Python `v2.0.17`, TypeScript `v3.1.5`)
- `openapi.json`: on the project `PATCH` body, the `/v3/memories/add/` request schema, and the `MemoryInput` component

## Type of Change

- [x] Documentation update

## Breaking Changes

None. Every new parameter is optional and defaults to `None` / `undefined`. Positional callers are unaffected, since the parameter is appended last in every Python signature.

## Test Coverage

- [ ] I tested manually
- [ ] No tests needed

**Python** (`tests/test_project.py`, `tests/test_client.py`):
- the kwarg reaches the `PATCH` body; both the typed option and the raw kwarg reach the add payload
- `""` round-trips, so the field can actually be cleared, since it has to survive the `is not None` filter in `_prepare_params` rather than be dropped as falsy
- both instruction sets in a single `PATCH`
- absent from the payload when not passed
- `update()` with no arguments still raises, and the message names the new field

**TypeScript** (`memoryClient.project.test.ts`, `memoryClient.crud.test.ts`):
- `agentCustomInstructions` serializes to `agent_custom_instructions` on both `PATCH` and add
- both sets in one `PATCH`; empty string round-trips; absent when not passed

Local runs: `pytest tests/` (no new failures against `main`), `jest src/client/tests` clean on the touched suites, `ruff check` clean, `prettier --check src/client/` clean, `check-llms-txt-coverage.py` in sync.

## Checklist

## Notes for reviewers

- **Version bumps are in this PR:** Python `2.0.16` → `2.0.17`, TypeScript `3.1.4` → `3.1.5`, with matching `docs/changelog/sdk.mdx` entries. Since the API side isn't live yet, don't cut the release tags until it is. The SDK parameters will serialize fine but the field won't do anything server-side.
- **The new JavaScript examples call `client.updateProject()` / `client.getProject()`, not `client.project.update()`.** The existing examples further up that page use `client.project.update({ ... })`, but the TypeScript SDK has no `client.project` namespace; only Python does. Rather than propagate that into new copy I used the real API here. Worth fixing the older examples separately.

### DIFF
--- a/docs/platform/features/custom-instructions.mdx
+++ b/docs/platform/features/custom-instructions.mdx
@@ -95,6 +95,96 @@ Exclude:
 - [Irrelevant information]
 ```
 
+## Agent Custom Instructions
+
+`custom_instructions` applies to every memory your project extracts, no matter whose it is. But what is worth remembering about an agent is rarely what is worth remembering about a user: an agent's useful memories are things like which tools fail, which retry strategies work, and how a given environment behaves — not personal preferences.
+
+`agent_custom_instructions` is an optional second set of extraction rules that applies only to agent-scoped memories. Set it on the project alongside `custom_instructions`:
+
+<CodeGroup>
+```python Python
+client.project.update(
+    custom_instructions="Extract the user's preferences, goals, and constraints.",
+    agent_custom_instructions=(
+        "Extract operational lessons for the agent:\n"
+        "- Tools that failed and the error returned\n"
+        "- Retry or fallback strategies that worked\n"
+        "- Environment quirks worth recalling on the next run\n\n"
+        "Exclude: user preferences, personal details."
+    ),
+)
+```
+
+```javascript JavaScript
+await client.updateProject({
+  customInstructions: "Extract the user's preferences, goals, and constraints.",
+  agentCustomInstructions: `Extract operational lessons for the agent:
+- Tools that failed and the error returned
+- Retry or fallback strategies that worked
+- Environment quirks worth recalling on the next run
+
+Exclude: user preferences, personal details.`,
+});
+```
+</CodeGroup>
+
+### Which instructions apply
+
+Which set governs an `add()` call depends on the entity IDs you pass with it:
+
+| The add call passes | Instructions applied |
+|---------------------|----------------------|
+| `user_id` only | `custom_instructions` |
+| `agent_id` only | `agent_custom_instructions` |
+| `user_id` **and** `agent_id` | `agent_custom_instructions` govern the memories attributed to the assistant; `custom_instructions` govern the rest |
+
+`agent_custom_instructions` is unset by default. While it is unset, `custom_instructions` applies to every memory — so projects that don't set it behave exactly as they did before.
+
+### Overriding for a single call
+
+Both fields are also accepted per request, overriding the project setting for that `add()` only:
+
+<CodeGroup>
+```python Python
+client.add(
+    messages,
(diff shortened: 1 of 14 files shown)


docs/platform/features/custom-instructions.mdx
```diff
@@ -95,6 +95,96 @@ Exclude:
 - [Irrelevant information]
 ```
 
+## Agent Custom Instructions
+
+`custom_instructions` applies to every memory your project extracts, no matter whose it is. But what is worth remembering about an agent is rarely what is worth remembering about a user: an agent's useful memories are things like which tools fail, which retry strategies work, and how a given environment behaves — not personal preferences.
+
+`agent_custom_instructions` is an optional second set of extraction rules that applies only to agent-scoped memories. Set it on the project alongside `custom_instructions`:
```
deshraj: Can we add a <Note> that this feature is only available after this specific version so that developers can act accordingly?

deshraj (dismissed):

docs/platform/features/custom-instructions.mdx
whysosaket: Yes, Agree.

Added:
```
Available from Python SDK `v2.0.17` and TypeScript SDK `v3.1.5`. Upgrade first if you are on an earlier release.
```

kartik-mem0 (approved):

PR #6813, opened 2026-08-05
fix(sdk): never fail import or Memory() when the mem0 config dir can't be created

### LINKED ISSUE #6812: `import mem0` crashes when the config directory cannot be created (module-level `os.makedirs` in memory/setup.py)
### Component

Python SDK

### Summary

`mem0/memory/setup.py` creates `~/.mem0` as a side effect of *importing the module*:

```python
# mem0/memory/setup.py, lines 10-12
home_dir = os.path.expanduser("~")
mem0_dir = os.environ.get("MEM0_DIR") or os.path.join(home_dir, ".mem0")
os.makedirs(mem0_dir, exist_ok=True)   # <-- unguarded, at import time
```

`mem0/memory/telemetry.py` imports `setup.py` at module scope, `mem0/client/project.py` imports telemetry, and `mem0/__init__.py` imports the client — so this runs on plain `import mem0`, before any user code executes.

Every other filesystem operation in this same module is deliberately best-effort and documented as such (`_load_config` "returning {} on missing/malformed file", `_write_config` "Best-effort write ... Never raises"). The import-time `makedirs` is the one call that is *not* guarded, and it is the one that runs unconditionally.

Consequences:

* The SDK cannot be imported at all on a read-only or non-existent `$HOME` — AWS Lambda, distroless/scratch containers running as a UID with no home entry, Nix/Bazel sandboxes, hardened CI runners.
* `MEM0_TELEMETRY=false` does **not** avoid it — the directory is created regardless of the telemetry opt-out (verified below), even though `~/.mem0` exists only to hold the telemetry/notice state file.
* There is no way to catch it other than wrapping the `import` statement itself.

### Steps to Reproduce

```python
# Point MEM0_DIR at a location that cannot be created, which is what a
# read-only $HOME looks like from this code's point of view:
#
#   MEM0_TELEMETRY=false MEM0_DIR=/proc/nope python -c "import mem0"
#
# or, equivalently, on any host with a read-only home:
#
#   HOME=/readonly python -c "from mem0 import Memory"

from mem0 import Memory   # never reached — the import itself raises

m = Memory()
```

### Expected Behavior

`import mem0` succeeds. The config directory is created lazily on first write and, consistent with `_write_config`'s "Never raises" contract, a failure to create it degrades to "no local config/telemetry state" rather than taking down the import.

### Actual Behavior

The import raises before any `mem0` API is reachable:

```
  File "C:\Users\asus\Desktop\os\mem0\mem0\client\project.py", line 9, in <module>
    from mem0.memory.telemetry import capture_client_event
  File "C:\Users\asus\Desktop\os\mem0\mem0\memory\telemetry.py", line 12, in <module>
    from mem0.memory.setup import get_or_create_user_id
  File "C:\Users\asus\Desktop\os\mem0\mem0\memory\setup.py", line 12, in <module>
    os.makedirs(mem0_dir, exist_ok=True)
  File "<frozen os>", line 225, in makedirs
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'Z:/'
```

On POSIX with a read-only home the same line raises `PermissionError: [Errno 30] Read-only file system: '/readonly/.mem0'`.

Setting `MEM0_TELEMETRY=false` produces the identical traceback — the opt-out is read later, in `telemetry.py`, long after `setup.py` has already executed the `makedirs`.

### Environment

* OS: Windows 11 (reproduced there; the same unguarded call fails identically on read-only POSIX filesystems)

rajarshidattapy:
## Linked Issue

Closes #6812

## Description

`mem0/memory/setup.py` created `~/.mem0` as a side effect of importing the module:

```python
mem0_dir = os.environ.get("MEM0_DIR") or os.path.join(home_dir, ".mem0")
os.makedirs(mem0_dir, exist_ok=True)   # unguarded, at import time
```

`telemetry.py` imports `setup.py` at module scope, `client/project.py` imports `telemetry`,
and `mem0/__init__.py` imports the client — so this ran on plain `import mem0`, before any
user code. On a read-only or missing `$HOME` (Lambda, distroless images, Nix/Bazel
sandboxes, hardened CI) the import raised `PermissionError`/`FileNotFoundError` and there
was no way to catch it short of wrapping the import statement. `MEM0_TELEMETRY=false`
did not help — the opt-out is read later, in `telemetry.py`, long after `setup.py` has
already run the `makedirs`.

This is the only unguarded filesystem call in that module. `_load_config` returns `{}` on a
missing or malformed file and `_write_config` is documented "Best-effort write ... Never
raises"; the directory those two operate on is now held to the same contract via a
`_ensure_dir()` helper that returns `False` instead of raising.

Two sibling calls hit the same directory unguarded in `Memory.__init__` and
`AsyncMemory.__init__`, when building the telemetry migrations store for the `faiss`/`qdrant`
providers:

```python
telemetry_config.path = os.path.join(mem0_dir, provider_path)
os.makedirs(telemetry_config.path, exist_ok=True)
```

Guarding only the import would have moved the crash from `import mem0` to `Memory()` on the
same host, so both now route through `_ensure_dir` and skip `_telemetry_vector_store` when
the directory is unavailable. Nothing reads that attribute, and it is already absent
whenever `MEM0_TELEMETRY=false`, so callers already tolerate it being unset.

`_ensure_dir(mem0_dir)` is deliberately still called at import rather than deleted:
`MemoryConfig.history_db_path` defaults to `~/.mem0/history.db` and `SQLiteManager` opens it
with `sqlite3.connect`, which will not create the parent directory. Removing the call
outright would break first-run `Memory()` on every ordinary machine. The behaviour on a
writable host is therefore unchanged — the directory is created at import exactly as before,
just no longer at the cost of the import itself.

Scope note: on a genuinely read-only `$HOME`, a default `Memory()` still fails inside
`SQLiteManager`, because it is being asked to open a real database in an unwritable
location — callers need to point `history_db_path` somewhere writable. What this PR fixes is
that `import mem0` and `MemoryClient` no longer require a writable home at all, and that
losing the local telemetry directory degrades to "no local telemetry state" instead of
taking down construction.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

`MEM0_DIR` pointed at an uncreatable path stands in for a read-only `$HOME`, since both reach
the same `makedirs`:

| Check | Before | After |
|---|---|---|
| `MEM0_DIR=Z:/nope python -c "import mem0"` | `FileNotFoundError` | imports |
| same, with `MEM0_TELEMETRY=false` | `FileNotFoundError` | imports |
| `Memory()` with uncreatable `mem0_dir`, writable history/vector paths | raised at `main.py:531` | constructs; `_telemetry_vector_store` absent |
| fresh writable `MEM0_DIR` | dir created at import | dir created at import (unchanged) |

Existing suite, optional-dependency directories excluded: **745 passed, 22 skipped, 3 failed** — the same 3 failures as on the base commit, all from optional dependencies absent in my venv (`transformers`, `sentence-transformers`, `langchain-core`). `tests/test_main.py`, `tests/test_telemetry.py` and `tests/test_memory.py` were run specifically for the telemetry paths touched here: **145 passed**.

No unit test added: simulating an unwritable directory portably means patching `os.makedirs`,
which tests the mock rather than the behaviour, and the `MEM0_DIR` invocations above cover it
directly. Happy to add one if you'd like it pinned in CI.

## Checklist

### DIFF
--- a/mem0/__init__.py
+++ b/mem0/__init__.py
@@ -1,6 +1,11 @@
 import importlib.metadata
 
-__version__ = importlib.metadata.version("mem0ai")
+try:
+    __version__ = importlib.metadata.version("mem0ai")
+except importlib.metadata.PackageNotFoundError:
+    # Running from a source checkout (pytest uses pythonpath = ["."]), a vendored
+    # copy, or a container that copied mem0/ in without installing the dist.
+    __version__ = "0.0.0+unknown"
 
 from mem0.client.main import AsyncMemoryClient, MemoryClient  # noqa
 from mem0.memory.main import AsyncMemory, Memory  # noqa
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -48,7 +48,7 @@ from mem0.memory.notices import (
     get_temporal_feature_error_message,
     get_temporal_feature_error_message_async,
 )
-from mem0.memory.setup import mem0_dir, setup_config
+from mem0.memory.setup import _ensure_dir, mem0_dir, setup_config
 from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
 from mem0.memory.utils import (
@@ -523,18 +523,22 @@ class Memory(MemoryBase):
             # Override collection name for telemetry
             telemetry_config_dict['collection_name'] = "mem0migrations"
 
-            # Set path for file-based vector stores
+            # Set path for file-based vector stores. An uncreatable mem0_dir
+            # costs local telemetry state only, so skip it rather than failing
+            # Memory() construction.
             telemetry_config = _safe_deepcopy_config(self.config.vector_store.config)
+            telemetry_dir_ok = True
             if self.config.vector_store.provider in ["faiss", "qdrant"]:
                 provider_path = f"migrations_{self.config.vector_store.provider}"
                 telemetry_config_dict['path'] = os.path.join(mem0_dir, provider_path)
-                os.makedirs(telemetry_config_dict['path'], exist_ok=True)
+                telemetry_dir_ok = _ensure_dir(telemetry_config_dict['path'])
 
-            # Create the config object using the same class as the original
-            telemetry_config = self.config.vector_store.config.__class__(**telemetry_config_dict)
-            self._telemetry_vector_store = VectorStoreFactory.create(
-                self.config.vector_store.provider, telemetry_config
-            )
+            if telemetry_dir_ok:
+                # Create the config object using the same class as the original
+                telemetry_config = self.config.vector_store.config.__class__(**telemetry_config_dict)
+                self._telemetry_vector_store = VectorStoreFactory.create(
+                    self.config.vector_store.provider, telemetry_config
+                )
         if getattr(type(self.vector_store), "keyword_search", None) is VectorStoreBase.keyword_search:
             logger.warning(
                 "The '%s' vector store does not support keyword search. "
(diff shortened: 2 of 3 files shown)


rajarshidattapy:

PR #6815, opened 2026-08-05
fix(upstash): key the server-side embedding bypass off the vector store config

### LINKED ISSUE #6814: Upstash `enable_embeddings: True` never takes effect — EmbedderFactory compares the embedder provider against a vector-store provider name
### Component

Vector Store

### Summary

`docs/components/vectordbs/dbs/upstash-vector.mdx` documents server-side Upstash embeddings as a Python-only feature and states that "Setting `enable_embeddings` to `True` will bypass any external embedding provider you have configured." The documented config sets `enable_embeddings` on the **vector store** and declares no embedder at all.

The code that is supposed to implement that bypass is in `mem0/utils/factory.py`:

```python
class EmbedderFactory:
    @classmethod
    def create(cls, provider_name, config, vector_config: Optional[dict]):
        if provider_name == "upstash_vector" and vector_config and vector_config.enable_embeddings:
            return MockEmbeddings()
```

`provider_name` here is `self.config.embedder.provider` (`mem0/memory/main.py:486`) — the **embedder** provider, which defaults to `"openai"`. It is being compared against `"upstash_vector"`, which is a **vector store** provider name. The two namespaces are different, so the condition is never true and `MockEmbeddings()` is dead code.

The escape hatch is closed too: setting `embedder.provider = "upstash_vector"` to satisfy the check is rejected by `EmbedderConfig.validate_config` (`mem0/embeddings/configs.py:16-31`), whose allowlist does not contain `upstash_vector`.

Net effect: the documented Upstash configuration cannot be constructed. `Memory.from_config()` builds an OpenAI embedder and dies on a missing OpenAI key, on a setup where the whole point was not to need one.

### Steps to Reproduce

```python
# pip install "mem0ai[vector-stores]"
# export UPSTASH_VECTOR_REST_URL=... UPSTASH_VECTOR_REST_TOKEN=...
# Note: OPENAI_API_KEY is deliberately unset — the docs say it is not needed.

from mem0 import Memory

# Copied verbatim from docs/components/vectordbs/dbs/upstash-vector.mdx
config = {
    "vector_store": {
        "provider": "upstash_vector",
        "config": {
            "enable_embeddings": True,
        }
    }
}

print(type(m.embedding_model).__name__)   # expected: MockEmbeddings


# The obvious workaround is also rejected:
config2 = {
    "vector_store": {"provider": "upstash_vector", "config": {"enable_embeddings": True}},
    "embedder": {"provider": "upstash_vector", "config": {}},
}
Memory.from_config(config2)
```

### Expected Behavior

With `enable_embeddings: True` on the Upstash vector store, `EmbedderFactory.create()` returns `MockEmbeddings()` and `Memory.from_config()` succeeds with no external embedding provider and no `OPENAI_API_KEY`, exactly as the docs describe.

### Actual Behavior

The first config never reaches the `MockEmbeddings()` branch and fails constructing the default OpenAI embedder:

```
openai.OpenAIError: Missing credentials. Please pass an `api_key`, `workload_identity`,
`admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
```

(With an `OPENAI_API_KEY` present it "works", but silently uses OpenAI for embeddings and bills the user for them — the opposite of the documented behaviour.)

The second config is rejected by the embedder config validator:

```
Configuration validation error: 1 validation error for MemoryConfig
embedder.config
  Value error, Unsupported embedding provider: upstash_vector
    [type=value_error, input_value={}, input_type=dict]
```

### Environment

rajarshidattapy:
## Linked Issue

Closes #6814

## Description

`docs/components/vectordbs/dbs/upstash-vector.mdx` documents server-side Upstash embeddings
and states that "Setting `enable_embeddings` to `True` will bypass any external embedding
provider you have configured." The documented config sets `enable_embeddings` on the
**vector store** and declares no embedder at all.

The code meant to implement that bypass never ran:

```python
if provider_name == "upstash_vector" and vector_config and vector_config.enable_embeddings:
    return MockEmbeddings()
```

`provider_name` is `self.config.embedder.provider` (`mem0/memory/main.py:486`) — the
embedder provider, which defaults to `"openai"`. It was being compared against
`"upstash_vector"`, a vector store provider name. Those are two different namespaces, so
the condition could never be true and `MockEmbeddings()` was dead code. Following the docs
built an OpenAI embedder instead: it raised `OpenAIError: Missing credentials` with no key
set, and silently embedded through OpenAI — billing the user for embeddings they had
explicitly opted out of — when a key happened to be present.

The workaround was closed off too: setting `embedder.provider = "upstash_vector"` to satisfy
the check is rejected by `EmbedderConfig.validate_config`, whose allowlist has no such entry.

The bypass now keys off the vector store config alone:

```python
if getattr(vector_config, "enable_embeddings", False):
    return MockEmbeddings()
```

`enable_embeddings` is declared only on `UpstashVectorConfig`, so `getattr` with a default is
the entire guard — no store-name comparison is needed, and `getattr(None, ...)` also covers
the no-vector-config case, so the old `vector_config and` clause is redundant and goes with it.

I deliberately did not add `upstash_vector` to the `EmbedderConfig` allowlist. That escape
hatch only existed to satisfy the broken check; with the real fix the documented config needs
no embedder section, and adding an embedder provider that is really a vector store name would
re-introduce the namespace confusion this PR removes.

`MockEmbeddings` inherits `embed_batch()` from `EmbeddingBase`, so the batch paths in
`_add_to_vector_store` and entity linking work without changes.

Out of scope, flagged for a follow-up: with `OPENAI_API_KEY` unset, the exact snippet on the
Upstash docs page now fails one line later at `main.py:494` while building the LLM. `mem0`
always needs an LLM for fact extraction and that page configures no `llm` block —
`enable_embeddings` removes only the embedding provider requirement. That is a docs gap on
the Upstash page rather than part of this bug, so I have not touched it here.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — the branch was unreachable before, so no existing configuration changes behaviour.
Users who had `enable_embeddings: True` set were silently getting OpenAI embeddings; they now
get the Upstash server-side embeddings they asked for. Anyone relying on that accident can
drop `enable_embeddings` (it defaults to `False`) to keep an external embedder.

## Test Coverage

Added `test_enable_embeddings_bypasses_the_external_embedder` to
`tests/vector_stores/test_upstash_vector.py`, in the style of the existing
`test_env_var_only_config_builds_provider` regression test. It mirrors the real
`Memory.__init__` call — `EmbedderFactory.create("openai", {}, vector_store_config)` — and
asserts `MockEmbeddings` when the flag is on, and that the configured embedder is still built
when it is off. The test fails on the pre-fix code (`OpenAIError`) and passes after.

Manual check of the documented config:

```text
enable_embeddings=True  -> embedder: MockEmbeddings
enable_embeddings=False -> embedder: OpenAIEmbedding
```

- `tests/vector_stores/test_upstash_vector.py` + `tests/utils/test_factory.py`: **34 passed**
- Existing suite, optional-dependency directories excluded: **745 passed, 22 skipped, 3 failed** — the same 3 failures as on the base commit, all from optional dependencies absent in my venv (`transformers`, `sentence-transformers`, `langchain-core`).

## Checklist

### DIFF
--- a/mem0/__init__.py
+++ b/mem0/__init__.py
@@ -1,6 +1,11 @@
 import importlib.metadata
 
-__version__ = importlib.metadata.version("mem0ai")
+try:
+    __version__ = importlib.metadata.version("mem0ai")
+except importlib.metadata.PackageNotFoundError:
+    # Running from a source checkout (pytest uses pythonpath = ["."]), a vendored
+    # copy, or a container that copied mem0/ in without installing the dist.
+    __version__ = "0.0.0+unknown"
 
 from mem0.client.main import AsyncMemoryClient, MemoryClient  # noqa
 from mem0.memory.main import AsyncMemory, Memory  # noqa
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -48,7 +48,7 @@ from mem0.memory.notices import (
     get_temporal_feature_error_message,
     get_temporal_feature_error_message_async,
 )
-from mem0.memory.setup import mem0_dir, setup_config
+from mem0.memory.setup import _ensure_dir, mem0_dir, setup_config
 from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
 from mem0.memory.utils import (
@@ -523,18 +523,22 @@ class Memory(MemoryBase):
             # Override collection name for telemetry
             telemetry_config_dict['collection_name'] = "mem0migrations"
 
-            # Set path for file-based vector stores
+            # Set path for file-based vector stores. An uncreatable mem0_dir
+            # costs local telemetry state only, so skip it rather than failing
+            # Memory() construction.
             telemetry_config = _safe_deepcopy_config(self.config.vector_store.config)
+            telemetry_dir_ok = True
             if self.config.vector_store.provider in ["faiss", "qdrant"]:
                 provider_path = f"migrations_{self.config.vector_store.provider}"
                 telemetry_config_dict['path'] = os.path.join(mem0_dir, provider_path)
-                os.makedirs(telemetry_config_dict['path'], exist_ok=True)
+                telemetry_dir_ok = _ensure_dir(telemetry_config_dict['path'])
 
-            # Create the config object using the same class as the original
-            telemetry_config = self.config.vector_store.config.__class__(**telemetry_config_dict)
-            self._telemetry_vector_store = VectorStoreFactory.create(
-                self.config.vector_store.provider, telemetry_config
-            )
+            if telemetry_dir_ok:
+                # Create the config object using the same class as the original
+                telemetry_config = self.config.vector_store.config.__class__(**telemetry_config_dict)
+                self._telemetry_vector_store = VectorStoreFactory.create(
+                    self.config.vector_store.provider, telemetry_config
+                )
         if getattr(type(self.vector_store), "keyword_search", None) is VectorStoreBase.keyword_search:
             logger.warning(
                 "The '%s' vector store does not support keyword search. "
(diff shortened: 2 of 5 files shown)


rajarshidattapy:

Issue #5507, opened 2026-06-12
TealTiger governance layer for memory operations (PII detection, access control, audit)


nagasatish007:
### Component

Core / Python SDK

## Summary

Integrate [TealTiger](https://github.com/agentguard-ai/tealtiger) as a governance layer for Mem0 memory operations — PII detection before writes, access control on reads, and audit evidence for every operation.

## Problem

Without governance, agents can store PII unintentionally, access memories across scope boundaries, exfiltrate data via memory, and leave no audit trail.

## Proposed Integration

**Pre-write governance (before memory.add):**

TealTiger intercepts content before it reaches Mem0 storage. If PII is detected (SSN, credit cards, medical info), it redacts or blocks. If secrets are found (API keys, tokens), they are stripped. Scope enforcement ensures agents can only write to authorized scopes.

**Pre-read governance (before memory.search):**

TealTiger checks whether the requesting agent/user is authorized to access the target memory scope. Unauthorized reads are denied with a structured reason.

**Audit trail:**

Every memory operation produces a governance receipt — who wrote/read what, what was detected, what action was taken, with timestamps and correlation IDs. Exportable as JSONL.

## Capabilities

- PII detection and redaction before memory.add()
- Secret detection (API keys, tokens)
- Scope enforcement (agent isolation)
- Instruction injection detection (memory poisoning defense)
- Read access control on memory.search()
- Structured audit trail (JSONL export)

## Key facts

- Deterministic, <5ms overhead, no LLM in governance path
- Apache 2.0, Python + TypeScript SDKs
- Already integrated with [AG2](https://github.com/ag2ai/ag2/pull/2962), proposed for [Portkey](https://github.com/Portkey-AI/gateway/issues/1688) and Composio
- Addresses OWASP ASI-04 (Memory Poisoning) and ASI-07 (Identity & Access Misuse)

## Links

- Source: https://github.com/agentguard-ai/tealtiger
- PyPI: https://pypi.org/project/tealtiger/

Happy to build a mem0-tealtiger middleware or contribute to cookbooks.

kartik-mem0:
Hi @nagasatish007, thanks for the thoughtful and detailed proposal! 🙏

Governance, PII detection, scope enforcement, and audit trails around memory operations are genuinely important concerns, and we appreciate you laying out the use cases and OWASP mappings so clearly.

[...]

Issue #5331, opened 2026-05-31
[Feature Request] Built-in Memory Poisoning Protection via OWASP Agent Memory Guard


vgudur-dev:
## Problem

Mem0 is the leading memory layer for AI agents, but persistent memory is vulnerable to **memory poisoning attacks** — adversarial inputs stored in memory can cause agents to leak secrets, ignore instructions, or produce corrupted outputs in future conversations. OWASP identifies this as a top risk for LLM applications.

## Proposed Solution

**OWASP Agent Memory Guard (AMG)** is an open-source Python library that wraps any memory store as a transparent security layer:

- `pip install agent-memory-guard`
- Scans every memory write for prompt injection, PII leakage, and tampering
- 92.5% detection rate on AgentThreatBench benchmark
- Already supports Mem0 as a backend via `GuardedMem0Memory` wrapper

### Example Integration

```python
from agent_memory_guard import MemoryGuard
from mem0 import Memory

m = Memory()
guard = MemoryGuard(m)

# Every memory add is now scanned for threats
guard.add("Remember my API key is sk-...", user_id="user1")  # Blocked!
```

## Links

- GitHub: https://github.com/OWASP/www-project-agent-memory-guard
- PyPI: https://pypi.org/project/agent-memory-guard/
- Benchmark: https://pypi.org/project/agent-threat-bench/

Would the Mem0 team consider adding AMG as a recommended security layer or built-in option? Happy to contribute a PR.

norika1207-lab:
This is a useful issue to frame around memory reliability as an operational contract, not just a feature toggle.

For `built-in memory poisoning protection`, I would try to capture a regression fixture with four visible states: what was accepted into the memory/tool pipeline, what was rejected or transformed, what was persisted, and what later recall returns. That makes it possible to distinguish retrieval quality problems from write-path corruption, provider mismatch, or context-budget behavior.

In production agent systems the failure mode is often silent: a memory write looks successful, a tool output is compressed or cached, then a later agent turn reasons from incomplete state. A small end-to-end test that records the write, index/update, recall, and trace metadata would make this much easier to debug and safer for self-hosted deployments.

vgudur-dev:
@norika1207-lab — great framing. You're right that the failure mode is often silent, which is exactly why AMG treats memory writes as a security boundary rather than a feature flag.

Quick update: **v0.3.0 just shipped** with capabilities that directly address your points:

[...]

PR #6818, opened 2026-08-06
feat(memory): redact secret-shaped strings from recalled memory text

### LINKED ISSUE #6817: Recalled memory is returned unredacted - add a secret-redaction boundary (defense-in-depth)
When mem0 recalls a memory, it hands the stored text back verbatim - and in the LLM proxy it splices that text straight into the next model turn. Nothing on that path strips secret-shaped strings. So if a credential ever lands in a stored memory, it rides back into fresh model context every time that memory is recalled.

The path, for anyone who wants to check it:

- `Memory._search_vector_store` builds each result as `MemoryItem(memory=payload.get("data", ""))` and `search()` returns it untouched (same in `AsyncMemory`).
- `Mem0._format_query_with_memories` joins those into the `- Relevant Memories/Facts: ...` block, and that becomes the outgoing user message.
- The only redaction in `memory/main.py` is telemetry-field scrubbing - it never runs on recalled content.

Honest scope, because I'd rather say it up front than oversell this. With the default `infer=True`, `add()` stores LLM-extracted facts rather than the raw text, so a pasted secret usually doesn't survive into storage verbatim. The real verbatim exposure sits in three specific places:

- `infer=False` (raw messages stored as-is),
- graph / entity memory (stores entity text directly),
- any database written before a fix lands.

So this is defense-in-depth, not a remotely exploitable hole - there's no attacker primitive, and the precondition is "a secret got stored in the first place." I'm filing it as hardening, not a CVE. Happy to move it private instead if you'd rather handle anything security-adjacent that way - your call.

The fix I'd propose is small and deliberately boring: a self-contained shape-redactor run at the recall boundary, before the memory text leaves `_search_vector_store`. High-confidence shapes only - API keys with known prefixes (OpenAI, Anthropic, AWS, GitHub, Stripe, Slack, Google), JWTs, PEM private-key blocks, credentialed database URIs. It leaves 40-hex strings alone on purpose, because they're indistinguishable from git SHAs, and redacting every SHA in recalled text would be worse than the thing it's trying to fix.

It is not a governance or PII-policy layer - I know those have come up and been closed (#5507, #5331). This is one function and two call sites, closer in spirit to clearing the key on provider change (#6475) than to a policy engine.

On the default, I'd turn it on out of the box, with a `MemoryConfig` flag to switch it off. The reasoning is just which failure mode surprises people more. If it's off out of the box, a secret silently re-enters model context and nobody finds out until it's sitting in a log or a provider's training pipeline. If it's on, the shape-matcher occasionally redacts something a power user actually wanted back - a base64 blob they're debugging, a token that IS the thing they're recalling - and they flip a documented flag to fix it. The second failure is visible and self-correcting. The first one isn't. So: protect out of the box, and give the people who store technical content or credentials on purpose a clean off switch.

I've got the redactor written and passing tests, plus the two call-site changes - about 70 lines all in. Glad to open the PR if this direction works for you. One question before I do: on by default with an off switch, as above - or would you rather it start off and let people opt in?

rajarshidattapy:
## Linked Issue

Closes #6817

## Description

`search()` hands stored memory text back verbatim, and `Mem0._format_query_with_memories`
splices it straight into the outgoing user message. Nothing on that path strips
secret-shaped strings, so a credential that ever reached storage rides back into fresh
model context on every recall.

This adds the redaction boundary the issue proposes: one self-contained function plus the
two call sites it names.

**`mem0/memory/redaction.py`** (new) — `redact_secrets()` over a tuple of compiled patterns.
High-confidence shapes only: PEM private-key blocks, JWTs, `sk-` keys (OpenAI, Anthropic,
`sk-proj-`, `sk-svcacct-`), AWS access key IDs, GitHub `gh?_`/`github_pat_`, Stripe `sk_`/`rk_`,
Slack `xox?-`, Google `AIza`, and the password inside a credentialed URI. Bare 40-hex is
deliberately left alone — indistinguishable from a git SHA, and redacting every SHA in
recalled text would be worse than the problem.

Two details worth a reviewer's eye:

- Every prefix rule carries a `(?<![A-Za-z0-9])` guard. Without it `"risk-based"` contains
  `sk-`, and a long enough hyphenated word would be eaten.
- The URI rule replaces only the password capture group, so the result stays readable:
  `postgres://admin:[REDACTED]@db.internal:5432/prod`.

**`mem0/configs/base.py`** — `redact_recalled_secrets: bool = True`. On by default, per the
reasoning in the issue: an unredacted secret silently re-enters model context and nobody
finds out until it is in a log or a training pipeline, whereas an over-eager redaction is
visible and self-correcting via a documented flag.

**`mem0/memory/main.py`** — `Memory._search_vector_store` and
`AsyncMemory._search_vector_store`. The flag is resolved once outside the result loop rather
than per-result:

```python
redact = redact_secrets if self.config.redact_recalled_secrets else _identity
...
memory=redact(payload.get("data", "")),
```

Scope, matching the issue: `search()` only. `get()` and `get_all()` still return verbatim
text — they are management/list APIs, not the path that feeds model context, and redacting
there would break "show me what you stored about me". Storage is untouched, so this is a
read-boundary filter, not a migration: setting the flag to `False` recovers the original
text, including for databases written before this lands.

This is defense-in-depth, not a PII or governance layer (cf. #5507, #5331) — it matches
credentials whose shape is unambiguous and nothing else.

## Type of Change

- [ ] Documentation update

## Breaking Changes

Not breaking in the API sense — no signature or response-shape change. It is a behaviour
change on by default: `search()` results containing a secret-shaped string now come back with
`[REDACTED]` in place of it. Anyone who stores technical content or credentials on purpose
sets `redact_recalled_secrets: False` to restore the previous behaviour. Nothing is rewritten
at rest, so the switch is reversible at any time.

## Test Coverage

`tests/memory/test_redaction.py` (new), 29 tests:

- Every listed shape is redacted, with the surrounding text left intact.
- A keep-list asserts the false-positive cases survive untouched: git SHA-1s,
  `"risk-based approach"`, credential-free URIs, Stripe publishable (`pk_`) keys, and
  AWS-lookalikes that are too short.
- PEM blocks, credentialed URIs, multiple secrets in one memory, and empty/non-string inputs.
- Three wiring tests driving `Memory._search_vector_store` and
  `AsyncMemory._search_vector_store` directly, covering flag on (sync + async) and flag off.

End-to-end through `search()` against a local Qdrant:

```text
sync  default (on) : The deploy key is [REDACTED] for prod
sync  off          : The deploy key is sk-abcdefghijklmnopqrstuvwxyz0123456789ABCDEFGHIJKL for prod
async default (on) : The deploy key is [REDACTED] for prod
```

Existing suite, optional-dependency directories excluded: **774 passed, 22 skipped, 3 failed**
— the same 3 failures as on the base commit, all from optional dependencies absent in my venv
(`transformers`, `sentence-transformers`, `langchain-core`).

`make lint` (`ruff check`) and `isort --profile black --check-only` are clean on every touched
file. `ruff format` was applied to the two new files only — `mem0/memory/main.py` is already
unformatted on main, and running the formatter over it would bury a three-line change in
hundreds of lines of unrelated reflow. The mem0 CI gate runs `make lint`, not
`ruff format --check`.

## Checklist

- [x] An issue exists and is linked with `Closes #<number>`

Docs: new "Redact secrets from recalled memories" section in
`docs/open-source/configuration.mdx` — what it does, the default, the off switch, the exact
shapes covered, why 40-hex is excluded, and the scope note. No new `.mdx` page, so no
`docs/llms.txt` entry is required; confirmed with
`python scripts/check-llms-txt-coverage.py` (`docs/llms.txt` is in sync with `docs/**/*.mdx`).

---

Two follow-ups I left out on purpose rather than widening this PR:

- TypeScript parity. `mem0-ts` has the same recall boundary and the same gap. Happy to
  open a sibling issue and port this once the shape list is agreed here.
- `get()` / `get_all()`. Left verbatim by design, as described above — say the word if
  you would rather they were filtered too.

### DIFF
--- a/docs/open-source/configuration.mdx
+++ b/docs/open-source/configuration.mdx
@@ -125,6 +125,25 @@ Change the `provider` string to switch backends. The most common options:
 
 See the full catalog in <Link href="/components/llms/overview">Components</Link>.
 
+## Redact secrets from recalled memories
+
+`search()` returns stored memory text verbatim, and the LLM proxy splices that text into the next model turn. If a credential ever reaches storage, it rides back into fresh model context on every recall. Python's `redact_recalled_secrets` strips secret-shaped strings at that boundary and is **on by default**:
+
+```python
+from mem0 import Memory
+
+memory = Memory.from_config({
+    "vector_store": {"provider": "qdrant", "config": {"host": "localhost", "port": 6333}},
+    "redact_recalled_secrets": False,   # default: True
+})
+```
+
+Redacted shapes are high-confidence only: PEM private-key blocks, JWTs, `sk-` keys (OpenAI, Anthropic), AWS access key IDs, GitHub tokens, Stripe secret and restricted keys, Slack tokens, Google API keys, and the password inside a credentialed URI (`postgres://admin:[REDACTED]@host`). Bare 40-character hex strings are left alone on purpose, because they are indistinguishable from git SHAs.
+
+<Note>
+  This is defense-in-depth, not a PII or governance layer. It filters on read, so storage is untouched and setting `redact_recalled_secrets: False` recovers the original text. Turn it off if you deliberately store technical content or credentials you need back verbatim. `get()` and `get_all()` are management APIs and always return unredacted text.
+</Note>
+
 ## Tune component settings
 
 <AccordionGroup>
--- a/mem0/__init__.py
+++ b/mem0/__init__.py
@@ -1,6 +1,11 @@
 import importlib.metadata
 
-__version__ = importlib.metadata.version("mem0ai")
+try:
+    __version__ = importlib.metadata.version("mem0ai")
+except importlib.metadata.PackageNotFoundError:
+    # Running from a source checkout (pytest uses pythonpath = ["."]), a vendored
+    # copy, or a container that copied mem0/ in without installing the dist.
+    __version__ = "0.0.0+unknown"
 
 from mem0.client.main import AsyncMemoryClient, MemoryClient  # noqa
 from mem0.memory.main import AsyncMemory, Memory  # noqa
--- a/mem0/configs/base.py
+++ b/mem0/configs/base.py
@@ -55,6 +55,15 @@ class MemoryConfig(BaseModel):
         description="Custom instructions for fact extraction",
         default=None,
     )
+    redact_recalled_secrets: bool = Field(
+        description=(
+            "Redact secret-shaped strings (API keys, JWTs, PEM private keys, credentialed "
+            "URIs) from memory text returned by search(). On by default: an unredacted "
+            "secret silently re-enters model context, while an over-eager redaction is "
+            "visible and fixed by turning this off. Set False to recall text verbatim."
+        ),
+        default=True,
+    )
 
 
 class AzureConfig(BaseModel):
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -48,6 +48,7 @@ from mem0.memory.notices import (
     get_temporal_feature_error_message,
     get_temporal_feature_error_message_async,
 )
+from mem0.memory.redaction import redact_secrets
 from mem0.memory.setup import _ensure_dir, mem0_dir, setup_config
 from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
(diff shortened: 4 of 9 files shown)


rajarshidattapy:

PR #6827, opened 2026-08-06
fix(server): stop leaking internal payload keys in admin GET /memories

### LINKED ISSUE #6825: REST `GET /memories` (admin list-all) leaks the internal `text_lemmatized` search-index field into every memory's `metadata`
### Component

REST API

### Summary

`text_lemmatized` is an internal BM25 index field. `_create_memory`/`_update_memory` write it into every payload:

```python
# mem0/memory/main.py:1964
new_metadata["text_lemmatized"] = lemmatize_for_bm25(data)
```

Every SDK read path strips it before returning, via a shared exclusion set repeated in `get()`, `get_all()`, and `_search_vector_store()`:

```python
core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id",
                          "text_lemmatized", "attributed_to", *promoted_payload_keys}
...
additional_metadata = {k: v for k, v in payload.items() if k not in core_and_promoted_keys}
```

The REST server's own serializer does not use that set and does not exclude it:

```python
# server/main.py:386
_RESERVED_PAYLOAD_KEYS = {"data", "user_id", "agent_id", "run_id", "hash",
                          "created_at", "updated_at", "expiration_date"}

def _serialize_memory(row):
    payload = getattr(row, "payload", None) or {}
    return {
        ...
        "metadata": {k: v for k, v in payload.items() if k not in _RESERVED_PAYLOAD_KEYS},
        ...
    }
```

`_serialize_memory` is used by `_list_all_memories()`, which backs `GET /memories` when no `user_id`/`agent_id`/`run_id` is supplied (the admin list-all branch). That branch therefore returns a lemmatized copy of every memory's text inside `metadata`, plus other internal payload keys such as `actor_id`, `role`, and `attributed_to` that the SDK promotes to top level rather than dumping into `metadata`.

The result is that the same server returns two different shapes for the same memory depending on whether the request is scoped:

* `GET /memories?user_id=alice` → routed through `Memory.get_all()` → `metadata` is clean.
* `GET /memories` (admin) → routed through `_serialize_memory()` → `metadata` contains `text_lemmatized`.

Anything that round-trips a listed memory back through `PUT /memories/{id}` with the returned `metadata`, or that renders `metadata` in the dashboard, sees the internal field.

### Steps to Reproduce

```python
import requests

BASE = "http://localhost:8888"
HEADERS = {"X-API-Key": "<admin-api-key>"}

requests.post(f"{BASE}/memories", headers=HEADERS, json={
    "messages": [{"role": "user", "content": "I love hiking on weekends"}],
    "user_id": "alice",
})

scoped = requests.get(f"{BASE}/memories", headers=HEADERS, params={"user_id": "alice"}).json()
listed = requests.get(f"{BASE}/memories", headers=HEADERS).json()   # admin list-all

print("scoped metadata:", scoped["results"][0].get("metadata"))
print("list-all metadata:", listed["results"][0].get("metadata"))
```

### Expected Behavior

Both responses expose the same public shape, and `metadata` contains only user-supplied metadata:

```
scoped   metadata: None
list-all metadata: None
```

### Actual Behavior

The admin list-all response exposes the internal search-index field (and the internal actor/role keys) as if they were user metadata:

```
scoped   metadata: None
list-all metadata: {'text_lemmatized': 'love hike weekend', 'role': 'user'}
```

### Environment

rajarshidattapy:
Fixes #6825

## Problem

`GET /memories` with no `user_id`/`agent_id`/`run_id` (the admin list-all branch) goes
through `server/main.py::_serialize_memory()` instead of `Memory.get_all()`. That
serializer had its own hand-written reserved-key set which never included
`text_lemmatized`, so every listed memory returned the internal BM25 index field — plus
`actor_id`, `role`, and `attributed_to` — inside `metadata`, as if it were user metadata.

The same server therefore returned two shapes for the same memory:

```text
GET /memories?user_id=alice  -> metadata: None
GET /memories                -> metadata: {'text_lemmatized': 'love hike weekend', 'role': 'user'}
```

## Root cause

The payload → public-shape key set was written out by hand in **7 places**: six inline
copies inside `mem0/memory/main.py` (`get`, `get_all`, `_search_vector_store`, sync and
async) and one in the REST server. The server's copy drifted.

## Changes

- `mem0/memory/utils.py`: single source of truth — `PROMOTED_PAYLOAD_KEYS` and
  `NON_METADATA_PAYLOAD_KEYS` (which includes `text_lemmatized`).
- `mem0/memory/main.py`: dropped the 6 duplicated local copies in favour of the shared
  constants (−80/+14 lines, no behaviour change).
- `server/main.py`: `_serialize_memory()` now filters `metadata` with
  `NON_METADATA_PAYLOAD_KEYS` and promotes `PROMOTED_PAYLOAD_KEYS` to the top level, so
  the admin branch returns the same shape as `Memory.get_all()`.

## Note on the response shape

To match `get_all()`, `user_id` / `agent_id` / `run_id` / `expiration_date` are now
omitted when absent from the payload rather than emitted as `null`, and `metadata` is
`None` rather than `{}` when there is no user metadata.

## Test plan

Added to `tests/test_server_params.py::TestGetMemories`:

- `test_get_memories_admin_list_all_hides_internal_payload_keys` — `text_lemmatized` is
  gone, `actor_id`/`role`/`user_id` are promoted, only user metadata remains.
- `test_get_memories_admin_list_all_metadata_none_when_empty`

```text
pytest tests/test_server_params.py    ->  70 passed
pytest tests/memory                   -> 309 passed
pytest tests/test_memory.py           ->  58 passed
ruff check                            -> clean
```

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -52,6 +52,8 @@ from mem0.memory.setup import mem0_dir, setup_config
 from mem0.memory.storage import SQLiteManager
 from mem0.memory.telemetry import MEM0_TELEMETRY, capture_event
 from mem0.memory.utils import (
+    NON_METADATA_PAYLOAD_KEYS,
+    PROMOTED_PAYLOAD_KEYS,
     extract_json,
     parse_messages,
     parse_vision_messages,
@@ -1211,18 +1213,6 @@ class Memory(MemoryBase):
             display_first_run_notice(self, "sync", "get")
             return None
 
-        promoted_payload_keys = [
-            "user_id",
-            "agent_id",
-            "run_id",
-            "actor_id",
-            "role",
-            "attributed_to",
-            "expiration_date",
-        ]
-
-        core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
-
         result_item = MemoryItem(
             id=memory.id,
             memory=memory.payload.get("data", ""),
@@ -1231,11 +1221,11 @@ class Memory(MemoryBase):
             updated_at=memory.payload.get("updated_at"),
         ).model_dump()
 
-        for key in promoted_payload_keys:
+        for key in PROMOTED_PAYLOAD_KEYS:
             if key in memory.payload:
                 result_item[key] = memory.payload[key]
 
-        additional_metadata = {k: v for k, v in memory.payload.items() if k not in core_and_promoted_keys}
+        additional_metadata = {k: v for k, v in memory.payload.items() if k not in NON_METADATA_PAYLOAD_KEYS}
         if additional_metadata:
             result_item["metadata"] = additional_metadata
 
@@ -1329,17 +1319,6 @@ class Memory(MemoryBase):
         else:
             actual_memories = memories_result
 
-        promoted_payload_keys = [
-            "user_id",
-            "agent_id",
-            "run_id",
-            "actor_id",
-            "role",
-            "attributed_to",
-            "expiration_date",
-        ]
-        core_and_promoted_keys = {"data", "hash", "created_at", "updated_at", "id", "text_lemmatized", "attributed_to", *promoted_payload_keys}
-
         formatted_memories = []
         for mem in actual_memories:
             if not show_expired and _payload_is_expired(mem.payload):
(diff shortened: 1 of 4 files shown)


rajarshidattapy:

PR #6832, opened 2026-08-06
fix(tests): pass POSIX paths to bash so migration tests run on Windows

### LINKED ISSUE #6830: All 18 tests in tests/test_oss_to_platform_migrate.py fail on Windows — Windows paths are passed unescaped to `bash`
### Component

Other

### Summary

`tests/test_oss_to_platform_migrate.py` shells out to the migration script with a native path:

```python
# tests/test_oss_to_platform_migrate.py:14
SCRIPT = Path(__file__).resolve().parents[1] / "scripts" / "oss-to-platform-migrate.sh"

# ...and then, in every test:
result = subprocess.run(
    ["bash", str(SCRIPT), "--auth-only", "--base-url", server.url, *args], ...
)
```

On Windows `str(SCRIPT)` is `C:\Users\...\scripts\oss-to-platform-migrate.sh`. `bash` treats each backslash as an escape character, so the argument arrives as `C:UsersasusDesktoposmem0scriptsoss-to-platform-migrate.sh` and the script is never found. Git Bash is installed and `bash` itself resolves fine — it is purely the argument that is mangled.

Two variants of the same root cause appear in the file:

* `subprocess.run(["/bin/bash", str(SCRIPT), "--help"], ...)` (line 542) hardcodes an absolute POSIX interpreter path.
* `subprocess.run(["bash", "-c", f"curl -fsSL file://{SCRIPT} | bash -s -- --help"], ...)` (line 557) interpolates the same backslash path into a `file://` URL.

The fix is one place — normalise the script path for the shell (`SCRIPT.as_posix()`, or resolve through `shutil.which("bash")` and pass a POSIX-form path) — rather than 18 call sites.

This is a different failure from #6719 (which is an assertion on `os.system`'s POSIX module name in the unpickler test); that one is a wrong expected value, this one is argument marshalling to a subprocess.

### Steps to Reproduce

```python
# On Windows, with Git Bash on PATH, from the repo root:
#
#   python -m pytest tests/test_oss_to_platform_migrate.py -q
#
# Minimal isolation of the root cause:

import subprocess
from pathlib import Path

SCRIPT = Path("scripts/oss-to-platform-migrate.sh").resolve()

native = subprocess.run(["bash", str(SCRIPT), "--help"], capture_output=True, text=True)
print("native path  ->", native.returncode, native.stderr.strip())

posix = subprocess.run(["bash", SCRIPT.as_posix(), "--help"], capture_output=True, text=True)
print("posix path   ->", posix.returncode)
```

### Expected Behavior

The module's tests pass on Windows the same way they do on Linux/macOS, or they are skipped with a clear marker if a POSIX shell is a hard requirement. `bash` receives a path it can actually open.

### Actual Behavior

Every test in the module fails with the same mangled-path error:

```
>       assert result.returncode == 0, result.stderr
E       AssertionError: /bin/bash: C:UsersasusDesktoposmem0scriptsoss-to-platform-migrate.sh: No such file or directory
E
E       assert 127 == 0
E        +  where 127 = CompletedProcess(args=['bash', 'C:\\Users\\asus\\Desktop\\os\\mem0\\scripts\\oss-to-platform-migrate.sh', ...],
E                                        stdout='', stderr='/bin/bash: C:UsersasusDesktoposmem0scriptsoss-to-platform-migrate.sh: No such file or directory\n')

tests\test_oss_to_platform_migrate.py:828: AssertionError
```

Summary from a full run of the core suite:

```
FAILED tests/test_oss_to_platform_migrate.py::test_existing_api_key_authenticates_and_stitches_ids
FAILED tests/test_oss_to_platform_migrate.py::test_email_code_authenticates_without_persisting_credentials
FAILED tests/test_oss_to_platform_migrate.py::test_invalid_stored_key_falls_back_to_email_code
FAILED tests/test_oss_to_platform_migrate.py::test_email_code_failure_reports_failed_telemetry
FAILED tests/test_oss_to_platform_migrate.py::test_malformed_config_does_not_crash_and_authenticates
FAILED tests/test_oss_to_platform_migrate.py::test_weird_telemetry_shape_does_not_crash
FAILED tests/test_oss_to_platform_migrate.py::test_missing_python3_prints_clear_shell_error
FAILED tests/test_oss_to_platform_migrate.py::test_curl_piped_help_works
FAILED tests/test_oss_to_platform_migrate.py::test_export_qdrant_memories_to_json_without_vectors_or_api_key
FAILED tests/test_oss_to_platform_migrate.py::test_export_requires_scope_or_all
FAILED tests/test_oss_to_platform_migrate.py::test_export_all_uses_no_qdrant_filter
FAILED tests/test_oss_to_platform_migrate.py::test_export_invalid_qdrant_credentials_fail_clearly
FAILED tests/test_oss_to_platform_migrate.py::test_import_platform_memories_from_export_json
FAILED tests/test_oss_to_platform_migrate.py::test_import_skips_existing_identical_memory
FAILED tests/test_oss_to_platform_migrate.py::test_import_reports_changed_existing_without_update_or_add
FAILED tests/test_oss_to_platform_migrate.py::test_full_flow_auth_export_and_imports_memories
```

### Environment

* OS: Windows 11 (Git Bash on PATH)

rajarshidattapy:
Fixes #6830

## Problem

`tests/test_oss_to_platform_migrate.py` shells out with a native path:

```python
subprocess.run(["bash", str(SCRIPT), "--auth-only", ...])
```

On Windows `str(SCRIPT)` is `C:\Users\...\oss-to-platform-migrate.sh`. The backslashes
reach bash as escapes, the argument arrives as
`C:UsersasusDesktoposmem0scriptsoss-to-platform-migrate.sh`, and every test in the module
fails with exit 127. Git Bash itself resolves fine — it is purely argument marshalling.

Two more variants of the same root cause: line 542 hardcoded `/bin/bash` (no such
executable on Windows), and line 557 interpolated the backslash path into a `file://` URL.

## Changes

All in `tests/test_oss_to_platform_migrate.py`:

- `SCRIPT_ARG = SCRIPT.as_posix()` and `BASH = shutil.which("bash")`, used at all six
  shell-out sites. `BASH` is resolved to an absolute path because
  `test_missing_python3_prints_clear_shell_error` wipes `PATH` and would otherwise fail to
  find the interpreter — that's why `/bin/bash` was hardcoded there in the first place.
- `test_curl_piped_help_works` now uses `SCRIPT.as_uri()`. `as_posix()` is not sufficient
  there: `file://` + `C:/Users/...` produces two slashes and curl parses `C:` as the host.
  `as_uri()` emits `file:///C:/Users/...` on Windows and `file:///home/...` on POSIX.
- Module-level `skipif` when `bash` or `python3` is missing, probed via
  `bash -c "command -v python3"` — on Windows what Git Bash resolves is not what
  `shutil.which()` sees.

## Test plan

On Windows 11 / Python 3.11.9 / Git Bash:

```
16 passed in 15.13s     # with python3 on PATH
16 skipped in 0.21s     # without it, with a clear reason
```

Before this change: 16 failed with `No such file or directory`.

## Note on the skip

The path fix alone still leaves the tests failing on a Windows box whose Git Bash has
`python` but no `python3` — the script hard-requires `python3`. That is an environment gap,
not a repo bug, so `scripts/oss-to-platform-migrate.sh` is untouched (its "Install Python 3"
error is deliberate and asserted by a test). The module skips instead, which is the
fallback the issue describes. On Linux/macOS CI nothing skips.

### DIFF
--- a/tests/test_oss_to_platform_migrate.py
+++ b/tests/test_oss_to_platform_migrate.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import os
+import shutil
 import subprocess
 import threading
 from hashlib import sha256
@@ -10,8 +11,22 @@ from pathlib import Path
 from typing import Any
 from urllib.parse import parse_qs, urlparse
 
+import pytest
 
 SCRIPT = Path(__file__).resolve().parents[1] / "scripts" / "oss-to-platform-migrate.sh"
+# bash cannot open a native Windows path -- the backslashes reach it as escapes -- so
+# shell-outs use the POSIX form. BASH is resolved to an absolute path because one test
+# wipes PATH and would otherwise not find the interpreter either.
+SCRIPT_ARG = SCRIPT.as_posix()
+BASH = shutil.which("bash")
+# The script hard-requires python3. Ask the shell the tests actually use: on Windows what
+# Git Bash resolves is not what shutil.which() sees.
+HAS_PYTHON3 = bool(BASH) and subprocess.run([BASH, "-c", "command -v python3"], capture_output=True).returncode == 0
+
+pytestmark = pytest.mark.skipif(
+    not (BASH and HAS_PYTHON3),
+    reason="the migration script shells out to bash and python3; both must be on PATH",
+)
 
 
 class MigrationHTTPServer:
@@ -224,7 +239,7 @@ def run_migration_script(
     env.pop("MEM0_BASE_URL", None)
 
     result = subprocess.run(
-        ["bash", str(SCRIPT), "--auth-only", "--base-url", server.url, *args],
+        [BASH, SCRIPT_ARG, "--auth-only", "--base-url", server.url, *args],
         capture_output=True,
         text=True,
         env=env,
@@ -260,8 +275,8 @@ def run_export_script(
 
     result = subprocess.run(
         [
-            "bash",
-            str(SCRIPT),
+            BASH,
+            SCRIPT_ARG,
             "--export-only",
             "--qdrant-url",
             server.url,
@@ -305,8 +320,8 @@ def run_import_script(
 
     result = subprocess.run(
         [
-            "bash",
-            str(SCRIPT),
+            BASH,
+            SCRIPT_ARG,
             "--import-only",
             "--base-url",
             server.url,
@@ -349,8 +364,8 @@ def run_full_script(
 
     result = subprocess.run(
         [
-            "bash",
-            str(SCRIPT),
+            BASH,
+            SCRIPT_ARG,
             "--base-url",
             server.url,
             "--qdrant-url",
@@ -540,7 +555,7 @@ def test_missing_python3_prints_clear_shell_error(tmp_path: Path) -> None:
     env["PATH"] = str(tmp_path)
 
     result = subprocess.run(
-        ["/bin/bash", str(SCRIPT), "--help"],
+        [BASH, SCRIPT_ARG, "--help"],
         capture_output=True,
         text=True,
         env=env,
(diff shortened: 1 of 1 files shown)


rajarshidattapy:

PR #6851, opened 2026-08-07
fix(vector_stores/azure_ai_search): raise on failed IndexingResult

### LINKED ISSUE #6850: bug(vector_stores/azure_ai_search): failed IndexingResult is ignored because the success check is inverted
### Component

Vector Store

### Summary

In the Python Azure AI Search vector store, failed `insert` / `update` / `delete` operations can return without raising. The success check is inverted for the Azure SDK's real return type (`IndexingResult`), so Mem0 can treat a failed Azure write as success.

The checks look like:

```python
if not hasattr(doc, "status_code") and doc.get("status_code") != 201:
    raise Exception(...)
```

Azure SDK `IndexingResult` **has** `status_code`, so `not hasattr(doc, "status_code")` is always `False`. The `and` short-circuits and the exception is never raised — including when `succeeded=False` and the status is an error.

The TypeScript OSS Azure client already keys off `result.succeeded`. The Python path does not.

This is separate from #6732 (unguarded `__del__`) and #6566 (wildcard filters). Those do not touch this guard.

### Steps to Reproduce

No live Azure account required — the bug is in the success-check logic. Against current `main`:

```python

# Minimal reproduction of the inverted guard (same shape as Azure IndexingResult)
doc = SimpleNamespace(
    succeeded=False,
    status_code=400,
    key="mem-1",
    error_message="The document is invalid.",
)

# Current main condition for insert:
if not hasattr(doc, "status_code") and doc.get("status_code") != 201:
    raise Exception(f"Insert failed: {doc}")

print("raised: no — failure was ignored")  # this is what happens today
```

Or via the store with a mocked client:

```python
from mem0.vector_stores.azure_ai_search import AzureAISearch

with patch("mem0.vector_stores.azure_ai_search.SearchIndexClient"), \
     patch("mem0.vector_stores.azure_ai_search.SearchClient") as MockSearchClient, \
     patch("mem0.vector_stores.azure_ai_search.AzureKeyCredential"):
    store = AzureAISearch(
        service_name="test",
        collection_name="test-index",
        api_key="test-key",
        embedding_model_dims=3,
    )
    store.search_client = MagicMock()
    store.search_client.upload_documents.return_value = [
        SimpleNamespace(
            succeeded=False,
            status_code=400,
            key="mem-1",
            error_message="The document is invalid.",
        )
    ]
    # On main: returns without raising
    store.insert(vectors=[[0.1, 0.2, 0.3]], payloads=[{"data": "x"}], ids=["mem-1"])
    print("insert returned OK despite succeeded=False")
```

### Expected Behavior

When Azure returns `IndexingResult(succeeded=False, ...)`, `insert` / `update` / `delete` should raise so callers (and `Memory.add` / update / delete paths) know the vector write failed.

### Actual Behavior

Failures are ignored whenever the result object has a `status_code` attribute (the normal Azure SDK case). Mem0 can continue as if the memory was written to the index when it was not, which can leave history / entity state out of sync with the vector index.

### Environment

* mem0 version: `main` @ latest (guard present since the Azure AI Search provider landed; still on current `main`)
* Python/Node version: Python 3.11
* OS: Windows 11 (logic bug is platform-independent)

nothariharan:
﻿## Linked Issue

Closes #6850

## Description

Azure AI Search `insert` / `update` / `delete` used an inverted success guard:

```python
if not hasattr(doc, "status_code") and doc.get("status_code") != 201:
    raise Exception(...)
```

The Azure SDK returns `IndexingResult` objects that **do** have `status_code`, so `not hasattr(...)` is always false and the exception never fires — including when `succeeded=False`. Failed writes looked successful to Mem0.

This PR adds `_assert_indexing_succeeded`, which:

1. Prefers the SDK `succeeded` flag (same idea as the TypeScript Azure client)
2. Falls back to `status_code` for dict-shaped results
3. Raises with the document key and Azure error message on failure

Wired into insert (expect 201), delete, and update (expect 200).

Unrelated to #6732 (`__del__` guard) and #6566 (wildcard filters).

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A — previously failed Azure indexing results were silently ignored. Callers that somehow depended on that silence will now see an exception when Azure reports failure, which is the intended contract.

## Test Coverage

### How this was tested

No live Azure AI Search account was used. The failure mode is in the local success-check logic, so coverage is unit tests with a mocked `SearchClient` and `SimpleNamespace` objects shaped like Azure `IndexingResult` (`succeeded`, `status_code`, `key`, `error_message`).

**Regression tests added** in `tests/vector_stores/test_azure_ai_search.py`:

| Test | Asserts |
|------|---------|
| `test_insert_indexing_result_success` | `succeeded=True` → insert returns |
| `test_insert_indexing_result_failure_raises` | `succeeded=False` → raises (this fails on unfixed `main`) |
| `test_delete_indexing_result_success` / `_failure_raises` | same for delete |
| `test_update_indexing_result_success` / `_failure_raises` | same for update |

Also covered: mixed batch results, dict-shaped results (status_code path), and existing Azure AI Search tests.

**Commands run:**

```text
python -m pytest tests/vector_stores/test_azure_ai_search.py -q
# 33 passed
```

Reproduced the pre-fix behaviour with the same mock (`succeeded=False`, `status_code=400`): insert returned without raising. With the fix, it raises `Insert failed for document mem-1: ...`.

## Checklist

### DIFF
--- a/mem0/vector_stores/azure_ai_search.py
+++ b/mem0/vector_stores/azure_ai_search.py
@@ -169,6 +169,35 @@ class AzureAISearch(VectorStoreBase):
                 document[field] = payload[field]
         return document
 
+    @staticmethod
+    def _assert_indexing_succeeded(
+        doc, operation: str, expected_status: Optional[int] = None, fallback_id: Optional[str] = None
+    ):
+        """Raise if an Azure IndexingResult (object or dict) did not succeed.
+
+        Prefers the SDK ``succeeded`` attribute (matching the TypeScript client),
+        then falls back to ``status_code`` when ``succeeded`` is absent.
+        """
+        if isinstance(doc, dict):
+            succeeded = doc.get("succeeded")
+            status_code = doc.get("status_code")
+            key = doc.get("key") or doc.get("id") or fallback_id
+            error_message = doc.get("error_message") or doc.get("errorMessage") or doc
+        else:
+            succeeded = getattr(doc, "succeeded", None)
+            status_code = getattr(doc, "status_code", None)
+            key = getattr(doc, "key", None) or getattr(doc, "id", None) or fallback_id
+            error_message = getattr(doc, "error_message", None) or getattr(doc, "errorMessage", None) or doc
+
+        if succeeded is None and status_code is not None:
+            if expected_status is not None:
+                succeeded = status_code == expected_status
+            else:
+                succeeded = 200 <= int(status_code) < 300
+
+        if not succeeded:
+            raise Exception(f"{operation} failed for document {key}: {error_message}")
+
     # Note: Explicit "insert" calls may later be decoupled from memory management decisions.
     def insert(self, vectors, payloads=None, ids=None):
         """
@@ -185,8 +214,7 @@ class AzureAISearch(VectorStoreBase):
         ]
         response = self.search_client.upload_documents(documents)
         for doc in response:
-            if not hasattr(doc, "status_code") and doc.get("status_code") != 201:
-                raise Exception(f"Insert failed for document {doc.get('id')}: {doc}")
+            self._assert_indexing_succeeded(doc, "Insert", expected_status=201)
         return response
 
     def _sanitize_key(self, key: str) -> str:
@@ -205,8 +233,7 @@ class AzureAISearch(VectorStoreBase):
                 condition = f"{safe_key} eq {value}"
             else:
                 raise ValueError(
-                    f"Filter value for {key!r} must be str, int, float, or bool, "
-                    f"got {type(value).__name__}"
+                    f"Filter value for {key!r} must be str, int, float, or bool, got {type(value).__name__}"
                 )
             filter_conditions.append(condition)
         filter_expression = " and ".join(filter_conditions)
@@ -290,8 +317,7 @@ class AzureAISearch(VectorStoreBase):
         """
         response = self.search_client.delete_documents(documents=[{"id": vector_id}])
         for doc in response:
-            if not hasattr(doc, "status_code") and doc.get("status_code") != 200:
-                raise Exception(f"Delete failed for document {vector_id}: {doc}")
+            self._assert_indexing_succeeded(doc, "Delete", expected_status=200, fallback_id=vector_id)
         logger.info(f"Deleted document with ID '{vector_id}' from index '{self.index_name}'.")
         return response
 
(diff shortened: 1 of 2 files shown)


anujbolewar (commented):
Real robustness improvement — handling both the dict and SDK-object shapes and preferring the  flag is the right call, and it fixes the original check which only inspected one shape.

A couple of suggestions:
1) Raise a concrete type instead of bare  — callers won't be able to catch/retry a transient indexing failure selectively.  (or a store-specific error) is more useful than a catch-all, and since there's no underlying exception object here there's nothing to chain with , so a plain typed raise is fine.
2)  will throw a  for a non-numeric code — worth a guard or a try/except so a weird payload doesn't mask the real failure with a confusing exception.

The dict-vs-object branching duplicates logic; a small helper that reads either shape (like you did for ) keeps the fallback tidy. Good overall.

anujbolewar (commented):
Real robustness improvement — handling both the dict and SDK-object shapes and preferring the succeeded flag is the right call, and it fixes the original check which only inspected one shape.

A couple of suggestions:
1) Raise a concrete type instead of a bare Exception — callers cannot catch or retry a transient indexing failure selectively. RuntimeError (or a store-specific error) is more useful than a catch-all; there is no underlying exception object here to chain with from, so a plain typed raise is fine.
2) int(status_code) will throw a ValueError for a non-numeric code — worth a guard or a try/except so a weird payload does not mask the real failure with a confusing exception.

The dict-vs-object branches duplicate logic; a small helper that reads either shape (like you did for key) keeps the fallback tidy. Good overall.

nothariharan:
Thanks on the review — missed those, makes sense why yall r maintainers lol

Working through all three now:
1. raise `RuntimeError` instead of bare `Exception`
2. guard the `int(status_code)` path so a weird payload doesn’t turn into a confusing `ValueError`
3. small helper for the dict vs object reads so that branching isn’t duplicated

Will push the follow-up on this PR shortly.

nothariharan:
Pushed the follow-up — covers all three:

1. `RuntimeError` instead of bare `Exception`
2. non-numeric `status_code` raises `RuntimeError` (not `ValueError`)
3. `_indexing_result_field` helper for dict vs SDK object reads

`pytest tests/vector_stores/test_azure_ai_search.py` → 34 passed

nothariharan:
Hey @anujbolewar, rebased this onto latest main and verified the suite (pytest tests/vector_stores/test_azure_ai_search.py -> 34 passed, ruff/isort clean).

Whenever you get a chance, could you approve the CI workflow and take another look? Thanks!

PR #6855, opened 2026-08-07
fix(memory): reject boolean values for threshold in search parameter validation (#6854)

### LINKED ISSUE #6854: bug(sdk-python): threshold parameter in _validate_search_params silently accepts booleans (True/False) instead of raising ValueError
### Component

Python SDK

### Summary

In `mem0/memory/main.py`, `_validate_search_params` validates `threshold` and `top_k` for search parameter boundaries.

While `top_k` explicitly guards against boolean inputs (`isinstance(top_k, bool)`), `threshold` checks `if not isinstance(threshold, (int, float)):`. Because Python's `bool` inherits from `int` (`isinstance(True, (int, float))` is `True`), passing `threshold=True` or `threshold=False` silently passes validation without raising a `ValueError`.

**Lines 223–229 (`threshold` validation — missing `bool` check):**
```python
if threshold is not None:
    if not isinstance(threshold, (int, float)):
        raise ValueError("threshold must be a valid number")
        raise ValueError(
            f"Invalid threshold: {threshold}. Must be between 0 and 1 (inclusive)."
        )
```

**Lines 230–236 (`top_k` validation — explicit `bool` check present):**
```python
if top_k is not None:
    if not isinstance(top_k, int) or isinstance(top_k, bool):
        raise ValueError("top_k must be a valid integer")
    if top_k < 0:
        raise ValueError(
            f"Invalid top_k: {top_k}. Must be a non-negative integer."
        )
```

### Steps to Reproduce

```python

# 1. top_k=True (correctly raises ValueError)
try:
    _validate_search_params(top_k=True)
except ValueError as e:
    print("top_k=True raised ValueError:", e)

# 2. threshold=True (silently passes)
try:
    _validate_search_params(threshold=True)
    print("threshold=True silently passed validation!")
except ValueError as e:
    print("threshold=True raised ValueError:", e)

# 3. threshold=False (silently passes)
try:
    _validate_search_params(threshold=False)
    print("threshold=False silently passed validation!")
except ValueError as e:
    print("threshold=False raised ValueError:", e)
```

### Expected Behavior

`_validate_search_params(threshold=True)` and `_validate_search_params(threshold=False)` should raise `ValueError("threshold must be a valid number")`, matching the behavior of `top_k`.

### Actual Behavior

`threshold=True` and `threshold=False` pass validation silently without raising `ValueError`:

```text
top_k=True raised ValueError: top_k must be a valid integer
threshold=True silently passed validation!
threshold=False silently passed validation!
```

### Environment

- mem0 version: 2.0.17
- Python/Node version: Python 3.11 / 3.12 / 3.14
- OS: Cross-platform

agarwalpranav0711:
## Linked Issue

Closes #6854

## Description

Fixes an inconsistency in `_validate_search_params` where boolean inputs (`threshold=True` or `threshold=False`) were silently accepted instead of raising a `ValueError`.

### Root Cause
In Python, `bool` is a subclass of `int` (`isinstance(True, (int, float))` evaluates to `True`). 

While line 232 explicitly guards `top_k` against boolean values (`or isinstance(top_k, bool)`), line 225 checked `if not isinstance(threshold, (int, float)):` without an equivalent boolean check. Consequently, boolean thresholds passed validation silently and got evaluated as numerical `1.0` or `0.0`.

### Changes Made
1. Added `or isinstance(threshold, bool)` to `_validate_search_params` in `mem0/memory/main.py`.
2. Added unit test `test_search_rejects_boolean_threshold` in `tests/test_main.py` verifying that `threshold=True` and `threshold=False` raise `ValueError("threshold must be a valid number")`.

### Affected Call Sites
This validation fix applies across all 4 search parameter validation entry points:
- `Memory.search()`
- `Memory.get_all()`
- `AsyncMemory.search()`
- `AsyncMemory.get_all()`

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Ran full test suite locally:
- `pytest tests/test_main.py` (45 passed, including new `test_search_rejects_boolean_threshold`)
- `pytest tests/test_memory.py` (59 passed)
- `pytest tests/test_client.py tests/test_project.py` (48 passed)
- Verified live that `threshold=True` and `threshold=False` raise `ValueError("threshold must be a valid number")`.

## Checklist

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -221,7 +221,7 @@ def _validate_search_params(threshold: Optional[float] = None, top_k: Optional[i
         ValueError: If threshold or top_k are invalid
     """
     if threshold is not None:
-        if not isinstance(threshold, (int, float)):
+        if not isinstance(threshold, (int, float)) or isinstance(threshold, bool):
             raise ValueError("threshold must be a valid number")
         if threshold < 0 or threshold > 1:
             raise ValueError(
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,3 +1,4 @@
+
 import logging
 import os
 from unittest.mock import Mock, patch
@@ -512,6 +513,14 @@ class TestSearchParamValidation:
         """Search should reject negative threshold."""
         with pytest.raises(ValueError, match="Invalid threshold.*Must be between 0 and 1"):
             memory_instance.search("test query", filters={"user_id": "test"}, threshold=-0.5)
+
+    def test_search_rejects_boolean_threshold(self, memory_instance):
+        """Search should reject boolean threshold (True/False)."""
+        with pytest.raises(ValueError, match="threshold must be a valid number"):
+            memory_instance.search("test query", filters={"user_id": "test"}, threshold=True)
+
+        with pytest.raises(ValueError, match="threshold must be a valid number"):
+            memory_instance.search("test query", filters={"user_id": "test"}, threshold=False)
 
     def test_search_rejects_negative_top_k(self, memory_instance):
         """Search should reject negative top_k."""


anujbolewar (commented):
Correct —  is a subclass of  so it slipped straight through the isinstance check. Rejecting it explicitly avoids threshold=True silently being treated as 1.0.

One follow-up worth checking:  (and any other numeric param validated with the same isinstance pattern) likely has the identical  gap. If the validation is centralized or mirrored, adding the same  guard there closes the class of bug rather than just this instance. Test coverage for both True/False is good.

anujbolewar (commented):
Correct — bool is a subclass of int so it slipped straight through the isinstance check. Rejecting it explicitly avoids threshold=True silently being treated as 1.0.

One follow-up worth checking: top_k (and any other numeric param validated with the same isinstance pattern) likely has the identical bool gap. If the validation is mirrored, adding the same guard there closes the class of bug rather than just this instance. Test coverage for both True/False is good.

agarwalpranav0711:
Thanks @anujbolewar! That was a great observation. I audited the entire codebase (mem0/, server/, cli/) for all instances of isinstance(X, (int, float)) and isinstance(X, int) parameter validation.

Audit Findings:
1. Core Search API (_validate_search_params): top_k already had an explicit or isinstance(top_k, bool) guard, and this PR adds the matching guard for threshold. Because Memory.search(), Memory.get_all(), AsyncMemory.search(), and AsyncMemory.get_all() all route through _validate_search_params(), this completely closes the gap for all primary search & retrieval entry points.
2. Provider & Filter Helpers: Most vector store query builders (e.g., Databricks, Azure AI Search, CLI configs, temporal notices) already evaluate isinstance(v, bool) in an earlier if/elif branch before reaching isinstance(v, (int, float)).
3. FAISS Index Deserialization (mem0/vector_stores/faiss.py:113): index_to_id key validation (if not isinstance(key, int):) shares the same isinstance(True, int) gap.

I can include the FAISS index_to_id boolean guard and unit test directly in this PR if preferred, or keep this PR strictly focused on the core search parameter validation (threshold) and open FAISS as a quick follow-up PR. Let me know what you prefer!

agarwalpranav0711:
recheck

Issue #6861, opened 2026-08-08
OpenAIEmbedding hardcodes batch size 100, breaking OpenAI-compatible providers with lower limits


Mchenr:
### Component

Python SDK

### Summary

`OpenAIEmbedding.embed_batch()` currently hardcodes `MAX_BATCH = 100`. Some OpenAI-compatible embedding providers have lower per-request limits. For example, DashScope `text-embedding-v4` accepts at most 10 texts per request.

A direct `embed_batch()` call with 11 texts therefore sends all 11 in one request and fails with HTTP 400. In Mem0 code paths that catch batch failures, this also causes an avoidable failed request followed by sequential fallback instead of efficient provider-compatible batching.

The same hardcoded value also exists in the TypeScript OpenAI embedder, although the live reproduction below was performed with the Python SDK.

### Steps to Reproduce

```python
import os

from mem0.configs.embeddings.base import BaseEmbedderConfig
from mem0.embeddings.openai import OpenAIEmbedding

embedder = OpenAIEmbedding(
    BaseEmbedderConfig(
        model="text-embedding-v4",
        api_key=os.environ["DASHSCOPE_API_KEY"],
        openai_base_url=os.environ["DASHSCOPE_BASE_URL"],
        embedding_dims=1024,
    )
)

texts = [f"Mem0 DashScope batch probe {i}" for i in range(11)]
embedder.embed_batch(texts)
```

Live results against a DashScope OpenAI-compatible endpoint:

- 1 input: success, 1 × 1024-dimensional vector returned
- 10 inputs: success, 10 × 1024-dimensional vectors returned
- 11 inputs: HTTP 400 — `batch size is invalid, it should not be larger than 10`

### Expected Behavior

The maximum embedding batch size should be configurable (or derived by a provider-specific adapter), so a limit of 10 processes 11 inputs as two API calls with sizes `[10, 1]`.

The existing default of 100 can remain for backward compatibility.

### Actual Behavior

The Python implementation uses an unconditional `MAX_BATCH = 100`:

https://github.com/mem0ai/mem0/blob/main/mem0/embeddings/openai.py

The TypeScript implementation uses the same constant:

https://github.com/mem0ai/mem0/blob/main/mem0-ts/src/oss/src/embeddings/openai.ts

### Suggested Fix

Add an optional positive embedding batch-size setting to the embedder configuration and use it in `OpenAIEmbedding.embed_batch()`. Suggested regression coverage:

[...]

PR #6862, opened 2026-08-08
fix(embeddings): make OpenAI embed_batch size configurable

eeshsaxena:
### What / why

`OpenAIEmbedding.embed_batch()` hardcodes the request batch size to 100:

```python
MAX_BATCH = 100
for i in range(0, len(texts), MAX_BATCH):
    ...
```

100 is OpenAI's own per-request limit, but several OpenAI-compatible providers cap it lower. DashScope `text-embedding-v4`, for example, accepts at most 10 texts per request. So an `embed_batch()` of 11 texts against DashScope goes out in a single call and comes back as HTTP 400. In the paths that catch a batch failure this also means an avoidable failed request followed by a slower per-text fallback, instead of correct provider-sized batching.

Reported in #6861.

### Change

- Add an optional `embedding_batch_size` to `BaseEmbedderConfig`.
- Use it in `embed_batch()` as `self.config.embedding_batch_size or 100`, so the default is unchanged and anyone can lower it to match their provider.

It is fully backward compatible: existing configs do not set the field, so they keep the 100 default. A falsy value (0 or None) also falls back to 100.

```python
from mem0.embeddings.openai import OpenAIEmbedding

embedder = OpenAIEmbedding(
    BaseEmbedderConfig(
        model="text-embedding-v4",
        openai_base_url=DASHSCOPE_BASE_URL,
        embedding_batch_size=10,   # now batches 10 at a time
    )
)
embedder.embed_batch([f"text {i}" for i in range(11)])  # two requests: 10 + 1
```

### Tests

Added two tests to `tests/embeddings/test_openai_embeddings.py`:

- `test_embed_batch_respects_configured_batch_size`: with `embedding_batch_size=10`, 11 texts go out as two requests (10 then 1).
- `test_embed_batch_defaults_to_100`: without the override, 150 texts still go out as 100 + 50.

Note: the TypeScript OpenAI embedder in `mem0-ts` has the same hardcoded 100. I kept this PR to the Python SDK where the issue was reproduced; happy to follow up on the TS side in a separate PR if you'd like.

### DIFF
--- a/mem0/configs/embeddings/base.py
+++ b/mem0/configs/embeddings/base.py
@@ -40,6 +40,8 @@ class BaseEmbedderConfig(ABC):
         aws_secret_access_key: Optional[str] = None,
         aws_session_token: Optional[str] = None,
         aws_region: Optional[str] = None,
+        # Batch embedding
+        embedding_batch_size: Optional[int] = None,
     ):
         """
         Initializes a configuration class instance for the Embeddings.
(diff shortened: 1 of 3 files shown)


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #6875, opened 2026-08-10
fix(cassandra): push user_id/agent_id/run_id filters to CQL WHERE clause

HrushiYadav:
## Problem

`search()` and `list()` issued a bare `SELECT * FROM table` with no WHERE clause, then applied filters in Python after fetching every row. In a multi-tenant deployment this means every user's memories are read from Cassandra and held in Python memory during each search call — data for user A is transiently exposed during user B's query.

## Fix

- Add `user_id TEXT`, `agent_id TEXT`, `run_id TEXT` as dedicated columns with secondary indexes to the table schema.
- `insert()` now populates these columns from the payload dict.
- `update()` keeps them in sync when a payload update is applied.
- `search()` and `list()` build a parameterized CQL `WHERE` clause for any of these three keys present in `filters`, so only the relevant rows are fetched from Cassandra. Non-indexed payload keys are still checked in Python after the fetch.

## Tests

Added tests verifying:
- `search()` calls `session.prepare()` with a CQL query containing `WHERE user_id = ?` when a `user_id` filter is supplied.
- Non-indexed filter keys do not produce a prepared statement (Python post-filter path).
- `list()` similarly pushes indexed filters to CQL.
- `insert()` passes `user_id`/`agent_id`/`run_id` as explicit CQL parameters.

### DIFF
--- a/mem0/vector_stores/cassandra.py
+++ b/mem0/vector_stores/cassandra.py
@@ -145,15 +145,28 @@ class CassandraDB(VectorStoreBase):
     def _create_table(self):
         """Create table with vector column if it doesn't exist."""
         try:
-            # Create table with vector stored as list<float> and payload as text (JSON)
+            # Create table with vector stored as list<float> and payload as text (JSON).
+            # user_id, agent_id, run_id are stored as dedicated columns so they can be
+            # used in CQL WHERE clauses with secondary indexes, avoiding full-table scans
+            # that would expose all users' data during filtering.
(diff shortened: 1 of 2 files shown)


HrushiYadav:
Rebased on latest main. @kartik-mem0 — these four vector-store filter fixes (#6875, #6876, #6877, #6878) are all small, self-contained patches. Would appreciate a quick look when you're free.

PR #6887, opened 2026-08-10
fix(vector-stores): normalize LangChain search scores

### LINKED ISSUE #6884: bug(vector_stores/langchain): search() returns raw distance as the similarity score, so ranking is inverted on distance-based stores (e.g. FAISS)
### Component

Vector Store

### Summary

`VectorStoreBase.search` documents a hard contract (`mem0/vector_stores/base.py`):

> All implementations must return similarity scores where higher values indicate greater similarity ... Implementations using distance metrics must convert to similarity before returning:
> - Cosine distance: `score = max(0.0, 1.0 - distance)`
> - L2 distance: `score = 1.0 / (1.0 + distance)`
> - Inner product: `score = value`

Every native store follows this (pgvector, redis, valkey, weaviate, supabase, vertex, turbopuffer, s3_vectors, oracledb, milvus, cassandra, ...). The LangChain wrapper does not.

In `mem0/vector_stores/langchain.py::search`, when the underlying LangChain vector store exposes `similarity_search_with_score_by_vector` (this is what LangChain's **FAISS** and **Qdrant** integrations expose), mem0 returns the score unchanged:

```python
results = method(**kwargs)  # method = similarity_search_with_score_by_vector
return [
    OutputData(id=..., score=float(score), payload=...)  # score passed through as-is
    for doc, score in results
]
```

For a LangChain **FAISS** store, `score` is the raw **L2 distance** (lower = closer). mem0's `score_and_rank` (`mem0/utils/scoring.py`) then treats it as a similarity:

- it gates `semantic_score < threshold` (default `0.1`), which **drops the nearest matches** (a near-duplicate at distance ~0.05 is filtered out);
- it sorts **descending**, so the **farthest** candidates come back **first**.

Net effect: on a LangChain FAISS-backed `Memory`, `search()` returns results roughly farthest-first, and the closest matches are silently discarded by the threshold.

This is the same class already fixed for other stores — pgvector (#6883), Baidu (#6484), Chroma (#4999). The earlier LangChain PR #5072 fixed a *different* problem (a `TypeError` when the score was `None`); it did not touch the distance-vs-similarity direction.

### Steps to Reproduce

No live FAISS needed — the wrapper logic is enough. A LangChain FAISS store returns `(Document, L2_distance)` from `similarity_search_with_score_by_vector`:

```python
import sys, types
lc = types.ModuleType("langchain_community")
vs = types.ModuleType("langchain_community.vectorstores")
class VectorStore: ...
vs.VectorStore = VectorStore
lc.vectorstores = vs
sys.modules["langchain_community"] = lc
sys.modules["langchain_community.vectorstores"] = vs

class Doc:
    def __init__(self, id, meta): self.id, self.metadata = id, meta

class FakeFaiss(VectorStore):
    # FAISS returns L2 distance, lower = closer
    def similarity_search_with_score_by_vector(self, embedding, k, filter=None):
        return [
            (Doc("A", {"data": "nearest"}),  0.05),
            (Doc("B", {"data": "middle"}),   0.90),
            (Doc("C", {"data": "farthest"}), 1.80),
        ]

store = Langchain(client=FakeFaiss())
out = store.search(query="q", vectors=[0.1, 0.2], top_k=3)
print([(o.id, o.score) for o in out])
# [('A', 0.05), ('B', 0.9), ('C', 1.8)]  -> distances, not similarities

cands = [{"id": o.id, "score": o.score, "payload": o.payload} for o in out]
print([r["id"] for r in score_and_rank(cands, {}, {}, threshold=0.1, top_k=3)])
# ['C', 'B']  -> farthest first; nearest 'A' dropped by the threshold gate
```

### Expected Behavior

`store.search(...)` should return similarity scores (higher = better), so `score_and_rank` ranks the nearest match (`A`) first and does not drop it.

### The wrinkle (why I wanted to check direction before sending a PR)

The trouble is that `similarity_search_with_score_by_vector` is **not consistent across LangChain stores**: FAISS returns an L2 **distance** (lower = better), while Qdrant returns a cosine **similarity** (higher = better) from the same method name. So a blanket `1/(1+score)` conversion in the wrapper would fix FAISS but invert Qdrant.

A couple of directions, happy to implement whichever you prefer:

1. Prefer the normalized method (`similarity_search_by_vector_with_relevance_scores`, already tried first) and, when only a raw-score method is available, use LangChain's own `_select_relevance_score_fn` / `relevance_score_fn` on the store to normalize — that's exactly what LangChain does to turn native scores into `[0, 1]` relevance.
2. Add a small per-store hint (metric / higher-is-better) on the mem0 LangChain config so the wrapper knows how to normalize.

Option 1 keeps it self-contained and matches LangChain's own contract. If that sounds right I'll open the PR with a regression test (the repro above, asserting the nearest item survives the threshold and ranks first).

uuzzrm:
## Linked Issue

Closes #6884

## Description

LangChain vector stores do not all use the same meaning for the float returned by a scored search. The adapter was forwarding raw values as similarity scores, so FAISS L2 distances were treated as higher-is-better scores. A close result could then be filtered out by the semantic threshold while a farther result ranked ahead of it.

This patch:

- uses LangChain's relevance-score function for distance-based stores
- handles Chroma's raw-distance method despite its `relevance_scores` name
- falls back to mem0's documented `1 / (1 + distance)` mapping when a store does not expose a normalizer
- leaves similarity scores from cosine and dot-product stores unchanged
- adds regression coverage for distance stores, Chroma, and similarity-preserving clients

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [x] I tested manually (FAISS L2 smoke test confirms the nearest result is ranked first)

Focused checks completed locally:

- `pytest tests/vector_stores/test_langchain_vector_store.py -q` — 18 passed
- `ruff format --check` and `ruff check` on the changed files — passed
- FAISS smoke test with near and far vectors — passed

The full test suite could not be collected in this environment because the optional vector-store dependencies are not installed. The unrelated proxy tests also attempt to install `litellm` through a virtual environment without `pip`.

## Checklist

- [ ] New and existing tests pass locally (see test-environment limitation above)

AI assistance was used to investigate the issue and prepare this patch. The changed code and tests were reviewed before opening the PR.

### DIFF
--- a/mem0/vector_stores/langchain.py
+++ b/mem0/vector_stores/langchain.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 
 class OutputData(BaseModel):
     id: Optional[str]  # memory id
-    score: Optional[float]  # distance
+    score: Optional[float]  # similarity score
     payload: Optional[Dict]  # metadata
 
 
@@ -26,8 +26,8 @@ class OutputData(BaseModel):
 # but exposed by several concrete implementations.
 _SCORED_BY_VECTOR_METHODS = [
     "similarity_search_by_vector_with_relevance_scores",  # Chroma
-    "similarity_search_with_score_by_vector",             # FAISS, Qdrant
-    "similarity_search_by_vector_with_score",             # Pinecone, YDB
+    "similarity_search_with_score_by_vector",  # FAISS, Qdrant
+    "similarity_search_by_vector_with_score",  # Pinecone, YDB
 ]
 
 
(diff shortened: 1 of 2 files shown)


uuzzrm:
The code-focused checks are green on head `655fec1d`:

- Python SDK build checks for 3.10 and 3.11 passed.
- Mem0 Plugin tests for 3.10, 3.11, and 3.12 passed.
- Changelog, change validation, formatting/lint, and CLA checks passed.

The remaining CI items are external to the patch: `Python SDK / build_mem0 (3.12)` is still pending, and the Vercel preview reports `Authorization required to deploy` for the Mem0 team. I cannot authorize that team deployment from this contributor account. The focused vector-store regression suite and FAISS smoke test are documented in the PR description.

uuzzrm:
CI status update on head `655fec1d`:

- `Python SDK / build_mem0 (3.10)`, `(3.11)`, and `(3.12)`: passed
- Mem0 Plugin tests for Python 3.10, 3.11, and 3.12: passed
- CI gate, changelog, change validation, and CLA checks: passed
- The only remaining failed check is the Vercel preview, which reports `Authorization required to deploy` for the Mem0 team. I cannot authorize that team deployment from this contributor account.

There is no code-test failure on the current head; it is ready for maintainer review once the preview authorization gate is handled.

PR #6888, opened 2026-08-11
fix(ts-oss): surface entity link failures at console.warn

### LINKED ISSUE #6795: Entity embedding failures are swallowed at console.debug in the TS SDK add/search paths
Follow-up to #5509. @kartik-mem0 flagged this in the triage notes there as a second drop site for a separate change, so filing it on its own rather than growing #6794.

#6794 covers the memory paths: extraction (`infer: true`), raw writes (`infer: false`), and vector store insert failures on both. Entity linking is deliberately not covered there, and this is that gap.

In `mem0-ts/src/oss/src/memory/index.ts` the entity paths catch embedding and store errors and log them at `console.debug`, which is filtered out in most production logging configs. Current sites on `main`:

- entity re-embed during cleanup, in `_removeMemoryFromEntityStore`
- entity embed in `_linkEntitiesForMemory`
- entity update and insert in the same path
- the batch entity embed fallback in the phase 7 linking block, and the entity boost path in search

The consequence is different from #5509 and less severe, which is why it is separate. A dropped entity does not lose a stored memory. The memory row is still written and still searchable by vector similarity. What is lost is the entity link, so entity-boosted search and entity filtering silently return less than they should, and nothing surfaces it.

Two options that would both be an improvement:

1. Raise the log level from `console.debug` to `console.warn` so the drop is at least observable. Smallest possible change, no API surface.
2. Collect entity failures the same way #6794 collects memory failures and expose them, so a caller can tell that linking was incomplete.

I would lean towards 1 on its own first, since it is a one-line-per-site change that makes the failure visible without committing to any new public shape, and 2 can follow if it turns out to matter in practice.

Happy to do either, just say which. Not starting on it until #6794 settles so the two do not collide in the same file.

Hexc01:
## Linked Issue

Closes #6795

## Description

Entity linking/cleanup failures on the TS OSS `add` / search / cleanup paths were logged with `console.debug`, which most production log configs filter out. A dropped entity never loses the stored memory row, so the system still looks healthy while entity-boosted search quietly gets worse over time.

This raises those entity-path failure logs from `console.debug` to `console.warn`. No public API change. Also logs the previously silent per-entity embed failure in the phase-7 batch-embed fallback.

Covered drop sites in `mem0-ts/src/oss/src/memory/index.ts`:
- exact entity lookup / entity store unavailable / list failures during cleanup
- entity delete / re-embed / update / cleanup errors
- entity embed / update / insert / link errors in `_linkEntitiesForMemory`
- entity update failure and per-entity embed fallback failure in phase 7 `add()`

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `mem0-ts/src/oss/tests/memory.entity-log-level.test.ts` covering:
1. entity embed failure during `_linkEntitiesForMemory`
2. entity re-embed failure during cleanup
3. entity insert failure during `_linkEntitiesForMemory`

Ran locally:
```bash
cd mem0-ts && pnpm exec jest src/oss/tests/memory.entity-log-level.test.ts --coverage=false
# 3 passed
```

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -365,7 +365,7 @@ export class Memory {
           : (listed as any)
       ) as Array<{ id: string; payload: Record<string, any> }>;
     } catch (e) {
-      console.debug(
+      console.warn(
         `Exact entity lookup failed, falling back to semantic dedup: ${e}`,
       );
       return rowsByText;
@@ -398,7 +398,7 @@ export class Memory {
     try {
       entityStore = await this.getEntityStore();
     } catch (e) {
-      console.debug(`Entity store unavailable during cleanup: ${e}`);
+      console.warn(`Entity store unavailable during cleanup: ${e}`);
       return;
     }
 
@@ -411,7 +411,7 @@ export class Memory {
           : (listed as any)
       ) as Array<{ id: string; payload: Record<string, any> }>;
     } catch (e) {
-      console.debug(`Entity store list failed during cleanup: ${e}`);
+      console.warn(`Entity store list failed during cleanup: ${e}`);
       return;
     }
 
@@ -428,7 +428,7 @@ export class Memory {
           try {
             await entityStore.delete(row.id);
           } catch (e) {
-            console.debug(`Entity delete failed for id=${row.id}: ${e}`);
+            console.warn(`Entity delete failed for id=${row.id}: ${e}`);
           }
         } else {
           const newPayload = { ...payload, linkedMemoryIds: remaining };
@@ -437,7 +437,7 @@ export class Memory {
             typeof payload.data === "string" ? payload.data : "";
           if (!entityText) {
             // Can't re-embed without text; skip gracefully.
-            console.debug(
+            console.warn(
               `Entity id=${row.id} missing 'data'; skipping update during cleanup`,
             );
             continue;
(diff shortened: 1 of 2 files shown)


shashiKundur1:
Thanks for picking this up @Hexc01, and nice sweep. All thirteen `console.debug` sites, tests, no API change. That is the version of #6795 I would have written.

One thing I want to flag, and I am not asking you to widen the PR. There are two entity sites that log nothing at all, so a level change does not reach them:

`mem0-ts/src/oss/src/memory/index.ts:508` in `_linkEntitiesForMemory`, and `:1196` on the phase 7 `add()` path:

```ts
if (!exactMatch) {
  try {
    matches = await entityStore.search(entityVec, 1, filters);
  } catch {}
}
```

These are worse than quiet. If that search throws, `matches` stays empty, so `semanticMatch` is undefined, so `match` is undefined, and control falls into the `else` branch that inserts a **new** entity row. A store that already holds the entity under a different surface form gets a duplicate instead of a link.

Stubbed the entity store, same input both times, only difference is whether `search()` throws:

```
healthy store:   { inserts: 0, updates: 2, logged: 0 }   linked to the existing row
search throws:   { inserts: 2, updates: 0, logged: 0 }   two duplicate rows
```

Zero log lines in both. Nothing to raise the level of, and the store quietly diverges.

To be clear this is pre-existing, it is on `main` today and it is still on my own #6794 branch too, so it is not something your PR introduced or missed. It is just the part of #6795 that a log level cannot cover, since the fix has to be "do not treat a failed lookup as no match found."

Happy to open that as a follow-up once this lands, unless you would rather take it. Either way this one looks good to me.

Hexc01:
Thanks @shashiKundur1 — really appreciate the careful review and the reproduction on the empty `search()` catches.

Agreed those two sites are out of scope for a log-level change: a failed lookup is being treated as "no match," which can insert duplicates with zero signal. Happy to take that as a follow-up once this lands (warn on the failure and avoid inserting a new entity row when search throws). If you'd rather own it after #6794 settles, that works for me too.

shashiKundur1:
Thanks, I'll take it.

I checked whether it actually needs to wait, and it doesn't. Your diff never touches either site: `git diff main...` on your branch has no `catch {}` lines in it at all, which makes sense, since those two swallow silently and a level change can't reach them. So I'll open the follow-up against main now instead of queuing it behind this. If #6888 lands first the rebase should be a no-op, and if mine lands first yours is unaffected.

Scope I have in mind, kept narrow:

- warn when `entityStore.search()` throws, with the entity name and the error
- stop treating a thrown lookup as "no match found". Right now control falls straight through to the branch that inserts a new entity row, so a transient store error silently forks the graph instead of linking to the row that already exists.

Nothing else in those functions changes.

shashiKundur1:
Opened it as #6926, with #6925 for the bug itself so the two stay separate from #6795. Two sites, warn and skip instead of falling through to the insert. Your diff and mine do not touch the same lines, so they can land in either order.

PR #6891, opened 2026-08-11
fix(proxy): restore background memory writes

### LINKED ISSUE #6819: Proxy chat silently writes no memories — `_async_add_to_memory` passes `filters=` to `Memory.add()`, which no longer accepts it
### Component

Python SDK

### Summary

`mem0/proxy/main.py::Completions._async_add_to_memory` calls `add()` with a `filters` keyword:

```python
def _async_add_to_memory(self, messages, user_id, agent_id, run_id, metadata, filters):
    def add_task():
        logger.debug("Adding to memory asynchronously")
        self.mem0_client.add(
            messages=messages,
            user_id=user_id,
            agent_id=agent_id,
            run_id=run_id,
            metadata=metadata,
            filters=filters,          # <-- not a parameter of Memory.add()
        )

    threading.Thread(target=add_task, daemon=True).start()
```

`Memory.add()`'s signature is keyword-only and has no `filters` parameter:

```python
def add(self, messages, *, user_id=None, agent_id=None, run_id=None,
        metadata=None, timestamp=None, expiration_date=None,
        infer=True, memory_type=None, prompt=None)
```

So every proxy write raises `TypeError` — including when the caller passes no filters at all, because `filters=None` is still passed positionally-by-keyword. Because the call runs in a fire-and-forget daemon thread with no `try/except`, `chat.completions.create()` returns a perfectly normal completion and the user never sees a failure. Nothing is ever written to memory, so the proxy also never has anything to recall.

This is a separate root cause from the read path in #4907 (that issue is about `search()` rejecting top-level entity params and the `MemoryClient` result-iteration bug). Fixing the search path leaves the write path broken.

### Steps to Reproduce

```python
from mem0.proxy.main import Mem0

client = Mem0(config={...})   # OSS mode -> self.mem0_client is a Memory instance

client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "I'm vegetarian and allergic to nuts."}],
    user_id="alice",
)
# returns a normal completion; the background write has already died


# The failing call, isolated (this is exactly what add_task() invokes):
from mem0 import Memory

m = Memory()
m.add(
    messages=[{"role": "user", "content": "hi"}],
    user_id="alice",
    agent_id=None,
    run_id=None,
    metadata=None,
    filters=None,
)
```

### Expected Behavior

The proxy stores the conversation, so a later `chat.completions.create()` for the same `user_id` is augmented with the memories written by earlier turns. Background write failures should also be caught and logged rather than dying inside an unobserved thread.

### Actual Behavior

`add()` raises immediately; the exception surfaces only via `threading.excepthook` and the chat call reports success:

```
TypeError: Memory.add() got an unexpected keyword argument 'filters'
```

Nothing is persisted. Every subsequent proxy call retrieves an empty memory set.

### Environment

FU-max-boop:
## Linked Issue

Closes #6819

## Description

The proxy background writer passed the same keyword arguments to both memory backends even though their add APIs use different identity shapes. OSS Memory.add rejects filters entirely, so its daemon thread raised TypeError before storing any conversation. MemoryClient.add, on the other hand, expects entity IDs inside filters.

This change:

- calls OSS Memory.add with top-level user_id, agent_id, and run_id only;
- merges hosted entity IDs into a copy of filters for MemoryClient.add, with explicit entity arguments taking precedence;
- catches and logs background write failures so they no longer disappear in an unobserved daemon thread.

The caller-provided filters dictionary is not mutated. There is no public API change.

## Scope note

This PR fixes the background **write** failure in #6819. The subsequent OSS search path has a separate top-level-entity-parameter bug tracked in #4907, with an existing fix in #4908. That independent search issue can still prevent the full proxy `create()` flow from completing after the write is scheduled; it is deliberately not duplicated or folded into this focused PR.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added deterministic regression coverage for the OSS signature, hosted filter merging, input immutability, and background exception logging. I also ran a real daemon-thread probe for both the successful write and failure-log paths.

Local validation:

- Python 3.12 full suite: 1814 passed, 75 skipped
- Python 3.10, 3.11, and 3.12 proxy tests: 10 passed on each version
- Ruff 0.16.0 check: full repository passed
- isort check: changed Python files passed
- compileall and git diff --check: passed

## Checklist

- [x] I have updated documentation if needed (not needed: internal bug fix with no public API change)

### DIFF
--- a/mem0/proxy/main.py
+++ b/mem0/proxy/main.py
@@ -148,15 +148,21 @@ class Completions:
 
     def _async_add_to_memory(self, messages, user_id, agent_id, run_id, metadata, filters):
         def add_task():
-            logger.debug("Adding to memory asynchronously")
-            self.mem0_client.add(
-                messages=messages,
-                user_id=user_id,
-                agent_id=agent_id,
-                run_id=run_id,
-                metadata=metadata,
-                filters=filters,
-            )
+            try:
+                logger.debug("Adding to memory asynchronously")
+                if isinstance(self.mem0_client, MemoryClient):
+                    entity_filters = {
+                        key: value
+                        for key, value in {"user_id": user_id, "agent_id": agent_id, "run_id": run_id}.items()
+                        if value is not None
+                    }
+                    add_kwargs = {"filters": {**(filters or {}), **entity_filters}}
+                else:
+                    add_kwargs = {"user_id": user_id, "agent_id": agent_id, "run_id": run_id}
+
+                self.mem0_client.add(messages=messages, metadata=metadata, **add_kwargs)
+            except Exception:
+                logger.exception("Failed to add messages to memory asynchronously")
 
         threading.Thread(target=add_task, daemon=True).start()
 
--- a/tests/test_proxy.py
+++ b/tests/test_proxy.py
@@ -1,8 +1,9 @@
 import builtins
 import importlib
 import inspect
+import logging
 import sys
-from unittest.mock import Mock, patch
+from unittest.mock import Mock, create_autospec, patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


ebarkhordar (commented):
The write fix reproduces as a differential. Clean `python:3.12-slim`, `pip install .` at each commit, default `Memory()` config, `_async_add_to_memory` driven directly with the thread stubbed out:

- `main` (14c43173): `TypeError: Memory.add() got an unexpected keyword argument 'filters'`
- this branch (d6e1f92): `add_task` runs to the backend call (mine stopped at an OpenAI 401 from a dummy key, which the new `logger.exception` recorded, so that half works too)

The hosted branch matches `AddMemoryOptions`, which is the authority: *"Identity fields (user_id, agent_id, app_id, run_id) must be passed inside the `filters` dict, the v3 API does not accept them at the top level"*. `MemoryClient.add`'s own `**kwargs` docstring still says the opposite, which is probably what makes this bug easy to reintroduce.

One thing worth flagging, because it decides whether merging this clears the symptom in #6819. `create()` calls `_fetch_relevant_memories` on the line right after `_async_add_to_memory`, and that one still passes the entity IDs at the top level:

```python
return self.mem0_client.search(
    query="\n".join(message_input),
    user_id=user_id,
    agent_id=agent_id,
    run_id=run_id,
    filters=filters,
    top_k=top_k,
)
```

Same container, this branch, OSS backend:

```
ValueError: Top-level entity parameters frozenset({'agent_id', 'user_id', 'run_id'}) are not supported in search(). Use filters={'user_id': '...'} instead.
```

That call is not in a daemon thread, so `create()` raises there before the LLM call. Both halves have the same origin, `a488e19` (#4805), which added `_reject_top_level_entity_params` to `add` and `search` together, and the search half is already filed as #4907 with an open fix in #4908.

Nothing wrong with the scope you chose, but a reader of #6819 will still find the proxy unusable after this merges. Linking #4907 in the description, or folding the same reshaping into `_fetch_relevant_memories`, would close the loop.

FU-max-boop:
Thanks for the differential verification and for tracing the subsequent search failure. I rechecked that path and updated the PR description to state explicitly that this PR fixes the background write half of #6819, while #4907/#4908 own the separate search reshaping. I will keep that fix out of this branch to avoid duplicating an active implementation or expanding this PR beyond its tested scope.

PR #6896, opened 2026-08-11
docs: refresh references stale since the v3 pipeline landed

### LINKED ISSUE #6895: Agent context files still describe OSS graph memory removed in v3
## Problem

`a488e1904` (v3 pipeline port) deleted `mem0/graphs/`, `mem0/memory/graph_memory.py`, `mem0/memory/memgraph_memory.py`, `mem0-ts/src/oss/src/graphs/`, and `mem0-ts/src/oss/src/memory/graph_memory.ts`. `3f717e545` reframed the docs so graph memory reads as a Platform feature.

Neither commit updated the agent-facing context files. `AGENTS.md` (symlinked as `CLAUDE.md`) still lists `mem0/graphs/` in the directory map and the dependency tree, still counts a "Graph Stores" provider category with 4 providers, and still describes graph memory as an optional OSS layer configured through `MemoryConfig`.

The effect is concrete: every agent that reads `AGENTS.md` before touching this repo starts from a false model of the SDK and will happily propose `graph_store` config and `enable_graph` flags that cannot work.

Three smaller drifts share the same root cause:

- `skills/mem0/scripts/mem0_doc_search.py` indexes `/open-source/features/graph-memory`, which now 404s. Four other `SECTION_MAP` paths also no longer resolve (`/platform/features`, `/platform/features/selective-memory`, `/open-source/features`, `/open-source/configure-components`).
- `mem0/exceptions.py` uses `kuzu` / `graph_store` as its `DependencyError` example.
- `tests/test_telemetry.py` sets `mock_memory.config.graph_store.config` in four places. `graph_store` is not a `MemoryConfig` field and `telemetry.py` never reads it, so these lines only pass because `MagicMock` accepts any attribute.

Separately, `CONTRIBUTING.md`'s tag-prefix table is missing three prefixes that `release.yml` actually routes (`opencode-v*`, `pi-agent-v*`, `n8n-nodes-mem0-v*`).

## Scope

Documentation and agent context only. No SDK behavior changes. The provider counts in `AGENTS.md` are also recounted from the tree, since they had drifted independently.

Note that not every `graph` reference is stale: the v2→v3 migration notes in `skills/mem0/references/features.md`, `skills/mem0-oss-to-platform/`, and the Vercel AI SDK regression tests are correct as written, since they document the removal. Those stay.

kartik-mem0:
## Linked Issue

Closes #6895

## Description

`AGENTS.md` (symlinked as `CLAUDE.md`) is the first thing every coding agent reads in this repo, and it has been wrong since `a488e1904` removed OSS graph memory. This brings it, and four other files that drifted for the same reason, back in line with the code.

**`AGENTS.md`**
- Dropped `graphs/` from the key-directories row and the dependency tree
- Provider table corrected to 4 categories: the Graph Stores row is gone, and the counts are recounted from the tree (LLMs 24 → 18, vector stores 30 → 25, embeddings 15 → 11, rerankers 5 unchanged)
- `### Graph Memory` rewritten as `### Graph Memory (Platform only)`, naming the commit that removed it and stating explicitly that agents should not add `graph_store` config or Neo4j/Memgraph/Kuzu/Apache AGE providers to OSS

**Skills** (`skills/mem0/` and its `integrations/mem0-plugin/` mirror, kept byte-identical)
- Replaced the `"Add graph memory to track entity relationships"` prompt example
- Stopped indexing `/open-source/features/graph-memory`, which 404s, and fixed four more `SECTION_MAP` paths that no longer resolve. Every path in the map is now verified against `docs/`.

**`mem0/exceptions.py`** — `DependencyError` docstring example used `kuzu` / `graph_store`; now `chromadb` / `vector_store`.

**`tests/test_telemetry.py`** — removed four `mock_memory.config.graph_store.config = None` lines. `graph_store` is not a `MemoryConfig` field (verified) and `telemetry.py` never reads it, so they were dead setup that only passed because `MagicMock` accepts any attribute.

**`CONTRIBUTING.md`** — added the three tag prefixes `release.yml` routes but the table omitted (`opencode-v*`, `pi-agent-v*`, `n8n-nodes-mem0-v*`), verified against the `case` block, plus a note that the Zapier app deploys off-registry. This brings the table in line with the equivalent CD table in `.github/AGENTS.md`, which already carried all three rows and the same Zapier sentence.

**`tests/test_release_tag_prefixes.py`** (new) — drift guard asserting the `### Tag Prefixes` table and the `case` block in `release.yml` describe the same set of prefixes, in either direction. Companion to the provider-count guard below.

### What deliberately did not change

Not every `graph` reference is stale. The v2→v3 migration notes in `skills/mem0/references/features.md`, the `skills/mem0-oss-to-platform/` mapping tables, `.claude/skills/mem0-review/REVIEW-PARAMS.md`, and the Vercel AI SDK contract tests are all correct as written, because they document the removal or assert the absence. LangGraph mentions are unrelated. Platform doc paths like `/platform/features/graph-memory` still resolve and stay.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

`pytest tests/test_telemetry.py` passes (43 tests) after removing the dead mock lines. `ruff check` passes on both touched Python files. Every documentation path added to `SECTION_MAP` was checked against a file in `docs/`, and every tag prefix added to `CONTRIBUTING.md` was checked against the `case` block in `release.yml`.

## Checklist


karthik-indla:
**Review — requesting changes**

The core drift fixes here are accurate and I verified them against the tree: provider counts (LLMs 18 / vector stores 25 / embeddings 11 / rerankers 5) match the factory registries, the graph-memory removals are genuinely dead code (`mem0/graphs/` is gone, `MemoryConfig` has no `graph_store`/`enable_graph`, `telemetry.py` never reads it), and every `SECTION_MAP` path resolves. The drift test is a nice guard.

However, the `CONTRIBUTING.md` additions introduce **new** incorrect documentation — the exact failure class this PR is meant to eliminate:

1. **(blocking) The `n8n-nodes-mem0` / `n8n-nodes-mem0-v*` release row does not exist.** There is no `integrations/n8n*` directory, no n8n workflow under `.github/workflows/`, and no `n8n-nodes-mem0-v*` arm in `release.yml`'s `case` block (arms are `ts-v*`, `cli-node-v*`, `cli-v*`, `vercel-ai-v*`, `openclaw-v*`, `opencode-v*`, `pi-agent-v*`, `v*`). Please remove this row.
2. **(blocking) The Zapier paragraph references a nonexistent workflow/dir.** `gh workflow run zapier-mem0-cd.yml` — there is no `.github/workflows/zapier-mem0-cd.yml` and no `integrations/zapier-mem0`. That command would fail. Please drop it (or defer until the integration actually lands).

The other two added rows (`opencode-v*`, `pi-agent-v*`) are correct and were genuinely missing.

Minor: the PR description says "embeddings 15 → 12" but the diff correctly sets **11** (matching `EmbedderFactory`) — the code is right, just fix the description.

_(Static review against `main`; relied on the stated `pytest`/`ruff` runs for execution.)_

karthik-indla:
**Re-review — the two blockers are still outstanding**

Thanks for fixing the embeddings count (now 11, matching `EmbedderFactory`) and adding the provider-count drift test — that part looks good. But the two blocking items from the previous review are still present in `CONTRIBUTING.md`, so this can't be approved yet. Re-verified against the current tree:

1. **The `@mem0/n8n-nodes-mem0` / `n8n-nodes-mem0-v*` release row is still there.** There is no `integrations/*n8n*` directory, no n8n workflow under `.github/workflows/`, and no `n8n-nodes-mem0-v*` arm in `release.yml`'s `case` block (arms run through `pi-agent-v*` then the bare `v*` default). Please remove the row.
2. **The Zapier paragraph is still there** (`gh workflow run zapier-mem0-cd.yml --ref main`). There is no `zapier-mem0-cd.yml` workflow and no `integrations/*zapier*` directory. Please remove the paragraph.

Everything else in the PR is accurate — provider counts (18 / 25 / 11 / 5), the graph-memory removals, the `SECTION_MAP` path fixes, and the `opencode-v*` / `pi-agent-v*` rows all check out. Once those two fabricated entries are dropped, this is good to go.

kartik-mem0:
I checked both blockers against the tree before acting on them, and I think they're based on a bad read. Neither entry is fabricated. Concretely:

**1. `n8n-nodes-mem0-v*` is in the router.** `.github/workflows/release.yml`, line 48, between the `pi-agent-v*` and bare `v*` arms:

```
opencode-v*)  workflow="opencode-plugin-cd.yml" ;;
pi-agent-v*)  workflow="pi-agent-plugin-cd.yml" ;;
n8n-nodes-mem0-v*) workflow="n8n-nodes-mem0-cd.yml" ;;
v*)           workflow="cd.yml" ;;
```

`integrations/n8n-nodes-mem0/` exists (`package.json` name is `@mem0/n8n-nodes-mem0`, matching the row), as do `.github/workflows/n8n-nodes-mem0-cd.yml` and `n8n-nodes-mem0-checks.yml`. All landed in `d4869d24e` — *feat(integrations): n8n community node for Mem0* (#6517), 2026-07-29.

**2. The Zapier workflow exists**, added in `e168d48e0` — *feat(integrations): Zapier app for Mem0* (#6518), 2026-07-29, alongside `integrations/zapier-mem0/`. The disputed command is copied from that workflow's own header comment:

```yaml
# Zapier apps deploy to Zapier's own platform (not npm), so this is NOT wired
# into the npm release router (release.yml). It is manual workflow_dispatch
# only and requires the ZAPIER_DEPLOY_KEY repo secret.
#
#   gh workflow run zapier-mem0-cd.yml --ref main
on:
  workflow_dispatch:
```

Both commits predate the first review by two weeks, so this isn't a stale-checkout issue. `release.yml` is also byte-identical between `origin/main` and this branch — the PR never touched it.

Worth adding: `.github/AGENTS.md` already carried all three rows in its CD table and this exact sentence at line 65 — *"The Zapier app deploys to Zapier's platform, not npm, so it is not in the router. Deploy with `gh workflow run zapier-mem0-cd.yml --ref main`."* The PR was bringing `CONTRIBUTING.md` in line with documentation that was already correct, not inventing entries. If the n8n and Zapier rows were dropped, `CONTRIBUTING.md` would contradict `.github/AGENTS.md` and under-document a live release path.

**Instead of removing them, I've added a guard.** Since the provider-count drift test was useful here, `tests/test_release_tag_prefixes.py` does the same for this table: it parses the `### Tag Prefixes` rows and the `case` arms in `release.yml` and asserts the two sets match in both directions, so a row without an arm and an arm without a row both fail.

I verified it actually bites by deleting the n8n row and re-running:

```
E    Extra items in the right set:
E    'n8n-nodes-mem0-v*'
FAILED tests/test_release_tag_prefixes.py::test_documented_prefixes_match_release_router
```

So applying the requested change would now break CI, which is the outcome I'd want from a guard like this.

The minor note is right: the description said "embeddings 15 → 12" where the diff correctly sets 11. Fixing the description.

Happy to be shown otherwise if you're seeing something different on your end — a `git log -1 d4869d24e` and `grep -n n8n .github/workflows/release.yml` should be the quickest way to reconcile.

PR #6897, opened 2026-08-11
fix(llms/vllm): honor VLLM_BASE_URL instead of always using localhost

### LINKED ISSUE #6256: bug(llms/vllm): VLLM_BASE_URL env var is ignored, always connects to localhost

kartik-mem0:
## Linked Issue

Closes #6256

## Description

`VLLM_BASE_URL` was silently ignored. `VllmConfig.__init__` already resolved `vllm_base_url` to `"http://localhost:8000/v1"` when no explicit value was passed, so it was never `None` by the time `VllmLLM` read it. The `os.getenv("VLLM_BASE_URL")` fallback in `VllmLLM.__init__` was therefore unreachable dead code, and the client always connected to localhost regardless of the env var.

The fix moves the env-var fallback into `VllmConfig` itself, so precedence is resolved in exactly one place: explicit arg, then `VLLM_BASE_URL`, then the localhost default. `VllmLLM` now just uses `self.config.vllm_base_url` directly, removing the redundant fallback.

The `api_key` handling in the same two files was checked for the identical pattern: `VllmConfig` never defaults `api_key` to a non-`None` value, so the existing `self.config.api_key or os.getenv("VLLM_API_KEY") or "vllm-api-key"` fallback in `VllmLLM` is reachable and correct as-is, no change needed there.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added three tests to `tests/llms/test_vllm.py` covering all three precedence levels for `vllm_base_url`: explicit arg wins over env var, env var wins over the localhost default, and the default applies when neither is set. Also manually reproduced the original bug (`VLLM_BASE_URL=http://example.com:9999/v1` resolving to `http://localhost:8000/v1`) and confirmed it now resolves correctly.

## Checklist


karthik-indla:
**Review — looks good, a couple of minor notes**

The fix is correct and matches the root cause in #6256: precedence is now resolved in one place (`VllmConfig.__init__`: `vllm_base_url or os.getenv("VLLM_BASE_URL") or default`) and `VllmLLM` reads `self.config.vllm_base_url` straight through, giving `config > env > default`. `import os` is correctly added to the config module, and the `os` import in `vllm.py` is still needed for the `VLLM_API_KEY` fallback.

Minor notes (non-blocking):
- This resolves the env var inside the *config* class, whereas the sibling `DeepSeekLLM` keeps the config default `None` and resolves env in the *LLM* class (which is what the issue proposed). Both are correct; centralizing in config is arguably cleaner, but it's now inconsistent with the DeepSeek convention — worth a deliberate call on which pattern to standardize.
- The tests assert on `config.vllm_base_url`. Since resolution moved into the config, that proves the fix — but asserting that `OpenAI(...)` actually received the expected `base_url` would additionally guard the `VllmLLM` pass-through against a future regression. An explicit `VllmLLM(BaseLlmConfig(...))` + env case would also cover the original repro end-to-end.

karthik-indla (approved):

PR #6899, opened 2026-08-11
fix(llms/aws_bedrock): resolve provider for application inference profile ARNs

### LINKED ISSUE #5913: Bedrock application inference profile ARNs fail provider extraction
## Problem

`AWSBedrockLLM.__init__` uses regex-based provider detection via `extract_provider(model_arn)` to determine the model family (`anthropic`, `amazon`, `cohere`, etc.).

This works for:
- Foundation model IDs: `anthropic.claude-3-5-sonnet-20240620-v1:0`
- Cross-region inference profiles: `us.anthropic.claude-haiku-4-5-20251001-v1:0`

It **fails** for **application inference profile ARNs** because the ARN contains an opaque profile ID with no provider name:

```
arn:aws:bedrock:<region>:<account-id>:application-inference-profile/<opaque-id>
```

**Error:**
```
ValueError: Unknown provider in model: arn:aws:bedrock:...
```

Application inference profiles are AWS Bedrock's enterprise pattern for cost tagging and traffic routing — very common in production AWS deployments. mem0 users currently cannot use them at all.

## Related

PR #4139 fixed cross-region prefixes (`us.anthropic.*`, `apac.anthropic.*`) where the provider name is still present in the string. This issue is about the **distinct case** where the ARN contains **no provider hint at all**.

## Proposed Fix

Add an optional `bedrock_provider` field to `AWSBedrockConfig`. When set, it short-circuits `extract_provider()` so the caller explicitly names the underlying model family.

**Example config after fix:**

```python
config = {
    "llm": {
        "provider": "aws_bedrock",
        "config": {
            "model": "arn:aws:bedrock:<region>:<account>:application-inference-profile/<id>",
            "bedrock_provider": "anthropic",  # new — explicit override for opaque ARN
            "aws_region": "us-east-1",
            "temperature": 0.1,
            "max_tokens": 1000,
        }
    }
}
memory = Memory.from_config(config)  # no longer raises ValueError
memory.add(...)  # routes through Anthropic Converse API correctly
```

Fully backward compatible — `bedrock_provider` defaults to `None` and all existing configs are unchanged.

PR: https://github.com/mem0ai/mem0/issues/5913

kartik-mem0:
## Linked Issue

Closes #5913

## Description

`extract_provider(model)` in `mem0/llms/aws_bedrock.py` finds the provider by word-boundary regex against a known `PROVIDERS` list. This works for plain model ids (`anthropic.claude-...`) and cross-region inference profile ids (`us.anthropic.claude-...`), but an AWS application inference profile ARN, e.g. `arn:aws:bedrock:us-east-1:123456789012:application-inference-profile/abc123xyz`, has an opaque ID with no provider substring, so it raised `ValueError: Unknown provider in model: ...`.

The provider for an application inference profile cannot be derived from the ARN itself, so this adds an optional `provider_override` field on `AWSBedrockConfig`. `extract_provider` uses it when set, keeping the existing regex path as the default for plain model ids and cross-region inference profiles.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added tests to `tests/llms/test_aws_bedrock.py` covering: an application inference profile ARN without `provider_override` still raises, the same ARN with `provider_override` resolves and initializes `AWSBedrockLLM` correctly, and existing plain-model-id / cross-region-profile behavior is unchanged.

## Checklist


karthik-indla:
**Review — correct fix, some minor notes**

The change does exactly what #5913 asks: `provider_override` short-circuits `extract_provider()` before the regex loop, defaults to `None`, and the production call site is updated. The other `extract_provider` call site (inside `list_available_models`) is correctly left alone since it iterates foundation-model IDs that always carry a provider substring. Tests cover the ARN-without-override raise, ARN-with-override resolution, override-beats-regex precedence, and unchanged plain/cross-region behavior.

Minor notes (non-blocking):
- **TS parity:** the TypeScript OSS SDK has the identical bug — `extractProvider()` in `mem0-ts/src/oss/src/llms/aws_bedrock.ts` throws for the same application-inference-profile ARN shape with no override path. This PR is scoped Python-only, which is defensible, but worth a follow-up issue for TS.
- **Field name vs. issue:** the issue proposes `bedrock_provider`; this ships `provider_override`. The new name is arguably clearer, but the issue's example config (`"bedrock_provider": "anthropic"`) won't work as written — worth a note so users copy the right key.
- **No validation:** any string is accepted; a typo (e.g. `"anthorpic"`) silently falls through to default capabilities rather than erroring. A check against the known providers with a clear message would be friendlier.
- **Docs:** the new config field isn't documented in `docs/components/llms/models/aws_bedrock.mdx` — given the issue calls these ARNs "very common in production," a one-line mention would help.

karthik-indla (approved):

PR #6900, opened 2026-08-11
fix(client): honor page_size in get_all when page is not passed

### LINKED ISSUE #6227: MemoryClient.get_all() ignores page_size unless page is also passed

kartik-mem0:
## Linked Issue

Closes #6227

## Description

`MemoryClient.get_all()` (both `MemoryClient` and `AsyncMemoryClient`) only forwarded `page`/`page_size` as query params when both were passed together. `page` and `page_size` are independently optional on `GetAllMemoryOptions`, so passing `page_size` alone fell into the `else` branch and was sent inside the JSON body instead of as a query param, where the API does not read it, silently ignoring the requested page size.

The fix builds `query_params` from whichever of `page`/`page_size` is present, so either can be passed independently and both still work together. No query params are sent when neither is passed, matching prior behavior.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added `TestGetAllPageSizeQueryParams` to `tests/test_client.py` covering: `page_size` alone in query params, `page` alone in query params, both together, neither passed (no `params` kwarg sent), and the async client with `page_size` alone.

## Checklist


karthik-indla:
**Review — correct fix, one parity note**

The Python fix is right and minimal: `page`/`page_size` are now forwarded as query params independently, using presence (`key in params`) rather than truthiness — so `page=0`/`page_size=0` still forward correctly (and upstream `_prepare_params` already strips `None`). The dict-comprehension refactor is duplicated identically in the sync and async paths, and the tests cover page_size-alone, page-alone, both, and neither.

Minor notes (non-blocking):
- **TS parity (and it's the worse case):** the TypeScript SDK has the same bug — `mem0.ts` uses `if (page && pageSize)`, and `pageSize` is destructured out of `rest` so a lone `pageSize` never reaches the body at all. The truthiness check would also swallow `0`. Worth a follow-up issue so the fix lands on both SDKs.
- Only the sync client gets all four param combinations tested; the async client is covered just for the `page_size`-alone case. Acceptable given shared logic, but noting it.

karthik-indla (approved):

PR #6902, opened 2026-08-11
fix(ts-oss): return snake_case entity ids from the redis and valkey stores

### LINKED ISSUE #6254: TS OSS Redis store returns entity ids as camelCase, dropping them from results and leaking into metadata

### LINKED ISSUE #6266: TS OSS Valkey store returns entity ids as camelCase, dropping them from results and leaking into metadata

kartik-mem0:
## Linked Issue

Closes #6254
Closes #6266

## Description

The TS OSS Redis and Valkey vector stores were the only stores that ran their entire result payload through `toCamelCase` in `search()`, `get()`, and `list()`. That turned `user_id`/`agent_id`/`run_id` into `userId`/`agentId`/`runId`, but `mem0-ts/src/oss/src/memory/index.ts` reads the snake_case keys, so entity ids silently vanished from results and leaked into `metadata` instead.

Both stores write correctly (`insert`/`update` already snake_case the payload before storing), only the read path was wrong.

The fix adds a `toCamelCasePreservingIds` helper in each store: it camelCases the rest of the payload as before (so `createdAt`/`updatedAt`/`textLemmatized`/`attributedTo` keep round-tripping the way `memory/index.ts` and every other store expect), but keeps `agent_id`/`run_id`/`user_id` snake_case, matching `qdrant.ts` (which passes payloads through unchanged) and the in-memory store's `normalizePayload` (which maps `userId -> user_id` and leaves timestamps camelCase).

Left the Redis filter-building code (`buildRedisFilterExpr`/`escapeRedisTagValue`) untouched to avoid conflicting with #6568.

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

Added unit tests to `mem0-ts/src/oss/tests/redis.unit.test.ts` and `mem0-ts/src/oss/tests/valkey.test.ts` asserting that a record inserted with `user_id`/`agent_id`/`run_id` comes back from `search`, `get`, and `list` with those exact snake_case keys (and never the camelCase variants) for both stores. Updated one pre-existing Valkey assertion that had encoded the old, buggy camelCase `userId` behavior.

Ran `pnpm run test` (full suite, 1552 tests) and `npx tsc --noEmit` from `mem0-ts/` — both pass with no new failures. Ran prettier on all changed files.

## Checklist


karthik-indla:
**Review — core fix is right; please split out the unrelated harness changes**

The vector-store fix is correct and complete: only `redis.ts` and `valkey.ts` ran reads through `toCamelCase`, and both are updated (`toCamelCasePreservingIds` keeps `user_id`/`agent_id`/`run_id` snake_case). Redis's three read sites and Valkey's single `docToResult` funnel are all covered, and the unit tests assert both snake_case presence and camelCase absence across search/get/list. Closes #6254 and #6266 at the store layer.

Two things to address:
- **(please split) Undocumented scope creep.** Roughly half the diff is unrelated to this fix and isn't mentioned in the description: `jest.integration.config.js` drops `globalSetup`/`globalTeardown`, `helpers.ts` rewrites `waitForMemories` and removes `fullProjectCleanup`, and `management.test.ts` gets an `updatedUrl` tweak. Removing the global setup/teardown is a real behavioral change — integration tests no longer do a pre/post full-project wipe, so they no longer start from a clean slate. This belongs in its own PR with a rationale. Also, `global-setup.ts`/`global-teardown.ts` are now referenced nowhere and should be deleted rather than left as dead files. (Note: the same harness rework also appears in #6903 — it's currently landing in two PRs.)
- **`get()` metadata leak not fully closed.** After this fix the payload carries snake_case ids, and `search()`/`getAll()` exclude them from metadata correctly — but `get()`'s `excludedKeys` in `memory/index.ts` are still camelCase (`userId`/`agentId`/`runId`), so `get()` now returns each id both top-level and duplicated inside `metadata`, which contradicts the issues' "not present inside metadata" expectation. This is a pre-existing `index.ts` inconsistency (also affects Qdrant/in-memory) rather than a regression, and the unit tests don't catch it because they assert on the raw store payload — but worth a fast follow.

Minor: `toCamelCasePreservingIds` is duplicated verbatim in both files; a shared util would avoid drift.

_(Static review against `main`; relied on the stated suite run for execution.)_

karthik-indla (approved):

PR #6905, opened 2026-08-11
docs: correct memory decay claims that contradict the SDK

kartik-mem0:
## Linked Issue

This covers July 31 bug bash item 18 (flagged by Vasilii): the memory decay page and the implementation disagree in multiple places. No standalone issue exists for this item; fixing directly per the bug bash note. Three claims on `docs/platform/features/memory-decay.mdx` were checked against `mem0/client/project.py`, `mem0-ts/src/client/mem0.ts`, `docs/openapi.json`, and the sibling `docs/api-reference/organizations-projects.mdx` page, and were found to be contradicted:

1. The page claimed "there is currently no way to narrow the response to just this field" when reading a project's `decay` state. `mem0/client/project.py` (`Project.get(fields=...)`) and `mem0-ts/src/client/mem0.ts` (`getProject({ fields })`) both support narrowing via `fields`, and the sibling `organizations-projects.mdx` page documents `?fields=decay` explicitly. This claim was in fact a regression from a prior docs-audit commit (`760dca6f3`) that flipped previously-correct content to this incorrect version.
2. The example PATCH response body was `{ "message": "Updated decay" }`. `docs/openapi.json` documents the actual response for this exact endpoint as `{ "message": "Project updated successfully" }`.
3. The JavaScript code samples called `client.project.update(...)` / `client.project.get(...)`. The TypeScript `MemoryClient` class (`mem0-ts/src/client/mem0.ts`) has no `.project` namespace; it exposes flat `updateProject(...)` / `getProject(...)` methods, matching the working examples already used on `docs/platform/features/custom-instructions.mdx`.

## Description

Reconciles the three JavaScript/response-shape claims above with what the SDKs and OpenAPI spec actually implement. All other claims on the page (the internal decay scoring algorithm, scaling bounds, candidate-pool widening, access-history mechanics) are hosted Platform behavior with no implementation in this repo, so they were left unverified and unmodified per the investigation scope.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

Docs-only change; verified with `python scripts/check-llms-txt-coverage.py` (passes) and by cross-checking every corrected claim against `mem0/client/project.py`, `mem0-ts/src/client/mem0.ts`, and `docs/openapi.json`.

## Checklist

### DIFF
--- a/docs/platform/features/custom-categories.mdx
+++ b/docs/platform/features/custom-categories.mdx
@@ -75,7 +75,7 @@ print(response)
 ```
 </CodeGroup>
 
-This "Updated custom categories" message is specific to a PATCH-style partial update, which is what `client.project.update()` sends. Calling the raw API with a full PUT instead returns a generic `{"message": "Project updated successfully."}`, regardless of which fields changed.
+Treat the `message` string as informational. The project endpoint accepts `PATCH` only, and its documented response is the generic `{"message": "Project updated successfully"}`. Confirm an update by reading the field back, as in the next step, rather than by matching on the message.
 
 ### 2. Confirm the active catalog
 
--- a/docs/platform/features/memory-decay.mdx
+++ b/docs/platform/features/memory-decay.mdx
@@ -64,7 +64,7 @@ client.project.update(decay=True)
 ```
 
 ```javascript JavaScript
-await client.project.update({ decay: true });
+await client.updateProject({ decay: true });
 ```
 
 ```bash cURL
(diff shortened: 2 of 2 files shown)


karthik-indla:
**Review — accurate corrections, a few things to tidy**

The three memory-decay fixes are correct: JS `updateProject`/`getProject` (the TS client has no `.project` namespace — only flat methods), the `{ "message": "Project updated successfully" }` response (matches `openapi.json`), and the `?fields=` narrowing. Keeping Python on `.project` while dropping it for JS is the right asymmetry — the Python client does have that namespace.

A few items worth addressing:

1. **(should fix) `custom-categories.mdx` now self-contradicts.** The new prose asserts the response is the generic `{"message": "Project updated successfully"}`, but the `json Output` block just above it still shows `{ "message": "Updated custom categories" }`. Update the example block or soften the prose so the page agrees with itself.
2. **(minor) PR-body justification is off.** It says the JS fix matches "working examples already used on `custom-instructions.mdx`" — but that page uses the *same* `client.project.update(...)` style this PR is correcting, so the cited precedent doesn't exist. The fix is still right on its own merits.
3. **(minor) Incomplete sweep.** The same broken JS `client.project.*` calls remain on `custom-instructions.mdx` (~6 sites). After this PR the platform docs are inconsistent (one page flat methods, siblings still on the non-existent namespace). Worth folding in a one-line sweep so they're uniform.

No `llms.txt` change needed (no pages added/removed) — correctly omitted.

karthik-indla (approved):

PR #6906, opened 2026-08-11
docs: document the real search filter grammar

kartik-mem0:
## Linked Issue

No issue exists for this. This PR addresses item 30 from the July 31 bug bash (reported by Harsh): "The search docs do not make the filter syntax clear. This compounds bug 12(c): the docs show a flat filter shape the API rejects, and only the undocumented `{"AND":[...]}` wrapper works. Fix: document the real filter grammar with working examples for every operator, on one page."

## Description

The two canonical filter docs pages (`docs/platform/features/v2-memory-filters.mdx` for the hosted Platform API and `docs/open-source/features/metadata-filtering.mdx` for the self-hosted SDK) did not clearly show that the Platform API rejects a flat, multi-key filter object and requires an explicit `AND`/`OR`/`NOT` wrapper. The OSS page also contained a materially inaccurate vector-store support claim.

Every example and claim in this PR was cross-checked against the actual filter-handling code (`mem0/client/main.py`, `mem0/memory/main.py`, and the vector store filter-translation layers in `mem0/vector_stores/qdrant.py`, `pgvector.py`, `chroma.py`, `pinecone.py`, `weaviate.py`) rather than written from memory.

Platform page (`v2-memory-filters.mdx`):
- Added a worked example showing the exact flat, multi-key shape the API rejects, next to the correct `AND`-wrapped form.
- Added a callout clarifying that this wrapper requirement is Platform-specific: the OSS SDK is more forgiving because sibling top-level keys are implicitly ANDed there.
- Added a callout clarifying that `nin` is not a Platform operator (it's OSS-only), with the `NOT` + `in` workaround for exclude-list semantics on Platform.

OSS page (`metadata-filtering.mdx`):
- Added a cross-link callout to the Platform page explaining the grammar difference up front.
- Corrected the wildcard (`*`) operator's description and added a warning that its semantics vary by vector store: it's a real "field exists" check on pgvector, a no-op (matches everything, including missing fields) on Qdrant and Chroma, and a literal string match (not a real wildcard) on Pinecone.
- Added a warning that `AND`/`OR`/`NOT` wrapper support in `get_all()` depends on the vector store, because the wrapper is only resolved into vector-store-native syntax inside `search()`'s pre-processing, not `get_all()`.
- Replaced the "Vector store support" `AccordionGroup`, which claimed Weaviate has "Full operator coverage with advanced text filters," with a table verified directly against each provider's filter-translation code. Weaviate's `list()`/`search()`/`keyword_search()` only implement plain equality via `Filter.by_property(...).equal(...)`; there is no comparison, list, substring, logical, or wildcard support in that provider at all. The old claim was wrong and has been removed.
- Added a cross-link card back to the Platform page.

Both pages already existed and were already registered in `docs/docs.json` and `docs/llms.txt`, so no new page was created and no registration changes were needed. `python scripts/check-llms-txt-coverage.py` passes.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

Docs-only change, no executable code was modified. Verification consisted of cross-referencing every operator and behavior claim against the actual filter-handling implementation in `mem0/client/main.py`, `mem0/memory/main.py`, and the five vector store provider files listed above, plus running `python scripts/check-llms-txt-coverage.py` to confirm the docs index stays in sync.

## Checklist


karthik-indla:
**Review — strong, code-grounded OSS grammar; one Platform claim to re-verify**

The OSS side is excellent and I verified each store-specific claim against the filter-translation code: Weaviate (only user_id/agent_id/run_id exact-equality, everything else dropped), pgvector `*` = key-exists, Qdrant/Chroma `*` = no-op/dropped, Pinecone literal-match, and the `AND`/`OR`/`NOT` translates-in-`search()`-not-`get_all()` structural claim (raw wrapper on pgvector `get_all()` silently matches nothing). All accurate.

Two things on the Platform side:

1. **(please re-verify before merge) The "flat multi-key filter is rejected / a wrapper is required for more than one condition" claim looks overstated.** Against the current hosted v2 filter validation, a flat filter with multiple *allowed* top-level keys is accepted — each key is validated independently and the siblings are implicitly AND-ed; only *unknown* keys are rejected. So `{"user_id": "alice", "categories": {"in": ["work"]}}` (two allowed keys, no wrapper) validates fine. The explicit `AND`/`OR`/`NOT` wrapper is needed for OR/NOT semantics, not merely because there's more than one condition. If production genuinely rejects flat multi-key, please keep it but note that's stricter than the documented v2 validation; otherwise soften the "rejected"/"required" phrasing (the `# Rejected: multiple sibling keys` example especially).
2. **(confirmed correct) "No `nin` on Platform"** — verified: the hosted v2 comparison-operator set is `in, gte, lte, gt, lt, ne, contains, icontains`; `nin` is OSS-only. 

Minor: the OSS table's pgvector `in`/`nin` cell says a non-list "raises `ValueError`" — that guarantee holds for Qdrant but not pgvector (a non-list there raises `TypeError` or mis-iterates a string). Suggest "should be a list; a non-list value errors or misbehaves depending on store."

karthik-indla:
**Re-review — the main claim is fixed; one small cell still off**

Thanks — the central correction landed well. The "flat multi-key is rejected / wrapper required" framing is gone; the page now correctly says sibling top-level keys are implicitly AND-ed and reserves `AND`/`OR`/`NOT` for OR/NOT/nesting, with the only rejection example being an unknown top-level key. That matches the hosted v2 validation. The `nin`-is-OSS-only callout (with the `{"NOT": {…: {"in": […]}}}` workaround) is correct too.

One residual, not resolved (and now restated in a still-incorrect form): the OSS table's pgvector cell says `in`/`nin` "must be a list, or the call raises `ValueError`." pgvector doesn't raise `ValueError` — it does `[str(v) for v in op_value]` with no validation, so a bare string is silently iterated character-by-character (wrong results, no error) and a number raises `TypeError`. `ValueError` is Qdrant's behavior. Suggest: "`in`/`nin` expect a list; a bare string is silently iterated per-character and a number raises `TypeError`." One-cell fix and this is good to go.

karthik-indla (approved):

PR #6908, opened 2026-08-11
docs: ground the platform/OSS comparison and memory-type status in reality

kartik-mem0:
## Linked Issue

No linked issue. This covers July 31 bug bash items 62 and 31 (no GitHub issue was filed for either).

## Description

Fixes two related "positioning-page" defects flagged in the July 31 bug bash:

**Bug 62 (Deshraj)**: `docs/platform/platform-vs-oss.mdx` was stale and didn't make a real case for the Platform. Several comparison rows no longer matched the code: Graph Memory was listed as available on OSS via an external store (removed when the v3 pipeline landed), Memory Filters v2 implied a Platform-only advantage that OSS no longer has (both SDKs now support the full operator set), Custom Categories was marked "Limited" on OSS when it's actually unsupported, Support was presented as a Platform perk when the same channels are used on both, Analytics claimed a "Built-in" Platform feature that doesn't exist as a shipped product, and the provider counts were out of date.

Rebuilt the page around real, current deltas: hosted infra, `app_id` + org/project/member scoping, the v3-only ranking features (Graph Memory, Memory Decay, Temporal Reasoning, Dream), Custom Categories, Webhooks, and Memory Export. Every remaining row links to the specific feature doc or code path it's based on. Rows I could not substantiate were removed rather than kept as marketing copy (see below).

**Bug 31 (Younes)**: `docs/core-concepts/memory-types.mdx` described a four-tier "Conversation / Session / User / Organizational memory" hierarchy with automatic promotion between tiers. That hierarchy doesn't exist in the code, there is no automatic promotion logic anywhere in `mem0/memory/main.py`; scoping is purely the caller-supplied `user_id` / `agent_id` / `run_id` (and `app_id` on Platform).

Rewrote the page around what's actually real: the `MemoryType` enum has three values, but only `procedural_memory` is implemented (Python OSS only; `add()` validates against it directly and rejects anything else). `semantic_memory` and `episodic_memory` are defined in the enum but never referenced anywhere else in the codebase, they are dead code, not roadmap items (found no evidence of planned work on them). The page now states this plainly in a status table, then explains the real `user_id`/`agent_id`/`run_id`/`app_id` scoping model and the actual v3 extraction pipeline (context gathering, existing-memory retrieval, single LLM extraction call).

Both pages kept their existing Mintlify components only (`CardGroup`/`Card`/`AccordionGroup`/`Accordion` on platform-vs-oss; `Card`/`CardGroup`/`Tip`/`Warning`/`Link` on memory-types); no new component types were introduced.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A

## Test Coverage

Docs-only change (`.mdx` content), no code or public API behavior was modified. Verified with `python scripts/check-llms-txt-coverage.py` (passes) and manually checked every internal link referenced from both pages resolves to a real file in this repo.

## Checklist


karthik-indla:
**Review — big accuracy win; a couple of precision fixes + one Platform claim to re-verify**

This is a substantial improvement over the marketing-flavored original, and I verified the code-grounded claims: the `MemoryType` enum dead-code point (only `procedural_memory` is wired into `add()`; semantic/episodic are defined but unused), the OSS "not supported" error strings for decay/temporal/custom-categories/project-updates, `nin` in OSS filters, and "at least one identifier required." All accurate.

Items to address:

1. **(please re-verify) The "combining more than one condition requires an explicit AND/OR/NOT wrapper" row is overstated.** Against the current hosted v2 filter validation, a flat filter with multiple *allowed* top-level keys is accepted (validated per-key, implicitly AND-ed) — only unknown keys are rejected. The wrapper is needed for OR/NOT semantics, not merely for having more than one condition. Please confirm against production; if it doesn't actually reject flat multi-key, soften this row. (The companion claim on the same table — **no `nin` on Platform** — is correct; verified the hosted comparison-op set has no `nin`.)
2. **(should fix) Self-contradicting sentence in `memory-types.mdx`.** The procedural section says omitting `memory_type` *"or setting it to anything other than `procedural_memory`"* stores an ordinary memory — but any non-null value other than `procedural_memory` is rejected by validation, which contradicts the page's own Warning box. Tighten to: omitting it stores an ordinary memory; any other explicit value is rejected.
3. **(minor) Stale `llms.txt` description** for Memory Types still says "explaining working, factual, episodic, and semantic memory distinctions" — the opposite of the rewritten page's thesis. Update it.
4. **(minor) Exception-type precision:** the Warning says `Mem0ValidationError`, which is right for sync `Memory`, but `AsyncMemory.add()` raises a plain `ValueError`.

kartik-mem0:
Thanks for the review, this was a good catch list. All four items are addressed in the two commits above. Item 1 turned out to be right, and chasing it down surfaced two further inaccuracies at the source.

### 1. Flat multi-key v2 filters (re-verified against production)

You were right, and the row was wrong. I ran it against `api.mem0.ai` rather than reasoning from the validator:

```
POST /v2/memories/  {"filters": {"user_id": "...", "agent_id": "..."}}   -> 200
POST /v2/memories/  {"filters": {"user_id": "...", "nonsense_key": "x"}} -> 400
```

A flat filter with several allowed top-level keys is accepted and implicitly ANDed; only unknown keys are rejected. So the wrapper is needed for OR/NOT semantics and nested grouping, not for having more than one condition, exactly as you described.

The fix is not only in this page. The claim originates in `platform/features/v2-memory-filters.mdx`, which said the same thing in two places (the "Combining conditions" intro and the FAQ entry "Do I need to use AND/OR/NOT?"). Both are corrected there too, so the comparison page and its source page now agree.

What replaces the row: filter *grouping* moved up into "What's the same", since `AND`/`OR`/`NOT`, flat implicit AND, and `*` wildcards all behave the same on both. The genuine differences are narrower, so they are now two precise rows instead of one overstated one:

- **Filterable fields** - Platform validates top-level keys against a fixed allowlist and rejects anything else with a `400`; OSS lets you filter any metadata key directly.
- **Comparison operators** - Platform's operator set is per-field (`metadata` takes `eq`/`ne`/`contains`, `categories` takes `contains`/`in`, dates take the range operators); OSS accepts the full set on any key. Your `nin` point survives inside this row, it is in the OSS list and in none of the Platform per-field lists.

### 2. Second bug found while verifying item 1

`v2-memory-filters.mdx` listed `eq` and `ne` as valid operators on `categories`. Production rejects both with a `400`; only `in` and `contains` are accepted. Corrected in the same commit.

### 3. Self-contradicting sentence in `memory-types.mdx`

Fixed as you suggested. It now reads: omitting `memory_type` stores an ordinary memory, and any other explicit value is rejected by validation rather than quietly falling back.

### 4. Stale `llms.txt` description

Updated to match the rewritten page's thesis (`procedural_memory` implemented, `semantic_memory` / `episodic_memory` defined in the enum but rejected). `python3 scripts/check-llms-txt-coverage.py` reports in sync.

### 5. Exception-type precision

Corrected. The Warning now names both: sync `Memory.add()` raises `Mem0ValidationError` (`mem0/memory/main.py:826-830`), `AsyncMemory.add()` raises a plain `ValueError` (`mem0/memory/main.py:2475-2478`).

### Also added: three missing Platform-only rows

While auditing the client surface for anything the table omitted, three capabilities exist on `MemoryClient` with no OSS equivalent and were not mentioned: `batch_update`/`batch_delete`, `feedback`, and `get_summary`. Added to the "Configuration & data operations" accordion.

Deliberately not added: Direct Import. `infer=False` exists on OSS `Memory.add()` (`mem0/memory/main.py:765`), so it is not a Platform difference.

### Three more inaccuracies, found by probing rather than reading

Verifying your item 1 properly meant probing every field and operator the pages assert, and that turned up three more, two of them mine:

**The allowlist I wrote was the wrong endpoint's.** `/v2/memories/search/` and `/v2/memories/` do not share a field list:

```
POST /v2/memories/        {"filters": {"user_id": "...", "zzz": "x"}}
  -> 400  allowed: AND, OR, NOT, user_id, agent_id, app_id, run_id, created_at,
          updated_at, timestamp, expiration_date, categories, metadata, keywords,
          feedback, feedback_reason, memory_ids

POST /v2/memories/search/ {"query": "hi", "filters": {"user_id": "...", "zzz": "x"}}
  -> 400  ... expiration_date, text, categories, metadata, keywords_search,
          memory_ids, keywords
```

`text` and `keywords_search` are search-only, `feedback` and `feedback_reason` are get_all-only. The row now lists the thirteen keys common to both and says there are endpoint-specific extras, instead of presenting one endpoint's list as the whole allowlist. (`keywords_search` clears validation but then fails downstream with `Unsupported field: keywords_search`, so it is not worth documenting as filterable.)

**Every date field rejects explicit `eq`.** `v2-memory-filters.mdx` listed `eq` on all three time fields:

```
{"created_at": {"eq": "2026-01-01"}}
  -> 400  Invalid operator: eq. Allowed operators are ['in','gte','lte','gt','lt','ne','contains','icontains']
{"created_at": "2026-01-01"}   -> 200
```

Same for `updated_at` and `timestamp`. Corrected, and `expiration_date` added to that table since it is allowlisted, behaves identically, and was missing.

**The nested AND/OR example was broken.** The "Can I nest AND/OR?" FAQ filtered categories with a bare string:

```
{"AND":[{"user_id":"..."},{"OR":[{"categories":"finance"},{"categories":"health"}]}]}
  -> 400  'categories' value must be a dictionary.
```

It contradicted the `categories` row two sections above it in the same file. Rewritten to use `{"in": [...]}`.

### One correction to my own OSS claim

"The full set on any key" was unqualified, and that is only true of the SDK's filter layer. It is not true end to end: `mem0/vector_stores/chroma.py:280-283` falls back to equality for `contains`/`icontains` because Chroma has no substring match, and `mem0/vector_stores/pinecone.py:189-197` does not map them at all. Scoped the claim to the filter layer and pointed at `/open-source/features/metadata-filtering`, which already documents the per-store variance.

### Verified and deliberately left alone

The 25 vector stores / 18 LLM providers / 11 embedders counts, the OSS graph-memory removal note, and the project-scoped wording on webhooks (the client calls `api/v1/webhooks/projects/{project_id}/`).

One thing I could not settle: the `metadata` operator set. The validator returns `200` for `gt`, `nin`, and `icontains` on `metadata`, which is wider than what this page and its "Known limitations" section claim. Every probe account I have is empty, so a `200` there only proves the request validated, not that the operator filters anything. I left the existing narrower claim (`eq`, `ne`, `contains`) rather than widen it on evidence that thin. Worth a look from someone who can query seeded data.

karthik-indla:
**Re-review — all findings addressed, LGTM**

All five prior points are resolved and verified:
- The "Filter grouping grammar" row now correctly states several top-level keys in one flat filter are implicitly AND-ed, with the wrapper only needed for OR/NOT/nesting — matches the hosted v2 validation.
- The procedural `memory_type` sentence now reads "Any other explicit value is rejected by validation" — consistent with the page's own Warning box and the sync/async code.
- The `llms.txt` description is updated to reflect the real enum status (procedural implemented; semantic/episodic defined but rejected).
- The exception note now correctly distinguishes sync `Mem0ValidationError` from async `ValueError`.
- "No `nin` on Platform" confirmed correct.

Nice bonus: expanding scope to fix the v2-memory-filters date/`categories` operator rows — those check out against the validator too. No remaining issues from my side.

karthik-indla (approved):
Re-reviewed: all prior findings resolved and verified against the code (filter-grouping grammar, procedural memory_type sentence, llms.txt description, sync/async exception types, no-nin-on-Platform). The v2-memory-filters operator-row additions check out too. LGTM.

PR #6917, opened 2026-08-11
fix(embeddings/huggingface): resolve api_key from HUGGINGFACE_API_KEY and default it

### LINKED ISSUE #6301: bug(embeddings/huggingface): api_key ignored on base_url path, endpoint auth dropped

OfficialAbhinavSingh:
Closes #6301.

#6770 forwarded `api_key` to the `OpenAI` client on the `huggingface_base_url` path, which fixed the first failure mode in #6301. The second one is still live on `main`: the resolve stops at the config value, so `api_key` is `None` whenever the caller does not pass one explicitly, and the OpenAI client then resolves `OPENAI_API_KEY` and raises when that is unset.

Two consequences on current `main`:

1. `HUGGINGFACE_API_KEY` is a no-op in Python. Setting it and no config key still raises.
2. An unauthenticated local TEI endpoint cannot be constructed at all.

## Repro on current main

```python
import os
os.environ["HUGGINGFACE_API_KEY"] = "hf_fromEnvToken"

HuggingFaceEmbedding(BaseEmbedderConfig(huggingface_base_url="http://localhost:8080"))
```

```
openai.OpenAIError: Missing credentials. Please pass an `api_key`, `workload_identity`,
`admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
```

Same error with nothing set at all, which is the no-auth TEI case.

## Fix

```python
self.config.api_key = self.config.api_key or os.getenv("HUGGINGFACE_API_KEY") or "hf"
```

The three-step resolve and the `"hf"` placeholder are not invented here, they are what the rest of the codebase already does:

- `mem0/embeddings/lmstudio.py:15` does `self.config.api_key = self.config.api_key or "lm-studio"` for the same OpenAI-compatible-endpoint situation, including the same assignment back onto the config.
- `mem0-ts/src/oss/src/embeddings/huggingface.ts:39` does `apiKey: config.apiKey || process.env.HUGGINGFACE_API_KEY || "hf"`.
- The TypeScript tab of `docs/components/embedders/models/huggingface.mdx` already documents `apiKey` falling back to `HUGGINGFACE_API_KEY` with a `"hf"` default. The Python tab had no `api_key` row at all, so this PR adds one.

## Note on the changed assertion

`test_embed_with_huggingface_base_url` asserted `api_key=None`, which pins the behaviour this PR fixes, so it is updated to `api_key="hf"`. Both pre-existing base_url tests also gained a `monkeypatch.delenv("HUGGINGFACE_API_KEY")` so they do not depend on the ambient environment.

## Tests

Four new tests: env fallback, config wins over env, placeholder default, and one end-to-end guard that constructs against the real `OpenAI` client rather than a patched one and asserts a no-auth endpoint builds.

Verified red then green. With the source change reverted and the tests kept:

```
FAILED tests/embeddings/test_huggingface_embeddings.py::test_embed_with_huggingface_base_url
FAILED tests/embeddings/test_huggingface_embeddings.py::test_base_url_api_key_falls_back_to_env
FAILED tests/embeddings/test_huggingface_embeddings.py::test_base_url_api_key_defaults_to_placeholder
FAILED tests/embeddings/test_huggingface_embeddings.py::test_base_url_no_auth_endpoint_constructs_without_credentials
4 failed, 12 passed
```

With the fix applied:

```
16 passed
```

Suite also passes with `HUGGINGFACE_API_KEY` set in the ambient environment, confirming the new tests are not env-sensitive. `ruff check` is clean.

## History

#6301 drew four PRs: #6302, #6315, #6486 and #6755. All four independently included this same `config or HUGGINGFACE_API_KEY or "hf"` resolve, so this part is not a preference of mine. #6770 landed the first step of it and the rest was never picked up. I have closed my #6302 rather than rebasing it, since only this one line of it is still relevant and a fresh minimal diff is easier to review.

### DIFF
--- a/docs/components/embedders/models/huggingface.mdx
+++ b/docs/components/embedders/models/huggingface.mdx
@@ -99,6 +99,7 @@ Here are the parameters available for configuring the Hugging Face embedder:
 | `embedding_dims` | Dimensions of the embedding model | `selected_model_dimensions` |
 | `model_kwargs` | Additional arguments for the model | `None` |
 | `huggingface_base_url` | URL to connect to Text Embeddings Inference (TEI) API | `None` |
+| `api_key` | API key for the endpoint; falls back to the `HUGGINGFACE_API_KEY` env var. Only used on the `huggingface_base_url` path | `"hf"` |
 </Tab>
 <Tab title="TypeScript">
 | Parameter | Description | Default Value |
--- a/mem0/embeddings/huggingface.py
+++ b/mem0/embeddings/huggingface.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Literal, Optional
 
 from openai import OpenAI
@@ -17,6 +18,10 @@ class HuggingFaceEmbedding(EmbeddingBase):
         super().__init__(config)
 
         if self.config.huggingface_base_url:
+            # Without a placeholder the OpenAI client falls back to OPENAI_API_KEY and raises
+            # when that is unset, so an unauthenticated TEI endpoint cannot be constructed at
+            # all. Mirrors mem0/embeddings/lmstudio.py and the Node SDK's huggingface.ts.
+            self.config.api_key = self.config.api_key or os.getenv("HUGGINGFACE_API_KEY") or "hf"
             self.client = OpenAI(base_url=self.config.huggingface_base_url, api_key=self.config.api_key)
             self.config.model = self.config.model or "tei"
         else:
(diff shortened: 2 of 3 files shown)


OfficialAbhinavSingh:
Closing this. #6947 landed the same resolution on main:

```python
```

along with all four test cases this PR added (`..._forwards_api_key`, `..._falls_back_to_env_var`, `..._config_key_beats_env_var`, and the `"hf"` default), so there is nothing left in the code here. The only delta remaining was assigning back onto `self.config.api_key` instead of a local, and the local is the better shape.

One thing did not land: the config table documents `apiKey` on the TypeScript tab but has no `api_key` row on the Python tab, even though the behaviour now exists in both. Opening that as a docs-only follow-up.

Thanks @kartik-mem0.

PR #6932, opened 2026-08-13
fix: compare expiration date against local calendar day, not UTC (#6931)

### LINKED ISSUE #6931: bug(sdk-python): expiration check compares naive local date to UTC date, hiding memories early/late by UTC offset
### Component

Python SDK

### Summary

`expiration_date` is stored and provided as a **naive local date** (`YYYY-MM-DD`,
no timezone), but the expired check compares it against the **UTC** calendar date.
The decision therefore drifts by the user's UTC offset around every date boundary.

In `mem0/memory/main.py`:

```python
# _normalize_expiration_date (line 427) stores a NAIVE date:
return value.date().isoformat()          # "2026-08-13" — tz dropped

# _payload_is_expired (line 442) compares it to UTC:
return date.fromisoformat(str(expiration_date)) < datetime.now(timezone.utc).date()
```

The semantics of the `<` is "still valid on the expiration date itself, expired
after". The problem is that "after" is measured on the **UTC** clock, while the
stored date is the **user's local** calendar date.

`created_at` / `updated_at` are correctly written as UTC instants (this is right for
*timestamps*), but `expiration_date` is a *calendar date*, not an instant, so it has
no business being compared to UTC.

### Steps to Reproduce

No LLM or vector store needed — the helper is pure date logic. Verified against the
installed `mem0ai==2.0.18` package:

```python
from unittest.mock import patch
from mem0.memory.main import _payload_is_expired

def check(label, utc_now, exp_date, expected):
    payload = {"expiration_date": exp_date}
    with patch("mem0.memory.main.datetime") as mock_dt:
        mock_dt.now.return_value = utc_now
        got = _payload_is_expired(payload)
        print(f"{label}: got={got} expected={expected}")

# Direction A: UTC+8 user, local 08-14 00:30 (13th already over), UTC still 08-13
check("UTC+8 (past local, not past UTC)", datetime(2026,8,13,16,30,tzinfo=timezone.utc), "2026-08-13", True)

# Direction B: UTC-8 user, local 08-13 23:30 (13th still ongoing), UTC already 08-14
check("UTC-8 (not past local, past UTC)",  datetime(2026,8,14,7,30,tzinfo=timezone.utc),  "2026-08-13", False)
```

**Actual output (mem0ai 2.0.18):**

```
UTC+8 (past local, not past UTC): got=False expected=True
UTC-8 (not past local, past UTC): got=True expected=False
```

### Expected Behavior

Expiration is decided on the **user's local calendar date** (or a configurable
timezone). A memory with `expiration_date="2026-08-13"` hides starting the user's
local 2026-08-14 00:00 — not 08:00 (UTC+8), and not 16:00 the day before (UTC-8).

| scenario | local | UTC | actual | expected |
|---|---|---|---|---|
| UTC+8, past local day | 08-14 00:30 | 08-13 16:30 | `False` | `True` |
| UTC-8, still in local day | 08-13 23:30 | 08-14 07:30 | `True` | `False` |

### Actual Behavior

- **UTC+ users**: memories linger up to `offset` hours past their local expiry
  (delayed hiding).
- **UTC− users**: memories vanish up to `offset` hours before their local expiry
  (premature hiding).

The drift scales linearly with the user's UTC offset.

### Proposed Fix

Make "today" a single, consistent, documented notion and use it in
`_payload_is_expired`. Two consistent options:

1. **Compare against local date** (simplest, matches the stored naive value):
   ```python
   return date.fromisoformat(str(expiration_date)) < datetime.now().date()
   ```
2. **Normalize at input** to a timezone-aware UTC instant in
   `_normalize_expiration_date` (keep `_payload_is_expired` as-is). Pick the user's
   local midnight and convert to UTC.

Either is correct; the current mix ("naive local date in, UTC date compared") is
internally inconsistent.

### Environment

- mem0 version: `2.0.18` (current `main` behaves the same)
- Python: 3.13
- Reproducible on any OS; effect scales with UTC offset

icn5381:
Fixes #6931

## Summary

`_normalize_expiration_date` stores the caller's wall-clock calendar day (`value.date()` drops any timezone), but `_payload_is_expired` compared that stored date against the **UTC** calendar day. Around every date boundary the decision drifted by the local UTC offset: a memory could stay visible up to one day too long, or expire up to one day early — e.g. with a UTC+8 host, a memory expiring on Aug 13 was still visible until Aug 14 08:00 local.

## Change

- `_payload_is_expired` now compares against `datetime.now().date()` (local calendar day) instead of `datetime.now(timezone.utc).date()`.
- Rationale for this option (the issue also floated normalizing to UTC or a configurable timezone):
  - The stored value is already a **calendar day in the caller's frame** — `value.date()` on an aware datetime keeps the caller's wall-clock day and discards the zone. That information is gone at storage time.
  - Re-normalizing to UTC is therefore impossible without guessing the caller's timezone; a configurable timezone is a new knob with no consumer.
  - Comparing in the local calendar day makes both sides of the comparison use the same frame and preserves the documented semantics: *still valid on the expiration date itself, expired after*.

## Testing

New `tests/memory/test_expiration_tz.py`:

- red-before/green-after regression: local 2026-08-14 00:30 (+08:00, i.e. UTC still 08-13) with expiration 2026-08-13 must be expired — fails on the old UTC comparison, passes now
- still valid on the expiration day itself
- no-expiration payloads
- `_normalize_expiration_date` keeps the caller's calendar day for aware datetimes, and still accepts `date` / ISO strings / `None`

Validated with `pytest tests/memory/test_expiration_tz.py` (5 passed), `pytest tests/memory/test_main.py` (52 passed), and `ruff check` on both changed files.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** AI-assisted coding. It located the two helpers, and drafted the fix and the regression tests; I reviewed, chose between the timezone-semantics options described in the issue, and finalized every line.

Human verification performed for this change:

- Read and understood every line of the diff (2 files, 82 insertions, 1 deletion)
- Reproduced the drift reasoning by hand before coding (UTC+8 host, Aug-13 expiration stays visible until Aug-14 08:00 local)
- Ran the regression test against the old code and watched it fail, then watched it pass after the one-line comparison change
- Ran the neighboring test module and ruff on both files
- Verified the diff touches nothing beyond the comparison line, a clarifying comment, and the new test module

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -446,7 +446,11 @@ def _payload_is_expired(payload: Optional[Dict[str, Any]]) -> bool:
     if not expiration_date:
         return False
     try:
-        return date.fromisoformat(str(expiration_date)) < datetime.now(timezone.utc).date()
+        # ``_normalize_expiration_date`` stores the caller's wall-clock calendar
+        # day (``value.date()`` strips any timezone). Compare against the local
+        # calendar day for symmetry: comparing with the UTC day drifted the
+        # decision by the local UTC offset around every date boundary (#6931).
+        return date.fromisoformat(str(expiration_date)) < datetime.now().date()
     except ValueError:
         return False
 
--- a/tests/memory/test_expiration_tz.py
+++ b/tests/memory/test_expiration_tz.py
@@ -1,0 +1,77 @@
+"""Regression tests for expiration-date timezone handling (#6931).
+
+The stored expiration value is a calendar date taken from the caller's
+wall-clock (``value.date()``), but the expiry check compared it against the
+UTC calendar day. Around every date boundary the decision drifted by the
+local UTC offset: a memory could stay visible up to one day too long, or
+expire up to one day early.
+"""
+
+from datetime import date, datetime, timedelta, timezone
+from unittest.mock import patch
+
+from mem0.memory.main import _normalize_expiration_date, _payload_is_expired
+
+_UTC_PLUS_8 = timezone(timedelta(hours=8))
+
+
+class _FrozenDatetime(datetime):
+    """datetime subclass whose ``now()`` returns a frozen wall-clock instant.
+
+    ``now(tz=...)`` converts the frozen local instant to the requested zone,
+    mirroring the stdlib's behavior of ``datetime.now(timezone.utc)``.
+    """
+
+    frozen_local: datetime | None = None
+
+    @classmethod
+    def set_now(cls, local_dt: datetime) -> None:
+        cls.frozen_local = local_dt
+
+    @classmethod
+    def now(cls, tz=None):
+        if cls.frozen_local is None:
+            return datetime.now(tz)
+        if tz is not None:
+            return cls.frozen_local.astimezone(tz)
+        return cls.frozen_local
+
+
+def _freeze(local_dt: datetime):
+    _FrozenDatetime.set_now(local_dt)
+    return patch("mem0.memory.main.datetime", _FrozenDatetime)
+
+
+def test_expired_uses_local_calendar_day_not_utc():
+    """Local date has advanced past the expiration while UTC has not.
+
+    Local time is 2026-08-14 00:30 (+08:00) == 2026-08-13T16:30Z. The memory
+    expires on 2026-08-13, so it must be considered expired — but comparing
+    against the UTC calendar day (still 08-13) kept it visible for another
+    7.5 hours.
+    """
+    with _freeze(datetime(2026, 8, 14, 0, 30, tzinfo=_UTC_PLUS_8)):
+        assert _payload_is_expired({"expiration_date": "2026-08-13"})
+
+
+def test_not_expired_on_expiration_day():
+    """Still valid on the expiration date itself (local calendar day)."""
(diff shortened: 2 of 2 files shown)


786262170 (commented):
Independent verification of this fix (as the reporter of #6931):

I reproduced the original bug against the installed `mem0ai==2.0.18` package before this PR, and confirm the same root cause. I then applied your exact one-line change and verified it against **argument-sensitive** mock (local "now" vs UTC "now" differ), which is the key detail that distinguishes the fixed code from the old code:

- Old `datetime.now(timezone.utc).date()` with UTC+8 host at local 08-14 00:30 (UTC still 08-13) → `expiration_date="2026-08-13"` returns `False` (memory lingers ~7.5h past expiry) — bug.
- New `datetime.now().date()` with same host → `True` (correct).

Your `_FrozenDatetime.now()` handling of the `tz` argument (converting the frozen local instant via `astimezone`) is the right way to test this — a single fixed return value would not discriminate the two code paths, since `.date()` just extracts the datetime's own calendar day.

Local run of your `tests/memory/test_expiration_tz.py` + the neighboring suite passes here too. Looks correct — thanks for picking this up.

PR #6935, opened 2026-08-13
fix(milvus): convert COSINE distance to similarity in _parse_output

### LINKED ISSUE #6933: bug(sdk-python): MilvusVectorStore returns raw distance as score for COSINE metric — ranking inverted, best match dropped by default threshold
### Component

Python SDK / vector store (Milvus)

### Summary

`MilvusVectorStore._parse_output` returns the **raw Milvus distance** as the score for every metric except L2. For `COSINE` (the metric mem0's own test fixtures use, and the most common choice for text embeddings), Milvus returns **cosine distance** = `1 - cosine_similarity` in `[0, 2]`, where **lower is better**. `_parse_output` passes it through unchanged:

```python
# mem0/vector_stores/milvus.py:195-198
if raw_distance is not None and self.metric_type in (MetricType.L2, "L2"):
    score = 1.0 / (1.0 + raw_distance)
else:
    score = raw_distance
```

This violates the `VectorStoreBase` contract (`base.py:19-23`):

> All implementations must return similarity scores where higher values ... using distance metrics must convert to similarity before returning: **Cosine distance: `score = max(0.0, 1.0 - distance)`**

`pgvector.py:367` implements exactly this fix for cosine; **Milvus is the one store that does not.**

### Impact (confirmed by tracing downstream)

Because `score` is inverted and unbounded, the hybrid scoring in `score_and_rank` (`mem0/utils/scoring.py:110-112`) misbehaves:

1. **Best match dropped by the default threshold.** `score_and_rank` gates on `semantic_score < threshold` (search default `threshold=0.1`). For COSINE, the *closest* vector has distance ≈ 0 → score ≈ 0 → **below 0.1 → excluded**. The **least-similar** vectors (distance ≈ 2) score highest and are returned first.
2. **Ranking inverted.** `scored.sort(key=lambda x: x["score"], reverse=True)` sorts descending, so the returned order is the *worst* matches first.
3. **Entity-boost gate broken.** `_compute_entity_boosts` (`main.py:1792-1794`) gates on `similarity < 0.5`; real entity matches (distance ≈ 0) are skipped because their score is ≈ 0.

In short: with a COSINE-configured Milvus store, `search()` returns wrong results (best match dropped, worst ranked first) and entity boosting silently disables.

### Steps to Reproduce

Configure a Milvus store with `metric_type="COSINE"` (mem0's own tests do this — `test_milvus.py:39/50/363`), insert a memory whose text matches the query exactly, then:

```python
memory.search("the exact memory text", filters={"user_id": "u1"}, threshold=0.1)
```

Expected: the exact match is returned first with a high score.
Actual: the exact match has `score ≈ 0` (cosine distance ≈ 0), so it is **dropped** by the threshold, and whatever is returned is the least-similar vectors.

### Expected Behavior

For COSINE (and any other distance metric), `_parse_output` should convert distance to similarity per the contract before returning — e.g. `score = max(0.0, 1.0 - raw_distance)` for cosine, mirroring `pgvector.py:367`.

### Additional note

The same `_parse_output` is reused by `keyword_search` (Milvus BM25 hits), where the raw score is **higher-is-better**; on L2 configs it would be squashed to `1/(1+score)`. A proper fix should branch on the metric and/or search type.

### Environment

- mem0 version: current `main`
- Vector store: Milvus (COSINE metric)
- Reproducible with any embedding model

786262170:
Fixes #6933

## Summary

`MilvusVectorStore._parse_output` returned the **raw Milvus distance** as the score for every metric except L2. For `COSINE` (the default and most common metric for text embeddings), Milvus returns cosine **distance** = `1 - cosine_similarity` in `[0, 2]` where **lower is better**. Passing it through unchanged:

1. **Inverted the ranking** — `score_and_rank` sorts descending, so the least-similar vectors ranked first.
2. **Dropped the best match** — the closest vector (distance ≈ 0 → score ≈ 0) fell below the default `threshold=0.1` and was excluded.
3. **Broke entity boosting** — `_compute_entity_boosts` gates on `similarity < 0.5`, so real entity matches (distance ≈ 0) were skipped.

This violated the `VectorStoreBase` contract (`base.py:19-23`), which requires `cosine distance → score = max(0.0, 1.0 - distance)`. `pgvector.py:367` already implements this conversion; **Milvus was the one store that did not**.

## Change

`mem0/vector_stores/milvus.py` — `_parse_output` now:
- Converts **COSINE** distance to similarity (`max(0.0, 1.0 - distance)`), per the base contract.
- Keeps **L2** conversion (`1 / (1 + distance)`) unchanged.
- Passes through **IP** (already higher-is-better) and HAMMING/JACCARD (no defined contract conversion) unchanged.
- Adds an `is_keyword` flag so **BM25 keyword hits** keep their higher-is-better score instead of being distance-converted.

## Testing

`tests/vector_stores/test_milvus.py`:
- Fixed two assertions that had codified the buggy behavior (COSINE distance `0.8`/`0.9` were asserted as the score; now they assert the converted similarity `0.2`/`0.1`).
- Added `test_cosine_distance_converted_to_similarity` — verifies distance 0 (best match) → score 1.0, distance ≥1 → clamped to 0.
- Added `test_keyword_search_keeps_raw_bm25_score` — verifies BM25 scores pass through unconverted.

All 30 tests in `tests/vector_stores/test_milvus.py` pass locally. `ruff check` passes on both changed files. Verified the regression tests fail against the pre-fix code (red before / green after).

### DIFF
--- a/mem0/vector_stores/milvus.py
+++ b/mem0/vector_stores/milvus.py
@@ -175,12 +175,14 @@ class MilvusDB(VectorStoreBase):
 
         return " and ".join(operands)
 
-    def _parse_output(self, data: list):
+    def _parse_output(self, data: list, is_keyword: bool = False):
         """
         Parse the output data.
 
         Args:
             data (Dict): Output data.
+            is_keyword (bool): Whether this is a keyword (BM25) search result. BM25
+                scores are higher-is-better and must not be distance-converted.
 
         Returns:
             List[OutputData]: Parsed output data.
@@ -192,10 +194,23 @@ class MilvusDB(VectorStoreBase):
             raw_distance = value.get("distance")
             metadata = value.get("entity", {}).get("metadata")
 
-            if raw_distance is not None and self.metric_type in (MetricType.L2, "L2"):
-                score = 1.0 / (1.0 + raw_distance)
-            else:
+            if is_keyword:
+                # BM25 full-text hits: score is higher-is-better already.
                 score = raw_distance
+            elif raw_distance is not None:
+                # Dense vector hits: convert distance to similarity per the
+                # VectorStoreBase contract (base.py:19-23).
+                if self.metric_type in (MetricType.COSINE, "COSINE"):
+                    score = max(0.0, 1.0 - raw_distance)
+                elif self.metric_type in (MetricType.L2, "L2"):
+                    score = 1.0 / (1.0 + raw_distance)
+                else:
+                    # IP returns the inner product (higher = better); HAMMING/JACCARD
+                    # are distance-based but have no defined contract conversion here,
+                    # so pass the raw value through rather than inverting it.
+                    score = raw_distance
+            else:
+                score = None
 
             memory_obj = OutputData(id=uid, score=score, payload=metadata)
             memory.append(memory_obj)
@@ -263,7 +278,7 @@ class MilvusDB(VectorStoreBase):
                 filter=query_filter,
                 output_fields=["*"],
             )
-            result = self._parse_output(data=hits[0])
+            result = self._parse_output(data=hits[0], is_keyword=True)
             return result
         except Exception as e:
             logger.debug(f"Keyword search not available for collection {self.collection_name}: {e}")
--- a/tests/vector_stores/test_milvus.py
+++ b/tests/vector_stores/test_milvus.py
@@ -139,7 +139,8 @@ class TestMilvusDB:
         # Verify results are parsed correctly
         assert len(results) == 1
         assert results[0].id == "mem1"
-        assert results[0].score == 0.8
+        # fixture uses COSINE metric: distance must be converted to similarity
+        assert results[0].score == pytest.approx(0.2)  # max(0, 1 - 0.8)
 
     def test_search_different_user_ids(self, milvus_db, mock_milvus_client):
         """Test that search works with different user_ids (reproduces reported bug)."""
@@ -241,13 +242,14 @@ class TestMilvusDB:
         ]
         
         parsed = milvus_db._parse_output(raw_data)
-        
+
         assert len(parsed) == 2
         assert parsed[0].id == "mem1"
-        assert parsed[0].score == 0.9
+        # fixture uses COSINE metric: distance must be converted to similarity
+        assert parsed[0].score == pytest.approx(0.1)  # max(0, 1 - 0.9)
         assert parsed[0].payload == {"user_id": "alice"}
         assert parsed[1].id == "mem2"
-        assert parsed[1].score == 0.85
+        assert parsed[1].score == pytest.approx(0.15)  # max(0, 1 - 0.85)
 
     def test_update_with_none_vector_fetches_existing(self, milvus_db, mock_milvus_client):
         """Test that update with vector=None fetches the existing vector (fixes #3708)."""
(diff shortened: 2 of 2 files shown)


tonydzi (commented):
Hi, Mycroft here, TonyDzi's synthetic AI co-founder (an agent ran this; disclosed per CONTRIBUTING). Short version: I think this PR fixes a bug Milvus doesn't have, and it creates the bug the issue describes.

**On a real Milvus, `distance` for `COSINE` is already cosine *similarity*, not distance.** With Milvus Lite 3.2.1 / pymilvus 3.0.1, a FLAT index and query `[1,0,0,0]`, the raw hits are:

```
COSINE raw: exact=1.0  close=0.9939  orthogonal=0.0  opposite=-1.0   (server sorts descending)
L2 raw:     exact=0.0  close=0.02    orthogonal=2.0  opposite=4.0
```

The new branch `score = max(0.0, 1.0 - raw_distance)` turns that into `exact=0.0 … opposite=2.0`. The best match gets the lowest score, and `score_and_rank` drops it at the default `threshold=0.1`.

I dropped `mem0/vector_stores/milvus.py` from this PR's head (`63a8980`) into main `c7ee362`. Then I ran a Milvus Lite test that asserts "exact match scores highest and is ≥ 0.1" for COSINE, IP and L2:

```
main:            3 passed
PR head 63a8980: 1 failed, 2 passed   (COSINE: assert 'opposite' == 'exact'; IP and L2 controls pass)
```

It failed the same way in 2/2 runs on the PR head. Main was green before and after.

On the diff itself:

- `tests/vector_stores/test_milvus.py`: the change to `pytest.approx(0.2)  # max(0, 1 - 0.8)` encodes the mock's assumption, not server behaviour. The mocked client can't catch the inversion, which is why the suite stays green.
- The `is_keyword=True` split for BM25 is harmless and could stay on its own. The COSINE branch is the part that changes ranking.
- `else: score = None` for a missing distance is a behaviour change for callers doing arithmetic on `score`. That's worth a note if it's intended.

The full repro (script, 3 byte-identical runs, the test file) is in #6933. If the author saw inverted ranking on a real deployment, the server version and index type would be the first thing to compare. I only ran Lite.

— TonyDzi · repro from running a second brain on a multi-agent lab; more at github.com/tonydzi

kartik-mem0 (commented):
Reviewed at `63a8980b`.

The `is_keyword` fix for BM25 scores is correct and should be kept. BM25 scores are already higher-is-better and were previously not distinguished from vector distance at all.

However, the COSINE branch needs to be reverted. The PR's premise that Milvus returns cosine distance (lower=better) for `metric_type=COSINE` is not correct. Real Milvus returns raw cosine similarity (higher=better, range [-1, 1]):

1. Official Milvus docs (v2.5.x/v2.6.x metric reference): "For COSINE metric type: A greater value indicates a higher similarity. Range [-1, 1]."
2. Milvus maintainer `yhmo` in milvus-io/milvus#34037: the COSINE search field is explicitly a similarity value.
3. This repo's own pre-existing e2e test (`tests/vector_stores/test_score_normalization.py::TestMilvusCosine`): docstring reads "Milvus COSINE already returns similarity. No conversion needed."
4. Confirmed empirically against a real Milvus Docker server.

Ran a live scenario (`oss-milvus` lane, real Docker Milvus, `metric_type=COSINE`):
- `main` (unpatched): PASS, close vector correctly outscores far vector
- PR head: FAIL, far (opposite) vector scores 1.78 vs close vector's 0.21 (both outside [0,1])

The `pgvector.py:367` precedent cited in the PR is not analogous: pgvector's `<=>` operator genuinely returns cosine distance, while Milvus's `COSINE` metric returns similarity directly.

What would unblock: drop the `max(0.0, 1.0 - raw_distance)` branch for COSINE. Either pass the raw value through (as `main` does today) or verify against a real Milvus server before re-asserting. Keep the `is_keyword` BM25 fix.

PR #6936, opened 2026-08-13
fix(mem0-ts): don't insert duplicate entity row when entity lookup fails

### LINKED ISSUE #6925: TS OSS: a failed entity lookup inserts a duplicate entity row instead of linking
### What happened

In the TypeScript OSS SDK, a failed entity lookup is treated as "no matching entity exists", so `add()` inserts a duplicate entity row instead of linking to the row that is already there. Nothing is logged.

Two sites on current main (d70cc00), both in `mem0-ts/src/oss/src/memory/index.ts`:

- `:508` in `_linkEntitiesForMemory`
- `:1203` in the phase 7 batch path of `addToVectorStore`

Both look like this:

```ts
if (!exactMatch) {
  try {
    matches = await entityStore.search(entityVec, 1, filters);
  } catch {}
}
```

When `search()` throws, `matches` stays `[]`, so `semanticMatch` is `undefined`, `match` is `undefined`, and control reaches the `else` branch that builds an `entityPayload` and inserts a new row.

### Why it matters

The entity store is a separate vector store from the memory store, and it can fail on its own: a network blip, a rate limit, a collection that is briefly unavailable. When that happens the graph silently forks. The same entity ends up with two rows holding different `linkedMemoryIds`, and later lookups only find one of them, so the links from the other are effectively lost. There is no warning and no error, so nothing tells you it happened.

This is distinct from #6795, which is about entity failures being logged too quietly. These two sites log nothing at all, so a level change cannot reach them, and the consequence here is a wrong write rather than a missing message.

### Reproduction

Stubbed entity store, identical input both times, the only difference being whether `search()` resolves or throws:

```
healthy store:   { inserts: 0, updates: 2 }   linked to the existing rows
search throws:   { inserts: 2, updates: 0 }   two duplicate rows
```

### Expected

A lookup that failed is not a lookup that found nothing. Warn and skip the entity, leaving the existing row alone, rather than inserting a competing one.

### Version

`mem0-ts` on main at d70cc00. TypeScript OSS SDK only, the Python side does not share this code path.

xiangnuans:
## Linked Issue

Closes #6925

## Description

In the TypeScript OSS SDK, a failed entity-store `search()` was swallowed by an empty `catch {}`, leaving `matches = []`. Control then fell through to the *insert* branch and created a **duplicate entity row**, silently forking the graph: the same entity ends up with two rows holding different `linkedMemoryIds`, and later lookups only find one — so the links from the other are lost, with no warning.

A lookup that *failed* is not a lookup that *found nothing*. This PR treats the two differently: on a failed lookup we **warn and skip the entity**, leaving any existing row untouched, instead of inserting a competing one.

Both sites in `mem0-ts/src/oss/src/memory/index.ts` are fixed:
- `_linkEntitiesForMemory` (single-memory path)
- the phase-7 batch path in `addToVectorStore`

## Type of Change

- [ ] Documentation update

## Breaking Changes

N/A

## Test Coverage

- [ ] No tests needed

New regression test `mem0-ts/src/oss/tests/memory.entity-lookup-failure.test.ts`:
- when `search()` throws → asserts no `insert`/`update` happens and a warning is logged (this test **fails on `main`** without the fix, reproducing the duplicate insert);
- when `search()` legitimately returns `[]` → asserts a new entity is still inserted (guards against over-correcting).

Verified locally with `jest` (this suite + the neighbouring `memory.entity-boost.test.ts` pass); `prettier --check` and `tsc --noEmit` clean.

## Checklist

- [ ] I have updated documentation if needed (internal fix, no docs change)

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -505,7 +505,15 @@ export class Memory {
           if (!exactMatch) {
             try {
               matches = await entityStore.search(entityVec, 1, filters);
-            } catch {}
+            } catch (e) {
+              // A failed lookup is not the same as "no match found". Skip this
+              // entity instead of inserting a competing row that would silently
+              // fork the graph (mem0ai/mem0#6925).
+              console.warn(
+                `Entity lookup failed for '${entity.text}', skipping to avoid a duplicate entity row: ${e}`,
+              );
+              continue;
+            }
           }
 
           const semanticMatch =
@@ -1200,7 +1208,15 @@ export class Memory {
             if (!exactMatch) {
               try {
                 matches = await entityStore.search(entityVec, 1, filters);
-              } catch {}
+              } catch (e) {
+                // A failed lookup is not the same as "no match found". Skip this
+                // entity instead of collecting a competing row for batch insert
+                // that would silently fork the graph (mem0ai/mem0#6925).
+                console.warn(
+                  `Entity lookup failed for '${entityText}', skipping to avoid a duplicate entity row: ${e}`,
+                );
+                continue;
+              }
             }
 
             const semanticMatch =
--- a/mem0-ts/src/oss/tests/memory.entity-lookup-failure.test.ts
+++ b/mem0-ts/src/oss/tests/memory.entity-lookup-failure.test.ts
@@ -1,0 +1,141 @@
+/**
+ * Entity lookup failure handling (#6925).
+ *
+ * When the entity store's semantic `search()` throws (network blip, rate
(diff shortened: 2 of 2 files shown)


shashiKundur1:
Heads up, this is the same change as #6926, which has been open since 12 Aug against the same issue. Same two sites, same approach, warn and skip rather than falling through to the insert.

Flagging it mainly so @kartik-mem0 does not end up reviewing the same fix twice, given the size of the queue at the moment.

One thing worth carrying across whichever one lands: the tests in #6926 include a positive control, a case where `search()` resolves normally and genuinely returns nothing, asserting that the insert still happens. Without it, a change that simply stopped inserting altogether would pass the other two tests just as well.

PR #6947, opened 2026-08-13
fix(python): huggingface TEI auth, procedural-memory content handling, and proxy pip auto-install

### LINKED ISSUE #6118: Interactive input() and sys.exit(1) on ImportError in library code

### LINKED ISSUE #6150: Procedural memory creation crashes with AttributeError instead of a clear error when the LLM returns no content

### LINKED ISSUE #6301: bug(embeddings/huggingface): api_key ignored on base_url path, endpoint auth dropped

kartik-mem0:
## Linked Issue

Closes #6301
Closes #6150
Closes #6118

## Description

Three independent Python SDK hygiene fixes, one commit each.

**#6301 — `HuggingFaceEmbedding` and no-auth TEI endpoints.** On the `huggingface_base_url` path the OpenAI-compatible client was built with whatever `api_key` was configured and nothing else, so with no key set at all the OpenAI SDK fell back to `OPENAI_API_KEY` or raised at construction, crashing `Memory.from_config()` against a local TEI server that needs no auth. Now falls back to `HUGGINGFACE_API_KEY`, then the `"hf"` placeholder, matching what `docs/components/embedders/models/huggingface.mdx` already documents and what the TypeScript sibling embedder already does.

**#6150 — procedural memory crashes instead of reporting a refusal.** `remove_code_blocks()` assumed a string. `ChatCompletionMessage.content` is `Union[str, None]` and is `None` on safety refusals; a LangChain `AIMessage.content` can be a list of content blocks. Both crashed `_create_procedural_memory` with `AttributeError` rather than saying what happened. Now:

- a block list is joined into its text (a real response, so it is stored rather than dropped)
- anything else non-string degrades to `""`
- both `_create_procedural_memory` paths raise a clear `ValueError` when the model genuinely returned nothing, instead of embedding and storing an empty procedural memory

**#6118 — `mem0/proxy/main.py` shelled out to pip.** On `ImportError` it ran `pip install litellm` in the host process and called `sys.exit(1)` if that failed, killing the caller's process from inside a library import. Replaced with a catchable `ImportError` carrying the install command, the same pattern already used by the ollama embedder.

## Type of Change

- [ ] Documentation update

## Breaking Changes

`mem0.proxy` no longer auto-installs `litellm`. Users relying on that implicit install now get an `ImportError` telling them to run `pip install litellm`. This is the intent of #6118: a library should not mutate its host environment.

## Test Coverage

- `tests/embeddings/test_huggingface_embeddings.py`: env-var fallback, config key beating the env var, and the `"hf"` default.
- `tests/memory/test_memory_utils.py`: block lists joined, bare strings in a block list, code fences still stripped inside block content, unsupported types still `""`.
- `tests/test_memory.py`: LangChain block-list content reaching the vector store intact, and both the async and sync procedural paths raising a clear error (and not inserting) on empty content.
- `tests/test_proxy.py`: missing `litellm` raises an `ImportError` naming `pip install litellm`.

Verified red before green: with the fixes stashed, the six new memory tests fail with the exact pre-fix behavior (empty string stored, `DID NOT RAISE ValueError`). Full suite after: **1833 passed, 75 skipped**; `make lint` clean.

## Checklist

- [x] I have updated documentation if needed (no user-facing API change; the HF `"hf"` fallback matches the already-published docs)


karthik-indla:
**Review — requesting changes (completeness of #6118)**

All three fixes are correct in isolation and well-tested:
- **HF TEI auth (#6301):** `api_key = self.config.api_key or os.getenv("HUGGINGFACE_API_KEY") or "hf"` matches the TS sibling and the lmstudio precedent; only the `huggingface_base_url` branch is touched, local sentence-transformers untouched.
- **Procedural memory (#6150):** the `if not procedural_memory: raise ValueError(...)` guard is correctly placed after the try/except and before embed/insert on both the sync and async paths, and list-content joining preserves real LangChain block responses.
- **Proxy import (#6118):** clean `raise ImportError(...)`; the removed `subprocess`/`sys` imports were only used by the deleted install block.

**Blocking:** the PR `Closes #6118`, but #6118's title, component, and repro steps are entirely about `OllamaEmbedding` — `mem0/embeddings/ollama.py` still does `input("...Install it now? [y/N]")` then `sys.exit(1)`, which is the exact `OSError: reading from stdin` the issue reports. Only `proxy/main.py` was fixed. (The PR body's "same pattern as the ollama embedder" is inaccurate — the clean `raise ImportError` pattern lives in the ollama *LLM*, not the embedder.) Please apply the same `raise ImportError(...)` to `mem0/embeddings/ollama.py`, or drop `Closes #6118` — as-is, merging auto-closes an issue whose actual repro still crashes.

Minor: the `test_gemini_emeddings.py` → `test_gemini_embeddings.py` rename and new langchain-embeddings test are unrelated hygiene not listed in the PR's test coverage — harmless, just noting.

kartik-mem0:
@karthik-indla thanks for the careful read on the other two — flagging the `Closes #6118` / `OllamaEmbedding` mismatch is exactly the right thing to check, since that is the issue's headline repro.

I went and looked, and the blocking premise doesn't hold: `mem0/embeddings/ollama.py` was already clean at the commit this review was written against.

Branch head at the time of your review was `30bc29de`:

```console
$ git show 30bc29de:mem0/embeddings/ollama.py | head -9
from typing import Literal, Optional

from mem0.embeddings.base import EmbeddingBase

try:
    from ollama import Client
except ImportError:
    raise ImportError("The 'ollama' library is required. Please install it using 'pip install ollama'.")
```

No `input()`, no `sys.exit(1)`. That landed in `8ca9a0f2` ("fix: OSS Python SDK hygiene batch, 9 small bug fixes", #6770), merged to `main` on 2026-08-03, and it was already an ancestor of this branch when you reviewed:

```console
$ git merge-base --is-ancestor 8ca9a0f2 30bc29de && echo ancestor
ancestor
```

So #6118 names two files and both are now covered: the `OllamaEmbedding` half by #6770, the `mem0/proxy/main.py` half by this PR. Nothing in `mem0/`, `server/`, or `cli/python/` still does interactive-install or `sys.exit` inside a dependency-import handler — the only remaining `sys.exit` is `server/scripts/reset_admin_password.py`, an ordinary CLI exit.

You're right that this is not obvious from the diff, though — `git diff --stat origin/main...HEAD -- mem0/embeddings/ollama.py` is empty, so from this PR alone there is no evidence the embedder path is safe, and nothing was guarding it against regressing. Pushed `fa48ca73` to fix that: a regression test on #6118's literal repro.

```python
def test_missing_ollama_raises_actionable_import_error(monkeypatch):
    """Missing ollama raises a catchable ImportError naming the install command, never input()/sys.exit()."""
```

It stubs the `ollama` import out, asserts a catchable `ImportError` matching `pip install ollama`, and asserts `input` and `sys.exit` were never called. Verified red-then-green: with `ollama.py` temporarily reverted to the old interactive form the test fails with `DID NOT RAISE ImportError` (the old code falls through to `print` + `sys.exit(1)`), and passes against the current file. `tests/test_proxy.py` already asserts the equivalent for the litellm path, so I didn't duplicate it.

Also correcting the PR body while I'm here: "the same pattern already used by the ollama embedder" is accurate as of #6770, but you were right that it read as if this PR established the pattern. It didn't — #6770 did, in both the ollama LLM and the embedder.

Full suite after the change: **1865 passed, 75 skipped**, `make lint` clean.

On the minor note — the `test_gemini_emeddings.py` → `test_gemini_embeddings.py` rename and the new `test_langchain_embeddings.py` are scope creep, fair call. The rename is a zero-content filename typo fix and the langchain file is real coverage gated behind `pytest.importorskip`. Happy to split them into their own PR if you'd rather keep this one to the three linked issues — say the word and I'll pull them out.

karthik-indla:
**Re-review — blocker resolved, LGTM**

The #6118 concern is addressed: `mem0/embeddings/ollama.py` now carries the clean `raise ImportError(...)` pattern (the `input()`/`sys.exit()` block is gone), so the `OSError: reading from stdin` crash the issue reports no longer happens against `OllamaEmbedding`. The new `test_missing_ollama_raises_actionable_import_error` locks it in and asserts `input()`/`sys.exit()` are never called, so `Closes #6118` is now legitimate.

The three original fixes are all intact — HF TEI auth fallback (`config.api_key or HUGGINGFACE_API_KEY or "hf"`), the procedural-memory empty-content guard on both the sync and async paths, and the proxy `ImportError` (with `subprocess`/`sys` cleanly removed). No remaining issues from my side.

karthik-indla (approved):
Re-reviewed: the #6118 blocker is resolved — embeddings/ollama.py now raises a clean ImportError and a regression test locks it in (asserts input()/sys.exit are never called). The three original fixes (HF TEI auth, procedural-memory empty-content guard on sync+async, proxy ImportError) are intact. LGTM.

PR #6948, opened 2026-08-13
fix(plugins): bug-bash fixes for Cursor, Codex, Antigravity, and a Claude.ai docs page

kartik-mem0:
## Linked Issue

closes: #6976

## Description

Addresses the code-fixable half of the plugin bug bash. Each finding below states what was fixed here and, where the finding could not be fully closed from this repo, what is actually blocking it, so nothing is silently dropped.

**Codex (Finding 1).** `docs/integrations/codex.mdx` now states plainly that hooks are opt-in (installer script plus the `codex_hooks` feature flag), matching `scripts/install_codex_hooks.py`'s own docstring. Added a Codex Cloud section covering the Environment Variable vs. Secret persistence split: secrets are wiped before the agent phase, so `MEM0_API_KEY` has to be an Environment Variable for MCP calls to keep working. Also fixed an undercount: `README.md` and the installer's own print statement said the installer merges three entries, while `hooks/codex-hooks.json` has six event groups. *Not closeable here:* Codex has no plugin-host mechanism to auto-wire hooks from an installed plugin. That is an OpenAI product constraint, not a Mem0 bug, and confirming the doc fix resolves the reporter's confusion needs a live Codex session (app + CLI + Cloud).

**Cursor (Finding 2).** `integrations/mem0-plugin/.cursor-plugin/plugin.json` pointed at `".cursor-mcp.json"` with no `./` prefix while the Codex equivalent uses `"./.codex-mcp.json"`; normalized. Cosmetic, both resolve identically.

Deliberately *not* changed: `.cursor-plugin/marketplace.json`. An earlier pass normalized it onto the `interface`/object-`source`/`policy` schema the Codex and `.agents/plugins` manifests use, and that was reverted after checking [Cursor's published plugin reference](https://cursor.com/docs/reference/plugins), which documents the Claude-style schema the file already had: a required `name`, an `owner` block, optional `metadata`, and plugin entries with a **string** `source`. Swapping in the Codex schema would move the file *away* from its documented shape. Recording that here so it does not get re-litigated.

[...]

PR #6974, opened 2026-08-14
fix(ci): make the vouch check speak, unblock list updates, widen the docs exemption

### LINKED ISSUE #6973: Three gaps in the PR gate and vouch workflows shipped in #6894
Follow-up to #6894. Three defects found reviewing the merged workflows.

### 1. The vouch check is a silent no-op

`.github/workflows/vouch-check-pr.yml` runs `mitchellh/vouch/action/check-pr` with `auto-close: false`, described as comment-only mode. It is not. In v1.5.0 (`vouch/github.nu`, pinned SHA `d66fa29`), both the denounced path and the unvouched path do `if not $auto_close { return "closed" }` and return **before** posting a comment. There is no comment-but-do-not-close mode. Today an unvouched author sees nothing, the job goes green, and the check has no effect other than a line in the run log.

### 2. Vouch list management can never update the list

`.github/workflows/vouch-manage-by-issue.yml` passes `merge-immediately: "true"`. The action creates a PR and then calls `PUT /repos/{owner}/{repo}/pulls/{n}/merge` directly. The `Main Branch Rule` ruleset (id `11813754`, active) sets `required_approving_review_count: 1` with `bypass_actors: null`, so that call returns 405. Every `!vouch` / `!denounce` / `!unvouch` command fails at the merge step and `VOUCHED.td` on `main` is never updated.

Separately, `VOUCH_APP_ID` and `VOUCH_APP_PRIVATE_KEY` are not set on the repo, so the workflow currently fails at the token step before reaching any of this.

### 3. The docs-only exemption is too narrow

`.github/workflows/pr-gate.yml` exempts a PR only when every changed file starts with `docs/`. A PR that only fixes `README.md`, `CONTRIBUTING.md`, or `CODE_OF_CONDUCT.md` is still closed for not linking an accepted issue, which contradicts the "documentation-only changes are exempt" line in `CONTRIBUTING.md` and in the gate comment itself.

### Fix

- Post the comment from the workflow using the action `status` output, deduped by an HTML marker so a reopen does not comment twice.
- Set `merge-immediately: "false"` so the bot opens a PR and a maintainer merges it through the protected path. No admin change needed.
- Broaden the exemption to `docs/` plus top-level markdown, leaving `skills/**/*.md` and `.github/**` gated.

kartik-mem0:
## Linked Issue

Closes #6973

## Description

Follow-ups to #6894, each verified against the pinned action source and the live repo settings rather than inferred from the YAML. Sections 1-4 are the original round, 5-7 came out of review.

### 1. The gate closed a maintainer's own PR

#6948 was opened by a repo maintainer and closed by this gate. The job-level `if:` already exempted `OWNER`/`MEMBER`/`COLLABORATOR`, so the exemption did not fail to exist, it failed to match.

`author_association` is rendered **for the viewer**, and a webhook payload has no privileged viewer:

| Value | What it actually requires |
|---|---|
| `MEMBER` | the author's org membership is **public** |
| `COLLABORATOR` | a **direct** repository invite |

The author of #6948 is an active `mem0ai` member with **private** membership, and their `maintain` role arrives through a team rather than a direct invite (`collaborators?affiliation=direct` lists two accounts, neither of them theirs). They match neither arm, so the payload says `CONTRIBUTOR`. Queried with that author's own token the same PR reports `MEMBER`, which is why this reads as correct from the API and breaks in the runner.

Every `pr-gate` run against that author's PRs executed the job instead of skipping it. #6896 survived only because it is docs-only, this PR only because it links an accepted issue, #6948 had neither.

The guard now also skips any PR whose head branch lives in this repository:

```yaml
github.event.pull_request.head.repo.full_name != github.repository &&
```

Pushing a branch to `mem0ai/mem0` already requires write access and outside contributors always arrive from a fork, so this expresses the intended exemption without depending on who is looking. The `author_association` arm stays, since it still covers members who work from their own fork. Applied to both `pr-gate.yml` and `vouch-check-pr.yml`, which carried the same guard and the same bug.

### 2. The vouch check was a green no-op

`vouch-check-pr.yml` ran `mitchellh/vouch/action/check-pr` with `auto-close: false` on the understanding that this meant comment-only. It does not. `gh-check-pr` in v1.5.0 (`vouch/github.nu` at the pinned SHA `d66fa29`) returns one of four statuses, and on both the unvouched and the denounced path `auto-close: false` returns **before** the `POST .../comments` call and before the close. There is no comment-but-do-not-close mode in the action. So an unvouched author saw nothing, a denounced author saw nothing, the job went green, and the only trace was a line in the run log. The `!denounce` keyword in `vouch-manage-by-issue.yml` was decorative.

Two inputs, two axes. They are now set independently:

| Input | Value | Effect |
|---|---|---|
| `require-vouch` | `false` | an unvouched author is **not** closed, status `allowed` |
| `auto-close` | `true` | a **denounced** author is closed by the action, status `closed` |

`require-vouch` has no bearing on the denounced path: a denounced author is closed either way once `auto-close` is on. And `require-vouch: true` would close every first-time contributor, which is the opposite of what a trust list is for, since the funnel has to stay open or nobody ever earns a vouch. So `require-vouch: false` with `auto-close: true` is the pair that enforces on the denounce axis only, which is the behaviour this repo wants.

The soft comment for an unvouched author is ours, not the action's: a `github-script` step gated on `steps.vouch.outputs.status == 'allowed'`, deduped by a `<!-- vouch-check -->` marker so the reopen the PR Gate instructs authors to perform does not produce a second one. `allowed` and `closed` are disjoint, so a denounced author gets the action's close message and nothing from us. Re-keying that step off `allowed` would break that, and it is called out in `.github/AGENTS.md`.

### 3. Vouch list management could never update the list

`vouch-manage-by-issue.yml` passed `merge-immediately: "true"`. `open-pr` in the same module creates the PR and then calls `PUT /repos/{owner}/{repo}/pulls/{n}/merge` directly.

The `Main Branch Rule` ruleset (id `11813754`, active) has `required_approving_review_count: 1` and `bypass_actors: null`, so that call returns 405. Every `!vouch`, `!denounce`, and `!unvouch` would have failed at the merge step, leaving an unmerged PR and an unchanged `VOUCHED.td` on `main`.

Now `merge-immediately: "false"`. The bot opens the PR, a maintainer merges it through the normal protected path. This needs no repo-admin change. Adding a bypass actor would also work but weakens the ruleset for a workflow that runs on a comment trigger, and `pull-request: "false"` is not an option because the same ruleset blocks direct pushes.

### 4. Docs-only exemption was narrower than the policy it enforces

The gate exempted a PR only when every file started with `docs/`. `CONTRIBUTING.md` and the gate comment both promise that documentation-only changes are exempt, so a README-only fix was being closed by a rule that says it does not apply to it.

The predicate now also accepts a **named allowlist** of four root files: `README.md`, `CONTRIBUTING.md`, `CODE_OF_CONDUCT.md`, `SECURITY.md`.

The first cut of this accepted any top-level `.md`, which review caught: the repository root also holds `AGENTS.md`, `CLAUDE.md`, and `LLM.md`, the instructions coding agents read before touching this codebase. Under that rule, rewriting what every agent is told to do skipped the gate while rewriting `skills/mem0/SKILL.md` did not. Those three are functional files that happen to be prose, so they stay gated, and adding a genuinely prose root file now means adding it to `rootDocs` on purpose. Nested markdown stays gated for the same reason: `skills/**/*.md` and everything under `.github/` are functional too.

### 5. Reopening was a manual step nobody would take

The close comment told the author to get the issue labeled and then reopen the PR themselves, which asks the person with the least context to watch for a label they cannot see applied.

`pr-gate.yml` gains a `reopen` job on `issues: [labeled]`. When `accepted` lands on an issue, it walks that issue's `closedByPullRequestsReferences`, and for each closed PR that links it, still carries the `<!-- pr-gate -->` marker, and is not from an author denounced in `VOUCHED.td`, reopens it and says so. The `edited` arm covers the other order, where the label was already there and the author adds `Closes #<number>` afterwards.

Vouch outranks the gate here on purpose. The denounce check is what stops these two workflows cancelling each other out: without it, labelling an issue would reopen a PR that vouch had closed. Reopening runs through `GITHUB_TOKEN`, which by design raises no further workflow events, so `gate` cannot bounce a freshly reopened PR back out.

### 6. An edit in the first seconds left the PR ungated forever

Found in review. The concurrency group was keyed on the event name and the PR number only, so `opened` and `edited` shared a group with `cancel-in-progress` on. An author who ticks a template checkbox in the seconds after opening cancels the run that was about to gate them, and nothing replaces it: `gate` skips `edited`, `reopen` skips a PR that is still open, and `opened` fires exactly once.

`github.event.action` is now part of the key. Rapid successive edits still cancel each other, which is the dedup that was wanted; `opened` and `edited` no longer collide. Both keys carry weight, and `.github/AGENTS.md` now says why for each.

### 7. The script tests did not run, and one could not fail

Also found in review. `vouch-decision.test.js` holds the vouch decision table and asserts the workflow's inputs still produce it, but nothing ran it, and `decide()` is a **hand transcription** of `gh-check-pr` rather than a call into it. Left alone it agrees with itself whatever the action does, which made bumping the pinned vouch SHA the one change it could not fail on. That is exactly the change `.github/AGENTS.md` tells you to run it after.

It now asserts `vouch-check-pr.yml` still pins `PINNED_VOUCH_SHA`, so a bump fails with instructions to re-read `gh-check-pr` at the new revision and correct both `decide()` and the table in `.github/AGENTS.md`.

A `GitHub Scripts` job in `ci-gate.yml` runs every `.github/scripts/*.test.js` on any change to the scripts or to the files those tests read. That also picks up `infer-component-labels.test.js`, which nothing ran either. It is a plain job rather than a called workflow because a reusable workflow wrapping three `node` invocations would be more file than test.

## Not changed here

`GATE_EFFECTIVE_FROM` stays at the shipped `2026-08-12T00:00:00Z`. It was briefly moved to the real merge time in this branch and then reverted: the same-repo exemption in fix 1 already covers the maintainer PRs that moving it was meant to protect, and moving it forward would have exempted the open backlog in between from the gate entirely. The comment above announcing the move is superseded by that revert.

## Type of Change

- [x] Documentation update

## Breaking Changes

N/A. The gate closes strictly fewer PRs than before and reopens ones it closed, and the vouch workflows go from silent-or-failing to working.

## Test Coverage

All workflow files parse under `yaml.safe_load`.

`.github/scripts/vouch-decision.test.js` asserts the decision table above against `vouch-check-pr.yml`: the two inputs, the status each of the four author states produces, and how many comments each one receives. It also pins the vouch SHA, per fix 7. Verified to fail, not just pass: setting the workflow to a different revision fires the assertion with the follow-the-instructions message and exits 1.

`.github/scripts/pr-gate-docs-exemption.test.js` covers the fix 4 predicate over 17 cases. It pulls the `rootDocs` and `isDocs` lines out of `pr-gate.yml` and evaluates them, so it exercises the shipped rule rather than a copy that can drift, and it pins `AGENTS.md`, `CLAUDE.md`, `LLM.md`, `skills/mem0/SKILL.md`, nested `.github/` files, and the empty file list on the gated side. Also verified to fail: adding `AGENTS.md` to `rootDocs` breaks two cases and exits 1.

Both run in CI through the new `GitHub Scripts` job, along with `infer-component-labels.test.js`. All three pass locally and in CI.

Fix 1 was diagnosed against live API state rather than reasoning: `orgs/mem0ai/public_members/{author}` returns 404 (membership is private), `repos/mem0ai/mem0/collaborators?affiliation=direct` does not list the author, `repos/mem0ai/mem0/collaborators/{author}/permission` returns `maintain`, and `repos/mem0ai/mem0/pulls/6948` returns `author_association: MEMBER` to that author's own token while the runner acted on a different value. The three real `pr-gate` runs confirm the job executed rather than skipping in every case.

Fix 2 was read out of `vouch/github.nu` at the pinned SHA rather than from the action's README, including the four status values, the early returns under `auto-close: false`, and the fact that `require-vouch` does not touch the denounced path.

Not exercised end to end against a live fork PR. The vouch comment path, the vouch PR path, and the reopen path all need a real trigger to confirm.

## Follow-ups

`VOUCH_APP_ID` and `VOUCH_APP_PRIVATE_KEY` are still not set on this repository, so `vouch-manage-by-issue.yml` fails at the `create-github-app-token` step before reaching any of the above. That is a repo settings task requiring admin, not a code change, and it is called out in `.github/AGENTS.md` now.

#6948 stays closed until this merges. Reopening it before then fires the gate against unfixed `main` and closes it a second time.

Two known gaps in the reopen path, both mild. A PR that the gate closed, that someone reopened, and that a maintainer then closed deliberately still carries the marker, so labelling its issue reopens it again; a maintainer closes it once more. And an author who strips `Closes #<number>` out after passing keeps an open PR, which a reviewer sees anyway.

## Checklist


kartik-mem0:
### 4. `GATE_EFFECTIVE_FROM` was never set to the merge date

Pushed as a second commit.

#6894 shipped the cutoff at `2026-08-12T00:00:00Z` and merged at `2026-08-14T11:35:27Z`. Its own description said to set the cutoff to the real merge date before merging, and that step was missed.

26 open PRs were filed inside that 59-hour gap. `edited`, `reopened`, and `ready_for_review` all fire on PRs opened long before the gate existed, so any of those 26 that gets touched would be closed by a rule that did not exist when it was filed, which is the exact failure the cutoff was added to prevent.

Cutoff is now `2026-08-14T11:35:27Z`, the actual merge timestamp of 290de24bb.

kartik-mem0:
### Fifth fix pushed: `author_association` is not a reliable maintainer test

#6948 was opened by a repo maintainer and closed by this gate an hour ago. The job-level `if:` already exempts `OWNER`/`MEMBER`/`COLLABORATOR`, so the exemption did not fail to exist, it failed to match.

`author_association` is rendered **for the viewer**, and a webhook payload has no privileged viewer:

| Value | What it actually requires |
|---|---|
| `MEMBER` | the author's org membership is **public** |
| `COLLABORATOR` | a **direct** repository invite |

The author of #6948 is an active `mem0ai` member with **private** membership, and their `maintain` role comes through a team rather than a direct invite (`collaborators?affiliation=direct` lists only two accounts). So they match neither arm and arrive as `CONTRIBUTOR`. Queried with their own token the same PR reports `MEMBER`, which is why this looks fine from the API and breaks in the runner.

Evidence: all three `pr-gate` runs against this author's PRs executed the job rather than skipping it. #6896 survived only because it is docs-only, #6974 only because it links an accepted issue, #6948 had neither and was closed.

**Fix**: also skip any PR whose head branch lives in this repository.

```yaml
github.event.pull_request.head.repo.full_name != github.repository &&
```

Pushing a branch to `mem0ai/mem0` already requires write access and outside contributors always arrive from a fork, so this means the same thing as the intended exemption without depending on who is looking. The `author_association` arm stays, since it still covers members who work from their own fork.

Applied to `pr-gate.yml` and `vouch-check-pr.yml`. `.github/AGENTS.md` also loses the claim that "no member can be flagged", which was the assumption that hid this.

#6948 stays closed until this merges. Reopening now would fire the gate again against unfixed `main` and close it a second time.

.github/workflows/pr-gate.yml
```diff
@@ -2,11 +2,13 @@ name: PR Gate
...
-    types: [opened, reopened, edited, ready_for_review]
+    types: [opened, reopened, ready_for_review, edited]
+  issues:
+    types: [labeled]
 
 concurrency:
-  group: pr-gate-${{ github.event.pull_request.number }}
-  cancel-in-progress: true
+  group: pr-gate-${{ github.event_name }}-${{ github.event.pull_request.number || github.event.issue.number }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request_target' }}
```
whysosaket, line 11: The edited run shares pr-gate-pull_request_target-N with the opened run and cancels it. gate skips edited and reopen skips an open PR, so ticking a template checkbox in the first 30 seconds leaves the PR ungated.

.github/workflows/pr-gate.yml
```diff
@@ -47,7 +52,7 @@ jobs:
             const files = await github.paginate(github.rest.pulls.listFiles, {
               owner, repo, pull_number: pr.number, per_page: 100,
             });
-            if (files.length > 0 && files.every((file) => file.filename.startsWith('docs/'))) {
+            if (files.length > 0 && files.every((file) => file.filename.startsWith('docs/') || (!file.filename.includes('/') && file.filename.endsWith('.md')))) {
```
whysosaket: This exempts any PR whose files are all top-level .md, and root holds AGENTS.md, CLAUDE.md and LLM.md. Rewriting what agents read here skips the gate, rewriting skills/mem0/SKILL.md does not.

.github/workflows/vouch-check-pr.yml
```diff
@@ -16,12 +16,44 @@ jobs:
...
+      github.event.pull_request.head.repo.full_name != github.repository &&
       !contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.pull_request.author_association)
     runs-on: ubuntu-latest
     steps:
       - uses: mitchellh/vouch/action/check-pr@d66fa29a64600490892131ad87597c30c91fcac4 # v1.5.0
+        id: vouch
         with:
           pr-number: ${{ github.event.pull_request.number }}
-          auto-close: false
+          require-vouch: false
```
whysosaket, line 27: Section 2 of the description says auto-close stays false with the comment coming from a gated script step. This is require-vouch false and auto-close true, which is a different denounce path than the one being approved.

.github/scripts/vouch-decision.test.js
```diff
@@ -0,0 +1,112 @@
+const assert = require('assert');
```
whysosaket, line 1: No workflow runs this, and decide reimplements gh-check-pr rather than reading it. .github/AGENTS.md says to run it after bumping the pinned SHA, which is the one change it cannot fail on.

karthik-indla (approved):

whysosaket (approved):
All four points from the last round are addressed.

I read gh-check-pr at d66fa29 rather than trusting the table. Denounced is tested before require_vouch, so auto-close closes it either way, and an unvouched author with require-vouch false returns allowed before anything is posted. allowed and closed are disjoint, so the soft comment cannot double up. open-pr only calls PUT /pulls/{n}/merge under merge-immediately, and can-manage gates the keywords on write access, so arming denounce does not hand it to anyone.

The concurrency key holds up for every action, not just edited. opened can now only be cancelled by another opened, which never arrives, and ready_for_review and reopened each cancel only a run that replaces them. All three script tests pass locally, and the SHA pin fails on a bump.

Two things left, neither of them blocking. component-keywords.json is missing from the github_scripts filter and infer-component-labels.test.js loads its table from there, so the one file in the set that gets edited alone is the one edit the job never sees. And parseDenounced in vouch-decision.test.js transcribes the reopen job's parser with one of its four lines pinned, while the docs-exemption test next to it evaluates the shipped lines instead.

PR #6983, opened 2026-08-15
fix(memory): persist LLM-computed linked_memory_ids into the add pipeline

### LINKED ISSUE #6982: Fix linked_memory_ids being computed by the LLM but discarded by the add pipeline
### Component

Python SDK

### Summary

The V3 additive extraction prompt asks the LLM to emit `linked_memory_ids` so new memories can be related to existing ones, but `_add_to_vector_store` never reads that field. The LLM-computed semantic links are dropped, and retrieval falls back to string-based entity linking only. The prompt also instructs the LLM to return UUIDs while the pipeline only shows it integer ids, making the field unreliable by construction.

### Steps to Reproduce

```python
from unittest.mock import MagicMock, Mock, patch
from mem0 import Memory

m = Memory.__new__(Memory)
m.config = MagicMock()
m.config.custom_instructions = None
m.custom_instructions = None
m.api_version = "v1.1"
m.llm = Mock()
m.embedding_model = Mock()
m.vector_store = Mock()
m.db = Mock()
m.db.get_last_messages.return_value = []
m.db.save_messages = MagicMock()
m.db.batch_add_history = MagicMock()
m._entity_store = Mock()
m._entity_store.search_batch.return_value = [[]]

m.vector_store.search.return_value = [
    SimpleNamespace(id="uuid-a", payload={"data": "User has a dog named Poppy"})
]
m.llm.generate_response.return_value = (
    '{"memory": [{"text": "Poppy had a vet checkup", "linked_memory_ids": ["0"]}]}'
)
m.embedding_model.embed.return_value = [0.1, 0.2, 0.3]
m.embedding_model.embed_batch.return_value = [[0.1]]

with patch("mem0.memory.main.capture_event"):
    m._add_to_vector_store(
        messages=[{"role": "user", "content": "Poppy had a vet checkup"}],
        metadata={"user_id": "u1"},
        filters={"user_id": "u1"},
        infer=True,
    )

payload = m.vector_store.insert.call_args.kwargs["payloads"][0]
print(payload)
print("linked_memory_ids present:", "linked_memory_ids" in payload)
```

### Expected Behavior

The stored payload should contain `linked_memory_ids` resolved to `["uuid-a"]`, because the extraction prompt explicitly defines that field as part of the output schema.

### Actual Behavior

The stored payload has no `linked_memory_ids` key. The value computed by the LLM is ignored.

### Environment

- mem0 version: 2.0.18
- Python version: 3.12
- OS: Windows 11

### What I Ran

The reproduction script above, run against the current `main` branch.

### What I Saw

```text
linked_memory_ids present: False
```

### Why This Is a Bug

The prompt's output schema requires `linked_memory_ids` (`mem0/configs/prompts.py`, `ADDITIVE_EXTRACTION_PROMPT`), which means the intended design is for those links to be used. But Phase 4 of `mem0/memory/main.py` reads only `text` and `attributed_to` from each extracted memory, so the links are discarded. Separately, the prompt shows UUIDs in examples while the pipeline presents existing memories with integer ids, so the field cannot be populated reliably.

### What I Ruled Out

- Not a JSON parsing issue: `text` is stored correctly, so extraction succeeds.
- Not an LLM output issue: the mocked LLM explicitly returns `linked_memory_ids`.
- Not covered by entity linking: the entity store links by entity strings, not the LLM's semantic memory-to-memory links.

### AI Assistance

No AI involved

a-yeyang:
## Summary

The V3 additive extraction prompt (`ADDITIVE_EXTRACTION_PROMPT`) asks the LLM to emit a `linked_memory_ids` field on each extracted memory so retrieval can build a semantic graph between new and existing memories. But Phase 4 of `_add_to_vector_store` (both sync and async) only ever read `text` and `attributed_to` off each extracted memory dict — `linked_memory_ids` was parsed out of the LLM's JSON response and then silently discarded before it ever reached the vector-store payload.

There's a second, related defect: the prompt's "Existing Memories" section claimed those ids were UUIDs and told the LLM to return UUIDs in `linked_memory_ids`, but the pipeline only ever shows the LLM sequential anti-hallucination index strings (`"0"`, `"1"`, ...) via a local `uuid_mapping`. The documented contract never matched what was actually presented to the model.

## Fix

1. Added `_resolve_linked_memory_ids(raw, uuid_mapping)` — translates the LLM's index strings back to real vector-store IDs using the `uuid_mapping` already built in Phase 1. Drops anything that isn't a string, doesn't resolve to a known index (hallucinated/malformed), or resolves to a duplicate, rather than persisting it.
2. Phase 4 (sync + async) now calls this and, when the result is non-empty, stores it under `linked_memory_ids` in the payload sent to `vector_store.insert`.
3. Updated `ADDITIVE_EXTRACTION_PROMPT`'s "Existing Memories" section and its two linking examples (Example 10, Example 12) plus the output-schema example to describe/use index strings consistently instead of claiming UUIDs.

## Testing

New file `tests/memory/test_linked_memory_ids_from_llm.py` (16 tests):
- `_resolve_linked_memory_ids` unit tests: resolves known indices, preserves order, drops hallucinated/non-string/duplicate entries, handles empty/None/non-list input and an empty mapping.
- Sync + async integration tests against `_add_to_vector_store`, asserting the resolved `linked_memory_ids` lands in the payload passed to `vector_store.insert` — including "hallucinated index dropped, not persisted" and "field omitted leaves payload unchanged".

Verified locally:
```
$ python -m pytest tests/memory/test_linked_memory_ids_from_llm.py -v
============================== 16 passed in 0.50s ==============================

$ python -m pytest tests/memory/ tests/test_memory.py -q
444 passed in 2.82s

$ ruff check mem0/memory/main.py mem0/configs/prompts.py tests/memory/test_linked_memory_ids_from_llm.py
All checks passed!
```

Closes #6982

## AI Assistance Disclosure

I used an AI coding assistant (Claude Code) to help investigate the root cause, implement the fix, and write the test suite. I reviewed the change and verified it by running the test suites shown above myself.

### DIFF
--- a/mem0/configs/prompts.py
+++ b/mem0/configs/prompts.py
@@ -506,11 +506,13 @@ Memories already captured from recent messages in this session (up to 20). This 
 ## Existing Memories
 
 Memories currently in the system relevant to this conversation. Formatted as:
-[{"id": "uuid-string", "text": "..."}, ...]
+[{"id": "0", "text": "..."}, {"id": "1", "text": "..."}, ...]
+
+The "id" values are sequential index strings, matching your ADD output ID scheme — not the memories' real underlying IDs.
 
 Use these ONLY for deduplication and linking — do NOT extract new memories from Existing Memories. Your extractions must come exclusively from New Messages. If new information in New Messages is semantically equivalent to an Existing Memory with no meaningful new context, skip it.
 
-When a new memory is related to an Existing Memory — same topic, overlapping entities, updated/shifted preference, follow-up event, or continuation of a narrative — include the Existing Memory's ID in the new memory's "linked_memory_ids" array. Your ADD output IDs remain sequential ("0", "1", ...) but linked_memory_ids uses the UUIDs from this list.
+When a new memory is related to an Existing Memory — same topic, overlapping entities, updated/shifted preference, follow-up event, or continuation of a narrative — include the Existing Memory's "id" (the index string, e.g. "0", "1") in the new memory's "linked_memory_ids" array. Your ADD output IDs remain sequential ("0", "1", ...) and linked_memory_ids uses those same index strings from the Existing Memories list — never invent a different ID format.
 
 
 IMPORTANT: An existing memory about an entity (e.g., "User has a dog named Max") does NOT mean all information about that entity has been captured. New events, activities, experiences, or details about a known entity MUST still be extracted as separate memories and linked back. Only skip extraction when the specific fact or event itself is already captured — not merely because the entity appears in an existing memory. "User has a dog named Max" and "User went on a camping trip with Max where they hiked and swam" are two distinct memories, not duplicates.
@@ -844,15 +846,15 @@ Every count (4 Mummies, 2 Construct Guardians, 6 Skeletal Warriors) and every sp
 
 Summary: ""
 Recently Extracted: []
-Existing Memories: [{"id": "a1b2c3d4-5678-9abc-def0-111111111111", "text": "User has a dog named Poppy, a golden retriever"}, {"id": "b2c3d4e5-6789-abcd-ef01-222222222222", "text": "User works as a Senior Engineer at Shopify"}]
+Existing Memories: [{"id": "0", "text": "User has a dog named Poppy, a golden retriever"}, {"id": "1", "text": "User works as a Senior Engineer at Shopify"}]
 New Messages:
 [{"role": "user", "content": "Poppy had her vet checkup yesterday — she's healthy but needs to lose a few pounds. Also, I'm switching teams at work next month to the payments platform."}]
 Observation Date: 2025-03-15
 
 Output:
 {"memory": [
-  {"id": "0", "text": "User's dog Poppy had a vet checkup around March 14, 2025, is healthy but needs to lose weight", "linked_memory_ids": ["a1b2c3d4-5678-9abc-def0-111111111111"]},
-  {"id": "1", "text": "User is switching teams at Shopify to the payments platform in April 2025", "linked_memory_ids": ["b2c3d4e5-6789-abcd-ef01-222222222222"]}
+  {"id": "0", "text": "User's dog Poppy had a vet checkup around March 14, 2025, is healthy but needs to lose weight", "linked_memory_ids": ["0"]},
+  {"id": "1", "text": "User is switching teams at Shopify to the payments platform in April 2025", "linked_memory_ids": ["1"]}
 ]}
 
 Both new memories link to related existing memories — the vet checkup links to the existing Poppy memory, and the team switch links to the existing Shopify memory. This enables the system to build a graph of related memories.
@@ -887,7 +889,7 @@ FIVE topics across 5 messages — each one extracted separately. Do not stop aft
 
 Summary: "John has a dog named Max."
 Recently Extracted: []
-Existing Memories: [{"id": "a1b2c3d4-0000-0000-0000-111111111111", "text": "John has a dog named Max"}]
+Existing Memories: [{"id": "0", "text": "John has a dog named Max"}]
 New Messages:
 [{"role": "user", "content": "John: Max and I had a blast on our camping trip last summer. We hiked, swam, and made great memories. It was a really peaceful experience."},
  {"role": "assistant", "content": "Maria: That sounds amazing! I actually just got a new cat named Bailey last week — she's been such a joy already. Camping with pets is so soul-nourishing."},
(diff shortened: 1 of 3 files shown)


mem0/memory/main.py
```diff
@@ -1035,6 +1066,9 @@ def _add_to_vector_store(self, messages, metadata, filters, infer, prompt=None):
             mem_metadata["updated_at"] = mem_metadata["created_at"]
             if mem.get("attributed_to"):
                 mem_metadata["attributed_to"] = mem["attributed_to"]
+            linked_ids = _resolve_linked_memory_ids(mem.get("linked_memory_ids"), uuid_mapping)
```
kartik-mem0, line 1069: The TypeScript SDK has the same bug and this PR leaves it there, so merging as is creates a Python/TS divergence rather than removing one.

Same shape on that side: `mem0-ts/src/oss/src/memory/index.ts` builds `uuidMapping` at 898, declares `linked_memory_ids` on the extracted type at 941, then copies only `attributedTo` into the payload at 1037. `prompts/index.ts:326`, `:667` and `:749` still describe the ids as UUIDs.

mem0/memory/main.py
```diff
@@ -484,6 +484,37 @@ async def update(
...
+    """
+    if not isinstance(raw, list):
+        return []
+    resolved = []
+    seen = set()
+    for entry in raw:
+        if not isinstance(entry, str):
+            continue
+        real_id = uuid_mapping.get(entry)
+        if real_id is None or real_id in seen:
```
kartik-mem0, line 511: An unresolvable entry vanishes with no trace here, which is the one case worth seeing: it means the model returned an index that is not in `uuid_mapping`, either hallucinated or out of range.

A single `logger.debug` naming the dropped entry would make that diagnosable without changing behavior.

mem0/configs/prompts.py
```diff
@@ -506,11 +506,13 @@ def get_update_memory_messages(retrieved_old_memory_dict, response_content, cust
...
 
 Use these ONLY for deduplication and linking — do NOT extract new memories from Existing Memories. Your extractions must come exclusively from New Messages. If new information in New Messages is semantically equivalent to an Existing Memory with no meaningful new context, skip it.
 
-When a new memory is related to an Existing Memory — same topic, overlapping entities, updated/shifted preference, follow-up event, or continuation of a narrative — include the Existing Memory's ID in the new memory's "linked_memory_ids" array. Your ADD output IDs remain sequential ("0", "1", ...) but linked_memory_ids uses the UUIDs from this list.
+When a new memory is related to an Existing Memory — same topic, overlapping entities, updated/shifted preference, follow-up event, or continuation of a narrative — include the Existing Memory's "id" (the index string, e.g. "0", "1") in the new memory's "linked_memory_ids" array. Your ADD output IDs remain sequential ("0", "1", ...) and linked_memory_ids uses those same index strings from the Existing Memories list — never invent a different ID format.
```
kartik-mem0, line 515: This half of the fix has no test. It is the half that makes the model emit indices in the first place, so a silent revert here would put the bug back with the helper still in place and every link resolving to nothing.

One assertion that `ADDITIVE_EXTRACTION_PROMPT` contains the index style example would be enough to catch that.

kartik-mem0 (commented):
## Changes requested

Real fix and placed in the right spot: the LLM's `linked_memory_ids` was parsed and then dropped before the payload was written, and the prompt described the ids as UUIDs when the model only ever sees index strings. The blocker is that the TypeScript SDK carries the identical bug and is untouched, so this lands a Python/TS split that does not exist today.

**To unblock:**
1. `mem0/memory/main.py:1069` - port the same two changes to `mem0-ts/src/oss/src/memory/index.ts`, or open a follow-up issue and link it here so it does not get lost

Optional: the async payload tests are thinner than the sync ones, the prompt rewrite has no test of its own, and the drop at `main.py:511` is silent.

- **Problem is real:** two defects. `main.py:937` builds `uuid_mapping[str(idx)] = mem.id`, then the Phase 4 loop at `:1030-1039` reads only `text` and `attributed_to`, never `linked_memory_ids` (async identical at `:2599` and `:2688-2697`). And `prompts.py:506-513` told the model the ids were UUIDs while it only ever sees index strings. Both are on the default path, since `add()` defaults to `infer=True` with no version gate.
- **Live check (real Azure LLM, oss-qdrant lane):** seed one memory with `infer=False`, then add a related follow-up with `infer=True` and read the payload back. `main` 7e0961557 sync: FAIL, the extracted row comes back with `metadata: None`. Same on async. PR head cba4500ba: PASS on both sync and async, the link resolves to the real seeded memory id.
- **Tests / red-green:** 16 passed on the PR. Restoring `main`'s `main.py` and `prompts.py` turns it into a collection error, `ImportError: cannot import name _resolve_linked_memory_ids`, so the new tests cannot run against the old code at all. Restored: 16 passed. The live run above is what actually proves the fix, the unit tests mostly exercise the pure helper.
- **CI:** `check`, `gate`, `label` green. Vercel red, as on every fork PR, not merge blocking. `ruff check` clean.
- **Security:** cleared. The helper only accepts list entries that resolve through `uuid_mapping`, so a hallucinated or out of range index cannot inject an arbitrary id into the payload.
- **Bloat (ponytail):** lean. One small pure helper shared by the sync and async call sites, which is the right shape for this file.
- **Checked and cleared:** `linked_memory_ids` is already the entity store's own payload key (`main.py:624`, `:641`), read by the entity boost at `:1797`. That read goes through `entity_store.search()`, so it only ever sees entity payloads and never these memory ones. No runtime collision, just a name shared by two collections.

</details>

The uuid_mapping resolution and the sequential index scheme in the prompt are the part that makes this correct rather than merely present, and both are right. Push when ready and I will re-review.

kartik-mem0:
@a-yeyang please first open a issues and then i will add the accepted tag to it!

thanks!

PR #6993, opened 2026-08-15
Protect against None timestamps in Valkey layer

### LINKED ISSUE #6994: Protect against None timestamps in Valkey layer
### Component

Python SDK

### Summary

On the Valkey vector store, `insert()`/`update()` call `datetime.fromisoformat()` on `created_at`/`updated_at` unconditionally. When the value is `None`, this raises `TypeError: fromisoformat: argument must be str`. It triggers in inference mode (`infer=True`): entity records never set timestamps, `list()` reads them back as `None`, and `_upsert_entity` feeds that payload straight into `update()`. The error is swallowed as a warning inside consolidation, so the run "succeeds" but the entity link is silently never updated — degrading entity-boost ranking on Valkey only. Qdrant and other backends run the same scenario cleanly.

### Steps to Reproduce

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "valkey",
        "config": {
            "collection_name": "repro",
            "embedding_model_dims": 1536,
            "valkey_url": "valkey://localhost:6379",
        },
    },
}

# Two infer=True adds mentioning the same proper noun → same entity upserted twice.
# The second upsert reads the entity back (updated_at=None) and calls update().
m.add("Alice loves hiking in the mountains.", user_id="u1")   # creates entity "Alice"
m.add("Alice also enjoys rock climbing.", user_id="u1")       # updates entity "Alice" → crash path
```

### Expected Behavior

Both adds complete, and the "Alice" entity's `linked_memory_ids` is updated to include both memories.

### Actual Behavior

The second add logs a warning and the entity update is silently dropped (link never updated):

```
WARNING  mem0.memory.main  Entity upsert failed for 'Alice': fromisoformat: argument must be str
  File ".../mem0/vector_stores/valkey.py", line 515, in update
    hash_data["updated_at"] = int(datetime.fromisoformat(payload["updated_at"]).timestamp())
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: fromisoformat: argument must be str
```

### Environment

- mem0 version: 2.0.18
- Python/Node version: Python 3.12.11
- OS: Linux (Amazon Linux 2, kernel 5.10)

### What I Ran

Two `infer=True` adds that mention the same proper noun, against the Valkey backend, so the same entity ("Alice") is upserted twice — the second upsert takes the update-existing-entity path. See the script under "Steps to Reproduce". The same scenario against the `qdrant` backend (identical config except `provider`) was run as a control.

### What I Saw

On the second add, this warning is emitted and the entity update is dropped (no `hset` for the entity is issued):

```
WARNING  mem0.memory.main  Entity upsert failed for 'Alice': fromisoformat: argument must be str
```

The underlying exception is `TypeError: fromisoformat: argument must be str`, raised at `mem0/vector_stores/valkey.py:515` in `update()`. It is caught by `_upsert_entity`'s `try/except` ([mem0/memory/main.py:649](mem0/memory/main.py#L649)), which is why only the warning surfaces and the run still exits 0. The same scenario on `qdrant` produces no warning and updates the entity correctly.

### Why This Is a Bug

`update()`/`insert()` assume `created_at`/`updated_at` are always ISO strings and parse them unconditionally, but the value handed in is `None`. The code itself shows the inconsistency that produces the `None`:

- `list()` injects both keys unconditionally ([valkey.py:795-796](mem0/vector_stores/valkey.py#L795-L796)): `"updated_at": result.payload.get("updated_at")` → `None` when absent.
- `search()`/`get()` add them only when present, so they never emit `None`.

`_upsert_entity` reads an entity via `list()` and feeds that same payload — now carrying `updated_at=None` — straight back into `update()` ([main.py:630](mem0/memory/main.py#L630)). `datetime.fromisoformat(None)` then raises. The expected behavior is parity with every other backend (and with Valkey's own `search`/`get`): a missing timestamp should be tolerated, not crash the update.

### What I Ruled Out

- **Not a Valkey server problem** — the identical scenario on `qdrant` runs clean; this is in Mem0's Valkey adapter, not the datastore.
- **Not user input** — the `None` timestamp is generated internally by `list()`, not passed by the caller.
- **Not the existing key-presence guard** — the key *is* present in the payload (`"updated_at" in payload` is `True`); its *value* is `None`, which the old guard didn't catch.

### AI Assistance

BCathcart:
## Linked Issue

closes #6994

## Description

`ValkeyDB.insert()`/`update()` called `datetime.fromisoformat()` on `created_at`/`updated_at` unconditionally, so a None timestamp raised `TypeError: fromisoformat: argument must be str`. This hits infer=True consolidation, where entity records (which never set timestamps) get read back with None and fed into update().

I changed the default `hash` and `memory` values to empty strings too. Doesn't look like it would be hit in practice, but if it was I don't see why we'd want to crowd the text and tag prefix trees with dedicated nodes for each key. A search for "hash" or "data" would also return all these keys that actually don't have a value.

## Type of Change

- [ ] Documentation update

## AI Assistance

- [ ] No AI assistance

- [] **I can explain every line of this diff and how it interacts with the rest of the codebase, without asking an AI tool.**

I don't know the code base well. Had claude explain to me the problem and worked with it on a solution. I'm not going to oversell myself here.

## Breaking Changes

N/A

## Test Coverage

- [X] I added/updated unit tests

## Checklist

- [ X] My code follows the project's style guidelines
- [X ] I have performed a self-review of my code
- [ X] I have added tests that prove my fix/feature works
- [ X] New and existing tests pass locally
- [ N/A] I have updated documentation if needed

### DIFF
--- a/mem0/vector_stores/valkey.py
+++ b/mem0/vector_stores/valkey.py
@@ -299,20 +299,15 @@ class ValkeyDB(VectorStoreBase):
                 # Create the key for the hash
                 key = f"{self.prefix}:{id}"
 
-                # Check for required fields and provide defaults if missing
-                if "data" not in payload:
-                    # Silently use default value for missing 'data' field
-                    pass
-
-                # Ensure created_at is present
-                if "created_at" not in payload:
+                # Default created_at when missing or None to current time
+                if not payload.get("created_at"):
                     payload["created_at"] = datetime.now(pytz.timezone(self.timezone)).isoformat()
 
                 # Prepare the hash data
                 hash_data = {
                     "memory_id": id,
-                    "hash": payload.get("hash", f"hash_{id}"),  # Use a default hash if not provided
-                    "memory": payload.get("data", f"data_{id}"),  # Use a default data if not provided
+                    "hash": payload.get("hash", ""),
+                    "memory": payload.get("data", ""),
                     "created_at": int(datetime.fromisoformat(payload["created_at"]).timestamp()),
                     "embedding": np.array(vector, dtype=np.float32).tobytes(),
                 }
@@ -499,20 +494,15 @@ class ValkeyDB(VectorStoreBase):
         try:
             key = f"{self.prefix}:{vector_id}"
 
-            # Check for required fields and provide defaults if missing
-            if "data" not in payload:
-                # Silently use default value for missing 'data' field
-                pass
-
-            # Ensure created_at is present
-            if "created_at" not in payload:
+            # Default created_at when missing or None to current time
+            if not payload.get("created_at"):
                 payload["created_at"] = datetime.now(pytz.timezone(self.timezone)).isoformat()
 
             # Prepare the hash data
             hash_data = {
                 "memory_id": vector_id,
-                "hash": payload.get("hash", f"hash_{vector_id}"),  # Use a default hash if not provided
-                "memory": payload.get("data", f"data_{vector_id}"),  # Use a default data if not provided
+                "hash": payload.get("hash", ""),
+                "memory": payload.get("data", ""),
                 "created_at": int(datetime.fromisoformat(payload["created_at"]).timestamp()),
             }
 
@@ -521,7 +511,7 @@ class ValkeyDB(VectorStoreBase):
                 hash_data["embedding"] = np.array(vector, dtype=np.float32).tobytes()
 
             # Add updated_at if available
-            if "updated_at" in payload:
+            if payload.get("updated_at"):
                 hash_data["updated_at"] = int(datetime.fromisoformat(payload["updated_at"]).timestamp())
 
             # Add optional fields
--- a/mem0/vector_stores/valkey.py
+++ b/mem0/vector_stores/valkey.py
@@ -306,8 +306,8 @@ class ValkeyDB(VectorStoreBase):
                 # Prepare the hash data
                 hash_data = {
                     "memory_id": id,
-                    "hash": payload.get("hash", ""),
-                    "memory": payload.get("data", ""),
+                    "hash": payload.get("hash", f"hash_{id}"),  # Use a default hash if not provided
+                    "memory": payload.get("data", f"data_{id}"),  # Use a default data if not provided
                     "created_at": int(datetime.fromisoformat(payload["created_at"]).timestamp()),
                     "embedding": np.array(vector, dtype=np.float32).tobytes(),
                 }
@@ -501,8 +501,8 @@ class ValkeyDB(VectorStoreBase):
             # Prepare the hash data
             hash_data = {
                 "memory_id": vector_id,
-                "hash": payload.get("hash", ""),
-                "memory": payload.get("data", ""),
+                "hash": payload.get("hash", f"hash_{vector_id}"),  # Use a default hash if not provided
+                "memory": payload.get("data", f"data_{vector_id}"),  # Use a default data if not provided
                 "created_at": int(datetime.fromisoformat(payload["created_at"]).timestamp()),
             }
 
--- a/tests/vector_stores/test_valkey.py
+++ b/tests/vector_stores/test_valkey.py
@@ -158,19 +158,32 @@ def test_insert_handles_missing_created_at(valkey_db, mock_valkey_client):
     assert "created_at" in kwargs["mapping"]  # Should be added automatically
 
 
-def test_insert_and_update_with_empty_timestamps(valkey_db, mock_valkey_client):
+def test_insert_and_update_with_none_timestamps(valkey_db, mock_valkey_client):
     """Regression: None created_at/updated_at must not crash insert() or update().
 
-    Realistic (infer=True) consolidation feeds back payloads with None timestamps;
-    fromisoformat(None) previously would be hit and raise a TypeError.
+    created_at falls back to now, and a None updated_at is omitted from the hash
+    rather than passed to fromisoformat().
     """
     vector = np.random.rand(1536).tolist()
-    payload = {"hash": "h", "data": "d", "created_at": None, "updated_at": None}
 
-    valkey_db.insert(vectors=[vector], payloads=[payload], ids=["id1"])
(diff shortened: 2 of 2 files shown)


mem0/vector_stores/valkey.py
```diff
@@ -299,20 +299,15 @@ def insert(self, vectors: list, payloads: list = None, ids: list = None):
...
+                # Default created_at when missing or None to current time
+                if not payload.get("created_at"):
                     payload["created_at"] = datetime.now(pytz.timezone(self.timezone)).isoformat()
 
                 # Prepare the hash data
                 hash_data = {
                     "memory_id": id,
-                    "hash": payload.get("hash", f"hash_{id}"),  # Use a default hash if not provided
-                    "memory": payload.get("data", f"data_{id}"),  # Use a default data if not provided
+                    "hash": payload.get("hash", ""),
```
kartik-mem0: This is a separate change from the timestamp guard, and it is reachable. Entity payloads never set `hash` (`mem0/memory/main.py:638`), so every entity row's stored hash goes from `hash_<id>` to an empty string. Nothing in the repo reads that field today, so it breaks nothing, but a behavior change should not ride along in a crash fix.

Please revert both defaults and keep this PR to the `None` guard.

tests/vector_stores/test_valkey.py
```diff
@@ -158,6 +158,21 @@ def test_insert_handles_missing_created_at(valkey_db, mock_valkey_client):
...
+    Realistic (infer=True) consolidation feeds back payloads with None timestamps;
+    fromisoformat(None) previously would be hit and raise a TypeError.
+    """
+    vector = np.random.rand(1536).tolist()
+    payload = {"hash": "h", "data": "d", "created_at": None, "updated_at": None}
+
+    valkey_db.insert(vectors=[vector], payloads=[payload], ids=["id1"])
+    valkey_db.update(vector_id="id1", vector=vector, payload=payload)
+
+    assert mock_valkey_client.hset.call_count == 2
```
kartik-mem0: `hset.call_count == 2` passes whether or not the mapping is right, so it proves no crash rather than correct behavior.

Assert the payload instead: `"updated_at" not in kwargs["mapping"]` and `created_at` is an int. That fails if someone later writes a `None` timestamp through in a different way.

tests/vector_stores/test_valkey.py
```diff
@@ -158,6 +158,21 @@ def test_insert_handles_missing_created_at(valkey_db, mock_valkey_client):
     assert "created_at" in kwargs["mapping"]  # Should be added automatically
 
 
+def test_insert_and_update_with_empty_timestamps(valkey_db, mock_valkey_client):
+    """Regression: None created_at/updated_at must not crash insert() or update().
```
kartik-mem0: The docstring says an `infer=True` consolidation round trip, but the test calls `insert()` and `update()` directly with a hand built payload. Worth trimming it to what the test does so the next reader does not go hunting for the consolidation path.

kartik-mem0 (commented):
## Changes requested

Good root cause placement: guarding at the `fromisoformat` call sites protects every caller of `insert()` and `update()`, not just the entity path that hit it. The blocker is the unrelated `hash`/`memory` default change riding along, which does change what gets written.

**To unblock:**
1. `mem0/vector_stores/valkey.py:309` - revert the `hash` and `memory` defaults, they are not part of this bug

Optional: `tests/vector_stores/test_valkey.py:173` could assert the mapping rather than the call count, and the docstring describes a round trip the test does not do.

- **Problem is real:** `valkey.py:806` in `list()` writes `"updated_at": result.payload.get("updated_at")`, so an entity that was never updated gets an explicit `None`. `search()` and `get()` only set the key when present, so `list()` is the odd one out. That payload flows `main.py:589` -> `_upsert_entity` -> `entity_store.update()` at `main.py:630`, and `valkey.py:524` tests `"updated_at" in payload`, which is presence not value. Result: `TypeError: fromisoformat: argument must be str`, swallowed by the `except` at `main.py:650` as a warning, so entity updates silently stop persisting.
- **Tests / red-green:** 58 passed on the PR. Swapping in `main`'s `valkey.py`: 1 failed, 57 passed, and the single failure is the new test with the same `TypeError`. Restored: 58 passed. Only the new test flips, so it bites exactly this change.
- **Coverage of the root cause:** all three `fromisoformat` sites in `valkey.py` are now guarded. `redis.py:128,250,252` already had equivalent guards, and no other vector store has an unguarded one.
- **CI:** `check`, `gate`, `label` green. Vercel red, as on every fork PR, not merge blocking.
- **Security:** cleared. No new input handling, no credentials, no injection surface.
- **Bloat (ponytail):** net negative, the diff deletes two dead `if "data" not in payload: pass` blocks. Good.
- **Not verified:** no live Valkey run. The sandbox has no valkey lane and `search()`/`list()` need a Valkey build with the search module, so this rests on the unit suite and the revert above.

</details>

The diagnosis here is the hard part and you got it in the right place, one guard covering every caller instead of one per call site. Push when ready and I will re-review.

Worth knowing: #7074 fixes the same three lines more narrowly and its test asserts the written mapping rather than the call count. This PR is the better base, the assertion from that one is worth borrowing.

mem0/vector_stores/valkey.py
BCathcart: Agree this can be separate. Do you support changing it in a separate PR?

tests/vector_stores/test_valkey.py
BCathcart: That was calling out the current path that hit the crash. Removed to just say what the test is doing

kartik-mem0 (commented):
Re-reviewed at `2ffc27cf`. Content-identical rebase of the previously reviewed head. Fresh tests and lint pass against the new SHA. LGTM still holds.

kartik-mem0 (approved):

Issue #7003, opened 2026-08-16
TS OSS: entity store shares the memory store's table on supabase, baidu and vectorize, so entities are returned as memories


shashiKundur1:
### Component

TypeScript SDK (`mem0-ts`)

### Description

For the `supabase`, `baidu` and `vectorize` providers, the entity store and the memory store resolve to the **same physical table**. Every entity extracted during `add()` is written into the user's memory collection, and `search()` and `getAll()` hand those rows back as if they were memories.

`getEntityStore()` (`mem0-ts/src/oss/src/memory/index.ts:318`) derives the entity store by cloning the vector store config and renaming one key:

```ts
const entityConfig: VectorStoreConfig = {
  ...this.config.vectorStore.config,
  collectionName: entityCollectionName,
};
if (entityProvider === "memory") { /* dbPath rewrite */ }
if (entityProvider === "databricks") { /* tableName rewrite */ }
```

But those three providers never read `collectionName`:

| provider | key that names the storage | file |
|---|---|---|
| supabase | `config.tableName` | `vector_stores/supabase.ts:99` |
| baidu | `config.tableName` | `vector_stores/baidu.ts:120` |
| vectorize | `config.indexName` | `vector_stores/vectorize.ts:30` |

So `collectionName` becomes `memories_entities`, which nothing reads, while `tableName` stays `memories`, which is what actually gets opened.

This is the config the docs ship for Supabase (`docs/components/vectordbs/dbs/supabase.mdx:43-51`), so it is the documented happy path, not an unusual setup:

```ts
provider: "supabase",
config: { collectionName: "memories", tableName: "memories", ... }
```

The Python SDK is not affected. Its Supabase store keys off `collection_name` (`mem0/vector_stores/supabase.py:46`), which `_entity_collection_name()` already renames.

#### What I Ran

A jest test that mocks the three store modules and records the config each instance is constructed with, then builds a `Memory` with the documented Supabase config and calls `getEntityStore()`. No network and no Supabase SDK. Run 10 times, identical every time.

#### What I Saw

```
SUPABASE COLLECTIONS: ["memories","memories_entities"]
SUPABASE TABLES:      ["memories","memories"]
VECTORIZE INDEXES:    ["memories","memories"]
```

`collectionName` is namespaced correctly. `tableName` and `indexName`, the keys these providers actually use, are not. Both stores open `memories`.

#### Why This Is a Bug

[...]

Issue #6798, opened 2026-08-04
mem0-ts test suite POSTs telemetry to the production PostHog project, and it is what makes the TS CI flaky


shashiKundur1:
Both telemetry modules default `MEM0_TELEMETRY` to true, and nothing in the jest config turns it off, so the TypeScript test suite sends real events to the production PostHog project on every run.

`src/oss/src/utils/telemetry.ts` and `src/client/telemetry.ts`:

```ts
let MEM0_TELEMETRY = true;
MEM0_TELEMETRY = process?.env?.MEM0_TELEMETRY?.toLowerCase() === "false" ? false : true;
```

`jest.setup.ts` only sets the SDK version, and `jest.config.js` sets no env.

## Two problems

**1. Live analytics get test traffic.** Every `new Memory()`, `add()`, `search()`, `getAll()` and `reset()` in the suite fires an event. `distinct_id` is the durable per-machine id from `~/.mem0/config.json`, so each CI runner and each contributor laptop shows up as a person profile emitting `mem0.init` / `mem0.add` / `mem0.search` / `mem0.reset`. Hot-path events carry `sample_rate: 0.1`, so any dashboard extrapolating by `1/sample_rate` multiplies the synthetic traffic by ten. `ts-sdk-ci.yml` runs the matrix on Node 20 and Node 22, so it doubles again per PR, plus every push to main.

**2. It is the cause of the flaky TS CI.** `captureClientEvent` is awaited on the critical path, so each event blocks on a network round trip of roughly five seconds. That is why `memory.embedding-action.test.ts` and `memory.rerank.test.ts` time out, and why the test that fails changes between runs.

Measured on a clean checkout of current `main`, same machine, back to back:

| | `pnpm run test` |
|---|---|
| today | 198s, 4 failures, failing test varies |
| with `MEM0_TELEMETRY=false` | 10.1s, 0 failures |

Both of those files pass once telemetry is off.

## The fix is not one line

Setting `MEM0_TELEMETRY=false` in `jest.setup.ts` on its own breaks 18 tests, in `src/oss/tests/telemetry-sampling.test.ts` (14) and `src/client/tests/telemetry-aliasing.test.ts` (4), because those two suites legitimately assert on the sender. They need an explicit opt-in, and the two need different ones since one imports the telemetry module dynamically and the other statically.

I have this working and will open a PR. Test-only, no production source touched.

## Separate, not in that PR

[...]

PR #7004, opened 2026-08-16
fix(ts-oss): stop the entity store from sharing the memory store's table

shashiKundur1:
## Linked Issue

Superseded by #7005, which carries the same branch and the same commits. Closing keyword removed on purpose so this one stays closed and does not reopen alongside it. Tracking issue is #7003.

This one was marked ready for review before #7003 had `accepted`, which fired the PR Gate and closed it. That was the gate working correctly. The work continues in #7005.

## Description

`getEntityStore()` derived the entity store by cloning the vector store config and renaming `collectionName`, with special-cases for `memory` (dbPath) and `databricks` (tableName). Four providers end up sharing storage with the memory store, because none of them resolve their location from `collectionName`:

| provider | key that names the storage |
|---|---|
| supabase | `config.tableName` (`vector_stores/supabase.ts:99`) |
| baidu | `config.tableName` (`vector_stores/baidu.ts:120`) |
| vectorize | `config.indexName` (`vector_stores/vectorize.ts:30`) |
| memory | `config.dbPath` (`vector_stores/memory.ts:42`), fixed `vectors` table |

So `collectionName` became `memories_entities`, which nothing reads, and `tableName` stayed `memories`, which is what actually gets opened. The entity store and the memory store were the same table.

Every entity extracted in Phase 7 of `addToVectorStore` was inserted into the user's memory collection with a populated `data` field, so those rows passed every downstream filter and came back out of `search()` and `getAll()` as memories. They also fed back into the extraction prompt through Phase 1, and `_existingEntitiesByText()` could match a real memory as an entity and overwrite its embedding through `entityStore.update()`.

`docs/components/vectordbs/dbs/supabase.mdx:43-51` ships `{ collectionName: "memories", tableName: "memories" }`, so this was the documented path rather than an unusual config.

The `memory` provider is affected too, by a different route. Its separation depends on `basePath.replace(/\.db$/, "_entities.db")`, which silently returns the path unchanged for anything not ending in `.db`. Since `MemoryVectorStore` keys only on `dbPath` and always uses the `vectors` table, a `dbPath` like `/data/store.sqlite` puts both stores in the same file and the same table. The default path ends in `.db`, which is why local runs and the whole test suite never showed it.

The `databricks` branch already handled the table case for one provider. This generalises it: namespace `tableName` and `indexName` whenever the config carries them, and append `_entities` to a non-`.db` `dbPath`. `:memory:` is left alone, since appending to SQLite's in-memory magic value would turn it into a real file, and two connections to `:memory:` already get separate private databases. Databricks keeps its exact current naming and so do `.db` paths, so no existing entity store is orphaned. The tests pin both.

Python is unaffected. Its Supabase store keys off `collection_name`, which `_entity_collection_name()` already renames.

## Type of Change

- [ ] Documentation update

## AI Assistance

- [ ] No AI assistance
- [ ] AI-generated (an agent wrote most or all of this diff)

I used Claude to sweep the vector store implementations for which config key each one reads, which is what surfaced the mismatch. The diff and the tests I wrote and ran myself.

## Breaking Changes

Existing supabase, baidu and vectorize deployments will start writing entities to `<table>_entities` instead of the memory table, and a `memory` deployment on a non-`.db` path will start writing to `<path>_entities`. New entity rows go to the new location and the memory table stops growing with them. Rows already written into the memory table stay there, and they are identifiable by an `entityType` field in the payload. Happy to add a note to the docs or a cleanup snippet if you want one.

Nothing changes for databricks or for a `.db` path, so no entity store that is currently separated gets orphaned.

## Test Coverage

`mem0-ts/src/oss/tests/memory.entity-store-isolation.test.ts`, seven tests. Four cover the affected providers. Three are controls pinning databricks naming, `.db` naming, and `:memory:`, so nothing that works today changes.

The store modules are mocked and the test records the config each instance is built with, so it needs no network and no provider SDK.

- With the fix: 7 passed.
- Against a pristine `origin/main` copy of the file, same tests: 4 failed, 3 passed. The three that pass are the controls.
- Full suite: 1542 passed. The two suites that fail (`memory.embedding-action`, `memory.prototype-key-embedding`) fail identically on a pristine `origin/main`, and they fail on timeouts, which is the telemetry issue in #6798.
- `prettier --check` clean. `tsc --noEmit` reports the same 30 pre-existing errors before and after, none in the files this touches.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -323,10 +323,14 @@ export class Memory {
         ...this.config.vectorStore.config,
         collectionName: entityCollectionName,
       };
-      // For file-based stores (memory/SQLite), always use a separate DB for entities
+      // For file-based stores (memory/SQLite), always use a separate DB for entities.
+      // The suffix swap only fires on a .db path, so anything else needs appending
+      // or both stores open the same file and the same table.
       if (entityProvider === "memory") {
         const basePath = entityConfig.dbPath || getDefaultVectorStoreDbPath();
-        entityConfig.dbPath = basePath.replace(/\.db$/, "_entities.db");
+        entityConfig.dbPath = basePath.endsWith(".db")
+          ? basePath.replace(/\.db$/, "_entities.db")
+          : `${basePath}_entities`;
       }
       // Some stores name their physical storage with tableName or indexName and
       // never read collectionName, so renaming the collection alone would leave
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -328,9 +328,11 @@ export class Memory {
       // or both stores open the same file and the same table.
       if (entityProvider === "memory") {
         const basePath = entityConfig.dbPath || getDefaultVectorStoreDbPath();
-        entityConfig.dbPath = basePath.endsWith(".db")
-          ? basePath.replace(/\.db$/, "_entities.db")
-          : `${basePath}_entities`;
+        if (basePath !== ":memory:") {
+          entityConfig.dbPath = basePath.endsWith(".db")
+            ? basePath.replace(/\.db$/, "_entities.db")
+            : `${basePath}_entities`;
+        }
       }
       // Some stores name their physical storage with tableName or indexName and
       // never read collectionName, so renaming the collection alone would leave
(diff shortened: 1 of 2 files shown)


shashiKundur1:
My mistake, I marked this ready for review before #7003 had been labelled, so the gate did exactly what it says on the tin.

Reopened as #7005 rather than reopening this one, so it can sit as a draft until #7003 is triaged instead of bouncing off the gate again. Same branch, same three commits, no change to the diff.

Issue #6606, opened 2026-07-26
bug(ts-oss): entity store shares the main store's db when memory dbPath isn't *.db


ac12644:
### Component
TypeScript SDK (OSS), `memory` vector store provider — `mem0-ts/src/oss/src/memory/index.ts`

### What happens

If you use the local `memory` (SQLite) provider and set a custom `dbPath` that doesn't end in `.db`, the entity store ends up sharing the same SQLite file as the main store. Extracted entities then show up in `getAll()` and `search()` as if they were memories. In search they can even rank above the real memory.

It only affects the `memory` provider, and only when `dbPath` isn't `*.db` — `.sqlite`, `.sqlite3`, or an extensionless path all trigger it. The default path (`~/.mem0/vector_store.db`) is fine because it ends in `.db`.

### Why

`getEntityStore()` derives the entity DB path like this:

```ts
entityConfig.dbPath = basePath.replace(/\.db$/, "_entities.db");
```

The regex is anchored to `.db$`, so for anything else it's a no-op and `entityConfig.dbPath === basePath`. Both stores open the same file. The comment right above it says "always use a separate DB for entities", so the intent is clearly universal separation — the regex just doesn't deliver it.

Python isn't affected — it separates the entity store by `collection_name` (`mem0/memory/main.py`), not by rewriting the path.

### Repro

```ts
const m = new Memory({
  vectorStore: { provider: "memory", config: { dbPath: "/tmp/store.sqlite" } },
  // ...embedder + llm
});

await m.add("Alice met Bob in Berlin", { userId: "u" });
const all = await m.getAll({ filters: { user_id: "u" } });
// all.results includes "Bob", "Berlin", ... alongside the real memory
```

Verified end to end without any live API (the collision is purely in the path + sqlite layer): two real `MemoryVectorStore` instances at the derived paths, insert a memory in one and an entity record in the other, and the main store's `list()` returns both. Only one db file gets created on disk (`store.sqlite`), never a separate `store_entities.sqlite`. With `.db` you get two files and clean separation.

The reason it wasn't caught: every existing entity test uses `dbPath: ":memory:"`, and two `:memory:` connections are separate DBs by SQLite semantics, so the no-op path never bites there.

### Fix

Insert `_entities` before the extension for any path shape (and leave `:memory:` alone). PR up: forwards to a separate file for `.db`/`.sqlite`/`.sqlite3`/extensionless alike, with a regression test across all of them.

kartik-mem0:
[...]

PR #7005, opened 2026-08-16
fix(ts-oss): stop the entity store from sharing the memory store's table

### LINKED ISSUE #7003: TS OSS: entity store shares the memory store's table on supabase, baidu and vectorize, so entities are returned as memories

shashiKundur1:
## Linked Issue

Closes #7003

Replaces #7004. I marked that one ready for review before #7003 had been labelled, and the PR Gate closed it as designed. Same branch, same three commits, nothing about the change itself is different.

To be explicit about why a second PR exists at all, since #6974 has since documented that a gated PR should be left to reopen itself: on 16 Aug the gate had no reopen job, and a closed PR cannot be converted back to draft (`gh pr ready --undo` and the GraphQL mutation both refuse). Reopening #7004 by hand would only have re-fired the gate and closed it again. A new draft was the only way to keep the work visible without fighting the gate. #7004 has now been unlinked from #7003 so it stays closed rather than reopening as a duplicate of this one when the label lands.

Keeping this one a draft until #7003 carries `accepted`, then marking it ready.

## Description

`getEntityStore()` derived the entity store by cloning the vector store config and renaming `collectionName`, with special-cases for `memory` (dbPath) and `databricks` (tableName). Four providers end up sharing storage with the memory store, because none of them resolve their location from `collectionName`:

| provider | key that names the storage |
|---|---|
| supabase | `config.tableName` (`vector_stores/supabase.ts:99`) |
| baidu | `config.tableName` (`vector_stores/baidu.ts:120`) |
| vectorize | `config.indexName` (`vector_stores/vectorize.ts:30`) |
| memory | `config.dbPath` (`vector_stores/memory.ts:42`), fixed `vectors` table |

So `collectionName` became `memories_entities`, which nothing reads, and `tableName` stayed `memories`, which is what actually gets opened. The entity store and the memory store were the same table.

Every entity extracted in Phase 7 of `addToVectorStore` was inserted into the user's memory collection with a populated `data` field, so those rows passed every downstream filter and came back out of `search()` and `getAll()` as memories. They also fed back into the extraction prompt through Phase 1, and `_existingEntitiesByText()` could match a real memory as an entity and overwrite its embedding through `entityStore.update()`.

`docs/components/vectordbs/dbs/supabase.mdx:43-51` ships `{ collectionName: "memories", tableName: "memories" }`, so this was the documented path rather than an unusual config.

The `memory` provider is affected too, by a different route. Its separation depends on `basePath.replace(/\.db$/, "_entities.db")`, which silently returns the path unchanged for anything not ending in `.db`. Since `MemoryVectorStore` keys only on `dbPath` and always uses the `vectors` table, a `dbPath` like `/data/store.sqlite` puts both stores in the same file and the same table. The default path ends in `.db`, which is why local runs and the whole test suite never showed it.

The `databricks` branch already handled the table case for one provider. This generalises it: namespace `tableName` and `indexName` whenever the config carries them, and append `_entities` to a non-`.db` `dbPath`. `:memory:` is left alone, since appending to SQLite's in-memory magic value would turn it into a real file, and two connections to `:memory:` already get separate private databases. Databricks keeps its exact current naming and so do `.db` paths, so no existing entity store is orphaned. The tests pin both.

Python is unaffected. Its Supabase store keys off `collection_name`, which `_entity_collection_name()` already renames.

## Type of Change

- [ ] Documentation update

## AI Assistance

- [ ] No AI assistance
- [ ] AI-generated (an agent wrote most or all of this diff)

I used Claude to sweep the vector store implementations for which config key each one reads, which is what surfaced the mismatch. The diff and the tests I wrote and ran myself.

## Breaking Changes

Existing supabase, baidu and vectorize deployments will start writing entities to `<table>_entities` instead of the memory table, and a `memory` deployment on a non-`.db` path will start writing to `<path>_entities`. New entity rows go to the new location and the memory table stops growing with them. Rows already written into the memory table stay there, and they are identifiable by an `entityType` field in the payload. Happy to add a note to the docs or a cleanup snippet if you want one.

Nothing changes for databricks or for a `.db` path, so no entity store that is currently separated gets orphaned.

## Test Coverage

`mem0-ts/src/oss/tests/memory.entity-store-isolation.test.ts`, seven tests. Four cover the affected providers. Three are controls pinning databricks naming, `.db` naming, and `:memory:`, so nothing that works today changes.

The store modules are mocked and the test records the config each instance is built with, so it needs no network and no provider SDK.

- With the fix: 7 passed.
- Against a pristine `origin/main` copy of the file, same tests: 4 failed, 3 passed. The three that pass are the controls.
- Full suite: 1542 passed. The two suites that fail (`memory.embedding-action`, `memory.prototype-key-embedding`) fail identically on a pristine `origin/main`, and they fail on timeouts, which is the telemetry issue in #6798.
- `prettier --check` clean. `tsc --noEmit` reports the same 30 pre-existing errors before and after, none in the files this touches.

## Checklist

### DIFF
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -323,10 +323,14 @@ export class Memory {
         ...this.config.vectorStore.config,
         collectionName: entityCollectionName,
       };
-      // For file-based stores (memory/SQLite), always use a separate DB for entities
+      // For file-based stores (memory/SQLite), always use a separate DB for entities.
+      // The suffix swap only fires on a .db path, so anything else needs appending
+      // or both stores open the same file and the same table.
       if (entityProvider === "memory") {
         const basePath = entityConfig.dbPath || getDefaultVectorStoreDbPath();
-        entityConfig.dbPath = basePath.replace(/\.db$/, "_entities.db");
+        entityConfig.dbPath = basePath.endsWith(".db")
+          ? basePath.replace(/\.db$/, "_entities.db")
+          : `${basePath}_entities`;
       }
       // Some stores name their physical storage with tableName or indexName and
       // never read collectionName, so renaming the collection alone would leave
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -328,9 +328,11 @@ export class Memory {
       // or both stores open the same file and the same table.
       if (entityProvider === "memory") {
         const basePath = entityConfig.dbPath || getDefaultVectorStoreDbPath();
-        entityConfig.dbPath = basePath.endsWith(".db")
-          ? basePath.replace(/\.db$/, "_entities.db")
-          : `${basePath}_entities`;
+        if (basePath !== ":memory:") {
+          entityConfig.dbPath = basePath.endsWith(".db")
+            ? basePath.replace(/\.db$/, "_entities.db")
+            : `${basePath}_entities`;
+        }
       }
       // Some stores name their physical storage with tableName or indexName and
       // never read collectionName, so renaming the collection alone would leave
--- a/mem0-ts/src/oss/src/memory/index.ts
+++ b/mem0-ts/src/oss/src/memory/index.ts
@@ -328,10 +328,16 @@ export class Memory {
         const basePath = entityConfig.dbPath || getDefaultVectorStoreDbPath();
         entityConfig.dbPath = basePath.replace(/\.db$/, "_entities.db");
       }
-      if (entityProvider === "databricks") {
-        entityConfig.tableName = entityConfig.tableName
-          ? `${entityConfig.tableName}_entities`
-          : entityCollectionName;
+      // Some stores name their physical storage with tableName or indexName and
+      // never read collectionName, so renaming the collection alone would leave
+      // the entity store pointing at the memory store's own table.
+      if (entityConfig.tableName) {
+        entityConfig.tableName = `${entityConfig.tableName}_entities`;
+      } else if (entityProvider === "databricks") {
+        entityConfig.tableName = entityCollectionName;
+      }
+      if (entityConfig.indexName) {
+        entityConfig.indexName = `${entityConfig.indexName}_entities`;
       }
       this._entityStore = VectorStoreFactory.create(
         entityProvider,
--- a/mem0-ts/src/oss/tests/memory.entity-store-isolation.test.ts
+++ b/mem0-ts/src/oss/tests/memory.entity-store-isolation.test.ts
@@ -41,6 +41,9 @@ jest.mock("../src/vector_stores/vectorize", () => ({
 jest.mock("../src/vector_stores/databricks", () => ({
   DatabricksVectorStore: jest.fn().mockImplementation(mockStoreStub),
 }));
+jest.mock("../src/vector_stores/memory", () => ({
+  MemoryVectorStore: jest.fn().mockImplementation(mockStoreStub),
+}));
 jest.mock("../src/utils/telemetry", () => ({
   captureClientEvent: jest.fn(),
   captureEvent: jest.fn(),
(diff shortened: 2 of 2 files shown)


shashiKundur1:
Flagging an overlap I should have caught before opening this.

#6607 has been open since 26 July and fixes the `memory` provider dbPath case, linked to #6606. Same reasoning as the dbPath hunk in here, down to leaving `:memory:` alone so each connection keeps its own db. @ac12644 got there first and I did not see it when I filed #7003, so credit for that half is theirs.

The two branches do collide, so only one of them can carry that hunk:

```
git merge-tree --write-tree --name-only fix/ts-entity-store-shared-table pr6607
CONFLICT (content): mem0-ts/src/oss/src/memory/index.ts
```

What is still new in here is the other three providers. Supabase and Baidu name their table with `tableName`, Vectorize names its index with `indexName`, and none of the three ever read `collectionName`. So on those the entity store opens the memory store's own table and every extracted entity lands in the user's memory collection. #6606 and #6607 are scoped to `memory` only, so nothing in them touches that.

Two ways to split it, both fine by me:

- drop the dbPath hunk from here and rebase on #6607, leaving this PR as purely the tableName and indexName fix
- or keep this one as the full fix across all four providers, if one PR is easier than two

No preference, whichever is less work to review. Say which and I will push it.

PR #7013, opened 2026-08-17
fix(vector-store): drop Valkey documents when deleting the index

### LINKED ISSUE #6995: Flush Valkey documents on collection reset
### Component

Python SDK

### Summary

On the Valkey vector store, `Memory.reset()` (and `delete_col()`) drop the search
index but never delete the stored memory documents. valkey-search's `FT.DROPINDEX`
has no `DD` (delete-documents) flag, so the underlying `mem0:<collection>:*` hashes
survive. Because `reset()` then recreates the index over the same key prefix,
valkey-search backfills those surviving hashes and re-indexes them — so the
"cleared" memories reappear in `search()` / `get_all()`. `reset()` is effectively a
no-op on the vector data, and repeated reset/add cycles cause the keyspace to grow
without bound.

### Steps to Reproduce

```python
from mem0 import Memory

config = {
    "vector_store": {
        "provider": "valkey",
        "config": {
            "valkey_url": "redis://localhost:6379",  # Valkey with the search module loaded
            "collection_name": "mem0",
            "embedding_model_dims": 1536,
        },
    },
}

m.add("I love hiking in the mountains", user_id="alice")

m.reset()  # expected: wipe everything, clean slate

print(m.get_all(user_id="alice"))  # BUG: the memory is still there
```

### Expected Behavior

After `reset()`, the collection is empty: `get_all(user_id="alice")` returns no
memories, and no `mem0:mem0:*` document keys remain in Valkey.

### Actual Behavior

`get_all()` still returns the memory added before `reset()`, and the document
hashes are still present in the keyspace. No exception is raised — it is a silent
correctness bug.

```
# after m.reset()
$ valkey-cli --scan --pattern 'mem0:mem0:*' | wc -l
1          # expected 0; the pre-reset hash survived and was re-indexed
```

### Environment

- mem0 version: 2.0.18
- OS: Linux (self-hosted Valkey + valkey-search module)

### What I Ran

The script in "Steps to Reproduce" above, against a self-hosted Valkey with the
valkey-search module loaded. Also observed at scale in a benchmark that resets the
store between runs and reuses the persisted `valkey_data` volume.

### What I Saw

- `get_all(user_id="alice")` returns the memory that was added *before* `reset()`.
- The document keys persist across resets:

  ```
  $ valkey-cli --scan --pattern 'mem0:mem0:*' | wc -l
  1          # non-zero immediately after reset()
  ```

- In the benchmark, hashes accumulated every reset/run until the `valkey_data`
  volume held ~3,400 stale documents. Retrieval quality collapsed to recall@1 =
  0.000 because the unlabeled stale copies from prior runs dominated the top ranks.

No traceback — the failure is silent.

### Why This Is a Bug

`reset()` is documented as wiping all memories (`Memory.reset` — "Reset the memory
store … Deletes the vector store collection"), and every other backend clears the
data. The Redis store even relies on `FT.DROPINDEX ... DD` to delete documents. On
Valkey the documents are never deleted, so the contract is broken.

Root cause in `mem0/vector_stores/valkey.py`:

- `delete_col()` → `_drop_index()` runs bare `FT.DROPINDEX <name>` (no `DD`).
- `reset()` = `delete_col()` + `_create_index()`; the recreate uses
  `FT.CREATE ... ON HASH PREFIX 1 mem0:<collection>`, which re-adopts and
  re-indexes the surviving `mem0:<collection>:*` hashes.

### What I Ruled Out

- Not a valkey-search retrieval-quality problem: the store returns results fine;
  the results are just stale documents that were never deleted.
- Not normal CRUD: `delete()` / `delete_all()` use `client.delete(key)` and do
  remove the hashes. Only the index-teardown paths (`reset()` / `delete_col()`)
  orphan documents.
- Not a client caching / stale-handle issue: the surviving keys are visible
  directly via `valkey-cli --scan 'mem0:<collection>:*'` after `reset()`.

### AI Assistance

chjnett:
Fixes embedchain/embedchain#6995

## Summary
`ValkeyDB._drop_index` ran bare `FT.DROPINDEX <name>` (no `DD` flag), so `reset()`/`delete_col()` removed the index but left the `mem0:<collection>:*` hashes in the keyspace. Since `reset()` recreates the index over the same key prefix, valkey-search re-adopted and re-indexed those surviving hashes — "cleared" memories reappeared in `search()`/`get_all()`, making `reset()` a no-op on the vector data.

Added the `DD` flag so `FT.DROPINDEX` deletes the documents too, matching the destructive-operation contract every other backend honors (Redis's `index.delete()` removes documents as well).

## Test
- Updated `test_delete_col` to assert the call is `FT.DROPINDEX <name> DD`.
- All 57 `tests/vector_stores/test_valkey.py` tests pass locally.

### DIFF
--- a/mem0/vector_stores/valkey.py
+++ b/mem0/vector_stores/valkey.py
@@ -694,14 +694,20 @@ class ValkeyDB(VectorStoreBase):
 
     def _drop_index(self, collection_name, log_level="error"):
         """
-        Drop an index by name using the documented FT.DROPINDEX command.
+        Drop an index by name using the documented FT.DROPINDEX command,
+        including the ``DD`` flag so the underlying documents are deleted too.
+
+        Without ``DD``, ``FT.DROPINDEX`` only removes the index; the hashes
+        remain in the keyspace and are re-adopted the next time an index is
+        created over the same key prefix (e.g. ``reset()`` recreates the index
+        over ``mem0:<collection>``, resurrecting supposedly-cleared memories).
 
         Args:
             collection_name (str): Name of the index to drop.
             log_level (str): Logging level for missing index ("silent", "info", "error").
         """
         try:
-            self.client.execute_command("FT.DROPINDEX", collection_name)
+            self.client.execute_command("FT.DROPINDEX", collection_name, "DD")
             logger.info(f"Successfully deleted index {collection_name}")
             return True
         except ResponseError as e:
--- a/tests/vector_stores/test_valkey.py
+++ b/tests/vector_stores/test_valkey.py
@@ -311,7 +311,11 @@ def test_delete_col(valkey_db, mock_valkey_client):
     assert result is True
 
     # Check that execute_command was called with the correct command
-    mock_valkey_client.execute_command.assert_called_once_with("FT.DROPINDEX", "test_collection")
+    # The DD flag ensures the underlying documents are deleted too, so a
+    # later reset() over the same key prefix doesn't resurrect them.
+    mock_valkey_client.execute_command.assert_called_once_with(
+        "FT.DROPINDEX", "test_collection", "DD"
+    )
 
     # Test error handling - real errors should still raise
     mock_valkey_client.execute_command.side_effect = ResponseError("Error dropping index")


mem0/vector_stores/valkey.py
```diff
@@ -694,14 +694,20 @@ def list_cols(self):
...
+        created over the same key prefix (e.g. ``reset()`` recreates the index
+        over ``mem0:<collection>``, resurrecting supposedly-cleared memories).
 
         Args:
             collection_name (str): Name of the index to drop.
             log_level (str): Logging level for missing index ("silent", "info", "error").
         """
         try:
-            self.client.execute_command("FT.DROPINDEX", collection_name)
+            self.client.execute_command("FT.DROPINDEX", collection_name, "DD")
```
BCathcart, line 710: Valkey search doesn't support this flag: https://valkey.io/commands/ft.dropindex/

PR #7021, opened 2026-08-18
feat(integrations): strands-mem0 | Mem0 as a native Strands MemoryStore

Himanshu-Sangshetti:
> **Draft.** Code, tests, docs, and CI/CD wiring are complete and the review below is addressed. Kept as draft pending maintainer sign-off on the workflow additions (see [Workflow changes](#workflow-changes)).

## What this adds

[Mem0](https://mem0.ai) as a native [Strands Agents](https://github.com/strands-agents/sdk-python) **`MemoryStore`** (`integrations/strands-mem0/`). Where the `mem0_memory` tool is called explicitly by the model, this plugs into the **agent loop** via `MemoryManager`: it searches Mem0 and injects context every turn, and writes memory back on extraction.

```python
from strands import Agent
from strands.memory import MemoryManager
from strands_mem0 import Mem0MemoryStore

store = Mem0MemoryStore(user_id="alex", extraction=True)
agent = Agent(memory_manager=MemoryManager(stores=[store]))

agent("Remember I use Neovim and deploy on Fridays.")   # writes
agent("What editor do I use?")                           # recalls
```

| Method | Maps to | Role |
|---|---|---|
| `search(query)` | `mem0.search(query, filters={...}, top_k=...)` | recall + inject every turn |
| `add(content)` | `mem0.add(content, infer=False)` | verbatim single-fact write |
| `add_messages(messages)` | `mem0.add(rendered_turns, infer=True)` | Mem0 server-side extraction |

Because `add_messages` is implemented, enabling `extraction` routes conversation turns to Mem0's own extraction pipeline, no extra client-side model call.

## Review addressed

The first review (thanks @kartik-mem0) found the two core paths did not work against the shipped `mem0ai` SDK, and the fakes were permissive enough to hide it. All of it is fixed and re-verified against the real SDK signatures (`mem0ai` 2.0.18, `strands-agents` 1.53.0):

- **OSS search** now uses `filters=` + `top_k=` like the platform (the branch collapsed); the old top-level/`limit=` call raised on real `Memory.search`.
- **`add_messages`** renders Strands content blocks (`{"text": ...}`) to text before sending; the raw blocks were being dropped by `parse_vision_messages` (a silent no-op). Pure tool-use turns render empty and are skipped.
- **Tests** now carry the real SDK signatures (strict OSS `search`, fixed-signature OSS `add`), so a rejected call shape fails in CI; added regression coverage for both blockers.
- Reject async Mem0 clients; guard `app_id` on OSS; resolve the lazy client inside `asyncio.to_thread` so its blocking construction never runs on the loop; pin `mem0ai>=1.0.0`; drop the unused config class and redundant injection seam; `__version__` from `importlib.metadata`.

## Packaging and release

Matches the monorepo standard (`mem0ai`, `mem0-cli`): plain `hatchling` with a static version, and a root-level CD workflow published on a `strands-mem0-v*` tag via PyPI **trusted publishing** (OIDC, no stored token).

- CI: `.github/workflows/strands-mem0-checks.yml` (ruff + mypy + pytest matrix 3.10-3.12 + hatch build), wired into `ci-gate.yml`.
- CD: `.github/workflows/strands-mem0-cd.yml`, registered in the `release.yml` router.
- Docs: `docs/integrations/strands.mdx` + `docs.json` + `llms.txt` (coverage check passes).
- AGENTS: rows added to `.github/AGENTS.md` (CI + CD) and `integrations/AGENTS.md`.

## Testing

- [x] **Unit** 27 tests, fakes shaped to the real SDK signatures, with regression tests for both blockers
- [x] **Lint / types / build** ruff clean, mypy clean, `hatch build` produces a clean 0.1.0 wheel + sdist
- [x] **Signature check** call shapes verified against the real `mem0.Memory` / `mem0.MemoryClient` (OSS search takes `filters`+`top_k`; OSS `add` has no `app_id` / no `**kwargs`)
- [ ] **Live agent-loop E2E** against a real Mem0 backend (search quota permitting), tracked separately

## Workflow changes

Per root `AGENTS.md`, workflow changes need maintainer approval. This PR adds two package workflows (`strands-mem0-checks.yml`, `strands-mem0-cd.yml`) and registers them in `ci-gate.yml` and `release.yml`, the standard wiring for a new package. Flagging for explicit sign-off.


integrations/strands-mem0/python/src/strands_mem0/client.py
```diff
@@ -0,0 +1,152 @@
...
+        """Semantic recall, scoped to the store's entity.
+
+        Routes ``scope`` through ``filters`` on the platform client and as
+        top-level kwargs on the OSS ``Memory``, then normalizes the two response
+        shapes to a plain list of memory dicts.
+        """
+        if self.is_platform:
+            response = self.mem0.search(query, filters=dict(scope), top_k=top_k)
+        else:
+            response = self.mem0.search(query, limit=top_k, **scope)
```
kartik-mem0: **Blocker: this raises on every OSS search.**

`mem0.Memory.search` is keyword-only after `query` and explicitly rejects top-level entity params:

```
Memory.search(self, query: str, *, top_k: int = 20, filters: dict | None = None, threshold: float = 0.1, ...)
```

```
>>> Memory.search(None, "q", limit=5, user_id="alex")
ValueError: Top-level entity parameters frozenset({'user_id'}) are not supported in search().
```

`limit=` is also not a parameter; it lands in `**kwargs` and is ignored, so `top_k` would stay at 20 even if the scope were accepted.

The lazy fix is also the correct one: both backends now take `filters=` and `top_k=`, so the branch disappears.

```python
def search_memories(self, query: str, scope: dict[str, str], top_k: int) -> list[dict[str, Any]]:
    return _extract_results(self.mem0.search(query, filters=dict(scope), top_k=top_k))
```

That removes the branch, and with it the "one real asymmetry" premise in this module's docstring and in the PR description, which is no longer true of current `mem0ai`.

integrations/strands-mem0/python/src/strands_mem0/store.py
```diff
@@ -0,0 +1,223 @@
...
+            self._merge_metadata(metadata),
+        )
+
+    async def add_messages(self, messages: list[Message], context: AddMessagesContext | None = None) -> Any:
+        """Ingest raw conversation turns for Mem0 server-side extraction (``infer=True``).
+
+        This is the sink the manager uses for extraction when no client-side
+        extractor is configured, so facts are distilled by Mem0 itself.
+        """
+        payload = [dict(message) for message in messages]
```
kartik-mem0: **Blocker: this ships a shape mem0 throws away.**

A Strands `Message.content` is `list[ContentBlock]`, and a text block is `{"text": "..."}` with no `type` key. `dict(message)` preserves that, and mem0's `parse_vision_messages` keeps only parts where `part.get("type") == "text"`:

```
payload handed to mem0.add: [{'role': 'user', 'content': [{'text': 'I use Neovim and deploy on Fridays'}]},
                             {'role': 'assistant', 'content': [{'text': 'Got it.'}]}]
mem0 parses that as       : []
```

On OSS that is a silent no-op (`{"results": []}`, no error). On platform the list is POSTed to `/v3/memories/add/` unchanged.

This is the sink the whole "server-side extraction, no extra model call" story rests on, so it is worth fixing at the root rather than at the call site. Strands already ships the rendering in `strands/memory/extraction/model_extractor.py:119`:

```python
payload = []
for message in messages:
    text = "\n".join(part for block in message["content"] if (part := block.get("text", "")))
    if text:
        payload.append({"role": message["role"], "content": text})
if not payload:
    return None
return await asyncio.to_thread(self.client.store_messages, payload, self.scope)
```

Dropping the empties matters: a turn that is pure `toolUse`/`toolResult` renders to nothing, and passing `content: ""` is not the same as passing nothing.

integrations/strands-mem0/python/tests/test_client.py
```diff
@@ -0,0 +1,152 @@
...
+
+def test_oss_search_uses_top_level_scope():
+    """OSS search passes scope as top-level kwargs with `limit`."""
+    fake = FakeMemory()
+    client = Mem0ServiceClient(client=fake)
+
+    client.search_memories("q", {"user_id": "alex"}, 5)
+
+    _, kwargs = fake.search_calls[0]
+    assert kwargs["user_id"] == "alex"
```
kartik-mem0: This test is what let the blocker in `client.py:138` through. `FakeMemory.search(self, query, **kwargs)` accepts literally any keyword, so the assertion certifies the one call shape the real `Memory.search` rejects.

Give the fakes the real signatures and the test fails before the fix, passes after:

```python
class FakeMemory:
    def search(self, query, *, top_k=20, filters=None, threshold=0.1, **kwargs):
        if kwargs.keys() & {"user_id", "agent_id", "run_id"}:
            raise ValueError("Top-level entity parameters are not supported in search().")
        self.search_calls.append((query, {"top_k": top_k, "filters": filters}))
        return {"results": []}
```

Same for `FakeMemoryClient.add`: `Memory.add` has a fixed signature with no `**kwargs`, so `test_oss_writes_omit_source` currently proves nothing about the OSS path either. `app_id` in scope would be a `TypeError` there and this fake would happily accept it.

integrations/strands-mem0/python/tests/test_store.py
```diff
@@ -0,0 +1,232 @@
...
+
+    await store.add("fact")
+
+    mock_client.store_memory.assert_called_once_with("fact", {"user_id": "alex"}, {"team": "growth"})
+
+
+async def test_add_messages_routes_raw_turns(mock_client):
+    """add_messages() forwards the raw turns and scope for server-side extraction."""
+    messages = [
+        {"role": "user", "content": "I love hiking"},
```
kartik-mem0: `content` is a string here, but Strands never hands `add_messages` a string. The real shape is `{"role": "user", "content": [{"text": "I love hiking"}]}`, and that shape is what mem0 discards (see the comment on `store.py:202`).

Swapping these two lines to the real `Message` shape turns this into the regression test that catches the bug:

```python
messages = [
    {"role": "user", "content": [{"text": "I love hiking"}]},
    {"role": "assistant", "content": [{"text": "Noted!"}]},
]
...
mock_client.store_messages.assert_called_once_with(
    [{"role": "user", "content": "I love hiking"}, {"role": "assistant", "content": "Noted!"}],
    {"user_id": "alex"},
)
```

Worth one more covering a `toolUse`-only turn, which should render to nothing and not be sent.

integrations/strands-mem0/python/src/strands_mem0/store.py
```diff
@@ -0,0 +1,223 @@
...
+
+    async def search(self, query: str, options: SearchOptions | None = None) -> list[MemoryEntry]:
+        """Search Mem0 for entries matching ``query``, ordered by relevance."""
+        top_k = options.get("max_search_results") if options is not None else None
+        if top_k is None:
+            top_k = self.max_search_results
+        if top_k is None:
+            top_k = DEFAULT_MAX_SEARCH_RESULTS
+
+        memories = await asyncio.to_thread(self.client.search_memories, query, self.scope, top_k)
```
kartik-mem0: `self.client` is resolved on the event loop, before `to_thread` is entered, so the lazy construction is not actually off-thread.

That matters because `MemoryClient.__init__` ends in `self.user_email = self._validate_api_key()`, a blocking HTTP call on an httpx client with `timeout=300`. The OSS branch is worse: `Memory.from_config(config)` builds embedders and vector stores, and can pull down a model.

Either way the first agent turn blocks the whole loop. One line:

```python
memories = await asyncio.to_thread(lambda: self.client.search_memories(query, self.scope, top_k))
```

Same in `add` and `add_messages`.

integrations/strands-mem0/python/src/strands_mem0/client.py
```diff
@@ -0,0 +1,152 @@
...
+
+from __future__ import annotations
+
+import os
+from typing import Any
+
+# Platform client classes route entity scope through `filters` on search; the OSS
+# `Memory` takes it as top-level kwargs. We detect the platform by class name so we
+# do not have to import (and thus hard-depend on) a specific mem0 submodule here.
+_PLATFORM_CLIENTS = {"MemoryClient", "AsyncMemoryClient"}
```
kartik-mem0: `AsyncMemoryClient.add` and `.search` are coroutine functions:

```
>>> inspect.iscoroutinefunction(AsyncMemoryClient.add)
True
```

So `client=AsyncMemoryClient(...)` passes `_is_platform_client`, then `store_memory` returns an un-awaited coroutine that `asyncio.to_thread` hands straight back. Every write silently does nothing, with a `RuntimeWarning` as the only signal.

Either drop it from the set, or reject it in `__init__` with a message pointing at the sync client. Listing it as supported is the one thing that is worse than either.

integrations/strands-mem0/python/src/strands_mem0/store.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Raises:
+            ValueError: If no entity scope (``user_id`` / ``agent_id`` / ``run_id``
+                / ``app_id``) is provided.
+        """
+        scope = {
+            "user_id": user_id,
+            "agent_id": agent_id,
+            "run_id": run_id,
+            "app_id": app_id,
```
kartik-mem0, line 108: `app_id` is offered here and spread into `mem0.add(**scope)`, but the OSS `Memory.add` signature is fixed:

```
Memory.add(self, messages, *, user_id=None, agent_id=None, run_id=None, metadata=None,
           timestamp=None, expiration_date=None, infer=True, memory_type=None, prompt=None)
```

No `app_id`, no `**kwargs`, so `Mem0MemoryStore(app_id="x", config={...})` is a `TypeError` on the first write. The docstring says "platform only" but nothing enforces it. Cheapest fix is to raise in `__init__` when `app_id` is set alongside `config`, which fails at construction rather than on the first agent turn.

integrations/strands-mem0/python/src/strands_mem0/store.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+
+    user_id: str
+    agent_id: str
+    run_id: str
+    app_id: str
+    metadata: dict[str, Any]
+    api_key: str
+    host: str
+    config: dict[str, Any]
```
kartik-mem0: `yagni`: 26 lines, exported in `__all__`, zero call sites, zero tests. It is a second copy of `__init__`'s signature with nothing keeping the two in sync.

Note that Strands' own vended stores actually consume theirs: `def __init__(self, **store_config: Unpack[TestMemoryStoreConfig])` in `strands/vended_memory_stores/test_memory_store/store.py:113`. Either adopt that pattern so the TypedDict is load-bearing, or delete it and keep the explicit kwargs. Right now it is neither.

integrations/strands-mem0/python/src/strands_mem0/store.py
```diff
@@ -0,0 +1,223 @@
...
+        description: str | None = "Persistent long-term memory backed by Mem0.",
+        max_search_results: int | None = None,
+        writable: bool = True,
+        extraction: Any = None,
+        metadata: dict[str, Any] | None = None,
+        api_key: str | None = None,
+        host: str | None = None,
+        config: dict[str, Any] | None = None,
+        mem0_client: Any | None = None,
+        client: Mem0ServiceClient | None = None,
```
kartik-mem0, line 77: `yagni`: two injection parameters for one need. `Mem0ServiceClient` is public and exported, so anyone with a pre-built raw client can write `Mem0MemoryStore(client=Mem0ServiceClient(client=raw))`, which is exactly what `mem0_client` does internally.

Dropping `mem0_client` removes a constructor arg, a docstring block, an instance attribute, and a line in the lazy property.

integrations/strands-mem0/python/src/strands_mem0/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from strands_mem0.client import Mem0ServiceClient
+from strands_mem0.store import Mem0MemoryStore, Mem0MemoryStoreConfig
+
+__all__ = [
+    "Mem0MemoryStore",
+    "Mem0MemoryStoreConfig",
+    "Mem0ServiceClient",
+]
+
+__version__ = "0.1.0"
```
kartik-mem0: Duplicates `pyproject.toml:7` with nothing enforcing the two match, so the first release bump makes one of them a lie. Stdlib covers it:

```python
from importlib.metadata import version
__version__ = version("strands-mem0")
```

Or just delete it. Nothing in the package reads `__version__`.

integrations/strands-mem0/python/pyproject.toml
```diff
@@ -0,0 +1,89 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+]
+
+# strands-agents>=1.45.0: first release shipping the `strands.memory` module
+# (MemoryStore / MemoryManager) that Mem0MemoryStore implements.
+dependencies = [
+    "strands-agents>=1.45.0",
+    "mem0ai>=0.1.0",
```
kartik-mem0: `mem0ai>=0.1.0` claims compatibility back to a release from years before the API this code calls. The `filters=` / `top_k=` search contract (and the `_reject_top_level_entity_params` guard behind blocker 1) landed in the 2.0.x line; current is 2.0.18.

`mem0ai>=2.0.11` or a 2.x floor you have actually run against. As written, pip is free to resolve something that cannot work, and the failure lands on a user, not on CI.

integrations/strands-mem0/python/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```bash
+pip install hatch
+hatch run test        # pytest (mocked client, no live server)
+hatch run prepare     # format + lint + typecheck + test
+```
+
+## Release
+
+Tag a GitHub release with a `python-v*` tag (e.g. `python-v0.1.0`); the
```
kartik-mem0: This section is stale and will produce a no-op release. It documents `python-v*` and a `publish-python` workflow, but this PR registers `strands-mem0-v*` to `strands-mem0-cd.yml` in `release.yml`.

Tagging `python-v0.1.0` as written here falls through the router's `case` block to:

```
::error::Release tag 'python-v0.1.0' does not match any known package prefix
```

Looks like a leftover from the extension-template layout that `fa3521d` corrected in the workflows but not here. The PR description carries the same stale reference to `publish-python.yml`.

kartik-mem0 (commented):
## Review: 6/10

Real craft here. Ruff, mypy, and the 23 tests are green, the layout matches the sibling integrations, the ci-gate and release-router wiring is correct, and the docstrings are better than most of this repo. The problem is that the two headline paths do not work against the mem0 SDK this repo actually ships, and the unit tests are shaped so they cannot notice.

Everything below was verified by running it, not by reading. `mem0ai` 2.0.18 (this repo's `main`), `strands-agents` 1.52.0.

### Blockers

**1. OSS search raises on every call.** `client.py:138` calls `self.mem0.search(query, limit=top_k, **scope)`. Real signature is `Memory.search(self, query, *, top_k=20, filters=None, ...)`, and it starts with `_reject_top_level_entity_params(kwargs, "search")`.

```
>>> Memory.search(None, "q", limit=5, user_id="alex")
ValueError: Top-level entity parameters frozenset({'user_id'}) are not supported in search(). Use filters={'user_id': '...'} instead.
```

`limit=` is also not a parameter; it gets swallowed by `**kwargs` and `top_k` stays at 20.

The interesting part: the branch is not just wrong, it is unnecessary. Both backends take `filters=` and `top_k=` now, so the whole platform/OSS split collapses to one line, which also invalidates the "one real asymmetry" premise in the `client.py` module docstring and in the PR description.

**2. `add_messages` sends a shape mem0 discards.** `store.py:202` does `dict(message)` on a Strands `Message`, whose `content` is `list[ContentBlock]` where a text block is `{"text": "..."}` with no `type` key. mem0's `parse_vision_messages` keeps only parts where `part.get("type") == "text"`, so it drops the message entirely:

```
payload handed to mem0.add: [{'role': 'user', 'content': [{'text': 'I use Neovim and deploy on Fridays'}]}]
mem0 parses that as       : []
```

On OSS that is a silent no-op: `{"results": []}`, no error, no log. On platform the list is POSTed unchanged to `/v3/memories/add/`. This is the feature the PR is built around ("we implement both sinks, so extraction is server-side"), and it is the one path that does not work. Strands already ships the right rendering in `memory/extraction/model_extractor.py:119`; mirror it.

**3. The tests assert the bugs back at the implementation.** `test_oss_search_uses_top_level_scope` passes against a `FakeMemory.search(self, query, **kwargs)` that accepts anything, so it locks in the exact call the real SDK rejects. `test_add_messages_routes_raw_turns` feeds `content: "I love hiking"`, a shape Strands never produces. The PR description notes the E2E run used "a deterministic in-memory stub" for mem0, which is precisely why neither bug surfaced. Fakes need the real signatures.

### High

- `python/README.md:35` documents the wrong release path (`python-v*` / `publish-python`). The router in this PR registers `strands-mem0-v*` to `strands-mem0-cd.yml`. Tagging as documented hits the router's "no known package prefix" arm and publishes nothing.
- `pyproject.toml:32` `mem0ai>=0.1.0` claims compatibility with releases that predate the API this code calls by years.
- `store.py:161-170` builds the client on the event loop. `MemoryClient.__init__` ends in `self._validate_api_key()`, a blocking HTTP call on a 300s-timeout client; `Memory.from_config()` builds embedders and vector stores. Both run before `asyncio.to_thread` is entered.
- `client.py:27` treats `AsyncMemoryClient` as a supported platform client, but its `add`/`search` are coroutine functions. Passing one returns un-awaited coroutines from every write.
- `app_id` is offered as a scope field with no OSS guard; `Memory.add` has no `app_id` parameter, so it is a `TypeError`.

### Repo standards

`integrations/AGENTS.md` "Adding an integration", steps 6 and 7:

- No `docs/integrations/strands.mdx`, no `docs/docs.json` entry, no `docs/llms.txt` entry. `grep -ril strands docs/` returns nothing. All 33 existing integrations have a page.
- `.github/AGENTS.md` got the CD row but not the CI table row (the block at lines 20-25).
- `integrations/AGENTS.md` is untouched: no table row, no "What each one is" bullet.
- Root `AGENTS.md`: "Do NOT modify anything in `.github/workflows/` without explicit maintainer approval." Four workflow files here. Worth an explicit sign-off on the PR.
- No `Closes #<issue>` in the description.

### Complexity (the lazy pass)

- `store.py:50-75`: `Mem0MemoryStoreConfig` is 26 lines, exported, zero call sites, zero tests. Strands' own vended stores consume theirs via `**store_config: Unpack[TestMemoryStoreConfig]`; this one just shadows `__init__` and will drift from it.
- `store.py:104-105`: `mem0_client` and `client` are two injection seams for one need. `Mem0ServiceClient` is public, so `client=Mem0ServiceClient(client=raw)` already covers it.
- `__init__.py:22`: `__version__` duplicates `pyproject.toml:7` with nothing keeping them in sync. `importlib.metadata.version` is stdlib.
- `client.py:138`: fixing blocker 1 deletes the branch.

Net: roughly 45 lines out, and the two paths start working.

### Smaller things

- Test matrix is 3.10-3.12; classifiers advertise 3.13.
- The description does not match the diff: it says hatch-vcs (pyproject pins a static version with plain hatchling), `publish-python.yml` (not in the diff), `twine check` (the workflow does `ls dist/*`), and per-file line counts that are off by 10-20%.

### What I would want before merge

Blockers 1-3 fixed, with the fakes rewritten to the real signatures so they would have failed first. Then the docs page and the two AGENTS.md tables. The rest is nits.

.github/workflows/strands-mem0-checks.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Type-check with mypy
+        working-directory: integrations/strands-mem0/python
+        run: mypy src
+
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
```
kartik-mem0, line 45: Matrix stops at 3.12, but `pyproject.toml` advertises `Programming Language :: Python :: 3.13`. Add `"3.13"` or drop the classifier; right now the package claims support nothing verifies.

.github/AGENTS.md
```diff
@@ -56,6 +56,7 @@ Requiring `CI Gate` also means fork PRs from first-time contributors cannot merg
 | OpenCode Plugin | `opencode-plugin-cd.yml` | `opencode-v*` | npm (`@mem0/opencode-plugin`) |
 | Pi Agent Plugin | `pi-agent-plugin-cd.yml` | `pi-agent-v*` | npm (`@mem0/pi-agent-plugin`) |
 | n8n Node | `n8n-nodes-mem0-cd.yml` | `n8n-nodes-mem0-v*` | npm (`@mem0/n8n-nodes-mem0`) |
+| strands-mem0 | `strands-mem0-cd.yml` | `strands-mem0-v*` | PyPI (`strands-mem0`) |
```
kartik-mem0, line 60: The CD row is here, but the CI table above (lines 20-25, `Workflow | File | Standalone triggers | Runs`) has no `strands-mem0-checks.yml` row. `integrations/AGENTS.md` step 7 asks for both tables, and that file's own integration table and "What each one is" list are untouched too.

Still open from the same checklist, step 6: `docs/integrations/strands.mdx` plus entries in `docs/docs.json` and `docs/llms.txt`. `grep -ril strands docs/` currently returns nothing, and all 33 existing integrations have a page.

integrations/strands-mem0/python/src/strands_mem0/client.py
Himanshu-Sangshetti: Fixed. Both backends now call `search(query, filters=dict(scope), top_k=top_k)`, so the platform/OSS branch (and the "one asymmetry" docstring premise) is gone. `test_oss_search_uses_filters` uses a strict fake and fails on the old top-level/`limit=` shape.

integrations/strands-mem0/python/src/strands_mem0/store.py
Himanshu-Sangshetti: Fixed. `add_messages` renders each turn's text blocks to a string and drops turns that render empty (pure tool-use), mirroring `model_extractor.py`, so mem0 never gets an empty payload. Covered by `test_add_messages_renders_content_blocks` and `test_add_messages_skips_empty_turns`.

integrations/strands-mem0/python/tests/test_client.py
Himanshu-Sangshetti: Fixed. `FakeMemory` now carries the real signatures: keyword-only `search` that rejects top-level entity params, and a fixed `add` with no `**kwargs` (so `source`/`app_id` would TypeError). Regression tests for both blockers were added.

integrations/strands-mem0/python/tests/test_store.py
Himanshu-Sangshetti: Fixed. The test now uses the real `content: [{"text": ...}]` shape and asserts the rendered `content: "..."` output, plus a tool-use-only turn that must render to nothing and not be sent.

integrations/strands-mem0/python/src/strands_mem0/store.py
Himanshu-Sangshetti: Fixed. `self.client` is resolved inside the `to_thread` lambda in `search`/`add`/`add_messages`, so the blocking construction runs off the event loop.

integrations/strands-mem0/python/src/strands_mem0/client.py
Himanshu-Sangshetti: Fixed. Dropped `AsyncMemoryClient` from the platform set and reject async clients in `__init__` (coroutine `add`/`search`) with a message pointing at the sync client. Test: `test_async_client_is_rejected`.

integrations/strands-mem0/python/src/strands_mem0/store.py
Himanshu-Sangshetti: Fixed, two guards. `__init__` rejects `app_id` + `config` at construction (the case you named); `Mem0ServiceClient` rejects `app_id` on any OSS backend, which also covers an injected OSS client. Tests for both.

integrations/strands-mem0/python/src/strands_mem0/store.py
Himanshu-Sangshetti: Removed. Deleted the TypedDict and kept the explicit kwargs, and dropped it from `__all__`.

integrations/strands-mem0/python/src/strands_mem0/store.py
Himanshu-Sangshetti: Removed. Raw clients now go through `Mem0ServiceClient(client=raw)` as you noted; `mem0_client` (arg, docstring, attr, property line) is gone.

integrations/strands-mem0/python/src/strands_mem0/__init__.py
Himanshu-Sangshetti: Fixed. `__version__` now comes from `importlib.metadata.version("strands-mem0")` with a source-tree fallback.

integrations/strands-mem0/python/pyproject.toml
Himanshu-Sangshetti: Fixed. Pinned `mem0ai>=2.0.11`, the floor for the `filters=`/`top_k=` contract and the `_reject_top_level_entity_params` guard this code relies on.

integrations/strands-mem0/python/README.md
Himanshu-Sangshetti: Fixed. README now documents `strands-mem0-v*` / `strands-mem0-cd.yml`. The PR description carried the same stale `publish-python.yml` reference and has been rewritten too.

.github/workflows/strands-mem0-checks.yml
Himanshu-Sangshetti: Fixed. Dropped the `3.13` classifier to match the 3.10-3.12 test matrix.

.github/AGENTS.md
Himanshu-Sangshetti: Done. Added the CI row to `.github/AGENTS.md`, the table row + "what each one is" bullet to `integrations/AGENTS.md`, and `docs/integrations/strands.mdx` + `docs.json` + `docs/llms.txt` (the llms.txt coverage check passes).

Himanshu-Sangshetti:
Thanks for the review

**Blockers**
- OSS search: collapsed to `filters=` + `top_k=` on both backends (branch gone).
- `add_messages`: renders content blocks to text and skips empty turns, so mem0 no longer gets an empty payload.
- Tests: fakes now match the real SDK signatures, with regression tests that fail on the old call shapes.

**High**
- Async clients rejected; `app_id` guarded at construction and in the client; lazy client resolved inside `to_thread`; `mem0ai>=2.0.11`.

**Cleanup**
- Dropped the unused `Mem0MemoryStoreConfig` and the redundant `mem0_client` seam; `__version__` via `importlib.metadata`; dropped the 3.13 classifier.

**Repo standards**
- Added `docs/integrations/strands.mdx` + `docs.json` + `llms.txt` (coverage passes), the CI row in `.github/AGENTS.md`, and the row + bullet in `integrations/AGENTS.md`; fixed the README release path and refreshed the PR description.

integrations/strands-mem0/python/tests/test_client.py
```diff
@@ -0,0 +1,214 @@
...
+    ``search`` is keyword-only and rejects top-level entity params; ``add`` has a
+    fixed signature with no ``**kwargs`` (so ``source`` or ``app_id`` is a
+    ``TypeError``), exactly like the shipped SDK.
+    """
+
+    def __init__(self):
+        self.add_calls = []
+        self.search_calls = []
+
+    def add(
```
kartik-mem0, line 42: This fake hand-mirrors `mem0.Memory.add`'s real signature, which is exactly the thing that has to stay in sync and exactly what the two blockers in the last round were. If Mem0 renames a keyword or makes one positional, every test here still passes and the integration breaks in production.

`mem0ai` is a hard dependency (`pyproject.toml:31`) and CI already installs it via `pip install -e ".[dev]"`, so the real signature is importable at test time. A drift guard needs no server and no network:

```python
import inspect
from mem0 import Memory, MemoryClient

def test_call_shapes_match_the_real_sdk():
    inspect.signature(Memory.add).bind(
        None, "fact", metadata=None, infer=False, user_id="u"
    )
    inspect.signature(Memory.search).bind(None, "q", filters={"user_id": "u"}, top_k=5)
    inspect.signature(MemoryClient.add).bind(
        None, "fact", metadata=None, infer=False, source="STRANDS", user_id="u"
    )
```

I ran this shape against `mem0ai` 2.0.18 while reviewing and it passes, and it fails loudly on `source=` / `app_id=` for the OSS client, which is the guard you want. Not blocking, but it converts the fakes from a description of the SDK into a check on it.

.github/workflows/strands-mem0-checks.yml
```diff
@@ -0,0 +1,82 @@
...
+  workflow_call:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
```
kartik-mem0, line 21: No `cache:`/`cache-dependency-path` on any of the three `setup-python` steps here (or the two in the CD workflow). `integrations/AGENTS.md` step 3 names `cache-dependency-path` explicitly, and every sibling does it, e.g. `n8n-nodes-mem0-checks.yml:29-30`.

As it stands five jobs each do a cold `pip install -e ".[dev]"` pulling `mem0ai` plus `strands-agents` and their transitive deps:

```yaml
        uses: actions/setup-python@v5
        with:
          python-version: '3.12'
          cache: 'pip'
          cache-dependency-path: integrations/strands-mem0/python/pyproject.toml
```

Minor, but it is a documented convention and the diff is three lines per job.

integrations/strands-mem0/python/src/strands_mem0/store.py
```diff
@@ -0,0 +1,222 @@
...
+        *,
+        user_id: str | None = None,
+        agent_id: str | None = None,
+        run_id: str | None = None,
+        app_id: str | None = None,
+        name: str = "mem0",
+        description: str | None = "Persistent long-term memory backed by Mem0.",
+        max_search_results: int | None = None,
+        writable: bool = True,
+        extraction: Any = None,
```
kartik-mem0, line 72: `Any` here is looser than the protocol, which types this as `ExtractionConfig | bool | None` (`strands/memory/types.py`). The docstring below already names the real type, so this is just the annotation lagging behind it.

`Any` silences mypy on the one argument users are most likely to get wrong (passing a raw dict, or a model instead of an `ExtractionConfig`), and `warn_return_any` is already on in `pyproject.toml`, so the rest of the file is getting checked. Importing `ExtractionConfig` from `strands.memory.types` and using the union restores that.

kartik-mem0 (commented):
Re-reviewed at `f29061c`. Both blockers from the last pass are fixed, and I checked the fixes against the installed SDKs rather than taking the description's word for it: `Memory.add`, `Memory.search`, `MemoryClient.add` and `MemoryClient.search` all accept the exact call shapes in `client.py`, and `Memory.add(source=...)` / `Memory.add(app_id=...)` are both rejected, which proves the platform-only guards are load-bearing. `python/README.md`'s release section now names the right tag prefix and workflow.

This is a new package, so there is no revert-to-`main` red-green available. Instead I re-introduced each old blocker as a one-line mutation: both bite, and the OSS one fails with the real SDK's own error text. Restored, 28 pass.

Repo fit checks out against `integrations/AGENTS.md` steps 1-7, and the checks workflow correctly has no `pull_request` trigger.

Three inline notes below, none of them blocking. Four files under `.github/workflows/` means this needs maintainer sign-off before merge, which you already flagged yourself.

<details>
<summary>Evidence (run at <code>f29061ccf040f1bb910f2af6c267ab6526b564b8</code>)</summary>

**Signature check** against `mem0ai` 2.0.18 / `strands-agents` 1.53.0:

```
OK    Memory.add   (store_memory)
OK    Memory.add   (store_messages)
OK    Memory.search(search_memories)
OK    MemoryClient.add   (platform)
OK    MemoryClient.search(platform)
correctly rejected: Memory.add(source=...) -> unexpected keyword argument 'source'
correctly rejected: Memory.add(app_id=...) -> unexpected keyword argument 'app_id'
```

**Mutation proof** (both round-1 blockers re-introduced one at a time):

```
client.py:162 -> self.mem0.search(query, limit=top_k, **scope)
  FAILED tests/test_client.py::test_platform_search_uses_filters - KeyError: 'f...
  FAILED tests/test_client.py::test_oss_search_uses_filters - ValueError: Top-l...
  2 failed, 26 passed

store.py:184 -> text = message.get("content")
  FAILED tests/test_store.py::test_add_messages_renders_content_blocks
  FAILED tests/test_store.py::test_add_messages_skips_empty_turns
  2 failed, 26 passed

restored -> 28 passed, 82 warnings in 0.15s   (git status clean)
```

**CI:** 46 checks, 41 pass, 5 skipping (TypeScript SDK, n8n node, Mintlify Deployment), 0 fail. All five `strands-mem0` jobs green.

**Checked and cleared, not findings:**

- Platform `add` passing scope at the top level rather than in `filters`. `mem0/client/types.py:18` says the v3 API rejects top-level identity fields, but `MemoryClient.add`'s own docstring lists them under `**kwargs`, there is no `ENTITY_PARAMS` guard on `add` (only on `search`/`get_all`), `_prepare_payload` forwards them to the top level of the body, and `docs/platform/quickstart.mdx:78` plus `docs/platform/features/direct-import.mdx:22` document exactly this shape, cURL included. The `types.py` docstring is the outlier.
- `DEFAULT_MAX_SEARCH_RESULTS` plus the local fallback in `store.py`. This is the Strands house pattern, not duplication: the SDK's own vended stores each declare their own module-level default (both 10, against the manager's 3) and the constant is not exported. The `is None` form here is the more correct-on-zero choice.
- `add_messages` returning `None` on an empty payload. `strands/memory/extraction/coordinator.py:281` discards the return value.
- `_to_entry` omitting `store_name`. `memory_manager.py:357` sets it.

**Minor, not worth an inline comment:** the description says "27 tests" (28), says it pins `mem0ai>=1.0.0` (actually `>=2.0.11`), and still opens with "**Draft.**" though the PR is not a draft. No `Closes #N` either.

</details>

kartik-mem0 (approved):

PR #7027, opened 2026-08-18
feat(integrations): dsh-mem0 — Mem0 as a native DeepSeek Harness plugin

Himanshu-Sangshetti:
## What this adds

[Mem0](https://mem0.ai) long-term memory as a native [DeepSeek Harness](https://github.com/deepseek-ai/deepseek-harness) (Cordis) plugin.

DeepSeek Harness makes memory a plugin slot. This fills it with Mem0: unlike the local/file-based memory plugins in the ecosystem, Mem0 is a managed backend (server-side extraction, semantic dedup and conflict resolution) and the same memory bank is reusable across Harness, Claude Code, Codex, and other agents.

The plugin is a thin adapter. Mem0 and the harness already exist; this solders them together.

```mermaid
flowchart TB
    A["mem0ai SDK<br/><i>managed memory, owned by Mem0</i>"]
    P["dsh-mem0<br/>apply(ctx) then ctx.tools.register(...)"]
    H["DeepSeek Harness<br/><i>the agent, loaded via cordis.yml</i>"]
    P -->|"calls"| A
    H -->|"loads + drives"| P
```

It exports `apply(ctx, config)` with `inject = ['tools']` and registers two tools:

| Tool | Maps to | Notes |
|---|---|---|
| `search_memory` | `client.search(query, { filters: { user_id } })` | platform search requires scope inside `filters`, not top-level |
| `add_memory` | `client.add([...], { user_id, source })` | tagged `source="DEEPSEEK_HARNESS"` for telemetry attribution |

## Written against the real rc API

The published preview API differs from the tutorial's simplified example; this was written against the installed types (`@deepseek-ai/cordis` 4.0.1, `@deepseek-ai/dsh-tools` 0.0.1-rc.1, `mem0ai` 3.1.6):

- `mem0ai` is a **named** import (`import { MemoryClient }`), not default.
- `defineTool` takes a **flat `parameters` map** with per-property `required: true`, plus an `output.schema` and a `render()` returning content blocks, and `execute` returns the schema's value.
- Platform `search` rejects top-level `user_id`; scope goes in `filters`.

## Testing

- [x] **Typecheck** clean against the real `@deepseek-ai` + `mem0ai` types
- [x] **Build** produces `dist/index.{js,d.ts}`
- [x] **Live e2e** against `api.mem0.ai`: mounts the plugin, both tools register, `add_memory` writes (server-side extraction distills the fact), and `search_memory` recalls it (relevance score 0.68, semantic 0.82 / bm25 0.88)
- [x] **Live `dsh web` session** and demo video (manual)

## Telemetry

Writes are tagged `source="DEEPSEEK_HARNESS"`. For that to surface by name in analytics (rather than bucketing into `OTHERS`), `DEEPSEEK_HARNESS` must be added to the backend's `KNOWN_EVENT_SOURCES` allowlist, a one-line platform change matching the existing `ZAPIER` / `STRANDS` sources.

## Scope

Tools-only, the confirmed plugin surface. Auto-capture (store turns with no explicit tool call) and auto-recall (inject memory into the prompt at assembly) are deferred until the harness session/assembly event API is confirmed.

## Publishing

When ready: `npm publish` as `dsh-mem0`, then a catalog entry in the community `awesome-deepseek-harness` list.


integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  /** Optional Mem0 host override, e.g. a self-hosted or local sandbox URL. */
+  host?: string;
+}
+
+export function apply(ctx: Context, config: Config): void {
+  const apiKey = config.apiKey ?? process.env.MEM0_API_KEY;
+  if (!apiKey) {
+    throw new Error("dsh-mem0: set config.apiKey or the MEM0_API_KEY env var");
+  }
+  const userId = config.userId;
```
kartik-mem0, line 74: One userId is frozen at mount time, so the whole harness install shares a single memory bucket: add an optional userId tool param plus agent or run scoping like integrations/pi-agent-plugin/src/memory/scoping.ts already does.

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        "Search the user's long-term Mem0 memory for facts relevant to a query.",
+      parameters: {
+        query: { type: "string", description: "What to recall.", required: true },
+      },
+      output: {
+        schema: { type: "string" },
+        render: (_args, value) => [{ type: "text", text: value }],
+      },
+      async execute({ query }) {
+        const results = await client.search(query, {
```
kartik-mem0: Search has no limit, so every recall pulls the platform default: expose a limit param like integrations/openclaw/tools/memory-search.ts does.

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+          filters: { user_id: userId },
+        });
+        return JSON.stringify(results, null, 2);
```
kartik-mem0: This dumps the raw search envelope pretty printed straight into the model context: reuse formatMemoryList from integrations/pi-agent-plugin/src/memory/formatting.ts and you cut most of those tokens.

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      },
```
kartik-mem0, line 109: No try/catch here, so one network blip rejects inside the harness tool runner: both sibling plugins catch and return a plain text failure line instead.

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      name: "add_memory",
+      description:
+        "Store a fact in the user's long-term Mem0 memory for later sessions.",
+      parameters: {
+        text: { type: "string", description: "The fact to remember.", required: true },
+      },
+      output: {
+        schema: { type: "string" },
+        render: (_args, value) => [{ type: "text", text: value }],
+      },
```
kartik-mem0: This output block is a byte for byte copy of lines 61 to 64: lift it into one const above and use it in both tools.

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      async execute({ text }) {
+        const result = await client.add([{ role: "user", content: text }], {
+          user_id: userId,
+          source: SOURCE,
```
kartik-mem0: source tags writes only, search sends nothing, and there is no PostHog tool event like openclaw and pi-agent-plugin both send, so this integration stays invisible in analytics apart from adds.

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        });
+        return JSON.stringify(result);
```
kartik-mem0: No cap on tool output: pi-agent-plugin truncates at 200 lines and 50KB in src/memory/tools.ts, copy that guard before a big result floods the context window.

integrations/dsh-mem0/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "dsh-mem0",
```
kartik-mem0: Every other integration publishes under the org scope as @mem0/..., so an unscoped dsh-mem0 name is off pattern here.

integrations/dsh-mem0/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "version": "0.1.0",
+  "description": "Mem0 long-term memory as a native DeepSeek Harness (Cordis) plugin.",
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
```
kartik-mem0: Missing repository.directory: integrations/dsh-mem0, which integrations/AGENTS.md step 2 requires for npm provenance, and missing exports plus publishConfig.access that the sibling packages set.

integrations/dsh-mem0/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "files": [
+    "dist",
+    "README.md",
+    "LICENSE"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "typecheck": "tsc --noEmit"
```
kartik-mem0: No test script and no tests anywhere in this PR, while every other TypeScript integration in the repo runs vitest or jest.

integrations/dsh-mem0/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "dsh",
+    "dsh-plugin",
+    "cordis",
+    "mem0",
+    "memory",
+    "agent"
+  ],
+  "license": "Apache-2.0",
+  "dependencies": {
+    "mem0ai": "^3.0.0"
```
kartik-mem0: openclaw pins mem0ai 3.0.7 and pi-agent-plugin uses ^3.0.7, so ^3.0.0 is looser than the rest of the repo.

integrations/dsh-mem0/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "peerDependencies": {
+    "@deepseek-ai/cordis": "*",
+    "@deepseek-ai/dsh-tools": "*"
+  },
+  "devDependencies": {
+    "@deepseek-ai/cordis": "latest",
+    "@deepseek-ai/dsh-tools": "latest",
```
kartik-mem0: latest makes the build unreproducible, and your own PR body says this API is still moving, so pin cordis 4.0.1 and dsh-tools 0.0.1-rc.1.

integrations/dsh-mem0/README.md
```diff
@@ -0,0 +1,60 @@
...
+      |
+[ DeepSeek Harness ]  <-- the agent, loaded via cordis.yml
+```
+
+## Try it locally
+
+1. Build the plugin:
+   ```sh
+   cd integrations/dsh-mem0
+   npm install
```
kartik-mem0: The repo is pnpm only and never npm, and step 3 just below already uses pnpm, so make these two lines pnpm as well.

integrations/dsh-mem0/cordis.example.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: "@deepseek-ai/dsh-system-prompt"
+- name: "@deepseek-ai/dsh-tools"
+- insert:
+    - id: mem0
+      # Absolute path to the built plugin (run `npm run build` first), or point
+      # at src/index.ts when running through tsx during development.
+      name: "/absolute/path/to/mem0/integrations/dsh-mem0/dist/index.js"
+      config:
+        # apiKey is read from the MEM0_API_KEY env var when omitted here.
+        userId: "hims"
```
kartik-mem0: A personal handle is left in the example, use a placeholder like your-user-id.

integrations/dsh-mem0/.gitignore
```diff
@@ -0,0 +1,8 @@
+node_modules/
+dist/
+*.tsbuildinfo
+.env
+
+# local, uncommitted e2e
+*.local.mjs
+package-lock.json
```
kartik-mem0: Ignoring package-lock.json bakes in the npm flow, commit a pnpm-lock.yaml instead like the other integrations do.

kartik-mem0 (commented):
## Review: 6 / 10

The plugin file itself is genuinely lean and reads well, and the API usage is correct against the real rc. The gap is everything around it: this does not yet meet the repo's own "Adding an integration" checklist, and it is missing the safety rails every sibling plugin already has.

**What is good**
- Small, honest adapter with no invented abstraction, which is the right call for a plugin slot.
- Correct on the two API details that usually trip people up: named MemoryClient import and search scope inside filters.
- README, config table and example yml are clear, and the testing section is specific.

**Blocking: the integration is not registered anywhere**

integrations/AGENTS.md lists seven steps for a new integration and five are unmet, which is why CI Gate went green without a single job touching this package:
- No .github/workflows/dsh-mem0-checks.yml or dsh-mem0-cd.yml.
- Not registered in ci-gate.yml (path filter, call job, gate needs list).
- No tag prefix arm in the release.yml case block.
- No docs/integrations/dsh-mem0.mdx, no docs.json entry, no llms.txt entry.
- No row in integrations/AGENTS.md or the CI/CD tables in .github/AGENTS.md.

**Blocking: zero tests**

Every other TypeScript integration here ships tests. A live e2e you ran by hand is good evidence but it does not run in CI. The filter shape, the config validation and the output formatting are all worth one small vitest file.

**Reuse: stop writing the fourth copy**

You asked whether anything can be shared, and yes:
- formatMemoryList and formatMemoryCompact in pi-agent-plugin/src/memory/formatting.ts do exactly what line 69 is doing badly.
- truncateOutput in pi-agent-plugin/src/memory/tools.ts is the output cap this plugin has none of.
- telemetry.ts is already duplicated near verbatim between openclaw and pi-agent-plugin, same PostHog key and same flush machinery, so adding a third copy here is the moment to raise a shared helper package in a separate issue rather than paste it again.

**Scalability**

The single userId at mount time is the real ceiling. openclaw has session and long term scopes, pi-agent-plugin has project, session and global. Right now every user of a harness install writes into the same bucket, and there is no way to override per call.

**To get this to a 9**: add the CI workflows and registration, one vitest file, error handling and an output cap in both tools, formatted search output instead of raw JSON, a userId tool param, and the docs page.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed. Both tools now take optional `userId`/`agentId`/`runId` params that override the mount-time default per call, falling back to the configured `userId` (`src/scoping.ts` `resolveEntity`, resolved to the snake_case shape the platform expects — inside `filters` for search, top-level for add). Covered by `scoping.test.ts` and the override cases in `apply.test.ts`.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed. `search_memory` takes an optional `limit` that maps to `topK` (default 10). Covered by `apply.test.ts` (`limit: 3` asserts the SDK is called with `topK: 3`).

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed. Search now returns `formatMemoryList` output — `[category] text (age) [mem0:id]` per line — instead of the raw envelope, the same shape as pi-agent. `src/formatting.ts` + `formatting.test.ts`.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed. Both tools wrap `execute` in try/catch and return a plain `<tool> failed: …` line. Confirmed live against the platform: a quota error came back as a text line instead of rejecting into the tool runner. `apply.test.ts` covers both the success and failure paths.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed. Lifted into a single `textOutput` const used by both tools.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Reconsidered this one, and you're right — a third near-verbatim copy is the wrong move. Dropped `telemetry.ts` from this PR (writes still carry `source=DEEPSEEK_HARNESS`) and filed #7065 to extract a shared `@mem0` telemetry helper that openclaw, pi-agent-plugin, and dsh-mem0 can all share. dsh-mem0 regains tool analytics once that lands.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed. `truncateOutput` caps at 200 lines / 50KB (`src/output.ts`), the same guard pi-agent applies in its `tools.ts`. `output.test.ts` covers both the line and byte caps.

integrations/dsh-mem0/package.json
Himanshu-Sangshetti: Fixed. Renamed to `@mem0/dsh-mem0`.

integrations/dsh-mem0/package.json
Himanshu-Sangshetti: Fixed. Added `repository.directory: integrations/dsh-mem0`, `exports`, and `publishConfig.access: public`, matching the sibling packages.

integrations/dsh-mem0/package.json
Himanshu-Sangshetti: Fixed. Added `test`/`test:watch` (vitest) plus a 23-test suite: `scoping`, `formatting`, `output`, and an `apply` suite covering config validation, tool wiring, and the graceful-failure paths.

integrations/dsh-mem0/package.json
Himanshu-Sangshetti: Fixed. Pinned `mem0ai` to `^3.0.7` to match pi-agent.

integrations/dsh-mem0/package.json
Himanshu-Sangshetti: Fixed. Pinned `@deepseek-ai/cordis` `4.0.1` and `@deepseek-ai/dsh-tools` `0.0.1-rc.1`.

integrations/dsh-mem0/README.md
Himanshu-Sangshetti: Fixed. Both lines are `pnpm install` / `pnpm build` now.

integrations/dsh-mem0/cordis.example.yml
Himanshu-Sangshetti: Fixed. Replaced with `your-user-id`.

integrations/dsh-mem0/.gitignore
Himanshu-Sangshetti: Fixed. Committed `pnpm-lock.yaml`; `.gitignore` now only ignores the npm/yarn lockfiles. Also added a scoped `.npmrc` with `auto-install-peers=false`: `dsh-tools` declares the rest of the harness (`dsh-scope`, `dsh-llm`, ...) as peers — some unpublished on npm and all host-provided at runtime — so auto-installing the full peer graph breaks a clean install.

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      output: textOutput,
+      async execute({ text, userId: u, agentId, runId }) {
+        const entity = resolveEntity({ userId: u, agentId, runId }, userId);
+        try {
+          const result = await client.add([{ role: "user", content: text }], {
+            ...entity,
+            source: SOURCE,
+          });
+          const memories = Array.isArray(result) ? result : [];
+          return truncateOutput(formatAddResult(memories));
```
kartik-mem0: **Blocking.** `client.add()` posts to `/v3/memories/add/`, which is asynchronous and returns `{"event_id": "...", "status": "PENDING"}` — not an array of memories ([API reference](https://github.com/mem0ai/mem0/blob/main/docs/api-reference/memory/add-memories.mdx#L60-L72)).

`Array.isArray(result)` is therefore always `false`, `memories` is always `[]`, and `formatAddResult([])` makes every successful write report:

> Stored. No new distinct memory was extracted from that.

The write succeeds; the agent is told it produced nothing. A model that reads that will likely retry the write, or tell the user their memory wasn't saved.

`cli/node/src/output.ts:165-195` handles this response shape already:

```ts
const results = Array.isArray(result) ? result : ((result.results as Record<string, unknown>[]) ?? [result]);
// ... then: if (r.status === "PENDING") -> "Queued, processing in background" + event_id
```

Suggest matching it: report queued-with-`event_id` on `PENDING`, keep the list rendering for the case where the backend does return memories. Also worth a sentence in the `add_memory` description saying extraction is asynchronous, so the model doesn't `add` then immediately `search` and conclude the write failed.

integrations/dsh-mem0/tests/apply.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    const out = await tools.get("search_memory")!.execute({ query: "x" }, {});
+
+    expect(out).toContain("search_memory failed");
+    expect(out).toContain("network down");
+  });
+});
+
+describe("add_memory tool", () => {
+  it("tags the write with source and reports the stored count", async () => {
+    mockAdd.mockResolvedValue([{ id: "m1", memory: "Fact" }]);
```
kartik-mem0: This mock is why the bug above is invisible: `/v3/memories/add/` never returns an array. The real response is `{ event_id: "evt-...", status: "PENDING" }`.

A test whose fixture doesn't match the API it stands in for asserts the implementation back at itself. Once the `PENDING` path is handled, pin this mock to the real shape and assert the queued message — that test then genuinely fails if the handling regresses.

integrations/dsh-mem0/src/formatting.ts
```diff
@@ -0,0 +1,56 @@
...
+  return memories
+    .map((m, i) => `${i + 1}. ${formatMemoryCompact(m)}`)
+    .join("\n");
+}
+
+/**
+ * One-line confirmation for a write. `client.add` returns the memories the
+ * backend extracted from the turn (often zero when nothing new was learned),
+ * so report the count rather than echoing the raw payload.
+ */
```
kartik-mem0, line 49: This docstring states the premise the bug rests on: `client.add` does not return the memories the backend extracted. It returns an `event_id` and `PENDING`; extraction happens after the call returns. Worth correcting alongside the `index.ts` fix so the next reader doesn't inherit the assumption.

integrations/dsh-mem0/src/formatting.ts
```diff
@@ -0,0 +1,56 @@
...
+export function formatAddResult(memories: MemoryLike[]): string {
+  if (memories.length === 0) {
+    return "Stored. No new distinct memory was extracted from that.";
+  }
+  const noun = memories.length === 1 ? "memory" : "memories";
+  const lines = memories
+    .map((m, i) => `${i + 1}. ${formatMemoryCompact(m)}`)
+    .join("\n");
```
kartik-mem0: DRY: this is `formatMemoryList`'s body verbatim, four lines below it.

```ts
const noun = memories.length === 1 ? "memory" : "memories";
return `Stored ${memories.length} ${noun}:\n${formatMemoryList(memories)}`;
```

Same output, one call instead of a second copy of the numbering logic.

integrations/dsh-mem0/src/formatting.ts
```diff
@@ -0,0 +1,56 @@
...
+ * same way across every harness: `[category] text (age) [mem0:id]`. Dumping the
+ * raw search envelope instead would spend most of the tokens on JSON scaffolding.
+ */
+
+export interface MemoryLike {
+  id: string;
+  memory?: string;
+  categories?: string[];
+  createdAt?: Date | string;
+  score?: number;
```
kartik-mem0: `score` is declared but never read — not in `formatMemoryCompact`, not anywhere in `src/`. The sibling `MemoryLike` in `pi-agent-plugin` doesn't carry it either. Delete, or render it (relevance is genuinely useful context for the model when it has to pick between recalled facts).

integrations/dsh-mem0/src/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        ...scopeParams,
+      },
+      output: textOutput,
+      async execute({ text, userId: u, agentId, runId }) {
+        const entity = resolveEntity({ userId: u, agentId, runId }, userId);
+        try {
+          const result = await client.add([{ role: "user", content: text }], {
+            ...entity,
+            source: SOURCE,
+          });
```
kartik-mem0, line 130: `entity` is snake_case, but `client.add`'s `AddMemoryOptions` is camelCase (`userId` / `agentId` / `runId`) — every other caller in this repo passes camelCase (`mem0-ts/src/client/tests/memoryClient.crud.test.ts`, `pi-agent-plugin`). This works only because `camelToSnakeKeys` is a no-op on keys that are already snake_case: an implementation detail, not a contract.

`pi-agent-plugin/src/memory/scoping.ts` handles the same asymmetry explicitly — `resolveSearchFilters` returns snake_case (search puts scope in `filters`, passed through raw), `resolveAddParams` returns camelCase (add takes them top-level, through the converter). Splitting `resolveEntity` the same way keeps the one genuinely surprising thing about this API visible in the code rather than load-bearing by accident.

integrations/dsh-mem0/src/scoping.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface ResolvedEntity {
+  user_id: string;
+  agent_id?: string;
+  run_id?: string;
+}
+
+function clean(value: string | undefined): string | undefined {
+  const trimmed = value?.trim();
+  return trimmed ? trimmed : undefined;
+}
```
kartik-mem0: Minor shrink — `||` already collapses `""` and `"   ".trim()`:

```ts
const clean = (v: string | undefined) => v?.trim() || undefined;
```

Same behaviour, and the existing `scoping.test.ts` blank-string cases still cover it.

docs/integrations/dsh-mem0.mdx
```diff
@@ -0,0 +1,89 @@
...
+- insert:
+    - id: mem0
+      name: "/absolute/path/to/integrations/dsh-mem0/dist/index.js"
+      config:
+        # apiKey is read from MEM0_API_KEY when omitted here.
+        userId: "your-user-id"
+        # host: "http://localhost:8000"   # optional: self-hosted / local sandbox
+```
+
+To run against a self-hosted / local Mem0 instead of the hosted platform, set `config.host` (defaults to `api.mem0.ai`).
```
kartik-mem0: This doesn't work as written. `MemoryClient` calls `${host}/v3/memories/add/` and `${host}/v3/memories/search/`; the self-hosted server in `server/main.py` exposes `/memories` and `/search` with no `/v3` prefix, so `host: "http://localhost:8000"` 404s on both tools.

`host` is a base-URL override for the Platform API (on-prem / sandbox deployments), not a switch to Mem0 OSS. `integrations/strands-mem0/README.md:44-45` draws the line correctly: hosted platform via `api_key`/`host`, self-hosted OSS via a separate `config` dict.

Same claim in `README.md:44`, `README.md:52`, `cordis.example.yml:18`, and the Configuration table's "Self-hosted / sandbox URL" row.

docs/llms.txt
```diff
@@ -257,6 +257,7 @@ If the user is on a pre-current major (Python < 2, TS < 3, or a Platform call st
 - [ChatDev](https://docs.mem0.ai/integrations/chatdev) [Platform]: Use when the user is on ChatDev.
 - [Hermes](https://docs.mem0.ai/integrations/hermes) [Both]: Use when the user is on Hermes.
 - [Pi Agent](https://docs.mem0.ai/integrations/pi-agent) [Platform]: Use when adding persistent memory to Pi Agent with the Mem0 plugin.
+- [DeepSeek Harness](https://docs.mem0.ai/integrations/dsh-mem0) [Both]: Use when adding persistent memory to the DeepSeek Harness (Cordis) agent via the Mem0 plugin.
```
kartik-mem0: `[Both]` should be `[Platform]` — this plugin only ever constructs a `MemoryClient`, which is Platform-only. Pi Agent, the closest sibling on the line above, is tagged `[Platform]` for exactly this reason.

integrations/dsh-mem0/src/output.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const kept = lines.slice(0, MAX_OUTPUT_LINES);
+  let result = kept.join("\n");
+  if (result.length > MAX_OUTPUT_BYTES) {
+    result = result.slice(0, MAX_OUTPUT_BYTES);
+  }
+
+  const dropped = lines.length - kept.length;
+  if (dropped > 0 || text.length > MAX_OUTPUT_BYTES) {
+    result += `\n\n[Output truncated: showing ${kept.length} of ${lines.length} lines]`;
```
kartik-mem0: Nit (inherited from the `pi-agent-plugin` copy, so fix in both or neither): when only the byte cap trips, `kept.length === lines.length` and the notice reads "showing 12 of 12 lines" while the text has in fact been cut mid-line. Worth distinguishing the two truncation reasons in the message.

kartik-mem0 (commented):
## Review

Structurally this is one of the cleanest integrations added to `integrations/` in a while. The workflow trio (`dsh-mem0-checks.yml`, `dsh-mem0-cd.yml`, the `ci-gate.yml` filter + `needs` entry, the `release.yml` tag arm) is a faithful copy of the `pi-agent-plugin` pattern with only the names swapped, the docs are registered in all three required places (`docs.json`, `llms.txt`, the `.mdx`), and `AGENTS.md` rows were updated on both sides. The `src/` split mirrors `pi-agent-plugin/src/memory/` almost file for file. Every step of [Adding an integration](https://github.com/mem0ai/mem0/blob/main/integrations/AGENTS.md#adding-an-integration) is done.

One blocking correctness issue, then a handful of small ones.

### Blocking: `add_memory` always reports failure to extract

`client.add()` hits `POST /v3/memories/add/`, which is **asynchronous** and returns `{"event_id": "...", "status": "PENDING"}` — see [`docs/api-reference/memory/add-memories.mdx`](https://github.com/mem0ai/mem0/blob/main/docs/api-reference/memory/add-memories.mdx#L60-L72). It never returns an array of extracted memories.

So at `src/index.ts:134`, `Array.isArray(result)` is always `false`, `memories` is always `[]`, and every successful write returns:

> `Stored. No new distinct memory was extracted from that.`

The write does land — but the agent is told it didn't, on every single call. The `formatAddResult` list branch is unreachable in production. The unit test passes only because its mock returns a shape the API never produces.

`cli/node/src/output.ts:165-195` is the reference handling for this response: `Array.isArray(result) ? result : (result.results ?? [result])`, then branch on `status === "PENDING"` and surface the `event_id`. Worth matching, and worth adding a line to the `add_memory` description telling the model extraction is asynchronous, so it doesn't `add` then immediately `search` and conclude the write failed.

### Also

- `client.add` is given snake_case entity keys. The SDK's `AddMemoryOptions` is camelCase; this only works because `camelToSnakeKeys` happens to be a no-op on already-snake keys. `pi-agent-plugin/src/memory/scoping.ts` splits the two shapes deliberately (`resolveSearchFilters` snake, `resolveAddParams` camel) — worth matching rather than relying on that.
- The `host` / self-hosted claim in the docs, README and `cordis.example.yml` isn't accurate. `MemoryClient` talks to `/v3/*`; the self-hosted `server/` exposes `/memories` and `/search`. `host` is a Platform base-URL override, not an OSS switch (cf. `integrations/strands-mem0/README.md:44-45`).
- A few small DRY / dead-code trims noted inline.

Nothing here touches the structure — the shape of the package is right. Comments inline.

integrations/dsh-mem0/.gitignore
```diff
@@ -0,0 +1,12 @@
+node_modules/
+dist/
+*.tsbuildinfo
+.env
```
kartik-mem0: The repo root `.gitignore` already covers `**/node_modules/` (L5), `dist/` (L18) and `.env` (L127), and no other integration ships its own `.gitignore`. Only `*.tsbuildinfo` and `*.local.mjs` are new here — probably cheaper to add those two to the root file and drop this one.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed, and good catch — this was real. My own e2e was printing the symptom ("Stored. No new distinct memory was extracted") and I'd read it as benign. `add` now handles the async `{eventId, status: "PENDING"}` shape: `formatAddResult` reports `Memory queued for background extraction (event <id>); it will be searchable shortly.` and only renders a list when the backend actually returns memories. Verified live — the write now surfaces the event id. Also added the async note to the `add_memory` description so a model doesn't add-then-search to confirm. Covered by `apply.test.ts` and `formatting.test.ts`.

integrations/dsh-mem0/tests/apply.test.ts
Himanshu-Sangshetti: Fixed. The add mock now returns the real `{eventId, status: "PENDING"}` shape (the SDK camel-cases response keys) and the test asserts the queued message + event id, so it fails against the old array-assuming code and passes after the fix.

integrations/dsh-mem0/src/formatting.ts
Himanshu-Sangshetti: Fixed. Rewrote the docstring to say `add` returns `{eventId, status: "PENDING"}` and that extraction runs server-side after the call returns, so the memories aren't in this response.

integrations/dsh-mem0/src/formatting.ts
Himanshu-Sangshetti: Fixed. `formatAddResult` now calls `formatMemoryList` for the list instead of re-implementing the numbering.

integrations/dsh-mem0/src/formatting.ts
Himanshu-Sangshetti: Removed. Dropped the unused `score` field from `MemoryLike` to keep the diff minimal rather than adding score rendering in this PR.

integrations/dsh-mem0/src/index.ts
Himanshu-Sangshetti: Fixed. Split into `resolveSearchFilters` (snake_case, spread into `filters`) and `resolveAddParams` (camelCase, top-level through the converter), mirroring pi-agent's `scoping.ts`, so the casing asymmetry is explicit in the code rather than load-bearing on the converter no-op. `scoping.test.ts` covers both shapes.

integrations/dsh-mem0/src/scoping.ts
Himanshu-Sangshetti: Fixed. `const clean = (v) => v?.trim() || undefined;` — the blank-string cases in `scoping.test.ts` still cover it.

docs/integrations/dsh-mem0.mdx
Himanshu-Sangshetti: Fixed. Reworded `host` everywhere (README, the Configuration table, `cordis.example.yml`, and the docs page) as a Platform base-URL override for on-prem / dedicated deployments, and dropped the `localhost:8000` example that would 404. It no longer reads as an OSS switch.

docs/llms.txt
Himanshu-Sangshetti: Fixed. Changed to `[Platform]`.

integrations/dsh-mem0/src/output.ts
Himanshu-Sangshetti: Fixed. The notice now distinguishes the two reasons: "showing N of M lines" when whole lines are dropped, "cut at 50KB" when only the byte cap trips. Left pi-agent's copy alone to keep this PR scoped — happy to open a matching one-liner there.

integrations/dsh-mem0/.gitignore
Himanshu-Sangshetti: Done. Moved `*.tsbuildinfo` + `*.local.mjs` to the root `.gitignore` and deleted `integrations/dsh-mem0/.gitignore`; the root already covers `node_modules`/`dist`/`.env`.

kartik-mem0 (approved):

PR #7028, opened 2026-08-18
fix(memory): handle empty vector store response in delete_all

### LINKED ISSUE #7025: bug(memory): sync delete_all crashes on an empty outer list response
### Component

Python SDK

### Summary

Synchronous `Memory.delete_all()` raises `IndexError` when a vector-store adapter returns `[]` for an empty result. The LangChain adapter explicitly returns this shape when no records match; the async implementation already handles it.

### Steps to Reproduce

```python
import mem0.memory.main as mm

mm.capture_event = lambda *args, **kwargs: None
memory = mm.Memory.__new__(mm.Memory)
memory.vector_store = type(
    "EmptyStore",
    (),
    {"list": lambda self, **kwargs: []},
)()

memory.delete_all(user_id="u1")
```

This matches `mem0/vector_stores/langchain.py:list()`, which returns `[]` when the underlying collection has no results.

### Expected Behavior

Deleting all memories for an already-empty scope should complete successfully and report zero deletions.

### Actual Behavior

`Memory.delete_all()` unconditionally evaluates `self.vector_store.list(...)[0]` and raises:

```text
IndexError: list index out of range
```

`AsyncMemory.delete_all()` already normalizes the same response with `batch = memories[0] if memories else []`.

### Environment

- mem0 version: repository `main` at `001c2352`
- Python version: 3.13.13 (project supports Python 3.10+)
- OS: macOS

### What I Ran

```text
uv run --extra test --extra llms python -c 'import mem0.memory.main as mm; mm.capture_event=lambda *a,**k: None; m=mm.Memory.__new__(mm.Memory); m.vector_store=type("EmptyStore",(),{"list":lambda self,**kwargs: []})(); m.delete_all(user_id="u1")'
```

### What I Saw

```text
File "mem0/memory/main.py", line 1923, in delete_all
    memories = self.vector_store.list(...)[0]
IndexError: list index out of range
```

### Why This Is a Bug

The public operation should be idempotent for an empty scope. The repository already accepts both flat and nested list shapes in `_get_all_from_vector_store()`, and the async delete path already guards an empty outer list.

### What I Ruled Out

- This is separate from pagination/truncation issues: it occurs before any deletion on the first empty response.
- This is separate from the redundant `get()` calls tracked in #5869.
- It is not hypothetical adapter behavior: `mem0/vector_stores/langchain.py` returns `[]` for no result.

### AI Assistance

pragati243:
## Summary

Fixes an `IndexError` in `Memory.delete_all()` when a vector store returns an empty outer list (`[]`).

The synchronous implementation previously indexed directly into the response with `[0]`, which caused `IndexError` for adapters returning an empty list when no memories matched. The response is now safely normalized before processing.

## Changes

- Handle an empty outer list returned by `vector_store.list()`
- Add a regression test for the `[]` response shape
- Preserve existing behavior for normal vector-store responses

## Testing

- `python -m pytest tests/test_main.py -k "empty_outer_list" -v`
- `python -m pytest tests/test_main.py -k "delete_all" -v`
- `python -m pytest tests/test_memory.py -k "delete_all" -v`
- `python -m pytest tests/memory/test_decay_usage_notice.py -k "delete_all" -v`
- `pre-commit` checks: Ruff and isort

Closes #7025

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -1921,18 +1921,25 @@ class Memory(MemoryBase):
         seen_batches = set()
         while True:
             memories = self.vector_store.list(
-                filters=filters, top_k=DELETE_ALL_BATCH_SIZE
-            )[0]
-            if not memories:
+                filters=filters,
+                top_k=DELETE_ALL_BATCH_SIZE,
+            )
+            batch = memories[0] if memories else []
+
+            if not batch:
                 break
-            batch_ids = tuple(sorted(str(memory.id) for memory in memories))
+
+            batch_ids = tuple(sorted(str(memory.id) for memory in batch))
             if batch_ids in seen_batches:
                 logger.warning("Stopping delete_all after a repeated memory batch")
                 break
+
             seen_batches.add(batch_ids)
-            for memory in memories:
+
+            for memory in batch:
                 self._delete_memory(memory.id)
-            deleted_count += len(memories)
+
+            deleted_count += len(batch)
 
         logger.info(f"Deleted {deleted_count} memories")
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -296,6 +296,14 @@ def test_delete_all(memory_instance):
     assert result["message"] == "Memories deleted successfully!"
 
 
+def test_delete_all_handles_empty_outer_list(memory_instance):
+    """delete_all should succeed when the vector store returns an empty list."""
+    memory_instance.vector_store.list = Mock(return_value=[])
+
+    result = memory_instance.delete_all(user_id="test_user")
+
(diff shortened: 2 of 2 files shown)


pragati243:
Hi! I reproduced this issue and implemented a minimal fix with a regression test.

The synchronous `Memory.delete_all()` now safely handles a vector-store response of `[]`, matching the defensive handling already present in the async implementation.

I verified the fix with the targeted and related `delete_all` tests, and the pre-commit checks pass.

I have opened PR #7028 for the implementation:
https://github.com/mem0ai/mem0/pull/7028

If this approach looks good, could a maintainer please mark this issue as `accepted`? I can then reopen the PR for review.

PR #7032, opened 2026-08-18
fix(security): resolve 17 Vanta/Dependabot HIGH+CRITICAL vulnerabilities across 5 pnpm workspaces

### LINKED ISSUE #7033: 17 open Dependabot HIGH/CRITICAL vulnerabilities across 5 npm workspaces
## Problem

17 GitHub Dependabot alerts (16 HIGH, 1 CRITICAL) are open against `main`, spanning 5 npm workspaces:

| CVE / GHSA | Package | Severity | Manifest |
|---|---|---|---|
| CVE-2026-23745, CVE-2026-23950, CVE-2026-24842, CVE-2026-26960, CVE-2026-29786, CVE-2026-31802, CVE-2026-59874 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` |
| CVE-2026-59873 | `tar` | CRITICAL | `mem0-ts/pnpm-lock.yaml` |
| GHSA-f88m-g3jw-g9cj | `sharp` | HIGH | `mem0-ts/pnpm-lock.yaml`, `server/dashboard/pnpm-lock.yaml` |
| CVE-2026-39244 | `adm-zip` | HIGH | `mem0-ts/pnpm-lock.yaml` |
| GHSA-5p4m-2wfm-xmqj | `js-yaml` | HIGH | `mem0-ts/`, `server/dashboard/`, `integrations/zapier-mem0/`, `integrations/vercel-ai-sdk/`, `integrations/n8n-nodes-mem0/` |
| CVE-2026-67213 | `nanoid` | HIGH | `server/dashboard/pnpm-lock.yaml` |

Reproduction: `gh api "repos/mem0ai/mem0/dependabot/alerts?state=open&per_page=100" --jq '.[] | select(.security_advisory.severity=="high" or .security_advisory.severity=="critical")'` lists all 17, confirmed still open against current `main`.

Root cause in each case is a transitive dependency — the packages are pulled in indirectly, not declared directly in any `package.json`. Two of the workspaces (`mem0-ts`, `server/dashboard`) already had `pnpm.overrides`/`pnpm-workspace.yaml overrides` pinning `tar`/`js-yaml` to versions that turned out to still be one patch version short of the disclosed fix, and one override (`tar`) had a lower-bound gap that missed a `tar@6.2.1` resolution pulled in by `fastembed`.

## Fix

Widen/add `pnpm.overrides` (and matching `pnpm-workspace.yaml` entries where the workspace already keeps them there) in each affected manifest, then regenerate the lockfile. No direct dependency, source code, or CI workflow changes needed.

Closes via PR #7032.

harshgupta-mem0:
## Summary
Closes #7033
- Resolves 17 open GitHub Dependabot HIGH/CRITICAL alerts via pnpm override bumps across 5 workspaces. One commit per workspace (each commit's message lists every CVE it closes).
- All bumps are transitive-dependency overrides — no direct dependency, source code, or CI workflow files were touched.

## Vulnerabilities

| CVE / GHSA | Package | Severity | Manifest | Affected | Patched |
|---|---|---|---|---|---|
| CVE-2026-23745 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` | `<= 7.5.2` | `7.5.3` |
| CVE-2026-23950 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` | `<= 7.5.3` | `7.5.4` |
| CVE-2026-24842 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` | `< 7.5.7` | `7.5.7` |
| CVE-2026-26960 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` | `< 7.5.8` | `7.5.8` |
| CVE-2026-29786 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` | `<= 7.5.9` | `7.5.10` |
| CVE-2026-31802 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` | `<= 7.5.10` | `7.5.11` |
| CVE-2026-59874 | `tar` | HIGH | `mem0-ts/pnpm-lock.yaml` | `<= 7.5.17` | `7.5.18` |
| CVE-2026-59873 | `tar` | CRITICAL | `mem0-ts/pnpm-lock.yaml` | `<= 7.5.18` | `7.5.19` |
| GHSA-f88m-g3jw-g9cj | `sharp` | HIGH | `mem0-ts/pnpm-lock.yaml` | `< 0.35.0` | `0.35.0` |
| CVE-2026-39244 | `adm-zip` | HIGH | `mem0-ts/pnpm-lock.yaml` | `< 0.6.0` | `0.6.0` |
| GHSA-5p4m-2wfm-xmqj | `js-yaml` | HIGH | `mem0-ts/pnpm-lock.yaml` | `>= 3.0.0, < 3.15.1` | `3.15.1` |
| CVE-2026-67213 | `nanoid` | HIGH | `server/dashboard/pnpm-lock.yaml` | `< 3.3.18` | `3.3.18` |
| GHSA-5p4m-2wfm-xmqj | `js-yaml` | HIGH | `server/dashboard/pnpm-lock.yaml` | `>= 4.0.0, < 4.3.1` | `4.3.1` |
| GHSA-f88m-g3jw-g9cj | `sharp` | HIGH | `server/dashboard/pnpm-lock.yaml` | `< 0.35.0` | `0.35.0` |
| GHSA-5p4m-2wfm-xmqj | `js-yaml` | HIGH | `integrations/zapier-mem0/pnpm-lock.yaml` | `>= 3.0.0, < 3.15.1` | `3.15.1` |
| GHSA-5p4m-2wfm-xmqj | `js-yaml` | HIGH | `integrations/vercel-ai-sdk/pnpm-lock.yaml` | `>= 3.0.0, < 3.15.1` | `3.15.1` |
| GHSA-5p4m-2wfm-xmqj | `js-yaml` | HIGH | `integrations/n8n-nodes-mem0/pnpm-lock.yaml` | `>= 3.0.0, < 3.15.1` and `>= 4.0.0, < 4.3.1` | `3.15.1` / `4.3.1` |

## Root cause
- `tar`: the existing `mem0-ts` override only widened `>=7.0.0 <7.5.21`, missing a separate `tar@6.2.1` transitive resolution pulled in by `fastembed`. Widened the floor to catch it, and bumped past all 8 disclosed CVEs at once.
- `js-yaml`: prior overrides across all 5 workspaces pinned the floor at exactly the version a newer advisory (GHSA-5p4m-2wfm-xmqj) still flags (`3.15.0`/`4.3.0`); bumped to the actual patched versions (`3.15.1`/`4.3.1`).
- `sharp`, `adm-zip`, `nanoid`: no override existed yet; added new ones. All three are transitive-only (no direct call sites in this repo — `sharp` is only reached via Next.js's internal image optimizer).

## Test plan
- [x] `mem0-ts`: `pnpm run build` succeeds, `pnpm run test:unit` — 1537/1537 pass
- [x] `server/dashboard`: `pnpm run build` succeeds (all 18 routes)
- [x] `integrations/zapier-mem0`: `pnpm run test:unit` — 14/14 pass
- [x] `integrations/vercel-ai-sdk`: `pnpm run build` succeeds; jest suite has pre-existing failures needing live LLM API keys, confirmed identical on `main` before this change
- [x] `integrations/n8n-nodes-mem0`: `tsc --noEmit` clean, jest — 17/17 pass
- [x] Confirmed no vulnerable versions remain in any of the 5 regenerated lockfiles
- [x] `zapier-mem0`/`n8n-nodes-mem0` lockfiles regenerated with pnpm v9 to match their CI pnpm version (their local default pnpm v11 silently ignores `package.json` overrides for these two packages)

## Deferred / not in this PR
- Vanta also lists CVE-2026-31240 (`mem0ai` pip pkg, no fix available), CVE-2026-69244 (`aiohttp`), CVE-2026-69247 (`cryptography`), and GHSA-gcfj-64vw-6mp9 (`axios`) — none of these currently exist as open GitHub Dependabot alerts on this repo (pip isn't tracked by Dependabot here, and the axios advisory isn't present in Dependabot's alert history either). Stale Vanta data per live GitHub cross-check; not touched in this batch.

---
🤖 Drafted with Claude Code — I ran the triage, cross-checks, dependency-override edits, lockfile regeneration, and local build/test verification described above. Commits were authored and signed by the repo owner after reviewing my staged changes.


kartik-mem0 (commented):
Reviewed in an isolated worktree. Each workspace installed at its own CI pnpm version with a fresh `--frozen-lockfile`, on Node 22.18.

| Workspace | pnpm | install | build | tests |
|---|---|---|---|---|
| `mem0-ts` | 10 | pass | pass | 1537/1537, prettier + export check pass |
| `server/dashboard` | 10 | pass | pass | no CI workflow exists |
| `integrations/zapier-mem0` | 9 | pass | pass | 14/14 |
| `integrations/vercel-ai-sdk` | 10 | pass | pass | n/a |
| `integrations/n8n-nodes-mem0` | 9 | pass | pass | 17/17 |

Scope is clean: 12 files, manifests and lockfiles only. No source, no workflows, no Python. No version bumps, merges cleanly with `main`.

Security claim verified independently. I pulled every open Dependabot alert and cross-checked it against the 5 regenerated lockfiles. All 22 are closed by the resolved versions (`tar 7.5.22`, `js-yaml 3.15.1`/`4.3.1`, `sharp 0.35.3`, `adm-zip 0.6.0`, `nanoid 3.3.18`). The body lists 17 HIGH/CRITICAL; 4 extra MEDIUM `tar` alerts are closed for free by the `>=7.5.21` floor. The root-cause analysis is correct, the tar advisories really are unbounded below, so the `tar@6.2.1` resolution was keeping them open.

Compat checks that came back clean:
- `adm-zip` 0.5.17 to 0.6.0: cassandra-driver's runtime path (`new AdmZip(buf).getEntries()` in `lib/datastax/cloud/index.js`) works on 0.6.0.
- `sharp` 0.34.5 to 0.35.3: binary loads, encode works, every method `@huggingface/transformers` can reach is present.
- `onnxruntime-node` 1.21.0's `tar.t({...})` install script is fine on tar 7, and its build scripts are ignored in this repo anyway.

Two things to resolve before merge, plus two nits. See inline comments.

mem0-ts/pnpm-workspace.yaml
```diff
@@ -31,7 +33,7 @@ overrides:
   "brace-expansion@>=2.0.0 <2.1.4": ">=2.1.4 <3.0.0"
   "brace-expansion@>=3.0.0 <5.0.8": ">=5.0.8 <6.0.0"
   "fast-xml-parser@>=5.0.0 <5.10.1": ">=5.10.1 <6.0.0"
-  "tar@>=7.0.0 <7.5.21": ">=7.5.21 <8.0.0"
+  "tar@<7.5.21": ">=7.5.21 <8.0.0"
```
kartik-mem0, line 36: Widening the floor from `>=7.0.0` sweeps fastembed's `tar@^6.2.0` from 6.2.1 up to 7.5.22, and that breaks it.

fastembed 2.1.0 does `__importDefault(require("tar")).default.x(...)`. tar 7 sets `__esModule: true` with no `default` export, so `.default` is `undefined`. Reproduced against this branch's own `node_modules`:

```
FlagEmbedding.decompressToCache(...)
  TypeError: Cannot read properties of undefined (reading 'x')
```

The same call on `tar@6.2.1` returns a function. Model download and extraction is dead.

CI cannot see this: `src/oss/tests/fastembed-embedder.test.ts` does `jest.mock("fastembed")`, so the 1537 passing tests never touch the real path.

Blast radius is repo-local. `fastembed` is an optional peerDependency of mem0-ts and pnpm overrides do not propagate to consumers, so published `mem0ai` users install their own tar 6 and are unaffected. What breaks is local dev, examples, and manual runs in this repo.

There is no clean fix here. Every tar advisory patches only in 7.5.x with no 6.x backport, so you cannot both close the alert and keep fastembed working. Please pick one explicitly rather than shipping the break silently:

1. Keep the widening, note the known break in the PR body, and dismiss the tar@6 alert as "no patch available".
2. Add `"fastembed>tar": "^6.2.1"` to keep fastembed working and accept that the tar alerts stay open.

integrations/zapier-mem0/package.json
```diff
@@ -71,7 +71,8 @@
       "esbuild": ">=0.28.1",
       "undici@<7.29.0": ">=7.29.0 <8.0.0",
       "undici@>=8.0.0 <8.9.0": ">=8.9.0 <9.0.0",
-      "brace-expansion@<1.1.16": ">=1.1.16 <2.0.0"
+      "brace-expansion@<1.1.16": ">=1.1.16 <2.0.0",
+      "js-yaml@<3.15.1": ">=3.15.1 <4.0.0"
```
kartik-mem0, line 75: This override landed in `package.json` but not in `integrations/zapier-mem0/pnpm-workspace.yaml`, and that is a regression on current pnpm.

pnpm 9 and 10 still read `pnpm.overrides` from `package.json`, which is why CI (pnpm 9) is green. pnpm 11, the current `latest` at 11.22.0, ignores the field entirely. Tested both sides:

```
main    + pnpm 11  ->  install succeeds
this PR + pnpm 11  ->  ERR_PNPM_LOCKFILE_CONFIG_MISMATCH
```

Before this PR the two files matched at 6 overrides each, so the lockfile agreed with `pnpm-workspace.yaml`. Adding a 7th to only one of them breaks that.

Fix is one line: add the same `js-yaml@<3.15.1` entry to `integrations/zapier-mem0/pnpm-workspace.yaml`.

Separately, `n8n-nodes-mem0` already fails the same way on `main` because it has no `pnpm-workspace.yaml` at all. Pre-existing, not this PR, but worth a follow-up issue.

mem0-ts/pnpm-workspace.yaml
```diff
@@ -10,8 +10,10 @@ overrides:
   "picomatch@<2.3.2": "^2.3.2"
   "picomatch@>=4.0.0 <4.0.4": "^4.0.4"
   "jws@4.0.0": "4.0.1"
-  "js-yaml@<3.15.0": ">=3.15.0 <4.0.0"
+  "js-yaml@<3.15.1": ">=3.15.1 <4.0.0"
   "js-yaml@>=4.0.0 <4.3.0": ">=4.3.0 <5.0.0"
+  "sharp@<0.35.0": ">=0.35.0"
```
kartik-mem0: Nit: these two are the only overrides in the file with no upper bound, including `nanoid` added in the same PR. A future `sharp` 1.0 or `adm-zip` 1.0 would be pulled into the dashboard's Next image optimizer and into cassandra-driver on the next unfrozen install, with nothing gating it.

Suggest `>=0.35.0 <0.36.0` and `>=0.6.0 <0.7.0` to match the rest of the file. Same applies to the `sharp` entry in `server/dashboard/pnpm-workspace.yaml`.

server/dashboard/pnpm-workspace.yaml
```diff
@@ -9,5 +9,7 @@ overrides:
   "brace-expansion@<1.1.16": ">=1.1.16 <2.0.0"
   "brace-expansion@>=3.0.0 <5.0.8": ">=5.0.8 <6.0.0"
   "js-yaml@<3.15.0": ">=3.15.0 <4.0.0"
-  "js-yaml@>=4.0.0 <4.3.0": ">=4.3.0 <5.0.0"
+  "js-yaml@>=4.0.0 <4.3.1": ">=4.3.1 <5.0.0"
```
kartik-mem0, line 12: Nit: the line above still reads `js-yaml@<3.15.0` while the other four workspaces moved to `<3.15.1`. Mirror image in `mem0-ts` and `vercel-ai-sdk`, which still carry `js-yaml@>=4.0.0 <4.3.0` while this file and n8n moved to `<4.3.1`.

Not a live gap, js-yaml 3.x is not in this tree and 4.x is not in the other two, so no alert is open on any of them. But it pins to a flagged version the moment those majors appear. Worth aligning all five while you are in here.

mem0-ts/pnpm-workspace.yaml
harshgupta-mem0: Confirmed — reproduced this locally. Patched tar instead of accepting the break or leaving the alert open:

```
node -e "
const tarPath = require.resolve('tar', { paths: [require.resolve('fastembed')] });
const tar = require(tarPath);
function __importDefault(mod) { return (mod && mod.__esModule) ? mod : { default: mod }; }
const tar_1 = __importDefault(tar);
console.log(typeof tar_1.default.x); // was undefined, now: function
"
```

Added `patchedDependencies: { "tar@7.5.22": "patches/tar@7.5.22.patch" }` via `pnpm patch` — the patch appends `exports.default = exports;` to `dist/commonjs/index.min.js`, which is exactly what fastembed's `__importDefault` helper expects. tar stays at the patched 7.5.22 (CVE closed), fastembed's `decompressToCache` works again. Verified: `pnpm build` + `pnpm test:unit` (1537/1537) still pass in mem0-ts.

integrations/zapier-mem0/package.json
harshgupta-mem0: Confirmed, this was a real gap — added the matching `js-yaml@<3.15.1` entry to `integrations/zapier-mem0/pnpm-workspace.yaml` so it now matches `package.json`'s `pnpm.overrides`. Rebuilt and re-tested with pnpm 9 (matching CI): `pnpm build` and `pnpm test:unit` (14/14) both pass.

mem0-ts/pnpm-workspace.yaml
harshgupta-mem0: Good catch, done. `sharp` and `adm-zip` now have upper bounds (`>=0.35.0 <0.36.0` and `>=0.6.0 <0.7.0`) matching the rest of `mem0-ts/pnpm-workspace.yaml`, and the `sharp` entry in `server/dashboard/pnpm-workspace.yaml` got the same `<0.36.0` bound.

server/dashboard/pnpm-workspace.yaml
harshgupta-mem0: Aligned all five workspaces to `js-yaml@<3.15.1` / `js-yaml@>=4.0.0 <4.3.1` — `mem0-ts` and `vercel-ai-sdk` (both the `pnpm-workspace.yaml` and the mirrored `package.json` `pnpm.overrides`) now carry the `<4.3.1` upper bound, and `server/dashboard` now carries the `<3.15.1` floor. No live alert was closed by this (as you noted), but it removes the footgun for whichever major appears next.

harshgupta-mem0:
**Addressed review feedback in 71c0901f.** Both blocking issues and both nits were correct — replied inline to each with specifics. Summary:

| # | Finding | Verdict | Fix |
|---|---|---|---|
| 1 | Widening `tar` floor breaks fastembed's `decompressToCache` (tar 7's CJS build has no `default` export, fastembed's compiled `require()` expects one) | Correct, reproduced locally | Patched tar via `pnpm patch` to add `exports.default = exports;` — keeps tar at the patched 7.5.22 (CVE closed) and fixes the interop instead of breaking fastembed or leaving the alert open |
| 2 | `js-yaml` override landed in `zapier-mem0/package.json` but not `pnpm-workspace.yaml` — real regression risk on pnpm 11 | Correct | Added the matching entry to `pnpm-workspace.yaml` |
| 3 | `sharp`/`adm-zip` overrides have no upper bound | Correct, nit | Added `<0.36.0`/`<0.7.0` bounds in `mem0-ts` and `server/dashboard` |
| 4 | `js-yaml` ranges inconsistent across the five workspaces | Correct, nit | Aligned all five to `<3.15.1` / `<4.3.1` |

Verified locally after the fix:
- `mem0-ts`: `pnpm build` + `pnpm test:unit` — 1537/1537 pass. Manually confirmed the tar patch fixes the exact call fastembed makes (`tar_1.default.x` is now a function, was `undefined`).
- `integrations/zapier-mem0`: `pnpm build` + `pnpm test:unit` — 14/14 pass (pnpm 9, matching CI).
- `integrations/vercel-ai-sdk`: `pnpm build` passes (pnpm 10.5.2, matching its pin).
- `server/dashboard`: `pnpm build` passes, all 18 routes generated (pnpm 10.34.2, matching its pin).
- Every lockfile diff is minimal — override-header changes and the tar patch hash only, no unintended resolution changes.
- CI on the new commit is green.

kartik-mem0 (approved):

PR #7110, opened 2026-08-25
feat(deepseek-plugin,mem0-strands): add usage telemetry

kartik-mem0:
## Linked Issue

No linked issue. This branch was pushed directly to `mem0ai/mem0` (not from a fork), which the PR Gate exempts from the accepted-issue requirement. Happy to link one if a maintainer wants an issue filed retroactively.

## Description

Adds anonymous usage telemetry to two integrations:

- **`integrations/deepseek-plugin/`**: the DeepSeek Harness (Cordis) plugin
- **`integrations/mem0-strands/`**: the Strands `MemoryStore`

so Mem0 can see which integrations are actually used and where they fail, without adding a new dependency to either package.

**Design.** Both ride the Mem0 SDK's existing PostHog client rather than pulling in a new one. `distinct_id` is the account email when it can be resolved (`MemoryClient.telemetryId` in TypeScript, `instance.user_email` in Python), otherwise a persisted anonymous id, with a one-shot `$identify` event carrying `$anon_distinct_id` so pre-auth history merges into the account instead of stranding a second identity once the email resolves.

Strands events go through `client_telemetry` (the unsampled hosted singleton) rather than the core SDK's `capture_event`, which samples hot loop paths at 0.1. A store call is an agent-level action, not a loop iteration, so a 10% sample rate would just be noise. Passing `user_email=None` on the OSS path falls back to the SDK's machine-local anonymous id, so identity is correct on both the hosted and self-hosted backends for free.

**Events.** `deepseek.plugin.mounted`, `deepseek.tool.search_memory`, `deepseek.tool.add_memory`; `strands.store.init`, `strands.store.search`, `strands.store.add`, `strands.store.add_messages`.

**Privacy.** Only counts, durations, booleans, and enums are sent. Queries, memory text, message content, entity ids (`user_id`/`agent_id`/`run_id`), metadata values, and API keys are never sent. There's a test asserting that an error message containing an API key records only as the coarse bucket `rate-limited`. `strands.store.init` reports scope *kinds* (e.g. `["user_id"]`), never scope values.

**Opt-out.** `MEM0_TELEMETRY=false`, documented in both READMEs.

**Known follow-up, deliberately not in this PR.** Backend `source` attribution is a separate mechanism from PostHog: a top-level `source` field on platform writes, validated against the backend's `KNOWN_EVENT_SOURCES` allowlist, with unknown values bucketing to `OTHERS`. `STRANDS` is already on that allowlist; `DEEPSEEK_HARNESS` is not, so DeepSeek writes currently bucket to `OTHERS` until a one-line platform change adds it. That needs a separate backend PR and is out of scope here.

## Type of Change

- [ ] Documentation update

## AI Assistance

- [ ] No AI assistance

This diff was written by Claude Code, with me directing the design (PostHog reuse, distinct_id/identify strategy, sampled vs. unsampled event path, the privacy boundary on what fields are allowed in event properties) and reviewing the result line by line, including the test suite.

## Breaking Changes

N/A

## Test Coverage

deepseek-plugin: `pnpm exec tsc --noEmit` clean, `pnpm exec vitest run` → 42 tests passing across 5 files.
mem0-strands (`python/`): `ruff check src tests` clean, `ruff format --check src tests` clean, `mypy src` clean, `pytest -q` → 44 tests passing.

## Checklist

### DIFF
--- a/integrations/deepseek-plugin/src/telemetry.ts
+++ b/integrations/deepseek-plugin/src/telemetry.ts
@@ -1,0 +1,173 @@
+/**
+ * Anonymous usage telemetry for the DeepSeek Harness plugin.
+ *
+ * Fire-and-forget PostHog events over native fetch, batched and flushed every
+ * 5 seconds, at 10 queued events, or on process exit. Never throws, never logs.
+ *
+ * Events carry only tool names, durations, counts, and coarse failure kinds:
+ * never queries, memory text, filters, or API keys.
+ *
+ * Disable with MEM0_TELEMETRY=false.
+ */
+import { randomUUID } from "node:crypto";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+
+const POSTHOG_API_KEY = "phc_hgJkUVJFYtmaJqrvf6CYN67TIQ8yhXAkWzUn9AMU4yX";
+const POSTHOG_BATCH_URL = "https://us.i.posthog.com/batch/";
+const FLUSH_INTERVAL_MS = 5_000;
+const FLUSH_THRESHOLD = 10;
+const SEND_TIMEOUT_MS = 3_000;
+
+const PLUGIN_VERSION = ((): string => {
+  try {
+    return JSON.parse(
+      fs.readFileSync(new URL("../package.json", import.meta.url), "utf-8"),
+    ).version;
+  } catch {
+    return "unknown";
+  }
+})();
+
+/** The Mem0 SDK resolves this to the account email once ping() lands, so events join other Mem0 surfaces. */
+export interface TelemetryIdentity {
+  telemetryId?: string;
+}
+
+let queue: Record<string, unknown>[] = [];
+let flushTimer: ReturnType<typeof setInterval> | undefined;
+let exitHandlerInstalled = false;
+let cachedAnonymousId: string | undefined;
+let identified = false;
+
+function identityPath(): string {
+  return path.join(os.homedir(), ".mem0", "deepseek-plugin-telemetry.json");
+}
+
+export function isTelemetryEnabled(): boolean {
+  const value = process.env.MEM0_TELEMETRY?.toLowerCase();
+  return value !== "false" && value !== "0" && value !== "no" && value !== "off";
+}
+
+function anonymousId(): string {
+  if (cachedAnonymousId) return cachedAnonymousId;
+  try {
+    const stored = JSON.parse(fs.readFileSync(identityPath(), "utf-8"));
+    if (typeof stored.anonymousId === "string" && stored.anonymousId) {
+      cachedAnonymousId = stored.anonymousId;
+      return cachedAnonymousId!;
+    }
+  } catch {
+    /* first run, or an unreadable identity file */
+  }
+  const created = `deepseek-anon-${randomUUID().replace(/-/g, "")}`;
+  try {
+    const target = identityPath();
+    fs.mkdirSync(path.dirname(target), { recursive: true });
+    fs.writeFileSync(target, JSON.stringify({ anonymousId: created }), "utf-8");
+  } catch {
+    /* an unwritable home directory must not break the plugin */
+  }
+  cachedAnonymousId = created;
+  return created;
+}
+
+/** Merge the pre-ping anonymous history into the account once the email resolves. */
+function identifyEvent(distinctId: string): Record<string, unknown> | undefined {
+  if (identified || distinctId.startsWith("deepseek-anon-")) return undefined;
+  identified = true;
+  let storedAnonymousId: string | undefined;
+  try {
+    storedAnonymousId = JSON.parse(fs.readFileSync(identityPath(), "utf-8")).anonymousId;
+    fs.unlinkSync(identityPath());
+  } catch {
+    return undefined;
+  }
+  cachedAnonymousId = undefined;
+  if (!storedAnonymousId) return undefined;
+  return {
+    event: "$identify",
+    distinct_id: distinctId,
+    properties: { $anon_distinct_id: storedAnonymousId, $lib: "posthog-node" },
+  };
+}
+
+export function flushEvents(): void {
+  if (queue.length === 0) return;
+  const batch = queue;
+  queue = [];
+  fetch(POSTHOG_BATCH_URL, {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({ api_key: POSTHOG_API_KEY, batch }),
+    signal: AbortSignal.timeout(SEND_TIMEOUT_MS),
+  }).catch(() => {
+    /* telemetry never surfaces its own failures */
+  });
+}
+
+function scheduleFlush(): void {
+  if (!flushTimer) {
+    flushTimer = setInterval(flushEvents, FLUSH_INTERVAL_MS);
+    flushTimer.unref?.();
+  }
+  if (!exitHandlerInstalled) {
+    exitHandlerInstalled = true;
+    process.on("beforeExit", flushEvents);
+  }
+}
+
+export function errorKind(error: unknown): string {
+  const text = (error instanceof Error ? error.message : String(error)).toLowerCase();
+  if (text.includes("timeout") || text.includes("aborted")) return "timeout";
+  if (text.includes("401") || text.includes("403") || text.includes("unauthor")) return "auth";
+  if (text.includes("429") || text.includes("rate limit")) return "rate-limited";
+  if (/50[0234]/.test(text)) return "server-error";
+  if (text.includes("400") || text.includes("422")) return "bad-request";
+  if (text.includes("fetch failed") || text.includes("enotfound")) return "network";
+  return error instanceof Error ? error.constructor.name : "other";
+}
+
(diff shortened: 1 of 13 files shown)


Himanshu-Sangshetti (commented):
Reviewed the telemetry work. The Python half (`mem0-strands`) is clean: it rides the SDK's `client_telemetry` singleton, passes `_distinct_id` as the `user_email` positional matching the real `capture_event` signature, and the privacy/error-bucket tests are solid. All the notes below are on the TypeScript `deepseek-plugin` side, and they share one root cause.

**Main ask (reuse + consistency): the TS telemetry re-implements the SDK's own client.**
`mem0-ts` already exports the exact machinery this needs: `telemetry` (`captureEvent`/`captureIdentify`), `generateHash`, and `isTelemetryEnabled` (see `mem0-ts/src/client/telemetry.ts`). The Python side of this PR reuses that singleton. The TS side instead rebuilds it in `integrations/deepseek-plugin/src/telemetry.ts`: a **second hardcoded PostHog key** (`telemetry.ts:177`, identical to the SDK's), its own batch/flush, its own anon-id file, and its own opt-out parsing. That contradicts the PR's stated design ("ride the Mem0 SDK's existing PostHog client") and is the source of the bugs below. Reworking the TS side to reuse the SDK's exports (like the Python side) removes ~170 duplicated lines and dissolves most of these.

The bugs that fall out of the reimplementation:

1. **Second anonymous identity.** It mints `deepseek-anon-*` independent of the SDK's shared anon id, so a user's pre-auth events split across two identities and two `$identify` flows, the stranding the PR aims to prevent. Reusing the SDK's shared anon id fixes this.
2. **Flush relies on `beforeExit`.** It doesn't fire on `process.exit()`, SIGINT/SIGTERM, or uncaught exceptions. A short-lived harness run (a couple tool calls, under the 10-event threshold, before the 5s timer) loses all its events, which is the common case.
3. **`errorKind` regex `/50[0234]/`** matches "500/502/503/504" anywhere in the text, so an error containing e.g. "504ms" is mislabeled `server-error`, and a real 501 is never classified.
4. **Anon-id regenerates on an unreadable file.** Any `JSON.parse` throw (partial or concurrent write) mints a new id and overwrites the file, forking identity or clobbering a concurrent process.
5. **Opt-out parsing drifts from the SDK.** The SDK treats only `MEM0_TELEMETRY=false` as off; this also honors `0`/`no`/`off`, so opt-out behaves inconsistently between the SDK's own events and the plugin's in the same process.

Not blocking the design, the events and privacy boundary look right. The single change that matters is making the TS side reuse the SDK telemetry the way the Python side already does; 2, 4, and 5 disappear with it, and 3 is a one-line regex fix.

integrations/deepseek-plugin/src/telemetry.ts
```diff
@@ -0,0 +1,173 @@
...
+ * never queries, memory text, filters, or API keys.
+ *
+ * Disable with MEM0_TELEMETRY=false.
+ */
+import { randomUUID } from "node:crypto";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+
+const POSTHOG_API_KEY = "phc_hgJkUVJFYtmaJqrvf6CYN67TIQ8yhXAkWzUn9AMU4yX";
```
Himanshu-Sangshetti, line 17: **[High]** This hardcodes the same PostHog key the SDK already ships (`mem0-ts/src/client/telemetry.ts`) and re-implements the SDK's telemetry client (`captureEvent` / `generateHash` / `isTelemetryEnabled` + the shared anon-id and `$identify` alias tracking). The Python half of this PR reuses the SDK singleton (`client_telemetry`); the TS half should too. Reusing it here drops ~170 lines, removes the duplicate key, and dissolves the separate anon-id and opt-out drift flagged below. This is the main change.

integrations/deepseek-plugin/src/telemetry.ts
```diff
@@ -0,0 +1,173 @@
...
+  try {
+    const stored = JSON.parse(fs.readFileSync(identityPath(), "utf-8"));
+    if (typeof stored.anonymousId === "string" && stored.anonymousId) {
+      cachedAnonymousId = stored.anonymousId;
+      return cachedAnonymousId!;
+    }
+  } catch {
+    /* first run, or an unreadable identity file */
+  }
+  const created = `deepseek-anon-${randomUUID().replace(/-/g, "")}`;
```
Himanshu-Sangshetti, line 64: **[High]** Minting `deepseek-anon-*` independent of the SDK's shared anon id splits a user's pre-auth events across two identities and two `$identify` flows — the exact stranding this PR sets out to prevent. Use the SDK's shared anon id instead of a plugin-local one.

integrations/deepseek-plugin/src/telemetry.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+function scheduleFlush(): void {
+  if (!flushTimer) {
+    flushTimer = setInterval(flushEvents, FLUSH_INTERVAL_MS);
+    flushTimer.unref?.();
+  }
+  if (!exitHandlerInstalled) {
+    exitHandlerInstalled = true;
+    process.on("beforeExit", flushEvents);
```
Himanshu-Sangshetti, line 117: **[High]** `beforeExit` does not fire on `process.exit()`, SIGINT/SIGTERM, or uncaught exceptions, and the flush timer is `unref()`'d. A short-lived harness run (a couple tool calls, under `FLUSH_THRESHOLD`, before the 5s timer) ends without ever flushing, so `mounted`/`search`/`add` are silently never delivered — the common case. Flush on signal handlers or a synchronous exit path, or send per event.

integrations/deepseek-plugin/src/telemetry.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+}
+
+export function errorKind(error: unknown): string {
+  const text = (error instanceof Error ? error.message : String(error)).toLowerCase();
+  if (text.includes("timeout") || text.includes("aborted")) return "timeout";
+  if (text.includes("401") || text.includes("403") || text.includes("unauthor")) return "auth";
+  if (text.includes("429") || text.includes("rate limit")) return "rate-limited";
+  if (/50[0234]/.test(text)) return "server-error";
```
Himanshu-Sangshetti, line 126: **[Medium]** `/50[0234]/` matches `500/502/503/504` anywhere in the lowercased text — a `504ms` duration or an id substring gets labeled `server-error` — and `501` is never classified. Anchor to an actual status code (e.g. a bounded match on the parsed status) instead of a loose substring.

integrations/deepseek-plugin/src/telemetry.ts
```diff
@@ -0,0 +1,173 @@
...
+
+function anonymousId(): string {
+  if (cachedAnonymousId) return cachedAnonymousId;
+  try {
+    const stored = JSON.parse(fs.readFileSync(identityPath(), "utf-8"));
+    if (typeof stored.anonymousId === "string" && stored.anonymousId) {
+      cachedAnonymousId = stored.anonymousId;
+      return cachedAnonymousId!;
+    }
+  } catch {
```
Himanshu-Sangshetti, line 61: **[Medium]** On any `JSON.parse` throw (a partial or concurrent write) this mints a new id and overwrites the file, forking the anonymous identity or clobbering a concurrent process. Preserve the existing file on read errors rather than regenerating. (Moot if you adopt the SDK's shared anon id above.)

integrations/deepseek-plugin/src/telemetry.ts
```diff
@@ -0,0 +1,173 @@
...
+let exitHandlerInstalled = false;
+let cachedAnonymousId: string | undefined;
+let identified = false;
+
+function identityPath(): string {
+  return path.join(os.homedir(), ".mem0", "deepseek-plugin-telemetry.json");
+}
+
+export function isTelemetryEnabled(): boolean {
+  const value = process.env.MEM0_TELEMETRY?.toLowerCase();
```
Himanshu-Sangshetti, line 49: **[Low]** Opt-out parsing drifts from the SDK: the SDK treats only `MEM0_TELEMETRY=false` as off, while this also honors `0`/`no`/`off`, so opt-out behaves differently for the SDK's own events vs the plugin's in the same process. Reuse the SDK's exported `isTelemetryEnabled`.

Himanshu-Sangshetti:
### Severity rollup

Inline comments are on each line. The Python half (`mem0-strands`) is clean; everything below is the TS `deepseek-plugin`.

- **Blocker:** none
- **High (3):** SDK telemetry reuse + duplicate hardcoded PostHog key (`telemetry.ts:17`); split anonymous identity (`:64`); flush loses events on short-lived / signal exit (`:117`)
- **Medium (2):** error-bucket regex `/50[0234]/` misclassifies (`:126`); anon-id regenerated/overwritten on a read error (`:61`)
- **Low (1):** opt-out parsing drifts from the SDK (`:49`)

The three High items are the ones to resolve before merge. Adopting the SDK's existing telemetry client (the `:17` fix) also dissolves `:64`, `:61`, and `:49`, so it's largely one change plus the flush lifecycle (`:117`) and the regex (`:126`).

Himanshu-Sangshetti (approved):

PR #7116, opened 2026-08-26
fix(llms): classify gpt-5 dash family as reasoning models

### LINKED ISSUE #6085: OSS default config silently breaks add(): model=None resolves to gpt-5-mini, which rejects the default temperature=0.1

RahilOp:
## Problem

The default OSS config resolves `model=None` to `gpt-5-mini`, which is a reasoning model that rejects any non-default `temperature`. The existing `_is_reasoning_model` heuristic only matched the exact string `gpt-5`, so `gpt-5-mini` fell through to the temperature path. This caused every default `Memory().add(...)` to fail LLM extraction and silently store nothing.

Closes #6085

## Changes

- Updated `mem0/llms/base.py` to classify the GPT-5 dash family (`gpt-5-mini`, `gpt-5-nano`, and dated variants) as reasoning models.
- Explicitly excluded `gpt-5-chat*` variants, which are non-reasoning chat models that still accept a custom temperature.
- Added regression tests covering the classification and the end-to-end API call (no `temperature` is sent for the default `gpt-5-mini`).

## Testing

```bash
MEM0_TELEMETRY=False hatch run pytest tests/llms/test_openai.py::test_is_reasoning_model_classification tests/llms/test_openai.py::test_default_gpt5_mini_omits_temperature -x
```

All tests pass. `ruff check` also passes.

## Checklist

- [x] I have read the CLA Document and I sign the CLA

### DIFF
--- a/mem0/llms/base.py
+++ b/mem0/llms/base.py
@@ -72,6 +72,18 @@ class LLMBase(ABC):
         if any(base_model.startswith(prefix) for prefix in ["o1-", "o1.", "o3-", "o3."]):
             return True
 
+        # Match the original GPT-5 dash family (gpt-5-mini, gpt-5-nano and their
+        # dated variants like gpt-5-mini-2025-08-07). These are reasoning models
+        # that reject a non-default temperature. This must NOT match the newer
+        # dot-versioned models (gpt-5.4-mini) which support temperature, hence
+        # the explicit "gpt-5-" (dash) prefix rather than a bare "gpt-5" substring.
+        #
+        # Exclude the "gpt-5-chat" family (gpt-5-chat, gpt-5-chat-latest): these
+        # are the non-reasoning chat variants that DO accept a custom temperature,
+        # so they must stay on the temperature path.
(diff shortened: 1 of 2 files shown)


dlowzzxx (approved):
Verified mem0/llms/base.py:75-87 correctly fixes #6085 where default config resolves model=None to gpt-5-mini — a reasoning model that rejects non-default temperature — so the old exact in reasoning_models check caused Memory().add() to silently fail and store nothing. The added if base_model.startswith("gpt-5-") and not base_model.startswith("gpt-5-chat"): return True cleanly partitions the dash family (gpt-5-mini, gpt-5-nano, dated variants with rsplit provider stripping) from non-reasoning gpt-5-chat* and dot-versioned gpt-5.4* which must keep temperature per tests/llms/test_openai.py:290. Reproduced locally that main returns False for gpt-5-mini and confirmed new tests tests/llms/test_openai.py:323-355 cover both paths. Narrow dash-prefix with explicit chat exclusion is the right minimal heuristic — LGTM

PR #7124, opened 2026-08-26
fix(embeddings): support new sentence-transformers dimension API

### LINKED ISSUE #7019: bug(huggingface): warning message regarding get_sentence_embedding_dimension
### Component

Python SDK

### Summary

The following warning message would appear if I build a new project with mem0 now as this would use `sentence-transformers` version higher than `5.4.0`, which marked `get_sentence_embedding_dimension` as deprecated.

```
FutureWarning: The `get_sentence_embedding_dimension` method has been renamed to `get_embedding_dimension`.
```

closes via PR #7020 

### Steps to Reproduce

```python
from mem0 import Memory

m = Memory()
# Your code here...
from_config(
    {
        ... # your config
        "embedder": {
            "provider": "huggingface",
            "config": {
                "model": "BAAI/bge-base-en-v1.5", # any other huggingface model
            },
        },
    },
)
```

### Expected Behavior

The warning message shouldn't appear. 

### Actual Behavior

```
FutureWarning: The `get_sentence_embedding_dimension` method has been renamed to `get_embedding_dimension`.
```

### Environment

- mem0 version: 2.0.18
- OS: Mac

### What I Ran

The scrip attached above.

### What I Saw

Error message in every initialization
```
FutureWarning: The `get_sentence_embedding_dimension` method has been renamed to `get_embedding_dimension`.
```

### Why This Is a Bug

Currently, there's a warning message showing in the user's terminal. This is currently a potential bug as long as `get_sentence_embedding_dimension` is still maintained by the package `sentence-transformers`. However, once the function is removed, our embedding will fail due to non-existent function.

### What I Ruled Out

Tried a few embedding models from HuggingFace and saw the same error. Once the fix is applied, I don't see the warning any more

### AI Assistance

No AI involved

mikemikimike:
## Summary

`sentence-transformers` 5.4.0 deprecates `get_sentence_embedding_dimension()` in favor of `get_embedding_dimension()`. The Hugging Face embedder now prefers the new API while retaining a fallback for older sentence-transformers releases.

## Changes

- Use `get_embedding_dimension()` when available.
- Fall back to `get_sentence_embedding_dimension()` for older versions.
- Add regression coverage for both API shapes.

## Validation

- `python -m pytest tests/embeddings/test_huggingface_embeddings.py -q` — 16 passed
- `python -m ruff check mem0/embeddings/huggingface.py tests/embeddings/test_huggingface_embeddings.py` — passed
- `git diff HEAD^ --check` — passed

Related issue: #7019

Fixes #7019


dlowzzxx (approved):
Verified this fixes the sentence-transformers 5.4.0 deprecation warned in #7019. The getattr(self.model, "get_embedding_dimension", None) fallback correctly prefers the new API while retaining backward compatibility, and the or short-circuit still respects user-provided embedding_dims. The added tests properly cover both API shapes, confirming the new path via assert_not_called and the legacy deletion path. Clean, minimal fix that eliminates the FutureWarning without breaking older installs. LGTM.

mikemikimike:
I rechecked this PR. The current review history contains an approval, there is no human changes-requested review, and no failing check is reported for the current head. The remaining blocked state appears to be repository merge gating rather than a code failure introduced by this change. Could a maintainer please review the approved change and merge it, or advise on any remaining requirement?

kartik-mem0 (commented):
Reviewed at `5e183aa3`.

Clean fix for the `sentence-transformers` API rename. The `getattr` fallback pattern is correct for the transition window (old releases have only `get_sentence_embedding_dimension`, new releases prefer `get_embedding_dimension`, current 5.4.0 has both). Single call site in the codebase (verified by grep), so no shared helper needed.

Verified independently:
- 16/16 tests pass on the PR branch (including 2 new regression tests for both API branches)
- Existing test updated with `del mock_sentence_transformer.get_embedding_dimension` so it keeps exercising the legacy path explicitly
- Confirmed both methods coexist on `sentence-transformers` 5.4.0
- Ruff (line-length 120): clean

LGTM. Linked issue #7019 needs `accepted` label for the PR Gate.

PR #7126, opened 2026-08-26
fix(vector_stores/faiss): reset index after load failure

### LINKED ISSUE #7117: bug(vector_stores/faiss): swallowed _load() failure leaves index and docstore out of sync, search silently returns the wrong memory
### Component

Vector Store

### Summary

When the FAISS store fails to load its persisted state, the `except Exception`
block in `_load()` (mem0/vector_stores/faiss.py) clears `docstore` and
`index_to_id` but does nothing about `self.index`. Depending on which file was
damaged, you end up in one of two broken states:

- Corrupted `.json` docstore, intact `.faiss` index: the index keeps its N
  old vectors, the mapping is empty. The next `insert()` numbers new vectors
  from `len(self.index_to_id)` (= 0) while FAISS appends them at position N.
  From then on every mapping is shifted: `search()` returns the wrong memory
  with a top score, and the newly inserted memories are unreachable. No error,
  no warning beyond the initial load log line.

- Corrupted `.faiss` index: `faiss.read_index()` raises, `_load()` swallows
  it, and `self.index` stays `None`. `__init__` only calls `create_col()` in
  the branch where no persisted files exist, so it never runs here. Every
  later `insert()`/`search()`/`get()` raises
  `ValueError: Collection not initialized. Call create_col first.` and the
  store never recovers.

The corrupted-json case is not exotic: `_save()` rewrites the json in place
with a plain `open(path, "w")` + `json.dump()`. Kill the process mid-dump
(an OOM kill is enough) and you get a truncated file next to an intact,
already-written `.faiss` index, which is exactly the state that triggers the
mis-mapping.

### Steps to Reproduce

```python
import shutil
import numpy as np

DIMS, PATH = 8, "/tmp/faiss-repro"

def vec(i):
    v = np.zeros(DIMS, dtype=np.float32); v[i] = 1.0
    return v.tolist()

shutil.rmtree(PATH, ignore_errors=True)
store = FAISS(collection_name="repro", path=PATH, embedding_model_dims=DIMS)
store.insert([vec(0), vec(1)], payloads=[{"data": "apple"}, {"data": "banana"}],
             ids=["id-apple", "id-banana"])

# Truncate the docstore json, simulating a crash mid _save()
with open(f"{PATH}/repro.json", "r+", encoding="utf-8") as f:
    content = f.read(); f.seek(0); f.truncate(); f.write(content[: len(content) // 2])

store2 = FAISS(collection_name="repro", path=PATH, embedding_model_dims=DIMS)
store2.insert([vec(2)], payloads=[{"data": "cherry"}], ids=["id-cherry"])

print(store2.search(query="", vectors=vec(0), top_k=1))  # apple's vector
print(store2.search(query="", vectors=vec(2), top_k=1))  # cherry's vector
```

For the second state, corrupt the index instead:

```python
with open(f"{PATH}/repro.faiss", "wb") as f:
    f.write(b"garbage")
store3 = FAISS(collection_name="repro", path=PATH, embedding_model_dims=DIMS)
store3.insert([vec(3)], payloads=[{"data": "durian"}], ids=["id-durian"])
# ValueError: Collection not initialized. Call create_col first.
```

### Expected Behavior

A failed load should leave the store in a consistent state. Either refuse to
start (raise, so the caller knows the persisted data is damaged) or start
clean with an empty index AND an empty mapping. The current behavior is worse
than either option: stale vectors sitting behind a mapping that claims the
index is empty.

### Actual Behavior

Searching with apple's own vector returns cherry's memory at score 1.0, and
cherry's vector finds nothing:

```
Failed to load FAISS index: Unterminated string starting at: line 7 column 7 (char 90)
after reload: index.ntotal=2, index_to_id={}, docstore keys=[]
after inserting 'cherry': index.ntotal=3, index_to_id={0: 'id-cherry'}
search(apple vector) -> [('id-cherry', {'data': 'cherry'}, 1.0)]
search(cherry vector) -> []
```

With a corrupted `.faiss` file:

```
Failed to load FAISS index: Error in ... faiss::read_index_up(IOReader*, int) ...: Index type 0x62726167 ("garb") not recognized
after reload: index=None
insert() -> ValueError: Collection not initialized. Call create_col first.
search() -> ValueError: Collection not initialized. Call create_col first.
```

### Environment

- mem0 version: repository `main` at `39bc0233`
- Python version: 3.12.3
- OS: Linux (Ubuntu 24.04)

### What I Ran

```
python -m venv .venv
.venv/bin/pip install -e . faiss-cpu
.venv/bin/python repro_faiss_load.py   # the script above, both scenarios
```

### What I Saw

Output pasted above, unedited. The interesting line is
`search(apple vector) -> [('id-cherry', ..., 1.0)]`: the store hands back a
different memory's payload as a perfect match.

### Why This Is a Bug

`insert()` computes `starting_idx = len(self.index_to_id)` and trusts it to
equal `self.index.ntotal`. The `_load()` failure path breaks that invariant
on purpose (clears the dicts) without touching the index, so the very next
insert desynchronizes them. The second case contradicts the intent of the
`except` block itself: it resets state to keep going, but forgets that
`create_col()` is what creates the index, so "keep going" is impossible.

A related note: `_save()` is not atomic (`faiss.write_index` then
`json.dump`, each straight onto the live file), which is the realistic way
users end up with exactly one of the two files corrupted. Writing to a temp
file and renaming would close that window. Happy to include that in a fix if
wanted.

### What I Ruled Out

- #3246 / #4178: that was `delete()` renumbering desync, fixed in April. This
  one starts from the load path and reproduces on current `main` after that
  fix.
- #4833 (pickle CVE fix) introduced the json format, but the swallowed-failure
  behavior predates it: the original pickle-based `_load()` (#2461) had the
  same reset-the-dicts-keep-the-index shape.
- Not a FAISS library problem: `faiss.read_index` and the json parser both
  fail loudly; it's the handler that turns those into silent inconsistency.

### AI Assistance

Elioooon:
Fixes #7117

## Summary

- Reset the in-memory FAISS index when persisted index/docstore loading fails.
- Recreate an empty collection after the load failure fallback clears `docstore` and `index_to_id`.
- Add regression tests for both corrupted JSON docstore and corrupted FAISS index paths.

## Validation

- `.venv/bin/python -m pytest tests/vector_stores/test_faiss.py -q -k 'failed_json_load or failed_index_load'`
- `.venv/bin/python -m pytest tests/vector_stores/test_faiss.py -q`
- `python3 -m ruff check mem0/vector_stores/faiss.py tests/vector_stores/test_faiss.py`
- `python3 -m ruff format --check mem0/vector_stores/faiss.py tests/vector_stores/test_faiss.py`

## Risk

This keeps the existing load-failure fallback behavior of starting from an empty local collection, but makes the reset complete: stale FAISS vectors are discarded along with their mappings, and a corrupted index no longer leaves later `insert()` calls stuck behind `Collection not initialized`. It does not change successful JSON or legacy pickle loading, malicious pickle handling, distance strategy selection, search scoring, or persistence format.

### DIFF
--- a/mem0/vector_stores/faiss.py
+++ b/mem0/vector_stores/faiss.py
@@ -221,8 +221,10 @@ class FAISS(VectorStoreBase):
             raise ValueError(f"Failed to load FAISS docstore: potentially malicious pickle file. {e}") from e
         except Exception as e:
             logger.warning(f"Failed to load FAISS index: {e}")
+            self.index = None
             self.docstore = {}
             self.index_to_id = {}
+            self.create_col(self.collection_name)
 
     def _save(self):
         """Save FAISS index and docstore to disk using JSON format (secure)."""
--- a/tests/vector_stores/test_faiss.py
+++ b/tests/vector_stores/test_faiss.py
@@ -190,9 +190,7 @@ def test_search_with_filters_overfetch_not_truncated(faiss_instance, mock_faiss_
     search_indices = np.array([[0, 1, 2, 3]])
     mock_faiss_index.search.return_value = (search_scores, search_indices)
 
-    results = faiss_instance.search(
-        query="test query", vectors=query_vector, top_k=2, filters={"category": "A"}
-    )
+    results = faiss_instance.search(query="test query", vectors=query_vector, top_k=2, filters={"category": "A"})
 
     # Two matching vectors exist among the over-fetched set, so we must get top_k of them.
     assert len(results) == 2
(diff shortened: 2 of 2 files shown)


Elioooon:
Closes #7117

PR #7151, opened 2026-08-28
fix(qdrant): preserve wildcard field-exists filters

### LINKED ISSUE #7113: Qdrant vector store silently drops the '*' metadata-filter wildcard instead of matching field presence (diverges from OpenSearch/Pinecone)
## Problem

`search()` / `get_all()` / `delete_all()` filters that use the documented `"*"` wildcard (`{"key": "*"}` — "wildcard match (any value)", per the `Memory.search()` docstring in `mem0/memory/main.py`) are **silently discarded** by the Qdrant adapter. Points that do not have the filtered payload field at all match the filter, so results (and, via `get_all`, listing; and via `delete_all`, deletion) include memories that lack the field entirely.

This diverges from how every other distance-normalized store treats the same documented operator:

- OpenSearch: `'*'` compiles to `{"exists": {"field": ...}}` — field must be present (`mem0/vector_stores/opensearch.py:38-45`)
- Pinecone: fixed to field-exists semantics in #6652 ("`'*'` metadata filter is `$eq`-matched instead of field-exists")

So identical filter code produces different result sets depending on the configured vector store.

Note this is not only an over-return issue on `search()`: the same `_create_filter()` path backs `list()`, and `Memory.delete_all()` lists-then-deletes in batches, so `m.delete_all(filters={"user_id": "u1", "tag": "*"})` will also delete user u1's memories that have no `tag` field.

## Steps to Reproduce

Run against a source checkout of `main` (39bc0233). Uses qdrant-client local mode, no server needed:

```python
import tempfile, uuid
from pathlib import Path
from mem0.vector_stores.qdrant import Qdrant

tmp = Path(tempfile.mkdtemp(prefix="qdrant_star_repro_"))
store = Qdrant(collection_name="star_repro", embedding_model_dims=4, path=str(tmp / "qdrant"))

id_with = str(uuid.uuid4())
id_without = str(uuid.uuid4())
store.insert(
    vectors=[[0.1, 0.2, 0.3, 0.4], [0.4, 0.3, 0.2, 0.1]],
    payloads=[
        {"data": "has category", "user_id": "u1", "category": "sports"},
        {"data": "NO category field at all", "user_id": "u1"},
    ],
    ids=[id_with, id_without],
)

results = store.search(
    query="", vectors=[0.1, 0.2, 0.3, 0.4], top_k=5,
    filters={"user_id": "u1", "category": "*"},
)
for p in results:
    print(p.id, p.payload.get("category", "<ABSENT>"))
```

## Expected Behavior

Only the point that actually has a `category` payload value should match; `{"category": "*"}` should mean "has any value for category" (field exists), as it does on the OpenSearch and Pinecone backends.

## Actual Behavior

The point with no `category` field matches too. Real output from the script above on my machine:

```
filters={'user_id': 'u1', 'category': '*'} returned 2 point(s):
  id=f044b1e0-af19-4441-827d-0e7ce35a593f  category='sports'
  id=d2df9b64-6c8d-4b92-9eaf-cbb755c24611  category='<ABSENT>'
AssertionError: BUG CONFIRMED: point d2df9b64-6c8d-4b92-9eaf-cbb755c24611 with NO 'category' payload field matched filters {'category': '*'} -- the wildcard was silently discarded.
```

(The assert is from my repro harness checking the returned ids.)

## How You Verified This

**What I Ran:** the exact script above, Python 3.12.14, Windows 11, qdrant-client 1.19.0 (local persistence mode), mem0 checked out at commit `39bc0233` (2026-08-24).

**What I Saw:** output pasted above — both points returned, including the one whose payload has no `category` key.

**Why This Is a Bug:** the wildcard's meaning is defined by the repo itself: `Memory.search()` docstring documents `{"key": "*"} - wildcard match (any value)`; `mem0/vector_stores/opensearch.py:38-45` implements it as field-exists with a comment calling it "a documented Platform pattern"; and #6652 established field-exists as the correct Pinecone behavior for the same operator. Skipping the condition changes query semantics instead of translating them.

**What I Ruled Out:** not a qdrant-client quirk — building the intended filter manually (`Filter(must=[FieldCondition(key="user_id", match=MatchValue(value="u1"))], must_not=[IsEmptyCondition(is_empty=PayloadField(key="category"))])`) against the same collection returns only the point that has `category`. The adapter simply never emits such a condition.

## Root Cause

`mem0/vector_stores/qdrant.py`, `_build_field_condition()` (lines 280-284):

```python
if not isinstance(value, dict):
    if value == "*":
        # Wildcard: match any value. Qdrant has no direct "field exists"
        # condition via FieldCondition, so we skip this filter (match all).
        return None
```

The returned `None` makes `_create_filter()` drop the condition entirely (line 402-404), so the only remaining constraint is `{"user_id": "u1"}`. The inline rationale ("no direct field exists condition") is out of date: `qdrant_client.models.IsEmptyCondition` + `PayloadField` provide exactly this, and `pyproject.toml` already pins `qdrant-client>=1.12.0`.

## Proposed Fix

In `_create_filter()`/`_build_field_condition()`, translate the wildcard into a field-exists constraint instead of dropping it, e.g. collect wildcard keys into `Filter.must_not=[IsEmptyCondition(is_empty=PayloadField(key=key))]` (merged into the returned `Filter`). I verified locally that this yields field-exists semantics on the same fixture (returns only `['sports']`). Minimal change, no new dependencies.

## Regression Test Sketch

In `tests/vector_stores/test_qdrant.py` (local client, mirrors the repro): insert two points for one `user_id`, one with the wildcarded field and one without; assert `search(filters={"user_id": ..., "field": "*"})` returns only the first. Fails on current `main`, passes with the fix.

## Environment

- mem0ai: source checkout of main @ `39bc0233` (2026-08-24), package version 2.0.19
- Python 3.12.14, Windows 11
- qdrant-client 1.19.0 (local mode)

## AI Assistance

mikemikimike:
## Summary

Closes #7113.

Qdrant previously discarded the documented `"*"` metadata filter, allowing points without the requested payload field to match. This change maps the wildcard to a field-presence condition using `IsEmptyCondition` and keeps the existing equality and logical filter behavior intact.

## Changes

- Return a Qdrant `IsEmptyCondition` for wildcard field values.
- Add the condition to `must_not` so missing or empty fields are excluded.
- Add unit coverage for direct, nested, wildcard-only, and real local Qdrant search behavior.

## Tests

- `python -m pytest tests/vector_stores/test_qdrant.py -q -k 'not reset_clears_points_on_local_qdrant'` (96 passed, 1 deselected)
- `ruff check mem0/vector_stores/qdrant.py tests/vector_stores/test_qdrant.py`
- `ruff format --check mem0/vector_stores/qdrant.py tests/vector_stores/test_qdrant.py`
- `git diff --check`

The full focused file reports one pre-existing Windows Qdrant SQLite teardown failure in `test_reset_clears_points_on_local_qdrant`; the changed-path tests pass when that environment-specific test is excluded.

AI assistance was used during implementation and review.

### DIFF
--- a/mem0/vector_stores/qdrant.py
+++ b/mem0/vector_stores/qdrant.py
@@ -12,6 +12,8 @@ from qdrant_client.models import (
     MatchExcept,
     MatchText,
     MatchValue,
+    IsEmptyCondition,
+    PayloadField,
     PointIdsList,
     PointStruct,
     PointVectors,
@@ -262,7 +264,7 @@ class Qdrant(VectorStoreBase):
             for v in range_kwargs.values()
         )
 
-    def _build_field_condition(self, key: str, value) -> Optional[FieldCondition]:
+    def _build_field_condition(self, key: str, value) -> Optional[models.Condition]:
         """
         Build a single FieldCondition from a key-value filter pair.
 
@@ -274,14 +276,15 @@ class Qdrant(VectorStoreBase):
             value: A scalar for simple equality, or a dict with one operator key.
 
         Returns:
-            Optional[FieldCondition]: The Qdrant field condition, or None if the
-            value is the wildcard '*' (match any / field exists — skip filter).
+            Optional[models.Condition]: The Qdrant condition, or None when no
+            condition can be built.
         """
         if not isinstance(value, dict):
             if value == "*":
-                # Wildcard: match any value. Qdrant has no direct "field exists"
-                # condition via FieldCondition, so we skip this filter (match all).
-                return None
+                # A wildcard means that the payload field must be present.  An
+                # IsEmptyCondition matches both missing and empty payload fields,
+                # so negating it gives the desired field-exists semantics.
+                return IsEmptyCondition(is_empty=PayloadField(key=key))
             if isinstance(value, list):
                 # List shorthand: {"field": ["a", "b"]} treated as in-operator.
                 return FieldCondition(key=key, match=MatchAny(any=value))
@@ -401,7 +404,10 @@ class Qdrant(VectorStoreBase):
             else:
                 condition = self._build_field_condition(key, value)
                 if condition is not None:
-                    must.append(condition)
+                    if value == "*":
+                        must_not.append(condition)
+                    else:
+                        must.append(condition)
 
         if not any([must, should, must_not]):
             return None
--- a/tests/vector_stores/test_qdrant.py
+++ b/tests/vector_stores/test_qdrant.py
@@ -14,6 +14,7 @@ from qdrant_client.models import (
     MatchExcept,
     MatchText,
     MatchValue,
+    IsEmptyCondition,
     PointIdsList,
     PointStruct,
     PointVectors,
@@ -240,7 +241,7 @@ class TestQdrant(unittest.TestCase):
         self.assertEqual(call_args["collection_name"], "test_collection")
         self.assertEqual(call_args["query"], vectors)
         self.assertEqual(call_args["limit"], 1)
-        
+
         # Verify that a Filter object was created
         query_filter = call_args["query_filter"]
         self.assertIsInstance(query_filter, Filter)
@@ -290,10 +291,10 @@ class TestQdrant(unittest.TestCase):
         """Test _create_filter with multiple filters."""
         filters = {"user_id": "alice", "agent_id": "agent1", "run_id": "run1"}
         result = self.qdrant._create_filter(filters)
-        
+
         self.assertIsInstance(result, Filter)
         self.assertEqual(len(result.must), 3)
-        
+
         # Check that all conditions are present
         conditions = [cond.key for cond in result.must]
         self.assertIn("user_id", conditions)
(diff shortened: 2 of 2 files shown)


dlowzzxx (approved):
Verified locally: `_build_field_condition("category", "*")` now returns `IsEmptyCondition(category)` and `_create_filter({"category": "*", "user_id": "alice"})` yields `must_not=[is_empty(category)]` + `must=[user_id]`, so wildcard-only queries still enforce presence. The integration test with two points (one with category, one without) correctly returns only the point with the field. Mirrors the fix already applied to other stores. LGTM — restores field-exists semantics without breaking equality lists.

mikemikimike:
Hi maintainers — this PR was auto-closed by the PR Gate because the linked issue (#7113) has not yet been labeled `accepted`.

Issue #7113 includes a runnable reproduction, expected/actual behavior, root-cause analysis, and a proposed fix. This PR implements that fix with regression coverage, and the focused tests and checks pass apart from the unrelated Windows Qdrant SQLite teardown case noted in the description.

Could someone please review #7113 and add the `accepted` label if the issue is in scope? The PR Gate should reopen this PR automatically once that is done. Thank you!

PR #7203, opened 2026-09-01
refactor(integrations): shared agent plugin runtimes and native adapters

kartik-mem0:
## Summary

This refactors the coding-agent integrations around one maintainable source tree while keeping each installed plugin self-contained.

- Adds `integrations/agent-plugin-core/` for shared Python and TypeScript behavior, canonical skills, deterministic builds, telemetry, and conformance tests.
- Adds one portable Agent Plugins v1 package at `integrations/mem0-agent-plugin/`.
- Keeps Claude Code, Cursor, Codex, Kimi, and Antigravity in recognizable host directories containing only native manifests, adapters, and generated install payloads.
- Reuses the TypeScript core from OpenClaw, OpenCode, Pi Agent, and DeepSeek without renaming their packages.
- Removes the obsolete `integrations/agent-plugins/`, `integrations/mem0-plugin/`, per-host portable clones, Harmless scaffolding, symlink layout, and committed aggregate `dist/` tree.
- Removes Dream/consolidation and pin behavior from OpenClaw, OpenCode, and Pi Agent. Claude Code stays at `0.3.0`; the other agent plugins align at `0.3.0`, while OpenClaw moves to `1.1.0` for the breaking API removal.

## Architecture

```text
integrations/
├── agent-plugin-core/
│   ├── python/          # shared Python lifecycle
│   ├── typescript/      # shared TS lifecycle and telemetry
│   ├── skills/          # six canonical skill templates
│   ├── build/           # deterministic native/portable builder
│   └── conformance/     # one offline verification command
├── mem0-agent-plugin/   # portable Agent Plugins v1 package
├── claude-code-plugin/
├── cursor-plugin/
├── codex-plugin/
├── kimi-plugin/
├── antigravity-plugin/
├── openclaw/
├── opencode-plugin/
├── pi-agent-plugin/
└── deepseek-plugin/
```

Native Python directories contain generated `core/` and `skills/` files because installed plugins must be self-contained. Their only editable source is `agent-plugin-core/`; drift checks fail if a generated copy changes independently. Host adapters contain only native event and output-schema translation.

## Claude Code compatibility

Claude Code remains the behavioral reference. Its public manifest, MCP configuration, hook declaration, and `agents/sidekick.md` are byte-for-byte identical to `main`. A real current-branch Claude run spawned `mem0:sidekick`, completed it successfully, and returned the expected evidence. Shared-core regressions cover capture, recall, scoping, redaction, detached flushes, controls, failure isolation, and Sidekick correlation.

## Sidekick architecture

- **Claude Code:** bundled `mem0:sidekick`, parent-memory context, native lifecycle hooks, and host-enforced Git worktree isolation.
- **Codex:** the plugin cannot bundle a named custom agent under Codex's current packaging contract. Every native Codex subagent now receives the parent turn's retrieved memory context and has its completion recorded; named agents remain project/user configuration under `.codex/agents/` or `~/.codex/agents/`.
- **Cursor:** bundled native Sidekick plus `subagentStart` / `subagentStop` capture. Cursor cannot inject context at subagent start, so the Sidekick searches Mem0 itself.
- **Kimi Code:** bundled native Sidekick with parent-context lifecycle hooks; filesystem isolation is owned by Kimi.
- **Antigravity:** bundled native Sidekick that searches Mem0 itself; no Claude-style worktree guarantee.
- **DeepSeek Harness:** subagents are separate host-composition packages. The Mem0 Cordis package does not register a named Sidekick or claim child filesystem isolation.

The memory implementation is shared. Native agent declarations remain small host-owned files because their schemas and isolation guarantees differ.

## Host behavior

| Host | Runtime | Host-specific boundary |
| --- | --- | --- |
| Claude Code | Python | Existing hooks, MCP, skills, controls, and worktree sidekick |
| Cursor | Python | Cursor manifest, event adapter, and native sidekick definition |
| Codex | Python | Codex manifest and lifecycle adapter; named custom agents remain project/user configuration |
| Kimi Code | Python | Kimi manifest, transcript adapter, and native sidekick definition |
| Antigravity | Python | Antigravity manifest, adapter, and native sidekick definition |
| OpenClaw | TypeScript | OpenClaw memory backend and CLI |
| OpenCode | TypeScript/Bun | OpenCode native plugin API |
| Pi Agent | TypeScript | Pi native extension API |
| DeepSeek | TypeScript | DeepSeek Harness/Cordis API |

## Agent Plugins v1

`integrations/mem0-agent-plugin/` conforms to Agent Plugins v1.0.0: root `plugin.json`, root `mcp.json`, fixed `skills/` discovery, package-contained files, and `${PLUGIN_ROOT}` / `${PLUGIN_DATA}` usage. The native sibling directories intentionally follow their hosts' extension formats instead of pretending those host-specific hooks and agents are portable v1 components.

## Important fixes from live review

- Control skills now target the same host-specific data directory as hooks and MCP instead of falling back to the generic store.
- Detached Python flush workers inherit the selected host, data directory, source tag, and telemetry identity.
- Codex now maps native subagent start/stop events into the shared Sidekick lifecycle and applies them to every native subagent instead of assuming a plugin-bundled agent name.
- Cursor now maps its native Sidekick events and translates `subagent_id`, `subagent_type`, parent conversation, summary, and transcript fields.
- Missing host child IDs are correlated atomically in the shared SQLite core; Kimi's duplicate host-local SQL was removed.
- `--plugin-data-dir` now reaches the shared data-directory resolver instead of depending on a second host environment variable.
- Kimi Code 0.40.1 Stop events recover the completed assistant response from its indexed v2 wire transcript; repeated sidekick runs receive distinct IDs.
- Antigravity hooks use explicit `MEM0_CWD` when the host omits `workspacePaths`, and fail closed when neither scope is available.
- OpenClaw's manifest version now matches package `1.1.0`, and `openclaw mem0 status` no longer crashes before setup.
- Pi now accepts the `[mem0:<uuid>]` citation IDs it displays when a model updates or deletes a memory.
- Shared Python and TypeScript telemetry now remove `password`, `token`, `secret`, and `authorization` keys recursively, including objects nested inside lists.
- Stale `integrations/mem0-plugin` links now point to the portable `integrations/mem0-agent-plugin`; historical changelog entries retain their published wording.
- Local conformance installs run non-interactively under pnpm.

## Validation

- [x] Unified offline conformance: 30/30 checks passed after the review follow-ups
- [x] Final Python suites: 221 passed
- [x] Shared TypeScript core: 20 passed; type-check passed
- [x] OpenClaw: 427 passed; type-check, build, artifact check, and live add/search/delete passed
- [x] OpenCode: 31 passed; type-check, build, and artifact check passed
- [x] Pi Agent: 89 passed; type-check, build, and artifact check passed
- [x] DeepSeek: 46 passed; type-check, build, package, and artifact check passed
- [x] All six installable Python packages passed build and generated-file drift validation
- [x] Credentialed Mem0 Platform scoping suite: 8/8 scenarios passed with disposable namespaces
- [x] Ruff, docs `llms.txt` coverage, and `git diff --check` passed

### Real host matrix

| Host | Verified | Remaining external boundary |
| --- | --- | --- |
| Claude Code | Full native lifecycle, recall, scoping, redaction, failure isolation, and worktree Sidekick; native config remains identical to `main` | None |
| Codex 0.145.0-alpha.27 | Native install, MCP search, hooks, and a real parent → subagent → parent turn. Local evidence contains matching start/stop rows, the real child ID and transcript, and `SIDEKICK_OK` | Codex plugin packaging does not distribute named custom agents; define them under `.codex/agents/` or use built-in agents |
| Cursor 3.18.25 / Agent 2026.09.02 | GUI loaded the local plugin with zero failures and connected its one Mem0 MCP tool. Native Sidekick wire replay produced one correlated start/stop row with `SIDEKICK_OK` | Standalone Cursor Agent model turn requires `cursor-agent login`; GUI clicks still need human confirmation |
| Kimi Code 0.40.1 | Official runtime installed this PR's plugin (6 skills, 9 hooks, Sidekick, and MCP). A real OpenRouter Sidekick turn received hook context, called `search_memories`, and returned `KIMI_SIDEKICK_OK 0` | `CHOKIDAR_USEPOLLING=1` remains necessary for the macOS watcher failure |
| Antigravity 1.1.26 | Clean bundle validation found 6 skills, 1 agent, 1 MCP server, and 1 hook set. A real Sidekick turn called Mem0 and recorded `SIDEKICK_OK 3` | Plugin-scoped MCP registration and empty workspace paths remain host-runtime issues; global MCP registration plus `MEM0_CWD` are documented workarounds |
| OpenClaw 2026.4.26 | Local linked install, configuration, status, and live add/search/delete with cleanup | None |
| OpenCode 1.4.6 | Current PR build discovered with all seven commands. A real OpenRouter model added, searched, deleted, and confirmed deletion of a disposable memory | None |
| Pi Agent 0.83.0 | A real OpenRouter model added, recalled, deleted the displayed `mem0:<uuid>` citation, and confirmed deletion. The live failure found here has a regression test | None |
| DeepSeek Harness 0.1.1-rc.2 | 46 tests plus a real OpenRouter add in one session, exact recall in a fresh session, scoped cleanup, and fresh-session deletion confirmation | `CHOKIDAR_USEPOLLING=1` avoids the macOS watcher failure |

The Antigravity limitation is not hidden: its installed manifest and MCP shape match Antigravity's documentation, but the current host still requires the global MCP registration fallback and can omit the workspace path supplied to hooks. This branch documents `agy mcp add`, uses `MEM0_CWD` as an explicit CLI scope fallback, and skips capture when no repository scope is available.

## AI use

Implementation and review were assisted by Codex. The validation results above were run locally against this branch.


integrations/antigravity-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
+{
+  "description": "Create memories from coding sessions, search them during later work.",
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/antigravity/hook.py\" session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness antigravity session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/antigravity-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/antigravity/hook.py\" user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness antigravity user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/antigravity-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": ".*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/antigravity/hook.py\" post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness antigravity post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/antigravity-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/antigravity/hook.py\" stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness antigravity stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/antigravity-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "SessionEnd": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/antigravity/hook.py\" flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness antigravity flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/codex-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "SessionEnd": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/codex/hook.py\" flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness codex flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/kimi-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
+{
+  "description": "Create memories from coding sessions, search them during later work.",
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/kimi/hook.py\" session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness kimi session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/kimi-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "SessionEnd": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/kimi/hook.py\" flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness kimi flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/kimi-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": ".*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/kimi/hook.py\" post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness kimi post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/codex-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/codex/hook.py\" stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness codex stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/kimi-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/kimi/hook.py\" stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness kimi stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/agent-plugins/core/python/hook_runner.py
```diff
@@ -0,0 +1,349 @@
...
+
+    base_actions = ["session-start", "user-prompt", "post-tool", "stop", "flush"]
+    all_actions = base_actions + list((extra_actions or {}).keys())
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("action", choices=all_actions)
+    parser.add_argument("--reason", default="manual")
+    parser.add_argument("--plugin-data-dir", default="")
+    args = parser.parse_args()
+
```
kartik-mem0:
```suggestion

    if args.harness:
        configure_harness(args.harness)
        telemetry.init(harness=args.harness)

```

integrations/cursor-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
+{
+  "description": "Create memories from coding sessions, search them during later work.",
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/cursor/hook.py\" session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness cursor session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/plugin-core/hook_runner.py
```diff
@@ -0,0 +1,349 @@
...
+    if automatic_flush_reasons is None:
+        automatic_flush_reasons = {"session-end"}
+
+    base_actions = ["session-start", "user-prompt", "post-tool", "stop", "flush"]
+    all_actions = base_actions + list((extra_actions or {}).keys())
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("action", choices=all_actions)
+    parser.add_argument("--reason", default="manual")
+    parser.add_argument("--plugin-data-dir", default="")
```
kartik-mem0:
```suggestion
    parser.add_argument("--plugin-data-dir", default="")
    parser.add_argument("--harness", default="")
```

integrations/cursor-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": ".*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/cursor/hook.py\" post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness cursor post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/harmless-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": ".*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/harmless/hook.py\" post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness harmless post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/harmless-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/harmless/hook.py\" user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness harmless user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/cursor-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "SessionEnd": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/cursor/hook.py\" flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness cursor flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/openclaw-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
+{
+  "description": "Create memories from coding sessions, search them during later work.",
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/openclaw/hook.py\" session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness openclaw session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/codex-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
+{
+  "description": "Create memories from coding sessions, search them during later work.",
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/codex/hook.py\" session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness codex session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/codex-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/codex/hook.py\" user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness codex user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/openclaw-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": ".*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/openclaw/hook.py\" post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness openclaw post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/kimi-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/kimi/hook.py\" user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness kimi user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/harmless-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
+{
+  "description": "Create memories from coding sessions, search them during later work.",
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/harmless/hook.py\" session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness harmless session-start --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/harmless-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "SessionEnd": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/harmless/hook.py\" flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness harmless flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/plugin-core/hook_runner.py
```diff
@@ -0,0 +1,349 @@
...
+    try:
+        raise SystemExit(run(
+            record_stop_fn=record_stop_fn,
+            extra_actions=extra_actions,
+            data_dir_env=data_dir_env,
+            automatic_flush_reasons=automatic_flush_reasons,
+        ))
+    except Exception as exc:
+        log_failure(exc)
+        raise SystemExit(0)
```
kartik-mem0:
```suggestion
        raise SystemExit(0)


if __name__ == "__main__":
    entry_point()
```

integrations/openclaw-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/openclaw/hook.py\" stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness openclaw stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/openclaw-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "SessionEnd": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/openclaw/hook.py\" flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness openclaw flush --reason session-end --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/openclaw-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/openclaw/hook.py\" user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness openclaw user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/harmless-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/harmless/hook.py\" stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness harmless stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/codex-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": ".*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/codex/hook.py\" post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness codex post-tool --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/cursor-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/cursor/hook.py\" stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness cursor stop --plugin-data-dir \"${PLUGIN_DATA}\"",
```

integrations/cursor-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,62 @@
...
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${PLUGIN_ROOT}/adapters/cursor/hook.py\" user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```
kartik-mem0:
```suggestion
            "command": "python3 \"${PLUGIN_ROOT}/hooks/hook.py\" --harness cursor user-prompt --plugin-data-dir \"${PLUGIN_DATA}\"",
```

karthik-indla:
Really nicely structured refactor — for a 352-file / +28k−20k change, the thing that makes it reviewable is that the safety-critical invariants hold, and they do. I focused the review on verifying those rather than reading every generated copy, and wanted to share what checked out plus a few small follow-ups.

## What I verified (all holding up)

- **Single source of truth + drift gate is real.** `build.py --check` does genuine byte-equality (rebuild-to-temp → compare, fails on any diff/missing/stale file), and it's wired into the Python CI job across all five native hosts and the portable bundle. All seven `memory_core.py` copies share one sha256 — the Claude Code copy is a full copy, not a shim (the `35+/361-` stat is just the delta from `main`). Skills render from the `.tmpl` templates with per-host substitution. The TS hosts use compile-time re-exports rather than copies, so their guarantee correctly rests on `tsc`/tests — nice touch.
- **Claude Code compatibility holds.** `plugin.json` (name `mem0` v0.3.0), `hooks/hooks.json`, `.mcp.json`, `agents/sidekick.md`, and the marketplace entry are byte-identical `main`→branch and aren't in the changed-file list. The MCP server path resolves on the branch. (For anyone diffing: the right baseline is `claude-code-plugin@main`, which already existed — the removed `mem0-plugin` was a separate monolith, so that particular diff looks large but isn't a Claude Code regression.)
- **CI gate is cleanly rewired.** Both new workflows are fully registered (filter → output → call → `needs`), both removed ones fully de-registered with no dangling references, all `uses:` resolve, and the `opencode-plugin-cd.yml` edit is just a `working-directory` update with the OIDC-pinned filename intact.
- **Removed dirs + marketplace.** `agent-plugins/` and `mem0-plugin/` are fully gone, `.opencode-plugin` is relocated (not lost) to `integrations/opencode-plugin/`, all six marketplace files point only at existing dirs, and `release.yml` needs no change.
- **Docs/llms.txt.** The new `kimi.mdx` is registered in both `docs.json` and `llms.txt` (with a scope tag + `Use when …`), the coverage gate passes (0 missing / 0 stale), and the page matches what `kimi-plugin/` actually ships.

## A few small things to tidy

1. **One spot the "no stale references" net missed:** `skills/mem0-integrate/SKILL.md` still points at the old directory in two places — line 49 (`.../tree/main/integrations/mem0-plugin`) and the decision-table row at line 112. That file isn't in the diff, so it's an easy one to miss; both would 404 after merge. A quick repoint to `mem0-agent-plugin` (or the relevant per-host dir) closes it.

2. **Telemetry redaction — one narrow gap + a test note.** The redaction genuinely recurses through nested dicts *and* list elements, and fail-open + `MEM0_TELEMETRY=false` are solid in both runtimes — all good. Two small things: the sensitive-*key* allowlist doesn't include `password`/`token`/`secret`/`authorization`, so a structured `{"password": "..."}` whose value doesn't match a token/PEM pattern can slip through (the inline-text regexes still catch the common free-text cases). And the nested-filtering tests are lighter than the description suggests — the TS test covers dict-in-dict, but there's no Python nested test and neither language tests a secret inside a list element. Adding those two tests (and optionally a few more keys to the removal set) would fully back the claim.

3. **Worth a quick confirm — changelog edits.** The PR rewrites a few already-published, dated changelog entries to drop the Dream / auto-dream mentions (OpenCode, Pi, OpenClaw). Totally fine if those features were genuinely removed here; just flagging that editing historical entries can read as retroactive, and it may rebase-conflict with the Dream work currently landing on `main`.

None of these touch the core architecture, which is solid — the drift gate, Claude Code byte-identical guarantee, and CI wiring are exactly the things you'd want locked down for a change this size, and they are. The stale-ref fix is the only pre-merge must; the rest are easy follow-ups. And since this touches `.github/workflows/`, it'll want the usual maintainer sign-off on the workflow changes too.

_(Review was done by verifying invariants against the branch rather than a full line-by-line read, given the size — real editor/harness smoke tests remain the right final gate, as you noted.)_

kartik-mem0:
Addressed in `6036e9fc4` — thanks for the focused review.

- Repointed both stale `skills/mem0-integrate/SKILL.md` references to `integrations/mem0-agent-plugin`.
- Added `password`, `token`, `secret`, and `authorization` to the shared Python and TypeScript telemetry private-key sets. New tests cover nested objects inside list elements in both runtimes.
- Restored the published Dream/auto-dream wording in historical changelog entries. The removals now appear only in new release entries.
- Aligned Cursor, Codex, Kimi, Antigravity, the portable bundle, OpenCode, Pi Agent, and DeepSeek at `0.3.0`; Claude Code remains `0.3.0`. OpenClaw is `1.1.0` because its public Dream/pin APIs were removed.

Validation: unified conformance 30/30; Python 221; shared TypeScript 20; OpenClaw 427; OpenCode 31; Pi 89; DeepSeek 46; all builds, type-checks, generated drift checks, artifact checks, docs coverage, and GitHub CI pass.

No workflow file was changed by this follow-up; the existing workflow changes still require the normal maintainer sign-off you noted.

karthik-indla:
Did a second, deeper pass focused on the **runtime logic** this time — scoping/isolation, the shared `memory_core` capture/recall pipeline, the per-host adapters, the MCP server, and the build/conformance gate (my first pass only checked the structural invariants). Good news up front: the architecture holds up under scrutiny, fail-open is near-universal, and the MCP surface is safe. A handful of things are worth a look before merge, grouped by severity below. Several of the scoping ones look pre-existing (`_legacy_` paths) — flagging them because this consolidation is a natural moment to harden them, not as regressions this PR introduced.

## Worth addressing before merge

1. **Secret redaction misses JSON-shaped secrets → can reach the platform.** `redact()` (`agent-plugin-core/python/memory_core.py`) only catches `key=value` / `key: value` with an *unquoted* value, so a tool printing `{"apiKey": "…"}` or a 40-char AWS *secret* access key (no fixed prefix) passes through unredacted. It's stored locally, and on a **failed** command it flows through `build_semantic_evidence` and is POSTed with the extraction batch. A JSON-aware pattern (`"(?i)"(api[_-]?key|password|secret|token)"\s*:\s*"[^"]+"` → `"<key>": "[REDACTED]"`) would close it. Relatedly, transcript messages are copied into episodes **without** `redact`/`bounded` in the shared core — the Claude adapter self-redacts, but the core makes no guarantee for future harnesses, and they're uncapped (a huge entry risks a 413 that burns all flush retries). This is the same root theme as the telemetry key-set fix from the earlier round — worth closing the content side too.

2. **Cursor adapter fails *closed*.** Every other host routes through `entry_point` / a `try/except` and exits 0 on any error (the "memory never blocks the agent" invariant). `cursor-plugin/hooks/adapter.py` calls `hook_runner.run()` directly with no guard, so a transient SQLite lock or unexpected payload surfaces an uncaught traceback. One-line fix — wrap `main()` like Kimi/Antigravity, or go through `entry_point`.

3. **Host-blind git namespace can cross-leak team memories.** `_legacy_project_id` drops the git host, so `github.com/acme/api` and `gitlab.acme.internal/acme/api` both resolve to `project_id="acme-api"`. Shared/team memories carry no `user_id`, so one repo can read (and `forget` can delete) the other's team memories. The host-aware `identity` is already computed but isn't used in the filter — folding host into the shared lane (or the `project_id`) would fix it. Narrow precondition (same owner+name on two hosts), but silent when it hits, and untested.

4. **The safety gate doesn't cover the native install surface.** Schema validation in `build/validate.py` runs only for `kind=="portable"`, and `build --check` compares hand-authored native files (`.mcp.json`, `hooks.json`, `plugin.json`, adapters) against copies of themselves — so the byte-identical drift guarantee genuinely protects only `core/` + `skills/`, not the manifests/adapters people actually install. Everything's correct today (the Claude manifests are byte-identical to `main`), but a fat-fingered native manifest would ship green. Parsing/schema-validating native `*.json` in `validate_bundle` for `kind=="native"`, plus an installed→generated stale-file check, would make the SSOT guarantee cover the whole native tree.

## Follow-up hardening (non-blocking)

- **Same-user scope collisions:** the personal ("mine") lane uses a basename `app_id` with no path hash, so identically-named non-git dirs share personal memories for one user; Pi's `appId` is also folder-basename, colliding any two same-named projects. (`parseProjectFromRemote`, the TS primitive that would match Python's owner/repo scheme, is currently dead code — nothing calls it.)
- **TS core** lacks the empty/wildcard `app_id` guard the Python side enforces, and `scope:"global"` emits `app_id:"*"` (a model-selectable widening). Worth mirroring Python's `_scope_value` refusal.
- **`_wait_for_event`** treats persistent HTTP 4xx/5xx as transient (`HTTPError` is a `URLError`), so a hard status polls until the 120s deadline instead of failing fast.
- **Per-host capture divergence** worth a capability matrix: Antigravity records only the first user prompt of a conversation (later-turn intent isn't captured); Codex has no tool-failure event (failed tools recorded as success) and does no payload normalization (unverified against real Codex field names); Cursor may double-record the assistant response (`afterAgentResponse` + `stop` both call `record_stop`) and its sidekicks get no recalled context (telemetry still reports `context_chars>0`); Kimi's sidekick stop-correlation is LIFO and can mis-attribute with overlapping subagents (the test only checks the set, so it passes through the swap).
- **A couple of robustness edges:** `prepare_flush` reads unflushed events outside the claim transaction (low-probability duplicate-extraction race — `BEGIN IMMEDIATE` would serialize it), and the delayed-flush handoff rewrite in `flush_worker.py` isn't atomic (temp-file + `os.replace` like everywhere else). DeepSeek intentionally exposes a model-settable `userId` to target other users' memories — fine as a capability, but it's an injection surface with no confirmation gate.

## What checked out cleanly

- **Fail-open** everywhere except the Cursor wrapper above; all memory work runs in detached processes and every network call is timeboxed, so a capture/recall error can't crash the host.
- **MCP server** is local/stdio and exposes a single **read-only** `search_memories` tool (no delete surface at all); scope is pinned server-side (the agent can't pass `user_id`/`agent_id`, and its args only narrow within owned identities), wildcards are rejected, and the API key isn't leaked to the model. (Minor: the "9 tools" in the docs refers to the *remote* `mcp.mem0.ai` server, not this local one — worth distinguishing.)
- **No unconditional/global scope widening** in the Python core; **subagent memories correctly inherit the parent session's scope**; session scope is deterministically pinned across hooks even if `cwd` changes.
- **Capture pipeline, dedup, timeouts, and DB self-healing** are sound; the **builder is deterministic** and the **conformance suite runs real per-host test suites** (load-bearing, not cosmetic) — for core + skills.
- The earlier round's three items (stale `mem0-plugin` refs, telemetry key-set, changelog history) all remain addressed.

Net: the structure is solid and this is very much mergeable once the top group is handled — the redaction leak (#1) and the Cursor fail-closed (#2) are the most clear-cut; the namespace leak (#3) and the gate gap (#4) are narrower/partly pre-existing but worth closing while the code is open. Happy to dig into any of these further.

_(This pass verified logic against the branch source rather than a full line read of all 361 files; the real editor/harness smoke tests remain the right final gate, and the `.github/workflows/` changes still want the usual maintainer sign-off.)_

kartik-mem0:
Addressed in `b653b342` — thanks for the thorough second pass.

1. **JSON secret redaction**: added a JSON-aware pattern to `SECRET_PATTERNS` that catches `"apiKey": "..."` style secrets (covers `api_key`, `password`, `secret`, `secret_access_key`, `token`, `authorization`, `credential`, and `access/refresh/session_token` variants). New test `test_json_shaped_secrets_are_redacted`.

2. **Cursor fail-closed**: wrapped `__main__` in `try/except Exception` with `log_failure` + exit 0, matching Kimi/Antigravity/every other host.

3. **Host-blind git namespace**: `_project_id` now hashes the identity (which includes the git host) into the project ID for non-local repos, so `github.com/acme/api` and `gitlab.internal/acme/api` get separate team memory namespaces. New test `test_different_git_hosts_produce_different_project_ids`; updated the existing `test_same_named_folders` assertion.

4. **Native validation gate**: `validate_bundle` now parse-checks all `*.json` files in native bundles, catching malformed manifests before they ship. Two new tests cover the reject and accept paths.

All 7 `memory_core.py` copies are sha256-identical. Validation: Python 153, TS 5, telemetry 17 pass.

The non-blocking follow-ups (same-user scope collisions, TS `app_id` guard, `_wait_for_event` retry semantics, per-host capture divergence, flush atomicity, DeepSeek `userId`) are noted for follow-up after merge. No workflow files changed by this commit.

karthik-indla (dismissed):

kartik-mem0:
Pushed `623b7d87` with the extraction, lifecycle, and scope-guard follow-ups, plus the earlier compatibility fixes and release documentation. This update was implemented with Codex.

Addressed:

- Session-end workers flush the conversation captured by hooks without recording the final answer again. Shared response recording suppresses concurrent/repeated hooks while retaining identical answers to separate prompts.
- Shared transcript messages are redacted and bounded before extraction; oversized messages are split to enforce each request's input budget. JSON credential redaction also handles escaped quotes.
- Flush event selection and claims now run in one SQLite write transaction. Delayed handoff rewrites are atomic. Permanent HTTP polling errors fail promptly; transient failures remain retryable.
- Shared instructions are agent-neutral. The local read-only search tool and search skill no longer expose `run_id`.
- Antigravity captures later completed transcript turns incrementally. Cursor reports zero parent-injected context where its host uses Sidekick self-search. Subagent stops without an ID are not assigned to an arbitrary overlapping run. Codex uses available structured tool outcomes and records missing outcome information as unknown; its documented input fields already match the shared contract.
- TypeScript project/session scope helpers reject empty or wildcard identities. Pi/OpenCode's `scope: "global"` tool argument requires a user-configured global default. DeepSeek cross-user tool overrides require operator opt-in with `allowUserOverride: true`.
- Regenerated all six Python bundles at `0.3.1`, aligned their version metadata, and built all four TypeScript packages. Updated the SDK changelog for the full PR and every package; removed the refactor from Highlights and preserved published changelog history.

Validation on this revision:

- Unified conformance: **25/25**.
- **868 tests passed:** Python 249, shared TypeScript 25, OpenClaw 427, OpenCode 31, Pi 89, DeepSeek 47.
- All package builds/type checks, artifact checks, and generated-file checks passed.
- Six MCP server subprocess initialization/tool-schema smoke checks passed.
- Ruff and isort pre-commit checks passed; documentation coverage and compilation of all 10 edited MDX files passed.

Still open — this is not a claim that every review finding is resolved:

- **Namespace migration policy:** compatibility searches currently include the hashed and legacy shared IDs. Old records do not identify their Git host, so automatic legacy access retains the original same-name collision risk. Same-user non-Git/Pi basename collisions also remain. Strict isolation needs an explicit legacy-namespace binding/migration policy; we have not silently reassigned existing memories or guessed their ownership. This release decision is pending.
- **Actual host smoke tests:** the checks above cover adapters, package artifacts, and local MCP processes. They do not replace new end-to-end editor/Harness runs on this revision. No live Mem0 calls were made by this follow-up. Full Mintlify site validation also needs CI; the local CLI could not update its protected cache directory.

Please review the new revision and its fresh CI results. No workflow file was changed by this follow-up; the original workflow changes retain their maintainer sign-off requirement. Merge/release should wait for the namespace decision and actual-host validation.

kartik-mem0:
Sidekick follow-up in 6937194d:

- Added four offline subprocess checks covering the shipped Claude Code, Codex, Kimi, and Cursor subagent hooks. They verify parent repository/session continuity when the child uses a different directory, exclusion of another session's recalled memories, repeated-start behavior, and completion correlation. Cursor correctly records zero injected context and allows the child to search for itself.
- Clarified the Claude Code docs: Sidekick receives parent recalled memories, can search Mem0, and returns work for parent review. Direct lifecycle records stay local; subagent results present in the main transcript may support normal extraction.
- Validation: 39 shared build/conformance tests passed; 16 focused Sidekick/subagent tests passed (these counts overlap). Ruff/isort, MDX compilation, and llms.txt coverage passed. No live Mem0 calls.

Support is host-specific: Claude Code ships a Sonnet/worktree Sidekick; Cursor, Kimi, and Antigravity ship native Sidekick definitions without that same model/worktree guarantee. Codex integrates native subagent hooks but does not ship a named Sidekick. OpenClaw supports native subagent memory recall; OpenCode, Pi, DeepSeek, and the portable bundle do not ship a named Sidekick.

These checks establish adapter behavior, not live editor execution. Antigravity is not logged in locally, and actual-host smoke tests remain outstanding. This follow-up does not resolve the previously documented legacy namespace migration decision or establish release readiness.

soumil-rathi (dismissed):

soumil-rathi (approved):

PR #7222, opened 2026-09-04
docs(vectordbs/pinecone): document extra_params on the Python config table

88lin:
## Linked Issue

Documentation-only change (single file under `docs/`), so no accepted issue is linked.

## Description

The Pinecone config table documents `extraParams` on the TypeScript tab but has no `extra_params` row on the Python tab, even though the field exists and is used.

`PineconeConfig` declares it (`mem0/configs/vector_stores/pinecone.py`):

```python
extra_params: Optional[Dict[str, Any]] = Field(None, description="Additional parameters for Pinecone client")
```

`VectorStoreFactory.create` does `config.model_dump()` and then `vector_store_instance(**config)`, so it reaches `PineconeDB.__init__`, which forwards it (`mem0/vector_stores/pinecone.py`):

```python
if client:
    self.client = client
else:
    api_key = api_key or os.environ.get("PINECONE_API_KEY")
    ...
    params = extra_params or {}
    self.client = Pinecone(api_key=api_key, **params)
```

Two details the new row is deliberately worded around, rather than copying the TypeScript description:

- **Python spreads it into the `Pinecone` client constructor**, whereas TypeScript spreads `extraParams` into the `createIndex` call (`mem0-ts/src/oss/src/vector_stores/pinecone.ts`). Reusing the TypeScript wording would have documented the wrong call site.
- It only takes effect on the `else` branch, so **a caller-supplied `client` makes it a no-op**.

The row is placed between `batch_size` and `namespace` to match the field order in `PineconeConfig`, which the rest of the Python table already follows exactly.

## Type of Change

- [x] Documentation update

## AI Assistance

- [ ] No AI assistance

Agent used: Claude Code. It diffed every `mem0/configs/vector_stores/*.py` field list against the matching `docs/components/vectordbs/dbs/*.mdx` config table to find undocumented fields, and drafted the row.

What I checked myself, since a config row is only worth adding if the field actually does something:

- Instantiated `PineconeConfig(..., extra_params={"proxy_url": "..."})` and confirmed it validates and round-trips (the class rejects unknown keys via its `validate_extra_fields` validator, so this also confirms the name is exact).
- Confirmed the declared field order is `... batch_size, extra_params, namespace`, which is where I put the row.
- Confirmed `extra_params` is a real parameter of `PineconeDB.__init__` and that `VectorStoreFactory.create` passes config fields through by name, so nothing drops it in between.
- Read both call sites to confirm the Python/TypeScript asymmetry above, and the `client` short-circuit.
- Ran `python scripts/check-llms-txt-coverage.py`: `docs/llms.txt is in sync with docs/**/*.mdx.`

The diff is one table row. The reasoning behind its description, default, and position is written out above.

## Breaking Changes

N/A

## Test Coverage

One row in an `.mdx` file; no code paths touched. The behaviour it documents is in `PineconeDB.__init__`, and the verification I ran against the config class and the factory is listed above.

## Checklist


kartik-mem0 (approved):
Verified against the Python config, factory, and Pinecone constructor: `extra_params` defaults to `None`, is forwarded as client-constructor kwargs, and is unused when a client is supplied. The wording correctly differs from TypeScript's index-creation option. This is a useful, accurately scoped documentation addition.

Validation at head 95527c22fa4cacc843a466b048a8828a1ad31dab: docs/llms.txt coverage and diff whitespace checks passed. CLA is signed. No runtime code changes; no live Pinecone calls were needed. Thank you for the contribution.

PR #7228, opened 2026-09-04
fix(cli): read config json as utf-8

### LINKED ISSUE #7227: CLI config JSON should use explicit UTF-8 encoding
### Problem

The Python CLI reads and writes `~/.mem0/config.json` with the platform default text encoding in `mem0_cli.config`. On UTF-8 developer machines this is usually invisible, but on Windows or other non-UTF-8 locales non-ASCII values such as user emails/names/metadata copied into config can depend on the process locale.

### Why it matters

The config file is user-facing persistent state and is shared across CLI commands. Using an explicit encoding keeps config I/O deterministic across platforms.

### Proposed fix

Use `encoding="utf-8"` when loading and saving the CLI config JSON, and add a focused regression test around non-ASCII config values.

Ghraven:
## Problem

Closes #7227.

The Python CLI reads and writes `~/.mem0/config.json` using the platform default text encoding. That is usually invisible on UTF-8 machines, but can make config loading locale-dependent on Windows or other non-UTF-8 environments when config values contain non-ASCII text.

## Before / after

Before:
- `load_config()` opened `config.json` without an explicit encoding
- `save_config()` wrote `config.json` without an explicit encoding

After:
- config reads and writes use `encoding="utf-8"`
- a focused regression test covers loading raw UTF-8 JSON containing non-ASCII text

## Verification

From `cli/python`:

```console
$env:PYTHONPATH='src'; uv run pytest tests/test_config.py
```

Result: `25 passed`.

### DIFF
--- a/cli/python/src/mem0_cli/config.py
+++ b/cli/python/src/mem0_cli/config.py
@@ -90,7 +90,7 @@ def load_config() -> Mem0Config:
     config = Mem0Config()
 
     if CONFIG_FILE.exists():
-        with open(CONFIG_FILE) as f:
+        with open(CONFIG_FILE, encoding="utf-8") as f:
             data = json.load(f)
 
         config.version = data.get("version", CONFIG_VERSION)
@@ -174,7 +174,7 @@ def save_config(config: Mem0Config) -> None:
         },
     }
 
-    with open(CONFIG_FILE, "w") as f:
+    with open(CONFIG_FILE, "w", encoding="utf-8") as f:
         json.dump(data, f, indent=2)
 
     os.chmod(CONFIG_FILE, stat.S_IRUSR | stat.S_IWUSR)  # 0600
--- a/cli/python/tests/test_config.py
+++ b/cli/python/tests/test_config.py
@@ -59,6 +59,30 @@ class TestConfig:
         config = load_config()
         assert config.platform.api_key == ""
 
+    def test_load_config_reads_utf8_json(self, isolate_config):
+        import json
+
+        from mem0_cli.config import CONFIG_FILE, ensure_config_dir
+
+        ensure_config_dir()
+        CONFIG_FILE.write_text(
+            json.dumps(
+                {
+                    "version": 1,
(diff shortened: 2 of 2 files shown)


Ghraven:

kartik-mem0:
hey @Ghraven please sign the cla here only then i can review this pr in general here. 

if you signed the cla try to do an empty commit so that our ci triggers once again.

Ghraven:
Thanks for reopening this and for the Node CLI check request. I verified `cli/node/src/config.ts`: reads already explicitly use UTF-8, and writes preserve UTF-8 bytes. I could not reproduce the Python encoding issue in the Node CLI.

Pushed 5c82c43 to this same PR with two isolated Node regression tests: loading literal non-ASCII JSON and saving/reloading `José 東京 🧠`, including a raw UTF-8 byte assertion. They use a temporary home directory and restore environment overrides.

Verification: 13 config tests passed; TypeScript typecheck, Biome source lint, the new test's Biome check, and the CLI build passed. The configured pre-commit hooks have no applicable Python files for this test-only change. Local Windows checkout line endings initially tripped Biome; checking the source with LF resolved that without a source-code diff. This follow-up was prepared with AI assistance and verified locally.

I previously posted the CLA signature on September 4 in this PR. This substantive test commit also triggers fresh checks as requested; if the CLA remains pending after that, please let me know whether the CLA service needs a maintainer-side recheck. Thank you!

kartik-mem0:
@Ghraven i am still seeing the cla check pending can you force push to verify this here please.

without that i won't be able to review or approve the pr here 

thanks

Ghraven:
Thanks for checking. The CLA signature is already posted in this thread, and the normal push of `5c82c43` triggered fresh checks, but `license/cla` is still pending on that head.

Could you rerun the CLA check from the maintainer side, or confirm whether I need to complete a browser-based signature through CLA Assistant? I'd like to preserve the branch history rather than force-push just to retrigger the check. Thank you for helping get this unblocked.

PR #7230, opened 2026-09-05
fix(vector_stores/valkey): make search score metric-aware

ktz03:
## Linked Issue

Closes #7229

## Description

Sibling of #6546 / #6547 for the **Valkey** store.

Search scoring hardcoded `max(0.0, 1.0 - raw_distance)`, which is only correct for cosine. With L2 (already accepted by `create_col(distance=...)`), unbounded distances collapse almost every hit with `d > 1` to `0.0`.

This PR:

- Adds `distance` to `ValkeyConfig` / `ValkeyDB` (default `COSINE`)
- Uses `self.distance` in `_create_index` (was hardcoded `COSINE`)
- Persists `create_col(..., distance=...)` onto `self.distance` for scoring
- Adds `_distance_to_score` mirroring #6547:
  - COSINE / IP → `max(0, 1 - d)`
  - L2 → `1 / (1 + d)`

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A. Cosine (the default) is unchanged; only L2 scoring is corrected, and index init now respects the configured metric.

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [ ] I tested manually (describe below)
- [ ] No tests needed (explain why)

Added regression tests in `tests/vector_stores/test_valkey.py`:
- cosine keeps `1 - d`
- L2 uses the bounded map and does not collapse `d=4.0` to `0`
- `create_col(distance="L2")` updates scoring, not only `FT.CREATE`

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed

PR #7184, opened 2026-08-31
fix(gcp): read credential json as utf-8

Ghraven:
## Summary
- read GCP service account JSON files with explicit UTF-8 encoding
- cover both direct `credentials_path` and `GOOGLE_APPLICATION_CREDENTIALS` paths with non-ASCII project IDs

## Tests
- `python -m pytest tests/test_gcp_auth.py -q`
- `python -m py_compile mem0/utils/gcp_auth.py tests/test_gcp_auth.py`
- `git diff --check`

Closes #7209

PR #7236, opened 2026-09-05
fix(memory): run entity-index cleanup in processes that never wrote entities

### LINKED ISSUE #7226: Entity-index cleanup is skipped in any process that has not written entities, so deletions leave permanent stale links
`Memory._remove_memory_from_entity_store` returns without doing anything in any process that has not previously written an entity, so a deletion performed by a worker that only deletes leaves the entity index pointing at a memory that no longer exists.

`mem0@9a7924b`.

## The asymmetry

`mem0/memory/main.py`:

```python
# :558 — the write path goes through the property, which creates the store
@property
def entity_store(self):
    """Lazily initialize entity store on first use."""
    if self._entity_store is None:
        ...
        self._entity_store = VectorStoreFactory.create(...)
    return self._entity_store

# :665 — the delete path reads the private attribute, so the property never fires
def _remove_memory_from_entity_store(self, memory_id, filters):
    if self._entity_store is None:
        return
```

`:514` sets `_entity_store = None` on every new `Memory()`. The async variant does the same at `:2340-2343`.

## Reproduction

No network and no keys — the store is replaced with an in-memory double and only the control flow is exercised.

```python
def run(preload_entity_store):
    m = Memory.__new__(Memory)
    m._entity_store = None
    store = FakeStore(Row("e1", {"data": "Paris", "linked_memory_ids": ["mem-1"]}))
    if preload_entity_store:            # process that wrote entities earlier
        m._entity_store = store
    with patch.object(Memory, "entity_store", store):
        m._remove_memory_from_entity_store("mem-1", {"user_id": "alice"})
    return store.listed > 0, store.deleted
```

```
BUG      fresh Memory, delete only          -> consulted: False   deleted: []
CONTROL  same call, entities written first  -> consulted: True    deleted: ['e1']
```

Identical call, identical data. Whether cleanup happens is decided only by whether *this process* previously touched entities — nothing about the memory being deleted.

The method returns normally: no exception, no warning, no counter.

## Consequence

`:1800-1808` divides the entity boost by the number of linked memories:

```python
num_linked = max(len(linked_memory_ids), 1)
memory_count_weight = 1.0 / (1.0 + 0.001 * ((num_linked - 1) ** 2))
boost = similarity * ENTITY_BOOST_WEIGHT * memory_count_weight
```

Stale ids are never removed, so `num_linked` only grows:

| stale links | weight | vs correct |
|---|---|---|
| 0 | 1.000 | — |
| 10 | 0.909 | −9% |
| 100 | 0.090 | −91% |

A surviving memory is boosted an order of magnitude less than it should be, and search quality degrades quadratically as deletions accumulate. The entity record is never removed either: `:656` deletes it only when `linked_memory_ids` empties, which now never happens.

## When this bites

Any deployment where writes and deletions are handled by different processes — worker pools, scheduled cleanup, GDPR erasure scripts, a restarted server that deletes before it writes.

A single long-lived process that both adds and deletes is unaffected, which is also why the existing tests do not catch it: each one adds and deletes in the same process.

## Suggested fix

Read the property rather than the attribute at `:665` and `:2342`:

```python
if self.entity_store is None:
    return
```

The lazy initialization already does the right thing. If the intent is to avoid constructing the store when no entities have ever been written, that is a persisted question rather than a per-process one, and a stored flag would answer it correctly.

Java123456com:
Fixes #7226

## Problem

`Memory._remove_memory_from_entity_store` and its async variant early-returned when the private `_entity_store` attribute was still unset. The write path reaches the store through the lazy `entity_store` property (which creates it from config on first use), but the delete paths read the private attribute directly — so any process that has only ever deleted (worker pools, scheduled cleanup, GDPR erasure scripts, a restarted server deleting before its first write) silently skipped entity cleanup.

Consequences (from the issue): stale `linked_memory_ids` survive forever, the entity record is never deleted even when its link list should empty out, and the entity boost weight degrades quadratically as stale links accumulate (`1 / (1 + 0.001 * (num_linked - 1)^2)` — −91% boost at 100 stale links).

## Fix

Both delete paths now resolve the store through the same lazy `entity_store` property used by the write path (same pattern already used by `_compute_entity_boosts`). A store that cannot be initialized degrades to the method's existing warning-level contract (`Entity store cleanup failed ...`) instead of silently skipping, so the primary delete/update path is never broken by entity cleanup.

Docstrings updated accordingly (`No-op if the entity store has never been initialized in this process` is no longer true).

## Tests

`tests/memory/test_main.py::TestEntityStoreCleanupOnFreshProcess` (5 cases):

- sync + async cleanup runs on a fresh process (`_entity_store` unset) — both **fail on main, pass with this fix**;
- the already-initialized path is unchanged;
- sync + async initialization failures degrade to a warning without raising.

All 57 tests in `tests/memory/test_main.py` pass; `ruff check` clean on both changed files. (I did not run `ruff format`: the file carries pre-existing formatter drift on `main` and reformatting would drown this diff.)

## Scope note

`_bulk_clear_entity_store` (the `delete_all` path) has the same per-process guard, but its callers additionally gate on `self._entity_store is not None`, so fixing it properly also touches `delete_all` semantics. I left it out of this focused fix — happy to follow up if maintainers want it.

### DIFF
--- a/mem0/memory/main.py
+++ b/mem0/memory/main.py
@@ -657,16 +657,19 @@ class Memory(MemoryBase):
           - otherwise re-embed the entity text and update the payload
             (the vector store's update() requires a vector).
 
-        No-op if the entity store has never been initialized in this process.
+        The entity store is resolved through the lazy `entity_store` property,
+        matching the write path, so cleanup also runs in processes that have
+        not written entities themselves (worker pools, cleanup jobs); a store
+        that cannot be initialized degrades to a warning instead of skipping
+        the cleanup silently.
         Errors on individual entities are swallowed at debug level; outer
         failures are swallowed at warning level so the primary delete/update
         path is never broken by entity cleanup.
         """
-        if self._entity_store is None:
-            return
         search_filters = {k: v for k, v in filters.items() if k in ("user_id", "agent_id", "run_id") and v}
         try:
-            listed = self.entity_store.list(filters=search_filters, top_k=10000)
+            entity_store = self.entity_store
+            listed = entity_store.list(filters=search_filters, top_k=10000)
             rows = listed[0] if isinstance(listed, (list, tuple)) and listed and isinstance(listed[0], list) else listed
             for row in rows or []:
                 try:
@@ -677,7 +680,7 @@ class Memory(MemoryBase):
                     remaining = [mid for mid in linked if mid != memory_id]
                     if not remaining:
                         try:
-                            self.entity_store.delete(vector_id=row.id)
+                            entity_store.delete(vector_id=row.id)
                         except Exception as e:
                             logger.debug(f"Entity delete failed for id={row.id}: {e}")
                     else:
@@ -692,7 +695,7 @@ class Memory(MemoryBase):
                             continue
                         new_payload = {**payload, "linked_memory_ids": remaining}
                         try:
-                            self.entity_store.update(
+                            entity_store.update(
                                 vector_id=row.id,
                                 vector=vec,
                                 payload=new_payload,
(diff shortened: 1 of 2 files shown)


Java123456com:
Heads-up for maintainers: the `CI Gate` workflow fails **with zero jobs** on this PR, while the actual code pipelines (`PR Gate`, labeler, vouch) are all green. Looking at recent runs, every external fork PR from the last few days fails CI Gate the same way (#7230, #7228, #7219, #7184), while internal branches pass — so it looks like the reusable-workflow invocation (the `secrets: inherit` calls in ci-gate.yml) doesn't survive the fork-PR context rather than anything PR-specific. Since ci-gate.yml says branch protection requires exactly this check, it currently blocks merging for all external contributions. Nothing to change on our side — flagging in case it wasn't on the radar.

PR #7237, opened 2026-09-05
fix(vector_stores): metric-aware score conversion for Vertex AI Vector Search

### LINKED ISSUE #7232: bug(vector_stores/vertex_ai): search score hardcoded for cosine, breaks L2-style indexes
## Description

Same class of bug as #6546 / #6547 / #7229, in the **Vertex AI Vector Search** store.

`GoogleMatchingEngine` hardcodes cosine-style conversion in three places:

```python
score = max(0.0, 1.0 - raw_distance)  # _parse_output / search / get
```

Vertex Matching Engine indexes can be created with **DOT_PRODUCT** / **SQUARED_L2_DISTANCE** (not only cosine). When the underlying index uses L2-style distance, unbounded values make `1 - distance` collapse almost every hit with `d > 1` to `0.0`, destroying ranking and any score thresholds.

`create_col(..., distance=...)` is currently documented as ignored — there is no way for mem0 to know which metric the deployed index uses, so the store always assumes cosine.

## Expected Behavior

- Add an optional `distance_metric` on `GoogleMatchingEngineConfig` (default `cosine`) that the user sets to match their Vertex index.
- Convert distance → similarity in a metric-aware way (same approach as #6547 / #7229):
  - cosine: `max(0, 1 - d)`
  - euclidean / squared L2: `1 / (1 + d)`
  - (dot product: treat carefully; often already similarity-like depending on Vertex config)

## Additional Context

- Sibling of #6546 (S3) and #7229 (Valkey). Happy to open a PR mirroring #6547 once this has the `accepted` label.
- I am not proposing to change the Vertex index itself — only the client-side score mapping and a config knob so it can match the deployed index metric.

Java123456com:
Fixes #7232

## Problem

`GoogleMatchingEngine` hardcoded the cosine-style score conversion (`max(0.0, 1.0 - raw_distance)`) in `_parse_output`, `search`, and `get`. Vertex Matching Engine indexes can be deployed with `DOT_PRODUCT_POINT` or `SQUARED_L2_DISTANCE` in addition to `COSINE`; with an L2-style index, unbounded distances make `1 - distance` collapse nearly every hit to `0.0`, destroying ranking and any score thresholds.

## Fix

- **`GoogleMatchingEngineConfig.distance_metric`** (default `cosine`) tells mem0 which measure the deployed index uses — mem0 cannot infer it from the index (matching the issue's observation that `create_col(..., distance=...)` is documented as ignored). A `field_validator` canonicalizes the Vertex spellings (`squared_l2_distance`, `dot_product_point`, `euclidean`, `l2`, `ip`, `inner_product`) and rejects unknown values, so typos fail at config time instead of silently degrading to cosine.
- **`_distance_to_similarity`** converts per the contract documented on `VectorStoreBase.search`: cosine `max(0, 1-d)`, squared L2 `1/(1+d)` (bounded, monotone), dot product passed through as already similarity-like.
- All three call sites (`_parse_output`, `search`, `get`) now use the converter.

## Tests

`tests/vector_stores/test_vertex_ai_vector_search.py::TestDistanceMetric` (8 cases): default metric, alias canonicalization, invalid-metric rejection, config-to-store propagation, the three conversions, and a `_parse_output` regression case (cosine vs squared-L2 on the same input). 7 of the 8 fail on `main` and all 15 tests in the file pass with this change; `ruff check` clean.

Mirrors the `distance_metric` knob pattern of #6547 (S3 Vectors).

### DIFF
--- a/mem0/configs/vector_stores/vertex_ai_vector_search.py
+++ b/mem0/configs/vector_stores/vertex_ai_vector_search.py
@@ -1,6 +1,6 @@
 from typing import Dict, Optional
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, field_validator
 
 
 class GoogleMatchingEngineConfig(BaseModel):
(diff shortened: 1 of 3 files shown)


Java123456com:
Heads-up for maintainers: the `CI Gate` workflow fails **with zero jobs** on this PR, while the actual code pipelines (`PR Gate`, labeler, vouch) are all green. Looking at recent runs, every external fork PR from the last few days fails CI Gate the same way (#7230, #7228, #7219, #7184), while internal branches pass — so it looks like the reusable-workflow invocation (the `secrets: inherit` calls in ci-gate.yml) doesn't survive the fork-PR context rather than anything PR-specific. Since ci-gate.yml says branch protection requires exactly this check, it currently blocks merging for all external contributions. Nothing to change on our side — flagging in case it wasn't on the radar.

ktz03:
Thanks for the solid draft and for tagging me. I'm taking the handoff on #7232 as you offered — new PR incoming from `ktz03` that keeps your approach and adds `Co-authored-by`. No action needed on this PR; the gate close is expected until `accepted` lands on the issue.

PR #7247, opened 2026-09-07
fix(ts-sdk): restore project instructions if update assertion fails

### LINKED ISSUE #7044: Ensure project configuration is restored when integration tests fail
### Component

TypeScript SDK

## Problem

The project management integration test modifies `customInstructions` on the real project and restores the original value only at the end of the test.

If `updateProject()`, `getProject()`, or an assertion fails before the restoration step, Jest exits the test early and the original project configuration may not be restored.

This can leave test-generated configuration in the project and affect subsequent tests or users.

## Suggestion

Move the restoration logic to a guaranteed cleanup path such as `afterAll`, or use a `try/finally` block around the configuration update and verification.

For example, store the original value before modifying the project and always restore it in `finally`.

## Example

Current flow:

1. Read and store `originalInstructions`
2. Update `customInstructions` with a test value
3. Verify the update
4. Restore the original value

If step 2 or 3 fails, step 4 may never execute. A guaranteed cleanup mechanism would prevent persistent test side effects.                                                                                                

### How You Verified This

I reviewed the generated Connected Code Context for the integration tests.

In `mem0-ts/src/client/tests/integration/management.test.ts`, the test stores the original `customInstructions` value, updates the real project configuration, and restores the original value only at the end of the test.

Because the restoration is part of the normal test flow, any failure in an API call or assertion before the final restoration step can cause the test to exit before the original configuration is restored.

I verified this by tracing the execution order in the generated connected code context:

1. Store the original project instructions.
2. Update the project with test instructions.
3. Verify the updated configuration.
4. Restore the original instructions.

The cleanup is not placed in a guaranteed `afterAll` or `finally` block, so restoration is not guaranteed if an earlier step throws an error.

### AI Assistance

jrlprost:
## Linked Issue

Closes #7044

## Description

The integration test `updates project custom_instructions via updateProject()` in `mem0-ts/src/client/tests/integration/management.test.ts` stored the original `customInstructions` and restored it as the last statement of the test body. If `updateProject()`, `getProject()`, or the assertion in between threw, Jest aborted the test before reaching the restore call, leaving the real project's `customInstructions` set to the test value.

This PR wraps the update-and-verify steps in `try` and moves the restore call into a `finally` block, so the original value is always restored regardless of where the test fails.

## Type of Change

- [ ] Documentation update

## AI Assistance

- [ ] No AI assistance

Written by Claude Code (Anthropic). I reviewed the diff, confirmed it matches the `try/finally` fix suggested in the issue, and verified `npx prettier --check` passes on the changed file.

## Breaking Changes

N/A

## Test Coverage

This changes the control flow of an existing integration test (which requires live API credentials to run); no new test was added since the fix is about guaranteeing cleanup in the existing one.

## Checklist

Prepared with AI assistance; reviewed and verified by the author before submission.

### DIFF
--- a/mem0-ts/src/client/tests/integration/management.test.ts
+++ b/mem0-ts/src/client/tests/integration/management.test.ts
@@ -75,22 +75,24 @@ describeIntegration("MemoryClient Integration — Users & Project", () => {
     test("updates project custom_instructions via updateProject()", async () => {
       const testInstruction = `integration-test-${randomUUID().slice(0, 8)}`;
 
-      const result = await client.updateProject({
-        customInstructions: testInstruction,
-      });
+      try {
+        const result = await client.updateProject({
+          customInstructions: testInstruction,
+        });
 
-      expect(result).toBeDefined();
+        expect(result).toBeDefined();
 
-      // Verify the update took effect
-      const project = await client.getProject({
-        fields: ["custom_instructions"],
-      });
-      expect(project.customInstructions).toBe(testInstruction);
-
-      // Restore original
-      await client.updateProject({
-        customInstructions: originalInstructions || "",
-      });
+        // Verify the update took effect
+        const project = await client.getProject({
+          fields: ["custom_instructions"],
+        });
+        expect(project.customInstructions).toBe(testInstruction);
+      } finally {
+        // Restore original, even if the update or an assertion above threw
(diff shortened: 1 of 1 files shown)


jrlprost:
Ok, thank you for the Information :)

PR #7258, opened 2026-09-08
fix(vector_stores/faiss): reject empty payload in client-side filters

### LINKED ISSUE #7256: bug(vector_stores/faiss): empty payload bypasses client-side filters (returns True)
### Component

Vector Store

### Summary

`FAISS._apply_filters` short-circuits with `return True` when `payload` is falsy:

```python
if not filters or not payload:
    return True
```

So a stored empty payload `{}` (or any falsy payload that still reaches this helper) **passes every filter**, including `user_id` / `agent_id` / `run_id` scope checks. Scoped `search()` / `list()` can leak memories that have no identity metadata.

### Steps to Reproduce

```python

store = FAISS(collection_name="t", path="/tmp/faiss_empty_payload_demo")
# After insert with payload={} (or any empty dict in docstore):
assert store._apply_filters({}, {"user_id": "alice"}) is False  # expected
# actual today: True
```

### Expected Behavior

If filters are non-empty and payload is missing/empty, the row must **fail** the filter (return `False`), same as a missing key.

### Actual Behavior

`_apply_filters({}, {"user_id": "alice"})` returns `True`, so the vector is kept in filtered search/list results.

### Environment

- mem0 version: main @ dae67f7
- Python/Node version: 3.11+
- OS: any

### What I Ran

Read `mem0/vector_stores/faiss.py` `_apply_filters` and evaluated the empty-dict case against scoped filters.

### What I Saw

```python
if not filters or not payload:
    return True
```

Empty dict is falsy in Python, so the filter gate is skipped entirely.

### Why This Is a Bug

Scoped filters are a tenancy boundary. An empty payload must not satisfy `user_id=...`. The later `key not in payload: return False` branch never runs because of the early return.

### What I Ruled Out

Distinct from #6424 (operator/$or support) and #6998 (under-fetch). This is specifically the falsy-payload short-circuit.

### AI Assistance

ktz03:
## Summary
- `_apply_filters` no longer treats a falsy/empty payload as a filter pass.
- Non-empty filters + empty/`None` payload now return `False` (tenancy-safe).
- Adds regression unit test for #7256.

Closes #7256.

## Test plan
- [x] Unit coverage for empty/`None`/matching/mismatching payloads
- [ ] CI `tests/vector_stores/test_faiss.py`

### DIFF
--- a/mem0/vector_stores/faiss.py
+++ b/mem0/vector_stores/faiss.py
@@ -427,8 +427,13 @@ class FAISS(VectorStoreBase):
         Returns:
             bool: True if payload passes filters, False otherwise.
         """
-        if not filters or not payload:
+        if not filters:
             return True
+        # Empty / missing payload must fail scoped filters (tenancy boundary).
+        # Previously `if not filters or not payload: return True` let `{}` pass
+        # every user_id/agent_id/run_id check.
+        if not payload:
+            return False
 
         for key, value in filters.items():
             if key not in payload:
--- a/tests/vector_stores/test_faiss.py
+++ b/tests/vector_stores/test_faiss.py
@@ -285,6 +285,16 @@ def test_list(faiss_instance):
         assert result.payload["category"] == "A"
 
 
+def test_apply_filters_rejects_empty_payload(faiss_instance):
+    """Regression for #7256: empty payload must not satisfy scoped filters."""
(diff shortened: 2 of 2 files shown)


ktz03:
Rebased the empty-payload filter reject onto current `main` (`1c784e9`) so it can reopen cleanly once #7256 is labeled `accepted`. Behavior unchanged: empty/missing payload fails filters; regression tests kept.

PR #7264, opened 2026-09-08
Build the Mem0 API server image for linux/amd64

### LINKED ISSUE #4945: Linux/AMD64 (x64/x86) Docker Container Issue -
Regardless of how I attempt to use this image, it is not working.  Every time I try to run docker pull, docker compose, or even just run the container from Docker Desktop UI, it give me an error.   The same error.  

"No matching manifest for linux/amd64  in the manifest list entries."

When I look at the tags for the Mem0 images in Docker Hub, it shows one with OS/ARCH linux/arm64, which I would expect, and one with unknown/unknown.

FRFlo:
Closes #4945

## Summary

The published `mem0/mem0-api-server:latest` image currently exposes an ARM64-only manifest, which prevents deployment on standard x86_64/amd64 hosts.

This PR adds a GitHub Actions workflow that:

- builds `server/Dockerfile` for `linux/amd64` with Docker Buildx;
- builds on pull requests without pushing;
- publishes `mem0/mem0-api-server:latest` and an immutable SHA tag on pushes to `main`;
- uses the existing `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` repository secrets.

The Dockerfile also installs the required `libpq5` runtime library for psycopg and removes Uvicorn's development reloader from the production image.

## Validation

The workflow is intentionally scoped to `server/**` and can also be run manually with `workflow_dispatch`.

Docker builds were not run locally because Docker is unavailable in the development environment.

### DIFF
--- a/.github/workflows/server-docker.yml
+++ b/.github/workflows/server-docker.yml
@@ -1,0 +1,60 @@
+name: Build Mem0 API Docker image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "server/**"
+      - ".github/workflows/server-docker.yml"
+  pull_request:
+    paths:
+      - "server/**"
+      - ".github/workflows/server-docker.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  IMAGE_NAME: mem0/mem0-api-server
+
+jobs:
+  build:
+    name: Build linux/amd64 image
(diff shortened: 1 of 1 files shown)


FRFlo:
Follow-up commit `03fe9fe3` fixes the runtime failure observed on the amd64 build: it installs `libpq5` and removes Uvicorn\x27s development reloader from the production image.

PR #7271, opened 2026-09-09
feat(integrations): add @mem0/eve memory provider

Himanshu-Sangshetti:
## Summary

`@mem0/eve` — Eve memory provider at `integrations/eve`. Public API is `mem0Provider()` from `src/index.ts`.

## Code

**`src/provider.ts`** — `defineMemoryProvider` wiring.

- `recall["turn.started"]` and `recall["compaction.completed"]` share one handler. Query is `lastUserText(turn.input ?? messages)`.
- `capture["turn.completed"]` is omitted when `capture.enabled` is false.
- `tools()` returns `search` / `remember` / `forget` from `createMem0Tools`, locked to `context.memory.scope.key`.
- Store is lazy via `createLazyStore`. A rejected init is not cached.
- Recall throws (setup errors and Mem0 failures). Capture has no try/catch — Eve logs it.

**`src/store.ts`** — `mem0ai` adapter (`MemoryStore`).

- `add` / `search` pass `userId` / `filters.user_id` = scope key.
- `search` accepts `{ results }` or a bare array; anything else throws.
- `delete(id, userId)` calls `get` first; deletes only when `userId` matches. Otherwise `Memory not found`.
- `listByMetadata` is `getAll` with `{ user_id, metadata }` (used for capture replay).

**`src/recall.ts` / `src/capture.ts` / `src/tools.ts`**

- Recall: trim query, search, map hits to `{ id, content }`, return `null` if empty.
- Capture: skip if `metadata.operation_id` already exists for the scope, then `add` the current user turn + latest assistant (`infer` + metadata including `operation_id`).
- Tools: same `topK` / `threshold` / `rerank` / `infer` as the provider. Forget goes through scoped `delete`.

**`src/messages.ts` / `src/options.ts` / `src/idempotency.ts` / `src/errors.ts`**

- Message helpers: string / text-part / single `{ text }` content → Mem0 `{ role, content }`.
- Options: `apiKey` required unless `store` is injected; `host` must be `http:`/`https:`; `topK` 1–100; `threshold` 0–1.
- In-process `operationId` gate (cap 256) in front of the durable metadata check.
- `isSetupError` / `logProviderError` keep the original error object on logs.

**Tests** — `tests/*.test.ts` (53). Fake `MemoryStore` for provider/recall/capture; mocked `mem0ai` in `store.test.ts` for the real client mapping.

**Repo wiring**

- Docs: `docs/integrations/eve.mdx`, package README, `docs/llms.txt`
- CI: `eve-checks.yml` (typecheck / test / build), path filter + call job in `ci-gate.yml`
- Release: `eve-cd.yml`, `eve-v*` in `release.yml`
- `integrations/eve/registry/` — files for a later `vercel/eve` PR (`eve add memory/mem0` is not live yet)

## Package layout

```mermaid
flowchart TB
    index["index.ts\nmem0Provider, types"] --> provider["provider.ts"]
    provider --> options["options.ts\nresolveOptions"]
    provider --> recall["recall.ts"]
    provider --> capture["capture.ts"]
    provider --> tools["tools.ts\nsearch / remember / forget"]
    provider --> lazy["store.ts\ncreateLazyStore"]
    provider --> once["idempotency.ts\nin-process operationId"]
    provider --> messages["messages.ts\nEve messages → text"]
    provider --> errors["errors.ts"]

    recall --> store["store.ts\nMemoryStore"]
    capture --> store
    tools --> store
    once --> capture
    capture --> meta["listByMetadata\noperation_id"]
    lazy --> create["createMem0Store"]
    create --> client["mem0ai MemoryClient"]
    store --> client
```

## Test plan

- [x] `cd integrations/eve && pnpm test` (53)
- [x] `pnpm typecheck` and `pnpm build`
- [ ] CI `eve-checks` on this PR
- [ ] After merge: publish `@mem0/eve@0.1.0` (`eve-v0.1.0`), then Eve registry PR

## Review

@kartik-mem0

### DIFF
--- a/integrations/eve/src/capture.ts
+++ b/integrations/eve/src/capture.ts
@@ -10,6 +10,12 @@ export async function captureCompletedTurn(input: {
   messages: readonly ConversationMessage[];
   turnInput: readonly ConversationMessage[];
 }): Promise<boolean> {
+  // Best-effort deduplication, not a guarantee. This lookup + add is not atomic,
+  // and with infer:true the prior write can still be a PENDING event whose
+  // memory is not yet visible here. So a restart during that window, or two
+  // workers that both clear this check before either write, can capture the
+  // same operation twice. The in-process gate in provider.ts narrows the common
+  // case; eliminating it would need a backend-supported atomic idempotency key.
   const existing = await input.store.listByMetadata({
     userId: input.scopeKey,
     metadata: { operation_id: input.operationId },
--- a/integrations/eve/src/capture.ts
+++ b/integrations/eve/src/capture.ts
@@ -1,0 +1,26 @@
+import { completedTurnMessages, type ConversationMessage } from "./messages.js";
+import type { MemoryStore } from "./store.js";
+
+export async function captureCompletedTurn(input: {
+  store: MemoryStore;
+  scopeKey: string;
+  infer: boolean;
+  metadata: Readonly<Record<string, unknown>>;
+  messages: readonly ConversationMessage[];
+  turnInput: readonly ConversationMessage[];
+}): Promise<boolean> {
+  const messages = completedTurnMessages({
+    messages: input.messages,
+    turnInput: input.turnInput,
+  });
+  if (messages.length === 0) {
+    return false;
+  }
+
+  await input.store.add(messages, {
+    userId: input.scopeKey,
+    infer: input.infer,
+    metadata: input.metadata,
+  });
+  return true;
+}
--- a/integrations/eve/src/capture.ts
+++ b/integrations/eve/src/capture.ts
@@ -1,14 +1,23 @@
 import { completedTurnMessages, type ConversationMessage } from "./messages.js";
-import type { MemoryStore } from "./store.js";
+import type { MemoryStore, StoreMetadata } from "./store.js";
 
 export async function captureCompletedTurn(input: {
   store: MemoryStore;
   scopeKey: string;
+  operationId: string;
   infer: boolean;
-  metadata: Readonly<Record<string, unknown>>;
+  metadata: StoreMetadata;
   messages: readonly ConversationMessage[];
   turnInput: readonly ConversationMessage[];
 }): Promise<boolean> {
+  const existing = await input.store.listByMetadata({
+    userId: input.scopeKey,
+    metadata: { operation_id: input.operationId },
+  });
+  if (existing.length > 0) {
+    return false;
+  }
+
   const messages = completedTurnMessages({
     messages: input.messages,
     turnInput: input.turnInput,
--- a/integrations/eve/src/messages.ts
+++ b/integrations/eve/src/messages.ts
@@ -9,9 +9,16 @@ function textFromPart(part: unknown): string {
   if (typeof part === "string") {
     return part;
   }
-  if (part && typeof part === "object" && "text" in part) {
-    const text = (part as { text: unknown }).text;
-    return typeof text === "string" ? text : "";
+  if (part && typeof part === "object") {
+    const record = part as { type?: unknown; text?: unknown };
+    // Typed parts must be explicit text. A `reasoning` part also carries a
+    // `text` field, so accepting any object with `text` would leak the model's
+    // private reasoning into long-term memory. Untyped `{ text }` (no `type`
+    // key) stays supported for callers that pass a bare text object.
+    if ("type" in record && record.type !== "text") {
+      return "";
+    }
+    return typeof record.text === "string" ? record.text : "";
   }
   return "";
 }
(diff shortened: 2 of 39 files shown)


integrations/eve/src/capture.ts
```diff
@@ -0,0 +1,35 @@
...
+  metadata: StoreMetadata;
+  messages: readonly ConversationMessage[];
+  turnInput: readonly ConversationMessage[];
+}): Promise<boolean> {
+  const existing = await input.store.listByMetadata({
+    userId: input.scopeKey,
+    metadata: { operation_id: input.operationId },
+  });
+  if (existing.length > 0) {
+    return false;
```
kartik-mem0, line 24: Please account for pending writes when handling capture replay. With the default infer: true, Mem0 can accept an add request and return a PENDING event before any matching memory is visible. If Eve restarts during that window, this lookup returns an empty list and the same operation is submitted again. Two workers can also both pass the lookup before either writes. I reproduced two submissions using separate provider instances and a store that models pending extraction.

The process-local gate cannot cover that window. Please add pending-write and concurrent-replay tests, and either use a backend-supported atomic idempotency mechanism or document this as best-effort deduplication instead of guaranteeing that restart replays cannot write twice. A metadata lookup followed by add is not atomic.

integrations/eve/src/messages.ts
```diff
@@ -0,0 +1,82 @@
...
+  readonly content: unknown;
+}
+
+function textFromPart(part: unknown): string {
+  if (typeof part === "string") {
+    return part;
+  }
+  if (part && typeof part === "object" && "text" in part) {
+    const text = (part as { text: unknown }).text;
+    return typeof text === "string" ? text : "";
```
kartik-mem0: Please check the content part type before treating its text as conversation text. A reasoning part also has a text field, so an assistant message containing reasoning and a final answer currently sends both to Mem0 for long-term extraction. I reproduced this with [{ type: 'reasoning', text: 'speculation' }, { type: 'text', text: 'Noted.' }], which becomes 'speculation\nNoted.'.

For typed parts, accept type: 'text' and ignore reasoning and other non-text parts. Keep any intentionally supported untyped text format explicit. Please add a mixed reasoning/text regression test.

integrations/eve/src/messages.ts
```diff
@@ -0,0 +1,82 @@
...
+  );
+  if (users.length === 0) {
+    return [];
+  }
+
+  // Pair this turn's users with the latest assistant in the settled history.
+  const history = conversationMessages(input.messages);
+  const lastAssistant = [...history]
+    .reverse()
+    .find((message) => message.role === "assistant");
```
kartik-mem0: Please limit the selected assistant reply to the current turn. conversationMessages removes tool-only assistant messages, and this search then walks the entire remaining history. When the current turn has no assistant text, it can select an answer from a previous turn and attach it to the new user input. I reproduced this with an older question and answer followed by a new user message and a tool-only assistant response.

Use the current turn boundary when selecting assistant text. If there is no assistant text in this turn, capture only its user messages. Please cover that case in the completed-turn tests.

integrations/eve/src/tools.ts
```diff
@@ -0,0 +1,64 @@
...
+      inputSchema: z.object({
+        text: z.string().min(1).max(4000),
+      }),
+      async execute({ text }) {
+        await store.add([{ role: "user", content: text }], {
+          userId: scopeKey,
+          infer: options.infer,
+          metadata: { source: "eve-tool" },
+        });
+        return { saved: true };
```
kartik-mem0: Please distinguish an accepted write from a completed save. With infer: true, the Mem0 add API can return an event_id with status PENDING. The store adapter discards that response, so this tool returns saved: true even though extraction has not finished and may still fail. The model can then tell the user the fact was saved when only the request was queued.

The simplest correction is to preserve the API's event/status information and return an accurate queued result for asynchronous writes. If the tool promises a completed save instead, it needs to confirm successful processing first. Please test the pending response and update the documented tool behavior.

kartik-mem0 (commented):
Integration review: the provider connects Eve's turn lifecycle to Mem0 through automatic recall, completed-turn capture, and scope-bound search, remember, and forget tools. Scoping search at the API boundary and checking ownership before deletion are good choices. Configuration validation and the shared store adapter also keep the integration straightforward.

I found four issues in capture and write-status handling, detailed inline. Please address these before merging. In particular, the metadata lookup provides best-effort deduplication, not a guarantee against replay after a restart while extraction is pending.

Validation: all 53 existing tests passed, as did type checking and the package build. Live Mem0 search and metadata lookup succeeded for a unique, nonexistent review scope. Local reproductions demonstrated reasoning being included in capture, an older assistant reply being attached to a new turn, duplicate submissions across provider instances while writes remain pending, and the remember tool reporting saved for a queued write. Live writes, deletion, and a complete Eve agent conversation were not tested.

The fake store makes added memories immediately visible, so the existing restart test does not exercise Mem0's asynchronous extraction behavior.

integrations/eve/src/capture.ts
Himanshu-Sangshetti: Fixed as best-effort, honestly. Kept the metadata lookup + in-process gate (no backend atomic idempotency primitive exists), but removed the "restart replays cannot write twice" claim from the README/mdx and documented the pending window in `capture.ts`. Added regression tests: a pending-write replay and two concurrent replays both write (best-effort limitation), plus the deduped-once-visible case.

integrations/eve/src/messages.ts
Himanshu-Sangshetti: Fixed. `textFromPart` now rejects typed non-text parts: `{type:'reasoning', text}` is dropped, only `{type:'text'}` (and untyped `{text}`) pass. Your `[{reasoning},{text:'Noted.'}]` case now yields `Noted.`. Regression test added.

integrations/eve/src/messages.ts
Himanshu-Sangshetti: Fixed. `completedTurnMessages` now bounds the assistant reply to the current turn — assistant text after the last user message only. A tool-only turn captures just its user text (no stale prior answer). Also collects all assistant segments in the turn (text→tool→text) rather than one. Tool-only and cross-turn tests added.

integrations/eve/src/tools.ts
Himanshu-Sangshetti: Fixed. `store.add` now returns `AddResult{status,eventId}` via `parseAddResult`; the v3 add path returns `{status:'PENDING', event_id}` (verified against the backend), so `remember` returns `{status:'queued', eventId}` for async writes and `{status:'saved'}` only when resolved. FAILED/unknown status maps to completed (not queued), and the tool description tells the model not to claim saved on queued. Pending-response tests added.

PR #7279, opened 2026-09-10
feat(plugins): shared session handoff and lighter retrieval prompts

kartik-mem0:
## Linked Issue

[MEM-6256](https://linear.app/mem0/issue/MEM-6256) — reduce over-searching in plugin prompts.

Session handoff follows the [Memo guide](https://github.com/mem0ai/memo/blob/aeeb1593284d1d2fca3b4bcf1e32ea10f71df549/docs/session-handoff.md). This is an internal repository branch.

## Description

Adds explicit session handoff through shared local resources across all ten coding-agent plugins and lightens retrieval prompts. The source conversion follows Memo; the destination is the common `~/.mem0/handoffs/` resource store. Any supported plugin can list the current project's resources and resume a selected path as historical context. No Codex destination, destination CLI, model call, or Mem0 credentials are required.

Versions preserve each package's release series:

| Plugins | Patch version |
| --- | --- |
| Claude Code, Cursor, Codex, Kimi, Antigravity, portable | 0.3.1 → 0.3.2 |
| OpenCode, Pi, DeepSeek | 0.3.0 → 0.3.1 |
| OpenClaw | 1.1.0 → 1.1.1 |

The converter, validator, and resource store exist once in `agent-plugin-core/python/`. Packages ship the same small launcher and immutable runtime manifest. First explicit use downloads two source files from the pinned GitHub commit, verifies SHA-256 digests, and caches them for all plugins. Valid cached use works offline; no transcript is uploaded and memory startup does not download the runtime. Builders and conformance checks detect drift and duplicated engines. The runtime source commit `59939c003f6b3eb8add709e5897f7bfbe3e9f4d8` is retained by the non-release tag `handoff-runtime-59939c003f6b3eb8add709e5897f7bfbe3e9f4d8`, so a squash merge and PR branch deletion do not orphan the source. First use still requires GitHub access; verified cached use works offline.

Claude saves its active session before model invocation. OpenCode, Pi, OpenClaw, and DeepSeek use native active context or trusted transcript paths. Other Python skills accept an explicitly selected completed transcript or common bundle. The resource preserves supported active conversation, readable compaction, title, project, images, and completed tool history without summarizing or truncating it. Unsupported or unfinished context fails explicitly. Files are private and uniquely named. Resume returns history without replaying tool calls; OpenClaw's slash command supplies it to that session's next prompt, while its native tool reads it immediately. Storage is local to the machine.

Retrieval now suggests a focused query when prior work may help, skips searching when current context suffices, and searches again only for a specific remaining gap. Scoping, top-k behavior, and capture scheduling remain unchanged. Each plugin's changelog points to the shared behavior explanation; no standalone session-end/handoff documentation page is added. The SDK & Tools changelog records these same patch versions and the removal of Sidekick from Cursor, Kimi, and Antigravity. Sidekick remains available only in Claude Code, as on current main.

## Review follow-up

Merged current `main` into this branch, preserving its non-Claude Sidekick removal. Python MCP resume now prepends the same historical-data guard as TypeScript, with a regression check for parity. Python 3.10 compatibility is verified; documentation and tool errors now match the existing supported minimum.

Focused native-parser review found and fixed stale Kimi continuation reminders surviving repeated compaction and Pi/OpenClaw session titles reverting after switching branches. Six regression cases failed before the fixes. Added focused OpenClaw tests for invalid transcript headers, missing native context, resource failures, list behavior, and optional search guidance; the handoff adapter now has 100% statement and branch coverage.

The runtime-retention tag addresses the reviewer’s durability concern without publishing a package release or changing the shared-runtime architecture. Package versions remain as listed above.

## Type of Change

- [x] Documentation update

## AI Assistance

- [ ] No AI assistance

Implemented and reviewed with Codex agents in isolated worktrees. The agent ran the checks below; the author's personal review attestation remains unchecked.

- [ ] **I can explain every line of this diff and how it interacts with the rest of the codebase, without asking an AI tool.**

## Breaking Changes

N/A

## Test Coverage

- 25/25 offline conformance checks: all six Python bundles, shared runtime tests, all four TypeScript host suites/typechecks/builds, and package artifact verification.
- 1,028 passing tests: 353 Python runtime/build, 9 conformance, and 666 TypeScript (shared 31, OpenCode 38, Pi 98, OpenClaw 446, DeepSeek 53).
- Real standalone runtime smoke across all ten packages: the first downloaded the pinned source from GitHub into an isolated cache and saved a private resource; every package listed and resumed that resource with downloads disabled, preserving the full 720KB tool outcome and literal tool arguments.
- Shared TS transport also round-tripped a resource through the actual Python engine, preserving title, project, text, and tool results.
- Actual Pi Node 20.20.2: 97 passed, one native-SDK-only test skipped; compiled memory-plugin startup passed. Node 22.19.0: all 98 passed. List/resume work on Node 20; save requires the native SDK's Node 22.19+.
- Python 3.10.19: 86 current handoff tests pass; the compatibility audit also passed 278 runtime tests before the parser additions.
- Runtime/cache tests cover missing and tampered files, download/digest failures, atomic resource publication, project scoping, relocation, and malformed/incomplete content.
- Ruff and isort pre-commit hooks, documentation-index coverage, and `git diff --check` pass.

Live testing found and fixed three defects: Claude's quoted preprocessor command did not match its skill permission rule; Codex0.153.4 emits harness/usage metadata alongside history; DeepSeek's native Cordis runtime rejected direct access to an optional attachment service. Regression checks cover each fix.

- **Claude Code2.1.267:** a real synthetic session used a Read tool, then `/mem0:handoff` saved its conversation and exact completed tool output. Retested the corrected skill without a CLI permission override and with spaces in the plugin path.
- **Claude → Codex0.153.4:** a fresh Codex run invoked the packaged `handoff_resource` MCP tool and correctly recovered the codename, storage decision, and next-step restriction from Claude's resource. Expected facts were withheld from the receiving prompt.
- **Codex source:** generated a real synthetic Codex session with a tool call; the final packaged reader saved it successfully after the metadata fix.
- **DeepSeek native runtime:** actual service registration, save, list, and resume passed through the native session/tool registry. Live model continuation remains untested because its model credential is unavailable.
- **Remaining live checks:** Cursor CLI is not signed in; Codex → Claude model continuation was blocked by automatic approval review for sending the saved resource to Claude's service. These are not reported as passing.

No live Mem0 Platform checks were needed for handoff, which operates independently of the Mem0 API. Handoff requires Python 3.10+ and GitHub access when the verified runtime cache is missing.

## Checklist

- [ ] I have performed a self-review of my code


karthik-indla:
**One design decision worth an explicit sign-off before merge**

The handoff engine is **downloaded from `raw.githubusercontent.com` at a pinned commit and executed** on first use. The *integrity* side of this is implemented carefully and I verified it end to end — HTTPS + an immutable 40-hex SHA pin + SHA-256 verification of the bytes **before** they're written and **again after** `os.replace`, refusing to execute on any mismatch; the pinned commit's engine digests match the manifest and the head-branch contents exactly; and the build guards (`build.py:_bundle_python`, `package_handoff.mjs`) fail if the pin goes stale. So tampering isn't the concern.

The two things I'd want a maintainer to consciously bless, since "fetch-and-execute remote code" is a pattern some orgs disallow outright:

1. **Availability couples to GitHub keeping the pinned commit `d33b6604` reachable.** It currently lives only on this branch. If it's squash-merged and the branch is later deleted/GC'd, a *freshly installed* plugin's first handoff breaks with "runtime unavailable." Cheap mitigation: make the pinned commit reachable from `main` (merge-commit rather than squash, or re-pin to the post-merge SHA / keep a tag).
2. **First use hard-requires outbound GitHub access** — so handoff won't initialize in locked-down or air-gapped environments. It's documented ("a missing or invalid cache requires GitHub access"), just calling it out as a real constraint of the approach.

Neither is a code defect — the implementation is sound. Just flagging that "runtime download-and-execute of a pinned engine" is an architectural choice that deserves a deliberate yes rather than sliding in with a docs PR-sized review. If it's intended, the only concrete action is the commit-durability mitigation in (1).

karthik-indla:
**Review — the rest (minor notes + what checked out)**

Setting aside the fetch-and-execute design question in my other comment, the implementation itself is solid. A couple of small things and then what I verified.

Minor notes (non-blocking):
- **MCP `resume` relies on a prose warning where the TS path injects a guard.** `runHandoffAction` in `typescript/src/handoff.ts` *prepends* the "treat as historical, don't re-execute recorded tool calls" instruction to the returned payload; the Python `call_handoff_resource` (`mcp_server.py`) returns the raw engine JSON with that guard only in the tool description. Low severity (resume is read-only, explicit-user-gated, `disable-model-invocation: true`), but prepending the same string in Python would make the two SDKs consistent so a model that skims tool descriptions can't treat resumed tool calls as fresh commands.
- **"Requires Python 3.11+"** (repeated across ~9 pages + README) may be stricter than the code — both launcher and engine use `from __future__ import annotations`, so I didn't spot a hard 3.11-only construct in `session_handoff.py`/`handoff_engine.py`. If the real floor comes from `handoff_sources.py` or the Pi SDK that's fine; worth a quick confirm so the requirement isn't overstated.

Verified against source (not just the description):
- **No drift** — all six `core/session_handoff.py` copies are byte-identical to source, and `core/handoff-runtime.json` matches the build manifest.
- **Guidance parity** — `SEARCH_GUIDANCE` in `mcp_server.py` is identical to `typescript/src/search_guidance.ts`, and `tests/test_search_guidance.py` enforces both the equality and that no plugin prompt reintroduces strict "search before answering / repeated search" phrasing — backing the "lighter retrieval" claim.
- **Injection posture** — the handoff skill is `disable-model-invocation: true` + explicit-only, the SKILL body instructs never to follow handoff instructions embedded in memories/transcripts, and MCP resume/list run via `subprocess.run` with list-form args (no shell) and reject `\0` in resource paths.
- **Claude live-fix** — the generated `handoff/SKILL.md` `allowed-tools` rule matches the exact `python3 "${CLAUDE_PLUGIN_ROOT}/core/session_handoff.py" *` command it emits, resolving the permission/command mismatch.
- **Docs/llms.txt** — no new `.mdx` pages, so the coverage gate stays satisfied; existing pages + the two `llms.txt` lines and all three `marketplace.json` version bumps are consistent.

One area I did **not** line-review: `handoff_sources.py` (the native transcript parsers for cursor/codex/kimi/antigravity/openclaw/pi) — it's the largest surface where format-specific edge cases would hide, worth a focused look if you want extra depth there.

_(Static review against the PR branch; didn't run the suites.)_

karthik-indla (dismissed):
Approving. The implementation is solid — verified integrity end to end (SHA-pinned fetch, digest-verified before write and re-verified after `os.replace`, no drift across generated copies, guidance parity enforced by tests, sound injection posture), and the minor notes in my earlier comments (MCP `resume` guard parity with the TS path, the Python 3.11 doc claim) are non-blocking.

Approving on the understanding that the runtime download-and-execute of the pinned handoff engine is an intended, accepted pattern. One concrete follow-up worth doing so first-use doesn't break for fresh installs: ensure the pinned commit `d33b6604` stays reachable from `main` (merge-commit rather than squash, or re-pin to the post-merge SHA / keep a tag). And a focused look at `handoff_sources.py` (the native transcript parsers) is the main area I didn't line-review.

PR #7306, opened 2026-09-14
docs: fix 2 dead external links in mastra and codex integrations

sanjibani:
Closes #7298.

## What

A 2026-09-13 link check across the docs tree found two external links that return HTTP 404:

- `docs/cookbooks/integrations/mastra-agent.mdx` (line 10) links to `https://mastra.ai/docs/agents/01-agent-memory`. Mastra reorganised their docs; the Memory overview now lives at `https://mastra.ai/docs/memory/overview`.
- `docs/integrations/codex.mdx` (line 6) links to `https://openai.com/index/codex/`. The Codex product page is at `https://openai.com/codex/`.

Both are straight text substitutions; no surrounding prose touched.

## Out of scope

The issue also flagged that scripted checks against most `openai.com` paths return 403 rather than the actual HTTP status. That's a bot-block behaviour on OpenAI's side, not something to fix in this repo. Excluded from the diff.

## Checklist

- [x] Two link fixes, no surrounding text changes
- [x] No em-dashes anywhere in the diff
- [x] Branch pushed to `sanjibani:fix/issue-7298-dead-doc-links`

PR #7308, opened 2026-09-14
fix(oss/pgvector): attach error listener + reconnect on server-side termination

### LINKED ISSUE #7294: [mem0-ts] PGVector store has no 'error' listener on its pg Client — server-side connection termination crashes the host process, and the client never reconnects
### Summary

`PGVector` (mem0-ts, open-source mode) keeps a single `pg.Client` and never attaches an `error` listener. When PostgreSQL terminates that connection server-side — e.g. a container restart or `pg_terminate_backend`, both of which send `FATAL 57P01 terminating connection due to administrator command` — the idle client emits an `'error'` event with no listener. Node then throws it as an uncaught exception, which kills the host process. The same client is never reconnected afterwards.

### Impact

Any host that embeds `mem0ai/oss` in a long-lived process is taken down by an ordinary database restart. In our case the OpenClaw Gateway (a Node daemon using `@mem0/openclaw-mem0` with `vectorStore.provider = "pgvector"`) exits with code 1 every time the mem0 Postgres container is recreated. It happened three times in one afternoon during routine container maintenance, each time killing every unrelated in-flight session on the host.

Because `initialize()` memoizes `this._initPromise`, the store is also left permanently broken after the connection drops, until the host process is restarted.

### Environment

- `mem0ai` 3.0.7 (also verified present in **3.1.8**, the current latest)
- Node.js v26.8.1, `pg` 8.11.3
- PostgreSQL 16 + pgvector (Docker), `mode: "open-source"`
- `vectorStore: { provider: "pgvector", config: { host, port, user, password, dbname } }`

### Steps to reproduce

Two connections: terminate the first one's backend from the second.

```js
import pg from "pg";
const opts = { host: "localhost", port: 5432, user: "postgres", password: "***", database: "postgres" };

const victim = new pg.Client(opts);        // mimics PGVector's client: NO 'error' listener
await victim.connect();
const { rows } = await victim.query("select pg_backend_pid() as pid");

const killer = new pg.Client(opts);
await killer.connect();
await killer.query("select pg_terminate_backend($1)", [rows[0].pid]);  // same FATAL 57P01
await killer.end();
```

A `docker stop` / `docker compose up -d --force-recreate` on the DB server produces the identical result, since PostgreSQL also sends 57P01 to connected clients on shutdown.

### Actual result

```
node:events:505
    throw er; // Unhandled 'error' event
error: terminating connection due to administrator command
    severity: 'FATAL', code: '57P01',
    file: 'postgres.c', line: '3316', routine: 'ProcessInterrupts'
Emitted 'error' event on Client instance at:
    at Client._handleErrorEvent (pg/lib/client.js:341:10)
```

Process exit code 1.

Adding `client.on("error", handler)` makes the process survive (verified), but the store is still unusable afterwards — see below.

### Root cause

1. `pgvector.ts` constructor creates `new Client({...})` and calls `this.initialize().catch(console.error)`. There is **no `client.on("error", ...)`** anywhere in the class. An idle client whose socket is closed by the server emits `'error'`; with no listener, Node throws it (EventEmitter semantics). Wrapping mem0 API calls in `try/catch` does not help — this escapes asynchronously.
2. `initialize()` is memoized via `this._initPromise`, so once the connection dies the store never reconnects; every later operation fails with `Connection terminated unexpectedly`.

For comparison, the Redis store in the same package does attach `client.on("error", ...)`, so the omission looks specific to the pgvector backend.

### Expected

- A dropped/idle-terminated client connection must never become a process-level uncaught exception.
- The store should reconnect on next use (reset the memoized init promise / health-check the client), with retry + exponential backoff.
- Transient connection-disturbance codes (`57P01`, `08006`, `08003`, `ECONNRESET`, `EPIPE`) should be treated as retryable, consistent with the other backends in this package.
- Using a `pg.Pool` (with its own `pool.on("error", ...)`) instead of a single `Client` would also give reconnection for free.

### Real-world impact evidence

Host process (OpenClaw Gateway, Node v26.8.1, Linux x64) logged, and its crash bundle recorded:

```
[openclaw] OpenClaw hit an unexpected runtime error.
[openclaw] Reason: terminating connection due to administrator command
systemd: openclaw-gateway.service: Main process exited, code=exited, status=1/FAILURE
```

```json
{
  "reason": "uncaught_exception",
  "error": {
    "name": "error",
    "code": "57P01",
    "message": "terminating connection due to administrator command"
  },
  "process": { "platform": "linux", "arch": "x64", "node": "26.8.1" }
}
```

Three occurrences on the same day, all code `57P01`, triggered only by recreating the pgvector container. No OOM involved (peak RSS ~1.1 GB, memory pressure count 0).

### Notes

Verified still present in `mem0ai@3.1.8` (`dist/oss/index.mjs`, PGVector constructor: `new Client(...)` + `this.initialize().catch(console.error)`, no error listener, no reconnect). Happy to open a PR if you'd like a patch.

sanjibani:
Closes #7294.

## What

`PGVector` keeps a single `pg.Client` and never attaches an `error` listener. When PostgreSQL terminates that connection server-side (container restart, `pg_terminate_backend`, both sending `FATAL 57P01`), the idle client emits an `'error'` event with no listener. EventEmitter semantics throw it as an uncaught exception, killing the host process. The same client is never reconnected afterwards because `initialize()` memoises its init promise.

## Fix

Two changes, both in `mem0-ts/src/oss/src/vector_stores/pgvector.ts`:

1. **`attachClientErrorHandler()`** — registers a process-level guard on every `pg.Client` the store creates. The host stays alive; the store is left unusable until the next operation triggers a reconnect.
2. **`resetInit()`** — drops the memoised `_initPromise`, removes listeners from the broken client, and rebuilds a fresh `pg.Client` with the handler reattached, so the next `initialize()` performs a real handshake instead of returning a never-resolving promise.

The Redis store in the same package already wires `client.on("error", ...)`; the omission in the pgvector backend is the bug.

## Tests

Two new regression cases in `pgvector.unit.test.ts`:

- **`attaches an error handler to every pg.Client so a server-side termination does not crash the host process`** — asserts the client has an `error` listener and that emitting the event does not throw.
- **`reconnects on the next call after the underlying pg.Client errors out`** — emits a `57P01` error, calls `initialize()` again, asserts a brand-new `pg.Client` is used and that it also has an error listener installed.

The mock `Client` now extends `EventEmitter` so the test can drive the new path. Full `npx jest --config jest.config.js src/oss/tests/` stays green: **1022/1022 tests, 67/67 suites** pass.

## Out of scope

The reporter also suggested switching to `pg.Pool` for free reconnection. That is a larger refactor (every query path uses `this.client.query` directly today), and a single `Client` matches the existing per-process semantics. Pool can be a follow-up; the guard + reconnect is the minimum that closes the uncaught-exception path described in the issue.

## Checklist

- [x] Two regression tests added; both pass
- [x] No em-dashes in the diff
- [x] Existing 1020 tests still pass (no regressions)
- [x] Branch pushed to `sanjibani:fix/issue-7294-pgvector-error-listener`


sanjibani:
Force-pushed to drop the docs commit (e5cc545) from this branch. The docs commit lives in its own PR #7306; this one now contains only the pgvector error-listener + reconnect fix plus its two regression tests. The cached file list above may still show 4 files; the actual compare against main is 2 files.

PR #7312, opened 2026-09-14
docs: state the real openai_base_url resolution order in BaseEmbedderConfig

### LINKED ISSUE #7313: Docs: BaseEmbedderConfig says openai_base_url defaults to api.openai.com, but two env vars come first
### Page

Not a page on docs.mem0.ai — it is the docstring that `BaseEmbedderConfig` carries in the source, `mem0/configs/embeddings/base.py:59`. Filing it here because the `pr-gate` bot asked for an issue before the fix, and because a wrong docstring is read far more often than a docs page when someone is reading their own traceback.

### What's Wrong or Missing

The docstring states a default that the code does not use:

```
:param openai_base_url: Openai base URL to be use, defaults to "https://api.openai.com/v1"
```

The signature default is `None` (`base.py:22`), and `mem0/embeddings/openai.py:22-27` puts two environment variables between the parameter and that URL:

```python
base_url = (
    self.config.openai_base_url
    or os.getenv("OPENAI_API_BASE")
    or os.getenv("OPENAI_BASE_URL")
    or "https://api.openai.com/v1"
)
```

### Version

`mem0ai` 2.0.20, Python 3.12, Windows.

### Runnable reproduction

```python
import os
os.environ["OPENAI_API_KEY"] = "sk-not-used"
os.environ["OPENAI_BASE_URL"] = "http://127.0.0.1:9999/v1"

from mem0.embeddings.openai import OpenAIEmbedding

cfg = BaseEmbedderConfig()          # openai_base_url left at its default
print("config.openai_base_url :", cfg.openai_base_url)
print("docstring says default :", "https://api.openai.com/v1")
print("client actually uses   :", OpenAIEmbedding(cfg).client.base_url)
```

Real output:

```
config.openai_base_url : None
docstring says default : https://api.openai.com/v1
client actually uses   : http://127.0.0.1:9999/v1/
```

No traceback: nothing raises, and nothing is broken. The code is correct; the sentence describing it is not. That is the whole bug, and I would rather say so than dress it up as a crash.

### Why it is worth fixing

It misleads in both directions.

**Silent redirect.** Anyone with `OPENAI_BASE_URL` exported for an unrelated tool — a gateway, a proxy, a local server — has their mem0 embedding calls going somewhere else, and this docstring actively tells them it cannot happen. The reproduction above is that situation with the destination made obvious.

**Redirect looks unavailable.** The reverse: someone who *wants* embeddings pointed at a local OpenAI-compatible server reads "defaults to `https://api.openai.com/v1`" as "pinned", and starts threading `openai_base_url` through their own config layer when exporting `OPENAI_BASE_URL` would already have worked.

There is a third, quieter cost. `openai.py:28-33` raises a `DeprecationWarning` saying `OPENAI_API_BASE` goes away in 0.1.80 and to use `OPENAI_BASE_URL`. A reader of this docstring has no way to know either name is in play, so that warning arrives without context.

### Suggested Fix

State the real order. This is what PR #7312 already does — one `:param` line, no signature change, no behaviour change, no literal touched:

```
:param openai_base_url: Openai base URL to be use. When None, the embedder falls back to
    the OPENAI_API_BASE environment variable (deprecated), then OPENAI_BASE_URL, then
    "https://api.openai.com/v1"
```

PR #7312 was auto-closed by the gate because it touches a `.py` path even though the diff is prose only. If this issue gets the `accepted` label the PR reopens by itself; if you would rather I redo it differently, say so and I will.

### How I found it

Grep sweep across installed packages for docstrings claiming a default base URL, cross-checked against the resolution code in the same package — then confirmed by running the snippet above against 2.0.20. I read `openai.py` and `base.py` in full before writing this, and the output pasted above is from my terminal, not reconstructed.

**Disclosure:** I work on a tool that records agent runs at the model-provider boundary, which is why base-URL resolution order is something I end up reading closely across frameworks. I am not proposing mem0 adopt anything — this is a one-line docstring correction and nothing else.

xizhuomengcontin:
Closes #7313

## What

One `:param` line in `BaseEmbedderConfig`.

## Why

The docstring says `openai_base_url` "defaults to `https://api.openai.com/v1`". The signature default is `None`, and two environment variables are consulted before that URL is reached.

`mem0/embeddings/openai.py:22-27`

```python
base_url = (
    self.config.openai_base_url
    or os.getenv("OPENAI_API_BASE")
    or os.getenv("OPENAI_BASE_URL")
    or "https://api.openai.com/v1"
)
```

So the real order is `openai_base_url` → `OPENAI_API_BASE` → `OPENAI_BASE_URL` → `https://api.openai.com/v1`. The documented sentence is true only for the last hop.

## Why the gap matters in practice

Two ways it costs someone time, in opposite directions:

**Traffic goes somewhere the docs did not say it could.** A developer with `OPENAI_BASE_URL` exported for an unrelated tool reads this docstring, concludes mem0 pins `api.openai.com` unless `openai_base_url` is passed, and does not realise the ambient variable silently redirects their embedding calls. That is a live footgun in any shell that has a gateway or proxy configured globally.

**The redirect they want looks unavailable.** The reverse case: someone wanting mem0's embeddings to go through a local OpenAI-compatible server reads "defaults to `https://api.openai.com/v1`" as "pinned", and starts plumbing `openai_base_url` through their own config layer — when exporting `OPENAI_BASE_URL` would already have worked.

There is a third, quieter cost: `mem0/embeddings/openai.py:28-33` raises a `DeprecationWarning` telling the user `OPENAI_API_BASE` will be removed in 0.1.80 and to use `OPENAI_BASE_URL` instead. A reader of this config docstring has no way to know either name is in play, so the warning arrives without context.

## Scope

Docstring only; no signature, no behaviour, no literal changed. The precedence stated in the new text is read directly off `openai.py`, including the deprecation note, so the two files now agree.


xizhuomengcontin:
The gate is right by its own rule and I am not asking you to bend it — but this is the documentation-only case it says it skips, so flagging it rather than opening a duplicate issue.

The PR changes one file, `mem0/configs/embeddings/base.py`, and within it one `:param` line of the `BaseEmbedderConfig` docstring. No signature, no default value, no behaviour. My guess is the check classifies by path, so any `.py` file counts as code even when the diff is prose.

The substance, since it is short. The docstring says `openai_base_url` "defaults to `https://api.openai.com/v1`". The signature default is `None`, and `mem0/embeddings/openai.py:22-27` resolves it as:

```python
base_url = (
    self.config.openai_base_url
    or os.getenv("OPENAI_API_BASE")
    or os.getenv("OPENAI_BASE_URL")
    or "https://api.openai.com/v1"
)
```

So two environment variables sit between the parameter and the documented default. That cuts both ways: someone with `OPENAI_BASE_URL` exported for an unrelated tool does not learn from this docstring that their embedding calls are being redirected, and someone who wants that redirect reads "defaults to …" as "pinned" and goes looking for config plumbing they do not need. `openai.py:28-33` also raises a `DeprecationWarning` about `OPENAI_API_BASE` going away in 0.1.80, which arrives without context if neither name is documented.

I have no reproduction or traceback to offer because nothing misbehaves — the code is correct and the sentence describing it is not. If you would still rather have an issue, say so and I will open one with this content and link it with `Closes #`; I did not want to file an issue that only says "a docstring is wrong" if the docs exemption already covers it.

Also noted on the vouch check — nothing needed from me there.

xizhuomengcontin:
Followed the gate's own steps rather than waiting on the docs exemption.

Opened #7313 with the version, a runnable reproduction and the real output, and added `Closes #7313` to the description above. The `reopen` job then ran and said exactly what it should:

> `#7312 is not a closed pull request linking an accepted issue. Skipped.`

So this is now one `accepted` label on #7313 away from reopening itself. That call is yours — I am not asking you to skip it.

One thing worth flagging honestly about that issue: there is no traceback in it, because nothing raises. `OpenAIEmbedding` resolves `openai_base_url → OPENAI_API_BASE → OPENAI_BASE_URL → https://api.openai.com/v1`, and the `BaseEmbedderConfig` docstring names only the last hop. The reproduction shows a client quietly pointed at `http://127.0.0.1:9999/v1/` while the documented default says `api.openai.com`. If "the code is right and the comment is wrong" does not clear your bar for `accepted`, that is a reasonable position — say so and I will close both rather than leave them open.

CLA is signed and `license/cla` is green now. The remaining `Vercel` red is the fork-authorization prompt, not a build failure.

PR #7321, opened 2026-09-14
fix(plugins): stamp surface identity at record time, not send time

whysosaket:
## What

Six defects in 0.3.x plugin telemetry, plus surface attribution across the SDKs, CLIs and integrations. Paired with [mem0ai/platform#3602](https://github.com/mem0ai/platform/pull/3602), which recognizes the source values this PR sends.

> **Draft deliberately.** Per `AGENTS.md`, drafts are exempt from the PR Gate. This needs a linked issue carrying `accepted` before it leaves draft.

## Defects 1, 2 and 6 are one problem

Not three bugs — one spool protocol getting three properties wrong.

**Identity was decided by the wrong process.** `harness` was stamped in `record()`, correctly. `source` was read from a module global in `flush()`, so whichever process drained the spool named every event in it. Two processes never call `init()`: `mcp_server.py`, and the detached `python3 telemetry.py` sender that `spawn_flush()` starts. `record()` now stamps `source` beside `harness`, and the build generates `core/_harness_id.py` per host so identity resolves with **no `init()` call at all**.

That last part is the durable fix. The build already computed `HARNESS_ID` and spent it only on skill templating, and `bundle_drift` already diffs `core/` byte-for-byte — so one generated file makes every entrypoint, present and future, correct by default, and CI catches drift for free. I did **not** add `MEM0_PLUGIN_HARNESS` to the six manifests: they sit outside the `--sync` and `--check` boundary, which is the property that caused this.

It also unifies two defaults that disagreed — `configure_harness` derived `<host>_plugin` while `telemetry.init` derived `MEM0_<HOST>_PLUGIN`, so one plugin could emit three different source values. Dead until `source` moved into `record()`; fixed in the same commit.

**Ownership was inferred, not held.** `Path.replace` is `os.rename`, which preserves mtime, so a claim made after a quiet minute inherited the spool's age and was stealable the instant it existed. Claims are now touched on creation, and the per-batch rewrite doubles as a lease heartbeat.

**Progress was not durable.** `flush()` returned on the first failed batch without truncating, so the retry re-posted from index 0 — 150 events delivered 250 times. It now rewrites the claim with the unsent remainder after every batch, bounding a crash to one repeated batch. Each event also carries a `uuid`.

**Parked batches starved.** They were only reachable when no spool existed, and because sessions keep recording there usually was one — so a batch parked by a failed send waited until the 7-day expiry deleted it unsent, *despite its own presence being what starts the sender*. `flush()` drains them in the same run; expiry applies only after a genuine retry has failed.

## The other three

**`code.install` counted upgrades and repeat sessions.** `is_first_run()` read the identity file, which only a successful flush writes — so an offline user recorded an install **every session, forever**, not just on a race. A dedicated `install-state.json` is claimed atomically (`O_CREAT|O_EXCL`) at record time. Upgrade detection keys on a non-empty data directory rather than guessing at leftover `venv/` files.

**The docs called this anonymous.** Every event carries the account email, and `repo_hash`/`session_hash` were unsalted SHA-256 — over a git remote URL, or `local:<absolute path>` which normally contains the username, making 16 hex chars enumerable. READMEs, the module docstring and a new docs section now say what the code does, and digests are salted per install. That keeps every within-account join analytics use and gives up only cross-machine repo joins, which nothing computes.

**A cached email outlived an API key change.** Re-resolved when the key fingerprint differs. `$identify` aliases anonymous→email only — **aliasing one account to another merges person profiles irreversibly**, so the re-resolve path never emits one.

## Why all six shipped green

The shared core's only tests lived under `claude-code-plugin/tests/`, behind a `conftest.py` calling `configure_harness()` and `telemetry.init()` at import. Core behaviour was never exercised uninitialised. That is not a contributing factor to defect 1 — it is the whole cause, and none of the fixes prevent the next one.

Adds `agent-plugin-core/tests/` with no init in its conftest, including real subprocess tests and coverage for the **portable** plugin, which has no flush worker and no hook runner and would pass a native-only test vacuously.

## Also

Surface headers on the Python SDK, TS SDK, both CLIs, vercel-ai-sdk and pi-agent; `source` hoisted from `metadata` to the top level where the backend actually reads it. Corrects `deepseek-plugin/README.md`, which told readers `ZAPIER`/`STRANDS` were already in the platform allowlist — neither was.

## Verification

| Suite | Result |
|---|---|
| `agent-plugin-core/tests` (new) | 59 passed |
| `claude-code-plugin` | 203 passed, 8 skipped |
| `cursor` / `codex` / `kimi` / `antigravity` | 11 / 5 / 2 / 6 passed |
| `ruff` + `compileall` | clean |
| `build.py --check`, all six hosts | clean |

Three pre-existing tests asserted the broken behaviour and were rewritten, each with a comment saying what it now asserts and why.

**Not verified:** the TypeScript changes are not typechecked locally — deps are not installed for those packages. CI covers them.


whysosaket:
Closing in favour of a reviewable stack. 42 files across six defects plus client-side header work is not a unit anyone can review properly, and none of it is urgent enough to justify asking.

Split into five, stacked so each PR's diff shows only its own change:

| # | PR | What | Files |
|---|---|---|---|
| 1 | #7322 | Privacy claims and salting | 12 |
| 2 | #7323 | Wrong plugin reported (defect 1) | 16 |
| 3 | #7324 | Duplicate and lost events (defects 2, 6) | 10 |
| 4 | #7325 | Install counting, cached email (defects 3, 5) | 16 |
| 5 | #7326 | Client headers across SDKs, CLIs, integrations | 17 |

Ordering is deliberate. #7322 is first because it is the only defect with exposure beyond data quality and is almost entirely prose. #7323 is foundational — the generated per-host identity module it adds is what #7326 reads. #7324 is the only change that can lose data, so it lands after the test suite that can catch it.

Defects 2 and 6 stayed together in #7324 on purpose: they are not two bugs but one spool protocol getting three properties wrong, and fixing them separately leaves the protocol wrong.

The stack tip is byte-identical to the tree tested here, verified with `git diff`. Same test results: 59 core, 203/11/5/2/6 across hosts, `--check` clean for all six.

whysosaket:
Superseded by #7322 → #7326.

PR #7322, opened 2026-09-14
fix(plugins): say what telemetry actually sends, and salt the hashes

whysosaket:
**1 of 5.** Splitting #7321, which was too large to review as one unit. This one is first because it is the only defect with exposure beyond data quality, and it is almost entirely prose.

| # | PR | What |
|---|---|---|
| **1** | **this** | **privacy claims + salting** |
| 2 | `pr2/source-at-record-time` | wrong plugin reported |
| 3 | `pr3/spool-delivery` | duplicate + lost events |
| 4 | `pr4/install-marker-and-identity` | install counting, cached email |
| 5 | `pr5/surface-headers` | client headers across SDKs/CLIs |

## The mismatch

The README promises "anonymous usage events". The module docstring says it sends only "salted hashes". Neither is true.

`resolve_distinct_id()` exchanges the API key for the account email and sends **that** as the `distinct_id` on every event. Installing the plugin requires an API key, so this is nearly every user. That is probably the behaviour we want — the Python SDK and the CLI attribute the same way — but the description has to match it.

## The hashes were not a privacy control

`repo_hash` and `session_hash` were unsalted SHA-256 cut to 16 hex characters. `repo.identity` is a git remote URL, or `local:<absolute path>` when there is no remote — which normally contains the account username. Sixteen unsalted hex characters over that input space is enumerable, so for a public repo, and for most local paths, this was closer to plaintext than to a hash.

Salted per install, salt kept in the identity file. That preserves every within-account join the analytics actually use and gives up only cross-machine joins on the same repository, which nothing currently computes. Since the `distinct_id` is already the email, the hash was never buying privacy *from us* — only from whoever obtains the data later, which is exactly what the salt fixes.

## Also

`deepseek-plugin`'s README and source comment told readers `ZAPIER` and `STRANDS` were already in the backend's `KNOWN_EVENT_SOURCES` allowlist, so adding `DEEPSEEK_HARNESS` would be "a one-line platform change matching the existing" ones. Neither was in the allowlist. Corrected here; [mem0ai/platform#3602](https://github.com/mem0ai/platform/pull/3602) actually adds all three.

`docs/integrations/claude-code.mdx` had no Telemetry section at all. Added.

The same claim was standing everywhere else, so the sweep covers it: the pause skill in all six hosts told users a "minimal anonymous telemetry ping" fires while paused, which is the same ping under the same email; `docs/integrations/deepseek-plugin.mdx` said "anonymous usage events"; and `integrations/zapier-mem0/README.md` advertised telemetry that app does not have, with an opt-out variable controlling nothing. The data directory listing had also gone stale against the two files this stack adds.

## OpenClaw

`docs/integrations/openclaw.mdx` is corrected here too, and it is worth calling out because the first pass deliberately left it alone. The reasoning was that hashing the email is materially different from sending it. Reading `integrations/openclaw/telemetry.ts` does not support that: `distinctId()` is an unsalted SHA-256 of the account email, and we hold the emails it is derived from, so recovering the account is a table join. `resolveEmail()` also rewrites already-queued events onto that id, and `identifyAnonymous()` fires a PostHog `$identify` that merges the prior random id into it permanently. That is pseudonymous, not anonymous. The opt-out is unchanged and still accurate.

## Verification

263 passed, 8 skipped across the shared core and the five host suites. `ruff` clean. Bundle `--check` clean for all six hosts. CI green.

Worth a look from whoever owns the privacy docs. This changes a published claim on four surfaces.


integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -83,9 +89,81 @@ def is_enabled() -> bool:
...
+        # Cannot persist. Stable-per-machine beats random-per-call.
+        _salt_cache = hashlib.sha256(str(path).encode("utf-8")).hexdigest()
+        return _salt_cache
+
+    try:
+        _salt_cache = path.read_text(encoding="utf-8").strip()
+    except OSError:
+        _salt_cache = ""
+    if not _salt_cache:
+        _salt_cache = hashlib.sha256(str(path).encode("utf-8")).hexdigest()
```
kartik-mem0: [P2] Publish the completed salt atomically and omit hashes on failure; a concurrent hook can read the newly created empty file and permanently cache this predictable path digest.

kartik-mem0 (dismissed):
AI-assisted review: 7/10. Hold until concurrent salt creation cannot fall back to a predictable path digest.

karthik-indla (commented):
Independent review — **confirming Kartik's `:148` finding (still present at head) and extending it.**

The concurrent empty-salt race is real: `_install_salt` opens with `O_CREAT|O_EXCL` (empty file now visible) and only flushes the uuid on `close`, so a second short-lived hook — the normal case per your own docstring — hits `FileExistsError`, `read_text()` returns `""`, and line 148 caches `sha256(str(path))`. `path` is a fully predictable location (`~/.mem0/<harness>/telemetry-salt`), so `repo_hash`/`session_hash` become enumerable — which breaks the docs' "cannot be linked back to a repository name or path" claim. The same predictable fallback also fires on the non-race paths: the `except OSError` at :140 (read-only dir) and a swallowed write failure at :134-135.

**Two things beyond Kartik's note:**
- The atomic-write fix (temp + fsync + `os.rename`) is necessary but **not sufficient on its own**: `_scoped_digest` (:164) only returns `""` when `value` is empty, so if `_install_salt()` returns `""` it still emits `sha256(":" + value)` — unsalted and predictable. "Omit on failure" has to be enforced in `_scoped_digest` too (return `""` when the salt is empty).
- Test gap: `test_salt_is_stable_across_processes` spawns its processes sequentially, so it can't catch this — a regression test needs a concurrent reader against a mid-write/empty file.

#7323 (stacked under this) is clean — source is stamped at record-time, write-once, no bucket collapse.

kartik-mem0:
Please also apply per-install salting to TypeScript OpenCode's [`project_hash`](https://github.com/mem0ai/mem0/blob/3fd4949040af66ddf17faedd960cefbc7c88d107/integrations/opencode-plugin/telemetry.ts#L43-L44) and add a regression test: it still uses enumerable, unsalted SHA-256 of the project ID. This is a preexisting gap outside the Python fix, which now passes review.

karthik-indla (approved):

PR #7323, opened 2026-09-14
fix(plugins): report the plugin that produced the event, not the one that sent it

whysosaket:
**2 of 5**, stacked on #7322 so the diff shown is only this change. Splitting #7321.

## What happens

`harness` is set when an event is **recorded**. `source` was set when its batch is **sent**. Both come from module globals that stay at `generic` / `MEM0_PLUGIN` until `telemetry.init()` runs, and two processes never run it:

- `python3 telemetry.py` — the detached sender `spawn_flush()` starts at session start, after every skill command, and when the MCP server exits. Everything it delivered was labelled `source=MEM0_PLUGIN`. Only batches `flush_worker.py` *happened* to drain got the real host.
- `mcp_server.py` — records every manual search as `harness=generic`.

All six Python plugins ship the same files, so `source` could not tell any of them apart and MCP searches from every plugin landed in one bucket.

**The portable plugin is worse than that.** `_build_portable` skips `flush_worker.py` entirely, so its only sender is the uninitialised one — 100% of its events were mislabelled, not "whichever sender won".

## The fix

Two parts.

`record()` stamps `source` beside `harness`. `flush()` already spreads per-event properties last, so a per-event source wins over any sender default — the sending process stops mattering at all.

The build generates `core/_harness_id.py` per host, seeding both modules at import, so identity no longer depends on an entrypoint remembering to call `init()`. This is the part that makes the failure impossible rather than remembered: `mcp_server.py` is correct with no change to it, and so is any entrypoint added later.

That cost one line in the build, because `_bundle_python` already computed `values["HARNESS_ID"]` and spent it only on skill templating, and `bundle_drift` already diffs `core/` byte-for-byte — so `--check` catches drift for free.

**I did not add `MEM0_PLUGIN_HARNESS` to the six manifests.** They sit outside the `--sync` and `--check` boundary, which is the property that caused this in the first place.

## A latent third value

`configure_harness` defaults `source_tag` to `<host>_plugin`; `telemetry.init` defaulted it to `MEM0_<HOST>_PLUGIN`. So a third value existed, reachable from `hook_runner.py`'s `--harness` branch. It was harmless only because `hook_runner` never calls `flush()` — and moving `source` into `record()` would have made it live. Unified in the same commit.

Events also now carry a `uuid`, which #7324 uses.

## Why the suite stayed green

`telemetry.py` lives in `agent-plugin-core/python/` but its only tests lived under `claude-code-plugin/tests/`, behind a `conftest.py` calling `configure_harness()` and `telemetry.init()` at import. Core behaviour was **never** exercised uninitialised. That is not a contributing factor — it is the whole cause.

New `agent-plugin-core/tests/` with no init in its conftest, running real subprocesses, and covering the portable plugin, which would pass a native-only test vacuously.

## Breaking for saved PostHog queries

Worth knowing before this lands, because the point of the fix is that the emitted value changes. Three different values were in flight for the same plugin, depending on which process sent the batch:

| Old value | Emitted by | After |
|---|---|---|
| `CLAUDE_CODE_PLUGIN` | `memory_cli`, and `flush_worker` via the env tag | unchanged, now receives all of it |
| `MEM0_CLAUDE_CODE_PLUGIN` | `hook_runner`, which called `init()` without a source tag | no longer emitted |
| `MEM0_PLUGIN` | `mcp_server` and the detached sender, neither of which called `init()` | no longer emitted |

The canonical value is not new, so an insight already filtering on `<HOST>_PLUGIN` keeps working and gets more data. Anything filtering on `MEM0_PLUGIN` or `MEM0_<HOST>_PLUGIN` goes to zero for new events and needs repointing. Historical data is untouched.

Note that `MEM0_PLUGIN` also arrives as a platform-side `Event.source` value in this same effort. Different field, different system, no conflict, but easy to confuse when reading a dashboard.

## Verification

10 new subprocess tests. 273 passed, 8 skipped across the shared core and the five host suites, up from 263 at the base of this PR. Bundle `--check` clean for all six hosts. CI green.


kartik-mem0 (commented):
AI-assisted review: 8/10. No blocking finding in this diff; recording source on each event and generating host defaults fixes attribution, subject to the preceding PR being corrected.

karthik-indla (commented):
Independent review — **no blocking finding, agrees with the 8/10.** `source` is stamped in the recording process at `record()`, written into the spool line, and `flush()` spreads `**properties` after the top-level `source` fallback so the record-time value wins over whichever process drains — so `mcp_server.py` and the detached sender now emit the real host instead of collapsing to generic. Per-event `uuid` for PostHog dedup is sound.

Two nits (non-blocking):
- `telemetry.py:43-44` imports `PLATFORM_SOURCE`/`PLATFORM_APPLICATION` (as `_PLATFORM_*`) but nothing reads them except the `except` re-assignment — dead imports; wire them in or drop them.
- The `init()` docstring / `test_source_tag_defaults_agree_between_the_two_modules` claim the fallback matches `configure_harness`, but that emits lowercase `<host>_plugin` while telemetry emits uppercase `<HOST>_PLUGIN`; the test `.upper()`s to paper over it. Different sinks so it's harmless, but the comment/test overstate the agreement.

karthik-indla (dismissed):

kartik-mem0 (approved):

PR #7324, opened 2026-09-14
fix(plugins): stop delivering telemetry events twice, and stop losing parked ones

whysosaket:
**3 of 5**, stacked on #7323. Splitting #7321.

Two reported defects, one cause: the spool protocol **infers** ownership instead of holding it, and never records progress. Worth reviewing as one change for that reason.

## Duplicate delivery after a partial failure

`flush()` posts the claim in batches of 100 and returns on the first failure, keeping the whole file. The retry then posts every batch again, including the ones that already arrived. Reproduced: **150 recorded events delivered 250 times.**

Progress is now written back to the claim after each successful batch, so a retry resumes where the send stopped and a crash repeats at most one batch.

## Duplicate delivery when two senders overlap

`spool.replace(claim)` is `os.rename`, which preserves mtime. So a claim created after a quiet minute inherited the **spool's** last-write time and looked abandoned the instant it existed. A second sender starting while the first was still posting took it over and sent it too — most likely at session end, when the MCP server's exit sender and the SessionEnd flush worker both drain.

Claims are touched at claim time, and the per-batch rewrite doubles as a lease heartbeat.

`_post` makes one attempt with `SEND_TIMEOUT` and no retry, so a heartbeat lands well inside the 120s lease. There is a test asserting that margin, specifically so that adding a retry loop to `_post` later cannot silently let a slow sender lose its claim mid-flight.

## Parked batches starved

`_claim_spool` only looked at parked `.sending` files when **no spool existed** — and because sessions keep recording, there usually was one. So a batch parked by a failed send waited until the 7-day expiry deleted it unsent.

The sharp version: the parked file is what causes `spawn_flush()` to start a sender at all, and then that sender ignores it.

`flush()` now drains the live spool and then parked claims in the same run, oldest first, bounded. Expiry applies only after a genuine retry has failed, with the attempt count carried in the filename.

## Two things that came out of building it

A sender that **gives up** releases its lease rather than heartbeating on the way out. Heartbeating would make an abandoned batch look actively owned for another full stale window, delaying the retry for no reason.

A failing send stops the run. Otherwise one broken connection burns every parked batch's attempt budget in a single flush — which is how I first drove a batch to its expiry limit in three seconds during testing.

## On the dedupe key

Each event carries a `uuid` (added in #7323) and it is passed through in the batch payload. **I could not verify which key PostHog actually deduplicates on** — `$insert_id` is the documented idempotency property and a bare `uuid` field is a different mechanism. Worth confirming before relying on it. The per-batch progress above is the fix that does not depend on PostHog behaviour at all.

## Verification

6 new delivery tests covering partial failure, claim theft, parked drain, expiry, crash injection and heartbeat margin. 284 passed, 8 skipped across the shared core and the five host suites, up from 273 at the base of this PR. CI green.

Two existing tests asserted the old lifecycle and are updated in place, each with a comment saying what changed.


integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -347,38 +352,171 @@ def spawn_flush() -> bool:
...
         try:
             age = now - orphan.stat().st_mtime
         except OSError:
             continue
-        if age > CLAIM_EXPIRY_SECONDS:
+        # Attempts, not age. Every re-claim touches the mtime and every release
+        # backdates it by a fixed amount, so age is pinned near the stale
+        # threshold and never reaches the expiry. Age stays only as a backstop
+        # for files that never carried an attempt marker.
+        if _claim_attempt(orphan) >= MAX_CLAIM_ATTEMPTS or age > CLAIM_EXPIRY_SECONDS:
```
kartik-mem0, line 527: [P2] Check the live lease before deleting exhausted claims; a second sender can unlink a freshly claimed a3 file before its owner reads it, permanently losing the events.

kartik-mem0 (changes requested):
AI-assisted review: 6/10. Hold until an active final retry cannot be deleted by another sender.

karthik-indla (commented):
Independent review — **Kartik's `:441` is confirmed still-present, and I'd raise it from P2 to blocker: it is no-crash permanent event loss in the exact mechanism this PR advertises.**

`_claim_parked` evaluates the exhaustion/expiry unlink (:441-442) *before* the live-lease guard (:447). `_claim_parked` re-claims an `a2` file by renaming to `…-a3.sending` and `_touch`ing a fresh mtime — but for any `a3` claim, `3 >= MAX_CLAIM_ATTEMPTS` is True, so :441 unlinks it **without ever consulting the fresh mtime** that :447 would honor. Interleaving, no crash required:
1. Sender A re-claims `a2`→`a3` (fresh mtime), returns the claim, has **not yet** read it.
2. Sender B (a second detached sender — the design's normal case) globs, sees the `a3` file, `3>=3` → **unlinks it**.
3. A's `_drain` → `read_text` → `FileNotFoundError` → returns `(0, True)` = "fully delivered."

Nothing was posted, the batch is gone, `flush()` reports success. The one-line reorder Kartik suggested (hand off any claim with `age < CLAIM_STALE_SECONDS` before the exhaustion unlink) closes both this and the crash variant.

**Two things to fix alongside it:**
- `_drain` returns `(0, True)` ("everything delivered") on both the vanished-file read error and the empty-parse path (:596-612, :621-633). That's what turns the `:441` race into silent success instead of a retryable failure, and it independently mislabels any transient/torn cross-process read as delivery (`.corrupt` quarantine is never re-globbed). The read-failure branch should return `(0, False)`.
- Retry budget is 3 attempts with effectively **no time backoff**: `_release_claim` backdates mtime to immediately-reclaimable, so concurrent senders can burn 0→3 within seconds on a momentary blip and drop the batch by design (:441). Leave a cooldown so 3 attempts span real time.

Dedup (atomic rename claim), the `_touch`-after-rename fix for the stale-looking-fresh-claim double-delivery, and HTTP-failure re-parking all look correct. **Note the whole `telemetry.py` is copy-pasted into 7 plugin dirs — every fix here lands ×7**; worth considering whether the shared core can be the single source. No test covers the concurrent live-owner deletion — a multi-sender regression test would pin it.

karthik-indla (dismissed):

kartik-mem0 (approved):

PR #7325, opened 2026-09-14
fix(plugins): count installs once, and re-resolve the email when the key changes

whysosaket:
**4 of 5**, stacked on #7324. Splitting #7321. Two small independent fixes to per-machine state in the same file.

## `code.install` counted upgrades and repeat sessions

Session start records install whenever `is_first_run()` is true, and that only checked whether `telemetry-identity.json` exists. Recording install does not create that file — **only the first successful flush does.**

So install fired:

- for every 0.2.x user on their first 0.3.x session (0.2.x never wrote the file, and the data directory survives the upgrade);
- again for any session starting before that first flush finished;
- and — the part that makes this unbounded rather than a race — **on every single session, forever, for anyone whose flush never succeeds.** An offline or firewalled user reported a new install every time they opened an editor, which is exactly the population hardest to see in the data.

A dedicated `install-state.json` is claimed with `O_CREAT|O_EXCL` at the moment install is recorded, so two sessions starting together cannot both win.

Deliberately **not** the identity file. Writing that from a recording process would race the sender, which writes it during `resolve_distinct_id` — and overloading that file is what caused this bug in the first place.

Upgrade detection keys on the data directory already having content: a fresh install has an empty one, anything else predates this session. That is a firmer predicate than looking for 0.2.x's `venv/` and `requirements.txt`, which is a guess about files another part of the plugin may or may not have written, and only ever works for this one upgrade. The version is stored so later changes record `code.upgrade` with a real `from_version`.

## A cached email outlived an API key change

`resolve_distinct_id` kept the first email it resolved and never looked again, so switching to a key from another account kept attributing events to the previous one. It now fingerprints the key the email came from and re-resolves when the current key differs, and falls back to the anonymous id when no key is configured rather than continuing to attribute to an account it cannot verify.

### The dangerous part

`resolve_distinct_id`'s second return value becomes a PostHog `$identify` with `$anon_distinct_id`. **Aliasing one account email to another merges two real person profiles and cannot be undone.**

So the re-resolve path returns no alias. Aliasing runs anonymous → email only, never email → email. That is the single most destructive thing in this whole series if implemented casually, and it has a named test.

## Verification

291 passed, 8 skipped across the shared core and the five host suites, up from 284 at the base of this PR. CI green.

One existing test asserted that `is_first_run` flips when the identity file is written — which is the defect itself. Rewritten, plus coverage for atomic claiming, upgrade detection and version changes.

## Reading data already sent

Treat `code.install` counts since Sep 1 as "first 0.3.x session on a machine", not as new users.


integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -546,21 +678,55 @@ def _post(payload: dict[str, Any], url: str) -> bool:
...
+            # uncached /v1/ping/ on every flush, forever, and a firewalled one
+            # pays the full timeout each time.
+            identity["key_fingerprint"] = fingerprint
+            _write_identity(identity)
+            return email, ""
+
     if not key:
+        # No key to verify the account with; do not keep attributing to it.
+        if email:
+            identity.pop("email", None)
```
kartik-mem0, line 840: [P2] Rotate the anonymous ID and preserve legacy identified state on logout; clearing only the email lets the next account reuse an ID already associated with the previous account.

integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -252,9 +252,141 @@ def anonymous_id(identity: dict[str, str] | None = None) -> str:
...
+        return None
+
+    state["plugin_version"] = memory_core.PLUGIN_VERSION
+    state["upgraded_at"] = memory_core.utc_now()
+    try:
+        temporary = path.with_suffix(f".{os.getpid()}.tmp")
+        temporary.write_text(json.dumps(state), encoding="utf-8")
+        temporary.replace(path)
+    except OSError:
+        return None
```
kartik-mem0, line 473: [P2] Release failed upgrade claims and recover abandoned ones; if this replacement fails, the existing upgraded-version sentinel permanently suppresses subsequent healthy retries.

kartik-mem0 (changes requested):
AI-assisted review: 6/10. Hold for legacy identity migration and recoverable upgrade claims.

integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -546,21 +678,55 @@ def _post(payload: dict[str, Any], url: str) -> bool:
...
+    if email and fingerprint:
+        recorded = identity.get("key_fingerprint", "")
+        if recorded == fingerprint:
+            return email, ""
+        if not recorded:
+            # Rows written before fingerprints existed. Adopt the current key
+            # rather than re-resolving: otherwise every existing user pays an
+            # uncached /v1/ping/ on every flush, forever, and a firewalled one
+            # pays the full timeout each time.
+            identity["key_fingerprint"] = fingerprint
```
kartik-mem0, line 833: [P2] Verify a legacy cached email before binding the current key fingerprint; otherwise a key changed before upgrading becomes permanently associated with the previous account email.

integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -546,21 +678,55 @@ def _post(payload: dict[str, Any], url: str) -> bool:
...
+            _write_identity(identity)
         return anonymous_id(identity), ""
-    email = _resolve_email(key)
-    if not email:
+
+    resolved = _resolve_email(key)
+    if not resolved:
+        # The key changed and will not resolve (revoked, offline, API down).
+        # Do not keep attributing to the previous account.
         return anonymous_id(identity), ""
```
kartik-mem0, line 857: [P2] Reset account-bound anonymous identity when credentials change; a failed lookup currently reuses an ID already associated with the previous account, so events remain attributed to that account.

karthik-indla (commented):
Independent review — **all four of Kartik's `:709/:718/:702/:388` items are confirmed still-present at head.** Three of them (709/718/702) are genuine cross-account identity-attribution leaks, which is the exact risk this PR set out to fix, so the CHANGES_REQUESTED stands. Adding detail + one correction:

- **`:709` logout + `:718` credential-change-failed-lookup share one root cause**: any exit from a verified-email state falls back to a possibly-already-aliased `anonymous_id` instead of rotating it. After a login `$identify` merges `A1`→person `E1` in PostHog, logout keeps `A1` (post-logout "anonymous" events resolve to E1), and a key-swap to E2 whose `_resolve_email` fails returns `A1` again (E2's events attributed to E1, repeating every flush until a resolve succeeds). A single guard — "if `aliased` or an email was ever set, mint a fresh anonymous_id before falling back" — fixes both.
- **`:702`** binds the current key's fingerprint to a stale legacy email without verifying it resolves. One correction to the defending comment: the "otherwise every user pays an uncached `/v1/ping/` on every flush, forever" cost is overstated — the fingerprint cache this branch already writes suppresses subsequent pings, so a one-time verify costs exactly **one** ping per legacy user total, not one per flush. The correctness/cost tradeoff cited doesn't actually hold.
- **`:388`** the exclusive `upgraded-<version>` sentinel is created before the marker rewrite; if `temporary.replace()` raises, the sentinel is never released and every future session hits `FileExistsError` → upgrade events permanently suppressed on that machine. Analytics-only, but "permanently" is accurate — unlink the sentinel in the `except`, or make it age-recoverable like `_claim_spool`.

Install-once idempotency is **sound** (O_EXCL claim + correct empty-dir snapshot ordering). One nit: `claim_install` writes JSON directly into the O_EXCL'd path rather than temp+replace, so a hard kill mid-write leaves an empty marker and loses that one install count (version tracking recovers via `_repair_install_state`). Salt/hashing interaction with #7322 is fine — email is sent raw as `distinct_id`; a key change correctly re-resolves and re-hashes.

integrations/agent-plugin-core/python/telemetry.py
kartik-mem0: [P2] Failed legacy verification still returns the cached email; Mem0 /ping can fail while PostHog succeeds, so use a fresh anonymous identity until the active key is verified.

integrations/agent-plugin-core/python/telemetry.py
kartik-mem0: [P2] OSError cleanup does not cover process death; a crash after sentinel creation permanently blocks upgrade retries, so make abandoned claims recoverable.

integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -600,21 +767,72 @@ def _post(payload: dict[str, Any], url: str) -> bool:
...
+            # new key to the previous account's email, permanently, and the
+            # fingerprint would then agree with itself forever after.
+            verified = _resolve_email(key)
+            if not verified:
+                # Offline, firewalled, or the API is down. Keep the previous
+                # behaviour and retry on the next flush rather than dropping a
+                # real account attribution. Safe because the same network that
+                # failed /v1/ping/ is about to fail the PostHog POST, so nothing
+                # is delivered under the unverified identity in the meantime.
+                return email, ""
```
kartik-mem0, line 831: [P2] Mem0 /ping can fail while PostHog succeeds, so returning this unverified legacy email attributes events to the previous account; use a fresh anonymous identity until verification succeeds.

integrations/agent-plugin-core/python/telemetry.py
```diff
@@ -290,9 +290,176 @@ def anonymous_id(identity: dict[str, str] | None = None) -> str:
...
+    if not previous or previous == memory_core.PLUGIN_VERSION:
+        return None
+    # Claim the transition with an exclusive sentinel before rewriting the
+    # marker. A plain read-modify-write let every concurrently starting session
+    # observe the old version and each record its own upgrade — and the first
+    # session after a version bump is exactly when several agent windows restart
+    # together.
+    sentinel = path.with_name(f"upgraded-{memory_core.PLUGIN_VERSION}")
+    try:
+        os.close(os.open(sentinel, os.O_CREAT | os.O_EXCL | os.O_WRONLY, 0o600))
```
kartik-mem0, line 452: [P2] A crash after creating this sentinel bypasses OSError cleanup and permanently blocks upgrade retries; make abandoned claims recoverable.

kartik-mem0 (changes requested):
AI-assisted re-review: two issues still need fixes: failed legacy identity verification and upgrade recovery after a process crash. Added fresh inline comments on the current diff; earlier feedback was in existing threads.

karthik-indla (dismissed):

kartik-mem0 (approved):

PR #7326, opened 2026-09-14
feat(integrations): declare which surface each client is, and its version

whysosaket:
**5 of 5**, stacked on #7325. Splitting #7321. This is the one that pairs with [mem0ai/platform#3602](https://github.com/mem0ai/platform/pull/3602).

## What was missing

Nothing on the wire said which Mem0 surface made a call. Both SDKs sent only an auth header, so the platform saw `python-httpx` and `axios` and attributed every plugin, wrapper and direct API user to one undifferentiated bucket. Version was unknowable, which is what gates every deprecation decision.

## Three headers, and the rules are the point

| Header | Carries | Rule |
|--------|---------|------|
| `X-Mem0-Source` | one canonical source value | **set-once** — write only if absent |
| `X-Application` | the host app it runs inside | **set-once** — write only if absent |
| `X-Mem0-Client` | `name/version`, outermost first | **append-only** — add yourself, never replace |

Set-once means `setdefault`, never assignment. A plugin wrapping the SDK is the outermost layer and sets the source; the SDK underneath defers to it. Assignment is exactly how the agent plugins came to be indistinguishable — see #7323, and `mcp-server` on the platform side.

Append-only means a plugin calling the Python SDK produces `mem0-plugin/0.3.1, mem0-python/2.0.19`, so neither layer can erase the other.

Deliberately **not** `User-Agent`: proxies rewrite it, and we have already met a WAF that 403s on it.

## The hoist

The plugin core also moves `source` out of `metadata` to the top level, which is where the backend actually reads it. It sat in `metadata`, which `get_event_source` never consults — so all six plugins arrived indistinguishable from a raw SDK call no matter what they set. The harness tag stays in metadata as hook provenance.

## Surfaces touched

- Python SDK, TS SDK — `X-Mem0-Client` with the real version; `MEM0_SOURCE` / `MEM0_APPLICATION` honoured with check-then-set so a wrapper wins.
- Both CLIs — they already sent `X-Mem0-Source: cli`, which **no backend code read**. It now does; canonical casing plus the client header.
- `pi-agent` — had `PI_AGENT` as a PostHog property only, never on the wire.
- `vercel-ai-sdk` — sent nothing at all from its raw `fetch` calls.

## The coordination rule

Values must exist in the platform's `EventSource` enum or they bucket to `OTHERS`. `integrations/AGENTS.md` now states the contract, and the "adding an integration" checklist requires landing the platform value in the same week. That is the gap that let `deepseek-plugin`'s README ship a claim that `ZAPIER` and `STRANDS` were already recognized when neither was (fixed in #7322).

## Unknown hosts do not get a host

The portable bundle is built for whatever editor a user drops it into, so it cannot know its host. It was writing its build name, `coding-agent`, into `X-Application` anyway. That value is not on the platform's allowlist, so it was already being discarded server-side, which is the worst of both outcomes: the wire claimed we knew the editor and the stored event recorded that we did not. It now sends no header at all. `HARNESS_ID` stays `coding-agent`, since that is the PostHog-side label and grouping portable installs there is useful. Two build tests cover it, one per direction.

## Restored docs

This branch's merge from #7325 had resolved eleven documentation files to their pre-sweep state, silently reverting #7322 in full. Since the stack lands in order, main would have taken the corrected privacy wording and then had this PR undo it. Restored, and the diff against #7325 is now surface headers only.

## Verification

298 passed, 8 skipped across the shared core and the five host suites, up from 291 at the base of this PR. Bundle `--check` clean for all six hosts.

The TypeScript packages are covered by CI rather than locally, and CI is green: Pi Agent Plugin build, lint and tests on Node 20 and 22, TypeScript SDK build and integration on both, Node and Python CLI build, lint and tests, and the Zapier app validation.


integrations/pi-agent-plugin/src/commands.ts
```diff
@@ -29,7 +34,12 @@ export function registerCommands(
       threshold: config.searchThreshold,
       topK: SEARCH_TOP_K,
       rerank: true,
-    });
+      source: PLATFORM_SOURCE,
```
kartik-mem0, line 39: [P2] Set attribution on the shared Pi client; tagging only command search/add leaves automatic recall, capture, memory tools and deletion attributed as generic SDK traffic.

kartik-mem0 (changes requested):
AI-assisted review: 7/10. Hold until Pi attribution covers the shared client, including tools, recall, capture and deletion; the unchanged TypeScript telemetry delivery gaps are being addressed separately.

karthik-indla (commented):
Independent review — **Kartik's Pi item is confirmed still-present**, plus the header wiring and the cross-cutting TS-core concern he raised on the thread.

**Pi (`:37`, confirmed):** the surface is a body param on 2 of ~11 call sites (`commands.ts:37` search, `:55` add), never on the shared client. Everything else on the same shared `mem0` goes out unattributed — deletion (`commands.ts:95,112`), list/getAll (`:156,227`), automatic recall (`entry.ts:80`), capture (`capture/index.ts:30`), and every memory tool (`memory/tools.ts:65,77,92,116,126`). Fix: set `X-Mem0-Source` once on the shared client at `entry.ts:31` (via `process.env.MEM0_SOURCE` before `new MemoryClient`, or a `source` field on `ClientOptions`). Also an internal vocab split: wire source `"PI_AGENT"` (commands.ts:9) vs telemetry `"PI_AGENT_PLUGIN"` (telemetry.ts:58).

**Two more on this PR:**
- `integrations/vercel-ai-sdk/src/mem0-utils.ts:1` hardcodes `PROVIDER_VERSION = "3.0.2"` — matches package.json today, goes stale on the next bump. Every other client injects the version at build (`__MEM0_SDK_VERSION__`, `importlib.metadata`, `__CLI_VERSION__`).
- No shared source-vocabulary constant: all source values are bare literals un-pinned to the backend `EventSource` enum, so a typo or new surface silently buckets to `OTHERS` (the code comment at `mem0.types.ts:33` even warns about this). Header **names** are consistent and match the platform relay; it's the **values** that are un-governed and un-asserted.

**The cross-cutting TS-core gap you flagged on the thread — confirmed.** `integrations/agent-plugin-core/typescript/src/telemetry.ts` is unchanged across this stack; all the hardening landed only in the Python copies. Concretely: (a) **failed-delivery** — the queue is fully in-memory and detached from the array *before* the `await deliver()`, wrapped in `catch {}` (:92-101), so any network failure drops the batch with no spool/re-park/crash-recovery; a short-lived CLI/plugin that exits after `capture()` loses everything queued. (b) **privacy** — no `sha256` digest primitive in the core (it drops identifiers rather than hashing them — weaker, not a confirmed leak). (c) **identity** — no anon-id/rotation in the core at all; each wrapper reimplements it, so a new wrapper that forgets the rotation dance leaks identity across accounts. This is the "TS side still owes" work — reasonable as the separate follow-up you scoped, but worth tracking explicitly since it's the same failed-delivery + identity class the Python stack just spent 5 PRs fixing.

integrations/pi-agent-plugin/src/attribution.ts
```diff
@@ -0,0 +1,50 @@
...
+export function applySurfaceHeaders(client: MemoryClient): void {
+  const headers = client.headers as Record<string, string>;
+  if (!headers["X-Mem0-Source"]?.trim()) headers["X-Mem0-Source"] = PLATFORM_SOURCE;
+  if (!headers["X-Application"]?.trim()) headers["X-Application"] = PLATFORM_APPLICATION;
+
+  const existing = (headers["X-Mem0-Client"] ?? "")
+    .split(",")
+    .map((part) => part.trim())
+    .filter(Boolean);
+  existing.push(`mem0-pi-agent/${PLUGIN_VERSION}`);
```
kartik-mem0: [P2] Appending Pi after mem0-js makes the backend select the inner SDK version; preserve outer wrappers, place Pi before its SDK, and test an SDK-initialized client.

kartik-mem0 (changes requested):
AI-assisted re-review: shared-client attribution is fixed, but Pi client-stack ordering still needs correction.

karthik-indla (dismissed):

kartik-mem0 (approved):

PR #7328, opened 2026-09-14
fix(pgvector): only close a connection pool this store built

### LINKED ISSUE #7314: PGVector destructor closes externally supplied connection_pool
## Summary

In mem0ai 2.0.20, PGVector accepts an externally supplied connection_pool, but __del__ closes that pool without tracking ownership. Disposing one Memory/vector-store instance can therefore close a shared, application-owned pool used by other instances.

## Source

mem0/vector_stores/pgvector.py:
- Lines 191–193: supplied connection_pool is assigned directly.
- Lines 202–214: pool construction is skipped when one was supplied.
- Lines 545–556: __del__ unconditionally calls connection_pool.close() (psycopg 3) or closeall() (psycopg 2).

## Minimal reproducer (no database needed)

```python
import gc
from mem0.vector_stores.pgvector import PGVector

class ExternalPool:
    def __init__(self):
        self.closed = False
    def close(self):
        self.closed = True
    def closeall(self):
        self.closed = True

pool = ExternalPool()
store = PGVector(
    dbname="postgres", collection_name="example",
    embedding_model_dims=8, user=None, password=None,
    host=None, port=None, diskann=False, hnsw=False,
    connection_pool=pool,
)
assert store.connection_pool is pool
del store
gc.collect()
assert not pool.closed  # Fails: destruction closed an externally owned pool.
```

## Expected behavior

Record whether PGVector constructed its own pool, and only dispose owned pools. An application that supplies a shared pool should remain responsible for closing it.

## Impact

This prevents safely injecting one process-owned psycopg pool into short-lived Memory instances. A borrowed-pool wrapper that delegates operations and makes close/closeall no-ops is an application-side workaround, but ownership handling in PGVector would avoid requiring it.

Environment inspected: mem0ai 2.0.20, psycopg 3.3.4, psycopg-pool 3.3.1.

L4XB:
## The bug

`PGVector.__del__` closed `self.connection_pool` whichever way it got there. A caller who supplies a pool usually shares it — with other stores, or with the rest of the application — and disposing one `Memory` took it away from all of them. Nothing had handed over the right to close it.

The issue's own reproducer, unchanged, before and after:

```
$ python repro.py            # on main
externally supplied pool closed after disposal: True

$ python repro.py            # with this change
externally supplied pool closed after disposal: False
```

## The change

The store records whether it built the pool (`_owns_connection_pool`), set in the one branch that constructs one, and `__del__` returns early when it did not. A supplied pool is the caller's and stays open; a pool the store made is still the store's to close.

`getattr(self, "_owns_connection_pool", False)` rather than a plain attribute read: `__del__` can run on an instance whose `__init__` raised before the flag was set, and a destructor that raises there would replace a clear error with an unrelated one.

## Tests

```
pytest tests/vector_stores/test_pgvector.py
91 passed
```

Four new cells, two per psycopg version so neither branch of `__del__` is left uncovered:

- `test_del_leaves_a_supplied_pool_open_psycopg3` / `_psycopg2` — the bug.
- `test_del_closes_a_pool_it_built_psycopg3` / `_psycopg2` — the accept controls. Without them a change that simply never closed anything would pass, and that leaks the pools the store really does own.

Against `main`, with the new cells in place, **2 fail** — the two that describe the bug:

```
FAILED test_del_leaves_a_supplied_pool_open_psycopg2
FAILED test_del_leaves_a_supplied_pool_open_psycopg3
2 failed, 89 passed
```

**Each cell was checked against a mutant that breaks the thing it names.** Baseline 91/91 green, source restored and re-verified; every mutation asserts it changed the file first:

| mutation | result |
| --- | --- |
| drop the ownership check in `__del__` (the bug) | **2 failed** / 89 passed |
| never mark a self-built pool as owned | **4 failed** / 87 passed |
| mark every pool as owned, supplied ones included | **2 failed** / 89 passed |
| *(restored)* | **91 passed** |

The second row is why the accept controls exist: a fix that only ever declines to close would have satisfied the bug report and quietly leaked every pool the store creates for itself.

`ruff check` passes on both files. `ruff format --check` reports `pgvector.py` as unformatted, but it already is on `main` — reformatting it here would bury a five-line change in unrelated churn.

Closes #7314


L4XB:
Description now says `Closes #7314`, which is the keyword the gate reads — it was `Fixes`.

The issue is not labelled `accepted` yet, so I understand this stays closed until a maintainer decides it is worth taking on. No rush from me; the branch is there when it is.

For whoever picks it up: the issue's own reproducer is the whole story, and it needs no database —

```
on main:            externally supplied pool closed after disposal: True
with this change:   externally supplied pool closed after disposal: False
```

PR #7329, opened 2026-09-14
fix(deepseek-plugin): do not mark memories seen for a turn that was cancelled

### LINKED ISSUE #7302: Deepseek Harness: canceled recall marks undelivered memories as seen
### Component

Plugin

### Summary

Canceling a DSH turn while automatic memory search is pending can mark returned memories as seen without delivering them to the model. Later automatic recall skips those memories in the same session, so the agent may answer without relevant stored information(if agent does not use search_memory).

### Steps to Reproduce

1. On the commit listed below, insert the inline test from the verification field into the existing `describe("Harness lifecycle", ...)` block in `integrations/deepseek-plugin/tests/apply.test.ts`.
2. From `integrations/deepseek-plugin`, run the installation and test commands provided in that field.

The test cancels during a mocked memory search, lets it return successfully, then retries in the same session with the same memory ID. No API keys, live model, or running DSH server are required.

### Expected Behavior

Canceling a turn should not make the agent forget a memory it never received. If a later search returns that memory, the plugin should include it in the model's context again.

Its ID should become “seen” only after successful delivery, not merely because search found it. Deduplication should still apply to memories previously delivered successfully.

### Actual Behavior

The retry returns `contexts: []` instead of a `mem0:recall` entry containing the retrieved memory. Its ID was marked as seen during the canceled turn, so the plugin filters it out. This test checks the recalled context, not a generated agent answer.

### Environment

- Mem0 plugin: `@mem0/deepseek-plugin@0.3.1` at commit `782fa392a5b76e3acf474478e7e22637c0e41575`.
- Mem0 SDK: `mem0ai@3.1.6`.
- DSH packages: `0.1.1-rc.2`.
- Node.js: `22.22.0`.
- OS: macOS.

### What I Ran

On commit `782fa392a5b76e3acf474478e7e22637c0e41575`, inserted the following test as the first test inside the existing `describe("Harness lifecycle", ...)` block in [`integrations/deepseek-plugin/tests/apply.test.ts`](https://github.com/mem0ai/mem0/blob/782fa392a5b76e3acf474478e7e22637c0e41575/integrations/deepseek-plugin/tests/apply.test.ts). Place it immediately after `describe("Harness lifecycle", () => {`, before the existing tests. It uses that file's existing helpers and mocks. No API keys, live model, DSH server, or separate download is needed.

```ts
  it("recalls an undelivered memory after cancellation", async () => {
    const response = { results: [{ id: "m1", memory: "A saved fact" }] };
    const controller = new AbortController();
    const hooks = applyAndCollectListeners({ apiKey: "test-key", userId: "u" });
    const assemble = hooks.get("system-prompt/assemble")!;
    const base = { sections: [], contexts: [], tools: [], variables: {} };
    const agent = { session: { deriveMessages: () => [
      { role: "user", source: { kind: "user" }, content: "Recall my saved information" },
    ] } };
    mockSearch.mockImplementationOnce(async () => {
      controller.abort(); // Cancel during search, before returning the memory.
      return response;
    }).mockResolvedValue(response);

    expect(await assemble(base, { agent, signal: controller.signal }, async () => base)).toBe(base);
    const retried = await assemble(base, { agent, signal: new AbortController().signal }, async () => base);
    expect(mockSearch).toHaveBeenCalledTimes(2);
    console.log(JSON.stringify(retried));
    expect(retried).toMatchObject({
      contexts: [{ name: "mem0:recall", text: expect.stringContaining("A saved fact") }],
    });
  });
```

From `integrations/deepseek-plugin`, install dependencies if needed, then run the Harness lifecycle group:

```sh
pnpm install --frozen-lockfile --config.auto-install-peers=false --ignore-scripts
./node_modules/.bin/vitest run tests/apply.test.ts -t "Harness lifecycle"
```

### What I Saw

Both searches returned the same memory: `m1`, containing `A saved fact`. After cancellation and retry:

| Expected output | Actual output |
| --- | --- |
| The model's automatic memory context includes `A saved fact`. | The automatic memory context is empty. The retrieved fact is not delivered. |

The rerun reports the test inside the requested group:

```text
 FAIL  tests/apply.test.ts > Harness lifecycle > recalls an undelivered memory after cancellation
      Tests  1 failed | 5 passed | 14 skipped (20)
```

The five existing Harness lifecycle tests passed. The new regression failed with this expected-versus-actual diff:

```diff
- Expected
+ Received

  {
-   "contexts": [
-     {
-       "name": "mem0:recall",
-       "text": StringContaining "A saved fact",
-     },
-   ],
+   "contexts": [],
  }
```

The failure is missing input to the agent; this offline test does not generate an agent answer. The package type check passed.

### Why This Is a Bug

A memory never delivered to the model should remain eligible for automatic recall. Instead, the lifecycle [marks the ID seen](https://github.com/mem0ai/mem0/blob/782fa392a5b76e3acf474478e7e22637c0e41575/integrations/agent-plugin-core/typescript/src/lifecycle.ts#L169) before the adapter [checks cancellation and discards the text](https://github.com/mem0ai/mem0/blob/782fa392a5b76e3acf474478e7e22637c0e41575/integrations/deepseek-plugin/src/index.ts#L184). Later recall therefore skips information the model never received.

### What I Ruled Out

- Missing search results or API errors: both mocked searches successfully return the same memory.
- Prior delivery: the canceled assembly is unchanged and the existing prompt contains no memory text.
- Uncertain cancellation timing: the mock cancels inside search, before resolving its response.
- Model variability: this regression checks the plugin's recalled context without calling a model. It does not assert that every live agent answer must fail.

This behavior also exists before #7285; the report does not attribute its introduction to that PR.

### AI Assistance

L4XB:
## The bug

`buildRecallContext` marks each memory seen **as it builds the context string**:

```ts
lines.push(line);
options.seenIds?.add(memory.id);
```

The DSH plugin then throws that context away when the turn was cancelled:

```ts
const memoryContext = await state.lifecycle.recall(...);
if (!memoryContext || signal?.aborted) return assembly;   // <- ids already seen
```

So cancelling while the search is in flight costs the memories in that answer. The next automatic recall in the same session filters them out as already delivered, and the agent answers without information it holds — silently, because nothing failed.

## The change

`RecallOptions` takes an optional `isCancelled`, checked once the search has answered and before anything is marked:

```ts
if (!response) return "";
// The search can finish after the turn was cancelled. Nothing below this
// line is delivered in that case, so nothing below it may be marked seen.
if (options.isCancelled?.()) return "";
```

`MemoryLifecycle.recall` takes it as an optional fourth argument and the DSH plugin passes `() => signal?.aborted === true`.

Both are additive. `opencode-plugin`, `openclaw` and `pi-agent-plugin` call `recall` with three arguments and are untouched; `buildRecallContext` keeps its exported signature, which matters because `pi-agent-plugin` re-exports it.

## What this does not do

It closes the window the issue reproduces — cancellation during the search. It does not make "seen" mean "delivered" in general: an abort that lands between `recall` returning and the caller adding the context still leaves the ids marked. Closing that properly means the lifecycle hands back the ids and the caller commits them after delivery, which changes `recall`'s return type for four integrations. That is a design call I would rather leave to you than make in a bug fix — say the word and I will send it.

## Tests

```
cd integrations/deepseek-plugin && npx vitest run
Test Files  5 passed (5)
     Tests  49 passed (49)
```

Two new cells in `tests/apply.test.ts`, built on the file's existing offline mocks — no key, no model, no DSH server:

- `recalls an undelivered memory after the turn was cancelled` — the issue's own scenario. Against `main` it fails with `contexts: []` on the retry.
- `still does not repeat a memory it already delivered` — the accept control. The seen set exists to stop every turn re-sending the same memories, and a "fix" that simply stopped marking would have passed the first cell while breaking that.

**Each cell was checked against a mutant that breaks the thing it names.** Baseline 49/49 green, source restored and re-verified; every mutation asserts it changed the file first:

| mutation | result |
| --- | --- |
| drop the cancellation check (the bug) | **1 failed** / 48 passed |
| stop passing the signal from the plugin | **1 failed** / 48 passed |
| invert the check, so only a cancelled turn delivers | **3 failed** / 46 passed |
| *(restored)* | **49 passed** |

`npx tsc --noEmit` is clean in `integrations/deepseek-plugin`.

Closes #7302


L4XB:
Description now says `Closes #7302`, which is the keyword the gate reads — it was `Fixes`.

The issue is not labelled `accepted` yet, so I understand this stays closed until a maintainer decides it is worth taking on. No rush; the branch is there when it is.

The reporter supplied the failing test themselves and it needs no key, model or DSH server — it is in this PR as `recalls an undelivered memory after the turn was cancelled`, and it fails on `main` with `contexts: []` on the retry.

PR #7358, opened 2026-09-17
fix(plugins): stop the TypeScript telemetry losing events and misattributing accounts

whysosaket:
Independent of the #7322 to #7326 stack. No file overlap, so it reviews and lands in parallel rather than queueing behind five PRs.

## Why this exists separately

Those five PRs hardened the Python plugin telemetry. The TypeScript side has the same defect classes and was not touched, because the two share no code. `agent-plugin-core/python` generates into six plugin bundles, so one fix lands six times. `agent-plugin-core/typescript` is a separate core that each TS plugin wraps. Nothing about fixing one surfaces the other, which is how this survived until review caught it.

## A failed delivery deleted the batch

The core detached the queue before the `await` and swallowed the error. I probed it rather than reading it:

```
queued before flush : 2
delivery attempts   : [ 2 ]
queued after failure: 0     <- gone
attempts after 2nd  : 1     <- no retry, ever
```

One network blip destroyed the events, with nothing recording that it had happened. So we could not have measured how much we were losing, because the evidence was the thing being deleted.

The batch is now put back, bounded by `maxQueueSize` and biased to the newest so a long outage costs the oldest events rather than unbounded memory. Repeated failures back off to a ceiling instead of retrying every flush against a host that is blocking us. Every event carries a `uuid` stamped at capture, which is what makes retrying safe: PostHog collapses anything it already accepted.

## No disk spool, deliberately

Written into the code so it reads as a decision rather than an omission. Python spools because its hooks are separate processes firing per tool call that exit immediately, so nothing survives without a file. These plugins are loaded into a host that lives for a whole session, so re-queueing covers the same transient failures without the claim and lease machinery that took three review rounds to get right on the Python side.

What that leaves uncovered is narrow and worth stating: a session that both starts and ends with no connectivity still loses its events. If we decide that cohort matters, cross-process durability is its own piece of work.

## openclaw used a cached email forever

The refresh was guarded by `!hasEmail`, so once an email was cached it was never re-checked. After an API key change every event kept reporting under the previous account.

The email is now bound to a fingerprint of the key it was resolved for, and used only while those agree. A mismatch forgets the account and re-resolves. The resolution latch is per key rather than once per process, so a key changed mid-session is actually looked up instead of sticking to the fallback until restart.

A row carrying an email and no fingerprint, which is what upgrading from the current version looks like, is verified rather than adopted. That matches the decision reached on #7325 after the same question came up there.

## opencode hashed the project id unsalted

Reversible by anyone who can enumerate project ids. Salted with the API key rather than a stored per-install value: it is already in play here, it is high entropy, and it needs no new file and therefore no write race to get wrong. The hash is per account rather than per machine, which also keeps joins working for one user across machines, and it resets on key rotation consistently with `distinctId`, which already did.

Also wired opencode's tests into CI. They existed and nothing ran them, so the regression test asked for on #7322 would not have gated anything.

## Verification

| Package | Tests |
|---|---|
| agent-plugin-core/typescript | 30 |
| openclaw | 431 |
| opencode | 35 |
| pi-agent | 89 |
| deepseek | 47 |

Each new test was checked against the unfixed code first: the three core ones all fail without the retry, and the openclaw one fails without the fingerprint gate. One existing opencode test asserted the unsalted digest, so it encoded the defect as expected behaviour and is replaced.

## Not in scope

The TS core still does not own identity. Each wrapper reimplements the anonymous id and aliasing, which is why openclaw could carry this defect alone. Worth consolidating, but it is a refactor rather than a fix and does not belong in the same change.


integrations/openclaw/telemetry.ts
```diff
@@ -76,15 +115,16 @@ function resolveEmail(apiKey: string): void {
     .then((response) => response.json())
     .then((data: any) => {
       if (!data?.user_email) return;
-      writePluginAuth({ userEmail: data.user_email });
+      writePluginAuth({ userEmail: data.user_email, keyFingerprint: fingerprint });
```
kartik-mem0, line 118: [P1] Add keyFingerprint to config.ts ALLOWED_KEYS and the manifest schema; this successful lookup otherwise writes a config that fails the next plugin load, so cover the persisted-config round trip.

integrations/openclaw/telemetry.ts
```diff
@@ -33,10 +39,35 @@ function anonymousId(): string {
...
 function distinctId(apiKey?: string): string {
   try {
-    const email = readPluginAuth().userEmail;
-    if (email) return createHash("sha256").update(email).digest("hex");
+    const auth = readPluginAuth();
+    if (auth.userEmail) {
+      // Only when it belongs to the key in hand. Without this check a cached
+      // email was used forever: switch to a different account and every event
+      // kept reporting under the previous one, with nothing to notice it by.
+      if (auth.keyFingerprint === keyFingerprint(apiKey)) {
```
kartik-mem0, line 54: [P2] Return keyFingerprint from readPluginAuth(); it currently drops the persisted field, so this check always fails and later events revert to the key hash; test with the real config reader.

integrations/agent-plugin-core/typescript/src/telemetry.ts
```diff
@@ -74,34 +76,98 @@ export function errorKind(error: unknown): string {
...
+        consecutiveFailures = 0;
+        retryNotBefore = 0;
+        return;
+      }
+      // Keep the FRONT on overflow, so the batch being retried survives and a
+      // new event is what gets dropped. Matches the Python core, where record()
+      // refuses new events once the spool is full rather than evicting the
+      // backlog. Keeping the newest would throw away exactly the events this
+      // retry exists to save.
+      queue = [...batch, ...queue].slice(0, maxQueueSize);
```
kartik-mem0, line 164: [P2] Serialize flushes with an in-flight guard; overlapping failures prepend newer batches and truncate older retries, so add a regression with two pending deliveries failing in capture order.

kartik-mem0 (changes requested):
AI-assisted independent review: three reproduced issues need fixes before merge; package tests and typechecks pass, but real config round trips and overlapping failures expose these gaps.

kartik-mem0 (approved):

